PR #56155, opened 2024-01-31
P20-691: Stop sync-out the source language

artem-vavilov:
- jira ticket: [P20-691](https://codedotorg.atlassian.net/browse/P20-691)


bin/i18n/i18n_script_utils.rb
```diff
@@ -16,8 +16,8 @@ class I18nScriptUtils
   CROWDIN_CREDS_PATH = CDO.dir('bin/i18n/crowdin_credentials.yml').freeze
   PROGRESS_BAR_FORMAT = '%t: |%B| %p% %a'.freeze
   PARALLEL_PROCESSES = Parallel.processor_count.freeze
-  SOURCE_LOCALE = 'en-US'.freeze
-  TTS_LOCALES = (::TextToSpeech::VOICES.keys - %i[en-US]).freeze
+  SOURCE_LOCALE = I18n.default_locale.to_s.freeze
+  TTS_LOCALES = (::TextToSpeech::VOICES.keys - [SOURCE_LOCALE.to_i]).freeze
```
mgc1194: should it be `to_sym` instead of `to_i`? We could also use `%I[#{SOURCE_LOCALE}]`

bin/i18n/utils/sync_out_base.rb
```diff
@@ -19,7 +19,9 @@ def process(_language)
       end
 
       def languages
-        @languages ||= PegasusLanguages.all
+        @languages ||= I18nScriptUtils.cdo_languages.select do |cdo_language|
+          cdo_language[:locale_s] != I18nScriptUtils::SOURCE_LOCALE
+        end
```
mgc1194: Could you add a comment in the code explaining what is the purpose of the `languages` method? and explaining what it returns. I'm concerned other developers may assume it returns all languages.

PR #56158, opened 2024-01-31
Removing firehose logs for afe

hannahbergam:
Removes all but one firehose call for afe! The submit call still happens in amazon_future_engineer_controller.rb. Any events that had a redirect on callback now redirect after the amplitude event is called.

I was able to test locally that these windlow.location calls still work, both from an ineligible school entry and a signed out submit (which redirects to account creation). 

## Links

- spec: []()
- jira ticket: https://codedotorg.atlassian.net/browse/ACQ-1373

## PR Checklist:

- [ ] Code is well-commented


TurnerRiley (approved):
Woooo! Love to see these firehose deletion PR's

PR #56165, opened 2024-01-31
Storybook - Refactor choice_responses.story.jsx

kelbyhawn:
Refactors the `ChoiceResponses` component story.

_Note:_ It looks like there were two `storybook` elements in the original story with duplicated constants: one with and without facilitators. I renamed the duplicate constants to include `PerFacilitator`, but if anyone knows of a cleaner way to do this with the new API let me know!

**Jira ticket:** [ACQ-1433](https://codedotorg.atlassian.net/browse/ACQ-1433)

----

https://github.com/code-dot-org/code-dot-org/assets/9256643/f9d17ca9-b573-4a1a-a5a2-eb1c31bb8a1c


TurnerRiley (approved):
I think the way you set it up with perFacilitator looks great! Seems like an accurate refactoring of how it was set up!

PR #56167, opened 2024-01-31
Add HelpTip for Announcement link field

bethanyaconnor:
We've had a couple issues where, when a link should go to our code.org site, it ends up going to a broken link. I don't fully understand why but when the link is `code.org/afe`, it sends the user to `studio.code.org/s/code.org/afe`. If `https://` is added in front of the link, it works.

This PR simply adds a HelpTip with this information, as it's a bit of a sharp corner. Definitely open to feedback on wording or whether folks think this is useful.

Screenshot 2024-01-31 at 4 45 54 PM


davidsbailey (approved):
help tip is nice! some validation to make sure the field starts with `/` or `http` might be more effective, but up to you on whether to spend the time on that

bethanyaconnor:
haha, yeah I was going for lightweight as I don't think this is a common issue tbh

PR #53623, opened 2023-09-05
TypeScript locale completion

sanchitmalhotra126:
Leveraging TypeScript to get autocompletion and type checking for locale objects. This is done by creating an object type that uses the keys from a given JSON file as its own keys. I've included some examples of using these objects in the Music Lab directory as well. See videos below for examples.

The main open question I have here is if it's safe to import a file outside of `apps/src` in the `apps/src` directory. In this case, we're reading the keys from JSON files in `apps/i18n/...`. Since these files are only referenced in types however, these checks only happen compile-time, and at runtime, importing these locale objects is essentially equivalent to just `import <locale> from '@cdo/locale/<app>'`. 

For example, a TypeScript source file which references `apps/i18n/...` and looks like this

Screenshot 2023-09-05 at 3 47 14 PM

gets converted down to just this at runtime:

Screenshot 2023-09-05 at 3 46 17 PM

Autocompletion/type-checking feature examples in VS Code
- Using common strings:
 
https://github.com/code-dot-org/code-dot-org/assets/85528507/6060e6c5-62c9-4543-a95b-0cd2a27b6d62

- Using app-specific (music) strings:

https://github.com/code-dot-org/code-dot-org/assets/85528507/d13487d6-a835-44af-9e09-82032af7b3f6

PR #56171, opened 2024-01-31
Convert a few Blockly files to TypeScript

molly-moen:
Initial steps at adding some TypeScript to our Blockly code. I just converted a couple files here:
- `constants.js` -> `constants.ts` : the only change that was needed here was typing the parameters of two functions. We also were explicitly importing things from this file via `constants.js`, so I updated those imports to just be from `constants`.
- `ModalFunctionEditor.jsx` -> `ModalFunctionEditor.tsx`: our only React component in the blockly folder. All that was required here was typing the state from redux and changing the locale import to the TypeScript version (see #53623 for details on this)
- `utils.js` -> `utils.ts`: this one was more interesting:
  -  Most of the changes were adding types, either basic ones or ones provided by Blockly. 
  -  I also learned that `reduce` in TypeScript utilizes [generics](https://www.typescriptlang.org/docs/handbook/2/generics.html), so you need to add a type to your call of `reduce` to specify what type it will return.
  - Finally, we had a function, `findFlyout`, which relied on private fields in Blockly's `WorkspaceSvg` class (`flyout` and `toolbox_.flyout_`. If we want to use TypeScript here, we could not reference a private field. I looked at `WorkspaceSvg` to see if there was an alternative. Very conveniently, `WorkspaceSvg.getFlyout()` was doing the exact same thing as this entire function ([code](https://github.com/google/blockly/blob/0b25a8a27f156a8104cd242ee68de304506693cd/core/workspace_svg.ts#L999-L1007)). Therefore I replaced use of this function with blockly's function instead (it was only used in one place, `eventHandlers`). We may not always get such a neat solution for our use of private fields, but this one ended up making our code shorter.
 

## Links

- jira ticket: [CT-307](https://codedotorg.atlassian.net/browse/CT-307)

## Testing story
I tested locally that Google Blockly labs (Dance Party and Sprite Lab on google blockly) still worked as expected, specifically around the modal function editor and browser resizing, which is what this touched.

## Follow-up work
I am spending this week trying to move some of our Blockly code to TypeScript, so expect more PRs like this. I am trying to keep each PR relatively small and low risk.

## PR Checklist:

- [ ] Code is well-commented


apps/src/blockly/utils.ts
```diff
@@ -16,19 +21,19 @@ export const TRUTHY_DEFAULT = attributeValue => attributeValue !== 'false';
  * @returns {boolean} The boolean value of the attribute as determined by the callback function.
  */
 export function readBooleanAttribute(
-  xmlElement,
-  attribute,
+  xmlElement: Element,
+  attribute: string,
```
ebeastlake: For my own TS knowledge/outside the scope of this PR: Is there any value added by defining this as an ENUM or something so we're explicit about the attributes we could be reading/avoiding typos? IIRC, I defined one-off constants like `USER_CREATED_XML_ATTRIBUTE,` but I wonder if TS could make this cleaner/more succinct.

apps/src/blockly/utils.ts
```diff
@@ -16,19 +21,19 @@ export const TRUTHY_DEFAULT = attributeValue => attributeValue !== 'false';
...
   callback = FALSEY_DEFAULT
 ) {
   const attributeValue = xmlElement.getAttribute(attribute);
   return callback(attributeValue);
 }
 
-export function capitalizeFirstLetter(string) {
+export function capitalizeFirstLetter(string: string) {
```
ebeastlake: Should we be specifying return types?

apps/src/blockly/eventHandlers.js
```diff
@@ -97,15 +96,10 @@ export function adjustCalloutsOnViewportChange(event) {
 // When the browser is resized, we need to re-adjust the width of any open flyout.
 export function reflowToolbox(event) {
   const mainWorkspace = Blockly.getMainWorkspace();
-  const mainWorkspaceFlyout = findFlyout(mainWorkspace);
-  if (mainWorkspaceFlyout) {
-    mainWorkspaceFlyout.reflow();
-  }
+  mainWorkspace?.getFlyout()?.reflow();
```
ebeastlake: Nice find on `getFlyout`!

apps/src/blockly/components/ModalFunctionEditor.tsx
```diff
@@ -7,14 +7,15 @@ import {
 import moduleStyles from './modal-function-editor.module.scss';
 import classNames from 'classnames';
 import Button from '@cdo/apps/templates/Button';
-import msg from '@cdo/locale';
+import {commonI18n} from '@cdo/apps/types/locale';
```
ebeastlake: Can you confirm all `tsx` files use `commonI18n` instead of `msg`?

apps/src/blockly/utils.ts
molly-moen: good question, do we know that these will always be specific values? If we do, we can definitely specify the expected values as an enum or type.

apps/src/blockly/utils.ts
molly-moen: It's inferred here! Generally our rule of thumb (@sanchitmalhotra126  can correct me if he follows a different one) is if the type is inferred, no need to specify. It knows it's a string because it's a pretty simple string method.
Screenshot 2024-02-01 at 9 35 02 AM

apps/src/blockly/components/ModalFunctionEditor.tsx
molly-moen: Yep! As of #53623 this is an exported object with the specific name `commonI18n`.

apps/src/blockly/utils.ts
sanchitmalhotra126: Agreed, I typically also leave out the return type if it's inferred. Sometimes it helps to be explicit if a function is more complex (for example it can be helpful to specify the return type of async functions like `Promise<SomeDataType>` so it doesn't default to `Promise<void>`) which also helps in enforcing that constraint throughout the function, but yeah in cases like this it seems fine to not specify.

apps/src/blockly/utils.ts
ebeastlake: Yes and no. We could start reading a new XML attribute, but then we'd probably want to use that string in many different places and not risk misspelling it anywhere. So, at any given moment, we should be able to deduce that we're only reading specific values, but that list is not static.

ebeastlake (approved):

PR #56176, opened 2024-02-01
Music: Convert TimelineSimple2Events to TypeScript

sanchitmalhotra126:
Convert TimelineSimple2Events to TypeScript with some minor refactors - mostly wrapping memoizable values in `useMemo` and moving util functions out of the component to avoid extra renders.


apps/src/music/views/TimelineSimple2Events.tsx
```diff
@@ -0,0 +1,152 @@
...
+  orderedFunction: FunctionEvents,
+  uniqueSounds: string[],
+  orderedFunctions: FunctionEvents[]
+) => {
+  let left = Number.MAX_SAFE_INTEGER,
+    top = Number.MAX_SAFE_INTEGER,
+    right = 0,
+    bottom = 0;
+
+  orderedFunction.playbackEvents.forEach(playbackEvent => {
```
breville: Out of curiosity, do you have a preference for `forEach` over `for...of`?

breville (approved):
Nice!

apps/src/music/views/TimelineSimple2Events.tsx
sanchitmalhotra126: Oh no preference, I think this function was copied as-is - can switch to `for...of` if you think that's clearer!

breville: I used to have a slight perference for `for...of`, partly because I think it used to be easier to step into with a browser debugger, but these days either seems fine.

PR #56185, opened 2024-02-01
Migrate SpeedSlider storybook entry

fisher-alice:
This PR migrates the `SpeedSlider` component to the updated Storybook API.

Video of component in Storybook:

https://github.com/code-dot-org/code-dot-org/assets/107423305/bf81bb64-d192-4aa7-b208-ea42eb24098d

## Links
[jira](https://codedotorg.atlassian.net/browse/LABS-477)

## Follow-up work
I noticed in `SpeedSlider.jsx` that the [dark turtle and rabbit images](https://github.com/code-dot-org/code-dot-org/blob/staging/apps/src/templates/SpeedSlider.jsx#L24-L27) are not used. Currently, this component is used in the Debug Console area in App Lab and in the Game Lab animation tab if an animation with more than one frame is added. For both uses, the background is dark so the light color UI is appropriate.

Screenshot 2024-02-01 at 11 11 02 AM

Screenshot 2024-02-01 at 11 11 21 AM

But I wonder if we should update the `SpeedSlider` component so it has both white and dark UI available (similar to how `color` is implemented in [`SimpleDropdown.tsx`](https://github.com/code-dot-org/code-dot-org/blob/f109907bbc4d23238b184548a56f8a3b33da97c1/apps/src/componentLibrary/simpleDropdown/SimpleDropdown.tsx#L29)). 

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/templates/SpeedSlider.story.jsx
+++ b/apps/src/templates/SpeedSlider.story.jsx
@@ -23,7 +23,6 @@ class StorybookHarness extends React.Component {
 }
 
 export default {
-  title: 'SpeedSlider',
   component: SpeedSlider,
 };
 
--- a/apps/src/templates/SpeedSlider.story.jsx
+++ b/apps/src/templates/SpeedSlider.story.jsx
@@ -1,24 +1,32 @@
 import React from 'react';
 import SpeedSlider from './SpeedSlider';
 
-class StorybookHarness extends React.Component {
+class SpeedSliderHarness extends React.Component {
   state = {value: 0.5};
 
   onValueChange = newValue => this.setState({value: newValue});
 
   render() {
     return (
-      <SpeedSlider
-        hasFocus={false}
-        value={this.state.value}
-        onChange={this.onValueChange}
-      />
+      <div style={{display: 'flex', gap: '20px'}}>
+        <div style={{background: 'black', padding: 10}}>
+          <SpeedSlider
+            hasFocus={true}
+            value={this.state.value}
+            lineWidth={200}
+            onChange={this.onValueChange}
+          />
+        </div>
+      </div>
     );
   }
 }
 
-export default storybook => {
-  return storybook
-    .storiesOf('SpeedSlider', module)
-    .add('Default', () => <StorybookHarness />);
+export default {
+  title: 'SpeedSlider',
+  component: SpeedSlider,
 };
+
+const Template = () => <SpeedSliderHarness />;
+
+export const Default = Template.bind({});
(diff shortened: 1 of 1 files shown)


apps/src/templates/SpeedSlider.story.jsx
```diff
@@ -8,17 +8,25 @@ class StorybookHarness extends React.Component {
...
     );
   }
 }
 
-export default storybook => {
-  return storybook
-    .storiesOf('SpeedSlider', module)
-    .add('Default', () => <StorybookHarness />);
+export default {
+  title: 'SpeedSlider',
```
bencodeorg: Assuming we get no objections to storybook reorganization, remove this?

bencodeorg (approved):
LGTM -- curious if the harness effect could be done without react/state, ie:

```
let value = 0.5;
const onValueChange = newValue => value = newValue;
```

But fine as-is

fisher-alice:
> LGTM -- curious if the harness effect could be done without react/state, ie:
> 
> ```
> let value = 0.5;
> const onValueChange = newValue => value = newValue;
> ```
> 
> But fine as-is

Thanks! I tried your suggestion, but then the slider was not movable. So will leave as is.

Just noting that same harness approach is used in `PaginationWrapper.story.jsx`.

levadadenys (approved):
LGTM!

levadadenys:
> But I wonder if we should update the `SpeedSlider` component so it has both white and dark UI available (similar to how `color` is implemented in [`SimpleDropdown.tsx`](https://github.com/code-dot-org/code-dot-org/blob/f109907bbc4d23238b184548a56f8a3b33da97c1/apps/src/componentLibrary/simpleDropdown/SimpleDropdown.tsx#L29)).

We have `SpeedSlider` component implementatin called `Slider` in scope of Design System (in Figma only for now), designs of it are not final, but right now it has only light icons color, the only case where icons are slightly darker - disabled state of the `Slider`. 

So I guess for now we should be fine with what we have in SpeedSlider, no need/rush to implement dark styles right now.

fisher-alice:
> > But I wonder if we should update the `SpeedSlider` component so it has both white and dark UI available (similar to how `color` is implemented in [`SimpleDropdown.tsx`](https://github.com/code-dot-org/code-dot-org/blob/f109907bbc4d23238b184548a56f8a3b33da97c1/apps/src/componentLibrary/simpleDropdown/SimpleDropdown.tsx#L29)).
> 
> We have `SpeedSlider` component implementatin called `Slider` in scope of Design System (in Figma only for now), designs of it are not final, but right now it has only light icons color, the only case where icons are slightly darker - disabled state of the `Slider`.
> 
> So I guess for now we should be fine with what we have in SpeedSlider, no need/rush to implement dark styles right now.

Awesome - thanks for the update @levadadenys !

PR #56005, opened 2024-01-26
Revert "Add Amplitude events to the navbar "Create" menu"

TurnerRiley:
Reverts code-dot-org/code-dot-org#55843

Looks like it was [causing HoneyBadger errors](https://app.honeybadger.io/projects/3240/faults/103959996) on staging: `RuntimeError: Invalid webpack asset name: 'js/userHeaderEventLogger.js'`

PR #55843, opened 2024-01-19
Add Amplitude events to the navbar "Create" menu

TurnerRiley:
Add Amplitude events to the navbar to track when a signed-out user clicks the "Create" menu and if they click an option within the menu.

My main concern is that we've seen problems before with Amplitude not logging a click event when the click triggers a redirect, I think because it doesn't have time to fire off before leaving the page the user was on. The tricky part about it is that it appears to work locally (i.e. the Amplitude log appears in the console right before navigating to the given url) so it's tough to tell whether it will work on production. I have confidence that the menu click event will work fine, but since each menu option is an `<a>` tag, I worry that the option click events won't log on Amplitude. After discussing with Dani, the plan is to wait and see if it works or not, and if it doens't work we'll remove the option event and find a new way of tracking that information.

### Shows for signed-out user
click_create_signed_out

### Does not show for signed-in user
click_create_signed_in

### Shows for signed-out user
click_option_signed_out

### Does not show for signed-in user
click_option_signed_in

## Links
Jira ticket: [here](https://codedotorg.atlassian.net/jira/software/c/projects/ACQ/boards/64?assignee=60d62161f65054006980bd71&selectedIssue=ACQ-1349)

## Testing story
Local testing.

## Follow-up work
Observe whether "Signed Out User Selects Create Dropdown Option" events show up in Amplitude, and if they don't find a workaround to track that information.

PR #56195, opened 2024-02-01
Add files for Amplitude events in the navbar Create menu

TurnerRiley:
My [first attempt](https://github.com/code-dot-org/code-dot-org/pull/55843) at adding click events to the "Create" menu in the navbar seemed to work locally, but once it merged there were [HoneyBadger errors](https://app.honeybadger.io/projects/3240/faults/103959996) on staging relating to `webpack_asset_name`. Since the navbar is used all across our site, I was afraid it would break something so I rushed to revert. However, after having [a conversation with Elijah](https://codedotorg.slack.com/archives/C046G4TRLEN/p1706554417651349), it seems like it's possible that if I had let it play out after it finished building, it would've used the newly-built package and worked.

So, following Elijah's advice, I'm using this PR to create the files themselves without actually implementing them into the user navbar so that if anything goes wrong it won't cause errors on our pages. Currently, I'm just having it console log a message so that there's something in the `js` file (even though it won't be shown anywhere). Once I can confirm this is working, I'll set up the 2nd PR which will actually implement the event logger.

## Links
Jira ticket: [here](https://codedotorg.atlassian.net/jira/software/c/projects/ACQ/boards/64?assignee=60d62161f65054006980bd71&selectedIssue=ACQ-1349)
Previous attempt: [here](https://github.com/code-dot-org/code-dot-org/pull/55843)
Revert of attempt: [here](https://github.com/code-dot-org/code-dot-org/pull/56005)
Slack conversation with Elijah: [here](https://codedotorg.slack.com/archives/C046G4TRLEN/p1706554417651349)

## Testing story
While the event logger is not being implemented anywhere yet, I was able to confirm that adding it into the `user_header.haml` file showed the console log I would expect locally (I then deleted that line before committing this):
userlogger

## Deployment strategy
I'm going to merge this after a DTP so I can monitor #infra-staging and HoneyBadger for errors. Then, if anything breaks there's plenty of time to debug and/or revert.

## Follow-up work
Actually implement the event logger into the user navbar.


hannahbergam (approved):
I appreciate the conservative approach!

PR #56199, opened 2024-02-01
Edit username in Account Settings

juanmanzojr:
Allows users to edit their username within the Accounts Settings page. Adds a validation for `username` which checks if the username contains profanity.
Screenshot 2024-02-01 at 10 47 37 AM
Screenshot 2024-02-01 at 11 41 38 AM
Screenshot 2024-02-01 at 11 46 03 AM

## Links

- jira ticket: [here](https://codedotorg.atlassian.net/browse/P20-374)
Note: Mock ups in Jira contain the new styling which will be implemented in a different PR:
Slack thread for styling changes: [here](https://codedotorg.slack.com/archives/CFTFD6BPV/p1706810549232389)

## Testing story
New test within `user_test.rb` which tests for the new profanity validation.

## Follow-up work
Styling changes

## PR Checklist:

- [ ] Code is well-commented


dashboard/app/models/user.rb
```diff
@@ -537,6 +537,7 @@ def sections
   validates_uniqueness_of :username, allow_blank: true, case_sensitive: false, on: :create, if: -> {errors.blank?}
   validates_uniqueness_of :username, case_sensitive: false, on: :update, if: -> {errors.blank? && username_changed?}
   validates_presence_of :username, if: :username_required?
+  validate :check_profanity
```
carl-codeorg: We probably need an `if username_changed?` check here. Otherwise, this validation will run on any update, potentially causing an unintentional failure because an existing user already has a profane username.

dashboard/app/models/user.rb
```diff
@@ -968,6 +969,12 @@ def username_required?
     manual? || username_changed?
   end
 
+  private def check_profanity
```
carl-codeorg: Since this function specifically validates the username field, can we call it specifically check_username_profanity, or make the function generic so it can be passed a property name and validate that property for profanity? Minor naming nitpick, but the user model is messy enough that a random validation for `:check_profanity` doesn't tell the reader enough.

carl-codeorg (changes requested):

carl-codeorg (approved):
LGTM! Did we do a quick sanity check into WebPurify costs, just to make sure the change won't cause us to hit any limits and estimate costs?

PR #56200, opened 2024-02-01
Reorganize Storybook to use directory structure

bencodeorg:
Currently, when creating a new Storybook component, we include a `title` in the default export that typically matches the name of the component itself, something like:

```
export default {
  title: 'GDPRDialog',
  component: GDPRDialog,
};
```

According to Storybook, [this is an anti-pattern](https://github.com/storybookjs/eslint-plugin-storybook/blob/main/docs/rules/no-title-property-in-meta.md) -- this PR removes that property from (almost, see caveat below) all of our Storybook components.

Removing it gives us the goodness of using our directory structure to inform Storybook's structure. Admittedly, there's work to do in organizing our `apps` directory, but IMO this is easier to follow than what was here previously (an almost entirely unstructured organization of components), and any reorganization we do of `apps` will automagically be reflected in Storybook as well. It also alphabetizes everything automagically, which is nice!

I've also added some configuration that puts the new design system components @levadadenys has been working on at the top of Storybook, since I think of those as the first thing you'd want to see when opening it. The design system components still use the `title` property so it appears at the top level -- you can still include a `title` property in your component if you want to create a new top-level "root" node (as they call it in [Storybook docs](https://storybook.js.org/docs/6.5/writing-stories/naming-components-and-hierarchy#roots)) and override the default of using the directory structure.

The downside of this approach (IMO) is that there is more hierarchy, and this hierarchy reflects our `apps` hierarchy (see second screenshot), so you might not see a component that you might have stumbled across had it been at the root level like it was previously. I think the upside outweighs the downside, but open to other opinions.

| Before | After |
| - | - |
| image | image |
|(no equivalent, most components appear at the "root" level) | image |

## Testing story

Tested manually locally (see screenshots).

## Follow-up work

It might make sense to enable some linting that would prevent adding of a `title` property (like what's in the Storybook linter mentioned above)? But given our `apps` (dis)organization, I do like the idea of giving people the option to override, like what we do with design system components. Maybe a linting rule like "you must include a slash in the title if you want to include a title property"?

### DIFF
--- a/apps/.storybook/preview.js
+++ b/apps/.storybook/preview.js
@@ -5,3 +5,15 @@ initializeRTL();
 
 //Stub jquery fileupload library function
 $.fn.fileupload = () => {};
+
+export const parameters = {
+  options: {
+    storySort: {
+      order: [
+        'DesignSystem',
+        'templates',
+        'code-studio',
+      ],
+    }
+  }
+};
--- a/apps/src/applab/ExternalRedirectDialog.story.jsx
+++ b/apps/src/applab/ExternalRedirectDialog.story.jsx
@@ -3,7 +3,6 @@ import {UnconnectedExternalRedirectDialog as ExternalRedirectDialog} from '@cdo/
 import {REDIRECT_RESPONSE} from './redux/applab';
 
 export default {
-  title: 'ExternalRedirectDialog',
   component: ExternalRedirectDialog,
 };
 
--- a/apps/src/applab/ImportProjectDialog.story.jsx
+++ b/apps/src/applab/ImportProjectDialog.story.jsx
@@ -3,7 +3,6 @@ import {ImportProjectDialog} from './ImportProjectDialog';
 import {action} from '@storybook/addon-actions';
 
 export default {
-  title: 'ImportProjectDialog',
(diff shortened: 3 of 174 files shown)


sanchitmalhotra126 (approved):
Looks great! Loving all the storybook improvements!

fisher-alice (approved):
Great idea!

Hamms (approved):
LGTM! I personally view adding a hierarchy that mirrors the hierarchy of `apps/` as an upside; we might lose some discoverability, but we gain a whole lot of internal consistency and predictability. Plus, significantly DRYer code!

Thanks!

bencodeorg:
🤔 not sure why `Checkbox` and `Chips` components are at the bottom of the Design System section...

levadadenys (approved):
image
LGTM! Thanks a lot for doing this! This is awesome!

levadadenys:
> 🤔 not sure why `Checkbox` and `Chips` components are at the bottom of the Design System section...

Interesting, on my machine everything is sorted alphabetically
image

bencodeorg:
@levadadenys oh perfect! I dunno what was going on in my screenshot then 😆

PR #56205, opened 2024-02-01
[Blockly] update math_random_int generators

mikeharv:
During yesterday's Sprite Lab "final checks", @ebeastlake noticed several levels where the start program wouldn't run and a workspace alert banner showed that the function `math_random_int` was undefined.

- https://studio.code.org/s/csc-ecosystems-2023/lessons/3/levels/2/sublevel/4 
- https://studio.code.org/s/csc-ecosystems-2023/lessons/3/levels/2/sublevel/2 
- https://studio.code.org/s/csc-ecosystems-2023/lessons/3/levels/2/sublevel/3  

image

Some custom blocks from the CSC pools, such as [ecosystems_makeNumSpritesWithinRang](https://levelbuilder-studio.code.org/pools/ecosystems/blocks/ecosystems_makeNumSpritesWithinRange/edit) included calls to this function in their helper code:

```
function makeNumSpritesWithinRange(numSprites, costume, range){
  for (var count = 0; count < numSprites; count++) {
    makeNewSpriteAnon(costume, {
      x: math_random_int(range.x1, range.x2),
      y: 400 - math_random_int(range.y1, range.y2)
    });
  }
}
```

The reason `math_random_int` was `undefined` took some time to discover. One red herring is that several of Sprite Lab's helper libraries explicitly define `math_random_int`, but none of the libraries were included in the affected levels. Interestingly, levels with one of these libraries would work fine, but we couldn't just fix it by adding a library because that would introduce other undesired consequences.
By inspecting the final [code](https://github.com/code-dot-org/code-dot-org/blob/mike/math_random_int/apps/src/p5lab/P5Lab.js#L1125) that was being passed into the JSInterpreter, @molly-moen and I noticed that CDO Blockly had a function called `math_random_int` but Google Blockly had `mathRandomInt`. This is evidently because the block's generator function was renamed to use camelcase instead of snakecase at some point in mainline Blockly. If blocks of type `math_random_int` were used, they would generate valid but different JavaScript in each version of Blockly that we currently support. 

CDO Blockly:

https://github.com/code-dot-org/blockly/blob/v4.1.0/generators/javascript/math.js#L533-L573

Google Blockly:

https://github.com/google/blockly/blob/master/generators/javascript/math.ts#L406-L429

This means that the block was working just fine on its own. The problem is that we had other custom blocks that relied upon the legacy spelling of the function name. It's probably a bit a weird for one block to call a function that is supplied by another block's generator (as opposed to functions defined in the Sprite Lab library or helper libraries), but here we are.

The solution I found was to copy the new generator from mainline into our `customBlocks` file and simply change the string used to create the function name. I used the `newBlockGenerator` helper we recently created as the API for generators is different between CDO Blockly and v10+ Google Blockly. While I was here I also updated one other generator for seldom-used block that was defined nearby.

The custom CSC blocks now work as expected (as evidenced by the creation of lots of happy animal sprites)!

image|image|image
-|-|-

## Links

Jira -https://codedotorg.atlassian.net/browse/CT-310

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/p5lab/spritelab/blocks.js
+++ b/apps/src/p5lab/spritelab/blocks.js
@@ -501,19 +501,27 @@ export default {
       },
       removeVar: Blockly.Blocks.variables_get.removeVar,
     };
-    generator.sprite_variables_get = function () {
-      return [
-        `{name: '${Blockly.JavaScript.translateVarName(
-          this.getFieldValue('VAR')
-        )}'}`,
-        Blockly.JavaScript.ORDER_ATOMIC,
-      ];
-    };
+    Blockly.customBlocks.defineNewBlockGenerator(
+      generator,
+      'sprite_variables_get',
+      function () {
+        return [
+          `{name: '${Blockly.JavaScript.translateVarName(
+            this.getFieldValue('VAR')
+          )}'}`,
+          Blockly.JavaScript.ORDER_ATOMIC,
+        ];
+      }
+    );
     Blockly.Variables.registerGetter(
       Blockly.BlockValueType.SPRITE,
       'sprite_variables_get'
     );
-
+    Blockly.customBlocks.defineNewBlockGenerator(
+      generator,
+      'math_random_int',
+      Blockly.customBlocks.mathRandomIntGenerator
+    );
     // NOTE: On the page where behaviors are created (the functions/#/edit page)
     // blockInstallOptions is undefined.
     if (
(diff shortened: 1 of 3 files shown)


apps/src/p5lab/spritelab/blocks.js
```diff
@@ -501,19 +501,27 @@ export default {
       },
       removeVar: Blockly.Blocks.variables_get.removeVar,
     };
-    generator.sprite_variables_get = function () {
```
molly-moen, line 504: out of curiosity: was this block broken as-is or was this just a clean up to make things easier in the future?

molly-moen (approved):
Nice!

apps/src/p5lab/spritelab/blocks.js
mikeharv: Just clean up!

PR #56206, opened 2024-02-01
Splitting the landing_pages file into two

hannahbergam:
History: This test file has been through it. The most salient information: the test flakiness has quieted down since the fix to check for image.complete. Additionally, it's worth noting that the full test file runs in about 8 minutes, sometimes as high as 12. 

Unfortunately, image.complete is true for *any* of the following cases, source (https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/complete): 
Screenshot 2024-02-01 at 2 38 37 PM

This means that when an image errors or fails to load, the image will get marked complete and the test will move past that step and then fail. However, this is because the image has truly failed to load, which I believe is an issue with the images we should fix, not an issue with the test. The test *should* fail in these scenarios. 

What I propose is a simple solution that should help. By splitting the two most likely to fail (Dance and Oceans) across two files, if the tests continue to have some issues, the reruns are much more likely to pass. Additionally, if a DOTD should have to review them manually, the files will run much faster because they only contain half the cases. 

## Links

- spec: []()
- jira ticket: https://codedotorg.atlassian.net/browse/ACQ-1353

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/dashboard/test/ui/features/hour_of_code/tutorial_landing_pages.feature
+++ b/dashboard/test/ui/features/hour_of_code/tutorial_landing_pages.feature
@@ -15,7 +15,4 @@ Examples:
   | http://code.org/minecraft                                         | minecraft tutorial landing |
   | http://code.org/playlab                                           | playlab tutorial landing   |
   | http://code.org/starwars                                          | starwars tutorial landing  |
-  | http://code.org/athletes                                          | athletes tutorial landing  |
-  | http://code.org/educate/applab                                    | app lab tutorial landing   |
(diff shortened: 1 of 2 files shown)


kelbyhawn (approved):
This looks good to me! I don't fully understand how tests work so you might want to wait for someone else to review before merging, but this makes sense based on your comment.

TurnerRiley (approved):

TurnerRiley:
> Unfortunately, image.complete is true for _any_ of the following cases

Dang this makes it much more complicated :(

PR #56208, opened 2024-02-01
Make a type for our global redux store

molly-moen:
A proposal for typing our global redux store. Right now when using typescript with redux you have to do slightly messy code like this:

```
  const code = useSelector(
    (state: {pythonlab: PythonlabState}) => state.pythonlab.code
  );
```

This is because we don't have a global type for our entire redux store. Redux recommends [inferring](https://redux-toolkit.js.org/tutorials/typescript#define-root-state-and-dispatch-types) the type, which we cannot do because we dynamically add to our store. They also [recommend](https://redux-toolkit.js.org/tutorials/typescript#define-typed-hooks) creating a typed `useSelector` hook, which we can do and which I did here (thanks @sanchitmalhotra126 for pointing this out in a different pr!)

Now the above code is:

```
  const code = useAppSelector(state => state.pythonlab.code);
```

Much neater!
We will have to manually add to`RootState` whenever we add or migrate a new slice in order to make this work. I was also able.

## Testing story
Tested locally that `PythonEditor` still works (the only component I updated to use the new hook). I also verified that usage of `useAppDispatch` also still compiles.

## Follow-up work
I updated `PythonEditor` to show how we would use the new hook, but once this is in we can go in and clean up our usage of `useSelector` to `useAppSelector` instead.

## PR Checklist:

- [ ] Code is well-commented


apps/src/types/reduxTypes.ts
sanchitmalhotra126, line 1: minor: do we have a preference for how to name files in this directory, with or without "Types" in the filename? Looks like we do have a "progressTypes" but also a "util" and "locale". I think i have a slight preference for not using "Types" since it's already in the /types/ directory, but open to having it this way too

apps/src/util/reduxHooks.ts
```diff
@@ -1,14 +1,14 @@
...
-// Right now our store isn't in typescript, and even if it was we would face the issue
-// that we define it as a singleton and therefore need to type it manually anyway, as it
-// could be null. Therefore we are manually defining the App Dispatch type here, and need
-// to use any because that's what redux-toolkit uses.
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-export type AppDispatch = ThunkDispatch<any, undefined, AnyAction> &
+// Right now our full store isn't in typescript, and we manually type what is in TypeScript
+// using RootState because our store is updated programmatically.
+// Therefore we are manually defining the App Dispatch type here.
+export type AppDispatch = ThunkDispatch<RootState, undefined, AnyAction> &
```
sanchitmalhotra126, line 10: 🎉 for removing the `any` here

sanchitmalhotra126 (approved):
Awesome! Thanks for adding this!

apps/src/types/reduxTypes.ts
molly-moen: I mostly didn't want yet another file called `redux`, but I am open to updating! Any thoughts on what we could call it other than `redux` or `reduxTypes`?

molly-moen: I do agree with not having `types` in the filename, so I'll rename to `redux` unless we can think of a better name.

sanchitmalhotra126: Ah good point, yeah I don't have any better ideas either 🤔 I guess most consumers would only ever be importing `useAppSelector` though rather than this type directly, so it likely won't be referenced that much anyway?

PR #49133, opened 2022-11-14
Updated Personal Projects Table Story

jmkulwik:
Updated the storybook entry for the PersonalProjectsTable story to work on the newer versions of Storybook.

We may want to consider deleting this one as it's not a reusable component. However, updating it was good practice and it now works, so keeping it around for now.

PR #56213, opened 2024-02-02
Updated the Personal Projects Table storybook entry

jmkulwik:
This PR was originally here: https://github.com/code-dot-org/code-dot-org/pull/49133 - I found it while cleaning out old branches and PRs. Since the branch had diverged so far from staging, it was better to just recreate it. Here is the original description:

---

Updated the storybook entry for the PersonalProjectsTable story to work on the newer versions of Storybook.

We may want to consider deleting this one as it's not a reusable component. However, updating it was good practice and it now works, so keeping it around for now.


levadadenys (approved):
LGTM!

PR #56214, opened 2024-02-02
 [labs-275] Dance AI: Clean up unused strings (ex. Strings with ‘2’ suffix)

cnbrenci:
Remove unused strings and rename the new strings with "2" suffixes back to original names.

## Links

- jira ticket: [labs-275 Dance AI: Clean up unused strings (ex. Strings with ‘2’ suffix)](https://codedotorg.atlassian.net/browse/LABS-275)

## PR Checklist:

- [ ] Code is well-commented


cnbrenci (commented):
I'm not sure how this will work for the translated strings- is it fine to use the same names as previously existed? Any weird fallback that might show the old strings briefly or something until they're translated? I want to make sure it's OK to re-use the original string names.

apps/i18n/dance/en_us.json
```diff
@@ -18,24 +18,19 @@
...
-  "danceAiModalFinding": "A.I. is finding the best effect to generate.",
-  "danceAiModalFinding2": "A.I. is finding the best effect for your emojis.",
-  "danceAiModalGenerating": "A.I. is generating this effect.",
-  "danceAiModalGenerating2": "A.I. thinks this is the best effect for your emojis.",
+  "danceAiModalFinding": "A.I. is finding the best effect for your emojis.",
+  "danceAiModalGenerating": "A.I. thinks this is the best effect for your emojis.",
   "danceAiModalEffectButton": "Effect",
-  "danceAiModalEffect": "A.I. generated this effect.",
-  "danceAiModalEffect2": "A.I. thinks this is the best effect for your emojis and generated the code!",
+  "danceAiModalEffect": "A.I. thinks this is the best effect for your emojis and generated the code!",
```
breville, line 24: Will this rename to an existing key have an effect on our translations?

cnbrenci: Yeah I was hoping you'd know the answer to that :P @mgc1194 do you know if this is going to be problematic? Do I need to avoid re-using the string names? I wanted to keep the ones that don't have "2" at the end, but if that's going to cause problems with translations I can just clean up the unused ones and leave the "2" ones.

PR #56220, opened 2024-02-02
[Design2-93] Design System Checkbox - Update/fix vertical alignment of checkbox and it's label

levadadenys:
### [Design2-93] Design System Checkbox - Update/fix vertical alignment of checkbox and it's label
After the font change we have a slight misalignment of checkbox content, this PR fixes it.

Before:
Знімок екрана 2024-02-02 о 18 23 15

After:
Знімок екрана 2024-02-02 о 18 23 06

## Links
- jira ticket: [DESIGN2-93](https://codedotorg.atlassian.net/browse/DESIGN2-93)

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/componentLibrary/checkbox/checkbox.module.scss
+++ b/apps/src/componentLibrary/checkbox/checkbox.module.scss
@@ -4,7 +4,7 @@
 .label {
   display: flex;
   gap: 8px;
-  align-items: start;
+  align-items: center;
 
   i {
     vertical-align: bottom;
@@ -20,7 +20,6 @@
 
   span {
     margin-bottom: 0;
-    line-height: initial;
   }
 
   input[type="checkbox"] {
--- a/apps/src/componentLibrary/checkbox/CHANGELOG.md
+++ b/apps/src/componentLibrary/checkbox/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 All notable changes to this project will be documented in this file.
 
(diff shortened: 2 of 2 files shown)


apps/src/componentLibrary/checkbox/checkbox.module.scss
```diff
@@ -20,7 +20,6 @@
 
   span {
     margin-bottom: 0;
-    line-height: initial;
```
fisher-alice, line 23: I see removing this line seems to increase the `line-height`. Just curious why you included this change. Also, where is this`initial` value defined? Thanks!

fisher-alice (approved):
LGTM! Just left a question out of curiosity. Thanks!

apps/src/componentLibrary/checkbox/checkbox.module.scss
levadadenys: Yep, it increases the line height to be what initially it should've be :D. Reason I added this `initial` value was to fix vertical alignment for previous font we were using.

Styles for this span are set in tsx part, we're using `Typography` `bodyText-${size}` styles here (they have line height)

fisher-alice: Thanks for the explanation!

sanchitmalhotra126 (approved):

PR #56222, opened 2024-02-02
Fix account settings page that fails with too many students

nicklathe:
Currently when a teacher with a large number of students in sections they own opens their `/users/edit` page, the page fails to load, as the query to get all students times out. Ultimately, this list of students is used in a warning message if they try to delete their account. It's not necessary or even helpful to display a long list of students, rather the importance is in the message and knowing that deleting their account will delete some number of student accounts as well. _Also_, the vast majority of teachers navigating to their `/users/edit` page aren't going there to delete their account.

This PR removes the logic and UI that displays a list of students, get's rid of the expensive query, and only displays a number of students to be deleted, in the warning message. This query is also limited to 100, as that should improve efficiency, as well as communicate the danger of deleting their account.

- Add new method in user.rb to query for students with a limit, and return count
- Change React components to take a count vs. list of students
- Remove UI that displays list of students
- Only show number of students to be deleted in warning message
- If queried student count equals limit, a `+` gets appended to the student count
   in the warning message to signify the number of students to be deleted is larger
   than can be displayed

Old dialogue box
image

New dialogue box
Screenshot 2024-02-01 at 3 49 54 PM

## Links

[Jira ticket](https://codedotorg.atlassian.net/browse/P20-658)

## PR Checklist:

- [ ] Code is well-commented


apps/src/lib/ui/accounts/PersonalLoginDialog.jsx
```diff
@@ -42,22 +39,16 @@ export default class PersonalLoginDialog extends React.Component {
           <p>
             <strong style={styles.dangerText}>
               {i18n.personalLoginDialog_body1({
-                numStudents: dependentStudents.length,
+                numStudents:
+                  dependentStudentsCount === dependentStudentsCountLimit
+                    ? `${dependentStudentsCount}+`
```
nicklathe: There's a lot of required logic to pass `dependentStudentsCount` all the way up to this component, all in an effort to communicate that there are _over_ 100 student accounts to be deleted. If we decide this isn't necessary, it can be removed.

nicklathe: Putting the `+` here vs. in the i18n strings file also means this isn't getting translated, however since it's a symbol, how would this get translated if it _were_ included in the i18n file?

dashboard/app/models/user.rb
```diff
@@ -2570,6 +2571,11 @@ def dependent_students
     dependent_students
   end
 
+  def dependent_students_count
```
daynew: Consider adding this to `queries/user.rb`. That file doesn't exist, but it would be easy to setup.

daynew (approved):
Thank you for fixing the long time issue. 
I just had one request.

apps/src/lib/ui/accounts/PersonalLoginDialog.jsx
daynew: Other languages might have an alternative to writing '+'. If we don't know, we can leave it to the translators to decide. It should probably be moved into the translation file.

dashboard/app/models/user.rb
nicklathe: Good call, I agree.

dashboard/app/models/user.rb
```diff
@@ -2570,6 +2571,11 @@ def dependent_students
...
+    # Limit the number of students to mitigiate slow page loads on /users/edit
+    students.limit(Policies::User::DEPENDENT_STUDENTS_COUNT_LIMIT).uniq.count
```
wilkie: I'm gonna test and stress this query out myself, but I have one question: is it that big of a deal if this returns a value less than the `DEPENDENT_STUDENTS_COUNT_LIMIT`? I think the subquery would return at most that limit, but since there may be a few duplicates, it would return a count less than that after the `uniq` or does the `uniq` qualify the subquery as `SELECT DISTINCT`? I think this is just ActiveRecord trivia, but I'm curious.

apps/src/lib/ui/accounts/PersonalLoginDialog.jsx
wilkie: we could either make it "... {count} or greater students ..." or have two strings for each case and have one be "... {count} students ..." and the other be "... over {count} students"

dashboard/app/models/user.rb
nicklathe: Oh, that's a good question. If we never reach `DEPENDENT_STUDENTS_COUNT_LIMIT` due to `uniq` trimming some of the returned students, the whole logic around communicating the overage (using `+` or `over`) is pointless, as we'd never reach the case where `dependentStudentCount === dependentStudentCountLimit`.

Issue #55759, opened 2024-01-16
Git LFS Migration


snickell:
### Advance preparation of systems:
1. [x] Merge PR adding git-lfs to chef + drone dockerfile, see: https://github.com/code-dot-org/code-dot-org/pull/53310
2. [x] Check staging, test & prod to ensure `git-lfs --version` >= 3.0
3. [x] Check levelbuilder to ensure `git-lfs --version` >= 3.0
4. [x] Post to #developers suggesting folks install git-lfs and verify their version is >= 3.0
5. [x] ~Investigate if we'll need to migrate our drone AMI and/or master and autoscaler servers for git-lfs support too~, after discussion we don't think this will be an issue and its hard to check, so we'll try a migration and if it doesn't work with drone, we'll back out and look into this.

### Practice conversion:
1. [x] Update file patterns to match https://docs.google.com/document/d/1YpB9CN-jswgZ3XTh4BBNrK-rf4RblyLse2cJu88aUjg
2. [x] Run a test conversion against current `staging` to make sure it hasn't gone stale since sept (conclusion: old conversion technique with bfg didn't work anymore, had to switch to `git lfs import`)
3. [x] Build migration script, see: https://gist.github.com/snickell/ca717b3266568b4ee05b0af20f5a25d8
4. [x] Do a full dry run on https://github.com/snickell/code-dot-org
5. [x] Validate the new repository works and performs as expected

### GitHub Migration:
1. [x] Levelbuilder - commit content before starting & warn content editors not to make change until after the work is complete (changes made during the transition will not be stash/pop-able)
2. [x] Notify #developers the day before, and day of migration that they'll need to push any branches and will need to check them out again after migration
5. [x] Add branch protection rule with pattern `*` and `[x] Lock Branch`. This will prevent unexpected pushes.
3. [x] Create a `git clone --mirror` of the repo in multiple locations, including AWS
4. [x] Start running migration script https://gist.github.com/snickell/ca717b3266568b4ee05b0af20f5a25d8 
6. [x] Wait for migration script to complete (several hours)
7. [x] Disable --force branch protections on staging, test, production, levelbuilder, staging-next: UPDATE, we were forced to actually delete these due to a bug in github, the last step is now recreating the branch protection rules.
8. [x] Run push in chunks section of the migration script
9. [x] Do a fresh git clone of staging from github
  - [x] verify that LFS files are materialized in the fresh clone
[...]

PR #56230, opened 2024-02-02
Music: Convert SoundsPanel and PatternPanel to TypeScript

sanchitmalhotra126:
Convert `SoundsPanel` and `PatternPanel` to TypeScript. Mostly the same, with a few function optimizations.


apps/src/music/views/SoundsPanel.tsx
```diff
@@ -105,34 +128,29 @@ const SoundsPanelRow = ({
         </div>
         <div className={styles.previewContainer}>
           <FontAwesome
+            title={undefined}
```
bencodeorg, line 131: What does this do?

bencodeorg (approved):

apps/src/music/views/SoundsPanel.tsx
sanchitmalhotra126: Oh nothing, TS just complains because it thinks it's a required prop 🙂 ideally I'd use the `FontAwesomeV6Icon` component, but it currently doesn't accept an onClick (this should probably actually be a button component wrapping an icon) so I figured we could address it as a follow up.

snickell:
This PR was accidentally auto-closed during the Git LFS migration (https://github.com/code-dot-org/code-dot-org/issues/55759), reopening it, sorry bout that 😘, -Seth

snickell:
This PR was accidentally auto-closed during the Git LFS migration (https://github.com/code-dot-org/code-dot-org/issues/55759), reopening it, sorry bout that 😘, -Seth

PR #56244, opened 2024-02-03
Use Docker image for Drone clone step with LFS support

snickell:
While our main drone test images have LFS installed, our drone build currently uses the `docker:git` image (which may be deprecated?), which does not include Git LFS, during its clone steps.

This PR switches our clone step to run on the `bitnami/git` image maintained by VMWare instead (https://hub.docker.com/r/bitnami/git). Image is popular on dockerhub with 10M+ pulls and active usage.

As a side note, disappointingly but not shockingly switching to LFS has caused our `git clone` stage to regress by 10%, it now takes 10:30 minutes instead of 9:30 minutes.


cat5inthecradle (approved):

snickell:
Success! https://drone.cdn-code.org/code-dot-org/code-dot-org/39489/1/2

PR #55983, opened 2024-01-26
Fix Ability to Provision a Full Stack System via CloudFormation

sureshc:
Continuation of work started in #55788

There are numerous issues that prevent our CloudFormation template from successfully provisioning a new adhoc with a "Full Stack" (all optional AWS Resources are enabled such as CloudFront, load balancer, Auto Scaling Group, memcached, RDS Aurora database cluster, etc.).

## PR Checklist:

- [ ] Tests provide adequate coverage
- [ ] Privacy and Security impacts have been assessed
- [ ] Code is well-commented
- [ ] New features are translatable or updates will not break translations
- [ ] Relevant documentation has been added or updated
- [ ] User impact is well-understood and desirable
- [ ] Pull Request is labeled appropriately
- [ ] Follow-up work items (including potential tech debt) are tracked and linked

PR #56255, opened 2024-02-04
Fix Ability to Provision a Full Stack via CloudFormation

sureshc:
Continues work post git-lfs migration that was started in #55983 

From #55983:
> There are numerous issues that prevent our CloudFormation template from successfully provisioning a new adhoc with a "Full Stack" (all optional AWS Resources are enabled such as CloudFront, load balancer, Auto Scaling Group, memcached, RDS Aurora database cluster, etc.).

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/aws/cloudformation/cloud_formation_stack.yml.erb
+++ b/aws/cloudformation/cloud_formation_stack.yml.erb
@@ -30,7 +30,7 @@ Parameters:
   # This domain name needs to be referenced both in Cloudformation and in the code for MarketingRouterLambda.
   # As a result, it is declared here in Ruby so that it can be passed to the Lambda function.
   # WARNING: Changing this value is NOT sufficient to change the value in the stack. It ONLY changes the default value.
-  # The value must be changed manually in the AWS Console or CLI, followed by an update to the stack. 
+  # The value must be changed manually in the AWS Console or CLI, followed by an update to the stack.
   <% INTERNAL_MARKETING_DOMAIN_NAME = "code.marketing-sites.#{rack_env?(:production) ? 'code.org' : 'test-code.org'}" %>
   InternalMarketingDomainName:
     Type: String
@@ -499,7 +499,7 @@ Resources:
               - Effect: Allow
                 Action:
                   - logs:CreateLogGroup
-                Resource: 
+                Resource:
                   - !Sub "arn:aws:logs:*:*:log-group:/aws/lambda/us-east-1.${AWS::StackName}-marketing-router:*"
                   - !Sub "arn:aws:logs:*:*:log-group:/aws/lambda/${AWS::StackName}-marketing-router:*"
               - Effect: Allow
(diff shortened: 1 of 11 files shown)


aws/cloudformation/cloud_formation_stack.yml.erb
```diff
@@ -611,7 +612,11 @@ Resources:
       IamInstanceProfile: !Ref DaemonInstanceProfile
       KeyName: <%=SSH_KEY_NAME%>
       Tags:
-        - {Key: Name, Value: <%=frontends ? "#{environment}-daemon" : "!Ref 'AWS::StackName'"%>}
+<%   if frontends -%>
+        - {Key: Name, Value: !Sub "${AWS::StackName}-daemon"}
+<%   else -%>
+        - {Key: Name, Value: !Ref AWS::StackName}
+<%   end -%>
```
cat5inthecradle, line 751: I believe this would change `production-daemon` to be renamed `autoscale-prod-daemon`.

aws/cloudformation/cloud_formation_stack.yml.erb
```diff
@@ -643,7 +648,7 @@ Resources:
           #!/bin/bash -x
           <%=indent(erb_file(BOOTSTRAP_CHEF,
             resource_id: daemon,
-            node_name: frontends ? '$ENVIRONMENT-daemon' : '$STACK',
+            node_name: frontends ? '$STACK-daemon' : '$STACK',
```
cat5inthecradle, line 784: Minor nit: Could we rename `$STACK` to `$STACKNAME` to align with the cloudformation template attribute?

aws/cloudformation/cloud_formation_stack.yml.erb
```diff
@@ -669,9 +674,13 @@ Resources:
...
-# We don't yet provision the production database cluster/instances via CloudFormation.
+    DependsOn:
+<%# We don't yet provision the production database cluster/instances via CloudFormation. %>
 <% if database && !rack_env?(:production) %>
-    DependsOn: DBProxy
+      - DBProxy
+<% end -%>
+<% if frontends-%>
+      - Memcached
 <% end -%>
```
cat5inthecradle, line 818: This could possibly generate an empty `DependsOn` array, which I believe is invalid YAML. You could use `!Ref AWS::NoValue` if you don't want to conditionally render the whole `DependsOn`

```yml.erb
<% if !frontends && !database -%>
      - !Ref AWS::NoValue
<% end -%>
```

aws/cloudformation/components/ami.yml.erb
```diff
@@ -51,16 +51,22 @@ frontend_device_name ||= '/dev/sda1'
   FrontendInstanceProfile:
     Type: AWS::IAM::InstanceProfile
     Properties: {Roles: [!Ref FrontendRole]}
-  # Signal when the instance is fully provisioned and ready for AMI creation.
+  # The WebServerAMI UserData shell script sends a signal when it completes successfully to this Resource.
   AMICreate<%=commit_hash%>:
+    DependsOn: [WebServerAMI]
```
cat5inthecradle, line 56: `DependsOn` blocks the creation of this resource, which could potentially lead to a race condition where this WaitCondition is not yet created when the waited-upon resource sends it's signal - makes me nervous to list it here. Would it make sense to instead have this WaitCondition mirror the `DependsOn` of the WebServerAmi resource, so they start creation at about the same time (keeping your timeout relevant)?

aws/cloudformation/components/ami.yml.erb
```diff
@@ -112,11 +118,12 @@ frontend_device_name ||= '/dev/sda1'
     LatestStackUpdateDateTime: Time.now # Trigger update of this custom Resource (invocation of the Lambda) for each Stack update.
   %>
   Frontends:
+    DependsOn: [ASGCount]
```
cat5inthecradle, line 122: Any `GetAtt` or `Ref` calls to another resource implicitly create a `DependsOn` relationship, so you don't need to explicitly list it here.

cat5inthecradle (commented):
Left various comments. Several TODO's have been left in place that would need to be dealt with.

Issue #55189, opened 2023-11-28
Build tool to import firebase JSON backup to MySQL


snickell:
Firebase contains a terabyte of data. The easiest way to slurp it we've found is using one of the firebase backups, which is a giant terabytes JSON file. This presents challenges, since we need to streaming parse the JSON since it won't fit in RAM.

See: https://github.com/code-dot-org/firebase-migration-tool

snickell:
After trying and failing to parse this scale of data with the first-pick ruby streaming parser gem, we're trying https://github.com/uhop/stream-json which seems popular in the nodejs world and is supposed to have good performance.

See: https://github.com/code-dot-org/code-dot-org/tree/unfirebase/unfirebase

We've discovered a few caveats:
1. We're only getting about 10MB/s speed on (streaming parse json => mysql insert), with the bottleneck appearing to be the streaming JSON parse not the mysql insert.
2. This means we're looking at ~24hrs+ for a full import
3. Projects can currently use an arbitrary (? check firebase rules) amount of disk space. As a result, we're having to run `'SET GLOBAL max_allowed_packet=10000000000;'` to support 8MB+ inserts. Unfortunately this sets max_allowed_packet for the lifetime of the mysql instance. We couldn't find a way around this.

snickell:
If we want a faster parse, we may need to consider a parser written in a lower level language like C/C++/Rust. Unfortunately we require not just speed, but a streaming API since we can't hold a terabyte in RAM. 

A general overview of really fast JSON parser perf: https://github.com/simdjson/simdjson#performance-results suggests anywhere from 0.5GB/s (50x current) to 7GB/s is achievable, with the major caveat that these charts are for the NON streaming form of the API.

If we're willing to use C++, we could use the "event based parser" approach to the simdjson API: https://simdjson.github.io/simdjson/md_doc_ondemand_design.html . Or the simpler (and possibly slower, but at some point DB perf is gonna be the bottleneck) SAX API from RapidJSON: https://rapidjson.org/md_doc_sax.html

With our existing `stream-json` based parser, we get **13MB/s without SQL** inserts (=153s on the 2GB big-data.json) and **10MB/s with SQL inserts** (=205s on the 2GB big-data.json) .

snickell:
[...]

PR #54643, opened 2023-11-01
Deprecating Firebase

snickell:
Tracking issue for the firebase deprecation project: https://github.com/orgs/code-dot-org/projects/4

This PR will probably not be merged, but we're writing a fair bit of code to build a bulk importer etc that we'll probably throw away (see #55189 ). We could have opened a new repo, but it seemed reasonable just to keep it on this branch.

PR #56276, opened 2024-02-05
Deprecating Firebase

snickell:
Tracking issue for the firebase deprecation project: https://github.com/orgs/code-dot-org/projects/4

This PR will probably not be merged, but we're writing a fair bit of code to build a bulk importer etc that we'll probably throw away (see #55189 ). We could have opened a new repo, but it seemed reasonable just to keep it on this branch.


snickell:
This PR is a continuation of #54643, which was mis-closed by the Git LFS migration (#55759).

### Previous Comments:
- **snickell** [commented](https://github.com/code-dot-org/code-dot-org/pull/54643#issuecomment-1789825783) - Another way of looking at this is not just where we use "firebase" in method names, but where do we actually use the firebase(TM) library. That's limited to two files:
 image
 
- **snickell** [commented](https://github.com/code-dot-org/code-dot-org/pull/54643#issuecomment-1789850667) - This is the one function used outside of `apps/storage/firebaseMetadata.js` that I wonder if we could easily handle through SQL, it feels like a firebase "subscription".... is it?
```
export function onColumnsChange(database, tableName, callback) {
  getColumnsRef(database, tableName).on('value', snapshot => {
    const columnsData = snapshot.val() || {};
    const columnNames = _.values(columnsData).map(column => column.columnName);
    callback(columnNames);
  });
}
```

In applab.j... [more](https://github.com/code-dot-org/code-dot-org/pull/54643#issuecomment-1789850667)
- **snickell** [commented](https://github.com/code-dot-org/code-dot-org/pull/54643#issuecomment-1789893398) - image

The only thing that looks like it might need subscription like functionality is `onRecordEvent`:
```
onRecordEvent("MyNewTable", function(record, eventType) {
  console.log(eventType);
  console.log(record);
});
```

- All of our APIs update one record at a time, using the record ID, so its easy on the backend to know exactly what change message to publis... [more](https://github.com/code-dot-org/code-dot-org/pull/54643#issuecomment-1789893398)
- **snickell** [commented](https://github.com/code-dot-org/code-dot-org/pull/54643#issuecomment-1789893611) - Idea: could we /just/ use redis? Like actually use that as our DB?
- **snickell** [commented](https://github.com/code-dot-org/code-dot-org/pull/54643#issuecomment-1789919051) - Discussing with @cnbrenci we naively might consider four approaches:
1. Back our data feature with MySQL and deprecate support for `onRecordEvent` (see question 1)
2. Back our data feature with MySQL, applab projects hold an open websocket (or poll???), and use a redis pubsub channel to broadcast `(tableId, recordId)` change events (or potentially the whole row value, either way), and then rails backend to the websocket sends the new row data to the client along with a changeEvent
3. Like #2,... [more](https://github.com/code-dot-org/code-dot-org/pull/54643#issuecomment-1789919051)
- **snickell** [commented](https://github.com/code-dot-org/code-dot-org/pull/54643#issuecomment-1800683832) - Per first convo with product, we may be willing to just deprecate `onRecordEvent`, breaking existing apps that use it. Conventional wisdom is that its mainly used by chat apps that are technically not allowed under our terms of service anyway, which might explain some of the high cost issues with firebase.

I'd personally like to see how widely used the `onRecordEvent` block is in Applab projects that have been used/run/loaded in, say, the last 3 months. Ideally used would mean loaded even jus... [more](https://github.com/code-dot-org/code-dot-org/pull/54643#issuecomment-1800683832)
- **snickell** [commented](https://github.com/code-dot-org/code-dot-org/pull/54643#issuecomment-1800728176) - Step #1 then would be to implement a script to find "percent of active applab projects using the onRecord event block". Maybe as a first step, we work on getting access to S3, looking up paths to relevant main.json student saves. Say we start with implementing `blockFinder` / block_finder.rb. Then you could do, say:

```
require_relative './block_finder'
require 'active_support/time'
blockFinder(blockName: 'onRecord', game: 'applab', lastActiveAfter: 90.days.ago.to_date)
```

@cnbrenci a... [more](https://github.com/code-dot-org/code-dot-org/pull/54643#issuecomment-1800728176)
- **snickell** [commented](https://github.com/code-dot-org/code-dot-org/pull/54643#issuecomment-1800824525) - Say we start by writing a script to determine `percent_active_applabs_using_on_record_event()`:

```
def sample_random_student_source(game, last_active_after)
  # TO IMPLEMENT
end

def source_contains_block(block_type)
  # TO IMPLEMENT
end

def days_to_seconds (days)
  return days * 24 * 60 * 60
end

def percent_applabs_actively_using_on_record_event()
  num_samples = 1000
  num_samples_using = 0
  for i in O.num_samples do
    source = sample_random_student_source(game: appla... [more](https://github.com/code-dot-org/code-dot-org/pull/54643#issuecomment-1800824525)
- **snickell** [commented](https://github.com/code-dot-org/code-dot-org/pull/54643#issuecomment-1803024430) - The "Projects" table (project.rb, Project model) has all the data we need:
- The projects table includes both updated_at and project_type="applab"
- the ID column of the Projects table is the "storage_app_id", which is the SECOND number in the S3 paths
- the storage_id column of the projects table is the FIRST number in the S3 paths

So we can go:
```
project = Project.where(project_type: 'applab', updated_at: 30.days.ago...).take!
s3_path_prefix = "sources" # or in dev: "sources_develop... [more](https://github.com/code-dot-org/code-dot-org/pull/54643#issuecomment-1803024430)
- **cnbrenci** [commented](https://github.com/code-dot-org/code-dot-org/pull/54643#issuecomment-1804859773) - > oject = Project.where(project_type: 'applab', updated_at: 30.days.ago...).take!
 > s3_path_prefix = "sources" # or in dev: "sources_development"
 > s3_path = "#{s3_path_prefix}/#{project.storage_id}/#{project.id}/main.json"
 > s3_bucket = "cdo-v3-sources"
 > source_code = AWS::S3.download_from_bucket(‘cdo-v3-sources’, s3_path)
 
 Confirmed that this does work (I successfully downloaded the source)
- **cnbrenci** [commented](https://github.com/code-dot-org/code-dot-org/pull/54643#issuecomment-1804898889) - open Q to look into: can Project.state be used to determine if the project is active rather than updated time? (I see in my dev env all my projects are state='active')
- **cnbrenci** [commented](https://github.com/code-dot-org/code-dot-org/pull/54643#issuecomment-1804922481) - First pass on full script after everything we figured out yesterday.

The last step would be we need to run it against the prod dashboard DB with the "sources" s3_path_prefix instead of "sources_development"

```ruby
#!/usr/bin/env ruby
require_relative('../config/environment')

# Gets a sampling of applab source code from s3
# computes the percentage of those sources that use the onRecordEvent block

def percent_projects_actively_using_block(project_type, block_type, last_active_afte... [more](https://github.com/code-dot-org/code-dot-org/pull/54643#issuecomment-1804922481)
- **snickell** [commented](https://github.com/code-dot-org/code-dot-org/pull/54643#issuecomment-1805128299) - Nice!! Curious what percent we're gonna end up with 🤞 when we run on production.

- We should probably use random samples rather than whatever MySQL happens to give for a default ordering. I generally avoid using DB-specific SQL, but since we're just a one-off script.... `Project.where(...).order('RANDOM()').take(sample_size)` (https://www.geeksforgeeks.org/sql-select-random/). This does require the DB to internally walk the entire set of matching rows, join with a random number each, and then... [more](https://github.com/code-dot-org/code-dot-org/pull/54643#issuecomment-1805128299)
- **snickell** [commented](https://github.com/code-dot-org/code-dot-org/pull/54643#issuecomment-1809461422) - We sampled the prod DB to find instances of the `onRecordEvent` block being used in applab projects. Where we found instances, we read the code to see if they were broken or TOS violations.

TL;DR: **0.002% of applab projects (that aren't stubs or chat apps) created in the last 2 months use the `onRecordEvent` block, with a sample size of 100k.**

1. Sampling applab projects created in the last 2 months, 6 out of 100000 applab projects created in the last 2 months used `onRecordEvent`. We re... [more](https://github.com/code-dot-org/code-dot-org/pull/54643#issuecomment-1809461422)
- **snickell** [commented](https://github.com/code-dot-org/code-dot-org/pull/54643#issuecomment-1813699715) - Product noticed (TY @moneppo) we hadn't considered that students use different blocks throughout the year as they complete different curriculum. 

So we went back to the sampling, and grabbed 105k samples covering all of 2022...

We scanned 2022 (samples from each month), and **we found 3 "valid" apps in 2022 out of 105k samples (=0.003%, about the same as the "last two months" sample)**. "Valid" meant apps that weren't either chat apps, or stub code (e.g. "drag some blocks out"). We found 1... [more](https://github.com/code-dot-org/code-dot-org/pull/54643#issuecomment-1813699715)
- **snickell** [commented](https://github.com/code-dot-org/code-dot-org/pull/54643#issuecomment-1815763762) - We couldn't figure out how people were writing gamelab based chat apps. I found the "secret" gamelab APIs thanks to Hannah Nees giving me a great sample of "banned chat apps", including samples from Gamelab. These use "setKeyValue" and "getKeyValue" functions.... which aren't exposed as blocks (?!?).

Relevant code is here: https://github.com/code-dot-org/code-dot-org/blob/e6905b0c899383ca2c7e43885f847cedf84478b2/apps/src/p5lab/gamelab/commands.js/#L25-L47

The PR that added them is here: ht... [more](https://github.com/code-dot-org/code-dot-org/pull/54643#issuecomment-1815763762)
- **davidsbailey** [commented](https://github.com/code-dot-org/code-dot-org/pull/54643#issuecomment-1815816682) - > @davidsbailey any chance you remember why we were adding these APIs without adding them as blocks?
 
 no, sadly I do not think I was involved in this decision.
- **snickell** [commented](https://github.com/code-dot-org/code-dot-org/pull/54643#issuecomment-1899550847) - - Next we should start implementing more methods from the firebase interface we extracted, and get corresponding tests passing.
 - We should also consider adding tests that actually create student source using the data blocks, and thereby test end to end, from JS api through the Rails controller. I think these tests, while a little more bulky, will actually catch much more and more helpful potential breakages in the future.

### Previous Reviews:

Issue #55344, opened 2023-12-08
Design performant schema for formerly-Firebase data


snickell:
Part of the firebase deprecation project: https://github.com/code-dot-org/code-dot-org/issues/55084

We have approximately 4 billion records currently in Firebase (by record we mean one row in a student project dataset), spread across millions of student data tables (contained in millions of applab projects).

Now that we have full data imports going (#55189) we can start performance testing possible schemas to store this data in MySQL.

1. Store each student-channel (=student project) in a mysql row
2. Store each student-table (multiple tables per channel) in a JSON mysql row
3. Store each student-table in a TEXT mysql row?
4.  Store each student-record (multiple records per table) in a TEXT mysql row

snickell:
Basic SELECT statements on the record-per-row table `unfirebase.records` show excellent performance:
![Image](https://github.com/code-dot-org/code-dot-org/assets/223277/42ee260b-ac68-4009-a77e-53dde77b5b10)

I believe (?) this is only half the total rows (and includes stock table data), but should be generally representative:
![Image](https://github.com/code-dot-org/code-dot-org/assets/223277/4a127670-1f2a-49eb-95ed-f73c9c0d12b2)

Indices are on a composite PK (channel_id, table_name, record_id) as well as on (channel_id), and take about 20% of storage space. **Queries that use `channel_id=` take approximately 0.25s over 1.2 billion rows.**

snickell:
Row-per-record schema with table dedupe seems to give favorable query performance (of the queries we see in `apps/src/storage/firebaseStorage.js`) and reasonable storage
1. Uses row-per-record for table schema
2. Dedupes whole tables, and can even dedupe tables when running in `KIND_OF_ROW=Record`. Reduces total DB byte size to 1/5th of before.
3. Stores records as the MySQL JSON column type. I find this re-assuring because it means 100% of the individual records have passed MySQL's json validation.... ="our data is probably imported correctly :sweat_smile:"
4. Only takes mysql 370GB on-disk for the DB.... including totally working perf-wise indices!
5. Schema shows good performance: fetching a full table of records takes ~0.25s (=readRecord block), individual record inserts are 0.14s (=createRecord block), individual record updates are also 0.14s (=updateRecord block).

![Image](https://github.com/code-dot-org/code-dot-org/assets/223277/023866bb-c7d1-4bd4-9499-fa61b29777b6)

[...]

Issue #55481, opened 2023-12-15
Figure out Rate Limiting for Datasets on MySQL


snickell:
### Rate Limiting

In Firebase we maintain a rate limit per project/channel, 300 writes per 15 seconds, 600 writes per 60 seconds.

1. Where does this data rate come from? Its pretty high. Is this required by the microbit projects?
2. On MySQL we see inserts taking 0.15s per, which if we're not inserting in parallel gives us an insert rate of about 6 inserts/s, which is 90 writes per 15 seconds or 360 per 60 seconds. This is less than our existing limits.
3. Should we be batching these together?

```
{
  "counters": {
    "limits": {
      "15": {
        "lastResetTime": 1654619184297,
        "writeCount": 175
      },
      "60": {
        "lastResetTime": 1654524726608,
        "writeCount": 188
      }
    }
  },
  "serverTime": 1661532798036,
```

snickell:
We're using encrypted cookies for sessions storage, so it should be fast/performant to keep rate counters per-user if that's the approach we want to follow.

cnbrenci:
Firebase is doing async calls and using a websocket to be able to achieve the throughput. It doesn't have to do a roundtrip per transaction. We're investigating how we would deal with auto incrementing the record_id part of the composite primary key without  locking the entire table. Here's a working guess on a query to use:


 ```sql
BEGIN;
  SELECT @id := IFNULL(MAX(record_id),0) + 1 FROM unfirebase.records WHERE channel_id='shared' AND table_name='words' FOR UPDATE;
  INSERT INTO unfirebase.records
     VALUES
     ('shared', 'words', @id, '{}');
  COMMIT;
  
  SELECT * FROM unfirebase.records WHERE channel_id='shared' AND table_name='words' and record_id > 4983;
```
todo: update the quary when we figure out a better one

```sql
# set profiling=1;
BEGIN;
  SELECT MAX(record_id) FROM unfirebase.records WHERE channel_id='shared' AND table_name='Rabbits' FOR UPDATE;
  SELECT @id := IFNULL(MAX(record_id),0)+1 FROM unfirebase.records WHERE channel_id='shared' AND table_name='Rabbits';
  DO SLEEP(10);
  INSERT INTO unfirebase.records
     VALUES
     ('shared', 'Rabbits', @id, '{}');
COMMIT;
  
# show profiles;

# SELECT * FROM unfirebase.records WHERE channel_id='shared' AND table_name='Rabbits';
```

[...]

PR #51439, opened 2023-04-20
Create common util for retrieving CSRF token

sanchitmalhotra126:
Creates a common utility helper to fetch the Rails authenticity (CSRF) token. This is needed for most POST requests made by the browser. See [thread](https://codedotorg.slack.com/archives/C0T0PNR0D/p1681928133278129) for more discussion and reasoning.

As an eventual follow-up, we should replace existing usages of alternate routes to fetch the CSRF token with this method (ex. in Code Review, Teacher Feedback, project_commits#get_token, user_levels#get_token). We can also replace code that accesses the `meta` element directly with this util.

A possible future improvement could be to create a standalone HTTP client, specifically for communicating with dashboard endpoints, that fetches and adds the token to requests as needed. However, a lot of our API calls use fetch or $.ajax directly, so it would require converting all of those.

## Testing story

Tested locally.

PR #11951, opened 2016-11-17
Add Data APIs to Game Lab (no toolbox blocks or autocomplete)

joshlory:
Follow up to PR https://github.com/code-dot-org/code-dot-org/pull/11895.

Adds `getUserId`, `getKeyValue` & `setKeyValue` APIs to Game Lab.  Does not add the corresponding blocks to the toolbox or allow them to show up in the autocomplete list.

PR #56279, opened 2024-02-05
Datablock Storage: replace firebase for applab data features

snickell:
# Datablock Storage

This PR implements a **Rails/MySQL backend to the Applab data features** currently backed with Firebase:

1. The Data Library:
    image
2. The Data blocks:
    image
3. The data browser, including KVP editing, Table editing, etc:
    image

This PR **does not immediately switch to Datablock Storage**, in fact when this PR is merged by default 0% of Applab projects will use it. Instead a gradual rollout concurrent with bugfixing and optimization is envisioned, see "Rollout plan" below.

### Motivation

This project is part of the "Eng Excellence" series. Primary motivation is cost (Firebase is on track to cost us $20k/month and it rises significantly each year), secondary motivation is consolidating the technology we are using: datablock storage uses our existing MySQL DB. This provides small advantages like local-dev using your existing DB, vs firebase where all local dev actually shares a DB, leading to weird behavior like editing the dataset manifests in local dev would actually modify production (!!!). Firebase provides very cool live subscription features, but in practice we did not use these in our curriculum.

## File changes in this PR fall into one of three categories:
1. New MySQL backed storage implementation including new models, controller, and javascript API:
    - API [**datablock_storage_controller.rb**](https://github.com/code-dot-org/code-dot-org/blob/unfirebase/dashboard/app/controllers/datablock_storage_controller.rb) and its frontend wrapper [**datablockStorage.js**](https://github.com/code-dot-org/code-dot-org/blob/unfirebase/apps/src/storage/datablockStorage.js). One of the easiest places to see the overall API is in [routes.rb](https://github.com/code-dot-org/code-dot-org/blob/2727e5ee8eee5575e21957c867c95c9d81b2db45/dashboard/config/routes.rb#L1122-L1170)
    - Testing: [datablock_storage_controller_tests.rb](https://github.com/code-dot-org/code-dot-org/blob/unfirebase/dashboard/test/controllers/datablock_storage_controller_test.rb): also includes coverage for all conditions in FirebaseStorageTest.js
    - Backend models: [**datablock_storage_table.rb**](https://github.com/code-dot-org/code-dot-org/blob/unfirebase/dashboard/app/models/datablock_storage_table.rb), [datablock_storage_kvp.rb](https://github.com/code-dot-org/code-dot-org/blob/unfirebase/dashboard/app/models/datablock_storage_kvp.rb), [datablock_storage_record.rb](https://github.com/code-dot-org/code-dot-org/blob/unfirebase/dashboard/app/models/datablock_storage_record.rb)
6. Refactors to fully encapsulate Firebase-related calls behind an interface that could be implemented in parallel in `datablockStorage.js`, allowing for the approach of both code paths existing alongside each-other, switching on configuration when initializing the data store.
    - New methods added to: [firebaseStorage.js](https://github.com/code-dot-org/code-dot-org/blob/2727e5ee8eee5575e21957c867c95c9d81b2db45/apps/src/storage/firebaseStorage.js#L1097-L1250). These are extracted from Firebase-specific code scattered through JSX and applab.js, and created an interface that can then be implemented in datablockStorage.js.
    - [storage.js](https://github.com/code-dot-org/code-dot-org/blob/unfirebase/apps/src/storage/storage.js) breaks direct coupling between storage-accessing code and firebaseStorage.js. A lot of the changes to files outside this core set are simply replacing a hardcoded `import FirebaseStorage` import with `import {storageBackend} from '../storage'`
7. Todos & references to follow-up issues, and work to be done after a full rollout to remover Firebase code (`// TODO: post-firebase-cleanup`): #56994

## UX Changes

Ideally this is a backend-only change without student-facing, curriculum-facing or teacher-facing changes.

### New features
- Rather than copying static tables each time they are selected in the dataset browser, Datablock Storage implements a copy-on-write semantic. Only when a table is modified is it copied. This is because we analyzed the 1TB of JSON we currently have on Firebase and discovered that the majority of it is unchanged copies of our static tables.
- As a result, static tables and "current" tables (covid, spotify, daily weather) are equivalent
- An optimized `getColumn` block doesn't fetch the whole table client-side every time. This is important because curriculum has told us this is the main block they use, and many datasets like Ramen are challenging on student machines as a result (timing out, etc).

## Rollout plan

This code implements the core of the student-facing data-in-applab experience, but does NOT include curriculum-facing dataset editing interfaces. During rollout, we'll request curriculum authors to notify us if they edit datasets (only happens a few times a year and we've already started the conversation with them) and we'll manually sync the changes to Datablock Storage.

While the AP CSP create task may result in delaying the final rollout until its completed, we would like to do at least at least do a test-the-waters rollout prior to it. Its possible we'll decide at that point to fully switch back to Firebase storage, or might feel confident to go further depending on how things look.

1. Test the waters: this PR does NOT switch to Datablock Storage by default, instead a DCDO flag can be used to control the percentage of newly created projects that will use Datablock Storage (defaults to 0%). Initially, we plan to do a 1-day trial with <1% of projects created on that day using Datablock Storage. We'll triage any issues we find, and analyze the performance impact on our DB and rails backend.
6. Gradual to rollout, fixing and optimizing as needed: from there, we'll determine what we need to optimize now, what we need to fix now, and do so if necessary. Then, slowly day by day or week by week, we'll increase the percentage of new projects that run on Datablock Storage until 100% are on the new backend.
7. Once we've got a majority of new projects running on Datablock Storage, we'll migrate the Dataset editor code to Datablock Storage by default.
8. Migrate old projects: once 100% of new projects are on Datablock Storage, we'll start converting old projects. This process will include deduplicating the many copies of our static tables (currently most of the 1TB of data stored in firebase is copies). More information on migration work can be [found here](https://github.com/code-dot-org/code-dot-org/issues/55189)
9. Post-firebase-cleanup: we've marked extensively throughout the code what to remove when Firebase is no longer needed. After everything is migrated, we'll do a followup PR that removes all Firebase code paths.

## Data Model

The [data model](https://github.com/code-dot-org/code-dot-org/issues/55344) is very simple, we initially thought we might have to try fancy things to deal with the scale of data involved, but we tried importing all 1TB of Firebase data into various MySQL schemas and found a very basic approach to be performant (various query and data structures were timed and optimized, leading to the ActiveRecord calls we have now).

- Table/Record data: a single row-per-table in DatablockStorageTables, includes table metadata including its name, the columns on the table (json blob), and an indicator whether this is a pointer to a table, or a table. If its a pointer, row-data will be loaded from the corresponding "shared table". Table records are stored in DatablockStorageRecords table one-record-per-row as a record_id and a JSON typed column, keyed by a composite key: project_id and table_name.
- KVP data: a single row-per-kvp in the DatablockStorageKvps table
- Data Library manifest data: this is stored as a singleton JSON row in the DatablockStorageLibraryManifest table. We decided not to decompose this data because its already edited, saved, and loaded/used as a single JSON blob in the existing datasets controller.

The most challenging issue was dealing with auto-incrementing the record_id relative to the composite key. For this, we lock the first row in the table when computing a new record ID.

### Data Migrations

At @bethanyaconnor 's suggestion, we've split the migrations out into a separate PR to be applied first: https://github.com/code-dot-org/code-dot-org/pull/57392

## Followup Work

This PR implements the core student-facing data features. Anticipated followups include possible backend optimizations, and UX bug fixes. Followup work is [being tracked on a GitHub kanban board](https://github.com/orgs/code-dot-org/projects/4)

Major required followups include:
1. Rate limiting checks
2. Converting the datasets controller and associates to Datablock Storage

Release critical issues can be found here: https://github.com/orgs/code-dot-org/projects/4/views/10

Co-authored-by: Cassi Brenci <cassi.brenci@code.org>


snickell:
This PR is a continuation of #54643, which was mis-closed by the Git LFS migration (#55759).

### Previous Comments:
- **snickell** [commented](https://github.com/code-dot-org/code-dot-org/pull/54643#issuecomment-1789825783) - Another way of looking at this is not just where we use "firebase" in method names, but where do we actually use the firebase(TM) library. That's limited to two files:
 image
 
- **snickell** [commented](https://github.com/code-dot-org/code-dot-org/pull/54643#issuecomment-1789850667) - This is the one function used outside of `apps/storage/firebaseMetadata.js` that I wonder if we could easily handle through SQL, it feels like a firebase "subscription".... is it?
 ```
 export function onColumnsChange(database, tableName, callback) {
   getColumnsRef(database, tableName).on('value', snapshot => {
     const columnsData = snapshot.val() || {};
     const columnNames = _.values(columnsData).map(column => column.columnName);
     callback(columnNames);
   });
 }
 ```
 
 In applab.js this is just used to update data preview pages when columns are renamed, seems like that doesn't necessitate a full firebase-level DB.
- **snickell** [commented](https://github.com/code-dot-org/code-dot-org/pull/54643#issuecomment-1789893398) - image
 
 The only thing that looks like it might need subscription like functionality is `onRecordEvent`:
 ```
 onRecordEvent("MyNewTable", function(record, eventType) {
   console.log(eventType);
   console.log(record);
 });
 ```
 
 - All of our APIs update one record at a time, using the record ID, so its easy on the backend to know exactly what change message to publish
 - `onRecordEvent` gives you the JSON record contents, including the record.id, and whether it was a create, update, or delete event. Events are all about ONE ROW.
- **snickell** [commented](https://github.com/code-dot-org/code-dot-org/pull/54643#issuecomment-1789893611) - Idea: could we /just/ use redis? Like actually use that as our DB?
- **snickell** [commented](https://github.com/code-dot-org/code-dot-org/pull/54643#issuecomment-1789919051) - Discussing with @cnbrenci we naively might consider four approaches:
 1. Back our data feature with MySQL and deprecate support for `onRecordEvent` (see question 1)
 2. Back our data feature with MySQL, applab projects hold an open websocket (or poll???), and use a redis pubsub channel to broadcast `(tableId, recordId)` change events (or potentially the whole row value, either way), and then rails backend to the websocket sends the new row data to the client along with a changeEvent
 3. Like #2, but back our data feature directly with redis, skip the whole mysql in the middle bit. Like aurora for mysql, AWS has a hosted durable redis option, and the pricing is very reasonable.... something like $600/mo would get us 12gbps of redis bandwidth.
 
 Questions to answer:
 1. How actively used by student projects is `onRecordEvent`? How many calls to onRecordEvent do we see per day? How many projects use it? TODO: download all student code that uses onRecordEvent and skim through them, see how frequent it is overall, how its generally used, etc
 2. How many concurrent applab users do we see? Can we hold that many websockets open with our existing rails infra? If we don't have an approach to holding lots websockets open, is there an approach to this we can implement? How hard is it?
- **snickell** [commented](https://github.com/code-dot-org/code-dot-org/pull/54643#issuecomment-1800683832) - Per first convo with product, we may be willing to just deprecate `onRecordEvent`, breaking existing apps that use it. Conventional wisdom is that its mainly used by chat apps that are technically not allowed under our terms of service anyway, which might explain some of the high cost issues with firebase.
 
 I'd personally like to see how widely used the `onRecordEvent` block is in Applab projects that have been used/run/loaded in, say, the last 3 months. Ideally used would mean loaded even just to use the student applab app as a user, vs developed on, but if we can't get "when was this last run by anyone", we could use "code was changed in the last 6 months" as a proxy.
 
 From this we can compute: "**percent of active applab projects using the onRecord event block**", which can then inform a decision to just deprecate `onRecordEvent`, making option #1 above viable, and not having to add a need to hold constant websockets open to our rails server (expensive and possibly not scalable).
- **snickell** [commented](https://github.com/code-dot-org/code-dot-org/pull/54643#issuecomment-1800728176) - Step #1 then would be to implement a script to find "percent of active applab projects using the onRecord event block". Maybe as a first step, we work on getting access to S3, looking up paths to relevant main.json student saves. Say we start with implementing `blockFinder` / block_finder.rb. Then you could do, say:
 
 ```
 require_relative './block_finder'
 require 'active_support/time'
 blockFinder(blockName: 'onRecord', game: 'applab', lastActiveAfter: 90.days.ago.to_date)
 ```
 
 @cnbrenci and I tracked down where applab student source is being saved, and its in main.json files in S3, e.g.: https://s3.console.aws.amazon.com/s3/object/cdo-v3-sources?region=us-east-1&prefix=sources/50000063/131542020/main.json
- **snickell** [commented](https://github.com/code-dot-org/code-dot-org/pull/54643#issuecomment-1800824525) - Say we start by writing a script to determine `percent_active_applabs_using_on_record_event()`:
 
 ```
 def sample_random_student_source(game, last_active_after)
   # TO IMPLEMENT
 end
 
 def source_contains_block(block_type)
   # TO IMPLEMENT
 end
 
 def days_to_seconds (days)
   return days * 24 * 60 * 60
 end
 
 def percent_applabs_actively_using_on_record_event()
   num_samples = 1000
   num_samples_using = 0
   for i in O.num_samples do
     source = sample_random_student_source(game: applab, last_active_after: Time.now - days_to_seconds(90))
     if source_contains_block(source)
       nun_samples_using += 1
     end
     return num_samples_using / num_samples
   end
 
   puts "Percent of recent applab projects using 'onRecordEvent':"
   puts(percent_applabs_actively_using_on_record_event())
 end
 
 ```
 
 @cnbrenci and I tracked down where applab student source is being saved, and its in main.json files in S3, e.g.:
 https://s3.console.aws.amazon.com/s3/object/cdo-v3-sources?region=us-east-
 1&prefix=sources/50000063/131542020/main.json
 
 - [x] implement `sample_random_student_source()`
   - [x] research (then implement) how to get list of student applab sources active in last N days using rails DB models in `dashboard/app/models`
   - [x] research (then implement) how to get S3 path to `cdo-v3-sources` bucket for the `main.json` given applab source (using more rails DB models I think?)
   - [x] fetch relevant `main.json` path from S3, possibly using `lib/cdo/aws/s3.rb`
 - [x] implement `source_contains_block()`
   - [x] parse main.json, determine if it contains the 'onRecordEvent' block
 - [x] get relevant S3 credentials, and run sampler on real student source code
- **snickell** [commented](https://github.com/code-dot-org/code-dot-org/pull/54643#issuecomment-1803024430) - The "Projects" table (project.rb, Project model) has all the data we need:
 - The projects table includes both updated_at and project_type="applab"
 - the ID column of the Projects table is the "storage_app_id", which is the SECOND number in the S3 paths
 - the storage_id column of the projects table is the FIRST number in the S3 paths
 
 So we can go:
 ```
 project = Project.where(project_type: 'applab', updated_at: 30.days.ago...).take!
 s3_path_prefix = "sources" # or in dev: "sources_development"
 s3_path = "#{s3_path_prefix}/#{project.storage_id}/#{project.id}/main.json"
 s3_bucket = "cdo-v3-sources"
 source_code = AWS::S3.download_from_bucket(‘cdo-v3-sources’, s3_path)
 ```
 
 Now we need to figure out how to do `download_from_s3()`
- **cnbrenci** [commented](https://github.com/code-dot-org/code-dot-org/pull/54643#issuecomment-1804859773) - > oject = Project.where(project_type: 'applab', updated_at: 30.days.ago...).take!
 > s3_path_prefix = "sources" # or in dev: "sources_development"
 > s3_path = "#{s3_path_prefix}/#{project.storage_id}/#{project.id}/main.json"
 > s3_bucket = "cdo-v3-sources"
 > source_code = AWS::S3.download_from_bucket(‘cdo-v3-sources’, s3_path)
 
 Confirmed that this does work (I successfully downloaded the source)
- **cnbrenci** [commented](https://github.com/code-dot-org/code-dot-org/pull/54643#issuecomment-1804898889) - open Q to look into: can Project.state be used to determine if the project is active rather than updated time? (I see in my dev env all my projects are state='active')
- **cnbrenci** [commented](https://github.com/code-dot-org/code-dot-org/pull/54643#issuecomment-1804922481) - First pass on full script after everything we figured out yesterday.
 
 The last step would be we need to run it against the prod dashboard DB with the "sources" s3_path_prefix instead of "sources_development"
 
 ```ruby
 #!/usr/bin/env ruby
 require_relative('../config/environment')
 
 # Gets a sampling of applab source code from s3
 # computes the percentage of those sources that use the onRecordEvent block
 
 def percent_projects_actively_using_block(project_type, block_type, last_active_after, sample_size)
   projects = Project.where(project_type: project_type, updated_at: last_active_after...).take(sample_size)
   s3_path_prefix = "sources_development"
   # s3_path_prefix = "sources" # or in dev: "sources_development"
   s3_bucket = "cdo-v3-sources"
 
   projects_have_block = projects.map do |project|
     s3_path = "#{s3_path_prefix}/#{project.storage_id}/#{project.id}/main.json"
     file_contents = AWS::S3.download_from_bucket(s3_bucket, s3_path)
     contains_block = JSON.parse(file_contents)["source"].include?("onRecordEvent(")
     contains_block
   end
 
   percent_using_block = (projects_have_block.count(true).to_f / projects_have_block.size) * 100
   puts "Percentage of #{project_type} projects using the #{block_type} block is #{percent_using_block}"
 end
 
 def percent_applabs_actively_using_on_record_event
   num_samples = 1000
   timeframe = 30.days.ago
   percent_projects_actively_using_block("applab", "onRecordEvent", timeframe, num_samples)
 end
 
 puts "Percent of recent applab projects using 'onRecordEvent':"
 puts(percent_applabs_actively_using_on_record_event)
 ```
 
- **snickell** [commented](https://github.com/code-dot-org/code-dot-org/pull/54643#issuecomment-1805128299) - Nice!! Curious what percent we're gonna end up with 🤞 when we run on production.
 
 - We should probably use random samples rather than whatever MySQL happens to give for a default ordering. I generally avoid using DB-specific SQL, but since we're just a one-off script.... `Project.where(...).order('RANDOM()').take(sample_size)` (https://www.geeksforgeeks.org/sql-select-random/). This does require the DB to internally walk the entire set of matching rows, join with a random number each, and then sort N rows. So it would probably suck if we had a million results.... but in reality: I'm betting this will work in our case if "applabs updated in the last N days" isn't too huge.
 - In reality on prod we might want to do 10k samples instead of 1k since I'm guessing we're going to get a value thats less than 1%.
 - Thinking 90 days might be a good date range
 - I think we should save each file_contents to a .json file, and have a copy of those in a tarball or whatever, because the next question after we get a percentage is probably for you and I to pick a dozen or so "uses onRecord" code bits, and loosely read them to find out WHAT its being used for (notably: "is it a chat app?" since we don't consider nixing things against our TOS as a deprecation)
 
 I guess a next step is figuring out how/where we can run this to have production DB creds (preferably read-only) available. One option, which might actually be simplest, is to run this on our `production-console` instance.
- **snickell** [commented](https://github.com/code-dot-org/code-dot-org/pull/54643#issuecomment-1809461422) - We sampled the prod DB to find instances of the `onRecordEvent` block being used in applab projects. Where we found instances, we read the code to see if they were broken or TOS violations.
 
 TL;DR: **0.002% of applab projects (that aren't stubs or chat apps) created in the last 2 months use the `onRecordEvent` block, with a sample size of 100k.**
 
 1. Sampling applab projects created in the last 2 months, 6 out of 100000 applab projects created in the last 2 months used `onRecordEvent`. We read the source code for the 6 projects, and found 4 of the 6 were chat apps or non-functional stub code. Thus only **2 out of 100000 or 0.002%** of this sample would be of concern.
     - "who needs assistance" crud app, 1200LOC
     - daily habit tracker, 160LOC (single-user: could be easily written w/o onRecordEvent)
     - 3x non-functional stub code
     - 1x chat app
 
 2. In case onRecordEvent was more used for older projects with active current usage, we sampled older applab projects that were updated_at the last 3 months, 7 out of 9000 older projects used `onRecordEvent`. We read the source code for the 7 projects, and found 5 out of 7 where either chat apps or non-functional stub code. Thus only **2 out of 9000 or 0.02%** of this sample would be of concern if we remove the `onRecordEvent` block.
     - 4x chat app
     - 1x non-functional stub code
     - "multiplayer farming simulator", 600LOC
     - math game (used for high score table), 200LOC
  
 Also worth noting that none of the apps appear to be finished to an extent that I think its likely anyone is actively using them in a multiplayer context, including the chat apps.
- **snickell** [commented](https://github.com/code-dot-org/code-dot-org/pull/54643#issuecomment-1813699715) - Product noticed (TY @moneppo) we hadn't considered that students use different blocks throughout the year as they complete different curriculum. 
 
 So we went back to the sampling, and grabbed 105k samples covering all of 2022...
 
 We scanned 2022 (samples from each month), and **we found 3 "valid" apps in 2022 out of 105k samples (=0.003%, about the same as the "last two months" sample)**. "Valid" meant apps that weren't either chat apps, or stub code (e.g. "drag some blocks out"). We found 11 chat apps out of the 105k, so chat apps were 0.01%, though some of the 11 are variations on the same "WUT by Adam" chat.
 
 So basically, looks like our rate of apps that use the onRecordEvent block is ~~0.002 - 0.003% (pretty in the noise given how few examples there were even with 100k samples each time)
- **snickell** [commented](https://github.com/code-dot-org/code-dot-org/pull/54643#issuecomment-1815763762) - We couldn't figure out how people were writing gamelab based chat apps. I found the "secret" gamelab APIs thanks to Hannah Nees giving me a great sample of "banned chat apps", including samples from Gamelab. These use "setKeyValue" and "getKeyValue" functions.... which aren't exposed as blocks (?!?).
 
 Relevant code is here: https://github.com/code-dot-org/code-dot-org/blob/e6905b0c899383ca2c7e43885f847cedf84478b2/apps/src/p5lab/gamelab/commands.js/#L25-L47
 
 The PR that added them is here: https://github.com/code-dot-org/code-dot-org/pull/11951
 
 @davidsbailey any chance you remember why we were adding these APIs without adding them as blocks?
- **davidsbailey** [commented](https://github.com/code-dot-org/code-dot-org/pull/54643#issuecomment-1815816682) - > @davidsbailey any chance you remember why we were adding these APIs without adding them as blocks?
 
 no, sadly I do not think I was involved in this decision.
- **snickell** [commented](https://github.com/code-dot-org/code-dot-org/pull/54643#issuecomment-1899550847) - - Next we should start implementing more methods from the firebase interface we extracted, and get corresponding tests passing.
 - We should also consider adding tests that actually create student source using the data blocks, and thereby test end to end, from JS api through the Rails controller. I think these tests, while a little more bulky, will actually catch much more and more helpful potential breakages in the future.

### Previous Reviews:

snickell:
UI tests passed on af4ff0c, still seeing a MySQL timeout error on unit tests, but 3fc9b28 and e8bd7b1 are two different approaches to perhaps fix the timeout error.

Results:
1. [Use factory :project_with_fake_storage](https://github.com/code-dot-org/code-dot-org/pull/56279/commits/3fc9b28ebfa97aafc09ef050b4f6b8aeadec18e5): didn't work, same mysql timeout error
2. [Try setting project_storage id starting at 1000](https://github.com/code-dot-org/code-dot-org/pull/56279/commits/e8bd7b1e9e83cde4201c0e804393baf7065dcefb): seems like a different error, but still errored out, see drone logs: https://drone.cdn-code.org/code-dot-org/code-dot-org/2297/1/5 . New errors look like:
```
==ERROR["test_add_column", "DatablockStorageControllerTest", 709.740416873]
--
2406 | test_add_column#DatablockStorageControllerTest (709.74s)
2407 | Minitest::UnexpectedError:         ActiveRecord::RecordNotUnique: Mysql2::Error: Duplicate entry '1003' for key 'PRIMARY'
2408 | config/initializers/mysql_check_index_used.rb:33:in `block (3 levels) in execute'
2409 | config/initializers/mysql_check_index_used.rb:32:in `block (2 levels) in execute'
2410 | config/initializers/mysql_check_index_used.rb:31:in `block in execute'
2411 | config/initializers/mysql_check_index_used.rb:30:in `execute'
2412 | test/factories/factories.rb:1031:in `block (3 levels) in <top (required)>'
2413 | test/controllers/datablock_storage_controller_test.rb:8:in `block in <class:DatablockStorageControllerTest>'
2414 | test/testing/setup_all_and_teardown_all.rb:36:in `run'
```

First error is for:
```
ERROR["setup_all", CodeReviewCommentsControllerTest, 570.7491446580002]
--
2332 | setup_all#CodeReviewCommentsControllerTest (570.75s)
2333 | Minitest::UnexpectedError:         ActiveRecord::RecordNotUnique: Mysql2::Error: Duplicate entry '1002' for key 'PRIMARY'
```

Final error is for: 
```
ERROR["test_updates_a_record", "DatablockStorageControllerTest", 711.447429462]
--
2846 | test_updates_a_record#DatablockStorageControllerTest (711.45s)
2847 | Minitest::UnexpectedError:         ActiveRecord::RecordNotUnique: Mysql2::Error: Duplicate entry '1043' for key 'PRIMARY'
```

I wonder what pre-existing IDs this is conflicting with? There are only 41 tests in `datablock_storage_controller_test.rb` total, so why do these error out as duplicates?

We maybe should remove the print statements from the JS log: 
image

Trying to re-run with factory bot project_storage ID starting at 100k instead of 1k, lets see if its that specific ID that's used or if its internally duplicating IDs without clearing them fully from the DB?

cnbrenci:
> Trying to re-run with factory bot project_storage ID starting at 100k instead of 1k, lets see if its that specific ID that's used or if its internally duplicating IDs without clearing them fully from the DB?

Looks like the latter
ActiveRecord::RecordNotUnique: Mysql2::Error: Duplicate entry '100044' for key 'PRIMARY'

snickell:
Re-running Cassi's last one-in-a-million attempt: last test failed because we got a weird merge conflict with staging (for some reason onDataPreview and onDataViewChange.

EDIT: million index also failed :-( trying a totally new approach, I don't think the issue is conflicts with existing fixture IDs, I think its a SQL locking issue related to Projects.rb using direct and separate from ActiveRecord SQL calls

snickell:
Progress on getting tests to pass in [f2e6991](https://github.com/code-dot-org/code-dot-org/pull/56279/commits/f2e6991bce69444e2e1ef6c61abbbdca8ec66f9e), I believe we're now on to a new type of test failure and the "mysql timeout" errors are behind us:

```
/home/circleci/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/mysql2-0.5.5/lib/mysql2/client.rb:151:in `_query': Mysql2::Error: Cannot truncate a table referenced in a foreign key constraint (`dashboard_test`.`project_use_datablock_storages`, CONSTRAINT `fk_rails_3795c35b7c` FOREIGN KEY (`project_id`) REFERENCES `dashboard_test`.`projects` (`id`)) (Sequel::DatabaseError)
--
2662 | from /home/circleci/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/mysql2-0.5.5/lib/mysql2/client.rb:151:in `block in query'
2663 | from /home/circleci/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/mysql2-0.5.5/lib/mysql2/client.rb:150:in `handle_interrupt'
2664 | from /home/circleci/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/mysql2-0.5.5/lib/mysql2/client.rb:150:in `query'
2665 | from /home/circleci/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/sequel-5.67.0/lib/sequel/adapters/mysql2.rb:142:in `block in _execute'
2666 | from /home/circleci/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/sequel-5.67.0/lib/sequel/database/logging.rb:38:in `log_connection_yield'
2667 | from /home/circleci/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/sequel-5.67.0/lib/sequel/adapters/mysql2.rb:137:in `_execute'
2668 | from /home/circleci/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/sequel-5.67.0/lib/sequel/adapters/utils/mysql_mysql2.rb:40:in `block in execute'
2669 | from /home/circleci/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/sequel-5.67.0/lib/sequel/connection_pool/threaded.rb:92:in `hold'
2670 | from /home/circleci/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/sequel-5.67.0/lib/sequel/database/connecting.rb:293:in `synchronize'
2671 | from /home/circleci/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/sequel-5.67.0/lib/sequel/adapters/utils/mysql_mysql2.rb:40:in `execute'
2672 | from /home/circleci/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/sequel-5.67.0/lib/sequel/adapters/mysql2.rb:67:in `execute_dui'
2673 | from /home/circleci/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/sequel-5.67.0/lib/sequel/database/query.rb:43:in `execute_ddl'
2674 | from /home/circleci/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/sequel-5.67.0/lib/sequel/dataset/actions.rb:1192:in `execute_ddl'
2675 | from /home/circleci/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/sequel-5.67.0/lib/sequel/dataset/actions.rb:929:in `truncate'
2676 | from /drone/src/shared/test/common_test_helper.rb:48:in `block in <top (required)>'
2677 | from /drone/src/shared/test/common_test_helper.rb:46:in `each'
2678 | from /drone/src/shared/test/common_test_helper.rb:46:in `<top (required)>'
2679 | from /drone/src/dashboard/legacy/test/middleware/middleware_test_helper.rb:3:in `require_relative'
2680 | from /drone/src/dashboard/legacy/test/middleware/middleware_test_helper.rb:3:in `<top (required)>'
2681 | from /drone/src/dashboard/legacy/test/middleware/files_api_test_base.rb:1:in `require_relative'
2682 | from /drone/src/dashboard/legacy/test/middleware/files_api_test_base.rb:1:in `<top (required)>'
2683 | from /drone/src/dashboard/legacy/test/middleware/files_api/test_public_thumbnails.rb:1:in `require_relative'
2684 | from /drone/src/dashboard/legacy/test/middleware/files_api/test_public_thumbnails.rb:1:in `<top (required)>'
2685 | from /home/circleci/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/rake-13.1.0/lib/rake/rake_test_loader.rb:21:in `require'
2686 | from /home/circleci/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/rake-13.1.0/lib/rake/rake_test_loader.rb:21:in `block in <main>'
2687 | from /home/circleci/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/rake-13.1.0/lib/rake/rake_test_loader.rb:6:in `select'
2688 | from /home/circleci/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/rake-13.1.0/lib/rake/rake_test_loader.rb:6:in `<main>'
```

apps/src/applab/applab.js
```diff
@@ -824,18 +836,25 @@ async function initDataTab(levelOptions) {
...
+            } else {
+              Applab.storage.copyStaticTable(
+                table,
+                () => console.log('success'),
+                outputError
+              );
+            }
+          } else {
+            Applab.storage.addSharedTable(table);
+          }
```
snickell, line 848: Because DatablockStorage tables are de-duplicated by default, there's no distinction between a "current" table and a static table. Static tables are also "current", the only difference is if there's a cron job that periodically updates the static table.

apps/src/applab/applab.js
```diff
@@ -873,7 +892,8 @@ function setupReduxSubscribers(store) {
       (isDataMode && view !== lastView) ||
       changedToDataMode(state, lastState)
     ) {
-      onDataViewChange(
+      loadDataForView(
```
snickell, line 895: onDataViewChange has been renamed to `loadDataForView` and refactored into `loadDataView.js`. This name better reflects the usage in Datablock Storage in particular, but also gives a better sense of what the firebase code does.

apps/src/applab/applab.js
```diff
@@ -1349,90 +1337,6 @@ function onInterfaceModeChange(mode) {
...
-      getStore().dispatch(updateTableRecords(tableName, snapshot.val()));
-    }
-  );
-}
-
-/**
- * Handle a view change within data mode.
- * @param {DataView} view
- */
-function onDataViewChange(view, oldTableName, newTableName) {
```
snickell, line 1368: onDataViewChange has been renamed to loadDataForView and refactored into loadDataView.js. This name better reflects the usage in Datablock Storage in particular, but also gives a better sense of what the firebase code does.

Most of the code seen below is pulled out into Firebase-specific methods in firebaseStorage.js

apps/src/applab/commands.js
snickell, line 1: Changes in `applab/commands.js` are to implement a `getColumn` optimization. Whereas the Firebase method must pull down ALL data from a table to extract data from a single column, in Datablock Storage we have a backend method that extracts the specific column's data being requested.

apps/src/applab/applab.js
snickell, line 1: Most of the changes to `applab.js` can be characterized as extracting Firebase-specific code and moving it to a method in firebaseStorage.js available as Applab.storage.methodName. Then we can implement a datablockStorage.js equivalent method, and Applab doesn't have to know which storage backend it is using.

apps/src/applab/commands.js
snickell, line 1: Changes to `applab/command.js` are to implement an optimized version of the `getColumn` block for DatablockStorage.getColumn. Whereas the Firebase implementation must fetch the whole table client side to filter out the column, the Datablock Storage version has a direct backend call which does the filtering server-side.

apps/src/p5lab/P5Lab.js
```diff
@@ -254,7 +254,8 @@ export default class P5Lab {
     config.usesAssets = true;
 
     this.studioApp_.labUserId = config.labUserId;
-    this.studioApp_.storage = initFirebaseStorage({
+    // TODO: unfirebase, convert to datablock storage: #56995
+    this.studioApp_.storage = initStorage(FIREBASE_STORAGE, {
```
snickell, line 258: For now we are only enabling DatablockStorage for Applab. Changing this line to be like applab (https://github.com/code-dot-org/code-dot-org/blob/384f3f51cf53b9b820a18fbd3f1a503051f0d3b3/apps/src/applab/applab.js/#L421-L443) should be all that's required to switch gamelab over too.

apps/src/storage/GetColumnParamPicker.jsx
```diff
@@ -24,11 +25,13 @@ export default class GetColumnParamPicker extends React.Component {
...
+      // Only firebase needs tableType, datablock storage checks on the backend
+      const tableType = isFirebaseStorage()
+        ? reduxState.data.tableListMap[this.props.table]
+        : undefined;
       Applab.storage
-        .getColumnsForTable(
-          this.props.table,
-          reduxState.data.tableListMap[this.props.table]
-        )
+        .getColumnsForTable(this.props.table, tableType)
```
snickell, line 34: In firebaseStorage, the frontend tracks tableType from the backend: is this a current table or a regular static table, and passes the tableType back in to all calls. This is not necessary in datablockStorage, which does the check on the backend. As a result, datablockStorage ignores tableType.

apps/src/storage/dataBrowser/DataLibraryPane.jsx
```diff
@@ -35,14 +36,24 @@ class DataLibraryPane extends React.Component {
...
+        storageBackend().copyStaticTable(
+          datasetInfo.name,
+          () => {},
+          this.onError
+        );
+      }
     } else {
-      FirebaseStorage.copyStaticTable(datasetInfo.name, () => {}, this.onError);
+      storageBackend().addSharedTable(datasetInfo.name);
+      refreshCurrentDataView();
```
snickell, line 56: In datablock storage the difference between current and static tables is collapsed, see https://github.com/code-dot-org/code-dot-org/pull/56279/files#r1524559478

apps/src/storage/dataBrowser/DataTableView.jsx
```diff
@@ -65,18 +67,23 @@ class DataTableView extends React.Component {
...
+    // TODO: post-firebase-cleanup, remove this conditional, leave only logic from the else clause #56994
+    if (isFirebaseStorage()) {
+      const isSharedTable =
+        this.props.tableListMap[this.props.tableName] === tableType.SHARED;
+      const tableName = encodeURIComponent(this.props.tableName);
+      const channelId = isSharedTable ? 'shared' : Applab.channelId;
+      location.href = `/v3/export-firebase-tables/${channelId}/${tableName}`;
+    } else {
+      location.href = storageBackend().exportCsvUrl(this.props.tableName);
+    }
```
snickell, line 79: Firebase CSV export redirects to the URL added by an old middleware method in tables_api.rb. In contrast, Datablock Storage redirects to a new method on DatablockStorageController.

apps/src/storage/dataBrowser/EditTableListRow.jsx
```diff
@@ -19,7 +20,19 @@ class EditTableListRow extends React.Component {
...
+        this.props.tableName,
+        this.props.tableType,
+        refreshCurrentDataView
+      );
+    } else {
+      storageBackend().deleteTable(
+        this.props.tableName,
+        refreshCurrentDataView
+      );
+    }
```
snickell, line 35: As before, `tableType` is detected automatically by the datablock storage backend, and doesn't need to be passed around on the frontend.

apps/src/storage/dataBrowser/loadDataForView.js
```diff
@@ -0,0 +1,98 @@
...
+let lastView;
+let lastTableName;
+let lastStorage;
+
+export function refreshCurrentDataView() {
+  // TODO: post-firebase-cleanup, remove if guard, but keep the contents: #56994
+  if (isDatablockStorage()) {
+    loadDataForView(lastStorage, lastView, null, lastTableName);
+  }
+}
```
snickell, line 23: You'll see references to `refreshCurrentDataView()` throughout the code after UX actions, this is because in Firebase the callbacks resulted in reloading data after a state change, whereas in datablock storage, the reload must be manually kicked off.

apps/src/storage/dataBrowser/loadDataForView.js
```diff
@@ -0,0 +1,98 @@
...
+
+/**
+ * Given a particular view in the data sets browser, invoke appropriate
+ * commands to the storageBackend() loading the data needed for that view.
+ * @param {StorageBackend} storage - firebaseStorage or datablockStorage
+ * @param {DataView} view
+ * @param {string} oldTableName - a previously subscribed table name, for firebase to unsubscribe
+ * @param {string} newTableName - name of the table being displayed, if any
+ */
+export function loadDataForView(storage, view, oldTableName, newTableName) {
```
snickell, line 33: This code is extracted from applab.js

apps/src/storage/datablockStorage.js
snickell, line 1: This file implements equivalents to all pre-existing firebaseStorage.js methods, and some new ones too. Unlike Firebase where this code is "the real core", the methods in datablockStorage.js are thin layers that call back to corresponding methods in datablock_storage_controller.rb.

apps/src/storage/firebaseStorage.js
snickell, line 1: Existing methods in `firebaseStorage.js` are unchanged. All changes are either adding documentation/notes that will be helpful during the final cleanup, or additions of new methods to encapsulate firebase-specific code that used to be in, e.g. applab.js.

apps/src/storage/firebaseStorage.js
```diff
@@ -1069,6 +1094,164 @@ FirebaseStorage.importCsv = function (
     .then(onSuccess, onError);
 };
 
+// Initialize redux's list of tables from firebase, and keep it up to date as
+// new tables are added and removed.
+// TODO: unfirebase, this is ONLY implemented in FirebaseStorage, not DatablockStorage
+FirebaseStorage.subscribeToListOfProjectTables = function (
```
snickell, line 1100: This method, along with most of the new methods, was extracted from `applab.js`

apps/src/storage/storage.js
snickell, line 1: The core of `storage.js` is `export function storageBackend()`, which is re-imported in all storage-using code in the rest of the codebase, and returns either firebaseStorage.js or datablockStorage.js. Because they have the same interface, code can then be unaware of which backend its currently using.

dashboard/app/controllers/datablock_storage_controller.rb
snickell, line 1: `datablock_storage_controller.rb` and `datablock_storage_table.rb` are the core of Datablock Storage. There's an equivalent fetch wrapper for most of these functions in datablockStorage.js.

dashboard/app/models/project_use_datablock_storage.rb
snickell, line 1: ProjectUseDatablockStorage introduces a small table that controls whether a given project will be backed by datablockstorage or firebasestorage. While we transition, most projects will be backed by Firebase, but as we gain more confidence, we'll increasingly create entries in this table.

dashboard/app/views/datablock_storage/index.html.haml
snickell, line 1: This is an HTML debug interface to datablock_storage_controller.rb. Each method there has a corresponding form here. Its useful for quickly debugging what's happening. This was mostly built for dev, but we think it might be useful for chasing down issues as we migrate, and we'll probably remove this once we're done migrating.

dashboard/legacy/middleware/helpers/projects.rb
```diff
@@ -448,6 +451,22 @@ def self.in_restricted_share_mode(channel_id, project_type)
...
+      #
+      # Once 100% of the old projects are migrated, we're ready to remove code
+      # marked with: TODO: post-firebase-cleanup
+      ProjectUseDatablockStorage.find_or_create_by(project_id: project_id) do |storage|
+        fraction = DCDO.get('fraction_of_new_projects_use_datablock_storage', 0.0)
+        storage.use_datablock_storage = rand < fraction
+      end
+    end
+  end
+
```
snickell, line 473: After merging this PR, whenever we create a new Applab project, we roll the dice and check against `DCDO.get('fraction_of_new_projects_use_datablock_storage')` to decide if the new project will be backed by datablock storage or firebase. This allows us to do a gradual rollout and see how load etc works out at production scale.

dashboard/legacy/middleware/tables_api.rb
```diff
@@ -22,6 +24,7 @@ class TablesApi < Sinatra::Base
   # Exports a csv file from a table where the first row is the column names
   # and additional rows are the column values.
   #
+  # TODO: post-firebase-cleanup, remove this method at the least, probably remove whole file: #56994
   get %r{/v3/export-firebase-tables/([^/]+)/([^/]+)$} do |channel_id, table_name|
```
snickell, line 28: For DatablockStorage, this is instead implemented as a regular route on DatablockStorageController. After fully migrating, we can probably delete table_api.rb.

snickell:
Fix for mysql truncate errors seemed to work, and now there's only one error:
```
ERROR["test_data_is_passed_down_to_edit_page", "ProgrammingMethodsControllerTest", 10.69352294500004]
--
2269 | test_data_is_passed_down_to_edit_page#ProgrammingMethodsControllerTest (10.69s)
2270 | Minitest::UnexpectedError:         ActionView::Template::Error: No such file or directory @ rb_sysopen - /drone/src/dashboard/tmp/cache/assets/sprockets/v3.0/5I/5IAZ-k5OahuzILG8OLWQeKgp50IxppX9_rcWtMVobKY.cache.11380.3640.227945
2271 | app/views/layouts/application.html.haml:18
2272 | app/controllers/application_controller.rb:186:in `block in with_locale'
2273 | app/controllers/application_controller.rb:185:in `with_locale'
2274 | test/controllers/programming_methods_controller_test.rb:42:in `block in <class:ProgrammingMethodsControllerTest>'
2275 | test/testing/setup_all_and_teardown_all.rb:36:in `run'
```

I have no idea what this one is about....

davidsbailey:
> snickell https://github.com/code-dot-org/code-dot-org/pull/54643#issuecomment-1789850667 - This is the one function used outside of apps/storage/firebaseMetadata.js that I wonder if we could easily handle through SQL, it feels like a firebase "subscription".... is it?

yes, I believe that's accurate, though I'm not sure if there is a very specific meaning of "subscription" that you're referring to as opposed to the general concept.

other initial comments/questions:

at a high level, how will the onRecordEvent block work? does MySQL have some support for listening for changes?

we know when CSP create projects are due. do we know when these projects will be created? a project created on the day you move 1% of new projects to Datablock Storage, could end up being a CSP Create project.

how do you ensure consistency when two users update the same table?

snickell:
Now we see lots of `NameError: uninitialized constant Projects::DatablockStorageController` errors:
```
test_viewing_abusive_assets                                    ERROR (0.01s)
--
8654 | Minitest::UnexpectedError:         NameError: uninitialized constant Projects::DatablockStorageController
8655 | /drone/src/dashboard/legacy/middleware/helpers/projects.rb:456:in `set_use_datablock_storage'
8656 | /drone/src/dashboard/legacy/middleware/helpers/projects.rb:54:in `create'
8657 | /drone/src/dashboard/legacy/middleware/channels_api.rb:106:in `block in <class:ChannelsApi>'
```

dashboard/test/controllers/datablock_storage_controller_test.rb
```diff
@@ -0,0 +1,713 @@
...
+  setup do
+    sign_in @student
+  end
+
+  def _url(action)
+    return "/datablock_storage/#{@channel_id}/#{action}"
+  end
+
+  def create_bob_record
+    post _url(:create_record), params: {
```
bethanyaconnor, line 22: `post :create_record, params: { ... `

snickell: I tried just removing `_url()` and I got tons of errors because the generated URLs looked like: `https://test-studio.code.org:443create_record`, whereas the routes.rb entry looks like: `resources :datablock_storage, path: '/datablock_storage/:channel_id/'`.

More detail here: https://github.com/code-dot-org/code-dot-org/pull/57330#issuecomment-2002355449

Any suggestions @bethanyaconnor ? Given the `:channel_id` in the route path, I suspect we'd have to modify every post/get/delete/etc call to include a channel_id somehow, at which point I'm probably using a _url() helper anyway... it just might look a little different internally?

Mostly I'm wondering whether this signals there's something deeply wrong with our routes.rb configuration.

apps/src/applab/applab.js
```diff
@@ -422,14 +416,32 @@ Applab.init = function (config) {
...
-    firebaseName: config.firebaseName,
-    firebaseAuthToken: config.firebaseAuthToken,
-    firebaseSharedAuthToken: config.firebaseSharedAuthToken,
-    firebaseChannelIdSuffix: config.firebaseChannelIdSuffix || '',
-    showRateLimitAlert: studioApp().showRateLimitAlert,
-  });
+
+  // TODO: post-firebase-cleanup, remove this conditional when we're removing firebase: #56994
+  if (!!config.useDatablockStorage) {
+    // TODO: post-firebase-cleanup, remove this console.log: #56994
```
bethanyaconnor: Why are the console logs needed at all?

apps/src/storage/datablockStorage.js
```diff
@@ -0,0 +1,403 @@
...
+}
+
+DatablockStorage.createRecord = function (
+  tableName,
+  record,
+  onSuccess,
+  onError
+) {
+  createRecord(tableName, record).then(
+    newRecord => onSuccess(newRecord),
```
bethanyaconnor: nit: can't this just be `onSuccess`?

dashboard/app/controllers/datablock_storage_controller.rb
```diff
@@ -0,0 +1,323 @@
...
+# - DatablockStorageRecord: stores the records in each table
+# - DatablockStorageKvp: stores key-value pairs
+#
+# Metadata for the code.org defined datasets are stored in:
+# - DatablockStorageLibraryManifest
+
+class DatablockStorageController < ApplicationController
+  before_action :validate_channel_id
+  before_action :authenticate_user!
+  skip_before_action :verify_authenticity_token
```
bethanyaconnor, line 29: why is this needed? You likely can fetch the authenticity token on the frontend

dashboard/app/controllers/datablock_storage_controller.rb
```diff
@@ -0,0 +1,323 @@
...
+
+  StudentFacingError = DatablockStorageTable::StudentFacingError
+
+  rescue_from StudentFacingError do |exception|
+    render json: {msg: exception.message, type: exception.type}, status: :bad_request
+  end
+
+  SUPPORTED_PROJECT_TYPES = ['applab', 'gamelab']
```
bethanyaconnor: I thought gamelab wasn't supported yet?

dashboard/app/controllers/datablock_storage_controller.rb
```diff
@@ -0,0 +1,323 @@
...
+  ##########################################################
+  #   Debug View                                           #
+  ##########################################################
+  def index
+    @key_value_pairs = DatablockStorageKvp.where(project_id: @project_id)
+    @records = DatablockStorageRecord.where(project_id: @project_id)
+    @tables = DatablockStorageTable.where(project_id: @project_id)
+    @library_manifest = DatablockStorageLibraryManifest.instance.library_manifest
+    @storage_backend = ProjectUseDatablockStorage.use_data_block_storage_for?(params[:channel_id]) ? "Datablock Storage" : "Firebase"
+    puts "####################################################"
```
bethanyaconnor: remove?

dashboard/app/controllers/datablock_storage_controller.rb
bethanyaconnor, line 1: A couple things on this file:
- I'll re-iterate that this isn't very "Rails-y", though as we discussed, I think the argument y'all made that it's supposed to mirror a non-Rails API is valid and I don't think it's a blocker
- I should've mentioned this on Friday, but we've had a convention in the past (that we're sort of falling away from) to add the URL pattern above each controller action (like [so](https://github.com/code-dot-org/code-dot-org/blob/3dba46efeb212fb9e80d54bef08dd7a05661cc67/dashboard/app/controllers/datasets_controller.rb#L20)). IMO it might be helpful as it can make it clear what params the controller actions are expecting

dashboard/app/models/datablock_storage_library_manifest.rb
bethanyaconnor, line 1: I know I asked this last week but remind me why this needs to be a table and can't be a constant of some kind?

dashboard/app/models/datablock_storage_table.rb
```diff
@@ -0,0 +1,263 @@
...
+
+  def self.get_shared_table_names
+    get_table_names(SHARED_TABLE_PROJECT_ID)
+  end
+
+  def self.find_shared_table(table_name)
+    DatablockStorageTable.find_by(project_id: SHARED_TABLE_PROJECT_ID, table_name: table_name)
+  end
+
+  def self.add_shared_table(project_id, table_name)
```
bethanyaconnor, line 81: to check my understanding, this copies a shared table into a project right?

dashboard/app/models/game.rb
```diff
@@ -266,7 +266,7 @@ def has_i18n?
     !([NETSIM].include? app)
   end
 
-  def use_firebase?
+  def use_datablock_storage?
     [APPLAB, GAMELAB].include? app
```
bethanyaconnor: same question about gamelab. I'm having trouble understanding how this change and [this comment](https://github.com/code-dot-org/code-dot-org/pull/56279/files#r1524578742) go together

dashboard/test/controllers/datablock_storage_controller_test.rb
bethanyaconnor: hm let me think on this one a little...

bethanyaconnor (commented):
Added a few comments. I would also really like to see the migrations separated out -- this is a big PR and I think there's a reasonable likelihood it'll need to be reverted. Separating out the migrations helps with that.

apps/src/applab/applab.js
cnbrenci: Might be useful to have in a console screenshot if we get zendesk tickets, but we can also look it up for a given project. We were thinking of removing them anyway since they're kinda noisy.

dashboard/app/models/datablock_storage_library_manifest.rb
cnbrenci: Because it's editable in levelbuilder.

dashboard/app/controllers/datablock_storage_controller.rb
cnbrenci: My understanding is that since we cache script level pages in CloudFront, the CSRF token we would get from the frontend might be invalid, so we can't rely on it. More [context from slack conversation](https://codedotorg.slack.com/archives/C046G4TRLEN/p1708484595942659?thread_ts=1708478288.327859&cid=C046G4TRLEN)

dashboard/app/models/datablock_storage_table.rb
cnbrenci: It effectively adds a copy-on-write reference to the shared table into the project.  For read operations (see ln 80 `read_records`), if `is_shared_table` is set, we read from the shared table instead (note `is_shared_table` contains the name of the shared table, not a bool - naming was hard for this one).  The first write operation on the table will copy the table into the project and remove the reference (clear `is_shared_table`) (see refs of `if_shared_table_copy_on_write`)

dashboard/app/controllers/datablock_storage_controller.rb
cnbrenci: hrm, yeah, I can see how this is confusing. It's supported, but it's not enabled yet. 😅

dashboard/app/models/game.rb
snickell: Good question, confusing name. This will all be deleted post-firebase-cleanup, the real bug here is that we don't have it flagged as such! This is only called in one place (in level_helper.rb), and since its not long-term surviving code, I'm going to move it there and just remove this code.

dashboard/app/controllers/datablock_storage_controller.rb
snickell: The missing piece for gamelab is just that we haven't turned it on in gamelab itself. Mainly because we'd prefer to start with as small a test as possible.

apps/src/applab/applab.js
snickell: We chatted, and decided it'd be better to give Hannah a quick way to check if a given URL is on datablock storage or firebase storage, so removing console.logs

dashboard/app/models/datablock_storage_table.rb
```diff
@@ -0,0 +1,263 @@
...
+      logger.warn "Table already exists: not populating project_id=#{project_id}, table_name=#{table_name}"
+    end
+  end
+
+  def read_records
+    is_shared_table ? shared_table.records : records
+  end
+
+  def create_records(record_jsons)
+    if_shared_table_copy_on_write
```
cnbrenci, line 157: From Jessie:) Think about how to convey that order is very important here - 
1. group together all the data mutation methods w/ a short readme

readme would be helpful!

dashboard/app/models/game.rb
snickell: We removed this in favor of the version in DatablockStorageController, which now references Game::APPLAB and Game::GAMELAB, sort of giving us the best of bothe worldes

dashboard/app/models/datablock_storage_record.rb
```diff
@@ -0,0 +1,28 @@
...
+# Indexes
+#
+#  index_datablock_storage_records_on_project_id                 (project_id)
+#  index_datablock_storage_records_on_project_id_and_table_name  (project_id,table_name)
+#
+class DatablockStorageRecord < ApplicationRecord
+  # Stores student-owned records for App Lab's data features, see datablock_storage_controller.rb
+
+  self.primary_keys = :project_id, :table_name, :record_id
+
```
cnbrenci, line 19:
```suggestion
  # Most of the implementation for managing DatablockStorageRecords is in datablock_storage_table.rb

```

cnbrenci, line 19:
```suggestion
  # Most of the implementation for managing DatablockStorageRecords is in datablock_storage_table.rb
```

dashboard/app/models/datablock_storage_record.rb
```diff
@@ -0,0 +1,28 @@
...
+#
+# Indexes
+#
+#  index_datablock_storage_records_on_project_id                 (project_id)
+#  index_datablock_storage_records_on_project_id_and_table_name  (project_id,table_name)
+#
+class DatablockStorageRecord < ApplicationRecord
+  # Stores student-owned records for App Lab's data features, see datablock_storage_controller.rb
+
+  self.primary_keys = :project_id, :table_name, :record_id
```
cnbrenci, line 18:
```suggestion
  # Most of the implementation for managing DatablockStorageRecords is in datablock_storage_table.rb
```

dashboard/app/controllers/datablock_storage_controller.rb
```diff
@@ -0,0 +1,347 @@
...
+    @library_manifest = DatablockStorageLibraryManifest.instance.library_manifest
+    @storage_backend = ProjectUseDatablockStorage.use_data_block_storage_for?(params[:channel_id]) ? "Datablock Storage" : "Firebase"
+  end
+
+  ##########################################################
+  #   Key-Value-Pair API                                   #
+  ##########################################################
+
+  def set_key_value
+    raise StudentFacingError, "The value is too large. The maximum allowable size is #{DatablockStorageKvp::MAX_VALUE_LENGTH} bytes" if params[:value].length > DatablockStorageKvp::MAX_VALUE_LENGTH
```
jmkulwik, line 65: Just a nit here: Do students know about bytes at this point? Do they know how to convert characters or numbers to bytes? (Should we just tell them to try a smaller value)

dashboard/app/controllers/datablock_storage_controller.rb
```diff
@@ -0,0 +1,347 @@
...
+# - datablock_storage_kvp.rb: stores key-value pairs
+#
+# Metadata for the code.org defined datasets are stored in:
+# - datablock_storage_library_manifest.rb
+#
+# Methods are broken into sections, for example the Key-Value-Pair API, Table API,
+# Table Column API, Table Record API, Library Manifest API & Project API
+#
+# More details can be found in the PR that initially created Datablock Storage:
+# https://github.com/code-dot-org/code-dot-org/pull/56279
```
jmkulwik, line 24: As I was reading through and trying to learn about the data model, the main question in my head was "How are we doing both a KV store and a more traditional table (Relational DB?) at the same time?"

I couldn't find much about the KV store in the documentation (comments, issues, or PR description). In retrospect, this is probably because it's actually the much simpler of the two. But when I was learning, I assumed it would be the more complex because we're using mySql as our underlying DB and in my mind that's more similar to the tables students are making than to the KV stores they're making.

It would be really helpful to have a bit more description of the KV store that helps me know I shouldn't overthink it and it's pretty straightforward. A few main possibilities that come to mind are in the PR description, the header of this file, or in the `Key-Value-Pair API` section below, as those are where I looked for details before tracing through the code.

apps/src/applab/applab.js
```diff
@@ -1349,90 +1328,6 @@ function onInterfaceModeChange(mode) {
...
-
-      storageRef.on('value', snapshot => {
-        getStore().dispatch(updateTableRecords(newTableName, snapshot.val()));
-      });
-      return;
-    }
-    default:
-      return;
-  }
-}
```
jmkulwik, line 1434: OMG It's so nice to have all this removed

apps/src/applab/applab.js
```diff
@@ -422,14 +416,23 @@ Applab.init = function (config) {
...
+      channelId: config.channel,
+    });
+  } else {
+    Applab.storage = initStorage(FIREBASE_STORAGE, {
+      channelId: config.channel,
+      firebaseName: config.firebaseName,
+      firebaseAuthToken: config.firebaseAuthToken,
+      firebaseSharedAuthToken: config.firebaseSharedAuthToken,
+      firebaseChannelIdSuffix: config.firebaseChannelIdSuffix || '',
+      showRateLimitAlert: studioApp().showRateLimitAlert,
```
jmkulwik, line 432: Speaking of rate limiting, are we doing any sort of rate limiting in the new datablock storage?

dashboard/app/controllers/datablock_storage_controller.rb
```diff
@@ -0,0 +1,347 @@
...
+  end
+
+  private def table_or_create
+    where_table.first_or_create
+  end
+
+  private def validate_channel_id
+    project = Project.find_by_channel_id(params[:channel_id])
+    unless SUPPORTED_PROJECT_TYPES.include? project.project_type
+      raise "DatablockStorage is only available for applab and gamelab projects"
```
jmkulwik, line 343: For the future: I think these messages will eventually need to be translated

dashboard/app/models/datablock_storage_record.rb
```diff
@@ -0,0 +1,29 @@
...
+
+  # TODO: #57001, implement enforcement of MAX_RECORD_LENGTH, we already have
+  # a test for this, but we're skipping it until this is implemented. This
+  # should ensure the string form of .json is less than 4096 bytes.
+  MAX_RECORD_LENGTH = 4096
+
+  # Enabling this was adding an extra `SELECT `datablock_storage_tables`.*` query to update_record
+  # and we aren't using it, soooo...
+  # belongs_to :table, class_name: 'DatablockStorageTable', foreign_key: [:project_id, :table_name]
```
jmkulwik: Nit: Not quite sure what this comment is referring to :)

jmkulwik (approved):
Looks great! Two main thoughts: not blocking for this PR.

* Are we doing any sort of rate limiting? Should we be?
* I naively expected the KV store implementation to be the more complicated of the two, so I kept looking for hidden complexity. It'd be helpful having some comments explaining how it works it so the reader knows that it's actually pretty simple.

dashboard/app/models/datablock_storage_library_manifest.rb
bethanyaconnor: right. How are you syncing the data from levelbuilder to production?

dashboard/app/controllers/datablock_storage_controller.rb
bethanyaconnor: I guess maybe my question is why gamelab is added here but not turned on. But I also think that's a nit and not a blocker :)

dashboard/app/controllers/datablock_storage_controller.rb
bethanyaconnor: We have created ways of fetching an authenticity token on cached pages, see https://github.com/code-dot-org/code-dot-org/pull/51439, which might be worth a look for so many new routes.

dashboard/app/controllers/datablock_storage_controller.rb
cnbrenci: I think the best place to look to answer that would be in the model file for the key-value pairs model (datablock_storage_kvp.rb). This file describes the schema of the table where we're storing the data & implementation of each of the methods we're calling here on the model, & is guaranteed to not be out of date. It sounds like you found your way to the issue describing the models, but it was out of date and caused some confusion with channel_id vs project_id; But you eventually did find your way to the model file and find the answers you were looking for. Do you have any suggestions on what we could say here to lead you in the direction of looking at the model files first?

apps/src/applab/applab.js
cnbrenci: Not yet, but that's next. https://github.com/code-dot-org/code-dot-org/issues/55481

dashboard/app/controllers/datablock_storage_controller.rb
snickell: Our goal was to be a "bug compatible" replacement for Firebase, and we match Firebase exactly here, both in byte size and (I believe?) in not having error messages translated. We should do that as a follow-on PR after we migrate (100% compatible the whole time) from Firebsae.

https://github.com/code-dot-org/code-dot-org/blob/106c169006320e410e0c77e23fe6bca4f9e080b9/apps/src/storage/firebaseStorage.js/#L188-L190

dashboard/app/controllers/datablock_storage_controller.rb
cnbrenci: resolving since Seth addressed this in a different comment

dashboard/app/controllers/datablock_storage_controller.rb
cnbrenci: Yeah, it's going to get turned on soon

dashboard/app/models/datablock_storage_library_manifest.rb
cnbrenci: We are coordinating with curriculum (I believe @kakiha11 is the only person who edits this in levelbuilder) to tell us if they edit this, and we'll manually sync any changes to the table until the time that we're fully migrated to Datablock Storage.   Once we're fully migrated we'll switch levelbuilder over to write to this table.

dashboard/app/controllers/datablock_storage_controller.rb
cnbrenci: Given that it's the same URL pattern for all the actions, we added this line in the header of the file instead:

```
# Methods in this controller are available as (see :datablock_storage in routes.rb):
# /datablock_storage/:channel_id/:method_name
```

dashboard/app/controllers/datablock_storage_controller.rb
snickell: Yay! I looked so hard for an equivalent of this before evaluating whether we strictly speaking needed it. Given the number of routes, I agree there's a lot of space for bugs to hide. We've added a followup issue to address this: https://github.com/code-dot-org/code-dot-org/issues/57430

dashboard/app/models/datablock_storage_library_manifest.rb
snickell: This is very infrequently edited (and Ken said mostly by him), and we didn't want there to be "multiple sources of truth" while we transition, and fortunately its a single single blob update. We preferred doing the editing by hand vs writing throwaway code and potentially having bugs in it.

snickell:
> Are we doing any sort of rate limiting? Should we be?

We have a followup issue for rate limiting, its probably the highest priority followup for us after our experiment, I was curious to see natural traffic patterns before we decided which way we wanted to rate limit.

cnbrenci (approved):
🔥🧊🍾🚀

dashboard/app/helpers/levels_helper.rb
```diff
@@ -578,17 +578,26 @@ def question_options
     app_options
   end
 
-  def firebase_options
```
hannahbergam, line 581: Is this the line that changed to cause the name error?

dashboard/app/helpers/levels_helper.rb
```diff
@@ -677,7 +686,7 @@ def blockly_options
     app_options[:legacyShareStyle] = true if @legacy_share_style
     app_options[:isMobile] = true if browser.mobile?
     app_options[:labUserId] = lab_user_id if @game == Game.applab || @game == Game.gamelab
-    app_options.merge!(firebase_options)
```
hannahbergam, line 680: or more specifically, here?

PR #56283, opened 2024-02-05
[Design2-54] Checkbox Dropdown Component

levadadenys:
### [Design2-54] Checkbox Dropdown Component

* Created `CheckboxDropdown` component
* created stories for  `CheckboxDropdown` component
* created tests for  `CheckboxDropdown` component
* created documentation for  `CheckboxDropdown` component

`CheckboxDropdown` (apps/src/componentLibrary/checkboxDropdown/CheckboxDropdown.tsx) component is based on Turner's `CheckboxDropdown` (apps/src/templates/CheckboxDropdown.jsx). (It's already refactored, but still used it as a base).

https://github.com/code-dot-org/code-dot-org/assets/22244040/4ca3cc71-6700-4d6f-961e-32b0b703da1c

## Links
- jira ticket: [DESGN2-54](https://codedotorg.atlassian.net/browse/DESIGN2-54)

## PR Checklist:

- [ ] Code is well-commented


levadadenys:
This PR is a continuation of #55762, which was mis-closed by the Git LFS migration (#55759).

### Previous Comments:


### Previous Reviews:

apps/src/componentLibrary/checkboxDropdown/CheckboxDropdown.story.tsx
```diff
@@ -0,0 +1,280 @@
...
+// eslint-disable-next-line
+const SingleTemplate: Story<CheckboxDropdownProps> = args => {
+  const [selectedValues, setValues] = useState(
+    (args.checkedOptions = [] as string[])
+  );
+  const onChange = useCallback(
+    (e: React.ChangeEvent<HTMLInputElement>) => {
+      // e.stopPropagation();
+      // e.preventDefault();
+      console.log(e);
```
bencodeorg: Intended logging?

apps/src/componentLibrary/checkboxDropdown/CheckboxDropdown.story.tsx
```diff
@@ -0,0 +1,280 @@
+import React, {useState, useCallback} from 'react';
+import CheckboxDropdown, {CheckboxDropdownProps} from './index';
+import {Meta, Story} from '@storybook/react';
+
+export default {
+  title: 'DesignSystem/Checkbox Dropdown Component',
```
bencodeorg: Nit: I dropped the "Component" suffix from all of the design system components when we reorganized, thinking being that these are all components. Thoughts?

apps/src/componentLibrary/checkboxDropdown/CheckboxDropdown.tsx
```diff
@@ -0,0 +1,193 @@
...
+ * Used to render checkbox (multiple choice) dropdowns.
+ */
+const CheckboxDropdown: React.FunctionComponent<CheckboxDropdownProps> = ({
+  name,
+  labelText,
+  allOptions,
+  checkedOptions = [],
+  onChange,
+  onSelectAll,
+  onClearAll,
```
bencodeorg, line 68: Why do we pass this in as a prop (vs defining it in this component)? Isn't it kind of expected that a clear all button will...clear all? Same question re: onSelectAll

apps/src/componentLibrary/checkboxDropdown/CheckboxDropdown.story.tsx
```diff
@@ -0,0 +1,280 @@
...
+  component: CheckboxDropdown,
+} as Meta;
+
+//
+// TEMPLATE
+//
+// This is needed to fix children type error (passing string instead of React.ReactNode type)
```
bencodeorg: Can you elaborate? What is the issue here?

apps/src/componentLibrary/common/contexts/DropdownContext.tsx
```diff
@@ -0,0 +1,24 @@
+import React, {createContext, useContext, useState, ReactNode} from 'react';
+
+const DropdownContext = createContext({
+  activeDropdownName: '',
+  setActiveDropdownName: (name: string) => {
+    console.log(name);
```
bencodeorg: Intentional logging?

apps/src/componentLibrary/checkboxDropdown/CheckboxDropdown.story.tsx
```diff
@@ -0,0 +1,280 @@
...
+  components: [
+    {
+      name: 'default-dropdown-xs',
+      allOptions: [
+        {value: 'option-1', label: 'Option 1'},
+        {value: 'option-2', label: 'Option 2'},
+      ],
+      checkedOptions: ['option-1'],
+      labelText: 'XS Dropdown',
+      onChange: args => console.log(args),
```
bencodeorg: Do we need all of these console logs?

apps/src/componentLibrary/checkboxDropdown/CheckboxDropdown.tsx
```diff
@@ -0,0 +1,193 @@
...
+      ) {
+        setActiveDropdownName('');
+      }
+    },
+    [dropdownRef, setActiveDropdownName, activeDropdownName]
+  );
+
+  useEffect(() => {
+    document.addEventListener('mousedown', handleClickOutside);
+    document.addEventListener('keydown', handleClickOutside);
```
bencodeorg, line 93: Does this close the dropdown for any key press?

apps/src/componentLibrary/common/contexts/DropdownContext.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+export const useDropdownContext = () => useContext(DropdownContext);
+
+export const DropdownProvider: React.FC<{children: ReactNode}> = ({
+  children,
+}) => {
+  const [activeDropdownName, setActiveDropdownName] = useState('');
+
+  return (
+    <DropdownContext.Provider
```
bencodeorg, line 21: Not super familiar with context in React...help me understand what this does?

apps/test/unit/componentLibrary/CheckboxDropdownTest.jsx
```diff
@@ -0,0 +1,211 @@
...
+
+import CheckboxDropdown from '@cdo/apps/componentLibrary/checkboxDropdown';
+
+const allOptions = [
+  {value: 'option-1', label: 'option1'},
+  {value: 'option-2', label: 'option2'},
+  {value: 'option-3', label: 'option3'},
+];
+let selectedValues = [];
+const onChceckboxDropdownChange = (value, isChecked) => {
```
bencodeorg: typo

bencodeorg (commented):
A lot to parse here, but these look beautiful 😍 . I took a first pass but will keep looking thru this tomorrow!

apps/src/componentLibrary/checkboxDropdown/CheckboxDropdown.story.tsx
levadadenys: Those console log were added for develop,ent consistency, but I agree it's better to remove them from public version. My eyes are bluring already with this component :D

apps/src/componentLibrary/checkboxDropdown/CheckboxDropdown.story.tsx
levadadenys: Oh, thanks for the catch!
That's some template code following me from some old components. Don't remember the initial issue, but deleted this line now and looks like everything is working ok. Deleted.

apps/src/componentLibrary/checkboxDropdown/CheckboxDropdown.tsx
levadadenys: yes, for any key press. Works as expected since it checks the event target attribute

apps/src/componentLibrary/checkboxDropdown/CheckboxDropdown.tsx
levadadenys: Those are added to make component fully controllable from the parent component

apps/src/componentLibrary/common/contexts/DropdownContext.tsx
levadadenys: That's just a placeholder to mach typescript types, not used anywhere. Replace with     `return name as unknown as void;` , also added explaining jsdoc comment above the function

apps/src/componentLibrary/common/contexts/DropdownContext.tsx
levadadenys: This creates 'redux' like storage/context of data shared across components that are wrapped in DropdownProvider like this: 
```
<DropdownProvider>
    <CheckboxDropdown {...props} />
  </DropdownProvider>
```

and then accessed inside DropdownProvider child compoent like this:
`  const {activeDropdownName, setActiveDropdownName} = useDropdownContext();`

This serves a purpose of creating one state for components using `DropdownProvider` at once. Also allows us to share this context with IconsDropdown (to be implemented). (not applicable to `SimpleDropdown` cause native select elements)

And the initial reason behind all of this is that we want only one dropdown to be open at once.

apps/src/componentLibrary/checkboxDropdown/CheckboxDropdown.tsx
```diff
@@ -0,0 +1,193 @@
...
+import {
+  DropdownProvider,
+  useDropdownContext,
+} from '@cdo/apps/componentLibrary/common/contexts/DropdownContext';
+
+export interface CheckboxDropdownProps {
+  /** CheckboxDropdown name */
+  name: string;
+  /** CheckboxDropdown color */
+  color?: 'white' | 'black';
```
sanchitmalhotra126, line 29: nit: move this to a `type DropdownColor = 'white' | 'black'` since it's shared with other dropdowns?

apps/src/componentLibrary/common/contexts/DropdownContext.tsx
```diff
@@ -0,0 +1,27 @@
+import React, {createContext, useContext, useState, ReactNode} from 'react';
+
+const DropdownContext = createContext({
+  activeDropdownName: '',
+  /**
+   * This is a dummy function to satisfy the type checker
+   * */
+  setActiveDropdownName: (name: string) => {
+    return name as unknown as void;
```
sanchitmalhotra126, line 9: Why is this cast needed? Curious where the type checker is complaining here

apps/src/componentLibrary/common/contexts/DropdownContext.tsx
```diff
@@ -0,0 +1,27 @@
...
+  },
+});
+
+export const useDropdownContext = () => useContext(DropdownContext);
+
+export const DropdownProvider: React.FC<{children: ReactNode}> = ({
```
sanchitmalhotra126, line 15: super minor: is there another name we could use for this component (maybe `DropdownWrapper`?) just so it's not confused with the `DropdownContext.Provider`?

sanchitmalhotra126 (approved):
Generally LGTM! Giving this the go ahead since these needs to be merged ASAP; but will leave a few more comments for follow-up later.

apps/src/componentLibrary/checkboxDropdown/CheckboxDropdown.tsx
```diff
@@ -0,0 +1,193 @@
...
+  /** CheckboxDropdown size */
+  size: ComponentSizeXSToL;
+  /** CheckboxDropdown disabled state */
+  disabled?: boolean;
+  /** CheckboxDropdown label */
+  labelText: string;
+  /** CheckboxDropdown options */
+  allOptions: {value: string; label: string; isOptionDisabled?: boolean}[];
+  /** CheckboxDropdown checked options */
+  checkedOptions: string[];
```
sanchitmalhotra126, line 39: could we just have a `checked` property part of `allOptions` instead of a separate `checkedOptions` list?

apps/src/componentLibrary/checkboxDropdown/CheckboxDropdown.tsx
```diff
@@ -0,0 +1,193 @@
...
+  /** CheckboxDropdown color */
+  color?: 'white' | 'black';
+  /** CheckboxDropdown size */
+  size: ComponentSizeXSToL;
+  /** CheckboxDropdown disabled state */
+  disabled?: boolean;
+  /** CheckboxDropdown label */
+  labelText: string;
+  /** CheckboxDropdown options */
+  allOptions: {value: string; label: string; isOptionDisabled?: boolean}[];
```
sanchitmalhotra126, line 37: nit: can `isOptionDisabled` just be `disabled` so it's a little less wordy?

apps/src/componentLibrary/checkboxDropdown/CheckboxDropdown.tsx
```diff
@@ -0,0 +1,193 @@
...
+  /** CheckboxDropdown checked options */
+  checkedOptions: string[];
+  /** CheckboxDropdown onChange handler */
+  onChange: (args: React.ChangeEvent<HTMLInputElement>) => void;
```
sanchitmalhotra126, line 41: nit: `event` instead of `args`?

apps/src/componentLibrary/checkboxDropdown/CheckboxDropdown.story.tsx
bencodeorg: haha I can see why, a ton of work here!

apps/src/componentLibrary/checkboxDropdown/CheckboxDropdown.tsx
sanchitmalhotra126: Yeah Ben at first I had the same thought, but I do agree that deferring control to the parent would probably be better overall. I'm thinking of how this would probably be used with redux pretty often - in which case the clear all function would just look something like `const onClearAll = () => dispatch(setSelectedOptions([]));`

apps/src/componentLibrary/checkboxDropdown/checkboxDropdown.module.scss
```diff
@@ -0,0 +1,336 @@
+@import "color";
+@import "@cdo/apps/componentLibrary/common/styles/mixins";
+
+// Dropdown common styles
```
sanchitmalhotra126, line 4: Probably a follow-up once you've implemented all the dropdown variants, but how much of these styles can we factor out and share between the various dropdown components?

apps/src/componentLibrary/common/contexts/DropdownContext.tsx
bencodeorg: Got it, thanks! Is there a reason we're especially conscious not to have multiple of these open at once, especially considering `SimpleDropdown` won't have the same behavior?

sanchitmalhotra126 (commented):
Just leaving a few more comments for follow-up after merging

apps/src/componentLibrary/checkboxDropdown/CheckboxDropdown.tsx
bencodeorg: Ok, fair enough. I guess my concern is that I think my preference for these super reusable components is that they emphasize ease of use rather than customization, whereas I think this approach emphasizees the latter.

apps/src/componentLibrary/checkboxDropdown/checkboxDropdown.module.scss
levadadenys: Good question, plan is to factor out as much styles as possible in the process of working on `IconsDropdown`. For `SimpleDropdown` I feel like it's not applicable since though visually the same, semantically and by it's structutre they differ a lot.

apps/src/componentLibrary/checkboxDropdown/CheckboxDropdown.tsx
levadadenys: Good question, I made it more wordy to avoid confusion since we already have `disabled` prop for Dropdown itself

apps/src/componentLibrary/common/contexts/DropdownContext.tsx
levadadenys: Typescript wants this function to be of type `(name:string) => void`, but at the same moment typescript/eslint doesn't allow us to create empty functions like `(name:string) => {}`

apps/src/componentLibrary/common/contexts/DropdownContext.tsx
levadadenys: Not sure since all `DropdownProvider` do is adding `DropdownContext.Provider` around children components. `DropdownWrapper` might cause some confusions I guess since we don't have word Provider or Context in it.
Though, I think `DropdownProviderWrapper` or `DropdownContextWrapper` might be a good name giving the full context to the developer, leaning towards `DropdownProviderWrapper`. What do you think?

apps/src/componentLibrary/checkboxDropdown/CheckboxDropdown.tsx
levadadenys: Yeah, we can do that, not sure if that'll be easier for use with redux or parent component state though. For this specific case I decided to stick with what we have now in `apps/src/templates/CheckboxDropdown.jsx` since it looked like the simplest way for the end user/developer.

Have no strong opinion about this though, can update in a follow up PR.

apps/src/componentLibrary/checkboxDropdown/CheckboxDropdown.tsx
fisher-alice: Hi Denys, I pulled your branch and 'enter' and 'space bar' seemed like the only keys that to close an open dropdown.

fisher-alice: Not sure if this is within the scope of this PR, but currently when a dropdown option is highlighted, the 'Enter' key press does not check the checkbox.

apps/src/componentLibrary/checkboxDropdown/CheckboxDropdown.tsx
```diff
@@ -0,0 +1,193 @@
...
+  }, [name, activeDropdownName, setActiveDropdownName]);
+
+  const isOpen = useMemo(
+    () => activeDropdownName === name,
+    [activeDropdownName, name]
+  );
+
+  // Collapse dropdown if 'Escape' is pressed
+  const onKeyDown: (e: KeyboardEvent) => void = e => {
+    if (e.keyCode === 27) {
```
fisher-alice, line 115: nit: could add this key code to [constant `KeyCodes`](https://gitkraken.dev/link/dnNjb2RlOi8vZWFtb2Rpby5naXRsZW5zL2xpbmsvci8yNTcwZTNiNjNhMWE2ZTNiODgzZDQ5ODlkZTE3ZjFkOGNhMzkwYjIyP3VybD1naXQlNDBnaXRodWIuY29tJTNBY29kZS1kb3Qtb3JnJTJGY29kZS1kb3Qtb3JnLmdpdA%3D%3D?origin=gitlens)

apps/src/componentLibrary/common/contexts/DropdownContext.tsx
sanchitmalhotra126: Ah interesting - my preference would be to just use `eslint-disable-next-line @typescript-eslint/no-empty-function` since this is a dummy function anyway (though weirdly the TS compiler isn't complaining for me when I change it to an empty function? might just be something in my environment though).

apps/src/componentLibrary/common/contexts/DropdownContext.tsx
sanchitmalhotra126: Yeah, no strong opinion either way, especially since this should only be used internally by design system components anyway

apps/src/componentLibrary/checkboxDropdown/CheckboxDropdown.tsx
sanchitmalhotra126: Hmm yeah hard to say without some real world examples. It really depends on what the data looks like that consumers are passing in. But yeah, sticking with what the existing `CheckboxDropdown` uses makes sense.

apps/src/componentLibrary/checkboxDropdown/CheckboxDropdown.tsx
levadadenys: Hey @fisher-alice , for open/close dropdown - that's correct behavior. The part you comment refers only the logic behind closeing the dropdown when clicked/keydowned anywhere outside of the dropdown content (can be any button which is also ok)

As for checkbox - that's also an intended behaviour since that's default browser approach for input fields (which checkboxes are)

apps/src/componentLibrary/common/contexts/DropdownContext.tsx
levadadenys: Yep, thanks for the question!
`SimpleDropdown` has same functionality, but it's implemented natively by `<select>` html element.

Answering re `CheckboxDropdown` and `IconsDropdown` - since it's custom component it can behave however we want of course, but design team asked me to make these two in a way that only one of these can be opened at a time. (will be same logic across all dropdown usages for UX reasons)

apps/src/componentLibrary/checkboxDropdown/CheckboxDropdown.tsx
levadadenys: Thanks for the hint!
keyCode is now actually deprecated, so I've rewritten this to use e.key instead which is more clear (see followup pr)

PR #56119, opened 2024-01-30
Pass extra data down to My PL Page

bethanyaconnor:
Passes extra data down to the My Professional Learning page as described in [the tech spec](https://docs.google.com/document/d/1BpAP0e_BWeKV7WwafPrzW77a_DlwCbDt9xP6Dys0XTU/edit#heading=h.yz85y9qg6eg4). This data is not used yet but will be used to determine which sections we will see on the page, as well as some of the data we will actually show in those sections.

PR #56287, opened 2024-02-05
Pass extra data down to My PL Page

bethanyaconnor:
Passes extra data down to the My Professional Learning page as described in [the tech spec](https://docs.google.com/document/d/1BpAP0e_BWeKV7WwafPrzW77a_DlwCbDt9xP6Dys0XTU/edit#heading=h.yz85y9qg6eg4). This data is not used yet but will be used to determine which sections we will see on the page, as well as some of the data we will actually show in those sections.


bethanyaconnor:
This PR is a continuation of #56119, which was mis-closed by the Git LFS migration (#55759).

### Previous Comments:


### Previous Reviews:
- **bethanyaconnor commented** - 
- **TurnerRiley approved** - This looks great! Thanks for adding the test coverage too!

bethanyaconnor:
@TurnerRiley previously approved this PR in its [previous incarnation](https://github.com/code-dot-org/code-dot-org/pull/56119), so mostly adding acquistions-products as an fyi.

@kelbyhawn would still appreciate a quick review from you!

TurnerRiley (approved):

dashboard/app/controllers/pd/professional_learning_landing_controller.rb
```diff
@@ -9,20 +9,35 @@ def index
...
+    last_enrollment_with_pending_survey = enrollments_with_pending_surveys.max_by {|e| e.workshop.ended_at}
 
     summarized_plc_enrollments = Plc::UserCourseEnrollment.where(user: current_user).map(&:summarize).sort_by do |enrollment|
       PLC_COURSE_ORDERING.index(enrollment[:courseName]) || PLC_COURSE_ORDERING.size
     end
 
+    enrolled_workshop_data = Pd::Enrollment.for_user(current_user).map do |enrollment|
+      workshop = enrollment.workshop
+      {
+        course_name: workshop.course_name,
```
kelbyhawn: @bethanyaconnor are these the keys I'll use when calling data?

kelbyhawn (approved):
A lot of this is new to me, but I read through it and have one non-blocking question 😄

dashboard/app/controllers/pd/professional_learning_landing_controller.rb
bethanyaconnor: yep!

bethanyaconnor:
fyi, I'm going to need to make one update to this PR -- will retag you both when it's ready!

bethanyaconnor:
@kelbyhawn @TurnerRiley would love a review of [1454033](https://github.com/code-dot-org/code-dot-org/pull/56287/commits/14540336b85956ee7cb4c52ecdfe6e7625d0d994) when you get a chance! It updates this PR to send down even more data to the page to support the new tabs. Some sections will be duplicated across tabs (ie "Upcoming Workshop"), but should be relevant to the tab they're in (ie the facilitator tab should show upcoming workshops the user is facilitating)

kelbyhawn (commented):
Do we need any Instructor data as seen [here](https://www.figma.com/file/sZ6d1uhJWFksJ6j3Fnmbnc/PL-Flow-Improvements?type=design&node-id=2041%3A8929&mode=design&t=nVwDtOP6fSW9s90K-1), or is that coming later?

bethanyaconnor:
good question! On the teacher homepage, sections are requested async, so I had been planning on reusing that pattern, though we could re-evaluate as we get there. 

This doesn't perfectly answer your question, but Eric's [comment](https://docs.google.com/document/d/1BpAP0e_BWeKV7WwafPrzW77a_DlwCbDt9xP6Dys0XTU/edit?disco=AAABEEMlHYM) is good context. I don't know the extent to which these concerns affect the PL experience, though.

kelbyhawn:
> good question! On the teacher homepage, sections are requested async, so I had been planning on reusing that pattern, though we could re-evaluate as we get there…

Got it, thanks for sharing Eric's comment!

kelbyhawn (approved):

TurnerRiley (approved):

PR #56294, opened 2024-02-05
[Google Blockly] Supporting starting over and outdated block XML for behaviors

mikeharv:
While working on getting UI tests to pass for Sprite Lab on Google Blockly, @molly-moen noticed that `/s/allthethings/lessons/36/levels/1?blocklyVersion=google` was failing to load due to a missing behavior id. This is a a case of an old block that needs additional support.
Separate from this issue, but related to this scenario, the `Start Over` button in Version History was crashing because it turns out we weren't clear the new workspaces that are created for the modal function editor.

This PR addresses both issues which allows the UI test scenario to pass on Google Blockly.

Run-down of changes:
- **cdoXml:** Updated `addMutationToBehaviorDefBlocks` to also add mutators/ids to behavior get blocks. This function needs to be called on all blocks now, not just top blocks, so the call to (renamed) `addMutationToBehaviorBlocks` has been moved.
- **cdoXml:** Fixed a bug in `addMissingBehaviorId` where we might erroneously read a behavior name from a child block (ie. behavior getter) if was missing from the parent block (ie. behavior definition).
- **behaviorDefMutator:** updated `domToMutation` (XML deserializer) to find behavior ids on the title/field element if it's missing for any reason on the mutator element.
- **wrappers / StudioApp** - Created a new `clearAllStudentWorkspaces` to ensure we clear the hidden and function editor workspaces when we clear the primary workspace.
- **cdoXml:** When an issue was hitting after pressing Start Over, we would typically create a new `unknown` block. This process calls `textToDom` using some preset XML.  This was deprecated from the `Xml` namespace last year and removed for v10. It is now located at `utils.xml`. See more: https://github.com/google/blockly/pull/6818, https://github.com/google/blockly/pull/7077

## Links

JIRA - https://codedotorg.atlassian.net/browse/CT-317

## Testing story

This branch was created to get existing tests passing on Google Blockly which will be confirmed by @molly-moen. I manually tested that the `allthethings` level works as expected, including after pressing Start Over. No regressions are found in the Sprite Lab project level. It looks like all block mutators now are created with the expected behavior id.

## PR Checklist:

- [ ] Code is well-commented


molly-moen (approved):
Nice!

PR #56299, opened 2024-02-05
[Google Blockly] strip random block ids when saving xml

mikeharv:
Recently, there was concern about the potential for storing identical solutions in the `level_sources` table in DB-backed levels for Google Blockly labs.

@breville [said](https://codedotorg.slack.com/archives/C05DK21DAHK/p1702490028201299?thread_ts=1702489794.582939&cid=C05DK21DAHK):
>The `level_sources` table is designed to only save one copy of each unique solution.  The saves us from writing tons of duplicate entries.  However, if each solution has some random IDs in it, we suddenly are saving every copy of what's essentially the same source.

He added:
>I think x/y is probably not such a big issue for levels where students are unlikely (or unable) to drag the parent block to a new position.  However, if IDs are truly unique for every use, then they seem like a more concerning issue. 

Currently, this issue impacts DB-backed levels using Flappy, Bounce, and Dance. As it happens, Sprite Lab/Poetry are never DB-backed, always using S3 for project storage. Minecraft also uses the database for many levels, so it makes sense to address this before migrating that lab.

Stripping block ids at the time of saving is a reasonable step forward, however there are times when blocks need to have explicitly set ids. For example, block ids are used with callouts, automated tests, and possibly some older forms of Blockly code validation.

This branch creates a new property on the Blockly Wrapper that stores a list of block ids that were explicitly set in the level (start blocks or toolbox). Any other ids found can safely be considered randomly generated by Blockly and discarded.

## Testing story

To test this change, I solved a Minecraft puzzle the same way twice. I logged the XML before and after the stripping and observed the following:
* Before stripping, each program contained with completely different ids, except for a `when_run` block with id `whenRun`.
* All random/unique ids were detected and removed.
* After stripping, each program's XML was identical.

I had to get creative with the way the `appOptions.level` object was passed and accessed due to failing automated tests. While `appOptions` is typically available globally, this didn't prove true in the context of some are tests that needed to traverse these code changes.

## PR Checklist:

- [ ] Code is well-commented


mikeharv:
This PR is a continuation of #56192, which was mis-closed by the Git LFS migration (#55759).

### Previous Comments:


### Previous Reviews:

apps/src/blockly/addons/cdoXml.js
```diff
@@ -113,9 +113,31 @@ export function getProjectXml(workspace) {
...
+/**
+ * Removes the randomized 'id' attribute from all 'block' elements.
+ * This is intended to prevent writing duplicate entries to the level_sources table.
+ * @param {Element} element - The XML element to process.
+ * @param {string[]} levelBlockIds - An array of ids to preserve, if found.
+ */
+function removeIdsFromBlocks(element) {
+  if (element.nodeName === 'block') {
+    const id = element.getAttribute('id');
+    if (id && !Blockly.levelBlockIds.includes(id)) {
```
molly-moen, line 130: this function is only used in the context of google blockly right?

mikeharv: Correct. And true of everything in this file!

molly-moen (approved):
Nice!

mikeharv:
I just added one more commit: https://github.com/code-dot-org/code-dot-org/pull/56299/commits/c9c9475dc35e1f6f64e371005598ae5cf56d0a57

The `sharepage.feature` UI test was failing: https://drone.cdn-code.org/code-dot-org/code-dot-org/39546

The reason for this is that loads in some block XML containing block ids which are needed to check the arrangement of the blocks.

https://github.com/code-dot-org/code-dot-org/blob/mike/strip-random-block-ids/dashboard/test/ui/features/step_definitions/flappy_steps.rb#L17-L21

In practice, these blocks would have random ids which are now getting stripped out. To preserve the ids for the test, we can just add them to the toolbox and start blocks XML.

ebeastlake:
🎉 Should we hold off on merging this until after the Sprite Lab migration?

mikeharv:
> 🎉 Should we hold off on merging this until after the Sprite Lab migration?

@ebeastlake I don't think it matters. We'll only be saving JSON for Sprite Lab (outside of levelbuilder functions), and never writing Sprite Lab solutions to the database.

Edit: Let's hold this until after the migration!

breville (approved):
Thanks for doing this!

breville:
It might be good to check the production database before and after this is merged just to verify that we are indeed saving fewer unique solutions!

mikeharv:
@breville Agreed! That is the plan.

PR #56231, opened 2024-02-02
Storybook: migrate `GameButtons` component

bencodeorg:
Woof, bit of a doozy this one. A few notes:
- The `common.css` file was needed for the run/reset icons to show up -- these icons are put in place [via CSS here](https://github.com/code-dot-org/code-dot-org/blob/d185420ee51c89096f517782f332a3a7b4cc908e/apps/style/common.scss#L1523-L1529). This stylesheet I believe is [generally available in our labs](https://github.com/code-dot-org/code-dot-org/blob/d185420ee51c89096f517782f332a3a7b4cc908e/dashboard/app/views/levels/_apps_dependencies.html.haml#L16), but I don't think is available in other contexts (eg, the teacher dashboard). This change makes that stylesheet available to all Storybook components (including ones that aren't in our labs) -- I skimmed a few components and didn't see any obvious problems, but definitely possible there could be conflicting styling. FWIW, we have other CSS files that are for specific contexts (`pd.css`, `levelbuilder.css`) that are loaded for all Storybook components already.
- There's some custom work that was done previously to get Minecraft styling to show up in a few examples here. I couldn't figure out how to get that to load, and I don't think it's particularly important to have Minecraft-styled buttons here, so I removed it.
- The `ResetButton` is hidden by default. The old Storybook entry for `GameButtons` had entries for both `GameButtons` (parent of run/reset buttons) as well as some entries for run and reset buttons individually. The run button is rendered/visible in `GameButtons`, but the `ResetButton` is hidden by default and there's no mechanism via props to get it to show up thru `GameButtons` (in our actual code, we hide/show it via code in `studioApp` 🤦 ). So, I added a standalone entry for the reset button in a couple of configurations.
- Remove an unused prop (`playspacePhoneFrame`) in `GameButtons`.
- Removed a couple of `title` entries from other stories that snuck in since I merged our Storybook reorganization PR.

## Testing story

Tested manually via local Storybook.

PR #56304, opened 2024-02-05
Storybook: migrate `GameButtons` component

bencodeorg:
Woof, bit of a doozy this one. A few notes:
- The `common.css` file was needed for the run/reset icons to show up -- these icons are put in place [via CSS here](https://github.com/code-dot-org/code-dot-org/blob/d185420ee51c89096f517782f332a3a7b4cc908e/apps/style/common.scss#L1523-L1529). This stylesheet I believe is [generally available in our labs](https://github.com/code-dot-org/code-dot-org/blob/d185420ee51c89096f517782f332a3a7b4cc908e/dashboard/app/views/levels/_apps_dependencies.html.haml#L16), but I don't think is available in other contexts (eg, the teacher dashboard). This change makes that stylesheet available to all Storybook components (including ones that aren't in our labs) -- I skimmed a few components and didn't see any obvious problems, but definitely possible there could be conflicting styling. FWIW, we have other CSS files that are for specific contexts (`pd.css`, `levelbuilder.css`) that are loaded for all Storybook components already.
- There's some custom work that was done previously to get Minecraft styling to show up in a few examples here. I couldn't figure out how to get that to load, and I don't think it's particularly important to have Minecraft-styled buttons here, so I removed it.
- The `ResetButton` is hidden by default. The old Storybook entry for `GameButtons` had entries for both `GameButtons` (parent of run/reset buttons) as well as some entries for run and reset buttons individually. The run button is rendered/visible in `GameButtons`, but the `ResetButton` is hidden by default and there's no mechanism via props to get it to show up thru `GameButtons` (in our actual code, we hide/show it via code in `studioApp` 🤦 ). So, I added a standalone entry for the reset button in a couple of configurations.
- Remove an unused prop (`playspacePhoneFrame`) in `GameButtons`.
- Removed a couple of `title` entries from other stories that snuck in since I merged our Storybook reorganization PR.

## Testing story

Tested manually via local Storybook.


bencodeorg:
This PR is a continuation of #56231, which was mis-closed by the Git LFS migration (#55759).

### Previous Comments:
- **snickell** [commented](https://github.com/code-dot-org/code-dot-org/pull/56231#issuecomment-1925555317) - This PR was accidentally auto-closed during the Git LFS migration (https://github.com/code-dot-org/code-dot-org/issues/55759), reopening it, sorry bout that 😘, -Seth
- **snickell** [commented](https://github.com/code-dot-org/code-dot-org/pull/56231#issuecomment-1925555521) - This PR was accidentally auto-closed during the Git LFS migration (https://github.com/code-dot-org/code-dot-org/issues/55759), reopening it, sorry bout that 😘, -Seth
- **snickell** [commented](https://github.com/code-dot-org/code-dot-org/pull/56231#issuecomment-1925555577) - This PR was accidentally auto-closed during the Git LFS migration (https://github.com/code-dot-org/code-dot-org/issues/55759), reopening it, sorry bout that 😘, -Seth

### Previous Reviews:

apps/src/templates/projects/PersonalProjectsTable.story.jsx
```diff
@@ -7,7 +7,6 @@ import publishDialog from '@cdo/apps/templates/projects/publishDialog/publishDia
 import deleteDialog from '@cdo/apps/templates/projects/deleteDialog/deleteProjectDialogRedux';
 
 export default {
-  title: 'PersonalProjectsTable',
```
jmkulwik, line 10: Thanks for cleaning this up!

jmkulwik (approved):
Nice work!

apps/src/templates/ResetButton.story.jsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import {ResetButton} from './GameButtons';
```
fisher-alice, line 2: Nice solution!

apps/src/templates/GameButtons.story.jsx
```diff
@@ -1,111 +1,21 @@
...
+export default {
+  component: UnconnectedGameButtons,
+};
 
-  storybook.storiesOf('Buttons/GameButtons/ResetButton', module).addStoryTable([
-    {
-      name: 'default',
-      description:
-        'You have to explicitly set display: block to make this show up. It is hidden by default?!',
-      story: () => <ResetButton style={{display: 'block'}} />,
```
fisher-alice, line 31: You explained that the `ResetButton` is hidden by default and there's no mechanism via props to get it to show up thru GameButtons (in our actual code, we hide/show it via code in studioApp. Do you mind explaining where this happens?
I see in `TopInstructions.story.jsx`, [a `createCommonStore` function](https://github.com/code-dot-org/code-dot-org/blob/dc60f4c3a1df3ba491d6489c860754d48a1a5939/apps/src/templates/instructions/TopInstructions.story.jsx#L30) is used to initialize a Redux store and set `pageConstants` values. Is this possible for `pageConstants.hideResetButton` and `pageConstants.hideRunButton`?
I like your solution, but just curious if you think this approach would've also worked.

fisher-alice (approved):
Just left a question. Great work!

apps/src/templates/GameButtons.jsx
```diff
@@ -105,7 +104,6 @@ export default connect(state => ({
   hideRunButton: state.pageConstants.hideRunButton,
   hideResetButton: state.pageConstants.hideResetButton,
   runButtonText: state.pageConstants.runButtonText,
-  playspacePhoneFrame: state.pageConstants.playspacePhoneFrame,
```
fisher-alice, line 108: TIL about the `placespacePhoneFrame`!

apps/src/templates/GameButtons.story.jsx
bencodeorg: haha yeah it is pretty confusing...so `ResetButton` is hidden by default per the `style` prop:

https://github.com/code-dot-org/code-dot-org/blob/5ccf9d443372bf05ee37b1846f1fdedd21334709/apps/src/templates/GameButtons.jsx#L40-L51

That `hideResetButton` prop gets passed as the `hidden` prop to `ResetButton`, which applies a css class that will hide the button, but not show it.

`GameButtons` doesn't pass a `style` prop to `ResetButton`, so you can't get it to show up via props from `GameButtons`. I could, of course, make this possible, but didn't want to make changes for the sake of Storybook (maybe I should?)

So, all of that to say, the Redux approach I don't think would work -- separately, I'd also generally advocate for us to use as little Redux as possible in Storybook.

bencodeorg: Ah and sorry, I think in our code this is how we hide/show the reset button:

https://github.com/code-dot-org/code-dot-org/blob/dc60f4c3a1df3ba491d6489c860754d48a1a5939/apps/src/StudioApp.js#L1112-L1116

fisher-alice: Oh interesting - thanks!
Do we need this line?
https://github.com/code-dot-org/code-dot-org/blob/5ccf9d443372bf05ee37b1846f1fdedd21334709/apps/src/templates/GameButtons.jsx#L48
Seems unnecessary since style is set below by `commonStyles.hidden` and updated accordingly in `StudioApp.js`.
Maybe a comment here would be helpful that styling is updated in `StudioApp`?

bencodeorg: I'm not really sure, I'll add a comment though. Here's a PR where Maddie got into this a bit if you want to know more:

https://github.com/code-dot-org/code-dot-org/pull/41118

fisher-alice: Interesting to learn - pretty complicated scenarios in App Lab! Thanks for adding the comment and sharing the PR.

PR #56305, opened 2024-02-05
Add keyboard navigability to Authored Hints

fisher-alice:
This PR makes the Character icon and lightbulb in the Instructions panel tab-navigable and adds an 'aria-label' to the icon/lightbulb. When a user presses the 'Enter' key on the icon/lightbulb, the focus moves to the 'Yes' button in the hint prompt bubble.

This PR also includes styling for `focus` for the following:
- immersive reader button
- yes / no button in hint prompt
- more/ less collapser button

I also added `aria-labelledby` to the yes/no hint prompt button and an `aria-label` to the TTS button so that when a user uses a screen reader and focuses on the button, 'text to speech button' is read out loud. The immersive-reader button already has this feature.

~~I made the decision to place the `div` that contains hint prompt character/lightbulb BEFORE the `div` that contains the instructions, hints and hint prompt in `ThreeColumns` [here](https://github.com/code-dot-org/code-dot-org/blob/17317dd97fd02b7fa005b4d8886c79aaa207ccf2/apps/src/templates/instructions/ThreeColumns.jsx#L52). Since the character is placed in the left column, it seemed to make more sense that when a user is keyboard-navigating that the character is sequenced before the chat bubbles in the middle column.~~ @samantha-code  - lmk what you think!
Updated - swapping back to original sequence according to Sam's feedback. Thanks Sam!

Screencast BEFORE update:

https://github.com/code-dot-org/code-dot-org/assets/107423305/78a9934e-48bb-4423-8ea9-6c22682f0295

Screencast AFTER update:

https://github.com/code-dot-org/code-dot-org/assets/107423305/9e80c0e9-8922-4642-9be8-118686e37165

Styling changes:

| Element | Not focused | BEFORE update | AFTER update |
| -- | -- | -- | -- |
| collapser button | no-focus-collapser | before-collapser | after-collapser |
| immersive reader button | no-focus-immersive | before-immersive-reader | after-immersive-reader |

I did try adding the hint prompt underneath the icon/lightbulb in [this branch](https://github.com/code-dot-org/code-dot-org/pull/56356), but it looked too squished. I added a screenshot to the closed PR.

## Links
[jira - Make Authored Hints lightbulb and yes/no in Instructions pane tab-navigable](https://codedotorg.atlassian.net/browse/LABS-468)
[jira - [a11y][VPAT] [Alt-text]character that gives hints in instructions does not have accessible text](https://codedotorg.atlassian.net/browse/LABS-471)

## PR Checklist:

- [ ] Code is well-commented


apps/i18n/common/en_us.json
```diff
@@ -2018,6 +2018,7 @@
   "privacyPolicy": "Privacy Policy",
   "privacyPracticesForChildren": "See our privacy practices for children",
   "projectWarning": "Note: You are on a level that is part of a longer project. Changes made on this level will also appear in other levels in the project.",
+  "promptIconAndLightbulbInstructionHints": "A friendly character with lightbulb image - click to display instruction hints prompt (Do you want a hint? yes or no) if hints are available",
```
fisher-alice, line 2021: @samantha-code  - lmk what you think of this aria-label. Thanks!

apps/src/templates/instructions/ThreeColumns.jsx
```diff
@@ -49,10 +49,10 @@ const ThreeColumns = props => {
 
   return (
     <div style={{...defaultStyles.container, ...styles.container}}>
+      <div style={{...defaultStyles.left, ...styles.left}}>{children[0]}</div>
```
fisher-alice: @samantha-code - this is where I switch the order of the keyboard-navigation. I can easily switch it back so that the user navigates to all the chat bubbles (including instructions, hints, feedback) and then to the character/lightbulb.

apps/src/templates/instructions/InstructionsCsfLeftCol.jsx
```diff
@@ -51,12 +53,34 @@ class InstructionsCsfLeftCol extends React.Component {
...
+      e.keyCode === KeyCodes.ENTER &&
+      this.props.hasAuthoredHints &&
+      this.props.hasUnseenHint
+    ) {
+      e.preventDefault();
+      this.props.requestHint();
+      // Defer focus until the next pass of the event loop. This is necessary so that the element
+      // has been added to the dom. Note that there can only be at most one hint prompt on the page.
+      setTimeout(
+        () => document.getElementById('hint-prompt-yes-button').focus(),
```
fisher-alice: Thanks @thomasoniii for helping me with this focus change!

apps/i18n/common/en_us.json
samantha-code: Perfect Label!

apps/src/templates/instructions/ThreeColumns.jsx
samantha-code: I think you should consider moving it to the end of the flow - thinking about when a user would want a hint and it would most likely be after they have been in the workspace trying something, so I think the less tabs between the workspace and the hint the better

fisher-alice: Sounds good! I updated code and after-update video. I agree after seeing it in action. Thanks!

apps/src/templates/instructions/InstructionsCsfLeftCol.jsx
```diff
@@ -51,12 +53,34 @@ class InstructionsCsfLeftCol extends React.Component {
...
     }
   };
 
+  /**
+   * Handle an enter key press to the hint display lightbulb
+   */
+  handleKeyDownLightbulb = e => {
+    // If we don't have authored hints to display, keyboard-navigating to lightbulb
+    // and then pressing 'enter' shouldn't do anything.
+    if (
```
bencodeorg, line 68: Is it undesirable to have this behavior if you click on it?

bencodeorg (approved):
Nice!

apps/src/templates/instructions/InstructionsCsfLeftCol.jsx
fisher-alice: Good question - I don't think so and I think it makes sense to have similar behavior. Thanks! will update.

apps/src/templates/instructions/HintPrompt.jsx
```diff
@@ -26,7 +26,12 @@ const HintPrompt = ({
       textToSpeechEnabled={textToSpeechEnabled}
     >
       <p>{message}</p>
-      <LegacyButton type="cancel" onClick={onConfirm} style={{marginRight: 5}}>
+      <LegacyButton
```
bencodeorg, line 29: Sorry, one other thought -- should this have an `aria-labelledby`? Wondering what the experience is like for a screen reader here, where I think it would just read "Yes" to you, rather than what you're saying yes to. I think you could use the `p` tag above if you gave it an ID?

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-labelledby

(to be honest, only skimmed these docs, so totally might not be the right approach!)

bencodeorg: oic, the label you include for the lightbulb kind of handles this. Feels non-standard, but probably does the trick!

bencodeorg: (on the other hand, if you didn't click yes immediately, then navigated back to the "Yes" button, you'd have no context, I think. So maybe worth considering?)

fisher-alice: This is so cool! Added the label and then checked with the screenreader. Great suggestion!

bencodeorg (approved):

PR #56224, opened 2024-02-02
Music: consolidate BeatPad code and convert to TypeScript

sanchitmalhotra126:
While converting BeatPad to TypeScript I found that there was a lot of now unused or refactor-able code, left over from when the beat pad used to be in a different place with different styles and functionality. Thankfully we're able to trim it down quite a bit now. I also moved the calculation of the selected trigger ID into MusicView/MusicBlocklyWorkspace just to be more logically consistent with how we compute and access other blockly fields, and stored the value in redux.

PR #56308, opened 2024-02-05
Music: consolidate BeatPad code and convert to TypeScript

sanchitmalhotra126:
While converting BeatPad to TypeScript I found that there was a lot of now unused or refactor-able code, left over from when the beat pad used to be in a different place with different styles and functionality. Thankfully we're able to trim it down quite a bit now. I also moved the calculation of the selected trigger ID into MusicView/MusicBlocklyWorkspace just to be more logically consistent with how we compute and access other blockly fields, and stored the value in redux.


sanchitmalhotra126:
This PR is a continuation of #56224, which was mis-closed by the Git LFS migration (#55759).

### Previous Comments:


### Previous Reviews:
- **bencodeorg approved** - Less familiar with some of the Music Lab specifics playing out here (not sure if you want @breville for more familiar eyes), but the changes generally look good to me!

apps/src/music/blockly/MusicBlocklyWorkspace.js
```diff
@@ -324,28 +324,54 @@ export default class MusicBlocklyWorkspace {
   }
 
   updateHighlightedBlocks(playingBlockIds) {
+    if (!this.workspace) {
```
sanchitmalhotra126, line 333: Copying over comments from https://github.com/code-dot-org/code-dot-org/pull/56224 for visibility
> @bencodeorg: Any value to differentiating these warning messages?

Oh that's a good idea, I'll add some specifics

apps/src/music/views/MusicView.jsx
```diff
@@ -357,6 +367,14 @@ class UnconnectedMusicView extends React.Component {
       return;
     }
 
+    if (e.type === Blockly.Events.CHANGE) {
```
sanchitmalhotra126, line 370: > @bencodeorg: Nit: one if statement?

👍

apps/src/music/blockly/MusicBlocklyWorkspace.js
```diff
@@ -324,28 +324,54 @@ export default class MusicBlocklyWorkspace {
...
+      this.metricsReporter.logWarning('workspace not initialized.');
+      return undefined;
+    }
+    const block = this.workspace.getBlockById(blockId);
+    if (!block) {
+      return undefined;
+    }
+    const isSelectedBlockTriggerAt =
+      block.type === BlockTypes.TRIGGERED_AT_SIMPLE2;
```
sanchitmalhotra126, line 379: > @bencodeorg: For my understanding, do these 2 suffixed block types (I see a couple of them) represent?

Oh yeah, this is a relic of when we were supporting a few different execution models for the sake of comparison and testing ("Simple", "Simple2", "Tracks", etc) - the one we landed on and have used since is "Simple2" so a lot of the blocks are suffixed with "SIMPLE2".

We had plans to continue supporting the other models for a while in case we wanted to switch back or experiment further but I think at this point we're pretty locked into "Simple2" and could probably go in and clean up all of these names (@breville what do you think?)

apps/src/music/views/KeyHandler.tsx
```diff
@@ -73,10 +72,6 @@ const KeyHandler: React.FunctionComponent<KeyHandlerProps> = ({
           reportKeyPress('toggle-headers');
           dispatch(toggleHeaders());
         }
-        if (event.key === 'b') {
```
sanchitmalhotra126, line 76: > @bencodeorg: Why is this no longer needed?

So originally the beat pad was a panel that popped up above the controls bar during playback, and we wanted a way to close it once opened. Now, it's been moved into the left controls box under the Run button and it doesn't make as much sense to hide/show it manually. There's also not a way for users to hit this unless they know the special `ui-keyboard-shortcuts-enabled` URL param

apps/src/music/views/BeatPad.tsx
```diff
@@ -0,0 +1,47 @@
...
+          <button
+            type="button"
+            id={trigger.id}
+            key={trigger.id}
+            className={classNames(
+              styles.triggerButton,
+              trigger.id === selectedTriggerId && styles.triggerButtonSelected,
+              isPlaying && styles.triggerButtonActive
+            )}
+            onClick={!isPlaying ? undefined : () => playTrigger(trigger.id)}
```
sanchitmalhotra126: > @bencodeorg: Nit: easier to read if you don't invert isPlaying?

👍

bencodeorg (approved):

apps/src/music/blockly/MusicBlocklyWorkspace.js
breville: I would be comfortable cleaning up these names, now that we really have settled on `"Simple2"`!

PR #56309, opened 2024-02-05
Better prompt for reason when performing progress reset

dju90:
Right now, only Code.org employees can delete user progress. The only reason that Customer Success is deleting user progress is when a teacher requests it. 100% of the user progress that has been reset by Customer Success has had the reason: TR or Teacher Request. See [Customer Support Wishlist for additional context](https://docs.google.com/document/d/1QpZhvFDo4-xr3Yrd94-U2ULQEcldSMU8LiSIq5lKnSo/edit).

This change pre-fills the reason field to reduce friction for the CX team while also prompting for a zendesk ticket number. In the long term, we want better audit logs of use of this tool, but for now, this should address a pain point for the CX team while still requiring a reason.

## Testing story
Tested locally
Screenshot_7-3-2024_102640_localhost-studio code org

## PR Checklist:

- [ ] Code is well-commented


dju90:
This PR is a continuation of #56178, which was mis-closed by the Git LFS migration (#55759).

### Previous Comments:


### Previous Reviews:

etaderhold:
A reason for deleting data has been pretty standard practice at other places I've worked. It can (and should) be as simple as a link to the ticket requesting this deletion. It just helps make sure that the exercise of admin privileges is auditable in case someone abuses these privileges.

dju90:
> A reason for deleting data has been pretty standard practice at other places I've worked. It can (and should) be as simple as a link to the ticket requesting this deletion. It just helps make sure that the exercise of admin privileges is auditable in case someone abuses these privileges.

The rationale you explain makes sense to me, but I'd still like to move forward with this change to prioritize reducing friction for customer success, based on two facts:
1. Currently, the only people resetting progress via this UI are customer success, and for 100% of progress reset tickets they put "TR" (meaning "teacher request"). The ability to reset progress via this form has been around since 2021, with not one recorded case of abuse of these privileges - in my opinion, the time lost by CX outweighs the risk associated with not requiring a reason.
2. There's no validation on what is entered into the "reason" field - someone could just as easily put in a random letter and submit the form, which makes it far less auditable. If someone did abuse this privilege, I doubt the "reason" field would be the thing that would enable us to effectively audit the activity.

Open to hearing your thoughts on this though

etaderhold (approved):

PR #56314, opened 2024-02-05
Convert a few Blockly files to TypeScript

molly-moen:
Initial steps at adding some TypeScript to our Blockly code. I just converted a couple files here:
- `constants.js` -> `constants.ts` : the only change that was needed here was typing the parameters of two functions. We also were explicitly importing things from this file via `constants.js`, so I updated those imports to just be from `constants`.
- `ModalFunctionEditor.jsx` -> `ModalFunctionEditor.tsx`: our only React component in the blockly folder. All that was required here was typing the state from redux and changing the locale import to the TypeScript version (see #53623 for details on this)
- `utils.js` -> `utils.ts`: this one was more interesting:
  -  Most of the changes were adding types, either basic ones or ones provided by Blockly. 
  -  I also learned that `reduce` in TypeScript utilizes [generics](https://www.typescriptlang.org/docs/handbook/2/generics.html), so you need to add a type to your call of `reduce` to specify what type it will return.
  - Finally, we had a function, `findFlyout`, which relied on private fields in Blockly's `WorkspaceSvg` class (`flyout` and `toolbox_.flyout_`. If we want to use TypeScript here, we could not reference a private field. I looked at `WorkspaceSvg` to see if there was an alternative. Very conveniently, `WorkspaceSvg.getFlyout()` was doing the exact same thing as this entire function ([code](https://github.com/google/blockly/blob/0b25a8a27f156a8104cd242ee68de304506693cd/core/workspace_svg.ts#L999-L1007)). Therefore I replaced use of this function with blockly's function instead (it was only used in one place, `eventHandlers`). We may not always get such a neat solution for our use of private fields, but this one ended up making our code shorter.
 

## Links

- jira ticket: [CT-307](https://codedotorg.atlassian.net/browse/CT-307)

## Testing story
I tested locally that Google Blockly labs (Dance Party and Sprite Lab on google blockly) still worked as expected, specifically around the modal function editor and browser resizing, which is what this touched.

## Follow-up work
I am spending this week trying to move some of our Blockly code to TypeScript, so expect more PRs like this. I am trying to keep each PR relatively small and low risk.

## PR Checklist:

- [ ] Code is well-commented


molly-moen:
This PR is a continuation of #56171, which was mis-closed by the Git LFS migration (#55759).

### Previous Comments:


### Previous Reviews:
- **ebeastlake commented** - 
- **ebeastlake commented** - 
- **ebeastlake commented** - 
- **ebeastlake commented** - 
- **molly-moen commented** - 
- **molly-moen commented** - 
- **molly-moen commented** - 
- **sanchitmalhotra126 commented** - 
- **ebeastlake commented** - 
- **ebeastlake approved** -

apps/src/blockly/eventHandlers.js
```diff
@@ -97,15 +96,10 @@ export function adjustCalloutsOnViewportChange(event) {
 // When the browser is resized, we need to re-adjust the width of any open flyout.
 export function reflowToolbox(event) {
   const mainWorkspace = Blockly.getMainWorkspace();
-  const mainWorkspaceFlyout = findFlyout(mainWorkspace);
-  if (mainWorkspaceFlyout) {
-    mainWorkspaceFlyout.reflow();
-  }
+  mainWorkspace?.getFlyout()?.reflow();
```
mikeharv, line 99: Great find!

apps/src/blockly/components/ModalFunctionEditor.tsx
```diff
@@ -7,14 +7,15 @@ import {
 import moduleStyles from './modal-function-editor.module.scss';
 import classNames from 'classnames';
 import Button from '@cdo/apps/templates/Button';
-import msg from '@cdo/locale';
+import {commonI18n} from '@cdo/apps/types/locale';
```
mikeharv, line 10: Thanks for including the link about this particular change!

mikeharv (approved):
Cool! I'm excited to see that these changes look fairly straight-forward! I don't think it had occurred to me that we could start to tackle this with small pieces like this. :tada:

PR #56321, opened 2024-02-05
Fix Facebook icon not loading in Eyes tests

TurnerRiley:
We've been having a [flaky eyes test](https://eyes.applitools.com/app/test-results/00000251695572159815/00000251695572159676/steps/7?accountId=mG4DoRfv0ntEgI3XIYBulrKxlZqrHtGwbP3V103ojvLFU110&mode=step-editor) where the Facebook icon [doesn't seem to load fast enough](https://codedotorg.slack.com/archives/C0T0PNTM3/p1706729232836649) sometimes. At first, I tried adding a `ui-test` id to the Facebook icon element so that I could have the Eyes test wait until it found an element with that id on the page. However, for some reaosn it would find the icon by id in the first part of the test (when it directly navigated to the `/congrats` page), but not on any subsequent pages (even though the Facebook icon would be visible in the SauceLabs preview). I also double checked locally that the Facebook logo was showing up and had the correct id.

So, as an alternative safe fix, I'm adding some wait times to give the Facebook logo more time to load since that seems to be the source of the flakiness.

## Links
Example test failure: [here](https://eyes.applitools.com/app/test-results/00000251695572159815/00000251695572159676/steps/7?accountId=mG4DoRfv0ntEgI3XIYBulrKxlZqrHtGwbP3V103ojvLFU110&mode=step-editor)
Slack discussion: [here](https://codedotorg.slack.com/archives/C0T0PNTM3/p1706729232836649)


TurnerRiley:
This PR is a continuation of #56193, which was mis-closed by the Git LFS migration (#55759).

### Previous Comments:


### Previous Reviews:
- **kelbyhawn approved** -

bethanyaconnor (approved):

PR #56325, opened 2024-02-05
Fix account settings page that fails with too many students

nicklathe:
Currently when a teacher with a large number of students in sections they own opens their `/users/edit` page, the page fails to load, as the query to get all students times out. Ultimately, this list of students is used in a warning message if they try to delete their account. It's not necessary or even helpful to display a long list of students, rather the importance is in the message and knowing that deleting their account will delete some number of student accounts as well. _Also_, the vast majority of teachers navigating to their `/users/edit` page aren't going there to delete their account.

This PR removes the logic and UI that displays a list of students, get's rid of the expensive query, and only displays a number of students to be deleted, in the warning message.

- Add new method in user.rb to query for students and return count
- Change React components to take a count vs. list of students
- Remove UI that displays list of students
- Only show number of students to be deleted in warning message

Old dialogue box
image

New dialogue box
Screenshot 2024-02-21 at 9 45 38 AM

## Links

[Jira ticket](https://codedotorg.atlassian.net/browse/P20-658)

## PR Checklist:

- [ ] Code is well-commented


nicklathe:
This PR is a continuation of #56222, which was mis-closed by the Git LFS migration (#55759).

### Previous Comments:


### Previous Reviews:
- **nicklathe commented** - 
- **nicklathe commented** - 
- **daynew approved** - Thank you for fixing the long time issue. 
 I just had one request.
- **daynew commented** - 
- **nicklathe commented** - 
- **wilkie commented** - 
- **wilkie commented** - 
- **nicklathe commented** -

dashboard/lib/queries/user.rb
```diff
@@ -0,0 +1,9 @@
+require 'policies/user'
+
+class Queries::User
+  def self.dependent_students_count(user_id)
+    user = User.find(user_id)
+    # Limit the number of students to mitigiate slow page loads on /users/edit
+    user.students.distinct.limit(Policies::User::DEPENDENT_STUDENTS_COUNT_LIMIT).count
```
nicklathe: @wilkie, this query should do what we want now, vs. the old query where I was limiting first then calling uniq. This is the sql query for the above code:

```
u.students.distinct.limit(2).count

(18.2ms)  SELECT COUNT(count_column) FROM (SELECT DISTINCT `users`.`id` AS count_column FROM `users` INNER JOIN `followers` ON `users`.`id` = `followers`.`student_user_id` INNER JOIN `sections` ON `followers`.`section_id` = `sections`.`id` INNER JOIN `section_instructors` ON `sections`.`id` = `section_instructors`.`section_id` WHERE `section_instructors`.`instructor_id` = 18 AND `section_instructors`.`status` = 0 AND `sections`.`deleted_at` IS NULL LIMIT 2) subquery_for_count
```

wilkie:
So a few things! I've pulled things down and experimented on a local build.

I wrote a script to create hundreds of sections with dozens of students each. The students were created with usernames and passwords, so they got an email AuthenticationOption. This presented one issue... here is what the old code would do:

image
image

So, the old code took 13 seconds to render the edit page and it showed you that deleting your account would delete zero students. This is because no student is dependent on the teacher.

The new code was dramatically better, but mostly because it isn't looking for the same information:

image
image

It says 100+ students would be deleted when, in reality, no students will be deleted. It just assumes all students in your sections would lose their accounts.

We could get away with this if we say "You have 100+ students... some of them might lose access to their accounts..." etc.

wilkie:
Looking at it, it collects them in the old `dependent_students` call very inefficiently by looping through and evaluating `depends_on_teacher_for_login?` on each student `User`. Yikes!

This logic thoroughly checks that it is a:

1. student without password with migrated auth options
2. is either a teacher managed account (secret picture / words section and no auth options or user password)
3. or it is a roster managed account (part of a section that is a rostered via google classroom, etc) and the student has only one auth option (presumed to match the section login type)
4. the student only has one teacher

A query for that is just never going to be very good... I can't seem to get half of that without hitting an unindexed path.

We can find the number of students in an 'externally rostered' section (but this doesn't count students that have their own email nor students in extra sections... so it is an approximation of a maximum):

```
User.joins(:sections_as_student).where(users: {user_type: 'student'}, sections_as_student: {user: current_user}).where.not(sections_as_student: {login_type: ['email']}).distinct.count
```

Even this takes a few seconds in my stress test... so probably not worth doing this or much of anything more...
because to be pretty sufficient one would need to count the number of students in externally rostered sections with just one `AuthenticationOption` and in only one `Section` ... that query is way too much trouble. So we just stick to the simple query, I think.

A better way of writing that query would be:

```
current_user.sections.where.not(login_type: 'email').joins(:students).select(:'user.id').distinct.count
```

Which runs in no time at all on my stress test.

-----

So, I like how you are doing it... just how many students you have as a kind of upper bound plus a warning that any that would be orphaned will have their own accounts deleted.

I would just change the wording to be a little less dramatic and get the number of students by a rough count like the query above of simply how many students are in an externally rostered section:

"You have {} students that log into Code.org via secret words, secret pictures, or a third-party service and might not have set up a personal login. Deleting your account may also delete the accounts of many of those students." ... and then the normal "Please give those students... etc"

nicklathe:
Think we should throw a `.limit` on your suggested query, or will this query be efficient enough that we don't need to worry about it? Since we aren't looping over each student and doing some additional logic/query, my gut says we probably don't need to limit it, since this single query should be efficient enough.

Thanks for the awesome explanation!

wilkie:
The limit is probably not as necessary because it is counting on all indexed associations... I stress tested with 600 sections of different login types of 30 students each sampled out of thousands and it came up with the answer in around 20-50ms on my little VM.

wilkie:
If you do a limit just to get the 100 or 100+, it won't hurt to put the limit on the join... hmm where does that go... `joins(:students).distinct.limit(100).count` or something? It's really the student join that hurts the most, I think.

nicklathe:
We don't need the limit for the message anymore, so if this query is efficient enough without it (sounds like it is based on your stress testing), I think I'll leave it off.

wilkie (approved):
Looks great! Thanks for working on this and solving this longstanding problem for our teachers and customer support team! 👍

PR #56327, opened 2024-02-05
Fix Sprite Lab test for Google Blockly

molly-moen:
We are flipping Sprite Lab to Google Blockly later this week. We have one test that needed to be updated in order to work once we flip. This is because our previous dropdown logic does not work for Google Blockly. Since the new test only works on Google Blockly, I set `spritelab.feature` to explicitly run on google blockly for now. Once we have flipped we can remove the `?blocklyVersion` url parameter from this test and our other Google Blockly sprite lab tests.

This test does not work on iPad or iPhone; the selenium click event on a dropdown does not work on these emulators for some reason. Therefore I disabled it on mobile.

Once we flip we should see some eyes diffs as well, so we will need to approve the new eyes baselines.

## Links

- jira ticket: [CT-314](https://codedotorg.atlassian.net/browse/CT-314)

## Testing story
Tested locally that all sprite lab specific tests pass once we are on Google Blockly. There is a slight chance I missed one, if there is an non-labs specific test that uses a sprite lab level.

## Follow-up work
Flip Sprite Lab to Google Blockly!

## PR Checklist:

- [ ] Code is well-commented


dashboard/test/ui/features/star_labs/spritelab/spritelab.feature
```diff
@@ -22,8 +22,9 @@ Scenario: Losing the first level
   And element "#runButton" is visible
   And element "#resetButton" is hidden
 
+@no_mobile
 Scenario: Winning the first level
-  Then I press dropdown number 6
+  Then I click block field "[data-id='make-new-sprite'] > .blocklyEditableText" number 1
```
mikeharv, line 27: I'm used to these steps being more human-readable. Would it be worth abstracting out the selector to a new step helper so that this line could just say something like:

```
Then I click block make-new-sprite field number 1
```

I realize that this has implications for a bunch of dance-ai tests that would need to be revised, so feel free to acknowledge as out of scope.

mikeharv (approved):
Looks good to me, and the new step definitions make sense. Approving now, but I wonder what your thoughts are on the refactor I suggest as a possible clean up task.

dashboard/test/ui/features/star_labs/spritelab/spritelab.feature
molly-moen: yeah that is a good follow up task, [filed](https://codedotorg.atlassian.net/browse/CT-319)!

PR #56338, opened 2024-02-06
[WIP] Split up Music View

sanchitmalhotra126:
Main significant changes to look at:
- Pulled out the view code from `MusicView.jsx` into a separate, mostly pure-view `MusicLabView.tsx`. Since this is a functional component, I also absorbed the logic from `UpdateTimer` and `ValidatorProvider`, and removed those two util components.
- Pulled out most of the playback/execution logic from `MusicView.jsx` into `MusicProgramExecutor.ts`. Includes things like starting/stopping playback, compiling and executing code, interfacing with the player and blockly workspace components; the goal was to try to move as much non-React logic out of MusicView and into this non-React unit.
- At this point, `MusicView.jsx` primarily just loads the library and otherwise wraps callbacks to the program executor, and passes them into MusicLabView to render.
- I started a functional component rewrite of `MusicView` in `MusicViewContainer` but this is incomplete for now. As of now, the app still works as far as I can tell, with `MusicView` still being the top-level component.


sanchitmalhotra126:
This PR is a continuation of #56135, which was mis-closed by the Git LFS migration (#55759).

### Previous Comments:


### Previous Reviews:

PR #56339, opened 2024-02-06
 [labs-275] Dance AI: Clean up unused strings (ex. Strings with ‘2’ suffix)

cnbrenci:
Remove unused strings leftover from danceAI.  

Notably, I left in the "2" at the end of the strings. In order to keep the original names without the "2" we'd have to remove the unused strings (essentially this PR), wait for translations to sync, then remove the "2" from the ends of the strings (renaming them to the original names that we're removing in this PR) and re-translate them. I chose to leave the "2" at the end and just clean up the unused string versions.

## Links

- jira ticket: [labs-275 Dance AI: Clean up unused strings (ex. Strings with ‘2’ suffix)](https://codedotorg.atlassian.net/browse/LABS-275)

## Testing story

Searched codebase to validate no usages for the removed strings.
Ran through DanceAI locally.


cnbrenci:
This PR is a continuation of #56214, which was mis-closed by the Git LFS migration (#55759).

### Previous Comments:


### Previous Reviews:
- **cnbrenci commented** - I'm not sure how this will work for the translated strings- is it fine to use the same names as previously existed? Any weird fallback that might show the old strings briefly or something until they're translated? I want to make sure it's OK to re-use the original string names.
- **breville commented** - Will this rename to an existing key have an effect on our translations?
- **cnbrenci commented** - Yeah I was hoping you'd know the answer to that :P @mgc1194 do you know if this is going to be problematic? Do I need to avoid re-using the string names? I wanted to keep the ones that don't have "2" at the end, but if that's going to cause problems with translations I can just clean up the unused ones and leave the "2" ones.

mgc1194 (approved):

fisher-alice (approved):
Thanks for doing this! 
So just confirming that after this is merged and after the i18n sync, we could still update the string key names and remove the `2` later on. But the repercussions of this would be that the English string in non-English languages would be used until the strings are translated? Is that correct @mgc1194 ?

breville (approved):
Thanks for this tidy-up.  The background is that we were fine-tuning a variety of strings near the very end of the release last year, but wanted to keep the prior versions of them just in case we pivoted back at the last minute.

PR #56210, opened 2024-02-02
Draft of new setup instructions for maker/setup

cnbrenci:
Switching the CP Express to bootloader mode and re-installing the firmata firmware is not an alternative pathway like we previously assumed. It must be done every time one switches back to Code.org after using Microsoft MakerCode. With that in mind, we are including the instructions in our main maker/setup instructions. 

image

## Links

- jira ticket: [Labs-282 Improve UI/instructions how to update the Firmata firmware for CP Express using alternative path](https://codedotorg.atlassian.net/browse/LABS-282)

## PR Checklist:

- [ ] Tests provide adequate coverage
- [ ] Privacy and Security impacts have been assessed
- [ ] Code is well-commented
- [ ] New features are translatable or updates will not break translations
- [ ] Relevant documentation has been added or updated
- [ ] User impact is well-understood and desirable
- [ ] Pull Request is labeled appropriately
- [ ] Follow-up work items (including potential tech debt) are tracked and linked

PR #56340, opened 2024-02-06
[LABS-282] Improve UI/instructions how to update the Firmata firmware for CP Express using alternative path

cnbrenci:
Switching the CP Express to bootloader mode and re-installing the firmata firmware is not an alternative pathway like we previously assumed. It must be done every time one switches back to Code.org after using Microsoft MakerCode. With that in mind, we are including the instructions in our main maker/setup instructions. 

Shoutout to @fisher-alice for providing image assets and being a hand model!

The old page looked like this:
image

The new page adds a new section below the Connect your Board: Checklist - 

For Windows:
image

Not Windows (The only difference is the screenshot in step 2):
image
## Links

- jira ticket: [Labs-282 Improve UI/instructions how to update the Firmata firmware for CP Express using alternative path](https://codedotorg.atlassian.net/browse/LABS-282)

## Testing story

Tested locally in Edge to make sure it looks right and jumplinks work.


cnbrenci:
This PR is a continuation of #56210, which was mis-closed by the Git LFS migration (#55759).

### Previous Comments:


### Previous Reviews:
- **fisher-alice commented** - 
- **fisher-alice commented** - 
- **fisher-alice commented** - 
- **fisher-alice commented** - 
- **fisher-alice commented** - 
- **fisher-alice commented** - 
- **fisher-alice commented** - 
- **cnbrenci commented** - 
- **fisher-alice commented** - 
- **cnbrenci commented** -

dancodedotorg:
The is looks great! I love that we're able to differentiate instructions between devices. I didn't look too closely at the actual steps and trust they make sense from your testing, but let me know if you want more folks to test!

I wonder about this design pattern of having images break up lists. I was just thinking about this in another context - it shows up on our "how to have students join a section" page where the SSO providers appear as an image in a list (example - https://studio.code.org/teacher_dashboard/sections/5211092/login_info). And that felt a little awkward and disjointed to me when I saw it the other day. I wonder about that here too, with the images in the middle of a list. 

Are there other patterns we can try? Maybe clickable expandable images or using expandable < details > elements to see images? Those are just some initial ideas - in general I'm curious what folks think of that pattern.

apps/src/lib/kits/maker/ui/setup-instructions.module.scss
```diff
@@ -0,0 +1,24 @@
+:export {
+  checklistId: connectYourBoardChecklist;
```
fisher-alice: Just curious why you defined your constants here instead of in `SetupInstructions`. I see other constants defined in `module.scss` files but they seem to related more deriectly with styling.

apps/src/lib/kits/maker/ui/SetupGuide.jsx
```diff
@@ -71,43 +59,12 @@ export default class SetupGuide extends React.Component {
...
-  href: PropTypes.string.isRequired,
-  imgSrc: PropTypes.string.isRequired,
-  imgStyle: PropTypes.object,
-  description: PropTypes.string.isRequired,
-  divStyle: PropTypes.object,
-  alt: PropTypes.string.isRequired,
-};
-
-function HeaderCard(props) {
+function HeaderCard({id, title, description, divStyle}) {
```
fisher-alice, line 62: Nice cleanup!

apps/src/lib/kits/maker/ui/setup-instructions.module.scss
```diff
@@ -0,0 +1,24 @@
...
+.imageContainer img {
+  max-height: 160px;
+}
+
+li p {
+  margin-bottom: 0;
+}
+
+.troubleshootList {
+  list-style-type: "⚠️";
```
fisher-alice: Could we add a space in-between the warning emoji and text?

apps/src/lib/kits/maker/ui/SetupInstructions.jsx
```diff
@@ -82,16 +91,112 @@ class ConnectionInstructions extends React.Component {
...
+                        {applabI18n.makerSetupCPXInstallStep1c({
+                          buttonName: RESET_BUTTON_NAME,
+                        })}
+                      </li>
+                      <li>{applabI18n.makerSetupCPXInstallStep1d()}</li>
+                    </ul>
+                  </li>
+                </ul>
+              </div>
+              <figure className={styles.imageContainer}>
```
fisher-alice: What do you think about adding style to this `figure` element? Maybe a light gray border?

apps/src/lib/kits/maker/ui/SetupInstructions.jsx
```diff
@@ -82,16 +91,112 @@ class ConnectionInstructions extends React.Component {
...
+          openExternalLinksInNewTab={true}
         />
         {connectionState}
+        <CPExpressInstallInstructions />
+      </div>
+    );
+  }
+}
+
+class CPExpressInstallInstructions extends React.Component {
```
fisher-alice, line 121: You could consider using subrender functions within `CPExpressInstallInstructions` to help parse this pretty long`render` function. - maybe for step 1 and step 2?

fisher-alice, line 121: What do you think of moving `ConnectionInstructions` and this new component into separate files? Don't have strong opinion about this but `SetupInstructions` is starting to get large.

apps/src/lib/kits/maker/ui/SetupInstructions.jsx
```diff
@@ -82,16 +91,112 @@ class ConnectionInstructions extends React.Component {
...
+            </ul>
           </li>
           <li>{applabI18n.makerSetupWebSerialSuccessfulConnection()}</li>
         </ol>
         <SafeMarkdown
           markdown={applabI18n.makerSetupWebSerialSupportArticle()}
+          openExternalLinksInNewTab={true}
         />
         {connectionState}
+        <CPExpressInstallInstructions />
```
fisher-alice, line 115: It seems to me that these 'install' instructions are separate from the 'connection' instructions. What do you think of moving `CPExpressInstallInstructions` above to [line 31](https://github.com/code-dot-org/code-dot-org/blob/aebd9d2979ad70a665d2474a776ef236ad57e2c1/apps/src/lib/kits/maker/ui/SetupInstructions.jsx#L31).

fisher-alice:
> The is looks great! I love that we're able to differentiate instructions between devices. I didn't look too closely at the actual steps and trust they make sense from your testing, but let me know if you want more folks to test!
> 
> I wonder about this design pattern of having images break up lists. I was just thinking about this in another context - it shows up on our "how to have students join a section" page where the SSO providers appear as an image in a list (example - https://studio.code.org/teacher_dashboard/sections/5211092/login_info). And that felt a little awkward and disjointed to me when I saw it the other day. I wonder about that here too, with the images in the middle of a list.
> 
> Are there other patterns we can try? Maybe clickable expandable images or using expandable < details > elements to see images? Those are just some initial ideas - in general I'm curious what folks think of that pattern.

We could add text/button that says 'View screenshot image' or something along those lines.
Something similar to how ['View full list' works](https://github.com/code-dot-org/code-dot-org/blob/aebd9d2979ad70a665d2474a776ef236ad57e2c1/apps/src/templates/projects/StartNewProject.jsx#L87-L133) on '/projects' page.

Another thought is that we could make the ‘Install the Firmata firmware onto the CP Express’ expandable/collapsible? Then when the user clicks on ‘checklist’ above, the ‘Install the Firmata firmware onto the CP Express’ list could be collapsed so the user sees only one list - connecting your board.

fisher-alice (commented):
Awesome work! This is such an improvement. Just left a few minor comments/questions. Thanks!

cnbrenci:
> The is looks great! I love that we're able to differentiate instructions between devices. I didn't look too closely at the actual steps and trust they make sense from your testing, but let me know if you want more folks to test!
> 
> I wonder about this design pattern of having images break up lists. I was just thinking about this in another context - it shows up on our "how to have students join a section" page where the SSO providers appear as an image in a list (example - https://studio.code.org/teacher_dashboard/sections/5211092/login_info). And that felt a little awkward and disjointed to me when I saw it the other day. I wonder about that here too, with the images in the middle of a list.
> 
> Are there other patterns we can try? Maybe clickable expandable images or using expandable < details > elements to see images? Those are just some initial ideas - in general I'm curious what folks think of that pattern.

@dancodedotorg Could you share a screenshot of the example at https://studio.code.org/teacher_dashboard/sections/5211092/login_info? I'm getting 403 forbidden and a blank page

apps/src/lib/kits/maker/ui/SetupInstructions.jsx
cnbrenci: That is where I originally had it, but I moved it here because if you load the page from somewhere the doesn't support WebSerial (!shouldUseWebSerial()) then we don't show the ConnectionInstructions (return null on line 77) so I thought it'd be weird to show these instructions in that case.

Maybe it would be clearer to pull that logic up into the SetupInstructions

apps/src/lib/kits/maker/ui/setup-instructions.module.scss
cnbrenci: That's a good question tbh. I dunno, I guess I just was thinking they're id names so they're part of the CSS. but they aren't actually used anywhere in the CSS so maybe it makes more sense to put them in SetupInstructions.

apps/src/lib/kits/maker/ui/setup-instructions.module.scss
cnbrenci: Oh yeah it is real squished looking. Good catch.

apps/src/lib/kits/maker/ui/SetupInstructions.jsx
cnbrenci: Here is a treatment similar to the cards on code.org/maker - I was a little skeptical with just an image alone and without any other content but I think it looks nice! What do you think?
image

apps/src/lib/kits/maker/ui/SetupInstructions.jsx
cnbrenci: I kinda want to keep them together because they have a bit of shared information between them (the jumplinks between the two components and the styling for the troubleshooting lists). I spose I could refactor that stuff tho so both can use it. Lemme noodle on it.

fisher-alice: Keeping them together sounds reasonable to me!

fisher-alice:
> > The is looks great! I love that we're able to differentiate instructions between devices. I didn't look too closely at the actual steps and trust they make sense from your testing, but let me know if you want more folks to test!
> > I wonder about this design pattern of having images break up lists. I was just thinking about this in another context - it shows up on our "how to have students join a section" page where the SSO providers appear as an image in a list (example - https://studio.code.org/teacher_dashboard/sections/5211092/login_info). And that felt a little awkward and disjointed to me when I saw it the other day. I wonder about that here too, with the images in the middle of a list.
> > Are there other patterns we can try? Maybe clickable expandable images or using expandable < details > elements to see images? Those are just some initial ideas - in general I'm curious what folks think of that pattern.
> 
> @dancodedotorg Could you share a screenshot of the example at https://studio.code.org/teacher_dashboard/sections/5211092/login_info? I'm getting 403 forbidden and a blank page

Dan shared (via Slack) screenshot of an example where we break up steps in a list with an image. Posting here for visibility:

image (2)

apps/src/lib/kits/maker/ui/SetupInstructions.jsx
fisher-alice: Yeah - I agree it would make sense to pull logic into setup instructions and instead of returning `null` we could return an alternative message that states that the device/computer needs to support WebSerial in order to be able to use the Code.org Maker Toolkit.

apps/src/lib/kits/maker/ui/setup-instructions.module.scss
fisher-alice: Unless the `id` is assigned a previously defined variable/prop, it seems like you can just use the string like [this example](https://github.com/code-dot-org/code-dot-org/blob/f357947e70c84eafd0a0420f35b5e12a29e4ffa8/apps/src/lib/kits/maker/ui/SetupGuide.jsx#L66) in the same file. You can see other examples in [this file](https://github.com/code-dot-org/code-dot-org/blob/staging/apps/src/dance/ai/DanceAiModal.tsx).

cnbrenci: I do still want them to be defined as a variable since they are referenced in 2 different places (because the anchor tag uses the value of the id to jump there)

fisher-alice: Gotcha - that makes sense. Thanks!

apps/src/lib/kits/maker/ui/SetupInstructions.jsx
fisher-alice: Looks good! We both decided offline to decrease padding -thanks!

apps/src/lib/kits/maker/ui/SetupInstructions.jsx
fisher-alice: Discussed offline that returning an alternative message is outside scope of this task and could be a follow-up task.

fisher-alice:
> > The is looks great! I love that we're able to differentiate instructions between devices. I didn't look too closely at the actual steps and trust they make sense from your testing, but let me know if you want more folks to test!
> > I wonder about this design pattern of having images break up lists. I was just thinking about this in another context - it shows up on our "how to have students join a section" page where the SSO providers appear as an image in a list (example - https://studio.code.org/teacher_dashboard/sections/5211092/login_info). And that felt a little awkward and disjointed to me when I saw it the other day. I wonder about that here too, with the images in the middle of a list.
> > Are there other patterns we can try? Maybe clickable expandable images or using expandable < details > elements to see images? Those are just some initial ideas - in general I'm curious what folks think of that pattern.
> 
> We could add text/button that says 'View screenshot image' or something along those lines. Something similar to how ['View full list' works](https://github.com/code-dot-org/code-dot-org/blob/aebd9d2979ad70a665d2474a776ef236ad57e2c1/apps/src/templates/projects/StartNewProject.jsx#L87-L133) on '/projects' page.
> 
> Another thought is that we could make the ‘Install the Firmata firmware onto the CP Express’ expandable/collapsible? Then when the user clicks on ‘checklist’ above, the ‘Install the Firmata firmware onto the CP Express’ list could be collapsed so the user sees only one list - connecting your board.

Just wanted to add that this is a bit outside the scope of this task and could be logged as a follow-up task. Thanks!

apps/src/lib/kits/maker/ui/SetupInstructions.jsx
cnbrenci: Issue created to address the alternative message https://codedotorg.atlassian.net/browse/LABS-508

cnbrenci:
> > > The is looks great! I love that we're able to differentiate instructions between devices. I didn't look too closely at the actual steps and trust they make sense from your testing, but let me know if you want more folks to test!
> > > I wonder about this design pattern of having images break up lists. I was just thinking about this in another context - it shows up on our "how to have students join a section" page where the SSO providers appear as an image in a list (example - https://studio.code.org/teacher_dashboard/sections/5211092/login_info). And that felt a little awkward and disjointed to me when I saw it the other day. I wonder about that here too, with the images in the middle of a list.
> > > Are there other patterns we can try? Maybe clickable expandable images or using expandable < details > elements to see images? Those are just some initial ideas - in general I'm curious what folks think of that pattern.
> > 
> > 
> > We could add text/button that says 'View screenshot image' or something along those lines. Something similar to how ['View full list' works](https://github.com/code-dot-org/code-dot-org/blob/aebd9d2979ad70a665d2474a776ef236ad57e2c1/apps/src/templates/projects/StartNewProject.jsx#L87-L133) on '/projects' page.
> > Another thought is that we could make the ‘Install the Firmata firmware onto the CP Express’ expandable/collapsible? Then when the user clicks on ‘checklist’ above, the ‘Install the Firmata firmware onto the CP Express’ list could be collapsed so the user sees only one list - connecting your board.
> 
> Just wanted to add that this is a bit outside the scope of this task and could be logged as a follow-up task. Thanks!

Logged here!
https://codedotorg.atlassian.net/browse/LABS-509

apps/src/lib/kits/maker/ui/SetupInstructions.jsx
cnbrenci: done!

apps/src/lib/kits/maker/ui/setup-instructions.module.scss
fisher-alice: Nit: Could we add a space in between emoji and text?

fisher-alice (approved):
Looks great! 🎉 
Left just a small nit. This is such an improvement - thanks!

apps/src/lib/kits/maker/ui/setup-instructions.module.scss
cnbrenci: Realized that I hadn't pushed my latest commit that had this fix in it >.> This is fixed!

PR #56341, opened 2024-02-06
Don't LFS dashboard/app/assets/stylesheets

snickell:
This PR excludes from Git LFS:
```
dashboard/app/assets/stylesheets/** !text -filter -merge -diff
dashboard/app/assets/javascripts/** !text -filter -merge -diff
```

We previously included `dashboard/app/assets/**` in LFS, but that directory includes `.scss` files and a `.js.erb` file.

We could have just made the LFS inclusion line only specifically add the `./assets/images` subdir, but for longevity of keeping binary files out of LFS, its better to cast a wide net. Its MUCH easier to migrate files OUT of LFS than rewrite history (like we just did, very painful).

To make this change I:
1. Edited .gitattributes to add exclusion lines
2. Used `git add --renormalize dashboard/app/assets` to get the previously-LFS files added as part of the commit

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -3,6 +3,8 @@ dashboard/config/locales/** filter=lfs diff=lfs merge=lfs -text
 i18n/locales/** filter=lfs diff=lfs merge=lfs -text
 apps/static/** filter=lfs diff=lfs merge=lfs -text
 dashboard/app/assets/** filter=lfs diff=lfs merge=lfs -text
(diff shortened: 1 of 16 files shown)


snickell:
To verify the exclusion worked, I ran:
```
➜  code-dot-org git:(seth/de-lfs-application.scss) git lfs ls-files | grep application.scss
➜  code-dot-org git:(seth/de-lfs-application.scss) 
```

levadadenys (approved):
LGTM! Thanks a lot!

PR #55808, opened 2024-01-17
Adds AI TA line annotations.

wilkie:
This adds the ability to annotate lines in the editor with observations made from the AI agent.

When a learning goal is selected within the rubric assessment popup for teachers, any specially marked observations will highlight lines in the text editor. Such lines are 'annotated' such that an icon appears within the gutter to the left of the line itself. When one hovers over that icon, the text describing why that line is relevant, as provided by the AI, will appear as a popover similarly to the same process for warnings/errors in code.

To do this, several functions to parse out and call out to the annotation routines are found in the `LearningGoals` component. To aid in this, a utility class called `EditorAnnotator` was created to house functions to highlight and annotate lines and blocks within any editor context existing on the page via `StudioApp`. The line existing annotation routines used to mark errors and warnings in student code were also augmented to allow for an `'INFO'` type, which is a common type much like `ERROR` or `WARNING` provided by the Ace Editor. The default icon is a small lowercase `i`.

[Video found on Slack](https://codedotorg.slack.com/archives/C051P2V2RN0/p1704494617264189).

New things beyond the video are the functions that will 'dim' blocks which helps emphasize the highlighted "lines" when in the block rendering mode of Droplet. The `EditorAnnotator` can be powered by implementations of an abstract `Annotator` interface to provide functionality to query and affect blocks and lines and overall code in the editor context where those implementations differ based on the underlying editor technology (Droplet is implemented, here, but the alternative might be Blockly, etc).

image

image

In the future, we can augment this `EditorAnnotator` with separate implementations for Blockly and CodeMirror to serve SpriteLab and JavaLab, etc, at a later point. I've done [rough proofs of concept](https://codedotorg.slack.com/archives/C050HE4QDHR/p1706294593062199?thread_ts=1706216332.346339&cid=C050HE4QDHR) (Slack) showing how we could similarly affect Blockly environments to highlight blocks based on code snippets from, say, AI agents.

### Database Migration and Required System Prompt Changes

Observations require a change to the system prompts for the AI agent such that the AI reports particularly formatted text within the `observations` column. With this, a database migration adds the `observations` column to the `LearningGoalAiEvaluation` table to record the full text returned from the aiproxy service.

### Experiment Accessibility

This is only available with the 'carousel' mode for the rubric popup modal, so it is obscured and unused if that experiment isn't active.

## Links

- jira ticket: [AITT-416](https://codedotorg.atlassian.net/browse/AITT-416)

## Testing story

Only unit tests.

[...]

PR #56344, opened 2024-02-06
Adds AI TA line annotations

wilkie:
This adds the ability to annotate lines in the editor with observations made from the AI agent.

When a learning goal is selected within the rubric assessment popup for teachers, any specially marked observations will highlight lines in the text editor. Such lines are 'annotated' such that an icon appears within the gutter to the left of the line itself. When one hovers over that icon, the text describing why that line is relevant, as provided by the AI, will appear as a popover similarly to the same process for warnings/errors in code.

To do this, several functions to parse out and call out to the annotation routines are found in the `LearningGoals` component. To aid in this, a utility class called `EditorAnnotator` was created to house functions to highlight and annotate lines and blocks within any editor context existing on the page via `StudioApp`. The line existing annotation routines used to mark errors and warnings in student code were also augmented to allow for an `'INFO'` type, which is a common type much like `ERROR` or `WARNING` provided by the Ace Editor. The default icon is a small lowercase `i`.

[Video found on Slack](https://codedotorg.slack.com/archives/C051P2V2RN0/p1704494617264189).

New things beyond the video are the functions that will 'dim' blocks which helps emphasize the highlighted "lines" when in the block rendering mode of Droplet. The `EditorAnnotator` can be powered by implementations of an abstract `Annotator` interface to provide functionality to query and affect blocks and lines and overall code in the editor context where those implementations differ based on the underlying editor technology (Droplet is implemented, here, but the alternative might be Blockly, etc).

image

image

In the future, we can augment this `EditorAnnotator` with separate implementations for Blockly and CodeMirror to serve SpriteLab and JavaLab, etc, at a later point. I've done [rough proofs of concept](https://codedotorg.slack.com/archives/C050HE4QDHR/p1706294593062199?thread_ts=1706216332.346339&cid=C050HE4QDHR) (Slack) showing how we could similarly affect Blockly environments to highlight blocks based on code snippets from, say, AI agents.

### Database Migration and Required System Prompt Changes

Observations require a change to the system prompts for the AI agent such that the AI reports particularly formatted text within the `observations` column. With this, a database migration adds the `observations` column to the `LearningGoalAiEvaluation` table to record the full text returned from the aiproxy service.

### Experiment Accessibility

This is only available with the 'carousel' mode for the rubric popup modal, so it is obscured and unused if that experiment isn't active.

## Links

- jira ticket: [AITT-416](https://codedotorg.atlassian.net/browse/AITT-416)

## Testing story

Only unit tests.

Added some unit tests for older routines that are being used by this work.

## Follow-up work

The experiment for carousel work would have to be used to allow access to this to any pilot participant.

The system prompt for lesson 21 would be updated with the changes to add observations. This is the relevant change:

```
when choosing a grade for each Key Concept, please follow the following steps:
1. follow the instructions in the Instructions column from the rubric to generate observations about the student's program. Observations will always begin with a description of the number of lines in the student program. Use the form "There are 30 lines of code." where you replace 30 with the actual number of lines in the given student program. Observations will include actual lines of code from the student work. These observations are marked by the number of lines of code preceding the actual line of code within the student work. The code and line numbers will precede each observation in the form of this example: "Line 5: The sprite is defined here. `mySprite = createSprite(100, 100);`". Please include the actual code next to the line numbers in the Observations column in your response.
2. based on those observations, determine the highest grade which accurately describes the student's program. Write this result to the Grade column in your response.
3. write a reason for your grade in the Reason column, citing evidence from the Observations column when possible.
```

## PR Checklist:

- [ ] Code is well-commented


wilkie:
Previously: https://github.com/code-dot-org/code-dot-org/pull/55808

apps/src/templates/rubrics/LearningGoals.jsx
```diff
@@ -28,6 +29,106 @@ import ProgressRing from './ProgressRing';
...
+ *   velocity `p.velocityX = 0;`
+ *
+ * In this case, we expect to highlight lines 3 and 4 and to annotate each with
+ * the provided description. Although, we might find that this code actually
+ * exists on lines 5 and 6, since the AI is somewhat unreliable about line
+ * number reporting. The `findCodeRegion` function is responsible for finding
+ * the actual lines.
+ *
+ * When the AI gives us multiple lines, it will generally remove the whitespace.
+ * That it, the newlines are not within the observation text block. We have to
```
davidsbailey: nit: "That it"?

dashboard/db/migrate/20240104234208_add_observations_to_learning_goal_ai_evaluations.rb
```diff
@@ -0,0 +1,5 @@
+class AddObservationsToLearningGoalAiEvaluations < ActiveRecord::Migration[6.1]
+  def change
+    add_column :learning_goal_ai_evaluations, :observations, :text
+  end
+end
```
davidsbailey, line 5: our best practice would be to ship this migration in a separate PR ahead of this one

apps/src/EditorAnnotator.js
```diff
@@ -0,0 +1,961 @@
...
+
+  /**
+   * Gets a reference to the DropletBlock for the given line.
+   *
+   * If there is no block for that line it returns `null`.
+   *
+   * @param {number} lineNumber - The one-based line number.
+   * @returns {DropletBlock|null} A reference to the block or `null` if it doesn't exist.
+   */
+  getBlockForLine(lineNumber) {
```
davidsbailey, line 107: I like the abstract class approach! this is a bit philosophical, but for some of these it seems odd for the abstract class to define methods which (I assume) blockly will not be able to implement. I would expect this kind of method to be implemented only in the DropletAnnotator so that the future BlocklyAnnotator does not have to [refuse](https://luzkan.github.io/smells/refused-bequest) it.

if it won't break anything too badly, I'd suggest simply removing these from the abstract class. if you prefer the current pattern or if it will be a big project to change then I can live with it though :-) ok to think about in follow-up PR given time pressure.

apps/src/EditorAnnotator.js
```diff
@@ -0,0 +1,961 @@
...
+  EditorAnnotator.reset = function () {
+    EditorAnnotator.studioApp_ = undefined;
+    EditorAnnotator.annotator_ = undefined;
+    EditorAnnotator.code_ = undefined;
+    EditorAnnotator.lines_ = undefined;
+    EditorAnnotator.strippedCode_ = undefined;
+    EditorAnnotator.strippedLines_ = undefined;
+    EditorAnnotator.highlightedLines = [];
+  };
+}
```
davidsbailey, line 959: what is the mental model for why this should be inline, as opposed to in a method that you can call at a time of your choosing during testing? isn't this something that will need to be done more than once, between tests?

apps/src/EditorAnnotator.js
```diff
@@ -0,0 +1,961 @@
...
+
+    EditorAnnotator.highlightedLines.forEach(item => {
+      const block = this.getBlockForLine(item.lineNumber);
+      if (block) {
+        this.undimBlock(block);
+      }
+    });
+  }
+
+  patch() {
```
davidsbailey, line 206: 🤯 if we have anyone in the org who is familiar with droplet it will be great to get their review here. I know you've pulled out the code into a separate class, but since it is patching droplet it seems like the code is very much droplet territory.

I think you mentioned a convo with student learning, but just wondering since I didn't remember seeing any comments from them on the PR.

davidsbailey (approved):
Hey Wilkie, I'll be honest, this was a bear to review. I know the LFS migration caught us all by surprise, but reviewing the first rev of this PR was already quite and undertaking and it's been a challenge to sort out the new changes since that review from the old. Probably I have missed some things here due to the size. For next time, once a PR is this big, please no more adding big chunks of new functionality after review :-) 

That said I'm really happy with how well factored this is and the unit test coverage looks really good! I know this has been a big push, nice job getting this feature all the way from the idea stage to completion!

Please consider comments before merging.

apps/src/EditorAnnotator.js
wilkie: I did, and the splitting things up was in a way to ensure that ongoing work to encapsulate these libraries wasn't going to be an issue in the broad sense. It is worrying that the patching happens, of course. The other option is to permanently patch droplet.

I did not get a strong answer on this approach. I'll ping them in [this existing thread](https://codedotorg.slack.com/archives/C03DBDN67B7/p1706142470831659).

apps/src/EditorAnnotator.js
wilkie: Since it is singleton-like and its state is assumed to be set-and-forgot once, the tests needed a way to reset that state. That was so closely tied to implementation, that it seemed reasonable to write it near the implementation and expose the logic as a method to the test file. It could be implemented inside the test file, but I try to avoid leaking this level of detail into unit tests.

It is similar to the stubStudioApp/restoreStudioApp functions but I don't explicitly instantiate the singleton, so it's essentially just the restore half.

There was no way to conditionally define the function within the class definition that I know of. So it hangs out on its own in kind of a messy way.

apps/src/EditorAnnotator.js
wilkie: This a great thing for me to give a second look. It does feel like it is only useful for implementing the other actions... and, also, I could break down the abstract class further in terms of the "line-based" or "block-based" distinction. I'll give it a think. Straight-forward to move around at any point.

wilkie:
>Hey Wilkie, I'll be honest, this was a bear to review. I know the LFS migration caught us all by surprise, but reviewing the first rev of this PR was already quite and undertaking and it's been a challenge to sort out the new changes since that review from the old. Probably I have missed some things here due to the size. For next time, once a PR is this big, please no more adding big chunks of new functionality after review :-)

@davidsbailey I'm so sorry. I really do not know what I could have done differently. Probably broken the PR up? I felt some pressure.

Having the rolling review did feel like it offered the chance to get it in reasonable shape and I was able to respond to the comments in the first review. I got the pushback from altering the legacy class and the push to abstract the implementations which led to a more complicated result. Of course, if I would have known that the implementation of this would be written in a different way, I would not have wasted your time with that initial review and would have taken on the delay risk as much on my own as I could.

If you want a diff of the differences between this and the last PR, you can diff that commit with the PR branch. I'm happy to help direct your review, and I suppose should have done so. The things got lost when the PR was reopened so it was hard to direct you to changes based on feedback.

For CLI reference to the relative diff: `git diff f1b9d9c aitt-line-annotations` 

For a visual reference: https://github.com/code-dot-org/code-dot-org/compare/f1b9d9c...aitt-line-annotations for the changes between your last review and this one.

Between these specific commits: https://github.com/code-dot-org/code-dot-org/compare/f1b9d9c...32b48493a336c0056a82c454c88f6a35ceff484c are testing asks you made about testing for whitespace and for testing that the default line ranges are respected when code cannot be found via the line detection.

I would say the rest: https://github.com/code-dot-org/code-dot-org/compare/32b48493a336c0056a82c454c88f6a35ceff484c...aitt-line-annotations is moving the annotations to their own class.

Specifically this commit: 81be3fa7c11e7ea28c8f28f1fdb35065d7020b4b adds the patching for dimming blocks and is where much of that is thrown in. I had asked to drop it and maybe I should have just done so to reduce the complexity and how much I'm asking for.

I very much appreciate the effort you have gone through to make this a better result. It is very anxiety-inducing to touch things which feel both important and load-bearing and fragile at the same time.

davidsbailey:
Thanks for the detailed explanation Wilkie. I think I missed that the big changes were requested by the student learning team, since I did not see those come through in code review. given those new constraints, I agree that it would have been hard to break into smaller pieces. Without knowing that, it seemed like adding more new work onto an otherwise mergeable PR. 

I did find some of those commit ranges, after trying (and mostly failing) to go back and find useful views of the first part of my review in the original PR (thanks to LFS).

Overall I feel good about the code quality here and the amount of review it has gotten, especially hearing that you got substantive feedback on the student learning side. This really is nicely factored and tested!

wilkie:
Phew, I hear you though-- Navigating the new PR was complicated. I did not realize that the old PR would become so corrupted and unusable or I think I would have done something differently. Not sure what. Should have linked the conversation with SL, though I did not get much direct feedback so I do feel like I was / still am winging it a bit. Thanks for your feedback, as always.

wilkie:
Flakey tests plaguing me:

This one I've not seen a mention of before:

```
2028 | test_teacher_based_experiment_at_50_percent_is_enabled_for_only_some_users#ExperimentTest (1291.98s)
2029 | Minitest::UnexpectedError:         ActiveRecord::RecordNotUnique: Mysql2::Error: Duplicate entry '1028' for key 'PRIMARY'
2030 | config/initializers/mysql_check_index_used.rb:33:in `block (3 levels) in execute'
2031 | config/initializers/mysql_check_index_used.rb:32:in `block (2 levels) in execute'
2032 | config/initializers/mysql_check_index_used.rb:31:in `block in execute'
2033 | config/initializers/mysql_check_index_used.rb:30:in `execute'
2034 | test/models/experiment_test.rb:68:in `block in <class:ExperimentTest>'
2035 | test/testing/setup_all_and_teardown_all.rb:36:in `run'
```

and then the more common one seen about once a month or so even on our testing server:

```
2070 | test_anonymizes_and_deletes_code_review_comments_written_by_user#DeleteAccountsHelperTest (931.65s)
2071 | Minitest::UnexpectedError:         ActiveRecord::RecordInvalid: Validation failed: User already has an open code review for this project
2072 | test/helpers/delete_accounts_helper_test.rb:873:in `block in <class:DeleteAccountsHelperTest>'
2073 | test/helpers/delete_accounts_helper_test.rb:31:in `block (2 levels) in run'
2074 | test/helpers/delete_accounts_helper_test.rb:30:in `block in run'
2075 | test/helpers/delete_accounts_helper_test.rb:29:in `run'
2076 | test/testing/setup_all_and_teardown_all.rb:36:in `run'
```

wilkie:
Drone seems unsafe to restart at the moment with the last few drone runs seeing clone issues due to the bandwidth cap for our repo's lfs data. I might merge anyway?? I got a passing test run before I modified a comment. https://drone.cdn-code.org/code-dot-org/code-dot-org/39636 (which shows the clone it did: `HEAD is now at 3ef148e6e3 Fixes show code toggling.`)

davidsbailey:
> Flakey tests plaguing me:

Thanks for flagging these!
 
> This one I've not seen a mention of before:
> 
> ```
> 2028 | test_teacher_based_experiment_at_50_percent_is_enabled_for_only_some_users#ExperimentTest (1291.98s)
> 2029 | Minitest::UnexpectedError:         ActiveRecord::RecordNotUnique: Mysql2::Error: Duplicate entry '1028' for key 'PRIMARY'
> 2030 | config/initializers/mysql_check_index_used.rb:33:in `block (3 levels) in execute'
> 2031 | config/initializers/mysql_check_index_used.rb:32:in `block (2 levels) in execute'
> 2032 | config/initializers/mysql_check_index_used.rb:31:in `block in execute'
> 2033 | config/initializers/mysql_check_index_used.rb:30:in `execute'
> 2034 | test/models/experiment_test.rb:68:in `block in <class:ExperimentTest>'
> 2035 | test/testing/setup_all_and_teardown_all.rb:36:in `run'
> ```

Seems like this test is hard-coding an id into the users table when it creates users, which is a no-no: https://github.com/code-dot-org/code-dot-org/blob/b1f8f0949fb08b9ba65baf7ac09d247352fa1e94/dashboard/test/models/experiment_test.rb#L29-L32

it looks like the methods being tested will call user.id to view the id (as opposed to requerying from the database), so this test could be implemented with simple stubbing of the user.id method instead.

Issue #56347, opened 2024-02-06
How do we update the view after doing a dataset browser action?


snickell:
See https://github.com/code-dot-org/code-dot-org/pull/56349 for a PR in this direction

The problem is that the way Firebase was coupled to the redux store was via subscription methods in applab.js. These are triggered by changes to the redux store (e.g. what data view is currently displayed), but flow through applab.js.

We don't have a subscription system like Firebase, instead we need to do a refresh of the data to load the current view in cases where, say, "Add Table" was pressed in DataTable.jsx. Basically any of the storageBackend().THINGHERE methods executed from `apps/src/storage/dataBrowser/*.jsx`.

snickell:
Relevant snippets from the codebase:
<img width="525" alt="image" src="https://github.com/code-dot-org/code-dot-org/assets/223277/88588f60-a7af-485b-92fe-e00f0113cb64">

snickell:
Fixed by #56349

PR #56349, opened 2024-02-06
Refresh DataViews after each data modifying action

snickell:
This is related to, and intended to fix https://github.com/code-dot-org/code-dot-org/issues/56347

### DIFF
--- a/apps/src/applab/applab.js
+++ b/apps/src/applab/applab.js
@@ -883,7 +883,6 @@ function setupReduxSubscribers(store) {
       (isDataMode && view !== lastView) ||
       changedToDataMode(state, lastState)
     ) {
-      console.log("calling onDataViewChanged() (now called refreshDataView())");
       loadDataForView(
         Applab.storage,
         state.data.view,
--- a/apps/src/applab/applab.js
+++ b/apps/src/applab/applab.js
@@ -52,7 +52,7 @@ import {
   setLibraryManifest,
   updateTableList,
 } from '../storage/redux/data';
-import {refreshDataView} from './refresh-data-view';
+import {loadDataForView} from './load-data-for-view';
 import {setStepSpeed} from '../redux/runState';
 import {
   getContainedLevelResultInfo,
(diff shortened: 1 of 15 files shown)


snickell:
@cnbrenci and I traced how calls flow when you change the DataView to the Key Value Pairs tab by clicking it:

1. In the browser, we click Key Value tab
2. The onClick handler runs: `this.props.onViewChange(DataView.PROPERTIES)`
3. onViewChange() is passed in from DataBrowser.jsx, which runs `dispatch(changeView(DataView.PROPERTIES, null))`
4. On dispatch, redux store runs `case CHANGE_VIEW` which runs state.set('view', DataView.PROPERTIES)
5. in applab, setupReduxSubscribers is subscribed to store.subscribe(), so this hook gets called because state changed
6. the setupReduxSubscribes subscription then inspects state.view and calls onDataViewChange(DataView.PROPERTIES, null, null)
7. onDataViewChange then invokes `Applab.storage.subscribeToKeyValuePairs()` to load relevant data, and sets callbacks when the data is loaded
8. The callbacks from Applab.storage.subscribeToKeyValuePairs() then calls redux to update the loaded kvp data: `getStore().dispatch(updateKeyValueData(keyValueData))`

snickell:
So lets focus on the Key Value Pair tab's "Add Pair button" 
1. Add Pair Buttons `onClick` handler in AddKeyRow.jsx calls `storageBackend().setKeyValue(newKey, newVal)`

IN FIREBASE MODE:
- `loadDataForView()` has setup a callback `storage.subscribeToKeyValuePairs()`, which sets up a Firebase Subscription to notify when any key value data changes in Firebase
- Firebase notices that a key value has changed, and calls the callback, which invokes a redux change: `getStore().dispatch(updateKeyValueData(keyValueData))`
- Which invokes `state.set('keyValueData')`
- This triggers redux's connect() function changing the keyValueData prop passed to KVPairs(), which then redraws the rows with the new Key Value Pair row.

IN DATABLOCK MODE:
TBD

snickell:
TODO next time: test implementation of clicking "Add pairs" button in Key Value Pair tab

cnbrenci (approved):
Yaaaaaaaaaaas!!

PR #56351, opened 2024-02-06
create pathway for updating rubric ai config

davidsbailey:
## background

Finishes https://codedotorg.atlassian.net/browse/AITT-446. 

Currently, we store our production ai config for rubric evaluation in a fixed location in S3. There is no safe way to modify it without directly updating config files which are being used to serve live traffic in our ai-rubrics pilot. Furthermore the lesson names are error-prone with a confusing mix of formats and 2022/2023 lesson numbers.

## description

done in this PR:
* added `s3://cdo-ai/teaching_assistant/releases/` directory to hold individual releases, each of which contains config for all lessons which ai can evaluate
* extract a ruby constant to hold the location of the current release, which can be updated to launch a new release
* standardized lesson names to `<unitname>-L<lesson-number>` format, e .g. `csd3-2023-L11` 
* added missing ai config for L24 and L28 to `interactive-games-animations-2023` unit
* copied the contents of `s3://cdo-ai/teaching_assistant/lessons/` into a first release, updating lesson names accordingly:

```
Dave-MBP:~$ aws s3 ls s3://cdo-ai/teaching_assistant/releases/2024-02-01-ai-rubrics-pilot-baseline/
                           PRE allthethings-L48/
                           PRE csd3-2023-L11/
                           PRE csd3-2023-L14/
                           PRE csd3-2023-L18/
                           PRE csd3-2023-L21/
                           PRE csd3-2023-L24/
                           PRE csd3-2023-L28/
Dave-MBP:~$ aws s3 ls --recursive s3://cdo-ai/teaching_assistant/releases/2024-02-01-ai-rubrics-pilot-baseline/csd3-2023-L11/
2024-02-06 04:46:38        174 teaching_assistant/releases/2024-02-01-ai-rubrics-pilot-baseline/csd3-2023-L11/confidence.json
2024-02-06 04:46:38        131 teaching_assistant/releases/2024-02-01-ai-rubrics-pilot-baseline/csd3-2023-L11/params.json
2024-02-06 04:46:38       1409 teaching_assistant/releases/2024-02-01-ai-rubrics-pilot-baseline/csd3-2023-L11/standard_rubric.csv
2024-02-06 04:46:38       3101 teaching_assistant/releases/2024-02-01-ai-rubrics-pilot-baseline/csd3-2023-L11/system_prompt.txt
```

shipping this PR shouldn't change ai behavior for our pilot users (besides enabling new lessons in interactive-games-animations-2023), it just sets us up to make ai changes in the future.

## Testing story

* added a UI test which validates that the current releases directory in S3 contains valid ai config for every lesson in the `UNIT_AND_LEVEL_TO_LESSON_S3_NAME` map. If you try to launch a new release and forget an entire lesson or a specific config file, you get a helpful error message like this:
Screenshot 2024-02-06 at 6 49 51 AM

* existing UI test coverage also validates that ai config for the allthethings lesson can be read from S3, and the key concepts there are correct.
* existing unit test coverage

## Future work

- [x] move lesson config (including dev/test datasets) out of gdrive and into S3 (Ceara)
* for more ideas, see: https://docs.google.com/document/d/1HxPgdoht5mzYSr8_HB0KTISt5rGS_648hVfD2ye2Jhs/edit


cat5inthecradle:
Infra would like to move towards per-environment S3 buckets. This bundles all resources for a given application environment into a single cloudformation stack, streamlining permissions and making cleanup & maintenance easier.

It looks like what you're doing here might not benefit from environment segregation, and all environments are referencing the same configuration. Is that right? This appears to be a case where S3 is being used for content/configuration, and while you might want to point an environment at a specific _version/release_ of the data, you're not really separating by _environment_. This is relevant to the implementation of AITT-345, and it probably makes the most sense to create this bucket once per AWS Account.

davidsbailey:
> Infra would like to move towards per-environment S3 buckets. This bundles all resources for a given application environment into a single cloudformation stack, streamlining permissions and making cleanup & maintenance easier.

What would this look like in this case?
1. `s3://cdo-ai-production/teaching_assistant/`, `s3://cdo-ai-development/teaching_assistant/`, etc, or
2. `s3://cdo-content-production/ai/teaching_assistant`, `s3://cdo-content-development/ai/teaching_assistant`
or something else?

> It looks like what you're doing here might not benefit from environment segregation, and all environments are referencing the same configuration. Is that right? This appears to be a case where S3 is being used for content/configuration, and while you might want to point an environment at a specific _version/release_ of the data, you're not really separating by _environment_. This is relevant to the implementation of AITT-345, and it probably makes the most sense to create this bucket once per AWS Account.

so the plan here looks something like:
* today, the development, test and production environments of dashboard need read access to this data
* tomorrow, rubric tester would like to read access this data too, to make sure we can run accuracy tests against whatever config is actually live in production
* in April/May, we'll want to move the pointers to this data so that it is only read by aiproxy, not by dashboard. all environments of aiproxy will want to access this data.

there are no future plans for aiproxy or dashboard services to need to write this data. that will be done manually, then validated with automated tools before the new release goes live.

davidsbailey:
> This is relevant to the implementation of AITT-345, and it probably makes the most sense to create this bucket once per AWS Account.

AITT-345 will access a different set of data for conducting ai config "experiments", and in that scenario we do want to be able to read/write experiment data, but only from development. perhaps we should use a different bucket for that, so that our development process cannot accidentally mess up our production release data?

what do you mean "once per AWS Account" -- once for codedotorg, and once for codedotorg-dev?

dashboard/app/jobs/evaluate_rubric_job.rb
```diff
@@ -5,30 +5,32 @@ class EvaluateRubricJob < ApplicationJob
...
   STUB_AI_PROXY_PATH = '/api/test/ai_proxy'.freeze
 
-  # 2D Map from unit name and level name, to the name of the lesson files in S3
-  # which will be used for AI evaluation.
-  # TODO: This is a temporary solution. After the pilot, we should at least make
-  # the S3 pointer editable on levelbuilder, and eventually make all of the data
-  # it points to editable there too.
+  # 2D Map from unit name and level name, to the name of the lesson files within
+  # the release dir in S3 which will be used for AI evaluation.
   UNIT_AND_LEVEL_TO_LESSON_S3_NAME = {
```
wilkie, line 20: what about potentially moving this to S3? or does that make it too hard to test or experiment locally? maybe a hybrid.

dashboard/test/ui/features/step_definitions/steps.rb
```diff
@@ -1466,3 +1466,11 @@ def refute_bad_gateway_or_site_unreachable
   expect(params['name']).to eq(name)
   expect(params['course']).to eq(course)
 end
+
+And(/^I validate rubric ai config for all lessons$/) do
```
wilkie, line 1470: would this UI test be possible on a development environment without AWS access and/or contributor access?

wilkie: I meant to ask this on the test and not the step 🥲

dashboard/app/jobs/evaluate_rubric_job.rb
davidsbailey: this is an interesting idea and I'm not sure where the best place for it is. one observation is that if we move this config to S3, then we lose some of the test coverage this PR adds which would ensure that any new lesson dirs added to this map have all the necessary files in the releases dir. so my inclination is there might need to be a compelling reason to move it there to overcome that?

for some context, where I think we want this to end up is in https://github.com/code-dot-org/aiproxy/blob/main/lib/assessment/config.py . then dashboard would just tell aiproxy what lesson to evaluate, and any updates to the latest release pointer could more easily be covered by the accuracy regression testing functionality which Ceara has added to the aiproxy repo. so, separately I am curious how that part of the plan sounds?

dashboard/test/ui/features/step_definitions/steps.rb
davidsbailey: no, this test would fail if you did not have S3 access. I don't think developers are often running all UI tests locally, but I could see how this could be a regression for a contributor working on ai rubric features.

other UI tests depend on S3 though (most labs features I think) so I'm not sure this is a big outlier in this way though.

per other comment, I hope to move this config into aiproxy, but i agree that i shouldn't leave things in a bad state in the interim.

davidsbailey: perhaps UI tests should be flagged with `@s3_access`, so we can intelligently skip them (or give helpful warning) when a developer without s3 access tries to run them? we do something similar for `@dashboard_db_access`, which has to do with whether you are running the runner.rb on the same machine as the rails server. so the implementation could be similar.

cat5inthecradle:
> what do you mean "once per AWS Account" -- once for codedotorg, and once for codedotorg-dev?

Yes, that's what I meant, though one globally would be fine too. Cross-account permissions are annoying, but solvable. We also don't have a strong need to stand this up in a different account right now.

dashboard/test/ui/features/step_definitions/steps.rb
wilkie: That feels like a nice thing to have! I try to run UI tests locally :) I'm part of the eng-excellence team for remove AWS deps so I'm looking for best practices moving forward. Things like skipping AWS dependent tests when we run them locally / don't have a valid key would be nice to have. Could we flag them even if we don't yet implement that check at least?

davidsbailey: to clarify, I agree that most engineers run individual UI tests locally for the features they are working on. However, it seems like it would be uncommon for a contributor to work on a feature which does not require S3 access and yet still need to run UI tests which require S3 access. I won't fight you on that overall eng initiative though. I'm happy to store the "releases" somewhere else, but not sure where that might be?

> Things like skipping AWS dependent tests when we run them locally / don't have a valid key would be nice to have. Could we flag them even if we don't yet implement that check at least?

we could start speculatively adding a flag to ui tests that use S3, but without anything checking this, it might be hard to keep this list in sync. I count 25 ui test files containing the string `studio.code.org/projects/`, and that's just one indicator of S3 access. Perhaps a more definitive way to tag them would be to take away S3 access and see which ones fail.

wilkie: I'd say that this shouldn't block merging this... there are already dozens that would be similarly broken for non-AWS access. I generally develop without AWS access, myself, so I do run into these from time to time, but I'm going to advocate we fix that separately!

wilkie (approved):
LGTM! 💯 thanks!

davidsbailey:
fast tracking this after making sure the following are passing:
* `Dave-MBP:~/src/cdo/dashboard (ta-ai-config-releases)$ bundle exec spring testunit ./test/jobs/evaluate_rubric_job_test.rb`
* `[development] dashboard > EvaluateRubricJob.new.validate_ai_config`

PR #56352, opened 2024-02-06
Fix negative accepted applications and show enrolled in status column

bethanyaconnor:
This PR ended up doing a few things:
- fixed a bug where accepted counts were negative. I used test driven development here by writing `enrollment only counts enrollments for that regional partner`, which failed before this change and passes after.
- Shows "Enrolled" on the quick and cohort views of the application dashboard (see screenshot below)
- Removes "Enrolled" from the dropdown of assignable statuses

The top two items share a solution: a new function, `status_including_enrolled` that is used instead of `status` for counting and display. This greatly simplified the logic for counting applications in each status (paired with my newfound favorite function `tally`).

Screenshot 2024-02-06 at 8 34 06 AM

## Links

[slack thread 1](https://codedotorg.slack.com/archives/C03SUS476R3/p1707168183512349)
[slack thread 2](https://codedotorg.slack.com/archives/C03SUS476R3/p1707168376839509)


dashboard/app/controllers/api/v1/pd/applications_controller.rb
```diff
@@ -22,21 +22,17 @@ def index
...
-          apps_statuses = apps_statuses.where(regional_partner_id: nil)
+          apps = apps.where(regional_partner_id: nil)
         elsif regional_partner_value && regional_partner_value != REGIONAL_PARTNERS_ALL
-          apps_statuses = apps_statuses.where(regional_partner_id: regional_partner_value)
+          apps = apps.where(regional_partner_id: regional_partner_value)
         end
 
-        apps_statuses.group(:status).each do |group|
-          application_data[role][group.status] = {total: group.total}
+        #apps_statuses = apps.select(:status, "count(id) AS total").group(:status)
```
TurnerRiley: Can this commented line be removed or is this being saved for something?

dashboard/app/controllers/api/v1/pd/applications_controller.rb
```diff
@@ -22,21 +22,17 @@ def index
...
+        apps_statuses = apps.map(&:status_including_enrolled).tally
```
TurnerRiley: In case `apps` returns an empty result from the above `where` clauses, should there be safe chaining for `map` and `tally` here or do those already protect against it?

TurnerRiley (approved):
Looks great! Just had a couple quick questions

dashboard/app/controllers/api/v1/pd/applications_controller.rb
bethanyaconnor: I think it would return an empty array so safe chaining isn't necessary. But, as I'm already going to update this PR to remove the above line, I'll add safe chaining just to be safe!

dashboard/app/controllers/api/v1/pd/applications_controller.rb
bethanyaconnor: 🤦‍♀️

PR #56218, opened 2024-02-02
AI Tutor: convert aiTutorFloatingActionButton and aiTutorPanel to TypeScript

Erin007:
[CT-302](https://codedotorg.atlassian.net/browse/CT-302)

Code Tools is working on converting our components (at least the new ones) to TypeScript based on this [proposal](https://docs.google.com/document/d/15KLtvSTHFOGIVwu60l1mFZscFL8miXvZHKNCKwuzf4E/edit#heading=h.a1916db8qd8j).  To support this initiative, as a mini-project for Meetings Week, the AI Tutor team is working on converting the files in `apps/src/code-studio/components/aiTutor` to TypeScript. 

This PR converts `aiTutorFloatingActionButton` and `aiTutorPanel`  to TypeScript.

- [ ] Tests provide adequate coverage
- [ ] Privacy and Security impacts have been assessed
- [ ] Code is well-commented
- [ ] New features are translatable or updates will not break translations
- [ ] Relevant documentation has been added or updated
- [ ] User impact is well-understood and desirable
- [ ] Pull Request is labeled appropriately
- [ ] Follow-up work items (including potential tech debt) are tracked and linked

PR #56353, opened 2024-02-06
AI Tutor: convert aiTutorFloatingActionButton and aiTutorPanel to TypeScript

Erin007:
[CT-302](https://codedotorg.atlassian.net/browse/CT-302)

Code Tools is working on converting our components (at least the new ones) to TypeScript based on this [proposal](https://docs.google.com/document/d/15KLtvSTHFOGIVwu60l1mFZscFL8miXvZHKNCKwuzf4E/edit#heading=h.a1916db8qd8j).  To support this initiative, as a mini-project for Meetings Week, the AI Tutor team is working on converting the files in `apps/src/code-studio/components/aiTutor` to TypeScript. 

This PR converts `aiTutorFloatingActionButton` and `aiTutorPanel`  to TypeScript.

- [ ] Code is well-commented


Erin007:
This PR is a continuation of #56218, which was mis-closed by the Git LFS migration (#55759).

### Previous Comments:


### Previous Reviews:
- **Erin007 commented** - 
- **sanchitmalhotra126 commented** -

apps/src/sites/studio/pages/levels/show.js
```diff
@@ -59,12 +60,18 @@ function initPage() {
 
   if (hasScriptData('script[data-aitutorleveldata]')) {
     const aiTutorLevelData = getScriptData('aitutorleveldata');
+    const {id, type, hasValidation} = aiTutorLevelData;
+    const level = {
+      id: id,
+      type: type,
+      hasValidation: hasValidation,
+    };
```
Erin007, line 69: ❓ Is there a way to call setLevel with the aitutorleveldata in levels/show.js rather than passing it down to all of the children and grandchildren components? Basically something like [this](https://github.com/code-dot-org/code-dot-org/blob/75443933100d3d3e7bd45da546588379c7d18cc7/apps/src/sites/studio/pages/levels/show.js#L84) but I wasn't sure what syntax to use with TypeScript and the way aiTutorRedux is set up.

Member
@[sanchitmalhotra126](https://github.com/sanchitmalhotra126) sanchitmalhotra126 [4 days ago](https://github.com/code-dot-org/code-dot-org/pull/56218#discussion_r1476728389)
I think we generally have a preference to avoid using redux code outside of React files (even though we have plenty of exceptions in older code 😅). I'm not as familiar with this feature, but where is the level information originating from? Is this tool meant to be used with any lab, or just Java Lab?

Erin007: @sanchitmalhotra126 

The level information is coming in from the view at [levels/show.haml.html](https://github.com/code-dot-org/code-dot-org/blob/344ff55a8ecc7decf5ab0c4151d04c0eff067969/dashboard/app/views/levels/show.html.haml#L19) and passed onto to show.js 

The short term answer is the tool is meant to used in the context of CSA levels. The eventual potential long term answer is that the tool is meant to be used much more broadly across labs, level types and maybe even outside of levels. 

I don't want to use an old/anti-pattern was just curious if there's a current preference to find a way to put data shared by multiple components into the store and accessing it from there or if it's ok, if a tiny bit cumbersome, to pass it around as needed. 🤷‍♀️

molly-moen: I think either way is fine, for lab2 levels we load the level data client side and immediately store in redux, but when it comes from the server the pattern is a little messier. This file is already adding things to redux, IMO it would be fine to set the level data here too. See line 32 in this file:
`getStore().dispatch(setCodeReviewEnabledForLevel(codeReviewEnabled));` This syntax works with either a js or ts redux store.

sanchitmalhotra126: Another option could be to use React [context](https://react.dev/learn/passing-data-deeply-with-context) to avoid passing props through multiple layers of components, but to Molly's point, looks like we're already passing server data into redux here, so that works too.

sanchitmalhotra126 (approved):
TypeScript changes LGTM!

apps/src/code-studio/components/aiTutor/aiTutorFloatingActionButton.tsx
```diff
@@ -1,28 +1,32 @@
...
 /**
  * Renders an AI Bot icon button in the bottom left corner over other UI elements that controls
  * toggling the AI Tutor Panel open and closed.
  */
-const AITutorFloatingActionButton = ({level}) => {
+
+const AITutorFloatingActionButton: React.FunctionComponent = () => {
   const store = getStore();
   const [isOpen, setIsOpen] = useState(false);
+  const level = useSelector(
```
molly-moen, line 19: (can be a follow up) we now have [useAppSelector](https://github.com/code-dot-org/code-dot-org/blob/7788d30a9e067ee634e17002998ca4da9d866557/apps/src/util/reduxHooks.ts#L14) which makes this syntax much nicer!

molly-moen (approved):
Nice!

PR #56096, opened 2024-01-30
Support pagination for LTI NRPS calls

carl-codeorg:
This adds handling for large courses, where the full roster can't be returned in a single NRPS response. [LTI specifies](https://www.imsglobal.org/spec/lti-nrps/v2p0/#limit-query-parameter) that the tool can request a certain number of users to be returned, and the platform can optionally respect the limit parameter. If another page of users is present, the tool must include a link header with a url for the next page.

This PR adds

- Adds the limit param to our NRPS calls, defaulting to 100 to ensure we don't get a huge response if a platform doesn't have a reasonable limit implemented. Canvas maxes at 50, so we won't get more than 50 back for now.
- Checks for a next page link and calls it. Keeps calling for next pages until no next page link is included or we hit the max course size in the LTI policy.

## Links

[JIRA](https://codedotorg.atlassian.net/browse/P20-602)

## Testing story

Since uploading a large number of students to Canvas and force-enrolling them in a class requires uploading an SIS flatfile, I found that the easiest way to test locally was to lower the default page limit in `get_context_membership` to a very low number, like 2. Then, even with a relatively small course, you can get multiple pages. If you put some `puts` calls in, you can verify that you're making multiple calls with query params like `page=2`, `page=3`, etc.

Unit Tests:
```
bundle exec spring testunit test/lib/clients/lti_advantage_client_test.rb
Running via Spring preloader in process 42072
Started with run options --seed 13043

  3/3: [====================================================] 100% Time: 00:00:08
```

## PR Checklist:

- [ ] Tests provide adequate coverage
- [ ] Privacy and Security impacts have been assessed
- [ ] Code is well-commented
- [ ] New features are translatable or updates will not break translations
- [ ] Relevant documentation has been added or updated
- [ ] User impact is well-understood and desirable
- [ ] Pull Request is labeled appropriately
- [ ] Follow-up work items (including potential tech debt) are tracked and linked

PR #56355, opened 2024-02-06
Support pagination for LTI NRPS calls

carl-codeorg:
This adds handling for large courses, where the full roster can't be returned in a single NRPS response. [LTI specifies](https://www.imsglobal.org/spec/lti-nrps/v2p0/#limit-query-parameter) that the tool can request a certain number of users to be returned, and the platform can optionally respect the limit parameter. If another page of users is present, the tool must include a link header with a url for the next page.

This PR adds

- Adds the limit param to our NRPS calls, defaulting to 100 to ensure we don't get a huge response if a platform doesn't have a reasonable limit implemented. Canvas maxes at 50, so we won't get more than 50 back for now.
- Checks for a next page link and calls it. Keeps calling for next pages until no next page link is included or we hit the max course size in the LTI policy.

## Links

[JIRA](https://codedotorg.atlassian.net/browse/P20-602)

## Testing story

Since uploading a large number of students to Canvas and force-enrolling them in a class requires uploading an SIS flatfile, I found that the easiest way to test locally was to lower the default page limit in `get_context_membership` to a very low number, like 2. Then, even with a relatively small course, you can get multiple pages. If you put some `puts` calls in, you can verify that you're making multiple calls with query params like `page=2`, `page=3`, etc.

Unit Tests:
```
bundle exec spring testunit test/lib/clients/lti_advantage_client_test.rb
Running via Spring preloader in process 42072
Started with run options --seed 13043

  3/3: [====================================================] 100% Time: 00:00:08
```

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/dashboard/lib/clients/lti_advantage_client.rb
+++ b/dashboard/lib/clients/lti_advantage_client.rb
@@ -11,7 +11,7 @@ class LtiAdvantageClient
   # The URL for the API will vary between LMS platforms, and is initially contained in the launch context.
   # See the LTI spec for details:
   # https://www.imsglobal.org/spec/lti-nrps/v2p0/
-  def get_context_membership(url, resource_link_id)
+  def get_context_membership(url, resource_link_id, page_limit = 100)
     options = {
       headers: {
         'Accept' => Policies::Lti::MEMBERSHIP_CONTAINER_CONTENT_TYPE,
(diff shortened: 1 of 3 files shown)


carl-codeorg:
This PR is a continuation of #56096, which was mis-closed by the Git LFS migration (#55759).

### Previous Comments:


### Previous Reviews:
- **carl-codeorg commented** - 
- **bdmesh commented** - 
- **carl-codeorg commented** - 
- **nicklathe approved** -

nicklathe (approved):

PR #56357, opened 2024-02-06
Legend responds to levels viewed

kobryan0619:
Modifies which icons appear depending on what lessons and levels are open.  The video below shows this change working.  

https://github.com/code-dot-org/code-dot-org/assets/24883357/02e726e6-024c-476a-a952-b1ebf0348dc4

[The original figma](https://www.figma.com/file/gCktlnafGYQSNJafD3FUgA/Progress-Tracking?node-id=2173%3A33166&mode=dev) shows what it should look like in these different states.

image

image

image

## Links

[Ticket.](https://codedotorg.atlassian.net/browse/TEACH-800)

## Testing story

I tested it locally.  I have another ticket to write up tests and other pieces of the icon key feature. 

## Follow-up work

Future work will address the spacing between the boxes in the icon key. 

## PR Checklist:

- [ ] Code is well-commented


apps/src/templates/sectionProgressV2/SectionProgressV2.jsx
```diff
@@ -28,10 +30,26 @@ function SectionProgressV2({
     }
   }, [unitData, isLoadingProgress, isRefreshingProgress, scriptId, sectionId]);
 
+  React.useEffect(() => {
```
kobryan0619: Note: At one point I wanted to move all this logic to the `IconKey` component but given that would require passing all the `unitData` to the child component, putting the logic here felt like a better choice.

apps/src/templates/sectionProgressV2/IconKey.jsx
```diff
@@ -8,8 +8,13 @@ import styles from './progress-table-legend.module.scss';
...
 import FontAwesome from '../FontAwesome';
 
-export default function IconKey({isViewingLevelProgress, hasLevelValidation}) {
+export default function IconKey({isViewingValidatedLevel, expandedLessonIds}) {
   const [isOpen, setIsOpen] = useState(false);
+  const [isViewingLevelProgress, setIsViewingLevelProgress] = useState(false);
+
+  React.useEffect(() => {
+    setIsViewingLevelProgress(expandedLessonIds.length > 0);
+  }, [expandedLessonIds]);
```
etaderhold: I don't think this case calls for a `useEffect`. There's no side effect here, it's just a very simple bit of logic to determine rendering within this component.

```suggestion
  const isViewingLevelProgress = expandedLessonIds.length > 0;
```

apps/src/templates/sectionProgressV2/SectionProgressV2.jsx
```diff
@@ -28,10 +30,26 @@ function SectionProgressV2({
...
+      const hasValidatedLevel = unitData.lessons.some(
+        lesson =>
+          lesson.levels.some(level => level.isValidated) &&
+          expandedLessonIds.includes(lesson.id)
+      );
+      setIsViewingValidatedLevel(hasValidatedLevel);
+    } else {
+      setIsViewingValidatedLevel(false);
+    }
+  }, [expandedLessonIds, unitData]);
```
etaderhold: Nice job using functional array methods here! I think `useMemo` is more appropriate for this case. Again, we're not causing any side effects such as calling an API, we're saving the results of a relatively expensive calculation. That's exactly what `useMemo` is for.
```suggestion
  const isViewingValidatedLevel = React.useMemo(() => {
    return unitData.lessons
      .filter(lesson => expandedLessonIds.includes(lesson.id))
      .some(lesson => lesson.levels.some(level => level.isValidated));
  }, [expandedLessonIds, unitData]);
```

kobryan0619: Thanks - this makes sense to me after reading the `useMemo` documentation.

etaderhold (approved):

PR #56358, opened 2024-02-06
P20-688: Fix restoration of Dashboard/course_content

artem-vavilov:
Since we skip updating `source` files with changed dirs, we shouldn't redact these files either
to avoid writing already redacted i18n data to the `original` files that we use for restoration.

https://github.com/code-dot-org/code-dot-org/blob/a0d62a0fecc9cb3f3914047f450cdb87e5c055d0/bin/i18n/resources/dashboard/course_content/sync_in.rb#L343-L345

https://github.com/code-dot-org/code-dot-org/blob/a0d62a0fecc9cb3f3914047f450cdb87e5c055d0/bin/i18n/resources/dashboard/course_content/sync_in.rb#L402-L404

- jira ticket: [P20-688](https://codedotorg.atlassian.net/browse/P20-688)

`Note:` This fix requires manual updating of the corrupted `original` files on the `i18n server`:
- /i18n/locales/original/course_content/2023/dance-ai-2023.json
- /i18n/locales/original/course_content/other/K5-OnlinePD.json
- /i18n/locales/original/course_content/other/k5-onlinepd-2019.json
- /i18n/locales/original/course_content/other/kodea-pd-2021.json
- /i18n/locales/original/course_content/other/outbreak.json


bin/i18n/resources/dashboard/course_content/sync_in.rb
```diff
@@ -32,6 +32,10 @@ def process
 
           private
 
+          def valid_source_file?(source_file_path)
+            !I18nScriptUtils.unit_directory_change?(I18N_SOURCE_DIR_PATH, source_file_path)
+          end
+
```
mgc1194, line 38: Here we are saying that a source file is valid if it's not a copied version of a file in a different directory. 
What happens if the valid source file is updated? Since it would already exist in the source directory, would it be skipped?

mgc1194 (approved):
Tricky behaviour, but changes look good. If we are looking for same files in different directories, we dont need to proved the full file path and also the file name, so this change makes sense.
I just have a question about updating any of the files.

bin/i18n/resources/dashboard/course_content/sync_in.rb
artem-vavilov: Correct, updating the previous version of the file will be skipped until we resolve this manually

daynew (approved):
The change makes sense to me.

PR #56359, opened 2024-02-06
[Google Blockly] Gracefully fail if we can't generate code

molly-moen:
Previously, if `getWorkspaceCode` from the google blockly wrapper threw an exception, the entire page would crash. This can happen if we hit a bug in our code (for example, we often hit it while developing with functions and behaviors), or if the level configuration has invalid blocks. With this change, we still allow the page to load, but show the user an error message and log the error to CloudWatch so we can fix it.

### Screenshot
Screenshot 2024-02-06 at 10 05 05 AM

## Links
- jira ticket: [CT-318](https://codedotorg.atlassian.net/browse/CT-318)

## Testing story
I tested this locally with a previously broken level, fixed by [this PR](https://github.com/code-dot-org/code-dot-org/pull/56294). 

## Follow-up work
If we discover other ways blockly can crash the page, we will file follow up tasks and fix them. This change doesn't cover the case when CDO Blockly crashes due to some issue, I will investigate that as a follow up to see if it's worth fixing now.

## PR Checklist:

- [ ] Code is well-commented


apps/src/blockly/googleBlocklyWrapper.js
```diff
@@ -160,13 +163,28 @@ function initializeBlocklyWrapper(blocklyInstance) {
...
+      if (this.getHiddenDefinitionWorkspace()) {
+        workspaceCode += Blockly.JavaScript.workspaceToCode(
+          this.getHiddenDefinitionWorkspace()
+        );
+      }
+      getStore().dispatch(setFailedToGenerateSources(false));
+    } catch (e) {
+      // We only want to log the error once per failure since getWorkspaceCode
+      // gets called many times and the error will be the same every time.
+      if (!getStore().getState().blockly.failedToGenerateSources) {
```
mikeharv: Nit: Would `failedToGenerateCode` be more accurate?

mikeharv (approved):
Looks good!

apps/src/blockly/googleBlocklyWrapper.js
molly-moen: yep I'll rename

ebeastlake:
Thanks for adding this, Molly! 🎉

PR #56360, opened 2024-02-06
Music: move view components out of MusicView

sanchitmalhotra126:
First part of some refactors to split up MusicView a bit. This moves all view rendering components out of MusicView into a separate, functional TS component, for now just called `MusicLabView` (I don't love the name but couldn't think of a better idea for now 🙂 ideally, once MusicView is just a container, we could rename it something like `MusicViewContainer` and this `MusicLabView` could then be called `MusicView`).

Since `MusicLabView` is a functional component, we can also absorb the logic from `UpdateTimer` and `ValidatorProvider`, both of which were created mostly just to take advantage of cleaner functional component syntax.

## Testing story

Tested locally on projectbeats and intro script. No functional changes.


breville (approved):
Looks great!

PR #56362, opened 2024-02-06
Web Interface to Edit DCDO Settings

sureshc:
The managed test server's web application server utilizes DCDO settings that are stored in DynamoDB, but unit tests (and any Ruby logic that runs outside the pegasus/dashboard puma web servers such as cron jobs or the dashboard console) utilize an in-memory DCDO data store. There's no way to configure the value of a DCDO setting that's used by the managed test server's web application server (and would thus control the functionality presented to browser user interface tests).

Provide a web interface (`/admin/dcdo`) linked from the Admin popup/overlay that enables admins to edit the DCDO settings utilized by that environment's web application server.

image

image

## PR Checklist:

- [ ] Code is well-commented


breville:
> There's no way to configure the value of a DCDO setting that's used by the managed test server's web application server

> Provide a web interface that enables admins to edit the DCDO settings utilized by that environment's web application server.

Is this saying we can now set permanent DCDO changes on the test server via this new UI?

lib/dynamic_config/datastore_cache.rb
```diff
@@ -21,6 +21,8 @@
 class DatastoreCache
   SHARED_CACHE_NAMESPACE = "DynamicConfigData".freeze
 
+  attr_reader :datastore
```
Hamms: Directly exposing the data that this class is intended to wrap seems like an anti-pattern; can this be removed? It's also not clear to me whether or not it's being used.

lib/dynamic_config/dcdo.rb
```diff
@@ -45,6 +45,14 @@ def frontend_config
...
     }
   end
+
+  def to_h
+    @datastore_cache.all
+  end
+
+  def refresh
+    @datastore_cache.update_local_cache
+  end
```
Hamms: nit; can these helper methods be defined on `dynamic_config_base` instead? Seems like they apply to Gatekeeper just as well as DCDO

dashboard/app/controllers/dynamic_config_controller.rb
```diff
@@ -65,4 +66,54 @@ def gatekeeper_set
...
+          raw_value.to_bool
+        when "String"
+          raw_value
+        else
+          new_value = Oj.load(raw_value)
+          if new_value.class.to_s != data_type
+            raise "#{new_value} does not match data type \"#{data_type}\""
+          end
+          new_value
+        end
```
Hamms, line 103: Is there a specific reason why we're doing a bunch of manual type checking rather than just using JSON directly?

dashboard/app/views/dynamic_config/dcdo_show.html.haml
```diff
@@ -0,0 +1,68 @@
...
+
+.dcdo-list
+  %h3 DCDO Flags
+  %div
+    %div{:class => %w[dcdo-row]}
+      %div{:class => %w[dcdo-label dcdo-padding]} Key
+      %div{:class => %w[dcdo-label dcdo-padding]} Value
+      %div{:class => %w[dcdo-label dcdo-padding]} Data Type
+      %div{:class => %w[dcdo-label]}
+    = form_for :add, :url => {:controller => 'dynamic_config', :action => 'dcdo_set'} do |f|
```
Hamms: I'm not familiar with this usage of `form_for`; I expect the first argument to that method to be a Rails model object whose attributes the form is providing a means to update, and then for us to use the form value `f` within the method block to create the inputs. It doesn't look like we're doing either of those things here; should this be [`form_tag`](https://api.rubyonrails.org/v6.1.7.3/classes/ActionView/Helpers/FormTagHelper.html#method-i-form_tag) instead?

snickell:
> The managed test server's web application server utilizes DCDO settings that are stored in DynamoDB, but unit tests (and any Ruby logic that runs outside the pegasus/dashboard puma web servers such as cron jobs or the dashboard console) utilize an in-memory DCDO data store. There's no way to configure the value of a DCDO setting that's used by the managed test server's web application server (and would thus control the functionality presented to browser user interface tests).

This is very much my ignorance speaking, but I would love an unpacking of this. Is this providing an edit-form for DCDO values stored in DynamoDB? How does this related to the unit test in-memory store? Can settings be set on that?

Hamms:
>  Is this providing an edit-form for DCDO values stored in DynamoDB?

Yep!

> How does this related to the unit test in-memory store? Can settings be set on that?

Only related in that both are using DCDO, but unrelated in that this form is using a different data store and so can in no way influence the actual values that the unit tests will retrieve. Currently, the store used by unit tests can be interacted with by executing `DCDO.set(...)` and `DCDO.get(...)` commands from a Ruby console; the only way currently to interact with the store used by the UI/Eyes tests is via the AWS console.

Hamms:
See https://github.com/code-dot-org/code-dot-org/blob/7e8bd89106b4350a6737d647fbc4a2638b9cd1be/lib/dynamic_config/environment_aware_dynamic_config_helper.rb#L5-L18 for the dirty details

snickell:
Thanks @Hamms, that was exactly the knowledge I needed.

lib/dynamic_config/datastore_cache.rb
sureshc: ooh. Good catch. I think I added this to debug some issues and didn't put in a TODO comment about removing it.

lib/dynamic_config/dcdo.rb
sureshc: Yes! That's definitely where they belong.

dashboard/app/controllers/dynamic_config_controller.rb
sureshc: Yes. Currently DCDO stores everything in DynamoDB as a string. When an engineer manually executes `DCDO.set` they can pass in a value that is an instance of a specific Ruby class: `DCDO.set('my_key', true)` and our dynamic configuration setting converts that to a string before persisting it, and converts it back when returning it.

The web user interface needs to ask the user to explicitly state the datatype of the value so that it can persist it in a way that ensures the datatype is preserved when it's retrieved.

It's been a couple of years since @KatieShipley and I implemented this in a hack-a-thon, so I may have forgotten some of the details, but I'm pretty certain that the core constraint is that we convert all values to string before saving to DynamoDB.

dashboard/app/views/dynamic_config/dcdo_show.html.haml
sureshc: Ah. Looks like starting with Rails 5.1, [both are soft-deprecated](https://guides.rubyonrails.org/v6.1/form_helpers.html#using-form-tag-and-form-for) in favor of [`form_with`](https://api.rubyonrails.org/v6.1.7.6/classes/ActionView/Helpers/FormHelper.html#method-i-form_with). I'll update to use that.

sureshc: I switched to `form_with` but after hours poring over the docs for the `url` parameter, I could not figure out how they work and went with a hard coded URL path. I could use some guidance here.

sureshc:
This is ready for another review.

breville:
Does this impact the approach we should take for updating DCDO values on test, as used by UI tests?  For example, we currently have some explanations in code for updating HOC-related values:

https://github.com/code-dot-org/code-dot-org/blob/51860e73d84658cdb3c8e655cd1c4e357cfde350/aws/ci_build#L214-L238

sureshc:
> Does this impact the approach we should take for updating DCDO values on test, as used by UI tests? For example, we currently have some explanations in code for updating HOC-related values:
> 
> https://github.com/code-dot-org/code-dot-org/blob/51860e73d84658cdb3c8e655cd1c4e357cfde350/aws/ci_build#L214-L238

No. This does not change how that logic works. It just provides a web interface for editing DCDO entries, and those values are persisted in whatever data store that environment is configured to use when Ruby logic is executing in a web application server. https://github.com/code-dot-org/code-dot-org/blob/b2ec55f87559fb268842979b4bf7992f0f4b1635/lib/dynamic_config/environment_aware_dynamic_config_helper.rb#L5-L18

dashboard/app/controllers/dynamic_config_controller.rb
```diff
@@ -65,4 +66,54 @@ def gatekeeper_set
...
+
+      DCDO.set(key, value)
+
+      log_msg = "<b>DCDO - #{key}</b> #{current_user.name} set to #{value}"
```
cat5inthecradle, line 107: Will this start creating an _incomplete_ log of DCDO changes?  Do all DCDO set actions get logged this way (other than a direct edit in DynamoDB)?

cat5inthecradle (approved):
LGTM!

dashboard/app/controllers/dynamic_config_controller.rb
sureshc: Dang. Yes. We copied the Gatekeeper web interface code, and brought over this functionality. I agree that the logging should be carried out by the underlying DCDO and Gatekeeper libraries. I think I want to save this for a future refactor of our dynamic configuration libraries.

PR #56364, opened 2024-02-06
Avoid `utf8` Alias in Schema Dump

Hamms:
In preparation for the eventual upgrade to MySQL 8, we'd like to be able to simultaneously support both MySQL 5.7 and MySQL 8. Unfortunately, the two versions treat the alias `utf8` differently; in 5.7, it equates to `utf8mb3` and in 8 to `utf8mb4`. To avoid potential confusion, we'd like to avoid using the alias at all and always specify which version we mean.

This PR proposes we implement that by modifying [`ActiveRecord::SchemaDumper::format_options`](https://github.com/rails/rails/blob/v6.1.7.3/activerecord/lib/active_record/schema_dumper.rb#L276-L278) to output the specific value rather than the alias when generating `schema.rb`. It also updates `schema.rb` to match the new expected results when dumping the schema as part of the staging build.

This should as a side effect reduce the frequency with which developers who are running MySQL 8 locally encounter unexpected `schema.rb` differences in their local environments.

## Links

- https://dev.mysql.com/doc/refman/8.0/en/charset-unicode-utf8.html
- https://dev.mysql.com/doc/refman/8.0/en/charset-unicode-utf8mb3.html
- https://dev.mysql.com/doc/refman/8.0/en/charset-unicode-conversion.html
- https://dev.mysql.com/blog-archive/mysql-8-0-when-to-use-utf8mb3-over-utf8mb4/
- https://dev.mysql.com/blog-archive/mysql-8-0-collations-migrating-from-older-collations/

## Testing story

Verified with a quick `grep` that we don't use any other `utf8` aliases:

```bash
$ grep -o "charset: [^,]*," dashboard/db/schema.rb | sort -u
charset: "utf8",
charset: "utf8mb4",
$ grep -o "collation: [^,]*," dashboard/db/schema.rb | sort -u
collation: "utf8mb4_bin",
collation: "utf8mb4_unicode_520_ci",
collation: "utf8_unicode_ci",
```

Also verified both locally and on an adhoc that creating a database from this schema with `db:schema:load` continues to work as expected.

## Deployment strategy

Usually the staging server is responsible for building and committing changes to `schema.rb`; this PR attempts to anticipate those changes, but it's possible there will be some details I overlooked. We should plan to keep an eye on the first DTS which includes this PR to verify that it worked as expected.

## Follow-up work

Next up, `schema_cache.rb`!

After that, we should plan to start converting some of these tables from `utf8mb3` to `utf8mb4`

## PR Checklist:

- [ ] Code is well-commented


Hamms:
This PR is a continuation of #56177, which was mis-closed by the Git LFS migration (#55759).

### Previous Comments:


### Previous Reviews:

dashboard/config/initializers/mysql_avoid_utf8_alias_in_schema_dump.rb
```diff
@@ -0,0 +1,49 @@
+# In MySQL 5.7 and earlier, `utf8` is an alias for `utf8mb3`, the three-byte
+# variant of the unicode character set. Starting with MySQL 8, `utf8` is now an
+# alias for `utf8mb4`, the four-byte variant.
+#
+# To help ease our transition from 5.7 to 8, avoid the use of the alias
+# entirely and instead use the appropriate explicit value for the current
+# verison.
+#
+# TODO infra: once we have fully upgraded to MySQL 8+, we can remove this.
```
pablo-code-org, line 9: <3

cat5inthecradle (approved):

dashboard/config/initializers/mysql_avoid_utf8_alias_in_schema_dump.rb
```diff
@@ -0,0 +1,49 @@
...
+      when /^8.0.\d+/
+        8.0
+      when /^5.7.\d+/
+        5.7
+      else
+        raise "cannot parse MySQL version #{raw_version.inspect}"
+      end
+    end
+  end
+end
```
pablo-code-org: My implementation
```
    @mysql_charset ||= begin
      version = ActiveRecord::Base.connection.select_value('SELECT VERSION()')
      major_version = version.split('.').first.to_i
      major_version >= 8 ? 'utf8mb4' : 'utf8mb3'
    rescue => e
      HoneyBadger.notify "Failed to fetch MySQL version for charset determination: #{e.message}"  
      'utf8mb3' # Default to 'utf8mb3' as a safer fallback for older versions of MySQL.
    end
```

The exception handling allows you to still get honeybadger errors but prevent any issues that may arise with a default value to the old one

dashboard/config/initializers/mysql_avoid_utf8_alias_in_schema_dump.rb
```diff
@@ -0,0 +1,49 @@
...
+        value = current_mysql_version < 8 ? 'utf8mb3' : 'utf8mb4'
+      end
+
+      if key == :collation && value == 'utf8_unicode_ci'
+        value = current_mysql_version < 8 ? 'utf8mb3_unicode_ci' : 'utf8mb4_unicode_ci'
+      end
+
+      "#{key}: #{value.inspect}"
+    end.join(', ')
+  end
```
pablo-code-org:
```
My own suggestion based on my own coding style.  


def format_options(options)
  charset, collation = charset_and_collation_values

  options.map do |key, value|
    value = case key
            when :charset then charset
            when :collation then collation
            else value
            end

    "#{key}: #{value.inspect}"
  end.join(', ')
end

def charset_and_collation_values
  if current_mysql_version < 8
    ['utf8mb3', 'utf8mb3_unicode_ci']
  else
    ['utf8mb4', 'utf8mb4_unicode_ci']
  end
end


```

dashboard/config/initializers/mysql_avoid_utf8_alias_in_schema_dump.rb
```diff
@@ -0,0 +1,49 @@
...
+      raw_version = ActiveRecord::Base.connection.select_value('SELECT VERSION()')
+      case raw_version
+      when /^8.0.\d+/
+        8.0
+      when /^5.7.\d+/
+        5.7
+      else
+        raise "cannot parse MySQL version #{raw_version.inspect}"
+      end
+    end
```
pablo-code-org:
```
    version_string = ActiveRecord::Base.connection.select_value('SELECT VERSION()')
    version_number = version_string.match(/\A(\d+\.\d+)/)&.[](1)&.to_f
    raise "Cannot parse MySQL version from '#{version_string}'" unless version_number
```    
   
This regex allows you to capture any version? I like that it can parse in more generic way and keeps it consistent and supports the format they use for naming releases

dashboard/test/config/initializers/mysql_avoid_utf8_alias_in_schema_dump_test.rb
```diff
@@ -0,0 +1,36 @@
+require 'test_helper'
+
+class MysqlAvoidUtf8AliasInSchemaDumpTest < ActiveSupport::TestCase
```
pablo-code-org, line 3: Some not happy test I would recommend
-Testing for bad versions 
- Testing for exceptions when requesting the version
- Default values if you decide to add this

```
test 'handles unexpected mysql version formats gracefully' do
  ActiveRecord::Base.connection.stubs(:select_value).returns('unexpected-format')
  assert_raises(RuntimeError) { @schema_dumper.send(:current_mysql_version) }
end

test 'uses fallback version on error fetching mysql version' do
  ActiveRecord::Base.connection.stubs(:select_value).raises('Simulated connection error')
  assert_equal(5.7, @schema_dumper.send(:current_mysql_version))
end

```

pablo-code-org, line 3: You might want to add a new integration test that allows you to test the attributes in the configuration

```
require 'active_record'
require 'minitest/autorun'

class SchemaDumperIntegrationTest < ActiveSupport::TestCase
  setup do
    @connection = ActiveRecord::Base.connection

    # Ensure any previous test schema is cleaned up
    @connection.drop_table('test_table') if @connection.table_exists?('test_table')

    # Create a simple table to test schema dumping
    @connection.create_table :test_table do |t|
      t.string :name
    end

    # Stub the current_mysql_version to control the test scenario
    ActiveRecord::SchemaDumper.prepend(Module.new do
      def current_mysql_version
        5.7
      end
    end)
  end

  teardown do
    @connection.drop_table('test_table') if @connection.table_exists?('test_table')
  end

  test 'schema dump includes utf8mb3 for MySQL 5.7' do
    schema_dump = dump_schema

    # Assert the schema dump includes the expected charset and collation
    assert_match /charset: "utf8mb3"/, schema_dump
    assert_match /collation: "utf8mb3_unicode_ci"/, schema_dump
  end

  private

  def dump_schema
    stream = StringIO.new
    ActiveRecord::SchemaDumper.dump(@connection, stream)
    stream.string
  end
end
```

pablo-code-org (approved):

pablo-code-org:
I am approving because my comments are enhancers and not blockers that can be included in other iterations. What you have is in a mergable state in my opinion and looks good. 

Feel free to integrate any suggestions here or in another PR or refute everything.

dashboard/config/initializers/mysql_avoid_utf8_alias_in_schema_dump.rb
Hamms: Oh, I had an underlying design decision here that I now realize is not at all clear. I'll add some comments to the code explaining this, but I'll also repeat myself here:

I want this code to be explicitly temporary, and scoped only to the MySQL 5.7 -> 8.0 transition. Later versions of MySQL than 8.0 don't support `utf8mb3` in the way we need them to, and I suspect earlier versions than 5.7 also would have issues (although I haven't verified that half). My goal with this method was not to create something that could identify versions of MySQL in general, my goal was to create something that could specifically distinguish between 5.7 and 8.0, and would alarm if anything other than that showed up. If this code at any point tries to run against a MySQL database running any other version, I _want_ it to error out.

Of course, that error should probably contain more of this context and explanation than just "cannot parse". I'll make that update; thank you for calling out this lack of communication!

dashboard/config/initializers/mysql_avoid_utf8_alias_in_schema_dump.rb
Hamms: I see where you're going with this; it's certainly more extensible than my approach, and if this code was intended to live for anything longer than as short a time as possible, I think that would be valuable. But I do personally find this harder to quickly understand at a glance, and for this kind of short-term transitional code, I think that kind of simple readability wins out over strict adherence to the principles of isolation.

That being said, I do agree that there's a bit too much complexity in my `map` block. What do you think about something like,

```ruby
def format_options(options)
  options.map do |key, value|
    value = resolve_alias_for_mysql_version(key, value)
    "#{key}: #{value.inspect}"
  end.join(', ')
end

def resolve_alias_for_mysql_version(key, alias)
  if key == :charset && alias == 'utf8'
    return current_mysql_version < 8 ? 'utf8mb3' : 'utf8mb4'
  end

  if key == :collation && alias == 'utf8_unicode_ci'
    return current_mysql_version < 8 ? 'utf8mb3_unicode_ci' : 'utf8mb4_unicode_ci'
  end

  return alias
end
```

dashboard/test/config/initializers/mysql_avoid_utf8_alias_in_schema_dump_test.rb
Hamms: I like this idea, and I put together a minimal implementation of it as an experiment. Unfortunately, it turns out that on my machine, calling `schema_dumper.dump` adds about *two seconds* to the runtime for just this one test! That seems excessive 

I'm sure it would be faster if we could use a mock connection that just dumps a single test table rather than the whole db, but after a bit of experimentation it looks like that would be a lot more complicated than it's worth. I think I'm ultimately going to skip implementing this integration test, despite liking the idea. Thanks for the suggestion, and the sample implementation!

PR #56370, opened 2024-02-06
[Google Blockly] Add keyboard navigation cursor options to workspace context menu

mikeharv:
As part of the work to [Create (and implement?) plan for making modal function editor keyboard navigable](https://codedotorg.atlassian.net/browse/CT-284), we should make it possible to utilize all three of Blockly's included cursor modes, at least for the purposes of collecting data about which mode(s) users prefer.

### Cursors
_(Descriptions come from Blockly's [demo](https://google.github.io/blockly-samples/plugins/keyboard-navigation/test/index.html) page)_
The cursor controls how the user navigates the blocks, inputs, fields and connections on a workspace. The demo shows three different cursors:

- **Default Cursor:** This cursor uses previous, next, in, and out to navigate through the different parts of a block. See the [developer documentation](https://developers.google.com/blockly/guides/configure/web/keyboard-nav) for more information.
- **Basic Cursor:** Uses pre order traversal to allow users to navigate through everything using only the previous and next command.
- **Line Cursor:** We tried to make this cursor mimic a text editor. Navigating up and down will take the cursor to the next and previous "line" of code. Navigating in and out will move the cursor through all the fields and inputs in that "line" of code.

This PR adds all three cursors to the wrapper. For some reason, the line cursor isn't already available directly from the Blockly instance, but it can be imported from the plugin. This is a bit odd, but it matches how the three cursor types are accessed in their own tests: https://github.com/google/blockly-samples/blob/435f28200f75362afce241922bb233f9fad9a36d/plugins/keyboard-navigation/test/index.js#L62-L71
In any case, we can add them as read-only properties on the wrapper by assigning them directly or using `wrapReadOnlyProperty`. While I was here, I took a minute to re-alphabetize the list.

In addition to this, I wanted to make it possible/somewhat easy for users to try different cursors so I added some new workspace context menu options which are hidden behind an experiment flag. These new options will show the current cursor in use (as a disabled option) and allow for toggling between the other two. Like all context menu options, they aren't super discoverable; our platform never actually directs users to interact with the menu, so this isn't in any way a final UI.

While I was in this space, I got frustrated by the convoluted way in which context menu option weights were mostly hard-coded. I did a quick audit of the standard options that we expect from core Blockly/plugins and added counters that are incremented programmatically to `registerAllContextMenuItems` that ensure that the options show up in the right order. Note: It is okay to use the same weight for multiple options if they have different scopes (block or workspace) as they won't appear together.

Similar to our accessible themes, here we're leveraging local storage to store the user's selected cursor type. This way, their preference will be remembered between levels and page refreshes.

Initial state (keyboard navigation off) - No changes|Keyboard navigation on, experiment disabled - no changes|Keyboard navigation on, experiment enabled, default cursor selected
-|-|-
image|image|image

@samantha-code suggested she would like to bring the cursor options to some keyboard users and ask about their preferences. These new options make it possible to conduct that research so we can hopefully learn which option(s) are most intuitive and preferred.

## Links

Relates to:
- https://codedotorg.atlassian.net/browse/CT-284

## Testing story

Tested that the new options show when and how expected and that they set the new cursor mode without messing anything related up. Tested that the cursor mode is preserved across workspaces and page loads.

Here's a long-ish video that demonstrates its use. I used additional software to monitor key presses on screen.

https://github.com/code-dot-org/code-dot-org/assets/43474485/8226543a-1f06-4459-9457-3d932a5c3d7f

## Follow-up work

At this time, it is not expected that all UI and block fields will be navigable, but that's a gap we'd ideally close before broadcasting this feature more broadly.

## PR Checklist:

- [ ] Code is well-commented


apps/src/blockly/addons/cdoUtils.js
```diff
@@ -261,6 +261,15 @@ export function getUserTheme(themeOption) {
   return Blockly.themes[localStorage.blocklyTheme] || themeOption || cdoTheme;
 }
 
+/**
+ * Returns a cursor type, based on the presence of an option in the browser's localStorage.
+ * @param {string} type
+ * @returns {string} one of 'default', 'basic', or 'line'
+ */
+export function getUserCursorType() {
+  return localStorage.blocklyCursor || 'default';
```
ebeastlake: Am I interpreting correctly that making the cursor type persistent for a given user is a follow-up task?

apps/src/blockly/addons/contextMenu.js
```diff
@@ -148,21 +153,59 @@ const registerKeyboardNavigation = function () {
...
-    weight: 11,
+    weight,
   };
   GoogleBlockly.ContextMenuRegistry.registry.register(keyboardNavigationOption);
 };
 
+/**
+ * Change cursor type for keyboard navigation
+ */
+const registerCursor = function (cursorType, weight) {
```
ebeastlake: nit: Could we call this `registerCursorOptions`?

apps/src/blockly/addons/contextMenu.js
```diff
@@ -148,21 +153,59 @@ const registerKeyboardNavigation = function () {
     },
     callback: function (scope) {
       const controller = Blockly.navigationController;
-      scope.workspace.keyboardAccessibilityMode
-        ? controller.disable(scope.workspace)
-        : controller.enable(scope.workspace);
+      if (scope.workspace.keyboardAccessibilityMode) {
```
ebeastlake: nit: It might be work creating a couple of helpers here to improve readability, like:
```
const isAccessibilityModeEnabled = (scope) => scope.workspace.keyboardAccessibilityMode;
const isNavigationEnabled = () => experiments.isEnabled(experiments.KEYBOARD_NAVIGATION);
```
But no strong feelings either way.

apps/src/blockly/addons/contextMenu.js
```diff
@@ -148,21 +153,59 @@ const registerKeyboardNavigation = function () {
...
+      }
+    },
+    callback: function (scope) {
+      localStorage.setItem(BLOCKLY_CURSOR, cursorType);
+      Blockly.navigationController.cursorType = cursorType;
+      setNewCursor(cursorType, scope);
+      Blockly.navigationController.navigation.focusWorkspace(scope.workspace);
+    },
+    scopeType: GoogleBlockly.ContextMenuRegistry.ScopeType.WORKSPACE,
+    id: cursorType + 'CursorOption',
```
ebeastlake: nit: `${cursorType}CursorOption`

apps/src/blockly/addons/contextMenu.js
mikeharv: It can be `registerCursorOption` (not plural). However, none of the other functions in this file use that convention. Possibly worth noting, we borrow the current convention from core Blockly: https://github.com/google/blockly/blob/develop/core/contextmenu_items.ts

apps/src/blockly/addons/contextMenu.js
```diff
@@ -148,21 +153,59 @@ const registerKeyboardNavigation = function () {
...
+        return MenuOptionStates.HIDDEN;
+      } else if (isCurrentCursor(cursorType)) {
+        return MenuOptionStates.DISABLED;
+      } else {
+        return MenuOptionStates.ENABLED;
+      }
+    },
+    callback: function (scope) {
+      localStorage.setItem(BLOCKLY_CURSOR, cursorType);
+      Blockly.navigationController.cursorType = cursorType;
```
ebeastlake: `setNewCursor` already handles this line, right?

apps/src/blockly/addons/contextMenu.js
```diff
@@ -148,21 +153,59 @@ const registerKeyboardNavigation = function () {
...
+      setNewCursor(cursorType, scope);
+      Blockly.navigationController.navigation.focusWorkspace(scope.workspace);
```
ebeastlake: Can you clarify what this line does, and under what conditions we have to refocus?

apps/src/blockly/addons/contextMenu.js
mikeharv: Ah yes, good catch! Will clean up.

apps/src/blockly/addons/contextMenu.js
```diff
@@ -209,10 +252,11 @@ const themes = [
     label: msg.blocklyTritanopiaTheme(),
   },
 ];
+const cursors = ['default', 'basic', 'line'];
```
ebeastlake: nit: Consider moving this to a Blockly-related constants file and making it all caps.
```
const CURSOR_TYPES = [...]
```

apps/src/blockly/addons/contextMenu.js
mikeharv: Will add a comment. It sets the navigation state to this workspace and moves the cursor to the top block.

apps/src/blockly/addons/contextMenu.js
```diff
@@ -286,22 +331,41 @@ export function registerHelp() {
...
+  registerEditable(nextBlockOptionWeight++);
+  registerShadow(nextBlockOptionWeight++);
+  registerUnshadow(nextBlockOptionWeight);
+
+  // Custom workspace options
+  registerKeyboardNavigation(nextWorkspaceOptionWeight++);
+  registerAllCursors(nextWorkspaceOptionWeight, cursors);
+  nextWorkspaceOptionWeight += cursors.length;
+  registerDarkMode(nextWorkspaceOptionWeight++);
+  registerThemes(nextWorkspaceOptionWeight, themes);
```
ebeastlake: Take or leave an alternate organization idea here:

```
// TODO: Add comments on why magic numbers make sense
const INITIAL_BLOCK_OPTION_WEIGHT = 7;
const INITIAL_WORKSPACE_OPTION_WEIGHT = 3;

const registerAllContextMenuItems = function () {
  unregisterDefaultOptions();
  
  let nextBlockOptionWeight = INITIAL_BLOCK_OPTION_WEIGHT;
  registerCustomBlockOptions(nextBlockOptionWeight);

  let nextWorkspaceOptionWeight = INITIAL_WORKSPACE_OPTION_WEIGHT;
  registerCustomWorkspaceOptions(nextWorkspaceOptionWeight, cursors, themes);
};

function registerCustomBlockOptions(startWeight) {
  const blockOptions = [registerHelp, registerDeletable, registerMovable, registerEditable, registerShadow, registerUnshadow];
  blockOptions.forEach(registerOption => {
    registerOption(startWeight);
    startWeight += 1;
  });
}

function registerCustomWorkspaceOptions(startWeight, cursors, themes) {
  const workspaceOptions = [registerKeyboardNavigation, registerDarkMode];
  workspaceOptions.forEach(registerOption => {
    registerOption(startWeight);
    startWeight += 1;
  });

  registerAllCursors(startWeight, cursors);
  startWeight += cursors.length;

  // No explicit increment needed here since registerThemes internally handles the increment
  registerThemes(startWeight, themes);
}
```

ebeastlake: I'd probably avoid using pre/post-increment operators as an argument because it feels a little weird to me to remember the timing...
```
If used postfix, with operator after operand (for example, x++), the increment operator increments and returns the value before incrementing.
```
But that's just one person's opinion!

ebeastlake:
Mainly nits. I know you've got a major rebase ahead of you, but nice work! 🎉

apps/src/blockly/addons/cdoUtils.js
mikeharv: Yes. Since this is just needed for testing, `localStorage` should be sufficient for getting it to work across levels/projects for a single user session. This is consistent with how we remember themes. I would imagine that we'd implement something better for user settings when we create the accessibility panel to manage these options.

apps/src/blockly/addons/contextMenu.ts
```diff
@@ -404,3 +471,35 @@ function unregisterDefaultOptions() {
...
+  // 2: redoWorkspace
+  let nextWeight = 3;
+
+  // Custom workspace options
+  registerKeyboardNavigation(nextWeight++);
+  registerAllCursors(nextWeight, NAVIGATION_CURSORS);
+  nextWeight += NAVIGATION_CURSORS.length;
+  registerDarkMode(nextWeight++);
+  registerThemes(nextWeight, themes);
+}
```
mikeharv, line 524: @ebeastlake I moved things out into two scope-specific functions here (lines 475-505). I'm still using increment operators, but let me know what you think.

apps/i18n/common/en_us.json
```diff
@@ -469,8 +469,6 @@
   "congratsCertificateThanks": "Thanks for submitting!",
   "congratsCertificateContinue": "Now, keep going with our other courses, or see more options to Learn beyond an Hour below.",
   "congratsNextLevelHeading": "Graduate to the next level",
-  "congratsSelfPacedPlTitle": "Explore more self-paced professional learning modules",
-  "congratsSelfPacedPlDescription": "Keep learning with our robust selection of self-paced professional learning modules. Our self-paced learning works in tandem with our curriculum so you can empower yourself and your students at the pace that’s best for you.",
```
ebeastlake: It looks like there's still unintended diff here @mikeharv?

apps/src/blockly/constants.ts
```diff
@@ -65,6 +66,7 @@ export const Renderers = {
   DEFAULT: 'cdo_renderer_thrasos',
 };
 
+export const NAVIGATION_CURSORS = ['default', 'basic', 'line'];
```
ebeastlake: Why not refer to these as cursor types? (Since they're not cursor instances.)

apps/src/blockly/addons/contextMenu.ts
```diff
@@ -404,3 +474,35 @@ function unregisterDefaultOptions() {
     GoogleBlockly.ContextMenuRegistry.registry.unregister('workspaceDelete');
   } catch (error) {}
 }
+
+function registerCustomBlockOptions() {
```
ebeastlake, line 483: Could you add a general comment about how the weights work/what they are for? (Or maybe JS Docs-style comments by default in Blockly functions would be useful generally, which would explain the weights on the functions that use them?)

ebeastlake:
It looks like there's still some unnecessary diff, but otherwise, this LGTM!

apps/src/blockly/constants.ts
mikeharv: Sure!

apps/src/blockly/addons/contextMenu.ts
mikeharv: Can do!

apps/i18n/common/en_us.json
mikeharv: Oh, hmm. Must be due to merging in Molly’s branch and staging. I’ll make sure it gets cleaned up.

ebeastlake (approved):

apps/src/blockly/addons/contextMenu.ts
```diff
@@ -264,16 +309,20 @@ const registerTheme = function (name: Themes, label: string, index: number) {
...
 
-function registerThemes(themes: {name: Themes; label: string}[]) {
+function registerThemes(
+  startingWeight: number,
+  themes: {name: Themes; label: string}[]
+) {
   themes.forEach((theme, index) => {
-    registerTheme(theme.name, theme.label, index);
+    const weight = startingWeight + index;
+    registerTheme(theme.name, theme.label, weight);
```
molly-moen, line 317: do you need to update `registerTheme` to take in `weight` rather than `index`?

apps/src/blockly/addons/contextMenu.ts
```diff
@@ -313,22 +362,47 @@ export function registerHelp() {
...
+    .setCursor(Blockly.getNewCursor(type));
+  if (Blockly.getFunctionEditorWorkspace()) {
+    Blockly.getFunctionEditorWorkspace()
+      ?.getMarkerManager()
+      .setCursor(Blockly.getNewCursor(type));
+  }
+  if (oldCurNode) {
+    markerManager.getCursor()?.setCurNode(oldCurNode);
+  }
+  // Set the navigation state to the workspace and moves the cursor to the top block.
```
molly-moen, line 391: for my information: if this moves the cursor to the top block, what does `setCurNode` do?

apps/src/blockly/addons/contextMenu.ts
```diff
@@ -386,21 +460,75 @@ function setAllWorkspacesTheme(
...
+  // sorted in the order listed here.
+  registerHelp(nextWeight++);
+  registerDeletable(nextWeight++);
+  registerMovable(nextWeight++);
+  registerEditable(nextWeight++);
+  registerShadow(nextWeight++);
+  registerUnshadow(nextWeight);
+}
+
+function registerCustomWorkspaceOptions() {
```
molly-moen, line 506: can you add comment(s) explaining the difference between workspace and block options?

apps/src/blockly/addons/contextMenu.ts
```diff
@@ -386,21 +460,75 @@ function setAllWorkspacesTheme(
...
+  let nextWeight = 3;
+
+  // Custom workspace options. We increment the weight for each so they are automatically
+  // sorted in the order listed here. Cursors and Themes have multiple options so we need to adjust
+  // the weight accordingly.
+  registerKeyboardNavigation(nextWeight++);
+  registerAllCursors(nextWeight, NAVIGATION_CURSOR_TYPES);
+  nextWeight += NAVIGATION_CURSOR_TYPES.length;
+  registerDarkMode(nextWeight++);
+  registerThemes(nextWeight, themes);
```
molly-moen: should this be `nextWeight++`?

apps/src/blockly/addons/functionEditor.ts
```diff
@@ -125,6 +125,18 @@ export default class FunctionEditor {
 
     const functionEditorTrashcan = new CdoTrashcan(this.editorWorkspace);
     functionEditorTrashcan.init();
+    // Set primary workspace to be active (until a function is shown).
+    Blockly.common.setMainWorkspace(this.primaryWorkspace);
```
molly-moen, line 129: why is this needed now?

apps/src/blockly/googleBlocklyWrapper.ts
```diff
@@ -706,6 +710,18 @@ function initializeBlocklyWrapper(blocklyInstance: GoogleBlocklyInstance) {
 
     blocklyWrapper.navigationController.addWorkspace(workspace);
 
+    blocklyWrapper.getNewCursor = function (type) {
+      switch (type) {
+        case 'basic':
+          return new blocklyWrapper.BasicCursor() as Cursor;
```
molly-moen: do we need to cast all of these cursor types? `blocklyWrapper.Cursor()` should be the right type already, right? Does `BasicCursor` extend from `Cursor`?

molly-moen (commented):
Mostly questions for my understanding, looks good!

apps/src/blockly/addons/contextMenu.ts
mikeharv: As far as I understand it, `setCurNode` just sets the location of the marker - the actual blinking SVG.

apps/src/blockly/addons/contextMenu.ts
mikeharv: Not needed. The `++` incrementation happens after the value is accessed so that it's ready for the next option to be registered. This is the last option, so we don't need to increment at all. The comment below this line explains that we'd want to count up (by the number of theme options) if we add anything else to the menu.

apps/src/blockly/addons/contextMenu.ts
mikeharv: Yes, but it really comes down to whether it's an option for when you right-click a block vs right-click the workspace.

apps/src/blockly/addons/contextMenu.ts
ebeastlake:
```
The ++ incrementation happens after the value is accessed 
```
Molly and I both asked about this, so it might be worth mentioning this in a comment?

apps/src/blockly/addons/contextMenu.ts
molly-moen: Are the marker and the cursor different? What I'm trying to understand is the comment says we move the cursor to the top block, but the marker is wherever it was before?

apps/src/blockly/googleBlocklyWrapper.ts
mikeharv: It looks like it's happy without the casting. `LineCursor` extends `BasicCursor` which extends `Cursor`. Will remove!

apps/src/blockly/addons/functionEditor.ts
mikeharv: The active ("main") workspace is the one used for keyboard shortcuts, including the one used to turn on keyboard navigation. Without this, if you toggled navigation on immediately upon page load it would actually enable it for the editor workspace since it was the active one. 

We also set the primary workspace to be active when the editor is closed, and do the opposite when we show the editor for a function.

apps/src/blockly/addons/contextMenu.ts
mikeharv: Agreed. Will add!

apps/src/blockly/addons/contextMenu.ts
```diff
@@ -523,12 +521,9 @@ function registerCustomWorkspaceOptions() {
   let nextWeight = 3;
 
   // Custom workspace options. We increment the weight for each so they are automatically
-  // sorted in the order listed here. Cursors and Themes have multiple options so we need to adjust
-  // the weight accordingly.
+  // sorted in the order listed here. The ++ incrementation happens after the value is accessed.
   registerKeyboardNavigation(nextWeight++);
-  registerAllCursors(nextWeight, NAVIGATION_CURSOR_TYPES);
-  nextWeight += NAVIGATION_CURSOR_TYPES.length;
```
molly-moen: why could this be removed?

mikeharv: I opted to give each cursor (and theme) option the same menu weight for simplicity.

apps/src/blockly/addons/contextMenu.ts
mikeharv: Oh good question. I just realized we're doing two contradictory things. When we say `setCurNode(oldCurNode)`, we're basically saying that we should keep the cursor on the same node (e.g. stack/block/field/connection) when the user changes from one cursor type to another. However, we're also moving the cursor to the top block after that. I think the latter makes more sense for the purposes of the testing that Sam wants to conduct.

The code for the preservation of the current node comes from the [test playground](https://google.github.io/blockly-samples/plugins/keyboard-navigation/test/index.html) for the plugin: https://github.com/google/blockly-samples/blob/0d09c5d2dfd31d370ada8393cb002a39ef3d028e/plugins/keyboard-navigation/test/index.js#L72-L74

Will remove logic that cared about the old node!

PR #56379, opened 2024-02-07
Add Storybook linter

bencodeorg:
As a follow-up to Storybook cleanup last week, adds a linter for our Storybook entries. Main goal here was to prevent adding entries with a `title` attribute on the default export, which undo some of the reorganization work done [here](https://github.com/code-dot-org/code-dot-org/pull/56200) when added. The Storybook linter enforces this rule -- more detail on the linter here:

https://github.com/storybookjs/eslint-plugin-storybook

This enables all recommended Storybook rules, and cleans up or makes exceptions for any existing violations. It also adds the rule that prevents adding a `title` property (`storybook/no-title-property-in-meta`), which isn't in the recommended set (it's in `csf-strict`), but I wanted to include since it's what sent me down this path :).

Some of the issues fixed here are:
- fixes stories using camelCase instead of PascalCase
- adds a couple of default exports where possible
- makes exceptions for a couple of components where using default/named exports doesn't really work (eg, when stories are generated in a loop)
- makes exceptions for stories with the `title` property where a top-level category was provided for the story (eg, `DesignSystem/Radio Button`)


apps/.eslintignore
```diff
@@ -5,3 +5,5 @@
 /lib
 /build
 node_modules
+
+!.storybook
```
fisher-alice, line 9: Was curious why this needed to be added and was going to ask about it, but then saw the explanation [here](https://github.com/storybookjs/eslint-plugin-storybook/blob/a053ba9ec53fa782467fdf25c58231ecf24b5277/README.md?plain=1#L58). 😄

apps/src/code-studio/pd/form_components/FieldGroup.story.jsx
```diff
@@ -51,22 +51,22 @@ TestWrapper.propTypes = {
...
-  title: 'FormComponents/FieldGroup',
+  title: 'FormComponents/FieldGroup', // eslint-disable-line storybook/no-title-property-in-meta
   component: FieldGroup,
 };
 
 const Template = args => <TestWrapper {...args} />;
 
-export const basicExample = Template.bind({});
-basicExample.args = {
+export const BasicExample = Template.bind({});
```
fisher-alice, line 60: I didn't know this was called pascal case!

apps/src/code-studio/pd/form_components/QuestionsTable.story.jsx
```diff
@@ -43,14 +43,14 @@ TestWrapper.propTypes = {
 };
 
 export default {
-  title: 'FormComponents/QuestionsTable',
+  title: 'FormComponents/QuestionsTable', // eslint-disable-line storybook/no-title-property-in-meta
```
fisher-alice, line 46: I think this is an example of a component for which the default/named exports doesn't work. You said this story is generated in a loop - do you mind explaining what you mean by this? Just wondering how we can prevent this in the future.

apps/src/componentLibrary/checkbox/Checkbox.story.tsx
```diff
@@ -4,7 +4,7 @@ import {Meta, Story} from '@storybook/react';
 import {ComponentSizeXSToL} from '@cdo/apps/componentLibrary/common/types';
 
 export default {
-  title: 'DesignSystem/Checkbox',
+  title: 'DesignSystem/Checkbox', // eslint-disable-line storybook/no-title-property-in-meta
```
fisher-alice, line 7: Just confirming that you are assigning the title for 'DesignSystem' components so that this name is used instead of 'componentlLibrary'. Could we just rename the 'apps/src/componentLibrary' directory to 'apps/src/designComponents' and then not have to include this ignore rule each time we add a new design system component?

fisher-alice: I'm also curious about the order of the directories in storybook. I see 'DesignSystem' first, but then 'Templates'. I thought the order would be alphabetical since it is in apps.

fisher-alice (approved):
Nice! Just left a couple questions. Thanks for doing this!

apps/src/componentLibrary/checkbox/Checkbox.story.tsx
bencodeorg: I added some configuration that allows us to force design system components (and the templates and code-studio directories after that) to the top of Storybook (I feel like this is desirable, at least the design system part, not sure if others agree):

https://github.com/code-dot-org/code-dot-org/pull/56200/files#diff-6547070ce8b34eb95836426236464d696f30d39efe116477c3bbb02c61d569c9

No strong opinions on whether its called "design system" or "component library"

apps/src/code-studio/pd/form_components/QuestionsTable.story.jsx
bencodeorg: This actually isn't an example of that -- this is just allowing the use of a `title` property that someone configured manually. I decided to leave those in place if someone had added a title with a directory structure (ie, has a slash in it).

An example of the loop thing is here:

https://github.com/code-dot-org/code-dot-org/blob/c13389b41c82801f74e0cd890f6e02f381d7e07b/apps/src/templates/progress/ProgressBubble.story.jsx#L17-L37

The problem is that you can only add an export in module scope (ie, the top level in a file, not inside a function call), which isn't possible if you're generating stories in a loop. You can check out the export in that file, which includes the stories as part of the ~~default export (instead of as named exports like we typically do in .story files). I'm actually not 100% sure why that structure works, but it seems to...~~ nvm, makes sense. See comment below!

Hamms (approved):
I'm not sure I use Storybook enough these days to review this from a storybook-specific perspective, but from a general tooling and process perspective, this looks great! Always happy to see more linting, and all of the individual fixes look like improvements to me :smile:

apps/src/code-studio/pd/form_components/FieldGroup.story.jsx
bencodeorg: haha me neither until I saw the error messages :)

apps/src/code-studio/pd/form_components/QuestionsTable.story.jsx
fisher-alice: Gotcha! This makes sense now `/* eslint-disable storybook/default-exports */` - thanks!

levadadenys (approved):
Checked locally, LGTM! Thanks for these updates, that's awesome!

One nit only: I see that now we have a light-gray backgorund color for 'component render area' from common.scss file. Is that intentional?
image

apps/src/code-studio/pd/form_components/QuestionsTable.story.jsx
bencodeorg: 👍 yep -- and realizing I think it makes sense that this export structure works...we provide the default export (assigned to the key `default` in `module.exports`), and named exports for every property in that `stories` object

apps/src/componentLibrary/checkbox/Checkbox.story.tsx
fisher-alice: Any idea on why the listing of the non-configured directories are not in alpha order as they are in apps?

bencodeorg: I don't know 😬

bencodeorg:
> Checked locally, LGTM! Thanks for these updates, that's awesome!
> 
> One nit only: I see that now we have a light-gray backgorund color for 'component render area' from common.scss file. Is that intentional?

Good q, yeah I think that's a result of this PR:

https://github.com/code-dot-org/code-dot-org/pull/56304

Not intentional really -- have you noticed any other styling issues? We could switch back to a white background by reordering this file so application.css is loaded last:

https://github.com/code-dot-org/code-dot-org/blob/staging/apps/.storybook/preview-head.html

But I looked at one of our labs (dance), and saw that we load common.css after application.css. So the current version is closer to what's actually happening in one of our labs, so it might be better to leave as-is? Do you have a strong preference either way?

levadadenys:
> > Checked locally, LGTM! Thanks for these updates, that's awesome!
> > One nit only: I see that now we have a light-gray backgorund color for 'component render area' from common.scss file. Is that intentional?
> 
> Good q, yeah I think that's a result of this PR:
> 
> #56304
> 
> Not intentional really -- have you noticed any other styling issues? We could switch back to a white background by reordering this file so application.css is loaded last:
> 
> https://github.com/code-dot-org/code-dot-org/blob/staging/apps/.storybook/preview-head.html
> 
> But I looked at one of our labs (dance), and saw that we load common.css after application.css. So the current version is closer to what's actually happening in one of our labs, so it might be better to leave as-is? Do you have a strong preference either way?

Thanks for the context! +1 to keeping storybook styles loading as close as possible to what we have on dashboard pages, +1 to keep things as is.

PR #56380, opened 2024-02-07
Fix Violations of and Reenable `Performance/MethodObjectAsBlock`

Hamms:
> Identifies places where methods are converted to blocks, with the use of `&method`, and passed as arguments to method calls. It is faster to replace those with explicit blocks, calling those methods inside.

Fixes mostly applied in vim with sed: `%s/(\?&\(\w*::\)\?\(\w*\.\)\?method(:\([^)]*\)))\?/ {|TODO| \1\2\3(TODO)}/gc`

In each case, I then manually replaced `TODO` with an appropriate variable name. In some cases, I did some additional manual cleanup.

I do personally like the cleanliness of `&method`, and it's a pattern that makes sense to my JavaScript-writing brain. If this was a `Style/` cop, I'd probably be arguing against enabling it. From a performance perspective, however, the difference is significant (see links below). I don't know that any of these instances currently represent actual performance issues in practice, but as a general principle I think we should always enable `Performance/` cops unless we have a very good reason not to.

## Links

Rubocop Documentation:

- https://msp-greg.github.io/rubocop-performance/RuboCop/Cop/Performance/MethodObjectAsBlock.html
- https://docs.rubocop.org/rubocop-performance/cops_performance.html#performancemethodobjectasblock

Performance Comparison:

- https://github.com/fastruby/fast-ruby#normal-way-to-apply-method-vs-method-code
- https://github.com/fastruby/fast-ruby/blob/main/code/general/block-apply-method.rb


Hamms:
This PR is a continuation of #52795, which was mis-closed by the Git LFS migration (#55759).

### Previous Comments:


### Previous Reviews:

davidsbailey (approved):
Great analysis -- 100% agree on sadness over losing the conciseness, but priority going to performance.

PR #56384, opened 2024-02-07
Fix Violations of and Reenable `jsx-a11y/no-noninteractive-tabindex`

Hamms:
Specifically, for all but one violation I fixed it by removing the nonfunctional `tabindex` attribute from the non-interactive element. In that last case, the element had a click handler so I added a `role` attribute to make it explicitly interactive.

## Testing story

I have not manually tested these changes; I checked few of them and confirmed this change is functionally a no-op, but wasn't sure how to find all of them for testing. In particular, the one element for which I added a `role` rather than removing a `tabindex` remains untested.

I feel reasonably confident this change is safe despite the lack of testing, but am very open to disagreement. Let me know what you think!

## PR Checklist:

- [ ] Code is well-commented


Hamms:
This PR is a continuation of #55558, which was mis-closed by the Git LFS migration (#55759).

### Previous Comments:


### Previous Reviews:

bencodeorg (approved):
LGTM! I looked at the `DanceAiModalHeader` component and confirmed no regression from this change (ie, can still tab nav to that component).

Hamms:
Great, thank you!

PR #56385, opened 2024-02-07
Install MySQL 8 in Docker

Hamms:
In preparation for eventually updating to MySQL 8 everywhere, for testing purposes.

## PR Checklist:

- [ ] Code is well-commented


Hamms:
This PR is a continuation of #55893, which was mis-closed by the Git LFS migration (#55759).

### Previous Comments:


### Previous Reviews:

cat5inthecradle:
- [x] Built and pushed this new image as 1.9.4
- [x] Updated versions used
- [x] resigned drone config.

This can be merged if drone passes.

docker/dockerfiles/Dockerfile
```diff
@@ -105,23 +105,8 @@ WORKDIR /tmp
...
-RUN curl -sSLOJ https://dev.mysql.com/get/mysql-apt-config_0.8.22-1_all.deb && \
-    echo "mysql-apt-config mysql-apt-config/select-server select mysql-5.7" | /usr/bin/debconf-set-selections && \
-    echo "mysql-apt-config mysql-apt-config/repo-codename select bionic" | /usr/bin/debconf-set-selections && \
-    dpkg -i mysql-apt-config_0.8.22-1_all.deb || apt-get --fix-broken --yes install && \
-    rm -rf mysql-apt-config_0.8.22-1_all.deb && \
-    apt-key adv --keyserver keyserver.ubuntu.com --recv-keys B7B3B788A8D3785C && \
-    apt-get update && \
-    apt-get install --yes libmysqlclient-dev mysql-client=5.7.42-1ubuntu18.04 mysql-server=5.7.42-1ubuntu18.04
+RUN apt-get update && \
+    apt-get install --yes libmysqlclient-dev mysql-client mysql-server
```
snickell, line 109: Well that's a big improvement 😅 I think its worth adding a comment above this line mentioning that we're specifically targeting MySQL v8, its implicit here, but sometimes its good to know where the version has been thoughtfully chosen vs "whatever's in ubuntu".

docker/dockerfiles/Dockerfile
```diff
@@ -130,6 +115,23 @@ RUN service mysql start && \
...
     service mysql stop
 
+# For some reason, MySQL 8 on Ubuntu 20 does not have a home directory by
+# default, so we explicity specify one to avoid the error:
+#
+#   su: warning: cannot change directory to /nonexistent: No such file or directory
+#
+# For more, see:
+#  https://serverfault.com/questions/755763/mysql-nonexistent-home-vs-no-directory-logging-in-with-home/942422
+#  https://stackoverflow.com/a/63040661/1810460
```
snickell, line 125: 🥇 for linking to relevant stackoverflow when adding necessary complexity

docker/dockerfiles/Dockerfile
```diff
@@ -105,23 +105,8 @@ WORKDIR /tmp
...
-    dpkg -i mysql-apt-config_0.8.22-1_all.deb || apt-get --fix-broken --yes install && \
-    rm -rf mysql-apt-config_0.8.22-1_all.deb && \
-    apt-key adv --keyserver keyserver.ubuntu.com --recv-keys B7B3B788A8D3785C && \
-    apt-get update && \
-    apt-get install --yes libmysqlclient-dev mysql-client=5.7.42-1ubuntu18.04 mysql-server=5.7.42-1ubuntu18.04
+RUN apt-get update && \
+    apt-get install --yes libmysqlclient-dev mysql-client mysql-server
 
 # disable auth for local dashboard mysql db
 RUN service mysql start && \
```
snickell, line 112: Does it matter that this `service mysql start` comes before the mysql homedir changes below?

snickell (approved):
LGTM

docker/dockerfiles/Dockerfile
Hamms: I don't think so; I don't remember exactly where we encountered the homedir error, but it doesn't prevent the commands in this block from executing successfully, and because we stop the service again afterwards this block shouldn't disrupt the homedir changes from being applied successfully. It seems likely to me that they would also work in the other order, but because they're working as-is, I'm inclined to leave it.

jordan-springer (approved):

PR #56386, opened 2024-02-07
Add Ability to Install MySQL 8.0 with Chef

Hamms:
Make it possible to install MySQL 8.0 via cookbooks, although do not yet do so by default. Environments can be updated individually by overriding the `cdo-mysql` cookbook's `target_version` Chef attribute to `"8.0"`.

## Testing story

Tested on an adhoc that this change can be applied cleanly to a new environment without impact (ie, MySQL 5.7 is still installed by default). Then verified that updating the target version and rerunning `chef-client` will update MySQL to 8.0.

## Deployment strategy

This change should be safe to deploy as-is, since all of our managed servers are still running MySQL 5.7.

## Follow-up work

After merging, we can begin selectively updating environments to MySQL 8.0

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/cookbooks/cdo-mysql/attributes/default.rb
+++ b/cookbooks/cdo-mysql/attributes/default.rb
@@ -1,4 +1,5 @@
 default['cdo-mysql'] = {
+  target_version: '5.7',
   proxy: {
     # Enable ProxySQL on non-daemon app-servers by default.
     enabled: node['cdo-apps'] && !node['cdo-apps']['daemon'],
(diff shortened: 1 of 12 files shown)


Hamms:
This PR is a continuation of #56086, which was mis-closed by the Git LFS migration (#55759).

### Previous Comments:


### Previous Reviews:

sureshc:
The 2nd test case did not pass. After merging this branch into the adhoc's branch, the adhoc updated successfully, but did not upgrade to MySQL8. This is possibly acceptable because adhocs are intended to be short lived, so we don't have that many existing adhocs that need to upgrade to MySQL 8. Inspecting the Chef log file (`/var/log/chef-bootstrap-debug.log`) it's possible that in an adhoc that's not Chef-managed, changes to Cookbooks after the initial provisioning of the adhoc are not deployed and executed on the adhoc.

sureshc (approved):

sureshc (approved):

PR #56388, opened 2024-02-07
updated version of puppeteer

kobryan0619:
This PR is intended to move us one step closer to being able to easily re-generate PDFs for lesson plans and resources in the curriculum.  

Back story:  PDFs have been not auto-generating since the fall when it started blocking the staging build.  Since then we have updated a google drive gem that works with our Ruby upgrade which helped get PDFs of resources generated.  However, we still struggled with the PDF generation timing out with puppeteer.  The "work around" was to just repeatedly try generating PDFs until they worked.  This PR updates the version of Puppeteer.  

## Links

[Ticket](https://codedotorg.atlassian.net/jira/software/projects/TEACH/boards/65?selectedIssue=TEACH-860)

## Testing story

I tested this on an adhoc. 

- PDFs of lesson plans were consistently uploaded to S3 successfully.
- PDFs of resources were less successful on the adhoc.   This will be investigated further in [TEACH-862](https://codedotorg.atlassian.net/browse/TEACH-862)
- PDFs of state fact sheets were able to be generated, however, due to the way the adhoc was setup (mainly, I didn't have access to the access key) I was unable to transfer and then open the PDFs locally.  I am not 100% sure of the right way to proceed on this.  I see three options:

1. I can try merging this onto staging and then testing there, while being prepare to revert if needed.
2. Try spinning up another adhoc using a key I own, but I didn't see directions for doing that in the documentation provided. 
3. Find someone who has access to the server key automatically created and have them verify that the content is accurate. 

## PR Checklist:

- [ ] Code is well-commented


lib/pdf/conversion.rb
```diff
@@ -23,13 +23,17 @@ def self.generate_from_html(html, outpath, options = {})
...
+  def self.invoke_generation_script(args, options = {}, attempts: 3)
     script_path = "#{deploy_dir}/bin/generate-pdf"
     cmd = (['timeout', PDF_GENERATION_TIMEOUT.to_s, 'node', script_path] + args).join(" ")
     puts cmd if options[:verbose]
     @result = `#{cmd}`
     if $?.exitstatus == 124
       raise "pdf generation timed out after #{PDF_GENERATION_TIMEOUT} seconds. cmd: #{cmd}"
+    elsif $?.exitstatus == 1
+      warn "Exit status 1.  Re-attempting."
+      raise "pdf generation failed with status #{$?.exitstatus}. cmd: #{cmd}" if attempts <= 1
```
bethanyaconnor: nit: flip these so that it doesn't look like we're reattempting when we're not

lib/pdf/conversion.rb
```diff
@@ -23,13 +23,17 @@ def self.generate_from_html(html, outpath, options = {})
...
+      invoke_generation_script(args, options, attempts: attempts - 1)
```
bethanyaconnor: Recursion always sets off my code smell detectors. This works but I'm curious why not simply use a loop? To me, that's generally more readable, but I'm open to discussion

kobryan0619: Good question - I was leveraging [this](https://github.com/code-dot-org/code-dot-org/blob/fc372301b0b1a4543fe5560b9f91d20afb5c8338/lib/cdo/crowdin/project.rb#L42) and [this](https://github.com/code-dot-org/code-dot-org/blob/fc372301b0b1a4543fe5560b9f91d20afb5c8338/lib/cdo/crowdin/legacy_utils.rb#L88) as examples. 

That being said, a loop seems like an equally good solution here.

Hamms (commented):
+1 to Bethany's comments, but otherwise this LGTM! I also tend to reach for recursion first as a solution to these kinds of problems, but it does unfortunately tend to lead to long-term maintenance issues, no matter how pretty it always looks in the moment.

Can you clarify what access key you're missing to be able to test PDFs on the adhoc? It should be possible to `scp` files from the adhoc to gateway and then from gateway to your machine using the same permissions we use to `ssh` between the same servers. Granted, it is a bit of an annoying manual process, but just let me know if you want to quickly pair sometime and I'd be happy to show you!

lib/pdf/conversion.rb
bethanyaconnor: If we used a loop, could we include the timeout above in the retry logic? Do we want to?

If the answer to both those questions is yes, then I think this should be turned into a loop. If the answer to one (or both) of those questions is no, I think this is fine to avoid a refactor / increased complexity outside of this change. This recursion is straightforward and not going to have a performance impact, so it's not something I'm going to block on. :)

kobryan0619: I did try changing this to a loop inside the `elsif $?.exitstatus == 1` case, it looked something like this:

```
def self.invoke_generation_script(args, options = {})
    script_path = "#{deploy_dir}/bin/generate-pdf"
    cmd = (['timeout', PDF_GENERATION_TIMEOUT.to_s, 'node', script_path] + args).join(" ")
    puts cmd if options[:verbose]
    @result = `#{cmd}`
    if $?.exitstatus == 124
      raise "pdf generation timed out after #{PDF_GENERATION_TIMEOUT} seconds. cmd: #{cmd}"
    elsif $?.exitstatus == 1
      retryAttempts = 2
      while $?.exitStatus == 1 && retryAttempts > 0
        retryAttempts = retryAttempts - 1
        warn "Exit status 1.  Re-attempting."
        cmd = (['timeout', PDF_GENERATION_TIMEOUT.to_s, 'node', script_path] + args).join(" ")
        puts cmd if options[:verbose]
        @result = `#{cmd}`
      end
    elsif $?.exitstatus != 0
      raise "pdf generation failed with status #{$?.exitstatus}. cmd: #{cmd}"
    end
    puts "  #{@result}" if options[:verbose]
  end
```
but I ran into an issue that if the re-attempt failed for a different reason, that reason wouldn't be captured in the output.  

That led me to try wrapping nearly all of the if-else logic inside a while loop but it got pretty ugly looking by the end of that... so I changed it back to use recursion.  

That being said, I am not so sure how important it is to raise all the different exit statuses in this function.  It is possible that we just say "if you fail once, try 2 more times" and then raise the last failure that took place.   I also might be totally overthinking this.

Thoughts?

kobryan0619:
Thanks @Hamms for the tip on moving files to my local machine.  Bryan showed me how to do it on a call and I added some documentation on doing that in the How To Provision an Adhoc document as well.

kobryan0619:
Note: I was able to verify that generated state fact sheets were created accurately with the new puppeteer version as well.

lib/pdf/conversion.rb
Hamms: Yeah, I suppose it is a bit awkward to implement a looped retry for just that one failure case. If we instead wanted to just always retry, we could do something like:

```ruby
# warning: entirely untested code; just a theoretical example
def self.invoke_generation_script(args, options = {}, retry_attempts: 3)
  script_path = "#{deploy_dir}/bin/generate-pdf"
  cmd = (['timeout', PDF_GENERATION_TIMEOUT.to_s, 'node', script_path] + args).join(" ")
  puts cmd if options[:verbose]
  current_attempt = 0
  while current_attempt < retry_attempts
    warn "Retrying; attempt #{current_attempt}" if current_attempt > 0
    current_attempt += 1
    @result = `#{cmd}`
    if $?.exitstatus == 124 # Timeout
      message = "pdf generation timed out after #{PDF_GENERATION_TIMEOUT} seconds. cmd: #{cmd}"
    elsif $?.exitstatus != 0 # Any other non-success
      message = "pdf generation failed with status #{$?.exitstatus}. cmd: #{cmd}"
    elsif $?.exitstatus == 0 # Success!
      break
    end

    if current_attempt < retry_attempts
      warn message
    else
      raise message
    end
  end
  puts "  #{@result}" if options[:verbose]
end
```

But that's starting to get a bit silly and I think your first implementation is probably cleaner.

What about this plan: keep your original recursive implementation, but add a comment which calls out that we are intentionally using recursion here even though it is often a code smell, and explains that we are doing so because in this specific case we are confident that performance is not a concern and we found that our attempts to implement this with a loop just led to even worse-smelling code. Just a little note to whatever poor soul finds themselves trying to iterate on this method in the future :smile:

bethanyaconnor: I like Elijah's suggestion! I'm satisfied that this code is more readable this way and, with a comment that performance isn't a concern here, that a future engineer will feel confident. :)

lib/pdf/conversion.rb
```diff
@@ -23,13 +23,24 @@ def self.generate_from_html(html, outpath, options = {})
...
     cmd = (['timeout', PDF_GENERATION_TIMEOUT.to_s, 'node', script_path] + args).join(" ")
     puts cmd if options[:verbose]
     @result = `#{cmd}`
     if $?.exitstatus == 124
       raise "pdf generation timed out after #{PDF_GENERATION_TIMEOUT} seconds. cmd: #{cmd}"
+    elsif $?.exitstatus == 1
+      # We have had inconsistent issues with pdf generation timing out thus here 
+      # we are re-attempting generation using recursion.  We considered using a loop
+      # instead of reucursion, but the code was less readable or had side effects that 
+      # were not addressing this specific issue with time out.
```
bethanyaconnor: can you add a note that performance isn't a concern in this case?

bethanyaconnor (approved):
🎉

lib/pdf/conversion.rb
```diff
@@ -23,13 +23,25 @@ def self.generate_from_html(html, outpath, options = {})
...
     puts cmd if options[:verbose]
     @result = `#{cmd}`
     if $?.exitstatus == 124
       raise "pdf generation timed out after #{PDF_GENERATION_TIMEOUT} seconds. cmd: #{cmd}"
+    elsif $?.exitstatus == 1
+      # We have had inconsistent issues with pdf generation timing out thus here 
+      # we are re-attempting generation using recursion.  We considered using a loop
+      # instead of reucursion, but the code was less readable or had side effects that 
+      # were not addressing this specific issue with time out. We do not see an issue 
+      # with performance in this case.
```
Hamms: Super-tiny nit; what about something like,

```suggestion
      # were not addressing this specific issue with time out. Because this method is
      # invoked as part of the staging build rather than anything user-facing, we do
      # not see an issue with performance in this case.
```

Hamms (approved):
One tiny little comment phrasing suggestion, but it's non-blocking. LGTM; thanks for all the care you put into this update!

PR #56389, opened 2024-02-07
Implement `cache-staging-clone` pipeline for Drone

### LINKED ISSUE #56375: Cache repo between drone runs
- We currently get ~~200 clones per day of our repo: https://github.com/code-dot-org/code-dot-org/graphs/traffic
- Each clone will consume more than 20GB of LFS data (possibly much more, tbd)
- The majority of our clones are from Drone (see below)
- If we implement a caching mechanism, we can dramatically reduce the number of clones.

Elijah had the idea that we could cache a zip download of the tagged releases from https://github.com/code-dot-org/code-dot-org/releases

By implementing a mechanism for caching git checkouts and "fast forwarding them" to the current branch, we can dramatically reduce LFS data usage, at leas the usage attributable to drone.

Fixed by #56389

snickell:
This PR uses the drone s3 caching plugin to save a copy of `staging` every time we push to the branch, and then use that cached copy of staging to build. This approach was @sureshc 's idea.

Should substantially reduce Git LFS bandwidth costs see issue #56375 for graphs and details. Could easily save $100+/day in LFS data packs.

Loading staging from a cache also **improves effective git checkout time on Drone by 5 minutes** on every PR run (from 10m to 4:30m).

This PR currently includes #56394 which caches in the same region as drone workers , which ideally would be reviewed and merged first.

### How tested
- I setup two extra branches, `seth/fake-staging` and `seth/fake-feature`. With `seth/fake-feature` I opened a corresponding PR which targets merging into `seth/fake-staging` (https://github.com/code-dot-org/code-dot-org/pull/56398). This allowed testing the new trigger-event-push that will be on `staging` once merged. 
- Here's an example of the fake feature branch running: https://drone.cdn-code.org/code-dot-org/code-dot-org/39783
- Performance was tested with a variety of cloning approaches, and in the end a full clone proved simple and performant.
- Tested it still runs when the cache file is deleted from S3, fallbacks to cloning if cache fails (works)
- Tested on a cache of staging where the feature branch was behind the $DRONE_COMMIT_ID
- Tested on a cache of staging where the feature branch did not exist yet
- Inspected output after checkout steps to make sure feature branch was being merged on top of the target branch correctly.

### Deploy checklist

- [x] Verify that `seth/fake-staging` also passes (that's a cache hit condition, this PR is a cache miss condition) https://github.com/code-dot-org/code-dot-org/pull/56400
- [x] Remove references to `seth/fake-staging` right before merge
- [x] Drone sign
- [ ] Notify #developers of lookalike jobs: this PR generates a new type of drone job. This can potentially be confusing, so we should send #developers this image that shows the difference:
image


snickell:
Fake feature branch/PR for testing new drone.yml is: https://github.com/code-dot-org/code-dot-org/pull/56390

Which merge's into a fake staging branch: `seth/fake-staging`

snickell:
The `cache-staging-clone` pipeline does the caching, and takes about 10 minutes to run a depth=100 clone + upload:
image

Without shallow cloning (experiment) we get a longer cache time:
image

A depth=200 clone takes somewhere between the full clone and shallow clone times:
image

snickell:
The fat clones take almost 4 extra minutes to cache, HOWEVER, the PR hooks (what we're used to running in drone) run a minute+ faster each time.

```
### WITH SHALLOW CLONE (DEPTH=100) CLONE CACHED:
drone fetch: 3:44

drone checkout:
  git fetch 2m10s
  checkout 50s
  total: 3m
  
TOTAL cache shallow clone: **6:44**


### WITH FULL CLONE CACHED:
drone fetch: 4:55

drone checkout:
  git fetch (not required)
  git checkout 50s
  
TOTAL cache fat clone: **5:45**

### WITH SHALLOW(200) CLONE CACHED:

drone fetch: 4:30

drone checkout:
  git fetch 11s
  checkout 50s
  total: 68s
  
TOTAL cache shallow (200) clone: **5:40**
```

The reason is that with a shallow clone, you're forced to fetch the branch vs already having its objects enumerated, and even though we're fetching it with `--depth 200`, its still having to do a big download of data.

snickell:
A depth=200 clone takes somewhere between the full clone and shallow clone times:
image

snickell:
```
### WITH SHALLOW(200) CLONE CACHED:

drone fetch: 4:30

drone checkout:
  git fetch 11s
  checkout 50s
  total: 68s
  
TOTAL cache shallow (200) clone: **5:40**
```

snickell:
ANALYSIS: although the full clone takes a couple minutes longer to run than a (depth=200) clone, it executes in about the same time (5:45 vs 5:40), and the drone code is much simpler, and therefore easier to reason about. Additionally it should be more robust in the face of unexpectedly deep branches, etc. Its not clear that its particularly important to have staging cache pipelines merge 2 minutes faster, and code simplicity is good.

DECISION: since complicated is not getting us big perf wins, lets use simple: FULL CLONE

For historical reference in case we want to resurrect shallow cloning, the last shallow clone was: d012d3d3bb47296992bfe39c70ee37eaa4e6def2

.drone.yml
```diff
@@ -212,8 +228,44 @@ trigger:
...
+clone:
+  disable: true
+
+steps:
+
+  - name: clone
+    image: bitnami/git
+    commands:
+      - git lfs install
+      - git clone --branch $DRONE_BRANCH --depth 100 $DRONE_GIT_HTTP_URL .
```
snickell: Opted to cache a full clone instead of using `--depth` fetches because the perf speedup was tiny and it increased complexity, see: https://github.com/code-dot-org/code-dot-org/pull/56389#issuecomment-1931622715

.drone.yml
```diff
@@ -39,7 +54,7 @@ steps:
   - name: restore-cache
     image: plugins/s3-cache
     settings:
-      endpoint: s3://code-dot-org
+      endpoint: s3://code-dot-org-drone
```
snickell: This is from #56394

.drone.yml
```diff
@@ -210,10 +243,47 @@ trigger:
...
+clone:
+  disable: true
+
+steps:
+
+  - name: clone
+    image: bitnami/git
+    commands:
+      - git lfs install
+      - git clone --branch $DRONE_BRANCH $DRONE_GIT_HTTP_URL .
```
snickell, line 265: Opted to cache a full clone instead of using --depth fetches because the perf speedup was tiny and it increased complexity, see: https://github.com/code-dot-org/code-dot-org/pull/56389#issuecomment-1931622715

.drone.yml
```diff
@@ -22,13 +22,29 @@ steps:
...
+        - .
+
+  - name: checkout
     image: bitnami/git
     commands:
-      - git lfs install
-      - git clone --branch $DRONE_SOURCE_BRANCH  --depth 200 $DRONE_GIT_HTTP_URL .
+      - git lfs install --skip-repo
+      # We should be in a cached git repo, but clone if the cache failed
+      - "[ -d .git ] || (echo 'WARNING: get-cached-staging failed: checkout will be SLOW and LFS $$$!' && git clone --branch $DRONE_BRANCH $DRONE_GIT_HTTP_URL .)"
```
snickell, line 42: Fallback method so we're resilient (but eventually aware, thinking racking up LFS GB fast if caching fails) in case the cache file disappears from S3, the caching pipeline breaks, or something else fails

.drone.yml
```diff
@@ -96,6 +112,7 @@ trigger:
   branch:
     - "staging"
     - "staging-next"
+    - "seth/fake-staging"
```
snickell: Will remove 2x `seth/fake-staging` refs before merge, useful to have until this is final.

.drone.yml
```diff
@@ -210,10 +243,47 @@ trigger:
...
```
snickell: Will remove 2x seth/fake-staging refs before merge, useful to have until this is final.

cat5inthecradle:
For consideration: Moving the S3 bucket and role definition to cloudformation: https://github.com/code-dot-org/infrastructure/pull/5

.drone.yml
cat5inthecradle: Will this warning surface anywhere useful?

.drone.yml
```diff
@@ -210,10 +243,47 @@ trigger:
...
+      debug: true
+      pull: true
+      mount:
+        - .
+
+trigger:
+  branch:
+    - "seth/fake-staging"
+  event:
+    - push
```
cat5inthecradle: nit: move `trigger` to be right after `kind`, `type`, and `name`. Then the comment become less necessary.

sureshc (approved):

cat5inthecradle (approved):

.drone.yml
snickell: This shows up in the drone logs, its not an instant catch, but I think I phrased it so there's a chance when somebody notices the checkout step is 5 minutes slower while waiting for drone, they can tie this back to why its slower.

.drone.yml
snickell: We're keeping the same ordering of the keys as the other pipelines, discussed with @sureshc

PR #56391, opened 2024-02-07
Revert "remove Dedicated Reporting Database Instance from ProxySQL Configuration"

sureshc:
Reverts code-dot-org/code-dot-org#56334

Preparing this in case the production release of #56334 fails or causes problems.


sureshc:
Not needed! Removing the reporting database was successful and did not need to be reverted.

PR #56403, opened 2024-02-07
added id check to investigate test failure

Nokondi:
Removed skip from ai_evaluate_student_code ui tests. Tests are passing. Based on investigation, it seems like the failure was caused by two reused failed AI assessments. Deleted the failed assessments from the test DB and added a new step that prints debug information on the current user to the cucumber logs. This will make it easier to identify this issue if it arises again.

Screenshot of debug results in cucumber logs
image


davidsbailey (commented):
could you please include a screenshot of how this shows up in the ui test output?

dashboard/test/ui/features/step_definitions/section_management_steps.rb
```diff
@@ -103,7 +103,9 @@
   section = JSON.parse(browser_request(url: '/dashboardapi/sections', method: 'POST', body: {login_type: 'email', participant_type: 'student'}))
   section_code = section['code']
   @section_url = "http://studio.code.org/join/#{section_code}"
-  create_user(name, url: "/join/#{section_code}", code: 200, age: under_13 ? '10' : '16')
+  @student = create_user(name, url: "/join/#{section_code}", code: 200, age: under_13 ? '10' : '16')
+  puts @student
```
pablo-code-org: Is this needed?

dashboard/test/ui/features/step_definitions/section_management_steps.rb
```diff
@@ -103,7 +103,8 @@
   section = JSON.parse(browser_request(url: '/dashboardapi/sections', method: 'POST', body: {login_type: 'email', participant_type: 'student'}))
   section_code = section['code']
   @section_url = "http://studio.code.org/join/#{section_code}"
-  create_user(name, url: "/join/#{section_code}", code: 200, age: under_13 ? '10' : '16')
+  @student = create_user(name, url: "/join/#{section_code}", code: 200, age: under_13 ? '10' : '16')
+  @student
```
davidsbailey: @Nokondi I thought the `puts` statement was needed to increase our ability to debug the problem in the future. Is there something that has changed since we discussed that?

also, one nit: there's no need to use an `@` variable if you are only referencing it locally. so this can be:
```suggestion
  student = create_user(name, url: "/join/#{section_code}", code: 200, age: under_13 ? '10' : '16')
  puts student
```

davidsbailey (commented):
as mentioned a very old comment, please include a screenshot of the UI test output verifying that this will show the info we need (user id).

dashboard/test/ui/features/step_definitions/section_management_steps.rb
Nokondi: RE: our conversation on slack, I've added a new step to account_steps that retrieves that current user's information and displays it in the cucumber logs for debugging.

davidsbailey (approved):
looks great! Thank you for your perseverance!

PR #56414, opened 2024-02-07
[Progress] style sort-by-name dropdown using new design component

lfryemason:
Note: Eyes changes for sort dropdown on teacher panel and progress v1 page

Changes the sort-by-name dropdown to use the SimpleDropdown so that it is easier to style for the new progress page.

## Before
image
image

## After
image

image
image

## Links
[TEACH-793](https://codedotorg.atlassian.net/browse/TEACH-793)

## PR Checklist:

- [ ] Code is well-commented


lfryemason:
@levadadenys the label seems a bit far away from the dropdown in the small version. Especially when compared to other dropdowns that haven't been migrated yet. Do you agree?

apps/src/code-studio/components/progress/teacherPanel/TeacherPanel.jsx
```diff
@@ -246,13 +247,16 @@ class TeacherPanel extends React.Component {
...
-            sortByStyles={styles.sortBy}
-            selectStyles={styles.select}
-            sectionId={sectionId}
-            unitName={unitName}
-            source={TEACHER_PANEL}
-          />
+          <div className={moduleStyles.sortDropdownContainer}>
+            <SortByNameDropdown
+              sortByStyles={styles.sortBy}
+              selectStyles={styles.select}
```
etaderhold: Looks like these props have been deleted from the component.

```suggestion
```

apps/src/code-studio/components/progress/teacherPanel/TeacherPanel.jsx
```diff
@@ -305,12 +309,8 @@ const styles = {
...
   },
   sortBy: {
-    display: 'block',
     textAlign: 'center',
-  },
-  select: {
-    width: 180,
-    margin: '0px 10px 5px',
+    marginBottom: '10px',
   },
```
etaderhold: No longer used.

etaderhold (approved):

apps/src/templates/sectionProgressV2/progress-table-v2.module.scss
```diff
@@ -227,14 +227,15 @@ $level-cell-width: 52;
 
 .sortDropdown {
   height: 76px;
-  overflow-x: hidden;
+  width: fit-content;
```
lfryemason, line 230: @levadadenys I noticed that this was required to not have the caret floating in space next to the div. If width isn't variable, maybe we should add this to the design component itself.

levadadenys:
> @levadadenys the label seems a bit far away from the dropdown in the small version. Especially when compared to other dropdowns that haven't been migrated yet. Do you agree?

Agree, though not sure if that's ok or not. Might worth asking design team in slack re this (@markabarrett)

apps/src/templates/sectionProgressV2/progress-table-v2.module.scss
levadadenys: Interesting, haven't encounetered this issue in storybook. Are you sure it's not use case specific thing?
If it can be used anywhere I think we can update SimpleDropdown component styles in a follow up PR (also updating changelog.md)

levadadenys (approved):
LGTM in general. Happy to see Design System components being used live!

One non bliocking nit, that can/should be treated in a follow up pr

PR #56421, opened 2024-02-07
Remove direct link to internal doc

dju90:
...to mitigate external users from requesting google doc access.

## Testing story
**Tested locally**
Before
image

After
image


cat5inthecradle (approved):
Thanks!

PR #56425, opened 2024-02-07
Autosave before closing applab

bencodeorg:
A teacher reported that they experienced data loss in Applab several times as they navigated between levels. Turns out we have autosave set up every 30 seconds (and triggered by other actions, like clicking the run button), but no autosave when navigating away from the page.

We have logic in place in Javalab that saves a project before leaving the page -- this PR refactors that logic into `project.js` and makes use of in Applab.

Javalab sets a project's edited/not state to be changed on each change in its editor (Codemirror), but we have no such logic set up for Applab. In this PR, I add a change handler to Applab's editor (Droplet) such that the project's edited/not state is changed each time an update happens in the editor. It's worth noting that this change doesn't trigger a write to S3 with every key stroke or anything -- it just sets the state within `project.js` of `hasProjectChanged` to true, which then is used to decide when to write to S3 (either via our every 30 second autosave, or when navigating away from the page).

[update 2/20/24]: a couple other changes were required in order to make this work with some existing code that handles multi-tab situations -- first, the possibility that a user is logged out when they try to save (401), and second, the possibility that a user makes a change in one tab that we might attempt to override in another tab (409). In both cases, I've updated the logic to not show the pop-up when these are observed.

## Links

- Jira (with Zendesk ticket): [LABS-256](https://codedotorg.atlassian.net/browse/LABS-256)

## Testing story

Tested manually in an Applab level (/s/csp3-2023/lessons/5/levels/3) that before this change, I could navigate away and lose progress. With his change, we present the user with the browser default "changes may not be saved if you navigate away" dialog. You can then cancel, close again, and changes are saved. Also tested no behavior change in a Javalab allthethings level.

## Follow-up work

Not sure if there's any reason we wouldn't want this for all labs? It's a little tricky because I think we have to plug into each editor (Blockly, Codemirror, Droplet, etc), but I think that is mostly handled by StudioApp's `addChangeHandler`?


apps/src/code-studio/initApp/project.js
```diff
@@ -1923,6 +1923,22 @@ var projects = (module.exports = {
     }
     return sourcesApi;
   },
+
+  // Sets a callback to save the project before unloading the page.
+  registerSaveOnUnload() {
+    const unloadHandler = event => {
+      if (this.hasOwnerChangedProject()) {
```
fisher-alice: Cool! Just tracing through how this works. I see that `hasProjectChanged` is set to `true` when the `events.workspaceChange` is detected.
This event is defined at https://github.com/code-dot-org/code-dot-org/blob/3b049ee796cd976c4235996d9f57e7e3738adb21/apps/src/code-studio/initApp/project.js#L46-L47.
I see where the other events are triggered like `appModeChanged` [Example](https://github.com/code-dot-org/code-dot-org/blob/3b049ee796cd976c4235996d9f57e7e3738adb21/apps/src/StudioApp.js#L2055-L2056)
But could you point out where the `events.workspaceChange` is actually triggered? Just trying to understand how this is working. Thanks!

bencodeorg: Yeah I was confused by this as well! I just did some digging, and I was able to trigger it on a CDO Blockly level (/s/outbreak/lessons/1/levels/2), but not Google Blockly (I tried dance). It was triggered here:

https://github.com/code-dot-org/blockly/blob/f012d8262f21bae3e54fb11dd8bc29cf0d29f3cd/core/ui/block_space/block_space.js#L831

So, I think that event is only relevant for old Blockly labs.

For Applab, that line I added above (`studioApp().addChangeHandler(project.projectChanged);`) is what triggers the change to the `hasProjectChanged` state:

https://github.com/code-dot-org/code-dot-org/blob/41dc24fe1c3a371469ca95e1e46326a57120b1cf/apps/src/StudioApp.js#L1044-L1075

fisher-alice: Nice - this makes sense. Thanks!

apps/src/javalab/Javalab.js
```diff
@@ -329,20 +329,7 @@ Javalab.prototype.init = function (config) {
...
-  if (project.hasOwnerChangedProject()) {
-    // Manually trigger an autosave instead of waiting for the next autosave.
-    project.autosave();
-
-    event.preventDefault();
-    event.returnValue = '';
-  } else {
-    delete event.returnValue;
-  }
+  project.registerSaveOnUnload();
```
fisher-alice: Nit - would this line make sense further up near where you [register reducers](https://github.com/code-dot-org/code-dot-org/blob/3b049ee796cd976c4235996d9f57e7e3738adb21/apps/src/javalab/Javalab.js#L201)? Seems a little out of place here.

fisher-alice (approved):
Really nice that you add a change handler to set the`hasProjectChanged` and not add additional writing to S3. Cool!
Other than the unit test that needs to be fixed - LGTM!

I've had ample experience when i've lost work on Sprite Lab after refreshing or navigating way from a level and forgetting to hit run/reset so think this would be great to implement for other labs.

sanchitmalhotra126 (approved):
Nice! Could other labs also simply add `project.registerSaveOnUnload();` to get this behavior?

bencodeorg:
> Nice! Could other labs also simply add `project.registerSaveOnUnload();` to get this behavior?

That, and making sure the editor has a change handler set up that marks a project as changed every time an edit is made (I think this would work the same way in Google Blockly labs as here, like `studioApp().addChangeHandler(project.projectChanged);`)

It looks like the `studioApp().setupChangeHandlers` thing I used is set up to handle Blockly and Droplet, so it might require additional work if we wanted to support another editor (weblab)?

bencodeorg:
@fisher-alice @sanchitmalhotra126 had to move some stuff around here for testing purposes (and add a couple new tests) since y'all last looked, lmk what you think! Sorry to have not had this sorted out the first time around.

apps/src/javalab/Javalab.js
bencodeorg: Moved!

apps/src/code-studio/initApp/project.js
```diff
@@ -1923,6 +1923,23 @@ var projects = (module.exports = {
     }
     return sourcesApi;
   },
+
+  // Sets a callback to save the project before unloading the page.
+  registerSaveOnUnload() {
+    window.addEventListener('beforeunload', this.unloadHandler_.bind(this));
```
levadadenys, line 1934: For the context, do we need and/or have a removeEventListener for this callback somewhere?
Here's a docs about it https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeunload_event#usage_notes

bencodeorg: We do not...is there a specific problem you're concerned about / alternative approach you'd suggest?

We add the listener once on mount, and basically want to keep it there until you navigate away from the page, so I'm not sure when we'd remove it?

apps/test/unit/code-studio/initApp/projectTest.js
```diff
@@ -180,6 +180,7 @@ describe('project.js', () => {
     });
 
     it('default case', () => {
+      window.appOptions.app = 'someOtherType';
```
fisher-alice: Why do you have this line twice?

apps/test/unit/code-studio/initApp/projectTest.js
```diff
@@ -1100,6 +1101,42 @@ describe('project.js', () => {
         .catch(err => done(err));
     });
   });
+
+  describe('registerSaveOnUnload', () => {
+    it('sets up event handler on unload', () => {
+      const unloadHandlerSpy = sinon.spy();
+      sinon.stub(project.unloadHandler_, 'bind').returns(unloadHandlerSpy);
```
fisher-alice, line 1107: Nice stubbing!

dashboard/test/ui/features/star_labs/applab/shared_apps.feature
```diff
@@ -5,6 +5,8 @@ Feature: App Lab Scenarios
   Background:
     Given I start a new Applab project
     And I wait for the page to fully load
+    # Make sure we can navigate to another page without getting an alert
+    And I disable onBeforeUnload
```
fisher-alice: Do you mind explaining why we're getting an alert without including this line?

fisher-alice (approved):
Updating the tests seemed pretty tricky - nice job! Just had a couple questions

apps/src/code-studio/initApp/project.js
levadadenys: Tested locally, looks like my concerns do not confirm since there's no way to naigate away from app lab screen rather then reloading the page fully (native web page chenge vs react no reload page/screens change).

I was worried about that this eventListener might be triggered on pages you visit after the app lab page, but looks like it's not possible. Also, if project is not changed, autosave on close is not triggered, which is awesome! In docs, removeEventListener was involved in that logic, but in our case we've did that without removeEventListener so we don't need one in the current code.

Though, reading more through docs - if we do want to add this functionality for mobile/tablets also - looks like it might be worth using `visibilitychange` instead of `beforeunload`. Since:

```
The beforeunload event suffers from some problems:

It is not reliably fired, especially on mobile platforms. For example, the beforeunload event is not fired at all in the following scenario:
A mobile user visits your page.
The user then switches to a different app.
Later, the user closes the browser from the app manager.
Note: It is recommended to use the [visibilitychange](https://developer.mozilla.org/en-US/docs/Web/API/Document/visibilitychange_event)
 event as a more reliable signal for automatic app state saving that gets around problems like the above.
 See [Don't lose user and app state, use Page Visibility](https://www.igvita.com/2015/11/20/dont-lose-user-and-app-state-use-page-visibility/) for more details.
```


https://developer.mozilla.org/en-US/docs/Web/API/Document/visibilitychange_event

What do you think?

apps/test/unit/code-studio/initApp/projectTest.js
bencodeorg: 😅 I have no idea how that got there, removed it!

dashboard/test/ui/features/star_labs/applab/shared_apps.feature
bencodeorg: Welp I hit so many of these across many tests I ended up deciding to disable this popup in UI tests. This step just runs some code that sets a flag:

https://github.com/code-dot-org/code-dot-org/blob/67907f0f23f60705d0ba44d60def05231f4187bf/dashboard/test/ui/features/step_definitions/steps.rb#L1226-L1228

Which then gets used to disable the event handler logic.

I ended up setting this flag to true by default in tests (which made the UI test step to set it to true unnecessary). The downside here is that UI tests wouldn't catch if this popup was appearing unexpectedly, but I think it is pretty annoying to have to disable it in any Applab test, and we do have unit tests that test the core behavior. It also does not disable the new change handler that marks a project as having changed on any change in the editor (ie, `studioApp().addChangeHandler(project.projectChanged);`), and no tests failed.

apps/src/code-studio/initApp/project.js
bencodeorg: I'm not super worried about it (this is only relevant for a maximum of 30 seconds of work, since we autosave on that interval), the mobile/tablet case sounds pretty niche. Good to know re: visibilitychange though for future reference!

dashboard/test/ui/features/star_labs/applab/shared_apps.feature
bencodeorg: Curious if you have thoughts about whether disabling these popups in UI tests is a bad idea though Alice!

fisher-alice: I pulled the latest version of your branch, and for me, it seems like the auto-saving is not working locally. I think it's because `ignoreOnBeforeUnload` is set to `true` by default. It looks like `window.__TestInterface` is defined not just for UI tests but in general when the app is loaded. When I set it to `false` the autosave works as expected.

It looks like there's been online discussion on how to disable this `onbeforeunload` popup in UI tests so it seems like it's okay to disable. I agree it's annoying to have to disable it in any Applab test.

I thought maybe for the UI tests, we could run/reset after a project change, but this also doesn't prevent the popup window, I think because `hasProjectChanged` is still set to `true`. Maybe you could add an update so that when a program is run/reset, the `projectHasChanged` property is set to `false`? Then we could just add a step in UI tests to run/reset the program to avoid the popup. 

This would also mean that user would not be presented with the pop-up unnecessarily if the project is already saved after run/reset and no change in the workspace has been detected.

bencodeorg: 🤦 ughh good catch Alice! I'm surprised we don't set hasProjectChanged to false after any project save, lemme look into it...

apps/src/code-studio/initApp/project.js
```diff
@@ -986,7 +986,10 @@ var projects = (module.exports = {
           JSON.stringify(currentSources) !== JSON.stringify(newSources);
         if (sourcesChanged || thumbnailChanged) {
           thumbnailChanged = false;
-          this.saveSourceAndHtml_(newSources, resolve);
+          this.saveSourceAndHtml_(newSources, () => {
+            hasProjectChanged = false;
```
fisher-alice, line 990: What do you think of assigning `hasProjectChanged = false;` within the function definition of `saveSourceAndHtml_`?
This would help include cases such as when version history is reset [here](https://github.com/code-dot-org/code-dot-org/blob/73e7ff9844ef17642ec4985e7766aa148eada6c3/apps/src/templates/VersionHistory.jsx#L134-L135) and sources are not changed from these defined events:
https://github.com/code-dot-org/code-dot-org/blob/73e7ff9844ef17642ec4985e7766aa148eada6c3/apps/src/code-studio/initApp/project.js#L42-L47

apps/src/code-studio/initApp/project.js
```diff
@@ -1138,13 +1141,15 @@ var projects = (module.exports = {
           if (err) {
             if (err.message.includes('httpStatusCode: 401')) {
               this.showSaveError_();
+              callback(err);
```
fisher-alice, line 1144: I don't think this and the other `callback(err)` are needed if you set `hasProjectChanged` to `false` within this function.

apps/src/code-studio/initApp/project.js
```diff
@@ -1495,9 +1500,9 @@ var projects = (module.exports = {
    * @param {function} callback Function to be called after saving.
    */
   autosave(callback) {
-    const callCallback = () => {
+    const callCallback = skipHandleUnload => {
```
fisher-alice, line 1503: I don't think `skipHandleUnload` is ever defined?

apps/src/code-studio/initApp/project.js
```diff
@@ -1538,7 +1543,7 @@ var projects = (module.exports = {
         return;
       }
 
-      this.saveSourceAndHtml_(newSources, () => {
+      this.saveSourceAndHtml_(newSources, skipHandleUnload => {
```
fisher-alice, line 1546: Same comment as above for `skiphandleUnload`.

apps/src/code-studio/initApp/project.js
```diff
@@ -1923,6 +1928,32 @@ var projects = (module.exports = {
...
+  unloadHandler_(event) {
+    // Skipped for UI tests
+    if (window.__TestInterface && window.__TestInterface.ignoreOnBeforeUnload) {
+      return;
+    }
+
+    if (this.hasOwnerChangedProject()) {
+      // Manually trigger an autosave instead of waiting for the next autosave.
+      this.autosave(skipHandleUnload => {
+        if (skipHandleUnload) {
```
fisher-alice, line 1946: Same comment about `skipHandleUnload`.

apps/src/code-studio/initApp/project.js
```diff
@@ -1923,6 +1928,32 @@ var projects = (module.exports = {
...
   },
+
+  // Sets a callback to save the project before unloading the page.
+  registerSaveOnUnload() {
+    window.addEventListener('beforeunload', this.unloadHandler_.bind(this));
+  },
+
+  unloadHandler_(event) {
+    // Skipped for UI tests
+    if (window.__TestInterface && window.__TestInterface.ignoreOnBeforeUnload) {
```
fisher-alice, line 1939: Is this still needed?

fisher-alice (commented):
Just sharing some observations - this really made me appreciate good tests that cover edge cases you don't immediately think of!

PR #56428, opened 2024-02-08
Add links and fix link interpolation

nicklathe:
This PR adds appropriate links to LTI integration portal and email, as well as fixes the string interpolation to use markdown, which is safer.

Continuing to debug why the email mailer isn't including links, but unfortunately the only way to test it is in prod. I'd like to merge this, and if the email template still doesn't contain the links as expected, I'll continue to debug.

## PR Checklist:

- [ ] Code is well-commented


dashboard/app/views/lti/v1/integration_status.html.haml
```diff
@@ -8,16 +8,18 @@
               %h1= t('lti.integration.success_header')
           .row
             .span10
-              %p= t('lti.integration.success_body').html_safe
-              %p= t('lti.integration.guide_instructions').html_safe
+              %p= t('lti.integration.success_body')
+              %p
+                != I18n.t('lti.integration.guide_instructions', lms_guides: 'https://support.code.org/hc/en-us/articles/23621907533965-Install-Code-org-Integrations-for-your-Learning-Management-System', contact_us: 'https://code.org/contact', markdown: true)
```
carl-codeorg: For the reused links like `contact_us`, you could put them into [shared_constants](https://github.com/code-dot-org/code-dot-org/blob/staging/lib/cdo/shared_constants.rb#L624-L632). For the CPA parent mailer we had a bunch of reused links that actually had to be changed, and updating them all individually can get messy.

carl-codeorg (approved):

dashboard/app/views/lti/v1/integration_status.html.haml
nicklathe: Good call, I moved these to shared_constants.

PR #56429, opened 2024-02-08
Always reporting language in GA

mgc1194:
This PR does two things:
1. Removes a condition that prevents logging user language to Google Analytics in some cases.
2. Refactores the logic used in the `locale_helper` to provide accurately accepted locales and languages supported in our system.

Under the removed condition, a list of accepted languages is requested from the HTTP header using varnish in some legacy code:
https://github.com/code-dot-org/code-dot-org/blob/cfde1a3ef544cb8855949ef2d20abb64f18ebef7/dashboard/app/helpers/locale_helper.rb#L45-L59

The `ACCEPT_LANGUAGE` header looks as follows:
Screenshot 2024-02-07 at 15 44 58

When the [locale language](https://github.com/code-dot-org/code-dot-org/blob/cfde1a3ef544cb8855949ef2d20abb64f18ebef7/dashboard/app/helpers/locale_helper.rb#L15-L19) coincides with the top language returned in the header, no value is assigned to the `selected_language_dim` resulting into a [(not set)](https://support.google.com/analytics/answer/13504892?hl=en#zippy=%2Cin-this-article) value in Google Analytics. 

By changing the logic, we are attempting to reduce or eliminate the number of (not set) values, decreasing the amount of noise in Google Analytics reporting data.

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/dashboard/app/helpers/locale_helper.rb
+++ b/dashboard/app/helpers/locale_helper.rb
@@ -1,3 +1,5 @@
+require_relative '../../lib/dashboard_languages'
+
 module LocaleHelper
   # Symbol of best valid locale code to be used for I18n.locale.
   def locale
@@ -42,25 +44,18 @@ module LocaleHelper
     options
   end
 
-  # Parses and ranks locale code strings from the Accept-Language header.
+  # Returns an Array of supported locale codes in dashboard.
   def accepted_locales
-    header = request.env.fetch('HTTP_X_VARNISH_ACCEPT_LANGUAGE', '')
-    begin
-      locale_codes = header.split(',').map do |entry|
-        locale, weight = entry.split(';')
-        weight = (weight || 'q=1').split('=')[1].to_f
-        [locale, weight]
-      end
-      locale_codes.sort_by {|_, weight| -weight}.map {|locale, _| locale.strip}
-    rescue
-      Logger.warn "Error parsing Accept-Language header: #{header}"
-      []
+    locale_codes = DashboardLanguages.table.select(:locale_s).map do |locale|
+      locale[:locale_s]
     end
+    locale_codes
   end
 
   # Strips regions off of accepted_locales.
   def accepted_languages
-    accepted_locales.map {|locale| locale.split('-')[0]}
+    accepted_languages = accepted_locales.map {|locale| locale.split('-')[0]}
+    accepted_languages.uniq
   end
 
   # Looks up a localized string driven by a database value.
--- a/dashboard/app/helpers/locale_helper.rb
+++ b/dashboard/app/helpers/locale_helper.rb
@@ -44,12 +44,9 @@ module LocaleHelper
     options
   end
 
-  # Returns an Array of supported locale codes in dashboard.
+  # Returns an Array of supported locale codes.
   def accepted_locales
-    locale_codes = DashboardLanguages.table.select(:locale_s).map do |locale|
-      locale[:locale_s]
-    end
-    locale_codes
+    I18n.available_locales.map(&:to_s)
   end
 
   # Strips regions off of accepted_locales.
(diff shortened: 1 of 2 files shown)


pablo-code-org:
Whats the maximum number of languages you may have?

Now that I read it, I think its also a prevention mechanism against malicious attacks. For example, someone could get different strings and killed the cardinality value on that dimension. 

I am being catastrophic here...

pablo-code-org:
I am trying to do some research and see if there is a solution that allows you to have accurate data but with the protection for the dimension value 

The only reason this rings in my head is because language is something more about user settings and reading sometimes a parameter from the URL (I haven't read our implementation)... digging into the code, sharing the ideas here  so you can see why the concern

daynew:
> I am trying to do some research and see if there is a solution that allows you to have accurate data but with the protection for the dimension value
> 
> The only reason this rings in my head is because language is something more about user settings and reading sometimes a parameter from the URL (I haven't read our implementation)... digging into the code, sharing the ideas here so you can see why the concern

I think it should be safe because `language` is always filtered to known values. The filtering is done in https://github.com/code-dot-org/code-dot-org/blob/4abada8adfabe99289e55dab76d474c78b10e88c/shared/middleware/varnish_environment.rb#L34

daynew (approved):

sureshc:
Not sure if this helps, but keep in mind that we have logic that executes in our Content Distribution Network ([a CloudFront Function](https://github.com/code-dot-org/code-dot-org/blob/staging/aws/cloudformation/accept-language.js.erb)) that normalizes the Accept-Language HTTP request header, so any logic in our Ruby web application servers that inspects this header should be seeing a relatively small set of predictable values.

dashboard/app/helpers/locale_helper.rb
```diff
@@ -1,3 +1,5 @@
+require_relative '../../lib/dashboard_languages'
+
```
mgc1194: I will remove this line if using `I18n.available_locales.map(&:to_s)`. Check [this commit](https://github.com/code-dot-org/code-dot-org/pull/56429/commits/133bc876218a58c06e5cff3d0e9ff30969371d0f) where I was using `DashboardLanguages.table.select(:locale_s).map`.

dashboard/app/helpers/locale_helper.rb
```diff
@@ -42,25 +44,15 @@ def options_for_locale_code_select
...
-      []
-    end
+    I18n.available_locales.map(&:to_s)
   end
 
   # Strips regions off of accepted_locales.
   def accepted_languages
-    accepted_locales.map {|locale| locale.split('-')[0]}
+    accepted_languages = accepted_locales.map {|locale| locale.split('-')[0]}
+    accepted_languages.uniq
```
artem-vavilov: I would suggest modifying `#language` a little bit and using it in `#accepted_languages` to  point out on the dependency between the two methods.
```ruby
def language(locale_code = locale)
  locale_code.to_s.split('-').first
end
```

```suggestion
    accepted_languages = accepted_locales.map(&method(:language)).uniq
```

artem-vavilov (approved):
Nice job 👍 Just a small suggestion

dashboard/app/helpers/locale_helper.rb
artem-vavilov: Memoize it once to not collect these locales on each call
```suggestion
    @accepted_languages ||= accepted_locales.map(&method(:language)).uniq
```

mgc1194: Thanks! I looked at this [memoization article](https://www.honeybadger.io/blog/ruby-rails-memoization/), which was useful for understanding how it works in the background.

PR #56431, opened 2024-02-08
Provision reader and writer SQL Users via CloudFormation

sureshc:
Provision the application reader and writer SQL users and grant them permissions via CloudFormation. These two new users (`writer` and `reader`) are NOT used by our web application server, nor any of our cron jobs. They are only used by RDS Proxy and the `bin/mysql-client-*` shell scripts.

RDS Proxy requires that if we want to use standard MySQL username/password authentication to connect to it, and in turn, for it to connect to our MySQL database, that we pass it AWS Secrets specifying those credentials, and those credentials. This requires us to provision the SQL users as well, whose credentials we're passing.

## Testing story

Successfully provisioned a new adhoc on this feature branch.

Successfully provisioned an adhoc from the staging branch, then merged in this feature branch and verified that it created the 2 SQL users and updated the RDS Proxies. Then reverted the commit and verified that it deleted the SQL users.

## PR Checklist:

- [ ] Code is well-commented


aws/cloudformation/components/database.yml.erb
```diff
@@ -53,7 +53,7 @@
       Description: !Sub "Database credentials for ${AWS::StackName} writer user."
 <% if database -%>
       GenerateSecretString:
-        SecretStringTemplate: !Sub '{"username": "application-writer"}'
+        SecretStringTemplate: '{"username": "writer"}'
```
cat5inthecradle, line 56: nothing is using these old usernames?

cat5inthecradle (approved):

aws/cloudformation/components/database.yml.erb
sureshc: The only components that utilize the SQL login credentials provisioned in this template are RDS Proxy and the `bin/mysql-client-` scripts. The `mysql-client-` reader & writer scripts currently do not work because the passwords generated by CloudFormation when it provisioned the AWS Secrets Manager Secret Resources in this template are different from the actual passwords of the `readonly` and `application-writer` SQL users.

Deploying this change will NOT change the `readonly` and `application-writer` SQL users that our web application servers and cron jobs utilize, nor does it change the Chef Attributes and the `CDO.` configuration settings where our Ruby code retrieves those credentials.

It will change:

- Anyone connecting to the Reporting RDS Proxy endpoint as the user `readonly` will be rejected. They'll either need the new `reader` credentials, or to use the `mysql-client-dashboard-reporting` script
- All of the `bin/mysql-client-*` scripts will now work. This the only Ruby code that utilizes Stack-specific Secrets to retrieve database connection credentials.

PR #56433, opened 2024-02-08
Add Initial Curriculum Recommender files with a test recommender

TurnerRiley:
This PR sets up the general framework that the curriculum recommenders will be built in. It creates a test recommender that utilizies every scoring question, a fake scoring framework, and fake curricula to test each scoring question. Here's the general format of how the files are/will be used:
1. A file calls one of the [recommender functions](https://github.com/code-dot-org/code-dot-org/pull/56433/files#diff-2224441e2997a6a0fab65ea4daecb046890777d103de7ff245689da1be5a6e0fR16-R51) in `curriculumRecommender.js` and is responsible for passing in any curricula the recommender should consider and the desired property values to match to (e.g. we want to show a user other curricula that have a duration of "month").
2. For each curriculum passed in, the recommender scores it based on the questions associated with that recommender
   a. For a given question, if the curriculum satisfies the condition of it (e.g. ["Does the curriculum have any specified school subjects?"](https://github.com/code-dot-org/code-dot-org/pull/56433/files#diff-2224441e2997a6a0fab65ea4daecb046890777d103de7ff245689da1be5a6e0fR79-R111)), it will receive a higher score according to that recommender's [scoring framework](https://github.com/code-dot-org/code-dot-org/pull/56433/files#diff-def804592183403d1b188f0cff251bcf68adab28e4948ee9f74994b52edcbd31R10-R19)
3. Once each curriculum is scored, they are then sorted in an array in descending order by score. That array is returned so the file that uses it can grab the first n curricula they want to show and those will be the most-recommended ones.

## Links
Jira ticket for creating the recommender file with a test recommender: [here](https://codedotorg.atlassian.net/jira/software/c/projects/ACQ/boards/64?assignee=60d62161f65054006980bd71&selectedIssue=ACQ-1374)
Jira ticket for creating the constants file with the fake scoring framework: [here](https://codedotorg.atlassian.net/jira/software/c/projects/ACQ/boards/64?assignee=60d62161f65054006980bd71&selectedIssue=ACQ-1376)
Jira ticket for creating a recommender test file: [here](https://codedotorg.atlassian.net/jira/software/c/projects/ACQ/boards/64?assignee=60d62161f65054006980bd71&selectedIssue=ACQ-1375)

## Testing story
I wrote unit tests with fake curricula to isolate each field and test that each scoring question was working. I modeled creating the fake curricula after the [Curriculum Catalog test helper file](https://github.com/code-dot-org/code-dot-org/blob/staging/apps/test/unit/templates/curriculumCatalog/CurriculumCatalogTestHelper.jsx) (but removed fields that didn't affect scoring, such as `image`).

## Follow-up work
The next step after this will be setting up the first actual recommender: the Similar Recommender for the Curriculum Catalog page.

## Caching
Once I start working on implementing the recommender into the Curriculum Catalog, I'll also start investigating ways I can potentially cache results to avoid duplicate queries to the recommender (tracked in [this Jira ticket](https://codedotorg.atlassian.net/jira/software/c/projects/ACQ/boards/64?assignee=60d62161f65054006980bd71&selectedIssue=ACQ-1398)).


apps/src/util/curriculumRecommender/curriculumRecommender.js
```diff
@@ -0,0 +1,130 @@
...
+      return b[1] - a[1];
+    }
+    if (a[0].is_featured !== b[0].is_featured) {
+      return a[0].is_featured ? -1 : 1;
+    }
+    if (a[0].published_date === b[0].published_date) {
+      return 0;
+    } else {
+      return a[0].published_date > b[0].published_date ? -1 : 1;
+    }
```
pablo-code-org: you can simplicity this
```
    return new Date(b[0].published_date) - new Date(a[0].published_date);
```

apps/src/util/curriculumRecommender/curriculumRecommender.js
```diff
@@ -0,0 +1,130 @@
...
+  const desiredTopics = csTopics.split(',');
+  let total = 0;
+
+  curriculumTopics?.forEach(currTopic => {
+    desiredTopics?.forEach(desTopic => {
+      total +=
+        currTopic === desTopic ? scoring_framework['hasDesiredTopics'] : 0;
+    });
+  });
+  return total;
```
pablo-code-org:
```
  return curriculumTopics.reduce((total, currTopic) => 
    total + (desiredTopics.has(currTopic) ? scoringFramework['hasDesiredTopics'] : 0), 0);
};
```

apps/src/util/curriculumRecommender/curriculumRecommender.js
```diff
@@ -0,0 +1,161 @@
...
+  curricula,
+  duration,
+  marketingInitiative,
+  schoolSubjects,
+  csTopics
+) => {
+  const curriculaScores = [];
+  curricula.forEach(curriculum => {
+    let score = 0;
+    score += hasDesiredDuration(FAKE_RECOMMENDER_SCORING, curriculum, duration);
```
bethanyaconnor: I assume this scoring is temporary? How will this be set in the future?

TurnerRiley: I was thinking it wouldn't be temporary since `getTestRecommendations` will only ever use the test scoring system. Then future recommenders will have their own function that only uses their respective scoring framework.

bethanyaconnor: ah got it -- so this function will be re-implemented for each scoring system, but the scoring functions below will be shared??

TurnerRiley: Yep! Each scoring system uses a different subset of those scoring functions so each one will just call the ones relevant to it

bethanyaconnor: got it got it. Thanks for clarifying!

bethanyaconnor (approved):
LGTM!

apps/src/util/curriculumRecommender/curriculumRecommender.js
```diff
@@ -0,0 +1,173 @@
...
+  .year(now.year() - 1);
+const twoYearsAgo = moment()
+  .utc()
+  .year(now.year() - 2);
+
+/*
+ * Curriculum recommenders: Each recommender receives an array of all curricula to consider and any preferences it should prioritize. The given
+ * recommender will score each curriculum based on the questions associated with that recommender. If a curriculum satisfies the conditions of a
+ * question, it receives points based on that recommender's scoring framework found in 'curriculumRecommenderConstants.js'. Once all curricula
+ * have been scored, the recommender returns an array of the passed-in curricula sorted in descending order by score (ties are broken by
```
hannahbergam, line 14: Love this attention to detail and the comment to help me understand🤩

apps/src/util/curriculumRecommender/curriculumRecommender.js
```diff
@@ -0,0 +1,173 @@
...
+    score += hasImportantButNotDesiredTopic(
+      FAKE_RECOMMENDER_SCORING,
+      curriculum,
+      csTopics
+    );
+    score += hasDesiredTopics(FAKE_RECOMMENDER_SCORING, curriculum, csTopics);
+    score += howRecentlyPublished(FAKE_RECOMMENDER_SCORING, curriculum);
+    curriculaScores.push([curriculum, score]);
+  });
+  console.log('SEPARATOR');
```
hannahbergam: Some console logs here:)

apps/src/util/curriculumRecommender/curriculumRecommenderConstants.js
```diff
@@ -0,0 +1,19 @@
+// Topics we care about promoting more than others
+export const IMPORTANT_TOPICS = [
+  'artificial_intelligence',
+  'physical_computing',
+];
+
+export const UTC_PUBLISHED_DATE_FORMAT = 'YYYY-MM-DD hh:mm:ss UTC';
+
+// Test recommender scoring framework.
+export const FAKE_RECOMMENDER_SCORING = {
```
hannahbergam: I know this is a bother, but is there any way we could rename this something like TEST_RECOMMENDER_SCORING?

apps/test/unit/util/curriculumRecommenderTest.js
```diff
@@ -0,0 +1,175 @@
...
+  DURATION_TEST_COURSES,
+  MARKETING_INIT_TEST_COURSES,
+  SCHOOL_SUBJECT_TEST_COURSES,
+  TOPICS_TEST_COURSES,
+  PUBLISHED_DATE_TEST_COURSES,
+} from './curriculumRecommenderFakeCurricula';
+import {IMPORTANT_TOPICS} from '@cdo/apps/util/curriculumRecommender/curriculumRecommenderConstants';
+import {expect} from '../../util/reconfiguredChai';
+
+describe('testRecommender', () => {
```
hannahbergam, line 13: I'm LOVING all these tests!!!

hannahbergam (approved):
Awesome work!! A couple small nits

apps/src/util/curriculumRecommender/curriculumRecommender.js
TurnerRiley: Whoops! I removed them in a branch I started off this one but not the original, thank you!

apps/src/util/curriculumRecommender/curriculumRecommenderConstants.js
TurnerRiley: Not a bother at all, super open to renames! Yes I like this!

PR #56436, opened 2024-02-08
[Game Design] K-5 variable display blocks

ebeastlake:
This PR does the following:
- Adds `showVariable` and `hideVariable` blocks to the GameLabJr pool
- Adds corresponding `showVariable` and `hideVariable` commands to `p5lab/spritelab/commands/variableCommands.js`, which make calls to `CoreLibrary.js`
- Passes the `JSInterpreter` instance from `SpriteLab` to `CoreLibrary` so it can be used there to evaluate the value of a variable with a given name (see `getVariableValue`)
- Adds the functionality to `CoreLibrary` and `drawUtils` to add, remove, and draw "variable bubbles" (the tentative name I've chosen for variable displays)
- Extracts some location-related logic into its own `locationUtils` file (Note: This might be overkill, but I noticed some location-related logic for addSprite that I figured we'd want to replicate for variable bubbles and possibly other display elements in the future)

Images of the new variable blocks:
Screenshot 2024-02-14 at 7 39 25 PM
Screenshot 2024-02-14 at 7 39 38 PM

Current demo of blocks in use

https://github.com/code-dot-org/code-dot-org/assets/26844240/e6b383e5-aaba-4e36-953b-c6a252f70798

TODO:
- [x] Update video once toolbox variable issues are solved 
- [x] Ask @mikeharv whether we need to fix the flicker/blank screen caused by using the blank pin block (it looks like normally these blocks get a value by default?)
- [x] Ask about i vs. j/where I need to add missing field value

## PR Checklist:

- [ ] Code is well-commented


apps/src/p5lab/spritelab/CoreLibrary.js
```diff
@@ -410,10 +471,7 @@ export default class CoreLibrary {
       return;
     }
     let name = opts.name;
-    let location = opts.location || {x: 200, y: 200};
-    if (typeof location === 'function') {
-      location = location();
```
ebeastlake, line 415: @mikeharv Do you have any idea why this is here/when the location will be a function and not just an Object with the shape `{x, y}`?

apps/src/p5lab/spritelab/CoreLibrary.js
```diff
@@ -11,7 +12,7 @@ import {
 import msg from '@cdo/locale';
 
 export default class CoreLibrary {
-  constructor(p5) {
+  constructor(p5, jsInterpreter) {
```
ebeastlake, line 15: This is where it feels like there must be a better way to do it. I passed the `jsInterpreter` instance here to call `evaluateWatchExpression` below.

apps/src/p5lab/spritelab/CoreLibrary.js
mikeharv: Yes, but it's confusing that we are overloading the `location` variable like this. 

Lots of blocks call `addSprite()`. `opts.location` should be a function if it's defined. This could be a separate location block, or a field on the block, or neither. It might be clearer to write it like this:

```
const locationGenerator = opts.location || (() => ({x: 200, y: 200}));
const location = locationGenerator();
// ...
const sprite = this.p5.createSprite(location.x, location.y);
```

ebeastlake:
Move `actionCommands.js` to `worldCommands.js` 😄 

Edited to add: I moved them to a new `variableCommands.js` file. I'd slightly prefer the commands.js file they live in match their toolbox category 😅

apps/src/p5lab/spritelab/CoreLibrary.js
ebeastlake: When I use the pin block with the show/hide variable blocks, the location is an object. You're saying that's different for sprite locations? I'm torn on leaving the refactor as-is and converting the Sprite logic to what you have above, removing the `locationUtils` file entirely, and just duplicating it with the variable bubbles. Thoughts?

ebeastlake: I'm leaning towards leaving the `resolveLocation` the way it is because it works for everything, but 🤷‍♀️

apps/src/blockly/addons/cdoUtils.js
```diff
@@ -395,13 +395,21 @@ function simplifyBlockState(block, variableMap) {
...
   const result = {...block};
 
-  // For VAR fields, look up the name of the variable to use instead of the id.
-  if (block.fields && block.fields.VAR) {
-    result.fields.VAR = {
-      name: variableMap[block.fields.VAR.id] || '',
-      type: '',
-    };
-  }
+  // For variable fields, look up the name of the variable to use instead of the id.
```
mikeharv: I think this diff might go away if you merge in staging.

apps/src/p5lab/drawUtils.js
```diff
@@ -22,6 +22,39 @@ export function getTextWidth(p5, text, size) {
...
+ * The size of the bubble is dynamically calculated based on the width of the text and padding.
+ *
+ * @param {p5} p5 - The p5 instance used to draw the bubble.
+ * @param {number} x - The x-coordinate of the center of the bubble.
+ * @param {number} y - The y-coordinate of the center of the bubble.
+ * @param {string} text - The text to display inside the bubble. This should represent the variable's name and value.
+ */
+export function variableBubble(p5, x, y, text) {
+  const padding = 15;
+  const textSizeValue = 35; // Text size matches the text size used for title page subtitles
```
mikeharv: This feels too big to me. Not a blocker since it's easy to iterate on. FWIW, we use size 20 for speech bubbles.

apps/src/p5lab/locationUtils.js
```diff
@@ -0,0 +1,17 @@
...
+ * Resolves the location value from the provided input.
+ * The input can be either a location object directly, or a function that returns a location object.
+ * If no input is provided and a fallback location exists, use the fallback.
+ * Otherwise, use a default location object with x and y coordinates set to 200.
+ *
+ * @param {Object|Function} locationInput - The location object or a function returning the location object.
+ * @param {Object} [fallback] - Optional. Ex. The current location of an existing variable bubble, used as a fallback.
+ * @returns {Object} The resolved location object with x and y coordinates.
+ */
+export function resolveLocation(locationInput, fallback) {
```
mikeharv, line 11: The fallback here is an interesting idea and seems to handle the edge case well. I compared this with the behavior of attempting to "move" a sprite to new undefined location. I was pleased to see it behaves similarly.

image

apps/src/p5lab/spritelab/CoreLibrary.js
mikeharv: I don't have a better idea of how to accomplish this, so I think it works for now.

apps/src/p5lab/spritelab/CoreLibrary.js
```diff
@@ -107,6 +111,39 @@ export default class CoreLibrary {
...
+  getVariableValue(variableName) {
+    if (!this.jsInterpreter) {
+      console.error('JS Interpreter not set in CoreLibrary');
+      return;
+    }
+
+    try {
+      // TODO: Check for existence of this.REFERENCE_ERROR and others in result,
+      // Or move this logic into the JSInterpreter and guarantee that it always returns a string
+      // TODO: Decide on what to display for undefined variables
```
mikeharv: For other blocks (title screen, print, sprite say) we just show `''` for undefined values.

apps/src/p5lab/spritelab/CoreLibrary.js
```diff
@@ -107,6 +111,39 @@ export default class CoreLibrary {
...
+      const result = this.jsInterpreter.evaluateWatchExpression(variableName);
```
mikeharv, line 138: I strongly feel it would be worth getting the Blockly variable name at this point so we can avoid reference errors. Instead of passing `variableName`, we'd basically want to pass `Blockly.JavaScript.getVariableName(variableName)` instead. I've confirmed locally that this works:

**Without suggested change**
image


**With change**
image

apps/src/p5lab/spritelab/commands/variableCommands.js
```diff
@@ -0,0 +1,16 @@
+export const commands = {
+  hideVariable(nameArg) {
+    if (!nameArg) {
```
mikeharv: I wonder if we have to worry about the `!nameArg` case now that the field is part of the block. I don't think variable fields have the ability to return something falsey. Probably doesn't hurt to leave it just in case though!

dashboard/config/blocks/GamelabJr/gamelab_hideVariable.json
```diff
@@ -0,0 +1,20 @@
+{
+  "category": "Variables",
+  "config": {
+    "func": "hideVariable",
+    "blockText": "hide variable {VAR}",
+    "color": [
```
mikeharv: Nit: Could we remove this color property now (from both blocks)? The preview in the block editor on your branch was probably showing it teal because it was still using CDO Blockly. If you merge in staging, it should load http://localhost-studio.code.org:3000/pools/GamelabJr/blocks/gamelab_hideVariable/edit with Google Blockly and you'll be able to confirm that the style is sufficient.

mikeharv (commented):
I played around with this for a good while. It works great! I even made a little game:

https://github.com/code-dot-org/code-dot-org/assets/43474485/f0bc2950-716b-4fb4-97eb-c68757a6db85

My main subjective feedback is that we could use a smaller size since the canvas is going to feel crowded quickly, depending on the types of games the curriculum will guide students towards making. (We could wait for this feedback from them or err on this side now.) I found that using 20 for text size and 10 for padding was still very readable and resembled sprite speech bubbles. I think this makes more sense than aligning with the much older (and less used in lessons) title screen block.

My main objective feedback is that we should prevent reference errors for student variables that contain invalid characters. We _could_ save this for a follow-up, but I don't see any real reason to wait.

apps/src/p5lab/spritelab/CoreLibrary.js
molly-moen, line 138: for my information: what is the reason we need to use the js interpreter for this? Can you add a comment explaining why we need it?

apps/src/p5lab/spritelab/CoreLibrary.js
molly-moen: Agree that I think this works for now

apps/src/p5lab/spritelab/CoreLibrary.js
ebeastlake: I'm tracking this as a follow-up!

apps/src/p5lab/spritelab/CoreLibrary.js
ebeastlake: Yep, I got this working locally yesterday. I'll push it!

apps/src/p5lab/spritelab/CoreLibrary.js
mikeharv: That works!

apps/src/p5lab/spritelab/CoreLibrary.js
ebeastlake: Added
```
// Blockly does not execute code or track the runtime values of variables, so we need to
// evaluate the variable's value using the JSInterpreter.
```

Does that cover it?

dashboard/config/blocks/GamelabJr/gamelab_hideVariable.json
ebeastlake: Ah, good catch, thank you!

ebeastlake:
Link to diff based on feedback: https://github.com/code-dot-org/code-dot-org/pull/56436/files/9f00f950614a7ee9c325173abb1a1be9a3257f72..260e562ec0934496aff6c7ee7d9e76f8cb6710ee

apps/src/p5lab/spritelab/CoreLibrary.js
mikeharv: This solution works for me!

Apologies for missing your other question. Blocks that return a location do so either as a simple object value or as a function that returns the object value, e.g. `randomLocation`. In the case of the latter it's usually because we'd expect a potentially different position each time the code for a block is executed. I don't think it should be different for sprites and variable bubbles.

apps/src/p5lab/drawUtils.js
```diff
@@ -36,8 +36,8 @@ export function getTextWidth(p5, text, size) {
  * @param {string} text - The text to display inside the bubble. This should represent the variable's name and value.
  */
 export function variableBubble(p5, x, y, text) {
-  const padding = 15;
-  const textSizeValue = 35; // Text size matches the text size used for title page subtitles
+  const padding = 10;
+  const textSizeValue = 20;
```
mikeharv, line 40: 👍

apps/src/p5lab/spritelab/blocks.js
```diff
@@ -442,10 +442,11 @@ const customInputTypes = {
...
     generateCode(block, arg) {
       const id = block.getFieldValue(arg.name);
-      const variable = Blockly.getMainWorkspace()
+      const label = Blockly.getMainWorkspace()
         .getVariableMap()
-        .getVariableById(id);
-      return [`"${variable.name}"`];
+        .getVariableById(id).name;
+      const name = Blockly.JavaScript.getVariableName(id);
+      return [`"${label}"`, `"${name}"`];
```
mikeharv, line 449: This looks reasonable to me!

dashboard/config/blocks/GamelabJr/gamelab_hideVariable.json
```diff
@@ -17,4 +12,4 @@
     "inline": true,
     "style": "variable_blocks"
   }
-}
\ No newline at end of file
+}
```
mikeharv: Nit: This line should change back if you run `bundle exec rake seed:blocks` from /dashboard. You can also make these changes from the levelbuilder interface if you want to avoid this. I don't think it matters so much, but if we leave it like this, it will probably show up as an extra diff back to the original end-of-file somewhere in our deploy pipeline.

mikeharv: Here's a commit from my recent branch where I undid some of VSCode's changes by reseeding. You can see my editor also changed some other whitespace stuff, which doesn't apply to your new blocks.  https://github.com/code-dot-org/code-dot-org/commit/2a3947f53ccb948935eb47af3385a14617622e12

mikeharv (approved):
This looks good! 🎉

molly-moen (approved):
looks good to me!

PR #33181, opened 2020-02-18
refuse to sync in duplicates of existing script files

Hamms:
otherwise, if a script is updated such that its strings move to a different path in crowdin, we'll end up with both copies of the script getting synced to crowdin. This can cause conflicts down the line when we try to sync out.

We have a couple options for what to do in this situation:

1. Clear the entire directory before regenerating script files, to ensure that everything is an exact match
2. Detect when we're about to create a conflict and remove the old versions of the file
3. Detect when we're about to create a conflict and warn about it, so the problem can be resolved manually.

The problem with 1 and 2 is that although our scripts have lots of ability to manipulate the filesystem and make git commits, mirroring those changes onto crowdin is more complicated. We could use [the crowdin API](https://support.crowdin.com/api/delete-file/) to do this, but that's a significant dependency that I don't think is worth adding for just this issue.

I propose we move forward with solution 3 for now, and if it becomes a significant pain to manage duplicates manually we can reconsider.

## Links

- [jira](https://codedotorg.atlassian.net/browse/FND-928)

## Testing story

Manually tested by simulating an extra file:

```bash
[~/cdo (dont-sync-duplicate-script-files)]$ cp i18n/locales/source/course_content/2017/coursea-2017.json i18n/locales/source/course_content/2018
[~/cdo (dont-sync-duplicate-script-files)]$ bin/i18n/sync-in.rb
Preparing level content
Script "coursea-2017" wants to output strings to 2017/coursea-2017.json, but 2018/coursea-2017.json already exists
```

And then a second extra file:

```bash
[~/cdo (dont-sync-duplicate-script-files)]$ cp i18n/locales/source/course_content/2017/coursea-2017.json i18n/locales/source/course_content/2019
[~/cdo (dont-sync-duplicate-script-files)]$ bin/i18n/sync-in.rb
Preparing level content
Script "coursea-2017" wants to output strings to 2017/coursea-2017.json, but 2019/coursea-2017.json and 2018/coursea-2017.json already exists
```

# Reviewer Checklist:

- [ ] Tests provide adequate coverage
- [ ] Code is well-commented
- [ ] New features are translatable or updates will not break translations
- [ ] Relevant documentation has been added or updated
- [ ] User impact is well-understood and desirable
- [ ] Pull Request is labeled appropriately
- [ ] Follow-up work items (including potential tech debt) are tracked and linked

PR #56446, opened 2024-02-08
P20-688: Fix redaction of Dashboard/course_content

artem-vavilov:
When we detect that a unit path has changed we skip the process of that file for translation because we need to manually delete the old file from Crowdin and the git repo first to avoid i18n strings duplication.
This PR fixes the bug where we were still redacting these skipped files.

- jira ticket: [P20-688](https://codedotorg.atlassian.net/browse/P20-688)

Related PRs:
- https://github.com/code-dot-org/code-dot-org/pull/33181
- https://github.com/code-dot-org/code-dot-org/pull/56358

Note: This fix requires manual updating of the corrupted `original` files on the `i18n server`:
- /i18n/locales/original/course_content/2023/dance-ai-2023.json
- /i18n/locales/original/course_content/other/K5-OnlinePD.json
- /i18n/locales/original/course_content/other/k5-onlinepd-2019.json
- /i18n/locales/original/course_content/other/kodea-pd-2021.json
- /i18n/locales/original/course_content/other/outbreak.json

### DIFF
--- a/bin/i18n/resources/dashboard/course_content/sync_in.rb
+++ b/bin/i18n/resources/dashboard/course_content/sync_in.rb
@@ -16,25 +16,18 @@ module I18n
         class SyncIn < I18n::Utils::SyncInBase
           def process
             prepare_level_content
-            progress_bar.progress = 25
-
-            prepare_project_content
             progress_bar.progress = 50
 
+            prepare_project_content
+            progress_bar.progress = 75
+
             write_to_yml(VARIABLE_NAMES_TYPE, variable_strings)
-            progress_bar.progress = 65
+            progress_bar.progress = 90
 
             write_to_yml(PARAMETER_NAMES_TYPE, parameter_strings)
-            progress_bar.progress = 80
-
-            redact_level_content
           end
 
           private
-
-          def valid_source_file?(source_file_path)
-            !I18nScriptUtils.unit_directory_change?(I18N_SOURCE_DIR_PATH, source_file_path)
-          end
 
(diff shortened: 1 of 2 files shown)


bin/i18n/resources/dashboard/course_content/sync_in.rb
```diff
@@ -16,26 +16,19 @@ module CourseContent
...
 
             write_to_yml(PARAMETER_NAMES_TYPE, parameter_strings)
-            progress_bar.progress = 80
-
-            redact_level_content
           end
 
           private
 
-          def valid_source_file?(source_file_path)
```
daynew, line 35: I'm guessing you removed this method because now it is only used in one place. However, I think it could be helpful to keep the method name  because it can make the code a little bit more readable. It would also be nice if there the was some documentation in the code about why we are skipping a script.

daynew (approved):
Could you update this PR description with an overview of what is happening? Something like "When we detect that the level's path has changed we skip process that file for translation because someone need to manually delete the old path from Crowdin and the git repo. This PR fixes the bug where we were still trying to redact these skipped files."

bin/i18n/resources/dashboard/course_content/sync_in.rb
artem-vavilov: I'm afraid that on the contrary, it will make the code a little bit more difficult to understand because we need to search to check what "valid source file" means. But I'm not against this idea.
This method would make more sense if it consisted of several conditions.

PR #56447, opened 2024-02-08
ToneJS Prototype

sanchitmalhotra126:
Prototype replacing our player implementation with ToneJS. We initially started looking into this library for its pitch/time stretching features, but it has several others that we can take advantage of. Of course, there are some considerations and open questions that should be addressed if we choose to move forward this.

### Benefits
- In-built transport control: this simplifies our internal playback logic quite a bit as we no longer need to convert between playhead time and Web Audio time, and don't need to hand-manage sample playback as closely. This enables some cool features like looping, and skipping around mid-playback with minimal code required on our part.
- Sampler support: currently when playing chords and beat patterns, we manually schedule each individual sample. ToneJS includes a pretty robust sampler API we can use instead. This simplifies our code but it can also automatically pitch individual samples up/down to fill in any gaps in the sample list. I increased the range of our chord player in this PR to take advantage of this.
- Effects: I haven't yet gotten to playing around with these, but there appear to be a wide array of effects we can use.
- Pitch/time stretch support: also satisfies our initial criteria of needing a library that can shift sample pitch and time stretch.
- Code simplification/reduction: the core ToneJS API (in ToneJSPlayer.ts) can effectively replace all of SamplePlayer.ts, SoundPlayer.ts, and soundSub.js (nearly 500 lines replaced with under 200!)

### Drawbacks/Questions
- For some reason this library isn't playing nicely with our build system. It seems to be potentially an issue with CommonJS vs ES modules. In this PR it's instead just manually added to the DOM via a script tag (see lab2.html.haml)
- The pitch shift/time stretch algorithm doesn't sound great. I don't really know what an alternative is here but the further you stray from the original pitch and bpm, the worse the audio quality. This may just be what we have to deal with since we're building on web audio, but worth noting.
- Open question: does this perform any worse (or better) on low power devices? I don't know how much stress testing we've done with our current player as it is, but it would definitely be worth checking before moving forward with this.

Simple looping/skipping demo:

https://github.com/code-dot-org/code-dot-org/assets/85528507/56949ccc-a80a-4367-925e-9f611159d7b5


sanchitmalhotra126:
This PR is a continuation of #55858, which was mis-closed by the Git LFS migration (#55759).

### Previous Comments:


### Previous Reviews:

apps/src/music/player/ToneJSPlayer.ts
```diff
@@ -0,0 +1,193 @@
...
+  async loadSounds(sampleIds: string[], callbacks?: SoundLoadCallbacks) {
+    return this.soundCache.loadSounds(sampleIds, callbacks);
+  }
+
+  async loadInstrument(
+    instrumentName: string,
+    sampleMap: {[note: number]: string},
+    callbacks?: SoundLoadCallbacks
+  ) {
+    const urls: {[note: number]: AudioBuffer} = {};
```
pablo-code-org, line 76: question.. do you need to load the instrument if previously loaded?

Also, is this the only reference to the instrument object? I was thinking this could be a good idea
```
    if (this.samplers[instrumentName]) {
      this.samplers[instrumentName].dispose();
    }
```

apps/src/music/player/ToneJSPlayer.ts
```diff
@@ -0,0 +1,193 @@
...
+  }
+
+  async playSampleImmediately(sound: SampleEvent, onStop?: () => void) {
+    if (this.currentPreview) {
+      this.currentPreview.player.stop();
+    }
+
+    const buffer = await this.soundCache.loadSound(sound.id);
+    if (!buffer) {
+      console.log('not in cache');
```
pablo-code-org, line 97: Was this left by mistake? If not, can a more explicit message be added?

apps/src/music/player/ToneJSPlayer.ts
```diff
@@ -0,0 +1,193 @@
...
+        notes,
+        `+${Tone.Transport.toSeconds(transportTime)}`
+      );
+    });
+  }
+
+  cancelPreviews() {
+    if (this.currentPreview) {
+      this.currentPreview.player.stop();
+    }
```
pablo-code-org, line 139: Instead of calling this.currentPreview, consider wrapping it around a function that will wrap any disposing operations. 

Even if you decide to ignore the memory leak suggestion, it can still help you adding more features later like logging when issues happening here. 

```
  private disposeCurrentPreview() {
    if (this.currentPreview) {
      this.currentPreview.player.stop();
      this.currentPreview.player.dispose();
      this.currentPreview = null;
    }
  }
```

pablo-code-org (commented):
I just read ToneJSPlayer and added feedback for it mostly around some memory management concerns when disposing of instrument and player objects
But This is my unexperience with JS

- The other comment about a method to take down previews I think will be useful in the future

sanchitmalhotra126:
> I just read ToneJSPlayer and added feedback for it mostly around some memory management concerns when disposing of instrument and player objects But This is my unexperience with JS
> 
> * The other comment about a method to take down previews I think will be useful in the future

Hey @pablo-code-org, thanks for the review! This is still somewhat prototype code that I'm experimenting with. I'm currently doing some load testing to see if there are any negative performance impacts with using this library. If we find the performance acceptable however, I will go ahead and integrate this library and can address your comments in a subsequent PR.

PR #56457, opened 2024-02-08
Music: more intro updates

breville:
Some small updates to `/s/music-intro-2024` based on internal feedback:

- Hover text for the four small buttons in the workspace header.
- Hide the "Version: " caption in the small footer to avoid overflow.
- Add a callout arrow that points to the number inside the workspace copy of the repeat block.  (Points to the one inside the toolbox if there's none inside the workspace yet, which is common fallback behavior for this system due to the selectors we are using.)
- No longer incorrectly show an arrow in the top-left corner if the functions category had been opened and then closed again, and we were trying to point at the function call block inside of it.
- Remove bolded text from panels' text, since it looked clickable like the instructions' bold text, but wasn't.  
- Add modern heading support to panels.


breville:
Thanks to @jmkulwik and @amy-b for the feedback!

apps/src/panels/panels.module.scss
```diff
@@ -45,6 +46,13 @@
       padding: 20px;
       animation: fade-in 1.5s;
 
+      h1, h2, h3, h4 {
```
sanchitmalhotra126, line 49: Just flagging this for @levadadenys - wondering if we should include our typography styles in our existing markdown components

sanchitmalhotra126 (approved):
Nice!

apps/src/panels/panels.module.scss
levadadenys: @sanchitmalhotra126 @breville   Good question,
It depends on designs you have for screens you develop. From what I see here in styles - these h1-h4 will look different from what we have in typography therefore it's ok not to use typography.

One nit though(for future probably since it's already merged): for design system we decided to use `rem` units for sizings where possible instead of `px` and `em` (for example for font-size, marging, padding properties, etc)

PR #56458, opened 2024-02-08
Catch cdo blockly code generation error

molly-moen:
Similar to #56359, if we hit an error while generation code for CDO Blockly, catch the error, log it, and show a banner to the user that we are working on it.

## Testing story
Tested locally. I had a hard time reproing this error on CDO blockly, but verified that code generation still works for the happy path, and if I force an exception to be thrown it logs appropriately and shows the banner.

## PR Checklist:

- [ ] Code is well-commented


apps/src/blockly/cdoBlocklyWrapper.js
```diff
@@ -220,12 +224,28 @@ function initializeBlocklyWrapper(blocklyInstance) {
...
+        null,
+        !!opt_showHidden
+      );
+      code = strip(code);
+      getStore().dispatch(setFailedToGenerateCode(false));
+    } catch (e) {
+      // We only want to log the error once per failure since getWorkspaceCode
+      // gets called many times and the error will be the same every time.
+      if (!getStore().getState().blockly.failedToGenerateCode) {
+        getStore().dispatch(setFailedToGenerateCode(true));
```
pablo-code-org: Assign the result of getStore() to a constant variable at the beginning of the scope, I believe this might prevent  unnecessary recalculations.

apps/src/blockly/cdoBlocklyWrapper.js
```diff
@@ -220,12 +224,28 @@ function initializeBlocklyWrapper(blocklyInstance) {
...
+        MetricsReporter.logError({
+          event: 'CDO_BLOCKLY_GET_CODE_ERROR',
```
pablo-code-org: Nit: I usually recommend avoiding literal and make them constants
```
    const ERROR_EVENT_NAME = 'CDO_BLOCKLY_GET_CODE_ERROR';
        MetricsReporter.logError({
      event: ERROR_EVENT_NAME,
      errorMessage: e.message,
      stackTrace: e.stack,
    });
    ```
    
By keeping this style, one day you will run into a value you will need to change and remember this because it will make it easier :D It has happened to me a couple of times. 

I prefer all constant for metric events into a single file. They all reference it, that way, you have a way to know which events are being logged too.

mikeharv (approved):

apps/src/blockly/cdoBlocklyWrapper.js
molly-moen: Good idea, I did a little bit of refactoring here. @sanchitmalhotra126 I added a new file `metrics/events.ts` that starts collecting event names. Let me know if that seems like a reasonable place for this information!

sanchitmalhotra126: Seems fine to me!

PR #56464, opened 2024-02-09
Use DataStorageTable.columns in all controller methods instead of inferring from data

### LINKED ISSUE #56463: DatablockStorage.addColumn doesn't seem to work
The problem is that column names comes from a `DatablockStorage.loadTableAndColumns` call and our implementation is inferring the columns from the records (when it should be used DatablockStorageTables.columns field in the DB).

We should instead be getting the columns from a call to datablock_storage_controller.rb's `get_columns_for_table` 

See: https://github.com/code-dot-org/code-dot-org/blob/8039db32b60a94f47c44a39d237689510f8a4ef1/apps/src/storage/datablockStorage.js/#L233-L240

PR to fix this: https://github.com/code-dot-org/code-dot-org/pull/56464

snickell:
We have a schema for defining columns, `DataStorageTable.columns`, but our current controller code is a mix of using this schema item, and inferring the column names from the actual records. This PR will switch us all the way over to using the columns schema.

Fixes https://github.com/code-dot-org/code-dot-org/issues/56463

### DIFF
--- a/dashboard/app/controllers/datablock_storage_controller.rb
+++ b/dashboard/app/controllers/datablock_storage_controller.rb
@@ -46,9 +46,6 @@ class DatablockStorageController < ApplicationController
     table = DatablockStorageTable.where(channel_id: params[:channel_id], table_name: params[:table_name]).first_or_create
     table.create_records([record_json])
 
-    # FIXME: unfirebase, must check Table record to see if we should update the table.columns
-    # column based on this record_json's keys
-
     render json: record_json
   end
 
@@ -87,8 +84,11 @@ class DatablockStorageController < ApplicationController
   def update_record
     raise "record_json must be less than 4096 bytes" if params[:record_json].length > 4096
 
-    # FIXME: unfirebase, must check Table record to see if we should update the table.columns
-    # column based on this record_json's keys
+    # FIXME: datablock/use-column-defs
+    # When we update a record, we need to check the Table.columns field, and see if
+    # we need to add new columns that are contained in the records but not in  Table.columns yet
+    #
+    # Also, we should extract this logic and move it to DatablockStorageTable.update_record instead
 
     record = DatablockStorageRecord.find_by(channel_id: params[:channel_id], table_name: params[:table_name], record_id: params[:record_id])
     if record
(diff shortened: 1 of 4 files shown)


dashboard/app/controllers/datablock_storage_controller.rb
```diff
@@ -155,10 +154,8 @@ def add_column
     column_name = params[:column_name]
 
     table = DatablockStorageTable.find([params[:channel_id], params[:table_name]])
-    columns = JSON.parse(table.columns)
-    unless columns.include? column_name
-      columns << column_name
-      table.columns = columns.to_json
+    unless table.columns.include? column_name
+      table.columns << column_name
```
snickell, line 158: At some point we screwed up and started saving a serialized-JSON string to table.columns, instead of saving the straight JSON. The result was that columns was in our DB as a JSON string like `'["id","name"]'`, not as a JSON object like `["id","name"]`.

Because this PR tries to "get columns right", I've unwound all those mistakes, and now table.columns is an object, not a string of a JSON-ified object.

Sorry for the extra noise in the PR tho lol

dashboard/app/models/datablock_storage_record.rb
```diff
@@ -13,4 +13,5 @@
 #
 class DatablockStorageRecord < ApplicationRecord
   self.primary_keys = :channel_id, :table_name, :record_id
+  belongs_to :table, class_name: 'DatablockStorageTable', foreign_key: [:channel_id, :table_name]
```
snickell, line 16: This lets us do: `record.table`

dashboard/app/models/datablock_storage_table.rb
```diff
@@ -11,6 +11,8 @@
 #
 class DatablockStorageTable < ApplicationRecord
   self.primary_keys = :channel_id, :table_name
+  has_many :records, class_name: 'DatablockStorageRecord', foreign_key: [:channel_id, :table_name]
```
snickell, line 14: this lets us do: `table.records`

dashboard/app/models/datablock_storage_table.rb
```diff
@@ -11,6 +11,8 @@
...
+  after_initialize -> {self.columns ||= ['id']}, if: :new_record?
```
snickell, line 15: This sets table.columns = ['id] when you create a new table, we used to do this in the controller create_table before, but it would overwrite the existing columns if. you tried to create_table with an existing column name

PR #55240, opened 2023-11-30
Updates to admin approval form

bethanyaconnor:
Updates to the admin approval form (see [spec](https://docs.google.com/document/d/1uFHbZuP7xkNLIOG4FHNouTdYmnCuEB3JbG7SMC_XRj8/edit)). Most of the edits were deleting questions, hence the largely negative diff.

Not sure if it's helpful, but I did a quick screen capture of the form:

https://github.com/code-dot-org/code-dot-org/assets/46464143/7604a4e1-235b-42a3-bd5a-661fe3c05902

PR #55191, opened 2023-11-28
Updates to the teacher application for 24-25

bethanyaconnor:
Updates to the teacher application based on [this spec](https://docs.google.com/document/d/1jrmraQqFuCxEBOqpYuZpg01sg1H8PMeT09j8AyweKLw/edit#heading=h.oapsklkkcf8e) as well as removing the question about whether CS will replace another course as suggested by Dani in [this spec](https://docs.google.com/document/d/1uFHbZuP7xkNLIOG4FHNouTdYmnCuEB3JbG7SMC_XRj8/edit) (unfortunately, I can't link to the comment itself as it's a suggestion).

I'll probably have one more PR to address [this conversation](https://docs.google.com/document/d/1jrmraQqFuCxEBOqpYuZpg01sg1H8PMeT09j8AyweKLw/edit?disco=AAABBO8MS0I).

PR #56474, opened 2024-02-09
Clean up references to removed questions

bethanyaconnor:
In https://github.com/code-dot-org/code-dot-org/pull/55191 and https://github.com/code-dot-org/code-dot-org/pull/55240, we removed a handful of questions from the teacher application and admin form. One of those questions was whether or not this course would replace an existing course, but I missed removing it from the application dashboard.

I also removed some references to invoicing data as we also removed that.


TurnerRiley (approved):
💯

PR #56475, opened 2024-02-09
[Google Blockly] add behaviorId when missing, or generate undefined

mikeharv:
After flipping Sprite Lab to Google Blockly, we started to receive some logs of projects that could generate code. In some cases, such as https://studio.code.org/projects/spritelab/ySPCqU53PnQ7XdwdEOgaMwFM4SKa-DORO_TI8i3uNUs/view this was due to behavior definitions with names like `do something` and `do something2` which were missing behavior ids. 
This change would provide a failsafe where definitions that are missing ids (for any reason) will have them created just prior to code generation. We use randomly generated alphanumeric strings for these.
While I've haven't yet found a similar repro for a behavior get block, we can also introduce a similar failsafe, just in case. We don't want to randomly generate ids for these blocks because a call to a random function name wouldn't exist. Instead, we look up the definition block (always on the hidden workspace), and copy it from there. If for some reason we can't find one there, we'd instead generate the code `undefined`, which is a valid value for the blocks with behavior inputs. (This is the generated code if e.g. a `sprite begins` block doesn't have a connected behavior block.)

## Testing story

I copied the project sources and animations from S3 into the folders for a new project I created locally. 

https://github.com/code-dot-org/code-dot-org/assets/43474485/0ca3530b-624d-4500-ae03-8bdbdee707ef

Also fixes:
- https://studio.code.org/projects/spritelab/qt5zPEzxJg5MezzOgAjBrE-xm4hdge8SI_QZzYlpjGY/view

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/blockly/customBlocks/googleBlockly/spritelabBlocks.js
+++ b/apps/src/blockly/customBlocks/googleBlockly/spritelabBlocks.js
@@ -12,6 +12,7 @@ import {
   editButtonHandler,
   toolboxConfigurationSupportsEditButton,
 } from './proceduresBlocks';
+import {getAlphanumericId} from '@cdo/apps/utils';
 
 const INPUTS = {
   FLYOUT: 'flyout_input',
@@ -222,6 +223,11 @@ export const blocks = {
 
     const generator = Blockly.getGenerator();
     generator.behavior_definition = function (block) {
+      // If we don't have a behavior id, generate a random id.
+      // This ensures the hidden definition block will generate valid code.
+      if (!block.behaviorId) {
+        block.behaviorId = getAlphanumericId();
+      }
       // Define a procedure with a return value.
       const funcName = generator.nameDB_.getName(
         block.behaviorId,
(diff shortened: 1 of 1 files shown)


apps/src/blockly/customBlocks/googleBlockly/spritelabBlocks.js
```diff
@@ -293,6 +299,20 @@ export const blocks = {
       return null;
     };
     generator.gamelab_behavior_get = function () {
+      // If we don't have a behavior Id, find on the definition block.
+      if (!this.behaviorId) {
+        const procedureModelName = this.getProcedureModel().name;
```
molly-moen: could `getProcedureModel()` return `null`?

mikeharv: `getProcedureModel()` _can_ return `null` if it can't find a model.  This one was a bit hard to test because I couldn't find any case where there wasn't already a behavior id and a model. If we somehow got to this point with neither, I'm not honestly sure how we should attempt to recover.

molly-moen: Should we just default to `unknown` in this case? That way we won't throw a null exception here.

mikeharv: If there's no model, we could generate the code `'undefined'` similar to further down when return early with that.

molly-moen: yep that's what I meant, reviewing too quickly 😬

mikeharv: Sounds good - updated!

molly-moen (approved):

PR #56478, opened 2024-02-09
[Google Blockly] add behavior id when deserializing xml

mikeharv:
In old levels with behavior blocks explicitly set in toolbox XML with no behavior ids. We add them to project source XML if needed, but I overlooked uncategorized toolboxes with behaviors. This causes us to be unable to generated code at the `block.behaviorId` ends up being `null`:
image

We can fix this easily by calling our existing helper to add behavior ids during deserialization. Now the toolbox blocks have behavior ids and can be used as normal:

image

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/blockly/addons/cdoXml.js
+++ b/apps/src/blockly/addons/cdoXml.js
@@ -312,7 +312,7 @@ export function addNameToBlockFunctionCallBlock(blockElement) {
  *
  * @param {Element} blockElement - The XML element for a single block.
  */
-function addMissingBehaviorId(blockElement) {
+export function addMissingBehaviorId(blockElement, trace = 'local') {
   const blockType = blockElement.getAttribute('type');
   if (blockType === BLOCK_TYPES.behaviorGet) {
     const behaviorNameField =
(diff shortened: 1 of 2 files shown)


apps/src/blockly/addons/cdoXml.js
```diff
@@ -312,7 +312,7 @@ export function addNameToBlockFunctionCallBlock(blockElement) {
  *
  * @param {Element} blockElement - The XML element for a single block.
  */
-function addMissingBehaviorId(blockElement) {
+export function addMissingBehaviorId(blockElement, trace = 'local') {
```
molly-moen: what is the new parameter used for?

molly-moen (commented):
Looks good, question about the second parameter

apps/src/blockly/addons/cdoXml.js
mikeharv: debugging, sorry. didn't mean to commit

molly-moen (approved):
Nice!

mikeharv:
@molly-moen I think I found a better and more thorough solution for this problem. Please see below when you get a chance:
- https://github.com/code-dot-org/code-dot-org/pull/56483

PR #56483, opened 2024-02-09
[Google Blockly] Process toolbox XML

mikeharv:
After flipping Sprite Lab to mainline Blockly, we noticed some levels were having new issues due to XML behavior blocks in flyouts not being created as expected. Unfortunately, this is because we weren't processing toolbox block XML with the same rigor as sources (start blocks or saved student code).

For example, `unnamed` behavior blocks were showing up at https://studio.code.org/s/express-2018/lessons/35/levels/4?no_redirect=true
due to to XML like this:
```
<block type="gamelab_behavior_get">
  <title name="VAR">spinning right</title>
</block>
```
image

We can fix this and other similar issues by treating our toolbox XML with the same care given to sources.
After processing the xml, it looks how we'd expect, including `userCreated` and `behaviorId` attributes:
```
<block type="gamelab_behavior_get">
  <mutation userCreated="false" behaviorId="spinning right"/>
  <title name="VAR" id="spinning right">spinning right</title>
</block>
```
image

Conveniently, this should resolve other issues unrelated to the level shown above. For example, it also fixes an issue that was potentially going to be addressed with 
- https://github.com/code-dot-org/code-dot-org/pull/56478

For the level in the PR above, https://studio.code.org/s/express-2018/lessons/35/levels/4?no_redirect=true
we see that this branch fixes it too, but more cleanly. By updating the XML right away we don't need to rely on the mutator hooks to handle a missing behavior id.

image

## Deployment strategy

If this PR is approved, I will close this branch: https://github.com/code-dot-org/code-dot-org/pull/56478

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/blockly/addons/cdoUtils.js
+++ b/apps/src/blockly/addons/cdoUtils.js
@@ -21,7 +21,10 @@ import {
 } from './cdoSerializationHelpers';
 import {parseElement as parseXmlElement} from '../../xml';
 import * as blockUtils from '../../block_utils';
-import {getProjectXml} from '@cdo/apps/blockly/addons/cdoXml';
+import {
+  getProjectXml,
+  processIndividualBlock,
+} from '@cdo/apps/blockly/addons/cdoXml';
 
 /**
  * Loads blocks to a workspace.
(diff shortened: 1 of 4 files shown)


mikeharv:
Conveniently, this is also fixes an issue with the "teacher sandbox level" at: https://studio.code.org/levels/18623

This level can be visited locally from here: http://localhost-studio.code.org:3000/levels?name=teacherSandboxFish&level_type=&script_id=&owner_id=&commit=Search

**Before:**
image

**After:**
image

molly-moen (approved):
Nice find!

apps/src/blockly/addons/cdoUtils.js
```diff
@@ -456,3 +459,30 @@ export function appendSharedFunctions(startBlocksSource, functionsXml) {
...
+ * We also process block xml during domToBlockSpace, which is called to
+ * convert sources from XML to JSON.
+ *
+ * @param {string} toolboxString - The XML string representing toolbox data.
+ * @returns {string} The modified XML string after processing.
+ */
+export function processToolboxXml(toolboxString) {
+  const parser = new DOMParser();
+  const xmlDoc = parser.parseFromString(toolboxString, 'text/xml');
+  const xmlRoot = xmlDoc.documentElement;
```
pablo-code-org: Nit: adding parsing errors check
```
    if (xmlDoc.querySelector('parsererror')) {
      throw new Error('Error parsing XML');
    }

```

apps/src/blockly/addons/cdoUtils.js
```diff
@@ -456,3 +459,30 @@ export function appendSharedFunctions(startBlocksSource, functionsXml) {
...
+  const blocks = Array.from(xmlRoot.childNodes).filter(
+    node => node.nodeName === 'block'
+  );
+
+  blocks.forEach(block => {
+    processIndividualBlock(block);
+  });
+
```
pablo-code-org: Do you need to do the array conversion? Not sure if something more direct like this would work

```
    const blocks = xmlDoc.querySelectorAll('block'); // Simplify node filtering
    blocks.forEach(processIndividualBlock); // Process each block directly
```

mikeharv: Much less verbose! Thank you. :)

PR #56495, opened 2024-02-09
Ai rubric redesign rollout

Nokondi:
This PR removes version 1 of the AI rubrics and replaces it with the new redesign. Update includes the following:

- Removed all experiment flags
- Removed frontend code for original rubrics (did not delete v1 rubric component files in case these are needed for reference)
- Updated unit and UI tests for v2 rubric compatibility
- Made general updates to styling to better match v2 figma
- Set learning goal tips to display: none (did not delete in case we need them in the future)
- Removed status text from header (to be replaced in the future)

https://github.com/code-dot-org/code-dot-org/assets/5552007/d23a7aae-4b33-4a78-ac21-930880de3304

### Links
[Figma](https://www.figma.com/file/WBcs7BY9IC6Nq5rG4V1SWF/AI-in-Teacher-Tools?type=design&node-id=3587-4370&mode=design&t=gFCYHbMgkUG1kaEJ-0)
[Jira Ticket](https://codedotorg.atlassian.net/browse/AITT-437)

### Future Work
- Update AI evaluation styling
- Add summary page
- Add footer element


apps/src/templates/rubrics/EvidenceLevelsForTeachersV2.jsx
```diff
@@ -29,7 +29,7 @@ export default function EvidenceLevelsForTeachersV2({
   // around evidence level.
   const passFail = useMemo(() => {
     if (!!aiEvalInfo) {
-      if (understanding === INVALID_UNDERSTANDING) {
+      if (understanding !== INVALID_UNDERSTANDING) {
```
kobryan0619, line 32: I am sure this is intentional... but can you explain the reasoning for the change?

apps/src/templates/rubrics/RubricSettings.jsx
```diff
@@ -299,69 +172,44 @@ export default function RubricSettings({
...
+      <div className={style.settingsGroup}>
+        <Heading2>{i18n.aiAssessment()}</Heading2>
+        <div className={style.settingsContainers}>
+          <div className={style.runAiAllStatuses}>
+            <BodyTwoText className="uitest-eval-status-all-text">
+              <StrongText>{summaryText()}</StrongText>
             </BodyTwoText>
-            <BodyTwoText>{i18n.runAiAssessmentDescriptionAll()}</BodyTwoText>
+            {statusAllText() && (
+              <BodyTwoText>{statusAllText() || ''}</BodyTwoText>
```
kobryan0619: Can you explain why the `|| ''` is needed?  it seems as if `statusAllText()` returns null this `BodyTwoText` wouldn't show up and if if `statusAllText()` returns `''` that would be printed here anyway...

apps/src/templates/rubrics/rubrics.module.scss
```diff
@@ -332,25 +329,19 @@ button.rubricHeaderTab {
...
-  padding: 4px 8px;
+  padding: 3px 8px 0;
   justify-content: center;
   align-items: center;
   gap: 4px;
-
   border-radius: 30px;
-  border: 1px solid $product_information_default;
-  background: $lightest_cyan;
+  background: #3cfff8;
```
kobryan0619: Just checking... should this be a color from our `color` file?

apps/src/templates/rubrics/rubrics.module.scss
```diff
@@ -678,13 +675,6 @@ i.infoAlertIcon {
 
 // Start of rubrics redesign
```
kobryan0619: Do you think this comment is still needed?

apps/test/unit/templates/rubrics/RubricContentTest.jsx
```diff
@@ -53,38 +54,56 @@ describe('RubricContent', () => {
...
+
   const processEventLoop = () => new Promise(resolve => setTimeout(resolve, 0));
 
-  it('shows learning goals with correct props when viewing student work on assessment level', () => {
-    const wrapper = shallow(
-      <RubricContent
-        {...defaultProps}
-        studentLevelInfo={{name: 'Grace Hopper', timeSpent: 706}}
-      />
+  it('displays LearningGoals component with correct props when viewing student work on assessment level', () => {
```
kobryan0619: I like this refactor!

dashboard/app/controllers/rubrics_controller.rb
```diff
@@ -118,7 +118,8 @@ def run_ai_evaluations_for_user
 
     script_level = @rubric.lesson.script_levels.find {|sl| sl.levels.include?(@rubric.level)}
 
-    is_ai_experiment_enabled = current_user && Experiment.enabled?(user: current_user, script: script_level.script, experiment_name: 'ai-rubrics')
+    is_ai_experiment_enabled = true
+    #current_user && Experiment.enabled?(user: current_user, script: script_level.script, experiment_name: 'ai-rubrics')
```
kobryan0619: Consider removing this comment?

dashboard/app/controllers/rubrics_controller.rb
```diff
@@ -145,7 +146,8 @@ def run_ai_evaluations_for_all
...
```
kobryan0619: Same thing - remove this comment?

dashboard/app/controllers/rubrics_controller.rb
```diff
@@ -185,7 +187,8 @@ def ai_evaluation_status_for_user
 
     script_level = @rubric.lesson.script_levels.find {|sl| sl.levels.include?(@rubric.level)}
 
-    is_ai_experiment_enabled = current_user && Experiment.enabled?(user: current_user, script: script_level&.script, experiment_name: 'ai-rubrics')
+    is_ai_experiment_enabled = true
+    #current_user && Experiment.enabled?(user: current_user, script: script_level&.script, experiment_name: 'ai-rubrics')
```
kobryan0619: One more

dashboard/app/controllers/rubrics_controller.rb
```diff
@@ -216,7 +219,8 @@ def ai_evaluation_status_for_all
...
```
kobryan0619: One more...

kobryan0619 (commented):
Left a few comments - this is massive but exciting to see the official change over to V2!

apps/src/templates/rubrics/EvidenceLevelsForTeachersV2.jsx
Nokondi: Had it the wrong way before, so it was not showing the correct style when no understanding was selected.

apps/src/templates/rubrics/RubricSettings.jsx
Nokondi: That's a good question. I believe that was there before I started on this, but you're right that it should be removed.

apps/src/templates/rubrics/rubrics.module.scss
Nokondi: That color isn't in the file yet. I suppose I could add it? Wasn't sure who managed that.

apps/src/templates/rubrics/rubrics.module.scss
Nokondi: Probably not

dashboard/app/controllers/rubrics_controller.rb
Nokondi: That's actually an error I need to fix! I commented that line out to enable ai analysis locally for testing.

Nokondi:
@kobryan0619 I think I addressed all of your feedback. Thanks!

kobryan0619 (approved):

PR #56498, opened 2024-02-10
Removing iframe to cbsnews

daynew:
On https://code.org/csforgood, we have an embedded iframe for a video hosted on cbsnews.com. This results in many cookies being added to the student's browser which we need to maintain a categorization for and try to block if the user doesn't want to be tracked. A much simpler solution is to remove the embedded video which is what this PR does.

## Testing
* Verified fix on http://localhost.code.org:3000/csforgood


TurnerRiley (approved):
LGTM!

PR #56515, opened 2024-02-12
[DESIGN2-92] Javalab - revert purple color back to original value

levadadenys:
### [DESIGN2-92] Javalab - revert purple color back to original value

* revert purple color back to original value for javalab

After https://github.com/code-dot-org/code-dot-org/pull/56281 we've found out that for Javalab new purple looks like too much. Decided to revert purple back to original value for Javalab for now. Also there're plans to make a restyle/rebrand of this page in the future.

Before:
Знімок екрана 2024-02-12 о 15 35 10

After:
Знімок екрана 2024-02-12 о 15 33 16

## Links
- jira ticket: [Design2-92](https://codedotorg.atlassian.net/browse/DESIGN2-92)

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/common-styles.module.scss
+++ b/apps/src/common-styles.module.scss
@@ -34,7 +34,8 @@
 //  More info here: https://github.com/code-dot-org/code-dot-org/pull/50895
 .purpleHeader {
   height: style-constants.$workspace-headers-height;
-  background-color: color.$purple;
+  // THIS IS AN OLD PURPLE COLOR. DO NOT USE THIS COLOR FOR NEW STYLES. ONLY USE IT FOR JAVALAB HEADER.
+  background-color: #7665A0;
   color: color.$white;
   overflow-y: hidden;
   -webkit-user-select: none;


apps/src/common-styles.module.scss
```diff
@@ -34,7 +34,8 @@
 //  More info here: https://github.com/code-dot-org/code-dot-org/pull/50895
 .purpleHeader {
   height: style-constants.$workspace-headers-height;
-  background-color: color.$purple;
+  // THIS IS AN OLD PURPLE COLOR. DO NOT USE THIS COLOR FOR NEW STYLES. ONLY USE IT FOR JAVALAB HEADER.
+  background-color: #7665A0;
```
sanchitmalhotra126, line 38: Worth making a constant (ex. `color.$deprecated-purple`) to make this extra clear? Not required though, if this is the only usage of it and we want to avoid it being used elsewhere.

sanchitmalhotra126 (approved):

apps/src/common-styles.module.scss
levadadenys: Yeah, that's the only place in the repo we use this deprecated-purple, so I think it's better to leave it as is

PR #56516, opened 2024-02-12
Customize PL congrats page

bethanyaconnor:
This PR:
- actually passes down whether the congrats page is for a PL Course! This effectively invisibly launches this page.
- shows new links for next steps after finishing self-paced PL

Screenshot 2024-02-14 at 11 31 19 AM


dashboard/app/models/course_offering.rb
```diff
@@ -445,6 +449,18 @@ def high_school_level?
     grade_levels_list.any? {|g| HIGH_SCHOOL_GRADES.include?(g)}
   end
 
+  def find_corresponding_offerings_for_pl_course
```
bethanyaconnor, line 452: This will also be useful for showing which courses to encourage users to assign to sections.

TurnerRiley (approved):

PR #56519, opened 2024-02-12
[Google Blockly] Cleanup - use BLOCK_TYPES constant

mikeharv:
While troubleshooting another issue, I came across a few instances of string literals being used to reference some common block types. We can use our constant for this instead.


molly-moen (approved):
🧹

PR #56526, opened 2024-02-12
Add Schools string for top nav

kelbyhawn:
Adds a new string "Schools" for use in the signed out top nav on code.org. Once this string is translated I will replace the Projects link with this Schools link in this PR: https://github.com/code-dot-org/code-dot-org/pull/56527

**Jira ticket:** [ACQ-1464](https://codedotorg.atlassian.net/browse/ACQ-1464)


bethanyaconnor (approved):
I assume we just need a full sync in/out? Said another way, do we have this string translated _somewhere_, just maybe not in this file?

kelbyhawn:
> I assume we just need a full sync in/out? Said another way, do we have this string translated _somewhere_, just maybe not in this file?

The word "schools" is in other strings in this file and other i18n files, but not by itself — does that influence things?

bethanyaconnor:
Yeah -- it means we can't rely on a translation already existing and will have to wait a bit longer to make the change. If it existed by itself, we'd have translations next week.

kelbyhawn:
> Yeah -- it means we can't rely on a translation already existing and will have to wait a bit longer to make the change. If it existed by itself, we'd have translations next week.

Got it, and bummer!

castro-jorge:
@kelbyhawn translations are in for 33 languages!

kelbyhawn:
> @kelbyhawn translations are in for 33 languages!

Woo hoo, thanks @castro-jorge!

kelbyhawn:
@mgc1194 ~do you know when should I be able to see this string locally? I'm still seeing the `translation missing: en-US.header_schools` error.~ Figured this out! I needed to add the string `header_schools` to the i18n gsheet 🙃

PR #56527, opened 2024-02-12
Replace Projects link with Schools link in signed out top nav on code.org

kelbyhawn:
Replaces the `Projects` link with `Schools` in the top navigation on code.org when signed out.
- Also added the `header_schools` string to the i18n gsheet so this shows up on Pegasus

_Note:_ I'm waiting on translations to go live before merging this.

**Related PR:**
- https://github.com/code-dot-org/code-dot-org/pull/56526

**Jira ticket:** [ACQ-1464](https://codedotorg.atlassian.net/browse/ACQ-1464)

----

## Before
Before

## After
After

## Hamburger

| Before | After | 
| ---- | ---- |
| Before | After |


lib/cdo/hamburger.rb
```diff
@@ -252,7 +252,7 @@ def self.get_header_contents(options)
       # When signed out, "Learn" will take all users to the student view of /courses.
       {title: I18n.t("#{loc_prefix}learn"), url: CDO.studio_url("/courses"), id: "header-learn"},
       {title: I18n.t("#{loc_prefix}teach"), url: CDO.code_org_url("/teach"), id: "header-teach"},
-      {title: I18n.t("#{loc_prefix}project_gallery"), url: CDO.studio_url("/projects/public"), id: "header-projects"},
+      {title: I18n.t("#{loc_prefix}schools"), url: CDO.code_org_url("/administrators"), id: "header-schools"},
       {title: I18n.t("#{loc_prefix}stats"), url: CDO.code_org_url("/promote"), id: "header-stats"},
       {title: I18n.t("#{loc_prefix}help_us"), url: CDO.code_org_url("/help"), id: "header-help"},
       {title: I18n.t("#{loc_prefix}incubator"), url: CDO.studio_url("/incubator"), id: "header-incubator"},
```
mgc1194, line 258: Changes look good, just a cleanup comment.
Could we have the same translation patterns for all links? It can be confusing as to what pattern to use if someone isn't familiar with this code.

Sometimes we do:
```
{title: I18n.t("#{loc_prefix}i18n_key"), ...},
```
and sometimes:
```

[...].each do |entry|
      entry[:title] = I18n.t("#{loc_prefix}#{entry[:title]}")
end
```

kelbyhawn: As someone who is new to this code I can confirm it's confusing! 😅 

I have a couple followups questions:
1. Do you have a preferred pattern?
2. Will this require a significant refactor of parts of the file? If yes, I'll make a ticket for it since I want to keep this PR simple in case we need to revert.

lib/test/test_hamburger.rb
```diff
@@ -184,6 +184,6 @@ def test_header_content_nobody_en
 
   def test_header_content_nobody_nonen
     contents = Hamburger.get_header_contents({user_type: nil, language: "fr"})
-    assert_includes_id contents, "header-projects"
+    assert_includes_id contents, "header-schools"
```
bethanyaconnor, line 187: Just checking if this makes sense in non-en? /administrators seems pretty US-centric?

dmcavoy: I was actually think today about how we should add a callout on that page to the international partner pages. 

However we should not special case the header navigation for that. I don't want to go back down that path now that they are unified.

dmcavoy: I added a ticket for Moshe to look at it

bethanyaconnor (approved):
This will be a fun set of eyes diffs 😅

kelbyhawn:
Closing this PR: O&A wants to use "Districts" instead of "Schools" here, so I'll make a new PR for this since we'll need to wait for translations again.

PR #56530, opened 2024-02-12
Add status column to featured_projects table

fisher-alice:
This PR adds the `status` column to the `featured_projects` table in anticipation of changes to the public project gallery.
This is part 1 of 3 steps described [here](https://codedotorg.atlassian.net/browse/LABS-497).

## Links
[jira](https://codedotorg.atlassian.net/browse/LABS-497)

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/dashboard/app/models/featured_project.rb
+++ b/dashboard/app/models/featured_project.rb
@@ -7,6 +7,7 @@
 #  featured_at    :datetime
 #  unfeatured_at  :datetime
 #  topic          :string(255)
+#  status         :string(255)      default("saved")
 #
 # Indexes
 #
--- a/dashboard/db/migrate/20240212234425_add_status_to_featured_projects.rb
+++ b/dashboard/db/migrate/20240212234425_add_status_to_featured_projects.rb
@@ -1,0 +1,5 @@
+class AddStatusToFeaturedProjects < ActiveRecord::Migration[6.1]
+  def change
+    add_column :featured_projects, :status, :string, default: "saved"
+  end
+end
--- a/dashboard/db/schema.rb
+++ b/dashboard/db/schema.rb
@@ -507,6 +507,7 @@ ActiveRecord::Schema.define(version: 2024_01_22_193916) do
     t.datetime "featured_at"
     t.datetime "unfeatured_at"
     t.string "topic"
+    t.string "status", default: "saved"
     t.index ["storage_app_id"], name: "index_featured_projects_on_storage_app_id", unique: true
     t.index ["topic"], name: "index_featured_projects_on_topic"
   end
--- a/dashboard/db/schema.rb
+++ b/dashboard/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2024_01_22_193916) do
+ActiveRecord::Schema.define(version: 2024_02_12_234425) do
 
   create_table "activities", id: :integer, charset: "utf8", collation: "utf8_unicode_ci", force: :cascade do |t|
     t.integer "user_id"


dashboard/app/models/featured_project.rb
```diff
@@ -7,6 +7,7 @@
 #  featured_at    :datetime
 #  unfeatured_at  :datetime
 #  topic          :string(255)
+#  status         :string(255)      default("saved")
```
bencodeorg, line 10: Will we ever want to do something like "return all projects with status = saved"? If so, we probably want an index on it, like we have on topic

bencodeorg (commented):
The little diagram in the Jira ticket is very helpful! One clarification -- what is the distinction between the "archived" and "featured" statuses and the current featured / unfeatured implementation?

dashboard/app/models/featured_project.rb
fisher-alice: Thanks. Do I just add the following at line 16?
```
index_featured_projects_on_status           (status)
```

fisher-alice:
> The little diagram in the Jira ticket is very helpful! One clarification -- what is the distinction between the "archived" and "featured" statuses and the current featured / unfeatured implementation?

"featured" status will denote projects that are currently featured at '/projects/public'.
"archived" status will denote projects that were featured at some point in the past and no longer featured.
"saved" status is the default. Instead of marking a project as "featured" immediately, an admin user will be able to mark a project as "saved". Then at "projects/featured" the admin user can update a project's status to "featured".
There will be 3 tables displayed at 'projects/featured'.

So adding the index on "status" is key. Thanks for that!

And @samantha-code created the diagram after we discussed the updates. Thanks Sam!

dashboard/app/models/featured_project.rb
bencodeorg: No, you need to update your migration to add an index, something like this:

https://github.com/code-dot-org/code-dot-org/blob/staging/dashboard/db/migrate/20210126021131_add_topic_to_featured_projects.rb

I believe you can rerun it by doing something like `bundle exec rake db migrate:down` then `bundle exec rails db migrate` again.

I'm not sure about the character set step though in this migration, that's a new one to me...looks like newer migrations aren't doing that, so I think you can ignore.

bencodeorg:
Maybe to clarify my question -- are the featured/archived states any different than what is currently derived from the featured_at / unfeatured_ at timestamps?

fisher-alice:
> Maybe to clarify my question -- are the featured/archived states any different than what is currently derived from the featured_at / unfeatured_ at timestamps?

Gotcha. The 'featured' state will continue to be derived as it currently is [here](https://github.com/code-dot-org/code-dot-org/blob/0939e83d482ef04c26714bb75cace8e3534c2011/dashboard/app/models/featured_project.rb#L24-L26).

In a follow-up PR, I will add a `save` method in `featured_projects_controller` and a new featured project's `featured_at` and `unfeatured_at` properties will be assigned `NULL`.

| status | featured_at | unfeatured_at |
| -- | -- | -- |
| 'saved' | NULL | NULL |
| 'featured' | most recently featured at time stamp | NULL |
| 'archived' | most recently featured at time stamp | most recently unfeatured at time stamp |

Does this answer your question? thanks!

bencodeorg:
It seems like your desired statuses can be achieved without a new column? Like you describe in the table in your comment, each state can be derived from the combination of featured_at / unfeatured_at. The new "saved" state is just a row with null for both timestamps?

fisher-alice:
> It seems like your desired statuses can be achieved without a new column? Like you describe in the table in your comment, each state can be derived from the combination of featured_at / unfeatured_at. The new "saved" state is just a row with null for both timestamps?

That's a great point! I'm glad you asked that question which led to the table above 😄 . 

You brought up indexing by status. I believe we'll still be able to filter featured projects into 3 groups depending on their `featured_at` and `unfeatured_at` properties. Could you explain indexing and why we need it for 'topic'?

fisher-alice:
Turns out adding `status` column is not needed when we add a third category of featured projects 'saved'.

PR #56532, opened 2024-02-13
Add ability to delete a project from the featured_projects table

fisher-alice:
This PR allows an admin user to delete a project from the `featured_projects` table displayed at ['projects/featured'](https://studio.code.org/projects/featured) (must have project validator permission to have access to this page). Currently there are 232 featured projects, and 550 projects in the ‘Archive of Previously Featured Projects’.

Many of the projects are no longer available (project deleted) and some of them were added unintentionally. There are currently too many to be able to manage effectively for an admin user.

Note that deleting a project from the `featured_projects` does NOT delete the project itself, and it is possible to re-add it to the `featured_projects` table.

Updates include:
- adding a 'delete' route for featured projects
- adding the 'Remove' option for projects listed in the 'Archive of previously featured projects' table at 'project/featured'.
- Deletes the record when button is clicked.

This is part 1 of 2 steps described [here](https://codedotorg.atlassian.net/browse/LABS-497).

Screencast video AFTER update: Note that button now displays 'Remove' instead of 'Delete' based on [this feedback](https://github.com/code-dot-org/code-dot-org/pull/56532#issuecomment-1942192942).

https://github.com/code-dot-org/code-dot-org/assets/107423305/3003393f-48a8-4cae-8dc4-7a732e3ecb0c

## Links

[jira](https://codedotorg.atlassian.net/browse/LABS-497)

## Testing story
I tested locally, and I also checked the `featured_projects` table using `mysql-client-dashboard-reader` to confirm the entry was removed from the table.

## Follow-up work
~~Follow-up work includes adding a confirmation modal when an admin users clicks on the 'Delete' button to delete a featured project. https://codedotorg.atlassian.net/browse/LABS-507~~
Asked @samantha-code about this and she prefers no confirmation dialog for 'Delete/Remove' button.

## PR Checklist:

- [ ] Code is well-commented


apps/src/templates/projects/FeaturedProjectsTable.jsx
```diff
@@ -151,6 +151,17 @@ const feature = (channel, publishedAt) => {
     .fail(handleFeatureFailure);
 };
 
+const onDelete = channel => {
+  var url = `/featured_projects/${channel}`;
+  $.ajax({
```
sanchitmalhotra126: nit: can we use `fetch` here? looks like we have other usages of `$.ajax` in this file though so also okay if you want to keep it consistent (and we can upgrade this file up later as a follow-up)

apps/src/templates/projects/FeaturedProjectsTable.jsx
```diff
@@ -151,6 +151,17 @@ const feature = (channel, publishedAt) => {
     .fail(handleFeatureFailure);
 };
 
+const onDelete = channel => {
+  var url = `/featured_projects/${channel}`;
```
sanchitmalhotra126: nit `const`

sanchitmalhotra126 (approved):
LGTM! I'm wondering if "Delete" feels a bit too intense since this doesn't actually delete the project from our system? But so long as users know what's happening it's probably fine

fisher-alice:
> LGTM! I'm wondering if "Delete" feels a bit too intense since this doesn't actually delete the project from our system? But so long as users know what's happening it's probably fine

Good point! I asked @samantha-code about this and she said 'Remove' sounds good. Thanks!

apps/src/templates/projects/FeaturedProjectsTable.jsx
fisher-alice: Sure! I'll wait till https://github.com/code-dot-org/code-dot-org/pull/56555 is merged since the authenticity token was required when `fetch` was used.

PR #56533, opened 2024-02-13
P20-706: Removes 'All' page, adds more page sizes, and adds CSV generation to permissions query page

wilkie:
Reported first in Slack: https://codedotorg.slack.com/archives/C04540KNGDQ/p1707240207180699

Essentially, we had a "All" button for the pagination... which failed due to the logic trying to use "All" as a number. However, it became clear that what they really needed was a larger pagination beyond 500. And they were working around it by manually forcing a page size of that magnitude.

Ultimately they want the data to create a CSV. Therefore, I asked if we could just... provide a download button.

This does the following:

* Removes the ill-advised "All" page option
* Adds page sizes in the thousands
* Adds a CSV download that will download a CSV representing the table on the current page

## Links

- jira ticket: [P20-706](https://codedotorg.atlassian.net/browse/P20-706)

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/dashboard/app/views/admin_users/permissions_form.html.haml
+++ b/dashboard/app/views/admin_users/permissions_form.html.haml
@@ -55,7 +55,8 @@
   %button.btn{type: 'submit'}
     %i.fa.fa-search
 -if @users_with_permission
-  = page_header 'users', @users_with_permission, permitted_params: [:permission]
+  - download_csv = link_to('Page as CSV', permissions_csv_path(params.permit([:permission, :page_size, :page, :search_term])), class: 'btn btn-default')
+  = page_header 'users', @users_with_permission, permitted_params: [:permission], extra_buttons: [download_csv]
   %table.table.table-hover.table-condensed.table-auto-width
     %thead
       %th ID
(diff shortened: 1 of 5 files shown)


dashboard/app/views/admin_users/permissions_form.html.haml
```diff
@@ -55,7 +55,8 @@
   %button.btn{type: 'submit'}
     %i.fa.fa-search
 -if @users_with_permission
-  = page_header 'users', @users_with_permission, permitted_params: [:permission]
+  - download_csv = link_to('Page as CSV', permissions_csv_path(params.permit([:permission, :page_size, :page, :search_term])), class: 'btn btn-default')
```
carl-codeorg, line 58: This button text might be more intuitive as "Download CSV"  or "Download Page as CSV".

carl-codeorg (approved):
LGTM, just one minor suggestion for the button text.

dashboard/app/views/admin_users/permissions_form.html.haml
wilkie: Yeah, I appreciate this comment... I had "Download Page as CSV" at one point and it felt comically long and would get in the way of adding any buttons to that header in the future. And just "Download CSV" might be surprising that the CSV is just the subset on the screen. It's not a user-facing button, so after thinking it through, I think I'll keep it as "Page as CSV" and not overthink it too much.

wilkie: Plus I'm a tad lazy to think too much about a feature for an internal tool. :)

Issue #56534, opened 2024-02-13
Allow marking applab projects as using datablock storage


snickell:
As part of the firebase migration for student data storage (https://github.com/orgs/code-dot-org/projects/4/), we'd like a way to progressively roll-out Datablock Storage.

A major implementation goal has been to keep the old firebase code paths present in the same codebase. For this issue, we just need a DB-persisted way for applab.js to determine if this project is stored in firebase, or if its been migrated to Datablock Storage.

Current idea:

1. add a "use_datablock_storage" boolean column to the projects table, default to false
2. add a DCDO field for "applab_percent_use_datablock_storage" to roll the dice over whether a new project is created in firebase or datablock.
3. Inject `use_datablock_storage` into AppOptions (how?)
4. increase percentage gradually, fixing any issues that come up, until we hit 100% datablock for new projects
5. convert old projects, until 100% of project rows have `use_datablock_storage=true`
6. Merge a cleanup PR that removes the `use_datablock_storage` column, and removes the firebase code paths

PR that implements part of this idea: https://github.com/code-dot-org/code-dot-org/pull/56536

PR #56536, opened 2024-02-13
Per-project `appOptions.useDatablockStorage` stored in DB

snickell:
1) Helps with progressive rollout of FirebaseStorage vs DatablockStorage.
2) Relevant to `APPLAB` and `GAMELAB`
3) Adds a new table `project_use_datablock_storages` with a `use_datablock_storage` column. This table should be removed by a follow-up PR after we've switched 100% to DatablockStorage.

Fixes https://github.com/code-dot-org/code-dot-org/issues/56534

### DIFF
--- a/apps/src/applab/applab.js
+++ b/apps/src/applab/applab.js
@@ -422,10 +422,10 @@ Applab.init = function (config) {
   Applab.channelId = config.channel;
 
   if (!!config.useDatablockStorage) {
-    console.warn('Initializing DATABLOCK_STORAGE');
+    console.log('Initializing DATABLOCK_STORAGE');
     Applab.storage = initStorage(DATABLOCK_STORAGE, {});
   } else {
-    console.warn('Initializing FIREBASE_STORAGE');
+    console.log('Initializing FIREBASE_STORAGE');
     Applab.storage = initStorage(FIREBASE_STORAGE, {
       // TODO: unfirebase
       channelId: config.channel,
--- a/apps/src/applab/applab.js
+++ b/apps/src/applab/applab.js
@@ -422,10 +422,10 @@ Applab.init = function (config) {
   Applab.channelId = config.channel;
 
   if (!!config.useDatablockStorage) {
-    console.log('Initializing DATABLOCK_STORAGE');
+    console.error('Initializing DATABLOCK_STORAGE');
     Applab.storage = initStorage(DATABLOCK_STORAGE, {});
   } else {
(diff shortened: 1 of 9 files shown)


bencodeorg:
The `project = Project.find_by_channel_id(channel_id); project&.use_datablock_storage` approach looks reasonable to me. I think part of the reason that you don't see much logic like that in our code base is that it's relatively recent (~1.5 years) that the project table got moved into dashboard, so I'm hoping we see more Rails-y stuff related to projects in the future.

Separately, have you chatted with infra about adding a column to a table of this size? I thought it was potentially a bit hairy to do so, docs are a little vague here:

https://dev.mysql.com/doc/refman/5.7/en/innodb-online-ddl-operations.html#online-ddl-column-operations

Says you can continue to write to the table during the migration, but "Data is reorganized substantially, making it an expensive operation." Wondering if having it in a separate table might make sense if this feature is used by a minority of projects (not sure if we see opportunities to expand its use in other cases)?

snickell:
> relatively recent (~1.5 years) that the project table got moved into dashboard

@bencodeorg woah! Where/how was it before? Was it in SQL but didn't have an ActiveRecord model? This is explains a lot of my wonderings why project didn't have relationships to other tables.

Appreciate the thinking on column adding vs adding a new table, I'm going to try this on a prod RDS clone and see what happens. I have a lot less experience with MySQL than I'd like. I think a separate table is a very worthy idea that I should investigate. 

In this case, eventually, every single Project will have `use_datablock_storage=true`, so a majority will use it, but I think a separate table still might make sense. After doing a rollout for new projects, we'd then run a huge migration back converting the many existing projects, and then we'd finally be able remove the column: another expensive operation according to https://dev.mysql.com/doc/refman/5.7/en/innodb-online-ddl-operations.html#online-ddl-column-operations

@code-dot-org/infrastructure any thoughts?

bencodeorg:
> @bencodeorg woah! Where/how was it before? Was it in SQL but didn't have an ActiveRecord model? This is explains a lot of my wonderings why project didn't have relationships to other tables.

Yep, it was in pegasus -- I just searched "pegasus dashboard projects" in Google Drive and came up with this, which at a skim looks like it has more details if you're interested!

https://docs.google.com/document/d/1-IdAlRCQFxPwcA4iUVy2Hav7_J6IOSoMc9Xt8t4Iluo/edit

snickell:
merged into `unfirebase`

PR #56546, opened 2024-02-13
Add `section_instructors` to redshift DMS repl. task

jordan-springer:
adding section_instructors table to redshift replication job.

left comments for suggested masking, I can't see that table in the app so not sure what all is needed to remove.

## PR Checklist:

- [ ] Code is well-commented


jordan-springer:
any other items I need to update to allow this table to pull through?

Hamms (commented):
Here are the columns in this table, for visibility:

```
+---------------+-------------+------+-----+---------+----------------+
| Field         | Type        | Null | Key | Default | Extra          |
+---------------+-------------+------+-----+---------+----------------+
| id            | bigint(20)  | NO   | PRI | NULL    | auto_increment |
| instructor_id | int(11)     | NO   | MUL | NULL    |                |
| section_id    | int(11)     | NO   | MUL | NULL    |                |
| invited_by_id | int(11)     | YES  | MUL | NULL    |                |
| deleted_at    | datetime    | YES  | MUL | NULL    |                |
| status        | int(11)     | NO   |     | NULL    |                |
| created_at    | datetime(6) | NO   |     | NULL    |                |
| updated_at    | datetime(6) | NO   |     | NULL    |                |
+---------------+-------------+------+-----+---------+----------------+
8 rows in set (0.01 sec)
```

I'm not familiar with this particular table or how it's being used in our application; it's entirely possible that I'm missing something subtle, but based on what I can see nothing in here looks like PII to me.

I think you can safely remove the `remove_columns` comment, at which point this will LGTM!

sureshc:
@code-dot-org/red-team Nearly every database table in dashboard has an ActiveRecord model class, and the comments in that class will show you the underlying database schema. Here’s how to view the schema on the `section_instructors` table https://github.com/code-dot-org/code-dot-org/blob/3796c462cd69e3e81f487fc6e17357c6e95d2570/dashboard/app/models/sections/section_instructor.rb#L5-L12

aws/dms/tasks.yml
```diff
@@ -58,6 +58,11 @@ cron:
 - dashboard.course_scripts
 - dashboard.survey_results
 - dashboard.sections
+- dashboard.section_instructors
+    # remove_columns:
```
sureshc: It appears that this table does not store any sensitive information (just links to entities in other tables), so we don’t need any `remove_columns` entries. Can you take that out? I can then get this change deployed to our data export system.

cat5inthecradle (approved):
Approved, but take a look at Suresh's comment and please remove those unnecessary comments from the yaml file.

I think you should have permission to click the Merge button yourself once you've done that. Let me know if that doesn't work for you.

Once merged, this should go out with the next DTP, and apply to the next overnight data migration.

cat5inthecradle:
LGTM! Ship it!

PR #56547, opened 2024-02-13
Update Self-paced PL CSC broken URLs

TurnerRiley:
Krystin [spotted](https://codedotorg.slack.com/archives/C04540KNGDQ/p1707851978128309) a broken link on the [CSC educate page](https://code.org/educate/csc#csc-modules) that sent the user to https://studio.code.org/s/self-paced-pl-csc. If the user is signed in, this redirects them automatically to https://studio.code.org/s/self-paced-pl-csc-2023. But if they're not signed in, then they see the sad bee. It seems that the url is just missing `-2023` so as a quick fix, this PR adds that to all uses of `/s/self-paced-pl-csc` to avoid any broken links.

## Links
Slack thread: [here](https://codedotorg.slack.com/archives/C04540KNGDQ/p1707851978128309)

## Testing story
CSC page:
csc_page

CSC faq:
csc_faq

Professional learning page carousel:
prof_learning_page

## Follow-up work
A more longterm fix that uses the most recent version rather than a hard-coded `-2023` will be trakced in [this Jira ticket](https://codedotorg.atlassian.net/browse/ACQ-1475).

### DIFF
--- a/pegasus/sites.v3/code.org/public/educate/csc.haml
+++ b/pegasus/sites.v3/code.org/public/educate/csc.haml
@@ -134,7 +134,7 @@ theme: responsive_full_width
         %h3.heading-sm
           =hoc_s(:self_paced_pl_modules_heading)
         %p=hoc_s(:self_paced_pl_modules_desc)
-        %a.link-button{href: CDO.studio_url("/s/self-paced-pl-csc")}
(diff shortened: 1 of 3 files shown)


kelbyhawn (approved):
Thanks for updating this!

PR #56551, opened 2024-02-13
Initial Music Mini Player Setup

sanchitmalhotra126:
Sets up a basic bare bones view of a "mini player" with most of the logic necessary to play a song given a list of channel IDs. This will be the basis of further work that @breville is doing to create the full mini player feature.

Also includes a few minor changes to other classes, namely allowing for our blockly workspace to be initialized in a "headless" mode (since we only need code execution capabilities for this), and tweaking Simple2Sequencer so we don't have to pass in the music library in the constructor.

### DIFF
--- a/apps/src/music/MiniPlayerView.tsx
+++ b/apps/src/music/MiniPlayerView.tsx
@@ -6,8 +6,6 @@ import Simple2Sequencer from './player/sequencer/Simple2Sequencer';
 import {RemoteSourcesStore, SourcesStore} from '../lab2/projects/SourcesStore';
 import {loadLibrary} from './utils/Loader';
 import MusicLibrary from './player/MusicLibrary';
-import {Triggers} from './constants';
-import {PlaybackEvent} from './player/interfaces/PlaybackEvent';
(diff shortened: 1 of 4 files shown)


apps/src/music/MiniPlayerView.tsx
```diff
@@ -0,0 +1,109 @@
...
+  }, [libraryName]);
+
+  useEffect(() => {
+    onMount();
+  }, [onMount]);
+
+  // This is the main function that is called when a song is played in the mini player
+  // Loads code from the server, compiles the song, executes it to generate events,
+  // and then plays the events.
+  // Optimization: cache code and/or compiled song after played once.
```
breville, line 45: Is this a potential optimization, or one already in place?

breville (approved):

apps/src/music/MiniPlayerView.tsx
sanchitmalhotra126: Oh this is a potential optimization, left a note here to add this as a follow-up. Currently we would always fetch sources for a project even if it's already been played.

PR #56555, opened 2024-02-13
Add delete request to HttpClient

fisher-alice:
This PR adds a send DELETE request function to `HttpClient`. I plan to use this new function in https://github.com/code-dot-org/code-dot-org/pull/56532. Other requests made in `FeaturedProjectsTable` use `ajax` and when I attempted to use `fetch` instead, I received a 422 (unprocessable entity) server error because the authenticity token was required. Thus, I wanted to use a request function from `HttpClient` but 'delete' was not yet available.

## Testing story
I tested locally in conjunction with changes in https://github.com/code-dot-org/code-dot-org/pull/56532 checking both DELETE and PUT requests.

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/util/HttpClient.ts
+++ b/apps/src/util/HttpClient.ts
@@ -74,12 +74,13 @@ async function fetchJson<ResponseType>(
 }
 
 /**
- * POST to the given endpoint. Adds the Rails authenticity
+ * o to the given endpoint. Adds the Rails authenticity
  * token if useAuthenticityToken is true.
  */
-async function post(
+async function sendRequest(
+  method: string,
   endpoint: string,
-  body: string,
+  body?: string,
   useAuthenticityToken = false,
   headers: Record<string, string> = {}
 ): Promise<Response> {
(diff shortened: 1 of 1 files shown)


apps/src/util/HttpClient.ts
```diff
@@ -74,12 +74,13 @@ async function fetchJson<ResponseType>(
 }
 
 /**
- * POST to the given endpoint. Adds the Rails authenticity
+ * o to the given endpoint. Adds the Rails authenticity
```
sanchitmalhotra126: nit: update comment - maybe "Sends a request to the given endpoint"?

sanchitmalhotra126 (approved):

apps/src/util/HttpClient.ts
fisher-alice: Oh yeah! Forgot to fix this - updated. Thanks!

fisher-alice: And thanks for pairing with me on this @sanchitmalhotra126 !

PR #56559, opened 2024-02-13
Put padlock icons in lesson headers where appropriate

etaderhold:
This change puts lock icons in the progress header for the non-numbered lessons: closed lock if the lesson is locked for all students in the section, open lock if the lesson is unlocked for any student in the section.

image

image

## Links
[JIRA](https://codedotorg.atlassian.net/browse/TEACH-858)

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/templates/sectionProgressV2/LessonProgressColumnHeader.jsx
+++ b/apps/src/templates/sectionProgressV2/LessonProgressColumnHeader.jsx
@@ -63,5 +63,5 @@ export default function LessonProgressColumnHeader({
 LessonProgressColumnHeader.propTypes = {
   lesson: PropTypes.object.isRequired,
   addExpandedLesson: PropTypes.func.isRequired,
-  allLocked: PropTypes.bool.isRequired,
+  allLocked: PropTypes.bool,
 };
--- a/apps/src/templates/sectionProgressV2/LessonProgressColumnHeader.jsx
+++ b/apps/src/templates/sectionProgressV2/LessonProgressColumnHeader.jsx
@@ -6,13 +6,16 @@ import FontAwesome from '../FontAwesome';
 import {lessonHasLevels} from '../progress/progressHelpers';
 import skeletonizeContent from '@cdo/apps/componentLibrary/skeletonize-content.module.scss';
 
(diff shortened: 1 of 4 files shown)


dju90:
This looks great Eric!

kobryan0619 (commented):
This code all looks good and makes sense to me, but I have a few questions after testing the branch out on my local machine:

- Which lessons did you test this out on locally?  When I tried assigning CSP Unit 1, which has an un-numbered lesson that is a locked set of levels first up in the series of lessons, I saw this in the progress view:

image

I saw something similar with CSD Unit 3 but with the last lesson in the set:

image

- I assume there is also follow up work for numbered lessons in the future, is that correct?

etaderhold:
> * Which lessons did you test this out on locally?

The screenshots in my original PR description are from CSD Unit 1.

> When I tried assigning CSP Unit 1, which has an un-numbered lesson that is a locked set of levels first up in the series of lessons, I saw this in the progress view:

My local environment encounters some `nil` errors trying to summarize the script structure for that unit. Strange.

> I saw something similar with CSD Unit 3 but with the last lesson in the set:

Here's what I see on CSD Unit 3. I suspect you need to rebuild the JavaScript after switching to my branch.

image

> * I assume there is also follow up work for numbered lessons in the future, is that correct?

That's a good question. For example on the CSP "Data" unit I see the current view has a lock replacing the "9" in between lessons 8 and 10, while my current implementation shows the 9 instead.

image
image

kobryan0619:
I am also seeing the number for the lesson in your implementation and comparing it to the Figma (below) - I am not really sure what it is supposed to be (I was kind of expecting it to be a lock... but looking at the Figma, I am less sure...)

image

[According to this Slack thread,](https://codedotorg.slack.com/archives/C06ERUABG01/p1706904098244879) I would think that there should be a lock in the lesson header in place of the number.  Perhaps this is already documented as follow up work.  I am assuming it is documented, or that this is part 1 of a few parts for this specific task.

kobryan0619 (approved):

dju90:
Yes, the lock icon should replace the lesson number in all cases, maintaining parity with the current view

PR #56560, opened 2024-02-14
Add new fields to Pop-up when missing due to LTI rostering (Pt 1)

juanmanzojr:
Incorporates the 'US State' and 'Gender' fields into the pop-up that appears when students, whose accounts were auto-generated through LTI rostering, log in. Student accounts which are not LTI and are missing the age field will still be prompted to enter their age but will not have access to the 'US State' and 'Gender' fields.
## LTI Student if age or US state is missing
Screenshot 2024-02-28 at 9 38 00 AM

## Non-LTI student with missing age
Screenshot 2024-02-16 at 10 50 33 AM

## Links

- jira ticket: [here](https://codedotorg.atlassian.net/browse/P20-697)

## Follow up Work
Front end changes done [here](https://github.com/code-dot-org/code-dot-org/pull/56802)
## PR Checklist:

- [ ] Code is well-commented


dashboard/app/views/layouts/application.html.haml
```diff
@@ -56,8 +56,8 @@
     - else
       - unless Rails.env.production?
         = render partial: 'layouts/environment_tag'
-    - if current_user && current_user.age.nil?
-      = render partial: 'layouts/age_interstitial'
+    - if current_user && (current_user.age.nil? || current_user.us_state.nil?)
```
carl-codeorg: Do we also need to do a `student?` check here?

dashboard/app/controllers/registrations_controller.rb
```diff
@@ -209,12 +209,18 @@ def begin_sign_up_params
...
   # checks since those depend on the age being set.
-  def set_age
+  def set_student_information
     return head(:forbidden) unless current_user
-    current_user.update(age: params[:user][:age]) if current_user.age.blank?
+    student_information = {}
+
+    student_information[:age] = params[:user][:age] if current_user.age.blank?
+    student_information[:us_state] = params[:user][:us_state] if current_user.us_state.blank?
+    student_information[:gender_student_input] = params[:user][:gender_student_input] if current_user.gender.blank?
```
carl-codeorg, line 221: Why do we only set these if the existing value is blank? Is there a reason we don't want to let them overwrite existing values?

juanmanzojr: No reason i can think of, I was following the previous method which only set the age if the exisitng value was blank

dashboard/app/views/layouts/application.html.haml
```diff
@@ -56,8 +56,8 @@
     - else
       - unless Rails.env.production?
         = render partial: 'layouts/environment_tag'
-    - if current_user && current_user.age.nil?
-      = render partial: 'layouts/age_interstitial'
+    - if current_user&.student? && (current_user.age.nil? || current_user.us_state.nil?)
```
carl-codeorg: According to the comment in the Jira ticket, the new data collection is supposed to be for LTI students only. You can use `Policies::Lti.lti?` to determine if a user is an LTI user. Since we probably still want to collect age for non-LTI users, you might have to add some conditional logic to the popup itself, and only show the fields for gender and state to LTI users.

dashboard/app/views/layouts/_student_information_interstitial.html.haml
```diff
@@ -0,0 +1,47 @@
...
+        .form-group
+          = f.label :age, t('signup_form.age')
+          = f.select :age, User::AGE_DROPDOWN_OPTIONS, include_blank: t('signup_form.select_age')
+        - if Policies::Lti.lti?(current_user)
+          .form-group
+            = f.label :us_state, t('activerecord.attributes.user.us_state')
+            = f.select :us_state, User::US_STATE_DROPDOWN_OPTIONS, include_blank: t('school_info.select_state')
+          .field
+            = f.label :gender_student_input, t('gender')
+            = f.text_field :gender_student_input, placeholder: t('type_here')
```
carl-codeorg: In other areas where we present a gender input, we don't provide any placeholder text. We received feedback during internal testing that the tone is too imperative for an optional field. Can you remove the placeholder text?

carl-codeorg (changes requested):

dashboard/app/views/layouts/_student_information_interstitial.html.haml
juanmanzojr: Will do!

carl-codeorg (approved):

PR #52992, opened 2023-07-25
Music: internal gallery

breville:
A very simple gallery for admin users only, which lists the last 50 `"music"` projects saved, with a link to view each.  Designed for internal use so we can peruse content currently being created.  Accessible at https://studio.code.org/musiclab/gallery.

Screenshot 2023-07-25 at 9 58 43 AM

PR #56561, opened 2024-02-14
Music: initial mini player UI

breville:
This is some initial work on a **Music Lab** mini player UI.

It leverages excellent foundational work by @sanchitmalhotra126 in https://github.com/code-dot-org/code-dot-org/pull/56551.

It also adapts the internal gallery we built in https://github.com/code-dot-org/code-dot-org/pull/52992.

The UI will undergo significant revision soon, but with this change, an admin user can visit https://studio.code.org/musiclab/gallery and see the 15 most recent Music Lab projects.  Each entry can be played, and a new play will stop an existing one.  A playing entry can also be stopped.  Finally, a link is available for each entry to open the source project in a new browser tab (though, as usual, a user will not be able to view the lab while signed in as an admin).

Screenshot 2024-02-14 at 1 29 07 PM


apps/src/music/views/MiniMusicPlayer.module.scss
```diff
@@ -0,0 +1,48 @@
...
+        transition: opacity 0.1s ease-in;
+        font-size: 20px;
+      }
+    }
+
+    .name {
+      flex-grow: 1;
+    }
+
+    .other {
```
sanchitmalhotra126, line 35: nit: more descriptive class name than `other`?

apps/src/music/views/MiniMusicPlayer.tsx
```diff
@@ -81,25 +89,55 @@ const MiniPlayerView: React.FunctionComponent<MiniPlayerViewProps> = ({
...
+              project.id === currentProjectId
+                ? onStopSong()
+                : onPlaySong(project);
+            }}
           >
-            Play! {project.name}
-          </button>
+            <div className={moduleStyles.control}>
+              <FontAwesomeV6Icon
+                iconName={project.id === currentProjectId ? 'stop' : 'play'}
```
sanchitmalhotra126, line 120: Can definitely be a follow-up - but what do you think of having a loading spinner here while the code/samples are loading?

apps/src/music/views/MiniMusicPlayer.tsx
```diff
@@ -81,25 +89,55 @@ const MiniPlayerView: React.FunctionComponent<MiniPlayerViewProps> = ({
...
+                iconStyle="solid"
+                className={moduleStyles.icon}
+              />
+            </div>
+            <div className={moduleStyles.name}>{project.name}</div>
+            <a
```
sanchitmalhotra126, line 126: could potentially use the design system `Link.tsx` here (if it works with these styles)

sanchitmalhotra126 (approved):
Looks great!

apps/src/music/views/MiniMusicPlayer.tsx
breville: Oh, good call.  Will do in follow-up.

apps/src/music/views/MiniMusicPlayer.tsx
breville: Yes, for sure.  Will investigate in follow-up.

apps/src/music/views/MiniMusicPlayer.module.scss
breville: Good point.  Will investigate in follow-up when we know what else we'll have.

PR #56562, opened 2024-02-14
[Web Lab 2] Add html and css linting prototype

molly-moen:
Add html and css error highlighting to Web Lab 2. The editor currently only shows html, so the code to configure it to lint css is commented out. 

HTML is linted with [HtmlHint](https://htmlhint.com/), and css is linted with [stylelint-bundle](https://github.com/openstyles/stylelint-bundle), which is a version of [stylelint](https://stylelint.io/) set up to run in browser. Getting stylelint-bundle imported was tricky, the best I could do was use a script tag and import from their cdn. Since that import needed to go in the `lab2` entrypoint, I've left it commented out for now. We should verify it's ok to have this in the lab2 entrypoint, and failing to fetch doesn't cause issues (if, for example, that url is blocked by schools). If we want to go with stylelint-bundle long-term, we would want to self host this.

The main challenge with getting linting working is finding the right package. We need something that runs in the browser, and can take in a string and output the violations in that string. We can then transform that output into an array of [Diagnostics](https://codemirror.net/docs/ref/#lint.Diagnostic), which is defined by codemirror.

Note: this PR looks like it's a lot of lines, but most of those lines are css rules.

### HTML Linting
Screenshot 2024-02-13 at 4 39 52 PM

HTML Hint only outputs 1 location for errors, so we can't get a "red squiggle" but instead get a single point.

### CSS Linting
Screenshot 2024-02-13 at 4 28 09 PM

stylelint does output a start and end for errors, so we get "red squiggles".

## Links
- jira ticket: [CT-289](https://codedotorg.atlassian.net/browse/CT-289)

## Testing story
Tested locally. This only impacts Web Lab 2, which is not in use yet.

## Follow-up work
Long term, we will want to decide if we like these linters, and if we want to adjust the ruleset we are evaluating with. We also need to figure out a better way to use the stylelint-bundle, ideally by self-hosting it.

## PR Checklist:

- [ ] Code is well-commented


Erin007 (approved):
Thank you for all of the comments! Makes it super clear what we'd need to update and where you are in the process.

PR #56564, opened 2024-02-14
Icon Key Tests

kobryan0619:
Adding tests to the `IconKey` for the new progress view.

Note, this is my first go with the React Testing Library.  It seems like there are only a handful of components that are using it right now - most of which Denys has written for our branding redesign.  I tried to avoid needing to add in any `data-testid` attributes, but sometimes I struggled to test the components without adding those in.  I am really interested in better understanding what kinds of norms and styles we want to develop as we convert to using React Testing Library across our code base.  For that reason, I have left some questions in the PR that will hopefully start some discussion around how we want to approach writing tests using React Testing Library.

## Links

[Ticket](https://codedotorg.atlassian.net/jira/software/projects/TEACH/boards/65?selectedIssue=TEACH-802)

## Testing story

It is all tests!

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/templates/sectionProgress/ProgressBox.jsx
+++ b/apps/src/templates/sectionProgress/ProgressBox.jsx
@@ -76,7 +76,7 @@ export default class ProgressBox extends Component {
     };
 
     return (
-      <div style={boxWithBorderStyle}>
+      <div style={boxWithBorderStyle} data-testid="progress-box">
         {this.props.lessonNumber && (
           <div style={lessonNumberStyle}>{this.props.lessonNumber}</div>
         )}
--- a/apps/src/templates/sectionProgressV2/AssignmentCompletionStatesBox.jsx
+++ b/apps/src/templates/sectionProgressV2/AssignmentCompletionStatesBox.jsx
@@ -6,9 +6,7 @@ import {ITEM_TYPE} from './ItemType';
 import styles from './progress-table-legend.module.scss';
 import {StrongText} from '@cdo/apps/componentLibrary/typography';
 
-export default function AssignmentCompletionStatesBox({
-  hasValidatedLevels,
-}) {
+export default function AssignmentCompletionStatesBox({hasValidatedLevels}) {
   // TO-DO (TEACH-800): Fix spacing on validated levels once width on page is set
   const legendIcons = () => {
     return (
--- a/apps/src/templates/sectionProgressV2/AssignmentCompletionStatesBox.jsx
+++ b/apps/src/templates/sectionProgressV2/AssignmentCompletionStatesBox.jsx
@@ -7,7 +7,6 @@ import styles from './progress-table-legend.module.scss';
 import {StrongText} from '@cdo/apps/componentLibrary/typography';
 
 export default function AssignmentCompletionStatesBox({
-  isViewingLevelProgress,
   hasValidatedLevels,
 }) {
   // TO-DO (TEACH-800): Fix spacing on validated levels once width on page is set
@@ -34,7 +33,7 @@ export default function AssignmentCompletionStatesBox({
             labelText={i18n.submitted()}
           />
         </div>
-        {isViewingLevelProgress && hasValidatedLevels && (
+        {hasValidatedLevels && (
           <div className={styles.legendColumn}>
             <LegendItem
               itemType={ITEM_TYPE.VALIDATED}
@@ -57,6 +56,5 @@ export default function AssignmentCompletionStatesBox({
 }
 
 AssignmentCompletionStatesBox.propTypes = {
-  isViewingLevelProgress: PropTypes.bool,
   hasValidatedLevels: PropTypes.bool,
 };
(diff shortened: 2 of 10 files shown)


apps/src/templates/sectionProgress/ProgressBox.jsx
```diff
@@ -76,7 +76,7 @@ export default class ProgressBox extends Component {
     };
 
     return (
-      <div style={boxWithBorderStyle}>
+      <div style={boxWithBorderStyle} data-testid="progress-box">
```
kobryan0619, line 79: Since this `ProgressBox` can look like a lot of different things, but for the ProgressIcon component to render correctly I wanted to see that there existed a ProgressBox at certain points.  That is why I added in this `testid` here.

apps/src/templates/sectionProgressV2/AssignmentCompletionStatesBox.jsx
```diff
@@ -34,7 +33,7 @@ export default function AssignmentCompletionStatesBox({
             labelText={i18n.submitted()}
           />
         </div>
-        {isViewingLevelProgress && hasValidatedLevels && (
+        {hasValidatedLevels && (
```
kobryan0619, line 34: Refactored because of how this was implemented at a higher level component.

apps/src/templates/sectionProgressV2/IconKey.jsx
```diff
@@ -35,7 +34,7 @@ export default function IconKey({isViewingValidatedLevel, expandedLessonIds}) {
       aria-expanded={isOpen}
       aria-label={i18n.iconKey()}
     >
-      <div onClick={clickListener} className={styles.iconKeyTitle}>
+      <div onClick={clickListener} className={styles.iconKeyTitle} data-testid="expandable-container">
```
kobryan0619: I wanted to be able to target this `div` for clicking which is why I added in this `testid` here.

apps/src/templates/sectionProgressV2/ProgressIcon.jsx
```diff
@@ -9,11 +9,11 @@ export const PROGRESS_ICON_TITLE_PREFIX = 'progressicon-';
 
 export default function ProgressIcon({itemType}) {
   const needsFeedbackTriangle = () => (
-    <div className={classNames(styles.needsFeedback, styles.cornerBox)} />
+    <div className={classNames(styles.needsFeedback, styles.cornerBox)} data-testid="needs-feedback-triangle"/>
   );
 
   const feedbackGivenTriangle = () => (
-    <div className={classNames(styles.feedbackGiven, styles.cornerBox)} />
+    <div className={classNames(styles.feedbackGiven, styles.cornerBox)} data-testid="feedback-given-triangle"/>
```
kobryan0619: Again, with both of these I couldn't find a single attribute that really spoke to what it was supposed to render, thus I added these in.

apps/src/templates/sectionProgressV2/ProgressIcon.jsx
```diff
@@ -38,7 +38,7 @@ export default function ProgressIcon({itemType}) {
   );
 
   return (
-    <>
+    <div data-testid="progress-icon">
```
kobryan0619, line 47: I tried adding this `testid` to the `React.Fragment` but apparently `React.Fragments` can only have a `key` component.  Thus I changed this to a `div`.  It did not seem to impact rendering at all.  Open to other suggestions here though.  It is possible that we remove this and just spot check that one of the `ITEM_TYPES` gets rendered specifically.

apps/test/unit/templates/sectionProgressV2/LegendItemTest.jsx
```diff
@@ -0,0 +1,21 @@
...
+    it('renders with provided label text', () => {
+        const labelText = 'Test Label';
+        render(<LegendItem labelText={labelText} />);
+        expect(screen.getByText(labelText)).to.exist;
+    });
+
+    it('renders ProgressIcon with correct itemType', () => {
+        const itemType = ITEM_TYPE.VALIDATED;
+        render(<LegendItem labelText="Test Label" itemType={itemType} />);
+        const progressIcon = screen.getByTestId('progress-icon');
```
kobryan0619: This is a spot where we could see that the specific correct `ITEM_TYPE` symbol shows up.  Right now this is more of a spot check that a `ProgressIcon` is rendered.

apps/test/unit/templates/sectionProgressV2/LevelTypesBoxTest.jsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import {render, screen} from '@testing-library/react';
+import {expect} from '../../../util/reconfiguredChai';
+import LevelTypesBox from '@cdo/apps/templates/sectionProgressV2/LevelTypesBox';
+
+describe('LevelTypesBox Component', () => {
+    it('renders three options', () => {
+      render(<LevelTypesBox />);
+      expect(screen.getByText('Level Types')).to.exist;
```
kobryan0619: Side note, in all of these tests I rely more on testing that the text shows up rather than the symbol.  I am interested in knowing if that feels sufficient, given I test the symbols show up in the test of `ProgressIcon`

etaderhold (approved):
I like it! I will say that I'm still new at the React Testing Library as well. From reading through the docs it seems they discourage the use of `testid` when other ways to test are possible. I'm not seeing a straightforward way to do better in the cases where you put one of these IDs in, but it might be worth having a chat with Liam about that when he returns.

PR #56569, opened 2024-02-14
Move deeper learning course info to separate page

bethanyaconnor:
This was a fun task in the end. It turns out we already have an existing index page for deeper learning courses at `/plc/user_course_enrollments`, which was completely written in haml. I pulled out all of the haml, used the existing component on the PL landing page to display the same data, and created a route `/deeper-learning` that also shows the same info.

Before:
Screenshot 2024-02-14 at 1 16 51 PM

After:
Screenshot 2024-02-14 at 1 31 28 PM

The existing route isn't in use as far as I can tell, but the courses page shows this template for deeper learning courses. This is well tested by our UI tests -- it took a while for me to get them to pass. 😅 [This](https://app.saucelabs.com/tests/af13704a2f1e4ce5891501e04408a10e#28) is the green test run from the green Drone run and [this](https://app.saucelabs.com/tests/8d6a82f61e4043cab884d44b1cabb184#24) is the run from a recent DTT.

### DIFF
--- a/apps/src/code-studio/pd/professional_learning_landing/LandingPage.jsx
+++ b/apps/src/code-studio/pd/professional_learning_landing/LandingPage.jsx
@@ -38,9 +38,12 @@ export default class LandingPage extends Component {
           )}
           <EnrolledWorkshops />
           {this.props.deeperLearningCourseData && (
-            <ProfessionalLearningCourseProgress
-              deeperLearningCourseData={this.props.deeperLearningCourseData}
-            />
+            <div>
+              <h2>Online Professional Learning Courses</h2>
+              <ProfessionalLearningCourseProgress
+                deeperLearningCourseData={this.props.deeperLearningCourseData}
+              />
+            </div>
           )}
         </main>
       </div>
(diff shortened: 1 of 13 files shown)


apps/src/code-studio/pd/professional_learning_landing/LandingPage.jsx
```diff
@@ -38,9 +38,13 @@ export default class LandingPage extends Component {
           )}
           <EnrolledWorkshops />
           {this.props.deeperLearningCourseData && (
-            <ProfessionalLearningCourseProgress
-              deeperLearningCourseData={this.props.deeperLearningCourseData}
-            />
+            <div>
+              <Heading2>Online Professional Learning Courses</Heading2>
```
bethanyaconnor, line 42: I didn't translate this text because it's temporary and will be removed when we add the facilitator page.

apps/src/code-studio/pd/professional_learning_landing/LandingPage.jsx
```diff
@@ -38,9 +38,12 @@ export default class LandingPage extends Component {
           )}
           <EnrolledWorkshops />
           {this.props.deeperLearningCourseData && (
-            <ProfessionalLearningCourseProgress
-              deeperLearningCourseData={this.props.deeperLearningCourseData}
-            />
+            <div>
+              <h2>Online Professional Learning Courses</h2>
```
bethanyaconnor, line 42: This wasn't translated before and is temporary, so I didn't bother to translate it now.

TurnerRiley (approved):
Nice job tracking all this down and thank you for the detailed test updates!

PR #54302, opened 2023-10-17
[Google Blockly] Merge level toolbox xml blocks into auto-populated variables category

mikeharv:
This makes some changes to the auto-populated variable flyout so that the blocks specified in the level XML are also included. We had to do a similar patch for the behaviors category, so I reused some logic to accomplish this.

For comparison purposes, I'm including videos where I will perform the same basic operations, comparing the toolbox contents as I go:
1. Open "Variables" category in brand new project
2. Create a variable called "test"
3. Create another variable called "test2"
4. Compare the flyout contents

CDO Blockly does not support the "Create a variable button...", but the curriculum team has expressed interest in using this feature as it feels more modern and better aligns to tools like scratch.

### CDO Blockly (baseline)
_In order to create a new variable, one must drag out a block with a variable picker input, choose "**Rename this variable**", then enter a new name. The block used for this operation is updated and the toolbox is updated behind the scenes._
* `change counter by 1` and `???` are both provided by the level's toolbox XML
* a pair of `set var to` and `var` (getter) blocks are added to the toolbox for each variable present in the main workspace

https://github.com/code-dot-org/code-dot-org/assets/43474485/8503d793-830d-40f1-89a8-ef3e83bcd50e

### Google Blockly (staging)
_In order to create a new variable, you either do use the method above (not shown here), or press the "**Create variable...**" button which creates a modal. The flyout stays open and is immediately updated.
* The `change counter by 1` and `???` blocks specified in the level's toolbox XML are not shown.
* If variables are present, `set var to` and `change var by 1` blocks are created, along with a `var` (getter) block for each block. Note that the `change test2 by 1` block in the video below comes from a different place than the `change counter by 1` block in the video above.

https://github.com/code-dot-org/code-dot-org/assets/43474485/d6ea4cd6-8181-488f-8949-b711723a4688

### Google Blockly (feature branch)
_Variables can be created using either method, as above._
* `change counter by 1` and `???` are both provided by the level's toolbox XML
* If variables are present, a single `set var to` block is created, along with a `var` (getter) block for each block. Because levels commonly include the `change var by` block (since CDO Blockly didn't provide it), we prevent Blockly from creating extra copies of the block when a variable is created. If no `change by` block is listed in the toolbox XML, we will now create one. This means the curriculum team won't have to manually include this block any more, but still can if they want to customize its default values.

https://github.com/code-dot-org/code-dot-org/assets/43474485/f7b08dba-af18-407c-ab2e-5293d47250b5

[...]

PR #56571, opened 2024-02-14
[Google Blockly] simplify VARIABLE field values for toolbox

mikeharv:
@ebeastlake noticed an error when trying to modify a toolbox with certain variable-related blocks:
```
variable_map.ts:186 Uncaught Error: Variable "i" is already in use and its id is "BY?TWU)X/R}36]#AMXP." 
which conflicts with the passed in id, "8c{g(FOfdV@:v^nWW=f[".
```
Toolbox Mode| Result in level
-|-
Screenshot 2024-02-13 at 4 53 10 PM|Screenshot 2024-02-13 at 4 53 50 PM

### Context
In Core Blockly, the blocks in the Variables category are auto-populated. On Code.org, we need to support a combination of auto-populated blocks and level-defined blocks in the XML. This was implemented for our Google Blockly labs here:
- https://github.com/code-dot-org/code-dot-org/pull/54302

Part of that work involves stripping unneeded block and variable ids before adding the level-defined blocks to the XML. (The ids show up in the first place because we convert the XML to JSON to make it compatible with the auto-populated blocks. We don't want random variable ids, because they might mismatch the defined variables on the main workspace's variable map, which would lead to the error shown above.) 
In that work, I had assumed we always used `VAR` for the field name. However, the `gamelab_changeVarBy` block uses `VARIABLE` (https://github.com/code-dot-org/code-dot-org/blob/mike/simplify-VARIABLE-field-values-for-toolbox/dashboard/config/blocks/GamelabJr/gamelab_changeVarBy.json#L14). Technically, this is supported as fields can have any name. Checking for both variants of the field name makes it so this block is properly simplified for the toolbox.

image

## Links

https://codedotorg.atlassian.net/browse/CT-337

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/blockly/addons/cdoUtils.js
+++ b/apps/src/blockly/addons/cdoUtils.js
@@ -398,13 +398,21 @@ function simplifyBlockState(block, variableMap) {
   // Create a copy of the block so we can modify certain fields.
   const result = {...block};
 
-  // For VAR fields, look up the name of the variable to use instead of the id.
-  if (block.fields && block.fields.VAR) {
-    result.fields.VAR = {
-      name: variableMap[block.fields.VAR.id] || '',
-      type: '',
-    };
-  }
+  // For variable fields, look up the name of the variable to use instead of the id.
+  const variableFields = [
+    'VAR' /* most common */,
+    'VARIABLE' /* used in gamelab_changeVarBy */,
+  ];
+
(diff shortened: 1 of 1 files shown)


ebeastlake (approved):

ebeastlake:
Thank you and nice find, Mike! 🎉

PR #56330, opened 2024-02-05
 Change teacher dashboard page width

lfryemason:
**WILL CAUSE EYES DIFFS**

Removes the set page-width from teacher dashboard pages.

All existing pages add back the width to make the content changes minimal. This should only effect:
* Progress page v2 (no ui tests yet)
* Header of each teacher dashboard page

https://github.com/code-dot-org/code-dot-org/assets/25193259/6ed2bbb0-9875-427f-854c-7fdc67be08aa

## Links

[TEACH-852](https://codedotorg.atlassian.net/browse/TEACH-852)

## PR Checklist:

- [ ] Tests provide adequate coverage
- [ ] Privacy and Security impacts have been assessed
- [ ] Code is well-commented
- [ ] New features are translatable or updates will not break translations
- [ ] Relevant documentation has been added or updated
- [ ] User impact is well-understood and desirable
- [ ] Pull Request is labeled appropriately
- [ ] Follow-up work items (including potential tech debt) are tracked and linked

PR #56581, opened 2024-02-14
modified spacing in icon key

kobryan0619:
This adds spacing to the boxes on the legend.  Note that [Liam has been working on a PR](https://github.com/code-dot-org/code-dot-org/pull/56330) to change the width of the screen this component will be on.  Once that goes live, there will be enough space for all of the boxes - I had to modify that width in the developer console for the image below.  Since this is behind a DCDO flag, we are able to merge this before the screen width gets changed with Liam's PR.

Old look:
image

New look:
image

## Links
[Figma](https://www.figma.com/file/gCktlnafGYQSNJafD3FUgA/Progress-Tracking?node-id=2173%3A33166&mode=dev)
[Slack thread with Molly](https://codedotorg.slack.com/archives/C06ERUABG01/p1707837856028289)

[
Ticket](https://codedotorg.atlassian.net/jira/software/projects/TEACH/boards/65?selectedIssue=TEACH-801)

### DIFF
--- a/apps/src/templates/sectionProgressV2/progress-table-legend.module.scss
+++ b/apps/src/templates/sectionProgressV2/progress-table-legend.module.scss
@@ -32,7 +32,7 @@
 }
 
 .legend.levelTypes {
-  margin-right: 0px;
+  margin-right: 0;
 }
 
 .legendColumn {
--- a/apps/src/templates/sectionProgressV2/progress-table-legend.module.scss
+++ b/apps/src/templates/sectionProgressV2/progress-table-legend.module.scss
@@ -21,6 +21,7 @@
   display: inline-flex;
   flex-direction: column;
   white-space: nowrap;
+  margin-right: 16px;
 }
 
 .legend .headerContainer {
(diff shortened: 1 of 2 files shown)


apps/src/templates/sectionProgressV2/progress-table-legend.module.scss
```diff
@@ -30,6 +31,10 @@
   box-sizing: border-box;
 }
 
+.legend.levelTypes {
```
kobryan0619, line 34: Note that having this margin on the last box in the row pushed it over the edge to the next line.  This was my solution to this.  Initially thought about just adding margin to the "middle box" on both the left and right side.  However, after feedback from Liam at one point, I like the idea of `.legend` as being the core style for most of the boxes and that style including padding.

THUS, this rule was added.  I'd love feedback on this though.

etaderhold: I agree with this approach. Give `.legend` a margin by default, that is overridden to zero in this specific case. Looks great!

etaderhold (approved):

PR #56517, opened 2024-02-12
Revert "Revert "Update student view of Curriculum Catalog cards""

TurnerRiley:
Reverts code-dot-org/code-dot-org#56507 (which reverted code-dot-org/code-dot-org#56322 due to a failing mobile UI test). Since the mobile UI test wasn't run on drone, I didn't know it was failing until it hit the test server. For this re-revert I pushed the empty commit to `test ios` and ensure that this PR runs mobile UI tests within its drone runs.

The issue was that I updated [this aria-label](https://github.com/code-dot-org/code-dot-org/pull/56322#discussion_r1483556585) but then forgot to update the [UI test](https://github.com/code-dot-org/code-dot-org/pull/56322/files#diff-a0df19ba121b7ce1a2175e6c7774ee756143c2f35954729ef7df98f80896ca23R320) to accomodate, so it was unable to find the element by aria-label.

PR #56583, opened 2024-02-14
Revert "Revert "Revert "Revert "Update student view of Curriculum Catalog cards""""

TurnerRiley:
Reverts code-dot-org/code-dot-org#56582

Sorry for the re-re-re-revert. I thought the [previous change](https://github.com/code-dot-org/code-dot-org/pull/56517) would work because of the passing test following the `[test ios]` commit, I didn't realize you needed to include `[test ios]` per commit you wanted tested. This PR does that and I found that the issue was [whether "now" was capitalized or not](https://github.com/code-dot-org/code-dot-org/pull/56583/commits/9c211c52c6b56a6351b1dd1c4668fc5bd20e6482).

As shown below, the [show failing baseline [test ios]](https://github.com/code-dot-org/code-dot-org/pull/56583/commits/a023445b157d172d8000f5e4a3061254f6c2b15c) commit fails, and once [the change is implemented](https://github.com/code-dot-org/code-dot-org/pull/56583/commits/9c211c52c6b56a6351b1dd1c4668fc5bd20e6482) it now passes. The reason for the [baseline [test ios]](https://github.com/code-dot-org/code-dot-org/pull/56583/commits/bb0860ce0e8ea95dbea09d80f55cbee10d67b98a) commit followed by making the fix then undoing the fix was because I wanted to show a failing commit without the change and a successful one with it, but I pushed both at the same time so drone only ran on the successful one. So, I just undid the fix to get a failing drone build then reimplemented the fix to get a successful one to feel confident that it'll work this time.

### DIFF
--- a/apps/i18n/common/en_us.json
+++ b/apps/i18n/common/en_us.json
@@ -2612,6 +2612,7 @@
   "tryIt": "Try it",
   "tryItOut": "Try it out",
   "tryNow": "Try Now",
+  "tryCourseNow": "Try {course_name} now",
   "turtleProgrammingInAppLab": "Turtle Programming in App Lab",
   "turtleProgrammingInAppLabDescription": "Unlock the ability to make rich, interactive apps with JavaScript in the App Lab!",
   "tutorialUnavailable": "Tutorial unavailable for younger students",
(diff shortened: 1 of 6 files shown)


bethanyaconnor (approved):
🤞 thanks for the persistence here!

PR #56584, opened 2024-02-14
Music: scroll timeline during playback

sanchitmalhotra126:
Automatically scroll the timeline during playback if it's near the edge of the screen. Disables user scrolling during playback.

Also pulled the playback update timer logic into a hook so it could be used by both `MusicLabView` and `Timeline`.

https://github.com/code-dot-org/code-dot-org/assets/85528507/7dfcedbc-d952-40b7-bbc4-36795598cd46

## Links

https://codedotorg.atlassian.net/browse/LABS-456

## Testing story

Tested locally with various screen sizes.


breville (approved):
This is great.  I did a bunch of browser resizing while playing and the approach held up nicely.  It's a really solid user experience.

PR #56586, opened 2024-02-15
Small progress lock icons

etaderhold:
This change puts small lock icons in the corner of the lesson cell for lockable lessons when they are locked for that particular student. Example below (the lesson is unlocked for the first three students and locked for the rest). The column header has an unlocked icon since the lesson is unlocked for some students.

image

## Links

[Figma](https://www.figma.com/file/gCktlnafGYQSNJafD3FUgA/Progress-Tracking?type=design&node-id=3618-13063&mode=design&t=ooE8AyKDvaGAPCl2-0)

## PR Checklist:

- [ ] Code is well-commented


apps/src/templates/sectionProgressV2/LessonProgressDataColumn.jsx
```diff
@@ -17,18 +17,26 @@ function LessonProgressDataColumn({
...
   addExpandedLesson,
 }) {
+  const lockedPerStudent = React.useMemo(
+    () =>
+      Object.fromEntries(
+        sortedStudents.map(student => [
+          student.id,
+          lesson.lockable &&
+            lesson.levels.every(
+              level => levelProgressByStudent[student.id][level.id]?.locked
```
kobryan0619, line 27: I don't think checking every level to see if it is locked is needed.  I believe teachers can only lock and unlock from the lesson (not at the level...).

etaderhold: Yes, the UI implements locking at the per-lesson level. We then store this state in the `UserLevel` object, which feeds back into the `levelProgressByStudent` in our redux store. `lessonProgressByStudent` does not contain this information, we have to look at the level information instead.

kobryan0619 (approved):
Approved with the note about "every level" being locked or not.  It might be a good question for Bryan - in someways it might be good to keep this logic here in case we do move to teachers being able to lock individual questions or portions of a survey/quiz, but I am not sure if that would really be useful to a teacher.

apps/src/templates/sectionProgressV2/LessonProgressDataColumn.jsx
kobryan0619: Ah... that makes sense - you needed to do that because of the way the data was organized... thanks for clarifying!  Looks good!

PR #56589, opened 2024-02-15
[Google Blockly] only apply color to block if it doesn't have a style

mikeharv:
Yesterday, @ebeastlake observed on Slack that there was a slight mis-match in block color for similar blocks that are custom vs included-with-core:
> Are these `set i to` blocks from core Blockly and the others are custom, which is why their colors are slightly off/different? Do we care?

Screenshot 2024-02-14 at 9 49 09 AM

Google Blockly supports both colors and styles, but if both are present the color is used. This has the drawback that even if a block is meant to be using the style `'variable_blocks'` it might not. HSV color values are also harder to reason about. Most importantly, because the colors are fixed, they don't change when the user switches to an accessible theme:
Here is a sampling of other blocks as they currently appear in the Protanopia theme (production):
image|image|image
-|-|-
These blocks all use the default style, but the two teal blocks also specify that they should use the default color. | The `being draggable` block is a custom (uneditable) behavior block that specifies both a style and color.| These blocks all use the logic_blocks style, but the light color ones also specify a color. (The darker blocks don't because they come from Core Blockly.)

As a immediate protection against this, we can simply stop setting a color if the block already has a style. This results in a more consistent color scheme:
image|image|image
-|-|-

Next, now that Sprite Lab has been migrated, we can clean up the block definitions in the default pool. 80+ custom blocks had both colors and styles defined. (If block didn't have a `style`, I left its `color` in place.) This PR removes the unneeded colors. 

To promote the cleanup these outdated color values in other block pools, we can also log a warning when the color has been ignored in favor of the existing style. For example, we will ignore the color and issue a warning for these block in the story lab pool:
imageimage

After adding this warning, I noticed that we were trying to apply the default color to these edited blocks (in `block_utils.js`). Since we shouldn't need to set a color on a block that has a style, I updated the logic to reflect that, using mainline's `block.getStyleName()` function. To prevent crashes in CDO Blockly, I added this to the block prototype in that wrapper.

## PR Checklist:

- [ ] Code is well-commented


apps/src/block_utils.js
```diff
@@ -1085,13 +1085,13 @@ exports.createJsWrapperBlockCreator = function (
         // automatically based on a block's returnType (e.g. yellow for "Location").
         if (color) {
           Blockly.cdoUtils.setHSV(this, ...color);
-        } else if (!returnType) {
+        } else if (!returnType && !this.getStyleName()) {
```
ebeastlake: What's the difference between checking `style` and `this.getStyleName()`?

I'm just curious why the logic you changed here and the logic related to styles above didn't get combined into something like...

```
const style = this.getStyleName();
if (style) {
  this.setStyle(style);
} else if (color) {
  Blockly.cdoUtils.setHSV(...);
} else {
  Blockly.cdoUtils.setHSV(this, ...DEFAULT_COLOR);
  this.setStyle('default');
}

Comment (mikeharv) [reply apps/src/block_utils.js]: This is the joy of `block_utils.js` madness. In this scope, `style` refers to the value set in the block config JSON. Above (https://github.com/code-dot-org/code-dot-org/commit/53e3d31e00c44814cdab0cd60affe59583e0a18d), we use this value to actually set the style on the block:

```
        if (style) {
          // Google Blockly method. No-op for CDO Blockly.
          this.setStyle(style);
        }
```

`setStyle` does nothing for CDO Blockly (e.g. Minecraft). Similiarly, `getStyleName` returns `undefiend` for CDO Blockly.  The line you tagged is using `this.getStyleName()` to check whether a block actually has a style. This will be falsey if `style` is missing OR if we're in CDO Blockly. In either situation, it's safe to apply the default color.

File: apps/src/block_utils.js [hunk 2]
```diff
@@ -1085,13 +1085,13 @@ exports.createJsWrapperBlockCreator = function (
         // automatically based on a block's returnType (e.g. yellow for "Location").
         if (color) {
```
molly-moen: could this just be an `else if`? We wouldn't get the warning log in that case (is it useful?), but it would make this logic clearer that style overrides color.

molly-moen (commented):
Thanks for doing this cleanup! One question on logic

apps/src/block_utils.js
mikeharv: It can't be an `else if` because of CDO Blockly. Consider a Minecraft block with both a `color` and `style`. We'd try to set a style (no-op), and then we'd be done.

molly-moen: ah gotcha I didn't realize cdo blockly blocks had styles.

mikeharv: Minecraft blocks do, as part of the work to get them ready for migration. This code makes it safe for a CDO Blockly block to have both for now (even though we only use the color), but it will automatically start using using the style when we flip the lab.

molly-moen: Could we use `getStyleName` instead?

molly-moen: (in place of the check on `style`)

mikeharv: We could move the `getStyleName` up earlier like this:

```
if (style) {
  // Attempt to set a style. Only works on Google Blockly
  this.setStyle(style);
}
if (!this.getStyleName()) {
  if (color) {
    Blockly.cdoUtils.setHSV(this, ...color);
  } else if (!returnType) {
    Blockly.cdoUtils.setHSV(this, ...DEFAULT_COLOR);
    this.setStyle('default');
  } 
} // else: block has a style 
```

`this.getStyleName()` will tell us if we've already successfully set a style on the block, but only after we've called `setStyle` on it.

mikeharv: See this other comment: https://github.com/code-dot-org/code-dot-org/pull/56589#discussion_r1491166313

molly-moen: do you think that's cleaner? I'm not sure I see a clear value from a warning log when we ignore a color value on a block, but if that's helpful for curriculum debugging we can keep it as-is!

mikeharv: Moved the logic out to a util for each wrapper in hopes that it's clearer:
https://github.com/code-dot-org/code-dot-org/pull/56589/commits/330868f8b4a15d6aa9af0d2f366cacd7fe24fd38

molly-moen (approved):
Nice! This solution is much easier to read

PR #56599, opened 2024-02-15
Remove principal_schedule_confirmed from scoreable questions

bethanyaconnor:
Applications that met all guidelines were seeing "Reviewing Incomplete" on the application dashboard. Looking at an application for each of CSD, CSP, and CSA, the issue was that `principal_schedule_confirmed` was still in `SCOREABLE_QUESTIONS` but removed in #55240. The applications I checked had all other fields set, so this is the only one that needed to be removed.

### DIFF
--- a/lib/cdo/shared_constants/pd/teacher_application_constants.rb
+++ b/lib/cdo/shared_constants/pd/teacher_application_constants.rb
@@ -290,7 +290,6 @@ module Pd
         :committed,
         :previous_yearlong_cdo_pd,
         :principal_approval,
-        :principal_schedule_confirmed,
       ],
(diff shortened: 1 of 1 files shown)


TurnerRiley (approved):
Looks great! Will this require re-running `auto_score!` for the applications affected or will it auto-update?

bethanyaconnor:
I don't believe so because this is all calculated in `meets_criteria`: https://github.com/code-dot-org/code-dot-org/blob/825e0b6ac4b3a765d1c0c7bdf03170e9cbe48364/dashboard/app/models/pd/application/teacher_application.rb#L1100. I'll double check after this PR goes live though!

PR #56601, opened 2024-02-15
[Google Blockly] skip finding model in embedded workspaces

mikeharv:
Resolves [CT-339: Hints are broken if they contain behavior blocks](https://codedotorg.atlassian.net/browse/CT-339)
Slack thread: https://codedotorg.slack.com/archives/C05DK21DAHK/p1707936093182039
ZenDesk ticket: https://codeorg.zendesk.com/agent/tickets/485498

A user reported that users were experiencing an unrecoverable crash at `s/coursef-2023/lessons/8/levels/4/sublevel/6`. 

image (178)

The cause of this evidently relates to how shareable procedure blocks interact across workspaces. If we are working on an embedded workspace, behavior blocks do not need to actually have models or be connected to the student workspaces. When authored hints include block XML, we create an embedded workspace. We then convert the XML to JSON using a temporary headless workspace. Finally, we load the JSON onto the rendered embedded workspace.

To solve this problem, we can pass the `embedded` state of the hint workspace to the temporary workspace. Then, if we know we're an a temporary workspace _for an embedded _workspace_ we can skip trying to find the procedure model altogether. (Again, this is safe because embedded blocks are effectively read-only so they don't need to have an actual model associated with them.)

Now, the page loads as expected, including using the first impacted student's exact source and animation files:

image

## Links

Resolves [CT-339: Hints are broken if they contain behavior blocks](https://codedotorg.atlassian.net/browse/CT-339)
Slack thread: https://codedotorg.slack.com/archives/C05DK21DAHK/p1707936093182039
ZenDesk ticket: https://codeorg.zendesk.com/agent/tickets/485498

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/blockly/addons/cdoSerializationHelpers.js
+++ b/apps/src/blockly/addons/cdoSerializationHelpers.js
@@ -64,10 +64,18 @@ function getSpaceBetweenBlocks(block) {
 /**
  * Converts XML serialization to JSON using a temporary unrendered workspace.
  * @param {xml} xml - workspace serialization, current/legacy format
+ * @param {boolean} [embedded] - indicates whether the source will be converted
+ * for an embedded workspace for not.
  * @returns {json} stateToLoad - modern workspace serialization
  */
-export function convertXmlToJson(xml) {
+export function convertXmlToJson(xml, embedded) {
   const tempWorkspace = new Blockly.Workspace();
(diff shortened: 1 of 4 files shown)


molly-moen (approved):
🎉

PR #56602, opened 2024-02-15
replace crypto-js with browserify-crypto as crypto-js is no longer maintained

stephenliang:
Previously, crypto-js was used solely for the md5 hashing function. Since node.js has a native crypto library, we can utilize the native md5 hasher. For browsers, a polyfill `browserify-crypto` can be used which is actively being maintained.

fixes dependabot moderate [Dependabot #434](https://github.com/code-dot-org/code-dot-org/security/dependabot/434) and critical  [Dependabot #486](https://github.com/code-dot-org/code-dot-org/security/dependabot/486)

## Links

1. [crypto-js discontinued announcement](https://github.com/brix/crypto-js)
2. [crypto-browserify statement of active project status](https://github.com/browserify/crypto-browserify/issues/229)
3. [Dependabot #486](https://github.com/code-dot-org/code-dot-org/security/dependabot/486)
4. [Dependabot #434](https://github.com/code-dot-org/code-dot-org/security/dependabot/434)

## Testing story

I ran rake test and the UI tests. The following tests were run to ensure that the native md5 hasher is consistent with crypto-js:

```
apps/test/unit/templates/instructions/InlineAudioTest.js
apps/test/unit/templates/curriculumCatalog/curriculumCatalogExpandedCardTest.jsx
```

## Follow-up work

No follow up necessary.

## Privacy

This does not change any user models or interact with privacy.

## Security

This addresses the discontinued crypto-js library by switching to an actively maintained and more mainline browserify library.

## PR Checklist:

- [x] Code is well-commented

### DIFF
--- a/apps/package.json
+++ b/apps/package.json
@@ -281,7 +281,7 @@
     "ace-builds": "^1.4.12",
     "blockly": "10.3.1",
     "chart.js": "^4.4.0",
-    "crypto-js": "^3.1.9-1",
+    "crypto-browserify": "^3.12.0",
     "dapjs": "^2.3.0",
     "details-element-polyfill": "https://github.com/javan/details-element-polyfill",
     "filesaver.js": "0.2.0",
--- a/apps/src/code-studio/hashEmail.js
+++ b/apps/src/code-studio/hashEmail.js
@@ -1,5 +1,5 @@
 import $ from 'jquery';
-import MD5 from 'crypto-js/md5';
+import {md5} from '../util/crypto';
 
 const EMAIL_REGEX = /^[A-Z0-9._%+-]+@[A-Z0-9.-]+\.[A-Z]{2,}$/i;
 
@@ -21,7 +21,7 @@ export default function (options) {
 }
 
 export function hashEmail(cleartextEmail) {
-  return MD5(normalizeEmail(cleartextEmail)).toString();
+  return md5(normalizeEmail(cleartextEmail));
 }
 
 function normalizeEmail(rawEmail) {
--- a/apps/src/sites/studio/pages/shared/_check_admin.js
+++ b/apps/src/sites/studio/pages/shared/_check_admin.js
@@ -1,8 +1,8 @@
-import MD5 from 'crypto-js/md5';
 import getScriptData from '@cdo/apps/util/getScriptData';
+import {md5} from '@cdo/apps/util/crypto';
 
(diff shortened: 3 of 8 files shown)


wilkie:
We're replacing a deprecated library with an older unmaintained library, that feels odd but I guess that's ok.

It would be nice, if we are going to the trouble, to just have a library with an md5 implementation instead of seemingly including an entire cryptography package for the one simple hashing function. Is there one?

wilkie:
I say unmaintained since it was pushed to npm 6 years ago and the links to the repo and homepage are broken: https://www.npmjs.com/package/crypto-browserify when I actually visit the place the repo actually is now, which is https://github.com/browserify/crypto-browserify it is still 6 years since the last push. if anything, this library is potentially _worse_ than the one we are replacing on that basis... so I'm asking if there is a more maintained, smaller package.

stephenliang:
@wilkie browserify-crypto isn't unmaintained, however it is seemingly unmaintained by the lack of activity. This is seen from the statement of maintenance from browserify [[1]](https://github.com/browserify/crypto-browserify/issues/229)  and [[2]](https://github.com/browserify/crypto-browserify/issues/223)

For a crypto library, it's a good thing to have a lack of activity since the algorithms (such as md5) haven't changed and won't be changing and you want to have a higher standard of scrutiny. 

Regarding importing a cryptography library, it is a polyfill of the node.js Crypto library which is native built on Node.js. Since we'd do tree shaking via webpack, the rest of the algorithms which are unused will be removed out from the bundle and only md5 will make it in the end. 

Let me know if you have strong concerns about the library though, although my preference is on browserify's suite of libraries due the cross-compatability with native Node.js' core libraries, I'd be happy to take a look at alternatives if necessary.

stephenliang:
One more thing: [crypto-js](https://github.com/brix/crypto-js) does explicitly state that we should migrate toward the node.js native crypto library:

> Active development of CryptoJS has been discontinued. This library is no longer maintained.
> 
> Nowadays, NodeJS and modern browsers have a native Crypto module. The latest version of CryptoJS already uses the > native Crypto module for random number generation, since Math.random() is not crypto-safe. Further development of > CryptoJS would result in it only being a wrapper of native Crypto. Therefore, development and maintenance has been > discontinued, it is time to go for the native crypto module.

daynew (approved):
The maintainer seems to be active given that they recently responded to questions.

PR #56607, opened 2024-02-15
[JIMT][CDOIDE] IDE POC

thomasoniii:
For Web Lab 2, and presumably eventually upper grades, here is the extremely rough, extremely ugly implementation of an IDE.

Screenshot 2024-02-15 at 3 45 39 PM

Right now, this _only_ really works with html/css pages. It has some placeholders for running more advanced languages, but they do nothing. There's a readme in the directory. So we have a working file browser, editor (though not with formatting atm, unfortunately), and preview. The internal code is roughly knocked out and needs a ton of iteration, this is really pre-alpha code even.

## What does this do

It lays out a full IDE from some of the upper grade POC docs, but all it actually implements is a file browser, editor (with tabs for open files) and preview of HTML pages w/CSS. Everything else is placeholders.

## What does this not do

Anything else.

- It's ugly as sin. There's little in the way of styling, it needs to be re-written into a pretty component library.
- Work will need to be done on the preview to make sure it does everything we may need it to do. For example, right now it doesn't have a concept of knowing where to find images or link to other pages in the project.
- All the logic needs to be prettied up. For example, there is a shared context with the `project` and `config` props, but no pretty reducers to easily update them.
- Other bits of proposed functionality aren't implemented - such as a search box or more robust instructions.
- The input structure is somewhat arbitrary. Maybe we want to adapt to a pre-existing project.json structure? Maybe we want to redefine it? Totally plausible.

## Where do I start?

- Take a look at `Weblab2View.tsx` for the wrapper code that just lays out an IDE with some pre-defined `config` and `project` values.
- In the `CDOIDE` directory, `CDOIDE/CDOIDE.tsx` is your starting point with the overall layout.
- There are `left-pane`, `center-pane`, `right-pane` folders laying out the content within each of them.
- The center pane contains the `Editor.tsx` component which is the interface to `CodeMirror` and the `FileNav.tsx` component which gives the file tabs at the top.
- The left pane contains the `Files.tsx` component which sets up the file browser.
- _Don't put any external CDO end points into here._ The `CDOIDE` folder should be completely agnostic of anything on our backend (ideally, at least), and we should only interact with the rest of our systems by handing in data and callbacks (i.e., `config`, `project`, `setProject`, `setConfig`). If we need more interaction, we should skew towards enhancing that interface instead of tightly coupling code internally.


apps/src/weblab2/README.md
```diff
@@ -0,0 +1,47 @@
+# CDO IDE
```
molly-moen, line 1: thanks for adding this!

molly-moen (approved):
I pulled locally, good progress!

PR #56609, opened 2024-02-15
Datablock Storage: switch DB from channel_id to project_id

snickell:
This PR switches us to use project_id instead of channel_id for DB storage.

- As per @cnbrenci discussion with @code-dot-org/infrastructure , we want to persist the project_id instead of the channel_id to DB. Long term, it is possible we'll need to re-numerate channel_id, and the name is confusing in terms of "what is this???" anyway.
- We're not creating a foreign key to the projects table at this time, can evaluate that later (might depend on if we're same-db in the end or not)
- Shared tables are stored as `DatablockStorageTable.project_id: 0`, since we start numbering projects at 1.

### DIFF
--- a/apps/src/applab/applab.js
+++ b/apps/src/applab/applab.js
@@ -421,14 +421,11 @@ Applab.init = function (config) {
   }
   Applab.channelId = config.channel;
 
-  function shouldUseDatablockStorage() {
-    // FIXME: unfirebase, how do we implement this?
-    return true;
-  }
-
-  if (shouldUseDatablockStorage()) {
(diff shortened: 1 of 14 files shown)


snickell:
After merging, this may be useful for seeding a couple rows (`bin/dashboard-console
`) of a shared table for experimentation:
```
table = DatablockStorageTable.create(project_id: 0, table_name: "100 Birds of the World")
table.create_records([{ name: "Robin", age: 31 }, { name: "Parrot", age: 104 }])
```

PR #56612, opened 2024-02-16
Add Explanation About How to Deploy Database Migration Service Task

sureshc:
Add comments warning that merging changes to `tasks.yml` doesn't deploy them and explain how to validate and deploy the changes.

## Testing story

```
bundle exec rake stack:data:validate RAILS_ENV=production ADMIN=true

Pending update for stack `DATA-production`:
No changes
Finished stack:data:validate (less than a minute)
```

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/aws/dms/tasks.yml
+++ b/aws/dms/tasks.yml
@@ -6,7 +6,7 @@
 #
 #    VALIDATE:
 #
-#    code-dot-org % export AWS_PROFILE=codeorg-admin
+#    code-dot-org$ export AWS_PROFILE=codeorg-admin
 #    code-dot-org % bin/aws_access
 #    AWS access: GoogleSignInAdmin/suresh@code.org
 #    code-dot-org % bundle exec rake stack:data:validate RAILS_ENV=production ADMIN=true
--- a/aws/dms/tasks.yml
+++ b/aws/dms/tasks.yml
@@ -4,12 +4,15 @@
 # a Pull Request that changes this file, an Infrastructure Engineer must execute the following steps on their workstation
 # utilizing their admin AWS credentials to validate and then deploy these changes:
 #
-#    VALIDATE:
+#    VALIDATE: The output should only specify AWS::DMS::ReplicationTask Resources that will change and only the specific
+#    Tasks that are being modified in this feature branch. If there are any other changes, do not deploy, and instead
+#    determine the source of the differences between template configuration in the current commit and the state of the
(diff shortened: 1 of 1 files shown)


aws/dms/tasks.yml
```diff
@@ -1,3 +1,25 @@
+# These configuration settings are utilized by data.yml.erb to generate configuration of the AWS Database Migration
+# Service (DMS) Tasks. Merging changes to this file does NOT update the `DATA-production` Stack that provisions
+# our AWS Redshift Cluster and the DMS Replication Tasks that export data into the analytics system. Before merging
+# a Pull Request that changes this file, an Infrastructure Engineer must execute the following steps on their workstation
+# utilizing their admin AWS credentials to validate and then deploy these changes:
+#
+#    VALIDATE:
+#
+#    code-dot-org % export AWS_PROFILE=codeorg-admin
```
Hamms: tiny nit, both here and below:

I'd recommend `$` as a more-standard alternative for this documentation. Although `%` does sometimes get used as a dividing character between path and command in the bash prompt, it's unusual to do so; it took me a second to realize that was the intent.

```suggestion
#    code-dot-org$ export AWS_PROFILE=codeorg-admin
```

aws/dms/tasks.yml
```diff
@@ -1,3 +1,25 @@
+# These configuration settings are utilized by data.yml.erb to generate configuration of the AWS Database Migration
+# Service (DMS) Tasks. Merging changes to this file does NOT update the `DATA-production` Stack that provisions
+# our AWS Redshift Cluster and the DMS Replication Tasks that export data into the analytics system. Before merging
+# a Pull Request that changes this file, an Infrastructure Engineer must execute the following steps on their workstation
+# utilizing their admin AWS credentials to validate and then deploy these changes:
```
Hamms, line 5: Is it worth adding an example of what a validation error looks like, or specifying exactly what in the output the developer should be looking for to determine that validation was successful? I expect it should be relatively obvious, but might be nice to be explicit anyway.

Hamms (approved):
One question and one minor little style nit, but neither are blocking and generally this looks great; thanks for the documentation!

aws/dms/tasks.yml
sureshc: Definitely worth doing that. I added a small explanation of what expected an unxpected output of the validation Task might be.

PR #55309, opened 2023-12-05
Revert "Comment out AFE amplitude events temporarily"

hannahbergam:
Reverts code-dot-org/code-dot-org#55157

(The revert removes the commented out Amplitude events.

The second commit is the actual review- adding in the Amplitude API key to the upstream haml files.

Testing:
I no longer see the 'api key missing' message when filling out the form locally. I was able to verify the events send as expected. See photos for examples below.

I believe adding the two keys on those two upstream haml files should cover all cases. All eight events are downstream from the two haml files:

AFE_SIGN_UP_BUTTON_PRESS
AFE_SIGN_IN_BUTTON_PRESS
AFE_START
AFE_CONTINUE
AFE_SUBMIT_SCHOOL_INFO
AFE_INELIGIBLE
AFE_SUBMIT
AFE_HOMEPAGE_BANNER_SUBMIT

Screenshot 2023-12-05 at 12 16 49 PM
Screenshot 2023-12-05 at 12 25 02 PM

https://codedotorg.atlassian.net/browse/ACQ-1209

PR #56638, opened 2024-02-16
Add Amplitude API key to the user header for the "Create" menu events

TurnerRiley:
Currently on production there is a "missing API key" error on pegasus pages with the "Create" menu after this PR was merged to log click events on the menu. We've seen this issue before like in [this PR](https://github.com/code-dot-org/code-dot-org/pull/55309) so adding the Amplitude api-key should resolve this!

Error message:
missing api key

## Links
PR this follows up: [here](https://github.com/code-dot-org/code-dot-org/pull/56319)

## Testing story
Tested locally to make sure there are no errors in the console and the Amplitude events still log as expected.

Pegasus:
pegasus

Dashboard:
dashboard

## Followup work
I'm going to monitor staging and check that the staging Amplitude event counts increase once this PR hits staging to double check.

### DIFF
--- a/shared/haml/user_header.haml
+++ b/shared/haml/user_header.haml
@@ -27,6 +27,8 @@
   cookie_key = environment_specific_cookie_name('_user_type')
   user_type = request.cookies[cookie_key]
 
+%script{data: {"amplitude-api-key": CDO.safe_amplitude_api_key}}
+
 - if show_create_menu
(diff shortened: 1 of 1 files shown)


bethanyaconnor (approved):
Is there any user impact or are we just not receiving the events??

TurnerRiley:
> Is there any user impact or are we just not receiving the events??

No user impact, just not receiving the events!

PR #56651, opened 2024-02-16
Convert Blockly code to typescript part 2

molly-moen:
Convert the following Google Blockly files to typescript:
- eventHandlers
- googleBlocklyWrapper

Having `googleBlocklyWrapper` converted and (mostly) typed should unblock a lot of other conversions. It looks like a lot of changes but almost every change is just adding types or casting to types. I added PR comments where there was any code change beyond a new type or where I removed an unused parameter. The only code changes were:
- removing unused parameters
- adding null checks
- converting a couple constant objects (already in TypeScript from a prior PR) to enums for easier typing
- removing an unnecessary error throw because we were already doing a null check
- getting rid of non-functional code: one parameter that did not exist on the blockly api, and one override method that did nothing
- using instance methods rather than accessing private fields

As part of this I added a bunch of types to `blockly/types`. We now have a type for the Blockly Wrapper, as well as our customized versions of Blockly classes such as `WorkspaceSvg`. We extend and override methods in `WorkspaceSvg`, `Block`, `BlockSvg`, and `Input` so we can have custom behavior that exists in CDO Blockly (see [this slack thread](https://codedotorg.slack.com/archives/C05DK21DAHK/p1708043301061919) for details on why we do this). Now that behavior is at least documented in one place.

## Links

## Testing story
Tested locally in Sprite Lab and Google Blockly. Very little actual code has changed here!

## Follow-up work
Convert more of our Blockly code to TypeScript! I am guessing we will need to tweak some types as we go, but this should be a good start towards unblocking a lot of work.

## PR Checklist:

- [ ] Code is well-commented


apps/src/blockly/addons/contextMenu.js
```diff
@@ -291,7 +291,7 @@ export function registerHelp() {
   GoogleBlockly.ContextMenuRegistry.registry.register(helpOption);
 }
 
-const registerAllContextMenuItems = function () {
+export const registerAllContextMenuItems = function () {
```
molly-moen, line 294: needed to change the syntax here in order to import into a ts file

apps/src/blockly/constants.ts
```diff
@@ -1,10 +1,10 @@
 import {makeEnum} from '@cdo/apps/utils';
 import {parseElement as parseXmlElement} from '../xml';
 
-export const BlocklyVersion = {
-  CDO: 'CDO',
-  GOOGLE: 'Google',
-};
+export enum BlocklyVersion {
```
molly-moen, line 4: changed this and `Themes` to enums to make typing easier

apps/src/blockly/eventHandlers.ts
```diff
@@ -23,34 +28,42 @@ import {BLOCK_TYPES} from './constants';
...
     blockEvent.type === Blockly.Events.BLOCK_CHANGE &&
-    blockEvent.element === 'disabled' &&
-    !blockEvent.newValue &&
-    blockEvent.oldValue;
-  if (!blockEvent.workspaceId) {
+    (blockEvent as BlockChange).element === 'disabled' &&
+    !(blockEvent as BlockChange).newValue &&
+    (blockEvent as BlockChange).oldValue;
+
+  if (!blockEvent.blockId || !blockEvent.workspaceId) {
```
molly-moen, line 48: consolidating checks here.

apps/src/blockly/eventHandlers.ts
```diff
@@ -23,34 +28,42 @@ import {BLOCK_TYPES} from './constants';
...
   const eventWorkspace = Blockly.Workspace.getById(blockEvent.workspaceId);
-  const block = eventWorkspace.getBlockById(blockEvent.blockId);
+  const block = eventWorkspace?.getBlockById(blockEvent.blockId);
   if (
     blockEvent.type === Blockly.Events.BLOCK_MOVE ||
     blockEvent.type === Blockly.Events.BLOCK_CREATE ||
     isEnabledEvent
   ) {
-    if (!blockEvent.blockId) {
-      throw new Error('Encountered a blockEvent without a proper blockId');
```
molly-moen, line 45: I added an early return if there's no block id instead. It seems better to fail silently if we can't find the block id rather than crash the page.

apps/src/blockly/eventHandlers.ts
```diff
@@ -23,34 +28,42 @@ import {BLOCK_TYPES} from './constants';
...
-    }
     if (block) {
       updateBlockEnabled(block);
     }
   } else if (
     blockEvent.type === Blockly.Events.BLOCK_DRAG &&
     block &&
-    block.type === BLOCK_TYPES.procedureDefinition
+    block.type === BLOCK_TYPES.procedureDefinition &&
+    eventWorkspace
```
molly-moen, line 66: new null check

apps/src/blockly/eventHandlers.ts
```diff
@@ -75,10 +88,11 @@ function updateBlockEnabled(block) {
         child.setEnabled(true);
       }
     } else if (block.outputConnection || block.previousConnection) {
+      let currentBlock: Block | null = block;
```
molly-moen, line 91: the passed in parameter has to be a `Block`, but we want to support a variable here that could also be `null`

apps/src/blockly/googleBlocklyWrapper.ts
```diff
@@ -51,7 +52,7 @@ import initializeCss from './addons/cdoCss';
 import CdoConnectionChecker from './addons/cdoConnectionChecker';
 import {UNKNOWN_BLOCK} from './addons/unknownBlock';
 import {registerAllContextMenuItems} from './addons/contextMenu';
-import {ToolboxType, Themes, Renderers} from './constants';
```
molly-moen, line 54: `ToolboxType` was previously unused

apps/src/blockly/googleBlocklyWrapper.ts
```diff
@@ -93,7 +107,10 @@ const INFINITE_LOOP_TRAP =
  * This wrapper will contain all of our customizations to Google Blockly.
  * See also ./cdoBlocklyWrapper.js
  */
-const BlocklyWrapper = function (blocklyInstance) {
+const BlocklyWrapper = function (
+  this: BlocklyWrapperType,
```
molly-moen, line 111: this is typescript syntax to type the `this` variable.

apps/src/blockly/googleBlocklyWrapper.ts
```diff
@@ -324,8 +356,7 @@ function initializeBlocklyWrapper(blocklyInstance) {
   blocklyWrapper.blockly_.serialization.registry.unregister('blocks');
   blocklyWrapper.blockly_.serialization.registry.register(
     'blocks',
-    new CdoBlockSerializer(),
-    true /* opt_allowOverrides */
```
molly-moen, line 328: there is no `opt_allowOverrides` parameter on this function

apps/src/blockly/googleBlocklyWrapper.ts
```diff
@@ -420,47 +451,57 @@ function initializeBlocklyWrapper(blocklyInstance) {
...
     googleBlocklyMixin.call(this, mixinObj, true);
   };
 
-  blocklyWrapper.BlockSvg.prototype.isDisabled = function () {
-    return this.disabled;
+  const extendedBlockSvg = blocklyWrapper.BlockSvg
+    .prototype as ExtendedBlockSvg;
+
+  extendedBlockSvg.isDisabled = function () {
+    return !this.isEnabled();
```
molly-moen, line 462: `this.disabled` is a private field, but we can use the method `isEnabled()` instead.

apps/src/blockly/googleBlocklyWrapper.ts
```diff
@@ -420,47 +451,57 @@ function initializeBlocklyWrapper(blocklyInstance) {
...
-  blocklyWrapper.Input.prototype.setStrictCheck = function (check) {
+  const extendedInput = blocklyWrapper.Input.prototype as ExtendedInput;
+
+  extendedInput.setStrictCheck = function (check) {
     return this.setCheck(check);
   };
-  blocklyWrapper.Block.prototype.setStrictOutput = function (isOutput, check) {
+
+  // We use fieldRow because it is public.
+  extendedInput.getFieldRow = function () {
```
molly-moen, line 491: this is just moved from below.

apps/src/blockly/googleBlocklyWrapper.ts
```diff
@@ -469,26 +510,23 @@ function initializeBlocklyWrapper(blocklyInstance) {
...
 
   // Called by StudioApp, but only implemented for CDO Blockly.
-  blocklyWrapper.WorkspaceSvg.prototype.addUnusedBlocksHelpListener =
-    function () {};
+  extendedWorkspaceSvg.addUnusedBlocksHelpListener = function () {};
 
-  blocklyWrapper.WorkspaceSvg.prototype.getAllUsedBlocks = function () {
-    return this.getAllBlocks().filter(block => !block.disabled);
+  extendedWorkspaceSvg.getAllUsedBlocks = function () {
+    return this.getAllBlocks().filter(block => block.isEnabled());
```
molly-moen, line 524: same use of an instance method rather than a private field.

apps/src/blockly/googleBlocklyWrapper.ts
```diff
@@ -498,45 +536,39 @@ function initializeBlocklyWrapper(blocklyInstance) {
...
     this.globalVariables = variableList;
     this.getVariableMap().addVariables(variableList);
   };
 
-  blocklyWrapper.WorkspaceSvg.prototype.getContainer = function () {
+  extendedWorkspaceSvg.getContainer = function () {
     return this.svgGroup_.parentNode;
   };
 
-  const googleBlocklyBlocklyResize =
```
molly-moen, line 512: [discussed in slack](https://codedotorg.slack.com/archives/C05DK21DAHK/p1708117535064739), `this.flyout_` is not a valid field anymore, and flyouts do not have a `resize` method, so this was not doing anything.

apps/src/blockly/googleBlocklyWrapper.ts
```diff
@@ -622,24 +654,29 @@ function initializeBlocklyWrapper(blocklyInstance) {
     blocksInWorkspace
       .filter(block => block.getParent() === null)
       .forEach(block => {
-        const svgTransformList = block.svgGroup_.transform.baseVal;
+        const svgTransformList = block.getSvgRoot().transform.baseVal;
```
molly-moen, line 657: `getSvgRoot()` returns `svgGroup_`, which is a private field we should not be directly accessing.

apps/src/blockly/googleBlocklyWrapper.ts
```diff
@@ -622,24 +654,29 @@ function initializeBlocklyWrapper(blocklyInstance) {
...
         const svgTranslationX = svgTransform.matrix.e;
         svgTransform.setTranslate(svgTranslationX, 0);
       });
 
     // Shrink SVG to size of the block
-    const bbox = svg.getBBox();
-    svg.setAttribute('height', bbox.height + bbox.y);
-    svg.setAttribute('width', bbox.width + bbox.x);
+    const bbox = (svg as SVGGraphicsElement).getBBox();
+    svg.setAttribute('height', `${bbox.height + bbox.y}`);
```
molly-moen, line 665: `setAttribute` expects a string, not a number.

apps/src/blockly/customBlocks/googleBlockly/behaviorBlocks.js
```diff
@@ -208,7 +208,7 @@ GoogleBlockly.Extensions.registerMixin(
  * Derived from core Google Blockly:
  * https://github.com/google/blockly/blob/5a23c84e6ef9c0b2bbd503ad9f58fa86db1232a8/core/procedures.ts#L202-L287
  * @param {WorkspaceSvg} workspace The workspace containing procedures.
- * @returns an array of XML block elements
+ * @returns {import('blockly/core/utils/toolbox').FlyoutDefinition} an array of XML block elements
```
mikeharv, line 211: I've never seen something like this before (import statement within a JSDoc). Could you explain why we're doing this a little more? Would an alternative be to import this type at the top of the file(s)? (Like what you did for the eventHandlers file.)

apps/src/blockly/googleBlocklyWrapper.ts
```diff
@@ -73,6 +74,19 @@ import {getStore} from '@cdo/apps/redux';
 import {setFailedToGenerateCode} from '@cdo/apps/redux/blockly';
 import {handleCodeGenerationFailure} from './utils';
 import {MetricEvent} from '@cdo/apps/lib/metrics/events';
+import {
+  BlocklyWrapperType,
```
mikeharv, line 78: Nit: Is it redundant to include `"Type"` in the name of this label? I think each import from this file is a type.

apps/src/blockly/googleBlocklyWrapper.ts
mikeharv: Confirmed

apps/src/blockly/googleBlocklyWrapper.ts
mikeharv: Do we even need this at all? I'm pretty sure that the only place we are calling `isDisabled` in a Blockly context is here:

https://github.com/code-dot-org/code-dot-org/blob/65b3753faa7ed11056ef03dc08728552f7282232/apps/src/blockly/addons/cdoGenerator.js#L51

If we updated the above call to use the built-in `isEnabled`, I think we could remove this particular extended function. 

Not sure if you're try trying to avoid diffs in other files here or not. If you think it's cleaner to leave for now and do it as a cleanup that's fine with me.

apps/src/blockly/googleBlocklyWrapper.ts
```diff
@@ -420,47 +451,57 @@ function initializeBlocklyWrapper(blocklyInstance) {
...
-    return this.disabled;
+  const extendedBlockSvg = blocklyWrapper.BlockSvg
+    .prototype as ExtendedBlockSvg;
+
+  extendedBlockSvg.isDisabled = function () {
+    return !this.isEnabled();
   };
 
-  blocklyWrapper.BlockSvg.prototype.getHexColour = function () {
+  extendedBlockSvg.getHexColour = function () {
```
mikeharv, line 465: It looks like this one is never called. The comment below suggests it's expected that we wouldn't need it by now, but I'm not seeing evidence we were ever using calling it in the first place.

apps/src/blockly/googleBlocklyWrapper.ts
```diff
@@ -498,45 +536,39 @@ function initializeBlocklyWrapper(blocklyInstance) {
...
 
-  blocklyWrapper.VariableMap.prototype.addVariables = function (variableList) {
+  extendedVariableMap.addVariables = function (variableList) {
     variableList.forEach(varName => this.createVariable(varName));
   };
 
-  // TODO - used for spritelab behavior blocks
-  blocklyWrapper.Block.createProcedureDefinitionBlock = function (config) {};
+  // Used for spritelab behavior blocks.
+  // We can remove this once we are ready to no longer support sprite lab on CDO Blockly.
```
mikeharv, line 564: I'm seeing this called once in `customBlocks/cdoBlockly` and one in a Dance file for behavior definition blocks. We don't actually have behavior definition blocks in Dance, so we could probably clean this up quickly as a follow up. I don't think that we need to keep it here just for Sprite Lab on CDO Blockly because this wrapper isn't used in that circumstance any way.

apps/src/blockly/googleBlocklyWrapper.ts
```diff
@@ -498,45 +536,39 @@ function initializeBlocklyWrapper(blocklyInstance) {
...
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  (blocklyWrapper.Block as any).createProcedureDefinitionBlock = function () {};
 
-  // TODO - used to add "create a behavior" button to the toolbox
-  blocklyWrapper.Flyout.configure = function (type, config) {};
+  // In cdo this is used to add "create a behavior" button to the toolbox
+  // Once we have fully moved to Google Blockly we can remove this.
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  (blocklyWrapper.Flyout as any).configure = function () {};
```
mikeharv, line 571: Similar to above comment. There's also a reference to this in a dance file that I bet we don't need.

apps/src/blockly/googleBlocklyWrapper.ts
mikeharv: Was this just implicitly converting to a string before then? Seems better to be explicit if we can.

mikeharv (commented):
Great to see the entire wrapper getting tackled here. Just a few comments/questions.

apps/src/blockly/customBlocks/googleBlockly/behaviorBlocks.js
molly-moen: Yep, we have to use this wonky syntax because `FlyoutDefinition` is only used in a comment, so the linter will think the import is unused. If we don't import anything and just write `{FlyoutDefinition}` here the type will be `any`. Therefore we import it directly so we can utilize the type in other files. Autocomplete will fill in this syntax thankfully so you don't have to memorize how to use it!

apps/src/blockly/googleBlocklyWrapper.ts
molly-moen: I made this `BlocklyWrapperType` to avoid confusion with the variable `BlocklyWrapper` (which did not seem like a good idea to rename). If you have an idea for a better name for the type I am open to it, I couldn't think of anything.

apps/src/blockly/googleBlocklyWrapper.ts
molly-moen: Yeah I wanted to keep this as scoped as possible. I made a [follow-up ticket](https://codedotorg.atlassian.net/browse/CT-343) and will add your suggestions to it!

apps/src/blockly/googleBlocklyWrapper.ts
molly-moen: added to follow-up ticket!

apps/src/blockly/googleBlocklyWrapper.ts
molly-moen: added this to the follow up ticket!

apps/src/blockly/googleBlocklyWrapper.ts
molly-moen: Yep I think it just implicitly converted before, it's the kind of thing that works in js but typescript complains and it's probably better to use the correct type.

apps/src/blockly/customBlocks/googleBlockly/behaviorBlocks.js
mikeharv: Thank you for explaining!

apps/src/blockly/googleBlocklyWrapper.ts
mikeharv: I don't have a better idea. :D 
This works for me.

mikeharv (approved):
🎉

PR #56655, opened 2024-02-17
Apply tab styles to navigation bar

etaderhold:
This change redoes the teacher dashboard navigation bar to use the new tab layout.

Before:
image

After:
image

The `TeacherDashboardNavigation` component currently has a behavior that adds some carat widgets to the edges of the navigation bar to aid in scrolling when the links are too long to fit on the screen. It assumes a fixed page width and the carat display is not responsive to resizing. This was previously never used in practice (at least in English) as the fixed dashboard page width was wide enough to show all of the link text. I removed all that in this PR and converted it to a much simpler, functional component.

Some decisions may still need to be made about how the nav bar should behave in a narrow-screen situation. The current implementation just puts a horizontal scroll bar in there, but perhaps a second row of tabs would be preferable.

## Links

[Figma](https://www.figma.com/file/gCktlnafGYQSNJafD3FUgA/Progress-Tracking?type=design&node-id=3618-13063&mode=design&t=ldXEcZGTknZW87R5-0)
[JIRA](https://codedotorg.atlassian.net/browse/TEACH-760)

## Testing story

Added a new test to validate the links show up as expected given different permutations of properties.

## PR Checklist:

- [ ] Code is well-commented


apps/src/templates/teacherDashboard/TeacherDashboardNavigation.jsx
```diff
@@ -46,193 +41,41 @@ const teacherDashboardLinks = [
...
+  return (
+    <div id="uitest-teacher-dashboard-nav" className={styles.navContainer}>
+      {renderedLinks.map(link => (
+        <NavLink
+          key={link.url}
+          to={link.url}
+          className={styles.linkContainer}
+          activeClassName={styles.activeLinkContainer}
+        >
+          <div className={styles.link}>{link.label}</div>
```
lfryemason, line 66: I think these are pretty important places to add aria-roles and aria-labels

apps/test/unit/templates/teacherDashboard/TeacherDashboardNavigationTest.jsx
```diff
@@ -0,0 +1,79 @@
...
+
+describe('TeacherDashboardNavigation', () => {
+  it('shows standard links with no AI tutor when no props given', () => {
+    render(
+      <Router>
+        <TeacherDashboardNavigation />
+      </Router>
+    );
+
+    screen.getByText(MANAGE_STUDENTS_LABEL);
```
lfryemason, line 28: While tedious, do you think it is worth it to have at least one test that checks if each of the standard tabs is shown? Just to make sure in a future PR the `progress` tab doesn't fall off

lfryemason (commented):
Just a couple small things

apps/src/templates/teacherDashboard/TeacherDashboardNavigation.jsx
etaderhold: Reading up on `aria-label`s, I see this note in the [MDN page about them](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-label)

> Note: aria-label is intended for use on interactive elements, or elements made to be interactive via other ARIA declarations, when there is no appropriate text visible in the DOM that could be referenced as a label

These links have clear text labels, so it seems the advice here is to not use them in this case. What do you think?

As for the labels, the component renders the link text inside a `div` nested under an `a` tag, and `a` tags have [the `link` role by default](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/link_role). Does the nested `div` create a need to explicitly tag it with `aria-role="link"` again?

apps/test/unit/templates/teacherDashboard/TeacherDashboardNavigationTest.jsx
etaderhold: I have a hard time imagining how someone would remove that link accidentally. I think checking one link is sufficient.

etaderhold:
I'm changing the selected tab color to white per [this Slack thread](https://codedotorg.slack.com/archives/C0SUN2SSF/p1709077632566619) where we were advised to make the selected tab match the background color of the page below. Once more than just a closed beta page is off-white we can revisit this.

image

lfryemason (approved):

PR #56657, opened 2024-02-17
Implement get_column 

cnbrenci:
Implement get_column on the backend.

Implemented the ability to switch a project to datablock_storage and back to firebase. Keep in mind that when doing this switch, the data that exists in either storage system will stay the same. This would let us pretty easily test in a bug bash scenario without the need for a DCDO flag. Folks could make a new project, navigate to /datablock_storage/CHANNEL_ID/ and swap the storage for the project to test.  This might be especially interesting to swap over existing projects that are exclusively using stock tables, because you could switch the project over, import the tables needed for the project, and run the project.

### DIFF
--- a/dashboard/app/controllers/datablock_storage_controller.rb
+++ b/dashboard/app/controllers/datablock_storage_controller.rb
@@ -236,11 +236,13 @@ class DatablockStorageController < ApplicationController
   #   Project Use Datablock Storage API                    #
   ##########################################################
 
+  # TODO: post-firebase-cleanup, remove
   def use_datablock_storage
     ProjectUseDatablockStorage.set_data_block_storage_for!(params[:channel_id], true)
     render json: true
   end
 
+  # TODO: post-firebase-cleanup, remove
   def use_firebase_storage
     ProjectUseDatablockStorage.set_data_block_storage_for!(params[:channel_id], false)
     render json: true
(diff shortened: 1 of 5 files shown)


dashboard/app/controllers/datablock_storage_controller.rb
```diff
@@ -225,6 +232,23 @@ def clear_all_data
...
+
+  def use_datablock_storage
+    ProjectUseDatablockStorage.set_data_block_storage_for!(params[:channel_id], true)
+    render json: true
+  end
+
+  def use_firebase_storage
+    ProjectUseDatablockStorage.set_data_block_storage_for!(params[:channel_id], false)
+    render json: true
+  end
```
snickell, line 249: We should add our TODO (can't remember exact) for post-merge cleanup/removal to these

dashboard/app/models/project_use_datablock_storage.rb
```diff
@@ -23,4 +23,9 @@ def self.use_data_block_storage_for?(channel_id)
     project = Project.find_by_channel_id(channel_id)
     find_by(project: project)&.use_datablock_storage || false
   end
+
+  def self.set_data_block_storage_for!(channel_id, use_datablock_storage)
+    project = Project.find_by_channel_id(channel_id)
+    find_or_create_by(project: project).update!(use_datablock_storage: ActiveRecord::Type::Boolean.new.cast(use_datablock_storage))
+  end
```
snickell, line 30: post-merge-cleanup TODO on this one

dashboard/config/routes.rb
```diff
@@ -1143,6 +1144,10 @@
         # Datablock Storage: Channel API
         get :channel_exists
         delete :clear_all_data
+
+        # Project Use Datablock Storage API
+        put :use_datablock_storage
+        put :use_firebase_storage
```
snickell, line 1151: post merge cleanup TODO

snickell (approved):
LGTM, we'll handle frontend changes in a separate PR

PR #56673, opened 2024-02-20
[Sprite Lab] Preserve previous drawing style settings when making variable bubbles

mikeharv:
The new K5 blocks shipped at the end of last week. This morning I noticed a regression that only impacted animation frames where both a variable bubble and a speech bubble were being drawn. 

The cause of this is that the new variable button use a different `rectMode` style: https://p5js.org/reference/#/p5/rectMode

The solution is to call `push()` and `pop()` around the bubble's drawing commands. These allow us to change the style and transformation settings and later return to what you had. https://p5js.org/reference/#/p5/push

**Before:**
image

**After:**
image

Note that push and pop were already implemented for other draw utils, e.g. speech bubbles, multiline text, and validation bars.

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/p5lab/drawUtils.js
+++ b/apps/src/p5lab/drawUtils.js
@@ -42,6 +42,7 @@ export function variableBubble(p5, x, y, text) {
   const textWidthValue = textWidth + 2 * padding;
   const textHeightValue = textSizeValue + 2 * padding;
 
+  p5.push();
   p5.fill(colors.darkest_gray);
   p5.stroke('white');
   p5.strokeWeight(3);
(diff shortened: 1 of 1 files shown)


apps/src/p5lab/drawUtils.js
```diff
@@ -42,6 +42,7 @@ export function variableBubble(p5, x, y, text) {
   const textWidthValue = textWidth + 2 * padding;
   const textHeightValue = textSizeValue + 2 * padding;
 
+  p5.push();
```
molly-moen, line 45: mind adding a comment here explaining why we need this?

molly-moen (approved):
Nice fix!

PR #56674, opened 2024-02-20
AI Tutor: UI update - tutorTypeSelector rather than radio button toggles 

Erin007:
## PR Checklist:

- [ ] Code is well-commented


apps/src/code-studio/components/aiTutor/aiTutorPanel.tsx
```diff
@@ -50,15 +55,9 @@ const AITutorPanel: React.FunctionComponent<AITutorPanelProps> = ({open}) => {
     },
   ];
 
-  const onChange = (e: ChangeEvent<HTMLInputElement>) => {
-    e.preventDefault();
-    dispatch(addAIResponse(''));
-    setSelected(e.target.value);
-  };
-
```
kakiha11, line 54: @Erin007 I wasn't 100% sure about removing the dispatch line but I figured we could probably move that functionality somewhere else if it's still needed now that the onChange doesn't apply

kakiha11:
@Erin007 here is where I got the panel to with these updates! 

A coding level with validation

aiTutorAllThree

A coding level without validation

aiTutorOnlyTwo

Turns out there aren't any non-coding non-assessment levels in CSA that I could find so hard to demo the just "General Chat " possibility.

apps/src/code-studio/components/aiTutor/aiTutorPanel.tsx
Erin007: I think that's totally fine. The dispatch line was just setting the AI's response to an empty string, which you're right, we'll end up doing somewhere else if we need it.

Erin007 (commented):
I think this is a big step in the right direction and we can continue to iterate from here. Nice job, Ken! Github won't let me approve the PR because I opened it, but 👍 🚀 ✅

kakiha11 (approved):

PR #56677, opened 2024-02-20
Music: sound library updates

breville:
This PR makes changes to the Music Lab sound library system and its UI.

Screenshot 2024-02-29 at 11 01 53 PM

## Sound selection UI

- The sound selection UI now begins with two columns, the left showing "packs" (internally "folders") and the right showing the sounds inside the currently-selected "pack".
- If the URL parameter `show-sound-filters=true` is used, the selection UI will also include two `SegmentedButton` groups allowing the user to switch between a two-column "per pack" view, and a single-column "all sounds" view, and allowing the user to filter down to a single sound type, respectively.
  - When we move to the next version of the library (`"launch2024"`), we can default to showing these elements all the time.
    - In the meantime, preview these filters with `?library=launch2024&show-sound-filters=true`.
- The UI is now more accessible:
  - It now features a focus trap, ensuring that tab/shift-tab navigates only through the panel's elements. (`react-focus-trap` didn't seem to work, but `react-focus-lock` did.)  
  - The Enter key can be used to make a selection.
  - The `:focus-visible` styling has been updated from the default.
  - We should discuss further, but due to aesthetic concern, there is a hidden element which gets the initial focus.  Pressing tab once navigates to the first element that can be selected.
- The two columns now scroll to the current folder and sound, respectively, when the UI is first shown.
- There is now support for a "vocal" sound type.

## Library format

A lot of possible paths were explored in updating the library's format, both its JSON and its directory structure.  In the end, the updates are relatively modest, maintaining compatibility with existing code that uses a folder+sounds paradigm, and favouring user-friendly management of both the JSON and the directories themselves.

- The JSON file now has three top-level keys - `"instruments"`, `"kits"`, `"packs"` - each of which contain folders of sounds.  A practical benefit of separating these is the ability to collapse each individually in many modern editors.
- The directory structure gets a similar hierarchy, and will typically now look like this:
```
  /library
    /instruments
      /instrument0
      /instrument1
    /kits
      /kit0
      /kit1
    /packs
      /pack0
        /bass
        /leads
      /pack1
```
- The JSON no longer has an array of `groups` at its root, since we never ended up using this idea.
- Existing user code should continue to work with this new format.  Specifically:
  - `sound` events continue to use "folder/sample" for sound IDs.
  - `pattern` events continue to use a kit name.
  - `chord` events continue to use an instrument name.
- For our higher-level music player code, we maintain the concept of folders, each containing sounds.
  - While folders are now split up amongst three categories in the library JSON, these are all combined into a flat list of folders when loaded.  They continue to share a namespace.
- To facilitate the new directory structure, each folder now has both an ID and a path.
- The sample management system utilizes the folder path to load the sounds from the correct directories.
- Each sound in a "pack" can have an optional `path`, so that sounds can be stored in subdirectories like `/bass` and `/leads`, as shown in the above sample.
- The `MusicPlayer` now uses a helper from `MusicLibrary` to generate full sound URLs which are stored in each `SampleEvent`. The `SoundCache` uses the full URL as a key, and when it loads the sounds, it no longer generates the full URL or has to ask the library whether the sound is restricted. Instead, if it sees a restricted URL, it does the appropriate work.

## Library content

- The default library until now had an ID of `"default"` and mapped to `music-library.json` on S3.
- Any time we attempt to load that library, we now load the new default, which is `music-library-intro2024.json` on S3.
- This `"intro2024"` library uses the new format but is content-compatible with the previous default library.
- The music levels used in `/s/allthethings` for UI testing, and in `/s/music-intro-2024` for our onboarding script, have been updated in this PR to specifically use this new `"intro2024"` library.  
- A new sample library called `"launch2024"` is also on S3.  It will form the basis for our May release.  It has 5 in-house packs, and 5 artist packs, with placeholder content for now.

## Misc

- Added support for a custom base URL using the `base-asset-url` URL parameter.
- Cleaned up the content of https://studio.code.org/musiclab/menu

## Next

Some follow-up work will include:

- Localizing the new strings once we settle on their content.
- Updating the sound selection UI with additional visual elements, metadata, etc.

## Screenshots

Using library `"launch2024"` with filters showing:
Screenshot 2024-02-29 at 10 57 33 PM

Using library `"launch2024"` without filters showing:
Screenshot 2024-02-29 at 10 57 55 PM

Keyboard navigation with `:focus-visible`  styling:
Screenshot 2024-02-29 at 10 58 28 PM


apps/src/music/player/MusicPlayer.ts
```diff
@@ -371,7 +385,8 @@ export default class MusicPlayer {
       return null;
     }
 
-    return `${instrument}/${sound.src}`;
+    //   return `instruments/${instrument}/${sound.src}`;
```
sanchitmalhotra126: nit: remove

apps/src/music/views/SoundsPanel.tsx
```diff
@@ -22,21 +25,29 @@ interface FolderPanelRowProps {
...
   playingPreview,
   folder,
+  currentValue,
+  onSelect,
   onPreview,
+  currentFolderRefCallback,
 }) => {
   const previewSound = folder.sounds.find(sound => sound.preview);
-  const soundPath = previewSound && folder.path + '/' + previewSound.src;
+  const soundPath = previewSound && folder.id + '/' + previewSound.src;
```
sanchitmalhotra126, line 51: Just checking my understanding here - we still need to store these event IDs as `folder.id/sound.src` to maintain backwards compatibility, as old projects would have sounds in the format `beats/groovy_beat` and we don't want to break those, right? However then in the MusicPlayer, we replace the `folder.id/sound.src` path with a `folder.path/sound.src` path which becomes something like `paths/beats/groovy_beat`?

Once we've fully switched projects over, just thinking it would be good to unify these so we're always storing either `folder.path/sound.src` as the sample ID consistently, or maybe just storing the `sound.src` by itself

apps/src/music/views/SoundsPanel.tsx
```diff
@@ -158,50 +195,135 @@ const SoundsPanel: React.FunctionComponent<SoundsPanelProps> = ({
...
-  folders.forEach(folder => {
-    entries.push({type: 'folder', folder: folder});
-    folder.sounds.forEach(sound => {
-      entries.push({type: 'sound', folder: folder, sound: sound});
+  const libraryGroupPath = library.libraryJson.path;
+
+  const [selectedFolder, setSelectedFolder] = useState<SoundFolder>(
+    library.getFolderForSoundId(currentValue) || folders[0]
+  );
+  const [mode, setMode] = useState<string>('packs');
```
sanchitmalhotra126: minor: create type for this? `type Mode = "packs" | "sounds"`

apps/src/music/views/SoundsPanel.tsx
```diff
@@ -158,50 +195,135 @@ const SoundsPanel: React.FunctionComponent<SoundsPanelProps> = ({
...
+  const [filter, setFilter] = useState<string>('all');
```
sanchitmalhotra126: same here - and this can be a union with the existing `SoundType` type, `type Filter = 'all' | SoundType`

apps/src/music/views/SoundsPanel.tsx
```diff
@@ -158,50 +195,135 @@ const SoundsPanel: React.FunctionComponent<SoundsPanelProps> = ({
...
+
+  const currentFolderRef: React.MutableRefObject<HTMLDivElement | null> =
+    useRef(null);
+  const currentSoundRef: React.MutableRefObject<HTMLDivElement | null> =
```
sanchitmalhotra126, line 215: minor: might be cleaner to parameterize the `useRef` function call instead of typing the variable, since the type can be inferred that way, i.e. - `const currentFolderRef = useRef<HTMLDivElement>(null)`

apps/src/music/views/SoundsPanel.tsx
```diff
@@ -158,50 +195,135 @@ const SoundsPanel: React.FunctionComponent<SoundsPanelProps> = ({
...
+    useRef(null);
+
+  const onModeChange = (value: string) => {
+    setMode(value);
+  };
+
+  const onFilterChange = (value: string) => {
+    setFilter(value);
+  };
```
sanchitmalhotra126: nit: `useCallback`s for these?

apps/src/music/views/SoundsPanel.tsx
```diff
@@ -158,50 +195,135 @@ const SoundsPanel: React.FunctionComponent<SoundsPanelProps> = ({
...
+      });
     });
-  });
+  }
+
+  if (filter === 'all') {
+    rightColumnSounds = possibleSounds;
+  } else {
+    rightColumnSounds = possibleSounds.filter(
+      sound => filter === 'all' || sound.type === filter
```
sanchitmalhotra126: nit: can remove the `filter === 'all'` here, right?

apps/src/music/views/SoundsPanel.tsx
```diff
@@ -158,50 +195,135 @@ const SoundsPanel: React.FunctionComponent<SoundsPanelProps> = ({
...
+    <FocusLock>
+      <div id="sounds-panel" className={styles.soundsPanel} aria-modal>
+        <div id="hidden-item" tabIndex={0} role="button" />
+        {showSoundFilters && (
+          <div id="sounds-panel-top" className={styles.soundsPanelTop}>
+            <SegmentedButtons
+              selectedButtonValue={mode}
+              buttons={[
+                {label: 'Packs', value: 'packs'},
+                {label: 'Sounds', value: 'sounds'},
```
sanchitmalhotra126, line 273: (can be a follow-up) - localized strings?

apps/src/music/views/SoundsPanel.tsx
```diff
@@ -158,50 +195,135 @@ const SoundsPanel: React.FunctionComponent<SoundsPanelProps> = ({
...
+
+            <SegmentedButtons
+              selectedButtonValue={filter}
+              buttons={[
+                {label: 'All', value: 'all'},
+                {label: 'Beats', value: 'beat'},
+                {label: 'Bass', value: 'bass'},
+                {label: 'Leads', value: 'lead'},
+                {label: 'Effects', value: 'fx'},
+                {label: 'Vocals', value: 'vocal'},
```
sanchitmalhotra126, line 287: Same for these, localized strings?

sanchitmalhotra126 (commented):
This looks great, nice work! Just a few minor comments

apps/src/music/views/SoundsPanel.tsx
breville: > Just checking my understanding here - we still need to store these event IDs as `folder.id/sound.src` to maintain backwards compatibility, as old projects would have sounds in the format `beats/groovy_beat` and we don't want to break those, right? However then in the MusicPlayer, we replace the `folder.id/sound.src` path with a `folder.path/sound.src` path which becomes something like `paths/beats/groovy_beat`?

That is correct.  And with the latest work just merged, there is also an optional per-sound `path`, so that it could even become `packs/pack1/beats/groovy_beat`.

> Once we've fully switched projects over, just thinking it would be good to unify these so we're always storing either `folder.path/sound.src` as the sample ID consistently, or maybe just storing the `sound.src` by itself

Yeah, I went back and forth quite a bit on the ideal form for the identifiers in the user's source code, and it would be great to talk through it some more.  We could do any of the following, with pros and cons for each:

- `groovy_beat`
- `pack1/groovy_beat` (existing)
- `packs/pack1/groovy_beat`
- `packs/pack1/beats/groovy_beat`

apps/src/music/views/SoundsPanel.tsx
breville: There were some issues with making that work, and I didn't find a really crisp solution online.  I see we use this technique [here](https://github.com/code-dot-org/code-dot-org/blob/staging/apps/src/componentLibrary/checkboxDropdown/CheckboxDropdown.tsx#L78-L79).  Maybe @levadadenys has thoughts?

apps/src/music/views/SoundsPanel.tsx
breville: Done.

apps/src/music/views/SoundsPanel.tsx
breville: Done.

apps/src/music/views/SoundsPanel.tsx
breville: Done.

apps/src/music/views/SoundsPanel.tsx
breville: Good catch.

apps/src/music/views/SoundsPanel.tsx
breville: Yes, I think best to localize once we settle on this UI after a bit of usage.

apps/src/music/views/SoundsPanel.tsx
sanchitmalhotra126: Oh interesting - we do something similar in Timeline.tsx, [here](https://github.com/code-dot-org/code-dot-org/blob/07d70ff3ecca702fc76b587bf447bd8ffffb8dc1/apps/src/music/views/Timeline.tsx#L59C23-L59C29) but I could be missing something. Not a huge deal either way though!

apps/src/music/appConfig.ts
```diff
@@ -1,12 +1,14 @@
...
+type ConfigValues = {[key: string]: string};
 
 // Static config values passed in via a call to setAppConfig.
 // If these are set, we ignore URL parameters.
-let configValues = null;
+let configValues: ConfigValues | null = null;
 
 // Called to set config values.
-export function setAppConfig(values) {
+export function setAppConfig(values: ConfigValues) {
```
sanchitmalhotra126, line 11: Not required for this change, but could we remove this actually? I don't think we still use this right?

apps/src/music/player/MusicLibrary.ts
```diff
@@ -62,23 +70,40 @@ export default class MusicLibrary {
...
   }
 
+  generateSoundUrl(folder: SoundFolder, soundData: SoundData): string {
+    const baseUrl = soundData.restricted
+      ? baseAssetUrlRestricted
+      : getBaseAssetUrl();
+
+    const optionalSoundPath = soundData.path ? `${soundData.path}/` : '';
+    const url = `${baseUrl}${this.libraryJson.path}/${folder.path}/${optionalSoundPath}${soundData.src}.mp3`;
+    return url;
```
sanchitmalhotra126: nit: inline return?

apps/src/music/player/MusicPlayer.ts
```diff
@@ -287,9 +285,18 @@ export default class MusicPlayer {
...
+        this.metricsReporter.logWarning(`No folder for ${soundEvent.id}`);
+        return [];
+      }
+
+      const url = library.generateSoundUrl(folder, soundData);
+
       return [
         {
-          sampleId: soundEvent.id,
+          sampleUrl: url,
```
sanchitmalhotra126: nit: inline this (`sampleUrl: library.generateSoundUrl...`)? unless `url` is being used somewhere else

apps/src/music/player/MusicPlayer.ts
```diff
@@ -365,22 +386,23 @@ export default class MusicPlayer {
...
+    if (soundData === null) {
       this.metricsReporter.logWarning(
         `No sound for note value ${note} on instrument ${instrument}`
       );
       return null;
     }
 
-    return `${instrument}/${sound.src}`;
+    const url = library.generateSoundUrl(folder, soundData);
+    return url;
```
sanchitmalhotra126, line 401: nit: inline here as well?

apps/src/music/player/ToneJSPlayer.ts
```diff
@@ -19,7 +19,7 @@ class ToneJSPlayer implements AudioPlayer {
   private samplers: {[instrument: string]: Sampler};
   private activePlayers: Source<SourceOptions>[];
   private currentPreview: {
-    id: string;
+    url: string;
```
sanchitmalhotra126, line 22: Thanks for following through with all these name updates!

sanchitmalhotra126 (approved):
LGTM! Looking forward to getting this in!

apps/src/music/player/ToneJSPlayer.ts
breville: Sure thing.  I figured they should be done, but wanted to make sure the general approach was approved before making all of them. :)

apps/src/music/appConfig.ts
breville: Oh good call.  Let's do in a follow-up.

apps/src/music/player/MusicLibrary.ts
breville: Hah, I think I do these for easier stepping and inspecting in the debugger, but can indeed tidy them up now.

apps/src/music/player/MusicPlayer.ts
breville: Done.

apps/src/music/player/MusicPlayer.ts
breville: Done.

apps/src/music/appConfig.ts
breville: Prepped in https://github.com/code-dot-org/code-dot-org/pull/57100.

apps/src/music/views/SoundsPanel.tsx
levadadenys: @breville code you've mentioned is now stored [here](https://github.com/code-dot-org/code-dot-org/blob/1f59a11a940e1cf79d8373a934de9db74b91651e/apps/src/componentLibrary/dropdown/_CustomDropdown.tsx#L59).
First I used your approach, then tried what @sanchitmalhotra126 suggested and everything worked fine out of the box. Might depend on specific use case scenarios though

PR #56679, opened 2024-02-20
Add Amplitude events for Congrats and Certificates

bethanyaconnor:
Adds Amplitude events for the congrats page. Details are in [the ticket](https://codedotorg.atlassian.net/browse/ACQ-1495), but I did decide to remove the event `TEACHER_HOC_CONGRATS_PAGE_VISITED`, which fired even when the curriculum wasn't HoC, and replace it with a more generic event `TEACHER_VISITED_CONGRATS_PAGE`, which fires whenever a teacher views the congrats page, with info about what the page is for. 

Example events:
```
[AMPLITUDE ANALYTICS EVENT]: Teacher Hour of Code Congrats Page Visited . Payload: {"payload":{"isHocTutorial":false,"isPlCourse":true,"isK5PlCourse":false}}
[AMPLITUDE ANALYTICS EVENT]: Certificate Shared. Payload: {"payload":{"platform":"twitter"}}
[AMPLITUDE ANALYTICS EVENT]: Certificate Printed. Payload: {}
```

### DIFF
--- a/apps/src/lib/util/AnalyticsConstants.js
+++ b/apps/src/lib/util/AnalyticsConstants.js
@@ -142,9 +142,11 @@ const EVENTS = {
   VIDEO_PAUSED: 'Video Paused',
   VIDEO_ENDED: 'Video Played To Completion',
 
+  // congrats and certificates
   BATCH_CERTIFICATES_PAGE_VIEWED: 'Batch Certificates Page Viewed',
-  TEACHER_HOC_CONGRATS_PAGE_VISITED:
-    'Teacher Hour of Code Congrats Page Visited ',
+  TEACHER_VISITED_CONGRATS_PAGE: 'Teacher Visited Congrats Page',
+  CERTIFICATE_SHARED: 'Certificate Shared',
+  CERTIFICATE_PRINTED: 'Certificate Printed',
 
   // Coteacher
   COTEACHER_INVITE_SENT: 'Coteacher Invite Sent',
--- a/apps/src/lib/util/AnalyticsConstants.js
+++ b/apps/src/lib/util/AnalyticsConstants.js
@@ -146,7 +146,7 @@ const EVENTS = {
   BATCH_CERTIFICATES_PAGE_VIEWED: 'Batch Certificates Page Viewed',
   TEACHER_VISITED_CONGRATS_PAGE: 'Teacher Visited Congrats Page',
   CERTIFICATE_SHARED: 'Certificate Shared',
-  CERTIFICATE_PRINTED: 'Certificate Printed',
+  CERTIFICATE_PRINT_PAGE_VISITED: 'Certificate Print Page Visited',
 
   // Coteacher
   COTEACHER_INVITE_SENT: 'Coteacher Invite Sent',
--- a/apps/src/sites/studio/pages/congrats/index.js
+++ b/apps/src/sites/studio/pages/congrats/index.js
@@ -43,7 +43,7 @@ $(document).ready(function () {
   const mcShareLink = tryGetLocalStorage('craftHeroShareLink', '');
 
   userType === 'teacher' &&
-    analyticsReporter.sendEvent(EVENTS.TEACHER_HOC_CONGRATS_PAGE_VISITED, {
+    analyticsReporter.sendEvent(EVENTS.TEACHER_VISITED_CONGRATS_PAGE, {
       isHocTutorial,
       isPlCourse,
       isK5PlCourse,
--- a/apps/src/sites/studio/pages/congrats/index.js
+++ b/apps/src/sites/studio/pages/congrats/index.js
@@ -43,7 +43,11 @@ $(document).ready(function () {
   const mcShareLink = tryGetLocalStorage('craftHeroShareLink', '');
 
   userType === 'teacher' &&
-    analyticsReporter.sendEvent(EVENTS.TEACHER_HOC_CONGRATS_PAGE_VISITED);
+    analyticsReporter.sendEvent(EVENTS.TEACHER_HOC_CONGRATS_PAGE_VISITED, {
+      isHocTutorial,
+      isPlCourse,
+      isK5PlCourse,
+    });
 
   ReactDOM.render(
     <Provider store={store}>
--- a/apps/src/sites/studio/pages/print_certificates/show.js
+++ b/apps/src/sites/studio/pages/print_certificates/show.js
@@ -1,0 +1,6 @@
+import analyticsReporter from '@cdo/apps/lib/util/AnalyticsReporter';
+import {EVENTS} from '@cdo/apps/lib/util/AnalyticsConstants';
+
+$(() => {
+  analyticsReporter.sendEvent(EVENTS.CERTIFICATE_PRINT_PAGE_VISITED);
+});
--- a/apps/src/templates/certificates/SocialShare.jsx
+++ b/apps/src/templates/certificates/SocialShare.jsx
@@ -42,11 +42,6 @@ export default function SocialShare({
     window.dashboard?.popupWindow(e);
   };
 
-  const onPrint = () => {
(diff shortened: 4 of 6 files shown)


apps/src/sites/studio/pages/congrats/index.js
```diff
@@ -43,7 +43,11 @@ $(document).ready(function () {
   const mcShareLink = tryGetLocalStorage('craftHeroShareLink', '');
 
   userType === 'teacher' &&
-    analyticsReporter.sendEvent(EVENTS.TEACHER_HOC_CONGRATS_PAGE_VISITED);
+    analyticsReporter.sendEvent(EVENTS.TEACHER_HOC_CONGRATS_PAGE_VISITED, {
```
TurnerRiley: Should this be `TEACHER_VISITED_CONGRATS_PAGE` instead of `TEACHER_HOC_CONGRATS_PAGE_VISITED`?

apps/src/templates/certificates/SocialShare.jsx
```diff
@@ -93,7 +103,7 @@ export default function SocialShare({
           </button>
         </a>
       )}
-      <a href={print} className="social-print-link">
+      <a href={print} className="social-print-link" onClick={onPrint}>
```
TurnerRiley: Since attaching an Amplitude logging event to a link can be inconsistent with it sometimes not logging fast enough, would it be possible to have the print event trigger when the user visits the print page (I think in [this file](https://github.com/code-dot-org/code-dot-org/blob/staging/apps/src/sites/studio/pages/print_certificates/batch.js))? Then you also wouldn't need an `onClick` in an `a` tag.

bethanyaconnor: Hm, I think this link actually uses [this HAML file](https://github.com/code-dot-org/code-dot-org/blob/staging/dashboard/app/views/print_certificates/show.html.haml), which doesn't run any JS, so I'd have to add that in. I believe the function to log to Amplitude will run before the navigation happens, but I think you're right that there's a race condition here. 

@dmcavoy thoughts on whether it would be better to possibly lose events on the button click or to log when someone visits the page to print a certificate?

apps/src/sites/studio/pages/congrats/index.js
bethanyaconnor: 🤦‍♀️

bethanyaconnor: Example event:
```
[AMPLITUDE ANALYTICS EVENT]: Teacher Visited Congrats Page. Payload: {"payload":{"isHocTutorial":false,"isPlCourse":true,"isK5PlCourse":false}}
```

apps/src/templates/certificates/SocialShare.jsx
TurnerRiley: I don't think we would need JS since it's just a page visit event like [this](https://github.com/code-dot-org/code-dot-org/blob/staging/pegasus/sites.v3/code.org/public/administrators.haml#L111)! But for sure that depends on whether we would want to switch the a click event for a page visit event

bethanyaconnor: oh that does make it easier! I want to wait for Dani to weigh in though as there is a difference between the page view and the click event

dmcavoy: If the button would always open the page then I think it makes sense to just do the page view. Is there another way they would view the page?

bethanyaconnor: The only other way I know of after a quick search is the path in this video. 

https://github.com/code-dot-org/code-dot-org/assets/46464143/6fc2bb82-5b6c-4bdf-9b92-8a582bb4cc5e

dmcavoy: Cool. Lets go with view because either of those are getting the result we want

apps/src/sites/studio/pages/print_certificates/show.js
```diff
@@ -0,0 +1,6 @@
+import analyticsReporter from '@cdo/apps/lib/util/AnalyticsReporter';
+import {EVENTS} from '@cdo/apps/lib/util/AnalyticsConstants';
+
+$(() => {
+  analyticsReporter.sendEvent(EVENTS.CERTIFICATE_PRINT_PAGE_VISITED);
+});
```
kelbyhawn, line 6: Is this file a good reference for how to add Amplitude events on studio.code.org pages?

kelbyhawn (approved):
This looks good to me, but might be good to have Hannah take a look too!

apps/src/sites/studio/pages/print_certificates/show.js
bethanyaconnor: I would say so! Most dashboard pages have associated JS these days, so normally I wouldn't need to add a new file and would just add this event to the existing JS file. In this case, Turner and I debated a bit offline about whether to add a partial like we have for pegasus or just add the JS like this and settled on adding a JS file. Definitely open to push back here!

apps/src/lib/util/AnalyticsConstants.js
```diff
@@ -142,9 +142,11 @@ const EVENTS = {
   VIDEO_PAUSED: 'Video Paused',
   VIDEO_ENDED: 'Video Played To Completion',
 
+  // congrats and certificates
   BATCH_CERTIFICATES_PAGE_VIEWED: 'Batch Certificates Page Viewed',
-  TEACHER_HOC_CONGRATS_PAGE_VISITED:
-    'Teacher Hour of Code Congrats Page Visited ',
```
hannahbergam, line 147: I assume the screenshot in the description is from before this string was updated?

apps/src/templates/certificates/SocialShare.jsx
```diff
@@ -35,12 +37,15 @@ export default function SocialShare({
     );
   }, []);
 
+  const onShare = (e, platform) => {
```
hannahbergam, line 40: This is nicely extracted from the return

dashboard/app/views/print_certificates/show.html.haml
```diff
@@ -1,2 +1,4 @@
 - alt_text = @student_name ? "Certificate for #{@student_name}" : 'Certificate for Completion of One Hour of Code'
 %img{src: @certificate_image_url, alt: alt_text, width:"100%", style: "margin-top: 25px"}
+
+%script{src: webpack_asset_path('js/print_certificates/show.js')}
```
hannahbergam, line 4: I'm down with this approach! I was just looking through some haml this week and realized we didn't yet have an example of Amplitude events from dashboard haml files.

hannahbergam (approved):
LGTM!

apps/src/sites/studio/pages/print_certificates/show.js
kelbyhawn: Cool, ty! What was the reasoning for going with the JS file?

apps/src/lib/util/AnalyticsConstants.js
bethanyaconnor: yep -- I put the new event in [this message](https://github.com/code-dot-org/code-dot-org/pull/56679#discussion_r1497511377)

apps/src/sites/studio/pages/print_certificates/show.js
bethanyaconnor: We'd need to add a javascript entry point either way. The pegasus helper mostly [loads a JS file](https://github.com/code-dot-org/code-dot-org/blob/37147bc79aef7bf0c0fe83bc546f2de8b319da13/pegasus/sites.v3/code.org/views/analytics_event_log_helper.haml#L2). The thinking here was that we want dashboard pages to move away from haml in general, so most will already load some JS. There are obviously exceptions (older pages, pages like this one that's so simple it doesn't need much), but it didn't seem worth the effort to create a customizable view for those exceptions.

PR #56684, opened 2024-02-20
[Blockly] Use handleColorAndStyle for all non-migrated labs

mikeharv:
Relates to:
- https://github.com/code-dot-org/code-dot-org/pull/56589

As part of the work to standardize our approach to setting colors and styles onto blocks, a new util `handleColorAndStyle` was created for each Blockly wrapper. In Google Blockly, we prioritize styles as they are compatible with themes. CDO Blockly only support colors. In anticipation of migrating more labs to Google Blockly, we can use this new util wherever we would set a block color. In most cases, the HSV values used map smoothly to the default color for a support block style. As there were a ton of repeated HSV arrays, I created a constant to handle these. I also created an enum for the matching block styles.
In cases where a block color does not map to a style (e.g. Jigsaw, unknown/comment blocks) I left the HSV array as-is and didn't provide a style argument. 

I also left behind approximately 50 direct calls to `setHSV` in labs we no longer support (Calc/Eval and Minecraft Code Connection). Since we don't have plans to migrate these labs it didn't make sense to take time to assign styles to their blocks.

I tested this locally by opening one of each Blockly project type (as listed on the projects page), as well as spot-checking Maze and Jigsaw levels.

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/blocksCommon.js
+++ b/apps/src/blocksCommon.js
@@ -2,6 +2,8 @@
  * Defines blocks useful in multiple blockly apps
  */
 var commonMsg = require('@cdo/locale');
+var BlockStyles = require('./blockly/constants').BlockStyles;
+var BlockColors = require('./blockly/constants').BlockColors;
 
 /**
  * Install extensions to Blockly's language and JavaScript generator
@@ -38,8 +40,11 @@ function installControlsRepeatSimplified(blockly, skin) {
     // Repeat n times (internal number) with simplified UI
     init: function () {
       this.setHelpUrl(blockly.Msg.CONTROLS_REPEAT_HELPURL);
-      Blockly.cdoUtils.setHSV(this, 322, 0.9, 0.95);
-      this.setStyle('loop_blocks');
+      Blockly.cdoUtils.handleColorAndStyle(
+        this,
+        BlockColors.LOOP,
+        BlockStyles.LOOP
+      );
       this.appendDummyInput()
         .appendField(
           blockly.Msg.CONTROLS_REPEAT_TITLE_REPEAT || commonMsg.repeat()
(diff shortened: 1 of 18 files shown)


apps/src/blocksCommon.js
```diff
@@ -250,7 +279,8 @@ function installJoinBlock(blockly) {
 function installCommentBlock(blockly) {
   blockly.Blocks.comment = {
     init: function () {
-      Blockly.cdoUtils.setHSV(this, 0, 0, 0.6);
+      // Comment blocks use a hard-coded HSV color and are not compatible with themes.
+      Blockly.cdoUtils.handleColorAndStyle(this, [0, 0, 0.6]);
```
molly-moen: Would it still be worth adding a constant for the hsv for comment and unknown blocks? That way the colors are in one place.

apps/src/dance/blockly/blocks.js
```diff
@@ -205,7 +205,7 @@ export default {
         // Must be marked EDITABLE so that cloned blocks share the same var name
         fieldLabel.EDITABLE = true;
         this.setHelpUrl(Blockly.Msg.VARIABLES_GET_HELPURL);
-        Blockly.cdoUtils.setHSV(this, 631, 0.84, 0.8);
+        Blockly.cdoUtils.handleColorAndStyle(this, [631, 0.84, 0.8]);
```
molly-moen: add a constant for this color too?

apps/src/maze/beeBlocks.js
```diff
@@ -4,6 +4,8 @@
 
 var msg = require('./locale');
 var blockUtils = require('../block_utils');
+var BlockStyles = require('../blockly/constants').BlockStyles;
```
molly-moen, line 7: can we use more modern import syntax here?

molly-moen (approved):
This is great! A couple constant suggestions

apps/src/maze/beeBlocks.js
mikeharv: I was worried about potential implications of changing the existing `require` statements to imports, so I thought for now I would just match what I saw rather than try to update the whole file. Do you think it would be straight forward to just convert all requires to imports in these files?

apps/src/blocksCommon.js
mikeharv: Yeah, that makes sense. I left them out because they were only used once but it makes sense to keep them all in one place either way.

apps/src/dance/blockly/blocks.js
mikeharv: This one is interesting because the hue value is outside the expected range (0-360º), and as far as I can tell this code isn't actually used. Looking more closely, it's for behavior_get blocks, which don't actually exist in Dance. Further, I think it's a typo and was meant to be `136, 0.84, 0.8` which aligns with the hue used for behavior blocks in Sprite Lab. I can update.

apps/src/maze/beeBlocks.js
molly-moen: I _think_ it would be straightforward (worth testing to make sure). Also fine to leave it as-is to keep with the current pattern.

PR #56686, opened 2024-02-20
Follow ups to googleBlocklyWrapper TypeScript migration

molly-moen:
A few clean ups that we noticed from the `googleBlocklyWrapper` TypeScript migration:
- replace usage of `block.isDisabled` with `!block.isEnabled` (and remove custom `isDisabled` function)
- get rid of `block.getHexColour`, as it was not used
- get rid of Dance usage of `block.createProcedureDefinitionBlock` and `Flyout.configure`. These two functions have been empty functions in the google blockly wrapper, and since Dance has been on google blockly for over a year we can safely remove that usage. I kept the usage of these functions in Sprite Lab, and the empty functions in the wrapper, because removing them breaks cdo blockly Sprite Lab. Since we are early post-migration we don't want to fully break the cdo version yet (mostly for our own debugging purposes).

## Links
- jira ticket: [CT-343](https://codedotorg.atlassian.net/browse/CT-343)

## Testing story
Tested locally that Sprite Lab still behaves the same on CDO and Google Blockly, and Dance still behaves the same on Google Blockly.

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/blockly/customBlocks/cdoBlockly/spritelabBlocks.js
+++ b/apps/src/blockly/customBlocks/cdoBlockly/spritelabBlocks.js
@@ -1,4 +1,5 @@
 // This file contains customizations to CDO Blockly Sprite Lab blocks.
+// When we are ready to remove support for CDO Blockly we can remove this file.
 import i18n from '@cdo/locale';
 import {BLOCK_TYPES, NO_OPTIONS_MESSAGE} from '@cdo/apps/blockly/constants';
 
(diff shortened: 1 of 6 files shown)


apps/src/blockly/customBlocks/cdoBlockly/spritelabBlocks.js
```diff
@@ -233,6 +233,8 @@ export const blocks = {
       ];
     };
 
+    // This is only used by CDO Blockly. When we are ready to remove support
+    // for CDO Blockly we can remove this call.
```
mikeharv: This entire file is only used by CDO Blockly so the statement really extends to the entire file as well.

mikeharv (approved):
Nice!

apps/src/blockly/customBlocks/cdoBlockly/spritelabBlocks.js
molly-moen: ah good point, will update

PR #56688, opened 2024-02-20
commented out failing tests

mgc1194:
The test environment cannot load certain classes used in the i18n test.
While the test locally succeeds, so does the Drone build, when running in the Test server, we get errors.

The test causing troubles is `cdo_languages` which requires the `CdoLanguages` class.

## PR Checklist:

- [ ] Code is well-commented


breville (approved):
Not familiar with this code, but if it helps us pass DTT then that will be good.  However, the priority should be making sure we don't ship something bad due to losing test coverage.

PR #56647, opened 2024-02-16
Set up Similar Curriculum recommender

TurnerRiley:
Sets up the Similar Curriculum Recommendation function which will be used on the [Curriculum Catalog page](https://studio.code.org/catalog) in the expanded card to show a similar curriculum to the one being viewed. This PR just adds the function and scoring framework and a followup PR will implement it on the catalog page. The scoring framework matches that of the Figma flowchart here.

## Links
Spec: [here](https://docs.google.com/document/d/1Ya0FaDTugMVLxAV_2OxdYgP4kkch-BGeq4prH4wzd7I/edit)
Figma of scoring flowchart: [here](https://www.figma.com/file/JIzLzUAVjLCBKX9Jit1FRD/Curriculum-Recommendation-System?type=whiteboard&node-id=0-1&t=DWP23wsqW46KO0t8-0)

## Testing story
Adds a test for the Similar Curriculum Recommender. More tests will be added in `CurriculumCatalogTest.jsx` once it's implemented in the next PR.

## Follow-up work
Implement the recommender on the curriculum catalog page and show the top recommended similar curriculum within the expanded card.

PR #56690, opened 2024-02-20
Show Similar Recommended Curriculum on catalog card

TurnerRiley:
This PR uses the Similar Curriculum Recommender function created in [this PR](https://github.com/code-dot-org/code-dot-org/pull/56647) to display a given Curriculum Catalog card's recommended similar curriculum. The order of events in how its used are:
1. A user selects the "Quick View" button of a Curriculum Catalog card on the [Curriculum Catalog page](https://studio.code.org/catalog) to view the expanded card
2. A copy is made of the list of curricula on the page. That list then filters out the curriculum of the current card the user is looking at and any curricula that don't support any of the same grade levels that the current card does.
3. That remaining list of curricula and the current curriculum's info is passed into the Similar Curriculum Recommender function to get an array of the curricula sorted by score.
4. Return the top-scoring curriculum and display its name, image, and link under the "Related Curricula" section of the expanded card

### Examples
ex1
ex2

### Demo of recommended text button opening the corresponding expanded card
https://github.com/code-dot-org/code-dot-org/assets/56283563/e7b31103-3c9a-45b8-890e-1927bf2b4e6f

## Links
Jira ticket for showing the recommended similar curriculum on the expanded card: [here](https://codedotorg.atlassian.net/jira/software/c/projects/ACQ/boards/64?assignee=60d62161f65054006980bd71&selectedIssue=ACQ-1378)
Jira ticket for adding an end-to-end test: [here](https://codedotorg.atlassian.net/jira/software/c/projects/ACQ/boards/64?assignee=60d62161f65054006980bd71&selectedIssue=ACQ-1380)
Curriculum Catalog Figma (for the "Related Curricula" section): [here](https://www.figma.com/file/w5R6KSZF01FAysZor6j4sa/Curriculum-Catalog?type=design&node-id=0-1&mode=design&t=0FPgXod01R2VLv6L-0)

## Testing story
Local testing and adding an end-to-end test.

## Follow-up work
My immediate 3 followup tasks for the Similar Curriculum recommender are:
1. If the user is signed-in, filter out curricula they have taught before
2. Look into implementing a way to cache recommender results to avoid rerunning with the same parameters
3. Implement eyes tests


apps/src/templates/curriculumCatalog/CurriculumCatalog.jsx
```diff
@@ -71,6 +72,41 @@ const CurriculumCatalog = ({
     setExpandedCardKey(expandedCardKey === key ? null : key);
   };
 
+  // Get the top recommended similar curriculum based on the curriculum with the given
+  // curriculumKey
+  const getRecommendedSimilarCurriculum = curriculumKey => {
```
TurnerRiley, line 77: I'm defining the `getRecommendedSimilarCurriculum` function in `CurriculumCatalog.jsx` and passing it down to `ExpandedCurriculumCatalogCard.jsx` so that later I can more easily implement a way to cache recommender results.

apps/test/unit/util/curriculumRecommenderTest.js
```diff
@@ -181,29 +181,24 @@ describe('testRecommender', () => {
...
+      'fullTestCourse2' /* 10 points = hasDesiredDuration(2) + hasDesiredMarketingInitiative(2) + hasAnySchoolSubject(1) +
+                          (2 overlapping topics * overlappingDesiredTopic(2)) + publishedWithinTwoYearsAgo(1) */,
+      'fullTestCourse1' /* 9 points = hasDesiredMarketingInitiative(2) + hasAnySchoolSubject(1) + (2 overlapping topics * overlappingDesiredTopic(2))
+                          + publishedWithinOneYearAgo(2) */,
+      'fullTestCourse5' /* 6 points = hasDesiredDuration(2) + (1 overlapping subject * overlappingDesiredSchoolSubject(2)) +
+                          (1 overlapping topic * overlappingDesiredTopic(2)) [sorted highest of 6-point courses because it is featured] */,
+      'fullTestCourse3' /* 6 points = hasDesiredDuration(2) + (1 overlapping topic * overlappingDesiredTopic(2)) + hasAnySchoolSubject(1) +
+                          publishedWithinTwoYearsAgo(1) [sorted next-highest of 6-point courses by recent published_date] */,
+      'fullTestCourse4' /* 6 points = hasDesiredDuration(2) + (1 overlapping subject * overlappingDesiredSchoolSubject(2)) +
+                          (1 overlapping topic * overlappingDesiredTopic(2)) */,
```
TurnerRiley: Since the Curriculum Catalog requires a lot of non-null fields (and many of the previous test curricula had only 1-2 fields set for the purposes of testing individual scoring questions), I had to add more test curricula that had all their fields set to be able to use the same test set here and in the end-to-end test in `CurriculumCatalogTest.jsx`.

dmcavoy:
@TurnerRiley so excited to see this! I'm wondering about the use of the icon that indicates that its taking you away from the page. I would love to keep people on the page when they click the recommended curriculum. We had spec'd that it would find that curriculum in the the catalog and open the expanded card for it (and put it in focus). I'm guessing this is more work but wondering how much.

TurnerRiley:
> @TurnerRiley so excited to see this! I'm wondering about the use of the icon that indicates that its taking you away from the page. I would love to keep people on the page when they click the recommended curriculum. We had spec'd that it would find that curriculum in the the catalog and open the expanded card for it (and put it in focus). I'm guessing this is more work but wondering how much.

@dmcavoy Done! I also added a video demo in the PR description of this behavior working.

dmcavoy:
> > @TurnerRiley so excited to see this! I'm wondering about the use of the icon that indicates that its taking you away from the page. I would love to keep people on the page when they click the recommended curriculum. We had spec'd that it would find that curriculum in the the catalog and open the expanded card for it (and put it in focus). I'm guessing this is more work but wondering how much.
> 
> @dmcavoy Done! I also added a video demo in the PR description of this behavior working.

Amazing!  Thank you Turner for that fast follow up.

apps/src/templates/curriculumCatalog/CurriculumCatalog.jsx
```diff
@@ -12,6 +12,7 @@ import CurriculumCatalogFilters from './CurriculumCatalogFilters';
 import CurriculumCatalogCard from '@cdo/apps/templates/curriculumCatalog/CurriculumCatalogCard';
 import analyticsReporter from '@cdo/apps/lib/util/AnalyticsReporter';
 import {EVENTS} from '@cdo/apps/lib/util/AnalyticsConstants';
+import {getSimilarRecommendations} from '../../util/curriculumRecommender/curriculumRecommender';
```
bethanyaconnor: nit: use absolute imports here

apps/test/unit/templates/curriculumCatalog/CurriculumCatalogTest.jsx
```diff
@@ -36,6 +36,8 @@ import teacherSections, {
   setSections,
 } from '@cdo/apps/templates/teacherDashboard/teacherSectionsRedux';
 import {sections} from '../studioHomepages/fakeSectionUtils';
+import {getSimilarRecommendations} from '@cdo/apps/util/curriculumRecommender/curriculumRecommender';
+import {FULL_TEST_COURSES} from '../../util/curriculumRecommenderTestCurricula';
```
bethanyaconnor, line 40: nit: use absolute imports

apps/src/templates/curriculumCatalog/CurriculumCatalog.jsx
```diff
@@ -71,6 +72,41 @@ const CurriculumCatalog = ({
...
+    // recommendable curriculum
+    const currCurriculumIndex = recommendableCurricula.findIndex(
+      e => e.key === curriculumKey
+    );
+    const currCurriculum = recommendableCurricula.splice(
+      currCurriculumIndex,
+      1
+    )[0];
+
+    // Filter out curricula that don't support any of the same grade levels
```
bethanyaconnor: Should this not be done in the `getSimilarRecommendations` function?

apps/src/templates/curriculumCatalog/CurriculumCatalog.jsx
```diff
@@ -71,6 +72,41 @@ const CurriculumCatalog = ({
...
+    // Get top recommended similar curriculum
+    const recommendations = getSimilarRecommendations(
+      recommendableCurricula,
+      currCurriculum.duration,
+      currCurriculum.marketing_initiative,
+      currCurriculum.school_subject,
+      currCurriculum.cs_topic
+    );
+
+    return recommendations[0];
```
bethanyaconnor, line 82: Are we only showing one recommendation? I see a couple in the figmas but not sure if we landed somewhere else

apps/src/templates/curriculumCatalog/CurriculumCatalog.jsx
TurnerRiley: Will do!

apps/test/unit/templates/curriculumCatalog/CurriculumCatalogTest.jsx
TurnerRiley: Will do!

apps/src/templates/curriculumCatalog/CurriculumCatalog.jsx
TurnerRiley: This is the part of the black box definition of the recommender that I'm the least confident about. Since the recommender takes in an array of curricula to consider (so that it's possible to pass in test curricula), I figured it might make more sense to have the black box defined as passing in an array of curricula to consider and returning an array of the same length with those curricula ranked by how recommended they are. Plus, then the recommender takes fewer parameters. However, this was mostly an arbitrary decision and I'm very open to pushback if you think it makes more sense to have the filtering + ranking all encolsed within the recommender!

apps/src/templates/curriculumCatalog/CurriculumCatalog.jsx
TurnerRiley: The Figma shows 2 curricula because one will be the recommended "Similar" curriculum and one will be the recommended "Stretch" curriculum. So we should only show 1 for now!

apps/src/templates/curriculumCatalog/CurriculumCatalog.jsx
TurnerRiley: In thinking a bit more, I'm swayed towards including the filtering in the black box so I'll update this pronto!

apps/test/unit/templates/curriculumCatalog/CurriculumCatalogTest.jsx
TurnerRiley: I'm having some odd result where when I switch it to an absolute import it doesn't work. I'm noticing the other files in the test directory are referred to with relative paths, is there some kind of special formatting I need to do to import from the `test` directory or should I stick with relative for this one?

bethanyaconnor: weird. I think using the relative imports is fine in a test file if needed though

apps/src/templates/curriculumCatalog/CurriculumCatalog.jsx
bethanyaconnor: ooo thanks!

bethanyaconnor (approved):
🎉

bethanyaconnor:
Just checking that it's fine that this will now be shown to users? I know you have some PRs queued on this, so I want to check that it's okay that this launches before those (ie filtering out already taught curriculum)

TurnerRiley:
> Just checking that it's fine that this will now be shown to users? I know you have some PRs queued on this, so I want to check that it's okay that this launches before those (ie filtering out already taught curriculum)

Yep! Everything following are smaller tweaks but the basic functionality is all here!

PR #56693, opened 2024-02-20
upgrade a student to a teacher if the LTI user is a teacher but on code.org the user is a student.

stephenliang:
upgrade a student to a teacher if the LTI user is a teacher but on code.org the user is a student

## Links

- [jira ticket](https://codedotorg.atlassian.net/browse/P20-677)

## Testing story

Tests added:

1. When authenticating user,  if LTI role contains instructor and on Code.org, the user is a student, render the "upgrade account to teacher account" popup.
2. If e-mail is present, prefill this popup with the email address.
3.  Upon upgrade, the user is updated to a teacher account.

[LMS Upgrade to Teacher.webm](https://github.com/code-dot-org/code-dot-org/assets/538214/6d36c1cb-27f7-4064-b8c3-77f1c0ac8894)

## Privacy

No additional privacy considerations are needed, the email is optionally provided by the LMS.

## PR Checklist:

- [x] Code is well-commented


dashboard/lib/policies/lti.rb
```diff
@@ -69,13 +69,19 @@ module AccessTokenScopes
 
   MAX_COURSE_MEMBERSHIP = 650
 
-  def self.get_account_type(id_token)
-    id_token[LTI_ROLES_KEY].each do |role|
+  def self.get_account_type(roles)
```
nicklathe, line 72: You're probably working on it (as this is a draft) but doing a search in the repo, this function gets called twice in `lti_test.rb`, so you'll want to update the tests with the new input for this function.

dashboard/app/controllers/lti_v1_controller.rb
```diff
@@ -123,6 +123,8 @@ def authenticate
       resource_link_id = decoded_jwt[Policies::Lti::LTI_RESOURCE_LINK_CLAIM][:id]
       deployment_id = decoded_jwt[Policies::Lti::LTI_DEPLOYMENT_ID_CLAIM]
       deployment = Queries::Lti.get_deployment(integration.id, deployment_id)
+      nrps_user_type = Policies::Lti.get_account_type(decoded_jwt[Policies::Lti::LTI_ROLES_KEY])
```
nicklathe: Since this is in the context of an LTI Launch, NRPS doesn't come into play yet. I would prefer we call this variable `lti_user_type` or `account_type`, or even `user_type`.

dashboard/lib/policies/lti.rb
nicklathe: Thinking about this more, I wonder if it's cleaner to leave this as it was, as it arguably makes calling the function cleaner, by passing in `id_token` rather than `id_token[Policies::Lti::LTI_ROLES_KEY]` from another file. I could be convinced either way, wondering what your reason was for changing this?

stephenliang: I added a usage in `services/lti.rb` which pulls the user type based on the roles from the NRPS response instead of the id token. Since this object is different than the one provided by the id token, I thought it'd be better to make this def a little more generic to accept the roles array rather than the entire object.

dashboard/app/controllers/lti_v1_controller.rb
stephenliang: Good call! I went with `lti_account_type` to differentiate it against other account types.

dashboard/lib/policies/lti.rb
nicklathe: Ah yes, sorry. Missed that usage down in `services/lti.rb`. This change looks good to me 👍

dashboard/config/locales/en.yml
```diff
@@ -1434,3 +1434,8 @@ en:
       success_body: "Great news! Your LMS has been registered with Code.org, and you can continue installing and using your LMS integration. You should receive a confirmation email shortly."
       success_header: "Thank you for registering your LMS!"
       register: "Register LMS"
+    upgrade_to_teacher_account:
+      heading: "Upgrade to a teacher account"
+      description: "Looks like you are logging in as a teacher. Please fill out the following information to finish creating your teacher account. Code.org will never rent, sell, or share this information with any third parties."
+      error:
+        missing_email: "Missing required parameter: Email"
```
carl-codeorg: The linter might complain without a newline at the end of the file

dashboard/lib/policies/lti.rb
```diff
@@ -69,13 +69,19 @@ module AccessTokenScopes
...
+    roles.each do |role|
       return User::TYPE_TEACHER if TEACHER_ROLES.include? role
     end
     return User::TYPE_STUDENT
   end
 
+  def self.get_email(id_token)
```
carl-codeorg: There is some [similar functionality](https://github.com/code-dot-org/code-dot-org/blob/staging/dashboard/lib/services/lti.rb#L69-L71) implemented in `services/lti.rb` that can be used to fetch a claim from the JWT, either at the top-level or from the custom claims. If it's sufficient, we might not need to implement a separate function for retrieving email.

stephenliang: I updated it to use `get_claim`!

carl-codeorg (approved):
LGTM!

nicklathe (approved):
LGTM :shipit:

PR #56701, opened 2024-02-21
Added a levelbuilder field for AI Tutor and a property to Level model

kakiha11:
Added a field into levelbuilder for all level types that have a "Help and Tips" section. 

Screenshot 2024-02-20 at 10 52 00 PM

When checking this box, the `true` gets assigned the property `ai_tutor_available` for that level

Screenshot 2024-02-20 at 10 59 41 PM

## Follow-up work

We may want to think about moving the section in levelbuilder to somewhere even more universal like "Bubble Choice and Lesson Extras" which appears even in multi levels and other non-app levels.

## PR Checklist:

- [ ] Code is well-commented


dashboard/app/views/levels/editors/fields/_help_and_tips.html.haml
```diff
@@ -18,3 +18,7 @@
       .span8
         - @level.reference_links = [''] unless @level.reference_links.try(:present?)
         = text_area_tag 'level[reference_links]', @level.reference_links.join("\r\n"), rows: 6, class: "input-block-level", placeholder: 'Additional Reference Link'
+
+  .field
+    = render partial: 'levels/editors/fields/checkboxes', locals: {f: f, field_name: :ai_tutor_available, description: "AI Tutor Available"}
+
```
molly-moen: nit: add a blank line to the end of this file

dashboard/app/views/levels/editors/fields/_help_and_tips.html.haml
```diff
@@ -18,3 +18,7 @@
       .span8
         - @level.reference_links = [''] unless @level.reference_links.try(:present?)
         = text_area_tag 'level[reference_links]', @level.reference_links.join("\r\n"), rows: 6, class: "input-block-level", placeholder: 'Additional Reference Link'
+
+  .field
+    = render partial: 'levels/editors/fields/checkboxes', locals: {f: f, field_name: :ai_tutor_available, description: "AI Tutor Available"}
```
molly-moen: Will AI tutor only show up for CSA levels or does this potentially enable it for all types of levels?

dashboard/app/views/levels/editors/fields/_help_and_tips.html.haml
molly-moen: actually nvm, looks like you did this. Not sure why git is showing the no blank line symbol

dashboard/app/views/levels/editors/fields/_help_and_tips.html.haml
kakiha11: This enables it for all types of lab levels

dashboard/app/views/levels/editors/fields/_help_and_tips.html.haml
kakiha11: @molly-moen Do you have an auto-formatter for haml files that you use? Looks like Prettier doesn't cover this type

dashboard/app/views/levels/editors/fields/_help_and_tips.html.haml
molly-moen: Is there a risk of it showing up in a lab type that doesn't support it or where we don't want it to show up? Or does only Java Lab support it right now?

kakiha11: Good question. It's all under an experiment now but I can't find anything in my quickish look to see something that would strictly limit it to JavaLab if a teacher was enrolled in the experiment and was teaching/assigning levels from other labs. @Erin007 Would you describe AI Tutor as only supported in Java Lab right now?

mikeharv: I was curious about this too. Here is the logic: https://github.com/code-dot-org/code-dot-org/blob/b7f380da20f7f295484502739fbb439677cdd28b/dashboard/app/models/levels/level.rb#L697-L700

I pulled the branch and confirmed that the new field does show up in some other level types, e.g. App Lab, Play Lab.

kakiha11: Yes, true Mike. However, even though we've added this to the help and tips section of levelbuilder, the AI Tutor isn't something that is going to show up in the Help and Tips tab on a level, it has it's own UI similar to AI TA with the little AI Bot button down the bottom left. Our thought about enabling it for more level types was that if CSA curriculum writers wanted to enable/disable the tutor for non-Java Lab levels like a multi or free response level we wanted to make that an option.

These comments are making me it's worth creating an entirely new section on levelbuilder that we can control more precisely where it shows up and update as AI Tutor potentially expands beyond CSA.

mikeharv: Do we know all of the level types used in CSA? Love the idea of making this a distinct section since then we'd have full control over the level types that can use it.

kakiha11: Off the top of my head CSA has JavaLab, multi, matching, and free response. I don't think there are any others. The other question I can ask of Jamila though is: are there any multi, matching, or free response that aren't marked/will be marked as assessments in CSA? If the answer is no, we can just make this a Javalab thing for now since assessment levels are already disabled for AI Tutor.

Erin007: All good questions! Here are my thoughts: 

-  In addition to the points raised here, I'm in favor of moving the AI Tutor available checkbox to its own area of the level edit page because the vision is that eventually Levelbuilders will also be able to edit system prompts on a per level basis, which I suspect could be done via a text input box that would also make sense to put in an "AI Tutor management" area of the level edit page 

- I think it's more accurate to say that AI Tutor is available for CSA levels rather than just Javalab levels at the moment. @kakiha11 I like your suggestion about determining if there are any non-Javalab, non-assessment levels in CSA where we'd want the checkbox to show.  Based on what you find out, it makes sense to me to do something similar to what @mikeharv linked where the logic is either 
```
def show_ai_tutor_managment_in_level_editor? 
    is_a?(Javalab)
end 
```
OR if we need to support more level types but still only in CSA scripts 
```
def show_ai_tutor_managment_in_level_editor? 
   script_levels.any? {|sl| sl.script.csa?}
end 
```
The second case seems tricky if the flow of levelbuilders is to build levels and _then_ associate them with scripts. When in the process of curriculum development a levelbuilder would make AI Tutor available on a level?

Regardless, I like the idea of some kind of `show_ai_tutor_managment_in_level_editor? ` method that we can modify over time as AI Tutor becomes more widely available (eventually...maybe? 🙃)

Erin007: The logic that hides/shows AI Tutor on a level is set here: https://github.com/code-dot-org/code-dot-org/blob/23c7a6e770f762520a03eb862b20a474dbcc9636/dashboard/app/views/levels/show.html.haml#L89

So even if a levelbuilder accidentally set AI Tutor available for a level we weren't anticipating there'd be no harm until we changed the logic in levels/show

dashboard/app/views/levels/editors/fields/_help_and_tips.html.haml
molly-moen: I don't think so :( I don't touch haml often

molly-moen (approved):
Nice!

Erin007:
Wahoo! @kakiha11 tiny request can you add a 📷 screenshot of the new checkbox in the new location either in the PR description or comments just so it's extra clear for future spelunkers where we landed. Thanks!

kakiha11:
In the end, decided to put the field only in JavaLab level editor for now with it's own field that lives between "Example Solutions" and "Sharing"

image

PR #56703, opened 2024-02-21
Upgrade eslint and @typescript-eslint

snickell:
Fixes warnings that appeared after https://github.com/code-dot-org/code-dot-org/pull/56067 when running `yarn lint` (now its silent):
```
Deprecation warnings:
 - The "property-case" rule is deprecated.

=============

WARNING: You are currently running a version of TypeScript which is not officially supported by @typescript-eslint/typescript-estree.

You may find that it works just fine, or you may not.

SUPPORTED TYPESCRIPT VERSIONS: >=3.3.1 <5.1.0

YOUR TYPESCRIPT VERSION: 5.3.3

Please only submit bug reports when using the officially supported version.

=============

Please only submit bug reports when using the officially supported version.
```

1. Fixed `yarn lint` TS warning by upgrading @typescript-eslint packages to latest: 7.0.2
2. Upgrading @typescript-eslint required upgrading eslint (used latest: 8.56.0)
3. Upgrading eslint to 8.x required addressing breaking plugin changes in `apps/.eslintCustomMessagesFormatter.js`, see https://eslint.org/docs/latest/use/migrating-to-7.0.0#deprecate-cliengine
4. Fixed several new `yarn lint` warnings in apps/ sourcecode
5. Fixed `"property-case" rule is deprecated` warning by removing `property-case` rule from `stylelint.config.js`

PR #56704, opened 2024-02-21
package.json: sort devDependencies from dependencies

snickell:
We had a bunch of package.json devDependencies sneak into the dependencies section: like typescript or source-map-loader

We also had a BUNCH of things in devDependencies that don't appear to be build/test related: like react or coremirror.

This PR is my best attempt to split these into the appropriate categories by skimming.

The way we do our webpack (and since nothing depends on our package) the difference between these isn't necessarily huge: the best proof of this is that everything was working with everything in devDependencies. Still, it might be tidier to keep them moderately well separated, and at least it'll make it clearer where to add my-new-dep.


apps/package.json
```diff
@@ -50,20 +50,7 @@
     "@babel/polyfill": "^7.12.1",
```
snickell, line 50: Starting here we're in the devDependencies section of package.json, red here means I moved something TO deps FROM devDeps

apps/package.json
```diff
@@ -253,8 +168,20 @@
     "@blockly/plugin-scroll-options": "~5.0.8",
```
snickell, line 174: Starting here we're in the dependencies section of package.json, red here means I moved something TO devDeps FROM deps

apps/package.json
```diff
@@ -112,9 +90,7 @@
     "eslint-plugin-react-hooks": "^4.6.0",
     "eslint-plugin-storybook": "^0.6.15",
     "events": "^3.3.0",
-    "fast-memoize": "2.0.2",
     "file-loader": "^6.2.0",
-    "firebase": "^2.4.2",
     "firebase-bolt": "^0.8.0",
     "firebase-mock": "^1.0.5",
     "firmata": "^2.2.0",
```
fisher-alice: I wonder if `firmata` could be moved to dependencies. https://github.com/code-dot-org/code-dot-org/blob/33615dc476053b78d47deda1a63aa46a6b2eb834/apps/src/lib/kits/maker/boards/circuitPlayground/CircuitPlaygroundBoard.js#L6

snickell: Oh! Probably, I didn't know what `firmata` was, but I see you're the expert :-)

snickell: done!

apps/webpack.config.js
```diff
@@ -136,6 +136,7 @@ const nodePolyfillConfig = {
       stream: require.resolve('stream-browserify'),
       timers: require.resolve('timers-browserify'),
       crypto: require.resolve('crypto-browserify'),
+      vm: require.resolve('vm-browserify'),
```
snickell, line 139: This fixes a `yarn build` webpack warning I /thought/ I had generated on my branch as a result of moving deps, but I'm actually seeing it on other branches too, I just hadn't noticed. Anyway, this fixes it.

apps/package.json
```diff
@@ -269,6 +195,7 @@
     "@dsco_/link": "^1.1.2",
     "@lezer/highlight": "^1.2.0",
     "@microsoft/immersive-reader-sdk": "^1.1.0",
+    "@react-bootstrap/pagination": "^1.0.0",
     "@reduxjs/toolkit": "^1.9.3",
     "@types/chai": "^4.3.5",
```
jmkulwik: Since Chai, Mocha, and Sinon are all used for testing, should the @types/chai, etc. all be up in devDependencies?

apps/package.json
```diff
@@ -279,54 +206,128 @@
...
+    "lazysizes": "^4.0.0-rc1",
+    "loadable-components": "2.2.3",
+    "lodash": "^4.17.21",
     "memoize-one": "^5.1.1",
+    "messageformat": "2.3.0",
     "ml-knn": "^3.0.0",
+    "moment": "^2.29.4",
+    "newrelic": "^1.27.2",
     "object-fit-images": "^3.2.3",
     "pa11y-ci": "^2.3.0",
```
jmkulwik, line 254: devDependencies?

apps/package.json
```diff
@@ -279,54 +206,128 @@
     "@typescript-eslint/eslint-plugin": "^5.57.0",
     "@typescript-eslint/parser": "^5.57.0",
```
jmkulwik: Should these two typescript linters be moved up?

apps/package.json
```diff
@@ -279,54 +206,128 @@
...
     "prop-types": "^15.6.2",
+    "pusher-js": "4.1.0",
     "pyodide": "^0.24.1",
     "qrcode.react": "^0.8.0",
+    "qtip2": "2.2.0",
     "query-string": "4.1.0",
+    "radium": "^0.25.2",
+    "react": "~16.14.0",
     "react-ace": "^9.2.1",
+    "react-addons-test-utils": "~15.4.0",
```
jmkulwik: Why move this one down?

apps/package.json
```diff
@@ -279,54 +206,128 @@
...
     "timers-browserify": "^2.0.12",
-    "ts-loader": "^9.4.2",
-    "typescript": "^5.3.3",
     "usb": "^2.8.1",
+    "video.js": "7.6.6",
+    "vm-browserify": "^1.1.2",
+    "vmsg": "^0.3.6",
+    "web-streams-polyfill": "^3.1.0",
     "wgxpath": "^1.2.0",
     "whatwg-fetch": "^2.0.3"
```
jmkulwik, line 333: Do you know if there's any way to check if any of these are unused? Might be nice to clear a few out if we can.

apps/package.json
snickell: That makes sense to me, I'll move them and see if anything freaks out :)

apps/package.json
snickell: My thinking was: while newrelic is a tool for devs, as far as I could tell (a little hard since I couldn't actually find the import/require for newrelic, all the accesses were to `window.newrelic`) its exclusively used at runtime vs from webpack or test code. You would know more, does that match your understanding?

apps/package.json
snickell: 🦅 eyes ;-) I split this PR out of https://github.com/code-dot-org/code-dot-org/pull/56703, which was all about upgrading these (and I noticed they should be moved in the process, and one thing led to another...). I didn't move them here just because I moved them there, and didn't feel like a merge conflict.

apps/package.json
snickell: Totally agree, a number of these I couldn't quickly find where they were being used. That said, this is a big list, and I erred on the side of a non-breaking change here. 

I think we should address removals in a followup PR, since I think I'd like the ability to easily revert one of those if something breaks (vs revert this comparatively safer change). That make sense to you?

apps/package.json
snickell: oopsy... fixed!

apps/package.json
snickell: Just merged #56703

apps/package.json
snickell: seemed to work ok, fixed!

molly-moen (approved):
This looks good to me!

apps/package.json
jmkulwik: Works for me! I started doing this a bit today. Doubt I'll finish, but it's nice to clear a few out.

apps/package.json
jmkulwik: Oops, I meant the pa11y-ci package.

PR #56707, opened 2024-02-21
Generate certificates for units within courses

bethanyaconnor:
Allow certificates to be generated for units within unit groups. This moves us away from strictly tying certificates to course versions and will allow us to generate a certificate for a unit within a course.

In the case of a certificate of a unit within a course, we'll also show the course name. Some examples are below:

CSD Unit 1:
csd-unit1-certificate

Teaching CSD, Module 1:
teaching-csd-module-1-certificate

CourseA (unchanged):
coursea-certificate


dashboard/app/controllers/certificate_images_controller.rb
```diff
@@ -27,8 +27,8 @@ def show
     # code tutorial.
     course_name = recognized_course_name?(data['course']) ? data['course'] : 'hourofcode'
 
-    course_version = CurriculumHelper.find_matching_course_version(course_name)
-    course_title = course_version&.localized_title
+    unit_or_unit_group = CurriculumHelper.find_matching_unit_or_unit_group(course_name)
+    course_title = unit_or_unit_group&.localized_title
```
hannahbergam, line 31: I assume localized_title still works on a unit_or_unit_group and not just course_version?

dashboard/app/helpers/curriculum_helper.rb
```diff
@@ -26,12 +26,20 @@ def validate_key_format
...
+  # retrieves the course version matching the UnitGroup or Unit with a name of course_name
   def self.find_matching_course_version(course_name)
     matching_unit_group = UnitGroup.get_from_cache(course_name)
     return matching_unit_group.course_version if matching_unit_group
     matching_standalone_course = Unit.get_from_cache(course_name, raise_exceptions: false)
     return matching_standalone_course.course_version if matching_standalone_course&.is_course
     return nil
   end
+
+  # retrieves a unit or unit group matching course name
```
hannahbergam, line 38: I think I understand the necessity of this in addition to the function above because we won't necessarily know if it's a unit or unitgroup? But we know from the start if we want unit/unit group or the course version?

dashboard/lib/certificate_image.rb
```diff
@@ -161,25 +161,53 @@ def self.create_course_certificate_image(name, course = nil, donor_name = nil, c
...
       image = Magick::Image.read(path).first
-      apply_text(image, name, 75, 'Helvetica bold', 'rgb(118,101,160)', 0, -135, CERT_NAME_AREA_WIDTH, CERT_NAME_AREA_HEIGHT)
-      # The area in pixels which will display the course title.
-      course_title_width = 1000
-      course_title_height = 60
-      apply_text(image, course_title, 47, 'Helvetica bold', 'rgb(29,173,186)', 0, 15, course_title_width, course_title_height)
-      donor_text_y_offset = 447
+      if template_file == 'self_paced_pl_certificate.png'
+        apply_text(image, name, 62, 'Helvetica bold', 'rgb(118,101,160)', 0, -248, CERT_NAME_AREA_WIDTH, 70)
+
```
hannahbergam: So most of this is copied from above- no need for course_title_height?

dashboard/lib/certificate_image.rb
```diff
@@ -161,25 +161,53 @@ def self.create_course_certificate_image(name, course = nil, donor_name = nil, c
...
+        course_title_width = 1400
+        if unit_or_unit_group.is_a?(Unit) && unit_or_unit_group.unit_group.present?
```
hannahbergam: Just because I'm confusing myself here- what happens here if it is a unit but not a unit_group?

dashboard/lib/certificate_image.rb
```diff
@@ -161,25 +161,53 @@ def self.create_course_certificate_image(name, course = nil, donor_name = nil, c
...
+        else
+          course_title_height = 171
+          apply_text(image, course_title, 62, 'Helvetica bold', 'rgb(29,173,186)', 0, 0, course_title_width, course_title_height)
+        end
+
+        total_minutes = unit_or_unit_group&.duration_in_minutes || 0
+        total_hours = (total_minutes / 60).floor
+        apply_text(image, total_hours.to_s, 30, 'Times bold', 'rgb(87,87,87)', -248, 124, 80, 30)
+        donor_text_y_offset = 611
+      else # all other courses use a certificate image where the course name is also blank
```
hannahbergam: Is this else case a direct copy/paste of the lines above but with a different image?

hannahbergam (commented):
Wow this was a bear logic-wise! I left a few questions, I suspect only because it was confusing trying to get acquainted with the space in a first read through. I don't think they're blocking, but would love to hear your thoughts!

dashboard/app/controllers/certificate_images_controller.rb
bethanyaconnor: Yes! course_version actually [delegates that function](https://github.com/code-dot-org/code-dot-org/blob/e4e4c475df0449dd304fa68d20cf87ee1289fd4e/dashboard/app/models/course_version.rb#L62) to either the unit or the unit group anyway

dashboard/app/helpers/curriculum_helper.rb
bethanyaconnor: Yes, pretty much. We don't know whether we want the content root of a course version (ie the unit group or standalone unit). This function will also find, say, `csd2-2022`, which the above function would not.

Definitely open to suggestions on how to improve the comments here!

dashboard/lib/certificate_image.rb
bethanyaconnor: The width is consistent across all these scenarios, but the height is different based on the conditions below.

dashboard/lib/certificate_image.rb
bethanyaconnor: So, there three cases:
- unit in a unit group
- standalone unit
- unit group

In the first case, we'll put both the unit name and the unit group name on the certificate. In the second and third case, we only put the name of `unit_or_unit_group` (no second name). I'll add a comment!

dashboard/lib/certificate_image.rb
bethanyaconnor: Yes, mostly. The heights and widths of the text also differ but I wonder if there's a way to pull out those constants and have this logic be shared 🤔

bethanyaconnor:
Okay, I made two changes:
- cddeba2d89840e4654f30364fdf71f5fde79fa2e adds a comment. Definitely open to feedback on wording/content but this is a pretty straightforward commit :)
- 56194b3524590ac1691e740f1dd76c39b04136cf refactors the logic to be shared for non-prefilled templates. I'm very curious for your thoughts, but it was surprisingly easy to combine (all the certificates pretty much "just worked")

<details>
<summary>Certificate images with change</summary>

<details>
<summary>PL</summary>

pl-unitgroup-cert
pl-unitgroup-and-unit-cert

</details>

<details>
<summary>Non-PL</summary>

csd-cert
csd1-cert
coursea-cert

</details>

</details>

dashboard/lib/certificate_image.rb
```diff
@@ -161,25 +161,125 @@ def self.create_course_certificate_image(name, course = nil, donor_name = nil, c
...
+      unit_or_unit_group = CurriculumHelper.find_matching_unit_or_unit_group(course)
       image = Magick::Image.read(path).first
-      apply_text(image, name, 75, 'Helvetica bold', 'rgb(118,101,160)', 0, -135, CERT_NAME_AREA_WIDTH, CERT_NAME_AREA_HEIGHT)
-      # The area in pixels which will display the course title.
-      course_title_width = 1000
-      course_title_height = 60
-      apply_text(image, course_title, 47, 'Helvetica bold', 'rgb(29,173,186)', 0, 15, course_title_width, course_title_height)
-      donor_text_y_offset = 447
+
+      text_constants = {
```
bethanyaconnor, line 168: I also wonder if these constants may have a better spot than here 🤔

hannahbergam: This seems like they could live in some sort of certificates helper file, but that's definitely a non-blocking (and preference, so feel free to disregard) nit!

dashboard/lib/certificate_image.rb
```diff
@@ -161,25 +161,125 @@ def self.create_course_certificate_image(name, course = nil, donor_name = nil, c
...
+          donor_text_y_offset: 447,
+        },
+      }
+      cert_text_constants = text_constants[template_file.to_sym]
+
+      name_constants = cert_text_constants[:name]
+      apply_text(image, name, name_constants[:font_size], 'Helvetica bold', 'rgb(118,101,160)', name_constants[:x_offset], name_constants[:y_offset], name_constants[:width], name_constants[:height])
+
+      # When we have a unit within a unit_group, we want to display both the unit and unit_group titles.
+      # When we have a standalone unit or the unit group, we only display the localized title of unit_or_unit_group.
```
hannahbergam, line 232: Thank you for refactoring this! It's easier for me to follow now🎉

hannahbergam (approved):
Thank you for your work on this! LGTM

dashboard/lib/certificate_image.rb
bethanyaconnor: Yay! Thank you for the great review 😀

bethanyaconnor:
> This seems like they could live in some sort of certificates helper file, but that's definitely a non-blocking (and preference, so feel free to disregard) nit!

I also don't love them in the middle of the function, but I think I'm going to merge as-is as this is on a somewhat tight deadline. I might follow up with a PR to move this somewhere though!

PR #56711, opened 2024-02-21
Flaky eyes test Teacher dashboard assessments tab survey submissions

vijayamanohararaj:
Eyes test `Firefox_teacher_tools_teacher_dashboard_teacher_dashboard_assessments2_Assessments tab survey submissions` intermittently fails with `ReferenceError: $ is not defined` error when waiting for an element on teacher dashboard page.

### Root cause analysis:
Spot checked multiple failure examples (https://app.saucelabs.com/tests/770b726d37fe4e8d9c9be7d3ab95d017#308, https://app.saucelabs.com/tests/9b28115386bd480f8adfa517907b6a89#280) of this test in sauce labs and noticed the below pattern,

1. Checking for jQuery loaded `"return (typeof jQuery !== 'undefined');"` returns `true`
image

2. Checking for element visibility `"return $(\"#classroom-sections\").is(':visible') && $(\"#classroom-sections\").css('visibility') !== 'hidden';"` returns `false` (this happens 0 or more times)
image

3. Same call to check for element visibility throws error that jQuery isn't loaded. 
 image

Looking at the screen shots for these steps, we see that when step #2 is executing, it is still in the section set up page and not transitioned to the dashboard. So, the jQuery load check isn't happening on the correct page leading to this issue.

### Fix
Wait for the new page to load after hitting save button.
 
## Links

JIRA https://codedotorg.atlassian.net/browse/TEACH-830?atlOrigin=eyJpIjoiZjUwZTc3NmRiNDVkNDNlMWExNDlmNDEzZGU4NDQ3ZGYiLCJwIjoiaiJ9

## Testing story

Validated the teacher dashboard test locally. Relying on drone to catch any issues this change causes on other tests.

## Deployment strategy
Regular DTP

## Follow-up work

None at this time. Curious if folks think we should follow up this change in other jQueries if this improves reliability

## Privacy

N/A

## Security

N/A

## Caching

N/A

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/dashboard/test/ui/features/step_definitions/steps.rb
+++ b/dashboard/test/ui/features/step_definitions/steps.rb
@@ -236,8 +236,13 @@ When /^I wait until (?:element )?"([^"]*)" is (not )?checked$/ do |selector, neg
   wait_until {@browser.execute_script("return $(\"#{selector}\").is(':checked');") == negation.nil?}
 end
 
+def jquery_is_defined
+  "(typeof jQuery !== 'undefined') && ($ !== 'undefined')"
+end
+
 def jquery_is_element_visible(selector)
-  "return (typeof jQuery !== 'undefined') && $(#{selector.dump}).is(':visible') && $(#{selector.dump}).css('visibility') !== 'hidden';"
+  "return #{jquery_is_defined} &&" +
+  " $(#{selector.dump}).is(':visible') && $(#{selector.dump}).css('visibility') !== 'hidden';"
 end
 
 def jquery_is_element_displayed(selector)
@@ -960,7 +965,7 @@ end
 
 def wait_for_jquery
   wait_until do
-    @browser.execute_script("return (typeof jQuery !== 'undefined');")
+    @browser.execute_script("return #{jquery_is_defined};")
   rescue Selenium::WebDriver::Error::ScriptTimeoutError
     puts "execute_script timed out after 30 seconds, likely because this is \
 Safari and the browser was still on about:blank when wait_for_jquery \
--- a/dashboard/test/ui/features/step_definitions/steps.rb
+++ b/dashboard/test/ui/features/step_definitions/steps.rb
@@ -241,8 +241,7 @@ def jquery_is_defined
 end
 
 def jquery_is_element_visible(selector)
-  "return #{jquery_is_defined} &&" +
-  " $(#{selector.dump}).is(':visible') && $(#{selector.dump}).css('visibility') !== 'hidden';"
+  "return #{jquery_is_defined} && $(#{selector.dump}).is(':visible') && $(#{selector.dump}).css('visibility') !== 'hidden';"
 end
 
 def jquery_is_element_displayed(selector)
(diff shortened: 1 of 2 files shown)


dashboard/test/ui/features/step_definitions/steps.rb
```diff
@@ -236,8 +236,12 @@ def navigate_to(url)
   wait_until {@browser.execute_script("return $(\"#{selector}\").is(':checked');") == negation.nil?}
 end
 
+def jquery_is_defined
+  "(typeof jQuery !== 'undefined') && ($ !== 'undefined')"
+end
+
 def jquery_is_element_visible(selector)
-  "return $(#{selector.dump}).is(':visible') && $(#{selector.dump}).css('visibility') !== 'hidden';"
+  "return #{jquery_is_defined} && $(#{selector.dump}).is(':visible') && $(#{selector.dump}).css('visibility') !== 'hidden';"
```
stephenliang: I don't think we need to test that jquery is defined when verifying that an element is visible. It will cause this to return false.

Instead, the act of loading a page should wait for jquery to become available. So I think we can remove this check.

vijayamanohararaj: The exception is thrown after step #2 (detailed in PR description), where a jQuery seems to have been executed successfully, but some page reload seems to have happened since leading to the exception. I think adding the additional check for $ increases our resilience in general, but don't think it would address this specific behavior we are seeing.

vijayamanohararaj: These tests are also executed as part of drone and seem to work working successfully. So, the overall condition is working as expected.

bethanyaconnor:
> It looks like between steps 2 & 3, something happens that results in jQuery not being available. There wasn't anything obvious from looking at the video that could explain what this could be.

I always figured this happened when the browser going to a new page (this is consistent with one of the videos you posted -- didn't check all of them). But, I also always wondered why this check didn't work as-is, so I'm super excited to have a more robust check for jquery to be present!

bethanyaconnor (approved):

molly-moen (approved):

davidsbailey:
Hi folks, I like the creativity here, but I suspect that the issue is something different. Here are my observations:
* our version of jQuery appears to set both globals at the same time: https://github.com/jquery/jquery/blob/bf11739f6c6926bc9bc1b5a1460505d3b7ef8b01/src/exports/global.js
* when we fail to find jQuery, you can see we are still looking at the previous page:
Screenshot 2024-02-25 at 9 45 34 PM

my recommendation would be to revert the changes in this PR, and update this call as follows:
```diff
-And I press the first "#uitest-save-section-changes" element
+And I press the first "#uitest-save-section-changes" element to load a new page
```

vijayamanohararaj:
> Hi folks, I like the creativity here, but I suspect that the issue is something different. Here are my observations:
> 
> * our version of jQuery appears to set both globals at the same time: https://github.com/jquery/jquery/blob/bf11739f6c6926bc9bc1b5a1460505d3b7ef8b01/src/exports/global.js
> * when we fail to find jQuery, you can see we are still looking at the previous page:
>   Screenshot 2024-02-25 at 9 45 34 PM
> 
> my recommendation would be to revert the changes in this PR, and update this call as follows:
> 
> ```diff
> -And I press the first "#uitest-save-section-changes" element
> +And I press the first "#uitest-save-section-changes" element to load a new page
> ```

Thank you @davidsbailey for taking a look and providing detailed insights. This makes a lot of sense! My bad on overseeing that the page had not transitioned. 

Updated the PR to wait for page load.

davidsbailey (approved):

PR #56712, opened 2024-02-21
[Google Blockly] Convert svgFrame and FunctionEditor to TypeScript

molly-moen:
Convert the following Blockly classes to TypeScript:
- `svgFrame` and its subclasses `blockSvgFrame` and `workspaceSvg`
- `functionEditor`
- `cdoScrollbar` (it is used by `functionEditor`)

As part of this I also updated the global `Blockly` type definition to be the type definition created by #56651. In order to do this I had to create a separate `.d.ts` file. This is because the type definition relies on importing the new type, and once you add an import to the `.d.ts` file it becomes a module file and the syntax changes. In order to not break the existing file I created a new file `GlobalsModule.d.ts`. There isn't much documentation on this pattern, I gleaned it from various stack overflow posts (in general TypeScript does not approve of global variables). 

As part of this I also found and fixed a bug where when you delete a function from the function editor, the call block gets deleted (expected), and any blocks connected to the call block also get deleted (not expected). The fix for this was to use the `healStack` parameter to the `dispose` function.

Most of the other changes are adding types and null checks.

## Links

## Testing story
Tested locally. I tested that the function editor still works as expected (and deleting functions works better), and the svg frames work as expected.

## Follow-up work
More migration!

## PR Checklist:

- [ ] Code is well-commented


apps/src/blockly/addons/functionEditor.ts
```diff
@@ -23,28 +23,30 @@ import {frameSizes} from './cdoConstants';
...
-    opt_paramTypes
-  ) {
-    // TODO: Are these options from the fork still relevant?
-    this.msgOverrides_ = opt_msgOverrides || {};
-    if (opt_definitionBlockType) {
-      this.definitionBlockType = opt_definitionBlockType;
-    }
-    this.parameterBlockTypes = opt_parameterBlockTypes || {};
-    this.disableParamEditing = opt_disableParamEditing || false;
-    this.paramTypes = opt_paramTypes || [];
```
mikeharv, line 43: Nice!

apps/src/blockly/addons/cdoScrollbar.ts
```diff
@@ -1,19 +1,25 @@
...
   // https://blockly-demo.appspot.com/static/tests/playground.html
   // Only non-paired scrollbars can have their visibility toggled.
-  workspace.scrollbar.hScroll.pair = false;
-  workspace.scrollbar.vScroll.pair = false;
+  // This is a hack, Blockly has pair set to be readonly. We force it to change here
+  // so we can unpair the two scrollbars.
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  (workspace.scrollbar.hScroll as any).pair = false;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  (workspace.scrollbar.vScroll as any).pair = false;
```
mikeharv, line 22: I'm not sure if it's helpful or not, but these should both be instances of the `Blockly.Scrollbar` class.

apps/src/blockly/addons/functionEditor.ts
```diff
@@ -146,20 +150,21 @@ export default class FunctionEditor {
...
 
-  // TODO
-  renameParameter(oldName, newName) {}
+  // Leaving these two functions as placeholders for when we implement parameters.
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars, @typescript-eslint/no-empty-function
+  renameParameter(_oldName: string, _newName: string) {}
 
-  // TODO
+  // eslint-disable-next-line @typescript-eslint/no-empty-function
   refreshParamsEverywhere() {}
```
mikeharv, line 166: Just noting that `refreshParamsEverywhere` and `renameParameter` are supposedly used in Dance by `sprite_parameter_get` blocks, but I don't think those even exist in that lab. It would be great for us to clean up some of the dance block logic that never actually got used, because then we could remove some of these methods (from here and the wrapper).

apps/src/blockly/addons/functionEditor.ts
```diff
@@ -355,61 +379,68 @@ export default class FunctionEditor {
...
+      block.dispose(true /* healStack */);
     });
 
     // delete all caller blocks from the main workspace
     Blockly.Procedures.getCallers(
       this.block.getProcedureModel().getName(),
       Blockly.mainBlockSpace
     ).forEach(block => {
-      block.dispose();
+      block.dispose(true /* healStack */);
```
mikeharv, line 424: :tada:

apps/src/blockly/addons/functionEditor.ts
```diff
@@ -355,61 +379,68 @@ export default class FunctionEditor {
...
     this.editorWorkspace
-      .getProcedureMap()
+      ?.getProcedureMap()
       .delete(this.block.getProcedureModel().getId());
 
-    // delete the block from the editor workspace and hide the modal
-    this.block.dispose();
+    // Delete the block from the editor workspace and hide the modal.
+    // We don't need to heal the stack here because we always clear the editor workspace.
+    this.block.dispose(false /* healStack */);
```
mikeharv, line 436: If we don't need to heal the stack, do we need to change this at all?

mikeharv (commented):
At this time I've reviewed as far down in the files list as functionEditor so I'll need to come back to this. A couple questions so far, but great work!

apps/src/blockly/addons/functionEditor.ts
molly-moen: yeah I left these for now because they are theoretically used, but I'm open to removing these as a follow up!

apps/src/blockly/addons/functionEditor.ts
molly-moen: Yeah, technically the parameter is required.

mikeharv: Oh, that makes sense!

apps/src/blockly/addons/cdoScrollbar.ts
molly-moen: Unfortunately `Blockly.scrollbar` does not allow resetting of `pair` so I had to cast it to hack around that.

apps/src/blockly/constants.ts
```diff
@@ -96,6 +96,14 @@ export const PROCEDURE_DEFINITION_TYPES = [
 // A list of block types associated with the Run button.
 export const SETUP_TYPES = [BLOCK_TYPES.whenRun, BLOCK_TYPES.danceWhenSetup];
 
+export const WORKSPACE_EVENTS = {
+  MAIN_BLOCK_SPACE_CREATED: 'mainBlockSpaceCreated',
+  EVENT_BLOCKS_IMPORTED: 'blocksImported',
+  BLOCK_SPACE_CHANGE: 'blockSpaceChange',
+  BLOCK_SPACE_SCROLLED: 'blockSpaceScrolled',
+  RUN_BUTTON_CLICKED: 'runButtonClicked',
```
mikeharv, line 135: It could be a good follow-up to audit these for use. I'd expect we're not using them all.

mikeharv (approved):
Looks good!

apps/src/blockly/addons/cdoScrollbar.ts
mikeharv: Ah, that tracks. Thanks for explaining.

PR #56717, opened 2024-02-21
[Google Blockly] Functions: fix destroy override; add doProcedureUpdate override

mikeharv:
Last month, @sanchitmalhotra126 [noticed](https://codedotorg.slack.com/archives/C05DK21DAHK/p1706554295359669) that if you delete a function definition from the main workspace an error was caused because we weren't disposing the associated call blocks. This was observed in Music Lab as well as Dance but can really happen in any level that doesn't use the modal function editor.

In Core Blockly, deleting a definition deletes any call blocks:

The reason we weren't doing the same was due to an override of the definition blocks' `destroy` function. We knew we needed to no-op this function for the modal function editor so that closing it wouldn't cause us to delete the shared procedures. What we missed is that we _should_ delete the procedures if we're on the main workspace, ie. when we're not using the modal editor.

One curious discovery while fixing this was that insertion markers (the gray block outlines that show where a real block is about to snap into place) also get this function. We don't want an automatic insertion marker deletion to trigger a procedure to get deleted either.

Once that logic was added, things were working mostly as expected, except that I discovered that the stack was not being "healed", ie. all child blocks were also getting disposed. (It's only a coincidence that @molly-moen had to fix the same issue for our modal editor's delete button: https://github.com/code-dot-org/code-dot-org/pull/56712/commits/25ebd30df5c4d35d2723ca759505f0d5dd9c14e3) What is strange about this is that the plugin assumes a different default behavior here compared with core. Notice that that a `healStack` boolean is not passed: 
https://github.com/BeksOmega/blockly-samples/blob/7954a8fff50e41fa7c0f891e957bf9ed616361d6/plugins/block-shareable-procedures/src/blocks.ts#L1068
You can see that regardless of which workspace, deleting the definition causes the children of the associated call block to be disposed.

We don't want this behavior, which unfortunately means we have to override `doProcedureUpdate` as well.

These changes are only needed for functions and not behaviors as behavior definitions should never show up on the primary workspace.
While testing this, I noticed a new crash when opening the behaviors category. The cause was that the modified destroy function was unnecessarily being added to behavior get blocks as well. (The error was because they didn't have an original `this.destroy` to store.) We don't ever need to no-op a function that shouldn't be there in the first place so I removed the mixin from those blocks.

## Links

Slack thread: https://codedotorg.slack.com/archives/C05DK21DAHK/p1706554295359669

Jira: https://codedotorg.atlassian.net/browse/CT-303

## Testing story

I tested that this is working as expected in a no-modal Sprite Lab level and in Music Lab. No regressions were observed using the modal editor in the Sprite Lab project level.

|
-|-

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/blockly/customBlocks/googleBlockly/proceduresBlocks.js
+++ b/apps/src/blockly/customBlocks/googleBlockly/proceduresBlocks.js
@@ -87,6 +87,7 @@ export const blocks = GoogleBlockly.common.createBlockDefinitionsFromJsonArray([
       'procedure_caller_context_menu_mixin',
       'procedure_caller_onchange_mixin',
       'procedure_callernoreturn_get_def_block_mixin',
+      'procedure_call_heal_stack',
     ],
     mutator: 'procedure_caller_mutator',
   },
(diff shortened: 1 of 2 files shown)


apps/src/blockly/customBlocks/googleBlockly/proceduresBlocks.js
```diff
@@ -187,15 +188,41 @@ GoogleBlockly.Extensions.register('procedures_block_frame', function () {
...
+      }
+    },
+  };
+  // We can't register this as a mixin since we're overwriting existing methods
+  Object.assign(this, mixin);
+});
+
+// Override the doProcedureUpdate function to heal the stack. Without this,
+// any child blocks connected to a call block would also get deleted.
+GoogleBlockly.Extensions.register('procedure_call_heal_stack', function () {
```
molly-moen: nit: `procedure_call_heal_stack` doesn't imply to me that this overrides `doProcedureUpdate`. Could we name it something like `procedure_caller_do_update`?

molly-moen (approved):
Nice!

apps/src/blockly/customBlocks/googleBlockly/proceduresBlocks.js
mikeharv: Good call. Updated!

PR #56719, opened 2024-02-21
Aitt basic updates to UI rubric component

wilkie:
Updates UI to match new design. This represents the following set of changes:

* Margins around header tightened
* Header text is truncated to fit instead of expanding the header height
* Fixes progress ring to fix fence-post error regarding radius length
* Uses standard colors
* Removes red/green colored boxes in favor of white inset box.
* Moves thumbs up/down feedback radio to row with score text.
* Adjusts 'USES AI' text and moves it alongside the title
* Adds 'temperature' widget for expressing 'accuracy'
* Adds new AI Bot image and adjusts its size
* **Updated**: Also updates the thumbs down feedback text from "has achieved" to Incorrectly score too high/low.
* **Updated**: Also removes redundant calls to gather prior feedback as one progresses through the carousel
* **Updated**: Also tackles the race condition between updating due to prior feedback and anybody writing new feedback
* **Updated**: Also eliminates the race condition where it updates the current view with queued network requests if they pile up (quickly going through the carousel faster than it can load a learning goal)

In order to move the thumbs up/down buttons in line with the score and temperature widget, a provider is introduced to maintain the state of the current AI feedback. This was also good to fix the current bug where that in-progress feedback was maintained if you switched from one learning goal to another without submitting.

## I18n

Adds strings: `aiAssessmentScore`, `aiAssessmentEvidence`, `aiAssessmentEvidenceLine`, `aiAssessmentEvidenceLines`, `aiFeedbackSubmit`, `feedbackHeader`, `feedbackPlaceholderShort`

And changes / fixes strings: `aiAssessmentFeedbackAsk`, `aiFeedbackFalsePos`, `aiFeedbackFalseNeg`, `aiConfidence`

## Links

- figma: [Design](https://www.figma.com/file/WBcs7BY9IC6Nq5rG4V1SWF/AI-in-Teacher-Tools?type=design&node-id=3525-5421&mode=design&t=eY7OKbyCaPxnJvuc-0)
- jira ticket: [AITT-368](https://codedotorg.atlassian.net/browse/AITT-368)

## Images

Before:

image

After Mark's recent PR:

image

After:

image

Scrolling down a bit:

image

Unlike above, this one has single lines for evidence instead of a range:

image

Pressing Thumbs up on feedback:

image

Pressing Thumbs down on feedback:

image

Pressing "other" on feedback:

image

AI Suggested assessments:

image

Hovering over evidence level:

image

Too subjective:

image

High confidence temperature bar:

image

Tooltip for explaining accuracy: (Unchanged, but moved to temperature bar. Not in figma design.)

image

## Testing story

Updated the unit tests to reflect the new state and to add in the mock provider.

## Deployment strategy

Pilot members should see the changes immediately if they can access the rubrics since Mark removed the carousel experiment restriction.

## PR Checklist:

- [ ] Code is well-commented


apps/src/templates/rubrics/ProgressRing.jsx
```diff
@@ -4,15 +4,17 @@ import i18n from '@cdo/locale';
 import style from './rubrics.module.scss';
 import {learningGoalShape} from './rubricShapes';
 import color from '@cdo/apps/util/color';
+import classnames from 'classnames';
```
wilkie, line 7: I needed the class added to style the margins.

apps/src/templates/rubrics/ProgressRing.jsx
```diff
@@ -4,15 +4,17 @@ import i18n from '@cdo/locale';
...
 export default function ProgressRing({
+  className,
   learningGoals,
   currentLearningGoal,
   understandingLevels,
   radius,
   stroke,
 }) {
-  const normalizedRadius = radius - stroke * 2;
+  const normalizedRadius = radius - stroke;
```
wilkie, line 17: You only need to account for the width of the stroke once since this is the radius and not the diameter and the stroke radiates from the outline of the circle. So there's only one half of it extended beyond the radius.

shared/css/color.scss
```diff
@@ -2,7 +2,7 @@
 
 // Please keep this file simple; it should only contain color definitions
 //   and aliases.  Variables defined in this file are exported to
-//   apps/src/color.js during the apps build process.
+//   apps/src/util/color.js during the apps build process.
```
wilkie, line 5: Happened to notice that this was inaccurate.

shared/css/color.scss
```diff
@@ -203,6 +203,11 @@ $light_gray_800: #54595E;
 $light_gray_900: #3F444B;
 
 $light_affirmative_100: #E2F1E2;
+$light_affirmative_500: #3EA33E;
+
+$light_negative_500: #E5311A;
+
+$light_caution_500: #F9CB28;
```
wilkie, line 210: Standard named colors from the figma design doc that are used in the temperature widget.

apps/src/templates/rubrics/AiAssessmentFeedbackContext.jsx
```diff
@@ -0,0 +1,5 @@
+import {createContext} from 'react';
+
+const AiAssessmentFeedbackContext = createContext(null);
+
+export default AiAssessmentFeedbackContext;
```
wilkie, line 9: This is the central coordination of the feedback context that both the feedback radio (thumbs up/down buttons) and actual feedback component (submit form) use to set the state of the radio buttons.

This was necessary to add because the thumbs up and down buttons needed to be placed in a separate branch of the UI so that they could be rendered in line with the score and temperature widget.

apps/src/templates/rubrics/AiAssessmentFeedback.jsx
```diff
@@ -57,100 +43,43 @@ export default function AiAssessmentFeedback({aiEvalInfo}) {
...
-            checked={aiFeedback === thumbsdownval}
-          />
-        </label>
-        {aiFeedbackReceived && (
-          <EmText className={style.aiFeedbackReceived}>
-            {i18n.aiFeedbackReceived()}
-          </EmText>
-        )}
-      </div>
-
```
wilkie, line 147: A lot of this got moved to `AiAssessmentFeedbackRadio` (thumbs up/down buttons) and the submitting happens higher up via the context.

apps/src/templates/rubrics/LearningGoal.jsx
```diff
@@ -56,7 +58,7 @@ export default function LearningGoal({
   const teacherFeedback = useRef('');
   const understandingLevel = useRef(invalidUnderstanding);
 
-  const aiEnabled = learningGoal.aiEnabled && teacherHasEnabledAi;
+  const aiEnabled = !!(learningGoal.aiEnabled && teacherHasEnabledAi);
```
wilkie, line 61: I made this a required bool, so it needed to be explicitly coerced to a bool.

apps/src/templates/rubrics/LearningGoals.jsx
```diff
@@ -124,8 +136,15 @@ export function annotateLines(observations) {
...
       }
+      ret.push({
+        firstLine: lineNumber,
+        lastLine: lastLineNumber,
+        message: message,
+      });
     }
   }
+
+  return ret;
```
wilkie, line 147: Need to now return the lines that we annotated to write them out in the assessment box.

apps/src/templates/rubrics/LearningGoals.jsx
```diff
@@ -246,11 +267,15 @@ export default function LearningGoals({
             if (json.understanding >= 0 && json.understanding !== null) {
               understandingLevels.current[index] = json.understanding;
             }
+            if (index === currentLearningGoal) {
+              setDisplayFeedback(teacherFeedbacks.current[currentLearningGoal]);
+              setDisplayUnderstanding(
+                understandingLevels.current[currentLearningGoal]
+              );
+            }
```
wilkie, line 276: This fixes an issue where the feedback would not get set if you opened the rubric container too quickly since it would otherwise rely on it polling the feedback again... which it would not do until you switched learning goals.

This fixes a flaky UI test as well... it would unsuccessfully race this behavior around 50% of the time.

davidsbailey: great minds think alike! https://github.com/code-dot-org/code-dot-org/pull/56888/files#diff-e87016cd4255be29b93b0c4fb42f5d4119064bad6fc271fceaf989693d13dbc5R249-R254

davidsbailey (commented):
Awesome work Wilkie! I love the temperature bar! This round of top-level feedback turns out to be for Tess. I'll follow up with some top-level comments for you:

> Hovering over evidence level:
>
> image

I think you've implemented this correctly, but this is weird to me because you've hovering over Convincing but the giant ^ is pointing at Extensive. UX feedback @tess323 

> Pressing Thumbs down on feedback:
>
>image

@tess323 will the "has achieved" text need to change once we are showing exact match scores?

dashboard/test/ui/features/teacher_tools/rubrics/teacher_view_of_rubric.feature
```diff
@@ -17,16 +17,19 @@ Scenario: Teachers can give and send feedback on the rubric to students.
...
   And I wait until element ".student-table" is visible
   And I click selector ".student-table tr:nth(1)" to load a new page
   And I click selector "#ui-floatingActionButton" once I see it
-  And I wait until element "h6:contains(Code Quality)" is visible
+  And I wait until element "h5:contains(Code Quality)" is visible
   And I wait until element "button:contains(Extensive)" is visible
   Then I click selector "button:contains(Extensive)"
   And I click selector "#ui-teacherFeedback" once I see it
+
+  When I wait for 3 seconds
```
davidsbailey: an explicit wait like this is a sign of future flakiness. is there anything in the UI where you can wait for it to change or become visible instead?

davidsbailey (approved):
Code changes look good, nice use of react context! one UI test request before merging.

tess323:
> 
> > Hovering over evidence level:
> > image
> 
> I think you've implemented this correctly, but this is weird to me because you've hovering over Convincing but the giant ^ is pointing at Extensive. UX feedback @tess323

**This looks right to me.  I agree it may feel slightly awkward so sent a request to Molly for her input.  I think its fine to launch this and we can update as a P2 **

tess323:
> 
> > Pressing Thumbs down on feedback:
> > image
> 
> @tess323 will the "has achieved" text need to change once we are showing exact match scores?


Can we update the "has achieved" options to "Incorrectly scored too high" and "Incorrectly scored too low"

apps/src/templates/rubrics/LearningGoals.jsx
wilkie: Wow!

wilkie:
Rebased from staging to resolve merge conflict.

wilkie:
I'll be looking at improving the UI test as mentioned.

apps/test/unit/templates/rubrics/RubricContainerTest.jsx
```diff
@@ -157,9 +157,19 @@ describe('RubricContainer', () => {
...
       await Promise.resolve();
     });
+    // Perform the json() call from the fetch
+    await act(async () => {
+      await Promise.resolve();
+    });
+    // Perform the data call from the fetch
+    await act(async () => {
+      await Promise.resolve();
+    });
```
davidsbailey, line 171: looking forward to someday not having to do this in RTL tests!

wilkie: Yes! And I test in Firefox locally instead of Chrome... and that usually means I need one more of these for it to pass... and it is a reason why I can't get a full test run to pass locally as I will hit another component with too few of these. I think Chrome might do the response half of the fetch in one go? I'm not sure.

I've considered adding a helper that just loops 3 or 4 times and does this and making that more of a standard than the duplication, here, but I'm 100% on board if RTL just solves this outright somehow.

wilkie:
> > > Hovering over evidence level:
> > > image
> > 
> > 
> > I think you've implemented this correctly, but this is weird to me because you've hovering over Convincing but the giant ^ is pointing at Extensive. UX feedback @tess323
> 
> **This looks right to me. I agree it may feel slightly awkward so sent a request to Molly for her input. I think its fine to launch this and we can update as a P2 **

Yep. The design mock-ups, to be fair, have this off-center and exactly between a couple of the buttons. This is pretty challenging to do in practice and felt like more work than warranted. It seemed more intentional to have it off-center (that is, it felt like it _meant_ to be pointing at something) and it was going to be hard to account for localized text if I had used, say, a strict by-the-pixel `left` position for it to reproduce the mock up.

wilkie:
Added the new strings for accessing thumbs-down feedback: 
image

davidsbailey:
> Yep. The design mock-ups, to be fair, have this off-center and exactly between a couple of the buttons. This is pretty challenging to do in practice and felt like more work than warranted. It seemed more intentional to have it off-center (that is, it felt like it _meant_ to be pointing at something) and it was going to be hard to account for localized text if I had used, say, a strict by-the-pixel `left` position for it to reproduce the mock up.

Thank you for reminding me about that Wilkie, I didn't check the spec when reviewing this PR, but the idea of having the pointer go directly between 2 values sounds familiar now. I agree with proceeding here and following up as P2! In hindsight this is probably more confusing to me (trying to decipher a cursorless screenshot where the user is hovering) than it will be for the user who knows where they are hovering.

dashboard/test/ui/features/teacher_tools/rubrics/teacher_view_of_rubric.feature
wilkie: I think I can just delete that line... it is racing the update to the textarea from opening the modal with the old data... so the problem might have been that you type in the thing but it gets replaced with empty string.

To be entirely certain, it's probably good to disable the textarea until it negotiates prior feedback at least once.

apps/src/templates/rubrics/LearningGoals.jsx
wilkie: I think we're both wrong because `currentLearningGoal` is state and not a ref... when it performs the fetch, `currentLearningGoal` is baked and bound as a particular value (whatever it happens to be when you call fetch). If you quickly flip through the carousel, it will queue a bunch of these fetches and they all race. Each one thinks it is the current one and ALL of them will update the display potentially showing the wrong value.

`currentLearningGoal` should be a ref such that this callback pulls its current value and the callback is dependent on the `current` of the ref. I might need the state still to trigger the re-render though... I think it will be fine because _something_ will trigger it.

I think.

dashboard/test/ui/features/teacher_tools/rubrics/teacher_view_of_rubric.feature
davidsbailey: disabling the textarea while the data loads makes sense to me! https://codedotorg.atlassian.net/browse/AITT-498

apps/src/templates/rubrics/LearningGoals.jsx
wilkie: I went ahead and added a very shallow ref that also tracks the current learning goal so that we don't get that update race when you click through the carousel quickly.

wilkie:
Sorry to add a little more complexity here than it really originally warranted. Very strange behavior was happening when I was testing locally that I wanted to quickly address.

The last two commits fix some network and render race conditions that made tests flaky and the presentation a little wonky. If you were to quickly navigate through the learning goals in the carousel view, it would queue ALL of the learning goals EACH time and those would spam re-renders of the modal that quickly flash through the different settings. I've fixed that with a shallow `ref` that keeps track of the current learning goal index for use basically just within the `fetch` for those. Also, queuing a dozen network requests for every press of the arrows was very expensive especially since if you were navigating to the last learning goal it could wait for the loading of all the other ones before it updated your view. I've added something to track that it loaded prior feedback and only download any it hasn't seen... so it really only loads once on revealing the floating modal.

The risk is that they've updated the information on a different tab... it wouldn't be reflected until a page load. but a page load happens every time you switch students, so that risk feels low. You could treat it as a cache and have it query again after 30 or 40 seconds maybe.

I've also disabled the textarea for teacher feedback until it loads the prior feedback to cut off a race condition almost entirely subject to the UI test which can navigate fast enough for it to matter, but you never know. Without this, you could technically type into the textarea only to have it erased when the feedback loaded.

apps/src/templates/rubrics/LearningGoals.jsx
davidsbailey: CC @cearachew and @Nokondi who looked at this more closely than I did (race condition)

dashboard/test/ui/features/teacher_tools/rubrics/teacher_view_of_rubric.feature
wilkie: Wasn't very hard. Could pull out into a separate PR if you'd like.

wilkie:
Perfectly willing to pull these last two commits into their own PR.

dashboard/test/ui/features/teacher_tools/rubrics/teacher_view_of_rubric.feature
davidsbailey: no that's fine, I was just hedging my bets in case you didn't have time to get to it this week. I can close out that issue when this PR merges

apps/src/templates/rubrics/LearningGoals.jsx
```diff
@@ -176,6 +177,9 @@ export default function LearningGoals({
   const [currentLearningGoal, setCurrentLearningGoal] = useState(0);
   const learningGoalEvalIds = useRef(Array(learningGoals.length).fill(null));
   const teacherFeedbacks = useRef(Array(learningGoals.length).fill(''));
+  const teacherFeedbacksLoaded = useRef(
+    Array(learningGoals.length).fill(false)
+  );
```
davidsbailey, line 185: @wilkie  I think you called this a shallow ref, is that right and if so what does that mean? it's not jumping to the top of my search results

dashboard/test/ui/features/teacher_tools/rubrics/teacher_view_of_rubric.feature
```diff
@@ -27,9 +27,8 @@ Scenario: Teachers can give and send feedback on the rubric to students.
   And I wait until element "h5:contains(Code Quality)" is visible
   And I wait until element "button:contains(Extensive)" is visible
   Then I click selector "button:contains(Extensive)"
+  And I wait until element "#ui-teacherFeedback" is enabled
```
davidsbailey, line 30: 😍

apps/src/templates/rubrics/LearningGoals.jsx
```diff
@@ -245,37 +250,52 @@ export default function LearningGoals({
...
+              }
+              if (index === currentLearningGoalRef.current) {
+                setDisplayFeedback(
+                  teacherFeedbacks.current[currentLearningGoalRef.current]
+                );
+                setLoaded(
+                  teacherFeedbacksLoaded.current[currentLearningGoalRef.current]
+                );
+                setDisplayUnderstanding(
+                  understandingLevels.current[currentLearningGoalRef.current]
```
davidsbailey: I think this block is worth highlighting with a brief comment explaining its role in preventing race conditions

davidsbailey (approved):
updates look great, nice sleuthing and nice fixes!

apps/src/templates/rubrics/LearningGoals.jsx
wilkie: For the context now lost to my update, I made the exact same fix as Ceara: https://github.com/code-dot-org/code-dot-org/blob/bdbc3c7793585240305a0ef187724f21f5a05445/apps/src/templates/rubrics/LearningGoals.jsx#L249-L254

But when the callback happens `currentLearningGoal` is bound to the fetch... if it changes after the fetch happens, it will not be reflected in the fetch `then` so multiple `fetch` calls might all update and re-render because _their own_ idea (outdated) of the `currentLearningGoal` will line up with some `index`.

apps/src/templates/rubrics/LearningGoals.jsx
wilkie: Oh, this is just representing that it had loaded the saved feedback for a particular learning goal. That's a copy from the similar ref storing the actual saved feedback.

The `useRef` creating a ref here is useful because you can set a ref without causing a re-render of a component. That's good because we load these all in the background and only selectively show the current one. So there's logic in the fetch that will put this to `true` for the learning goal and then logic there to, if it is the current index, set the `loaded` state to true.

I said it is 'shallow' because the variable doesn't not represent the value itself... you have to query it via `.current`. So when you pass around the variable (or trap it inside a Promise), that value is still negotiated when it is needed.

When you, say, hit the `index === currentLearningGoal` inside the fetch, which I said is a problem, the value that `currentLearningGoal` represents is trapped inside the fetch... frozen at that value when the fetch originally happened. When the fetch finishes, `setCurrentLearningGoal` might have been called, but it won't know that and it will be using the old value. So it will update the screen with the wrong info.

Since it freezes variables inside the Promise, we need a shallow representation that freezes only the function to retrieve the value instead of the value itself. That's what `useRef` is for... but since references cannot re-render components... and we want to re-render when the current learning goal changes, we annoyingly seem to need both a ref and a state to represent the same thing. But with the ref in place, the `then` clauses of the `fetch` will query the value at the current moment in time, and appropriate update the display.

apps/src/templates/rubrics/LearningGoals.jsx
wilkie: Agreed. Though. the 'brief' you threw in there is the real challenge.

apps/src/templates/rubrics/LearningGoals.jsx
wilkie: To be clear `currentLearningGoalRef` was what I referred to colloquially as a 'shallow ref'.

wilkie: Even though, hmm, all refs are technically shallow haha.

apps/src/templates/rubrics/LearningGoals.jsx
davidsbailey: "make sure the currently visible learning goal is updated when the data fetch completes"

wilkie:
That took me awhile to figure out that `props` was not seeing the change of `disabled`... it always reports the original value. Which makes sense in hindsight. I needed to check the current value of that attribute on the actual node to fix the failing test. Phew.

PR #56722, opened 2024-02-21
Convert cdoBlockFlyout, cdoFieldFlyout and cdoBlockSerializer to TypeScript

molly-moen:
More TypeScript Migration
- `cdoBlockFlyout`: it had an useless function (`setSourceBlock`, `sourceBlock_` used to be needed in `reflowInternal` but is no longer used) and an unused field `flyoutBlockPadding` that I could remove. Otherwise the changes were types and null checks.
- `cdoFieldFlyout`: got rid of a pointless call to `setSourceBlock` on `cdoBlockFlyout`. Types and null checks
- `cdoBlockSerializer`: just types

## Links

## Testing story
Tested that mini flyouts still work.

## PR Checklist:

- [ ] Code is well-commented


apps/src/blockly/addons/cdoBlockFlyout.ts
```diff
@@ -1,19 +1,27 @@
...
    * This flyout is intended to be placed inside of a block's FieldFlyout.
    *
-   * @param {Object} workspaceOptions - The options for constructing the class.
+   * @param workspaceOptions - The options for constructing the class.
    */
-  constructor(workspaceOptions) {
+  constructor(workspaceOptions: CdoBlockFlyoutOptions) {
     super(workspaceOptions);
-    this.horizontalLayout_ = true;
+    this.horizontalLayout = true;
```
mikeharv: I was wondering how this was working now if it had been renamed so I inspected the field and saw this:
image
 It looks like it's already being set by the parent class here, fortunately: https://github.com/google/blockly/blob/a517317b77c1e58727b748a99f9ae70b2c27d26a/core/flyout_horizontal.ts#L30

I suspect we could just delete this line instead since the old version wasn't accomplishing anything and your changed version is redundant.

apps/src/blockly/addons/cdoBlockFlyout.ts
```diff
@@ -232,7 +239,7 @@ export default class CdoBlockFlyout extends GoogleBlockly.HorizontalFlyout {
         cursorX += blockHW.width + gaps[i];
 
         this.addBlockListeners_(root, block, rect);
-      } else if (item.type === 'button') {
+      } else if (item.type === 'button' && item.button) {
```
mikeharv: I think that we can theoretically expect both of these conditions to be the same in terms of truthiness, but that it isn't made explicit by the `FlyoutItem` interface. Does that track?

https://github.com/google/blockly/blob/a517317b77c1e58727b748a99f9ae70b2c27d26a/core/flyout_base.ts#L1311-L1315

mikeharv (approved):
Looks good! 

I only wonder if it's worth removing `this.horizontalLayout = true;` while you're here. (See comment.)

apps/src/blockly/addons/cdoBlockFlyout.ts
molly-moen: oh nice, will remove that then

apps/src/blockly/addons/cdoBlockFlyout.ts
molly-moen: Correct, sometimes TypeScript makes you put in explicit null checks even when you know the thing won't be null because of some other condition. Seems like I could replace this check with just `else if(item.button)`?

mikeharv: Oh maybe! Either way, should we use the same convention here and at line 213 above?

```
      if (item.type === 'block') {
        const block = item.block;
        if (!block) {
          continue;
        }
...
```

mikeharv: Oh wait, there's also this:

```
   */
  // TODO: once we upgrade to v10 of blockly we should not need this anymore
  // and can rely on the parent version of this function.
  // This is copied from the core blockly repo to include a fix from this PR on Blockly:
  // https://github.com/google/blockly/pull/7333
```

molly-moen: Unfortunately that PR got reverted due to rtl issues (I updated the comment to this effect). As far as I can tell the fix from the pr is still fine for this use case, even in rtl, probably because this is a tiny flyout.
And yes, good catch on line 213, will update both.

mikeharv: Should we add a comment with the link to the open issue so we can track more easily?

mikeharv: https://github.com/google/blockly/issues/6280

molly-moen: updated comment and fixed logic!

mikeharv: LGTM!

PR #56724, opened 2024-02-21
Adding optgroups to SimpleDropdown

lfryemason:
Adds `<optgroups>` ([docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/optgroup)) to SimpleDropdown design component. 

This was requested for migrating the [UnitSelector](https://github.com/code-dot-org/code-dot-org/blob/staging/apps/src/templates/sectionProgress/UnitSelector.js) to use SimpleDropdown

## PR Checklist:

- [ ] Code is well-commented


apps/src/componentLibrary/simpleDropdown/SimpleDropdown.tsx
```diff
@@ -81,11 +84,21 @@ const SimpleDropdown: React.FunctionComponent<SimpleDropdownProps> = ({
           className={moduleStyles.dropdown}
           disabled={disabled}
         >
-          {items.map(({value, text}) => (
-            <option value={value} key={value}>
-              {text}
-            </option>
-          ))}
+          {itemGroups && itemGroups.length > 0
```
levadadenys: Awesome improvement 100%, have a couple suggestions though

apps/src/componentLibrary/simpleDropdown/SimpleDropdown.tsx
```diff
@@ -7,6 +7,8 @@ import moduleStyles from './simpleDropdown.module.scss';
 export interface SimpleDropdownProps {
   /** SimpleDropdown items list */
   items: {value: string; text: string}[];
+  /** SimpleDropdown grouped list of items */
+  itemGroups: {label: string; groupItems: {value: string; text: string}[]}[];
```
levadadenys: with current PropsType, both `items` and `itemGroups` are required fields, meaning we'll need to pass both of them always with one of them being empty array and second - actual array data.

levadadenys: I think there's two ways to improve:
1) Make both props optional and set a default value as a `[]` in component itself. So that end user will only need to pass one prop with actual data.
2) Use generics or support both types for one prop (`items`) so that we won't need the second one

apps/test/unit/componentLibrary/SimpleDropdownTest.jsx
```diff
@@ -142,4 +142,40 @@ describe('Design System - Dropdown Select Component', () => {
     expect(spyOnChange).to.have.not.been.called;
     expect(dropdownValue).to.equal('');
   });
+
+  it('SimpleDropdown - renders with correct text and options with grouped items', () => {
+    render(
+      <SimpleDropdown
+        name="test4-dropdown"
```
levadadenys, line 149: Since this file is `jsx`, not `tsx`, we don't have a type error here, however we need to pass items={[]} here with our current PropType

apps/src/componentLibrary/simpleDropdown/SimpleDropdown.story.tsx
```diff
@@ -100,13 +101,34 @@ DisabledDropdown.args = {
...
   labelText: 'Disabled Dropdown',
   onChange: args => console.log(args),
   disabled: true,
   size: 'm',
 };
 
+export const GroupedOptionsDropdown = SingleTemplate.bind({});
+GroupedOptionsDropdown.args = {
+  name: 'grouped-options-dropdown',
+  itemGroups: [
```
levadadenys, line 113: Interesting, aren't we having an error here about `items` being a required prop here?

levadadenys (commented):
100% great update, have polishing/improvement related comments only

apps/src/componentLibrary/simpleDropdown/SimpleDropdown.story.tsx
lfryemason: Funnily enough, we are not. It complains about not having `itemGroups` when not specified, but doesn't for `items`.

apps/src/componentLibrary/simpleDropdown/SimpleDropdown.tsx
lfryemason: I think it makes the code a bit more readable to just make them optionals, but if you would prefer to use generics to support both types in one prop I can change it.

Optionals have been added.

apps/src/componentLibrary/simpleDropdown/SimpleDropdown.story.tsx
levadadenys: oh, interesting

apps/src/componentLibrary/simpleDropdown/SimpleDropdown.tsx
levadadenys: nit: since itemGroups will always be defined now, we can simplify it to `{itemGroups.length > 0 ?`

levadadenys (approved):
LGTM!
Have one small nit, other than that everything is cool, thanks for these updates!

PR #56726, opened 2024-02-22
starting to translate school inputs to functional component

hannahbergam:
This PR does a few things:

1. Adds translation strings for the new school association flow
2. Refactors the header in the finish_sign_up.js file so that the two components can live side by side for a/b testing
3. Starts working on a new version of school association as a functional component, with its own stylesheet (NOTE: THIS IS NOT IN USE YET) called SchoolDataInputs.jsx
     This component is working toward a cleaner UI, without including any validation yet. I'm looking to get this piece reviewed new instead of waiting so that each PR has fewer lines to look through. 

To make myself feel better, here is the before and after of the school portion of sign in:

Before/On Prod:
Screenshot 2024-02-26 at 10 27 25 AM

After:
Screenshot 2024-02-26 at 10 27 32 AM

For posterity: this is the current state of the ui if you call it from finish_sign_up.js:
Screenshot 2024-02-27 at 4 10 10 PM

## Links

- spec: []()
- jira ticket: []()
https://codedotorg.atlassian.net/browse/ACQ-1532
https://codedotorg.atlassian.net/browse/ACQ-1531
https://codedotorg.atlassian.net/browse/ACQ-1530

## PR Checklist:

- [ ] Code is well-commented


apps/src/templates/SchoolDataInputs.jsx
```diff
@@ -0,0 +1,141 @@
+import PropTypes from 'prop-types';
```
hannahbergam, line 1: Worth noting: this file is a WIP. I initially converted the functionality of the old component while maintaining the validation. I have not yet fixed all the styling issues, and there are still some pieces I need to build out. Is it better to check this in or not merge it at all until it's done? Again, not actually in use yet.

apps/src/templates/school-association.module.scss
```diff
@@ -0,0 +1,13 @@
+@import "color.scss";
+@import "font.scss";
+
+.outerContainer {
+    text-align: center !important;
```
bethanyaconnor: why is important needed here? can you leave a comment as to why?

apps/src/templates/SchoolDataInputs.jsx
bethanyaconnor: maybe asked a different way -- what's the benefit of merging this now? Early PR feedback, simpler PRs in the future, something else? (This would also help reviewers know what to focus on)

apps/src/templates/school-association.module.scss
hannahbergam: Yes I can leave a comment in there! Was just talking to Kelby about this and it might be worth a broader conversation. I copied from the pegasus design system spreadsheet [here](https://github.com/code-dot-org/code-dot-org/blob/staging/pegasus/sites.v3/code.org/public/css/design-system-pegasus.scss#L65) and we don't have a studio equivalent. I'm not sure if that's in our future!

bethanyaconnor: I don't think so -- we don't use unified classes much in dashboard, instead opting for unified components (when they're implemented/available).

If you remove the `!important` does it break? If it does, a comment is great! Thanks!

apps/src/templates/SchoolDataInputs.jsx
hannahbergam: Very helpful question- yes I was hoping for simpler PR's instead of just opening one up that's a ton of changes and validation at once. I'll update the description and resolve this

apps/src/templates/school-association.module.scss
hannahbergam: Gotcha. Is this partly because of the magnitude of the work? Again, just thinking about an ideal approach from a design perspective (and learning from Kelby about how beneficial it could be to have a kind of cascading styling approach where a parent stylesheet could set up a component library for success). Anyway though I know that is beside the point here, and will either remove !important or add a comment.

bethanyaconnor: I might not be understanding -- what piece would be useful to have in dashboard? (this is not completely outside of the review of this PR -- just trying to understand!)

apps/src/templates/SchoolDataInputs.jsx
```diff
@@ -0,0 +1,91 @@
...
+import React from 'react';
+import i18n from '@cdo/locale';
+import {Heading1, Heading2, Heading3} from '@cdo/apps/lib/ui/Headings';
```
bethanyaconnor: I think you want to use the Headings from the typography component: https://github.com/code-dot-org/code-dot-org/tree/staging/apps/src/componentLibrary/typography. Now I'm doubting which is right but I hate that there are multiple 😬

apps/src/templates/SchoolDataInputs.jsx
```diff
@@ -0,0 +1,91 @@
...
+import style from './school-association.module.scss';
+import SimpleDropdown from '../componentLibrary/simpleDropdown/SimpleDropdown';
+import {COUNTRIES} from '@cdo/apps/geographyConstants';
+import SearchBar from './SearchBar';
+
+const TEST_SCHOOL_DATA = [{value: 'test', text: 'TestSchool'}];
```
bethanyaconnor, line 13: just checking -- this is just here for now right?

apps/src/templates/SchoolInfoInputs.jsx
```diff
@@ -125,6 +125,8 @@ export default class SchoolInfoInputs extends Component {
 
     return (
       <div style={{width: 600, ...this.props.styles}}>
+        <h5>{i18n.schoolInformationHeader()}</h5>
```
bethanyaconnor, line 128: nit: use `Heading5` here

bethanyaconnor (approved):
a couple small things about headings but otherwise looks good to me!

apps/src/templates/school-association.module.scss
hannahbergam: For example, there are a few things in the pegasus design system like this (because I'm sure these are common needs all over the site): 
Screenshot 2024-02-27 at 3 36 13 PM

apps/src/templates/SchoolDataInputs.jsx
hannahbergam: Yes! Just for me to see the UI clearly before I actually have to get the zipcode field filtering all the schools

apps/src/templates/SchoolDataInputs.jsx
hannahbergam: Oh dang hahah okay we've moved to T-shirt sizing then! Going to redo all of them 
Screenshot 2024-02-27 at 3 40 18 PM

apps/src/templates/SchoolInfoInputs.jsx
hannahbergam: There was no Heading5 in that file, but that's no longer an issue since the styling I should be using is t-shirt sizing anyway😅

apps/src/templates/SchoolDataInputs.jsx
hannahbergam: Okay but now I'm seeing folks import 'Heading1' from that file... I juuuust didn't understand how the scss was working

apps/src/templates/SchoolInfoInputs.jsx
hannahbergam: Because this is on the old/existing flow, I'd rather not change the styling. These lines are just moved from the file they were rendered in before. Does that make sense? Heading5 looks fairly different from the built in h5

hannahbergam: (For more context- I moved these two lines into this component so that the new one could also have a formatted (and different string) title and it wouldn't be duplicated across the AB test

bethanyaconnor: yep thats fine!

PR #56727, opened 2024-02-22
[GoogleBlockly] use modalInputs setting instead of overriding field classes

molly-moen:
Previously we were overriding `FieldAngle`, `FieldMultilineInput`, `FieldNumber` and `FieldTextInput` in order to override `showEditor_` to not show a modal on mobile devices. However, Blockly [added a workspace flag](https://github.com/google/blockly/pull/6625), `modalInputs` that we can set to false to achieve the same effect. If we use this flag when initializing our main workspace we don't need these overrides anymore.

I discovered this while trying to migrate these files to TypeScript, but instead got to delete them 🎉 

## Testing story
Tested that when simulating a mobile device we still don't show a pop-up when editing a field in Sprite Lab or Music Lab. If I change `modalInputs` to true I do see the pop-up. I also verified this behavior on the modal function editor.

## PR Checklist:

- [ ] Code is well-commented


mikeharv (approved):
Awesome!

PR #56732, opened 2024-02-22
Lab2: no fade parameter

breville:
We're considering removing the fade between Lab2 levels, and in many such debates, the decision becomes clear when we can try out the alternatives.

With this change, a `lab2-no-fade` URL parameter can be specified, to see what level switching feels like without the fade, like this: https://studio.code.org/s/music-intro-2024/lessons/1/levels/1?lab2-no-fade=true

### Details

In this initial implementation, as soon as the load begins, the app area becomes unclickable, but continues showing until the next level is ready and appears.  We might want to adjust what is shown, and maybe even add a spinner, but interestingly this early implementation is very similar to what a modern browser does upon a link click.  It will be interesting to see how this feels on our production enviroment, with its relatively fast loading time.


apps/src/lab2/views/Lab2Wrapper.tsx
```diff
@@ -29,9 +29,14 @@ const Lab2Wrapper: React.FunctionComponent<Lab2WrapperProps> = ({children}) => {
     (state: {lab: LabState}) =>
       state.lab.pageError?.errorMessage || state.lab.pageError?.error?.message
   );
-  const overlayStyle: string = isLoading
-    ? moduleStyles.showingBlock
-    : moduleStyles.fadeInBlock;
+  const noFade = window.location.href.includes('no-fade');
```
sanchitmalhotra126: minor: maybe `'lab2-no-fade'` since it could be used across many different lab2 labs/scripts?

sanchitmalhotra126 (approved):
Cool! Played around with this a bit and it's definitely interesting. I think we'd want to have some sort of loading UI still for the workspace, but cool that this does give individual labs the opportunity manage loading UI themselves (especially if parts of the screen don't actually change)

apps/src/lab2/views/Lab2Wrapper.tsx
breville: Done.  Also moved this code to only run once.

PR #56733, opened 2024-02-22
Lab2: prevent clicks when loading

breville:
Prevent clicks from reaching UI elements that are visually hidden (but still in existence) when loading a new Lab2 level.

Prior to this change, it was possible to rapidly click on a "Continue" button in Music Lab and, even though it was no longer visible after the first click, it would keep receiving clicks and jump to a new level.

With this change, the "fade overlay" `div` consumes those clicks while loading is happening.


thomasoniii (approved):
lgtm 🚀

sanchitmalhotra126 (approved):

PR #56744, opened 2024-02-22
[Progress] Add user select dropdown

lfryemason:
## PR Checklist:

- [ ] Code is well-commented


apps/src/templates/UnitSelectorV2.jsx
```diff
@@ -0,0 +1,77 @@
...
+      const newScriptId = e.target.value;
+      setScriptId(newScriptId);
+      loadUnitProgress(newScriptId, sectionId);
+
+      this.recordEvent('change_script_v2', {
+        old_script_id: scriptId,
+        new_script_id: newScriptId,
+      });
+
+      analyticsReporter.sendEvent(EVENTS.PROGRESS_CHANGE_UNIT, {
```
lfryemason: Change to new event

apps/src/templates/UnitSelectorV2.jsx
```diff
@@ -0,0 +1,77 @@
...
+      });
+
+      analyticsReporter.sendEvent(EVENTS.PROGRESS_V2_CHANGE_UNIT, {
+        sectionId: sectionId,
+        oldUnitId: scriptId,
+        unitId: newScriptId,
+      });
+    },
+    [scriptId, setScriptId, sectionId]
+  );
```
etaderhold: The script/unit synonym makes this a bit hard to follow. Could you perhaps replace `script` with `unit` wherever it's easy? Like, `setScriptId` is probably used in a bunch of places so that can be a refactor for another time, but I'd love to see us standardize on `unit` for new local variables, event names, etc.

etaderhold (approved):

apps/src/templates/UnitSelectorV2.jsx
lfryemason: the scriptId is used in 17 files and `setScriptId` is used in 16. I think I'd prefer to put that into a different PR. I did however change the newScriptId and also set `unitId = scriptId` in the file.

PR #56745, opened 2024-02-22
[Google Blockly] Migrate renderer code to typescript

molly-moen:
Migrate our renderer-related files to TypeScript. This was mostly straightforward adding types and null checks. Our custom constant provider was not returning a `type` for its custom shapes, which I synced with the Blockly team about. It seems like `type` is not used anywhere, but to make TypeScript happy they recommended overriding the parent `SHAPES` object to add new shape types.

## Links

## Testing story
Tested that blocks and connections still render correctly. The only real code change was in the constants provider, so I ensured all our connections still have the correct shapes and shadows when connecting.

## PR Checklist:

- [ ] Code is well-commented


mikeharv (approved):
Looks good!

PR #56763, opened 2024-02-23
Revert "replace crypto-js with browserify-crypto as crypto-js is no longer maintained"

hannahbergam:
Reverts code-dot-org/code-dot-org#56602


stephenliang:
I was able to merge in #56781 so this shouldn't be needed anymore. Sorry about all the trouble with Storybook!

PR #56775, opened 2024-02-23
Convert to 100% ESM import/export

snickell:
- Hand convert 99%+ of JS/TS files to use only ECMAScript (ESM) style import/export statements.
- Enable eslint-import with use of CommonJS or AMD as an error (and some other nice import linting)
- Flag remaining CommonJS usage (sync dynamic requires that couldn't be converted to async import() statements) with `// disable-line import/no-commonjs`
- As files compile as ESM, they also become implicitly `use strict`, convert usage of globals to be explicit.
- This PR will lets us optionally build with `esbuild` and other next gen higher performance loaders which don't like mixing ESM and other module styles in single file. `esbuild` cuts build time in half, this PR was split out of a bigger PR experimenting with switching to `esbuild` (https://github.com/code-dot-org/code-dot-org/pull/53693). The `esbuild` PR gets us a 2x speed boost for `npm start`, and this is a necessary stepping stone.
- `esbuild` has a stricter reference checker, which was very helpful in finding and correctly linking all instances.

TODO:
- [x] Get all karma unit tests running
- [x] Get karma integration tests running
- [x] Check how much extra time is taken by Karma when generating `inline-source-maps`
- [ ] get dashboard ui tests working
- [ ] Get eyes tests running (remember commit tag to make drone run them)


Hamms:
I like the goals that this PR sets out to achieve, but I am a bit worried about trying to land this many changes all at once. Particularly looking at it after a few weeks and seeing how many conflicts there are; what do you think about batching this? First land a handful of individual PRs updating CommonJS imports to ECMAScript imports wherever we can before eventually enabling eslint only once the number of remaining violations is manageable.

snickell:
Oops, this was a resurrection of an old PR, and didn't properly set it as a draft, sorry for the misflag! Its not at a mergeable stage at all - that said, its unfortunately not breakable into sub-PRs, which is part of what makes this a nightmare (long story: but we actually have a bizarre setup where we sniff whether its an ES module or CJS, and removing the last require will flip it to ES, and cause a cascade of changes in subtle situations for everyone who imports it...)

Hamms:
9f8x6MF

Issue #56774, opened 2024-02-23
Chrome users are logged out of code.org when the browser is closed


snickell:
Users of other browsers stay logged into code.org (Edge, Firefox), but it seems Chrome users are logged out whenever the browser closes.

Per https://developer.chrome.com/blog/cookie-max-age-expires :
> As of Chrome release [M104 (August 2022)](https://chromiumdash.appspot.com/schedule) cookies can no longer set an expiration date more than 400 days in the future.
>
> This change does not impact session cookies—cookies that do not explicitly set an expiration date with Max-Age or Expires—as these are instead cleared when the browsing session ends.

This seems potentially related.

I'm not sure if we overload the session cookie for auth as well (???), but at the very least, we set no expiration on it, so if it is related to auth chrome would be clearing it on each browser close: https://github.com/code-dot-org/code-dot-org/blob/fdf7a2efa1e634497e993f6fae48e216dbe99c58/dashboard/config/initializers/session_store.rb/#L3-L6

snickell:
We're using devise for auth, and its entirely possible it sets its own cookies, which could ALSO be the source of the surprise expiry, see: https://stackoverflow.com/questions/26564174/user-logged-out-automatically-when-browser-is-closed-when-using-devise-gem

1. We do enable devise's :rememberable hooks in user.rb: https://github.com/code-dot-org/code-dot-org/blob/fdf7a2efa1e634497e993f6fae48e216dbe99c58/dashboard/app/models/user.rb/#L174-L175
2. In `devise.rb` we configure the browser to be remembered for 2 weeks (https://github.com/code-dot-org/code-dot-org/blob/fdf7a2efa1e634497e993f6fae48e216dbe99c58/dashboard/config/initializers/devise.rb/#L170-L172):
```
  # ==> Configuration for :rememberable
  # The time the user will be remembered without asking for credentials again.
  config.remember_for = 2.weeks
```

When logged out of studio.code.org, here's the potentially relevant cookies:
<img width="483" alt="image" src="https://github.com/code-dot-org/code-dot-org/assets/223277/06a1bb21-9e73-4d5b-a3b8-9172c666bcf1">

snickell:
Changing the session store cookie to a 200 day expiration ("one login per school year") results in being able to quit chrome, and when its re-opened we're still logged in:
```
Dashboard::Application.config.session_store :cookie_store,
  key: session_cookie_key,
[...]

PR #56777, opened 2024-02-23
Don't logout chrome browsers on close

snickell:
We discovered Chrome browsers are logged out of studio.code.org every time they are closed. This does not appear to be true for Firefox and Edge, and is not the expected or desired behavior. See issue #56774 for details.

I immediately suspected this issue: https://developer.chrome.com/blog/cookie-max-age-expires. TL;DR: if you set a cookie expiration date over 400 days, Chrome will expire your cookie every time the browser is fully closed (like a session cookie).

Editing our primary cookie in session_store.rb produced the desired behavior: Chrome now preserved logins after I restarted the browser.

However, this behavior is confusing because:
1. Naively I would expect session_store.rb to be a "per session" cookie. However, since we're using it to store /everything/ not just session data, perhaps there's another place where we're setting it to be a general "everything while I'm logged in" cookie? This is actually quite common in a modern context, I believe session cookies have fallen out of vogue for storing "rails session" like values.
2. We also have some usage and configuration of a separate auth cookie in user.rb, but I cannot find it actually working or being created on any browser. I think this is historical cruft. (more to come on this)

I don't have much more time to spend on this, but despite the ambiguity, it may be worth merging this PR because:
1. It works on Firefox, Edge and Chrome
2. Gives the desired behavior on Chrome
3. Its a major student impact if we have Chrome browsers requesting login every time they are opened.
4. If we don't fix it soon, odds are it will get forgotten again. This is a very important UX detail that's not visible in a screenshot.

@dancodedotorg @dju90

### DIFF
--- a/dashboard/config/initializers/session_store.rb
+++ b/dashboard/config/initializers/session_store.rb
@@ -3,4 +3,5 @@ session_cookie_key = environment_specific_cookie_name('_learn_session')
 Dashboard::Application.config.session_store :cookie_store,
   key: session_cookie_key,
   secure: !CDO.no_https_store && (!Rails.env.development? || CDO.https_development),
-  domain: :all
+  domain: :all,
+  expire_after: 200.days # don't expire in the same school year


snickell:
@jmkulwik mentions cookie overflow could be relevant, it does seem like having our cookie last longer would increase the risk of overflow on Chrome. OTOH, it would match the risk on Firefox and Edge.

dju90:
Wondering if Platform team (@alivira and @bdmesh) could take this across the finish line

snickell:
Jess suggested Dave take a look, since he's dealt with cookie issues before.

davidsbailey (approved):
nice find! agreed, the name "session" is ambiguous because it could be referring to the rails session (persistent) or session storage (transient).

PR #56781, opened 2024-02-23
fix(crypto): replace crypto-browserify with the third party md5 dependency

stephenliang:
crypto-browserify was a polyfill added to replace the crypto-js dependency which is no longer maintained. This polyfill did not work properly in storybook. Since Code.org does not currently use any other hashing library other than md5, it makes sense to scope down and use a purpose built third party dependency whose sole responsibility is to create md5 hashes.

fixes dependabot moderate [Dependabot #434](https://github.com/code-dot-org/code-dot-org/security/dependabot/434) and critical  [Dependabot #486](https://github.com/code-dot-org/code-dot-org/security/dependabot/486)

## Links

1. [crypto-js discontinued announcement](https://github.com/brix/crypto-js)
2. [Dependabot #486](https://github.com/code-dot-org/code-dot-org/security/dependabot/486)
3. [Dependabot #434](https://github.com/code-dot-org/code-dot-org/security/dependabot/434)

## Testing story

I ran storybook and verified stories loaded correctly. I also ran rake test and the UI tests. The following tests were run to ensure that the md5 hasher is consistent with crypto-js:

```
apps/test/unit/templates/instructions/InlineAudioTest.js
apps/test/unit/templates/curriculumCatalog/curriculumCatalogExpandedCardTest.jsx
```

## Follow-up work

No follow up necessary.

## Privacy

This does not change any user models or interact with privacy.

## Security

This addresses the discontinued crypto-js library by switching to an actively maintained and more mainline browserify library.

## PR Checklist:

- [x] Code is well-commented


apps/webpack.config.js
```diff
@@ -135,7 +135,7 @@ const nodePolyfillConfig = {
       'process/browser': require.resolve('process/browser'),
       stream: require.resolve('stream-browserify'),
       timers: require.resolve('timers-browserify'),
-      crypto: require.resolve('crypto-browserify'),
+      crypto: false,
```
bencodeorg, line 138: What does this do now?

bencodeorg (approved):
Looks reasonable to me -- thanks for taking care of this so quickly!

apps/webpack.config.js
stephenliang: This was introduced when I swapped from crypto-js to crypto-browserify. Since I'm swapping from crypto-browserify to md5, the polyfill is no longer necessary and this reverts to the previous behavior which is not to fallback to a polyfill if crypto cannot be resolved.

bencodeorg: ah right, forgot this is a semi-revert. Out of curiosity, is this different than this not being here at all? Sounds like this maybe was necesssary in webpack 4, but not 5?

https://webpack.js.org/configuration/resolve/#resolvefallback

stephenliang: I believe it is no different than leaving it out (the value would be `undefined`). Webpack 5 dropped automatic polyfills which essentially made them false/undefined by default.

bencodeorg: Ok cool -- I'd vote to delete it then (less is more)?

stephenliang: I'd prefer to make this change separately so we can unblock storybook, just to make sure removing this statement doesn't cause another breakage. Does that work for you? I can create a followup PR.

bencodeorg: Fine with me!

wilkie: I'm also in favor of removing any line from the webpack config we don't need. sounds ok to make it its own PR if you'd like.

wilkie (approved):
LGTM 👍

hannahbergam (commented):
Hi! I checked out this branch locally. I restarted dahsboard-server, ran a yarn clean and a yarn build, and then restarted yarn and storybook. I'm still getting 'Module not found: Error: [CaseSensitivePathsPlugin] `/Users/hannahbergam/code-dot-org/apps/node_modules/md5/md5.js` does not match the corresponding path on disk `MD5`.' Are there other steps I need to take to get this working locally? Thanks in advance!

stephenliang:
@hannahbergam I think this is due to MacOS case insensitive volumes clashing between firebase-mock's dependency on MD5 and our use of the package md5. MD5 was renamed to md5 so upgrading firebase-mock is necessary to fix it for Mac users. I just pushed a new change that fixes this by upgrading firebase-mock as well.

hannahbergam:
> @hannahbergam I think this is due to MacOS case insensitive volumes clashing between firebase-mock's dependency on MD5 and our use of the package md5. MD5 was renamed to md5 so upgrading firebase-mock is necessary to fix it for Mac users. I just pushed a new change that fixes this by upgrading firebase-mock as well.

This most recent commit fixed storybook for me! Thanks!

apps/webpack.config.js
snickell: Given how touchy this part of the code has been, I think Stephen's instinct to make it a separate PR is good

snickell (approved):
image

and.... SOLD! :-)

stephenliang:
# Background

A change was made to replace crypto-js with the native node.js crypto library. This was done because crypto-js is no longer maintained and has a number of critical dependabot vulnerabilities assigned to it. Code.org uses crypto-js solely for its md5 hashing functionality. A couple alternative dependencies were selected:

* The native node.js crypto library 
   * This was selected as using the native library would eliminate the need for a third party dependency as it comes with node.js. However, browsers do not have this library and so the polyfill “browserify-crypto” is needed for client side rendering.
   * This was also suggested by crypto-js’ deprecation documentation
* The “md5” library
   * This would introduce a new third party dependency whose scope is strictly related to md5 hashing.
   * This library would be fairly lightweight
   * If Code.org were to use a different crypto hashing function, a new library would need to be brought in (or the crypto library). To avoid redundancies, the native crypto library for node.js was selected with the crypto-browserify polyfill for the browser.

# What broke?

In storybook, several stories were broken by this change:

```
Unexpected error while loading ./weblab/dialogs/ResetSuccessDialog.story.js: Buffer is not defined
 ReferenceError: Buffer is not defined
    at ./node_modules/create-hash/make-hash.js (http://localhost:9001/vendors-node_modules_amplitude_analytics-browser_lib_esm_index_js-node_modules_code-dot-org_j-22d7f3.iframe.bundle.js:101904:18)
    at __webpack_require__ (http://localhost:9001/runtime~main.iframe.bundle.js:28:33)
    at fn (http://localhost:9001/runtime~main.iframe.bundle.js:351:21)
    at ./node_modules/create-hash/md5.js (http://localhost:9001/vendors-node_modules_amplitude_analytics-browser_lib_esm_index_js-node_modules_code-dot-org_j-22d7f3.iframe.bundle.js:101953:16)
    at __webpack_require__ (http://localhost:9001/runtime~main.iframe.bundle.js:28:33)
    at fn (http://localhost:9001/runtime~main.iframe.bundle.js:351:21)
    at ./node_modules/create-hash/browser.js (http://localhost:9001/vendors-node_modules_amplitude_analytics-browser_lib_esm_index_js-node_modules_code-dot-org_j-22d7f3.iframe.bundle.js:101841:11)
    at __webpack_require__ (http://localhost:9001/runtime~main.iframe.bundle.js:28:33)
    at fn (http://localhost:9001/runtime~main.iframe.bundle.js:351:21)
    at ./node_modules/crypto-browserify/index.js (http://localhost:9001/vendors-node_modules_amplitude_analytics-browser_lib_esm_index_js-node_modules_code-dot-org_j-22d7f3.iframe.bundle.js:103212:37)
```

This error occurs because the md5 hashing function comes from the node.js native crypto library. However, browsers do not execute on node.js and therefore require a [browser polyfill](https://en.wikipedia.org/wiki/Polyfill_(programming)). When polyfilling the crypto library, the polyfill in itself uses another native node.js library “Buffer”.

Buffer is defined as a global variable in Node.js and therefore requires the injection into the window object using the [Webpack Provide Plugin](https://webpack.js.org/plugins/provide-plugin/). Our usage of md5, however, does not require the usage of the provide plugin as this is imported via the crypto package. (Buffer can be instantiated as new Buffer() without an import)

Code.org does currently utilize a [Buffer polyfill fallback](https://github.com/code-dot-org/code-dot-org/blob/staging/apps/webpack.config.js#L132) and the fallback is [provided by the webpack provide plugin to the browser](https://github.com/code-dot-org/code-dot-org/blob/staging/apps/webpack.config.js#L82).

However, while Storybook’s webpack config does inherit the base webpack config for polyfill fallbacks, it [does not inherit the Provide configuration](https://github.com/code-dot-org/code-dot-org/blob/staging/apps/webpackStorybook.config.js#L34). (Note the lack of …WEBPACK_BASE_CONFIG.plugins)

Therefore, while the fallback for Buffer is enabled, the Buffer fallback is not provided as a global, resulting in the Storybook crash.

# Why did Storybook break after the crypto dependency change?
The CI/CD process does not currently test that Storybook loads without errors. 

# Why did the crypto change break Storybook but not Code.org?
The crypto library uses a global “Buffer” class which can be used without an import. Code.org’s webpack config provides this global but storybook does not. Therefore, when Code.org executes the md5 hasher, it will not crash as the class exists on window. On storybook, the class is not provided and results in the error.

# Why did the md5 hashing library need to be changed?
The crypto-js library is no longer maintained and has resulted in numerous critical security vulnerabilities on dependabot. 

# Why not use the ‘md5’ library to replace crypto-js?
Rather than introducing another dependency to Code.org, preferring to use the native Crypto library would eliminate the need to have a dependency on the server side. However, a polyfill would still be needed on the client side. The crypto library being supported by Node.js and the popular browserify organization would help improve our security by being actively maintained by fundamental libraries in the javascript ecosystem. Adding a third party library could result in an explosion of other hashing libraries in our dependency tree. (Adding shasum, etc. as third party libraries)

Temporarily, the “md5” library was used to resolve this issue, however further discussion is warranted as to the most appropriate library to use in this case.

PR #56788, opened 2024-02-23
[Google Blockly] Convert remaining fields to TypeScript

molly-moen:
Convert the remaining Field subclasses to TypeScript. The only changes here are types and null checks, with the occasional `any` escape hatch because we are accessing a private field we cannot access otherwise. The files are
- cdoFieldImage
- cdoFieldImageDropdown
- cdoFieldToggle
- cdoFieldVariable: unfortunately github thinks I added and removed a file rather than renaming it. You can see the actual changes in [this commit](https://github.com/code-dot-org/code-dot-org/pull/56788/commits/da49c91fcbdefa0ad457663d66509eceba8c4b20)

## Links

## Testing story
Tested that the affected fields still work.

## PR Checklist:

- [ ] Code is well-commented


mikeharv (approved):
Changes look good. There are quite a few places here and in your previous PRs where we have to use `any` as a workaround. Are we going to share a list of these with Blockly so that they can potentially be filed as bugs?

molly-moen:
> Changes look good. There are quite a few places here and in your previous PRs where we have to use `any` as a workaround. Are we going to share a list of these with Blockly so that they can potentially be filed as bugs?

@mikeharv good idea. I'll summarize and share these once I reach a stopping point!

PR #9315, opened 2016-07-05
New Maker API blocks (including `onBoardEvent`, `accelerometer`, `sensor`s)

bcjordan:

PR #56419, opened 2024-02-07
Remove unused Maker getAveragedValue command

fisher-alice:
This PR removes the `getAveragedValue` command from the App Lab toolbox. The block is currently included in the 'Circuit' and 'micro:bit' drawers. Note that `LookbackLogger` was only used for `getAveragedValue` in `PlaygroundComponents.js` so it is deleted by this PR.

A user reported in [this ZenDesk ticket](https://codeorg.zendesk.com/agent/tickets/483773) that the function did not seem to be working when the Circuit Playground is connected. I am able to repro and the only value returned is `0`.
After investigation which is documented [in this Slack thread](https://codedotorg.slack.com/archives/C03DBDN67B7/p1707149892882419), it was agreed by curriculum and product to remove this function for both the CP and micro:bit as it is not used in the curriculum nor are there plans to use this function in the future.

BEFORE update:

App Lab toolbox - Circuit Playground drawer - function block is included
CP-before

AFTER update:

App Lab toolbox - Circuit Playground drawer - function block NOT included

CP-after

App Lab toolbox - micro:bit drawer - function block NOT included

microbit-after

App Lab program run with `getAveragedValue` function - not defined
after-cp-not-defined

## Links
[jira](https://codedotorg.atlassian.net/browse/LABS-481)
[ZenDesk ticket](https://codeorg.zendesk.com/agent/tickets/483773)
[Slack thread](https://codedotorg.slack.com/archives/C03DBDN67B7/p1707149892882419)

## Testing story
Tested locally and removed unit tests for the `getAveragedValue` function.

## PR Checklist:

- [ ] Tests provide adequate coverage
- [ ] Privacy and Security impacts have been assessed
- [ ] Code is well-commented
- [ ] New features are translatable or updates will not break translations
- [ ] Relevant documentation has been added or updated
- [ ] User impact is well-understood and desirable
- [ ] Pull Request is labeled appropriately
- [ ] Follow-up work items (including potential tech debt) are tracked and linked

PR #56792, opened 2024-02-23
Removed package previously used in maker offerings.

jmkulwik:

the data-collection package was used in the early days of our Maker project, but it doesn't appear to be used anymore. Here is the original PR: https://github.com/code-dot-org/code-dot-org/pull/9315

### DIFF
--- a/apps/package.json
+++ b/apps/package.json
@@ -220,7 +220,6 @@
     "crypto-browserify": "^3.12.0",
     "csv-parse": "^4.4.6",
     "dapjs": "^2.3.0",
-    "data-collection": "^1.1.6",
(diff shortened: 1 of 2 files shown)


sanchitmalhotra126 (approved):

snickell (approved):

fisher-alice (approved):
Thanks for taking care of this! Just noting that we removed the import to `data-collection` in https://github.com/code-dot-org/code-dot-org/pull/56419.

jmkulwik:
> Thanks for taking care of this! Just noting that we removed the import to `data-collection` in #56419.

Oh wow! That was very recent!! Thank you for solving the mystery of where that got removed!!

PR #56671, opened 2024-02-20
P20-629: Implement independent i18n sync-down of Dashboard resources

artem-vavilov:
1. [Sync-down source file translations in alphabetical order](https://github.com/code-dot-org/code-dot-org/commit/251631637ef3a1a520fb3e2d2ba46f61c5d8bf73)
2. [Created the `Dashboard/BaseContent` i18n sync-down script](https://github.com/code-dot-org/code-dot-org/commit/e1dfc7f44fb5cbff96888f4e1e55837a7097dd0c)
3. [Created the `Dashboard/Blocks` i18n sync-down script](https://github.com/code-dot-org/code-dot-org/commit/b293cd3b7e559ce9edf3ae13c63f3ec08d2d65f3)
4. [Created the `Dashboard/CourseContent` i18n sync-down script](https://github.com/code-dot-org/code-dot-org/commit/ddf403dcfd11944ba6644a187c1b936e2070b59b)
5. [Created the `Dashboard/CourseOfferings` i18n sync-down script](https://github.com/code-dot-org/code-dot-org/commit/f0b72f7fd6a1b1b759dfe86a30eecf592fc5ece1)
6. [Created the `Dashboard/Courses` i18n sync-down script](https://github.com/code-dot-org/code-dot-org/commit/3c86ec82d14750854a8d9bd1d38bfe2a430afc63)
7. [Created the `Dashboard/CurriculumContent` i18n sync-down script](https://github.com/code-dot-org/code-dot-org/commit/0089d4e7cffeb1bc20e3ee05c24f98ee8cbbd635)
8. [Created the `Dashboard/DataContent` i18n sync-down script](https://github.com/code-dot-org/code-dot-org/commit/e5b6753a2dca49fe9acf0a2cacbe1dfb6d461180)
9. [Created the `Dashboard/DeviseContent` i18n sync-down script](https://github.com/code-dot-org/code-dot-org/commit/bc75e784b575bd6670c74fe99f591671744e6ca6)
10. [Created the `Dashboard/Docs` i18n sync-down script](https://github.com/code-dot-org/code-dot-org/commit/f2f7795232ec766071531b94ef81345c4f19f4b3)
11. [Created the `Dashboard/MarketingAnnouncements` i18n sync-down script](https://github.com/code-dot-org/code-dot-org/commit/a4a0b35aaec9055f9e40e4d4a9f0c4088a9e2ea2)
[...]

PR #56630, opened 2024-02-16
P20-628: Implement independent i18n sync-down of Apps resources

artem-vavilov:
1. Created independent `sync-down` script for the `Apps/Animations` i18n resource.
2. Created independent `sync-down` script for the `Apps/ExternalSources` i18n resource.
3. Created independent `sync-down` script for the `Apps/Labs` i18n resource.
4. Sync-down translation files to the separate `i18n/crowdin` dir instead of `i18n/locales`.

- jira ticket: [P20-628](https://codedotorg.atlassian.net/browse/P20-628)

References:
- https://github.com/crowdin/crowdin-api-client-ruby/pull/69
- https://github.com/crowdin/crowdin-api-client-ruby/pull/70
- https://github.com/crowdin/crowdin-api-client-ruby/pull/75

PR #56796, opened 2024-02-23
P20-630: Apply new sync-down changes

artem-vavilov:
- JIRA ticket: [P20-630](https://codedotorg.atlassian.net/browse/P20-630)
- https://github.com/code-dot-org/code-dot-org/pull/56630
- https://github.com/code-dot-org/code-dot-org/pull/56671
- https://github.com/code-dot-org/code-dot-org/pull/56739


bin/i18n/utils/sync_out_base.rb
```diff
@@ -43,7 +43,7 @@ def perform
         I18nScriptUtils.process_in_threads(languages) do |language|
           process(language)
         ensure
-          I18nScriptUtils.remove_empty_dir I18nScriptUtils.locale_dir(language[:crowdin_name_s])
+          I18nScriptUtils.remove_empty_dir I18nScriptUtils.crowdin_locale_dir(language[:locale_s])
```
mgc1194, line 46: Synced down files go to a temporal directory `i18n/crowdin` instead of `i18n/locales`

bin/i18n/sync-out.rb
```diff
@@ -6,9 +6,6 @@
 require File.expand_path('../../../dashboard/config/environment', __FILE__)
 require 'cdo/languages'
 
-require 'cdo/crowdin/legacy_utils'
-require 'cdo/crowdin/project'
-
```
mgc1194, line 11: Are these just unused `require` or have they been moved somewhere else?

artem-vavilov: It's just something I forgot to remove during the sync-out refactoring

mgc1194 (approved):
I tested all changes locally. it took me a while to sun the whole sync but it was worth it!
I approve the changes, the only comment I have is, that I believe, we don't remove some of the temporary folders. 
 
Screenshot 2024-02-26 at 16 49 10

We should make sure they get removed after sync-out is completed or include them to git ignore.

artem-vavilov:
> We should make sure they get removed after sync-out is completed or include them to git ignore.

They are already in .`gitignore`. But it would be better if we removed them completely. I'll take a look at this.
https://github.com/code-dot-org/code-dot-org/blob/f07b45c262aeeaae0ed425547ca3bab7e3b8a794/.gitignore#L101

PR #34207, opened 2020-04-13
render FreeResponse content side-by-side

Hamms:
a la https://github.com/code-dot-org/code-dot-org/pull/33865

Compare the results of rendering FreeResponse levels as we would if rendering with SafeMarkdown with the current rendering strategy, and log discrepancies to firehose for eventual cleanup.

## Links

- [jira](https://codedotorg.atlassian.net/browse/FND-1100)

## Testing story

# Reviewer Checklist:

- [ ] Tests provide adequate coverage
- [ ] Code is well-commented
- [ ] New features are translatable or updates will not break translations
- [ ] Relevant documentation has been added or updated
- [ ] User impact is well-understood and desirable
- [ ] Pull Request is labeled appropriately
- [ ] Follow-up work items (including potential tech debt) are tracked and linked

PR #56800, opened 2024-02-23
Removing package that was previously used in SafeMarkdown project.

jmkulwik:

The rehype-stringify package was used in the safeMarkdown project, but it doesn't appear to be used anymore. Here is the original PR: https://github.com/code-dot-org/code-dot-org/pull/34207

Usage of rehype-stringify [was removed](https://github.com/code-dot-org/code-dot-org/pull/34784) when the project was finished and was not used in the final [SafeMarkdown component](https://github.com/code-dot-org/code-dot-org/blob/staging/apps/src/templates/SafeMarkdown.jsx).

### DIFF
--- a/apps/package.json
+++ b/apps/package.json
@@ -306,7 +306,6 @@
     "rehype-raw": "^4.0.0",
     "rehype-react": "^3.1.0",
     "rehype-sanitize": "^2.0.2",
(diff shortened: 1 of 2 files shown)


Hamms (approved):
Hey, nice catch! Thanks for the cleanup; LGTM

PR #50601, opened 2023-03-06
[Micro:bit] Add button on maker/setup page to flash Firmata

fisher-alice:
This PR adds a button on the 'maker/setup' page which appears if the connected micro:bit fails the third 'Board connectable' setup check.

Screen Shot 2023-03-06 at 5 26 20 PM

When the user clicks on the button, the user is prompted to select the micro:bit via webUSB. Once the device is selected, the correct version of the Firmata .hex file is transferred to the micro:bit. The process takes between 30 - 45 seconds so the percent update completion is displayed. 

Screen Shot 2023-03-06 at 5 26 34 PM

Once the file is transferred, the user is prompted to click on the re-detect button to confirm that the transfer was successful and their board is now 'connectable'. 

Screen Shot 2023-03-06 at 7 11 20 PM

Two new packages are used in this PR:
- [`dapjs`](https://www.npmjs.com/package/dapjs) - an interface to 'CMSIS-DAP' which enables access to the micro:bit using [`WebUSB`](https://developer.mozilla.org/en-US/docs/Web/API/WebUSB_API),
- [`usb`](https://www.npmjs.com/package/usb) - enables communication with USB devices

Prior to this update, if the micro:bit's Firmata is outdated, users are prompted to:
1. Click on a [link to instructions on GitHub](https://github.com/microbit-foundation/microbit-firmata#installing-firmata-on-your-bbc-microbit).
2. Click on a [link that goes to latest release of Firmata](https://github.com/microbit-foundation/microbit-firmata/releases).
3. Click on a link to download the .hex file.
4. Open two windows - one with downloaded file and one with micro:bit virtual USB drive.
5. Click and drag the .hex file from downloads to micro:bit window.

Note that we will need to explain this pathway to update the Firmata if the new pathway of using `WebUSB` does not work for the user.

Video before update:

https://user-images.githubusercontent.com/107423305/223189105-266f2cda-afc0-459a-b255-462ca6fa5623.mp4

micro:bit created [this website](https://microbit-carlos.github.io/dapjs/examples/universal-flash/web.html) which allows a user to drop the Firmata .hex file onto the page in order to flash the micro:bit via WebUSB. However, the user would still need to download the .hex file before clicking and dragging the file icon to the web page. 

In order to streamline this process, product and curriculum asked that we add a button that when clicked will automatically transfer the .hex file to the micro:bit without the user having to click and drag the file icon onto a web page. 

[...]

PR #56801, opened 2024-02-23
Remove unused usb package

jmkulwik:

It looks like usb was added as part of the recent Maker firmware work, however, it looks like it didn't end up being used. @fisher-alice, would you mind checking my work here? Here's where the usb package was added: https://github.com/code-dot-org/code-dot-org/pull/50601

### DIFF
--- a/apps/package.json
+++ b/apps/package.json
@@ -322,7 +322,6 @@
     "swiper": "^11.0.5",
     "timers-browserify": "^2.0.12",
     "tone": "14.7.77",
-    "usb": "^2.8.1",
     "video.js": "7.6.6",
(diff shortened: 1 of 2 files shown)


snickell (commented):
I think my review is gonna be useless here, I don't see this package getting used either, but much more interested in what Alice thinks

fisher-alice (approved):
LGTM! Thanks for cleaning this up.  😁 
I ended up using `dapjs` which includes `webusb` [as a dependency](https://github.com/ARMmbed/dapjs/blob/6a33cc90437b1f0cf3860d31c5b94cca8854eeae/package.json#L71).

PR #56803, opened 2024-02-23
Removed unused web-streams-polyfill package

jmkulwik:

From what I can tell, the web-streams-polyfill package was a candidate for polyfilling some lost functionality when we upgraded from Webpack 4 to Webpack 5. However, I _think_ we got that functionality via stream-browserify instead. Maddie left some great notes when she did this work: https://github.com/code-dot-org/code-dot-org/pull/48105. @snickell and @bencodeorg, I could definitely use a hand checking my assumptions on this one if you get the chance.

### DIFF
--- a/apps/package.json
+++ b/apps/package.json
@@ -326,7 +326,6 @@
     "video.js": "7.6.6",
     "vm-browserify": "^1.1.2",
     "vmsg": "^0.3.6",
-    "web-streams-polyfill": "^3.1.0",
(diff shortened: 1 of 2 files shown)


bencodeorg (approved):
Looks right to me!

snickell (approved):
I did a double check and agree that we don't seem to be using this pollyfill. This package weighs in at an astonishing 10MB, so pretty glad we /didn't/ use this pollyfill ;-)

PR #56805, opened 2024-02-23
[Google Blockly] Convert serialization files to TypeScript

molly-moen:
Migrate serialization-related files to TypeScript. This is mostly typing and null checks, I added PR comments to the more interesting changes. I probably did not type everything perfectly here. Part of the problem is Blockly does not define the serialization state as anything by `any`, so I used `any` for some objects to make progress. I also made some (likely) incomplete types for xml and json configurations, which we can update as we find more properties we need. This is safe to do because of TypeScript's [type compatability](https://www.typescriptlang.org/docs/handbook/type-compatibility.html) rules.

Files migrated
- `cdoGenerator`
- `cdoMetricsManager`. This one isn't as tied to serialization, but was a simple migration.
- `cdoSerializationHelpers`
- `cdoXml`
## Links

## Testing story
Tested locally. I verified the meaningful code changes did not cause regressions.

## PR Checklist:

- [ ] Code is well-commented


apps/src/blockly/addons/cdoGenerator.ts
```diff
@@ -8,11 +13,9 @@ export default function initializeGenerator(blocklyWrapper) {
 
   // This function was a custom addition in CDO Blockly, so we need to add it here
   // so that our code generation logic still works with Google Blockly
-  blocklyWrapper.Generator.xmlToBlocks = function (name, xml) {
+  blocklyWrapper.Generator.xmlToBlocks = function (_name, xml) {
     const div = document.createElement('div');
-    const workspace = blocklyWrapper.createEmbeddedWorkspace(div, xml, {
-      disableEventBindings: true,
```
molly-moen, line 14: this is a cdo-specific setting that we don't need on Google Blockly.

apps/src/blockly/addons/cdoSerializationHelpers.ts
```diff
@@ -166,7 +185,7 @@ function adjustBlockPositions(blocks, workspace) {
 
     // Set initial position; collision area must be updated to account for new position
     // every time block is moved
-    block.moveTo({x, y});
+    block.moveTo(new Blockly.utils.Coordinate(x, y));
```
molly-moen, line 188: `moveTo` [expects a Coordinate](https://developers.google.com/blockly/reference/js/blockly.blocksvg_class.moveto_1_method.md). Our previous method worked because the [method](https://github.com/google/blockly/blob/a517317b77c1e58727b748a99f9ae70b2c27d26a/core/block_svg.ts#L440) only references `.x` and `.y`, but it isn't guaranteed to continue to work so we should use the correct object

apps/src/blockly/addons/cdoXml.ts
```diff
@@ -10,39 +12,33 @@ import {
...
+export default function initializeBlocklyXml(
+  blocklyWrapper: BlocklyWrapperType
+) {
+  // Clear xml namespace. This property is readonly in Google Blockly.
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  (blocklyWrapper.utils.xml as any).NAME_SPACE = '';
 
   // Aliasing Google's domToBlock() so that we can override it, but still be able
   // to call Google's domToBlock() in the override function.
-  blocklyWrapper.Xml.originalDomToBlock = blocklyWrapper.Xml.domToBlock;
```
molly-moen, line 19: there wasn't a reason to assign this to the object since we only use it below, so I updated this to just be a constant.

apps/src/blockly/addons/cdoGenerator.ts
```diff
@@ -8,11 +13,9 @@ export default function initializeGenerator(blocklyWrapper) {
 
   // This function was a custom addition in CDO Blockly, so we need to add it here
   // so that our code generation logic still works with Google Blockly
-  blocklyWrapper.Generator.xmlToBlocks = function (name, xml) {
+  blocklyWrapper.Generator.xmlToBlocks = function (_name, xml) {
```
Erin007, line 16: Just so I can learn, what's the convention that prepends  _?

molly-moen: It means the parameter is unused.

apps/src/blockly/addons/cdoSerializationHelpers.ts
```diff
@@ -405,13 +429,21 @@ export function convertFunctionsXmlToJson(functionsXml) {
...
  * @param {Object} proceduresState - The shared procedures in JSON (blocks and procedures).
  * @returns {Object} - The updated project state with shared procedures appended.
+ * TODO: define a type for projectState and proceduresState. Blockly defines these as any.
  */
-export function appendProceduresToState(projectState, proceduresState) {
+export function appendProceduresToState(
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  projectState: any,
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  proceduresState: any
```
Erin007, line 438: I assume `projectState` and `proceduresState` are some kind of complex nested objects where we can't predict what exactly will be in there?

apps/src/blockly/types.ts
```diff
@@ -146,6 +151,8 @@ export type GoogleBlocklyInstance = typeof GoogleBlockly;
 export interface ExtendedBlockSvg extends BlockSvg {
   isVisible: () => boolean;
   isUserVisible: () => boolean;
+  // used for function blocks
+  functionalSvg_?: BlockSvgFrame;
```
Erin007, line 155: do we need the underscore at the end here? similar to above - still learning 😃

Erin007 (approved):
I'm neither a Typescript nor Blockly expert, but the changes seem logical and well commented. A couple of questions for my own edification. Presumably these changes would break tests if there was an issue?

apps/src/blockly/addons/cdoGenerator.ts
ebeastlake: I feel like we have a lot of unused parameters in Blockly, so this is useful for me to know; thanks, Molly!

apps/src/blockly/addons/cdoXml.ts
```diff
@@ -10,39 +12,33 @@ import {
...
+  const originalDomToBlock = blocklyWrapper.Xml.domToBlock;
   // Override domToBlock so that we can gracefully handle unknown blocks.
   blocklyWrapper.Xml.domToBlock = function (
-    xmlBlock,
-    workspace,
-    parentConnection,
```
ebeastlake, line 24: We don't need these arguments anymore?

apps/src/blockly/addons/cdoXml.ts
```diff
@@ -216,7 +213,7 @@ export function addMutationToBehaviorBlocks(blockElement) {
     USER_CREATED_XML_ATTRIBUTE,
     FALSEY_DEFAULT
   );
-  mutationElement.setAttribute('userCreated', userCreated);
+  mutationElement.setAttribute('userCreated', `${userCreated}`);
```
ebeastlake, line 216: Nice catch! Can you confirm these should be strings because the `mutationElement` is still an XML element? (We're moving it from the block XML element to the mutation XML element?)

ebeastlake (approved):

apps/src/blockly/addons/cdoXml.ts
molly-moen: they should be strings because all `setAttribute` values should be strings. It's a bit of a TypeScript annoyance, the function says [string](https://developer.mozilla.org/en-US/docs/Web/API/Element/setAttribute#value) but it will coerce any value to a string. But TypeScript enforces string.

apps/src/blockly/addons/cdoXml.ts
molly-moen: they are not part of the [original api](https://developers.google.com/blockly/reference/js/blockly.xml_namespace.domtoblock_1_function.md) so they weren't doing anything

apps/src/blockly/addons/cdoSerializationHelpers.ts
molly-moen: Kind of, I am actually working on a follow-up that defines these precisely. I just needed to put `any` here to make progress.

apps/src/blockly/types.ts
molly-moen: Underscore at the end is just a convention. This is a variable name we are already using so I didn't want to change it. Blockly had a lot of variables with `_` at the end, and I think we adopted that in some places. But they are moving away from that syntax, and I think we could too.

PR #52230, opened 2023-06-06
P20-216: Correct sync-all script algorithm

artem-vavilov:
## Issue
Because we squash-merge the `i18n-sync-down-out-mm-dd-yyyy` PR, the following `i18n-sync-in-up-mm-dd-yyyy` PR includes all the commits and changes made during the previous `sync-down-out`.

## Soluction
It makes more sense first run `sync-down-out` on the previous `I18n-sync-in-up-*` or `staging` branch to download existing translations for the last `sync-in-up` and then run `sync-in-up` on the `staging` branch as it contains the most relevant English content.

## JIRA
[P20-216](https://codedotorg.atlassian.net/browse/P20-216)

- [ ] Tests provide adequate coverage
- [ ] Privacy and Security impacts have been assessed
- [ ] Code is well-commented
- [ ] New features are translatable or updates will not break translations
- [ ] Relevant documentation has been added or updated
- [ ] User impact is well-understood and desirable
- [ ] Pull Request is labeled appropriately
- [ ] Follow-up work items (including potential tech debt) are tracked and linked

PR #56810, opened 2024-02-26
P20-739: Fix `sync-all` after resetting the `i18n` server

artem-vavilov:
### Issue
The clean run of `sync-all` always fails after resetting the `i18n` server because it executes `sync-out` first, which depends on the `original` files generated by `sync-in`.
### Solution
Run `sync-in` first to generate `original` files for the restoration of redacted files during `sync-out`.

- JIRA ticket: [P20-739](https://codedotorg.atlassian.net/browse/P20-739)

### DIFF
--- a/bin/i18n/sync-all.rb
+++ b/bin/i18n/sync-all.rb
@@ -40,17 +40,17 @@ class I18nSync
 
   def run
     if @options[:interactive]
-      # download and distribute translations from the previous sync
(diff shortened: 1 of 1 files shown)


daynew (approved):
Looks good!

mgc1194 (approved):
LGTM ✨ 
We switch the order of sync steps back in June 2023 #52230.
It's been working fine, but in situations where we update the i18n dev server, and untracked files get removed from the repo, we need to run the sync in before we can execute the sync out.

PR #56791, opened 2024-02-23
Show multiple certificates on Unit Group congrats page

bethanyaconnor:
This ended up being a doozy of a PR! It does a few things:
- Checks progress before showing certificates for non-HoC courses. If a course name is not given or we have no curriculum associated with that course name, we show a certificate
  - note: I didn't want to increase the number of queries this route does for HoC courses, so the order of some of the conditionals is important. I added a test to enforce this.
- For the congrats page for a unit group:
  - If the user has completed all units, shows a certificate for the unit group
  - If the user has completed some but not all units, shows a certificate for each completed unit
  - If the user has completed no units, shows a page telling them to return to the curriculum
  - note: I did not handle the HoC case here as we have no HoC unit groups. 

Video for multiple certificates:

https://github.com/code-dot-org/code-dot-org/assets/46464143/a7ba699b-33d7-4b4b-82fd-73122ac28597

Message to users to complete units:

Screenshot 2024-02-29 at 3 26 50 PM

## Links

[figma](https://www.figma.com/file/sZ6d1uhJWFksJ6j3Fnmbnc/PL-Flow-Improvements?node-id=1958%3A10514&mode=dev)

## Testing story

- local testing
- relying on existing testing. There is significant existing UI test coverage

## Follow-up work

- [add arrows to the carousel](https://codedotorg.atlassian.net/browse/ACQ-1557)
- [updating alt text](https://codedotorg.atlassian.net/browse/ACQ-1554)
- [fixing component widths](https://codedotorg.atlassian.net/browse/ACQ-1553)

## PR Checklist:

- [ ] Tests provide adequate coverage
- [ ] Privacy and Security impacts have been assessed
- [ ] Code is well-commented
- [ ] New features are translatable or updates will not break translations
- [ ] Relevant documentation has been added or updated
- [ ] User impact is well-understood and desirable
- [ ] Pull Request is labeled appropriately
- [ ] Follow-up work items (including potential tech debt) are tracked and linked

PR #56816, opened 2024-02-26
Pass down course name for certificates and pass to Amplitude

bethanyaconnor:
This PR does a couple things:
- adds courseNames to the amplitude event for visiting the congrats page
  - I made this plural because I'm working on showing multiple certificates but I'm open to push back!
- Uses the `tutorial` slug to generate the image URL. We already do this when the certificate is personalized, so I didn't see the reason for the redundancy is decoding the URL param in both [the controller](https://github.com/code-dot-org/code-dot-org/blob/e4e4c475df0449dd304fa68d20cf87ee1289fd4e/dashboard/app/controllers/congrats_controller.rb#L12) and on the client
  - This did require updating some UI tests, which I decided to turn into eyes tests (we are interested in how things look). I'm open for feedback here -- I decided on eyes tests as the image srcs are now obfuscated URLs, but I believe they are stable so I could keep these as UI tests. 

Some successful UI test runs with this change:
- https://app.saucelabs.com/tests/a7812e8a92244b6f9de146de68001aa5#72
- https://app.saucelabs.com/tests/6b3f12ba4461486880fb191e6bc71eee#17
- https://app.saucelabs.com/tests/6e945d5ac60244388150496dfdb62244#24
- https://app.saucelabs.com/tests/adea0fa1403546288153f69559b2cc0c#54

### DIFF
--- a/apps/src/sites/studio/pages/congrats/index.js
+++ b/apps/src/sites/studio/pages/congrats/index.js
@@ -26,18 +26,15 @@ $(document).ready(function () {
   // Allows us to conditionally hide the promotional card for the Dance Party
   // Extras tutorial if we have problems during Hour of Code.
   const hideDancePartyFollowUp = congratsData.hide_dance_followup;
-  const certificateImageUrl = congratsData.certificate_image_url;
   const isHocTutorial = congratsData.is_hoc_tutorial;
   const isPlCourse = congratsData.is_pl_course;
   const isK5PlCourse = congratsData.is_k5_pl_course;
+  const courseName = congratsData.course_name || 'hourofcode';
 
   let certificateId = '';
-  let tutorial = '';
   try {
     const params = queryString.parse(window.location.search);
     certificateId = params['i'] && params['i'].replace(/[^a-z0-9_]/g, '');
-    const s = params['s'];
-    tutorial = s ? atob(s).replace(/[^A-Za-z0-9_\- ]/g, '') : 'hourofcode';
   } catch (e) {}
 
   const mcShareLink = tryGetLocalStorage('craftHeroShareLink', '');
@@ -47,13 +44,14 @@ $(document).ready(function () {
       isHocTutorial,
       isPlCourse,
       isK5PlCourse,
+      courseNames: [courseName],
     });
 
   ReactDOM.render(
     <Provider store={store}>
       <Congrats
         certificateId={certificateId}
-        tutorial={tutorial}
+        tutorial={courseName}
         userType={userType}
         under13={under13}
         language={language}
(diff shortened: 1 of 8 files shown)


apps/src/sites/studio/pages/congrats/index.js
```diff
@@ -47,21 +44,21 @@ $(document).ready(function () {
       isHocTutorial,
       isPlCourse,
       isK5PlCourse,
+      courseNames: [courseName],
```
bethanyaconnor, line 47: I made this plural as we'll pass down multiple course names when multiple certificates come in

dashboard/test/ui/features/hour_of_code/hour_of_code_finish.feature
```diff
@@ -17,20 +17,23 @@ Scenario: Completing Minecraft HoC should go to certificate page and generate a
   And I press "button:contains(Submit)" using jQuery
   And I wait to see element with ID "uitest-thanks"
 
+@eyes
```
kelbyhawn, line 20: Learning Q: is this how you create an Eyes test? 🙀 If yes, is it the only line you need to add before the rest of the test (`Scenario: Flappy…`), or are there other steps?

TurnerRiley: You need the `@eyes` and then the lines involving `When I open my eyes...`, `And I see no difference...`, and `And I close my eyes`

apps/src/sites/studio/pages/congrats/index.js
```diff
@@ -47,21 +44,21 @@ $(document).ready(function () {
...
     });
 
   ReactDOM.render(
     <Provider store={store}>
       <Congrats
         certificateId={certificateId}
-        tutorial={tutorial}
+        tutorial={courseName}
```
TurnerRiley, line 54: If it's possible that we pass down multiple course names, should this be `{[courseName]}` as well?

TurnerRiley (approved):
Looks great! I like the shift to eyes test to prioritize what we're actually looking for. Just had a quick question!

apps/src/sites/studio/pages/congrats/index.js
bethanyaconnor: yes, sort of! I'm actually changing it to be a list of objects (I need the course name and the course path), but I'll do that in a future PR (you can see the in progress work [here](https://github.com/code-dot-org/code-dot-org/pull/56791/files), but that PR is still in draft). But, I didn't want to add an Amplitude attribute then change it two days later.

dashboard/test/ui/features/hour_of_code/hour_of_code_finish.feature
kelbyhawn: Oooo cool, thanks!

PR #56820, opened 2024-02-26
Disabling the error debug pages on LevelBuilder

cat5inthecradle:
NOTE: This PR belongs to @unlox775-code-dot-org , I just recreated it after the LFS Change.

---

Summary: debug pages show up whenever there is an error on levelbuilder, e.g: https://levelbuilder-studio.code.org/not/a/real/page

Issue: These pages are information disclosure vulnerabilities, on a production system. Even though levelbuilder is an internal tool, it is still logged into regularly by [Code.org](http://code.org/) admins, and thus it houses sensitive data we need to protect.

The Wrinkle: at rare times when debugging something new on levelbuilder, we do want to see the verbose errors. See this Slack thread [here](https://codedotorg.slack.com/archives/CA3KCSGTD/p1566230447210900)

Proposed Solution: turn off debug on levelbuilder, but setup new conditional bypass using a secret cookie

Override the behavior conditionally, doing the rescue_from in ApplicationControl if a certain config variable is set, a custom one `config.local_requests_conditional = true`.

When that is set to a truthy value:
1. If `config.consider_all_requests_local = false` then this has no effect.
2. If  `config.consider_all_requests_local = true`, then it will do the rescue idea above.
3. When it does the rescue, it checks for a cookie value `local_request_override_token`
   - If that cookie is set, with the value  "aaa111" (replace with the actual secret value) then it lets the normal verbose error be printed as normal with config.consider_all_requests_local = true
   - Otherwise, it interrupts and gives the generic "Internal Server Error" message with status 500
   
If that variable is false, then consider_all_requests_local behaves as normal.

Then we just setup the initializer to set `local_requests_conditional` to true if we are in `levelbuilder_mode`

## Links

- jira ticket: [INF-1177](https://codedotorg.atlassian.net/browse/INF-1177)

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/config.yml.erb
+++ b/config.yml.erb
@@ -363,6 +363,9 @@ skip_seed_all:
 # false reduces memory usage, but significantly slows down dev server startup.
 cache_assets_in_memory:
 
+# Magic key to set the local_request_override_token to, to see debugging on levelbuilder
+local_request_override_token: !Secret
+
 # Encryption key required for decoding certain protected level sources.
 # Contributors should ask a Code.org engineer for this if needed.
 properties_encryption_key: !Secret
--- a/dashboard/app/controllers/application_controller.rb
+++ b/dashboard/app/controllers/application_controller.rb
@@ -12,6 +12,11 @@ class ApplicationController < ActionController::Base
 
   include Services::DatabaseConnections::ControllerFilter
 
+  # Apply custom error handling based on configuration
+  if Rails.application.config.consider_all_requests_local && Rails.application.config.local_requests_conditional
+    rescue_from StandardError, with: :conditional_error_handling
+  end
+
   # Prevent CSRF attacks by raising an exception.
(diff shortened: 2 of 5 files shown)


cat5inthecradle:
This PR is a continuation of #55653, which was mis-closed by the Git LFS migration (#55759).

### Previous Comments:


### Previous Reviews:

cat5inthecradle:
Added this to the next Dev Tools Working Group meeting, but I'm in favor of turning off error/debut output on levelbuilder until we discuss a way to conditionally enable it.

cat5inthecradle:
CodeAI's Slack workspace moved from codedotorg.slack.com to codeai.slack.com. Here are updated URL's for any Slack links in this PR:

- https://codeai.slack.com/archives/CA3KCSGTD/p1566230447210900

PR #51013, opened 2023-03-29
Increase Archive time period to 15 months and Deep Archive to 3 years

sureshc:
We deliver CloudFront Real-Time HTTP Access Logs to an S3 Bucket where they can be searched via AWS Athena. Switch from Transitioning these log files to the S3 Intelligent-Tiering `Archive Access` tier (3-5 hours to Restore) from 3 months to 15 months to align with how long CloudWatch Metrics are retained. And increase the length of time from 6 months to 3 years when Objects are moved to the `Deep Archive Access` tier.

Note that we want to discourage Athena searches across wide date ranges due to the Athena pricing structure (we pay per GB of files scanned during each query execution).

## Deployment strategy

Note the drift between the template and the current configuration of the bucket
image

[This Stack Change Set](https://us-east-1.console.aws.amazon.com/cloudformation/home?region=us-east-1#/stacks/changesets/changes?stackId=arn%3Aaws%3Acloudformation%3Aus-east-1%3A475661607190%3Astack%2FAccessLogs%2Fe0c55ee0-c88c-11eb-bf48-0a6e9e609c1f&changeSetId=arn%3Aaws%3Acloudformation%3Aus-east-1%3A475661607190%3AchangeSet%2FAccessLogs-1e3erpmvnda-h5een9csl2q%2Fee716a2e-8f67-4589-aef6-5c38e89ebf3a) would potentially modify or replace a wide set of resources
image

## PR Checklist:

- [ ] Tests provide adequate coverage
- [ ] Privacy and Security impacts have been assessed
- [ ] Code is well-commented
- [ ] New features are translatable or updates will not break translations
- [ ] Relevant documentation has been added or updated
- [ ] User impact is well-understood and desirable
- [ ] Pull Request is labeled appropriately
- [ ] Follow-up work items (including potential tech debt) are tracked and linked

PR #41119, opened 2021-06-15
CloudFront real-time access log processor

wjordan:
This PR provides a CloudFormation stack that exports CloudFront real-time access logs to Parquet files in an S3 bucket, supporting efficient queries from Athena or Redshift Spectrum.

## Links

CloudFront:
- https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/real-time-logs.html

Kinesis Firehose:
- https://docs.aws.amazon.com/firehose/latest/dev/basic-deliver.html
- https://docs.aws.amazon.com/firehose/latest/dev/data-transformation.html
- https://docs.aws.amazon.com/firehose/latest/dev/record-format-conversion.html

Jira:
- [`INF-381`](https://codedotorg.atlassian.net/browse/INF-381) ('Set up Real-time logs for CloudFront distributions')
- [`INF-423`](https://codedotorg.atlassian.net/browse/INF-423) ('Log Pegasus HTTP logs to AWS Athena')

## Testing story

Manually tested, via the following steps:
- Provision CloudFormation stack in dev account
- Attach the exported configuration to an existing CloudFront distribution, and generate some HTTP requests against it
- Query the access logs using Athena (via management console), e.g.: `SELECT * FROM cdo_access_logs.access_logs WHERE datehour > '2021/06/15' ORDER BY timestamp DESC LIMIT 10`
- Create a test Redshift cluster and query against it:
  - Associate the exported Redshift Spectrum IAM role
  - Manually create the external schema using [`CREATE EXTERNAL SCHEMA`](https://docs.aws.amazon.com/redshift/latest/dg/r_CREATE_EXTERNAL_SCHEMA.html) query, e.g., `CREATE EXTERNAL SCHEMA cdo_access_logs FROM DATA CATALOG DATABASE 'cdo_access_logs_dev' IAM_ROLE '[role_arn]';`
  - Query the access logs using the external schema, e.g.: `SELECT * FROM cdo_access_logs.access_logs WHERE datehour > '2021/06/15' LIMIT 10`

## Deployment strategy

- Manually create stack in production account
  - NOTE: Packaging and uploading the Lambda functions associated with the stack template requires the use of [`aws cloudformation package`](https://awscli.amazonaws.com/v2/documentation/api/latest/reference/cloudformation/package.html) command.
- Update CloudFront distributions to associate the exported realtime log configuration
- Monitor metrics to make sure everything is working properly

## Follow-up work

- (eventually) disable classic access logs once they're no longer needed

[...]

PR #56823, opened 2024-02-26
Increase Archive time period to 15 months and Deep Archive to 3 years

bethanyaconnor:
This PR is a revival of https://github.com/code-dot-org/code-dot-org/pull/51013 (https://github.com/code-dot-org/code-dot-org/commit/ec671512f846a9837286ec4e83a1c6d86920ca8d#diff-36e7a5950fff79c82d69deec73fb7be176df02a96173bc67bfd5f23cec94b44fL14 is the relevant commit). I checked in S3 and the current "Days until transition to Archive Access tier" is 475 and the "Days until transition to Deep Archive Access tier" is 730 (both matching what is in this PR but not what is in staging). I adjusts the latter to match production, instead of Suresh's original commit, but happy to revert that.

tl;dr I _think_ this brings this cloudformation template in line with what is currently configured on S3.

### Other background

[Slack thread for original issue](https://codedotorg.slack.com/archives/C0T0PNTM3/p1679605311308529)

[Slack thread for reviving the PR](https://codedotorg.slack.com/archives/C03CK49G9/p1708966831159949)

[Jira ticket](https://codedotorg.atlassian.net/browse/ACQ-1642)


Hamms (approved):
I see the same numbers in S3 that you saw, so I'd say I also _think_ that this brings the template in line with what we actually have.

image

It would ideally be nice to get confirmation from @cat5inthecradle that we did indeed successfully deploy the changes from the last PR but just didn't get it merged in time; I don't think that's blocking, however.

Thanks for cleaning up some of our loose threads!

sureshc:
My memory is that we tried to apply updates to this template and encountered issues that prevented the CloudFormation template from updating successfully due to drift between the actually provisioned resources and what was specified in the template or some other technical debt.

cat5inthecradle:
Regarding deployment, this template doesn't appear to be documented anywhere. It references two .rb files, which would need to be packaged with the template to deploy.

I propose we move these three files to "aws/cloudformation/standalone/access_logs" where we can also include a readme or bash script to explain how to run the `cloudformation package` and `cloudformation deploy`.  ([See the PR where this was added](https://github.com/code-dot-org/code-dot-org/pull/41119))

cat5inthecradle:
Running drift detection on the AccessLogs stack reveals what we expect. Only this setting seems to have drifted:
image

cat5inthecradle (approved):
Approving this, however, we should update our documentation with the correct deploy procedure.

The way we typically handle changes like this, that don't have automatic deployment, is to manually deploy the change, and only merge once it's successful. That way if we need to iterate or fix anything, we get it all into one PR.

aws/cloudformation/standalone/access_logs/access_logs.yml
```diff
@@ -132,7 +132,7 @@ Resources:
           DATABASE: !Ref AccessLogDatabase
           TABLE: !Ref AccessLogTable
           PREFIX: !Ref BucketPrefix
-      Runtime: ruby2.7
+      Runtime: ruby3.3
```
cat5inthecradle, line 136: We cannot deploy new Ruby 2.7 lambdas, so we need to update the runtime. I have not actually tested to see if the code is valid Ruby 3.3.

lib/cdo/aws/cloud_formation.rb
```diff
@@ -23,7 +23,7 @@ class CloudFormation
     # @return [Hash]
     attr_reader :options
 
-    # Default S3 bucket for uploading temporary template files.
+    # Default S3 bucket for uploading template files.
     TEMP_S3_BUCKET = 'cf-templates-p9nfb0gyyrpf-us-east-1'
```
cat5inthecradle, line 27: This is actually the permanent location of uploaded lambda code, it's inaccurate to refer to this as temporary storage. Deleting the contents would break things.

cat5inthecradle:
@bethanyaconnor Apologies for pushing to your branch, but I went ahead and added a deploy script and made the necessary Ruby runtime version upgrade.

I think we need to test somehow that these ruby scripts are valid with Ruby 3.3. I'm not totally sure the best way to do that. We really don't want them to break, but unfortunately they must be updated if we want to run a stack update here.

bethanyaconnor:
no worries on pushing to this branch -- this PR is mainly to tie up a 1.5 year loose end that ended up in our backlog 😅 Is this PR good to merge once I get a green Drone run? (the run failed due to the sauce connect upgrade, so hopefully just need to merge in staging)

cat5inthecradle:
We'll want to deploy before we merge this. I think there's a pretty good chance there's something we overlooked that will fail during the deploy, and it's nice to keep it all wrapped up in one PR.

Once the drone run is green, we can give it a try.

bethanyaconnor:
Drone is green!

cat5inthecradle:
Some improvements were made to this lambda stack last week, I've resolved those conflicts. I think we're ready to just give this a try.

We should deploy it manually per the local [README](https://github.com/code-dot-org/code-dot-org/blob/91e79aba0ab251959b241127f937e7ec232caabf/aws/cloudformation/standalone/access_logs/README.md) before merging. Resolve problems in this PR before merging.

cat5inthecradle:
Actually, we did this last week - the only effective change in this PR now is the ruby version upgrade - we can close this and do the upgrade in it's own PR.

PR #56830, opened 2024-02-27
Check if channel_id is already appended before appending it to WebLab footer links

ebeastlake:
Here's the diff using an external tool since Github won't show it for an LFS file:
Screenshot 2024-02-26 at 4 58 13 PM

We've got a long-standing and intermittent issue where the "View Code" button on the Web Lab footer sometimes generates invalid links with the channel ID and `/view` suffix appended twice.

My guess at the root cause: Given the setup with a `readystatechange` event listener and an asynchronous XHR request loading the footer content, I think there are multiple pathways for applyFooter to be called. If both these events happen at times when applyFooter’s conditions are met (i.e., both footerContents and ready are truthy), the function might run more than once, appending the channel_id and /view to the URLs twice.

However, as I have not been able to reproduce locally, this is a speculative fix. 

## Links

Slack thread: https://codedotorg.slack.com/archives/C03DBDN67B7/p1708992954857889?thread_ts=1708988274.961119&cid=C03DBDN67B7
Most recent Zendesk ticket: https://codeorg.zendesk.com/agent/tickets/486903

## Testing story

This is a speculative fix because I haven't repro'ed locally. I might be able to repro locally by forcing some of these conditions to be true. 

## PR Checklist:

- [x] Code is well-commented

### DIFF
--- a/dashboard/public/weblab/footer.js
+++ b/dashboard/public/weblab/footer.js
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:fa9dfbba08578f25e24a97535e3c950edde6b52d1e86e9bc66465215560f1c81
-size 1546
+oid sha256:8b35c465973cb975a3012c5c4fd991476f9cf821c389f68fc94d1f102853ddad
+size 1535
--- a/dashboard/public/weblab/footer.js
+++ b/dashboard/public/weblab/footer.js
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:77561bc0aaaff298c949abcbffd627bdb0de330b4995eac5792ea8665ee4a46f
-size 1140
+oid sha256:fa9dfbba08578f25e24a97535e3c950edde6b52d1e86e9bc66465215560f1c81
+size 1546


molly-moen (approved):
Nice!

bethanyaconnor (approved):
I don't think this file should be in LFS -- maybe a follow up? But the diff in the screenshot looks good!

ebeastlake:
@snickell How do we get a file removed from LFS, assuming this one is a candidate?

snickell:
So generally you can remove files from LFS by adding an exclusion pattern in .gitattributes. I should comment that file, the original was commented, but during the transition weekend I ended up letting a tool rewrite it and I never got the comments back.

This is an "include in LFS" .gitattributes pattern (specifically the one that puts dashboard/public/weblab/footer.js in LFS):
`dashboard/public/** filter=lfs diff=lfs merge=lfs -text`

This is an "exclude from LFS" .gitattributes pattern:
`pegasus/sites/**/*.html !text !filter !merge !diff`

Generally you want the excludes to come AFTER the includes.

The big issue is that removing from LFS is a one-way-journey: its really easy to add things to LFS, and same-as-last-time-hell to remove them from LFS.

Looking at `dashboard/public`, maybe we want to exclude `dashboard/public/**/*.html, dashboard/public/**/*.css, dashboard/public/**/*.css` ?

snickell:
Here's a PR that excludes ALL css, html and js files from LFS (these patterns were safe to apply repo-wide): https://github.com/code-dot-org/code-dot-org/pull/56891

PR #38888, opened 2021-02-03
Basic Java IDE Setup

molly-moen:
Add the plumbing for a java ide. The IDE will be available at `studio.code.org/java_ide` for admins and levelbuilders (this is a temporary url until we integrate more fully with code studio).
The ide currently consists of only an [Ace editor](https://ace.c9.io/#nav=about) in Java mode--more will be added as a follow up. The editor will copy its contents out to the console.

## Links

- [jira](https://codedotorg.atlassian.net/browse/CSA-66)

## Testing story
Tested locally that page is appropriately locked down and the editor is functional.

# Reviewer Checklist:

- [ ] Tests provide adequate coverage
- [ ] Privacy and Security impacts have been assessed
- [ ] Code is well-commented
- [ ] New features are translatable or updates will not break translations
- [ ] Relevant documentation has been added or updated
- [ ] User impact is well-understood and desirable
- [ ] Pull Request is labeled appropriately
- [ ] Follow-up work items (including potential tech debt) are tracked and linked

PR #56833, opened 2024-02-27
Removed unused packages ace-builds and react-ace

jmkulwik:

The ace-builds and react-ace packages were used in the early days of our Java Lab project, but were removed when we moved to codemirror. Here is the original PR: https://github.com/code-dot-org/code-dot-org/pull/38888


molly-moen (approved):
🧹

PR #50406, opened 2023-02-23
Add TypeScript support in apps/

maddiedierker:
Allows for incremental TypeScript adoption with .ts and .tsx files. Changing TypeScript files will also cause the frontend server to rebuild automatically.

The Student Learning team has proposed that we use TypeScript for the "Lab Lab" project (an effort to build a shared platform for new/existing labs). More details about the project can be found in [Lab Lab Tenets](https://docs.google.com/document/d/1M-tHYj7fcfmuVBWZKjVX0hx3-jQNua8OdzlxgFnpCF8/edit).

**Follow-up work:**
- We'll likely need to add some linting rules or extend existing ones to work for TypeScript, but that can come later.

PR #56835, opened 2024-02-27
Removed unused grunt-ts package

jmkulwik:

The grunt-ts package looks like it was used when we initially started using typescript, but if I'm understanding right, we moved the compile step to webpack instead of grunt. I could definitely use an extra set of eyes or two on this one. To test it out, I successfully ran `yarn build` after deleting my build directory. Happy to run other tests if it would be useful. Here's where the package was first added: https://github.com/code-dot-org/code-dot-org/pull/50406


snickell (approved):
Yup, this is done in webpack currently, not using this.

PR #51009, opened 2023-03-29
Use React Testing Library for Curriculum Catalog Card

megcrenshaw:
Adds React Testing Library to our repo! And tests the curriculum catalog card with it. Once the button functionality is implemented, the tests will be more interesting. For now, focused on ensuring the text is rendering correctly.

Some highlights:
- Caught a bug with aria labels not working appropriately. Below is the log when the test failed –– notice that accessibility is prioritized and that the entire component is rendered.
- There was no need to test the child component –– everything that needed a test was captured here.
- The documentation is quite helpful and prioritizes accessibility: https://testing-library.com/docs/queries/about#priority
- By default, unmounts the component from the container and destroys the container after each unit test. See https://testing-library.com/docs/svelte-testing-library/api/#cleanup
- The site https://testing-playground.com/ allowed me to copy and paste the (child) component into the browser and get queries back for the tests.
```
TestingLibraryElementError: Unable to find an accessible element with the role "button" and name `/Assign AI for Oceans to your classroom/`

Here are the accessible roles:

  img:

  Name "":
  <img
    alt=""
    src="/webpack_output/anotherhocwpaf86497935e32baa094324e08ebd98e5.png"
  />

  --------------------------------------------------
  heading:

  Name "AI for Oceans":
  <h4 />

  --------------------------------------------------
  button:

  Name "View details about undefined":
  <button
    aria-label="View details about undefined"
    class="RgEAWGSURVHXgsym55ZC HQMxS6q4VCbKRv45lt8t SpnpMKQTi1rIE2H6S2st pUlkLwlwitRrefPiTBSj"
    style="box-shadow: none;"
  />

  Name "Assign undefined to your classroom":
  <button
    aria-label="Assign undefined to your classroom"
    class="RgEAWGSURVHXgsym55ZC qW47iNnhv1vfNqsF8sJq SpnpMKQTi1rIE2H6S2st pUlkLwlwitRrefPiTBSj"
    style="box-shadow: none;"
  />

  --------------------------------------------------

Ignored nodes: comments, script, style
<body>
  <div
    id="codeApp"
  >
    <div
      id="app"
    />
  </div>
  
  
  



  <div>
    <div
      class="curriculumCatalogCardContainer"
    >
      <img
        alt=""
        src="/webpack_output/anotherhocwpaf86497935e32baa094324e08ebd98e5.png"
      />
      <div
        class="curriculumInfoContainer"
      >
        <p
          class="overline"
        >
          Math
        </p>
        <h4>
[...]

PR #56836, opened 2024-02-27
Removing unused jsdom and global-jsdom packages

jmkulwik:

The jsdom and global-jsdom packages look like they were planned to be used when we started using react-testing library, but from what I can tell, we never ended up using them. I could definitely use an extra set of eyes on this one, in case we actually are using them in a way I missed. Additionally, I'm curious if anyone is planning to use these in the near-term future. Here's where the packages were first added: https://github.com/code-dot-org/code-dot-org/pull/51009


jmkulwik:
Note: Passed the original drone run. Rerunning to resolve some later merge conflicts. https://drone.cdn-code.org/code-dot-org/code-dot-org/1059

lfryemason:
I found this documentation that suggests that `jsdom` and `global-jsdom` are required for running RTL in Mocha:
[docs](https://testing-library.com/docs/react-testing-library/setup/#using-without-jest).

The docs might be wrong since I would expect this PR to fail, but I would want further confirmation that this isn't required before merging

jmkulwik:
> I found this documentation that suggests that `jsdom` and `global-jsdom` are required for running RTL in Mocha: [docs](https://testing-library.com/docs/react-testing-library/setup/#using-without-jest).
> 
> The docs might be wrong since I would expect this PR to fail, but I would want further confirmation that this isn't required before merging

Good call. That's a super helpful link. I took a look at that as well as the [jsdom](https://www.npmjs.com/package/jsdom) docs and the [global-jsdom](https://www.npmjs.com/package/global-jsdom) docs. From what I'm interpreting, it seems that jsdom and global-jsdom aren't required for all uses of RTL in Mocha; however, they are required if you want the same functionality as if you had a Jest setup with the `testEnvironment` set to `'jest-environment-jsdom'`. It looks like the two packages are useful in tests that need direct access to the document object (which we definitely have done in the past). However, since none of our RTL tests have utilized jsdom yet, and we haven't set it up globally, we don't hit test failures by removing it. 

My thought: jsdom appears to be the current industry standard for manipulation of the document object within RTL tests. However, since we aren't using it yet, I don't think we should tie ourselves to it yet. My vote would be to remove it: perhaps by the time we end up needing that functionality, it'll be built directly into RTL, or we'll be using Jest, or the industry will have moved to a different standard.

snickell (approved):
👍, if we're worried mostly about whether this breaks testing, its hard to argue with the tests passing 🤣 Agree with Jess that we can always add the dep back if we decide we want to use it in 2 weeks or 2 years.

Also... even if we kept it, now knowing what its for, it should be / have been in devDependencies.

levadadenys (approved):
We have quite unique RTL usage case 😄 (rtl+reconfigured chai) , not surprised we might not need something from documentation.
+1 to delete these if it doesn't break anything.
Also +1 to adding these as a devDependencies if we need it sometime.


Also a sidenote - RTL suggests not to manipulate DOM directly and use it's own queries instead, meaning removing these might help us in the future (one will try to use RTL native queries/functionality first before installing jsdom 😄 )

Issue #56766, opened 2024-02-23
Can't use emojis in table names


snickell:
original: From Jessie: Random UI bug: I created a table with a couple columns. Then I created a new table with an invalid table name of “😍😂😋” that opened a preview of a new table with that name even though that table wasn’t actually created. Strangely, it also had the column headers from the previous table. On top of that, when I try to add a row to that table, it hangs. Then, when I switch to the original table, the row I added is there, still hanging.

Firebase lets you use emojis in the name, no problem. So there might be tables with names like this that we have to port over. 

The error we get is:

ActiveRecord::StatementInvalid in DatablockStorageController#create_table

Mysql2::Error: Illegal mix of collations (utf8_unicode_ci,IMPLICIT) and (utf8mb4_unicode_ci,COERCIBLE) for operation '='

app/controllers/datablock_storage_controller.rb:278:in `table_or_create'
app/controllers/datablock_storage_controller.rb:64:in `create_table'

snickell:
I believe this is the line that's failing:
```
  def where_table
    DatablockStorageTable.where(project_id: @project_id, table_name: params[:table_name])
  end
```

Table on the adhoc looks like:
![Image](https://github.com/code-dot-org/code-dot-org/assets/223277/97b39766-6be1-4a39-a416-2a06452ac8cf)

snickell:
Can repro with:
```
DatablockStorageTable.where(project_id: 5, table_name: "🎉").first_or_create

  DatablockStorageTable Load (0.7ms)  SELECT `datablock_storage_tables`.* FROM `datablock_storage_tables` WHERE `datablock_storage_tables`.`project_id` = 5 AND `datablock_storage_tables`.`table_name` = '🎉' ORDER BY `datablock_storage_tables`.`project_id` ASC, `datablock_storage_tables`.`table_name` ASC LIMIT 1
/Users/seth/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/mysql2-0.5.5/lib/mysql2/client.rb:151:in `_query': Mysql2::Error: Illegal mix of collations (utf8_unicode_ci,IMPLICIT) and (utf8mb4_unicode_ci,COERCIBLE) for operation '=' (ActiveRecord::StatementInvalid)
[...]

Issue #56761, opened 2024-02-23
Setting 👁️🫦👁️ as a key in kvp results in 500 from controller


snickell:
ActiveRecord::StatementInvalid in DatablockStorageController#set_key_value

Mysql2::Error: Incorrect string value: '\xF0\x9F\x91\x81\xEF\xB8...' for column 'key' at row 1

Extracted source (around line #151):

#149     def query(sql, options = {})
#150       Thread.handle_interrupt(::Mysql2::Util::TIMEOUT_ERROR_NEVER) do
*151         _query(sql, @query_options.merge(options))
#152       end
#153     end
#154 

Extracted source (around line #150):

#148 
#149     def query(sql, options = {})
*150       Thread.handle_interrupt(::Mysql2::Util::TIMEOUT_ERROR_NEVER) do
#151         _query(sql, @query_options.merge(options))
#152       end
#153     end

Rails.root: /home/ubuntu/adhoc/dashboard

Application Trace
config/initializers/mysql_check_index_used.rb:33:in `block (3 levels) in execute'
config/initializers/mysql_check_index_used.rb:32:in `block (2 levels) in execute'
config/initializers/mysql_check_index_used.rb:31:in `block in execute'
config/initializers/mysql_check_index_used.rb:30:in `execute'
app/models/datablock_storage_kvp.rb:23:in `set_kvps'
app/models/datablock_storage_kvp.rb:31:in `set_kvp'
app/controllers/datablock_storage_controller.rb:26:in `set_key_value'
app/controllers/application_controller.rb:186:in `block in with_locale'
app/controllers/application_controller.rb:185:in `with_locale'

snickell:
Fixed by [#56837](https://github.com/code-dot-org/code-dot-org/pull/56837)

PR #56837, opened 2024-02-27
Allow emojis in DatablockStorage key names and table names

snickell:
Our existing Firebase storage allows full JSON, which means full support for UTF-8 characters, including 4-byte characters like Emoji. To match Firebase, and allow import of existing data, we need to support the full 4-byte characters, but code.org's DBs default to using the mysql `utf8` type, which unfortunately is only 3-byte UTF-8 (UGH!). Thus we need to explicitly specify that our tables will be `utf8mb4`, and support emojis. That is all.

We also shortened the length for table_names and keys to 700 characters, see: https://github.com/code-dot-org/code-dot-org/issues/55344#issuecomment-1965686475 for details.

### DIFF
--- a/dashboard/app/models/datablock_storage_kvp.rb
+++ b/dashboard/app/models/datablock_storage_kvp.rb
@@ -3,7 +3,7 @@
 # Table name: datablock_storage_kvps
 #
 #  project_id :integer          not null, primary key
-#  key        :string(768)      not null, primary key
+#  key        :string(700)      not null, primary key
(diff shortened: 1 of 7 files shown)


snickell:
Fixes: https://github.com/code-dot-org/code-dot-org/issues/56766
and
Fixes: https://github.com/code-dot-org/code-dot-org/issues/56761

snickell:
It works!

image

PR #56737, opened 2024-02-22
Fix git diff and merge by switching .gitattributes strategy for LFS

snickell:
After the Git LFS migration #55759 , some git tools like `git diff` or merge were treating explicitly excluded-from-LFS files like `apps/i18n/common/en_us.json` as binary files. Other tools like VSCode diff or github's merge treated them like text files as expected. See issue #56501 for details.

This PR fixes the issues with git diff and merge, which now work as expected.

Fix details: It turns out `.gitattributes` has separate 'unspecified' and 'unset' states. The gitattributes pattern produced by `git lfs import` was using unset to remove lfs attributes like `merge=lfs`, but tools appear to diverge on interpreting files with unset diff, merge and filter attributes as binary vs text. In contrast, overriding git lfs attributes with `unspecified` appears to produce consistent "its a text file" behavior.

### How tested
I created two test branches off this one with the updated .gitattributes `seth/test-parent` and `seth/test-child`. In each branch, I made a different change to `apps/i18n/common/en_us.json`.

Diffing now works:
```
➜  code-dot-org git:(seth/test-child) git diff seth/test-parent
diff --git a/apps/i18n/common/en_us.json b/apps/i18n/common/en_us.json
index f79e33cb032..43b550cc650 100644
--- a/apps/i18n/common/en_us.json
+++ b/apps/i18n/common/en_us.json
@@ -2575,6 +2575,7 @@
   "text": "Text",
   "textToSpeech": "Text to speech",
   "textVariable": "text",
+  "other": "change",
   "textToSpeechProfanity": "Your text contains inappropriate language so it will not be played. Please update your project to remove the {profanityCount, plural, one {word} other {words}} \"{profaneWords}\".\n",
   "textToSpeechTruncation": "Your text exceeds our 750 character limit so we have shortened it for you.",
   "thanksForYourResponse": "Thanks for your response!",
@@ -2782,7 +2783,6 @@
   "yourEmailPlaceholder": "you@example.com",
   "yourName": "Your name",
   "yourNameCaps": "Your Name",
-  "boop": "snek",
   "yourResponseSubmitted": "Your response has been submitted!",
   "yourSchoolAdminButton": "View administrator guide",
   "yourSchoolAdminDesc": "View resources on how to get started or expand opportunities",
```

Merging now works:
```
➜  code-dot-org git:(seth/test-child) git merge seth/test-parent
Auto-merging apps/i18n/common/en_us.json
Merge made by the 'ort' strategy.
[...]

PR #56841, opened 2024-02-27
Exlude `dashboard/config/locales/en.yml` from Git LFS

artem-vavilov:
[`dashboard/config/locales/en.yml`](https://github.com/code-dot-org/code-dot-org/blob/staging/dashboard/config/locales/en.yml) is a system-important file which not so often updated manually by developers and is used as a source for the `i18n sync`, so it's important to track its changes.
- Reference: https://github.com/code-dot-org/code-dot-org/pull/56737

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,5 +1,6 @@
 apps/i18n/** filter=lfs diff=lfs merge=lfs -text
 dashboard/config/locales/** filter=lfs diff=lfs merge=lfs -text
+dashboard/config/locales/en.yml !text !filter !merge !diff
 i18n/locales/** filter=lfs diff=lfs merge=lfs -text
 apps/static/** filter=lfs diff=lfs merge=lfs -text
 dashboard/app/assets/** filter=lfs diff=lfs merge=lfs -text


.gitattributes
```diff
@@ -1,5 +1,6 @@
 apps/i18n/** filter=lfs diff=lfs merge=lfs -text
 dashboard/config/locales/** filter=lfs diff=lfs merge=lfs -text
+dashboard/config/locales/en.yml !text !filter !merge !diff
```
snickell, line 3: This /should/ be redundant with line 14 bellow, which sets the same attributes:
```
dashboard/config/locales/**/*en.yml !text !filter !merge !diff
```

But maybe something is making this not work like I expect?

snickell (commented):
@artem-vavilov what issues are you experiencing after merging https://github.com/code-dot-org/code-dot-org/pull/56737 into your branches? I would have expected that PR to fix `dashboard/config/locales/en.yml`, but it is totally possible something is happening. I just want to learn what!

There's two different (but related) issues here:
1. Is `dashboard/config/locales/en.yml` in Git LFS: **it is not in LFS**, you can verify by running `git lfs ls-files | grep dashboard/config/locales/en.yml`
2. Are the git attributes on `dashboard/config/locales/en.yml` set in a way that makes your tools and workflow work: **maybe not**? 

That said, I would not expect the line you added here to do anything, it should be redundant with line 14 below.

Are you sure the branches. you are still having issues with have #56737 merged into them?

.gitattributes
artem-vavilov: It's just something I wanted to do from the beginning of the LFS migration and didn't notice that we already excluded the file. Sorry about that!

snickell: I just tested without this new line, and after **#56737 I can do this command-line diffs and merges on `dashboard/config/locales/en.yml`**.

### Testing diffs on `dashboard/config/locales/en.yml`: works
```
git checkout staging
git checkout -b seth/test-en-yml-diffing
# Make a small change to the file
echo "  small_change: 'By seth'" >> dashboard/config/locales/en.yml
git diff staging
```

Which gives me this diff:
```
diff --git a/dashboard/config/locales/en.yml b/dashboard/config/locales/en.yml
index 708dbc66dc8..b728dc7acc4 100644
--- a/dashboard/config/locales/en.yml
+++ b/dashboard/config/locales/en.yml
@@ -1434,3 +1434,4 @@ en:
       success_body: "Great news! Your LMS has been registered with Code.org, and you can continue installing and using your LMS integration. You should receive a confirmation email shortly."
       success_header: "Thank you for registering your LMS!"
       register: "Register LMS"
+  small_change: 'By seth'
```

### Testing merges between diverging branches on `dashboard/config/locales/en.yml`: works

I also tested merges between two branches that change en.yml in different places as well, which also seem to be working:

1. Create branch off staging: `seth/change_en_yml`, and modify top of en.yml, `git diff staging` is:
```
diff --git a/dashboard/config/locales/en.yml b/dashboard/config/locales/en.yml
index 708dbc66dc8..b728dc7acc4 100644
--- a/dashboard/config/locales/en.yml
+++ b/dashboard/config/locales/en.yml
@@ -1434,3 +1434,4 @@ en:
       success_body: "Great news! Your LMS has been registered with Code.org, and you can continue installing and using your LMS integration. You should receive a confirmation email shortly."
       success_header: "Thank you for registering your LMS!"
       register: "Register LMS"
+  small_change: 'By seth'
```

2. Create branch off staging: `seth/change_en_yml_2` and modify bottom of en.yml, `git diff staging` is:
```
diff --git a/dashboard/config/locales/en.yml b/dashboard/config/locales/en.yml
index 708dbc66dc8..b8feed39788 100644
--- a/dashboard/config/locales/en.yml
+++ b/dashboard/config/locales/en.yml
@@ -31,6 +31,7 @@
 
 en:
   errors:
+    another: 'small change by seth'
     messages:
       not_saved: 'Please correct the following errors:'
       too_young: 'This content has age restrictions in place and is not available for younger students.'
```

4. Now try to successfully merge both changes:
```
git checkout seth/change_en_yml_2
git merge seth/change_en_yml
# Auto-merging dashboard/config/locales/en.yml
# Merge made by the 'ort' strategy.
# dashboard/config/locales/en.yml | 1 +
# 1 file changed, 1 insertion(+)
```

5. Finally, verify that both changes are included `git diff staging`:
```
diff --git a/dashboard/config/locales/en.yml b/dashboard/config/locales/en.yml
index 708dbc66dc8..b8baca1c837 100644
--- a/dashboard/config/locales/en.yml
+++ b/dashboard/config/locales/en.yml
@@ -31,6 +31,7 @@
 
 en:
   errors:
+    another: 'small change by seth'
     messages:
       not_saved: 'Please correct the following errors:'
       too_young: 'This content has age restrictions in place and is not available for younger students.'
@@ -1434,3 +1435,4 @@ en:
       success_body: "Great news! Your LMS has been registered with Code.org, and you can continue installing and using your LMS integration. You should receive a confirmation email shortly."
       success_header: "Thank you for registering your LMS!"
       register: "Register LMS"
+  small_change: 'By seth'
```

I should add that none of this worked before #56737, so if the branch you are trying to merge in doesn't have #56737 yet, this might not work until its merged in.

PR #56842, opened 2024-02-27
Remove JS after removing corresponding HAML

bethanyaconnor:
The `public/education/courses.haml` file was removed in the [staging scoop](https://github.com/code-dot-org/code-dot-org/commit/8fe1f1f1a513f319d73b6c5a1da8bc7e1cc9c142#diff-df1cb23efd931bae9ca6828f4055481c1754f0ea4303eb0db101140c33811f7e), without removing the corresponding JS file, which caused the build to fail. Unfortunately, Drone does not run on these scoops, so this is currently blocking all Drone runs.

This has happened in some fashion a few times in the past several years and is unfortunately a flaw in our pegasus / dropbox system. :(

## Testing story

`yarn build` succeeded locally after this change

### DIFF
--- a/apps/src/sites/code.org/pages/public/educate/curriculum/courses.js
+++ b/apps/src/sites/code.org/pages/public/educate/curriculum/courses.js
@@ -1,3 +1,0 @@
-import {initCourseExplorer} from '@cdo/apps/courseExplorer/courseExplorer';
-
-$(document).ready(initCourseExplorer);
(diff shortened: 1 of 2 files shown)


snickell (approved):
LGTM... I'm just amazed we actually detect (and unfortunately fail) on the entry-point not being used

snickell:
For anyone looking in the future, the error looks like this (from: https://drone.cdn-code.org/code-dot-org/code-dot-org/1069/1/5):
```
Running "lint-entry-points" task
--
55 | ✘ code.org/public/educate/curriculum/courses ➜ ./src/sites/code.org/pages/public/educate/curriculum/courses.js
56 | ✘ Entry points should only be used by one template but this one is used by 0!
57 | [179 passed] [34 silenced] [1 failed]
58 | Warning: 1 entry points do not conform to naming conventions.
59 | Run grunt check-entry-points for details.
60 |  Use --force to continue.
```

bethanyaconnor:
Thanks for posting the error message Seth!

Merging ahead of a full green Drone run as unit tests passed.

PR #56844, opened 2024-02-27
AI Tutor: delete AiTutorInteractions when a user is purged

Erin007:
[CT-250](https://codedotorg.atlassian.net/browse/CT-250)
[AI Tutor Product Spec ](https://docs.google.com/document/d/1rEWpIA6zXzevldiyi6DH-Pt4UU7uqPlWjTnUbz9uVLE/edit) req 4.7 

Deletes a user's chat messages with the AI Tutor when the user is purged from our system.  We'll still have a record in Amplitude for aggregate metrics regarding the relative uses of compilation, validation and general chat support; but we will no longer have database records that contain the content of the messages the user sent to and received from AI Tutor. This alleviates a privacy concern for the tool since the `AiTutorInteractions` table can contain PII. 

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/dashboard/test/helpers/delete_accounts_helper_test.rb
+++ b/dashboard/test/helpers/delete_accounts_helper_test.rb
@@ -1588,7 +1588,7 @@ class DeleteAccountsHelperTest < ActionView::TestCase
   # Table: dashboard.ai_tutor_interactions
   #
 
-  test "deletes all of a soft-deleted user's ai tutor interactions (chat messages)" do
+  test "deletes all of a purged user's ai tutor interactions (chat messages)" do
     student = create :student_with_ai_tutor_access
     num_ai_tutor_interactions = 3
     create_list :ai_tutor_interaction, num_ai_tutor_interactions, user: student
--- a/dashboard/test/helpers/delete_accounts_helper_test.rb
+++ b/dashboard/test/helpers/delete_accounts_helper_test.rb
@@ -1593,11 +1593,9 @@ class DeleteAccountsHelperTest < ActionView::TestCase
     num_ai_tutor_interactions = 3
     create_list :ai_tutor_interaction, num_ai_tutor_interactions, user: student
 
-    assert AiTutorInteraction.where(user_id: student.id).count, num_ai_tutor_interactions
-
-    student.destroy
-
-    assert AiTutorInteraction.where(user_id: student.id).count, 0
+    assert_changes -> {AiTutorInteraction.where(user: student).count}, from: num_ai_tutor_interactions, to: 0 do
+      student.destroy
+    end
   end
 
   #
--- a/dashboard/test/helpers/delete_accounts_helper_test.rb
+++ b/dashboard/test/helpers/delete_accounts_helper_test.rb
@@ -1585,6 +1585,24 @@ class DeleteAccountsHelperTest < ActionView::TestCase
   end
 
   #
+  # Table: dashboard.ai_tutor_interactions
+  #
+
+  test "deletes all of a soft-deleted user's ai tutor interactions (chat messages)" do
+    student = create :student_with_ai_tutor_access
+    num_ai_tutor_interactions = 3
+    num_ai_tutor_interactions.times do
+      create :ai_tutor_interaction, user: student
+    end
+
+    assert AiTutorInteraction.where(user_id: student.id).count, num_ai_tutor_interactions
+
+    student.destroy
+
+    assert AiTutorInteraction.where(user_id: student.id).count, 0
+  end
+
+  #
   # Table: dashboard.projects
   #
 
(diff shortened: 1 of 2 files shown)


molly-moen (approved):

lib/cdo/delete_accounts_helper.rb
```diff
@@ -308,6 +308,13 @@ def purge_teacher_feedbacks(user_id)
     @log.puts "Cleared #{as_student_count} TeacherFeedback" if as_student_count > 0
   end
 
+  def delete_ai_tutor_interactions(user_id)
+    chat_messages_to_delete = AiTutorInteraction.where(user_id: user_id)
+    count = chat_messages_to_delete.count
+    chat_messages_to_delete.destroy_all
```
artem-vavilov: It's safer to perform such operations `in batches`
```suggestion
    chat_messages_to_delete.in_batches.destroy_all
```

dashboard/test/helpers/delete_accounts_helper_test.rb
```diff
@@ -1584,6 +1584,24 @@ def get_record_with_sql(select_attribute, table_name, selector)
...
+    num_ai_tutor_interactions = 3
+    num_ai_tutor_interactions.times do
+      create :ai_tutor_interaction, user: student
+    end
+
+    assert AiTutorInteraction.where(user_id: student.id).count, num_ai_tutor_interactions
+
+    student.destroy
+
+    assert AiTutorInteraction.where(user_id: student.id).count, 0
```
artem-vavilov:
```suggestion
    assert_changes -> {AiTutorInteraction.where(user: student).count}, from: num_ai_tutor_interactions, to: 0 do
      student.destroy
    end
```

dashboard/test/helpers/delete_accounts_helper_test.rb
```diff
@@ -1584,6 +1584,24 @@ def get_record_with_sql(select_attribute, table_name, selector)
...
+  #
+  # Table: dashboard.ai_tutor_interactions
+  #
+
+  test "deletes all of a soft-deleted user's ai tutor interactions (chat messages)" do
+    student = create :student_with_ai_tutor_access
+    num_ai_tutor_interactions = 3
+    num_ai_tutor_interactions.times do
+      create :ai_tutor_interaction, user: student
+    end
```
artem-vavilov:
```suggestion
    create_list :ai_tutor_interaction, num_ai_tutor_interactions, user: student
```

artem-vavilov (approved):

dashboard/test/helpers/delete_accounts_helper_test.rb
```diff
@@ -1584,6 +1584,20 @@ def get_record_with_sql(select_attribute, table_name, selector)
...
+  # Table: dashboard.ai_tutor_interactions
+  #
+
+  test "deletes all of a soft-deleted user's ai tutor interactions (chat messages)" do
+    student = create :student_with_ai_tutor_access
+    num_ai_tutor_interactions = 3
+    create_list :ai_tutor_interaction, num_ai_tutor_interactions, user: student
+
+    assert_changes -> {AiTutorInteraction.where(user: student).count}, from: num_ai_tutor_interactions, to: 0 do
+      student.destroy
```
artem-vavilov, line 1597: I believe that `User#destroy` doesn't trigger the `DeleteAccountsHelper#purge_user` which calls `DeleteAccountsHelper#delete_ai_tutor_interactions`, so it won't work

artem-vavilov: I believe that `User#destroy` doesn't trigger `DeleteAccountsHelper#purge_user` which calls `DeleteAccountsHelper#delete_ai_tutor_interactions`, so it won't work
```suggestion
      purge_user(student)
```
https://github.com/code-dot-org/code-dot-org/blob/5e6b5c62d0975bcc54b4573091c1634c2148b441/dashboard/test/helpers/delete_accounts_helper_test.rb#L2185-L2202

daynew: @artem-vavilov is right that `purge_user` is only called when the user is hard-deleted after 30 days of being soft-deleted. @Erin007 I think there is a disconnect between the PR description and the tests written. Do you want the AI messages to be deleted when the user is soft-deleted or when the user is hard-deleted?

If you want the messages to be deleted when the user is hard-deleted, then the code in this PR is good, but the PR description should be updated to reflect that.

Erin007: Thanks for the clarification @daynew! I was missing that nuance. I updated the PR title and description as well as test name to reflect that we are indeed deleting the chat message records when the user is purged.

PR #56851, opened 2024-02-27
[Google Blockly] Allow hidden variable_set blocks

mikeharv:
Similar to:
https://github.com/code-dot-org/code-dot-org/pull/56487

@onlinecsteacher noticed that previously "invisible" variable blocks were showing up on workspaces in Course F: 
https://studio.code.org/s/coursef-2023/lessons/10/levels/4

https://codedotorg.slack.com/archives/C060J3A3BBK/p1709063154569829?thread_ts=1707508025.412189&cid=C060J3A3BBK

CDO Blockly didn't provide levelbuilders with a way to pre-populate variable dropdown fields. As a workaround, they would add invisible variables get/set blocks to the workspace which would add those variables to the project.
Since invisible blocks are not supported, these blocks are now showing up on the workspace in a confusing way.

image

The easiest solution is to allow them to be loaded onto the hidden workspace instead. We have a set of serialization hooks that can be added to any generic block that will support read the uservisible block attribute and then handle this state thereafter. After adding these hooks, the blocks are now hidden and the toolbox works as expected.

imageimage

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/blockly/addons/cdoVariables.js
+++ b/apps/src/blockly/addons/cdoVariables.js
@@ -1,3 +1,5 @@
+import {VARIABLE_BLOCK_TYPES} from '../constants';
+
 export default function initializeVariables(blocklyWrapper) {
   blocklyWrapper.Variables.DEFAULT_CATEGORY = 'Default';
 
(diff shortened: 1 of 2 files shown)


molly-moen (approved):
Nice!

PR #56858, opened 2024-02-27
[Dance AI] use queueRender when converting blocks

mikeharv:
Emma reported a high-severity bug where clicking the "Use Code" button in the Dance AI modal caused an error and resulted in blocks failing to render.

https://codedotorg.slack.com/archives/C05DK21DAHK/p1709055539887369

image

This seems to be related to an explicit `render()` call on some newly created blocks. Typically we don't need to handle rendering blocks directly like this, but it has worked so far. With Blockly v10, there's a new render management system that is designed to handle rendering blocks more efficiently. If we use `queueRender()` instead, it still triggers a render right away and makes things work again:

image

For more details on these two methods:
https://github.com/google/blockly/blob/bdc416f6912106535c9ba9f09421ad107e8cf847/core/block_svg.ts#L1592-L1611

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/dance/ai/DanceAiModal.tsx
+++ b/apps/src/dance/ai/DanceAiModal.tsx
@@ -538,7 +538,7 @@ const DanceAiModal: React.FunctionComponent<DanceAiModalProps> = ({
 
       blocksSvg.forEach(blockSvg => {
         blockSvg.initSvg();
-        blockSvg.render();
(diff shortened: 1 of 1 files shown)


molly-moen (approved):
🎉

breville (approved):

PR #56863, opened 2024-02-28
Levelbuilder cannot view as peer reviewer on non-javalab level

davidsbailey:
## background

@cearachew noticed that the teacher panel sometimes does not show up on localhost, and @Nokondi  and I have also been able to repro. The root cause of the issue is as follows:
* `shouldRenderTeacherPanel` is `false` here: https://github.com/code-dot-org/code-dot-org/blob/64661ce07b845d8de18b2286a414d29899a64f66/apps/src/sites/studio/pages/levels/_teacher_panel.js#L44-L46
* because `is_code_reviewing` is set here: https://github.com/code-dot-org/code-dot-org/blob/49cd73cbf430a43ba21b23faecf015c84a12e480/dashboard/app/controllers/script_levels_controller.rb#L424-L426
* we try to limit this permission to javalab levels here: https://github.com/code-dot-org/code-dot-org/blob/3f187477d45a6245a45f76a9696ae000881c41b7/dashboard/app/models/ability.rb#L166-L174
* however, for levelbuilders, this gets overridden here: https://github.com/code-dot-org/code-dot-org/blob/3f187477d45a6245a45f76a9696ae000881c41b7/dashboard/app/models/ability.rb#L389-L410

For additional context on how ability.rb works, see: https://github.com/CanCanCommunity/cancancan/blob/develop/docs/combine_abilities.md#abilities-precedence

> An ability rule will override a previous one.

## description

the two changes needed to fix this issue are:
1. move the definition of the `view_as_user_for_code_review` rule to after levelbuilder permissions are assigned
2. remove `view_as_user_for_code_review` permissions from everyone before assigning `view_as_user_for_code_review` permissions to anyone

## Testing story

* extensive existing test coverage of `view_as_user_for_code_review` in ability_test.rb
* new unit test covering the levelbuilder case
* verified new unit test fails without the code change and passes with it

### DIFF
--- a/dashboard/app/models/ability.rb
+++ b/dashboard/app/models/ability.rb
@@ -157,39 +157,6 @@ class Ability
         can?(:manage, section) || user.sections_as_student.include?(section)
       end
 
-      can :view_as_user, ScriptLevel do |script_level, user_to_assume, sublevel_to_view|
-        user.project_validator? ||
-          user_to_assume.student_of?(user) ||
-          can?(:view_as_user_for_code_review, script_level, user_to_assume, sublevel_to_view)
-      end
-
-      can :view_as_user_for_code_review, ScriptLevel do |script_level, user_to_assume, level_to_view|
-        can_view_as_user_for_code_review = false
-
-        level_to_view ||= script_level&.oldest_active_level
-
-        # Only allow a student to view another student's project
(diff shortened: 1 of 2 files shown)


cearachew (approved):
Excellent find for this weird problem!!

PR #56866, opened 2024-02-28
P20-714: Prevent uploading the `animations` localized manifest during sync-out when testing

artem-vavilov:
The AWS S3 instance, to which we upload the `animations` localized manifest, is used in production.
So we must prevent the uploading of any testing data to it.

- jira ticket: [P20-714](https://codedotorg.atlassian.net/browse/P20-714)

### DIFF
--- a/bin/i18n/i18n_script_utils.rb
+++ b/bin/i18n/i18n_script_utils.rb
@@ -28,6 +28,24 @@ class I18nScriptUtils
     @crowdin_creds ||= YAML.load_file(CROWDIN_CREDS_PATH).freeze
   end
 
+  # Parses sync options from the command line
+  #
+  # @param options [Hash] the default options to populate
+  # @return [Hash] the parsed options
+  def self.parse_options(argv = ARGV, options: {})
+    options[:testing] = TESTING_BY_DEFAULT if options[:testing].nil?
+
+    OptionParser.new do |parser|
+      parser.on('-t', '--testing', 'Run in testing mode') do
+        options[:testing] = true
+      end
+
+      yield(parser, options) if block_given?
+    end.parse!(argv)
+
+    options
+  end
+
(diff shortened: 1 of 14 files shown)


bin/i18n/sync-all.rb
```diff
@@ -98,11 +98,12 @@ def sync_out
   end
```
mgc1194, line 98: Should we switch to this?
```suggestion
  def sync_out
    I18n::SyncOut.perform(options.slice(:testing))
  end
```

bin/i18n/sync-all.rb
```diff
@@ -130,12 +131,7 @@ def parse_options(args)
       opts.on("-y", "--yes", "Run without confirmation") do
         options[:yes] = true
       end
```
mgc1194, line 133: I wonder if we should move the rest of the options to the one parser. On one hand, having one parsing method could be cleaner, on the other hand, most options are only used in the `sync-all` script, and could get confusing to have options available that are not used in other modules.

artem-vavilov: We shouldn't add options that we are not going to use in the scripts

bin/i18n/i18n_script_utils.rb
```diff
@@ -28,6 +28,24 @@ def self.crowdin_creds
...
+      parser.on('-t', '--testing', 'Run in testing mode') do
+        options[:testing] = true
+      end
+
+      yield(parser, options) if block_given?
+    end.parse!(argv)
+
+    options
+  end
+
```
mgc1194, line 48: When I read the description of the PR moving the `option parser` to the `i18n_script_utils` was my first thought. I'm really glad we were thinking the same way!

mgc1194 (approved):
LGTM!!

PR #56879, opened 2024-02-28
Initialize and configure Statsig

nicklathe:
This PR adds configuration for [Statsig](https://statsig.com/). It adds the following:

- Create a module to init Statsig
- Initialize Statstig in AppServerHooks
- CDO variables for secret and client keys for environments

Statsig should run in `local_mode` for any environment that is not `production` or `test` (managed web server). This mode disables any network calls.

The following secrets have **NOT** been created yet, but will be run **before** this PR is merged.

`secrets.yml`
```
---
statsig_api_client_key:
  production:
    statsig_api_client_key: <prod-client-key>
  test:
    statsig_api_client_key: <non-prod-client-key>
statsig_server_secret_key:
  production:
    statsig_server_secret_key: <prod-server-key>
  test:
    statsig_server_secret_key: <non-prod-server-key>

```
NOTE: For environments where Statsig will run in local mode, the Ruby SDK still expects a non-empty server key, with `server-` prefixed to the key string. [See code](https://github.com/statsig-io/ruby-sdk/blob/5b373ab9f2555d18bcb957b2ff70a4b0ad3a931f/lib/statsig_driver.rb#L20). The Javascript SDK [does not](https://github.com/statsig-io/js-client/blob/706d359f588f01719448269c26d70a851666591f/src/StatsigClient.ts#L241) require this for the client key if the SDK is initialized in local mode, however for consistency, it was added to the `<environment>.yml.erb` files in the dummy keys. I'd like to open a PR against the Ruby SDK to change this behavior (should not require a key if in local_mode), however I don't want to block on this PR.

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/config.yml.erb
+++ b/config.yml.erb
@@ -631,6 +631,11 @@ openai_evaluate_rubric_org_id:
 # AI Proxy for OpenAI API
 ai_proxy_origin: https://aiproxy-test.code.org
 
-# Statsig keys
-statsig_server_secret_key: !Secret
-statsig_api_client_key: !Secret
+# Statsig API keys
+# These are dummy values for all environments except Production and our managed
+# Test web application server, which are overridden in production.yml.erb
+# and test.yml.erb. The Statsig SDK expects the prefixed key type (secret-) in
+# the string, otherwise it will throw an error on initialization, even in
+# local_mode.
+statsig_server_secret_key: 'secret-fake-key'
+statsig_api_client_key: 'client-fake-key'
--- a/config.yml.erb
+++ b/config.yml.erb
@@ -630,3 +630,7 @@ openai_evaluate_rubric_org_id:
 
 # AI Proxy for OpenAI API
 ai_proxy_origin: https://aiproxy-test.code.org
+
+# Statsig keys
+statsig_server_secret_key: !Secret
+statsig_api_client_key: !Secret
--- a/dashboard/config/initializers/statsig.rb
+++ b/dashboard/config/initializers/statsig.rb
@@ -1,5 +1,8 @@
 require 'statsig'
 
-local_mode = CDO.rack_env?(:production) || CDO.rack_env?(:staging) ? false : true
-options = StatsigOptions.new({'tier' => CDO.rack_env}, network_timeout: 5, local_mode: local_mode)
-Statsig.initialize(CDO.statsig_server_secret_key, options)
+# Statsig is initialized here for the development environment. In managed
+# environments, it is initialized in config/puma.rb
+if CDO.rack_env?(:development)
+  options = StatsigOptions.new({'tier' => :development}, network_timeout: 5, local_mode: true)
+  Statsig.initialize(CDO.statsig_server_secret_key, options)
+end
--- a/dashboard/config/initializers/statsig.rb
+++ b/dashboard/config/initializers/statsig.rb
@@ -1,0 +1,5 @@
+require 'statsig'
+
+local_mode = CDO.rack_env?(:production) || CDO.rack_env?(:staging) ? false : true
+options = StatsigOptions.new({'tier' => CDO.rack_env}, network_timeout: 5, local_mode: local_mode)
+Statsig.initialize(CDO.statsig_server_secret_key, options)
--- a/dashboard/config/initializers/statsig.rb
+++ b/dashboard/config/initializers/statsig.rb
@@ -1,8 +1,6 @@
-require 'statsig'
-
+require "cdo/statsig"
 # Statsig is initialized here for the development environment. In managed
-# environments, it is initialized in config/puma.rb
+# environments, it is initialized in lib/cdo/app_server_hooks
 if CDO.rack_env?(:development)
-  options = StatsigOptions.new({'tier' => :development}, network_timeout: 5, local_mode: true)
-  Statsig.initialize(CDO.statsig_server_secret_key, options)
+  Cdo::StatsigInitializer.init
 end
--- a/dashboard/config/puma.rb
+++ b/dashboard/config/puma.rb
@@ -26,7 +26,18 @@ before_fork do
   Cdo::AppServerHooks.before_fork
 end
 
+# Statsig is initialized here for managed environments. For development, it is
+# intialized in config/initializers/statsig.rb
+require 'statsig'
+# Determine if this is a managed test environment
+managed_test_environment = CDO.running_web_application? && CDO.test_system?
+# Enable local_mode for all environments except prod and test. This limits the
+# amount of metrics we emit, thereby lowering our credit usage.
+local_mode = CDO.rack_env?(:production) || managed_test_environment ? false : true
+options = StatsigOptions.new({'tier' => CDO.rack_env}, network_timeout: 5, local_mode: local_mode)
+
 on_worker_boot do |_index|
+  Statsig.initialize(CDO.statsig_server_secret_key, options)
   Cdo::AppServerHooks.after_fork(host: CDO.dashboard_hostname)
   ActiveRecord::Base.establish_connection
 end
--- a/dashboard/config/puma.rb
+++ b/dashboard/config/puma.rb
@@ -26,18 +26,7 @@ before_fork do
   Cdo::AppServerHooks.before_fork
 end
 
-# Statsig is initialized here for managed environments. For development, it is
-# intialized in config/initializers/statsig.rb
-require 'statsig'
-# Determine if this is a managed test environment
-managed_test_environment = CDO.running_web_application? && CDO.test_system?
-# Enable local_mode for all environments except prod and test. This limits the
-# amount of metrics we emit, thereby lowering our credit usage.
-local_mode = CDO.rack_env?(:production) || managed_test_environment ? false : true
-options = StatsigOptions.new({'tier' => CDO.rack_env}, network_timeout: 5, local_mode: local_mode)
-
 on_worker_boot do |_index|
-  Statsig.initialize(CDO.statsig_server_secret_key, options)
   Cdo::AppServerHooks.after_fork(host: CDO.dashboard_hostname)
   ActiveRecord::Base.establish_connection
 end
--- a/Gemfile
+++ b/Gemfile
@@ -354,3 +354,5 @@ gem 'crowdin-api', '~> 1.8.1'
 gem "delayed_job_active_record", "~> 4.1"
 
 gem 'rack-cors', '~> 2.0.1'
+
+gem 'statsig', '~> 1.33'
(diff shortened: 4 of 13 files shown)


cat5inthecradle:
Why the choice to use local mode in Test and live mode in Staging?

nicklathe:
> Why the choice to use local mode in Test and live mode in Staging?

I'm actually open to suggestions here. I'm not sure what environments we actually log events from today (in Amplitude). Should we enable this for `test` or disable for everything `non-prod`? I'm not sure what pricing plan we have for Statsig, figured it would be good to limit our requests where it made sense.

cat5inthecradle:
The more prod-like that we can make Test, the more likely we are to discover problems. That said, for something like StatSig, it might not be that risky.

We should find and document the newer to the pricing question. Also, does data sent to StatSig have the appropriate metadata to distinguish different environments?

dashboard/config/initializers/statsig.rb
```diff
@@ -0,0 +1,5 @@
+require 'statsig'
+
+local_mode = CDO.rack_env?(:production) || CDO.rack_env?(:staging) ? false : true
```
nicklathe: Should we log events from test as well or just production?

nicklathe:
> The more prod-like that we can make Test, the more likely we are to discover problems. That said, for something like StatSig, it might not be that risky.
> 
> We should find and document the newer to the pricing question. Also, does data sent to StatSig have the appropriate metadata to distinguish different environments?

To your second question, yes Statsig can contain metadata to distinguish different envs. I'm configuring that in the initializer via the `{tier => CDO.rack_env}`. You can configure environments in the Statsig settings page. See [docs](https://docs.statsig.com/guides/using-environments#configuring-environments).

cat5inthecradle:
Slack Discussion: https://codedotorg.slack.com/archives/C03CK49G9/p1709149118589539

dashboard/config/initializers/statsig.rb
```diff
@@ -0,0 +1,5 @@
...
+options = StatsigOptions.new({'tier' => CDO.rack_env}, network_timeout: 5, local_mode: local_mode)
+Statsig.initialize(CDO.statsig_server_secret_key, options)
```
sureshc: In our managed environments (adhocs, staging, test, levelbuilder, production), we use the `puma` Rack web application server to process multiple HTTP requests simultaneously across multiple child puma Unix processes, and multiple threads per child puma process. It looks like [there are special initialization](https://docs.statsig.com/server/rubySDK) instructions for StatSig when using `puma` to ensure it is initialized in each child `puma` process when the child process is forked / boots up:

``` ruby
on_worker_boot do
  Statsig.initialize('server-secret-key', options)
end
```

Here’s where we implement worker boot initialization logic: https://github.com/code-dot-org/code-dot-org/blob/92a89e8d1d94cdd6eda58543965e26e2f250c467/dashboard/config/puma.rb#L29-L32

And if we want StatSig on Pegasus as well, we could put that in this shared module https://github.com/code-dot-org/code-dot-org/blob/92a89e8d1d94cdd6eda58543965e26e2f250c467/lib/cdo/app_server_hooks.rb#L29-L32

nicklathe: Thanks @sureshc, I'll move the initialization over to Puma. Should I go ahead and initialize it for Pegasus while I'm at it, that way it's available for anyone to use?

nicklathe: Thinking through this, it seems like we should initialize in puma.rb, and then conditionally as an initializer if it's not a managed environment?

Statsig is [idempotent in it's initialization](https://github.com/statsig-io/ruby-sdk/blob/main/lib/statsig.rb#L14) (will only initialize if not already running). I'm not sure of the order of Rails bootstrapping. Does puma get start _before_ the initializers? Based on the [Rails guide](https://guides.rubyonrails.org/v6.1/initialization.html), it seems initializers get called first, so we would need to start Statsig in the initializers _only_ for non-managed environments, otherwise it wouldn't ever get initialized in Puma, since it would already be running.

nicklathe:
Since Statsig batches events to be dispatched to the API, it is recommended you [tell Statsig to shutdown](https://docs.statsig.com/server/rubySDK#shutting-statsig-down) when the app server is shutting down, so it can flush/send any unsent metrics. Are there any existing shutdown hooks in our code we could tie into?

config.yml.erb
```diff
@@ -630,3 +630,7 @@ openai_evaluate_rubric_org_id:
 
 # AI Proxy for OpenAI API
 ai_proxy_origin: https://aiproxy-test.code.org
+
+# Statsig keys
```
nicklathe: Will Pegasus pick up `production` keys? Or is there an additional configuration I need to add?

config.yml.erb
```diff
@@ -630,3 +630,7 @@ openai_evaluate_rubric_org_id:
...
+statsig_server_secret_key: !Secret
```
sureshc: We want to use StatSig `local_mode` on ALL environments except our managed test server (`test.code.org / test-studio.code.org`) and production. It’s probably simpler to declare the fake secrets in the parent `config.yml.erb` file and then override in just the `test.yml.erb` and `production.yml.erb`.

Less important … to help others recognize that the key/secret are fake for testing, I suggest using values that clearly indicate that, like we do in similar cases https://github.com/code-dot-org/code-dot-org/blob/ed073c68cd44c34023f05327d11f17b2f07d39ce/config/development.yml.erb#L40-L41

nicklathe: Ah yes, refactoring this to default to fake secrets in config.yml.erb and override in test and production makes sense. Will make this change.

I added a comment above the fake secret sections in each config file. One of the issues is that Statsig expects the secret to be prefixed with `secret-`. I could do something like `secret-fake-key`. I've tested this locally and Statsig doesn't error on init with this fake key.

dashboard/config/puma.rb
```diff
@@ -26,7 +26,18 @@
   Cdo::AppServerHooks.before_fork
 end
 
+# Statsig is initialized here for managed environments. For development, it is
+# intialized in config/initializers/statsig.rb
+require 'statsig'
+# Determine if this is a managed test environment
+managed_test_environment = CDO.running_web_application? && CDO.test_system?
+# Enable local_mode for all environments except prod and test. This limits the
```
sureshc: Probably worth being very specific in the comments about what we mean when we say “test”. What we mean is that StatSig will be enabled on the managed test server’s web application server. We should be clear that unit tests, ActiveJobs, cron jobs, and the interactive ruby console (`dashboard-console`, `pegasus-console`) that execute on the managed test server will use StatSig in stubbed (local mode) and also that all code that executes in continuous integration builds (“Drone”) will use StatSig local mode.

config.yml.erb
sureshc: Yes. The `CDO` singleton is initialized by both Pegasus and Dashboard web application servers (and also any code the requires [deployment.rb](https://github.com/code-dot-org/code-dot-org/blob/b6f7c574e799d54afdfe8dd633d01f9308dc3f5a/deployment.rb#L10) such as cron jobs).

config.yml.erb
sureshc: In `test.yml.erb`, we’ll need something like this before [the section dedicated to configuration settings](https://github.com/code-dot-org/code-dot-org/blob/b6f7c574e799d54afdfe8dd633d01f9308dc3f5a/config/test.yml.erb#L13-L29) for continuous integration builds.

``` erb
# The only type of test environment we enable StatSig is the managed test server (`test.code.org / test-studio.code.org`)
# and only in Ruby code executing within the web application server.
<% if running_web_application? && test_system? -%>
statsig_server_secret_key: !Secret
statsig_api_client_key: !Secret
<% else %>
statsig_server_secret_key: 'secret-fake-key'
statsig_api_client_key: 'fclient-fake-key'
<% end -%>
```

sureshc: > I added a comment above the fake secret sections in each config file. One of the issues is that Statsig expects the secret to be prefixed with `secret-`. I could do something like `secret-fake-key`. I've tested this locally and Statsig doesn't error on init with this fake key.

Apologies - you took the time to document the prefix requirement in the Pull Request, and I didn’t read closely! 😞

dashboard/config/puma.rb
```diff
@@ -26,7 +26,18 @@
...
+# amount of metrics we emit, thereby lowering our credit usage.
+local_mode = CDO.rack_env?(:production) || managed_test_environment ? false : true
+options = StatsigOptions.new({'tier' => CDO.rack_env}, network_timeout: 5, local_mode: local_mode)
+
 on_worker_boot do |_index|
+  Statsig.initialize(CDO.statsig_server_secret_key, options)
```
sureshc: Do we need to initialize `Statsig` here if we also doing it within `Cdo::AppServerHooks`?

nicklathe: Ah you're right, I'm just now seeing the `Cdo::AppserverHooks.after_fork` get's called in puma. I'll go ahead and remove this explicit initialization in puma, and keep it in `AppserverHooks`.

sureshc:
We’re currently initializing `Statsig` in 3 different places. One case (dashboard puma) is likely duplicate logic that could be eliminated, but I think it’s worth putting all the initializing logic into `/lib/cdo/statsig.rb` and then calling that method once from the dashboard initializer and once from `Cdo::AppServerHooks`. That way we only have to tweak things like the connection timeout in one place and all of the conditional logic (development vs test vs managed test server vs production etc) is in one place.

sureshc:
> Since Statsig batches events to be dispatched to the API, it is recommended you [tell Statsig to shutdown](https://docs.statsig.com/server/rubySDK#shutting-statsig-down) when the app server is shutting down, so it can flush/send any unsent metrics. Are there any existing shutdown hooks in our code we could tie into?

Our Firehose and CloudWatch wrappers use `Cdo::Buffer` which solves this problem. `Cdo::Buffer` seems to register a kernel `at_exit` hook. I don’t know enough about best practices with using that Kernel method, and think it’s fine to initially launch with a system where we lose a few events that were queued up to be published to Statsig. https://github.com/code-dot-org/code-dot-org/blob/e621f0ad155a47a152a180eaee4dd72d5077c39a/lib/cdo/buffer.rb#L45

nicklathe:
@sureshc, thanks so much for all your feedback! I've made the appropriate changes, let me know if there's anything else that could be improved. Thanks!

sureshc (approved):

nicklathe:
Secrets have been created in Secrets Manager (pre-req to merging this PR). Merging PR now.

PR #56882, opened 2024-02-28
Webpack hot module replacement (HMR)

snickell:
# Enable React "Fast Refresh" aka HMR
Reloads react components as soon as they build, without having to manually reload the page, works with both JSX and TSX.

## Demo
https://github.com/code-dot-org/code-dot-org/assets/223277/8beebf42-ca4c-487a-83d8-8b2876507f04

## What's Changed
1. Uses [react-refresh](https://www.npmjs.com/package/react-refresh), [@pmmmwh/react-refresh-webpack-plugin](https://github.com/pmmmwh/react-refresh-webpack-plugin), and [webpack-dev-server](https://www.npmjs.com/package/webpack-dev-server)
2. `yarn start` now launches [webpack-dev-server](https://webpack.js.org/configuration/dev-server/), which serves JS requests from memory and includes hooks to update the browser when source changes. It additionally proxies requests from port 9000 to port 3000 to provide a full dashboard server.
3. Prints new Dashboard URL at the end of `yarn start`: to use HMR, Dashboard must be accessed on port 9000: http://localhost-studio.code.org:9000. 
    image
4. Removes createWebpackConfig() watch argument/mode, replace with `writeToDisk: true`, so those wishing to bypass HMR can still access pages directly on http://localhost-studio.code.org:3000. Also removes WebpackNotifierPlugin, as watch mode is now gone (and didn't work anyway).
5. Removes LiveReloadPlugin and envConstants.AUTO_RELOAD, which didn't seem to be working anyway, and are outdated approaches compared to HMR.
6. Removes `react-hot-loader`, another dependency previously used for hot reloading, but also wasn't working currently and has been deprecated by the maintainers.
7. Supports TSX by adding react-refresh-typescript plugin to ts-loader
8. Supports JSX by adding react-refresh/babel plugin to babel-loader
9. Renames `cdoBlockStyles.mjs` to `cdoBlockStyles.js`, converts to CJS style module, and updates references. This was required because `"type": "module"` (which is implied by the .mjs prefix) in the browser triggers issues with `react-refresh`. Since we only include this one .mjs file in the browser (and here only because the file is shared between node and browser), it was easy to rename to workaround.
10. Moves babel-cache from .babel-cache into the build dir, now is: build/babel-cache, so it gets cleared by yarn clean, etc
11. Simplify `createWebpackConfig()` calls using default values, emphasizes what makes targets different in Gruntfile.js
12. When running `grunt dev`, sets HOT=1 and DEV=1 in Gruntfile.js, removing repetition from package.json
13. Creates a `yarn grunt` target that sets `NODE_OPTIONS=--max-old-space-size=4096`, and use to remove 5x repetition from package.json, as well as no longer requiring explicit paths into node_modules, e.g.
    - Before: `"start": "DEV=1 node --max_old_space_size=4096 ./node_modules/.bin/grunt dev;"`
    - After: `"start": "yarn grunt dev"`
    - Before: `"build": "DEV=1 node --max_old_space_size=4096 ./node_modules/.bin/grunt build"`
    - After: `"build": "DEV=1 yarn grunt build"`

## Testing

1. JSX hot replacement:
    - Load: http://localhost-studio.code.org:9000/s/allthethings/lessons/18/levels/20
    - modify `GameButtons.jsx`, change `<div>{props.runButtonText || msg.runProgram()}</div>` to `<div>FunButton</div>`
    - Verify [Run] button changes to [Fun Button] without a reload
    - Verify watch mode works too, load: http://localhost-studio.code.org:3000/s/allthethings/lessons/18/levels/20 (note: port 3000), and verify that [Run] button has changed to [Fun Button] there too
2. TSX hot replacement:
    - Load: http://localhost-studio.code.org:9000/s/dance-ai-2023/lessons/1/levels/3
    - modify `SongSelector.tsx`, change `fa-play-pause` to `fa-play`
    - Verify the play/pause icon changes to a play icon without a reload
3. Storybook hot replacement:
    - Run `yarn storybook`, and open the Checkbox design component
    - Edit `Checkbox.tsx` and add text right before the `<input`
    - Verify that text appears in storybook without a reload
5. Verify `yarn build:dist` works and does NOT include HMR hooks
6. Verify `yarn build` works and does NOT include HMR hooks

Co-authored-by: Ben Brooks <ben@code.org>


snickell:
Added a FIXME about `skipEnvCheck` having to be set here: https://github.com/code-dot-org/code-dot-org/blob/592675a37a951b7086b300725a955cfcef503e57/apps/webpack.config.js/#L318-L321

If we remove `{skipEnvCheck: true}`, we get weird errors like:
```
ERROR in ./src/weblab/brambleHost.js
Module build failed (from ./node_modules/babel-loader/lib/index.js):
Error: [BABEL] /Users/seth/src/code-dot-org/apps/src/weblab/brambleHost.js: React Refresh Babel transform should only be enabled in development environment. Instead, the environment is: "test". If you want to override this check, pass {skipEnvCheck: true} as plugin options. (While processing: "/Users/seth/src/code-dot-org/apps/node_modules/react-refresh/babel.js")
    at ReactFreshBabelPlugin (/Users/seth/src/code-dot-org/apps/node_modules/react-refresh/cjs/react-refresh-babel.development.js:25:13)
    at async (/Users/seth/src/code-dot-org/apps/node_modules/@babel/core/lib/gensync-utils/async.js:46:33)
    at async (/Users/seth/src/code-dot-org/apps/node_modules/gensync/index.js:186:15)
    at /Users/seth/src/code-dot-org/apps/node_modules/gensync/index.js:216:13
    at Generator.next (<anonymous>)
    at /Users/seth/src/code-dot-org/apps/node_modules/@babel/core/lib/config/full.js:229:21
    at Generator.next (<anonymous>)
    at Function.<anonymous> (/Users/seth/src/code-dot-org/apps/node_modules/@babel/core/lib/gensync-utils/async.js:27:3)
    at Generator.next (<anonymous>)
    at step (/Users/seth/src/code-dot-org/apps/node_modules/gensync/index.js:269:25)
```

They are thrown from react-refresh-babel.development.js, which looks like this:
```
    // Only available in Babel 7.
    var env = babel.env();

    if (env !== 'development' && !opts.skipEnvCheck) {
      throw new Error('React Refresh Babel transform should only be enabled in development environment. ' + 'Instead, the environment is: "' + env + '". If you want to override this check, pass {skipEnvCheck: true} as plugin options.');
    }
```

Basically, the question is: why is `babel.env()` set to test when we run `yarn start`? Something in our code is setting this, its not the default.

In babel.config.json we find:
```
  "env": {
    "test": {
      "plugins": [
        // make dynamic imports work in unit tests
        "dynamic-import-node"
      ]
    }
  }
```

But this is not setting it (at least, removing this doesn't change anything), just changing plugins with it set.... 

Totally reasonable to decide we're not going to dig too deep into this, and just remove the FIXME, but wanted to maybe look at it for another hour or two first.

snickell:
Tests are breaking on drone on the first webpack, before running any real tests:
```
HeadlessChrome 0.0.0 (Linux 0.0.0) ERROR
--
154 | Uncaught ReferenceError: $RefreshReg$ is not defined
155 | at build/karma/commons.js:403249
```

What's weird is that tests pass locally, AND when I ssh into the drone instance and run `yarn test` from apps, they also pass.

However, ssh-ing in, setting env vars from unit_tests.sh and then running tests does REPO:
```
export CI=true
export RAILS_ENV=test
export RACK_ENV=test
export DISABLE_SPRING=1
export LD_LIBRARY_PATH=/usr/local/lib

# Repros:
bundle exec rake circle:run_tests
```

Looking at `commons.js` gives us:
```
/***/ "./src/imageUtils.js":
/*!***************************!*\
  !*** ./src/imageUtils.js ***!
  \***************************/

// SNIP!

function downloadBlobAsPng(blob) {
  var filename = arguments.length > 1 && arguments[1] !== undefined ? arguments[1] : 'image.png';
  var download = document.createElement('a');
  download.href = URL.createObjectURL(blob);
  download.download = filename;
  download.click();
}

var _c;

$RefreshReg$(_c, "URIFromImageData");

/***/ }),

/***/ "./src/javalab/Backpack.jsx":
/*!**********************************!*\
```

There are a lot of RefreshReg instances:
```
circleci@fb9a206ff2db:/drone/src/apps$ grep RefreshReg build/karma/commons.js | wc -l
951
`

**Comment by snickell** [comment]
Finding a similar issue here: https://stackoverflow.com/a/73000810

The issue seems to be potentially tied to the `skipEnvCheck: true` workaround/hack: https://github.com/code-dot-org/code-dot-org/pull/56882#issuecomment-2002483958

1. In this case, `URIFromImageData` is interpreted as a react component based on its capitalization. We should make sure that URIFromImageData works in a dev build.
2. We're doing karma builds with HOT=1.... maybe we should reconsider making HOT=1 the default if DEV=1.  For now, trying to set HOT=0 in karma.config.js

**Comment by snickell** [comment]
In theory, setting `devServer: { hot: true }` should also enable the `HotModuleReplacementPlugin`:
image

However, in practice with this setup, we were still getting the following error on `yarn start`:
```
    <w> [ReactRefreshPlugin] Hot Module Replacement (HMR) is not enabled! React Refresh requires HMR to function properly.
```

This went away when adding an explicit `new webpack.HotModuleReplacementPlugin({}),` in commit [e27e5a8](https://github.com/code-dot-org/code-dot-org/pull/56882/commits/e27e5a8fbb1a56de90af234ddefa4934b064ee0c). I'm not sure the warning was entirely valid, but I'd rather not have a spurious warning.

File: apps/webpack.config.js [hunk 1]
```diff
@@ -341,20 +349,6 @@ const WEBPACK_BASE_CONFIG = {
   ignoreWarnings: [/Failed to parse source map/],
 };
 
-// FIXME: figure out how to re-enable hot reloading with
```
bencodeorg, line 343: Note to remove

apps/webpack.config.js
```diff
@@ -670,18 +659,58 @@ function createWebpackConfig({
...
-            }),
+            new webpack.HotModuleReplacementPlugin({}),
+            new ReactRefreshWebpackPlugin(),
+            // Prints a URL for accessing the Dashboard via webpack-dev-server
+            {
+              apply: compiler => {
+                compiler.hooks.afterDone.tap('PrintDashboardURL', stats => {
+                  if (stats.hasErrors()) return;
+
+                  if (!process.env.WEBPACK_SERVE) {
```
bencodeorg, line 672: Is this set by webpack itself, or by us?

apps/envConstants.js
```diff
@@ -39,7 +37,7 @@ module.exports = {
   CIRCLECI: process.env.CIRCLECI,
   DRONE: process.env.DRONE,
   CIRCLE_TEST_REPORTS: process.env.CIRCLE_TEST_REPORTS,
-  // If set, will turn on react hot loader and run the webpack dev server
+  // If set, run webpack dev server (with hot reloading)
   HOT: !!process.env.HOT,
```
bencodeorg, line 41: ...not our change, but how does one turn off this bool? Setting HOT=0 still returns true, does this need to be `process.env.HOT !== '0'`?

I'm a little confused about our proposed behavior for the "not hot" case -- it seems like webpack dev server is still running when I set hot to 0 (incl. the proposed change above to actually make the HOT=0 flag do something). Is that desirable?

apps/webpack.config.js
```diff
@@ -670,18 +659,58 @@ function createWebpackConfig({
...
+          proxy: [
+            {
+              context: ['**'],
+              target: 'http://localhost-studio.code.org:3000',
+              changeOrigin: false,
+              logLevel: 'debug',
+            },
+          ],
+          host: '0.0.0.0',
+          hot: true,
```
bencodeorg, line 707: Nice to be explicit here, but I get this warning when running `yarn start`, so maybe remove?:

`<w> [webpack-dev-server] "hot: true" automatically applies HMR plugin, you don't have to add it manually to your webpack configuration.`

bencodeorg (commented):
We are very close, thank you for putting this all together! Just a bit of cleanup -- my main question is about what our expected/desired behavior is when `HOT=0`?

apps/webpack.config.js
snickell: I believe this is complaining _not_ that we set `hot: true` explicitly here despite it being the default, but instead its complaining that we manually add the HMR plugin in this line:
```
new webpack.HotModuleReplacementPlugin({}),
```

However, not having that line there was triggering a different and slightly scarier warning in react-refresh-webpack-plugin discussed here: https://github.com/code-dot-org/code-dot-org/pull/56882#issuecomment-2002595741

In theory setting hot adds that plugin, but I think its doing so too late for react-refresh-webpack-plugin to detect it. In my reading of the code, I /think/ that react-refresh-webpack-plugin is actually at fault, and this may be a spurious warning, but it was complex enough I didn't feel confident, and given a choice of which warning-that-probably-causes-no-harm to have, I thought this was the less scary. What you think?

apps/envConstants.js
snickell: Most of our env vars have the property "having them set to anything makes them true", and going back to this just returns HOT to working like our other env vars. DEV works this way, etc, and its pretty common for env vars. It basically is equivalent to a flag like --hot, where absence means the opposite.

I don't object to having the check be `HOT !== "0"`, but I didn't see anything about this variable that particularly called for that behavior, now that we don't default to enabling HOT (it was important when not-specified=true to have a way to specify false, now that not-specified=false, its ok to have anything-specified=true).

snickell:
Drone just started failing on new lint checks added in https://github.com/code-dot-org/code-dot-org/pull/57247 that were not applied fully to existing code in /apps/*js (to webpack.config.js, Gruntfile.js, etc... basically same files we are editing).

I have a fix here: #57369

snickell:
I think my biggest concern left is related to the '$RefreshReg$ is not defined' errors we were getting when our karma tests were built with HMR enabled. Now, obviously we don't want karma tests built with HMR, but it makes me want to check if HMR possibly broke some of the same values that weren't working in tests. Just seeing one of them work would make me feel better.

 In particular, I noticed that it was just assuming variables using Capitalization were React components. This is /probably/ ok, and they probably planned all this out, but I'd just like to see it work.

I propose we try to access/use `URIFromImageData`, as that was one of the non-react components getting marked for HMR. If we can use `URIFromImageData` normally, I think that's a sign that we're ok on this front.

apps/webpack.config.js
bencodeorg: Ah that makes sense rereading the warning message, thanks for the explanation!

apps/envConstants.js
bencodeorg: Chatted offline -- just noting for posterity that there's not really a way (nor do we really want there to be) a way to run your dev server in "not hot mode". eg, `HOT=0 yarn start` doesn't actually turn off hot reloading.

bencodeorg (approved):
This looks good to me! I say 🚀

bencodeorg:
> I think my biggest concern left is related to the '$RefreshReg$ is not defined' errors we were getting when our karma tests were built with HMR enabled. Now, obviously we don't want karma tests built with HMR, but it makes me want to check if HMR possibly broke some of the same values that weren't working in tests. Just seeing one of them work would make me feel better.
> 
> In particular, I noticed that it was just assuming variables using Capitalization were React components. This is /probably/ ok, and they probably planned all this out, but I'd just like to see it work.
> 
> I propose we try to access/use `URIFromImageData`, as that was one of the non-react components getting marked for HMR. If we can use `URIFromImageData` normally, I think that's a sign that we're ok on this front.

For posterity, we tested manually and were able to use `URIFromImageData` successfully with HMR enabled.

apps/webpack.config.js
snickell: this is set by `webpack serve`, but annoyingly, it also gets tripped in a weird way that I couldn't figure out by the concurrent task in Gruntfile.js forks. Otherwise, I would have used it more heavily (e.g. keying printing this message off WEBPACK_SERVE instead of HOT)

PR #56883, opened 2024-02-28
[Google Blockly] Convert cdoUtils to TypeScript

molly-moen:
Convert `cdoUtils` to Typescript. As part of this work I also improved the serialization types I initially created in #56805, so they no longer include `any`. In addition, I updated an incorrect typing from that PR, where we were using `Document` when really we should have been using `Element`. Because of the typing changes this includes some changes to `cdoSerializationHelpers` as well. I have added inline comments where there are interesting code changes.

## Links

## Testing story
Tested locally. I tested various scenarios where these functions get hit, such as converting from xml to json, loading an existing json project, and loading a project with shared functions.

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/blockly/addons/cdoFieldButton.ts
+++ b/apps/src/blockly/addons/cdoFieldButton.ts
@@ -7,13 +7,13 @@ interface ColorOverrides {
 }
 
 interface CdoFieldButtonOptions {
-  value: string | undefined;
-  validator: FieldValidator<string> | null;
+  value?: string;
+  validator?: FieldValidator<string> | null;
   onClick: () => void;
-  transformText: ((text: string) => string) | undefined;
-  icon: SVGElement | undefined;
-  colorOverrides: ColorOverrides | undefined;
-  allowReadOnlyClick: boolean | undefined;
+  transformText?: (text: string) => string;
+  icon?: SVGElement;
+  colorOverrides?: ColorOverrides;
+  allowReadOnlyClick?: boolean;
 }
 
 export default class CdoFieldButton extends GoogleBlockly.Field {
--- a/apps/src/blockly/addons/cdoSerializationHelpers.ts
+++ b/apps/src/blockly/addons/cdoSerializationHelpers.ts
@@ -439,9 +439,7 @@ export function convertFunctionsXmlToJson(functionsXml: string) {
  * TODO: define a type for projectState and proceduresState. Blockly defines these as any.
  */
 export function appendProceduresToState(
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
   projectState: WorkspaceSerialization,
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
   proceduresState: WorkspaceSerialization
 ) {
   console.log({projectState, proceduresState});
@@ -472,8 +470,7 @@ export function appendProceduresToState(
 
 // Function to check if a block with the given behaviorId exists in the project
 // TODO: define type for projectBlocks.
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-function blockExists(behaviorId: string, projectBlocks: any[]) {
+function blockExists(behaviorId: string, projectBlocks: JsonBlockConfig[]) {
   return projectBlocks.some(
     block =>
       block.type === BLOCK_TYPES.behaviorDefinition &&
--- a/apps/src/blockly/addons/cdoSerializationHelpers.ts
+++ b/apps/src/blockly/addons/cdoSerializationHelpers.ts
@@ -98,7 +98,7 @@ export function convertXmlToJson(xml: Element, embedded: boolean) {
     // Create a map of ids (key) and block serializations (value).
     const blockIdMap = stateToLoad.blocks.blocks.reduce(
       (map: Map<string, JsonBlockConfig>, blockJson: JsonBlockConfig) =>
-        map.set(blockJson.id, blockJson),
+        map.set(blockJson.id || '', blockJson),
       new Map()
     );
 
--- a/apps/src/blockly/addons/cdoSerializationHelpers.ts
+++ b/apps/src/blockly/addons/cdoSerializationHelpers.ts
@@ -92,7 +92,9 @@ export function convertXmlToJson(xml: Element, embedded: boolean) {
   //   x: the x attribute found in <block/> element
   //   y: the y attribute found in <block/> element
   const xmlBlocks = Blockly.Xml.domToBlockSpace(tempWorkspace, xml);
-  const stateToLoad = Blockly.serialization.workspaces.save(tempWorkspace);
+  const stateToLoad = Blockly.serialization.workspaces.save(
+    tempWorkspace
+  ) as WorkspaceSerialization;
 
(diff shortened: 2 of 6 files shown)


apps/src/blockly/addons/cdoSerializationHelpers.ts
```diff
@@ -414,8 +419,10 @@ export function convertFunctionsXmlToJson(functionsXml: string) {
   const parser = new DOMParser();
   const xml = parser.parseFromString(`<xml>${functionsXml}</xml>`, 'text/xml');
   const tempWorkspace = new Blockly.Workspace();
-  Blockly.Xml.domToBlockSpace(tempWorkspace, xml);
-  const proceduresState = Blockly.serialization.workspaces.save(tempWorkspace);
+  Blockly.Xml.domToBlockSpace(tempWorkspace, xml.children[0]);
```
molly-moen, line 422: I needed to change this because we were passing a `Document` when we should have been passing an `Element`. This is safe to do because the parameter is parsed (in `cdoXml.ts`) with `Array.from(xml.querySelectorAll('xml > block'));` which returns the same array for both.

apps/src/blockly/addons/cdoUtils.ts
```diff
@@ -353,13 +378,16 @@ export function registerCustomProcedureBlocks() {
...
   const parser = new DOMParser();
+  if (!Blockly.toolboxBlocks) {
+    return;
+  }
   // TODO: Update this to support JSON once https://codedotorg.atlassian.net/browse/CT-8 is merged
   const xmlDoc = parser.parseFromString(Blockly.toolboxBlocks, 'text/xml');
 
   // Find the category based on the custom attribute
-  const categories = xmlDoc.getElementsByTagName('category');
+  const categories = Array.from(xmlDoc.getElementsByTagName('category'));
```
molly-moen, line 391: We needed to use `Array.from` here (and a few other places) because TypeScript did not believe `categories` was an iterable type. The returned list is the same data as the original object.

apps/src/blockly/addons/cdoFieldButton.ts
```diff
@@ -7,13 +7,13 @@ interface ColorOverrides {
 }
 
 interface CdoFieldButtonOptions {
-  value: string | undefined;
-  validator: FieldValidator<string> | null;
+  value?: string;
```
mikeharv, line 10: This feels like a semantic difference? Previously, we were saying that value will be a string or undefined, and now we're saying it's a string but it's optional

apps/src/blockly/addons/cdoSerializationHelpers.ts
```diff
@@ -432,28 +439,27 @@ export function convertFunctionsXmlToJson(functionsXml: string) {
  * TODO: define a type for projectState and proceduresState. Blockly defines these as any.
  */
 export function appendProceduresToState(
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  projectState: any,
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  proceduresState: any
+  projectState: WorkspaceSerialization,
+  proceduresState: WorkspaceSerialization
```
mikeharv, line 443: Nice!

apps/src/blockly/addons/cdoSerializationHelpers.ts
```diff
@@ -432,28 +439,27 @@ export function convertFunctionsXmlToJson(functionsXml: string) {
...
       // If the block doesn't exist, add it to the student project
       projectBlocks.push(block);
-      projectProcedures.push(
-        sharedProcedures.find(procedure => procedure.id === procedureId)
+      const procedure = sharedProcedures.find(
+        procedure => procedure.id === procedureId
       );
+      if (procedure) {
+        projectProcedures.push(procedure);
+      }
```
mikeharv, line 462: I think we're doing this because `.find()` doesn't guarantee a result that we'd want, even though in our case we'd expect to always find one.

apps/src/blockly/addons/cdoUtils.ts
```diff
@@ -335,6 +359,7 @@ export function locationField(icon, onClick) {
         // Just ignore bad values
       }
     }
+    return '';
```
mikeharv, line 362: Could this have also been placed in the catch?

apps/src/blockly/types.ts
```diff
@@ -87,7 +86,7 @@ export interface BlocklyWrapperType extends GoogleBlocklyType {
   levelBlockIds: string[];
   isStartMode: boolean;
   isToolboxMode: boolean;
-  toolboxBlocks: Element | ToolboxDefinition | undefined;
+  toolboxBlocks: string | undefined;
```
mikeharv: Does this cause any problems for Music Lab where toolboxes are defined in JSON? Did the imported `ToolboxDefinition` type  cover this?

From the browser console on a music lab page:
image

mikeharv (approved):
Looks good to me. Most questions are for my own knowledge. One question around JSON toolboxes in Music Lab that I think is worth considering.

apps/src/blockly/addons/cdoFieldButton.ts
molly-moen: The reason I changed it is because we were already treating these as optional when initializing `FieldButton`. For example [here](https://github.com/code-dot-org/code-dot-org/blob/c6abf21d2cbfecb0cf94cef22e7708711cc92d53/apps/src/blockly/addons/cdoUtils.js#L339) we aren't passing value, validator, colorOverrides or allowReadOnlyClick. In order to make the compiler happy I had to make it optional.

apps/src/blockly/addons/cdoSerializationHelpers.ts
molly-moen: Can you clarify what you mean here? Is the new null check an issue?

mikeharv: No I don't think so! I was just making sure I understood why we were adding it.

apps/src/blockly/addons/cdoUtils.ts
molly-moen: No because we surround the `try/catch` with an `if`, so I put this at the end to cover both the catch and the else case.

mikeharv: Oh, got it!

apps/src/blockly/addons/cdoSerializationHelpers.ts
molly-moen: Yeah `find` could theoretically find nothing.

apps/src/blockly/types.ts
molly-moen: Good call out, I will double check this.

molly-moen: I don't think it causes issues but it is incorrect so I'll update this back.

molly-moen: I changed it to be just `ToolboxDefinition | undefined` because `ToolboxDefinition` includes `Node` as a possible type.

mikeharv: I was thinking the same thing!

PR #56666, opened 2024-02-19
Part 1: Update public projects gallery / Remove ability to publish projects

fisher-alice:
This PR updates the public gallery so that only featured projects are displayed. It also removes the ability for a user to publish their project.

This PR consists of 4 parts:
Part 1: Remove ability to publish projects (description and write-up included below)
Part 2: Update projects public gallery to display only featured projects https://github.com/code-dot-org/code-dot-org/pull/56886
Part 3: Add information banner about updated public gallery for signed-in users https://github.com/code-dot-org/code-dot-org/pull/56918
Part 4: Notify user if their project is selected to be featured since the project will be frozen https://github.com/code-dot-org/code-dot-org/pull/56932

Description of part 1 - removing ability for a user to publish their project:
- removing the 'Publish' button from the `ShareAllowedDialog`,
- removing the 'Publish' button from the sharing dialog that is displayed after clicking on the 'Finish' button. This is available on certain progressions usually in the last freeplay level,
- removing the 'Published' column in the Personal Projects table at '/projects'.

This PR is the compilation of a sequence of PRs that will include converting the public project gallery to a featured project gallery as according to [this spec](https://docs.google.com/document/d/1vU68tzXG72Z80MvM6xYEDwF5hLDf7V097P-0dDfpouc/edit). This PR WILL NOT BE MERGED until the sequence of PRs are approved.

## Before Part 1 Update

Personal Projects Table Header:
Screenshot 2024-02-19 at 2 35 50 PM

## After Part 1 Update

Personal Projects Table Header:
Screenshot 2024-02-19 at 11 31 18 AM

## Before Part 1 Update

Project share dialog - Artist:
Screenshot 2024-02-19 at 2 41 09 PM

Project share dialog - Dance Party:
Screenshot 2024-02-19 at 2 41 09 PM

Project share dialog with uploaded image for student account:
Screenshot 2024-02-19 at 3 03 36 PM

## After Part 1 Update

Project share dialog - Artist:
Screenshot 2024-02-19 at 2 38 54 PM

Project share dialog - Dance Party:
Screenshot 2024-02-19 at 2 40 18 PM

Project share dialog with uploaded image for student account:
Screenshot 2024-02-19 at 3 03 58 PM

## Before Part 1 Update
A signed out user on a level attempting to share:
Screenshot 2024-02-19 at 3 40 14 PM

## After Part 1 Update
A signed out user on a level attempting to share:
Screenshot 2024-02-19 at 3 40 06 PM

## Before Part 1 Update
[...]

PR #56886, opened 2024-02-28
Part 2: Update projects public gallery to display only featured projects

fisher-alice:
This PR updates the projects public gallery at 'projects/public' so that that the gallery displays only active featured projects. It is a follow-up to https://github.com/code-dot-org/code-dot-org/pull/56666 which removes the ability for a user to publish projects. Note that this PR (part 2) and part 1 will not be merged until parts 3 and 4 are completed and approved (described in follow-up section).

From a test student account, here is a list of projects that are published but NOT featured projects for the purpose of showing before/after updates below:
Screenshot 2024-02-28 at 7 00 24 PM

## Before Update
These are the projects displayed in the public project gallery at 'projects/public': Note that projects include both published featured projects AND published projects that are NOT featured.
Screenshot 2024-02-28 at 6 59 29 PM

There are 10 project app types and initially 4 project cards per project app type are displayed. A user can click on 'View more ...' and more published projects of that app type are displayed. This shows the projects displayed after a user click on 'View more Dance Party projects'. Note that you can view published projects that are not featured and featured project that are 'archived' because all published projects are displayed.
Screenshot 2024-02-28 at 6 59 59 PM

## After Update
These are the featured projects listed in the admin table at 'projects/featured' (admin user must have project validator permission): Note that the titles of the projects reflect the status of the featured project.
Screenshot 2024-02-28 at 7 05 17 PM

These are the projects listed in at 'projects/public' for both signed-in and signed-out users: There are only 16 projects displayed and they are all published, active featured projects. There is NO 'View more...' button.
Screenshot 2024-02-28 at 7 03 51 PM

## Links
[jira](https://codedotorg.atlassian.net/browse/LABS-487)
[product spec](https://docs.google.com/document/d/1vU68tzXG72Z80MvM6xYEDwF5hLDf7V097P-0dDfpouc/edit)

## Testing story
I updated multiple dashboard unit tests and UI tests. I also tested locally.
You can see drone build passed at https://github.com/code-dot-org/code-dot-org/pull/57019.

## Follow-up work to complete P0
- PART 3:  Projects page for signed in users (student and teachers) should have a banner up (temporary for ~ 4 weeks after public gallery is updated) updating them that the gallery has been removed, they will no longer be able to publish projects publicly and we now have a featured gallery available). 
- PART 4: Admin users will freeze a featured project once it is selected to be included in the public project gallery, i.e., have 'active' status. Therefore, add UI to notify a user if their project is frozen.

[jira ticket](https://codedotorg.atlassian.net/browse/LABS-536) - to refactor `projects_list` to use Rails/ActiveRecord.

## Alternative solution to PART 4
This pathway will be investigated. If we receive reports about a 'bad' user experience as a result of an active featured project being frozen:
- Use a versioned URL to share featured project in public gallery, i.e., a snapshot of the project once the project is displayed or 'active'.
- Remove warning banner for versioned URL when project is being featured in public gallery.
- Tag featured projects on S3 with a new object tag and create a new lifecycle rule to delete old versions after 1 year. Currently old versions are deleted after 60 days.

## PR Checklist:

- [ ] Code is well-commented


dashboard/test/controllers/api/v1/projects/public_gallery_controller_test.rb
```diff
@@ -88,62 +42,44 @@ class Api::V1::Projects::PublicGalleryControllerTest < ActionController::TestCas
...
+    assert_equal 22, project_row['storage_id']
+    assert_equal 33, project_row['id']
+    assert_equal '/v3/files/thumbnail.png', project_row['thumbnailUrl']
     assert_equal 'applab', project_row['type']
-    assert_equal '2017-03-03T00:00:00.000-08:00', project_row['publishedAt']
-    assert_equal '13+', project_row['studentAgeRange']
-    assert_equal 'P', project_row['studentName']
-  end
-
-  test 'project details are correct without thumbnail' do
```
fisher-alice, line 109: Replaced this test with a new one in [projects_list_test.rb](https://github.com/code-dot-org/code-dot-org/blob/414ddba7b9723f8f7f4ab6c0a5b4fe65b2503688/dashboard/test/lib/projects_list_test.rb#L246-L268).

dashboard/test/controllers/api/v1/projects/public_gallery_controller_test.rb
```diff
@@ -67,13 +28,6 @@ class Api::V1::Projects::PublicGalleryControllerTest < ActionController::TestCas
     response: :bad_request,
   )
 
-  test_user_gets_response_for(
```
fisher-alice, line 70: Removed `published_before` param.

apps/src/templates/projects/ProjectAppTypeArea.jsx
```diff
@@ -17,26 +17,26 @@ class ProjectAppTypeArea extends React.Component {
   static propTypes = {
     labKey: PropTypes.string.isRequired,
     labName: PropTypes.string.isRequired,
-    labViewMoreString: PropTypes.string.isRequired,
+    labViewMoreString: PropTypes.string,
```
fisher-alice: I am not removing these props because the featured gallery may be updated in the future according to [this product spec](https://docs.google.com/document/d/1vU68tzXG72Z80MvM6xYEDwF5hLDf7V097P-0dDfpouc/edit?disco=AAABGLQtJe0).

thomasoniii (approved):
lgtm 🚀

apps/src/templates/projects/ProjectAppTypeArea.jsx
bencodeorg: Could you throw a comment in this file to that effect? I feel a little torn about leaving this stuff hanging around that may or may not get used (but I'm down with the decision to keep it). I also think the "view more" button would break if it were used with the labKey of "featured" because "featured" isn't a supported `projectType` here (projectType gets copied from labKey):

https://github.com/code-dot-org/code-dot-org/blob/3342df315b85926b0401a066fed5df7867d562b2/apps/src/templates/projects/ProjectAppTypeArea.jsx#L119

So, kind of a lot of confusing stuff for future viewers to work through.

apps/src/templates/projects/PublicGallery.jsx
```diff
@@ -42,18 +42,25 @@ class PublicGallery extends Component {
    * See the PropTypes of each component for a definition of each format.
    */
   mapProjectData(projectLists) {
-    return _.mapValues(projectLists, projectList => {
-      return projectList.map(projectData => {
-        return {
+    let allFeaturedProjects = [];
+    for (const projectListName in projectLists) {
+      if (projectListName === 'library') {
```
bencodeorg: We're returning libraries from the server when we request featured projects? Do we actually feature libraries anywhere?

dashboard/lib/projects_list.rb
```diff
@@ -236,6 +247,7 @@ def fetch_featured_projects_by_type(project_type)
           project_type: project_type.to_s,
           state: 'active'
         ).
+        exclude(featured_at: nil).
```
bencodeorg, line 258: Why didn't we need this before?

dashboard/test/controllers/api/v1/projects/public_gallery_controller_test.rb
```diff
@@ -1,27 +1,12 @@
...
-      birthday: 13.years.ago.to_datetime,
-      abuse_score: 0
-    }
-    stub_projects = [published_applab_project]
-
-    # return 1 applab project for initial DB request, then an empty set
-    # for all subsequent db requests. Subsequent requests are only expected
-    # when 'all' project types are requested.
-    Projects.stubs(:table).returns(db_result(stub_projects)).then.returns(db_result([]))
+    published_applab_project = {storage_id: 22, id: 33, name: "Applab1_to_be_featured", thumbnailUrl: "/v3/files/thumbnail.png", type: "applab", publishedAt: "2023-02-14T02:14:51.000+00:00", studentName: "A", studentAgeRange: "18+", isFeatured: true}
```
bencodeorg: Nit: readability

dashboard/test/controllers/api/v1/projects/public_gallery_controller_test.rb
bencodeorg: This maybe adds to confusion with what works/doesn't in `ProjectAppTypeArea`, since we still have code there that is passing in a `published_before` param. How'd you decide to remove this, but keep the UI around?

bencodeorg (commented):
A couple clarifying questions but looking good!

apps/src/templates/projects/PublicGallery.jsx
fisher-alice: Good point! I don't believe so. I think published libraries are fetch by the call to `fetch_section_libraries` since only libraries within a user's same section can be shared.
I'll go ahead and add this condition to the backend in `fetch_featured_published_projects`.

apps/src/templates/projects/ProjectAppTypeArea.jsx
fisher-alice: Good points! I'll update so that a call to this api endpoint works if labKey is assigned `featured`. Will also add a comment.

dashboard/lib/projects_list.rb
fisher-alice: We didn't have 'bookmarked' featured projects before.

fisher-alice: status | featured_at | unfeatured_at
-- | -- | --
'bookmarked' | NULL | NULL
'active' | most recently featured at time stamp | NULL
'archived' | most recently featured at time stamp | most recently unfeatured at time stamp

bencodeorg: ah right...another instance where Rails would be a helpful (ie, it won't be super obvious to future readers that the combination of where(unfeatured_at: nil...) and exclude(featured_at: nil) is what gets you the right set of projects. Alas...

dashboard/test/controllers/api/v1/projects/public_gallery_controller_test.rb
fisher-alice: Since this update to the public gallery to only display featured projects will go through another iteration this summer, I wanted to leave intact the code which fetched additional projects when the 'View more' link is clicked. See [spec doc](https://docs.google.com/document/d/1vU68tzXG72Z80MvM6xYEDwF5hLDf7V097P-0dDfpouc/edit?disco=AAABGLQtJe0)
However this current implementation does not include the 'View more' button so `published_before` param is not being used.

However, I can see how this can be confusing to a reader without the context of the project. ~~I'll go ahead and leave the param and when we re-visit the public gallery this summer, we can update as needed.~~
Updates explained below!

Thanks!

fisher-alice: Actually - I'm going to see if I can get some clarification on the next iteration of the public featured projects gallery from @samantha-code before updating here. Thanks!

samantha-code: The next iteration of the project gallery (something @moneppo believes will be needed for music lab launch in May), will be the ability to feature different categories of projects. From the validators perspective - this would be a column to pick (or create) a new category in the gallery for each  project (where applicable). 

From the student perspective, this is a SUPER lightweight mockup of what this could look like: 
**One featured projects section:**

image (11)


 **more than 1 featured projects section**
image (10)



**3. no featured projects sections**
image (12)

Happy to provide anymore clarity here if needed!

dashboard/lib/projects_list.rb
fisher-alice: I created a follow-up ticket to refactor `projects_list.rb` -> https://codedotorg.atlassian.net/browse/LABS-536
Thanks!

dashboard/test/controllers/api/v1/projects/public_gallery_controller_test.rb
```diff
@@ -1,149 +1,93 @@
...
     :index,
     name: 'bad request when project type is invalid',
-    params: -> {{project_type: 'bogus', limit: 1}},
-    response: :bad_request,
-  )
-
-  test_user_gets_response_for(
-    :index,
-    name: 'bad request when published_before is specified for project type all',
-    params: -> {{project_type: 'all', limit: 1, published_before: '2017-02-17T13:59:11.000-08:00'}},
```
fisher-alice, line 73: Now `featured_before` can be specified for project type all because in the future, there may be a 'View more' link for featured projects according to @samantha-code .

dashboard/test/controllers/api/v1/projects/public_gallery_controller_test.rb
```diff
@@ -1,149 +1,93 @@
...
+    params: -> {{project_type: 'bogus'}},
     response: :bad_request,
   )
 
   test_user_gets_response_for(
     :index,
-    name: 'bad request when limit is less than 1',
```
fisher-alice, line 79: `limit` no longer passed as param since defined as a shared constant.

dashboard/test/controllers/api/v1/projects/public_gallery_controller_test.rb
fisher-alice: Thanks so much Sam! This is very helpful.
I've made a few updates including adding the optional `featured_before` param when fetching active published featured projects in anticipation of the next iteration of the public gallery when the 'View more' button may be added to either the general featured project public gallery or a special section of the public gallery.

apps/src/templates/projects/ProjectAppTypeArea.jsx
```diff
@@ -109,17 +110,19 @@ class ProjectAppTypeArea extends React.Component {
...
    */
   fetchOlderProjects() {
-    const {projectList, labKey: projectType} = this.props;
+    const projectType =
+      this.props.labKey === 'featured' ? 'all' : this.props.labKey;
+    const {projectList} = this.props;
     const oldestProject = projectList[projectList.length - 1];
-    const oldestPublishedAt =
-      oldestProject && oldestProject.projectData.publishedAt;
+    const oldestFeaturedAt =
```
bencodeorg, line 127: Just looking at prop types, and `featuredAt` doesn't look like it's listed as available in what gets passed to this component...might just need to update the type?

fisher-alice: Thanks! I added `featured_at` as a property to extract  in `projects_list`, and then added as a property to the renamed `publishedFeaturedProjectDataPropType`. I went ahead and sorted list as well.

Thanks!

bencodeorg (approved):
LGTM!

PR #56888, opened 2024-02-28
Ceara/aitt 460 rubric sticky fix reload

cearachew:
Added states to session storage to make the open state of the rubric FAB sticky on a page reload. When a new student or section is selected, the page reload, this stores the state of the FAB to keep it open after the reload so teachers can make fewer clicks when going through student feedback. I also (bonus!) added a state to remember which tab it was open to.

This also fixes a race condition that this revealed, where the displayed feedback could be set before receiving all feedback from the http post.

## Links

Jira ticket: https://codedotorg.atlassian.net/browse/AITT-460

## Testing story

Updated a UI test to remove the click to open the rubric FAB

## Follow-up work

There's possible follow-up work to make the page not refresh at all when a new student is selected, but that would be a large amount of work because most parts of the page (outside of the FAB) do need to re-render

## PR Checklist:

- [X] Tests provide adequate coverage
- [ ] Code is well-commented


apps/src/templates/rubrics/RubricFloatingActionButton.jsx
```diff
@@ -60,6 +60,14 @@ function RubricFloatingActionButton({
     }
   }, [eventData, studentLevelInfo]); // Neither of these should change, so this should run once
 
+  useEffect(() => {
+    const eventName = isOpen
+      ? EVENTS.TA_RUBRIC_OPENED_FROM_FAB_EVENT
+      : EVENTS.TA_RUBRIC_CLOSED_FROM_FAB_EVENT;
+    analyticsReporter.sendEvent(eventName, eventData);
+    trySetSessionStorage(sessionStorageKey, isOpen);
+  }, [isOpen, eventData]);
```
davidsbailey: it seems like before this event would only be fired when the user clicked to open the FAB, but now (please correct if if I'm mistaken) it looks like we are firing the event when they load the page after previously having it open. @tess323 can you confirm whether we want the amplitude event to fire only when the user clicks?

davidsbailey (commented):
A few initial comments:

Do we also want to update this text to no longer mention the teacher panel? e.g. "Select a section and student to view and evaluate their work."
Screenshot 2024-02-28 at 3 25 59 PM

apps/src/templates/rubrics/LearningGoals.jsx
```diff
@@ -246,11 +246,15 @@ export default function LearningGoals({
             if (json.understanding >= 0 && json.understanding !== null) {
               understandingLevels.current[index] = json.understanding;
             }
+            if (index === currentLearningGoal) {
+              setDisplayFeedback(teacherFeedbacks.current[currentLearningGoal]);
+              setDisplayUnderstanding(
+                understandingLevels.current[currentLearningGoal]
+              );
+            }
```
davidsbailey, line 254: This threw me for a loop because I am still seeing references to these setters in LearningGoal.jsx. @Nokondi can you please take a follow-up to remove dead code from the experiment launch?

davidsbailey (approved):
Nice job y'all tracking down the race condition! Code changes look good. product questions could be addressed now or in a follow-up conversation/PR.

apps/src/templates/rubrics/RubricFloatingActionButton.jsx
cearachew: That would be easy to change back, I assumed it was more of "we want to know whether the FAB is open or closed" rather thatn "we want to know when it's clicked"

tess323: Original intent was when the FAB was clicked to indicate the start of using the TA.  I would be interested in having both events if possible - fine to add whichever we don't currently have to next sprint.

cearachew: I rolled back the change I made so that it only fires when the user makes a click, and we can add a ticket to make a different event for if the page loads with it open or something?

davidsbailey: Sounds great @cearachew ! Can you please make sure that ticket is opened before moving on?

cearachew: Issue created: https://codedotorg.atlassian.net/browse/AITT-493

apps/src/templates/rubrics/LearningGoals.jsx
davidsbailey: https://codedotorg.atlassian.net/browse/AITT-495

PR #56890, opened 2024-02-28
Add initial types for Gen AI tool

sanchitmalhotra126:
Add some initial types for the new Gen AI unit chatbot tool, which will repurpose the AI chat created for EduBot. These will most likely be tweaked as we go along, but these should hopefully give us a good starting point to start creating UI and mock levels.

## Links
[Gen AI Tool Spec](https://docs.google.com/document/d/1yIez6ooHbfwfM2W16pQvGHgKTe97AdGJN7YECTOpmBU/edit)

## Testing story

Tested locally; not actually used anywhere so no user-facing impact.

### DIFF
--- a/apps/src/aichat/types.ts
+++ b/apps/src/aichat/types.ts
@@ -1,16 +1,6 @@
 import {LevelProperties} from '@cdo/apps/lab2/types';
 import {AiTutorInteractionSaveStatus} from '@cdo/apps/util/sharedConstants';
 
-// TODO: Ideally this type would only contain keys present in
-// translated string JSON files (ex. apps/i18n/aichat/en_us.json).
-// However, this requires depending on files outside of apps/src,
-// so this approach is still being investigated. For now, this type
-// is an object whose keys are all functions which return strings,
-// matching what we expect for a locale object.
-export type AichatLocale = {
-  [key: string]: () => string;
-};
-
 export type ChatCompletionMessage = {
   id: number;
   role: Role;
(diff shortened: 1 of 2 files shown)


apps/src/aichat/types.ts
```diff
@@ -31,7 +21,52 @@ export const Status = AiTutorInteractionSaveStatus;
...
+ * Each field is the same as AiCustomizations, but with an additional visibility property.
+ */
+export type LevelAiCustomizations = {
+  [key in keyof AiCustomizations]: {
+    value: AiCustomizations[key];
+    visibility: Visibility;
+  };
+} & {
+  /** The initial panel to show when displaying the level */
+  initialPanel?: 'edit' | 'presentation';
```
fisher-alice, line 71: Is this being assigned the mode of the `initialPanel`, i.e., edit mode vs presentation mode?

apps/src/aichat/types.ts
```diff
@@ -31,7 +21,52 @@ export const Status = AiTutorInteractionSaveStatus;
...
 
 export interface AichatLevelProperties extends LevelProperties {
+  // --- DEPRECATED - used for old AI Chat
   systemPrompt: string;
   botTitle?: string;
   botDescription?: string;
+  // ---
+
+  /** Initial AI customizations set by the level. */
+  initialAiCustomizations?: LevelAiCustomizations;
```
fisher-alice, line 36: Curious why this named `initialAiCustomization` instead of `aiCustomization`? Would the visibility of the ai customization change at some point after being set by the levelbuilder?

apps/src/aichat/types.ts
sanchitmalhotra126: Yeah, we want a way for levelbuilders to specify which panel (edit/presentation) should show when first loading the level.

apps/src/aichat/types.ts
sanchitmalhotra126: Yeah was having a hard time naming this 🙂 the visibility won't change, but the AI customizations will as students make edits. I wanted to make clear the distinction between the initial/starting set of customizations that the levelbuilder sets on the level and the student-specific customizations that are then saved to their specific project. Lmk if that doesn't make sense or if you think another name would be better!

apps/src/aichat/types.ts
fisher-alice: What do you think of `initialPanelMode`?

apps/src/aichat/types.ts
fisher-alice: Oh gotcha. So visibility is not editable by the student user but the customization values are depending on the visibility set by the levelbuilder. Maybe add a comment to clarify this?

fisher-alice (approved):
Thanks for answering my questions!

apps/src/aichat/types.ts
sanchitmalhotra126: Sure thing - I have a comment down below on the `Visibility` type, but will add to this comment here as well.

apps/src/aichat/types.ts
sanchitmalhotra126: My thinking was that the edit and presentation panels are two different distinct panels? And so this value tells us which panel to show first. I'm just a bit unsure if `initialPanelMode` implies that different panels can also have different modes?

fisher-alice: Oh interesting. Thanks for explaining! Then keeping as is makes sense.

PR #56891, opened 2024-02-28
Refine .gitattributes: "What's in LFS?"

snickell:
This PR modifies .gitattributes to refine our patterns and document it. All the other file changes are a result of hundreds of small files (mostly PNGs) moving into or out of Git LFS, these should be non-changes for all purposes except LFS.

Specific changes to .gitattributes:
- split .gitattributes into sections of inclusion/exclusion with more narrowly defined purpose
- document our .gitattributes approach in a way that hopefully makes it long-term stable
- add several file types that are ALWAYS in Git LFS (adds several hundred small PNGs to LFS)
- JS, HTML and CSS are ALWAYS excluded from Git LFS (removes a dozen files from LFS)
- collapse almost identical sites and sites.v3 patterns together to simplify and shorten file
- exclude `pegasus/sites/**/*.json` from LFS @kelbyhawn

@code-dot-org/i18n : this PR does NOT actually modify i18n files, its just moves a few of them them out of LFS.


.gitattributes
```diff
@@ -1,39 +1,64 @@
+# Git LFS inclusion / exclusion is controlled by this file. Our approach is to:
+# INCLUDE dirs in Git LFS broadly, and then EXCLUDE text file-types narrowly.
+#
+# Why whitelist? Once a large binary file is committed to LFS, it is VERY
```
ebeastlake: nit: Prefer use of an alternative to whitelist/blacklist. Maybe "allow list"?

ebeastlake: Also, something about these two sentences I need clarification on. I understand broadly that your approach prioritizes managing repository size and complexity while not worrying about the ability to correct errors with narrowly excluded smaller files. But your comment says, "Why whitelist?" which, upon first read, I think refers to "including in LFS", and then you describe how accidental inclusion in LFS is really hard to undo. So I'm reading:
```
why be inclusive of large files? --> because accidentally including large files is a really tough mistake to undo
```
But maybe how I'm meant to read this is,
```
why narrowly exclude? --> unlike big files, undoing a mistake with a small file is easy
```

I think there's something about the "whitelist" meant to stand for "exclusion" that is making this hard for me to read, and maybe I'm the only person who would get hung up on this, but since I had to sit and think about it for a few minutes, I figured it's worth mentioning 🤷‍♀️

.gitattributes
```diff
@@ -1,39 +1,64 @@
...
+# hard to remove it from the git history as it requires a history rewrite.
+# Once merged to `staging`, that NNN MB file makes our git repo that much
+# bigger forever".
+#
+# In contrast, if a text file is accidentally included in LFS, adding it back to
```
ebeastlake: Since text is broadly defined here in a way that may not be familiar to someone with Git experience, is it worth leaving a comment about what it includes? "In this context, 'text file' refers to any file that Git normally recognizes as text because it contains only printable characters and white space. It covers many file types, including source code, HTML, CSS, JSON, and YAML files, assuming they don't contain binary data."?

ebeastlake: Also, TBD if my Googling there is correct 😅

.gitattributes
```diff
@@ -1,39 +1,64 @@
...
+pegasus/sites*/**/*.collate !text !filter !merge !diff
+pegasus/sites*/**/*.erb !text !filter !merge !diff
+pegasus/sites*/**/*.fetch !text !filter !merge !diff
+pegasus/sites*/**/*.haml !text !filter !merge !diff
+pegasus/sites*/**/*.json !text !filter !merge !diff
+pegasus/sites*/**/*.md !text !filter !merge !diff
+pegasus/sites*/**/*.moved !text !filter !merge !diff
+pegasus/sites*/**/*.partial !text !filter !merge !diff
+pegasus/sites*/**/*.redirect !text !filter !merge !diff
+pegasus/sites*/**/*.scss !text !filter !merge !diff
```
ebeastlake, line 51: Nice condensing!

.gitattributes
```diff
@@ -1,39 +1,64 @@
...
-pegasus/sites/**/*.md !text !filter !merge !diff
-pegasus/sites/**/*.partial !text !filter !merge !diff
-pegasus/sites/**/*.moved !text !filter !merge !diff
-pegasus/sites/**/*.redirect !text !filter !merge !diff
-pegasus/sites/**/*.scss !text !filter !merge !diff
+pegasus/sites*/**/*.collate !text !filter !merge !diff
+pegasus/sites*/**/*.erb !text !filter !merge !diff
+pegasus/sites*/**/*.fetch !text !filter !merge !diff
+pegasus/sites*/**/*.haml !text !filter !merge !diff
+pegasus/sites*/**/*.json !text !filter !merge !diff
```
ebeastlake, line 44: Did `.js` get lost in the jump?

dashboard/public/google3d42ce63ecebc76e.html
ebeastlake, line 1: Why did this `html` file not get added? 🤔

pegasus/sites.v3/code.org/config.json
ebeastlake, line 1: Similar question about these `json` files.

ebeastlake (commented):
Couple comments (some nitpicky because I'm relying on eagle eyes in this review as much as having any special LFS context), but basically this seems LGTM!

dashboard/public/google3d42ce63ecebc76e.html
snickell: Because HTML is in the:
```
# Small list of very common text file types to EXCLUDE from Git LFS:
*.css !text !filter !merge !diff
*.js !text !filter !merge !diff
*.html !text !filter !merge !diff
``

File: pegasus/sites.v3/code.org/config.json
Comment (snickell) [reply pegasus/sites.v3/code.org/config.json:1]: Because *.json is in the:
```
# EXCLUDE these files from our Git LFS dirs:
# snip
pegasus/sites*/**/*.json !text !filter !merge !diff
```

File: .gitattributes (hunk 1 above)
Comment (snickell) [reply .gitattributes]: It was hard to read because you read it backwards.... whitelist here was intended to refer to "narrow inclusion in LFS". I'll just use include/exclude.

Its not that small vs large file mistakes are easy to correct. Its that INCLUDING IN LFS is very hard to undo, but EXCLUDING FROM LFS is easy to undo.

File: .gitattributes (hunk 2 above)
Comment (snickell) [reply .gitattributes]: Hot take: people without the experience to interpret "text file" shouldn't be in .gitattributes 😅. I think what really matters here is "MY FILE IS NOT DIFFING!!!", so I've rewritten to explain text file in terms of what you're trying to do:
> In contrast, if a text file (= a file that you want to be able to diff)
> is accidentally included in LFS, adding it back to git is easy,

File: .gitattributes (hunk 1 above)
Comment (snickell) [reply .gitattributes]: Does this read more clearly for you?
```
# Git LFS inclusion / exclusion is controlled by this file. Our approach is to:
# INCLUDE dirs in Git LFS broadly, and then EXCLUDE text file-types narrowly.
#
# Why include in LFS broadly? Once a large binary file is comitted without lfs,
# it is VERY hard to remove it from the git history as it requires a history
# rewrite. This cannot be done without major disruption to all devs. Thus,
# once merged to `staging`, that NNN MB file that missed LFS makes our git repo
# that much bigger forever.
#
# In contrast, if a text file (= a file that you want to be able to diff)
# is accidentally included in LFS, adding it back to git is easy,
# see: "HOW TO EXCLUDE A FILE/FILES THATS IN AN LFS DIR" below.
```

I've spent quite a lot of time nuancing this text, I think to some extent it falls under "this is complicated and can't be explained simply". What I've tried to do is not fully explain (i.e. not going to redocument .gitattributes), but provide enough pointers that somebody could figure out what's going on and why if they do some research.

The "sections" I created are there to guide people without full understanding to being able to usefully add/remove things from Git LFS that meet certain patterns without breaking everything.

snickell: The explanation of include vs exclude was mostly intended to defend against a well-meaning dev 5 years from now deciding to flip the logic (the reverse really is simpler, but I think it makes git history mistakes much more likely and thus raises the chance that somebody 5 years from now will need to do another git history rewrite because .blorgh files became a thing and they're 5GB each and somebody started just checking them in)

dashboard/public/google3d42ce63ecebc76e.html
snickell: Its being re-added-to-normal-git (i.e. removed from LFS) in this PR because it /used/ to be in LFS before this PR. This PR adds some broader LFS-exclusions, and thus pulls a number of small files out of LFS that previously were there (but probably shouldn't have been).

.gitattributes
ebeastlake: The modification is clear to me! You're right; I was struggling not to read it backward. Framing in terms of include/exclude helps.

.gitattributes
ebeastlake: Corollary to hot take: People without the experience to interpret "text file" shouldn't be reviewing the PR to `.gitattributes` 😂 🙈

ebeastlake (approved):

.gitattributes
snickell: 🔥 BURN 🔥

PR #56897, opened 2024-02-29
Updating a11y linter to remove unnecessary toggled off blob rule

hannahbergam:
I was taking another look at how many errors arise from various rules if we were to turn them on from the accessibility linter. When I turned off the line disabling the /blob rule, the build passed. Through some research of the [linter documentation](https://www.npmjs.com/package/eslint-plugin-jsx-a11y) and [this log of errors](https://gist.github.com/bethanyaconnor/7f59c9f892a290f863a9692472512472) when the linter was first implemented, it appears the blob rule is actually from the 'anchor-is-valid' rule, which has its own line. When you add up all the lines in that log, it totals 615, which would align with the 69 errors being counted twice. 

This is from the initial sorting the team did when implementing the linter:
Screenshot 2024-02-28 at 10 50 22 PM

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/.eslintrc.js
+++ b/apps/.eslintrc.js
@@ -7,8 +7,8 @@
 // reenabling.
 const rulesToEventuallyReenable = {
   'jsx-a11y/anchor-is-valid': 'off',
+  'jsx-a11y/blob': 'off',
   'jsx-a11y/click-events-have-key-events': 'off',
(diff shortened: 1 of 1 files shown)


bethanyaconnor (approved):
If the build passes, that's good enough for me! Great catch!

PR #56736, opened 2024-02-22
Skip failing test approaching leap year

bethanyaconnor:
I added some logging to this test and determined this is a leap year issue. I'm not sure if it'll only happen today or for the next week, but I don't believe this is failing due to some recent change.

Logging output:
`Time.now`: 2024-02-22 14:20:13 
`Timecop.travel(7.days)`: 2024-02-29 14:20:13 +0000
`Timecop.travel(1.year)`: 2025-02-28 14:20:13 +0000

## PR Checklist:

- [ ] Tests provide adequate coverage
- [ ] Privacy and Security impacts have been assessed
- [ ] Code is well-commented
- [ ] New features are translatable or updates will not break translations
- [ ] Relevant documentation has been added or updated
- [ ] User impact is well-understood and desirable
- [ ] Pull Request is labeled appropriately
- [ ] Follow-up work items (including potential tech debt) are tracked and linked

PR #56903, opened 2024-02-29
Skip failing test for leap year

bethanyaconnor:
This test started failing recently with no clear reason why. I've already had to skip a different test from this file for leap year in https://github.com/code-dot-org/code-dot-org/pull/56736, so I'm skipping this one as well. I have a task for next week to re-enable these tests: https://codedotorg.atlassian.net/browse/ACQ-1526

Unfortunately, I haven't actually tested this change yet as I'm working on something else. I'll merge ahead of Drone if I or someone else gets a chance to verify locally.


fisher-alice (approved):
Checked locally (w/o and then with skip) and looks good! Thanks for taking care of this.

PR #56906, opened 2024-02-29
AI Tutor: add level type to Amplitude logs for AI Tutor so we can use for groupBy analysis 

Erin007:
We had real users in a classroom visit today! As I was reviewing the Amplitude AI Tutor usage logs, I realized that we aren't tracking `levelType` when we log what type of tutor the student is using. We are already tracking `levelType` when the [panel is opened or closed](https://github.com/code-dot-org/code-dot-org/blob/f76661a133b2661fe8ef55463e7b4fcd308d2ec4/apps/src/code-studio/components/aiTutor/aiTutorFloatingActionButton.tsx#L27). Adding in `levelType` so we can use it in groupBy analysis for each tutor type.


molly-moen (approved):
🎉

PR #37002, opened 2020-09-30
[Google Blockly] Implement trashcan

ajpal:
This sort of splits the difference between our existing trashcan and Google's default one.
- It's the same trashcan shape as Google Blockly's, though we change the color to Code.org purple
- It's positioned over the toolbox instead of in the bottom corner
- I added the 🚫  signifier for when you can't delete the block. Google's UI hint is just that the lid doesn't open, but I figure we might want our more explicit UI here.
- The trashcan shows any time you're dragging a block. I haven't implemented the fade in UI that we used to have. I think this is okay for now, and can be added later if needed.

Sep-30-2020 15-53-51

PR #56907, opened 2024-02-29
[Google Blockly] Convert remaining* addons to TypeScript

molly-moen:
With this PR every file directly in the `blockly/addons` file will be converted to TypeScript 🎉 I did not convert the files under `blockly/addons/plusMinusBlocks` because they are mostly copied from a Blockly plugin that has not been converted to TypeScript. For now I'm not prioritizing converting these.

These changes were pretty straightforward, mostly types and null checks. I added comments where things were more interesting. Of note is I was able to delete `cdoTouch` because it was overriding a defunct method.
The files converted are:
- `cdoSpritePointer`
- `cdoTrashcan`
- `cdoVariables`
- `cdoVerticalFlyout`
- `contextMenu`
- `functionEditorConstants`
- `unknownBlock`

## Links

## Testing story
I tested the affected functionality locally.

## Follow-up work
Here is what is left under `blockly` and whether I think it's worth converting:
- (folder) `/addons/plusMinusBlocks`: No, it's mostly copied over from Blockly and is not in TypeScript on their end. If they convert `@blockly/block-plus-minus` to TypeScript we could pull that in.
- (folder) `/customBlocks/cdoBlockly`: No, I don't think it's worth our time to convert cdo Blockly code
- (file) `cdoBlocklyWrapper`: No, also cdo Blockly code
- (folder) `/customBlocks/googleBlockly`: Yes. There are 15 files in here of varying sizes. It's a non-zero amount of work but _hopefully_ the work to convert everything else will make these straightforward.
- (folder) `/themes`: Maybe. These are mostly constants, and we are doing some weird things here with module files to convert styles to accessible styles. We could convert the theme files and not the constant files. Overall these are lower priority.

TL;DR, the Google Blockly custom block files are what we should convert next, and we can decide at that point if we want to do any of the remaining theme files.

## PR Checklist:

- [ ] Code is well-commented


apps/src/blockly/addons/cdoTouch.js
```diff
@@ -1,15 +0,0 @@
-export default function initializeTouch(blocklyWrapper) {
-  // Aliasing Google's setClientFromTouch() so that we can override it, but still be able
-  // to call Google's setClientFromTouch() in the override function.
-  blocklyWrapper.Touch.originalSetClientFromTouch =
-    blocklyWrapper.Touch.setClientFromTouch;
```
molly-moen, line 5: This function does not exist on `Blockly.Touch` anymore. We added this override to fix flappy tests. I verified that they pass without this change, which makes sense as since this function doesn't exist, overriding it does nothing.

apps/src/blockly/addons/cdoSpritePointer.ts
```diff
@@ -68,53 +75,72 @@ export function updatePointerBlockImage(
...
+  imageSourceBlock: ExtendedBlockSvg,
+  imageIndexOnSource: number
+) {
   const targetConnection =
     imageSourceBlock.inputList[imageIndexOnSource]?.connection
       ?.targetConnection;
   // We only want to get the image from a connection block that is not an insertion marker.
   // If the block is an insertion marker that means the block is being dragged
   // and is not yet connected to the image source block.
+  const sourceBlock = targetConnection?.getSourceBlock();
```
molly-moen, line 93: `sourceBlock_` is private, so we use `getSourceBlock()` instead (which returns `sourceBlock_`)

apps/src/blockly/addons/cdoSpritePointer.ts
```diff
@@ -68,53 +75,72 @@ export function updatePointerBlockImage(
...
-function changePointerImage(imageUrl, block) {
+function changePointerImage(
+  imageUrl: string | undefined,
+  block: ExtendedBlockSvg
+) {
   if (!imageUrl || imageUrl.length === 0) {
     resetPointerImageToLongString(block);
     return;
   }
-  const textInput = block.inputList[0].fieldRow[0];
```
molly-moen, line 108: almost the exact same code was used here and in `resetPointerImageToLongString`. I pulled it into a helper method `updatePointerImageHelper`.

apps/src/blockly/addons/cdoTrashcan.ts
```diff
@@ -360,13 +397,11 @@ export default class CdoTrashcan extends GoogleBlockly.DeleteArea {
    * @return {boolean} Whether the block or bubble provided should be returned to
    *     drag start.
    */
-  shouldPreventMove(_dragElement) {
+  shouldPreventMove(_dragElement: IDraggable) {
     return false;
   }
 }
 
-CdoTrashcan.TRASH_URL = '/blockly/media/trash.png';
```
molly-moen, line 368: I verified that this was not used anywhere. It was originally used in this class in [this PR](https://github.com/code-dot-org/code-dot-org/pull/37002/files), but as some point we moved to use `this.TRASH_URL`.

apps/src/blockly/addons/cdoVerticalFlyout.ts
```diff
@@ -49,22 +52,22 @@ export default class VerticalFlyout extends GoogleBlockly.VerticalFlyout {
...
+          const oldX = block.getRelativeToSurfaceXY().x;
+          let newX = flyoutWidth / this.workspace_.scale - this.MARGIN;
           if (!block.outputConnection) {
             newX -= this.tabWidth_;
           }
           block.moveBy(newX - oldX, 0);
         }
-        if (block.flyoutRect_) {
-          this.moveRectToBlock_(block.flyoutRect_, block);
+        if (this.rectMap_.has(block)) {
```
molly-moen, line 62: This change comes from the [original blockly function](https://github.com/google/blockly/blob/71185b5582401cd499f0e4a8ed444e7c4d531580/core/flyout_vertical.ts#L361-L363). It has been updated since we copied it over. `block.flyoutRect_` does not exist anymore.

apps/src/blockly/addons/cdoSpritePointer.ts
```diff
@@ -68,53 +75,72 @@ export function updatePointerBlockImage(
...
     // Blocks of type gamelab_allSpritesWithAnimation have an input with one field (the costume
     // picker dropdown).
-    return (
-      targetConnection.sourceBlock_.inputList[0].fieldRow[0].imageElement.getAttribute(
-        'xlink:href'
-      ) || ''
-    );
+    // imageElement is a private property on FieldDropdown.
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const imageElement = (sourceBlock.inputList[0].fieldRow[0] as any)
```
mikeharv: Unrelated to the TS changes, but I think we could replace `inputList[0].fieldRow[0]` with `getField('ANIMATION')`. Would that feel cleaner and/or easier to read?

image

I don't expect this block to change much, but if, for example, we ever added a label before the field ("any [costume]" or something), then this getting the field we want by name would be more reliable.

apps/src/blockly/addons/cdoTouch.js
mikeharv: There was a significant update a few Blockly versions back that changed how touch/mouse/pointer events. In fact, I believe Blockly always uses Pointer now, so it makes sense that this might not be doing anything.

apps/src/blockly/addons/cdoTrashcan.ts
```diff
@@ -159,39 +174,50 @@ export default class CdoTrashcan extends GoogleBlockly.DeleteArea {
...
     if (blocklyEvent.type === Blockly.Events.BLOCK_DRAG) {
+      const event = blocklyEvent as BlockDrag;
       let trashcanVisibility = 'hidden';
       let toolboxVisibility = 'visible';
       // Don't show the trashcan if the block is being dragged out of the toolbox.
-      const isDraggingFromToolbox = !!this.workspace.currentGesture_?.flyout_;
-      if (!isDraggingFromToolbox && blocklyEvent.isStart) {
+      // This property is private, but this is the only way to find out this information.
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      const isDraggingFromToolbox = !!(this.workspace.currentGesture_ as any)
```
mikeharv, line 185: It looks like we might be able to use `getGesture` here: 
image

https://github.com/google/blockly/blob/71185b5582401cd499f0e4a8ed444e7c4d531580/core/workspace_svg.ts#L2471-L2503

apps/src/blockly/addons/cdoTrashcan.ts
```diff
@@ -159,39 +174,50 @@ export default class CdoTrashcan extends GoogleBlockly.DeleteArea {
...
 
       // Shadow blocks can/should be successfully deleted
       // when dragged in conjunction with another deletable block;
       // however, isDeletable() returns false for shadow blocks,
       // so we manually override here.
-      const isDeletable = blocklyEvent.blocks.every(
+      const isDeletable = event.blocks?.every(
         block => block.isDeletable() || block.isShadow()
       );
+      if (!this.notAllowed_) {
```
mikeharv, line 218: The `notAllowed_` svg is created during `init()` -> `this.createTrashcanSvg()`, and this `workspaceChangeHandler` method is added a change listener a few lines earlier in `init()`. Is it possible that changing the order of these two thing in `init()` would help us to guarantee that the SVG exists at this point?

apps/src/blockly/addons/cdoTrashcan.ts
mikeharv: Confirmed that this looks redundant and unused now.
https://github.com/code-dot-org/code-dot-org/commit/b41904699ebd3a321e1cbb9ac419292afd30c778

mikeharv (approved):
A couple suggestions and questions but no blockers!

apps/src/blockly/addons/cdoTrashcan.ts
molly-moen: `currentGesture_` is actually a public field, it's `flyout` that is the private field. `getGesture()` can [change/create](https://github.com/google/blockly/blob/71185b5582401cd499f0e4a8ed444e7c4d531580/core/workspace_svg.ts#L2480) `currentGesture_` and I'm not sure that's what we would want.

apps/src/blockly/addons/cdoTrashcan.ts
molly-moen: I don't think TypeScript is smart enough for that. It will only guarantee things aren't null if we initialize them in the constructor.

apps/src/blockly/addons/cdoSpritePointer.ts
molly-moen: updated!

PR #56908, opened 2024-02-29
P20-746: Fix the `studio_saySpriteChoices` Blockly function

artem-vavilov:
The function `studio_saySpriteChoices` dynamically invokes the function `Studio.saySprite` with varying arguments. 
The invocation fails when any argument includes an apostrophe (`'`), due to the lack of proper escaping.
| Before  | After |
| ------- | ----- |
| before | after |

- jira ticket: [P20-746](https://codedotorg.atlassian.net/browse/P20-746)

### DIFF
--- a/apps/src/studio/blocks.js
+++ b/apps/src/studio/blocks.js
@@ -3208,9 +3208,9 @@ exports.install = function (blockly, blockInstallOptions) {
       this.id +
       "', " +
       (this.getFieldValue('SPRITE') || '0') +
-      ", '" +
-      (this.getFieldValue('VALUE') || ' ') +
-      "');\n"
+      ', ' +
+      JSON.stringify(this.getFieldValue('VALUE') || ' ') +
+      ');\n'
     );
   };
 
--- a/apps/src/studio/blocks.js
+++ b/apps/src/studio/blocks.js
@@ -3202,16 +3202,11 @@ exports.install = function (blockly, blockInstallOptions) {
   };
 
   generator.studio_saySpriteChoices = function () {
+    const sprite = JSON.stringify(String(this.getFieldValue('SPRITE') || '0'));
+    const value = JSON.stringify(String(this.getFieldValue('VALUE') || ' '));
+
     // Generate JavaScript for saying (choices version).
-    return (
-      "Studio.saySprite('block_id_" +
-      this.id +
-      "', " +
-      (this.getFieldValue('SPRITE') || '0') +
-      ', ' +
-      JSON.stringify(String(this.getFieldValue('VALUE') || ' ')) +
-      ');\n'
-    );
+    return `Studio.saySprite('block_id_${this.id}', ${sprite}, ${value});\n`;
   };
 
   generator.studio_saySpriteParams = function () {
--- a/apps/src/studio/blocks.js
+++ b/apps/src/studio/blocks.js
@@ -3209,7 +3209,7 @@ exports.install = function (blockly, blockInstallOptions) {
       "', " +
       (this.getFieldValue('SPRITE') || '0') +
       ', ' +
-      JSON.stringify(this.getFieldValue('VALUE') || ' ') +
+      JSON.stringify(String(this.getFieldValue('VALUE') || ' ')) +
       ');\n'
     );
   };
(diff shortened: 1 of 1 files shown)


wilkie (approved):
This is a good change. The only potential problem is if `this.getFieldValue` returns an integer or something since it will no longer be a double-quoted string. I think the field value is always a string... is that right? If so, or it won't be a problem... LGTM!

apps/src/studio/blocks.js
```diff
@@ -3208,9 +3208,9 @@ exports.install = function (blockly, blockInstallOptions) {
       this.id +
       "', " +
       (this.getFieldValue('SPRITE') || '0') +
-      ", '" +
-      (this.getFieldValue('VALUE') || ' ') +
-      "');\n"
+      ', ' +
+      JSON.stringify(this.getFieldValue('VALUE') || ' ') +
+      ');\n'
```
mgc1194: I see some of `studio_saySpriteChoices` still uses double quotes (`"`), should we cleanup the rest of the method?
```
"Studio.saySprite('block_id_" +
      this.id +
      "', " +
      (this.getFieldValue('SPRITE') || '0') +
```

artem-vavilov: I believe it's fine if I understand you correctly.

Screenshot 2024-02-29 at 20 27 45

apps/src/studio/blocks.js
```diff
@@ -3208,9 +3208,9 @@ exports.install = function (blockly, blockInstallOptions) {
       this.id +
       "', " +
       (this.getFieldValue('SPRITE') || '0') +
-      ", '" +
-      (this.getFieldValue('VALUE') || ' ') +
-      "');\n"
```
daynew, line 3213: Would this be more readable?
```
  generator.studio_saySpriteChoices = function () {
    const sprite = this.getFieldValue('SPRITE') || '0';
    const value = JSON.stringify(String(this.getFieldValue('VALUE') || ' '));
    // Generate JavaScript for saying (choices version).
    return `Studio.saySprite('block_id_${this.id}', '${sprite}', '${value}');\n`;
  };
```

apps/src/studio/blocks.js
wilkie: I noticed this too and think we should when it makes sense. However, these often have a single quote inside of them which does mean it should probably use a `"` (double quote) to enclose it. The linter _should_ catch any that _should_ be changed.

apps/src/studio/blocks.js
wilkie: The `value` would have the quotes already. The fix would require, perhaps confusingly, but no more than the original change: 

```
  generator.studio_saySpriteChoices = function () {
    const sprite = this.getFieldValue('SPRITE') || '0';
    const value = JSON.stringify(String(this.getFieldValue('VALUE') || ' '));
    // Generate JavaScript for saying (choices version).
    return `Studio.saySprite('block_id_${this.id}', '${sprite}', ${value});\n`;
  };
```

apps/src/studio/blocks.js
```diff
@@ -3202,16 +3202,11 @@ exports.install = function (blockly, blockInstallOptions) {
...
-    return (
-      "Studio.saySprite('block_id_" +
-      this.id +
-      "', " +
-      (this.getFieldValue('SPRITE') || '0') +
-      ', ' +
-      JSON.stringify(String(this.getFieldValue('VALUE') || ' ')) +
-      ');\n'
-    );
+    return `Studio.saySprite('block_id_${this.id}', ${sprite}, ${value});\n`;
```
wilkie, line 3209: Interestingly the `sprite` has the opposite behavior as value... looks like the `SPRITE` value was actually allowed to be an integer and not a string. I think the `spriteIndex` in `saySprite` needs to be an integer.

apps/src/studio/blocks.js
```diff
@@ -3202,7 +3202,7 @@ exports.install = function (blockly, blockInstallOptions) {
   };
 
   generator.studio_saySpriteChoices = function () {
-    const sprite = JSON.stringify(String(this.getFieldValue('SPRITE') || '0'));
+    const sprite = this.getFieldValue('SPRITE') || '0';
```
wilkie, line 3205: Looks good!

wilkie:
My approval stands. Looks good. I also like the slight refactor. Thanks!

PR #56821, opened 2024-02-26
Music: add sampler playback support

sanchitmalhotra126:
Part 4 of adding ToneJS player support, follow-up to https://github.com/code-dot-org/code-dot-org/pull/56807. This adds support for using samplers to play instrument sounds (notes/beats) which will be used by the ToneJS player when it is available.

## Links

https://codedotorg.atlassian.net/browse/LABS-513

## Testing story

Tested on /projectbeats, /music-intro-2024, standalone projects

PR #56912, opened 2024-02-29
Music: enable ToneJS player behind url param

sanchitmalhotra126:
Last part (for now) for adding ToneJS player support (follow-up to https://github.com/code-dot-org/code-dot-org/pull/56821). This enables the ToneJS player behind a url param. There are still a number of follow-up tasks (plus testing) we should finish before formally switching over, but this gives us the option to start trying it out in production. One significant callout here is that, as a temporary workaround, we're loading all samples for all instruments up-front to pre-create all the samplers in the ToneJS player. Instead, we should load instrument samples and create samplers on-demand as those blocks are added to a project. I held off on that for now since it was proving to be a little tricky with our current caching logic, and required some additional UI work to show some loading UI on the chord and drum panels. This work (and others) have been captured as follow-ups in JIRAs below.

To enable the ToneJS player, include `?player=tonejs` in the URL. Can confirm that this is working by looking for `[MusicPlayer] using ToneJSPlayer` in the console.

## Links

https://codedotorg.atlassian.net/browse/LABS-513

## Testing story

Tested on /projectbeats, /music-intro-2024, and standalone projects.

## Follow-up work

- Add effects support: https://codedotorg.atlassian.net/browse/LABS-528
- Add library-defined sound sequence support: https://codedotorg.atlassian.net/browse/LABS-529
- Add "advanced controls" panel behind URL param: https://codedotorg.atlassian.net/browse/LABS-530
- Load instruments on demand instead of all at once: https://codedotorg.atlassian.net/browse/LABS-531

### DIFF
--- a/apps/src/music/player/MusicPlayer.ts
+++ b/apps/src/music/player/MusicPlayer.ts
@@ -16,6 +16,8 @@ import Lab2Registry from '@cdo/apps/lab2/Lab2Registry';
 import {LoadFinishedCallback, UpdateLoadProgressCallback} from '../types';
 import {AudioPlayer, SampleEvent, SamplerSequence} from './types';
 import SamplePlayerWrapper from './SamplePlayerWrapper';
+import appConfig from '../appConfig';
+import ToneJSPlayer from './ToneJSPlayer';
 
 // Using require() to import JS in TS files
 const constants = require('../constants');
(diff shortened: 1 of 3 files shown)


breville (approved):
This has been a great journey.

Could you add the new `player` flag to `/musiclab/menu`?

sanchitmalhotra126:
> Could you add the new `player` flag to `/musiclab/menu`?

Done! Also updated the path in the menu from `/musiclab` to `/projectbeats` for easier copying.

PR #56916, opened 2024-02-29
Lab2: convert disable_projects to uses_projects

sanchitmalhotra126:
Change the opt-out `disable_projects` flag on lab2 labs to an opt-in `uses_projects` flag that can be computed from `is_project_level || channel_backed`. We've had a desire to update this for a while, but this recently resurfaced because I noticed we were getting 404s on the new "panels" level type in the music lab intro script, because we forgot the `disable_projects` flag there. With this change, most labs shouldn't need to specify if they support projects or not, because we can infer that from other properties. This is especially helpful, for example, with Dance Party, where certain levels (free play, standalone) use projects, while others don't.

**Question for @molly-moen / code-tools**: currently, pythonlab and weblab2 are not in the "always channel backed" list [here](https://github.com/code-dot-org/code-dot-org/blob/4c3897673c20d615f952bc5a604bcc2e6eded7c4/dashboard/app/models/game.rb#L277) - should they be? With this change, we would stop loading projects for those level types, but I wasn't sure if we've actually started using projects on these levels or not.

## Links

https://codedotorg.atlassian.net/browse/LABS-177

## Testing story

Tested this across all Lab2 labs -
- AIChat - always `false` since there are no standalone levels and for now projects are not supported (this will change though)
- Dance - `true` if free play or standalone, `false` otherwise
- Music - always `true`
- Panels - always `false`
- PythonLab - always `false` because it's currently not part of the `channel_backed` list in `game.rb` (but will update if that's not right)
- Weblab2 - always `false`, same as above
- StandaloneVideo - always `false`

### DIFF
--- a/apps/src/lab2/lab2Redux.ts
+++ b/apps/src/lab2/lab2Redux.ts
@@ -106,9 +106,9 @@ export const setUpWithLevel = createAsyncThunk(
         .getMetricsReporter()
         .updateProperties({appName: levelProperties.appName});
 
-      const {isProjectLevel, disableProjects} = levelProperties;
+      const {isProjectLevel, usesProjects} = levelProperties;
 
-      if (disableProjects) {
+      if (!usesProjects) {
         // If projects are disabled on this level, we can skip loading projects data.
         setProjectAndLevelData(
           {levelProperties},
--- a/apps/src/lab2/types.ts
+++ b/apps/src/lab2/types.ts
@@ -92,11 +92,7 @@ export interface LevelProperties {
   // Not a complete list; add properties as needed.
   isProjectLevel?: boolean;
   hideShareAndRemix?: boolean;
-  // TODO: Rework this field into an "enableProjects" or more complex list of
-  // "enabledFeatures" that is calculated on the back end. For now, since
-  // the only labs we support have projects enabled, it's easier to make this a
-  // disabled flag for specific exceptions.
-  disableProjects?: boolean;
+  usesProjects?: boolean;
   levelData?: LevelData;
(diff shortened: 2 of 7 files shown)


molly-moen:
> Currently, pythonlab and weblab2 are not in the "always channel backed" list [here](https://github.com/code-dot-org/code-dot-org/blob/4c3897673c20d615f952bc5a604bcc2e6eded7c4/dashboard/app/models/game.rb#L277) - should they be?

@sanchitmalhotra126 nope, we haven't set up projects yet for those level types. With this change, would we just need to add those labs to the list to get projects working?

sanchitmalhotra126:
> > Currently, pythonlab and weblab2 are not in the "always channel backed" list [here](https://github.com/code-dot-org/code-dot-org/blob/4c3897673c20d615f952bc5a604bcc2e6eded7c4/dashboard/app/models/game.rb#L277) - should they be?
> 
> @sanchitmalhotra126 nope, we haven't set up projects yet for those level types. With this change, would we just need to add those labs to the list to get projects working?

Yep! Once they're added to that list, we'll try to load projects for them. We can also update the channel_backed? method [here](https://github.com/code-dot-org/code-dot-org/blob/f3640be7611e7b09e872987352ba4497574a9374/dashboard/app/models/levels/level.rb#L400) if it needs to be a more custom computation (like `dance_party_free_play` for example)

bencodeorg (approved):

breville (approved):
Thanks for catching this issue and fixing!

This code reads nicely now that it's an "opt in" versus an "opt out" flag.

PR #56918, opened 2024-02-29
Part 3: Add information banner about updated public gallery for signed-in users

fisher-alice:
This PR adds an information banner about the public gallery update (only featured projects displayed) and that users will no longer be able to publish their projects. This banner will be displayed only for signed-in users.
The plan is to post this banner when the public gallery is updated and leave the banner up for ~ 4 weeks.

This is part 3 of 4 PRs that update the public gallery.
Part 1: Remove ability to publish projects https://github.com/code-dot-org/code-dot-org/pull/56666
Part 2: Update projects public gallery to display only featured projects https://github.com/code-dot-org/code-dot-org/pull/56886
Part 3: Add information banner about updated public gallery for signed-in users https://github.com/code-dot-org/code-dot-org/pull/56918
Part 4: Notify user if their project is selected to be featured since the project will be frozen https://github.com/code-dot-org/code-dot-org/pull/56932

After update:
Screenshot 2024-02-29 at 3 51 08 PM

Screencast of update:

https://github.com/code-dot-org/code-dot-org/assets/107423305/5f8a763c-3848-45fb-9874-b6532c796dd8

## PR Checklist:

- [ ] Code is well-commented


dashboard/app/views/projects/index.html.haml
```diff
@@ -11,6 +11,7 @@
   end
 
   projects_data[:specialAnnouncement] = Announcements.get_announcement_for_page("/projects")
+  projects_data[:showPublicGalleryUpdatedInfo] = current_user
```
bencodeorg: Is this supposed to be `!!current_user`? Or, maybe simpler, just grab `pageConstants.isSignedIn` from redux instead of having to pass this thru? I guess this is probably temporary / will be reverted, so whatever's easier.

bencodeorg (approved):
One cleanup worth doing I think before merging, but otherwise LGTM

dashboard/app/views/projects/index.html.haml
fisher-alice: Thanks! I don't think `pageConstants` are set when at '/projects' since they're set during level init.
https://github.com/code-dot-org/code-dot-org/blob/0e810f695ed3d979224b2e242865340c65d04bf6/apps/src/redux/pageConstants.js#L119

I updated to `!!current_user` here. Thanks!

fisher-alice: Actually, I think I could use `isSignedIn` from `currentUserRedux`. `PersonalProjectsTable` uses this value.
Since this is temporary, I think I'll leave as is, but just wanted to leave a note here. Thanks!

bencodeorg: Good to know, thanks for noting! What's available where in Redux is always hard for me to remember :)

PR #56925, opened 2024-03-01
[Design2-52] DSCO Button component

levadadenys:
### [Design2-52] DSCO Button component

* Created `Button` and `LinkButton` components
* created stories for `Button` and `LinkButton` components
* created tests for `_BaseButton` component
* created documentation for `Button`, `LinkButton`, `_BaseButton` component

`_BaseButton` is a private component for Designs System use only. It provides a base for `Button`
and `LinkButton` components, implementing all of the logic and styles for them.

`Button`:

https://github.com/code-dot-org/code-dot-org/assets/22244040/70f68d60-441a-44bd-a7a9-78872047e583

`LinkButton`:

https://github.com/code-dot-org/code-dot-org/assets/22244040/d588ce10-26dc-4b1d-974e-bf714ef56564

`_BaseButton`:

https://github.com/code-dot-org/code-dot-org/assets/22244040/79d976de-4cb9-4404-8085-2cf168d0b73e

## Links
- jira ticket: [DESIGN2-52](https://codedotorg.atlassian.net/browse/DESIGN2-52)

## PR Checklist:

- [ ] Code is well-commented


apps/src/componentLibrary/button/Button.tsx
```diff
@@ -1,36 +1,250 @@
...
+
+type ButtonColor = 'purple' | 'black' | 'white';
+
+export const buttonColors: {[key in ButtonColor]: ButtonColor} = {
+  purple: 'purple',
+  black: 'black',
+  white: 'white',
+};
+
 export interface ButtonProps {
```
sanchitmalhotra126: There are a lot of props here with a lot of underlying assumptions about which props are intended for which purpose. What do you think about exporting separate components, ex. `LinkButton`, `IconButton`, and `Button`/`TextButton` (for text + icons left/right) for the different variants, where we can ensure that the right props are being passed? Then we could turn this component into an internal-only component (`ButtonInternal` or `BaseButton` or maybe `CustomButton` to match the dropdown naming)

apps/src/componentLibrary/button/Button.tsx
```diff
@@ -1,36 +1,250 @@
...
+      aria-disabled={disabled}
+      aria-label={ariaLabel}
+      {...tagSpecificProps}
+    >
+      {isPending && spinnerPosition === 'left' && (
+        <FontAwesomeV6Icon {...spinnerIcon} />
+      )}
+      {showIconLeft && <FontAwesomeV6Icon {...iconLeft} />}
+      {showIcon && <FontAwesomeV6Icon {...icon} />}
+      {text && <span>{text}</span>}
```
sanchitmalhotra126: Out of curiosity - why are we using CSS to hide the text span when showing the loading spinner, but using conditionals to hide the icons?

sanchitmalhotra126: Could we in theory have an iconLeft, iconRight, icon, and text all at once? And if so is that desired? I'd imagine we'd probably only want either `icon` or `text` but not both right?

apps/src/componentLibrary/button/Button.tsx
```diff
@@ -1,36 +1,250 @@
...
+    [type, icon, useAsLink, buttonType, onClick, href, download]
+  );
+
+  /** Handling isPending state content & spinner show logic here.
+     - If there's only text - we show only spinner.
+     - If there's only icon - we show only spinner.
+     - If there's text and iconLeft or both iconLeft and iconRight -> we show spinner on the left + text + iconRight (if it's present).
+     - If there's text and iconRight - we show text + spinner on the right.
+     */
+  const showIcon = icon && !isPending;
```
sanchitmalhotra126: One thought - what do you think about turning this into a function that outputs a list of JSX elements? If i'm understanding this correctly, we'd always have 3 elements at most right? Could we do something like
```
const elements = useMemo(() => {
   const leftElement = isPending ? ...
   const middleElement = ...
   const rightElement = ...
   return [leftElement, middleElement, rightElement];
})
```

apps/src/componentLibrary/button/Button.tsx
levadadenys: It's for the case when we have only text in the button  - for the pending state we want to hide the text and only show spinner. We hide text with css in order to keep button sizes the same when we hide the text (if we don't render text as hidden for pending state - button size will be changed)

For render/not render icons - size of the buttons will remain the same cause pending icon will simply replace the not rendered icon

levadadenys: Correct, we want either want `icon` or `text` but not both.
it's being validated In `checkButtonPropsForErrors` function (validation can be extended)

```
  if (type !== 'iconOnly' && type !== 'iconBorder' && icon) {
    throw new Error(
      'Expect icon prop to be undefined when type is not iconOnly or iconBorder'
    );
  }
```

apps/src/componentLibrary/button/Button.tsx
levadadenys: Correct, good idea, thanks!
Will investigate. Trying to achieve balance between cleanness and readability, not sure if it'll be more readable/understandable based on our logic complexity

apps/src/componentLibrary/button/Button.tsx
levadadenys: Yeah, that's an option I think. Not sure though, since it might bring additional extra complexity to the code and will also require additional tests and stories.

The initial reason I decided to implement Button as is right now - following the skeleton of our current `Button` component, meaning transition to new `Button` component will be smother.

Also, with `Dropdown` the logic behind was that every dropdown had some differences visually and structurally, but here with `Button` all three types visually and structurally are almost identical.

Also, additional validation is provided here with help of `checkButtonPropsForErrors` function that can be extended as much as we want/need.

Finally, don't have a strong/leaning towards opinion right now, need to think what will be more convinient for future development and better consumer experience

apps/src/componentLibrary/button/Button.tsx
```diff
@@ -1,36 +1,82 @@
...
+  iconLeft?: FontAwesomeV6IconProps;
+  /** Button icon (When used in IconOnly mode)*/
+  icon?: FontAwesomeV6IconProps;
+  /** Left Button icon */
+  iconRight?: FontAwesomeV6IconProps;
+  /** Whether we use \<a> (when set to true) or \<button> (when false) html tag for Button component.
+   * If we want button to redirect to another page or download some file we should use \<a> tag.
+   * If we want button to call some function or submit some form we should use \<button> tag.
+   * */
+  useAsLink?: boolean;
```
sanchitmalhotra126: We can remove this prop right? Shouldn't all links use `LinkButton`?

apps/src/componentLibrary/button/types.ts
```diff
@@ -0,0 +1,8 @@
+export type ButtonType =
+  | 'primary'
+  | 'secondary'
+  | 'tertiary'
+  | 'iconBorder'
+  | 'iconOnly';
```
sanchitmalhotra126: Should these be two different dimensions actually (`primary | secondary | tertiary` and `icon | text`)? Couldn't primary/secondary apply to both text buttons and icon buttons? Looking at the storybook, it looks we use iconBorder and iconOnly to set the border for icon buttons, but could we accomplish that with primary/secondary/tertiary instead?

apps/src/componentLibrary/button/Button.tsx
```diff
@@ -1,36 +1,82 @@
...
+  /** (\<button> specific prop)
+   * Button html element type */
+  buttonType?: 'submit' | 'button';
```
sanchitmalhotra126: nit: I feel like the word `type` is getting a bit overloaded in these props. Maybe `tagType` or `buttonTag` for more specificity?

apps/src/componentLibrary/button/_baseButton/_BaseButton.tsx
```diff
@@ -0,0 +1,248 @@
...
+
+import {ButtonType, ButtonColor} from '@cdo/apps/componentLibrary/button';
+import FontAwesomeV6Icon, {
+  FontAwesomeV6IconProps,
+} from '@cdo/apps/componentLibrary/fontAwesomeV6Icon';
+import {ComponentSizeXSToL} from '@cdo/apps/componentLibrary/common/types';
+
+import moduleStyles from './_baseButton.module.scss';
+
+export interface _BaseButtonProps {
```
sanchitmalhotra126: I was thinking a little more about how to organize these props and enforce some of these contracts through typescript...here's an idea if you'd like to do something similar:

```
// First, create the common props and extensions for each Button variant

interface CommonButtonProps {
  className?: string;
  id: string;
  disabled?: boolean;
  isPending?: boolean;
  // ... others
}

interface LinkButtonProps extends CommonButtonProps {
  href: string;
  download?: string | boolean;
  // ... others
}

interface ButtonProps extends CommonButtonProps {
  onClick: () => void;
  buttonType?: 'submit' | 'button';
  // ... others
}

// Then, we can create separate interfaces to distinguish between buttons with text and buttons with an icon only

interface IconProps {
  icon: FontAwesomeV6IconProps;
}

interface TextProps {
  text: string;
  iconLeft?: FontAwesomeV6IconProps;
  iconRight?: FontAwesomeV6IconProps;
}

// We can use this XOR utility (found on SO, link below) to make sure that  the combined props use text or an icon but not both

type Without<T, U> = {[P in Exclude<keyof T, keyof U>]?: never};
type XOR<T, U> = T | U extends object
  ? (Without<T, U> & U) | (Without<U, T> & T)
  : T | U;

type CombinedButtonProps<T> = XOR<T & IconProps, T & TextProps>;

// Finally, create our Button variants

const Button: React.FunctionComponent<CombinedButtonProps<ButtonProps>> = props => { ... };

const LinkButton: React.FunctionComponent<CombinedButtonProps<LinkButtonProps>> = props => { ... };


// Now TS should be able to check at compile time that we're using the right prop combinations:


// X Doesn't compile - missing text and/or icon
<Button id="id" onClick={() => {}} />;

// ✓ Compiles - using only icon
<Button id="id" onClick={() => {}} icon={{iconName: 'icon'}} />;

// ✓ Compiles - using only text
<Button id="id" onClick={() => {}} text="text" />;

// X Doesn't compile - using both text and icon
<Button id="id" onClick={() => {}} icon={{iconName: 'icon'}} text="text" />;

// X Doesn't compile - using iconLeft (text only prop) and icon
<Button id="id" onClick={() => {}} icon={{iconName: 'icon'}} iconLeft={{iconName: 'icon'}} />;


// X Doesn't compile - missing text and/or icon
<LinkButton id="id" href="href" />;

// ✓ Compiles - using only icon
<LinkButton id="id" href="href" icon={{iconName: 'icon'}} />;

// ✓ Compiles - using only text
<LinkButton id="id" href="href" text="text" />;

// X Doesn't compile - using both text and icon
<LinkButton id="id" href="href" icon={{iconName: 'icon'}} text="text" />;

// X Doesn't compile - using iconRight (text only prop) and icon
<LinkButton id="id" href="href" icon={{iconName: 'icon'}} iconRight={{iconName: 'icon'}} />;
```

XOR type found on SO: https://stackoverflow.com/questions/42123407/does-typescript-support-mutually-exclusive-types

sanchitmalhotra126 (commented):
Thanks for splitting up the Button variants - I think that'll be clearer and easier to use in the long run. Had a few more minor comments plus an idea for enforcing some of the prop combinations through typescript.

apps/src/componentLibrary/button/types.ts
levadadenys: From what I've seen in Figma:
`'primary' | 'secondary' | 'tertiary' ` - can be text
`'iconBorder' | 'iconOnly';` - can be icon only

Will doublecheck with design team though, good question, thank you!

apps/src/componentLibrary/button/Button.tsx
levadadenys: Not sure about this one since it specifies `type` attribute of `<button>` element as documented.
`buttonTagType` or even `buttonTagButtonType` might be selfdocumenting names, though I think they're kinda too much 😄 
`tagType` or `buttonTag` might lead to even more confusion I think since that's not a tag prop and not related to a link button

apps/src/componentLibrary/button/_baseButton/_BaseButton.tsx
levadadenys: Great idea and structure, thank you!

Though, not sure that will help, since I've tried XOR logic before and that lead to types collisions in BaseButton still.
Might work for `Button` and `LinkButton`, but will still cause typerrors with `_BaseButton` probably

levadadenys: image
Checked locally, unfortunately it didn't help and caused type errors in `_BaseButton` component again. (removing 1 error raises other errors)

I've reorganized the structure of types to be reusable and better documented, I think it might be better to merge as is

sanchitmalhotra126 (approved):
LGTM, thanks for the back and forth on this!

PR #56926, opened 2024-03-01
[ProgressV2] Update script to use id instead of email

lfryemason:
I realized we wouldn't want to upload a file with emails and switched the script to take a csv file with user IDs instead. This also only adds the flag to teacher accounts and now prints how many accounts were added to the closed beta.

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/bin/oneoff/add_progress_v2_closed_beta_users.rb
+++ b/bin/oneoff/add_progress_v2_closed_beta_users.rb
@@ -4,16 +4,23 @@ require_relative '../../../dashboard/config/environment'
 require 'csv'
 
 # Run using `ruby add_progress_v2_closed_beta_users.rb <path_to_csv>`
-# For a given csv file, find all users by email and update their `progress_table_v2_closed_beta` user preference to true.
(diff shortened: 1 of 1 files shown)


bin/oneoff/add_progress_v2_closed_beta_users.rb
```diff
@@ -4,16 +4,23 @@
...
+unless teacher_id_csv
+  puts 'Usage: add_progress_v2_closed_beta_users <teacher_id_csv>'
   exit
 end
 
-CSV.table(email_csv).each do |row|
-  User.where(email: row[:email]).map {|u| u.update!(progress_table_v2_closed_beta: true)}
+count = 0
+CSV.table(teacher_id_csv).each do |row|
+  user = User.find(id: row[:id])
```
etaderhold: I believe `find` always takes an ID or list of IDs, while `find_by` is what you would use with named fields. Also `find` raises an exception if the record isn't found, which you should probably handle here.

```suggestion
  user = User.find(row[:id])
```

etaderhold (approved):

PR #56931, opened 2024-03-01
Refactor 404 page on code.org 

kelbyhawn:
Updates the 404 page on https://code.org to fix top spacing and use rebrand markup.

**Jira ticket:** [ACQ-1566](https://codedotorg.atlassian.net/browse/ACQ-1566)

---- 

## Before
Before

## After
After


pegasus/sites.v3/code.org/views/page_not_available.haml
kelbyhawn, line 1: I don't know where/how this is used, but I updated it to match the `404.haml` view. The only place I found it in the code outside of translation files is [/pegasus/routes/poste_routes.rb#L20](https://github.com/code-dot-org/code-dot-org/blob/dc69e27d9cc782e4a3168ded08ff31100b4f4795/pegasus/routes/poste_routes.rb#L20). If anyone knows if this can be deleted, pls let me know!

pegasus/sites.v3/code.org/styles_min/040-page.css
```diff
@@ -406,6 +406,11 @@ td {
   padding: 15px;
 }
 
+/*Centers a block-level element */
+.centered {
+  text-align: center !important;
+}
+
```
kelbyhawn, line 413: I moved this here so it can be used on legacy pages too.

pegasus/sites.v3/code.org/views/page_not_available.haml
TurnerRiley: I think if it's still being used somewhere, then we should keep it unless it's identical to the other page and can be replaced!

TurnerRiley (approved):

pegasus/sites.v3/code.org/views/page_not_available.haml
kelbyhawn: It looks like that `poste_routes.rb` file has something to do with the Dashboard Pegasus decoupling: [Slack convo](https://codedotorg.slack.com/archives/C03CK49G9/p1672862384027409?thread_ts=1672861260.350699&cid=C03CK49G9) — I'll keep it for now since I don't want to potentially break something.

PR #56753, opened 2024-02-22
Refactor TwoColumnActionBlock component

kelbyhawn:
Refactors the `TwoColumnActionBlock` component that's used on various pages on https://studio.code.org to look more like the ones used on https://code.org.
- Refactored to use a functional component
- Removed the following outdated components in the same file and the `CoursesTeacherEnglish.jsx` component: 
  - `AdministratorResourcesActionBlock`
  - `CscInfoActionBlock`
  - `SpecialAnnouncementActionBlock`
- Uses new Typography components and removes unused props
- Replaces a lot of styling logic with CSS
  - Works with LTR and RTL languages
  - Fully responsive (keeps the current UI pattern that hides the image at smaller screens)
- Added container padding to allow the image to keep it's aspect ratio if the text next to it varies in length due to translations

**Jira ticket:** [ACQ-1527](https://codedotorg.atlassian.net/browse/ACQ-1527)

----

### Before:
Before

### After:
After

### Before
Before-2

### After
After-2

## Responsive 

https://github.com/code-dot-org/code-dot-org/assets/9256643/cb1d7283-ae36-4b7f-bd91-a01d6ff94079

## RTL
Screenshot 2024-02-26 at 12 02 33 PM

## Updated Storybook
Screenshot 2024-02-27 at 10 00 59 AM

PR #56934, opened 2024-03-01
Fix Safari image bug on TwoColumnActionBlock component

kelbyhawn:
Makes the image on the TwoColumnActionBlock component not stretched out on Safari! This was my bad for not cross-browser checking [this PR](https://github.com/code-dot-org/code-dot-org/pull/56753) 🤦‍♀️ I checked it in Chrome, Safari, and Firefox this time and it looks good.

**Jira ticket:** [ACQ-1567](https://codedotorg.atlassian.net/browse/ACQ-1567)

---- 

## Before
Before

## After
After

### DIFF
--- a/apps/src/templates/studioHomepages/twoColumnActionBlock.module.scss
+++ b/apps/src/templates/studioHomepages/twoColumnActionBlock.module.scss
@@ -22,7 +22,6 @@ $breakpoint: 993px;
   .image {
     margin: 1.25rem;
     width: calc(50% - 1.25rem);
-    height: fit-content;
 
     @media (max-width: $breakpoint) {
       display: none;


TurnerRiley (approved):

apps/src/templates/studioHomepages/twoColumnActionBlock.module.scss
```diff
@@ -22,7 +22,6 @@ $breakpoint: 993px;
   .image {
     margin: 1.25rem;
     width: calc(50% - 1.25rem);
-    height: fit-content;
```
hannahbergam, line 25: omg good find. Such a frustrating bug!

hannahbergam (approved):
🎉

apps/src/templates/studioHomepages/twoColumnActionBlock.module.scss
kelbyhawn: Ack I know!! I forgot to cross-browser check 🙃

PR #29597, opened 2019-07-10
Delete unused log build times from apps 

Erin007:
While upgrading yarn I saw this message: 
Screen Shot 2019-07-10 at 1 06 14 PM
Given that Paul no longer works at Code.org, I [asked](https://codedotorg.slack.com/archives/C0T0PNR0D/p1562777338099100) whether we still need the message. 

Turns out we don't really use `logBuildTimes` as is any more. So I deleted it.

PR #8294, opened 2016-05-09
Log build stats to new relic

pcardune:
If you add `new_relic_license_key: <your key here>` to the `locals.yml`
file in the root directory, build times will be logged to new relic. If
you do not have this in locals.yml, build times will be logged
locally. If you later add the license key, then all locally logged build
times will be uploaded to new relic as well.

PR #56938, opened 2024-03-01
Removed unused newrelic package

jmkulwik:

The newrelic npm package was [added a while back](https://github.com/code-dot-org/code-dot-org/pull/8294) in order to track our build times. The usage was [removed a few years ago](https://github.com/code-dot-org/code-dot-org/pull/29597), but the npm package and one of the scripts were missed in the cleanup. This just does a little more cleanup to clear those out. This shouldn't impact our production monitoring via newrelic - from my investigation that's added to the window object from the backend and this package was only used for the specific scenario mentioned earlier - however, @sureshc, I could definitely use a second pair of eyes to validate that understanding.

This was discovered using [depcheck](https://www.npmjs.com/package/depcheck)


Erin007 (approved):
Good find! Thanks for clearing out the rest of these long-forgotten tendrils. 🧹

davidsbailey (approved):
🚮 💯

jmkulwik:
@sanchitmalhotra126 - I think you might be the right person to tag - Suresh said he only knows about how the backend NewRelic logging works. Can you take a quick look and let me know if I'm overlooking anything from your perspective? I couldn't find any usage of the newrelic NPM package and am pretty sure that we use the ruby gem to set up our front end monitoring. But could use a second pair of eyes.

sanchitmalhotra126:
> @sanchitmalhotra126 - I think you might be the right person to tag - Suresh said he only knows about how the backend NewRelic logging works. Can you take a quick look and let me know if I'm overlooking anything from your perspective? I couldn't find any usage of the newrelic NPM package and am pretty sure that we use the ruby gem to set up our front end monitoring. But could use a second pair of eyes.

Yeah, this sounds right to me - I'm only aware of [logToCloud.js](https://github.com/code-dot-org/code-dot-org/blob/staging/apps/src/logToCloud.js) that reports to New Relic from the front end, and that uses the newrelic object on the window as you mentioned.

PR #56942, opened 2024-03-01
Filter out curricula user has already taught for Similar Curriculum Recommender

TurnerRiley:
This PR updates the Similar Curriculum Recommender to not recommend any curricula the user has taught before (if they are a signed-in teacher with active/past sections).

## Demo
For a signed-out user, Course B's recommended similar curriculum is Course A:
signedOut

However, for a signed-in teacher who has taught Course A before, Course B's recommended similar curriculum is Course C:
signedInTaught

## Links
Jira ticket: [here](https://codedotorg.atlassian.net/jira/software/c/projects/ACQ/boards/64?assignee=60d62161f65054006980bd71&selectedIssue=ACQ-1379)

## Testing story
Local testing and adding tests.


apps/test/unit/templates/curriculumCatalog/CurriculumCatalogTest.jsx
```diff
@@ -690,5 +691,58 @@ describe('CurriculumCatalog', () => {
...
+        let recommendedSimilarCurriculum = similarCurriculumRecommendations[0];
+        if (
+          curriculaTaughtBefore.includes(
+            recommendedSimilarCurriculum.course_offering_id
+          )
+        ) {
+          recommendedSimilarCurriculum = similarCurriculumRecommendations[1];
+        }
+
+        screen.getByAltText(recommendedSimilarCurriculum.display_name); //Image's alt text is the curriculum's display name
```
bethanyaconnor: nit: put the comment above this line

apps/test/unit/util/curriculumRecommenderTestCurricula.js
```diff
@@ -292,6 +296,7 @@ const FULL_TEST_COURSES = [
   },
   {
     key: 'fullTestCourse5',
+    course_offering_id: 5,
```
bethanyaconnor, line 299: Are these already present on the objects used by the components? (ie outside of the tests)

dashboard/app/controllers/curriculum_catalog_controller.rb
```diff
@@ -23,7 +23,8 @@ def index
       isSignedOut: @is_signed_out,
       isTeacher: @is_teacher,
       sections: @sections_for_teacher,
-      isInUS: request.country.to_s.casecmp?('rd') || request.country.to_s.casecmp?('us')
+      isInUS: request.country.to_s.casecmp?('rd') || request.country.to_s.casecmp?('us'),
+      curriculaTaught: current_user&.sections&.map {|s| s.summarize_without_students[:course_offering_id]}
```
bethanyaconnor: That summarize method is doing quite a lot of things you don't need here, so I wonder if it would be better to extract the logic in `summarize_without_students` that fetches the course offering id into its own method then use it both in that method and here.

So, said another way, in `section.rb`, have a function that looks like: (feel free to change the name, etc)

```
def course_offering_for_assignment
  unit_group ? unit_group&.course_version&.course_offering : script&.course_version&.course_offering
end
```

And then use this function both here and in `summarize_without_students`. Thoughts?

TurnerRiley: Good call! On it!

apps/test/unit/util/curriculumRecommenderTestCurricula.js
TurnerRiley: Yep! As part of the `summarize_for_catalog` method in `course_offering.rb`, the [curricula data](https://github.com/code-dot-org/code-dot-org/blob/3345d7cdb3050c04a63f19b11a12c829ec190a3e/dashboard/app/models/course_offering.rb#L328) includes `course_offering_id`.

TurnerRiley: I'll also re-test everything to make sure it works as expected!

apps/test/unit/templates/curriculumCatalog/CurriculumCatalogTest.jsx
```diff
@@ -671,15 +672,62 @@ describe('CurriculumCatalog', () => {
...
         )[0];
 
         // Open expanded card of the current test curriculum
         fireEvent.click(quickViewButtons[i]);
         screen.getByText(currCurriculum.description);
 
         // Check that the recommended similar curriculum's image and link are present on the current test curriculum's expanded card.
-        //Image's alt text is the curriculum's display name.
+        // Image's alt text is the curriculum's display name.
+        screen.getByAltText(recommendedSimilarCurriculum.display_name);
```
bethanyaconnor, line 685: prefacing this with the fact that I'm learning RTL, but can you use `screen.getByRole('img', {name: recommendedSimilarCurriculum.display_name})` here instead? Also, my understanding was that `ByRole` was preferable but that might be wrong!

bethanyaconnor (approved):
one RTL question but otherwise looks good!

apps/test/unit/templates/curriculumCatalog/CurriculumCatalogTest.jsx
TurnerRiley: I'm not sure it will work becasue I think it will only find it by name if the image tag has no 'alt' tag but the image needs an 'alt' tag to be accessible. I'm also learning RTL so I might be wrong!

bethanyaconnor: oh interesting! good to know!

PR #56943, opened 2024-03-01
Add basic generic Tabs component and AI customizaton panel

bencodeorg:
Adds a basic version of a tabs component inside of a new panel to house our AI customization UI.

I've tried to follow the basics outlined in [ARIA standards for accessible tabs](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/), but there's definitely follow-up work to do here (see follow-up section).

I'm hoping this can be a starting point for adding it to the design system while unblocking some of our initial work in our generative AI lab. I've left the tabs components in our `aichat` directory for now so that no one sees it in our `componentLibrary` directory and thinks that it's ready for broader use. We can definitely move into that directory once it's firmed up.

image

## Testing story

Not much to test here yet, but can navigate between tabs and see the appropriate content rendered.

 ## Follow-up work

Things I think we'd still need to do to make this generalizable (cc @levadadenys):

- finer details of font weight / color / etc to match Figma
- managing what to do when tabs are larger than the container
- update to support multiple tab navs per page (tab IDs probably should be unique)
- collapsing/expanding tabs with icons (our approach is still being decided here by the design team)
- one "best practice" example would be to automatically open the "panel" controlled by a tab once you've focused on it (not done currently)
- [also calls for left/right arrows to allow you to move between tabs](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/examples/tabs-automatic/) -- I thought you might get this for free if you used the right aria properties and such, but this example makes me think you have to implement it yourself.
  - Alice also [noted in this comment](https://github.com/code-dot-org/code-dot-org/pull/56943#pullrequestreview-1920535253) that, once left/right arrows are implemented, that tab navigation should take you to the content of the panel, rather than to the next tab.


apps/src/aichat/views/Tabs.tsx
```diff
@@ -0,0 +1,69 @@
...
+  title: string;
+  content: React.ReactNode;
+};
+
+interface TabsProps {
+  tabs: Tab[];
+}
+
+const Tabs: React.FunctionComponent<TabsProps> = ({tabs}) => {
+  const [activeIndex, setActiveIndex] = useState<number>(0);
```
sanchitmalhotra126: Might be good to have an optional `initialActiveIndex` prop for this that defaults to 0 in case we want to open to a specific tab?

apps/src/aichat/views/Tabs.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+  const handleTabClick = (index: number) => {
+    setActiveIndex(index);
+  };
+
+  return (
+    <div>
```
sanchitmalhotra126: What's this outer `div` for?

sanchitmalhotra126 (commented):
Nice, this looks like a great start! I'll defer to @levadadenys from the design system perspective.

Minor - should we put the `Tab` files in their own subfolder here just so it's easier to identify and move when we migrate to design system tabs?

apps/src/aichat/views/Tabs.tsx
bencodeorg: Done! I also added a `name` prop that makes the tab and tab panel IDs between `Tabs` instances unique (assuming they don't use the same `name`). Not a guarantee obviously, but I'd expect not to have more than a couple instances of this on a page.

apps/src/aichat/views/Tabs.tsx
bencodeorg: Nothing :) Removed!

apps/src/aichat/views/tabs/Tabs.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+import styles from './tabs.module.scss';
+import TabPanel from './TabPanel';
+
+type Tab = {
+  title: string;
+  content: React.ReactNode;
+};
+
+interface TabsProps {
```
sanchitmalhotra126, line 12: nit: mind adding some comments to explain the props (mostly for the intended use of `name`)?

sanchitmalhotra126 (approved):

apps/src/aichat/views/tabs/Tabs.tsx
```diff
@@ -0,0 +1,85 @@
...
+                  )}
+                >
+                  {tab.title}
+                </button>
+              </li>
+            );
+          })}
+        </ul>
+      </div>
+      {tabs.map((tab, index) => {
```
levadadenys, line 71: nit: Do we need to render all of the tabs here? (since we only see 1 active tab at a time as far as I understand)

apps/src/aichat/views/tabs/tabs.module.scss
```diff
@@ -0,0 +1,37 @@
...
+  white-space: nowrap;
+  background-color: $neutral_dark70;
+  margin: 5px 8px 0 0;
+  padding: 0 10px;
+  border-radius: 4px 4px 0 0;
+
+  button {
+    @include remove-button-styles;
+    height: 30px;
+    font-size: 16px;
```
levadadenys, line 33: nit for the future, no need to update it right now:
feel fre to use `apps/src/componentLibrary/common/styles/mixins.scss`, for this specific case I guess `button-two-text` should work fine (-two size is a default, 'm' size of the text/components)

apps/src/aichat/views/tabs/Tabs.tsx
```diff
@@ -0,0 +1,85 @@
...
+      ? initialActiveIndex
+      : 0
+  );
+
+  const handleTabClick = (index: number) => {
+    setActiveIndex(index);
+  };
+
+  const nameStripped = name.replace(' ', '-');
+  const getTabButtonId = (index: number) => `${nameStripped}-tab-${index}`;
```
levadadenys, line 36: nit: use `useMemo` for `getTabPanelId` and `getTabButtonId` here?

apps/src/aichat/views/tabs/TabPanel.tsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+
+interface TabPanelProps {
+  content: React.ReactNode;
+  isActive: boolean;
+  id: string;
+  labelledBy: string;
```
levadadenys, line 8: nit: For design system version of Tabs I'll be writing comments describing every prop, but for now might be good to add a comment describing at least this one maybe?

levadadenys (approved):
Looks good overall to me, thanks a lot for this!
Have some nit comments, but they are non blocking

levadadenys:
> * one "best practice" example would be to automatically open the "panel" controlled by a tab once you've focused on it (not done currently)
> * [also calls for left/right arrows to allow you to move between tabs](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/examples/tabs-automatic/) -- I thought you might get this for free if you used the right aria properties and such, but this example makes me think you have to implement it yourself.

interesting, will need to investigate.
note for the future investigation, looks like manual activation tabs are also an option - more info here https://www.w3.org/WAI/ARIA/apg/patterns/tabs/examples/tabs-manual/

apps/src/aichat/views/tabs/Tabs.tsx
bencodeorg: Good q -- I was thinking we'd try to follow the [ARIA recommended pattern](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/) of loading all tab panels so we can do speedy nav between tabs via arrow keys. Not sure how frequently we'd put stuff that might be slower to load behind tabs (maybe we could handle this case by giving the developer using the tabs component the option to load the tab async if they'd like?)

image

apps/src/aichat/views/tabs/Tabs.tsx
bencodeorg: I don't think this is an expensive enough operation to make `useMemo` worthwhile?

apps/src/aichat/views/tabs/TabPanel.tsx
bencodeorg: Done!

apps/src/aichat/views/tabs/Tabs.tsx
levadadenys: Gotcha, thanks for the context!
+1 to allow developer to choose/switch between sync/async tabs load
Might also be worth to talk with design team about their opinion (maybe we want different behaviour based on the content of the tabs)

apps/src/aichat/views/tabs/Tabs.tsx
levadadenys: +1, fair enough

apps/src/aichat/views/ModelCustomizationWorkspace.tsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+
+import Tabs from './tabs/Tabs';
+
+const ModelCustomizationWorkspace: React.FunctionComponent = () => {
+  return (
+    <div style={{width: 400}}>
```
fisher-alice, line 7: nit: what do you think of using `aichatView.module.scss` as we do for `ChatWorkspace` at https://github.com/code-dot-org/code-dot-org/blob/53b937168ad0533fbecebe73d27f6964cc438141/apps/src/aichat/views/AichatView.tsx#L41

fisher-alice (approved):
Looks good! I always appreciate when you include the documentation you referred to for the PR, in this case, [ARIA standards for accessible tabs](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/).

I have a question about keyboard-navigation of tabs. I found it interesting that in the reference doc under 'Keyboard Interaction', they describe that when the focus moves into the tab list, pressing the tab key again 'moves focus to the next element in the page tab sequence outside the tablist, which is the tabpanel unless the first element containing meaningful content inside the tabpanel is focusable'.

 In order to navigate the tabs, a user uses the arrows which is included in your follow-up work. [Example](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/examples/tabs-automatic/)

Currently, pressing the tab key is the only way to keyboard-navigate from tab to tab. Do you know if we'll plan to implement keyboard-navigation as described in the docs? Just curious.

bencodeorg:
> I have a question about keyboard-navigation of tabs. I found it interesting that in the reference doc under 'Keyboard Interaction', they describe that when the focus moves into the tab list, pressing the tab key again 'moves focus to the next element in the page tab sequence outside the tablist, which is the tabpanel unless the first element containing meaningful content inside the tabpanel is focusable'.

That's a good callout! FYI @levadadenys as another thing to consider as we build out this component. I didn't have that listed as a follow-up, but adding now. Let me know if I didn't summarize the thought correctly!

apps/src/aichat/views/ModelCustomizationWorkspace.tsx
bencodeorg: Sounds like a good idea to manage the width in AichatView! I have a lot of CSS changes in the follow-up PR adding content to the tab panels, so I just added [a commit there](https://github.com/code-dot-org/code-dot-org/pull/57095/commits/33906c50a74db44548135fe00fe6c9a65d467420).

PR #55711, opened 2024-01-11
re-enable PDF generation

kobryan0619:
This is PR will re-enable the PDF auto-generation for curriculum PDFs.  These were disabled in [this PR](https://github.com/code-dot-org/code-dot-org/pull/53936/files). 

I have tested PDF generation locally on staging, and it looks good to go.  I am prepared to revert this PR if it causes issues with the staging build.  I will also let the DoTD know before merging that this is coming. 

## PR Checklist:

- [ ] Tests provide adequate coverage
- [ ] Privacy and Security impacts have been assessed
- [ ] Code is well-commented
- [ ] New features are translatable or updates will not break translations
- [ ] Relevant documentation has been added or updated
- [ ] User impact is well-understood and desirable
- [ ] Pull Request is labeled appropriately
- [ ] Follow-up work items (including potential tech debt) are tracked and linked

PR #52960, opened 2023-07-24
Reenable curriculum PDF generation

bethanyaconnor:
Effectively a revert of https://github.com/code-dot-org/code-dot-org/pull/52774 (there were some linter errors with that PR so I couldn't do a clean revert). Ruby 3 only broke resource pdf generation so we can turn PDF generation back on as those lines were commented out in #52923.

I've generated PDFs on staging since #52923 was merged so I believe this PR is safe, but I'm still planning on merging it at a time when it won't delay the DTP.

PR #56948, opened 2024-03-01
re-enable PDF generation

kobryan0619:
This is PR will re-enable the PDF auto-generation for curriculum PDFs.  These were disabled in [this PR](https://github.com/code-dot-org/code-dot-org/pull/53936/files). 

I have tested PDF generation locally on staging, and it looks good to go.  I am prepared to revert this PR if it causes issues with the staging build.  I will also let the DoTD know before merging that this is coming. 

## PR Checklist:

- [ ] Code is well-commented


kobryan0619:
This PR is a continuation of #55711, which was mis-closed by the Git LFS migration (#55759).

### Previous Comments:
- **kobryan0619** [commented](https://github.com/code-dot-org/code-dot-org/pull/55711#issuecomment-1887593193) - Thanks for the heads up @bethanyaconnor!  I will shoot for after the DTP!
 
- **davidsbailey** [commented](https://github.com/code-dot-org/code-dot-org/pull/55711#issuecomment-1887608333) - It looks like PDF generation was:
 * disabled in https://github.com/code-dot-org/code-dot-org/pull/52774 after it was broken by the ruby upgrade
 * enabled in https://github.com/code-dot-org/code-dot-org/pull/52960 after disabling generation of resource PDFs
 * disabled in https://github.com/code-dot-org/code-dot-org/pull/53936 
 
 so, why was it disabled in #53936 and what has changed that makes us think it is safe to enable again?
- **bethanyaconnor** [commented](https://github.com/code-dot-org/code-dot-org/pull/55711#issuecomment-1887656422) - Long thread but some context is here: https://codedotorg.slack.com/archives/C0T0PNTM3/p1695669458162549. No one had a clear theory as to why it stopped working, other than fast that it stopped working right after the Ubuntu upgrade. If Kaitie has tested on staging and not seen the issue re-appear, then it seems reasonable to try again?
 
 I think I also started a thread in the private teacher tools channel, but I've since left that channel, so I can't link to it.
- **davidsbailey** [commented](https://github.com/code-dot-org/code-dot-org/pull/55711#issuecomment-1887702949) - Thanks Bethany. 2 takeaways from the thread:
 * https://github.com/code-dot-org/code-dot-org/pull/53936 disabled it due to issues around the time of the ubuntu upgrade, which may have been transient
 * the conclusion at the end of the thread is that PDF generation might be flaky rather than completely broken, and so retrying it is a good idea
 
 I think this change looks good, and let's keep in mind to try restarting the staging build if it fails again.
- **kobryan0619** [commented](https://github.com/code-dot-org/code-dot-org/pull/55711#issuecomment-1887949499) - Thanks all - I am going to pause this PR for the moment and consult with someone on infra to confirm that this is ready.  I went to test out a few more units on staging, and started to get some flaky issues again which is confusing to me, but given this, I don't want to merge this in today before I leave for a break.  

### Previous Reviews:
- **bethanyaconnor approved** - fyi that there will likely be a lot of pdfs generated after this is re-enabled, so it's worth trying to merge this either early in the day or after a DTP
- **davidsbailey approved** -

kobryan0619:
Re-tagging Bethany to review just as a precaution.  Just an overview with where we are on this:

- We updated puppeteer which helped with the timeouts we were seeing. 
- We also modified the google gem for resources PDFs
- In all of this we are still seeing an issue with PDFs getting generated but with missing symbols when generated on the staging machine.  We will be investigating this later [[ticket](https://codedotorg.atlassian.net/jira/software/projects/TEACH/boards/65?selectedIssue=TEACH-905)] - it primarily impacts CSA.  We have reached out to the curriculum team and decided that after this PR is merged, we will manually generate the CSA PDFs on our local machine (where that issue is not happening).  As long as the curriculum team doesn't update the scripts, we think those PDFs should then be fine.  The curriculum team is not planning on updating those scripts.  We generated PDFs for the scripts listed [here](https://docs.google.com/spreadsheets/d/18vWPwdZOdarcNkiIiF2TMjMgOBAqomWiqAfuKMjnpU0/edit#gid=0) without any additional issue.

This PR will allow us to turn on PDF generation with this issue still in place.

bethanyaconnor:
> In all of this we are still seeing an issue with PDFs getting generated but with missing symbols

This worries me a little bit. If it's bad enough that we're asking curriculum writers to not update curriculum (including typos), then are we okay with re-enabling this? This could end up affecting new curriculum without us noticing.

kobryan0619:
> This worries me a little bit. If it's bad enough that we're asking curriculum writers to not update curriculum (including typos), then are we okay with re-enabling this? This could end up affecting new curriculum without us noticing.

It certainly is an issue with CSA - we asked Curriculum Writers, and they seemed ok with the change.  It is perhaps helpful to note that we have gotten at least one ZenDesk ticket reporting a similar issue with symbols in a CSA pdf not showing up prior to this update too.  

We are also plan on providing a fix ([TICKET](https://codedotorg.atlassian.net/browse/TEACH-912)) for this prior to the release of new curriculum too.  We would not want to ship new curriculum with these errors.

bethanyaconnor (approved):

bethanyaconnor:
Ah, if it existed in the past that does change things for me. I approved this PR but my remaining wonder is around what the bar is for "new curriculum" that don't have these symbol issues. We launch curriculum pretty frequently these days so we might hit this issue before our big curriculum launch.

That said, it sounds like y'all have thought this through and this change looks good!

PR #56956, opened 2024-03-02
Music: sound library updates - optional sound path

breville:
This adds support for an optional `path` for each sound in a "pack", so that sounds can be stored in subdirectories like `/bass` and `/leads` in this example:

```
  /library
    /instruments
      /instrument0
      /instrument1
    /kits
      /kit0
      /kit1
    /packs
      /pack0
        /bass
        /leads
      /pack1
      /pack2
```

In this change, the `MusicPlayer` uses a helper from `MusicLibrary` to generate full sound URLs which are stored in each `SampleEvent`.  The `SoundCache` uses the full URL as a key, and when it loads the sounds, it no longer generates the full URL or has to ask the library whether the sound is restricted.  Instead, if it sees a restricted URL, it does the appropriate work.


apps/src/music/player/MusicPlayer.ts
```diff
@@ -16,6 +16,7 @@ import Lab2Registry from '@cdo/apps/lab2/Lab2Registry';
 import {LoadFinishedCallback, UpdateLoadProgressCallback} from '../types';
 import {AudioPlayer, SampleEvent, SamplerSequence} from './types';
 import SamplePlayerWrapper from './SamplePlayerWrapper';
+import {DEFAULT_PATTERN_LENGTH, DEFAULT_CHORD_LENGTH} from '../constants';
```
sanchitmalhotra126, line 19: Oh thanks for fixing this! Can we remove the `require` below too?

apps/src/music/player/MusicPlayer.ts
```diff
@@ -285,9 +286,11 @@ export default class MusicPlayer {
         return [];
       }
 
+      const url = library.generateSoundUrl(folder, soundData);
+
       return [
         {
-          sampleId: `${folder.path}/${soundData.src}`,
+          sampleId: url,
```
sanchitmalhotra126: might be good to just change this field name to `sampleUrl` or `samplePath` for clarity?

apps/src/music/player/MusicLibrary.ts
```diff
@@ -96,16 +83,21 @@ export default class MusicLibrary {
...
-  getFolderForFolderPath(folderPath: string): SoundFolder | null {
-    return this.folders.find(folder => folder.path === folderPath) || null;
-  }
-
   // A progression step might specify a smaller set of allowed sounds.
   setAllowedSounds(allowedSounds: Sounds): void {
     this.allowedSounds = allowedSounds;
   }
 
+  generateSoundUrl(folder: SoundFolder, soundData: SoundData): string {
```
sanchitmalhotra126, line 91: could this just be `generateSoundUrlForId(id: string)` so you don't have to manually look up the `folder` and `soundData` first?

apps/src/music/player/MusicPlayer.ts
breville: Done.

apps/src/music/player/MusicPlayer.ts
breville: Done.

apps/src/music/player/MusicLibrary.ts
breville: If we have the ID (which is just `folder.id/sound.src`), we'd have to re-split it and look up the `SoundData` entry so that we could get the optional `path` which isn't included in the ID.

breville:
Thanks for looking at this.  Since it seems an overall reasonable direction, I'll merge it into its parent PR (https://github.com/code-dot-org/code-dot-org/pull/56677) but certainly happy to keep discussing these changes as part of that.

PR #53142, opened 2023-08-03
Create React UI for editing validations in levelbuilder

sanchitmalhotra126:
This adds a React-based UI for editing validations to levelbuilder pages. This is in an effort to create a nicer UI for authoring validations (rather than hand-crafting code or JSON), and has another big benefit of being able to share front end code with other front end components. This means that we can reference other front-end types, like the list of valid validations, and shared typescript types and interfaces.

Currently, this adds a new field to music levels called "validations", but when saving the level, we merge the validations back into level data. I think we are looking to move towards storing validations on level properties directly, rather than in level data, so when that change is complete, we can tweak some of the code that copies over the validations JSON in levels_controller.

### Testing Notes
I've hand tested this on some Music all the things levels, but i'm sure we'll continue to iterate on this as we start using it more.

https://github.com/code-dot-org/code-dot-org/assets/85528507/790777ef-302d-462b-970a-4428cf94787d

PR #56958, opened 2024-03-02
Add Nav Bar to Levelbuilder Edit Pages

kakiha11:
Allows for levelbuilders to more easily navigate the level edit page

https://github.com/code-dot-org/code-dot-org/assets/95503833/af3fc494-10b9-4961-ba3f-f8bae81b2cf5

Links adapt to each level type. Javalab above, Multi, Gamelab, and SpriteLab below.

SpriteLab

image

GameLab

image

Multi

image

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/dashboard/app/views/levels/edit.html.haml
+++ b/dashboard/app/views/levels/edit.html.haml
@@ -158,6 +158,5 @@
     var navbar = document.getElementById('navbar');
     navbar.appendChild(ul);
   });
-#navbar{style: "position: fixed; top: 0; left: 0; width: 10%; height: 100%; color: #8c52ba; padding: 30px; overflow-wrap: break-word"}
+#navbar{style: "position: fixed; top: 0; left: 0; width: 10%; height: 100%; color: #8c52ba; padding: 30px 15px; overflow-wrap: break-word; border-right: 3px solid rgb(231, 232, 234)"}
   %h2 Navigation
-  
--- a/dashboard/app/views/levels/edit.html.haml
+++ b/dashboard/app/views/levels/edit.html.haml
@@ -158,5 +158,24 @@
     var navbar = document.getElementById('navbar');
     navbar.appendChild(ul);
   });
-#navbar{style: "position: fixed; top: 0; left: 0; width: 10%; height: 100%; color: #8c52ba; padding: 30px 15px; overflow-wrap: break-word; border-right: 3px solid rgb(231, 232, 234)"}
+#navbar.navbar
   %h2 Navigation
+
+:css
+  .navbar {
+    position: fixed;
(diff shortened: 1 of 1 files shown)


breville:
It's great to see a useful feature being added to levelbuilder, and more such improvements are most welcome.

Here are some initial notes on what would bring this up to production quality:

- We shouldn't do JS or CSS directly in HAML any more.  When these are inside HAML, we lose the linting and browser-independence of our `apps` build pipeline.  Check out https://github.com/code-dot-org/code-dot-org/pull/53142 for an example of how to build a levelbuilder component using our `apps` pipeline.
- Unless there is a special case, it's best to programmatically render HTML using React.  The above-mentioned PR is a good example of that, as well.  React gives us more browser-independence, a more robust rendering model, and is consistent with the rest of our site.
- There should be no need to render this UI as `position: fixed`, and indeed it has some artifacts such as overlapping the rest of the page content at certain browser widths.  Instead, it should just be a `div` on the left side of the page, with the remaining content in a `div` to its right.
- More an aesthetic choice, but the list might not need to show bullets for each item.  We have some other examples on the site, such as at https://code.org/about.

kakiha11:
Thanks for these comments @breville ! It's cool to see an example of this done using React. I appreciate the guidance about best practices and I wonder if there is documentation about the direction the team is going when it comes to levelbuilder. I haven't seen any and it would be helpful to inform this and future work on this tool.

breville:
> I appreciate the guidance about best practices and I wonder if there is documentation about the direction the team is going when it comes to levelbuilder. I haven't seen any and it would be helpful to inform this and future work on this tool.

Good question.  I think the @code-dot-org/teacher-tools team might know more.

davidsbailey:
> > I appreciate the guidance about best practices and I wonder if there is documentation about the direction the team is going when it comes to levelbuilder. I haven't seen any and it would be helpful to inform this and future work on this tool.
> 
> Good question. I think the @code-dot-org/teacher-tools team might know more.

I agree with the best practices that Brendan mentioned above. Unfortunately levelbuilder is a place where we have lots of code violating these best practices, so it is often a judgment call on whether to follow the existing pattern vs try to improve it. That said, any time you find yourself adding a new command like `:javascript` or `:css` to a haml file, that's a sign that you probably should be creating a new file instead. (adding a couple of lines to a big existing block of js-in-haml could sometimes be ok). 

regarding building the html via `document.createElement`, I would say that we do not have many examples of that in the codebase and pretty much always want to use React (or in some cases haml) to generate the html.

For levelbuilder changes where you are not sure how to proceed, feel free to reach out at the outset.

PR #56959, opened 2024-03-02
Experiment: collaborative coding with ActionCable

snickell:
With this experiment, you can open http://localhost-studio.code.org:3000/collab/some_document_name in multiple browsers, and type in one or the other. You can change the documentID as you see fit, particularly important when it gets so diverged that it doesn't work anymore.

## Demo Video:
https://github.com/code-dot-org/code-dot-org/assets/223277/983a4a77-73b3-49a7-b28c-87a34f9fc26d

### Two main files of non-demo interest implement `collaborativeEditorExtension()`:
1) [**collaborative_editor_channel.rb**](https://github.com/code-dot-org/code-dot-org/blob/collab/dashboard/app/channels/collaborative_editor_channel.rb): which implements a ruby `ActionCable::Channel` version of 70% of [worker.ts](https://github.com/codemirror/website/blob/master/site/examples/collab/worker.ts) from the @codemirror/collab example.... leaving just the hard bits, naturally 😅. In addition to using ActionCable for broadcast, it implements stateful caching of update and doc data directly using redis.
2) [**collaborativeEditorExtension.ts**](https://github.com/code-dot-org/code-dot-org/blob/collab/apps/src/lab2/views/components/editor/collaborativeEditorExtension.ts): which implements an `@rails/actioncable` version of 100% of [collab.ts](https://github.com/codemirror/website/blob/master/site/examples/collab/collab.ts) from the @codemirror/collab example. Its designed as a standalone CodeMirror extension we could mix into CodeEditor.tsx, and is hopefully useful enough to finish and reuse.

### How to run the demo locally:
1) **run rails using: `bin/rails server`** from the dashboard directory instead of bin/dashboard-server. The latter uses thin as its web server, which apparently can be configured to work with actioncable, but isn't working out of the box.
2) **have a local redis instance running on redis://localhost:6379** (default port), on mac you can start your redis with `brew services start redis`.
3) open http://localhost-studio.code.org:3000/collab/boo in your browser (or try it in two browsers!). Its getting more robust, but if it does lose sync, try reloading, or if need be move on to /collab/boo2 etc.
4) if you close all browsers and restart rails, state should still be preserved (its saved in redis).

### TODO:
- [ ] **Integrate with lab2 CodeEditor.tsx**: add to the extensions list, and select the appropriate documentID to use, probably channelID?
- [x] [**Fix painful race condition keeping typing slow:**](https://github.com/code-dot-org/code-dot-org/blob/7431d5e1e602f29bf1e0684d3269b23ac2b2cd21/apps/src/lab2/views/components/editor/collaborativeEditorExtension.ts/#L190-L204) I believe there's a crucial async difference between web workers and ActionCable that is currently causing serious buggage. When calling `sendPushUpdates()` it does NOT block until the reply is received. The result is that if you type even remotely fast, you will retransmit the same versions, and the collab protocol is not resilient to this. Everything will freak out and break.
- [ ] implement `update_doc` in collaborative_editor_channel.rb. Right now it does not coalesce update streams into a document, instead it replays the full update stream on first load. This makes initial loads brittle. Location still to implement is generously flagged with a `raise DocNotImplementedError` on the ruby side. Crazy idea: use execjs gem in ruby (already installed) to re-use @codemirror/collab code to implement this. Otherwise we basically are going to be re-implementing all of @codemirror/collab in ruby 👎 .
- [x] [Use `get_doc` on the JS side](https://github.com/code-dot-org/code-dot-org/blob/5483f530462aad825ba116ed102d26e5d3ab86cd/apps/src/lab2/views/components/editor/collaborativeEditorExtension.ts/#L219-L222) to seed the initial document text and crucially the initial version (this is already saved in redis). The challenge here is the `collab()` extension takes an initialVersion at a point where I'm not sure we can block and be async. Their example gets around this in a pretty horrible way that tightly couples the extension to the EditorView/EditorState outside. Not seeding/reseeding the initial doc on update is a major source of instability, it means when the state breaks, its permanently broken, and you have to reset the data in redis (or just use a new documentID).
- [ ] implement [`pending` logic](https://github.com/code-dot-org/code-dot-org/blob/5483f530462aad825ba116ed102d26e5d3ab86cd/dashboard/app/channels/collaborative_editor_channel.rb/#L36-L42) in collaborative_editor_channel.rb. I'm not 100% sure this is essential, but the collab protocol is in general very very fragile, a single missing update breaks it semi-permanently. The pending logic makes it so an initial state-seeding `pull_updates` doesn't return if your current document is already ahead of the redis version. It blocks until the next push_update is complete. we can't block in action_cable, but we can use a redis queue to achieve a similar effect.
- [ ] Get websockets working with `bin/dashboard-server`, see: https://github.com/macournoyer/thin?tab=readme-ov-file#usage-with-rails-action-cable
- [ ] Add a separate per-user-per-document channel, and use it to reply to private messages. There's no reason for everyone to get spammed with `pull_update` because one person requested it.
- [ ] Implement other channels like: "who else is viewing this document"
- [ ] Implement shared cursor support
- [ ] How do we detect if we should enable collaborative mode? We don't want to open a websocket for every editor by default and wait for somebody else to join. This should be something you flip on.
- [ ] How do we cache the initial state / source code of documents?
- [x] **Implement throttling for `update(update: ViewUpdate)` calls to `push()`**. Even though the best performance is updating as frequently as the connection allows, we probably want to keep updates bellow a certain threshold. Maybe once every 5000ms, maybe once every 1000ms.
- [ ] Major investigation: what do we have to do to support websockets on prod-like instances (e.g. adhocs would probably work, I think?)
- [ ] Ruby tests for `collaborative_editor_channel.rb`
- [ ] TS tests for `collaborativeEditorExtension.ts`


dashboard/app/channels/collaborative_editor_channel.rb
```diff
@@ -0,0 +1,140 @@
...
+  def redis_version
+    redis.get(redis_version_key).to_i
+  end
+
+  def fetch_cached_doc
+    redis.get(redis_doc_key)
+  end
+
+  def update_doc(updates, version)
+    raise DocNotImplementedError, "update_doc is not implemented, so get_doc can't work"
```
snickell, line 95: implement `update_doc`

dashboard/app/channels/collaborative_editor_channel.rb
```diff
@@ -0,0 +1,140 @@
...
+    # FIXME: pending not yet implemented:
+    # "pull_updates who's version is current didn't return until the next push_updates"
+    #
+    # Q: is this really essential? It might be, but I'm not confident
+    # while (pending.length) pending.pop()!(data)
+  rescue PendingNotImplementedError
+  end
+
+  def get_doc
+    raise DocNotImplementedError, "get_doc is semi-implemented, but update_doc is not implemented so its not tested yet"
```
snickell: implement `get_doc`

dashboard/app/channels/collaborative_editor_channel.rb
```diff
@@ -0,0 +1,140 @@
...
+      #
+      # pending.push(resp)
+    end
+  end
+
+  def push_updates(data)
+    cache_updates data.dig("data", "updates"), data.dig("data", "version").to_i
+    ActionCable.server.broadcast(collaborative_editor_channel, data)
+
+    raise PendingNotImplementedError, "pending not yet implemented"
```
snickell, line 38: implement `pending`?

dashboard/app/channels/collaborative_editor_channel.rb
```diff
@@ -0,0 +1,140 @@
...
+  def pull_updates(data)
+    version = data.dig("data", "version").to_i
+
+    if version < redis_version
+      data[:data] = fetch_updates_from_cache(version)
+      # FIXME: this broadcasts to everyone, but we should only broadcast to the client that requested it
+      # that will require another channel
+      ActionCable.server.broadcast(collaborative_editor_channel, data)
+    else
+      raise PendingNotImplementedError, "pending not yet implemented"
```
snickell, line 23: implement `pending`?

apps/src/lab2/views/components/editor/collaborativeEditorExtension.ts
```diff
@@ -0,0 +1,231 @@
...
+      // next broadcast received of `push_updates` with the same version sent here.
+      //
+      // The effect of this difference is that if push is ever called while
+      // waiting for the previous sendPushUpdates to be received (and thereby
+      // this.view.dispatch(receiveUpdates(this.view.state, updates.updates))
+      // will be called, which means that update won't show up in sendableUpdates()
+      // anymore), then sync will be totally lost, versions will diverge, and
+      // everything will freak out and usually break. i.e. typing fast-ish
+      // breaks everything and I think this section is to blame.
+      this.sendPushUpdates(new Updates(version, updates));
```
snickell: I believe this is the source of the largest async race condition

apps/src/lab2/views/components/editor/collaborativeEditorExtension.ts
```diff
@@ -0,0 +1,231 @@
...
+      this.view.dispatch(receiveUpdates(this.view.state, updates.updates));
+      this.currentVersion = updates.version;
+    }
+
+    onReceive(action: Action, data: any) {
+      if (action === Action.PUSH_UPDATES) {
+        this.receiveUpdates(Updates.fromJSON(data));
+      } else if (action === Action.PULL_UPDATES) {
+        this.receiveUpdates(Updates.fromJSON(data));
+      } else {
```
snickell, line 269: TODO: implement a get_doc receive handler

apps/src/lab2/views/components/editor/collaborativeEditorExtension.ts
```diff
@@ -0,0 +1,231 @@
...
+    destroy() {
+      this.channel.unsubscribe();
+      this.done = true;
+    }
+  }
+
+  // FIXME: we need to seed our doc and crucial startVersion using Action.GET_DOC
+  // Awkward: how do we block on waiting for the initial version from Action.GET_DOC?
+  // any chance CodeMirror extension creation functions can be async?
+  const startVersion = 0;
```
snickell: TODO: initiate a get_doc here

apps/src/lab2/views/components/editor/collaborativeEditorExtension.ts
```diff
@@ -0,0 +1,231 @@
...
+
+  connect() {
+    const consumer = createConsumer();
+    this.channel = consumer.subscriptions.create(
+      {
+        channel: 'CollaborativeEditorChannel',
+        document_id: this.documentID,
+        client_id: this.clientID,
+      },
+      {
```
snickell, line 106: This is where we get down to ActionCable business on the JS side

snickell:
Confirmed: collab.ts blocks until it gets a response from the webworker: https://github.com/codemirror/website/blob/b12407fa512a9ccaf4272a3ff308c2746987f029/site/examples/collab/collab.ts#L24-L28

This is almost certainly the source of the extreme instability.

apps/src/lab2/views/components/editor/collaborativeEditorExtension.ts
snickell: we're initiating a get_doc in onConnect(), now the question is how to get that data down here, and have everything wait for it....

PR #56960, opened 2024-03-04
Incubator: update music buttons

breville:
This updates the **Music Lab** buttons at https://studio.code.org/incubator to get some extra explanatory text.  The intention is that users will have a better idea where they are going, and the onboarding funnel will strengthen because less users will want to skip directly to their project.

This updates the `TwoColumnActionBlock` component so that if any button is given some `extraText`, it switches layout to a vertical stack.  Otherwise, the component remains the same, with the buttons arranged horizontally.  The Storybook has been updated to demonstrate this new capability.

Screenshot 2024-03-06 at 8 29 55 AM

### DIFF
--- a/apps/src/templates/studioHomepages/twoColumnActionBlock.module.scss
+++ b/apps/src/templates/studioHomepages/twoColumnActionBlock.module.scss
@@ -44,5 +44,9 @@ $breakpoint: 993px;
     display: flex;
     flex-wrap: wrap;
     gap: 1.25rem;
+
+    &VerticalButtons {
+      flex-direction: column;
+    }
   }
 }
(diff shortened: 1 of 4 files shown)


apps/src/templates/studioHomepages/twoColumnActionBlock.module.scss
```diff
@@ -44,5 +44,9 @@ $breakpoint: 993px;
     display: flex;
     flex-wrap: wrap;
     gap: 1.25rem;
+
+    &VerticalButtons {
```
kelbyhawn, line 48: This is a cool pattern!

kelbyhawn (approved):
This looks good to me 🙌

levadadenys (approved):
LGTM in scope of Typography usage!
I wonder if this `extra text` feature is something we should also adopt in scope of dsco `Button` component, will discuss it with design team, thanks for pinging me!

PR #56966, opened 2024-03-04
Only log events on congrats and certificates pages for teachers

bethanyaconnor:
Dani pointed out that these events are sending for all users. Because we do have an event quota with Amplitude, I'm updating this to only send events for teachers.


apps/src/sites/studio/pages/print_certificates/show.js
```diff
@@ -1,6 +1,10 @@
 import analyticsReporter from '@cdo/apps/lib/util/AnalyticsReporter';
 import {EVENTS} from '@cdo/apps/lib/util/AnalyticsConstants';
+import {getStore} from '@cdo/apps/redux';
 
 $(() => {
-  analyticsReporter.sendEvent(EVENTS.CERTIFICATE_PRINT_PAGE_VISITED);
+  const isTeacher = getStore().getState().currentUser.userType === 'teacher';
+  if (isTeacher) {
+    analyticsReporter.sendEvent(EVENTS.CERTIFICATE_PRINT_PAGE_VISITED);
+  }
```
kelbyhawn, line 9: Learning Q: is this how you add Amplitude events on studio.code.org? (new logic notwithstanding)

apps/src/sites/studio/pages/print_certificates/show.js
```diff
@@ -1,6 +1,10 @@
 import analyticsReporter from '@cdo/apps/lib/util/AnalyticsReporter';
 import {EVENTS} from '@cdo/apps/lib/util/AnalyticsConstants';
+import {getStore} from '@cdo/apps/redux';
 
 $(() => {
-  analyticsReporter.sendEvent(EVENTS.CERTIFICATE_PRINT_PAGE_VISITED);
+  const isTeacher = getStore().getState().currentUser.userType === 'teacher';
```
kelbyhawn, line 6: Learning Q: is this a good example of using data coming from the database?

kelbyhawn (approved):
Have a couple learning questions, but otherwise looks good! Someone else might want to take a look too though.

apps/src/sites/studio/pages/print_certificates/show.js
hannahbergam: yuuup it's v simple on dashboard!

hannahbergam (approved):
This makes sense! Just to confirm: were you able to test via both account types in the console?

apps/src/sites/studio/pages/print_certificates/show.js
bethanyaconnor: This is more of an example of getting data from Redux (which, in this case, does come from the database at some point). But, definitely an example of a case where this page will act differently for different users!

bethanyaconnor:
> Just to confirm: were you able to test via both account types in the console?

Yep! Tested for a teacher (event fired), as well as signed out and for a student (no event fired).

apps/src/sites/studio/pages/print_certificates/show.js
kelbyhawn: Cool, ty!

PR #56937, opened 2024-03-01
Add Districts string for top nav

kelbyhawn:
Adds a new string "Districts" for use in the signed out top nav on code.org. Once this string is translated I will replace the Projects link with this Districts link. Also added to the [I18n gsheet](https://docs.google.com/spreadsheets/d/1Tq7VqZALgRA0wYk0HDfEOTyRI0TM2Dir2rloXIPGCgU/edit#gid=0&range=1538:1538) for Peagsus sites.

_Note:_ this is related to [this ticket](https://codedotorg.atlassian.net/browse/ACQ-1464), where the team originally wanted the Projects link replaced with Schools.

Jira ticket: [ACQ-1568](https://codedotorg.atlassian.net/browse/ACQ-1568)

PR #56972, opened 2024-03-04
String translation for new Create Account button

hannahbergam:
Getting a head start on the process for the header AB test. I chose this file and header to nest the string over based on [this PR](https://github.com/code-dot-org/code-dot-org/pull/56937/files) and because the other strings in this portion of the header live in the 'user' section of the file as well:

Screenshot 2024-03-04 at 11 44 10 AM

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/dashboard/config/locales/en.yml
+++ b/dashboard/config/locales/en.yml
@@ -431,6 +431,7 @@ en:
       logout: 'Sign out'
       signin: 'Sign in'
       signup: 'Create an account'
+      create_account: 'Create Account'
       already_registered: 'Already registered?'
       pair_programming: 'Pair programming'
       team: 'Team'


dashboard/config/locales/en.yml
```diff
@@ -431,6 +431,7 @@ en:
       logout: 'Sign out'
       signin: 'Sign in'
       signup: 'Create an account'
+      create_account: 'Create Account'
```
bethanyaconnor, line 434: similar to when @kelbyhawn added the "Schools" string, I think you also need to add this to the [i18n gsheet](https://docs.google.com/spreadsheets/d/1Tq7VqZALgRA0wYk0HDfEOTyRI0TM2Dir2rloXIPGCgU/edit#gid=0) so that it works on pegasus

bethanyaconnor (approved):

dashboard/config/locales/en.yml
hannahbergam: Roger that!

hannahbergam: (added)

PR #56974, opened 2024-03-04
Kaitie remove viewed icon

kobryan0619:
- Removed "Viewed" from the IconKey and popup.
- Removed `ITEM_TYPE.VIEWED` from any tests and from the `ItemType.jsx` file.
- Updated any associated tests

Before:
image

After:
image

image

## Links

[Ticket](https://codedotorg.atlassian.net/jira/software/projects/TEACH/boards/65?selectedIssue=TEACH-900)

## Testing story

Modified exisiting tests

### DIFF
--- a/apps/src/templates/sectionProgressV2/ItemType.jsx
+++ b/apps/src/templates/sectionProgressV2/ItemType.jsx
@@ -3,7 +3,6 @@ import color from '@cdo/apps/util/color';
 
 export const ITEM_TYPE = Object.freeze({
   NOT_STARTED: 1,
-  VIEWED: 2,
   NEEDS_FEEDBACK: 3,
   FEEDBACK_GIVEN: 4,
   ASSESSMENT_LEVEL: ['star', color.neutral_dark],
--- a/apps/src/templates/sectionProgressV2/ItemType.jsx
+++ b/apps/src/templates/sectionProgressV2/ItemType.jsx
@@ -3,6 +3,7 @@ import color from '@cdo/apps/util/color';
 
 export const ITEM_TYPE = Object.freeze({
   NOT_STARTED: 1,
+  VIEWED: 2,
   NEEDS_FEEDBACK: 3,
   FEEDBACK_GIVEN: 4,
   ASSESSMENT_LEVEL: ['star', color.neutral_dark],
--- a/apps/src/templates/sectionProgressV2/ProgressIcon.jsx
+++ b/apps/src/templates/sectionProgressV2/ProgressIcon.jsx
@@ -32,6 +32,20 @@ export default function ProgressIcon({itemType}) {
     />
   );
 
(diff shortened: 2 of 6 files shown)


apps/src/templates/sectionProgressV2/ItemType.jsx
```diff
@@ -3,7 +3,6 @@ import color from '@cdo/apps/util/color';
 
 export const ITEM_TYPE = Object.freeze({
   NOT_STARTED: 1,
-  VIEWED: 2,
```
kobryan0619: I am not sure if we should re-number `NEEDS_FEEDBACK` or `FEEDBACK_GIVEN` - I suppose we can... thoughts?

Also, I did a search for `ITEM_TYPE.VIEWED` to make sure we modified anywhere we had this still

etaderhold: If we do plan to implement the "viewed" state in the relatively near future I'd probably lean toward leaving in the definition of the VIEWED state in this file, and the rendering logic in `ProgressIcon` and just take out the references to it in the icon key and popup window. Might be worth asking Bryan what the plan is here.

If we do want to delete `ITEM_TYPE.VIEWED` entirely, there's no harm in renumbering them, or in keeping 2 unused. Your call.

kobryan0619: @dju90 any thoughts on the future of the `VIEWED` state/icon?  In this PR I removed any reference to it, but if we think we might implement it in the future (as Eric pointed out), we should just not render it for now - which will make it easier to add it back in the future.  Do you have an idea of the future for this icon/state?

dju90: We might implement it in the future, though it's unclear when. During prototype testing, users found it intuitive and valuable. However, I'm not sure yet which iteration will include this, as it does require backend changes to support it, so it's a relatively more costly investment.

apps/src/templates/sectionProgressV2/ProgressIcon.jsx
```diff
@@ -32,6 +32,9 @@ export default function ProgressIcon({itemType}) {
     />
   );
 
+  /*   Note that we decided not to have a viewedBox icon in this iteration
```
kobryan0619, line 35: Given the desire to include VIEWED in the future, I kinda want to keep this here as a note to anyone who gets assigned this in the future.  I know it is a small task, but I think it would be helpful... BUT it could also gunk up the code a bit... thoughts?

apps/src/templates/sectionProgressV2/ItemType.jsx
kobryan0619: Thanks @dju90 and nice catch @etaderhold .  I made the changes to reflect this.

etaderhold (approved):

PR #56975, opened 2024-03-04
[JIMT][CDOIDE] - newest rev of CDOIDE POC

thomasoniii:
Most recent rev of the IDE POC,  Internally this is completely different. With highlights:
- more config is exposed to twiddle how the components work.
- All of the project/config io is routed through a context and a reducer, centralizing the code instead of being striped to prop drilled functions.
- you can add a `.js` file to the project browser and it'll actually run.
- File browser is more robust with more error handling - I think it's feature complete other than being able to move items into other folders.
- Better handling of opening/closing docs and better keeping the preview up to date. Just generally more polished around.

Since this is still basically pre-alpha code, feel free to give it a cursory look and punt it over the goal line. Lots of iteration is required, but a lot of it is TBD.

Code is temporarily moved into a separate repo @ https://github.com/code-dot-org/cdo-ide-poc for now.

It's not going to live there long term (probably?) but I wanted to keep it more isolated from the rest of our monolith and it's faster to iterate separately. Please by all means feel free to poke around in that repo and look at code and offer any suggestions, and as we get things firmed up we'll pull the actual code back into the repo.

As of this PR, it's actually pointing to a separate branch in the repo, and will presumably continue to do so until I figure out how to resolve the issue with `define module` not being automatically added to the built `index.d.ts` file.

Couple of known outstanding things to think about before doing that -
- ~I haven't made our version of prettier play nicely as `prettier-standalone` to run in the browser. So we either need to figure that out or upgrade.~
- I made the subrepo export out a prettify function that uses an updated prettier that plays nicely. This'll work for now, but should obviously be swapped out for something local when we can.
- It's also installing `websandbox`, `@codemirror/lang-json`, and is using `@uiw/react-codemirror` for the internal editor. We may want to consider that package? I'm not married to it.
- linting is different in the other repo, but that's just temporary until it's merged back in with main. Pay no attention to the linting rules there.
- Maybe we decide we _do_ want a separate repo? I dunno.

Anyway, beyond that, the only really necessary bits to be aware of are to look to `Weblab2View.tsx` which hands the data into the editor instance. It's using the locally build `weblab2/Editor.tsx` file for our editing

The repo vends out `ProjectFileType`, `ConfigFileType`, which define the formats for the project json and config json files, respectively. When the IDE is instantiated, you should give it a `setProject` and `setConfig` callback to be called with updates as those things change. JSON is only returned en masse with a full project file again, regardless of changes. And nothing is throttled. The assumption is that this will all be handled externally, but we can iterate on that as necessary.

So for now, "all"(?) that's needed to implement external loading/saving of projects is an endpoint to vend out JSON in that format, and something that'll retrieve the updated json given to the `setProject` callback and then write it to disk (presumably after debouncing it and/or calculating diffs)
## PR Checklist:

- [ ] Code is well-commented


apps/src/weblab2/Editor.tsx
```diff
@@ -0,0 +1,65 @@
...
+  // this is a stupid hack. the low level code-mirror editor won't update itself
+  // automatically if the doc is changed externally. So in lieu of doing it better,
+  // for now we just key the component off of the file ID + an incrementing value that
+  // hits every time the format button is pressed. That'll force a re-render and make it work.
+  // let's swap this out with something better.
+  const [formats, setFormats] = useState(1);
+
+  const onChange = useCallback(
+    (value: string) => {
+      saveFile(file.id, value);
```
molly-moen, line 32: something to note is we will always save all files together

thomasoniii: All files are only always saved together. :-) 

That `saveFile` function there is internal to the editor, and what it's doing is is just updating the project json that was handed into the IDE.

_That_ updated json is then handed back to the `setProject` call back that was handed in when the IDE was instantiated, and it relies upon an external process to write that whole JSON blob back out to disk.

molly-moen:
> It's also installing websandbox, @codemirror/lang-json, and is using @uiw/react-codemirror for the internal editor. We may want to consider that package? I'm not married to it.

`websandbox`: probably fine to start with? We can do further investigations there. `@codemirror/lang-json` makes sense. I'm undecided on `@uiw/react-codemirror`: if it makes codemirror easier that is nice, but it's another dependency we have to hope will continue to keep pace with codemirror updates.

> Maybe we decide we do want a separate repo? I dunno.

Also undecided on this. In general we stick with the monorepo, so we would need a good argument for splitting.

molly-moen (approved):
I pulled this and took a look at the all the things level, looks reasonable! I also checked our other usages of codemirror (most importantly Java Lab) didn't get messed up because of new dependencies, still looks good there.

PR #56976, opened 2024-03-04
Cache Similar Curriculum Recommender results to save on identical runs

TurnerRiley:
Locally stores Similar Curriculum Recommender results so that identical queries can return the same result quicker with fewer runs of the recommender algorithm.

### Demo
As a way to demo that it's pulling the result from `sessionStorage` rather than re-running the algorithm, I temporarily added a console log statement within the check to see if the Curriculum Catalog page found a result in `sessionStorage` for the query.

First time opening the "Computer Science Discoveries" expanded card:
first

Next time opening the "Computer Science Discoveries" expanded card:
second

## Links
Jira ticket: [here](https://codedotorg.atlassian.net/jira/software/c/projects/ACQ/boards/64?assignee=60d62161f65054006980bd71&selectedIssue=ACQ-1398)

## Testing story
Added unit tests and locally tested.

## Follow-up work
When the Stretch Curriculum Recommender is added to the Curriculum Catalog page, I'll mirror/update this behavior to do the same thing.


bethanyaconnor (commented):
I love the direction and how scoped this PR is! I'm curious, however, about what happens when we add new curriculum or a user starts teaching a course; will the results become stale in that case? 

One thought would be to use `sessionStorage` over `localStorage`, though that may not cache long enough and theoretically could have the same issue if a user never restarts their browser. Another is to use a cache key that is hashed on the parameters of that function; I haven't thought through that enough to know if that's a winner though. :)

Thoughts?

TurnerRiley:
> I love the direction and how scoped this PR is! I'm curious, however, about what happens when we add new curriculum or a user starts teaching a course; will the results become stale in that case?
> 
> One thought would be to use `sessionStorage` over `localStorage`, though that may not cache long enough and theoretically could have the same issue if a user never restarts their browser. Another is to use a cache key that is hashed on the parameters of that function; I haven't thought through that enough to know if that's a winner though. :)
> 
> Thoughts?

Oh good point I think `sessionStorage` may make more sense to prevent stale results because I imagine it would be unlikely for a user to have the catalog open, teach a course, then return to that window without a reload/close.

bethanyaconnor (approved):

PR #56983, opened 2024-03-04
New School Association Component Progress

hannahbergam:
This PR focuses on setting up the new parent component of the school association flow. Note- this is still not live. I am viewing this locally by calling this file instead of SchoolInfoInputs from _finish_sign_up.js. I elected to open this PR now even though the component isn't quite finished to prevent the PR from getting too big and too stale.

Highlights: 
Moves data from props to state values (while maintaining save functionality by utilizing fieldNames)
Allows component to own its own change/update functions (encapsulation)
Sets up new child components for future work (zip search and school name inputs as opposed to SchoolNotFound and SchoolAutocompleteDropdownWithLabels)
Starts a couple more translations
Starts a new test file using RTL that covers a good bit of the flow logic

Some current pics of its status:

(on first load):
Screenshot 2024-03-12 at 1 13 39 PM

Screenshot 2024-03-12 at 1 11 37 PM
Screenshot 2024-03-12 at 1 11 45 PM
Screenshot 2024-03-12 at 1 11 57 PM

## Links

- spec: https://docs.google.com/document/d/1P1RVtGDy4XbIs0t5hnHC3Li6ZDsCnjwMnOPMQqPXNyk/edit?pli=1#heading=h.qi8xxyq72bct
- figma: https://www.figma.com/file/88sbry7xnl1wcyrOgbgFtn/Sign-Up-Flow?type=design&node-id=2197-13152&mode=design
- jira ticket: https://codedotorg.atlassian.net/browse/ACQ-1463

## PR Checklist:

- [ ] Code is well-commented


apps/src/templates/SchoolDataInputs.jsx
```diff
@@ -1,99 +1,138 @@
...
+      setAskForZip(false);
+      setIsOutsideUS(true);
+    }
+  };
+
   return (
     <div className={style.outerContainer}>
-      <Heading2 className={style.topPadding}>{i18n.censusHeading()}</Heading2>
-      <BodyTwoText>{i18n.schoolInfoInterstitialTitle()}</BodyTwoText>
+      {includeHeaders && (
```
hannahbergam, line 59: I can imagine a future where we're looking to grab the school association fields without the big headers at the top

bethanyaconnor: Is that in the spec? I don't want to prematurely optimize for a case that doesn't exist yet (or won't soon)

apps/src/templates/SchoolDataInputs.jsx
```diff
@@ -1,99 +1,138 @@
...
+      )}
       <div className={style.inputContainer}>
         <BodyTwoText className={style.padding} visualAppearance={'heading-xs'}>
           {i18n.whatCountry()}
         </BodyTwoText>
         <SimpleDropdown
           className={style.dropdown}
+          name={fieldNames.country}
           items={COUNTRY_ITEMS}
-          name="countryDropdown"
```
bethanyaconnor, line 51: this might be out of scope for this PR, but do you need to pass down the form name field here? ie one of the values defined [here](https://github.com/code-dot-org/code-dot-org/blob/03b428298e8b2bcfdc8bf90ff70ac4dadc3adac1/apps/src/templates/SchoolInfoInputs.jsx#L53)

bethanyaconnor: jk never mind sorry I completely missed the fact that this line moved 🤦‍♀️ 🤦‍♀️

apps/src/templates/SchoolDataInputs.jsx
```diff
@@ -1,99 +1,138 @@
...
+            <BodyTwoText
+              className={style.padding}
+              visualAppearance={'heading-xs'}
+            >
+              {i18n.enterYourSchoolZip()}
+            </BodyTwoText>
+            <input
+              type="text"
+              name={fieldNames.schoolZip}
+              placeholder={'i.e. 98104'}
```
bethanyaconnor: genuinely don't know the answer to this but should this be translateable?

apps/src/templates/SchoolDataInputs.jsx
dmcavoy: We use this component in other places such as the teacher application, afe, yourschool, etc. Moshe is supposed to be look at how the component should work in those other locations but we haven't finalized that yet.

apps/test/unit/templates/SchoolDataInputsTest.jsx
```diff
@@ -0,0 +1,41 @@
...
+import {render, screen, fireEvent} from '@testing-library/react';
+import {expect} from '../../util/deprecatedChai';
+import SchoolDataInputs from '@cdo/apps/templates/SchoolDataInputs';
+import i18n from '@cdo/locale';
+
+describe('SchoolDataInputs', () => {
+  function renderDefault(propOverrides = {}) {
+    render(<SchoolDataInputs {...propOverrides} />);
+  }
+  it('displays headers in basic component render', () => {
```
bethanyaconnor, line 12: tiny nit: can you add a new line above this test?

apps/src/templates/SchoolDataInputs.jsx
hannahbergam: @bethanyaconnor because Dani and I had to address this already with the different applications in the areas that I would've expected to include headers, I thought it was a good idea (it was also a small change, so didn't take me much time to build in). I'm trying to design something that is somewhat flexible from the start (especially considering what we talked over yesterday about including this work in places like account settings as phase one of sign up work). But happy to remove it if we don't think we'll need it!

apps/src/templates/SchoolDataInputs.jsx
hannahbergam: I'm not sure! This is by definition only available to users who select 'US' as country, but I'm all for translating anyway! @dmcavoy does that seem right to you?

apps/src/templates/SchoolDataInputs.jsx
bethanyaconnor: Sounds good -- no need to remove! Thanks for thinking ahead here :)

apps/src/templates/SchoolDataInputs.jsx
hannahbergam: (Also happy to remove this entirely. I subconsciously added it in but realize it isn't in the figma)

bethanyaconnor: I vote for translating if it stays in -- not all US users use our site in English (though I believe the vast majority do). I don't have an opinion on whether to keep it or not though.

bethanyaconnor (approved):
Realized my last two comments have become nits, so approving now :)

apps/src/templates/SchoolDataInputs.jsx
hannahbergam: After doing some [reading about placeholders](https://www.smashingmagazine.com/2018/06/placeholder-attribute/), I'm removing it! A highlight: the contrast is too low to be accessible. A11y guidelines state that the <label> should be sufficient, and if more info is needed, it should live below the label but above the input so it doesn't disappear.

apps/test/unit/templates/SchoolDataInputsTest.jsx
hannahbergam: Done! Thanks!

PR #56988, opened 2024-03-05
Datablock/get column js

cnbrenci:
Implement js side of get_column
Fix the get_column backend to work against shared tables
Update logging to be a big loud message instead of a console.error.
Enforce sorting records by record_id
Add tests for get_column


snickell (approved):
nice work!!!

PR #57007, opened 2024-03-05
Tag prod autoscaled instances with Name and environment

snickell:
Right now our production web-request-serving dashboard instances, created by the `Frontends-autoscale-prod` autoscaling group have no EC2 name set/tagged. This makes it unclear what they are in the AWS console.

This PR is totally untested because I don't know how to test cloudformation safely, but its my first attempt to find the right place to set the EC2 Name tag to something sensible. In this case it would be like:
`Name=Frontends-autoscale-prod-i-0f1b7e30ce0b09d6a`

To match dashboard-console and dashboard-daemon, I'm also setting tags:
`environment=production`.

### DIFF
--- a/aws/cloudformation/components/ami.yml.erb
+++ b/aws/cloudformation/components/ami.yml.erb
@@ -170,8 +170,8 @@ frontend_device_name ||= '/dev/sda1'
 
           STACK=${AWS::StackName}
           REGION=${AWS::Region}
-          INSTANCE_ID=$(curl -s http://169.254.169.254/latest/meta-data/instance-id)
 
+          INSTANCE_ID=$(curl -s http://169.254.169.254/latest/meta-data/instance-id)
           <%=indent(build_frontend, 10)%>
           if [ $? -eq 0 ]
           then
--- a/aws/cloudformation/components/ami.yml.erb
+++ b/aws/cloudformation/components/ami.yml.erb
@@ -163,8 +163,18 @@ frontend_device_name ||= '/dev/sda1'
 
           STACK=${AWS::StackName}
           REGION=${AWS::Region}
+          AUTO_SCALING_GROUP=Frontends-${AWS::StackName}
+          ENVIRONMENT=<%=environment%>
 
           INSTANCE_ID=$(curl -s http://169.254.169.254/latest/meta-data/instance-id)
+
+          # Tag the new EC2 instance's Name and environment for easy recognition in the AWS console
+          aws ec2 create-tags --resources $INSTANCE_ID \
+            --tags  Key=Name,Value="${AUTO_SCALING_GROUP}-${INSTANCE_ID}" \
+                    Key=environment,Value="$ENVIRONMENT" \
+            --region $REGION \
+            || true
+
           <%=indent(build_frontend, 10)%>
           if [ $? -eq 0 ]
           then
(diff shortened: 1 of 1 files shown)


aws/cloudformation/components/ami.yml.erb
```diff
@@ -163,8 +163,18 @@ frontend_device_name ||= '/dev/sda1'
 
           STACK=${AWS::StackName}
           REGION=${AWS::Region}
+          AUTO_SCALING_GROUP=Frontends-${AWS::StackName}
+          ENVIRONMENT=<%=environment%>
```
snickell: I'm copying how the ENVIRONMENT env var is set in bootstrap_chef_stack.sh.erb: https://github.com/code-dot-org/code-dot-org/blob/697167311a5338bf7d443c0ac691f581fe06968c/aws/cloudformation/bootstrap_chef_stack.sh.erb/#L24

In cloud_formation_stack.yml.erb, it directly assumes that $ENVIRONMENT is already set https://github.com/code-dot-org/code-dot-org/blob/697167311a5338bf7d443c0ac691f581fe06968c/aws/cloudformation/cloud_formation_stack.yml.erb/#L606

Maybe this line isn't required? No idea how to test, so I went with what seemed most conservative.

aws/cloudformation/components/ami.yml.erb
```diff
@@ -177,7 +187,6 @@ frontend_device_name ||= '/dev/sda1'
 
           # Signal CompleteLifecycleAction, in case this instance was launched from an Auto Scaling process.
           LIFECYCLE_HOOK=WebServerHook-${AWS::StackName}
-          AUTO_SCALING_GROUP=Frontends-${AWS::StackName}
```
snickell: moving this declaration up in the file, since its also used in the `aws ec2 crreate-tags` call

aws/cloudformation/components/ami.yml.erb
cat5inthecradle: This file, "ami.yml.erb" is included in "cloud_formation_stack" yml, so the variable should be accessible here. `ENVIRONMENT=<%=$ENVIRONMENT=>` should work. Or just use `$ENVIRONMENT` below, though I prefer getting out an ERB context as soon as possible by assigning some variables as you did.

Since this is ERB, I think you could test by validating it locally:

`bundle exec rake stack:validate RAILS_ENV=production`

You might need to set `VERBOSE=1` first to get the template to print.

cat5inthecradle (approved):

snickell:
@cat5inthecradle what does the process for merging this look like? I'm nervous to click the merge button given how little I know

cat5inthecradle:
Merge it!

"ami.yml.erb" is a component template imported into our main cloudformation stack. The `FrontendLaunchConfig` that you're modifying is already modified with every new deploy as the latest AMI gets swapped in, so this will simply add the Tags configuration for the new instances. No special action required for this to go out.

If there's a bug here, we won't see it until production :-1:  but it will simply error out of the cloudformation update and cancel the deploy :+1: . The biggest risk there would be that the daemon and console servers would be running new code, while the frontends would still be on the previously deployed version. If someone ships a non-backwards-compatible database migration, then we'll scrambling to hotfix, but otherwise this is pretty safe and standard procedure.

PR #57011, opened 2024-03-05
I18n-sync-down: Limit the number of requests to around 20 per second to avoid hitting Crowdin's rate limit

artem-vavilov:
https://codedotorg.slack.com/archives/C99KAHFK9/p1709595232044199
```
bundler: failed to load command: ./bin/i18n/sync-all.rb (./bin/i18n/sync-all.rb)
/home/ubuntu/code-dot-org/bin/i18n/utils/crowdin_client.rb:287:in `request':
Something went wrong while response processing. Details - 859: unexpected token at 
<html>                                                                                                           
<head><title>429 Too Many Requests</title></head>
<body>
<center><h1>429 Too Many Requests</h1></center>
<hr><center>nginx</center>
</body>
</html>
```

### DIFF
--- a/bin/i18n/utils/sync_down_base.rb
+++ b/bin/i18n/utils/sync_down_base.rb
@@ -72,7 +72,12 @@ module I18n
 
               locale_etags[source_file['path']] = translation_etag
             ensure
-              mutex.synchronize {progress_bar.increment}
+              mutex.synchronize do
+                progress_bar.increment
+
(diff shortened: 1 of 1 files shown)


mgc1194 (approved):
We are overwhelming Crowdin's server now that we request files individually instead of the project in bulk.
Downloading individual files seems fast enough, whereas adding a small sleep time every 20 requests could prevent the problem.

PR #57012, opened 2024-03-05
Re-enable tests that were skipped for leap day

bethanyaconnor:
These tests were skipped in https://github.com/code-dot-org/code-dot-org/pull/56736 and https://github.com/code-dot-org/code-dot-org/pull/56903, as they were breaking due to leap day. Now that leap day is passed, I'm re-enabling them!


kelbyhawn (approved):
🐸

PR #57017, opened 2024-03-05
Aitt add ai rubric sticky submit

wilkie:
Adds a 'sticky' footer to the bottom of the rubric modal.

This:
* Adds this footer... moving it from `RubricContent` to `RubricContainer` so it can exist outside of the content level
* Fixes an issue where the tab does not show for teacher feedback, but the textarea will, causing a 'are you sure' to appear on every navigation once general feedback is added
* Adds a route for adding general feedback relative to a student and rubric that is used to mark the student progress as 'keep working'
* Currently, it will write blank general feedback for this using the pre-existing mechanism.
* Refactors a bit of the tests that were getting unwieldy due to lots of async network activity.

image

image

image

The 'keep working' checkbox causes the student's bubble to reflect this in the progress view:

image

**Updated:** New styling that pushes down the feedback text a little bit:

image

## Links

- figma; [design](https://www.figma.com/file/WBcs7BY9IC6Nq5rG4V1SWF/AI-TA%3A-Rubrics?type=design&node-id=3488-115&mode=design&t=HDR4VqGdJWqj0Vh0-0)
- jira ticket: [AITT-469](https://codedotorg.atlassian.net/browse/AITT-469)

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/templates/instructions/TopInstructions.jsx
+++ b/apps/src/templates/instructions/TopInstructions.jsx
@@ -759,7 +759,7 @@ class TopInstructions extends Component {
                 }
               />
             )}
-            {!fetchingData && (
+            {!fetchingData && displayFeedbackTab && (
               <TeacherFeedbackTab
                 teacherViewingStudentWork={teacherViewingStudentWork}
                 visible={tabSelected === TabType.COMMENTS}
--- a/apps/src/templates/rubrics/RubricContainer.jsx
+++ b/apps/src/templates/rubrics/RubricContainer.jsx
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types';
 import style from './rubrics.module.scss';
 import i18n from '@cdo/locale';
 import classnames from 'classnames';
-import {BodyFourText, Heading6} from '@cdo/apps/componentLibrary/typography';
+import {Heading6} from '@cdo/apps/componentLibrary/typography';
 import FontAwesome from '@cdo/apps/templates/FontAwesome';
 import {
   reportingDataShape,
@@ -13,14 +13,9 @@ import {
 import RubricContent from './RubricContent';
 import RubricSettings from './RubricSettings';
 import RubricTabButtons from './RubricTabButtons';
+import RubricSubmitFooter from './RubricSubmitFooter';
 import {tryGetSessionStorage, trySetSessionStorage} from '@cdo/apps/utils';
 import Draggable from 'react-draggable';
-
-import HttpClient from '@cdo/apps/util/HttpClient';
-import {EVENTS} from '@cdo/apps/lib/util/AnalyticsConstants';
-import Button from '@cdo/apps/templates/Button';
-import Checkbox from '@cdo/apps/componentLibrary/checkbox/Checkbox';
-import analyticsReporter from '@cdo/apps/lib/util/AnalyticsReporter';
 
 const TAB_NAMES = {
   RUBRIC: 'rubric',
@@ -47,64 +42,7 @@ export default function RubricContainer({
   );
   const [aiEvaluations, setAiEvaluations] = useState(null);
 
-  const [isSubmittingToStudent, setIsSubmittingToStudent] = useState(false);
-  const [errorSubmitting, setErrorSubmitting] = useState(false);
-  const [lastSubmittedTimestamp, setLastSubmittedTimestamp] = useState(false);
   const [feedbackAdded, setFeedbackAdded] = useState(false);
-  const [keepWorking, setKeepWorking] = useState(false);
-
-  useEffect(() => {
-    if (open && studentLevelInfo) {
-      // Get the teacher feedback
-      const studentId = studentLevelInfo.user_id;
-      const endPoint = `/rubrics/${rubric.id}/get_teacher_feedback?student=${studentId}`;
-      fetch(endPoint)
-        .then(response => response.json())
-        .then(json => {
-          setKeepWorking(json.review_state === 'keepWorking');
-          const lastSubmittedDateObj = new Date(json.updated_at);
-          setLastSubmittedTimestamp(lastSubmittedDateObj.toLocaleString());
-        });
-    }
-  }, [open, rubric.id, studentLevelInfo]);
-
-  const submitFeedbackToStudent = () => {
-    analyticsReporter.sendEvent(EVENTS.TA_RUBRIC_SUBMITTED, {
-      ...reportingData,
-      studentId: studentLevelInfo.user_id,
-    });
-    setIsSubmittingToStudent(true);
-    setErrorSubmitting(false);
-    const body = JSON.stringify({
-      student_id: studentLevelInfo.user_id,
-      keep_working: keepWorking,
-    });
-    const endPoint = `/rubrics/${rubric.id}/submit_evaluations`;
-    HttpClient.post(endPoint, body, true, {'Content-Type': 'application/json'})
-      .then(response => response.json())
-      .then(json => {
-        setIsSubmittingToStudent(false);
-        if (!json.submittedAt) {
-          throw new Error('Unexpected response object');
-        }
-        const lastSubmittedDateObj = new Date(json.submittedAt);
-        setLastSubmittedTimestamp(lastSubmittedDateObj.toLocaleString());
-      })
-      .catch(() => {
-        setIsSubmittingToStudent(false);
-        setErrorSubmitting(true);
-      });
-    if (feedbackAdded) {
-      analyticsReporter.sendEvent(
-        EVENTS.TA_RUBRIC_SUBMITTEED_WRITTEN_FEEDBACK,
-        {
-          ...reportingData,
-          studentId: studentLevelInfo.user_id,
-        }
-      );
-      setFeedbackAdded(false);
-    }
-  };
 
   const tabSelectCallback = tabSelection => {
     setSelectedTab(tabSelection);
@@ -196,6 +134,8 @@ export default function RubricContainer({
             reportingData={reportingData}
             visible={selectedTab === TAB_NAMES.RUBRIC}
             aiEvaluations={aiEvaluations}
+            feedbackAdded={feedbackAdded}
+            setFeedbackAdded={setFeedbackAdded}
           />
           {showSettings && (
             <RubricSettings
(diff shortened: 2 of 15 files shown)


apps/src/templates/instructions/TopInstructions.jsx
```diff
@@ -759,7 +759,7 @@ class TopInstructions extends Component {
                 }
               />
             )}
-            {!fetchingData && (
+            {!fetchingData && displayFeedbackTab && (
```
wilkie, line 762: This is a strange one, but if you look at what happens when there is feedback but `displayFeedbackTab` is false, it will render the TeacherFeedbackTab anyway even though there's no way to navigate to it or even see it. The invisible textarea in this component _always_ attaches an `onunload` event to the window that warns if you navigate away. It made the page unusable in the case where a student's work was marked 'keep working'

apps/src/templates/rubrics/RubricContainer.jsx
```diff
@@ -40,6 +46,65 @@ export default function RubricContainer({
...
+      analyticsReporter.sendEvent(
+        EVENTS.TA_RUBRIC_SUBMITTEED_WRITTEN_FEEDBACK,
+        {
+          ...reportingData,
+          studentId: studentLevelInfo.user_id,
+        }
+      );
+      setFeedbackAdded(false);
+    }
+  };
```
wilkie: The feedback was moved from RubricContent to here unchanged.

dashboard/app/controllers/rubrics_controller.rb
```diff
@@ -1,9 +1,9 @@
...
-  load_resource only: [:get_teacher_evaluations, :ai_evaluation_status_for_user, :ai_evaluation_status_for_all, :run_ai_evaluations_for_user, :run_ai_evaluations_for_all]
-  load_and_authorize_resource except: [:submit_evaluations, :get_ai_evaluations, :get_teacher_evaluations, :ai_evaluation_status_for_user, :ai_evaluation_status_for_all, :run_ai_evaluations_for_user, :run_ai_evaluations_for_all]
+  before_action :require_levelbuilder_mode_or_test_env, except: [:submit_evaluations, :get_ai_evaluations, :get_teacher_evaluations, :get_teacher_feedback, :ai_evaluation_status_for_user, :ai_evaluation_status_for_all, :run_ai_evaluations_for_user, :run_ai_evaluations_for_all]
+  load_resource only: [:get_teacher_evaluations, :get_teacher_feedback, :ai_evaluation_status_for_user, :ai_evaluation_status_for_all, :run_ai_evaluations_for_user, :run_ai_evaluations_for_all]
+  load_and_authorize_resource except: [:submit_evaluations, :get_ai_evaluations, :get_teacher_evaluations, :get_teacher_feedback, :ai_evaluation_status_for_user, :ai_evaluation_status_for_all, :run_ai_evaluations_for_user, :run_ai_evaluations_for_all]
```
davidsbailey: it seems important to properly authorize the new route since it could return PII. Please add auth tests like these and then make sure they pass: https://github.com/code-dot-org/code-dot-org/blob/1f7fb6c9aade6892401913edd1425d4cad6b108e/dashboard/test/controllers/rubrics_controller_test.rb#L39-L41

ideally the auth rules would go in ability.rb, but if you're not familiar with those then it seems fine to continue the existing pattern in this file with rules like these: https://github.com/code-dot-org/code-dot-org/blob/71f304d563ec0d13481e3785430eb503e85bcf8d/dashboard/app/controllers/rubrics_controller.rb#L141

davidsbailey (commented):
not done reviewing but sharing initial comments

dashboard/app/controllers/rubrics_controller.rb
```diff
@@ -68,6 +68,16 @@ def submit_evaluations
...
 
+    # Update the main TeacherFeedback
+    teacher_feedback = TeacherFeedback.find_or_create_by!(student_id: permitted_params[:student_id], teacher_id: current_user.id, script: rubric.get_script_level.script, level: rubric.level)
+
+    if permitted_params[:keep_working]
+      teacher_feedback.update(review_state: TeacherFeedback::REVIEW_STATES.keepWorking)
+    else
+      teacher_feedback.update(review_state: nil)
+    end
+    teacher_feedback.save!
```
davidsbailey: I think I am missing something here. why is there new backend functionality being added when we submit evaluations, was this not getting saved prior to this PR? the PR description and Jira item seem to mostly cover UI changes so I am not seeing the connection here.

davidsbailey (commented):
done reviewing, more questions 😁

dashboard/app/controllers/rubrics_controller.rb
wilkie: The 'Keep Working' checkbox requires that the student be marked as such. Progress is marked via teacher feedback. So, to support the keep working checkbox, it required the ability to post progress state relative to a rubric. Happy to remove the checkbox and this solution and move it to another PR.

dashboard/app/controllers/rubrics_controller.rb
wilkie: Is that done for `get_teacher_evaluations`?

dashboard/app/controllers/rubrics_controller.rb
wilkie: Another way to go about it would be to use the existing feedback api to set 'keep working' which would mean posting to both our submit and the teacher feedbacks api endpoint and then querying both on load. I'd need to pass down the script and level ids to craft the requests instead of using the rubric to find them. Could add them to the rubric info and make two requests and forgo the new routes.

dashboard/app/controllers/rubrics_controller.rb
davidsbailey: No, it looks like it is not! Apologies for not catching this earlier. I think the line quoted just above should do the trick.

wilkie: Sounds good! I just did what the evaluations path did and assumed it was good! what you are saying makes 100% sense... though we might just look at doing an initial POST to the feedback api path instead of combining it in the one submit call.

dashboard/app/controllers/rubrics_controller.rb
wilkie: This is the existing route: https://github.com/code-dot-org/code-dot-org/blob/b728fdb10e2134fb8dbeb3568668bda2550054be/dashboard/app/controllers/api/v1/teacher_feedbacks_controller.rb#L56-L69

So, an initial call to this to mark the progress state before submitting the rubric feedback would accomplish this just as well after updating it to pass down the level/script id.

tess323:
could we add just a touch more padding above the submitted at timestamp (~3px).  The font-weight on the submitted at looks heavy to me, but if that feels like the right component to use I am fine with it

dashboard/app/controllers/rubrics_controller.rb
davidsbailey: makes sense! I just didn't realize the Keep Working checkbox was new as opposed to being moved from somewhere else in our UI. 

I think it is is dealer's choice on whether to make 1 API call or 2. I kind of like how you have it because the client logic can be a bit simpler, not having to worry about different combinations of request failures. Though I suppose if the existing route got updated in an important way, then our route could get out of date. Always with the tradeoffs 😁

apps/src/templates/rubrics/RubricContainer.story.jsx
```diff
@@ -95,6 +99,7 @@ const defaultStudentLevelInfo = {
   attempts: 2,
   timeSpent: 404,
   lastAttempt: '5/26/23',
+  user_id: 107,
```
wilkie, line 102: We need these now to support the footer's keepWorking feedback fetches.

apps/src/templates/rubrics/RubricContent.jsx
```diff
@@ -48,51 +44,12 @@ export default function RubricContent({
...
-        EVENTS.TA_RUBRIC_SUBMITTEED_WRITTEN_FEEDBACK,
-        {
-          ...reportingData,
-          studentId: studentLevelInfo.user_id,
-        }
-      );
-      setFeedbackAdded(false);
-    }
-  };
-
```
wilkie, line 95: This is moved to the RubricSubmitFooter component along with the rendering below it.

apps/src/templates/rubrics/rubricShapes.jsx
```diff
@@ -19,6 +19,12 @@ export const rubricShape = PropTypes.shape({
     position: PropTypes.number,
     name: PropTypes.string,
   }),
+  script: PropTypes.shape({
+    id: PropTypes.number,
+  }),
+  level: PropTypes.shape({
+    id: PropTypes.number,
+  }),
```
wilkie, line 27: Now needed in order to support the teacherFeedback API for supporting setting progress state to 'keepWorking'

apps/test/unit/templates/rubrics/RubricContainerTest.jsx
```diff
@@ -195,17 +229,10 @@ describe('RubricContainer', () => {
...
-      .onCall(1)
-      .returns(Promise.resolve(new Response(JSON.stringify(successJsonAll))));
-    fetchStub
-      .onCall(2)
-      .returns(
-        Promise.resolve(new Response(JSON.stringify(mockAiEvaluations)))
-      );
+    stubFetchEvalStatusForUser(successJson);
+    stubFetchEvalStatusForAll(successJsonAll);
+    stubFetchAiEvaluations(mockAiEvaluations);
```
wilkie, line 238: A lot of the changes to this test module are to get rid of the arbitrary order of fetch calls. Adding the support for another API call throughout and it being rendered as a whole and not shallowly just meant that it was hurting my head to figure out what the order was supposed to be.

So I'm stubbing out the fetches filtered by their URL so I can see that they are also being passed the right arguments and then it doesn't matter what order they happen in. I just care that they _have happened_ at each interval the result matters.

apps/test/unit/templates/rubrics/RubricContainerTest.jsx
```diff
@@ -218,9 +245,7 @@ describe('RubricContainer', () => {
         />
       </Provider>
     );
-    await act(async () => {
-      await Promise.resolve();
-    });
+    await wait();
```
wilkie, line 252: I got tired of copy and pasting this... I do wonder what in the world of RTL could help, but I just used this pattern there, too. It still seems useful when you want to delay the action when it happens initially or based on the passage of time.

apps/test/unit/templates/rubrics/RubricContentTest.jsx
```diff
@@ -203,83 +200,6 @@ describe('RubricContent', () => {
...
-    await act(async () => {
-      await Promise.resolve();
-    });
-    wrapper.update();
-    expect(wrapper.find('BodyThreeText').at(1).props().children).to.equal(
-      'Error submitting feedback to student.'
-    );
-    postStub.restore();
-  });
-
```
wilkie, line 282: More or less moved to the RubricSubmitFooter tests.

apps/test/unit/templates/rubrics/RubricSettingsTest.jsx
```diff
@@ -29,10 +46,14 @@ describe('RubricSettings', () => {
   });
 
   afterEach(() => {
-    fetchStub.restore();
-    utils.queryParams.restore();
+    if (clock) {
+      clock.restore();
```
wilkie, line 50: Not restoring the use of the fake timers made the React Testing Library's waiting not work since it uses the timers internally to poll for changes. If one of my tests ran after this one, it would timeout.

apps/test/unit/templates/rubrics/RubricSettingsTest.jsx
```diff
@@ -16,11 +16,28 @@ import {Provider} from 'react-redux';
...
+      });
+    }
+  }
+
+  function stubFetchEvalStatusForAll(data) {
+    fetchStub
+      .withArgs(sinon.match(/rubrics\/\d+\/ai_evaluation_status_for_all.*/))
+      .returns(Promise.resolve(new Response(JSON.stringify(data))));
+  }
+
```
Nokondi, line 37: I like all of these test refactors. Really cleans things up.

Nokondi (approved):
Looks great! I have nothing to add that Dave didn't already cover.

PR #57021, opened 2024-03-05
Setup initial levelbuilder editing for Gen AI fields

sanchitmalhotra126:
This is primarily adding the plumbing and boilerplate to hook up the React-based editing UI with the levelbuilder system. I've also just added one field (bot name) to get things working end-to-end. The rest of the fields will come in a separate change.

@dancodedotorg let me know if this is looking ok so far and if I should reword any of the text/labels for clarity!

Screenshot 2024-03-05 at 11 52 53 AM

Once saved, this data is available in level properties (screenshot of the JSON returned from the /level_properties server call):

Screenshot 2024-03-05 at 11 48 19 AM

## Links

https://codedotorg.atlassian.net/browse/LABS-516

## Testing story

Tested locally on allthethings AI Chat level. Confirmed that level properties are updated and read correctly.

### DIFF
--- a/apps/src/lab2/levelEditors/aiCustomizations/EditAiCustomizations.tsx
+++ b/apps/src/lab2/levelEditors/aiCustomizations/EditAiCustomizations.tsx
@@ -1,0 +1,92 @@
+import {
+  AiCustomizations,
+  LevelAiCustomizations,
+  Visibility,
+} from '@cdo/apps/aichat/types';
+import React, {useCallback, useState} from 'react';
+import {AppName} from '../../types';
+import {Heading3} from '@cdo/apps/componentLibrary/typography';
+import SimpleDropdown from '@cdo/apps/componentLibrary/simpleDropdown/SimpleDropdown';
+
+interface EditAiCustomizationsProps {
+  initialCustomizations: LevelAiCustomizations;
+  levelName: string;
+  appName: AppName;
+}
+
+const EditAiCustomizations: React.FunctionComponent<
+  EditAiCustomizationsProps
+> = ({initialCustomizations, levelName, appName}) => {
+  const [aiCustomizations, setAiCustomizations] =
+    useState<LevelAiCustomizations>(initialCustomizations);
+
+  const setPropertyVisiblity = useCallback(
+    (property: keyof AiCustomizations, visibility: Visibility) => {
+      setAiCustomizations({
+        ...aiCustomizations,
+        [property]: {
+          value: aiCustomizations[property]?.value,
+          visibility,
+        },
+      });
+    },
+    [setAiCustomizations, aiCustomizations]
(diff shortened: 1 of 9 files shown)


dancodedotorg (approved):
Looks good to me! Don't have strong thoughts about the wording yet - the dropdown options you provided feel right to me!

Potentially a question for farther down the road, but I wonder your thoughts on how all these settings will be grouped within levelbuilder - ie: if _everything_ I have the option to customize will be under this single "AI ChatBot Customizations" collapsible heading, or if you plan to do multiple collapsible headings (like "Prompt panel customization" and "Retrieval customization" and "Model card customization")? That may be too much for this moment, so just putting that thought out there for now!

sanchitmalhotra126:
> Potentially a question for farther down the road, but I wonder your thoughts on how all these settings will be grouped within levelbuilder - ie: if _everything_ I have the option to customize will be under this single "AI ChatBot Customizations" collapsible heading, or if you plan to do multiple collapsible headings (like "Prompt panel customization" and "Retrieval customization" and "Model card customization")? That may be too much for this moment, so just putting that thought out there for now!

For now I was planning to just group everything under this single heading, but I could also split them up down the road if that's easier? Either way works.

dancodedotorg:
> For now I was planning to just group everything under this single heading, but I could also split them up down the road if that's easier? Either way works.

If it's easy to switch around as we build it, then we don't need to think about it now. Grouping under a single header makes sense for now!

apps/src/lab2/levelEditors/aiCustomizations/EditAiCustomizations.tsx
```diff
@@ -0,0 +1,112 @@
...
+    [setAiCustomizations, aiCustomizations]
+  );
+
+  return (
+    <div>
+      <input
+        type="hidden"
+        id="level_initial_ai_customizations"
+        name="level[initial_ai_customizations]"
+        value={JSON.stringify(aiCustomizations)}
```
bencodeorg, line 46: For my understanding, is this a standard levelbuilder pattern (hidden input)?

bencodeorg (approved):
LGTM!

apps/src/lab2/levelEditors/aiCustomizations/EditAiCustomizations.tsx
sanchitmalhotra126: I think when I was working on the Lab2 validations editor, I saw this in a few other places and just copied it. I'm not sure if it's the ideal method, but we basically need a way for the levelbuilder edit page to grab the data in this component when we click the save button (i.e. submit the form). So this allows us to effectively split up the different pieces of the data into their own inputs, and then combine them back together into a single JSON object when we actually want to save it

PR #57022, opened 2024-03-05
[Sprite Lab] player jump block and explicit sprite groups

mikeharv:
Note that most of the changes in this branch are two a level file that shouldn't need close review. This is just an internal testing level that is inaccessible to users. It's also worth keeping in mind that everything in `dashboard/config` will be ultimately owned by levelbuilders. The new blocks and level will see a lot of iteration between now and launch that will not go through our PR review process.

**Context:** The K5 team is building a new game development module using Sprite Lab that focuses on students learning to program with variables. Top on their lists of block requests was a block to make sprites "jump". However, this is an incredibly complicated action to get right. Learning to make a video game sprite jump requires thinking about everything from velocity to gravity to collisions. Jumping also isn't a "behavior" in the Sprite Lab sense, as its not a set of actions that repeat until the behavior ends. In the simplest form, jumping could work by increasing a sprite's vertical velocity when a button is pressed, however we'd also need a way to make sure the sprite comes back down. In most games, sprites can only jump when they're "standing" on a suitable surface. After discussing these challenges with Katie from the curriculum team, we arrived at some shared conclusions:
* The focus of the module (and therefore the coding) needs to be variables. Ie. students should be thinking about incrementing a score when things happen and not contemplating physics.
* Students need to be able to create sprites that function either as a player, as part of the environment, or as different kind of game "object" (e.g. enemies, collectibles, etc.)
* In a platform-style game, the player sprite should gradually fall towards the bottom of the screen.
* Environment sprites (e.g. walls, floors) should block the player and object sprites from moving.
* Students should be able to use events to program the player sprite to jump (when space is pressed: player jumps).
* The player should only be able to begin jumping when directly above an environment sprite, or at the bottom of the canvas.
* The jump strength should be configurable.

**Approach:**
The technical solution for this required me to make additions to the new GameDev block pool, helper code for new blocks, a custom helper library for the level (likely to become a new level subtype), and some very small additions to Sprite Lab's core library.

_Explicit Sprite Groups_
Since about 2020, Sprite Lab has implicitly grouped all sprites based on their costumes. For example, it's easy to resize all carrot sprites with a single block rather than having to reference each one by name or id individually.  In order to hit some of the goals stated above, we will need to allow students to assign new sprites to _explicit_ groups for things like the environment (walls/floors). This will make it possible for the libraries that power the level to handle certain collisions for students automatically. Sprites that are assigned to the "players" group will automatically be affected by the library's simulated gravity, and will be programmable to jump when at rest above an environment sprite.
Implementing this requires updating CoreLibrary so that it has the ability to find sprites based on a `group` property. It also requires new blocks for the GameDev pool.

_New Blocks_
New blocks were created that allow you to create an reference sprites in the new groups. `makeNewGroupSprite`'s function is defined in `SpriteCommands` because no other commands has the ability to assign a sprite to a new group. 
image
The blocks for getting these groups of sprites do not have commands because they just return simple values such as `{group: 'walls'}` for environment sprites. The group names can be any valid string and the specific strings chosen here all live on levelbuilder so that the curriculum team may iterate upon them.
image
image
The new `playerJump` block's function is defined right with the block as it works entirely by calling other functions defined elsewhere.
image

_Player sprite position checking_
In order for a sprite to only jump at the right times, we need a way to confirm whether it was directly above another sprite. A new helper command was created for this which compares the "colliders" of the sprites in each group. If the bottom of the player lines up with the top of the environment and the left-to-right bounds of the sprites overlap, it counts as being directly above.

_New Helper Library_
Any new logic that relates to the goals above was placed into levelbuilder files wherever possible, including specific gravity and collision logic. These rules and behaviors will need to be tweaked as the curriculum team designs their lesson, but the library provides a simple starting place. If things are solidified after pilot and feedback iteration, we can potentially move these rules into the core library for Sprite Lab or another new sub-lab type (similar to Poetry).

https://github.com/code-dot-org/code-dot-org/assets/43474485/c09a66b3-9412-4712-8c3a-80f9d0294f46

## Links

Jra - https://codedotorg.atlassian.net/browse/CT-355
GDoc - https://docs.google.com/document/d/1nbKmsEF10y6HZdoL_ZeKqKTBWWxHN1JQugpEYlLSMJY/edit

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/p5lab/spritelab/commands/actionCommands.js
+++ b/apps/src/p5lab/spritelab/commands/actionCommands.js
@@ -1,5 +1,7 @@
 import {commands as behaviorCommands} from './behaviorCommands';
 import {layoutSpriteGroup} from '../../layoutUtils';
+import {APP_HEIGHT} from '../../constants';
+import {createSpriteCollider} from '../../utils';
 
 function move(coreLibrary, spriteArg, distance) {
   let sprites = coreLibrary.getSpriteArray(spriteArg);
@@ -107,6 +109,14 @@ export const commands = {
     sprites.forEach(sprite => this.p5.edges.displace(sprite));
   },
 
+  edgesCollide(spriteArg) {
+    if (!this.p5.edges) {
+      this.p5.createEdgeSprites();
+    }
+    let sprites = this.getSpriteArray(spriteArg);
+    sprites.forEach(sprite => sprite.collide(this.p5.edges));
+  },
+
   glideTo(spriteArg, location) {
     if (!location) {
       return;
@@ -165,6 +175,32 @@ export const commands = {
           touching = true;
         }
       });
+    });
+    return touching;
+  },
+
+  isDirectlyAbove(spriteArg, targetArg) {
+    let sprites = this.getSpriteArray(spriteArg);
+    let targets = this.getSpriteArray(targetArg);
+    let touching = false;
+    sprites.forEach(sprite => {
+      const spriteCollider = createSpriteCollider(sprite);
+      if (spriteCollider.bottom >= APP_HEIGHT) {
+        touching = true;
+      } else {
+        for (const target of targets) {
+          const targetCollider = createSpriteCollider(target);
+
+          if (
+            spriteCollider.bottom === targetCollider.top &&
+            spriteCollider.left <= targetCollider.right &&
+            spriteCollider.right >= targetCollider.left
+          ) {
+            touching = true;
+            break;
+          }
+        }
+      }
     });
     return touching;
   },
--- a/apps/src/p5lab/spritelab/commands/actionCommands.js
+++ b/apps/src/p5lab/spritelab/commands/actionCommands.js
@@ -101,6 +101,7 @@ export const commands = {
       targets.forEach(target => sprite[collisionType](target));
     });
   },
+
   edgesDisplace(spriteArg) {
     if (!this.p5.edges) {
       this.p5.createEdgeSprites();
@@ -109,6 +110,7 @@ export const commands = {
     sprites.forEach(sprite => this.p5.edges.displace(sprite));
   },
 
+  // Causes the sprite to stop moving when it hits an edge sprite.
   edgesCollide(spriteArg) {
     if (!this.p5.edges) {
       this.p5.createEdgeSprites();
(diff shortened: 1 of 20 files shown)


apps/src/p5lab/spritelab/commands/actionCommands.js
```diff
@@ -169,6 +179,32 @@ export const commands = {
...
+    sprites.forEach(sprite => {
+      const spriteCollider = createSpriteCollider(sprite);
+      if (spriteCollider.bottom >= APP_HEIGHT) {
+        touching = true;
+      } else {
+        for (const target of targets) {
+          const targetCollider = createSpriteCollider(target);
+
+          if (
+            spriteCollider.bottom === targetCollider.top &&
```
molly-moen, line 197: Are these numbers pixel locations? Should we have some sort of buffer (ex top is between `bottom - 2` and `bottom + 2`)? This might just be something we figure out by using the blocks so it could be a follow up.

apps/src/p5lab/utils.js
```diff
@@ -17,3 +17,15 @@ export function arrayEquals(a, b) {
 
   return true;
 }
+
+// Creates and returns a helper object that represents the outer bounds of a sprite.
+// Used for comparing the positions of sprites. For example, checking whether
+// one sprite is "standing" on top of another.
```
molly-moen: nit: mind adding an explainer of what the numbers mean? This might be me not knowing how p5 works, are these pixel locations or some other coordinate system?

apps/src/p5lab/spritelab/commands/actionCommands.js
```diff
@@ -107,6 +109,14 @@ export const commands = {
     sprites.forEach(sprite => this.p5.edges.displace(sprite));
   },
 
+  edgesCollide(spriteArg) {
```
molly-moen, line 114: mind adding a comment explaining what this does?

molly-moen (approved):
Looks good, non-blocking comments to help me understand things!

apps/src/p5lab/spritelab/commands/actionCommands.js
mikeharv: Yes, they're essentially pixel locations. For now, it's working as expected, but I think we could certainly add a buffer if/when needed.

apps/src/p5lab/spritelab/commands/actionCommands.js
mikeharv: Sure, I can do that. Unfortunately none of the other commands in this directory are well commented at this time.

apps/src/p5lab/utils.js
mikeharv: Just pushed a comment to explain!

PR #56932, opened 2024-03-01
Part 4: Add notification when user's project is frozen

fisher-alice:
This PR notifies a user when their project is frozen. A project is frozen when either it is selected as:
- an active featured project (displayed in public featured project gallery) or 
- an exemplar project (usually a curriculum writer owns the project and selects their own project as an example for users)

This PR adds an information icon in the personal projects gallery where a user's personal projects are listed in a table (at '/projects' for signed in users'). 

After update: project that is frozen (by an admin user) because it is an active feature project - information icon is displayed next to project title
Screenshot 2024-03-04 at 11 26 28 AM

When a user clicks on the icon, a project information dialog is displayed that notifies the user that the project is frozen and cannot be edited. However, if the user wishes to remove the project from the featured gallery, they can request via email.

After update: project info dialog is displayed after user clicks on information icon
Screenshot 2024-03-04 at 11 28 34 AM

In addition, a frozen project's actions are updated so that 'Remix' is the only option avaiable' ('Delete' and 'Rename' are removed). I added styling so that the info icon becomes a darker gray on hover and that the outline is displayed so that it is keyboard-navigable.

After update: screencast video showing updates described above.

https://github.com/code-dot-org/code-dot-org/assets/107423305/9733603d-3932-46ea-a60f-2e48806bd8ad

## Links
[jira](https://codedotorg.atlassian.net/browse/LABS-525)

## Testing story
I updated the `PersonalProjectsTable` storybook to include the `projectInfoDialog` as well as updating `PersonalProjectsTableTest.js`. The project info dialog icon is added and works as expected in the storybook.

## Deployment strategy
This is part 4 of 4 PRs that update the public gallery:
Part 1: Remove ability to publish projects https://github.com/code-dot-org/code-dot-org/pull/56666
Part 2: Update projects public gallery to display only featured projects https://github.com/code-dot-org/code-dot-org/pull/56886
Part 3: Add information banner about updated public gallery for signed-in users https://github.com/code-dot-org/code-dot-org/pull/56918
Part 4: Notify user if their project is selected to be featured since the project will be frozen https://github.com/code-dot-org/code-dot-org/pull/56932

These PRs will be merged to staging on the same day - the plan is we will merge during the week of March 18 in conjunction with communications to the public that @samantha-code is preparing.

## PR Checklist:

- [ ] Tests provide adequate coverage
- [ ] Privacy and Security impacts have been assessed
- [ ] Code is well-commented
- [ ] New features are translatable or updates will not break translations
[...]

PR #57023, opened 2024-03-05
Remove important rule from icon style

fisher-alice:
This PR removes the `!important` rule from the style rules of icons used in a table row within id `classroom-sections` or `uitest-personal-projects`.

This is no longer needed since `!important` is no longer used for icons in `application.scss`.

This change is motivated by updates made in https://github.com/code-dot-org/code-dot-org/pull/56932. See discussion [here](https://github.com/code-dot-org/code-dot-org/pull/56932/files#r1511566155)

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/templates/tables/QuickActionsCell.jsx
+++ b/apps/src/templates/tables/QuickActionsCell.jsx
@@ -145,13 +145,11 @@ const styles = {
     [QuickActionsCellType.body]: {
       border: '1px solid ' + color.white,
       borderRadius: 5,
-      color: color.darker_gray,
       margin: 3,
     },
     [QuickActionsCellType.header]: {
       fontSize: 20,
       lineHeight: '15px',
-      color: color.charcoal,
     },
   },
   hoverFocus: {
@@ -159,7 +157,6 @@ const styles = {
       backgroundColor: color.lighter_gray,
       border: '1px solid ' + color.light_gray,
       borderRadius: 5,
-      color: color.white,
     },
   },
 };
--- a/shared/css/phase1-design-system.scss
+++ b/shared/css/phase1-design-system.scss
@@ -160,7 +160,7 @@ p {
       background-color: $neutral_dark !important;
 
       i {
-        color: $neutral_white;
+        color: $neutral_white !important;
       }
     }
 
--- a/shared/css/phase1-design-system.scss
+++ b/shared/css/phase1-design-system.scss
@@ -160,7 +160,7 @@ p {
       background-color: $neutral_dark !important;
 
       i {
-        color: $neutral_white !important;
+        color: $neutral_white;
       }
     }
 
@@ -169,7 +169,7 @@ p {
       background-color: $neutral_light;
 
       i {
-        color: $neutral_dark !important;
+        color: $neutral_dark;
       }
     }
   }


bencodeorg (approved):
Nice! I assume you were able to confirm that `#classroom-sections` is unchanged?

fisher-alice:
> Nice! I assume you were able to confirm that `#classroom-sections` is unchanged?

Thanks for the reminder! I hadn't confirmed - just assumed it would be okay based on Denys's comment at https://github.com/code-dot-org/code-dot-org/pull/56932#discussion_r1512968227

Checking where `#classroom-sections` is used in `TeacherSections.jsx`, this does change the settings icon to a slightly lighter gray.

BEFORE update:
darker-gray-with-important

AFTER update:
lighter-gray-removed-icon-style

@levadadenys - could you let me know if this color change is okay? If not, I can implement in a different way. Thanks @bencodeorg !

Denys - I'm actually wondering if we can remove: https://github.com/code-dot-org/code-dot-org/blob/8703117a48404682017d9298c1db0bfc98d3a105/shared/css/phase1-design-system.scss#L171-L173.
Then styling will fall back to that of the parent `td`. Then there is no need to override in the module with `!important`.

levadadenys:
Wow, turned out to be harder then I though. Going to push changes I was able to make locally to improve overall situation. @fisher-alice feel free to revert my commit in case of anything.

Also, if/when - merging - we'll need to doublecheck the eye tests to confirm that there'll be no visual changes in the table since our `!important` styles from `phase1-design-system.scss` makes sure everything looks as we want it to look in rebranded version.

levadadenys:
Pushed a commit from me.
* added !important back for #classroom-sections, #uitest-personal-projects th 
* updated QuickActionsCell.jsx so that we don't need !important for #classroom-sections, #uitest-personal-projects td

Reason why I added !important again for `#classroom-sections, #uitest-personal-projects th` is that when I open branch locally, sort buttons/icons looked dark on my machine (caused by some inline styles from `reactabular-table` styles I wasn't able to remove)

Before:
Знімок екрана 2024-03-06 о 13 07 22

After:
Знімок екрана 2024-03-06 о 13 13 23


@fisher-alice answering your question about completely removing icon styles from `phase1-design-system.scss` - I think it's better to keep styles for update there so that it'll be easier to maintain and make sure everything is up to date.
Also, keeping in mind believe that we'll eventually update most of the screens with design system components/elements - 
I feel like it's more long term time-efficient not to rework old tables too much, but to create a new tables and use them sooner

Initial reason of adding `!important` - inline styles usage

levadadenys (commented):
Added a commit from me, which probably prevents me from approving PR since I'm a coauthor now 😄 

Left couple comments with context, I feel like this is good solution for the current situation.


a) Feel free to revert my commit
b) When/if we merge this PR - doublecheck eyes test to make sure nothing has changed visually (since current styles are approved rebrand styles)

fisher-alice:
> * added !important back for #classroom-sections, #uitest-personal-projects th
> * updated QuickActionsCell.jsx so that we don't need !important for #classroom-sections, #uitest-personal-projects td
> 
> Reason why I added !important again for `#classroom-sections, #uitest-personal-projects th` is that when I open branch locally, sort buttons/icons looked dark on my machine (caused by some inline styles from `reactabular-table` styles I wasn't able to remove)

Interesting to know where inline styles are coming from - thanks for checking!

> @fisher-alice answering your question about completely removing icon styles from `phase1-design-system.scss` - I think it's better to keep styles for update there so that it'll be easier to maintain and make sure everything is up to date. Also, keeping in mind believe that we'll eventually update most of the screens with design system components/elements - I feel like it's more long term time-efficient not to rework old tables too much, but to create a new tables and use them sooner
> 
> Initial reason of adding `!important` - inline styles usage

This makes sense - thanks so much for your guidance!

fisher-alice:
> Added a commit from me, which probably prevents me from approving PR since I'm a coauthor now 😄
> 
> Left couple comments with context, I feel like this is good solution for the current situation.
> 
> a) Feel free to revert my commit b) When/if we merge this PR - doublecheck eyes test to make sure nothing has changed visually (since current styles are approved rebrand styles)

This looks great! @bencodeorg - I've re-requested review since there's been an update to PR since you approved.

bencodeorg (approved):

levadadenys:
> Interesting to know where inline styles are coming from - thanks for checking!

I'm 95% sure they're coming from `reactabular-table` lib (we're using it for tables right now) inner functionality for Table.Head/Header for sortable headerCell

PR #55855, opened 2024-01-19
Add LTI Integration portal UI

nicklathe:
This PR adds the views for the LTI Integration portal. The primary view is the form which is used to create a new integration. Upon submit, another view is rendered that displays a success message or a message stating the integration already exists.

UPDATE: I've updated the screen shots to reflect the current button styling changes. These changes have been applied directly to the `application.scss`, and are not pulling in the official style guide classes in `apps/src/templates/buttons.module.scss`. The right thing to do here is to refactor this feature into a React component, rather than a traditional Rails view.

Primary view (`/lti/v1/integrations`)
Screenshot 2024-01-25 at 1 40 35 PM

Success view
Screenshot 2024-01-24 at 4 42 23 PM

Integration already exists view
Screenshot 2024-01-24 at 4 42 32 PM

Alert banner when submitting with required fields empty
Screenshot 2024-01-25 at 9 11 15 AM

## Links

- [Jira](https://codedotorg.atlassian.net/browse/P20-648)
- [Figma mock-ups](https://www.figma.com/file/DophYsrydeBxakbt2qfV18/LTI-Mock-ups?type=design&node-id=0%3A1&mode=design&t=SDqqlvAT9ZLJ4smS-1)

## Testing story

```
bundle exec spring testunit ./test/controllers/lti_v1_controller_test.rb
Running via Spring preloader in process 37704
Started with run options --seed 4639

/Users/nicklathe/Code/code-dot-org/code-dot-org/dashboard/app/models/experiments/experiment.rb:30: warning: already initialized constant MAX_CACHE_AGE-=---=---=---=---=---=---] 0% Time: 00:00:00,  ETA: ??:??:??
/Users/nicklathe/Code/code-dot-org/code-dot-org/dashboard/app/models/experiments/experiment.rb:30: warning: previous definition of MAX_CACHE_AGE was here
Since there is no EDITOR or BETTER_ERRORS_EDITOR environment variable, using Textmate by default.======================================================================       ] 96% Time: 00:00:39,  ETA: 00:00:02
  29/29: [===================================================================================================================================================================] 100% Time: 00:00:39, Time: 00:00:39

Finished in 40.69869s
29 tests, 55 assertions, 0 failures, 0 errors, 0 skips
```

## Follow-up work
The i18n strings need to be updated with URL's to link to various resources. Currently using a placeholder.

## PR Checklist:

- [ ] Tests provide adequate coverage
- [ ] Privacy and Security impacts have been assessed
- [ ] Code is well-commented
[...]

PR #57024, opened 2024-03-05
P20-750: LTI integration early access

artem-vavilov:
## DCDO `lti_early_access_limit`
`DCDO.get('lti_early_access_limit', false)` options:
- `false` — Early Access is disabled
- `true` — Early Access is enabled
- `0` — Early Access is closed
- `100` — Early Access is closed when the LTI integration count reaches the provided limit

## Early Access disabled 
Screenshot 2024-03-05 at 22 39 54

## Early Access enabled
Screenshot 2024-03-05 at 22 40 32

## Early Access closed
Screenshot 2024-03-05 at 22 41 10

## Links
- JIRA ticket: [P20-750](https://codedotorg.atlassian.net/browse/P20-750)
- Related JIRA ticket: [P20-753](https://codedotorg.atlassian.net/browse/P20-753)
- Related PR: https://github.com/code-dot-org/code-dot-org/pull/55855

### DIFF
--- a/dashboard/lib/policies/lti.rb
+++ b/dashboard/lib/policies/lti.rb
@@ -117,4 +117,17 @@ class Policies::Lti
   def self.roster_sync_enabled?(user)
     user.teacher? && user.lti_roster_sync_enabled
   end
+
+  def self.early_access?
+    DCDO.get('lti_early_access_limit', false).present?
+  end
+
+  def self.early_access_closed?
+    return unless early_access?
+
+    lti_early_access_limit = DCDO.get('lti_early_access_limit', false)
+    return false unless lti_early_access_limit.is_a?(Integer)
+
+    LtiIntegration.count >= lti_early_access_limit
+  end
 end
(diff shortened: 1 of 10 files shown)


dashboard/lib/policies/lti.rb
```diff
@@ -117,4 +117,17 @@ def self.show_email_input?(user)
   def self.roster_sync_enabled?(user)
     user.teacher? && user.lti_roster_sync_enabled
   end
+
+  def self.early_access?
+    DCDO.get('lti_early_access_limit', false).present?
+  end
+
+  def self.early_access_closed?
+    return unless early_access?
```
carl-codeorg, line 126: Is it better to consistently return a boolean, i.e. return `false` here? Or was this intentional to distinguish between the limit being reached (false) vs. early access not being enabled at all (nil)? Non-blocker, just curious

carl-codeorg (approved):
Just one non-blocker, otherwise LGTM! :shipit:

dashboard/lib/policies/lti.rb
artem-vavilov: Correct, it was done intentionally. If early access isn't enabled, then logically, it can't be closed.
Since `nil` is also a falsy value, it doesn't compromise the method's purpose but rather provides an extra indicator.

PR #57027, opened 2024-03-05
[AI Tutor] CT-230: Support filtering AI tutor interactions on user id

Erin007:
From @ebeastlake: I did a big refactor here from what the PR had originally, so in addition to the main diff, it might be useful to view the diff of the refactor: https://github.com/code-dot-org/code-dot-org/pull/57027/files/9528eaec0840167e207eaf3a408d4858bfbdb6e2..dcf308a21d3b4b652fffc3fb63e2d620a2e33b7a

We were moving in the direction of having
- GET /ai_tutor_interactions, and
- GET /ai_tutor_interactions/get_for_student

I don't think we need these to be separate routes. We have a concept of a current user, and we can have one route that allows users to fetch whatever chats they currently have access to -- whether that's their students (if they're a teacher) or themselves (if they're a student). We can accept all sorts of options/filters on this route -- userId, sectionId, and eventually pagination parameters -- though, in this case, I only accept the currently supported userId/sectionId filters from teachers and reject any student attempts at filtering. 

<details>
  <summary>See Erin's original PR description</summary>

[CT-230](https://codedotorg.atlassian.net/browse/CT-230) 

This PR adds a method to the ai_tutor_interactions_controller to fetch all of the AItTutorInteractions for a single student. 

### Follow up work 
We don't, as of yet, have a specific place that we want to display these messages; however, this work seemed like a useful first step toward: 
- determining how many times an individual student has been flagged for inappropriate use of AI Tutor 
- displaying student chats for a single student in the context of a level. If, for example, we want to do something like: 
Screenshot 2024-03-06 at 11 49 24 AM

- providing students a way to access their own chat history 
- giving teachers a some version of student-specific filtered view of chat messages since as is the the Student Chat Messages table on the Teacher Dashboard can be quite unwieldy  
</details>


dashboard/app/controllers/ai_tutor_interactions_controller.rb
```diff
@@ -81,4 +81,21 @@ def index
...
   end
+
+  # GET /ai_tutor_interactions/get_for_student
+  def get_for_student
+    params.require([:userId])
+    student = User.find(params[:userId])
+    return render(status: :not_found, json: {error: 'Student not found'}) unless student
+    teacher_can_view_chats = current_user.students.include?(student)
+    student_can_view_own_chats = student.id == current_user.id
+    return render(status: :forbidden, json: {error: 'This user does not have access to these chats'}) unless teacher_can_view_chats || student_can_view_own_chats
```
molly-moen: can/should this logic go in `ability.rb`? We do similar logic around code review comments there [link](https://github.com/code-dot-org/code-dot-org/blob/33df8e52769132afdefd02af06e5bda3dae775ec/dashboard/app/models/ability.rb#L144)

molly-moen (commented):
Looks good! One question about where we put permission logic.

apps/src/aiTutor/interactionsApi.ts
```diff
@@ -18,30 +24,46 @@ export async function savePromptAndResponse(
...
+  options: FetchAITutorInteractionsOptions
+) => {
+  const baseUrl = `/ai_tutor_interactions`;
+  let queryParams = '';
+
+  if (options.sectionId) {
+    queryParams += `sectionId=${options.sectionId}`;
+  }
+  if (options.userId) {
+    queryParams += `userId=${options.userId}`;
```
ebeastlake: TODO: Need to update to add ampersands.

molly-moen (approved):

PR #57029, opened 2024-03-05
Force LTI launch out of iframe

nicklathe:
This PR adds a new route, controller, and view to mitigate when Schoology tries to open us in an iframe. This new flow will prompt the user to open in a new tab, and continue the LTI launch there. This will only happen for Schoology, as Canvas does not attempt to open us in an iframe.

Below is a screen shot of the UI that are rendered in an iframe, prompting the user open in a new tab
Screenshot 2024-03-08 at 9 20 43 AM

## Links

[Jira](https://codedotorg.atlassian.net/browse/P20-760)

## Testing story
```
bundle exec spring testunit ./test/controllers/lti_v1_controller_test.rb
Running via Spring preloader in process 41839
Started with run options --seed 25940

Since there is no EDITOR or BETTER_ERRORS_EDITOR environment variable, using Textmate by default.==========================================================                   ] 89% Time: 00:00:24,  ETA: 00:00:03
  38/38: [===================================================================================================================================================================] 100% Time: 00:00:25, Time: 00:00:25

Finished in 26.02297s
38 tests, 77 assertions, 0 failures, 0 errors, 0 skips
```
```
bundle exec spring testunit ./test/lib/policies/lti_test.rb
Running via Spring preloader in process 9317
sh: force_iframe_launch?: command not found
Started with run options --seed 23111

DEPRECATED: Use assert_nil if expecting nil from test/lib/policies/lti_test.rb:65. This will fail in Minitest 6.=====                                                         ] 66% Time: 00:00:12,  ETA: 00:00:07
  9/9: [=====================================================================================================================================================================] 100% Time: 00:00:13, Time: 00:00:13

Finished in 13.09107s
9 tests, 14 assertions, 0 failures, 0 errors, 0 skips
```

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/dashboard/app/controllers/lti_v1_controller.rb
+++ b/dashboard/app/controllers/lti_v1_controller.rb
@@ -70,7 +70,7 @@ class LtiV1Controller < ApplicationController
   end
 
   def authenticate
-    unless params[:open_in_new_tab]
+    unless params[:new_tab]
       id_token = params[:id_token]
       state = params[:state]
       token_hash = {id_token: id_token, state: state}
@@ -201,63 +201,15 @@ class LtiV1Controller < ApplicationController
   # Detects if LMS is trying open Code.org in an iframe, prompt user to open in
   # new tab. The experience is unchanged to non-iframe users.
   def iframe
-    render html: <<~HTML.html_safe, layout: false
-      <!DOCTYPE html>
-      <html lang="en">
-      <head>
-      <meta charset="UTF-8">
-      <meta name="viewport" content="width=device-width, initial-scale=1.0">
-      <title>Opening Code.org in New Tab</title>
-      <script>
-        window.onload = function() {
-          // Get the current URL
-          const fullUrl = `${window.location.protocol}//${window.location.host}/lti/v1/authenticate`;
-          const urlParams = new URLSearchParams(window.location.search);
-          const idTokenParam = `id_token=${urlParams.get('id_token')}`;
-          const stateParam = `state=${urlParams.get('state')}`;
-          var newTabParam = "open_in_new_tab=true";
+    # Define your variables
+    auth_url_base = CDO.studio_url('/lti/v1/authenticate', CDO.default_scheme)
+    id_token_param = params[:id_token]
+    state_param = params[:state]
+    new_tab_param = 'new_tab=true'
 
-          // Append the new query parameter appropriately
-          const auth_url = `${fullUrl}?${idTokenParam}&${stateParam}&${newTabParam}`
-          // Check if the current window is the top-level window
-          if (window.location !== window.parent.location) {
-            // Create a div to hold the message and button
-            var messageDiv = document.createElement("div");
-            messageDiv.style.position = "fixed";
-            messageDiv.style.left = "0";
-            messageDiv.style.top = "0";
-            messageDiv.style.width = "100%";
-            messageDiv.style.height = "100%";
-            messageDiv.style.backgroundColor = "white";
-            messageDiv.style.zIndex = "10000";
-            messageDiv.style.display = "flex";
-            messageDiv.style.justifyContent = "center";
-            messageDiv.style.alignItems = "center";
-            messageDiv.style.flexDirection = "column";
-            messageDiv.style.textAlign = "center";
-            messageDiv.innerHTML = `
-              <p>Code.org cannot be run in an embeded window. Please open it in a new tab.</p>
-              <button onclick="openInNewTab()">Open in New Tab</button>
-            `;
-
-            // Function to open the current URL in a new tab
-            window.openInNewTab = function(url) {
-              // Open the new URL in a new tab
-              window.open(auth_url, '_blank');
-            };
-            // Append the message div to the body
-            document.body.appendChild(messageDiv);
-          } else {
-            // Redirect to auth_url endpoint
-            window.location.replace(auth_url);
-          }
-        };
-      </script>
-      </head>
-      <body>
-      </body>
-      </html>
-    HTML
+    # Construct the URL with interpolated query parameters
+    @auth_url = "#{auth_url_base}?id_token=#{id_token_param}&state=#{state_param}&#{new_tab_param}"
+    render 'lti/v1/iframe', layout: false
   end
 
   # GET /lti/v1/sync_course
--- a/dashboard/app/controllers/lti_v1_controller.rb
+++ b/dashboard/app/controllers/lti_v1_controller.rb
@@ -77,10 +77,11 @@ class LtiV1Controller < ApplicationController
     rescue => exception
       return log_unauthorized(exception)
     end
-    # client_id is the aud[ience] in the JWT
-    extracted_client_id = decoded_jwt_no_auth[:aud]
+    # client_id is the aud[ience] in the JWT, it can be a string or an array
+    extracted_client_id = decoded_jwt_no_auth[:aud].is_a?(Array) ? decoded_jwt_no_auth[:aud].first : decoded_jwt_no_auth[:aud]
     extracted_issuer_id = decoded_jwt_no_auth[:iss]
 
+    return log_unauthorized('Missing "aud" or "iss" from ID token') unless extracted_client_id.present? && extracted_issuer_id.present?
     # set cache key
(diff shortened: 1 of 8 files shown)


dashboard/app/controllers/lti_v1_controller.rb
```diff
@@ -109,6 +109,18 @@ def authenticate
       return log_unauthorized(exception)
     end
 
+    # Schoology has multiple contexts that will launch LTI tools in an iframe.
```
nicklathe, line 120: Initially I had this iframe check run for all LTI launches, at the top of the `authenticate` controller action. However, this has a couple downsides.

1. It redirects the request _before_ doing any jwt validation and authentication checks
2. It becomes harder to unit test this controller action, as the first thing it does is redirect to another page, unless the `new_tab` param is included in the request

I indexed on doing all the jwt validation and auth checks before redirecting, as I would rather deny the call before redirecting. The downside of this is now the redirect happens after a db call to retrieve the LTI Integration, which means this db request will happen two times now per LTI launch for Schoology.

I'm open to discussing further options here. In theory, we should be able to determine if the LMS is opening us in an iframe by the `https://purl.imsglobal.org/spec/lti/claim/launch_presentation` claim in the id_token, however in every id_token I've examined from both Canvas and Schoology, no matter the placement, this value is always `iframe`.

dashboard/app/views/lti/v1/iframe.html.haml
```diff
@@ -0,0 +1,40 @@
...
+    %meta{name: "viewport", content: "width=device-width, initial-scale=1.0"}
+    %title= t('lti.iframe_title')
+    :javascript
+      window.onload = function() {
+        const auth_url = `#{@auth_url}`;
+        const iframe_message = `#{t('lti.iframe_message')}`;
+        const iframe_button = `#{t('lti.iframe_button')}`;
+        // Determine if we are running in an iframe
+        if (window.location !== window.parent.location) {
+          var messageDiv = document.createElement("div");
```
nicklathe, line 14: I'm open to styling this differently, this currently gets the job done.

mgc1194: Could we get someone from the design team to do this for us?

nicklathe: We could. It's a little unorthodox the way we have to serve this page outside the typical layout, and build the ui dynamically with JS. I don't think we need to spend too much time on this UI, since it's just meant to get the user to open us in a new tab, but I agree it could use improvements.

I also don't want to block us for too long waiting on the design team to do this.

carl-codeorg: Could always have Ali follow up with UX and see if they have improvements to make. For now, I don't think we need it to look pretty

dashboard/app/controllers/lti_v1_controller.rb
carl-codeorg: Could we cache integrations for use in the auth routes? Would save us some lookups if all we need to do is check that it exists and lookup the various URLs. Might not be needed right away, but maybe down the road.

dashboard/app/controllers/lti_v1_controller.rb
```diff
@@ -109,6 +109,18 @@ def authenticate
...
+    # In this case, we will redirect to the iframe route, to prompt user to open
+    # in a new tab. This flow appends a 'new_tab=true' query param, so it will
+    # pass this block once the iframe "jail break" has happened.
+    if decoded_jwt[:iss] == Policies::Lti::LMS_PLATFORMS[:schoology][:issuer] && !params[:new_tab]
```
carl-codeorg: Would it be worth making a policy method like `forces_iframe_launch?` similar to what [Juan did for `issuer_accepts_resource_link`](https://github.com/code-dot-org/code-dot-org/pull/57068/files#diff-4a6088c854aa5f6c3deca26139c544ed84030aebfe3ca8d679def7fde62ef0f3R135)? Would cover Schoology ever having another issue, like how Canvas has a separate test issuer, or future issuers that also force iframes. Might be YAGNI, idk.

carl-codeorg (approved):
Overall looks great to me. Had a couple non-blocking comments

dashboard/app/controllers/lti_v1_controller.rb
nicklathe: Oh interesting. Mario had suggested the idea of caching on the first "round" then reading from the cache when the iframe flow redirects back here. I like the idea of caching integrations as a whole, to be used by multiple LTI launch requests though. We already have the write_cache and read_cache helper functions. It may be an over optimization right now, but it might be the right thing to do.

dashboard/app/controllers/lti_v1_controller.rb
nicklathe: Yeah, I considered moving this to a policy method. I'll do it.

mgc1194 (approved):
Carl's suggestions covered all my questions/concerns. I really appreciate your patience in the review process.

dashboard/app/controllers/lti_v1_controller.rb
```diff
@@ -78,10 +78,14 @@ def authenticate
...
-      return log_unauthorized('LTI integration not found', {client_id: extracted_client_id, issuer: extracted_issuer_id})
-    end
+    # set cache key
+    integration_cache_key = "#{extracted_issuer_id}/#{extracted_client_id}"
+
+    integration = read_cache(integration_cache_key) || LtiIntegration.find_by({client_id: extracted_client_id, issuer: extracted_issuer_id})
+    return log_unauthorized('LTI integration not found', {client_id: extracted_client_id, issuer: extracted_issuer_id}) unless integration
+    # Cache integration for fast retrieval on subsequent LTI launches. Set
+    # expires_in to 1 week
+    write_cache(integration_cache_key, integration, 1.week)
```
carl-codeorg, line 92: Nice, this is pretty cool that you can store the integration object this easily. Was afraid we'd have to do some weird JSON stringification/parsing

nicklathe: I'm doing that down in the `write_function`, stringifying on the way in, parsing on the way out. Had to add some logic to account for a nil cached value.

dashboard/app/controllers/lti_v1_controller.rb
```diff
@@ -122,14 +138,14 @@ def authenticate
       nrps_url = decoded_jwt[Policies::Lti::LTI_NRPS_CLAIM][:context_memberships_url]
       resource_link_id = decoded_jwt[Policies::Lti::LTI_RESOURCE_LINK_CLAIM][:id]
       deployment_id = decoded_jwt[Policies::Lti::LTI_DEPLOYMENT_ID_CLAIM]
-      deployment = Queries::Lti.get_deployment(integration.id, deployment_id)
+      deployment = Queries::Lti.get_deployment(integration[:id], deployment_id)
```
nicklathe, line 145: Since we are now getting the `integration` back from either the cache or the db, I had to change to bracket notation, since the cache doesn't return a class object like Active Record does, and therefore doesn't support dot notation.

carl-codeorg: Does the bracket + symbol notation work for both cached and ActiveRecord values?

nicklathe: It does, ActiveRecord has a `[]` method (see [here](https://api.rubyonrails.org/classes/ActiveRecord/AttributeMethods.html#method-i-5B-5D)). I don't see a way around this if we implement the cache.

dashboard/app/views/lti/v1/iframe.html.haml
nicklathe: I talked to Ali offline, he's created a jira ticket to follow up with UI enhancements. https://codedotorg.atlassian.net/browse/P20-769

carl-codeorg (approved):
Still LGTM. Great job finding a solution for the iframe issue!

dashboard/app/views/lti/v1/iframe.html.haml
alivira: added some designs to the ticket

PR #57032, opened 2024-03-05
Direct units to course congrats page

bethanyaconnor:
Instead of unit-specific congrats pages for units-within-a-course, we actually want to send users to the course congrats page. Standalone units should still go to the unit congrats page.

https://github.com/code-dot-org/code-dot-org/assets/46464143/a1b65cba-78a6-420d-abf4-c785c8c794be


TurnerRiley (approved):
🎉

PR #57035, opened 2024-03-05
Removing redundant babel plugin

jmkulwik:
Cleaning out some unused NPM packages

I noticed this one while clearing out another package.  `@babel/plugin-transform-classes` [was added](https://github.com/code-dot-org/code-dot-org/pull/8614) to keep us supporting IE 9/10. We don't support any version of IE anymore, including IE 11 so the original reasons for keeping it no longer apply. Additionally, this package is now imported and used by `@babel/preset-env`, which transpires our JS code to align to the ES5 spec. 

Some additional details: The `@babel/plugin-transform-classes` package rarely gets imported on its own anymore, so the documentation is a little sparse, however, the [NPM page](https://www.npmjs.com/package/@babel/plugin-transform-classes) says it gets used to "Compile ES2015 classes to ES5." (The [Babel page](https://babeljs.io/docs/babel-plugin-transform-classes) mostly discusses how to use it, rather than what it does.) However, looking at the [Babel docs](https://babeljs.io/docs/babel-preset-env) for `@babel/preset-env`, it looks like that is already done by preset-env. Babel [recommends](https://babeljs.io/docs/options#no-targets) setting a target when using `present-env`, which we do not do. Specifically: "When no targets are specified: Babel will assume you are targeting the oldest browsers possible. For example, @babel/preset-env will transform all ES2015-ES2020 code to be ES5 compatible." All of that to say, this appears to no longer be needed.

### DIFF
--- a/apps/babel.config.json
+++ b/apps/babel.config.json
@@ -9,8 +9,6 @@
     "@babel/plugin-proposal-object-rest-spread",
     "@babel/plugin-proposal-class-properties",
     ["@babel/plugin-transform-regenerator", {"async": true, "asyncGenerators": true}],
-    // needed for IE 9/10 support
-    ["@babel/plugin-transform-classes", {"loose": true}],
     "@babel/plugin-transform-modules-commonjs",
     "@babel/plugin-proposal-optional-chaining"
   ],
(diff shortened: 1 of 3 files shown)


jmkulwik:
(Looks like this package does get used, but I think the recommended way of setting this configuration has changed. I'll go ahead and investigate whether we can use the new method)

PR #57041, opened 2024-03-06
Remove OnRecordEvent block from AppLab

### LINKED ISSUE #57097: Deprecate OnRecordEvent
This will be fixed by PR #57041

cnbrenci:
OnRecordEvent is a block that utilized special functionality of firebase. Since we are deprecating the use of firebase in the backend, we are also deprecating this block. It is not used in curriculum, and we have [reason to believe](https://github.com/code-dot-org/code-dot-org/pull/54643#issuecomment-1809461422) only 0.02% of projects use it.

The user impact of removal is that any projects that include a call to the OnRecordEvent API (read 0.02% of projects) will fail with a console error. Users will need to remove the call from their projects and re-think their implementation using the available data APIs if they wish to continue using the project. 

Users will see the following error when running a project that uses this block after the deprecation:
Screenshot 2024-03-08 165816

ERROR: Line: 1: It looks like you're trying to use the onRecordEvent function. This function is no longer available. Try using the available data blocks.

## Links

- jira ticket: [Applab: deprecate OnRecordEvent block](https://codedotorg.atlassian.net/browse/LABS-485)
- fixes [#57097](https://github.com/code-dot-org/code-dot-org/issues/57097)
- [Zen Desk article](https://codeorg.zendesk.com/knowledge/editor/01HRTP6JCXK5J82G2T9GR8Q8DQ/en-us?brand_id=1346&section_id=13983818989709)
- [Slack announcement](https://codedotorg.slack.com/archives/C02KUS52U3X/p1710351008877529)
 
## Follow-up work

 - [x] Close out [this ticket](https://codedotorg.atlassian.net/browse/LABSIDEAS-198) referenced in some of the comments removed in this PR


dashboard/test/fixtures/script_level.yml
```diff
@@ -39653,28 +39653,6 @@ script_level_68_11305:
   seed_key: 
   activity_section_position: 12
   levels: level_11305
-script_level_68_557:
```
cnbrenci, line 39656: @dmcavoy 

Hi Dani! I noticed you have edited this file a bunch and I was wondering if you could help me answer some questions with the change I'm making. 

I'm having trouble finding where this fixture is used. If I just remove this, are there any side effects? Am I going to break some test somewhere like eyes tests that don't run in drone? Is there anything I need to go clean up? (Like the scripts in s3? I noticed this points to a level with code) 

Thank you!!!

dmcavoy: Fixtures are used in our tests to be able to mimic our course structure without actually seeding all of it. I think its safe to remove as long as tests pass when you remove it.

bencodeorg (approved):
LGTM! Is the thinking behind not doing a custom error message that explains the deprecation that it's such a small set of projects it's not worth doing?

cnbrenci:
> LGTM! Is the thinking behind not doing a custom error message that explains the deprecation that it's such a small set of projects it's not worth doing?

Hmm. It would actually be really easy to add a special case error message for it. But we'd have to come back and clean it up later. What say you @moneppo? 

image

moneppo:
I like adding a customized message as a good user experience for deprecated APIs. Some systems keep deprecation warnings around for years. And the suggested text you added is great!

PR #57044, opened 2024-03-06
Lazy load python lab in LabViewsRenderer

molly-moen:
We saw an issue a while ago where pyodide in Python Lab was causing crashes in Music Lab on old iOS versions. This is because the assets for Python Lab were being loaded for every Lab2 lab due us importing all lab views in `LabViewsRenderer`. In order to re-enable pyodide I wrapped `PythonlabView` in [lazy](https://react.dev/reference/react/lazy) so it will only be loaded when we actually view Python Lab. We are ok with python lab potentially not working on old iOS versions since it is a prototype (I also haven't been able to repro this error anymore so it may be resolved). 

Getting lazy loading to work with a TypeScript component and our webpack config was a little tricky. I had to re-export the component in an `index.js` file in order to get things to work properly. The syntax is a little funky, but not too long. 

Using `lazy` requires wrapping the lazily-loaded component in a [Suspense](https://react.dev/reference/react/Suspense) component with a fallback. I refactored our loading spinner that fades in after 3 seconds and shows a slow loading message after 11 seconds into a separate component so it could be reused.

## Links
- jira ticket: [CT-299](https://codedotorg.atlassian.net/browse/CT-299)

## Testing story
Tested locally that Lab2 labs still all load as expected, and only Python Lab loads the pyodide-specific assets. 

## Follow-up work
If we like this, we can re-use this pattern for the other Lab2 views. I think it would be best to isolate each lab's assets so we can feel comfortable prototyping new labs without causing bugs on existing labs.

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/lab2/views/Lab2Wrapper.tsx
+++ b/apps/src/lab2/views/Lab2Wrapper.tsx
@@ -12,17 +12,14 @@ import classNames from 'classnames';
 import moduleStyles from './Lab2Wrapper.module.scss';
 import ErrorBoundary from '../ErrorBoundary';
 import {LabState, isLabLoading, hasPageError} from '../lab2Redux';
-import FontAwesome from '@cdo/apps/templates/FontAwesome';
-const i18n = require('@cdo/locale');
 
 import {ErrorFallbackPage, ErrorUI} from './ErrorFallbackPage';
 import Lab2Registry from '../Lab2Registry';
+import Loading from './Loading';
 
 export interface Lab2WrapperProps {
   children: React.ReactNode;
 }
-
-const noFade = window.location.href.includes('lab2-no-fade');
 
 const Lab2Wrapper: React.FunctionComponent<Lab2WrapperProps> = ({children}) => {
   const isLoading: boolean = useSelector(isLabLoading);
@@ -31,14 +28,6 @@ const Lab2Wrapper: React.FunctionComponent<Lab2WrapperProps> = ({children}) => {
     (state: {lab: LabState}) =>
       state.lab.pageError?.errorMessage || state.lab.pageError?.error?.message
   );
-
-  const overlayStyle: string = noFade
-    ? isLoading
-      ? moduleStyles.noFadeLoading
-      : moduleStyles.noFadeLoaded
-    : isLoading
-    ? moduleStyles.fadeLoading
-    : moduleStyles.fadeLoaded;
 
   return (
     <ErrorBoundary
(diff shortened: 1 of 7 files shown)


apps/src/lab2/views/Lab2Wrapper.tsx
```diff
@@ -12,18 +12,15 @@ import classNames from 'classnames';
...
 
 import {ErrorFallbackPage, ErrorUI} from './ErrorFallbackPage';
 import Lab2Registry from '../Lab2Registry';
+import Loading from './Loading';
 
 export interface Lab2WrapperProps {
   children: React.ReactNode;
 }
 
-const noFade = window.location.href.includes('lab2-no-fade');
```
molly-moen, line 25: @breville @sanchitmalhotra126 this didn't seem likely to be useful anymore so I removed it. Let me know if I should keep it.

breville: This was a recent addition in https://github.com/code-dot-org/code-dot-org/pull/56732 and is functionality we still want.

molly-moen: whoops will add back!

apps/src/lab2/views/LabViewsRenderer.tsx
```diff
@@ -67,7 +82,14 @@ const appsProperties: {[appName in AppName]?: AppProperties} = {
   },
   pythonlab: {
     backgroundMode: false,
-    node: <PythonlabView />,
+    node: <div />,
+    lazyNode: lazy(() =>
+      import(
+        /* webpackChunkName: "pythonlab" */ '../../pythonlab/index.js'
```
sanchitmalhotra126, line 88: Out of curiosity - does this actually create a separate webpack chunk, or is this still technically part of the lab2.js chunk?

molly-moen: My understanding is yes. This comment actually does some magic so we get an asset request for `pythonlab.js` over the network (without it the request is for a file with a number as a name like `215.js`).

apps/src/lab2/views/LabViewsRenderer.tsx
```diff
@@ -9,16 +9,24 @@ import {setUpBlocklyForMusicLab} from '@cdo/apps/music/blockly/setup';
 import MusicView from '@cdo/apps/music/views/MusicView';
 import StandaloneVideo from '@cdo/apps/standaloneVideo/StandaloneVideo';
 import classNames from 'classnames';
-import React, {useContext, useEffect, useState} from 'react';
+import React, {
+  ComponentType,
+  LazyExoticComponent,
```
ebeastlake, line 14: What a name 😂

ebeastlake (approved):

PR #57046, opened 2024-03-06
React Testing Library: Update ProjectCardTest 

Erin007:
Baby steps! 

This PR updates `ProjectCardTest.js` to use React Testing Library.

A couple of notes: 
1.) I needed to update to check for an exact string match "Published:" versus just the substring "Published" 
2.) This is a lower level component, so potentially the test content here could get rolled into a parent component now that we're using deep rendering, but it maintains test coverage parity with the existing test for now.


levadadenys (approved):
LGTM! Thanks for this update!

PR #57047, opened 2024-03-06
Improve `bin/dotd` first time experience

snickell:
The first few times you do DotD, it can make you nervous. This improves two small details of the DotD experience that might help a little:

1. Print a link to the DotD doc when running `bin/dotd`
2. Improve `bin/dotd` error messages when missing CDO keys like CDO.github_access_key. Previously a completely static message was printed for any missing key. It wasn't clear which keys were actually missing, and instructions for fixing each type of error were jumbled together.

Before, not clear which key(s) were missing:
image

After, shows only the missing keys (two of them in this case):
image

### DIFF
--- a/bin/dotd
+++ b/bin/dotd
@@ -40,7 +40,7 @@ def check_for_cdo_keys
 
   puts <<~EOS.unindent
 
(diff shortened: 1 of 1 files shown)


cnbrenci (approved):
Thank you!!!!!!! This is really helpful.

PR #57051, opened 2024-03-06
Added a level-specific prompt field to AI Tutor section of levelbuilder

kakiha11:
- Updated the AI Tutor section of level edit pages to be a React component
- Added the a section for levelbuilders to add on a level-specific prompt to the system prompt

levelspecific

Added the prompt as a property to levels (see test level below) and added that prompt, when present, to the end of the default system prompt for all three modes of AI Tutor.

Screenshot 2024-03-05 at 8 24 55 PM

## Links

- jira ticket: [CT-238](https://codedotorg.atlassian.net/browse/CT-238?atlOrigin=eyJpIjoiODhmMWIwMDEyYmYzNGE2NmJkNmE3MmFkYWE2NDNiZTEiLCJwIjoiaiJ9)


apps/src/aiTutor/redux/aiTutorRedux.ts
```diff
@@ -70,6 +70,9 @@ export const askAITutor = createAsyncThunk(
     } else {
       systemPrompt = generalChatSystemPrompt;
     }
+    if (state.aiTutor.level?.levelSpecificPrompt) {
+      systemPrompt += state.aiTutor.level?.levelSpecificPrompt;
+    }
```
kakiha11: @Erin007 Based on the discussion we had during our AI Tutor sync, I decided to add this prompt to the end of all the system prompts instead of overriding just the general chat prompt like outlined in the jira task. It seemed like more inputs to the model was the direction we are currently heading.

dashboard/app/views/levels/editors/fields/_aitutor_fields.html.haml
```diff
@@ -1,5 +1,10 @@
...
 #ai_tutor.in.collapse
-  .field
-    = render partial: 'levels/editors/fields/checkboxes', locals: {f: f, field_name: :ai_tutor_available, description: "AI Tutor Available"}
+  
+  = f.label :ai_tutor_level_specific_prompt, 'Level-Specific Prompt'
+  %p Prompt used to specify the behavior of AI Tutor on a per level basis.
+  = f.text_area :ai_tutor_level_specific_prompt, placeholder: 'Write your level-specific prompt here. If left blank, this level will use the default system prompt.', rows: 4, class:"input-block-level"
+
+  = render partial: 'levels/editors/fields/checkboxes', locals: {f: f, field_name: :ai_tutor_available, description: "Make AI Tutor Available"}
+
```
Erin007: What do you think about switching the order so the availability checkbox is first and the input box is conditionally shown if the AI Tutor is available? Is there a scenario where a Levelbuilder would want to add a custom prompt for a level where AI Tutor is unavailable?

dashboard/app/views/levels/editors/fields/_aitutor_fields.html.haml
```diff
@@ -1,5 +1,10 @@
 %h1.control-legend{data: {toggle: "collapse", target: "#ai_tutor"}}
   AI Tutor
 #ai_tutor.in.collapse
-  .field
-    = render partial: 'levels/editors/fields/checkboxes', locals: {f: f, field_name: :ai_tutor_available, description: "AI Tutor Available"}
+  
+  = f.label :ai_tutor_level_specific_prompt, 'Level-Specific Prompt'
+  %p Prompt used to specify the behavior of AI Tutor on a per level basis.
+  = f.text_area :ai_tutor_level_specific_prompt, placeholder: 'Write your level-specific prompt here. If left blank, this level will use the default system prompt.', rows: 4, class:"input-block-level"
```
Erin007: Should we say what the default system prompt is? Right now it's "'You are a tutor in a high school classroom where the students are learning Java using the Code.org curriculum. Answer their questions in plain, easy-to-understand English. Do not write any code. Do not answer the question if it is not about Java or computer programming."  

I think it would be totally reasonable to punt on this for now, but might be a nice-to-have down the road depending on how things evolve. One advantage of re-writing these fields in React, if we have time/inclination etc, would be that we could pull in the base prompts from aiTutor/constants.ts to display for context.

apps/src/aiTutor/types.ts
```diff
@@ -38,6 +38,7 @@ export type Level = {
   hasValidation: boolean;
   isAssessment: boolean;
   isProjectBacked: boolean;
+  levelSpecificPrompt: string;
```
Erin007: maybe `levelSpecificPrompt?` to indicate that this is an optional field.

apps/src/sites/studio/pages/levels/show.js
```diff
@@ -61,14 +61,21 @@ function initPage() {
...
+      isProjectBacked,
+      levelSpecificPrompt,
+    } = aiTutorData;
     const level = {
       id: levelId,
       type: type,
       hasValidation: hasValidation,
       isAssessment: isAssessment,
       isProjectBacked: isProjectBacked,
+      levelSpecificPrompt,
```
Erin007, line 82: ohh can we shortcut like this or do we need `levelSpecificPrompt: levelSpecificPrompt`? If the short form works, let's make the fields consistent.

apps/src/aiTutor/redux/aiTutorRedux.ts
Erin007: Let's double check the spacing here. If the systemPrompt and levelSpecificPrompt are smashed together it might be hard for the AI to interpret.

Erin007:
Looking good!  A couple of questions for discussion 👍

apps/src/sites/studio/pages/levels/editors/fields/_ai_tutor_fields.js
```diff
@@ -0,0 +1,23 @@
+import $ from 'jquery';
+import React from 'react';
+import ReactDOM from 'react-dom';
+//import {useAppSelector} from '@cdo/apps/util/reduxHooks';
+import getScriptData from '@cdo/apps/util/getScriptData';
+import AITutorFields from '@cdo/apps/aiTutor/levelEditors/AITutorFields';
+
+$(document).ready(function () {
+  const isAvailable = getScriptData('aitutoravailable');
+  console.log('isAvailable', isAvailable);
```
molly-moen: should these console logs be removed?

apps/src/sites/studio/pages/levels/editors/fields/_ai_tutor_fields.js
```diff
@@ -0,0 +1,23 @@
+import $ from 'jquery';
+import React from 'react';
+import ReactDOM from 'react-dom';
+//import {useAppSelector} from '@cdo/apps/util/reduxHooks';
```
molly-moen: remove this?

apps/src/aiTutor/levelEditors/AITutorFields.tsx
```diff
@@ -0,0 +1,75 @@
...
+            }}
+          />
+        </div>
+        <br />
+
+        <input
+          type="hidden"
+          id="level_ai_tutor_level_specific_prompt"
+          name="level[ai_tutor_level_specific_prompt]"
+          value={levelSpecificPrompt.toString()}
```
molly-moen: do we need `toString` here? The value is already a string, right?

apps/src/aiTutor/levelEditors/AITutorFields.tsx
```diff
@@ -0,0 +1,75 @@
...
+        />
+        {aiTutorAvailable && (
+          <div className={moduleStyles.fieldRow}>
+            <div className={moduleStyles['field-value']}>
+              <label htmlFor="levelPrompt">Level-Specific Prompt</label>
+              <textarea
+                id="levelPrompt"
+                value={levelSpecificPrompt || ''}
+                className={moduleStyles.textarea}
+                placeholder={`Enter a prompt specific to this level to be added with the default AI Tutor system prompt:\n${generalChatSystemPrompt}`}
```
molly-moen: would it be more helpful to have this information above the input box so users can see the general prompt while they are writing, rather than before they start?

dashboard/app/views/levels/editors/fields/_ai_tutor_fields.html.haml
```diff
@@ -0,0 +1,13 @@
...
+  AI Tutor
+#ai-tutor-container.in.collapse
+
+%script{src: webpack_asset_path('js/levels/editors/fields/_ai_tutor_fields.js'),
+  data: {aitutoravailable: (@level.properties["ai_tutor_available"] || true).to_json, aitutorlevelspecificprompt: (@level.properties["ai_tutor_level_specific_prompt"] || '').to_json}}
+
+  -# data: {aicustomizations: (@level.properties["initial_ai_customizations"] || {}).to_json, levelname: @level.name, appname: @level.game&.app}
+    = f.label :ai_tutor_level_specific_prompt, 'Level-Specific Prompt'
+    %p Prompt used to specify the behavior of AI Tutor on a per level basis.
+    = f.text_area :ai_tutor_level_specific_prompt, placeholder: 'Write your level-specific prompt here. If left blank, this level will use the default system prompt.', rows: 4, class:"input-block-level"
```
molly-moen: for my understanding: why do we need this and the react version?

molly-moen (commented):
Looks good! A couple clean up comments and one question for my understanding

apps/src/aiTutor/levelEditors/AITutorFields.tsx
kakiha11: When I tested it without `toString` it didn't work and when I added it, then it worked. I can't say I understand why at this point.

dashboard/app/views/levels/editors/fields/_ai_tutor_fields.html.haml
kakiha11: Good catch @molly-moen! That was the old version that I had commented out, I will remove.

apps/src/aiTutor/levelEditors/AITutorFields.tsx
molly-moen: I wonder if the issue was that you needed `toString` on `aiTutorAvailable` because it's a boolean, but here you shouldn't need it. I tested locally and removed `toString` here and it seemed to work.

apps/src/aiTutor/levelEditors/AITutorFields.tsx
kakiha11: @molly-moen What do you think of this? 
image

apps/src/sites/studio/pages/levels/editors/fields/_ai_tutor_fields.js
kakiha11: Removed.

apps/src/aiTutor/levelEditors/AITutorFields.tsx
kakiha11: Thanks! Made this update.

apps/src/aiTutor/levelEditors/AITutorFields.tsx
molly-moen: that looks good!

molly-moen (approved):
Nice work!

PR #57059, opened 2024-03-06
Redirect studio.code.org/courses to code.org/students

kelbyhawn:
Redirects [studio.code.org/courses](https://studio.code.org/courses) to [code.org/students](https://code.org/students).
- Replaces the links in the top nav
- Redirects everywhere else
  - You can test this by going to http://localhost-studio.code.org:3000/courses
- This change resulted in needing to update multiple Unit and UI tests

**Followup:** I'll deprecate the `Courses.jsx` component and other related components in a separate PR.

**Jira ticket:** [ACQ-1271](https://codedotorg.atlassian.net/browse/ACQ-1271)

----

## Signed out nav
Screenshot 2024-03-06 at 10 34 12 AM

## Signed in nav
Screenshot 2024-03-06 at 10 33 13 AM

## Redirect example

https://github.com/code-dot-org/code-dot-org/assets/9256643/7126c896-7228-4837-b7bd-e9b88ab43b20


pegasus/sites.v3/code.org/public/students.haml
```diff
@@ -3,18 +3,20 @@ title: Learn Computer Science
...
 %link{href: "/css/generated/design-system-pegasus.css", rel: "stylesheet"}
 
 %section
   .wrapper
     %h1=hoc_s(:heading_anyone_can_learn)
     %p.body-one.no-margin-bottom
       =hoc_s(:student_courses_top_desc)
-    -# rubocop:disable CustomCops/DashboardDbUsage
-    - if !current_user
+    - if !user_type
```
davidsbailey, line 18: Beautiful! Thank you Kelby!

davidsbailey (approved):
Woohoo, thanks for thinking of this and fixing it!

pegasus/sites.v3/code.org/public/students.haml
kelbyhawn: You're welcome!

pegasus/sites.v3/code.org/public/students.haml
```diff
@@ -3,18 +3,20 @@ title: Learn Computer Science
 theme: responsive_full_width
 ---
 
+-# Determine if a user is signed in
+- cookie_key = environment_specific_cookie_name("_user_type")
+- user_type = request.cookies[cookie_key]
```
bethanyaconnor, line 8: Just checking -- have you tested this locally? I was looking at a different `userType` thing recently and the user type for signed out users was "unknown". That was in our redux store instead of in the cookie though, so definitely doesn't mean that the same behavior applies here!

kelbyhawn: Yep! I tested locally, and it worked as expected.

bethanyaconnor (approved):

bethanyaconnor:
We may also want to change the redirect that happens when a signed out user goes to studio.code.org (that code is [here](https://github.com/code-dot-org/code-dot-org/blob/1f59a11a940e1cf79d8373a934de9db74b91651e/dashboard/app/controllers/home_controller.rb#L62)). We could update it to go to code.org/students but I don't know if we'd prefer to direct users to the sign in page instead?

dmcavoy:
> We may also want to change the redirect that happens when a signed out user goes to studio.code.org (that code is [here](https://github.com/code-dot-org/code-dot-org/blob/1f59a11a940e1cf79d8373a934de9db74b91651e/dashboard/app/controllers/home_controller.rb#L62)). We could update it to go to code.org/students but I don't know if we'd prefer to direct users to the sign in page instead?

Ya I've been wondering about this choice and why we have studio.code.org go to studio.code.org/courses for signed out users. I think it makes more sense to point them to sign in.

kelbyhawn:
> I think it makes more sense to point them to sign in.

@bethanyaconnor @dmcavoy I'll make this change too! ~How can I test this locally?~ Nevermind!

kelbyhawn:
@bethanyaconnor one other thing: this test is failing, and I tried updating line 69 to `Then I wait until I am on "http://studio.code.org/home"` but it still failed: https://github.com/code-dot-org/code-dot-org/blob/73e677ea888134d31c003af311942182b8e2ab06/dashboard/test/ui/features/foundations/signing_in.feature#L65-L71

Is this test still relevant now that the /courses page will be deprecated? Or is there something else that needs to change?

bethanyaconnor:
The signing_in test is still relevant -- we likely want to test what happens when a user types in an invalid section code. Looking at [the video](https://app.saucelabs.com/tests/76cc0ced7b2145a3b7d985fa2271da78#276), it redirects to code.org/students (not surprising), but we lose the message that a section wasn't found, which feels like a slight regression. We could [update this behavior](https://github.com/code-dot-org/code-dot-org/blob/73e677ea888134d31c003af311942182b8e2ab06/dashboard/app/controllers/followers_controller.rb#L100) to send the user to https://studio.code.org/join and show the same alert text. @dmcavoy thoughts?


For context, this is how this currently works in production:

https://github.com/code-dot-org/code-dot-org/assets/46464143/5cfe8e11-c247-4fb8-80a3-01b774352fa7

kelbyhawn:
Here's the updated experience from https://github.com/code-dot-org/code-dot-org/pull/57059/commits/309889cecf1c37a4ef06012348f7406b51d94f15 :

https://github.com/code-dot-org/code-dot-org/assets/9256643/37d18b33-c01f-46d6-9e5a-64ae536de1fe

dmcavoy:
Love the updated pattern! Thank you @kelbyhawn and @bethanyaconnor for thinking about the users!

bethanyaconnor (approved):
🎉 thank you so much for all your work on this!

kelbyhawn:
> 🎉 thank you so much for all your work on this!

Thanks for all your guidance on fixing the tests! (Also shoutout to @hannahbergam for Sauce Labs help!)

PR #57063, opened 2024-03-06
Modify icon sizes for legend and pop-up

kobryan0619:
This PR:

- Modifies the icon size for the pop-up and the legend to be 16px (vs. 18px in the table)
- Changes the color of the `CHOICE_LEVEL` icon to be black in the pop-up and legend
- Changes the background color of the boxes and the min height

image

Here's the change to the background color and min-height:

AFTER:
image

BEFORE:
image

## Links

[Ticket](https://codedotorg.atlassian.net/jira/software/projects/TEACH/boards/65?selectedIssue=TEACH-780)
[Polish ticket](https://codedotorg.atlassian.net/jira/software/projects/TEACH/boards/65?selectedIssue=TEACH-897)
[Figma](https://www.figma.com/file/gCktlnafGYQSNJafD3FUgA/Progress-Tracking?node-id=4128%3A5805&mode=dev)


apps/src/templates/sectionProgressV2/ItemType.jsx
```diff
@@ -7,7 +7,7 @@ export const ITEM_TYPE = Object.freeze({
   NEEDS_FEEDBACK: 3,
   FEEDBACK_GIVEN: 4,
   ASSESSMENT_LEVEL: ['star', color.neutral_dark],
-  CHOICE_LEVEL: ['split', color.neutral_dark60],
+  CHOICE_LEVEL: ['split', color.neutral_dark60, color.neutral_dark],
```
kobryan0619: This felt like the least invasive way to change this color.  It does mean it doesn't match the pattern of other icons here, but I could see a world where we rely on more colors in the future too...

apps/src/templates/sectionProgressV2/ProgressIcon.jsx
```diff
@@ -52,8 +52,12 @@ export default function ProgressIcon({itemType}) {
...
           id={'uitest-' + itemType[0]}
           icon={itemType[0]}
-          style={{color: itemType[1]}}
-          className={styles.fontAwesomeIcon}
+          style={
+            className && itemType === ITEM_TYPE.CHOICE_LEVEL
+              ? {color: itemType[2]}
+              : {color: itemType[1]}
+          }
+          className={className ? className : styles.fontAwesomeIcon}
```
kobryan0619: Currently every time we are using a different `className` it is because it is coming from the legend or pop-up.  I also considered creating a prop like `isForLegend` that would then determine the className but again could imagine a scenario where we want to use different classNames in the future.

apps/src/templates/sectionProgressV2/MoreDetailsDialog.jsx
```diff
@@ -12,11 +12,12 @@ import SafeMarkdown from '@cdo/apps/templates/SafeMarkdown';
 import {ITEM_TYPE} from './ItemType';
 import ProgressIcon from './ProgressIcon';
 import styles from './progress-key-popup.module.scss';
+import iconStyles from './progress-table-legend.module.scss';
 
 export default function MoreDetailsDialog({hasValidation, onClose}) {
-  const renderItem = (itemType, itemTitle, itemDetails) => (
+  const renderItem = (itemType, itemTitle, itemDetails, className) => (
```
etaderhold: For most of the icons you can omit this last parameter, and pass in `color.neutral_dark` for the choice level icon.
```suggestion
  const renderItem = (itemType, itemTitle, itemDetails, colorOverride) => (
```

apps/src/templates/sectionProgressV2/MoreDetailsDialog.jsx
```diff
@@ -12,11 +12,12 @@ import SafeMarkdown from '@cdo/apps/templates/SafeMarkdown';
...
     <div className={styles.item}>
-      <ProgressIcon itemType={itemType} />
+      <ProgressIcon itemType={itemType} className={className} />
```
etaderhold:
```suggestion
      <ProgressIcon itemType={itemType} colorOverride={colorOverride} />
```

apps/src/templates/sectionProgressV2/MoreDetailsDialog.jsx
```diff
@@ -89,12 +97,14 @@ export default function MoreDetailsDialog({hasValidation, onClose}) {
...
-          i18n.progressLegendDetailsAssessmentLevels()
+          i18n.progressLegendDetailsAssessmentLevels(),
+          iconStyles.fontAwesomeIconInLegend
         )}
         {renderItem(
           ITEM_TYPE.CHOICE_LEVEL,
           i18n.choiceLevel(),
-          i18n.progressLegendDetailsChoiceLevels()
+          i18n.progressLegendDetailsChoiceLevels(),
+          iconStyles.fontAwesomeIconInLegend
```
etaderhold:
```suggestion
          color.neutral_dark
```

apps/src/templates/sectionProgressV2/ProgressIcon.jsx
```diff
@@ -7,7 +7,7 @@ import classNames from 'classnames';
 
 export const PROGRESS_ICON_TITLE_PREFIX = 'progressicon-';
 
-export default function ProgressIcon({itemType}) {
+export default function ProgressIcon({itemType, className}) {
```
etaderhold:
```suggestion
export default function ProgressIcon({itemType, colorOverride}) {
```

apps/src/templates/sectionProgressV2/ProgressIcon.jsx
```diff
@@ -52,8 +52,12 @@ export default function ProgressIcon({itemType}) {
         <FontAwesome
           id={'uitest-' + itemType[0]}
           icon={itemType[0]}
-          style={{color: itemType[1]}}
-          className={styles.fontAwesomeIcon}
+          style={
+            className && itemType === ITEM_TYPE.CHOICE_LEVEL
+              ? {color: itemType[2]}
+              : {color: itemType[1]}
```
etaderhold:
```suggestion
          style={{color: colorOverride ? colorOverride : itemType[1]}}
```

apps/src/templates/sectionProgressV2/ProgressIcon.jsx
etaderhold: Instead of passing in a new class name for each icon to override the styles, I think it's cleaner to just assign different styles to the existing `fontAwesomeIcon` class depending on its particular location. See the suggestion in the CSS file for how this could fix up the legend. You'd need to do something similar for the popup window.

apps/src/templates/sectionProgressV2/progress-table-legend.module.scss
```diff
@@ -31,6 +31,7 @@
   white-space: nowrap;
   margin-right: 16px;
   margin-bottom: 16px;
+  background-color: white;
```
etaderhold, line 35:
```suggestion
  background-color: white;
  
  .fontAwesomeIcon {
    font-size: 16px;
    padding: 0 2px 0 2px;
    vertical-align: middle;
  }
```

apps/src/templates/sectionProgressV2/progress-table-legend.module.scss
```diff
@@ -78,6 +80,12 @@
   line-height: normal;
 }
 
+.fontAwesomeIconInLegend {
+  font-size: 16px;
+  padding: 0 2px 0 2px;
+  vertical-align: middle;
+}
+
```
etaderhold:
```suggestion
```

apps/src/templates/sectionProgressV2/ItemType.jsx
etaderhold: I suggest a `colorOverride` parameter instead. We'll keep the defaults here, and anywhere that we need to use a different color we can specify that when rendering the `ProgressIcon` component.

apps/src/templates/sectionProgressV2/ProgressIcon.jsx
```diff
@@ -1,13 +1,15 @@
 import React from 'react';
+import PropTypes from 'prop-types';
 import {ITEM_TYPE, ITEM_TYPE_SHAPE} from './ItemType';
 import styles from './progress-table-legend.module.scss';
+import popUpStyles from './progress-key-popup.module.scss';
```
kobryan0619: Since the styles for the pop-up are in another .scss file, I added this second style file here.  TBH, I haven't really seen this anywhere else in our codebase, but I also cannot see any reason why we cannot do this... the alternative was to collapse the two css files into one but that lead to a rather lengthy single css file.

etaderhold: Looks like we have four SCSS module files in this directory, and this wouldn't be the first time we've included at least two of them in a single component. For that reason I wouldn't be opposed at all to merging them together as they don't seem to be cleanly separated anyway, but this way works too!

etaderhold (approved):

apps/src/templates/sectionProgressV2/ProgressIcon.jsx
kobryan0619: I wonder if this is something we can talk about as a team - I kinda like having smaller `.scss` files because it makes it easier for me to read, but having multiple import statements for styles is suspicious.  I looked into alternatives and it seems like `mixins` might solve the problem here, but we also don't use them a ton in our codebase.

I am wondering if we want to have a team norm on when do you have separate .scss files and when to combine.

PR #57068, opened 2024-03-06
Parsing for Schoology nrps response

juanmanzojr:
Adds support for the Schoology nrps response. Schoology does not accept the `rlid` while retrieving the membership of a context thus changing the structure of the nrps_response.
This PR provides a separate flow for parsing the nrps response for those platforms that do not offer a resource link.

There is a new method within _dashboard/lib/policies/lti.rb_  `self.issuer_accepts_resource_link?` in which we can add those platforms that do not accept a Resource Link level membership service when retrieving membership for a context.

## Links

- jira ticket: [here](https://codedotorg.atlassian.net/jira/software/c/projects/P20/boards/69?assignee=712020%3Af0e2716f-09f9-4686-8490-9f1f9a08bb06&selectedIssue=P20-700)

## Testing story
Unit tests


carl-codeorg:
Curious how we get around the [params.require for rlid ](https://github.com/code-dot-org/code-dot-org/blob/staging/dashboard/app/controllers/lti_v1_controller.rb#L201) in the `sync_course` route. On a launch from Schoology, what value does the param have?

juanmanzojr:
> Curious how we get around the [params.require for rlid ](https://github.com/code-dot-org/code-dot-org/blob/staging/dashboard/app/controllers/lti_v1_controller.rb#L201) in the `sync_course` route. On a launch from Schoology, what value does the param have?

Sorry possibly worded this wrong, the rlid is present in the params but when trying to use it to retrieve the context membership we get the following error: `Error getting context membership: 400 {"error":"Invalid parameter: \"query.rlid\" is not allowed"}`

dashboard/lib/services/lti.rb
```diff
@@ -92,28 +92,34 @@ def self.initialize_lti_student_from_nrps(client_id:, issuer:, nrps_member:)
...
-  def self.parse_nrps_response(nrps_response)
+  def self.parse_nrps_response(nrps_response, issuer)
     sections = {}
     members = nrps_response[:members]
     context_title = nrps_response.dig(:context, :title)
     members.each do |member|
       next if member[:status] == 'Inactive' || member[:roles].exclude?(Policies::Lti::CONTEXT_LEARNER_ROLE)
       # TODO: handle multiple messages. Shouldn't be needed until we support Deep Linking.
-      message = member[:message].first
+      if Policies::Lti.issuer_accepts_resource_link?(issuer)
```
carl-codeorg, line 104: A comment might be helpful to explain the significance of this check, which is that if the LMS hasn't implemented the resource link level membership service, we don't get the `message` property in the member object, if I understand the docs correctly

carl-codeorg (approved):
Great job; LGTM once tests are passing

dashboard/lib/services/lti.rb
```diff
@@ -92,28 +92,36 @@ def self.initialize_lti_student_from_nrps(client_id:, issuer:, nrps_member:)
...
+        member_section_ids = [nrps_response.dig(:context, :id)]
+        member_section_names = [nil]
+      end
       member_section_ids.each_with_index do |section_id, index|
         if sections[section_id].present?
           sections[section_id][:members] << member
         else
           sections[section_id] = {
-            name: "#{context_title}: #{member_section_names[index]}",
+            name: member_section_names[index].nil? ? context_title.to_s : "#{context_title}: #{member_section_names[index]}",
```
nicklathe, line 125: Nit: might be helpful to have a comment here explaining that Schoology already returns the context_title in the format want?

nicklathe (approved):
Looks good! I just had a single non-blocking comment.

PR #57072, opened 2024-03-06
Convert ControlButtonsTest to React Testing Library

molly-moen:
Convert Java Lab's control buttons test to React Testing Library. This was pretty straightforward to convert! Some of the tests were checking for the value of the `onClick` attribute on the finish/submit button, which didn't seem very useful to me as the `onClick` attribute is controlled by the same logic that controls the button text, which we are already testing. Therefore I removed that check.

## Testing story
Tested that the new version of the tests passed.

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/test/unit/javalab/ControlButtonsTest.js
+++ b/apps/test/unit/javalab/ControlButtonsTest.js
@@ -39,50 +39,38 @@ describe('Java Lab Control Buttons Test', () => {
     restoreRedux();
   });
 
-  it('submit button for unsubmitted submittable level', () => {
-    render(
+  function renderDefault(props) {
+    return render(
       <Provider store={store}>
-        <ControlButtons {...defaultProps} isSubmittable isSubmitted={false} />
+        <ControlButtons {...defaultProps} {...props} />
       </Provider>
     );
+  }
+
+  it('submit button for unsubmitted submittable level', () => {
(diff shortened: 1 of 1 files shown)


apps/test/unit/javalab/ControlButtonsTest.js
```diff
@@ -21,49 +30,59 @@ describe('Java Lab Control Buttons Test', () => {
...
+
+  afterEach(() => {
+    restoreRedux();
   });
 
   it('submit button for unsubmitted submittable level', () => {
-    const wrapper = shallow(
-      <ControlButtons {...defaultProps} isSubmittable isSubmitted={false} />
+    render(
+      <Provider store={store}>
```
ebeastlake, line 44: nit: IIRC, I think Liam used a custom render method to encapsulate the Provider setup (this feels more "natural" to me than including it every time). Do you think we should use one here? Something like...

```
function renderDefault(propOverrides = {}) {
  render(
    <Provider store={store}>
      <ControlButtons {...defaultProps} {...propOverrides} />
    </Provider>
  );
}
```

ebeastlake (approved):
One nit, but LGTM!

apps/test/unit/javalab/ControlButtonsTest.js
molly-moen: good call, will update

PR #57074, opened 2024-03-06
Gen AI: add remaining levelbuilder editing capabilities

sanchitmalhotra126:
Now that the boilerplate is in place, this adds the full levelbuilder editing UI, working end-to-end. We can definitely refine this as needed (e.g. collapsible sections, different wording, etc) but this gives us a way to start adding this info to levels.

Screenshots (edit: update) (@dancodedotorg let me know what you think!):
Screenshot 2024-03-08 at 12 08 43 PM (2)
Screenshot 2024-03-08 at 12 08 49 PM (2)
Screenshot 2024-03-08 at 12 08 58 PM (2)
Screenshot 2024-03-08 at 12 09 06 PM (2)

Once saved, this data appears in level_properties:
Screenshot 2024-03-08 at 12 10 18 PM

## Links

https://codedotorg.atlassian.net/browse/LABS-516

## Testing story

Tested locally with editing and creating a new AI Chat level. Also tested creating and editing Music Lab levels to confirm no impact as a result of the changes to levels_controller.


apps/src/aichat/constants.ts
```diff
@@ -12,3 +12,8 @@ const initialChatMessages: ChatCompletionMessage[] = [
 ];
 
 export {initialChatMessages};
+
+export const MIN_TEMPERATURE = 0;
+export const MAX_TEMPERATURE = 2;
+export const MAX_RETRIEVAL_CONTEXTS = 20;
+export const MAX_ASK_ABOUT_TOPICS = 10;
```
sanchitmalhotra126, line 19: These limits are somewhat arbitrary - we can update these as needed.

dancodedotorg:
Not my lane, potential nit: do you mean for the temperature value to be a string?
image

apps/src/lab2/levelEditors/aiCustomizations/EditAiCustomizations.tsx
```diff
@@ -37,75 +81,161 @@ const EditAiCustomizations: React.FunctionComponent<
...
+          description="The base prompt applied to all inputs to the model."
+          inputType="textarea"
+        />
+        <FieldSection
+          fieldName="retrievalContexts"
+          labelText="Retrieval Contexts"
+          description="Pieces of information the model references when generating a response."
+          inputType="custom"
+          inputNode={
+            <MultiItemInput
```
dancodedotorg, line 168: Nit: if there's a way to reduce the default vertical space for these textboxes for retrieval specifically, I think I'd prefer that. I'd rather be able to see most of the retrieval items together on the screen rather than scroll to view them, but the current vertical default makes me think I'll quickly get overloaded with vertical space if I add too many.

I guess to test that, what would it look like with the 5 retrieval sentences I added in my Tucson bot prototype? How much vertical space is added, and how much of it is mostly empty?
image

apps/src/lab2/levelEditors/aiCustomizations/ModelCardFields.tsx
```diff
@@ -0,0 +1,62 @@
...
+              value={aiCustomizations.modelCardInfo?.value?.[property] || ''}
+              onChange={e =>
+                setModelCardPropertyValue(property, e.target.value)
+              }
+              className={moduleStyles.textarea}
+            />
+          </div>
+        );
+      })}
+      <label>Ask About Topics</label>
```
dancodedotorg: In some recent mockups we sent to Mark, Sam and I settled on these being called "Example Prompts and Topics" for the model card. Even though the design isn't settled, that phrasing feels set - can we update this label to match that?

dancodedotorg:
In general, vertical space starts to become an issue in levelbuilder very very quickly, and will likely be at the heart of different types of cosmetic feedback I have. Looking at this now, I'm worried about how much space Retrieval and Model Card can take up on the page. Would it be possible to make those headers collapsible? Or, if I'm understanding the layers of React components correctly, would it be possible to make all of these sub-headings collapsible (Bot Name, Temperature, System Prompt, etc)?

dancodedotorg (approved):
Left some language / formatting comments and a slight nit, but nothing blocking. LGTM! 

And - as clarified in Slack - this doesn't include fine tuning / curation updates that we're still settling. So once those get settled and spec'd, I imagine some new controls would be added!

dashboard/app/views/levels/editors/_aichat.html.haml
```diff
@@ -1,4 +1,3 @@
-= render partial: 'levels/editors/fields/aichat_fields', locals: {f: f}
```
bencodeorg, line 1: Should we remove this partial to avoid confusion?

dashboard/app/controllers/levels_controller.rb
```diff
@@ -576,6 +567,20 @@ def embed_level
     params[:level].permit(permitted_params)
   end
 
+  private def handle_json_params(level_params)
+    # Parse a few specific JSON fields used by modern (Lab2) labs so that they are
```
bencodeorg, line 571: Move comment above function? Out of scope for this PR since you're just moving stuff around, but do we test this at all?

apps/src/lab2/levelEditors/aiCustomizations/ModelCardFields.tsx
```diff
@@ -0,0 +1,62 @@
...
+  ['limitationsAndWarnings', 'Limitations and Warnings'],
+  ['testingAndEvaluation', 'Testing and Evaluation'],
+];
+
+const ModelCardFields: React.FunctionComponent = () => {
+  const {setModelCardPropertyValue, aiCustomizations} =
+    useContext(UpdateContext);
+  const askAboutTopics = aiCustomizations.modelCardInfo?.value?.askAboutTopics;
+  return (
+    <div className={moduleStyles['model-card-fields']}>
```
bencodeorg, line 20: This looks slick, and very similar (but better 😆 ) to what I had on the level side...should we see if we can reuse these across levelbuilder and the student-facing UI?

apps/src/lab2/levelEditors/aiCustomizations/UpdateContext.ts
```diff
@@ -0,0 +1,23 @@
...
+  AiCustomizations,
+  LevelAiCustomizations,
+  ModelCardInfo,
+  Visibility,
+} from '@cdo/apps/aichat/types';
+import {createContext} from 'react';
+
+export const UpdateContext = createContext({
+  aiCustomizations: {} as LevelAiCustomizations,
+  setPropertyValue: (
```
bencodeorg, line 11: For my understanding, can you point me towards any docs that explain this pattern? I'm looking at the basic [React `createContext` docs](https://react.dev/reference/react/createContext#createcontext) and doesn't seem to demonstrate this sort of setter approach (or maybe it does and I'm being dense).

apps/src/lab2/levelEditors/aiCustomizations/ModelCardFields.tsx
bencodeorg: I didn't see see this section in the mocks I was working off of on the student side...maybe the Figma I was using is out of date? [This is what I was using](https://www.figma.com/file/yGY2tcdEwvS3Xnox4eP4BO/Mockups-Discussion-2%2F21?type=whiteboard&node-id=0-1&t=R2YN8xazIKhcsPKr-0)

bencodeorg (approved):
Very nice! This does have a lot of overlayp with the UI I added on the lab side (especially the model card UI), so I do wonder if we should be sharing code between here and the lab? Definitely hear Dan's concerns about vertical space as well, I think we can sort that out though as we go.

sanchitmalhotra126:
> In general, vertical space starts to become an issue in levelbuilder very very quickly, and will likely be at the heart of different types of cosmetic feedback I have. Looking at this now, I'm worried about how much space Retrieval and Model Card can take up on the page. Would it be possible to make those headers collapsible? Or, if I'm understanding the layers of React components correctly, would it be possible to make all of these sub-headings collapsible (Bot Name, Temperature, System Prompt, etc)?

@dancodedotorg yep, we can definitely make these sub-headings collapsible. I'll follow-up with that in a separate change!

dashboard/app/controllers/levels_controller.rb
sanchitmalhotra126: Hmm yeah we probably should, let me see if that's straightforward or if not I can file a task

apps/src/lab2/levelEditors/aiCustomizations/ModelCardFields.tsx
sanchitmalhotra126: Yeah definitely! I was also thinking as I was doing this that there would probably be a lot of overlap with the UI...maybe we can revisit once the UI is in place? Though if we want to start using shared constants/fields now, we can do that as well.

sanchitmalhotra126:
> This does have a lot of overlayp with the UI I added on the lab side (especially the model card UI), so I do wonder if we should be sharing code between here and the lab?

@bencodeorg yeah, I was thinking the same 🙂 let's check back once we get this and the student-facing UI in and see what we can reuse?

sanchitmalhotra126:
> Not my lane, potential nit: do you mean for the temperature value to be a string? image

Ah good question - yeah this is happening because of how the JSON is being stored and parsed. We'll need to just make sure to parse the numeric field when we handle this in the student-facing UI.

apps/src/lab2/levelEditors/aiCustomizations/ModelCardFields.tsx
bencodeorg: Sounds good to give things a little more time to firm up, then see what we can share :)

apps/src/lab2/levelEditors/aiCustomizations/UpdateContext.ts
sanchitmalhotra126: Oh sure, this is similar to setting up a redux store/actions (in that there is a single state value, aiCustomizations, and some action functions to update it) - but without actually doing all the redux boilerplate because this is a relatively small surface area. The main benefit here is just that we don't need to thread the `aiCustomizations` value and `setXYZ` functions through multiple layers of props.

apps/src/lab2/levelEditors/aiCustomizations/EditAiCustomizations.tsx
sanchitmalhotra126: Tweaked this a bit to try to take more advantage of horizontal space and reduce vertical space. Thoughts? Also worth noting that all of these text boxes are resizable. I'll also follow up with a separate change to make these sections collapsible.
Screenshot 2024-03-08 at 11 47 31 AM (2)

sanchitmalhotra126:
@dancodedotorg made a couple edits if you'd like to take a look (screenshots should be updated in the description)
- made text areas wider and shorter to optimize vertical space
- "Can Switch Panels?" -> "Hide Presentation Panel"
- "Ask About Topics" -> "Example Prompts and Topics"
- Will add collapsibility as a separate follow-up

Let me know if there's any feedback I missed!

Edit: also as @bencodeorg alluded to, there's a lot of similar code here with the student-facing UI, so we may end up consolidating in the future in which case this edit page could look more similar to the student-facing page. But for now we just want to make sure that we have a working and easy to use level editing experience 🙂

dancodedotorg:
Oh awesome - really like the wider-but-shorter text boxes!! Thanks for incorporating that feedback!

Everything else looks good to me - thanks for iterating!

PR #57066, opened 2024-03-06
add ai_confidence_exact_match field to learning_goal_ai_evaluations table

davidsbailey:
starts https://codedotorg.atlassian.net/browse/AITT-489. This is a small first PR to follow the best practice of putting the migration in its own PR.

For now there will be 2 confidence-related fields on the LearningGoalAiEvalution table: 
* `ai_confidence` -- our confidence level in the AI score, interpreted as a pass-fail score
* `ai_confidence_exact_match` -- our confidence level in the exact AI score

`ai_confidence_exact_match` will contain the same format data as `ai_confidence`, i.e. null or an integer representing low, medium or high confidence.

## Testing story

existing test coverage

PR #57078, opened 2024-03-06
write exact-match ai scores to the database in eval rubric job

davidsbailey:
Depends on https://github.com/code-dot-org/code-dot-org/pull/57066. Continues https://codedotorg.atlassian.net/browse/AITT-489. This PR does the folllowing when the eval rubric job is run:
* read confidence-exact.json
* use learning goal and ai label to figure out which confidence value to pull out of the json
* write the value to the database

Also done behind the scenes to prop up this PR:
* [spreadsheet](https://docs.google.com/spreadsheets/d/1q852ucSF_BsKkKjozxTumLqq0_XORe0IliU6De3YDIU/edit#gid=0) summarizing which [lesson, learning goal, ai score] tuples are high confidence, following the criteria in the [Accuracy Thresholds doc](https://docs.google.com/document/d/1OCtVDskN0z4-_vYBgt3Bv7P9xB6IlIn5b3mEEeCR0nY/edit)
* created a new release `2024-03-06-ai-rubrics-json-reason-confidence-exact/` and uploaded it to S3

## Testing story

manual testing:
* temporarily pointed local dashboard to `2024-03-06-ai-rubrics-json-reason-confidence-exact/`, and manually verified that the correct values are getting written to the DB
* also used a work-in-progress UI PR to confirm that the ai_confidence_exact_match values are getting passed through to the UI and that the correct values are being set

unit test updates:
* updated s3 stub to provide fake confidence-exact.json
* updated all existing unit tests to verify that valid `ai_confidence_exact_match` values are written to the DB
* added a unit test covering the case where confidence-exact.json is not present, in which case `nil` values should be written to the db

## Future work

* launch confidence-exact.json files as part of a release. currently I am planning to wait for @wilkie's addition of evidence prompting to other lessons before doing this


dashboard/app/jobs/evaluate_rubric_job.rb
```diff
@@ -390,11 +399,21 @@ def validate_ai_config_for_lesson(lesson_s3_name, code)
...
+
+      if ai_confidence_levels_exact_match
+        label = evaluation['Label']
+        confidence_exact_match = ai_confidence_levels_exact_match[learning_goal][label]
+        unless confidence_exact_match
+          raise "No confidence_exact_match for learning goal: #{learning_goal}, label: #{label} evaluation: #{JSON.pretty_generate(evaluation)} ai_confidence_levels_exact_match: #{JSON.pretty_generate(ai_confidence_levels_exact_match)}"
+        end
+        evaluation['Confidence-Exact-Match'] = confidence_s_to_i(confidence_exact_match)
+      end
+      evaluation
```
cearachew, line 416: For my own education- what's the purpose of this evaluation statement?

dashboard/app/jobs/evaluate_rubric_job.rb
```diff
@@ -422,7 +441,8 @@ def validate_ai_config_for_lesson(lesson_s3_name, code)
           learning_goal_id: lg.id,
           rubric_ai_evaluation_id: rubric_ai_evaluation.id,
           understanding: understanding_s_to_i(label),
-          ai_confidence: ai_evaluation['Confidence'],
+          ai_confidence: ai_evaluation['Confidence-Pass-Fail'],
+          ai_confidence_exact_match: ai_evaluation['Confidence-Exact-Match'],
```
cearachew: nit: is this key casing/format consistent with other keys? I see 'Key Concept' above

cearachew:
Just to clarify- this is not pointing to a collection of lessons in s3 that include exact match confidence yet, right?

dashboard/app/jobs/evaluate_rubric_job.rb
davidsbailey: this is an implicit return value. https://franzejr.github.io/best-ruby/idiomatic_ruby/implicit_return.html in this particular case, because we are doing a map operation, the original object needs to be returned in order for the result to reflect the additions to the hashmap. then the entire ~~hashmap~~ array of hashmaps is implicitly returned as the return value for the enclosing method.

dashboard/app/jobs/evaluate_rubric_job.rb
davidsbailey: good point! fixed

davidsbailey (commented):
> Just to clarify- this is not pointing to a collection of lessons in s3 that include exact match confidence yet, right?

Correct!

cearachew (approved):
Looks great to me! Thanks for the explanation on implicit return!

PR #57079, opened 2024-03-06
LTI metrics and Statsig wrapper init

nicklathe:
This PR adds a wrapper to the Statsig Ruby SDK, and adds an example log event using the wrapper.

There will be a follow up PR adding more event logging throughout the LTI code using this wrapper.

(UPDATE 3.20.24):

Statsig has an interesting event log format. It's `sendEvent` function takes an `event name`, `event value`, and an optional `metadata` object that can contain arbitrary key/values. Their examples for usage are something like `event_name: item_added_to_cart`, `event_value: SKU_109483`, `metadata: {price: 1.99, item_name: 'diet_coke'}`.

From what I can foresee in our logging cases, we care about `event_name` as this is what we will index on when building dashboards, and the `metadata` field which we will put things like `lms_type`, `user_type`, etc. We may not always care about adding an `event_value` field to our log.

As a result, I refactored the sendEvent wrapper function to set `event_value` to nil as a default. If it is nil, it will set this value to be the same as `event_name`, otherwise it will pass the supplied `event_value` down to the Statsig sendEvent.

Example:

**Send event WITH NO `event_value`**
```
metadata = {
  'user_type' => user.user_type,
  'lms_type' => integration[:platform_name],
}
Metrics::Events.log_event(
  user: user,
  event_name: 'lti_user_signin',
  metadata: metadata,
)
```
Screenshot 2024-03-20 at 3 00 16 PM

**Send event WITH `event_value`**
```
metadata = {
  'user_type' => user.user_type,
  'lms_type' => integration[:platform_name],
}
Metrics::Events.log_event(
  user: user,
  event_name: 'lti_user_signin',
  event_value: 'some_value'
  metadata: metadata,
)
```
Screenshot 2024-03-20 at 3 00 16 PM copy

## Links

- [Previous PR configuring Statsig](https://github.com/code-dot-org/code-dot-org/pull/56879)
- [Jira ticket](https://codedotorg.atlassian.net/browse/P20-736)

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/dashboard/app/controllers/lti_v1_controller.rb
+++ b/dashboard/app/controllers/lti_v1_controller.rb
@@ -163,13 +163,12 @@ class LtiV1Controller < ApplicationController
         sign_in user
 
         metadata = {
-          'LMS_Type' => integration.platform_name,
           'User_Type' => user.user_type,
+          'LMS_Type' => integration[:platform_name],
         }
         Metrics::Events.log_event(
           user: user,
-          event_type: 'LTI_EVENT',
-          event_name: 'LTI_EXISTING_USER_LOGIN',
+          event_name: 'LTI_User_Signin',
           metadata: metadata,
         )
         # If on code.org, the user is a student and the LTI has the same user as a teacher, upgrade the student to a teacher.
--- a/dashboard/app/controllers/lti_v1_controller.rb
+++ b/dashboard/app/controllers/lti_v1_controller.rb
@@ -5,6 +5,7 @@ require "policies/lti"
 require "concerns/partial_registration"
 require "clients/lti_advantage_client"
 require "cdo/honeybadger"
+require 'metrics/events'
 
 class LtiV1Controller < ApplicationController
   before_action -> {redirect_to lti_v1_integrations_path, alert: I18n.t('lti.integration.early_access.closed')},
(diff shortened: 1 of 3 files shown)


dashboard/app/controllers/lti_v1_controller.rb
```diff
@@ -161,6 +162,16 @@ def authenticate
       if user
         sign_in user
 
+        metadata = {
+          'LMS_Type' => integration.platform_name,
+          'User_Type' => user.user_type,
```
carl-codeorg: Is this the preferred casing/naming convention for Statsig? Do we get any capability to search/index off these metadata fields or just top-level fields?

dashboard/lib/metrics/events.rb
```diff
@@ -0,0 +1,43 @@
...
+      def log_event(user:, event_type:, event_name:, metadata: {})
+        if CDO.rack_env?(:development)
+          log_event_to_stdout(user: user, event_type: event_type, event_name: event_name, metadata: metadata)
+        else
+          log_event_with_statsig(user: user, event_type: event_type, event_name: event_name, metadata: metadata)
+        end
+      rescue => exception
+        Honeybadger.notify(
+          exception,
+          error_message: 'Error logging event to Statsig',
```
carl-codeorg: Would it be better to make this message generic instead of Statsig-specific?

dashboard/app/controllers/lti_v1_controller.rb
nicklathe: Good question. I'm basing this off of [this doc](https://docs.google.com/document/d/19Yxl73BAA7lsA_s4s27rD2ITuvyb4U39iVSZETlDLOs/edit#heading=h.gdpxre740sd2), but will confirm with Ali this is the casing he wants. I don't think we can index off of the metadata fields, but I should confirm that is true.

nicklathe: I changed these (and all string keys) to lower snake_case.

nicklathe: To answer your question regarding indexing off the metadata fields, from what I can see this isn't supported. Mostly we'll be indexing off of the event name. You can add additional "filters" (they call them "dimensions") on the event value, but I haven't seen anything on being able to index on metadata.

carl-codeorg (approved):
:shipit:

PR #57082, opened 2024-03-06
JS integration for Statsig

hannahbergam:
Utilizing [Statsig documentation here](https://docs.statsig.com/client/jsClientSDK) and following along with how Amplitude was set up via the AnalyticsReporter.

I created a copy because we are going to continue to use the platforms side-by-side. This way, if and when we decide to delete Amplitude, we won't have to modify this file. This means I also ended up duplicating a couple helper functions, namely formatUserId and shouldPutRecord. 

I'm opening the PR to get feedback from various teams. The next step will be to do additional testing by converting some acq events to send to 'BOTH' to ensure it's working as expected.

Here is Nick's PR implementing Statsig for Ruby: https://github.com/code-dot-org/code-dot-org/pull/56879/files

Was able to verify that at least the first half of this piping works by adding a 3rd param to a TeacherDashboard Analytics event to send to 'BOTH' platforms, and can see this locally:
Screenshot 2024-03-15 at 11 53 08 AM

## PR Checklist:

- [ ] Code is well-commented


apps/src/lib/util/StatsigReporter.js
```diff
@@ -0,0 +1,100 @@
...
+  isDevelopmentEnvironment,
+} from '../../utils';
+
+// A flag that can be toggled to send events regardless of environment
+const ALWAYS_SEND = false;
+const Statsig = require('statsig-js');
+
+class StatsigReporter {
+  constructor() {
+    const local_mode = isProductionEnvironment ? false : true;
```
sanchitmalhotra126: Did you mean for this to be `isProductionEnvironment()`?

Edit: also nit, but could simplify to just `local_mode = !isProductionEnvironment()`? (and even just do this directly in the `options` object below?)

sanchitmalhotra126 (commented):
Don't have a ton of context on the Statsig API, but generally LGTM! Just one comment.

Also a couple totally optional ideas:
- Would it be worth creating some composite reporter class that stores an instance of StatsigReporter and AnalyticsReporter, and can internally decide to use one or the other (or both)? That way some of these common functions can be shared and the APIs can be closer together?
- Also not at all required, but if you were interested in porting some code to TypeScript, this would be a great use case, since you could use an [interface](https://www.typescriptlang.org/docs/handbook/interfaces.html) to ensure that the two classes shared the same functions.

apps/src/lib/util/StatsigReporter.js
nicklathe: When I was setting up the Ruby SDK, the Infrastructure team emphasized the importance of maintaining consistency between our managed server test environment and production by ensuring Statsig configurations are identical in both environments, I.e. we are setting `local_mode` to false in both those environments. [Here](https://github.com/code-dot-org/code-dot-org/pull/56879/files#diff-f78118b42ccf9fa4ed82455846845c0513a11f72c375070da73aa25531983475R6) is how I did that on the Rails side, to ensure I was targeting the managed server test environment, not other test environments (drone, etc). I'm not sure if there is similar methods we can use in JS?

hannahbergam: Nick- I saw that code and wasn't quite sure what the front end equivalent would be. I tried to write something that would seem to fit both your Ruby integration, and what we have already in AnalyticsReporter (Amplitude). That said- I'll poke around more. And good catch Sanchit- fixed!

hannahbergam:
> Don't have a ton of context on the Statsig API, but generally LGTM! Just one comment.
> 
> Also a couple totally optional ideas:
> 
> * Would it be worth creating some composite reporter class that stores an instance of StatsigReporter and AnalyticsReporter, and can internally decide to use one or the other (or both)? That way some of these common functions can be shared and the APIs can be closer together?
> * Also not at all required, but if you were interested in porting some code to TypeScript, this would be a great use case, since you could use an [interface](https://www.typescriptlang.org/docs/handbook/interfaces.html) to ensure that the two classes shared the same functions.

I spent some time thinking about this and decided to move forward with an abstracted function that can take in an optional param (default: send to Amplitude, but can be overridden to send to Statsig or both). From what I can gather, we don't want to live in a world where we have a duplicate front end analytics platform. It's unfortunate that we have to have both living side by side when we abstracted Amplitude for this exact case (to be swapped out). That said- I'll start writing the version I detailed and see if it makes sense to you.

apps/src/lib/util/StatsigReporter.js
hannahbergam: @nicklathe in my latest commit, I loaded the same logic into the application.html.haml file so I could query it from the DOM to mimic what you did for the ruby sdk. Let me know if it makes sense/looks right!

nicklathe (approved):
LGTM!

sanchitmalhotra126 (approved):

apps/src/lib/util/AnalyticsReporter.js
```diff
@@ -12,13 +12,17 @@ import {
...
 
 import {EVENT_GROUPS} from './AnalyticsConstants';
 
 import DCDO from '@cdo/apps/dcdo';
 
 // A flag that can be toggled to send events regardless of environment
 const ALWAYS_SEND = false;
+const AMPLITUDE = 'Amplitude';
+const STATSIG = 'Statsig';
+const BOTH = 'Both';
```
bethanyaconnor: Could be done in a follow-up but should these be accessible as constants from other places? So that the third param could be set to `ANALYTICS_REPORTER_CONSTANTS::AMPLITUDE` (or something)

hannahbergam: Yuuup good call. I moved them into the AnalyticsConstants file and imported them instead!

bethanyaconnor (approved):

PR #57083, opened 2024-03-06
Put 'needs feedback' indicators on progress lesson and level cells

etaderhold:
This change adds a triangle to progress cells to indicate a lesson or level needs feedback.

A level needs feedback when it's an assessement level that can receive feedback that the student has worked on and has not received feedback since their last changes. 
A lesson needs feedback when any of the levels do.

image

## Links

[JIRA](https://codedotorg.atlassian.net/browse/TEACH-878)

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/templates/progress/progressHelpers.js
+++ b/apps/src/templates/progress/progressHelpers.js
@@ -164,6 +164,7 @@ export const studentNeedsFeedback = (progress, level) =>
   progress &&
   progress.status !== LevelStatus.not_tried &&
   level.kind === 'assessment' &&
+  level.canHaveFeedback &&
   !feedbackLeft(progress);
 
 /**
--- a/apps/src/templates/progress/progressHelpers.js
+++ b/apps/src/templates/progress/progressHelpers.js
@@ -156,6 +156,16 @@ export function lessonHasLevels(lesson) {
   return !!lesson.levels?.length;
 }
 
+export const feedbackLeft = progress =>
+  progress?.teacherFeedbackReviewState !== 'keepWorking' &&
+  progress?.teacherFeedbackNew;
+
+export const studentNeedsFeedback = (progress, level) =>
+  progress &&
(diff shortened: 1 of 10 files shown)


apps/src/templates/progress/progressHelpers.js
```diff
@@ -156,6 +156,17 @@ export function lessonHasLevels(lesson) {
   return !!lesson.levels?.length;
 }
 
+export const feedbackLeft = progress =>
+  progress?.teacherFeedbackReviewState !== 'keepWorking' &&
+  progress?.teacherFeedbackNew;
+
+export const studentNeedsFeedback = (progress, level) =>
+  progress &&
+  progress.status !== LevelStatus.not_tried &&
```
lfryemason, line 165: Does a student need feedback if the level is in progress but not submitted?

apps/test/unit/code-studio/progressReduxTest.js
```diff
@@ -732,6 +737,7 @@ describe('progressReduxTest', () => {
             app: 'maze',
             usesLab2: false,
             isValidated: false,
+            canHaveFeedback: undefined,
```
lfryemason, line 740: nit: Will this reasonable ever be undefined or should we just make the tests false?

apps/src/templates/progress/progressHelpers.js
etaderhold: I thought that's what we landed on in the last office hours. I'm happy to revisit that during the bug bash or whatever next opportunity we have.

apps/test/unit/code-studio/progressReduxTest.js
etaderhold: The response from the server will never be undefined, but also the exact value of the parameter is irrelevant to the test and several others were marked as undefined instead of trying to craft a high-fidelity object.

lfryemason (approved):

PR #9455, opened 2016-07-12
Fix storybook

pcardune:
Something went haywire when we merged code-studio. I found the solution through binary search. Not sure why it works!

PR #57092, opened 2024-03-07
Removed unused progress-bar package

jmkulwik:
Cleaning up some old packages using depcheck.

Looks like we used a progress bar plugin for a while, but eventually stopped using it. Interestingly, we didn't even know why we were using it back then https://github.com/code-dot-org/code-dot-org/pull/9455 :) . Going ahead and removing it.


breville (approved):
Slightly cautious about this one.  If we had to import it to make something work, could it be that something breaks again when we no longer import it?

PR #57095, opened 2024-03-07
Gen AI Lab: initial UI for model customization panels

bencodeorg:
Adds initial UI for the "Prompt", "Retrieval" and "Publish" tabs of the new "Model Customization" panel in our new generative AI lab. Mostly HTML/CSS here, with a bit of state just to get the juices flowing (eg, I played with setting state with the "temperature" slider).

| Prompt | Retrieval | Publish |
| - | - | - |
image | image | image |

## Links

- [Figma mockups](https://www.figma.com/file/yGY2tcdEwvS3Xnox4eP4BO/Mockups-Discussion-2%2F21?type=whiteboard&node-id=2-8192&t=fMVPXGQjzZ8EJujW-0)

## Testing story

Tested manually -- I didn't add any tests for now, as it feels like we're still in a early phase and things may change.

## Follow-up work

Doesn't hook it up to a Redux store or consume configuration from levelbuilder, but that's what's up next!

It's not pixel perfect (eg, the buttons are pretty ugly -- I was kinda hoping we'd be able to plug in Denys's new `Button` component from the component library when it's ready, but we can totally restyle if necessary in the meantime). I also looked at restyling the slider input, but whoa it is tricky! So I figured we'd wait on that as well, [more info here](https://developer.mozilla.org/en-US/docs/Web/CSS/::-webkit-slider-runnable-track).

We also need to do something with the translation drop-up (see music lab for an idea), which overlaps the button at the bottom of the page.


apps/src/componentLibrary/fontAwesomeV6Icon/FontAwesomeV6Icon.tsx
```diff
@@ -38,7 +38,7 @@ export interface FontAwesomeV6IconProps {
  * Can can be used in any component in/out of the scope of Design System.
  */
 const FontAwesomeV6Icon: React.FunctionComponent<FontAwesomeV6IconProps> = ({
-  iconStyle,
+  iconStyle = 'solid',
```
bencodeorg, line 41: @levadadenys are you cool with me adding a default here? Is `solid` the right choice, or something else?

apps/src/aichat/views/model-customization-workspace.module.scss
```diff
@@ -0,0 +1,48 @@
...
+  width: 100%;
+  box-sizing: border-box;
+}
+
+.modelCustomizationWorkspace {
+  height: 100%;
+  display: flex;
+  flex-direction: column;
+
+  textarea {
```
sanchitmalhotra126, line 14: nit: personal preference, but should we use class names here instead of selecting by element type?

apps/src/aichat/views/modelCustomization/PromptCustomization.tsx
```diff
@@ -0,0 +1,52 @@
...
+const PromptCustomization: React.FunctionComponent = () => {
+  const [temperature, setTemperature] = useState('0.05');
+
+  const handleTemperatureChange = (event: ChangeEvent<HTMLInputElement>) => {
+    setTemperature(event.target.value);
+  };
+
+  return (
+    <div className={styles.verticalFlexContainer}>
+      <div>
```
sanchitmalhotra126, line 15: nit: extra div?

apps/src/aichat/views/modelCustomization/PromptCustomization.tsx
```diff
@@ -0,0 +1,52 @@
...
+        <div className={styles.inputContainer}>
+          <label htmlFor="chatbot-name">
+            <StrongText>Chatbot name</StrongText>
```
sanchitmalhotra126, line 18: Should we start dropping these in `aichat/en_us.json` since we already have that localized file set up? Also ok if we want to hold off until designs are more finalized.

apps/src/aichat/views/modelCustomization/RetrievalCustomization.tsx
```diff
@@ -0,0 +1,67 @@
...
+import FontAwesomeV6Icon from '@cdo/apps/componentLibrary/fontAwesomeV6Icon/FontAwesomeV6Icon';
+import {StrongText} from '@cdo/apps/componentLibrary/typography/TypographyElements';
+import modelCustomizationStyles from '../model-customization-workspace.module.scss';
+import styles from './retrieval-customization.module.scss';
+
+const RetrievalCustomization: React.FunctionComponent = () => {
+  const [messages, setMessages] = useState<string[]>([]);
+  const [newMessage, setNewMessage] = useState('');
+
+  const onAdd = () => {
```
sanchitmalhotra126: nit: wrap these in `useCallback`?

apps/src/aichat/views/modelCustomization/PublishNotes.tsx
```diff
@@ -0,0 +1,34 @@
...
+  {id: 'intended-uses', text: 'Intended Uses'},
+  {id: 'limitations-and-warnings', text: 'Limitations and Warnings'},
+  {id: 'testing-and-evaluation', text: 'Testing and Evaluation'},
+];
+
+const PublishNotes: React.FunctionComponent = () => {
+  return (
+    <div className={styles.verticalFlexContainer}>
+      <div>
+        {INPUTS.map(input => {
```
sanchitmalhotra126: nit: can destructure `input` to `{id, text}` if that feels cleaner

sanchitmalhotra126 (approved):
A few minor nits but looks good! Solid starting point to start hooking this up to some data/redux. I'm sure we'll continue to refine UI specifics as we go along

apps/src/aichat/views/model-customization-workspace.module.scss
bencodeorg: haha I thought I was being slick by not having to use class names :) Do you just think it's easier if you see the class name inline when you're looking at some component, vs. in this case you're seeing some styling applied without having an explicit class name applied?

apps/src/aichat/views/modelCustomization/PromptCustomization.tsx
bencodeorg: Putting them there starts creating work for translators, right? I'd vote to hold off until we have things more solidified

apps/src/aichat/views/modelCustomization/PromptCustomization.tsx
bencodeorg: I think we need this container for flex reasons (ie, to keep all inputs at top of panel and submit button at bottom)

apps/src/aichat/views/model-customization-workspace.module.scss
sanchitmalhotra126: Haha I usually just opt for classnames to have more control over what elements should receive which styles - just in case we add another `textarea` here but want to apply different styles, for example. To your question I personally like to see the classname inline since it's also easier for tracking down where a particular set of styles originate from in the CSS. That being said, I could go either way with this specific example because we'd probably want all `textarea`s to look the same in this component anyway.

bencodeorg: Ok cool -- I'm going to merge so I can untangle merge conflicts with this branch / your levelbuilder branch in a follow-up I have going to load in levelbuilder config. Happy to update this down the line though!

PR #57034, opened 2024-03-05
[Google Blockly] Bitmap field and new Sprite Lab blocks

mikeharv:
This branch adds a new Blockly plugin [@blockly/field-bitmap](https://www.npmjs.com/package/@blockly/field-bitmap) which allows for user-inputted pixel grids. This new field class is extended and customized in our usual fashion and hooked up as a custom input type for Sprite Lab blocks.  Three new blocks are created which allow for users to create a bitmap representation of a layout for sprites. These blocks are intended for use in the new game development module and build off of the work in https://github.com/code-dot-org/code-dot-org/pull/57022.

image
image

The two following animations were recorded before the "environment" label was converted to a dropdown field (seen above). However, there's no other difference in appearance. Standard block with built-in field:

https://github.com/code-dot-org/code-dot-org/assets/43474485/57b682a0-702e-4a5e-ab09-3a24f9314485

Block with separate Array input and shadow block:

Note that default size grid is 8x8. This was chosen as it aligns well to Sprite Lab's 400x400 canvas in that sprites are automatically scaled to 50%. Other grid sizes can be specified in the block XML (for start blocks or toolboxes), but students cannot currently adjust the size of the grid themselves. New sprites are scaled and positioned automatically regardless of grid size.

## Extended field class
As stated in the plugin's documentation, it does support serialization to XML. Fortunately, we can accomplish this easily by using JSON methods to stringify and parse the field value. The `toXml` and `fromXml` hooks work with the old XML serialization system and have been added to our class.

## Scale and position calculations
The scale is determined based on the app height (400) and the number of rows in the array. For example, with a 4x4 grid, the scale is 100 and the sprites are rendered with a size of 100x100. Sprite x/y coordinates represent the center of the sprite. So, to position the sprites we offset x and y by _half_ of the scale.
imageimage

As like the previous work in this space (https://github.com/code-dot-org/code-dot-org/pull/57022), the exact block configuration and helper code will live on levelbuilder and so we should expect some iteration outside of the normal PR flow between now and piloting.

## Follow-up work

Further customize the field. This work is also up for review in a branch based off of this one:
- https://github.com/code-dot-org/code-dot-org/pull/57107

## PR Checklist:

- [ ] Tests provide adequate coverage
- [ ] Privacy and Security impacts have been assessed
- [ ] Code is well-commented
- [ ] New features are translatable or updates will not break translations
- [ ] Relevant documentation has been added or updated
- [ ] User impact is well-understood and desirable
[...]

PR #57107, opened 2024-03-07
[Google Blockly] Bitmap field overrides

mikeharv:
This branch includes customizations for the new bitmap field added here:
- https://github.com/code-dot-org/code-dot-org/pull/57034

Specifically, the changes here accomplish the following:
- Use our translations for the button text in the field's editor
- Reconfigure the rendered size of the field and its "pixels" to prevent blocks from being too tall

While many Blockly field's support passing configuration options directly into the constructor, this plugin was created by a contributor who did not supply that path. This means that I instead had to copy and override some of the methods from the original `FieldBitmap` class.

## Translations
The button text in the plugin is set with hard-coded English string: https://github.com/google/blockly-samples/blob/498c36e7f7596b1577a46c58db3ca1ed0002cc6a/plugins/field-bitmap/src/field-bitmap.js#L286-L288
This happens whenever a user clicks the field (`showEditor_()` calls `dropdownCreate_`). Because `dropdownCreate_` is private, we can't override it directly, but we can override `showEditor_` in order to find the buttons after they are created and update them. 

image

## Sizing
After getting the field to work originally (https://github.com/code-dot-org/code-dot-org/pull/57034), I was disappointed to see how big it made the blocks when the grid was taller than 4 "pixels". Since the actual pixel toggling happens in the editor, there's no need to stretch the field itself just to accommodate a larger array. Katie and @amy-b both confirmed that it would be preferable for the field size to be static like shown:
image (182)

This means that we need the size of the pixels to be dynamic instead based on the available space. I went with a field height of 42 to align to the costume picker. This is a screenshot of a fake block that I include here just for illustration purposes: 
image

The original sizing methods can be found here:
https://github.com/google/blockly-samples/blob/498c36e7f7596b1577a46c58db3ca1ed0002cc6a/plugins/field-bitmap/src/field-bitmap.js#L312-L356

## PR Checklist:

- [ ] Code is well-commented


molly-moen (approved):
Nice!

PR #57120, opened 2024-03-07
[Progress V2] Add a scrollbar to progress table that sticks to the bottom of screen

lfryemason:
Adds a scrollbar that sticks to the bottom of the screen if the bottom of the table is not visible.

https://github.com/code-dot-org/code-dot-org/assets/25193259/5ade1dfb-54d7-4587-9c27-349c3b99a32e

## Links
[TEACH-887](https://codedotorg.atlassian.net/browse/TEACH-887)

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/templates/sectionProgressV2/floatingScrollbar/FloatingScrollbar.jsx
+++ b/apps/src/templates/sectionProgressV2/floatingScrollbar/FloatingScrollbar.jsx
@@ -77,8 +77,10 @@ export default function FloatingScrollbar({
     const maxVisibleY =
       window.innerHeight || document.documentElement.clientHeight;
 
+    // 20 is the height of the scrollbar.
     const isNowVisible =
-      childContainerRef?.current.getBoundingClientRect().bottom < maxVisibleY;
+      childContainerRef?.current.getBoundingClientRect().bottom + 20 <
+      maxVisibleY;
 
     if (isNowVisible !== scrollVisible) {
       setScrollVisible(isNowVisible);
@@ -121,7 +123,10 @@ export default function FloatingScrollbar({
         ref={scrollRef}
         style={{width: childWidth + 'px'}}
       >
-        <div style={{width: childScrollWidth + 'px'}} />
+        <div
+          style={{width: childScrollWidth + 'px'}}
+          className={styles.emptyScrollContents}
+        />
       </div>
     </div>
   );
--- a/apps/src/templates/sectionProgressV2/floatingScrollbar/FloatingScrollbar.jsx
+++ b/apps/src/templates/sectionProgressV2/floatingScrollbar/FloatingScrollbar.jsx
@@ -31,7 +31,7 @@ export default function FloatingScrollbar({
   const [childWidth, setChildWidth] = React.useState(0);
   const [scrollVisible, setScrollVisible] = React.useState(true);
 
-  if (children.length > 1) {
+  if (children.length !== 1) {
     throw new Error('FloatingScrollbar only supports a single child');
   }
 
--- a/apps/src/templates/sectionProgressV2/floatingScrollbar/FloatingScrollbar.jsx
+++ b/apps/src/templates/sectionProgressV2/floatingScrollbar/FloatingScrollbar.jsx
@@ -31,12 +31,6 @@ export default function FloatingScrollbar({
   const [childWidth, setChildWidth] = React.useState(0);
   const [scrollVisible, setScrollVisible] = React.useState(true);
 
-  if (children.length > 1) {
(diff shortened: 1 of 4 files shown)


apps/src/templates/sectionProgressV2/progress-table-v2.module.scss
```diff
@@ -270,7 +270,7 @@ $level-cell-width: 52;
 .studentColumn {
   width: 195px;
   min-width: 195px;
-  margin-right: 12px;
+  margin-inline-end: 12px;
```
kelbyhawn, line 277: 💖

apps/src/templates/sectionProgressV2/floatingScrollbar/FloatingScrollbar.jsx
```diff
@@ -0,0 +1,142 @@
...
+  childContainerRef,
+  childContentsRef,
+}) {
+  const scrollRef = React.useRef();
+
+  const [childScrollWidth, setChildScrollWidth] = React.useState(0);
+  const [childWidth, setChildWidth] = React.useState(0);
+  const [scrollVisible, setScrollVisible] = React.useState(true);
+
+  if (children.length > 1) {
```
etaderhold: Just for the sake of completeness?
```suggestion
  if (children.length !== 1) {
```

apps/src/templates/sectionProgressV2/floatingScrollbar/FloatingScrollbar.jsx
```diff
@@ -0,0 +1,142 @@
...
+      if (childContainerRef?.current) {
+        childContainerRef.current.scrollLeft = scroll.target.scrollLeft;
+      }
+    },
+    [childContainerRef]
+  );
+
+  React.useEffect(() => {
+    setOnScroll(() => scroll => {
+      scrollRef.current.scrollLeft = scroll.target.scrollLeft;
```
etaderhold, line 108: Have you tried this in RTL mode?

apps/src/templates/sectionProgressV2/floatingScrollbar/FloatingScrollbar.jsx
```diff
@@ -0,0 +1,142 @@
...
+        }
+      }),
+    [setChildWidth]
+  );
+
+  const childContentsResizeObserver = React.useMemo(
+    () =>
+      new ResizeObserver(([entry]) => {
+        if (entry.borderBoxSize) {
+          setChildScrollWidth(entry.borderBoxSize[0].inlineSize + 1);
```
etaderhold, line 48: What's the effect of the `+ 1` here? Would a different number need to be used for different border widths?

etaderhold (commented):
This looks lovely!

Just to make sure...you've tested this in all supported browsers?

apps/src/templates/sectionProgressV2/floatingScrollbar/FloatingScrollbar.jsx
lfryemason: `scrollLeft` is the keyword for horizontal scroll. There is no `scrollRight`.

apps/src/templates/sectionProgressV2/floatingScrollbar/FloatingScrollbar.jsx
lfryemason: I'm not entirely sure, because it's surprisingly hard to test these off by ones, but I am pretty sure it's because of this -1 margin-right: https://github.com/code-dot-org/code-dot-org/blob/staging/apps/src/templates/sectionProgressV2/progress-table-v2.module.scss#L127

lfryemason:
> This looks lovely!
> 
> Just to make sure...you've tested this in all supported browsers?

Because we have bug-bash on Monday, I want to merge this in so that we can test it more thoroughly as a group

etaderhold (approved):

apps/src/templates/sectionProgressV2/floatingScrollbar/FloatingScrollbar.jsx
etaderhold: Yep. I was reading the [documentation about this](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollLeft) and it says

> If the element's [direction](https://developer.mozilla.org/en-US/docs/Web/CSS/direction) is rtl (right-to-left), then scrollLeft is 0 when the scrollbar is at its rightmost position (at the start of the scrolled content), and then increasingly negative as you scroll towards the end of the content.

Seems worth a test.

PR #57121, opened 2024-03-07
Count level progress for contained_levels

TurnerRiley:
We've been receiving several Zendesk tickets (e.g. [here](https://codeorg.zendesk.com/agent/tickets/488769) and [here](https://codeorg.zendesk.com/agent/tickets/488982)) about users who complete certain courses but are not directed to print their certificates nor given access to do so. @bethanyaconnor tracked down that the issue was that if a `level` in a `script_level` had `contained_levels` it would check the outer level for completion rather than the inner `contained_levels`. This would always result in the level being deemed incomplete because it would search for `UserLevels` with progress with the outer level's id (which returned `null`) rather than the `contained_levels` ids.

This PR updates the logic for checking completion so that if a `UserLevel` can't be found with a given level id, it will check if that level has `contained_levels` and if those level ids result in completed `UserLevel` entries.

### Previous progress page show "Continue" rather than "Print Certificate"
BROKEN

### Previous page when a user attempted to view their certificate
Certificate Issue

### Fixed progress page shows "Print Certificate"
WORKS

### Now user is shown the certificate for completing all levels
WORKS_CERT

## Links
Jira ticket: [here](https://codedotorg.atlassian.net/jira/software/c/projects/ACQ/boards/64?assignee=60d62161f65054006980bd71&selectedIssue=ACQ-1591)
Zendesk ticket examples: [here](https://codeorg.zendesk.com/agent/tickets/488769) and [here](https://codeorg.zendesk.com/agent/tickets/488982)

## Testing story

### DIFF
--- a/dashboard/app/models/user.rb
+++ b/dashboard/app/models/user.rb
@@ -1365,7 +1365,19 @@ class User < ApplicationRecord
     user_levels_by_level = user_levels_by_level(script)
 
     script.script_levels.none? do |script_level|
-      user_levels = script_level.level_ids.map {|id| user_levels_by_level[id]}
+      user_levels = []
+      script_level.levels.each do |level|
+        curr_user_level = user_levels_by_level[level.id]
+
+        # If level.id is not present in user_levels_by_level, check if level has contained_levels with present ids
+        if !curr_user_level && !level.contained_levels.empty?
+          level.contained_levels.each do |contained_level|
+            user_levels.push(user_levels_by_level[contained_level.id])
+          end
+        else
+          user_levels.push(curr_user_level)
+        end
+      end
       unpassed_progression_level?(script_level, user_levels)
     end
   end
--- a/dashboard/test/models/user_test.rb
+++ b/dashboard/test/models/user_test.rb
@@ -1194,6 +1194,72 @@ class UserTest < ActiveSupport::TestCase
     refute next_script_level.nil?
   end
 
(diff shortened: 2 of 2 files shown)


dashboard/app/models/user.rb
```diff
@@ -1365,7 +1365,19 @@ def completed_progression_levels?(script)
     user_levels_by_level = user_levels_by_level(script)
 
     script.script_levels.none? do |script_level|
-      user_levels = script_level.level_ids.map {|id| user_levels_by_level[id]}
+      user_levels = []
+      script_level.levels.each do |level|
+        curr_user_level = user_levels_by_level[level.id]
+
+        # If level.id is not present in user_levels_by_level, check if level has contained_levels with present ids
+        if !curr_user_level && !level.contained_levels.empty?
```
bethanyaconnor, line 1373: As I thought about this, I'm also wondering how this works with choice levels, such as https://studio.code.org/s/csc-timecapsule-2023/lessons/3/levels/2. In this case, I think we'd only want one level to be passing 🤔

bethanyaconnor: I tested locally and it looks like we create a user level for the BubbleChoice level itself, so I think we're good there! 🤞

bethanyaconnor (approved):
Thank you for adding tests here and for your quick work on this!

dashboard/app/models/user.rb
TurnerRiley: Oh great! Thank you for checking!

PR #57125, opened 2024-03-07
Show exact ai score when exact-match confidence is high

davidsbailey:
Depends on https://github.com/code-dot-org/code-dot-org/pull/57078. Finishes https://codedotorg.atlassian.net/browse/AITT-489.

Done in this PR: 
* UI updates: use exact-match confidence to determine whether to show exact-match evidence level. when we do:
  * highlight only one "suggested" evidence level bubble
  * update text describing evidence level to name only one evidence level
  * make sure we are showing exact-match confidence, not pass-fail confidence
* refactoring / cleanup: 
  * renamed ai_confidence field to aiConfidencePassFail in the json that is passed to the client. this avoids some ambiguity on the client since aiConfidence is then used to represent whichever confidence value is to be shown
  * variable renames and other minor changes
  * [remove rating text from css classnames](https://github.com/code-dot-org/code-dot-org/pull/57125/commits/36fe969e82ca7074e442e81d9358fef0bcb263b1)
  * converted one existing test to React Testing Library

### before
before--aitt-basic-updates-to-ui-rubric-component

### after 

after-exact-extensive

after-exact-no-ev

## Testing story

* New RTL unit tests covering UI updates above
* end-to-end manual testing

## Follow-up work

For testing, I used the `2024-03-06-ai-rubrics-json-reason-confidence-exact` release in S3, which is the same as the `ai-rubrics-json-reason` release but with confidence-exact.json files added to it. The follow-up work to enable this feature is to ship a release containing these `confidence-exact.json` files.


apps/src/templates/rubrics/AiAssessmentBox.jsx
```diff
@@ -25,16 +26,22 @@ export default function AiAssessmentBox({
...
     return i18n.aiStudentAssessment({
       studentName: studentName,
       understandingLevel: assessment,
     });
   };
 
+  const getStudentAssessmentString = () => {
+    if (aiEvalInfo.showExactMatch) {
+      return UNDERSTANDING_LEVEL_STRINGS[aiUnderstandingLevel];
+    }
```
davidsbailey, line 42: this code limits the text we display to a single evidence level

apps/src/templates/rubrics/EvidenceLevelsForTeachersV2.jsx
```diff
@@ -28,16 +29,25 @@ export default function EvidenceLevelsForTeachersV2({
...
       if (understanding !== INVALID_UNDERSTANDING || !isAiAssessed) {
         return [];
       }
-      if (aiEvalInfo.understanding > 1) {
-        return [2, 3];
-      } else if (aiEvalInfo.understanding >= 0) {
-        return [0, 1];
+      if (aiEvalInfo.showExactMatch) {
+        return [aiEvalInfo.understanding];
+      }
```
davidsbailey, line 41: This code limits us to highlighting one suggested evidence level bubble

apps/src/templates/rubrics/LearningGoals.jsx
```diff
@@ -221,6 +224,13 @@ export default function LearningGoals({
 
   const aiEvalInfo = getAiInfo(learningGoals[currentLearningGoal].id);
 
+  let aiConfidence;
+  if (aiEvalInfo) {
+    aiConfidence = aiEvalInfo.showExactMatch
+      ? aiEvalInfo.aiConfidenceExactMatch
+      : aiEvalInfo.aiConfidencePassFail;
+  }
```
davidsbailey, line 242: this change updates aiConfidence to depend on whether we are showing pass-fail or exact-match AI scores. all uses of aiConfidence in client code should now refer to this computed value, not the `ai_confidence` value on the server which is actually the pass-fail confidence level.

apps/test/unit/templates/rubrics/LearningGoalsTest.jsx
```diff
@@ -65,6 +69,136 @@ describe('LearningGoals', () => {
...
+          learningGoals={learningGoals}
+          aiEvaluations={aiEvaluations}
+          studentLevelInfo={studentLevelInfo}
+          teacherHasEnabledAi
+          canProvideFeedback
+        />
+      );
+      screen.getByText(
+        'Stella has achieved Extensive or Convincing Evidence for this learning goal.'
+      );
```
Nokondi, line 169: Do you need an expect here?

apps/test/unit/templates/rubrics/LearningGoalsTest.jsx
```diff
@@ -65,6 +69,136 @@ describe('LearningGoals', () => {
...
+      render(
+        <LearningGoals
+          learningGoals={learningGoals}
+          aiEvaluations={aiEvaluations}
+          studentLevelInfo={studentLevelInfo}
+          teacherHasEnabledAi
+          canProvideFeedback
+        />
+      );
+      screen.getByText('AI Confidence: medium');
```
Nokondi, line 181: Same here (see previous comment)

apps/test/unit/templates/rubrics/LearningGoalsTest.jsx
```diff
@@ -65,6 +69,136 @@ describe('LearningGoals', () => {
...
+      render(
+        <LearningGoals
+          learningGoals={learningGoals}
+          aiEvaluations={myAiEvaluations}
+          studentLevelInfo={studentLevelInfo}
+          teacherHasEnabledAi
+          canProvideFeedback
+        />
+      );
+      screen.getByText(
```
Nokondi, line 124: Do you need an expect here?

Nokondi (commented):
Looks great! Just had a question about using expect in some of your tests. The comments you added here are very helpful - might also be useful to include some of them in the code. (I say this, but I'm terrible at remembering to document my own code 🤣 )

apps/test/unit/templates/rubrics/LearningGoalsTest.jsx
davidsbailey: I believe the answer is no! admittedly I had to look this up to know the answer... but it looks like RTL "queries" starting with "get" will throw if the element is not found: https://testing-library.com/docs/queries/about/

davidsbailey: I was able to confirm that all 3 of the tests using "get" queries do fail if I ask them to look for a nonexistent string.

davidsbailey:
Thanks for the feedback, Mark! I've included some updated code comments mirroring the notes in my inline PR comments.

apps/test/unit/templates/rubrics/LearningGoalsTest.jsx
Nokondi: OK, good to know!

Nokondi (approved):
Great work!

PR #56488, opened 2024-02-09
Add Amplitude analytics for project sharing

fisher-alice:
In anticipation of [upcoming changes to the Project public gallery](https://docs.google.com/document/d/1vU68tzXG72Z80MvM6xYEDwF5hLDf7V097P-0dDfpouc/edit#heading=h.qi8xxyq72bct), the product team would like to add these project sharing analytics in Amplitude to better understand the usage of the share feature.

Conveniently, there was already a `recordShare` function within `ShareAllowedDialog` to report analytics to firehose so I added the Amplitude analytics within the same function. Product team request that app type is included in the event.

I grouped the project sharing events as 'project-sharing-events'.

AFTER update: screncast of Amplitude logging in dev env.

https://github.com/code-dot-org/code-dot-org/assets/107423305/a483595c-dc20-47af-9d14-d51739e13722

## Links
[jira](https://codedotorg.atlassian.net/browse/LABS-496)

## Testing story
I tested locally with the dev console open with different types of labs.

## Follow-up work
I learned that Artist labs have a 'print' option. Thus, the Twitter button ends up being on the second line. ~~@samantha-code - is this something we should fix?~~

UPDATE: checked with @samantha-code and she brought up that once we remove 'Publish' button this should no longer be an issue. Yay!

Screenshot 2024-02-09 at 4 09 43 PM

## PR Checklist:

- [ ] Tests provide adequate coverage
- [ ] Privacy and Security impacts have been assessed
- [ ] Code is well-commented
- [ ] New features are translatable or updates will not break translations
- [ ] Relevant documentation has been added or updated
- [ ] User impact is well-understood and desirable
- [ ] Pull Request is labeled appropriately
- [ ] Follow-up work items (including potential tech debt) are tracked and linked

PR #57127, opened 2024-03-07
Amplitude analytics for Finish button (project sharing) and export app

fisher-alice:
On the last level of a progression, when the user clicks on the 'Finish' button, a Congrats dialog appears with sharing options. This PR adds Amplitude events to these sharing options. This is a follow up to https://github.com/code-dot-org/code-dot-org/pull/56488 which added events to sharing options after a user clicks on the 'Share' button.

I grouped the project sharing events as 'finish-project-sharing-events'.

This PR also adds an Amplitude event to the 'export app' feature since product would like to know how often this is being used. And updates the Amplitude sharing events sent via the 'Share' button so that the values are sent instead of the keys.

## Links
[jira](https://codedotorg.atlassian.net/browse/LABS-545)

## Testing story
I tested locally to see the expected events 'sent' in the dev console.

## PR Checklist:

- [ ] Code is well-commented


apps/src/lib/util/AnalyticsConstants.js
```diff
@@ -175,12 +175,26 @@ const EVENTS = {
...
+  FINISH_SHARING_PUBLISH: 'User Clicks Publish In Project Share Dialog',
+  FINISH_SHARING_FB: 'User Clicks Facebook Icon In Project Share Dialog',
+  FINISH_SHARING_TWITTER: 'User Clicks Twitter Icon In Project Share Dialog',
+  FINISH_SHARING_LINK_SEND_TO_PHONE:
+    'User Clicks Send To Phone In Finish Congrats Dialog',
+  FINISH_BUTTON_CERTIFICATE:
+    'User Clicks on Finish Button in Finish Congrats Dialog - Certificate',
+
+  // Export app
+  EXPORT_APP: 'User Exports App From Share Advanced Options',
```
bencodeorg, line 190: It's not a requirement that events are in a group, right?

apps/src/feedback.js
```diff
@@ -529,18 +521,10 @@ FeedbackUtils.prototype.displayFeedback = function (
...
-      study_group: dialog_type,
-      event: event,
-      data_json: JSON.stringify({
-        level_type: options.level ? options.level.skin : null,
-        level_id: options.response ? options.response.level_id : null,
-        level_path: options.response ? options.response.level_path : null,
-        isPerfectBlockCount: isPerfect,
-      }),
+function recordShare(type, appType) {
+  if (Object.prototype.hasOwnProperty.call(EVENTS, type)) {
```
bencodeorg: nit: is this different than `if EVENTS[type]`?

apps/src/feedback.js
```diff
@@ -448,8 +439,9 @@ FeedbackUtils.prototype.displayFeedback = function (
   const publishButtonSelector = '#publish-to-project-gallery-button';
   const publishButton = feedback.querySelector(publishButtonSelector);
   if (publishButton) {
-    dom.addClickTouchEvent(publishButton, () => {
+    dom.addClickTouchEvent(publishButton, () => { // TODOLOG
```
bencodeorg: is "TODOLOG" a convention of ours?

apps/src/feedback.js
```diff
@@ -529,18 +521,10 @@ FeedbackUtils.prototype.displayFeedback = function (
...
-      event: event,
-      data_json: JSON.stringify({
-        level_type: options.level ? options.level.skin : null,
-        level_id: options.response ? options.response.level_id : null,
-        level_path: options.response ? options.response.level_path : null,
-        isPerfectBlockCount: isPerfect,
-      }),
+function recordShare(type, appType) {
+  if (Object.prototype.hasOwnProperty.call(EVENTS, type)) {
+    analyticsReporter.sendEvent(type, {
```
bencodeorg: I think the first argument here needs to be EVENTS[type], right? Or, maybe easier, just use `EVENTS.FINISH_SHARING_PUBLISH`, etc when you call `recordShare` (and then you don't need the if check above)?

bencodeorg (commented):
Thanks for doing this! I think we might send the wrong text to Amplitude right now in `recordShare` (ie, the all-caps key instead of the human readable value), lmk if I'm mistaken though.

cnbrenci:
Thanks for including the export!!

apps/src/lib/util/AnalyticsConstants.js
fisher-alice: Correct. I asked @samantha-code if grouping is helpful and she said 'Grouping events is super helpful especially when making charts on amplitude because I can go straight to that group rather than searching through all events!'

apps/src/feedback.js
fisher-alice: Haha! No - just a note to myself when I first looked at this file to know where to add logging. Removed!

apps/src/feedback.js
fisher-alice: Oh yeah! That is more simple. Will just remove the condition and use the `EVENTS` values for the arguments.

apps/src/feedback.js
fisher-alice: No and `EVENTS[type]` is more simple. Removing condition based on your comment at https://github.com/code-dot-org/code-dot-org/pull/57127#discussion_r1516963716.
But updating the condition in `ShareAllowedDialog`.

apps/src/code-studio/components/ShareAllowedDialog.jsx
```diff
@@ -31,8 +31,8 @@ function recordShare(type, appType) {
   if (!window.dashboard) {
     return;
   }
-  if (Object.prototype.hasOwnProperty.call(EVENTS, type)) {
-    analyticsReporter.sendEvent(type, {
+  if (EVENTS[type]) {
+    analyticsReporter.sendEvent(EVENTS[type], {
```
fisher-alice, line 35: I let @samantha-code know about this upcoming change from key to value.

fisher-alice:
> Thanks for doing this! I think we might send the wrong text to Amplitude right now in `recordShare` (ie, the all-caps key instead of the human readable value), lmk if I'm mistaken though.

Thanks for pointing this out, Ben! I updated in `feedback.js` and in `ShareAllowedDialog.jsx` and let @samantha-code know about the upcoming change.

apps/src/feedback.js
```diff
@@ -529,20 +515,10 @@ FeedbackUtils.prototype.displayFeedback = function (
...
-      data_json: JSON.stringify({
-        level_type: options.level ? options.level.skin : null,
-        level_id: options.response ? options.response.level_id : null,
-        level_path: options.response ? options.response.level_path : null,
-        isPerfectBlockCount: isPerfect,
-      }),
-    });
-  }
+function recordFinishShare(type, appType) {
+  analyticsReporter.sendEvent(EVENTS.type, {
```
bencodeorg: Does this work? I thought you had to use bracket syntax like EVENTS[type] when using a variable as a key.

apps/src/feedback.js
```diff
@@ -259,15 +256,6 @@ FeedbackUtils.prototype.displayFeedback = function (
   const showPuzzleRatingButtons =
     options.response && options.response.puzzle_ratings_enabled;
 
-  if (showingSharing) {
```
bencodeorg, line 262: Nice cleanup!

bencodeorg (approved):
I think one thing to fix but other than that LGTM! Like Cassi said, thanks for adding the export logging -- that will be super helpful information to have!

apps/src/feedback.js
fisher-alice: Thanks Ben! 🙏  Just updated similar to how I updated the events for the 'share' button. 
I think I'm probably working on too many branches simultaneously rn!

PR #57130, opened 2024-03-07
feat(lms): do not sync if not section owner

stephenliang:
Previously, any teacher on the LMS for a course will trigger course syncing even if the section owner may not actually want syncing to occur.

This PR updates the behavior to only trigger course syncing if this is the section owner. The "Sync Course" button is also hidden unless you are the section owner.

## Links

- [jira ticket](https://codedotorg.atlassian.net/browse/P20-758)

## Testing story

1. Created unit test to assert that teacher of a course does not sync if not section owner
2. Manually tested the same scenario on canvas

## PR Checklist:

- [x] Code is well-commented

### DIFF
--- a/dashboard/lib/services/lti.rb
+++ b/dashboard/lib/services/lti.rb
@@ -189,6 +189,9 @@ class Services::Lti
     had_changes = false
     lti_sections = LtiSection.where(lti_course_id: lti_course.id)
 
+    # Only sync the section if the section's owner is equal to the current user
+    return false unless lti_sections.first.section.user_id == current_user_id
+
     # Prune sections that have been deleted in the LMS
     lti_sections.each do |lti_section|
       unless nrps_sections.key?(lti_section.lms_section_id)
@@ -218,6 +221,7 @@ class Services::Lti
         lti_section = LtiSection.create(lti_course_id: lti_course.id, lms_section_id: lms_section_id, section: section)
         had_changes = true
       end
+
       unless lti_section.section.name == section_name
         lti_section.section.update(name: section_name)
         had_changes = true
--- a/dashboard/lib/services/lti.rb
+++ b/dashboard/lib/services/lti.rb
@@ -189,8 +189,8 @@ class Services::Lti
     had_changes = false
     lti_sections = LtiSection.where(lti_course_id: lti_course.id)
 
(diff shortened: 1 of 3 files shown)


dashboard/lib/services/lti.rb
```diff
@@ -176,14 +176,18 @@ def self.sync_course_roster(lti_integration:, lti_course:, nrps_sections:, secti
...
             name: section_name,
             login_type: Section::LOGIN_TYPE_LTI_V1,
           }
         )
         lti_section = LtiSection.create(lti_course_id: lti_course.id, lms_section_id: lms_section_id, section: section)
         had_changes = true
       end
+
+      # Only sync the section if the section's owner is equal to the current user
+      next unless lti_section.section.user_id == current_user_id
```
carl-codeorg: Should this check be moved earlier in the function? It seems like after line 162 where we get the array of sections, we could check the owner of the first section and short-circuit there if we have a mismatch.

carl-codeorg (commented):
One minor comment about the placement of the check for section owner vs. current user. It got me thinking, though, if we can pull the check into the `sync_course` route handler. Would it work to check the current user against the owner of the first section associated with the course that we look up at the start of the db transaction in `sync_course`? Then we could save having to make the external call to the NRPS service if the current user isn't the section owner.

dashboard/lib/services/lti.rb
```diff
@@ -166,10 +166,13 @@ def self.sync_section_roster(lti_integration, lti_section, nrps_members)
   end
 
   # Syncs a course and all its sections from an NRPS response.
-  def self.sync_course_roster(lti_integration:, lti_course:, nrps_sections:, section_owner_id:)
+  def self.sync_course_roster(lti_integration:, lti_course:, nrps_sections:, current_user_id:)
     had_changes = false
     lti_sections = LtiSection.where(lti_course_id: lti_course.id)
 
+    # Only sync the section if there are no sections or the section's owner is equal to the current user
+    return false unless lti_sections.empty? || lti_sections.try(:first).section.user_id == current_user_id
```
carl-codeorg: Is the `.try(:first)` necessary here? Since `lti_sections` is the result of a `.where` it should always have a `first` method.

carl-codeorg (approved):
Had one small comment; overall looks great!

PR #57133, opened 2024-03-08
Cleanup adding Similar Recommender to the Curriculum Catalog

TurnerRiley:
After the Similar Curriculum Recommender launched on the Curriculum Catalog, there were 2 quick bugs that needed cleaning up.

### Set default image
When a curriculum without an image was recommended, there was no default image ready for it to use so this PR implements it. Since the same default image was being used in `CurriculumCatalogCard.jsx`, I moved it to a constants file (which was the `curriculumCatalogShapes.jsx` file). This accounts for nearly all changes in this PR.

Previous look:
missing_image

New look:
FIX_IMAGE

### Render expanded card for curricula without topics
When trying to view the expanded card of a curriculum without topics, it would throw an error from running the recommender within the `hasImportantButNotDesiredTopic` question. This PR ensures that the curriculum has topics before checking them for the recommender.

Currently, the "Nevada Standards Aligned: Beta - CSF Course C" expanded card throws an error. However, with this change it now renders fine:
NO_TOPICS

## Links
Jira ticket: [here](https://codedotorg.atlassian.net/jira/software/c/projects/ACQ/boards/64?assignee=60d62161f65054006980bd71&selectedIssue=ACQ-1594)

## Testing story
Local testing.

### DIFF
--- a/apps/src/templates/curriculumCatalog/CurriculumCatalog.jsx
+++ b/apps/src/templates/curriculumCatalog/CurriculumCatalog.jsx
@@ -1,6 +1,6 @@
 import React, {useState, useEffect} from 'react';
 import PropTypes from 'prop-types';
-import {curriculumDataShape} from './curriculumCatalogShapes';
+import {curriculumDataShape} from './curriculumCatalogConstants';
 import i18n from '@cdo/locale';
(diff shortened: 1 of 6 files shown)


bethanyaconnor (approved):
Thanks for the quick fixes!

Issue #56759, opened 2024-02-23
The value of "record" passed to the callback for "updateRecord" seems to be network response object rather than a record object


snickell:
Edit: this does not repro with Firebase

![Image](https://github.com/code-dot-org/code-dot-org/assets/85528507/0961491a-d6ce-4697-b6d1-f074f5033e88)

PR #57136, opened 2024-03-08
Fix bug where updateRecord returns network response rather than record object

cnbrenci:
Fixed bug where updateRecord was returning the network response rather than the record object.


snickell (approved):
lgtm

snickell:
Fixes [#56759](https://github.com/code-dot-org/code-dot-org/issues/56759)

PR #57137, opened 2024-03-08
Music: sound library tweaks

breville:
This makes some tweaks to the sound library content and UI.

The `launch2024` sound library is updated with 6 new in-house packs.  S3 has the corresponding updates.

The sound selection UI now shows the artist name when viewing packs, and the artist name and pack name when viewing all sounds.

### Screenshots

Screenshot 2024-03-08 at 4 10 15 PM

Screenshot 2024-03-08 at 4 16 41 PM

### DIFF
--- a/apps/src/music/views/SoundsPanel.tsx
+++ b/apps/src/music/views/SoundsPanel.tsx
@@ -254,26 +254,26 @@ const SoundsPanel: React.FunctionComponent<SoundsPanelProps> = ({
     currentSoundRef.current = ref;
   };
 
-  let possibleSounds: SoundEntry[] = [];
-  let rightColumnSounds: SoundEntry[] = [];
+  let possibleSoundEntries: SoundEntry[] = [];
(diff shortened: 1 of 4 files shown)


apps/src/music/views/SoundsPanel.tsx
```diff
@@ -239,23 +254,28 @@ const SoundsPanel: React.FunctionComponent<SoundsPanelProps> = ({
     currentSoundRef.current = ref;
   };
 
-  let possibleSounds: SoundData[] = [];
-  let rightColumnSounds: SoundData[] = [];
+  let possibleSounds: SoundEntry[] = [];
```
bencodeorg: Nit (same with var names below): `possibleSoundEntries`? Maybe too verbose 🤷

bencodeorg (approved):

apps/src/music/views/SoundsPanel.tsx
breville: Good call, will update.

PR #57144, opened 2024-03-08
[Levelbuilder] use Google Blockly for all migrated labs

mikeharv:
Now that Sprite Lab is reliably using Google Blockly, we should also use it for levelbuilder level edit pages. I updated this line to check for the full list of migrated labs. `Poetry` inherits from `GamelabJr` (Sprite Lab), so we don't need to list both. `Flappy` and `Bounce` could have been flipped in this way any time but AFAIK nobody has been editing these levels. I also added `Music` to the list despite the fact that it doesn't appear we ever attempt to render blocks for those levels.

## Links

Jira - https://codedotorg.atlassian.net/browse/CT-338

## Testing story

I tested this by confirming that `Blockly.version` returned `'CDO'` or `'Google'` as expected for each Blockly level type by using my browser console. I also tested that a new Google Blockly block could be rendered as expected in the long instructions markdown field for a Sprite Lab level: 
image
I also checked a level's hint that was listed in the Jira ticket and confirmed that it now uses Google Blockly:
image

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/dashboard/app/views/levels/editors/fields/_blockly.html.haml
+++ b/dashboard/app/views/levels/editors/fields/_blockly.html.haml
@@ -1,6 +1,6 @@
 - content_for(:head) do
   - if @level.is_a?(Blockly) && !@level.uses_droplet?
-    - if @level.is_a?(Dancelab) || @level.is_a?(Poetry)
+    - if [Bounce, Dancelab, Flappy, GamelabJr, Music].any? { |level_type| @level.is_a?(level_type) }
       %script{src: webpack_asset_path('js/googleblockly.js')}
     - else
       %script{src: webpack_asset_path('js/blockly.js')}
--- a/dashboard/app/views/levels/editors/fields/_blockly.html.haml
+++ b/dashboard/app/views/levels/editors/fields/_blockly.html.haml
@@ -1,6 +1,6 @@
 - content_for(:head) do
   - if @level.is_a?(Blockly) && !@level.uses_droplet?
-    - if [Bounce, Dancelab, Flappy, GamelabJr, Music].any? { |level_type| @level.is_a?(level_type) }
+    - if @level.uses_google_blockly?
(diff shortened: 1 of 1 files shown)


bencodeorg (approved):

dashboard/app/views/levels/editors/fields/_blockly.html.haml
```diff
@@ -1,6 +1,6 @@
 - content_for(:head) do
   - if @level.is_a?(Blockly) && !@level.uses_droplet?
-    - if @level.is_a?(Dancelab) || @level.is_a?(Poetry)
+    - if [Bounce, Dancelab, Flappy, GamelabJr, Music].any? { |level_type| @level.is_a?(level_type) }
```
ebeastlake: Outside the scope of this PR: Do we have multiple sources of truth in our codebase for the list of fully migrated labs? If so, how can we consolidate so that when we migrate a lab, the list of level edit pages/block pools/etc. (making up things I think might fall under this umbrella) automatically get updated without us having to remember all the spots to update?

ebeastlake (approved):

dashboard/app/views/levels/editors/fields/_blockly.html.haml
mikeharv: Oh could we use this from each level model?

```
  def uses_google_blockly?
    true
  end
```

mikeharv: I'm on another branch now, but I wonder if this would work:

```
if @level.is_a?(Blockly) && @level.uses_google_blockly?
```

mikeharv: Confirmed that ^ works! FYI @bencodeorg @ebeastlake

bencodeorg: Nice! Good callout Emily

PR #57146, opened 2024-03-08
Adds evidence column and uses it for line annotations.

wilkie:
This adds an `evidence` column to `LearningGoalAiEvaluation` table and fills it with the incoming 'evidence' column in the AI result returned from `aiproxy`. This just rewires the observations column to the new evidence column when it relates to parsing out line annotations.

## Links

- jira ticket: assists in resolving [AITT-430](https://codedotorg.atlassian.net/browse/AITT-430)

## Testing

Validated locally. Added the prompt for lesson 21 from the intended release to follow-up this PR. Performed it with the existing 'Evidence'-less prompt and it did not regress for the existing prompts. It showed no evidence bullet points in the UI and did not capture the bullet points. It also did this for the case where the evidence was put in the Observations column.

When tested against the new prompt, it successfully saved the evidence in the LearningGoalAiEvaluation records. It then rendered those to the UI in the evidence section successfully and accurately. (Though it required a change to the regular expression to allow "Lines" (i.e. "Lines 5-6" instead of just "Line 5-6".)

## PR Checklist:

- [ ] Code is well-commented


dashboard/app/models/learning_goal_ai_evaluation.rb
davidsbailey, line 1: due to the migration I would expect to see some annotations in the model file. I believe the migration needs to be run against the development database and not the test database in order to trigger model annotation.

davidsbailey (approved):
looks good, one nit inline. maybe you can also say what you did to validate?

dashboard/app/models/learning_goal_ai_evaluation.rb
wilkie: Good catch. Added it.

wilkie:
I will do a manual validation with a local aiproxy that uses the code to-be-merged there and have it run some student work. Will have by Monday.

wilkie:
I did validate this locally Monday and edited the top comment to reflect what I did. The internal spreadsheet linked in the channel shows the experimentation with the prompts. This was only tested using the normal pre-JSON prompts but since the JSON-output prompts are only going to be different on the proxy service, they should work very much the same. Specifically, this is testing with the "ai-rubrics-evidence-column" experiment prompt for lesson 21.

@davidsbailey Is this enough to validate the work?

davidsbailey:
I feel great about the amount of accuracy validation! As long as you've also done an end-to-end test of the evidence UI working locally then that seems about right for manual validation.

PR #57150, opened 2024-03-08
[Sprite Lab] Move foreground effects from Poetry to Sprite Lab

mikeharv:
Request from @amy-b:
> I was reviewing the Game Design levels and I thought it would be cool to have some of the `do effect` blocks in that progression for students who want to do confetti effect (etc.) for when a player wins the game, like for [this level](https://levelbuilder-studio.code.org/s/katies-playground/lessons/7/levels/6). The block exists in the [Poetry ](https://levelbuilder-studio.code.org/pools/Poetry/blocks/Poetry_setForegroundEffect/edit)block pool right now. I was having some trouble getting that block to work though. I enabled the Poetry block pool for the level but got the error that the function doesn't exist.

Screen Shot 2024-03-08 at 9 43 42 AM

## Context
Poetry inherits from GamelabJr but not the other way around. In order to make this effect command work in both we should move it up to the parent class. 

This PR adds a `do effect` block to the core Sprite Lab block pool. It solves the problem described above by either moving or copying logic from Poetry's commands/library into Sprite Lab's. For places where the code is copied rather than moved I left comments indicating the source and rationale.

Interestingly, this desire to use effects in Sprite Lab came up last year as well when the CSC team was developing a Story Lab unit. This level subtype is Sprite Lab with an extra `storylab` block pool and `zForegroundEffects` helper library. In that block pool, @dancodedotorg had created a copy of the foreground effect block. Dan initially managed to get it to work by copying ALL of the needed source code into the helper library. He then realized it was a little better to move it all to the helper code for the block. This code includes logic from the Poetry library as well from our Dance repo (the actual origin of some of these effects). As he [said](https://codedotorg.slack.com/archives/C060J3A3BBK/p1709921352155479?thread_ts=1709919876.620709&cid=C060J3A3BBK) on Slack:
> It's such a beautiful mess

Now that the Sprite Lab has its own native command for foreground effects we can also clean up the helper code for Dan's block and delete the library. I confirmed that levels that use the deleted library still load without issue. Unfortunately, the story block needs to retain the name `setForegroundEffect2` as it already exists in student work that we cannot break.

## Links

Slack thread - https://codedotorg.slack.com/archives/C060J3A3BBK/p1709919876620709

## Testing story

I tested the following scenarios:

`gamelab_setForegroundEffect` block in a Sprite Lab project:
image

`Poetry_setForegroundEffect` block in a Poetry project:
image

`storylab_setForegroundEffect2` block in a story project:
image

## Follow-up work

- Make sure the K5 team knows to use the new Sprite Lab block in all new levels (to avoid adding other block pools unnecessarily).

## PR Checklist:

- [ ] Code is well-commented


apps/src/p5lab/constants.js
```diff
@@ -28,3 +28,8 @@ export const EMPTY_IMAGE =
   'data:image/gif;base64,R0lGODlhAQABAIAAAP///wAAACH5BAEAAAAALAAAAAABAAEAAAICRAEAOw==';
 
 export const PlayBehavior = utils.makeEnum('ALWAYS_PLAY', 'NEVER_PLAY');
+
+export const EFFECTS_PALETTES = {
+  ocean: ['#7FD0F5', '#3FABE3', '#2C7DBB', '#1D57A0', '#144188', '#061F4B'],
+  twinkling: ['#FFC702', '#FC9103', '#F17302', '#B83604', '#7E1301'],
+};
```
mikeharv, line 35: Copied from https://github.com/code-dot-org/code-dot-org/blob/mike/effects-for-spritelab/apps/src/p5lab/poetry/constants.js#L7-L28

We only needed these two palettes. The others are exclusively used in Poetry so it felt right to leave them there. I renamed to `EFFECTS_PALETTES` here since this is the only context they are used for here.

apps/src/p5lab/spritelab/CoreLibrary.js
```diff
@@ -55,6 +56,10 @@ export default class CoreLibrary {
         this.p5.drawSprites();
         this.drawVariableBubbles();
         this.drawSpeechBubbles();
+        // Don't show foreground effect in preview
+        if (!this.isPreviewFrame()) {
+          this.foregroundEffects.forEach(effect => effect.func());
+        }
```
mikeharv, line 62: Similar to https://github.com/code-dot-org/code-dot-org/blob/mike/effects-for-spritelab/apps/src/p5lab/poetry/PoetryLibrary.js#L90-L92

This needs to live in both places because each lab has its own `executeDrawLoopAndCallbacks` function.

apps/src/p5lab/utils.js
```diff
@@ -38,3 +40,15 @@ export function createSpriteCollider(sprite) {
...
+  return p5
+    .color('hsb(' + randomInt(0, 359) + ', 100%, 100%)')
+    .toString('#rrggbb');
+}
+
+export function randomColorFromPalette(palette) {
+  return EFFECTS_PALETTES[palette][
+    randomInt(0, EFFECTS_PALETTES[palette].length - 1)
+  ];
+}
```
mikeharv, line 50: Copied from https://github.com/code-dot-org/code-dot-org/blob/mike/effects-for-spritelab/apps/src/p5lab/poetry/commands/utils.js#L8-L16

Poetry doesn't seem to directly import these utils anywhere so I left them in both places. Possibly worth noting is that `randomInt`, which is also used for effects, was already defined in both utils files. This suggests that there is precedent here.

dashboard/config/blocks/GamelabJr/gamelab_setForegroundEffect.json
```diff
@@ -0,0 +1,47 @@
...
+          [
+            "twinkling",
+            "\"twinkling\""
+          ]
+        ]
+      }
+    ],
+    "style": "world_blocks"
+  }
+}
```
mikeharv, line 47: Direct copy of https://github.com/code-dot-org/code-dot-org/blob/mike/effects-for-spritelab/dashboard/config/blocks/Poetry/Poetry_setForegroundEffect.json

The old block cannot be deleted because students already have `Poetry_setForegroundEffect` in their projects and level progress.

apps/src/p5lab/utils.js
ebeastlake: I have a slight preference for only having these defined one place, even if we define them in `p5lab/utils` and have `poetry/commands/utils` import them and re-export them. That would look something like this in `apps/src/p5lab/poetry/commands/utils.js`:

```
import {
  randomInt,
  randomColor, 
  randomColorFromPalette 
} from '../../utils';

import {PALETTES} from '../constants';

// Adjusts P5's randomColorFromPalette function for Poetry's palette
export function randomColorFromPalettePoetry(palette) {
  return randomColorFromPalette(PALETTES[palette]);
}

export { randomInt, randomColor };

// Other stuff goes here...
```

dashboard/config/blocks/storylab/storylab_setForegroundEffect2.js
```diff
@@ -1,381 +1,3 @@
...
-    pop();
-  }
-  drops = drops.filter(function(drop) {
-    return (drop.y < 420 && drop.x < 420);
-  });
-  pop();
-}
-
-//From: https://github.com/code-dot-org/code-dot-org/blob/staging/apps/src/p5lab/poetry/commands/foregroundEffects.js
 function setForegroundEffect2(option) {
```
ebeastlake, line 2: This is where we'd leave a comment for an engineer coming after you that we cannot delete this, right? Since we can't leave JSON comments on `dashboard/config/blocks/GamelabJr/gamelab_setForegroundEffect.json`. Should we add one?

apps/src/p5lab/constants.js
ebeastlake: I feel slightly less strongly about this one than the one below (where code needs to be maintained in two places instead of constants), but I propose the following in `apps/src/p5lab/poetry/constants.js` if we think it's worth having the code represent the hierarchical relationship between these two labs. 

```
 // Import the twinkling and ocean palettes already defined in Sprite Lab
import { EFFECTS_PALETTES } from '../constants';

export const PALETTES = {
  ...EFFECTS_PALETTES,
  grayscale: [
    '#000000', '#333333', '#666666', '#999999', '#CCCCCC', '#EEEEEE', '#FFFFFF',
  ],
  ...
};
```

ebeastlake (approved):
Screenshot 2024-03-08 at 1 48 48 PM

Woo-hoo! Thanks Mike! 🎉

amy-b:
Thank you so much, Mike!

PR #57151, opened 2024-03-08
fix(karma): clear the body tag in the dom on each test execution

stephenliang:
Some tests anchor to the body tag and is not reset per test execution, leading to a case where the DOM is full of elements from prior test runs. This leads to scenarios where a test runs if executed alone but fails if run as part of the whole test suite.

Ensure that the <body> tag is empty before each test execution.

## PR Checklist:

- [x] Code is well-commented

### DIFF
--- a/apps/test/tests-entry.js
+++ b/apps/test/tests-entry.js
@@ -40,6 +40,19 @@ if (testType('unit')) {
     // throwOnConsoleWarningsEverywhere();
 
     clearTimeoutsBetweenTests();
+
+    beforeEach(() => {
+      // Some tests anchor to the body tag and is not reset per test execution, leading to a case where the DOM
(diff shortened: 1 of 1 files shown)


hannahbergam (commented):
I was reading about a similar solution/approach on stack overflow. So at the very least, it sounds like this is an approach folks at other orgs have taken!

lfryemason:
@hannahbergam could you link to the similar approach?

hannahbergam:
> @hannahbergam could you link to the similar approach?

Redoing my google search! Here was one example (see the setDom function) https://github.com/testing-library/react-testing-library/issues/716#issuecomment-1465374138

lfryemason (approved):

PR #57157, opened 2024-03-08
Migrated LibraryIdCopierTest to RTL

jmkulwik:
Migrated LibraryIdCopierTest to React Testing Library.

While doing so, I also found Liam's [extremely helpful doc](https://docs.google.com/document/d/1Yxek5dqYKLJUraB17yCdEFLlJvlOiB1zhSWZAuu3zPU/edit) about our various in-process initiatives. I used that as a guide to move styles to scss, migrate LibraryIdCopier to a functional component, and migrate the component and its test file to typescript.

I hadn't been doing much front-end coding when these migrations were started, so I haven't written any code with styles in SCSS, Functional Components, or Typescript (except maybe a few small edits here and there), so please take a close look! Lots of feedback welcome!


apps/src/code-studio/components/libraries/LibraryIdCopier.tsx
```diff
@@ -0,0 +1,32 @@
...
+import i18n from '@cdo/locale';
+import Button from '@cdo/apps/templates/Button';
+
+export default function LibraryIdCopier({channelId}: {channelId: string}) {
+  const textContainer = useRef<HTMLInputElement>(null);
+  function handleCopyClick() {
+    // textContainer.current will be null prior to render. This is ok since nothing will be
+    // clickable at that point anyway.
+    textContainer.current?.select();
+    navigator.clipboard.writeText(channelId);
```
bencodeorg: Nit: looks like we also have a `copyToClipboard` util that includes a fallback for legacy browsers -- I see this isn't [available until 13.1 in Safari,](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard) and our IT requirements page currently says 13.x) Might not be worth the additional complexity though.

bencodeorg (approved):
I haven't written any RTL tests yet but the updated test looks reasonable, and the other updates look good!

apps/src/code-studio/components/libraries/LibraryIdCopier.tsx
jmkulwik: Good call. Probably worth it to use the util whether or not the workaround is still needed. This is also a good reminder to take another rev on our browser support matrix.

PR #57159, opened 2024-03-08
Error message for launching LTI app outside of the context of a class

juanmanzojr:
If initiating an LTI launch outside the context of a course, an error message will appear letting users know to launch from the context of a course.

Additionally, based on the issuer, an extra message will be displayed linking users to more details about placements for launching and LTI roster sync.
Screenshot 2024-03-21 at 1 17 29 PM

Also gracefully handles optional claims that may be missing within the decoded JWT.

## Links
- jira ticket: [here](https://codedotorg.atlassian.net/browse/P20-765)

## Testing story
New test in `lti_v1_controller_test.rb`


apps/src/lib/ui/lti/sync/LtiSectionSyncDialogHelpers.tsx
```diff
@@ -0,0 +1,21 @@
+import i18n from '@cdo/locale';
+
+export const getRosterSyncErrorMessage = (errorCode: number) => {
+  switch (errorCode) {
```
carl-codeorg: These error codes are a bit confusing to me because all the actual response codes we're handling are 400s. Since 401 and 404 are both standard HTTP error codes that we aren't actually handling here, picking different codes to switch on could ease this confusion. We don't necessarily have to use HTTP codes either, since we have multiple different messages we want to convey based on different 400s.

apps/src/lib/ui/lti/sync/LtiSectionSyncDialogHelpers.tsx
```diff
@@ -0,0 +1,21 @@
...
+  }
+};
+
+export const getRosterSyncIssuerErrorDetails = (issuer: string) => {
+  switch (issuer) {
+    case 'Schoology':
+      return i18n.ltiSectionSyncDialogErrorSchoologyDetails({
+        url: 'example.com',
+      });
+  }
```
carl-codeorg, line 13: Should probably add a case for Canvas, otherwise won't we get undefined?

dashboard/app/controllers/lti_v1_controller.rb
```diff
@@ -120,10 +120,9 @@ def authenticate
 
       user = Queries::Lti.get_user(decoded_jwt)
       target_link_uri = decoded_jwt[:'https://purl.imsglobal.org/spec/lti/claim/target_link_uri']
-
       launch_context = decoded_jwt[Policies::Lti::LTI_CONTEXT_CLAIM]
-      nrps_url = decoded_jwt[Policies::Lti::LTI_NRPS_CLAIM][:context_memberships_url]
-      resource_link_id = decoded_jwt[Policies::Lti::LTI_RESOURCE_LINK_CLAIM][:id]
+      nrps_url = decoded_jwt[Policies::Lti::LTI_NRPS_CLAIM] ? decoded_jwt[Policies::Lti::LTI_NRPS_CLAIM][:context_memberships_url] : nil
```
carl-codeorg: Can we take advantage of [get_claim ](https://github.com/code-dot-org/code-dot-org/blob/staging/dashboard/lib/services/lti.rb#L70-L72) to clean up these ternaries?

dashboard/app/controllers/lti_v1_controller.rb
```diff
@@ -134,12 +133,12 @@ def authenticate
       redirect_params = {
         lti_integration_id: integration.id,
         deployment_id: deployment.id,
-        context_id: launch_context[:id],
+        context_id: launch_context ? launch_context[:id] : nil,
```
carl-codeorg:
```suggestion
        context_id: launch_context&.[](:id),
```

dashboard/app/controllers/lti_v1_controller.rb
```diff
@@ -198,15 +197,22 @@ def sync_course
     unless Policies::Lti.roster_sync_enabled?(current_user)
       return redirect_to home_path
     end
-    params.require([:lti_integration_id, :deployment_id, :context_id, :rlid, :nrps_url]) if params[:section_code].blank?
+
+    if params[:section_code].blank?
+      begin
+        params.require([:lti_integration_id, :deployment_id, :context_id, :rlid, :nrps_url])
+      rescue ActionController::ParameterMissing => _exception
+        return render_sync_course_error("Attempting to sync a course or section from the wrong place.", :bad_request, params[:issuer], 400)
```
carl-codeorg: Would any of these params being missing imply an invalid link placement? Or only the context_id?

carl-codeorg (changes requested):

apps/i18n/common/en_us.json
```diff
@@ -1598,6 +1598,10 @@
   "ltiSectionSyncButtonDisabledAltText": "Roster sync is disabled. To re-enable, go to your account settings.",
   "ltiSectionSyncDialogDescription": "New to Code.org? A [section]({aboutSectionsUrl}) is used to manage your classroom and assign curricula. We've identified the following sections & students associated with your course.\n\nYou can re-sync your sections & students at any time! Learn more [about syncing here]({aboutSyncingUrl}).",
   "ltiSectionSyncDialogError": "We encountered an error when trying to sync with your LMS.",
+  "ltiSectionSyncDialogErrorNoIntegration": "LTI Integration not found",
+  "ltiSectionSyncDialogErrorNoSectionFound": "We couldn't find the given section.",
+  "ltiSectionSyncDialogErrorWrongContext": "Sorry, looks like you are attempting to sync a course or section from the wrong place. Please launch the Code.org Integration from inside the course or section you would like to sync.",
+  "ltiSectionSyncDialogErrorSchoologyDetails": "Read more about supported Schoology placements [here](%{url}).",
```
nicklathe: We don't want to hardcode to Schoology, rather the LMS name of whatever LMS Platform they are on.

nicklathe: Alternatively, we could just remove `Schoology` here, e.g. `Read more about supported placements here`

carl-codeorg: Yeah we could have one support page with placement lists for all supported platforms

nicklathe: Sorry, I see now that this string key is Schoology specific, but I still don't think it needs to be? We can either make it generic (by removing `Schoology` as mentioned), and still can interpolate the specific URL, _or_ interpolate the issuer name here as well.

nicklathe: Yeah, that's my vote. Which would make the URL universal, and we could remove the method `getRosterSyncErrorMessage` further down in the code.

apps/src/lib/ui/lti/sync/LtiSectionSyncDialogHelpers.tsx
nicklathe: Per the comment above, if we just make this universal for all LMS platforms, with a URL that is a landing page, we could remove this method.

apps/src/lib/ui/lti/sync/LtiSectionSyncDialogHelpers.tsx
```diff
@@ -0,0 +1,12 @@
...
+export const getRosterSyncErrorMessage = (error: string) => {
+  switch (error) {
+    case 'wrong_context':
+      return i18n.ltiSectionSyncDialogErrorWrongContext({url: 'example.com'});
+    case 'no_integration':
+      return i18n.ltiSectionSyncDialogErrorNoIntegration();
+    case 'no_section':
+      return i18n.ltiSectionSyncDialogErrorNoSectionFound();
+  }
+};
```
mgc1194: I like this pattern

apps/src/lib/ui/lti/sync/LtiSectionSyncDialogHelpers.tsx
```diff
@@ -1,21 +1,12 @@
 import i18n from '@cdo/locale';
 
-export const getRosterSyncErrorMessage = (errorCode: number) => {
-  switch (errorCode) {
-    case 400:
-      return i18n.ltiSectionSyncDialogErrorWrongContext();
-    case 401:
+export const getRosterSyncErrorMessage = (error: string) => {
+  switch (error) {
```
carl-codeorg, line 4: Since we're using a switch statement here, we get the benefit of falling through to a default case; might as well take advantage of that and include a generic default case.

dashboard/app/controllers/lti_v1_controller.rb
```diff
@@ -198,15 +197,22 @@ def sync_course
     unless Policies::Lti.roster_sync_enabled?(current_user)
       return redirect_to home_path
     end
-    params.require([:lti_integration_id, :deployment_id, :context_id, :rlid, :nrps_url]) if params[:section_code].blank?
+
+    if params[:section_code].blank?
+      begin
+        params.require([:lti_integration_id, :deployment_id, :context_id, :rlid, :nrps_url])
+      rescue ActionController::ParameterMissing => _exception
+        return render_sync_course_error("Attempting to sync a course or section from the wrong place.", :bad_request, "wrong_context")
```
nicklathe: Super nit, there are a mixture of single and double quotes on these new changes (here and below). Our [style guide](https://github.com/code-dot-org/code-dot-org/blob/staging/STYLEGUIDE.md#ruby) prefers single quotes for non-interpolated strings.

dashboard/app/controllers/lti_v1_controller.rb
juanmanzojr: I believe just the context_id, the nrps_url will also be nil as a result of this.

juanmanzojr: I plan to check for context_id and nrps_url here as this would mean they are launching from the wrong context then checking for deployment_id in another block. The rlid is interesting though as I believe it is an optional field. Currently both Canvas and Schoology provide this rlid but only Canvas actually makes use of it.

nicklathe (approved):

dashboard/app/controllers/lti_v1_controller.rb
juanmanzojr: Update: within this rescue I added a switch case based on the param that is missing.

dashboard/lib/services/lti.rb
```diff
@@ -68,7 +68,7 @@ def self.get_claim_from_list(id_token, keys_array)
   end
 
   def self.get_claim(id_token, key)
-    id_token[key] || id_token.dig(Policies::Lti::LTI_CUSTOM_CLAIMS.to_sym, key)
+    id_token[key] || id_token.dig(Policies::Lti::LTI_CUSTOM_CLAIMS.to_sym, key) || id_token.dig(Policies::Lti::LTI_RESOURCE_LINK_CLAIM.to_sym, key) || id_token.dig(Policies::Lti::LTI_NRPS_CLAIM.to_sym, key)
```
carl-codeorg: I'm not sure we should modify get_claim like this. I think it's meant to get a property at the top level or in custom_claims, but I'm thinking we shouldn't start adding more props to dig into. Sorry for misleading here - I think we should keep get_claim as it was, and only use it if it's useful on lines 124 & 125 in the controller

carl-codeorg (approved):

PR #57161, opened 2024-03-08
Use Policies::ScriptActivity.completed to determine whether to show congrats page

bethanyaconnor:
After Turner's PR that counted contained levels hit production, I noticed that the "Print Certificates" button would show up while the congrats page would show an error. This is because the logic to show the button uses `Policies::ScriptActivity.completed?` ([here](https://github.com/code-dot-org/code-dot-org/blob/27c8a57541312998394e24fa97f2a2a166eb3930/dashboard/app/helpers/users_helper.rb#L236)), which in turn calls `completed_progression_levels?` ([here](https://github.com/code-dot-org/code-dot-org/blob/27c8a57541312998394e24fa97f2a2a166eb3930/dashboard/lib/policies/script_activity.rb#L7)), which my logic for the congrats page didn't call. This PR simply updates the congrats_controller to use the same logic as we do to determine whether we should the print certificate button.

Funny enough, I had a local repro before this PR, so I was able to confirm that this works as intended, both when all levels are completed and when not all levels are completed.

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/dashboard/app/controllers/congrats_controller.rb
+++ b/dashboard/app/controllers/congrats_controller.rb
@@ -20,7 +20,7 @@ class CongratsController < ApplicationController
     if curriculum.is_a?(UnitGroup)
       @curriculum_url = course_path(curriculum)
       units = curriculum.units_for_user(current_user)
-      completed_units = UserScript.where(user: current_user, script: units).where.not(completed_at: nil).map(&:script)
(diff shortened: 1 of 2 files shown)


TurnerRiley (approved):
This looks great!

PR #57162, opened 2024-03-08
Load levelbuilder config in gen AI lab

bencodeorg:
Loads initial configuration from levelbuilder for gen AI lab. It loads values themselves (eg, a system prompt) and the visibility of that bit of configuration (ie, one of "editable", "readonly", or "hidden"). It also will hide an entire tab from the UI if all of the fields in a given tab are configured as "hidden".

It also does a bit of refactoring to share some simple code / constants with our levelbuilder configuration.

## Testing story

Tested manually that configuration from levelbuilder appears in the level.

## Follow-up work

Notably, this does not yet appropriately handle updates by the user (this is next!) There's also a bit of iteration happening on what fields go in which tabs, so there will likely be a bit of reorganization here as well.


apps/src/aichat/types.ts
```diff
@@ -43,7 +43,7 @@ export interface AiCustomizations {
   temperature: number;
   systemPrompt: string;
   retrievalContexts: string[];
-  modelCardInfo?: ModelCardInfo;
+  modelCardInfo: ModelCardInfo;
```
bencodeorg, line 46: This being not required generated some weirdness when these keys are used to generate `LevelAiCustomizations`. Easiest thing was to require it, but not sure if we need to keep it optional.

apps/src/aichat/views/modelCustomization/constants.ts
```diff
@@ -0,0 +1,23 @@
+import {Visibility} from '../../types';
+
+export const DEFAULT_RETRIEVAL_CONTEXTS = {
```
bencodeorg: I think we can share these with the levelbuilder config.

apps/src/aichat/views/ModelCustomizationWorkspace.tsx
```diff
@@ -1,21 +1,57 @@
...
+import {
+  DEFAULT_MODEL_CARD_INFO,
+  DEFAULT_PROMPT_CUSTOMIZATIONS,
+  DEFAULT_RETRIEVAL_CONTEXTS,
+} from './modelCustomization/constants';
+import {isHidden} from './modelCustomization/utils';
 
 const ModelCustomizationWorkspace: React.FunctionComponent = () => {
+  const {retrievalContexts, modelCardInfo, botName, temperature, systemPrompt} =
+    useSelector(
```
sanchitmalhotra126: minor: you can use our cool new `useAppSelector` here which has type info about all of the typescript redux stores, so you don't need to cast the `state` object

apps/src/aichat/views/modelCustomization/PublishNotes.tsx
```diff
@@ -1,31 +1,51 @@
...
+import {LabState} from '@cdo/apps/lab2/lab2Redux';
+import {AichatLevelProperties, ModelCardInfo} from '@cdo/apps/aichat/types';
+import {DEFAULT_MODEL_CARD_INFO} from './constants';
+import {isHidden, isDisabled} from './utils';
 
-const INPUTS = [
-  {id: 'intended-uses', text: 'Intended Uses'},
-  {id: 'limitations-and-warnings', text: 'Limitations and Warnings'},
-  {id: 'testing-and-evaluation', text: 'Testing and Evaluation'},
+const INPUTS: {id: keyof ModelCardInfo; text: string}[] = [
```
sanchitmalhotra126: We can probably move this into constants and reuse between here and the levelbuilder edit view?

apps/src/aichat/views/modelCustomization/utils.ts
```diff
@@ -0,0 +1,6 @@
+import {Visibility} from '../../types';
+
+export const isHidden = (visibility: Visibility) =>
+  visibility === Visibility.HIDDEN;
+export const isDisabled = (visibility: Visibility) =>
```
sanchitmalhotra126: nice!

apps/src/aichat/views/modelCustomization/PublishNotes.tsx
bencodeorg: Done!

apps/src/aichat/views/ModelCustomizationWorkspace.tsx
bencodeorg: Done, thanks!

apps/src/aichat/views/modelCustomization/constants.ts
bencodeorg: Updated these to share with levelbuilder.

apps/src/aichat/views/ModelCustomizationWorkspace.tsx
```diff
@@ -1,21 +1,48 @@
...
 const ModelCustomizationWorkspace: React.FunctionComponent = () => {
+  const {retrievalContexts, modelCardInfo, botName, temperature, systemPrompt} =
+    useAppSelector(
+      state =>
+        (state.lab.levelProperties as AichatLevelProperties | undefined)
+          ?.initialAiCustomizations || EMPTY_AI_CUSTOMIZATIONS
+    );
+
+  const showPromptCustomization =
+    isVisible(botName.visibility) ||
```
sanchitmalhotra126, line 22: Just flagging that we'll probably tweak this to remove `botName` from the "Prompt"  panel, but there's also some discussion around putting a "Base Model" dropdown on that panel so I suspect we'll have another round of revisions in general

apps/src/aichat/views/modelCustomization/PromptCustomization.tsx
```diff
@@ -1,49 +1,76 @@
...
+        {isVisible(botName.visibility) && (
+          <div className={styles.inputContainer}>
+            <label htmlFor="chatbot-name">
+              <StrongText>Chatbot name</StrongText>
+            </label>
+            <input
+              id="chatbot-name"
+              value={botName.value}
+              disabled={isDisabled(botName.visibility)}
+              // readOnly might be preferred property for disabling inputs?
```
sanchitmalhotra126, line 35: nit: is this comment still needed?

apps/src/aichat/views/modelCustomization/PromptCustomization.tsx
```diff
@@ -1,49 +1,76 @@
...
+            <div className={styles.horizontalFlexContainer}>
+              <label htmlFor="temperature">
+                <StrongText>Temperature</StrongText>
+              </label>
+              {temperature.value}
+            </div>
+            <input
+              type="range"
+              min="0"
+              max="1"
```
sanchitmalhotra126, line 50: Can use the MIN/MAX_TEMPERATURE constants here (I set those kind of arbitrarily for now but at least this will be consistent with the levelbuilder edit settings)

apps/src/aichat/views/modelCustomization/PromptCustomization.tsx
```diff
@@ -1,49 +1,76 @@
...
-            <StrongText>Chatbot name</StrongText>
-          </label>
-          <input id="chatbot-name" />
-        </div>
-        <div className={styles.inputContainer}>
-          <div className={styles.horizontalFlexContainer}>
-            <label htmlFor="temperature">
-              <StrongText>Temperature</StrongText>
+        {isVisible(botName.visibility) && (
+          <div className={styles.inputContainer}>
```
sanchitmalhotra126, line 27: One idea - would it be a little cleaner to handle the visibility in CSS with classnames? Something like  `<div className={classNames(styles.inputContainer, isHidden(botName.visibility) && moduleStyles.hidden)}>`, where `.hidden` is a class that just applies `display: none`? This way is fine too so no strong preference, but just had the thought while looking at this.

sanchitmalhotra126 (approved):
LGTM! Pulled this locally to play around with some saved levelbuilder settings and it works great!

apps/src/aichat/views/modelCustomization/PromptCustomization.tsx
bencodeorg: mmm yeah no strong preference either, I think I'll leave as-is for now but can definitely update down the line if we prefer.

apps/src/aichat/views/modelCustomization/PromptCustomization.tsx
bencodeorg: Yep good call, I've got this in the next revision 👍

apps/src/aichat/views/modelCustomization/PromptCustomization.tsx
bencodeorg: ah yes I need to sort this out, React was complaining that it should be readOnly. I'll also make I figure it out / remove this this in the next revision.

bencodeorg: Removed in this PR: https://github.com/code-dot-org/code-dot-org/pull/57318

apps/src/aichat/views/modelCustomization/PromptCustomization.tsx
bencodeorg: Updated in this PR: https://github.com/code-dot-org/code-dot-org/pull/57318

PR #47808, opened 2022-08-24
Music Lab: initial prototype work

breville:
This is the initial prototype of **Music Lab**.  

Screen Shot 2022-10-16 at 9 29 42 PM

This standalone app is designed to allow the creation of music with code.

- Initially available standalone at https://studio.code.org/musiclab.
- The current API model allows the scheduling of a sound play at a specific time, provided as a measure number; non-integer values may be specified.
- The timeline displays all audio plays.
- The code under the "when run" block is executed whenever the blocks are changed.  The timeline is updated in realtime to reflect these code changes, whether the song is currently playing or not.
- When the music is playing, the user may trigger events using the "beat pad" which executes additional code which can schedule additional audio plays.  The timeline is updated as soon as these are scheduled.
- The available sounds are specified with a JSON file stored on S3, along with the relevant sounds.
- The instructions are also specified with a JSON file stored on S3, along with the relevant images.
- Keyboard shortcuts:
  - `i`: show/hide instructions
  - `t`: toggle timeline location, above/below the blocks
  - `d`: sample block mode = play block + dropdown
  - `v`: sample block mode = values
  - `p`: sample block mode = play block + values
  - `space`: play/stop song
  - `1`, `2`, `3`, `4`, `5`, `6`: trigger buttons
- Not yet integrated with existing game/script/level system.
- Sharing is not yet implemented.

Some changes of interest:

- The work involved in integrating with the game/script/level system: https://github.com/code-dot-org/code-dot-org/pull/48456
- Discussion of the audio scheduling model: https://github.com/code-dot-org/code-dot-org/pull/48538

PR #57166, opened 2024-03-11
Music: revive the advanced model

breville:
This revives the original Music Lab programming model, which came to be known as the advanced model, which uses variables and absolute positioning of play sound events.  This is due to new interest from the curriculum team in using something like this for grades 6-12.  

This is designed for internal evaluation only.  It can be viewed by using the `?blocks=advanced` URL parameter.

It's also certainly possible that we could build a new programming model off of this one.

The same model can be seen working in the very first PR for Music Lab at https://github.com/code-dot-org/code-dot-org/pull/47808.  

Screenshot 2024-03-11 at 6 05 08 PM

### DIFF
--- a/apps/src/blockly/googleBlocklyWrapper.ts
+++ b/apps/src/blockly/googleBlocklyWrapper.ts
@@ -755,16 +755,18 @@ function initializeBlocklyWrapper(blocklyInstance: GoogleBlocklyInstance) {
 
     blocklyWrapper.setMainWorkspace(workspace);
 
-    // Same flyout callbacks are used for both main workspace categories
-    // and categories when modal function editor is enabled.
-    workspace.registerToolboxCategoryCallback(
-      'PROCEDURE',
-      functionsFlyoutCategory
-    );
-    workspace.registerToolboxCategoryCallback(
-      'VARIABLE',
-      variablesFlyoutCategory
-    );
+    if (!optOptionsExtended.useBlocklyDynamicCategories) {
+      // Same flyout callbacks are used for both main workspace categories
+      // and categories when modal function editor is enabled.
+      workspace.registerToolboxCategoryCallback(
+        'PROCEDURE',
+        functionsFlyoutCategory
+      );
+      workspace.registerToolboxCategoryCallback(
+        'VARIABLE',
+        variablesFlyoutCategory
+      );
+    }
 
     workspace.registerToolboxCategoryCallback(
       'Behavior',
--- a/apps/src/music/player/sequencer/AdvancedSequencer.ts
+++ b/apps/src/music/player/sequencer/AdvancedSequencer.ts
@@ -30,7 +30,7 @@ export default class AdvancedSequencer extends Sequencer {
       return;
     }
 
-    this.addNewEvent<SoundEvent>({
+    this.playbackEvents.push({
       id,
       type: 'sound',
       length: soundData.length,
@@ -38,7 +38,7 @@ export default class AdvancedSequencer extends Sequencer {
       blockId,
       triggered: false,
       when: measure,
-    });
+    } as PlaybackEvent);
   }
 
   createTrack() {
@@ -63,9 +63,5 @@ export default class AdvancedSequencer extends Sequencer {
 
   getLastMeasure(): number {
     return 0;
-  }
-
-  private addNewEvent<T extends PlaybackEvent>(event: T) {
-    this.playbackEvents.push(event);
   }
 }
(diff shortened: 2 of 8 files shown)


apps/src/blockly/googleBlocklyWrapper.ts
```diff
@@ -755,16 +755,18 @@ function initializeBlocklyWrapper(blocklyInstance: GoogleBlocklyInstance) {
...
+      // and categories when modal function editor is enabled.
+      workspace.registerToolboxCategoryCallback(
+        'PROCEDURE',
+        functionsFlyoutCategory
+      );
+      workspace.registerToolboxCategoryCallback(
+        'VARIABLE',
+        variablesFlyoutCategory
+      );
+    }
```
breville, line 769: @mikeharv This appeared to be necessary since our implementations don't yet work for non-string toolbox definitions, per:
https://github.com/code-dot-org/code-dot-org/blob/7728bbaa9a8c21bf46ef3fd1a10fcc0cc4382b7d/apps/src/blockly/addons/cdoUtils.ts#L398-L403

mikeharv: This feels reasonable to me to unblock adding a standard Variables dynamic category. For context, we can't use the included flyout callback in other labs because we need to support levels that also include hard-coded block XML in this category. This was a feature of CDO Blockly.

That being said, we don't do any merging of dynamically created blocks with level XML blocks in the Functions category. It should be safe to pull the `functionsFlyoutCategory` logic out of this new conditional. We've been using this to date in Music Lab. The main benefit it offers is to make things compatible with the modal function editor. I understand that Music Lab doesn't use that, but it shouldn't hurt anything to make it an option.

apps/src/music/player/sequencer/AdvancedSequencer.ts
```diff
@@ -0,0 +1,71 @@
...
+    isBlockInsideWhenRun: boolean,
+    blockId: string
+  ) {
+    const soundData = MusicLibrary.getInstance()?.getSoundForId(id);
+    if (!soundData) {
+      this.metricsReporter.logWarning('Could not find sound with ID: ' + id);
+      return;
+    }
+
+    this.addNewEvent<SoundEvent>({
```
sanchitmalhotra126: nit: might as well just call `playbackEvents.push` here instead of using the private method?

apps/src/music/views/MusicView.jsx
```diff
@@ -446,7 +450,9 @@ class UnconnectedMusicView extends React.Component {
       lastMeasure: this.sequencer.getLastMeasure(),
     });
     this.props.addOrderedFunctions({
-      orderedFunctions: this.sequencer.getOrderedFunctions(),
+      orderedFunctions:
+        this.sequencer.getOrderedFunctions &&
+        this.sequencer.getOrderedFunctions(),
```
sanchitmalhotra126: nit: should this be something like `this.sequencer.getOrderedFunctions?.() || []` to be safe?

sanchitmalhotra126 (approved):
Thanks for reviving this! Glad to see it wasn't as much code as I thought!

apps/src/blockly/googleBlocklyWrapper.ts
breville: That's fair.  That said, could it be more future-proof to use Blockly's implementation rather than our own "fork" of it, if we don't need to use anything in our fork?

apps/src/music/player/sequencer/AdvancedSequencer.ts
breville: Good call.

apps/src/music/views/MusicView.jsx
breville: TIL `?.()`.  Done.

apps/src/blockly/googleBlocklyWrapper.ts
mikeharv: The problem is that Blockly's implementation doesn't do what we need it to. If we just used Blockly's implementation, old levels would be missing blocks in the variables category. 
Blockly's implementation (defined [here](https://github.com/google/blockly/blob/84218fe9b1c82e4795cdd4c8e74aa73aac18a1aa/core/variables.ts#L84-L106)) will always populate the category with the button and a dynamic set of blocks (`variables_set`, `math_change`, and `variables_get`) based on the variables currently defined on the workspace. Our levels require that we do that and additionally include blocks from the xml for whatever lesson-specific context applies (e.g. prompt blocks or standard blocks with pre-set initial values).

breville: Oh sorry, I meant in the case of Music Lab, where we don't need these extra features thus far.

mikeharv: Oh, you were probably talking about my suggestion re: Functions, then? I guess I don't have a strong enough opinion then. My instinct is still that internal consistency would be better. I don't worry so much about our custom logic breaking Music Lab because we already depend on it for all other labs. If we think it's easy for other engineers to understand that Music Lab is an exception here that's fine with me.

breville: Ah yes, I should have been more clear.  Thanks!

PR #57173, opened 2024-03-11
Use .0 at the end of unplugged levels

etaderhold:
This change makes the level header for unplugged levels have a ".0" suffix (as these are always before level 1 in a lesson). Students who have marked this lesson complete will have that reflected in the table.

image

## Links

[JIRA](https://codedotorg.atlassian.net/browse/TEACH-903)

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/templates/sectionProgressV2/LevelProgressHeader.jsx
+++ b/apps/src/templates/sectionProgressV2/LevelProgressHeader.jsx
@@ -67,7 +67,7 @@ export default function ExpandedProgressColumnHeader({
       >
         {level.sublevels?.length > 0 && <FontAwesome icon="caret-right" />}
         <div>{`${lesson.relative_position}.${
-          level.kind === 'unplugged' ? 0 : level.bubbleText
+          level.isUnplugged ? 0 : level.bubbleText
         }`}</div>
         {level.kind === 'assessment' && (
           <FontAwesome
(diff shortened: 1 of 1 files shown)


apps/src/templates/sectionProgressV2/LevelProgressHeader.jsx
```diff
@@ -79,6 +81,8 @@ export default function ExpandedProgressColumnHeader({
     [lesson, level, toggleExpandedChoiceLevel, isExpandable]
   );
 
+  console.log(level);
```
lfryemason: Extra log line

apps/src/templates/sectionProgressV2/LevelProgressHeader.jsx
```diff
@@ -66,7 +66,9 @@ export default function ExpandedProgressColumnHeader({
         onClick={() => toggleExpandedChoiceLevel(level)}
       >
         {level.sublevels?.length > 0 && <FontAwesome icon="caret-right" />}
-        <div>{lesson.relative_position + '.' + level.bubbleText}</div>
+        <div>{`${lesson.relative_position}.${
+          level.isUnplugged ? 0 : level.bubbleText
```
lfryemason, line 70: Since this is returned as undefined, should we instead fix this on the backend?

etaderhold: I did some digging and found the `undefined` value is being set on the front end [here](https://github.com/code-dot-org/code-dot-org/blob/a78b1ef6d8cc2a88d10ccb07207791960b0236cd/apps/src/templates/progress/progressHelpers.js#L284). My only concern about changing it there would be unintended side effects on other pages.

lfryemason: Seems very intentional then :\

lfryemason (approved):

PR #57172, opened 2024-03-11
update TA FAB icon

davidsbailey:
Finishes https://codedotorg.atlassian.net/browse/AITT-475. Positions bot + background circle as one image and TA oval as another, because this will make it easier to animate the css in the next task (https://codedotorg.atlassian.net/browse/AITT-500). 

### before
Screenshot 2024-03-11 at 9 52 26 AM

Screenshot 2024-03-11 at 10 26 41 AM

### after
Screenshot 2024-03-11 at 9 52 08 AM

Screenshot 2024-03-11 at 10 53 30 AM

### spec

original figma:

Screenshot 2024-03-11 at 9 51 45 AM

figma scaled to 66x64:

AI Teaching Assistant teal 66x64

## Testing story

I expect some eyes tests to need updating when this PR ships.

PR #57182, opened 2024-03-11
use pulse animation on TA floating action button

davidsbailey:
Finishes https://codedotorg.atlassian.net/browse/AITT-500 . 

There is logic in this PR to make sure that we only show the pulse animation when/if:
* it is the user's first time viewing the TA bot this session
* the AI bot and TA overlay images have both finished loading

## screenshots

https://github.com/code-dot-org/code-dot-org/assets/8001765/86be4de6-baf2-4ac9-a7f0-c81060a8ab84

For past screenshots see https://github.com/code-dot-org/code-dot-org/pull/57172.

## Testing story

new unit tests including first RTL tests for this component.

RTL really encourages you to add tags like "name" and "alt" in order to be able to write your tests, which seems like a great forcing function for accessibility!

### DIFF
--- a/apps/src/templates/rubrics/RubricFloatingActionButton.jsx
+++ b/apps/src/templates/rubrics/RubricFloatingActionButton.jsx
@@ -29,6 +29,11 @@ function RubricFloatingActionButton({
   const [isOpen, setIsOpen] = useState(
     JSON.parse(tryGetSessionStorage(sessionStorageKey, false)) || false
   );
+  // Show the pulse if this is the first time the user has seen the FAB in this
+  // session. Depends on other logic which sets the open state in session storage.
+  const [showPulse] = useState(
+    JSON.parse(tryGetSessionStorage(sessionStorageKey, null)) === null
+  );
 
   const eventData = useMemo(() => {
     return {
@@ -72,11 +77,15 @@ function RubricFloatingActionButton({
 
   const fabIcon = aiEnabled ? aiFabIcon : rubricFabIcon;
 
+  const classes = showPulse
+    ? classNames(style.floatingActionButton, style.pulse)
+    : style.floatingActionButton;
+
   return (
     <div id="fab-contained">
       <button
         id="ui-floatingActionButton"
-        className={classNames(style.floatingActionButton, style.pulse)}
+        className={classes}
         // I couldn't get an image url to work in the SCSS module, so using an inline style for now
         style={{backgroundImage: `url(${fabIcon})`}}
         onClick={handleClick}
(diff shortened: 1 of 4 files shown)


apps/src/templates/rubrics/rubrics.module.scss
```diff
@@ -15,6 +29,11 @@
   border-radius: 30px;
 }
```
wilkie: Would this work to set the opacity on the `img` tag?

```suggestion
  border-radius: 30px;
  
  > img {
    opacity: 1.0;
  }
}
```

apps/src/templates/rubrics/RubricFloatingActionButton.jsx
```diff
@@ -71,21 +79,36 @@ function RubricFloatingActionButton({
...
+        className={classes}
         onClick={handleClick}
         type="button"
-      />
+      >
+        <img
+          alt="AI bot"
+          src={fabIcon}
+          onLoad={() => !isFabImageLoaded && setIsFabImageLoaded(true)}
+          style={{opacity: 1.0}}
```
wilkie: Maybe I'm missing a weird interaction with the animation. This could be in the stylesheet, no? Or, at least, why is it need to be set?

```suggestion
```

davidsbailey: good catch, moved to the stylesheet! it needs to be set in order to override this rule: https://github.com/code-dot-org/code-dot-org/blob/ec0a63a5085bae65f40a59e9ec40f737c2cf47ce/dashboard/app/assets/stylesheets/application.scss#L3330-L3331

apps/src/templates/rubrics/rubrics.module.scss
davidsbailey: this looks like it should work, but my first though was just `img {` since it is a direct descendant.

wilkie: That should work. `img {` is less specific than `> img {` which only matches `img` that is a direct child of the button. (but not in any way that overrides the other). CSS is complex. 😓

wilkie (approved):
LGTM! 👍 

I didn't spin it up to look at the visual, nor do I have the discerning eye for that. Code and tests look great! Happy to see more RTL testing creeping into our rubric components.

apps/src/templates/rubrics/rubrics.module.scss
davidsbailey: apologies, I had it backwards, you are correct!

PR #57183, opened 2024-03-11
Add project saving to python lab

molly-moen:
Enable project saving in Python lab. This was fairly straightforward thanks to the previous lab2 work. Some of the work here is python lab specific, but most should be reusable for other codemirror-based labs. This PR includes:
- Adding python lab to the list of channel backed levels in `game.rb`. This will need to happen every time we add a new project-backed lab2 lab.
- Refactor our source validator code. It was previously hard-coded to only support saving blockly sources. Now that we have python lab we needed to expand this logic. I created 3 separate source validators: blockly, python and default. On level load, we use the app type to decide which validator to use.
- Update `CodeEditor.tsx` (only used in Python Lab for now but should be reused/logic should be pulled out of it for future codemirror based labs) to update the editor with new code when we have a new channel id or initial sources. This logic is similar to what we do in [Music lab](https://github.com/code-dot-org/code-dot-org/blob/a78b1ef6d8cc2a88d10ccb07207791960b0236cd/apps/src/music/views/MusicView.jsx#L228). This component will probably be replaced/heavily reworked once we pull in the CDO IDE changes, but we can utilize the logic added here for handling page reloads.

I also added a second python lab all the things level so I could test switching levels without a page reload. Projects do reload correctly now! Right now the output doesn't reset, but since this is a dummy component I didn't think it was worth fixing this.

## Links
- jira ticket: [CT-404](https://codedotorg.atlassian.net/browse/CT-404)

## Testing story
Tested locally that python lab now has project saving, and music lab project saving is unaffected.

## PR Checklist:

- [ ] Code is well-commented


apps/src/lab2/projects/sourcesApi.ts
```diff
@@ -18,7 +19,8 @@ export async function get(
   return HttpClient.fetchJson<ProjectSources>(
     rootUrl(channelId),
     {},
-    SourceResponseValidator
+    Lab2Registry.getInstance().getSourceResponseValidator() ||
+      DefaultSourceResponseValidator
```
ebeastlake: super nit: Would it be better to break this up into two lines?
```
const sourceResponseValidator = Lab2Registry.getInstance().getSourceResponseValidator() ||
      DefaultSourceResponseValidator;
return HttpClient.fetchJson<ProjectSources>(
    rootUrl(channelId),
    {}, sourceResponseValidator);
```

apps/src/lab2/responseValidators.ts
```diff
@@ -52,3 +80,32 @@ export class ValidationError extends Error {
     Object.setPrototypeOf(this, ValidationError.prototype);
   }
 }
+
+export function setValidatorForAppType(appName: AppName) {
+  if (appName === 'pythonlab') {
+    Lab2Registry.getInstance().setSourceResponseValidator(
```
ebeastlake: If you store the result of `Lab2Registry.getInstance()` in a variable on line 85, can you simplify the subsequent code? i.e.,

```
const registry = Lab2Registry.getInstance();
if (appName === 'pythonlab') {
  registry.setSourceResponseValidator(PythonSourceResponseValidator);
} else if (BLOCKLY_LABS.includes(appName)) {
  registry.setSourceResponseValidator(BlocklySourceResponseValidator);
} else {
  registry.setSourceResponseValidator(DefaultSourceResponseValidator);
}
```

apps/src/lab2/responseValidators.ts
```diff
@@ -52,3 +80,32 @@ export class ValidationError extends Error {
...
+    );
+  } else {
+    // Everything else uses the default validator
+    Lab2Registry.getInstance().setSourceResponseValidator(
+      DefaultSourceResponseValidator
+    );
+  }
+}
+
+function sourceValidatorHelper(
```
ebeastlake, line 103: What do you think about something like this to reduce some of the repeated code?

```
function parseJSON(source: string): any {
  try {
    return JSON.parse(source);
  } catch (e) {
    throw new ValidationError(`Error parsing JSON: ${e}`);
  }
}

function checkRequiredField(object: Record<string, any>, field: string): void {
  if (object[field] === undefined) {
    throw new ValidationError(`Missing required field: ${field}`);
  }
}

function sourceValidatorHelper(
  response: Record<string, unknown>,
  validate: (source: any) => void
): ProjectSources {
  checkRequiredField(response, 'source');
  const parsedSource = parseJSON(response.source as string);
  validate(parsedSource);
  return response as ProjectSources;
}

export const BlocklySourceResponseValidator: ResponseValidator<ProjectSources> = response => {
  return sourceValidatorHelper(response, blocklySource => {
    checkRequiredField(blocklySource, 'blocks');
  });
};

export const PythonSourceResponseValidator: ResponseValidator<ProjectSources> = response => {
  return sourceValidatorHelper(response, pythonSource => {
    checkRequiredField(pythonSource, 'main.py');
  });
};

export const DefaultSourceResponseValidator: ResponseValidator<ProjectSources> = response => {
  checkRequiredField(response, 'source');
  return response;
};
```
Forgive any TypeScript mistakes here!

apps/src/lab2/responseValidators.ts
```diff
@@ -52,3 +80,32 @@ export class ValidationError extends Error {
...
+  response: Record<string, unknown>,
+  appSpecificValidator: (response: Record<string, unknown>) => void
+): ProjectSources {
+  if (!response.source) {
+    throw new ValidationError('Missing required field: source');
+  }
+  appSpecificValidator(response);
+  return response as unknown as ProjectSources;
```
ebeastlake, line 111: Why does this need to have `as unknown as`?

apps/src/pythonlab/PythonEditor.tsx
```diff
@@ -3,37 +3,61 @@ import {darkMode} from '@cdo/apps/lab2/views/components/editor/editorThemes';
...
   const handleRun = () => {
     const parsedData = data ? (data as PermissionResponse) : {permissions: []};
     // For now, restrict running python code to levelbuilders.
     if (parsedData.permissions.includes('levelbuilder')) {
       dispatch(appendOutput('Running code...'));
-      runPythonCode(code);
+      if (source) {
+        // TODO: will need to handle a potentially nested main.py
+        const code = source['main.py'];
+        if (typeof code === 'string') {
```
ebeastlake, line 40: What would this be if not a string?

ebeastlake (approved):
Nice work! It looks good to me generally; I just made a couple of suggestions for refactoring. You might get good additional input from a second set of eyes, but it also doesn't seem like a high-risk change.

apps/src/pythonlab/PythonEditor.tsx
molly-moen: `main.py` could also theoretically be a folder name which would mean the contents are an object.

apps/src/lab2/responseValidators.ts
molly-moen: This is how you force convert a type to another type that TypeScript thinks is not compatible. `unknown` is a version of `any` that can only be used with more type assertions [reference](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-0.html#new-unknown-top-type). Since we are getting data from an outside source, we validate it, then cast it once we know it's valid.

apps/src/lab2/responseValidators.ts
molly-moen: good call, will update

apps/src/lab2/responseValidators.ts
molly-moen: will update!

molly-moen: I did a modified version of this. Instead of a `checkRequiredField` method I made a `throwMissingFieldError` method that handles throwing the exception. The typing was different for each version of the method so it would be messy to try to make it work as one method. It will also get more complicated for python lab once we have nested folders.

apps/src/lab2/responseValidators.ts
```diff
@@ -1,32 +1,49 @@
 import {ResponseValidator} from '@cdo/apps/util/HttpClient';
-import {BlocklySource, LevelProperties, ProjectSources} from './types';
+import {AppName, BlocklySource, LevelProperties, ProjectSources} from './types';
+import Lab2Registry from './Lab2Registry';
+import {BLOCKLY_LABS} from './constants';
+import {NestedSourceCode} from '@cdo/apps/pythonlab/pythonlabRedux';
 
-export const SourceResponseValidator: ResponseValidator<
```
sanchitmalhotra126, line 4: Potentially overkill - but just as a thought, what do you think of keeping a top-level `SourceResponseValidator` that internally delegates to the appropriate validator for the current app name? That way instead of storing a validator object on the registry, we can just store the current app name, and this top-level validator could look up the app name in the registry and pick the right delegate internally. And then the sourcesApi could still use the top-level validator without needing to do a lookup in the registry.

apps/src/lab2/responseValidators.ts
```diff
@@ -52,3 +69,39 @@ export class ValidationError extends Error {
     Object.setPrototypeOf(this, ValidationError.prototype);
   }
 }
+
+export function setValidatorForAppType(appName: AppName) {
```
sanchitmalhotra126: Thanks for refactoring this code! Excited to see this list grow 🎉

apps/src/lab2/responseValidators.ts
```diff
@@ -1,32 +1,49 @@
...
+  return sourceValidatorHelper(response, blocklyValidator);
+};
 
-  return response as unknown as ProjectSources;
+// Validator for Python sources.
+export const PythonSourceResponseValidator: ResponseValidator<
+  ProjectSources
+> = response => {
+  const pythonValidator = (source: string) => {
+    const pythonSource = parseJSON<NestedSourceCode>(source);
```
sanchitmalhotra126: To confirm - is python source code stored as stringified JSON in the main.json file? Wondering because don't we store Java lab code as plain non-stringified JSON, and if so would we want to do the same for multi-file python projects?

sanchitmalhotra126 (approved):
Nice, exciting to see this! A couple comments/questions but nothing blocking

apps/src/lab2/responseValidators.ts
molly-moen: that's a good idea! I was stuck on how we get the app name but storing it on the registry makes sense, and it's likely some other feature will need that info in the future anyway.

apps/src/lab2/responseValidators.ts
molly-moen: very good point. I was reproducing the blockly pattern of stringified json but using the Java Lab pattern is better for multi file. I'll update.

thomasoniii (approved):
lgtm 🚀

apps/src/lab2/responseValidators.ts
molly-moen: I refactored this to match the Java Lab pattern!

apps/src/lab2/responseValidators.ts
```diff
@@ -1,32 +1,79 @@
...
+  LevelProperties,
+  NestedSourceCode,
+  ProjectSources,
+} from './types';
+import Lab2Registry from './Lab2Registry';
+import {BLOCKLY_LABS} from './constants';
 
-export const SourceResponseValidator: ResponseValidator<
+// Validator for Blockly sources.
+export const BlocklySourceResponseValidator: ResponseValidator<
```
sanchitmalhotra126: These no longer need to be exported right?

apps/src/lab2/Lab2Registry.ts
```diff
@@ -1,17 +1,24 @@
 // Registry for Lab singletons
 
+import {ResponseValidator} from '../util/HttpClient';
 import LabMetricsReporter from './Lab2MetricsReporter';
 import ProjectManager from './projects/ProjectManager';
+import {AppName, ProjectSources} from './types';
 
 export default class Lab2Registry {
   private projectManager: ProjectManager | null;
+  private sourceResponseValidator:
```
sanchitmalhotra126: can remove?

sanchitmalhotra126 (approved):
LGTM!

apps/src/lab2/responseValidators.ts
molly-moen: yep, good catch

apps/src/lab2/Lab2Registry.ts
molly-moen: yep!

apps/src/lab2/responseValidators.ts
ebeastlake: This was a nice refactor suggestion, thanks, @sanchitmalhotra126!

PR #57185, opened 2024-03-11
Finish converting Google Blockly custom blocks code to TypeScript

molly-moen:
Note: I would recommend viewing this with "hide whitespace" enabled.

This is the last of the planned work to convert our Google Blockly related code to TypeScript! There are a few files not 
converted, see #56907 for why we aren't converting every single file.

This PR is a little larger than I'd like, but all the changes are very low risk and straightforward. It is pretty much entirely typing and null checks, as well as using the TypeScript version of i18n strings.

Two interesting things to note:
- Blockly does not type their Javascript Generator; it is [typed as any](https://github.com/google/blockly/blob/84218fe9b1c82e4795cdd4c8e74aa73aac18a1aa/typings/javascript.d.ts#L45). Therefore we don't get nice typings on our generator code yet. 
- We have custom methods to generically add domToMutation, mutationToDom, loadExtraState and saveExtraState to various blocks. These don't use standard Block methods but instead access fields using `this[fieldName]`. This meant I had to type `this` as `Record<string, any>` rather than as `Block`.

## Links
- jira ticket: [CT-401](https://codedotorg.atlassian.net/browse/CT-401)

## Testing story
Tested locally. Most of the changes here were around behaviors, procedures, and mini toolboxes. I made sure to test those block types.

## Follow-up work
There are probably some minor improvements and clean up to do after this work is in. I haven't file a specific task for this, but is something to keep in mind as we work on Blockly code. The types may not yet be perfect, and there could be things I missed. But working with Blockly code should be easier now, with much better inline documentation!

## PR Checklist:

- [ ] Code is well-commented


apps/src/blockly/customBlocks/googleBlockly/spritelabBlocks.ts
```diff
@@ -24,11 +32,14 @@ export const blocks = {
...
-  initializeMiniToolbox(miniToolboxBlocks, renderToolboxBeforeStack = false) {
+  initializeMiniToolbox(
+    _miniToolboxBlocks: string[],
+    renderToolboxBeforeStack = false
+  ) {
     // Function to create the flyout
-    const createFlyoutField = function (block) {
+    const createFlyoutField = function (block: Block) {
       const flyoutKey = CdoFieldFlyout.getFlyoutId(block);
-      const flyoutField = new Blockly.FieldFlyout(_, {
```
molly-moen, line 31: This should be a string, for some reason (my guess is error on our end) we were passing in lodash. I set it to an empty string since this clearly isn't used for anything.

apps/tsconfig.build.json
```diff
@@ -32,5 +32,6 @@
     "resolveJsonModule": true,
     "incremental": true,
     "sourceMap": true,
+    "downlevelIteration": true,
```
molly-moen, line 35: this allows us to do `for...of` type iteration in typescript ([reference](https://www.typescriptlang.org/tsconfig#downlevelIteration))

apps/src/blockly/types.ts
```diff
@@ -87,7 +88,8 @@ export interface BlocklyWrapperType extends GoogleBlocklyType {
   FieldToggle: typeof CdoFieldToggle;
   FieldFlyout: typeof CdoFieldFlyout;
   FieldBitmap: typeof CdoFieldBitmap;
-  JavaScript: typeof javascriptGenerator;
+  FieldVariable: typeof CdoFieldVariable;
+  JavaScript: JavascriptGeneratorType;
```
ebeastlake, line 92: Did this get imported in this file, assuming it needs to be?

apps/src/blockly/types.ts
```diff
@@ -358,3 +364,6 @@ export type PointerMetadataMap = {
 };
 
 export type BlockColor = [number, number, number];
+
+// Blockly defines this as any.
+export type JavascriptGeneratorType = typeof javascriptGenerator;
```
ebeastlake, line 369: Ah, here it is.

ebeastlake (approved):

PR #57187, opened 2024-03-11
Music: enable advanced controls for ToneJS player behind URL param

sanchitmalhotra126:
Enable the "advanced controls" panel to expose some ToneJS player-specific advanced features, for internal experimentation only. These include manual BPM and key control, and loop settings. This can be enabled by `advanced-controls-enabled=true` in the URL, but requires `player=tonejs` as well to work (e.g. `studio.code.org/projectbeats?player=tonejs&advanced-controls-enabled=true`).

Screenshot 2024-03-12 at 11 19 45 AM

## Links

https://codedotorg.atlassian.net/browse/LABS-530

## Testing story

Tested on /projectbeats, intro script, and standalone projects

### DIFF
--- a/apps/src/music/player/SamplePlayerWrapper.ts
+++ b/apps/src/music/player/SamplePlayerWrapper.ts
@@ -39,7 +39,7 @@ class SamplePlayerWrapper implements AudioPlayer {
   }
 
   async loadInstrument(): Promise<void> {
-    console.log('Not supported');
+    console.warn('loadInstrument not supported');
   }
 
   isInstrumentLoaded(): boolean {
@@ -64,7 +64,7 @@ class SamplePlayerWrapper implements AudioPlayer {
   }
 
   async playSequenceImmediately(): Promise<void> {
-    console.log('Not supported');
+    console.warn('playSequenceImmediately not supported');
   }
 
   cancelPreviews(): void {
@@ -80,7 +80,7 @@ class SamplePlayerWrapper implements AudioPlayer {
   }
 
   scheduleSamplerSequence(): void {
-    console.log('Not supported');
+    console.warn('scheduleSamplerSequence not supported');
   }
 
   async start(startPosition = 1) {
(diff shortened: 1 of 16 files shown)


apps/src/music/player/SamplePlayerWrapper.ts
```diff
@@ -100,6 +100,22 @@ class SamplePlayerWrapper implements AudioPlayer {
...
+  }
+
+  setLoopEnd() {
+    console.log('Not supported');
+  }
+
+  jumpToPosition() {
+    console.log('Not supported');
+  }
+
```
thomasoniii: Would it better to make the messages descriptive so we know what's called? e.g.

```
setLoopEnabled(): void {
  console.log('setLoopEnabled is not supported');
}
```

But I'm not sure if that'd just add more noise to it, so I'll defer. Also, could these be better sent out via `console.error`? Again, no strong preference, just raising an option.

thomasoniii (approved):
lgtm 🚀

apps/src/music/player/SamplePlayerWrapper.ts
sanchitmalhotra126: Good point, I'll update these. In theory we should never hit these so probably better to make them more descriptive if it does somehow occur.

apps/src/music/views/advanced-controls.module.scss
```diff
@@ -0,0 +1,40 @@
+@import 'color.scss';
+
+.container {
+  display: flex;
+  flex-direction: column;
+  color: $neutral_light;
+  padding: 10px;
+  background: $neutral_dark70;
```
breville: minor: I'd suggest `$neutral_dark80`.

apps/src/music/player/ToneJSPlayer.ts
```diff
@@ -44,7 +44,7 @@ class ToneJSPlayer implements AudioPlayer {
     );
   }
 
-  goToPosition(position: number) {
+  jumpToPosition(position: number) {
```
breville, line 47: `GOTO` did get a bit unpopular after BASIC :)

apps/src/music/views/AdvancedControls.tsx
```diff
@@ -0,0 +1,146 @@
...
+        </select>
+      </div>
+      <div className={classNames(moduleStyles.section)}>
+        <div className={moduleStyles.row}>
+          <div className={moduleStyles.rowLabel}>Loop</div>
+          <Checkbox
+            name="loop"
+            checked={loopEnabled}
+            onChange={event => onLoopChange(event.target.checked)}
+          />
```
breville, line 111: It would be nice if this component supported a white border, but the current dark border is alright for internal use for now.

breville (approved):
Great functionality and implementation.  It's fun to use!

PR #54374, opened 2023-10-20
LTI SSO init

nicklathe:
This feature adds SSO functionality for LTI 1.3 launches. It supports account creation and login for Teachers and Students coming to Code.org from an LMS via an LTI launch.

This feature has been built and tested using Canvas. Later iterations will verify/add support for other LMS platforms.

This screenshot shows the view of a teacher launching into Code.org for the first time. Note the name field is pre-filled, as we can get this information from the LTI id_token in most cases. We have also determined this user is a teacher from the id_token, and hide the user type selector.

Screenshot 2023-10-31 at 8 13 34 AM

This screenshot shows the view of a student launching into Code.org for the first time. The view is rendered appropriately for the student, based on the id_token, and has also hidden the user type selector.

Screenshot 2023-10-31 at 8 16 43 AM

## Testing story

```
bundle exec spring testunit ./test/lib/queries/lti_test.rb ./test/lib/policies/lti_test.rb ./test/lib/services/lti_test.rb
Running via Spring preloader in process 43862
Started with run options --seed 37837

  8/8: [=============================================================================================] 100% Time: 00:00:15, Time: 00:00:15

Finished in 20.09827s
8 tests, 17 assertions, 0 failures, 0 errors, 0 skips
```

## Follow-up work

There is followup work needed to add a field to capture a teacher's email, as we can't guarantee that we will receive this value in the id_token from the LMS. If it is present, we will pre-fill the form value as we do with `name`

[Jira link](https://codedotorg.atlassian.net/browse/P20-556)

## PR Checklist:

- [ ] Tests provide adequate coverage
- [ ] Privacy and Security impacts have been assessed
- [ ] Code is well-commented
- [ ] New features are translatable or updates will not break translations
- [ ] Relevant documentation has been added or updated
- [ ] User impact is well-understood and desirable
- [ ] Pull Request is labeled appropriately
- [ ] Follow-up work items (including potential tech debt) are tracked and linked

PR #57190, opened 2024-03-12
Manage each audience format differently

mgc1194:
### Story
When creating an LTI Authentication Option key, we interpolate values from the id_token. 
https://github.com/code-dot-org/code-dot-org/blob/febfc6dd49422c04c8ae78053db1142594b49479/dashboard/lib/policies/lti.rb#L79-L81

However, Schoology returns an id_token where the `aud` (audience) value is an array: 
`aud: ['some-client-id'] `

Due to the way the Auth ID generation is implemented, we grab the values and interpolate the string we are going to use for the Auth Option (with the brackets around the client-id):
`https://some-issuer.com|[\"some-client-id\"]|user-id `

Later in the LTI flow, when an LTI User Identity is created here:
https://github.com/code-dot-org/code-dot-org/blob/febfc6dd49422c04c8ae78053db1142594b49479/dashboard/lib/services/lti.rb#L30-L35

We recover the `issuer`, `client_id`, and `subject` from the `authentication_id` and query the existing LTI integrations to find one that matches `issuer` and `client_id`.

Since the recovered `client_id` is wrapped in brackets (`[client_id]`), the query won't find a record, returning `nil`

The `create_lti_user_identity` fails silently when creating an LTI User Identity using a `nil` value for lti_integration.

### Changes
- Refactor `generate_auth_id` to pull the string value out of the array before interpolating it.
- Created a new service (`Services::Lti::AuthIdGeneartor`) that generates the auth ID given a valid ID token.  
- Refactor [`get_user` query](https://github.com/code-dot-org/code-dot-org/blob/staging/dashboard/lib/queries/lti.rb#L5-L8) to rely on LtiUserIdentity table instead of `User.find_by_credential`

## Links

- jira ticket: [P20-768](https://codedotorg.atlassian.net/browse/P20-768)

## Follow-up work
- [P20-788: Create lti_user_identity at the same time as we create a new LTI user](https://codedotorg.atlassian.net/browse/P20-788)
- [P20-787: Remove Honeybadger Log in auth_id_generator](https://codedotorg.atlassian.net/browse/P20-787)
- [P20-789: Add Issuer column to the LtiUserIdentity table.](https://codedotorg.atlassian.net/browse/P20-789)

## PR Checklist:

- [ ] Code is well-commented


dashboard/lib/policies/lti.rb
```diff
@@ -77,7 +77,14 @@ def self.get_account_type(roles)
   end
 
   def self.generate_auth_id(id_token)
-    "#{id_token[:iss]}|#{id_token[:aud]}|#{id_token[:sub]}"
+    case id_token[:aud]
+    when String
+      "#{id_token[:iss]}|#{id_token[:aud]}|#{id_token[:sub]}"
+    when Array
+      "#{id_token[:iss]}|#{id_token[:aud].first}|#{id_token[:sub]}"
```
stephenliang: It might be helpful to add a log statement or metric if there were more than one element in the array. If we encountered a similar situation in Production having the log stating which audience we picked and what all the audiences were would be helpful.

mgc1194: I'm reviewing the LTI spec and see what additional information they have on the Audience claim.
I will need to understand if just one client ID is expected, or if multiple client IDs can be in that list.
If I get a clear answer, I'll handle the different cases. Otherwise, I agree with having an output that monitors and collects data on unexpected values.

dashboard/lib/queries/lti.rb
```diff
@@ -3,8 +3,7 @@
 
 class Queries::Lti
   def self.get_user(id_token)
-    auth_id = Policies::Lti.generate_auth_id(id_token)
-    User.find_by_credential(type: AuthenticationOption::LTI_V1, id: auth_id)
+    LtiUserIdentity.find_by(subject: id_token[:sub])&.user
```
mgc1194: ## Previous query:
```
 User.find_by_credential(type: AuthenticationOption::LTI_V1, id: "https:
//schoology.schoology.com|7193897412|130741586::d39d7c532aad97805a8450ae61c71f08")

=> #<User id: 14, studio_person_id: 5, email: "**********", parent_email: nil, created_at: "2024-03-11 19:54:31.000000000 +0000", updated_at: "2024-03-12 03:01:32.000000000 +0000", username: "mario417", provider: "migrated", uid: nil, admin: nil, gender: nil, name: "Mario Gil Correa", locale: "en-US", birthday: "2003-03-12", user_type: "teacher", school: nil, full_address: nil, school_info_id: nil, total_lines: 0, secret_picture_id: 2, active: true, hashed_email: "5077cbf96acb22180888ddd54dab6113", deleted_at: nil, purged_at: nil, secret_words: "test test", properties: {"lti_roster_sync_enabled"=>true}, terms_of_service_version: 1, urm: nil, races: nil, primary_contact_info_id: 14>
```
## Updated Query
```
LtiUserIdentity.find_by(subject: "130741586::d39d7c532aad97805a8450ae61c71f08")&.user

=> #<User id: 14, studio_person_id: 5, email: "**********", parent_email: nil, created_at: "2024-03-11 19:54:31.000000000 +0000", updated_at: "2024-03-12 03:01:32.000000000 +0000", username: "mario417", provider: "migrated", uid: nil, admin: nil, gender: nil, name: "Mario Gil Correa", locale: "en-US", birthday: "2003-03-12", user_type: "teacher", school: nil, full_address: nil, school_info_id: nil, total_lines: 0, secret_picture_id: 2, active: true, hashed_email: "5077cbf96acb22180888ddd54dab6113", deleted_at: nil, purged_at: nil, secret_words: "test test", properties: {"lti_roster_sync_enabled"=>true}, terms_of_service_version: 1, urm: nil, races: nil, primary_contact_info_id: 14>
```

dashboard/lib/services/lti.rb
```diff
@@ -24,6 +24,9 @@ def self.initialize_lti_user(id_token)
       email: get_claim(id_token, :email),
     )
     user.authentication_options = [ao]
+
+    # As final step of the LTI user creation, a LtiUserIdentity is created for the user.
+    create_lti_user_identity(user)
```
nicklathe: This won't work here, because `user` hasn't been created yet, it's only been "newed up", I.e. we haven't called `user.save`. That doesn't happen until the finish_signup is completed, and is handled in another controller.

If we do want to add the user identity here, we would need to do something similar what we are doing with the authentication options above, including adding [accepts_nested_attributes_for](https://github.com/code-dot-org/code-dot-org/blob/staging/dashboard/app/models/user.rb#L221) to `user.rb`.

nicklathe: I also think we'll need to add the nested lti_user_identies_attributes to the object we're returning here. I need to look deeper into this to remember how this all works https://github.com/code-dot-org/code-dot-org/pull/54374/files#diff-d00fda1c5d36968c46dc34d1f192623e04c3e415612db0570917dbefb2cde443R1

mgc1194: Cool, the idea was to have a visual of what the desired behaviour is. I'll look into this pointer you gave me.

dashboard/lib/services/lti.rb
```diff
@@ -63,6 +65,31 @@ def self.create_lti_deployment(integration_id, deployment_id)
     )
   end
 
+  def self.generate_auth_id(id_token)
```
artem-vavilov: It should be one operation per [Service Object](https://blog.trk.in.rs/2018/11/30/object-oriented-design-principles-and-patterns/#service-objects):
```ruby
module Services
  module Lti
    class AuthIdGenerator
      def initialize(id_token)
        @id_token = id_token
      end

      def call
        "#{iss}|#{aud}|#{sub}"
      end

      private

      attr_reader :id_token

      def iss
        id_token[:iss]
      end

      def aud
        ...
      end

      def sub
        id_token[:sub]
      end
    end
  end
end
```

dashboard/lib/policies/lti.rb
```diff
@@ -84,10 +84,6 @@ def self.lti_teacher?(roles)
     (Set.new(roles) & TEACHER_ROLES).any?
   end
 
-  def self.generate_auth_id(id_token)
-    "#{id_token[:iss]}|#{id_token[:aud]}|#{id_token[:sub]}"
-  end
-
```
mgc1194, line 90: I removed this method in favor of the service `Service::Lti::AuthIdGenerator`

dashboard/lib/queries/lti.rb
```diff
@@ -2,9 +2,8 @@
 require 'authentication_option'
 
 class Queries::Lti
-  def self.get_user(id_token)
-    auth_id = Policies::Lti.generate_auth_id(id_token)
-    User.find_by_credential(type: AuthenticationOption::LTI_V1, id: auth_id)
+  def self.get_user(user_id)
+    LtiUserIdentity.find_by(subject: user_id)&.user
```
mgc1194, line 6: we can find a user in the `LtiUserIdentity` table by its `user_id` so I removed the `id_token` as parameter in favor of just the `user_id`

dashboard/lib/services/lti.rb
```diff
@@ -4,231 +4,235 @@
...
-    user = User.new
-    user.provider = User::PROVIDER_MIGRATED
-    user.user_type = user_type
-    if user_type == User::TYPE_TEACHER
-      user.age = '21+'
-      user.name = get_claim_from_list(id_token, Policies::Lti::TEACHER_NAME_KEYS)
-      user.lti_roster_sync_enabled = true
-    else
-      user.name = get_claim_from_list(id_token, Policies::Lti::STUDENT_NAME_KEYS)
-      user.family_name = get_claim(id_token, :family_name)
```
mgc1194, line 19: Refactored `Services::Lti` into a module.
It makes more sense for this service to be a module since we are not creating any instance of `Service::Lti` anywhere in the code.

dashboard/test/controllers/lti_v1_controller_test.rb
```diff
@@ -477,11 +477,16 @@ def create_valid_jwt_raise_error
...
+    ao = AuthenticationOption.new(
+      user: user,
+      email: Services::Lti.get_claim(payload, :email),
+      credential_type: AuthenticationOption::LTI_V1,
+      authentication_id: Services::Lti::AuthIdGenerator.new(payload).call
+    )
     ao.save!
 
     deployment = LtiDeployment.create(deployment_id: @deployment_id, lti_integration_id: @integration.id)
+    create :lti_user_identity, user: user, subject: payload[:sub]
```
mgc1194, line 489: since we query now the `LtiUserIdentity` table we need to create the identity.

dashboard/test/lib/policies/lti_test.rb
```diff
@@ -4,7 +4,7 @@
 
 class Policies::LtiTest < ActiveSupport::TestCase
   setup do
-    @ids = ['http://some-iss.com', 'some-aud', 'some-sub'].freeze
+    @ids = ['http://some-iss.com', ['some-aud'], 'some-sub'].freeze
```
mgc1194, line 7: These changes are arbitrary, to better reflect the id_token content. Thus not necessary, I could roll them back.

dashboard/test/lib/policies/lti_test.rb
```diff
@@ -40,10 +40,6 @@ class Policies::LtiTest < ActiveSupport::TestCase
     assert_equal Policies::Lti.get_account_type(@id_token[Policies::Lti::LTI_ROLES_KEY]), User::TYPE_STUDENT
   end
 
-  test 'generate_auth_id should create authentication_id string' do
-    assert_equal Policies::Lti.generate_auth_id(@id_token), @ids.join('|')
-  end
-
```
mgc1194, line 46: Test removed since the method does not exists in `Policies::Lti` anymore.

dashboard/lib/services/lti.rb
mgc1194: I´m leaving a TODO note instead of including those changes in this PR. I consider an update in the user model complex enough to have it's own PR in case a revert is needed and for documentation purposes.

dashboard/lib/services/lti.rb
carl-codeorg: Does it make sense to move this into the lti directory as well?

dashboard/lib/services/lti/auth_id_generator.rb
```diff
@@ -0,0 +1,52 @@
+require 'cdo/honeybadger'
+
+module Services
+  module Lti
+    class AuthIdGenerator
```
carl-codeorg, line 5: What's the benefit of having this as a class vs a module? It seems to act like a function that takes a JWT and returns an ID, and in the code we instantiate a fresh `AuthIdGenerator` each time we generate an ID, so would it fit better as a `generate` function within a `Services::Lti::AuthIdGenerator`? There is probably some Ruby reason why it's supposed to be a class, but it isn't jumping out at me.

mgc1194: Your question makes total sense, I wondered something along the same lines too. 
This proposal is based on @artem-vavilov [suggestion](https://github.com/code-dot-org/code-dot-org/pull/57190#discussion_r1525711867), following the [Service object structure pattern](https://blog.trk.in.rs/2018/11/30/object-oriented-design-principles-and-patterns/#service-objects)

dashboard/lib/services/lti.rb
mgc1194: potentially, but many more changes will be needed to further refactor `Servce::Lti`. I wanted to keep the changes related to the bug as much as possible.

carl-codeorg: We do probably want to refactor it eventually, but would moving it break anything? Does the location of the file affect how it's referenced?

mgc1194: In theory, Rails autoloads all modules, so a location change should not affect it.
We could also specify the new `require` path: `require 'services/lti/some_lti_service'`.
We would have to update all the calls from `Services::Lti` to `Services::Lti::SomeLtiService`. this is the reason I did not want to add more changes to this PR.

cnbrenci (approved):
Plz add links to tasks for the todos! otherwise, LGTM! 🚀

dashboard/lib/services/lti/auth_id_generator.rb
artem-vavilov: Ruby is a purely object-oriented language (literally everything in Ruby is an object). While there is nothing inherently wrong with employing a functional programming style in Ruby, especially considering its potential for greater memory efficiency and simplicity, the object-oriented approach is generally preferred.

I believe that, due to the complexity introduced by the recent update, the `.generate_auth_id` business logic now necessitates encapsulation in its own class.

I recommended employing an object-oriented approach for the `Services::Lti::AuthIdGenerator` because it centralizes all operations related to generating the authentication ID within a single, coherent class instance. This encapsulation makes the code more intuitive and manageable. Additionally, it facilitates sharing the `id_token` across various operations within the class, enhancing data consistency and accessibility.

dashboard/lib/services/lti.rb
artem-vavilov: We don't need to (and defenelty don't have to) move all of the methods.
But I'd suggest moving those that are complex enough to benefit from encapsulation into their own class.

artem-vavilov: We could transform `Services::Lti` into a [Facade](https://refactoring.guru/design-patterns/facade). This way, it would serve as the interface for LTI-related functionality, while the business logic would be split among separate service objects:
```ruby
module Services
  module Lti
    def self.initialize_lti_user(id_token)
      Services::Lti::UserInitializer.new(id_token).call
    end
    
    def self.create_lti_user_identity(user)
      Services::Lti::UserIdentetyCreator.new(user).call
    end
    
    ...
  end
end
```

artem-vavilov (approved):
Great job! 👍

dashboard/lib/services/lti/auth_id_generator.rb
carl-codeorg: I definitely agree it makes sense to break the code down into manageable chunks. I'm not fully convinced for this particular class since it looks to have more boilerplate than actual code, but since this is the preferred pattern let's roll with it and see. I imagine in other cases, the benefit will be clearer.

nicklathe: I recently was trying to decide between a class and a module on another feature. ChatGPT says: 
* Use modules for grouping functions, mixins, and namespaces without needing instances.
* Use classes when you need to instantiate objects, model entities with specific attributes and behaviors, or establish an inheritance hierarchy.

In our case, all we need this functionality to do is take in an id_token, and return a string. Instantiating an object (class) for this seems like overkill to me.

dashboard/lib/services/lti.rb
```diff
@@ -4,231 +4,236 @@
...
-    issuer, client_id, subject = auth_option.authentication_id.split('|')
-    lti_integration = Queries::Lti.get_lti_integration(issuer, client_id)
-    LtiUserIdentity.create(user: user, subject: subject, lti_integration: lti_integration)
-  end
+    def self.create_lti_user_identity(user)
+      auth_option = user.authentication_options.find(&:lti?)
+      issuer, client_id, subject = auth_option.authentication_id.split('|')
+      lti_integration = Queries::Lti.get_lti_integration(issuer, client_id)
+      identity = LtiUserIdentity.new(user: user, subject: subject, lti_integration: lti_integration)
+      identity.save!
```
nicklathe, line 37: Why split this into two `.new` and `.save!`? For this case, wouldn't just using `.create!` be sufficient, since we aren't modifying the newed up `identity` object here?

nicklathe: Both `create!` and `save!` will raise a `ActiveRecord::RecordInvalid` exception if it's missing required values.

mgc1194: I did not touch this part of the code, I only added an indent.

dashboard/lib/services/lti/auth_id_generator.rb
artem-vavilov: All we need to do now, is:
1. Take an `id_token`.
2. Subtract the `audience` depend on its type.
3. Validate the `audience` for each type according to specific conditions.
4. And only then return a string.

In my humble opinion, implementing this logic within an instance makes the code more maintainable, understandable, and extensible for possible future changes.

I would also like to hear the opinions of @snickell and @Hamms on this.

snickell: Probably it doesn't matter either way. When in doubt, I usually go with "whoever wrote the code did it the right way". I _personally_ would have most likely written it as a single method on a module, but not because it would be better, that's just how **I** would have happened to do it.

My personal opinion: this is totally reasonable as-is, no need for too many of us to measure the code too closely, just go with it as it is written

1. I like how clearly call reads in this instance
2. I like how clearly issuer, audience and subject are broken into named bits
3. I like that id_token doesn't need to be piped around to each method, which is part of why call is so clear
4. I don't love the idea of invoking it as `AuthIdGenerator.new.call()`, but everything is a balance, and it doesn't bother me either.
5. If it was twice as much code, I'd say class for sure, if it was half as much code, I'd say module with a single method for sure, at this size I feel 🤷

artem-vavilov: We could invoke the service as `AuthIdGenerator.call(...)`: https://thiagolimadeveloper.medium.com/calling-services-object-without-instance-in-ruby-ca8b3f95fd10

dashboard/lib/services/lti.rb
nicklathe: The old code on the left on line 34 calls `.create`, which instantiates the new LtiUserIdentity and saves it in a single command. In this new code, on line 36 and 37, it has been changed to use `.new` and `.save!`. I was asking why we moved away from using `.create`. If we want to raise an exception for a missing value, we can just add the `!` to the end of `.create`.

dashboard/lib/services/lti/auth_id_generator.rb
Hamms: :tada: for Service Objects!

+1 to what Artem and Seth are saying; I agree it would probably work fine to implement this as a stateless module that manually passes the token around to whatever needs it, but I think this implementation does a good job of reflecting standard Ruby design principles. It might take a few more lines of code than it otherwise would, but I think those lines are being spent to improve readability rather than hinder it.

PR #57193, opened 2024-03-12
Music: standalone projects

breville:
This exposes standalone Music Lab projects to end-users for the first time.  

- The links from https://studio.code.org/incubator and from the final level in https://studio.code.org/s/music-intro-2024 now go to https://studio.code.org/projects/music/new.  This takes the user to a https://studio.code.org/projects/music/[channel-id]/edit URL which shows `Rename`, `Share`, `Remix` buttons in the header.
- The level backing these projects now uses the `"launch2024"` sound library
- The level backing these projects also has a new `"showSoundFilters": true` parameter which tells Music Lab to show the filters in the sound selection UI.
- When on a Music Lab standalone project page, the `Create` menu lists `Project Beats` as an option, which is existing behaviour.
- The `My Projects` tab at https://studio.code.org/projects will list Music Lab projects created by the user, but with a `Type` of `Project Beats`.

### DIFF
--- a/apps/src/music/globals.js
+++ b/apps/src/music/globals.js
@@ -23,7 +23,7 @@ export default {
     showSoundFilters = show;
   },
 
-  getShowSoundFilters(get) {
+  getShowSoundFilters() {
     return showSoundFilters;
   },
 };
--- a/apps/src/music/globals.js
+++ b/apps/src/music/globals.js
@@ -8,6 +8,7 @@
 // this data.
 
 let playerRef = null;
+let showSoundFilters = false;
 
 export default {
   setPlayer(player) {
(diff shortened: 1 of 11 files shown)


apps/src/music/globals.js
```diff
@@ -17,4 +18,12 @@ export default {
   getPlayer() {
     return playerRef;
   },
+
+  setShowSoundFilters(show) {
+    showSoundFilters = show;
+  },
+
+  getShowSoundFilters(get) {
```
sanchitmalhotra126: nit: remove `get` param?

sanchitmalhotra126 (approved):
Looks good!

thomasoniii (approved):
lgtm 🚀

PR #57195, opened 2024-03-12
[DESIGN2-95] Tabs component

levadadenys:
### [DESIGN2-95] Tabs component
* Created `Tabs` component
* created stories for `Tabs` component
* created tests for `Tabs` component
* created documentation for `Tabs` component

https://github.com/code-dot-org/code-dot-org/assets/22244040/d1e28a15-94e0-4a1a-a1ab-98956ad80a72

https://github.com/code-dot-org/code-dot-org/assets/22244040/a5b5397c-c693-4278-8f0c-ce34dec43a5d

https://github.com/code-dot-org/code-dot-org/assets/22244040/a4ba1a91-2d5e-4eac-a5ef-55f37e58ddb7

## Links
- jira ticket: [Design2-95](https://codedotorg.atlassian.net/browse/DESIGN2-95)

## PR Checklist:

- [ ] Code is well-commented


apps/src/componentLibrary/tabs/Tabs.tsx
```diff
@@ -0,0 +1,101 @@
...
+import _Tab, {TabModel} from '@cdo/apps/componentLibrary/tabs/_Tab';
+import TabPanel from '@cdo/apps/componentLibrary/tabs/TabPanel';
+
+import moduleStyles from './tabs.module.scss';
+
+export interface TabsProps {
+  /** Array of props for Tabs to render */
+  tabs: TabModel[];
+  /** The function that is called when a Tab is clicked */
+  onChange: (value: string) => void;
```
bencodeorg, line 14: Can we set a default behavior that clicking a tab will just open the associated tabpanel? I feel like as a consumer of this component, I wouldn't want to have to write that logic every time I use it? Same with making `selectedTabValue` an optional prop (`initialTabValue`?), and having this component manage what the currently selected value is.

apps/src/componentLibrary/tabs/Tabs.tsx
```diff
@@ -0,0 +1,101 @@
...
+        className={classnames(tabPanelsContainerClassName)}
+        id={tabPanelsContainerId}
+      >
+        {tabs.map(tab => (
+          <TabPanel
+            key={tab.value}
+            content={tab.tabContent}
+            isActive={tab.value === selectedTabValue}
+            id={`${name}-panel-${tab.value}`}
+            labelledBy={`${name}-tab-${tab.value}`}
```
bencodeorg: Does this work if `tab.value` has spaces?

apps/src/componentLibrary/tabs/_Tab.tsx
```diff
@@ -0,0 +1,97 @@
...
+  tabButtonId: string;
+}
+
+const checkTabForErrors = (
+  isIconOnly: boolean,
+  icon: FontAwesomeV6IconProps | undefined,
+  text: string | undefined
+) => {
+  if (isIconOnly && !icon) {
+    throw new Error('IconOnly tabs must have an icon');
```
bencodeorg, line 46: Is throwing an error too aggressive here? I saw we do this in some other components too, so maybe it's a pattern we're adopting.

apps/src/componentLibrary/tabs/Tabs.tsx
bencodeorg: I know I've made a similar comment on previous design system components, but I'd love to emphasize simplicity of use in these components where possible.

apps/src/componentLibrary/tabs/Tabs.tsx
```diff
@@ -0,0 +1,101 @@
...
+              onClick={onChange}
+              tabPanelId={`${name}-panel-${tab.value}`}
+              tabButtonId={`${name}-tab-${tab.value}`}
+            />
+          ))}
+        </ul>
+      </div>
+      <div
+        className={classnames(tabPanelsContainerClassName)}
+        id={tabPanelsContainerId}
```
bencodeorg, line 101: Is there an expected use for this prop?

bencodeorg (commented):
A few questions but this is exciting!

apps/src/componentLibrary/tabs/Tabs.tsx
levadadenys: Yep, for styles and or custom logic(if needed)

apps/src/componentLibrary/tabs/_Tab.tsx
levadadenys: Yep, this here is preventing us from rendering impossible combinations of icons/text. Considering it to be something ike typescript type check and have same weigth. (Failing types in TS casuses js erors)

apps/src/componentLibrary/tabs/Tabs.tsx
levadadenys: Good question. ---In the initial implementation there was a replacement of `_` with `-`, + showing the active tab content is handled by other prop. 
I can add something like const nameWithoutSpaces = name.replace(' ', '-');---

levadadenys: ohhh, I see now. I thought nameStriped simple replaces _ with - (see schreenshot)
Знімок екрана 2024-04-02 о 18 44 36

Now I understand the reason for that. Yeah, I can add this back. Though, the question is then if it'll also work if `name` will have spaces. Though, I'm not sure if there's a reason someone will be passing `name` and/or `value` with spaces. What do you think?

apps/src/componentLibrary/tabs/Tabs.tsx
levadadenys: The general idea behind DSCO is to keep components dumb/controlled, so that all the controlled is handled in higher level smart component. With previous components it was better to delegate controls to the container/higher level smart component. (Investigating usages of existing design system components we can see that we need the data from design system components on a higher level(form values, segmented buttons, etc))

Though, I agree that with Tabs it's rather 50/50, since rendering of the tab content is also handled inside of this component. I think we can use something like we use in RadioButtonGroup here:

```
  const handleChange = useCallback(
    (event: React.ChangeEvent<HTMLInputElement>) => {
      setSelectedValue(event.target.value);
      if (onChange) {
        onChange(event);
      }
    },
    [setSelectedValue, onChange]
  );

```

apps/src/componentLibrary/tabs/_Tab.tsx
bencodeorg: TS type checks aren't compiled though into production JS, whereas these errors are...I would expect the severity of these to be more like proptypes in old React components, which gave warnings, but wouldn't fully break the page if they occurred.

To be fair, I see we do this [in other places](https://github.com/code-dot-org/code-dot-org/blob/95f488ffef1fe252a86552c4b3586d7c8fa3f0ab/apps/src/javalab/components/JavalabDropdown.jsx#L15-L25) outside of the design system (I even contributed to this code!) But I do wonder if it's a pattern we should continue to use or not.

apps/src/componentLibrary/tabs/Tabs.tsx
levadadenys: Updated the logic, let me know what you think.
Also will appreciate more opinions on this one since not sure which option is better

bencodeorg: Thanks for explaining -- the update looks great!

apps/src/componentLibrary/tabs/Tabs.tsx
bencodeorg: Yeah I think the consumer _should_ be passing a name without spaces, but there's nothing protecting against it (so maybe just add `name.replace(' ', '-')`?

apps/src/componentLibrary/tabs/TabPanel.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+interface TabPanelProps {
+  content: React.ReactNode;
+  isActive: boolean;
+  id: string;
+  /* The ID of the button element that controls this panel's visibility. */
+  labelledBy: string;
+}
+
+// maybe need to move inline style
```
bencodeorg: Should this comment still be here?

bencodeorg (approved):
Nice!

apps/src/componentLibrary/tabs/Tabs.tsx
```diff
@@ -0,0 +1,113 @@
...
+export interface TabsProps {
+  /** Array of props for Tabs to render */
+  tabs: TabModel[];
+  /** The function that is called when a Tab is clicked/selected tab is changed */
+  onChange: (value: string) => void;
+  /** The value of the default selected Tab */
+  defaultSelectedTabValue: string;
+  /** The name attribute specifies the name of a Tabs group.
+     The name attribute is used to reference elements in a JavaScript.
+     */
```
sanchitmalhotra126, line 19: nit: comment formatting

apps/src/componentLibrary/tabs/TabPanel.tsx
```diff
@@ -0,0 +1,23 @@
...
+  isActive,
+  id,
+  labelledBy,
+}) => (
+  <div role="tabpanel" id={id} aria-labelledby={labelledBy} hidden={!isActive}>
+    {content}
+  </div>
+);
+
+export default TabPanel;
```
sanchitmalhotra126: nit: should this also be `_TabPanel` as it's meant to be internal only?

apps/src/componentLibrary/tabs/Tabs.tsx
sanchitmalhotra126: Agree that it's convenient to have the default behavior baked in, but I'm also thinking, should we enforce that consumers only use one or the other? I'm thinking of cases where consumers might provide an onChange handler that updates the selected tab value some other way that conflicts with the internal handler, leading to unexpected behavior. Could we make it so that if an onChange handler is provided, the consumer must also handle the tab switching manually? I could see some potential usefulness of that behavior if a consumer wants to update some data/save state in redux for example. Curious what y'all think

Also, thoughts on renaming `onChange` -> `onTabSelected` or similar?

apps/src/componentLibrary/tabs/_Tab.tsx
```diff
@@ -0,0 +1,97 @@
...
+import React, {useCallback} from 'react';
+
+import FontAwesomeV6Icon, {
+  FontAwesomeV6IconProps,
+} from '@cdo/apps/componentLibrary/fontAwesomeV6Icon';
+import moduleStyles from '@cdo/apps/componentLibrary/tabs/tabs.module.scss';
+
+export interface TabModel {
+  /** Unique value of the tab */
+  value: string;
```
sanchitmalhotra126, line 12: would `key` be better here, since this is meant to uniquely identify the tab? I also feel like it would make more sense to see a change handler that looks like `onTabSelected: (tabKey: string) => void` vs `onChange: (value: string) => void`

apps/src/componentLibrary/tabs/_Tab.tsx
```diff
@@ -0,0 +1,97 @@
...
+  /** Tab text */
+  text?: string;
+  /** Icon left from label */
+  iconLeft?: FontAwesomeV6IconProps;
+  /** Icon right from label */
+  iconRight?: FontAwesomeV6IconProps;
+  /** Whether button should be icon only */
+  isIconOnly?: boolean;
+  /** Tab icon */
+  icon?: FontAwesomeV6IconProps;
```
sanchitmalhotra126, line 22: Not to go back in circles but I'm thinking if there's a way through TS or otherwise we could share this behavior with the `Button` component, because the logic is very similar (can only have icon or text, and if text then iconLeft/Right is valid, otherwise not)...maybe some internal `<IconTextField/>` that handles the props and throws errors internally? Definitely not a requirement for this PR, but just calling it out as it came to mind

apps/src/componentLibrary/tabs/tabs.module.scss
```diff
@@ -0,0 +1,320 @@
...
+    padding: 0;
+
+    li {
+      list-style: none;
+      margin: 0;
+      padding: 0;
+    }
+  }
+
+  button {
```
sanchitmalhotra126, line 19: Out of curiosity, why not use class names here?

apps/src/componentLibrary/tabs/Tabs.tsx
levadadenys: That's how our linter/prettier formats comment

apps/src/componentLibrary/tabs/tabs.module.scss
levadadenys: Not using classNames here since it's not required.(html structure of this component is pretty basic) Keeping css selectors priority as low as possible in order make things simple.

apps/src/componentLibrary/tabs/_Tab.tsx
levadadenys: Agree, might worth investigating in the future. Should be ok to do in a separate PR since it should affect any consumer-facing logic

apps/src/componentLibrary/tabs/Tabs.tsx
levadadenys: > Agree that it's convenient to have the default behavior baked in, but I'm also thinking, should we enforce that consumers only use one or the other? I'm thinking of cases where consumers might provide an onChange handler that updates the selected tab value some other way that conflicts with the internal handler, leading to unexpected behavior. Could we make it so that if an onChange handler is provided, the consumer must also handle the tab switching manually? I could see some potential usefulness of that behavior if a consumer wants to update some data/save state in redux for example. Curious what y'all think

I would say let's merge as, discuss it more and them make a separate follow up PR for this if needed. (Since it won't break the component logic and some teams already need this component it to be merged)

apps/src/componentLibrary/tabs/_Tab.tsx
levadadenys: Won't it be to 'repeating'? (passing `onTabSelected` into `Tabs` component with `TabModel`) I thought we decided not to use repitative names for props/variables to make it easier to read and more unified/standardized naming.
Also, `value` is also meant to be unique and it's using here in the same way it's used in other deisgn system components. I feel like `key` might be to confusing react's `key` prop and `tabKey` might be to 'repeating' name

sanchitmalhotra126 (approved):
LGTM!

PR #57198, opened 2024-03-12
Music: fix missing condition value in intro script validation

sanchitmalhotra126:
Quick fix for the validations editor on level 2 of the intro script. I noticed that the "value" field wasn't displaying for the `played_sound_id` validation condition, which was because in the validation configuration, `hasValue` was set to `false`. I update this to just remove the `hasValue` field and infer if we should display the value based on the presence of the `valueType` field. This way we'll always show the value field if a value type has been specified.

Before:
Screenshot 2024-03-12 at 11 48 19 AM

After:
Screenshot 2024-03-12 at 11 47 42 AM

## Testing story

Tested locally.


thomasoniii (approved):
lgtm 🚀

breville (approved):

PR #57203, opened 2024-03-12
Add privacy link in demographic pop-up

juanmanzojr:
Hyperlinks the privacy policy page to “will never rent, sell or share this information with any third parties” in dempgraphic pop-up.
Screenshot 2024-03-12 at 12 23 40 PM

## Links
- jira ticket: [here](https://codedotorg.atlassian.net/browse/P20-776)

## Testing story
Local

### DIFF
--- a/dashboard/app/views/layouts/_student_information_interstitial.html.haml
+++ b/dashboard/app/views/layouts/_student_information_interstitial.html.haml
@@ -6,7 +6,7 @@
       %p{style: 'font-size: 13px; line-height: 18px; color: #333'}
         = t('parent_mailer.parent_email_added_subject')
         = t('activerecord.attributes.user.finish_sign_up_header_account_information')
-        = t('age_interstitial.disclaimer')
+        = t('age_interstitial.disclaimer', url: CDO.code_org_url('/privacy'), markdown: :inline).html_safe
       = form_for(current_user, url: registration_url('user'), html: {id: 'edit_user'}) do |f|
         .form-group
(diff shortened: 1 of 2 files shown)


dashboard/app/views/layouts/_student_information_interstitial.html.haml
```diff
@@ -6,7 +6,7 @@
       %p{style: 'font-size: 13px; line-height: 18px; color: #333'}
         = t('parent_mailer.parent_email_added_subject')
         = t('activerecord.attributes.user.finish_sign_up_header_account_information')
-        = t('age_interstitial.disclaimer')
+        = t('age_interstitial.disclaimer', url: CDO.code_org_url('/privacy'), markdown: :inline).html_safe
```
carl-codeorg: This privacy URL is available [in shared_constants.rb](https://github.com/code-dot-org/code-dot-org/blob/staging/lib/cdo/shared_constants.rb#L635). If we use this constant, then if the url ever changes we only have to change it in one place. Seems like overkill but for CPA we had to iterate these URLs many times.

carl-codeorg (changes requested):

dashboard/app/views/layouts/_student_information_interstitial.html.haml
juanmanzojr: Ill make this change thanks!

carl-codeorg (approved):
:shipit:

PR #57204, opened 2024-03-12
Update Redis gem to 4.8.1

Hamms:
Specifically, update to the latest minor version of the next major version. With this change we will still be one major version behind the latest version; we could instead attempt to skip all the way to 5.1.0, but given that we have relatively few tests exercising Redis functionality I'm opting for the safer, slower approach. Open to feedback if anyone thinks we should be more aggressive here!

The only breaking change I can see in 4.0 is removal of official support for Ruby < 2.2.2, and removal of a few aliases that we don't appear to be using.

## Links

- [changelog](https://github.com/redis/redis-rb/blob/master/CHANGELOG.md#40)

## Testing story

Relying on existing tests to verify that this does not result in any change to functionality. Also verified on [an adhoc](https://adhoc-redis-4-8-1-studio.cdn-code.org/courses) that this does not break in any obvious way.

## Follow-up work

After this is deployed and we have a chance to verify in production that everything is still working as expected, we should plan to at least attempt an upgrade all the way to 5.x

## PR Checklist:

- [ ] Code is well-commented


breville (approved):
Thanks for doing this!  I don't have additional insight into how to know this is safe, though I recall that Redis is only used by NetSim?  But this seems like a thoughtful and cautious change.

sureshc:
NetSim gets utilized primarily at the beginning (Fall) of the US school year, so we may not find out about breaking changes for 6 months. Should we ask the team(s) that own NetSim to add more tests that exercise Redis?

PR #57205, opened 2024-03-12
Persistent icon key expansion state

kobryan0619:
This uses `localStorage` so that the iconKey state is persistent.

https://github.com/code-dot-org/code-dot-org/assets/24883357/a6740286-be12-484d-be2e-4a5f9d744f8d

## Links

- [Ticket](https://codedotorg.atlassian.net/jira/software/projects/TEACH/boards/65?selectedIssue=TEACH-898)


apps/test/unit/templates/sectionProgressV2/IconKeyTest.jsx
```diff
@@ -38,6 +29,15 @@ describe('IconKey Component', () => {
     expect(screen.getByText('Validated')).to.exist;
   });
 
+  it('expands collapses on click', () => {
```
kobryan0619: Note: I moved this test down in order to be able to test what was in the IconKey before we collapse it.  There were a few other options I considered:

1. Stubbing out `tryGetLocalStorage` in each test so that it sets the state to what we want before each test. 

2. Running this "collapse" test and then ending the test with it opening again to resume tests in an open state.  Probably accompanied by a comment about why we are doing it this way.

It is possible that this rearranging here is the best solution but might benefit from a comment explaining why these tests come first.  

Thoughts?

dju90 (approved):
So great that you were able to get to this stretch goal of persisting the state! Two quick questions - is it open by default when the user first sees the new progress view, and were you able to test it out with a page refresh?

kobryan0619:
> So great that you were able to get to this stretch goal of persisting the state! Two quick questions - is it open by default when the user first sees the new progress view, and were you able to test it out with a page refresh?

Yep! Open by default when a user first sees it and yes I tested it with a page refresh.

apps/test/unit/templates/sectionProgressV2/IconKeyTest.jsx
lfryemason: I am a big fan of making tests completely independent of order, especially because test order is not guaranteed. I think both of your other options would work great though. 

Do we need a test that it will start collapsed if the setting is false?

kobryan0619: Sounds good - I modified it based on these thoughts and did add a test to check when it starts collapsed.

apps/test/unit/templates/sectionProgressV2/IconKeyTest.jsx
```diff
@@ -1,41 +1,60 @@
...
+    sinon.stub(utils, 'tryGetLocalStorage').returns('false');
+    render(<IconKey isViewingValidatedLevel={false} expandedLessonIds={[]} />);
+    expect(screen.queryByText('Assignment Completion States')).to.be.null;
+    expect(screen.queryByText('Teacher Actions')).to.be.null;
+    expect(screen.queryByText('Level Types')).to.be.null;
+    utils.tryGetLocalStorage.restore();
   });
 
   it('expands collapses on click', () => {
+    sinon.stub(utils, 'tryGetLocalStorage').returns('true');
```
lfryemason, line 30: Optional nit: You can move `utils.tryGetLocalStorage.restore()` to an `afterEach()` method.

And you could also try `sinon.stub(utils, 'tryGetLocalStorage').returns('true');` in a beforeEach() and restub with false where necessary. Both optional

apps/test/unit/templates/sectionProgressV2/IconKeyTest.jsx
```diff
@@ -1,41 +1,60 @@
...
 import IconKey from '@cdo/apps/templates/sectionProgressV2/IconKey';
 
 describe('IconKey Component', () => {
   it('renders the open state initially', () => {
+    sinon.stub(utils, 'tryGetLocalStorage').returns('true');
     render(<IconKey isViewingValidatedLevel={false} expandedLessonIds={[]} />);
     expect(screen.getByLabelText('Icon key')).be.visible;
     expect(screen.queryByText('More Details')).be.visible;
     expect(screen.getByText('Assignment Completion States')).to.exist;
     expect(screen.getByText('Teacher Actions')).to.exist;
```
lfryemason:
```suggestion
    screen.getByText('Teacher Actions');
```
nit: We don't need the `expect` if we are just checking existence of a single item via a `getBy` query. This will also print out a much better error without the expect. I'd recommend removing all `expects().to.exist` in the file

lfryemason (approved):
Some non-blocking nits, otherwise very excited

apps/test/unit/templates/sectionProgressV2/IconKeyTest.jsx
kobryan0619: Yeah, I did consider both of these... good to know it is optional in your mind too.  It seemed more annoying to restub and we didn't need to do `utils.tryGetLocalStorage.restore()` after every test so I decided to list it out... I think I will keep it as is for now.  I was torn on it myself too.

apps/test/unit/templates/sectionProgressV2/IconKeyTest.jsx
kobryan0619: Thanks for this context!  I made this change!

PR #57206, opened 2024-03-12
Updates AI TA release to JSON + Evidence column

wilkie:
Just moves the release to the one that is using both JSON and the Evidence column.

Would:
* Enable JSON responses in `aiproxy`
* Enables evidence support in all lessons in Unit 3
* Enable improvements to chain-of-thought reasoning, by moving the reason column to before the Grade column

## Links

- jira ticket: [AITT-430](https://codedotorg.atlassian.net/browse/AITT-430)

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/dashboard/app/jobs/evaluate_rubric_job.rb
+++ b/dashboard/app/jobs/evaluate_rubric_job.rb
@@ -11,7 +11,7 @@ class EvaluateRubricJob < ApplicationJob
   #
   # Basic validation of the new AI config is done by UI tests, or can be done locally
   # by running `EvaluateRubricJob.new.validate_ai_config` from the rails console.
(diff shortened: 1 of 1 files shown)


davidsbailey (approved):
woohoo, excited to see this all ship!

just for completeness, this release also effectively launches:
3. enable improvements to chain-of-thought reasoning, by moving the reason column to before the Grade column

I looked through a diff of the new release vs 2024-02-28-ai-rubrics-json-reason and just noticed one small issue, which is that some of the mentions of "field" (more appropriate for json) have gone back to being called "column" (more appropriate for tsv/csv/markdown table response). I do not think this is a blocker by any means given the detailed accuracy report you've collected, but, could be an interesting future experiment to see if this wording affects accuracy.

PR #57207, opened 2024-03-12
[Sprite Lab] Criterion commands for new CSC module lesson 1

mikeharv:
The CSC Game Development module's first lesson is ready for validation. To support this, we need to add 3 new commands related to behavior criteria. A couple other commands are updated to support a new optional minimum parameter.

## Links

Validation Notes Doc: https://docs.google.com/document/d/175wWAuwkDDBDxG31Ed57E7bo5gzACGQ9ZcAHSgk54Xs/edit
Jira (lessons 1-2): https://codedotorg.atlassian.net/browse/CT-405

## Testing story

I tested this by cloning the last level in Katie's progression and adding validation criteria for each condition listed for the entire lesson according to the document above. I tested that level success or failure feedback was triggered as expected.

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/p5lab/spritelab/commands/criterionCommands.js
+++ b/apps/src/p5lab/spritelab/commands/criterionCommands.js
@@ -44,7 +44,7 @@ export const commands = {
     );
   },
 
-  // Return true if any sprite began speaking
+  // Return true if a minimum number of sprites began speaking
   // and the text is not an empty string.
   strictAnySpriteSpeaks(min = 1) {
     const spriteIds = this.getSpriteIdsInUse();
@@ -55,7 +55,7 @@ export const commands = {
     );
   },
 
-  // Return true if any sprite was speaking.
+  // Return true if a minimum number of sprites was speaking.
   anySpriteSpeaking(min = 1) {
     const spriteIds = this.getSpriteIdsInUse();
     return (
@@ -64,7 +64,7 @@ export const commands = {
     );
   },
 
-  // Return true if any sprite was speaking
+  // Return true if a minimum number of sprites was speaking
   // and the text is not an empty string.
   strictAnySpriteSpeaking(min = 1) {
     const spriteIds = this.getSpriteIdsInUse();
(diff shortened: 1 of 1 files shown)


apps/src/p5lab/spritelab/commands/criterionCommands.js
```diff
@@ -46,30 +46,31 @@ export const commands = {
 
   // Return true if any sprite began speaking
```
molly-moen: nit: update comment for this and the methods below to reference the min variable

apps/src/p5lab/spritelab/commands/criterionCommands.js
```diff
@@ -125,10 +126,26 @@ export const commands = {
   },
 
   // Returns true if some minimum number of sprites are in use.
-  minimumSprites(min) {
+  minimumSprites(min = 1) {
     return this.getSpriteIdsInUse().length >= min;
   },
 
+  // Returns true if some minimum number of sprites are in use.
```
molly-moen: update comment to include location requirement?

molly-moen (approved):
Looks good, just a couple comment suggestions

apps/src/p5lab/spritelab/commands/criterionCommands.js
```diff
@@ -351,6 +368,51 @@ export const commands = {
...
+      totalBehaviors += currentBehaviors.length;
+    }
+    return totalBehaviors >= min;
+  },
+
+  // Returns true if any sprite has the specified behavior.
+  minimumMatchingBehaviors(matchingBehavior, min = 1) {
+    const spriteIds = this.getSpriteIdsInUse();
+    let matchingBehaviors = 0;
+    for (let i = 0; i < spriteIds.length; i++) {
```
ebeastlake: Are we guaranteed that the sprite IDs are 0-indexed and contiguous? What if we delete one? Would using a `for...of` or a `forEach` as above make sense?

ebeastlake:
One question about sprite id management, but otherwise LGTM!

apps/src/p5lab/spritelab/commands/criterionCommands.js
mikeharv: Great point. We have other commands that seem to have been working (ie. no bug reports) that use similar logic, so at worst it's an edge case worth fixing. We could keep the loop as is but change `this.getBehaviorsForSpriteId(i)` to `this.getBehaviorsForSpriteId(spriteIds[i])`, but I think your suggestions will be more readable (and more modern). I'll update the new commands now and then start on a follow-up branch to fix any old commands that might have this same potential bug.

PR #57211, opened 2024-03-12
Lab2: begin moving validation out of level data and into level properties

sanchitmalhotra126:
The primary motivation behind this change is to make validation localization easier; instead of digging into level data to retrieve strings during sync-in, and reconstructing the level data object with translated strings when sending out level data, we can work with the validations property directly. This also simplifies code a bit, since we were previously plucking validations out of level properties and merging it back into the level data JSON which we no longer need to do.

This should be backwards compatible; we'll look for validations on level properties first, but fall back to the level data validations if they're not present in level properties. When levels with validations are edited/saved, we'll keep validations on level properties, and just delete them from level data.

Local diff for a level that was opened and saved without any manual edits - validations are moved to level properties and deleted from level data.

Screenshot 2024-03-12 at 2 25 58 PM

## Links

https://codedotorg.atlassian.net/browse/LABS-535

## Testing story

Tested locally with the intro script, with a mix of levels that have validation in level properties and level data.

### DIFF
--- a/apps/src/lab2/progress/ProgressContainer.tsx
+++ b/apps/src/lab2/progress/ProgressContainer.tsx
@@ -4,10 +4,10 @@ import {
   ProgressLevelType,
 } from '@cdo/apps/code-studio/progressReduxSelectors';
 import ProgressManager from '@cdo/apps/lab2/progress/ProgressManager';
-import {useAppDispatch} from '@cdo/apps/util/reduxHooks';
+import {useAppDispatch, useAppSelector} from '@cdo/apps/util/reduxHooks';
 import React, {useCallback, useEffect, useRef} from 'react';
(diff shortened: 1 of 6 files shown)


breville (approved):
Looks good.

Is there a broader conversation to be had around what, if anything, stays in `level_data` in the long term?

sanchitmalhotra126:
> Is there a broader conversation to be had around what, if anything, stays in `level_data` in the long term?

Yeah, definitely something we can discuss. I can see some of the other music level data fields moving into properties (like toolbox and startSources), but might still be worth having a generic JSON field that levels can use so we're not overloading the number of fields in level properties? Then again, maybe that's fine. In any case worth, worth discussing down the road for sure!

PR #57212, opened 2024-03-12
feat(lms): do not sync or create sections if on the LMS, the user is not an instructor

stephenliang:
Previously, the sync_course logic will operate regardless of the user's role on the LMS. This resulted in some cases where synchronization occurred when not a teacher (for example, an administrator inadvertently synchronizing a course, or a student creating a course). When a non-instructor creates a course, the section becomes "owned" by the non-instructor potentially locking out the actual instructor.

## Links

- [jira](https://codedotorg.atlassian.net/browse/P20-759)

## Testing story

1. I tested creating a new course in Canvas, and never visiting Code.org.
2. I added a student in this course and had the student visit Code.org

Previous: The student becomes the section owner
Now: The result was that the course did not get created and the student goes to the home page.

## PR Checklist:

- [x] Code is well-commented

### DIFF
--- a/dashboard/lib/policies/lti.rb
+++ b/dashboard/lib/policies/lti.rb
@@ -16,12 +16,12 @@ class Policies::Lti
   JWT_ISSUER = CDO.studio_url('', CDO.default_scheme).freeze
 
   MEMBERSHIP_CONTAINER_CONTENT_TYPE = 'application/vnd.ims.lti-nrps.v2.membershipcontainer+json'.freeze
-  TEACHER_ONLY_ROLES = Set.new(['http://purl.imsglobal.org/vocab/lis/v1/institution/person#Instructor',
-                                'http://purl.imsglobal.org/vocab/lis/v2/membership#Instructor']
+  TEACHER_ROLES = Set.new(['http://purl.imsglobal.org/vocab/lis/v1/institution/person#Instructor',
+                           'http://purl.imsglobal.org/vocab/lis/v2/membership#Instructor']
 ).freeze
-  TEACHER_ROLES = Set.new(
+  STAFF_ROLES = Set.new(
     [
-      *TEACHER_ONLY_ROLES,
+      *TEACHER_ROLES,
       'http://purl.imsglobal.org/vocab/lis/v2/institution/person#Administrator',
       'http://purl.imsglobal.org/vocab/lis/v2/membership#Administrator',
       'http://purl.imsglobal.org/vocab/lis/v2/system/person#Administrator',
@@ -74,17 +74,14 @@ class Policies::Lti
 
   def self.get_account_type(roles)
     roles.each do |role|
-      return User::TYPE_TEACHER if TEACHER_ROLES.include? role
+      return User::TYPE_TEACHER if STAFF_ROLES.include? role
     end
     return User::TYPE_STUDENT
   end
 
   # Returns true if any of the user's roles is the LTI instructor role
   def self.lti_teacher?(roles)
-    Array(roles).each do |role|
-      return true if TEACHER_ONLY_ROLES.include? role
-    end
-    return false
+    (Set.new(roles) & TEACHER_ROLES).any?
   end
 
   def self.generate_auth_id(id_token)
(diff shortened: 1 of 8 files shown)


dashboard/lib/policies/lti.rb
```diff
@@ -14,13 +14,16 @@ module AccessTokenScopes
   NAMESPACE = 'lti_v1_controller'.freeze
   JWT_CLIENT_ASSERTION_TYPE = 'urn:ietf:params:oauth:client-assertion-type:jwt-bearer'.freeze
   JWT_ISSUER = CDO.studio_url('', CDO.default_scheme).freeze
+
   MEMBERSHIP_CONTAINER_CONTENT_TYPE = 'application/vnd.ims.lti-nrps.v2.membershipcontainer+json'.freeze
+  TEACHER_ONLY_ROLES = Set.new(['http://purl.imsglobal.org/vocab/lis/v1/institution/person#Instructor',
```
carl-codeorg: I feel like our naming isn't the best for `TEACHER_ROLES` (which includes admin staff roles) and `TEACHER_ONLY_ROLES`. Is there a better set of names we can use? Maybe `INSTRUCTOR_ROLES` instead of `TEACHER_ONLY_ROLES`. Nothing good is coming to mind for the combo of teacher and admin roles though...

dashboard/lib/policies/lti.rb
```diff
@@ -76,6 +79,14 @@ def self.get_account_type(roles)
     return User::TYPE_STUDENT
   end
 
+  # Returns true if any of the user's roles is the LTI instructor role
+  def self.lti_teacher?(roles)
+    Array(roles).each do |role|
+      return true if TEACHER_ONLY_ROLES.include? role
+    end
+    return false
```
carl-codeorg: There's a shorthand for intersections in Ruby that might make this function slightly more concise
```suggestion
    (Set.new(roles) & TEACHER_ONLY_ROLES).any?
```

dashboard/lib/policies/lti.rb
```diff
@@ -101,6 +112,13 @@ def self.issuer_name(issuer)
     I18n.t(:lti_v1, scope: [:section, :type])
   end
 
+  # Returns the LTI user id for a particular code.org user and LTI integration
+  def self.lti_user_id(user, lti_integration)
```
carl-codeorg: This function might fit better in [queries/lti.rb](https://github.com/code-dot-org/code-dot-org/blob/staging/dashboard/lib/queries/lti.rb) since it's just retrieving data.

dashboard/lib/policies/lti.rb
```diff
@@ -101,6 +112,13 @@ def self.issuer_name(issuer)
...
+    lti_user_identities = user.lti_user_identities
+
+    lti_user_identities.find {|identity|  identity.lti_integration_id == lti_integration.id}.try(:subject)
```
carl-codeorg: Instead of `find` does `find_by` work, and keep it a little cleaner? I'm certainly no ActiveRecord expert so I'm not sure if it works or not:
```suggestion
    user.lti_user_identities.find_by(lti_integration_id: lti_integration.id)&.subject
```

dashboard/lib/services/lti.rb
```diff
@@ -179,13 +198,20 @@ def self.sync_course_roster(lti_integration:, lti_course:, nrps_sections:, secti
     end
 
     nrps_sections.keys.each do |lms_section_id|
-      section_name = nrps_sections[lms_section_id][:name]
+      nrps_section = nrps_sections[lms_section_id]
```
carl-codeorg: Now that we're using the value as well as the key as we iterate, is there a better function to use than `keys.each`? Like some kind of object function where we can get both the key and value in the params of the iterator function.

dashboard/lib/policies/lti.rb
stephenliang: Renamed `TEACHER_ROLES` to `STAFF_ROLES` and `TEACHER_ONLY_ROLES` to `TEACHER_ROLES` to clarify

dashboard/test/lib/services/lti_test.rb
```diff
@@ -374,38 +375,82 @@ class Services::LtiTest < ActiveSupport::TestCase
...
 
     # Remove empty sections with no users
     @nrps_full_response[:members] = []
     parsed_response = Services::Lti.parse_nrps_response(@nrps_full_response, @id_token[:iss])
-    Services::Lti.sync_course_roster(lti_integration: lti_integration, lti_course: lti_course, nrps_sections: parsed_response, section_owner_id: teacher.id)
+    Services::Lti.sync_course_roster(lti_integration: lti_integration, lti_course: lti_course, nrps_sections: parsed_response, current_user: teacher)
     assert lti_course.reload.sections.length, 2
   end
+
+  test 'should NOT sync if the user is not a teacher on LTI' do
```
carl-codeorg: Minor nitpick: should be not a teacher in the _LMS_

carl-codeorg (approved):
LGTM!

PR #57215, opened 2024-03-12
pin http gem to version 5 or greater

davidsbailey:
## short version

upgrades an old version of the `http` gem, and solves a developer build problem.

## long version

when the `statsig` gem was introduced in https://github.com/code-dot-org/code-dot-org/pull/56879, it broke `bundle install` for me locally:
Screenshot 2024-03-12 at 3 24 12 PM

I was not able to find a workaround to install the http-parser gem directly, so I found this way to avoid needing this gem instead.

statsig depends on http-parser as follows:
```
In Gemfile:
  statsig was resolved to 1.33.0, which depends on
    http was resolved to 4.4.1, which depends on
      http-parser
```

the above version of http is relatively old compared to other versions of http which statsig accepts: https://github.com/statsig-io/ruby-sdk/blob/d6db370a148defa6e71f5e090f2c33920e84feb7/statsig.gemspec#L30

starting in version 5.0, the `http` gem stops depending on the problem gem `http-parser`: https://github.com/httprb/http/blob/main/CHANGES_OLD.md#500-2021-05-12

pinning http to version 5 in the Gemfile causes http-parser to no longer be installed and fixes my local build, while also upgrading an old gem, so seems like a win-win.

## Testing story

relying on existing test coverage


Gemfile.lock
```diff
@@ -722,7 +728,7 @@ GEM
       thor (~> 1.0)
     rainbow (3.1.1)
     raindrops (0.20.0)
-    rake (12.3.3)
+    rake (13.1.0)
```
Hamms, line 731: Is this change intentional? I glanced at rake's changelog and it looks like a pretty safe change for a major version update, but rake is a sufficiently-fundamental dependency that it seems worth asking.

Hamms (commented):
One question about a possibly-unrelated dependency update, but the overall plan here sounds good!

Gemfile.lock
davidsbailey: yes, it turns out to be intentional! `llhttp-ffi` depends on it: https://github.com/bryanp/llhttp/blob/b4cf901ef8b4a8946fb6d2b18f6624c6e2708107/ffi/llhttp-ffi.gemspec#L25

I'm spinning up an adhoc now to see if that could catch some problems. I'm also open to shipping the rake upgrade ahead of this one.

Hamms: Ahh, I missed that. Spinning up an adhoc sounds great! I think given that we don't have a direct top-level dependency on rake in the Gemfile, it's probably fine to include the rake upgrade as part of this rather than splitting it out into its own PR.

nicklathe:
I'm on an M1 MBP, running Rosetta. I pulled this branch and ran `bundle install`. However, when I started `bin/dashboard-server`, I got the following error:

```
Could not open library '/Users/nicklathe/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/llhttp-ffi-0.5.0/ext/aarch64-darwin/libllhttp-ext.bundle': dlopen(/Users/nicklathe/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/llhttp-ffi-0.5.0/ext/aarch64-darwin/libllhttp-ext.bundle, 0x0005): tried: '/Users/nicklathe/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/llhttp-ffi-0.5.0/ext/aarch64-darwin/libllhttp-ext.bundle' (mach-o file, but is an incompatible architecture (have 'arm64', need 'x86_64'))
```

I tried un-installing and re-installing the gem `ffi`, which were suggested [here](https://github.com/bryanp/llhttp/issues/20), but this did not fix the error. I tried un/re-installing a couple other gems, thinking maybe that could fix things, however I still got the same error. 

Finally, I tried following [this SO answer](https://stackoverflow.com/a/78060414/4557426), which fixed my problem.

nicklathe (approved):
I was able to get this working on my machine. I left a comment explaining the error I was getting and the fix that worked. Hopefully that will help anyone else that might have a similar error.

Gemfile.lock
davidsbailey: adhoc is up: https://adhoc-pin-http-5-for-statsig.cdn-code.org/students

PR #57220, opened 2024-03-12
modify strings

kobryan0619:
As per a DM from Bryan, there are two strings changed here...

image

## Task

image

image


dju90:
No objections from T&L - go ahead and merge!

dju90 (approved):

PR #57224, opened 2024-03-13
Music: share dialog UI

breville:
Initial work on a new project share dialog UI for Music Lab.  

This leaves the current share dialog UI intact, but does some refactoring to enable some component reuse, and adds a "V2" variant which we can use for Music Lab and other labs if desired.

This initial version for Music Lab just has a "copy to clipboard" button, and a bigger QR code.

Screenshot 2024-03-14 at 10 50 18 AM

### DIFF
--- a/apps/src/code-studio/components/ShareAllowedDialog.jsx
+++ b/apps/src/code-studio/components/ShareAllowedDialog.jsx
@@ -323,6 +323,271 @@ class ShareAllowedDialog extends React.Component {
 
     const warningText = this.getWarningText(showPublishInfo);
 
+    const interfaceVersion = appType === 'music' ? 2 : 1;
+
+    const renderAbuseError = () => {
+      return (
+        <AbuseError
+          i18n={{
+            tos: i18n.tosLong({url: 'http://code.org/tos'}),
+            contact_us: i18n.contactUs({
+              url: `https://support.code.org/hc/en-us/requests/new?&description=${encodeURIComponent(
+                `Abuse error for project at url: ${shareUrl}`
+              )}`,
+            }),
+          }}
+          className="alert-error"
+          style={styles.abuseStyle}
+          textStyle={styles.abuseTextStyle}
+        />
+      );
+    };
+
+    const renderSharingDisabled = () => {
+      return (
+        <div style={{position: 'relative'}}>
+          <div style={{paddingRight: 10}}>
+            <p>{i18n.sharingBlockedByTeacher()}</p>
+          </div>
+          <div style={{clear: 'both', height: 40}}>
+            <button
+              type="button"
+              id="continue-button"
+              style={{position: 'absolute', right: 0, bottom: 0, margin: 0}}
+              onClick={this.close}
+            >
+              {i18n.dialogOK()}
+            </button>
+          </div>
+        </div>
+      );
+    };
+
+    const renderCopyButton = () => {
+      return (
+        <Button
+          color={Button.ButtonColor.brandSecondaryDefault}
+          id="sharing-dialog-copy-button"
+          icon="clipboard"
+          style={{
+            ...styles.button,
+            ...styles.copyButton,
+            ...(this.state.hasBeenCopied && styles.copyButtonLight),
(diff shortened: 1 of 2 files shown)


apps/src/code-studio/components/ShareAllowedDialog.jsx
```diff
@@ -323,227 +323,270 @@ class ShareAllowedDialog extends React.Component {
...
+            'SHARING_LINK_COPY',
+            this.props.appType
+          )}
+          text={i18n.copyLinkToProject()}
+          value={shareUrl}
+        />
+      );
+    };
+
+    const renderInterfaceV1 = () => {
```
thomasoniii, line 388: Not blocking, but could these just be a components? 

Maybe tweaked like this:

```
const AbuseErrorWrapper = ({ isAbusive }) => {
  if (!isAbusive) { return null}
  // return same as above
}

// and then later this:
{ isAbusive && renderAbuseError() }
// turns into this
<AbuseErrorWrapper {isAbusive={isAbusive} />
```

An argument could be made about handing in the `isAbusive` prop vs continuing to wrap it in the conditional. I'm also not completely convinced of what a good name for that wrapper component would be.

But in general, we should skew towards components instead of functions that return components.

Same argument and style for the other `render` functions. Obviously I wrote this up after just seeing the first one. 😇

apps/src/code-studio/components/ShareAllowedDialog.jsx
```diff
@@ -323,227 +323,270 @@ class ShareAllowedDialog extends React.Component {
...
+                {this.state.isTwitterAvailable && (
+                  <a
+                    href={twitterShareUrl}
+                    target="_blank"
+                    rel="noopener noreferrer"
                     onClick={wrapShareClick(
-                      this.showSendToPhone,
-                      'SHARING_LINK_SEND_TO_PHONE',
+                      onClickPopup.bind(this),
+                      'SHARING_TWITTER',
```
thomasoniii, line 498: Is there an actual `SHARING_TWITTER` constant that we can import in from somewhere instead of adding it as a string?

thomasoniii (approved):
couple of styling suggestions around making components instead of functions, but otherwise lgtm 🚀

apps/src/code-studio/components/ShareAllowedDialog.jsx
```diff
@@ -323,227 +323,270 @@ class ShareAllowedDialog extends React.Component {
...
+                embedOptions={embedOptions}
+                appType={this.props.appType}
+              />
+            )}
+          </div>
+        </div>
+      );
+    };
+
+    const renderInterfaceV2 = () => {
```
sanchitmalhotra126, line 555: Similar to Jim's comment, but thoughts on creating a separate component for this? And additionally, should we just handle the branching logic in `Lab2ShareDialog.tsx`? We can bypass rendering `ShareAllowedDialog` entirely then, since this share view is quite different and I'd expect it will continue to deviate from the existing share UI.

apps/src/code-studio/components/ShareAllowedDialog.jsx
breville: You're right.  I was trying to make minimal changes to the legacy code, but have built a revised approach which uses functional components.

apps/src/code-studio/components/ShareAllowedDialog.jsx
breville: You're right.  I had been concerned about supporting all of the variants of the legacy dialog (as can be seen in the Storybook [here](https://github.com/code-dot-org/code-dot-org/blob/5285a67de0eedb3d14b5bdf115c1b18c0a542619/apps/src/code-studio/components/ShareAllowedDialog.story.jsx)), though the more I looked, the more it seemed that probably only `isAbusive` was a likely candidate.  I have built an alternative approach in https://github.com/code-dot-org/code-dot-org/pull/57333 which features a new functional React component, and doesn't yet support any of those variants, though it certainly could as we learn more.

PR #57232, opened 2024-03-13
[Sprite Lab] refactor criterion commands

mikeharv:
While reviewing some new criterion commands recently @ebeastlake identified an edge case bug in the way I was iterating through sprite ids:
> Are we guaranteed that the sprite IDs are 0-indexed and contiguous? What if we delete one? Would using a `for...of` or a `forEach` as above make sense?
https://github.com/code-dot-org/code-dot-org/pull/57207#discussion_r1522499202

Unfortunately, this same edge case existed in a number of edge cases as well. This PR performs the following clean-up tasks on these functions:

* Ensures that we are iterating through the actual sprite ids that are in use, and not just the indices of the array containing the ids.
* Modernize some traditional loops to be more concise and readable.
* Eliminate some `result` variables aren't really needed. We can often return the value directly instead of assigning it to `result`, or in some cases we can actually improve efficiency by returning early.

## Testing Story
I've checked for regressions by putting the significantly refactored commands into a level's validation code and ensuring that we still pass or fail as expected.

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/p5lab/spritelab/commands/criterionCommands.js
+++ b/apps/src/p5lab/spritelab/commands/criterionCommands.js
@@ -80,21 +80,17 @@ export const commands = {
   // current or previous frame.
   anySpeechIncludesValues(currentVariables, previousVariables) {
     const spriteIds = this.getSpriteIdsInUse();
-    const values = Object.values(currentVariables);
-    if (previousVariables) {
-      values.push(...Object.values(previousVariables));
-    }
+    const values = Object.values(currentVariables).concat(
+      previousVariables ? Object.values(previousVariables) : []
+    );
 
     for (const spriteId of spriteIds) {
       const speechText = this.getLastSpeechBubbleForSpriteId(spriteId)?.text;
-      if (speechText) {
-        for (const value of values) {
-          if (typeof speechText === 'string' && speechText.includes(value)) {
-            return true;
-          } else if (typeof speechText === 'number' && speechText === value) {
-            return true;
-          }
-        }
+      if (
+        speechText &&
+        values.some(value => `${speechText}`.includes(`${value}`))
+      ) {
+        return true;
       }
     }
     return false;
@@ -231,7 +227,6 @@ export const commands = {
     const spriteIds = this.getSpriteIdsInUse();
     const eventSpriteIds = validationCommands.getEventSpriteIds.call(this);
     let foundEventSpriteChange = false;
-    let foundNoneventSpriteChange = false;
 
     for (const spriteId of spriteIds) {
       const currentCostume = this.nativeSpriteMap[spriteId].getAnimationLabel();
@@ -243,16 +238,13 @@ export const commands = {
         if (eventSpriteIds.includes(spriteId)) {
           foundEventSpriteChange = true;
         } else {
-          foundNoneventSpriteChange = true;
-        }
-
-        if (foundEventSpriteChange && foundNoneventSpriteChange) {
-          return false; // Found both event and non-event sprite changes, no need to continue
+          // Fail immediately if a non-event sprite changed.
+          return false;
         }
       }
     }
 
-    return foundEventSpriteChange && !foundNoneventSpriteChange;
+    return foundEventSpriteChange;
   },
 
   anyPropChanged(prop) {
@@ -291,7 +283,6 @@ export const commands = {
     const spriteIds = this.getSpriteIdsInUse();
     const eventSpriteIds = validationCommands.getEventSpriteIds.call(this);
     let foundEventSpriteChange = false;
-    let foundNoneventSpriteChange = false;
 
     for (const spriteId of spriteIds) {
       let currentProp;
@@ -319,16 +310,13 @@ export const commands = {
         if (eventSpriteIds.includes(spriteId)) {
           foundEventSpriteChange = true;
         } else {
-          foundNoneventSpriteChange = true;
-        }
-
-        if (foundEventSpriteChange && foundNoneventSpriteChange) {
-          return false; // Found both event and non-event sprite changes, no need to continue
+          // Fail immediately if a non-event sprite changed.
+          return false;
         }
       }
     }
 
-    return foundEventSpriteChange && !foundNoneventSpriteChange;
+    return foundEventSpriteChange;
   },
 
   // Returns true if any sprite changed (started or stopped) behaviors this frame.
@@ -415,7 +403,6 @@ export const commands = {
     const eventSpriteIds = validationCommands.getEventSpriteIds.call(this);
 
     let foundEventSpriteChange = false;
-    let foundNoneventSpriteChange = false;
 
     for (const spriteId of spriteIds) {
       const currentBehaviors = this.getBehaviorsForSpriteId(spriteId);
@@ -435,16 +422,13 @@ export const commands = {
         if (isEventSprite) {
           foundEventSpriteChange = true;
         } else {
-          foundNoneventSpriteChange = true;
+          // Fail immediately if a non-event sprite changed.
+          return false;
         }
-      }
-
-      if (foundEventSpriteChange && foundNoneventSpriteChange) {
-        return false; // Found both event and non-event sprite changes, no need to continue
       }
     }
 
-    return foundEventSpriteChange && !foundNoneventSpriteChange;
+    return foundEventSpriteChange;
   },
 
   // Returns true if all sprites have the default size (100 for students)
--- a/apps/src/p5lab/spritelab/commands/criterionCommands.js
+++ b/apps/src/p5lab/spritelab/commands/criterionCommands.js
@@ -80,49 +80,40 @@ export const commands = {
   // current or previous frame.
   anySpeechIncludesValues(currentVariables, previousVariables) {
     const spriteIds = this.getSpriteIdsInUse();
-    let result = false;
-    const values = previousVariables
-      ? Object.values(currentVariables).concat(Object.values(previousVariables))
-      : Object.values(currentVariables);
-    for (let i = 0; i < spriteIds.length; i++) {
-      values.forEach(value => {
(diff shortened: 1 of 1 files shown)


apps/src/p5lab/spritelab/commands/criterionCommands.js
```diff
@@ -80,49 +80,40 @@ export const commands = {
...
+    const values = Object.values(currentVariables);
+    if (previousVariables) {
+      values.push(...Object.values(previousVariables));
+    }
+
+    for (const spriteId of spriteIds) {
+      const speechText = this.getLastSpeechBubbleForSpriteId(spriteId)?.text;
+      if (speechText) {
+        for (const value of values) {
+          if (typeof value === 'string' && speechText.includes(value)) {
```
molly-moen: previously we were checking the type of `speechText` but now we are testing the type of `value`, is this intentional?

apps/src/p5lab/spritelab/commands/criterionCommands.js
```diff
@@ -232,141 +212,141 @@ export const commands = {
...
+        }
       }
     }
-    result = foundEventSpriteChange && !foundNoneventSpriteChange;
-    return result;
+
+    return foundEventSpriteChange && !foundNoneventSpriteChange;
   },
 
-  // Returns true if any-sprite specified-property changes this frame.
```
molly-moen, line 280: why was this comment removed?

apps/src/p5lab/spritelab/commands/criterionCommands.js
```diff
@@ -419,31 +399,8 @@ export const commands = {
...
-      ) {
-        result = true;
-      }
-    }
-    return result;
-  },
-
-  // Special function for lesson: Mini-Project: Collector Game
-  // Returns true if any sprite has one of the "non-interactive" behaviors.
-  itemBehaviorFound() {
```
molly-moen, line 442: why could this function be removed?

molly-moen (commented):
Nice refactor! A couple minor questions

apps/src/p5lab/spritelab/commands/criterionCommands.js
mikeharv: Oh! Meant to leave a comment explaining this. It's removed for two reasons:
* it's an exact copy of `interactiveBehaviorFound`
* it's not used in any levels

apps/src/p5lab/spritelab/commands/criterionCommands.js
mikeharv: Accidental!

apps/src/p5lab/spritelab/commands/criterionCommands.js
mikeharv: Great catch! This is a mistake on my part. It's probably more important to check the type of `speechText` since we're assuming it has an `.includes()` method. Usually these types will match (and they did when I tested it), but it's possible they won't. Will fix!

apps/src/p5lab/spritelab/commands/criterionCommands.js
```diff
@@ -80,49 +80,40 @@ export const commands = {
...
-          case 'number':
-            if (speechText === value) {
-              result = true;
-            }
-            break;
-          default:
-            break;
+    const values = Object.values(currentVariables);
+    if (previousVariables) {
+      values.push(...Object.values(previousVariables));
```
ebeastlake: I find this use of push slightly unusual, because it's pushing the variables into the first array as individual elements. For a small array like this, it doesn't matter, but if you're ever dealing with really large arrays, I think there might be performance issues with using `.push()` like this vs. the original `.concat()`. 

You could also do:
```
const values = [
  ...Object.values(currentVariables),
  ...(previousVariables ? Object.values(previousVariables) : [])
];
```

or

```
const values = Object.values(currentVariables).concat(previousVariables ? Object.values(previousVariables) : []);
```

apps/src/p5lab/spritelab/commands/criterionCommands.js
```diff
@@ -80,49 +80,40 @@ export const commands = {
...
+    }
+
+    for (const spriteId of spriteIds) {
+      const speechText = this.getLastSpeechBubbleForSpriteId(spriteId)?.text;
+      if (speechText) {
+        for (const value of values) {
+          if (typeof speechText === 'string' && speechText.includes(value)) {
+            return true;
+          } else if (typeof speechText === 'number' && speechText === value) {
```
ebeastlake: If you're looking for this to return true in the following condition
```
Return true if any sprite's speech include values from the given object(s)
```
then why is this checking `speechText === value` for numerical values? Actually, I guess my real question is, why is this called `speechText` if it's sometimes numerical?

I'm wondering if we could do something like...
```
for (const spriteId of spriteIds) {
  const speechText = this.getLastSpeechBubbleForSpriteId(spriteId)?.text;
  if (speechText && values.some(value => 
    `${speechText}`.includes(`${value}`)
  )) {
    return true;
  }
}
return false;
```

apps/src/p5lab/spritelab/commands/criterionCommands.js
```diff
@@ -232,141 +212,141 @@ export const commands = {
...
-      } else if (currentCostume !== previousCostume) {
-        foundNoneventSpriteChange = true;
+        this.previous.sprites?.find(sprite => sprite.id === spriteId)
+          ?.costume ?? currentCostume;
+
+      if (currentCostume !== previousCostume) {
+        if (eventSpriteIds.includes(spriteId)) {
+          foundEventSpriteChange = true;
+        } else {
+          foundNoneventSpriteChange = true;
```
ebeastlake: Couldn't we return `false` here? That would mean a non-event sprite changed, so the condition fails immediately?

apps/src/p5lab/spritelab/commands/criterionCommands.js
```diff
@@ -232,141 +212,141 @@ export const commands = {
...
+
+        if (foundEventSpriteChange && foundNoneventSpriteChange) {
+          return false; // Found both event and non-event sprite changes, no need to continue
+        }
       }
     }
-    result = foundEventSpriteChange && !foundNoneventSpriteChange;
-    return result;
+
+    return foundEventSpriteChange && !foundNoneventSpriteChange;
```
ebeastlake: And then here, I think you could just
```
return foundEventSpriteChange;
```

But I think I've got too sucked into reading these validation functions and this might be too granular 😅

apps/src/p5lab/spritelab/commands/criterionCommands.js
```diff
@@ -232,141 +212,141 @@ export const commands = {
...
```
ebeastlake: Similar feedback here as above, but maybe not worth it.

apps/src/p5lab/spritelab/commands/criterionCommands.js
```diff
@@ -453,64 +402,78 @@ export const commands = {
...
+        // use the current behaviors.
+        this.previous.sprites?.find(sprite => sprite.id === spriteId)
+          ?.behaviors ?? currentBehaviors;
+
+      const behaviorsChanged = !utils.arrayEquals(
+        currentBehaviors,
+        previousBehaviors
+      );
+
+      if (behaviorsChanged) {
```
ebeastlake, line 420: Feel free to leave this nit: I don't necessarily think you need to store these as named conditions. 

```
if (!utils.arrayEquals(currentBehaviors, previousBehaviors)) {
      if (eventSpriteIds.includes(spriteId)) {
        foundEventSpriteChange = true;
      } else {
        foundNoneventSpriteChange = true;
      }
```

ebeastlake:
Only one blocking comment I'm curious about: https://github.com/code-dot-org/code-dot-org/pull/57232#discussion_r1524258712

Nice work! 🎉

apps/src/p5lab/spritelab/commands/criterionCommands.js
mikeharv: I'll switch to `.concat`. I wasn't too worried about performance because I wouldn't expect most students to have more than a few variables.

apps/src/p5lab/spritelab/commands/criterionCommands.js
mikeharv: Agreed, this feels cleaner.

apps/src/p5lab/spritelab/commands/criterionCommands.js
mikeharv: Yes, good call.

apps/src/p5lab/spritelab/commands/criterionCommands.js
mikeharv: I was going for readability in this case. Will leave if you don't feel strongly opinionated!

molly-moen (approved):
Nice!

apps/src/p5lab/spritelab/commands/criterionCommands.js
ebeastlake: I don't!

ebeastlake (approved):
Thanks for doing this, @mikeharv!

apps/src/p5lab/spritelab/commands/criterionCommands.js
```diff
@@ -80,49 +80,36 @@ export const commands = {
...
-        }
-      });
+    const values = Object.values(currentVariables).concat(
+      previousVariables ? Object.values(previousVariables) : []
+    );
+
+    for (const spriteId of spriteIds) {
+      const speechText = this.getLastSpeechBubbleForSpriteId(spriteId)?.text;
+      if (
+        speechText &&
```
mikeharv, line 90: This line introduced a regression as we should expect and accept some falsey values for `speechText`. It needs to be updated to `speechText !== undefined`.

PR #57233, opened 2024-03-13
Add strings for updated code.org/applab page and shared sections

kelbyhawn:
Adds strings for the updated App Lab page on code.org.

**Jira tickets:**
- This PR: [ACQ-1619](https://codedotorg.atlassian.net/browse/ACQ-1619)
- Parent ticket: [ACQ-1479](https://codedotorg.atlassian.net/browse/ACQ-1479)


pegasus/sites.v3/hourofcode.com/i18n/en.yml
```diff
@@ -1756,6 +1756,110 @@
       heading: "Don't see your country?"
       desc: "While we currently do not actively recruit new partners into our professional learning program, we encourage organizations worldwide to join us in organizing the annual Hour of Code campaign as the first step to introducing computer science in your country. [Learn more about a partnership opportunity and apply here!](%{apply_url})"
 
+  tools_pages_shared:
```
kelbyhawn, line 1759: These strings will be used in views that can be shared across the three new tools pages.

bethanyaconnor: This may not be worth it, but I'm wondering if these could be combined? For example:

```
tools_pages:
  shared:
    glance:
      titles:
        ages: "Ages"
        level: "Level"
        make: "What you can make"
        devices: "Devices"
        browsers: "Browsers"
        languages: "Languages supported"
        accessiblity: "Accessibility"
      button: "View documentation"
    featured_projects:
      heading: "Featured student projects"
      desc: "Explore more featured projects built by students like you in our featured project gallery."
  applab_page:
    heading: "App Lab"
    subheading: "Create an app with Javascript"
```

It's only worth it if it would help the international team group all of the translations for these pages though.

bethanyaconnor (approved):
Looks good! I added one thought/wonder, but it's definitely not blocking and up to you whether you think it's a good idea. :)

pegasus/sites.v3/hourofcode.com/i18n/en.yml
kelbyhawn: Oh interesting! My only hesitation is that I think it'll make all of these strings extra long in the markup. 

@mgc1194 do you have any thoughts on whether this pattern would be beneficial?

kelbyhawn: I thought about this more, and I'm going to keep what I had originally. There will be three more similar pages added, and I worry it'll get too nesty and confusing!

PR #57234, opened 2024-03-13
Add mailjet gem

bethanyaconnor:
Adds the `mailjet` gem to our application ahead of using their API. This gem is maintained by MailJet and its repo is [here](https://github.com/mailjet/mailjet-gem).

The changes here are from adding `mailjet` to `Gemfile` and running `bundle install`.


cat5inthecradle:
Looks like [1.7.8 is available](https://rubygems.org/gems/mailjet), any reason not to use that?

Likewise, any reason not to pin to 1.x or 1.7.x and take the opportunity now to ensure we've got the test coverage to feel comfortable with that?

Gemfile
```diff
@@ -356,3 +356,5 @@ gem "delayed_job_active_record", "~> 4.1"
 gem 'rack-cors', '~> 2.0.1'
 
 gem 'statsig', '~> 1.33'
+
+gem 'mailjet', '~> 1.7.3'
```
unlox775-code-dot-org, line 364: What is the reason for v1.7.3?  No strong preference, just curious.
In the [diff for 1.7.8](https://github.com/mailjet/mailjet-gem/compare/v1.7.3...v1.7.8) it says 1.7.8 is compatible with Rails >= 5

cat5inthecradle: jinx!

bethanyaconnor: Oh 1.7.3 was the default when running `bundle install` without specifying a version 🤷‍♀️ I'll update this and, barring any issues, pin to the later version

bethanyaconnor: Ah this is why:

```
Bundler could not find compatible versions for gem "rest-client":
  In Gemfile:
    rest-client (~> 2.0.1)

    mailjet (~> 1.7.8) was resolved to 1.7.8, which depends on
      rest-client (>= 2.1.0)
```

`rest-client` was last bumped [5 years ago](https://github.com/code-dot-org/code-dot-org/commit/b5403b67d80f886fe3216e6b8e34ec47a210b104). 2.1.0 is the latest release for `rest-client` -- also 5 years ago -- and nothing looks particularly scary in the [release notes](https://github.com/rest-client/rest-client/blob/master/history.md#210). I upgraded both in f2f0a10109a; if Drone fails or there are concerns with bumping `rest-client` here, I can revert that commit.

bethanyaconnor: Well, upgrading rest-client totally broke pegasus 🙃 I reverted that commit

bethanyaconnor: Okay, more info -- `rest-client` was a complete red herring. The issue is actually version 1.7.8 of mailjet, which uses `yajl-ruby`. Unfortunately, that somehow replaces `JSON` in the entire application (see https://github.com/mailjet/mailjet-gem/issues/255), which is, in turn, breaking pegasus.

This is a screenshot of the error page:
Screenshot 2024-03-14 at 2 56 14 PM

All this to say, I vote to keep us on 1.7.3 for `mailjet`. If that bug is fixed, I'm happy to bump the version.

unlox775-code-dot-org (approved):

cat5inthecradle (approved):

PR #57235, opened 2024-03-13
[Google Blockly] Retry failed code generation

molly-moen:
Since we've been logging google blockly code generation failures, we've theorized there is a set of errors that go away automatically after `getWorkspaceCode` is called again. We theorized this because we can't reproduce them and google blockly has event handlers that run after the workspace is set up that may not be complete by the time `getWorkspaceCode` is called.

In order to validate this theory and to hopefully clean up our logging to be real errors only, I updated `getWorkspaceCode` to retry getting code twice with a 500 ms interval between each retry. Only if all requests fail do we then log the error. I was able to validate that this fixes one known project that has a load failure until you press run or change anything. 

In the last week we saw 277 code generation errors. If this is successful this number should go down drastically.

## Testing story
Tested locally; in a normal case we never hit this code path. I also manually threw errors and confirmed only 2 retries occurred. In addition, I verified that the known failure case project is fixed by this issue. [slack thread](https://codedotorg.slack.com/archives/C05DK21DAHK/p1710351427916039)

## Follow-up work
After this is deployed and has been out for a day or so we should check if the error rate decreased, and then investigate the remaining errors.

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/blockly/googleBlocklyWrapper.ts
+++ b/apps/src/blockly/googleBlocklyWrapper.ts
@@ -205,6 +205,8 @@ function initializeBlocklyWrapper(blocklyInstance: GoogleBlocklyInstance) {
       getStore().dispatch(setFailedToGenerateCode(false));
     } catch (e) {
       if (retryCount < MAX_GET_CODE_RETRIES) {
+        // Sometimes we need to wait for Google Blockly change handlers to complete
+        // before the code will generate correctly. Retry after a short delay.
         setTimeout(() => {
           return getWorkspaceCodeHelper(retryCount + 1, hiddenWorkspace);
         }, RETRY_GET_CODE_INTERVAL_MS);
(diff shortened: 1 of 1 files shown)


apps/src/blockly/googleBlocklyWrapper.ts
```diff
@@ -185,22 +187,35 @@ function initializeBlocklyWrapper(blocklyInstance: GoogleBlocklyInstance) {
 
   blocklyWrapper.loopHighlight = function () {}; // TODO
   blocklyWrapper.getWorkspaceCode = function () {
+    return getWorkspaceCodeHelper(0, this.getHiddenDefinitionWorkspace());
```
mikeharv, line 190: I could also see us passing in `MAX_GET_CODE_RETRIES` here and decrementing it. Then `retryCount` could be optional since we'd only need to try again if it was > 0.

I think this is functionally equivalent and definitely not a blocker. Just sharing how my brain processed it a little differently.

mikeharv (approved):
Awesome!

apps/src/blockly/googleBlocklyWrapper.ts
molly-moen: yep that would also work! I think I'm going to leave it as-is for now.

PR #57240, opened 2024-03-13
[Sprite Lab] Criterion commands for new CSC module lesson 2

mikeharv:
The CSC Game Development module's second lesson is ready for validation. To support this, we need to add new commands related to foreground effects and variables. This work also required:
* adding the initial render frame to an effects object (so that we could validate that an effect started in conjunction with an event)
* updating our "previous" state to track variable bubbles and their values. We need this information so that we have an efficient way to identify student variables of interest and to check for changed values.

## Links

Validation Notes Doc: https://docs.google.com/document/d/175wWAuwkDDBDxG31Ed57E7bo5gzACGQ9ZcAHSgk54Xs/edit
Jira (lessons 1-2): https://codedotorg.atlassian.net/browse/CT-405
Jira (variable bubbles): https://codedotorg.atlassian.net/browse/CT-407

## Testing story

I tested this by adding temporary validation to several levels in Katie's progression according to the document above. I tested that level success or failure feedback was triggered as expected.

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/p5lab/spritelab/commands/criterionCommands.js
+++ b/apps/src/p5lab/spritelab/commands/criterionCommands.js
@@ -618,6 +618,10 @@ export const commands = {
   // Optionally checks if the variable values have increased or decreased.
   // Since this command requires a variable bubble, it should be used in conjunction
   // with variableBubblesCreated.
+  // Example usage:
+  //   variableValueChanged(); // Checks if any watched variable changed.
+  //   variableValueChanged('increase'); // Checks if any watched variable increased.
+  //   variableValueChanged('decrease'); // Checks if any watched variable decreased.
   variableValueChanged(changeType = '') {
     const previousStudentVars = this.previous.variableBubbles;
     if (previousStudentVars) {
@@ -668,7 +672,7 @@ export const commands = {
         }
       }
     }
-    // If we reach this point, none of the watched variables changed
+    // If we reach this point, none of the watched variables changed in the specified way.
     return false;
   },
 };
--- a/apps/src/p5lab/spritelab/commands/criterionCommands.js
+++ b/apps/src/p5lab/spritelab/commands/criterionCommands.js
@@ -583,9 +583,92 @@ export const commands = {
     return result;
   },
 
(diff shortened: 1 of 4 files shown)


apps/src/p5lab/spritelab/commands/criterionCommands.js
```diff
@@ -583,9 +583,92 @@ export const commands = {
...
+      studentVar => this.getVariableValue(studentVar.name) !== ''
+    );
+    return filteredVars.length >= min;
+  },
+
+  // Returns true if a watched variable's value changed this frame.
+  // Optionally checks if the variable values have increased or decreased.
+  // Since this command requires a variable bubble, it should be used in conjunction
+  // with variableBubblesCreated.
+  variableValueChanged(changeType = '') {
```
molly-moen, line 625: is it worth documenting (in the comment above?) what the valid values for `changeType` are? I'm not sure how we generally document these functions

apps/src/p5lab/spritelab/commands/criterionCommands.js
```diff
@@ -583,9 +583,92 @@ export const commands = {
...
+                console.error(
+                  `Invalid argument for variableValueChanged: ${changeType}`
+                );
+                break;
+            }
+          }
+        }
+      }
+    }
+    // If we reach this point, none of the watched variables changed
```
molly-moen: comment nit: they also could have changed, but not the way we are testing for, right?

molly-moen (approved):
Looks good, couple comment suggestions

apps/src/p5lab/spritelab/commands/criterionCommands.js
mikeharv: Yeah I think that’s worth adding. Good idea! As far documentation goes, right now it’s really just the comments in this file. I’d love to eventually publish something easier to read for levelbuilder.

apps/src/p5lab/spritelab/commands/criterionCommands.js
mikeharv: That’s fair. I can clarify.

PR #57243, opened 2024-03-13
INF-979 Prefactor docker code to prepare for additional dockerfiles and images

cat5inthecradle:
Our existing dockerfile builds an image suitable for Drone tests and perhaps local development, but it's not great for production workloads. In preparation for adding additional Docker options, this PR removes the presumption that this is the only or primary dockerfile, making room for additional dockerfiles.

For example, I am working on a 'minimal' or 'dependencies only' Dockerfile that contains only system dependencies for building and running the rails application. It will not contain developer tools, a mysql server, or testing dependencies like browser drivers.

Two major things are happening here:
* renaming "docker/dockerfiles" to "docker/drone" to isolate the Dockerfile and entrypoint script that we use for drone. 
* renaming the image we push to Docker Hub from `codedotorg/code-dot-org` to `codedotorg/cdo-drone`. Chosing this name vs just `drone` because it's an image of the `code-dot-org` application, `cdo` for brevity, prepared specially for executing in drone. Future images might be `cdo-minimal` and `cdo-activejob` or `cdo-dashboard`.

## Follow-up work

After merging this, a new image should be pushed to Docker Hub with the new naming convention. After this is pushed, we can create a new Pull Request that updates the docker compose files and the drone configuration to use the new image. Then we can complete all the 'TODO' items in this PR.

## PR Checklist:

- [ ] Code is well-commented


unlox775-code-dot-org (approved):

docker/README.md
```diff
@@ -85,14 +85,18 @@ Build a full image to your local machine
 docker build .
 ```
 
+**Note:** We are moving to a new naming convention in order to use different docker images for different purposes. What was the `code-dot-org` image will become the `cdo-drone` image. For now, these docs advise publishing both names.
+
 Tag the new image with an incremented version number. You can get the image id from the end of the build log or from `docker image list`.
 
 ```
 docker tag <image_id> codedotorg/code-dot-org:<version>
+docker tag <image_id> codedotorg/cdo-drone:<version>
```
cat5inthecradle: We don't define what are versioning scheme is. I'd suggest a simple incrementing integer, or a date based scheme.

wilkie (commented):
Yes please. :)

wilkie (approved):

sureshc:
Given how many lines of code we have that still reference Circle CI, I think I'd prefer that we name things `continuous-integration` instead of `drone` :)

cat5inthecradle:
@wilkie I'll merge this after drone passes in prep for #60274

PR #57247, opened 2024-03-13
Add an eslint rule to order JS import statements

lfryemason:
Groups and orders all JS import statements.

Ordering:

```
// builtin and external imports
import React from 'react';
import _ from 'lodash';

// '@cdo' and other internal imports
import Test from '@cdo/apps/test/test.js';

// parent folder imports
import Parent from '../parent/parent.js'

// sibling folder imports
import Sibling from './sibling.js'
import Sibling2 from './sibling2/sibling2.js'

// "index" of the current directory (should be rare)
import main from './';

// scss files
import styles from './styles.module.scss'
import moreStyles from '../moreStyles/more-styles.module.scss'

// misc others
```

Before:
image

After:
image

Currently only turned on for `apps/src/templates/sectionProgressV2` and `apps/test/unit/templates/sectionProgressV2`. Nothing is blocking turning on for other folders, but we don't want to increase the size of this PR too much.

## Follow-up work

In order to disable the new import/order rule for a file, for any reason, add the following code to the top of the file above imports:
```
/* eslint-disable import/order */
```

We are going to be slowly working the number of files that have overrides down to zero. We are going to do this by:

Every few days, removing 1-3 folders from the overrides, running `yarn lint --fix` and committing the changes.

We are doing this is stages so that the amount of files touched in each PR is smaller and so that any problems are much easier to revert and debug.

Any developer can turn this on for any folder at any time by following the above steps.

## PR Checklist:

- [ ] Code is well-commented


apps/.eslintrc.js
```diff
@@ -142,5 +176,182 @@ module.exports = {
         'storybook/no-title-property-in-meta': 'error',
       },
     },
+    {
+      files: [
```
Hamms, line 183: could you add a comment with a TODO and an explanation of our plan to gradually reduce this list down to zero?

apps/test/unit/lib/tools/jsdebugger/JsDebuggerTest.js
```diff
@@ -18,7 +19,6 @@ import {
 } from '@cdo/apps/redux';
 import commonReducers from '@cdo/apps/redux/commonReducers';
 import {setPageConstants} from '@cdo/apps/redux/pageConstants';
-import {sandboxDocumentBody} from '../../../../util/testUtils';
```
Hamms, line 21: did we get this deduping for free with the auto-fix functionality, or was this a manual fix?

Hamms (commented):
Looks great for an initial pass! I'm definitely looking forward to shrinking that list of exceptions.

One thing I'd like to see to facilitate that shrinking is some brief documentation about how to apply the automatic fixes and the in-file ignore directives, possibly in the follow-up section of the PR description. Doesn't have to be much since I imagine it's mostly pretty intuitive; just something so that if this gets forgotten about for a couple weeks, it won't be hard to pick back up.

apps/test/unit/lib/tools/jsdebugger/JsDebuggerTest.js
lfryemason: This was manual, but it was the only failure from turning on the rule and it should at least show up as red squigglies if the IDE is configured correctly

cat5inthecradle:
As a fast-follow, would it be relatively safe to update all "test/**" files in a single PR? Manually exercising all functionality of those files (aka "running tests") is easy to do.

lfryemason:
> As a fast-follow, would it be relatively safe to update all "test/**" files in a single PR? Manually exercising all functionality of those files (aka "running tests") is easy to do.

That would probably be a good next step. Depending on how many files that is, I may break it into multiple PRs unless you think it is fine to do all test files at once

cat5inthecradle:
> Depending on how many files that is, I may break it into multiple PRs unless you think it is fine to do all test files at once

For test files, I've got pretty high confidence that if tests pass, then the change is safe. Implementation files definitely need eyes, so smaller PR's are helpful.

I don't work in our JS every day though, so I'm just one opinion.

Hamms (approved):
LGTM!

Definitely agree with Darin that tests make sense as a first place to start updating these, and with Bethany in the Slack thread that there's value in giving the engineering team broadly a chance to weigh in on this before merging.

snickell (approved):
I like this! ✅

At first I wasn't sure about separating . vs .. relatives, but it grew on me. I like how the reformatted code looks in general.

Question: how does the auto-reformatter handle comments in the imports? Not an important question, but curious since it will come up if we try reformatting en-masse. I like the ability to keep comments with imports, I think commenting imports is underpracticed and underrated for long term dep maintenance. But that's just my weird agenda / curiosity ;-)

lfryemason:
> Question: how does the auto-reformatter handle comments in the imports? Not an important question, but curious since it will come up if we try reformatting en-masse. I like the ability to keep comments with imports, I think commenting imports is underpracticed and underrated for long term dep maintenance. But that's just my weird agenda / curiosity ;-)

That's a good question. It looks like comments will not move. I think we will need to audit all of the import changes to make sure nothing like this goes out of date.

PR #57249, opened 2024-03-13
Revert "Tag prod autoscaled instances with Name and environment"

unlox775-code-dot-org:
Reverts code-dot-org/code-dot-org#57007


cat5inthecradle (approved):
Approved, however I think this fix forward is correct:
https://github.com/code-dot-org/code-dot-org/pull/57250

PR #57256, opened 2024-03-14
Music: Separate dark mode

breville:
Currently, when dark mode is either turned on or off using the Blockly context menu, there's no way to return to the "initial state" in which Music Lab (our first lab to have a default of dark mode) has a dark mode Blockly area while all other labs have a non-dark mode Blockly area, short of deleting a local storage object.

This issue is tracked [here](https://codedotorg.atlassian.net/browse/CT-327), and apparently a potential long-term fix is to implement a different menu which allows the selection of dark mode to be "on", "off", or "default", the latter of which would allow for returning to the "initial state" described above.

This PR proposes an interim solution, in which labs which have a default of dark mode (currently only Music Lab) have their theme tracked in a separate local storage object.

### DIFF
--- a/apps/src/blockly/addons/cdoUtils.ts
+++ b/apps/src/blockly/addons/cdoUtils.ts
@@ -11,6 +11,8 @@ import {
   stringIsXml,
   Themes,
   ToolboxType,
+  BLOCKLY_THEME,
+  BLOCKLY_THEME_DEFAULT_DARK_MODE,
 } from '../constants';
 import {
   appendProceduresToState,
@@ -297,11 +299,16 @@ export function getField(type: string) {
  * @returns {?Blockly.Field}
  */
 // Users can change their active theme using the context menu. Use this setting, if present.
-export function getUserTheme(themeOption: string | undefined) {
+export function getUserTheme(
+  themeOption: string | undefined,
+  isDefaultDarkMode: boolean = false
+) {
+  const keyName = isDefaultDarkMode
+    ? BLOCKLY_THEME_DEFAULT_DARK_MODE
+    : BLOCKLY_THEME;
+
   return (
-    Blockly.themes[localStorage.blocklyTheme as Themes] ||
-    themeOption ||
-    cdoTheme
+    Blockly.themes[localStorage[keyName] as Themes] || themeOption || cdoTheme
   );
 }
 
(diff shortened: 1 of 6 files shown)


mikeharv:
Could we not also consider just adding a “lab default theme” option to the context menu? This would also solve the problem of not being able to return to an “initial state” and keep things consistent across our labs. This feels preferable to me than tracking theme preferences in two separate places.

breville:
> Could we not also consider just adding a “lab default theme” option to the context menu? This would also solve the problem of not being able to return to an “initial state” and keep things consistent across our labs. This feels preferable to me than tracking theme preferences in two separate places.

I think a menu with:

> Return to lab default theme
> Turn off dark mode

is confusing.

I think the right long term fix is more likely a menu with three radio buttons/options:

> Light mode
> Dark mode
> Auto

If that's possible in the context menu, then that seems alright, though I think they probably belong outside of the context menu in a place that feels more site-wide.

But short of that, I think the interim solution is less confusing.  And once we build the proper solution, we can ignore (and erase) the extra local storage option.

thomasoniii (approved):
lgtm 🚀

apps/src/blockly/addons/cdoUtils.ts
```diff
@@ -297,11 +299,16 @@ export function getField(type: string) {
  * @returns {?Blockly.Field}
  */
 // Users can change their active theme using the context menu. Use this setting, if present.
-export function getUserTheme(themeOption: string | undefined) {
+export function getUserTheme(
+  themeOption: string | undefined,
+  isDefaultDarkMode: boolean = false
```
sanchitmalhotra126, line 304: Currently we don't have a way to change theme at all in Music Lab right? Could we just pass in some `ignoreLocalStorageTheme` param here or something for now?

breville: You can change theme (and toggle dark mode) using the right-click context menu in Blockly in Music Lab.

ebeastlake:
If I were proposing an interim solution, it would be to remove these options from the context menu and local storage altogether (i.e., use each lab's default color scheme exclusively) until dark and light modes are fully supported.

breville:
Closing in lieu of #57293.

PR #57257, opened 2024-03-14
Music: sound selection UI tweaks

breville:
Three tweaks to the sound selection UI:

- When using Blockly's keyboard navigation, pressing Enter on the `play sound` field would open the sound selection UI, but pressing Enter again would trigger an additional open.  It would be ideal if Blockly didn't try to handle additional Enter presses while we are in this UI, but in lieu of that, we now explicitly prevent additional opens.
- Fixes a bug in which the initial display of the sound selection UI would show all sounds in a folder, rather than the optional subset specified in level data.
- The cursor no longer shows as a pointer for the entire UI, but only on clickable elements.  
- Remove "vocals" filter option for now.

### DIFF
--- a/apps/src/music/blockly/FieldSounds.js
+++ b/apps/src/music/blockly/FieldSounds.js
@@ -25,6 +25,7 @@ class FieldSounds extends GoogleBlockly.Field {
     this.CURSOR = 'default';
     this.backgroundElement = null;
     this.currentFieldWidth = 0;
+    this.showingEditor = false;
   }
 
   saveState() {
@@ -69,6 +70,10 @@ class FieldSounds extends GoogleBlockly.Field {
   }
 
   showEditor_() {
+    if (this.showingEditor) {
+      return;
+    }
+
     super.showEditor_();
 
     const editor = this.dropdownCreate_();
@@ -83,6 +88,8 @@ class FieldSounds extends GoogleBlockly.Field {
       this,
       this.dropdownDispose_.bind(this)
     );
+
+    this.showingEditor = true;
   }
 
   dropdownCreate_() {
@@ -136,6 +143,7 @@ class FieldSounds extends GoogleBlockly.Field {
 
   dropdownDispose_() {
     this.newDiv_ = null;
+    this.showingEditor = false;
   }
 
   hide_() {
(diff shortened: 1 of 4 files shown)


bencodeorg (approved):

apps/src/music/views/SoundsPanel.tsx
```diff
@@ -220,7 +220,7 @@ const SoundsPanel: React.FunctionComponent<SoundsPanelProps> = ({
   const libraryGroupPath = library.libraryJson.path;
 
   const [selectedFolder, setSelectedFolder] = useState<SoundFolder>(
-    library.getFolderForSoundId(currentValue) || folders[0]
+    folders.find(folder => folder.id === currentValue) || folders[0]
```
sanchitmalhotra126: Checking my understanding - in which case would a folder's ID match `currentValue`? `currentValue` is a sound ID right?

apps/src/music/blockly/FieldSounds.js
```diff
@@ -69,6 +70,10 @@ class FieldSounds extends GoogleBlockly.Field {
   }
 
   showEditor_() {
+    if (this.showingEditor) {
```
sanchitmalhotra126, line 73: A bit surprising that Blockly doesn't handle this automatically - I would think it would call the `dropdownDispose` handler whenever the field is reopened?

molly-moen: the show editor logic is pretty [bare bones](https://github.com/google/blockly/blob/81e2203f7fd4f61d015bd6e82114c69da9656adf/core/field.ts#L778-L791), they expect the subclass to define `showEditor_`

molly-moen (approved):

apps/src/music/views/SoundsPanel.tsx
breville: Thanks for catching this.  This was indeed not correct.  I've made an update which takes the "allowed sounds" into account properly when initializing the UI with an existing sound selected.

While making the update, it occurred to me that we could use the allowed sounds for all enumerations of the library. This would lead to stronger enforcement around which sounds could be used, in contrast to the current approach which only does enforcement by limiting the sounds shown in the selection UI.  That said, there might also be some risk, which is why I think it could be left for a potential follow-up.

sanchitmalhotra126 (approved):

PR #57263, opened 2024-03-14
P20-749: Add the "Canvas Integration Early Access" block to the incubator page

artem-vavilov:
## Summary
The block will be displayed by default, but it can be removed by `DCDO.set('incubator-canvas-block-enabled', false)`

## Screenshot
incubator-canvas-integration

## Links
- JIRA ticket: [P20-749](https://codedotorg.atlassian.net/browse/P20-749)


apps/src/templates/studioHomepages/Incubator.jsx
```diff
@@ -34,6 +44,27 @@ class Incubator extends Component {
...
+              description={i18n.incubator_canvasIntegration_earlyAccess_desc()}
+              marginBottom="20px"
+              buttons={[
+                {
+                  url: 'https://forms.gle/x7EBBiC18yJysb5D7',
+                  text: i18n.incubator_canvasIntegration_earlyAccess_signIn_button(),
+                  target: '_blank',
+                },
+                {
+                  url: '#',
```
artem-vavilov: The demo link will be provided by @alivira this week

apps/i18n/common/en_us.json
```diff
@@ -1347,6 +1347,10 @@
   "incorrectAnswerUnmutableBody": "The answer you've entered is not correct.",
   "incorrectSolution": "Incorrect solution",
   "incorrectSolutionBody": "The solution you've entered is not correct.  Please try again!",
+  "incubator_canvasIntegration_earlyAccess_title": "Canvas Integration Early Access",
```
carl-codeorg, line 1357: Is this the recommended naming convention for i18n strings to mimic a hierarchical structure? Curious why we don't support nested objects in our front-end strings when we do for our backend strings.

apps/src/templates/studioHomepages/Incubator.jsx
```diff
@@ -2,8 +2,18 @@ import React, {Component} from 'react';
 import HeaderBanner from '../HeaderBanner';
 import TwoColumnActionBlock from '@cdo/apps/templates/studioHomepages/TwoColumnActionBlock';
 import Button from '@cdo/apps/templates/Button';
+import DCDO from '@cdo/apps/dcdo';
+import i18n from '@cdo/locale';
```
mgc1194, line 10: wow I can't believe we were not localizing strings here before.

mgc1194 (approved):
Looks great!
Thanks for adding theses changes!

apps/src/templates/studioHomepages/Incubator.jsx
artem-vavilov: Replaced with the "Giudes" button

apps/i18n/common/en_us.json
```diff
@@ -1353,6 +1353,10 @@
...
+  "incubator_canvasIntegration_earlyAccess_desc": "Try out our one-click Single Sign-On (SSO) and Roster Syncing integrations for Canvas now!",
+  "incubator_canvasIntegration_earlyAccess_signIn_button": "Sign-in now",
```
carl-codeorg: Should this actually be "Sign up now"? The link goes to the EA form

apps/i18n/common/en_us.json
artem-vavilov: Based on a discussion in one of our daily meetings, this is indeed the recommended naming convention for the i18n hierarchical structure.
@mgc1194, please correct me if I'm mistaken.

carl-codeorg (approved):
LGTM

apps/i18n/common/en_us.json
artem-vavilov: @carl-codeorg I'm not sure, but I believe it's related to the convertation of i18n strings into JavaScript functions.

apps/src/lib/util/AnalyticsConstants.js
```diff
@@ -220,6 +220,10 @@ const EVENTS = {
   RECOMMENDED_SIMILAR_CURRICULUM_SHOWN: 'Recommended Similar Curriculum Shown',
   RECOMMENDED_SIMILAR_CURRICULUM_CLICKED:
     'Recommended Similar Curriculum Clicked',
+
+  // LTI Incubator
+  LTI_INCUBATOR_SIGNUP_CLICKED: 'LTI_Incubator_SignUp_Clicked',
+  LTI_INCUBATOR_GUIDES_CLICKED: 'LTI_Incubator_Guides_Clicked',
```
nicklathe: Currently with other LTI event names, we've gone with lower_camel_case. I recommend converting to that so we are consistent.

apps/src/templates/studioHomepages/Incubator.jsx
```diff
@@ -2,8 +2,36 @@ import React, {Component} from 'react';
...
+  }
+
+  reportEvent(eventName, platform = PLATFORMS.STATSIG) {
+    analyticsReporter.sendEvent(
+      eventName,
+      {
+        userId: this.state.currentUser.userId,
+        userType: this.state.currentUser.userType,
+        userName: this.state.currentUser.userName,
+        userRoleInCourse: this.state.currentUser.userRoleInCourse,
```
nicklathe, line 28: The way Statsig is configured, it [builds a user](https://github.com/code-dot-org/code-dot-org/blob/staging/apps/src/lib/util/StatsigReporter.js#L23) object as part of the initialization process and includes it in the event log, so you don't need to add it here in the metadata. Also, I'm not sure we want to log the username, as I believe this is PII.

As a result, I'd remove everything here except userRoleInCourse.

nicklathe (commented):
Left a couple comments.

apps/src/templates/studioHomepages/Incubator.jsx
artem-vavilov: Got it, thanks!

apps/src/lib/util/AnalyticsConstants.js
nicklathe: I mistyped. I meant snake_case, not camel_case. So for example: `lti_incubator_signup_clicked`.

nicklathe (approved):
:shipit:

apps/i18n/common/en_us.json
mgc1194: It is related to what Artem said. I was trying to find where in the codebase we do the conversion of json to JS functions, but I could not get further than the [`safeLoadLocale`](https://github.com/code-dot-org/code-dot-org/blob/staging/apps/src/util/safeLoadLocale.js).

mgc1194: Alternatively, we could have a separate file for LTI strings as we do for labs specific strings. But I prefer this approach.

mgc1194 (approved):

PR #57267, opened 2024-03-14
P20-755: Add LMS Feedback modal

artem-vavilov:
## Summary
The modal will be displayed to LMS-integrated teachers based on the timing set in `DCDO.get('lti_feedback_show_in', false)` after their first sign-in. Once the modal is closed or submitted, it will not be shown to the teacher again.

## Screenshots
| Feedback Form  | Feedback Provided |
| ---------------- | ------------------- |
| form | provided |

## Overview
overview

## Links
- JIRA ticket: [P20-755](https://codedotorg.atlassian.net/browse/P20-755)
- [Figma](https://www.figma.com/file/DophYsrydeBxakbt2qfV18/LTI-Mock-ups?type=design&node-id=342-719&mode=design&t=Z8hdEz9s1muIZhGD-0)


dashboard/db/migrate/20240307095343_create_lti_feedbacks.rb
```diff
@@ -0,0 +1,13 @@
+class CreateLtiFeedbacks < ActiveRecord::Migration[6.1]
```
carl-codeorg, line 1: If possible, can you break the migration into a separate PR? The guidance from infra is that all DB migrations be done in separate PRs, then the associated code in follow-up PRs.

dashboard/db/migrate/20240307095343_create_lti_feedbacks.rb
```diff
@@ -0,0 +1,13 @@
...
+  def change
+    create_table :lti_feedbacks do |t|
```
carl-codeorg, line 3: Plural of "feedback" is still "feedback" I believe

dashboard/app/models/lti.rb
```diff
@@ -0,0 +1,5 @@
+module Lti
+  def self.table_name_prefix
```
mgc1194, line 2: I may be missing something, could you point out where this method is used?

carl-codeorg, line 2: Cool trick with the prefix. Wish we had modularized the other tables to start with!

carl-codeorg (commented):
I think we're generally supposed to add new components in React instead of haml + js. What's the reasoning behind not using react here?

dashboard/app/models/lti.rb
artem-vavilov: This is a part of the Rails model modularization functionality: [Using table_name_prefix and namespace to organize models](https://icelandcheng.medium.com/ruby-on-rails-using-table-name-prefix-and-namespace-to-organize-models-a08199deaa69)

dashboard/db/migrate/20240307095343_create_lti_feedbacks.rb
artem-vavilov: That's what Rails generated with `rails g model Lti::Feedback`. Do you think we should use `lti_feedback` or maybe something like `lti_feedback_items` for the table name?

carl-codeorg: Yeah I figured it was the automatic pluralization not knowing about "feedback." IMO just `lti_feedback` is fine.

artem-vavilov:
> I think we're generally supposed to add new components in React instead of haml + js. What's the reasoning behind not using react here?

I realized this after finishing the modal. At first, I thought it was the general approach since we already have a bunch of haml templates. Should I refactor this using ReactJS?

dashboard/app/models/lti.rb
artem-vavilov: We could still apply modularization to the existing models

carl-codeorg:
> > I think we're generally supposed to add new components in React instead of haml + js. What's the reasoning behind not using react here?
> 
> I realized after finishing the modal. At first, I thought it was the general approach since we already have a bunch of haml templates. Should I refactor this using ReactJS?

Maybe bring this up during parking lot for team feedback. For this component, which is pretty short-lived and not really reused anywhere, it doesn't seem that bad to keep it as-is. But some folks may have stronger reasoning for refactoring to React.

dashboard/db/migrate/20240307095343_create_lti_feedbacks.rb
artem-vavilov: I've noticed that we are already using the plural form of this word. Therefore, if we aim to apply the plural form fix correctly by adding it to the Rails [inflector](https://github.com/code-dot-org/code-dot-org/blob/staging/dashboard/config/initializers/inflections.rb#L7), we must also change the names of the remaining tables accordingly. Otherwise, we might break some ActiveRecord functionality for those models.
https://github.com/code-dot-org/code-dot-org/blob/fa9ebf57df02525072d11688f6cbfb4c008cf15a/dashboard/db/schema.rb#L2013
https://github.com/code-dot-org/code-dot-org/blob/fa9ebf57df02525072d11688f6cbfb4c008cf15a/dashboard/db/schema.rb#L614

carl-codeorg (approved):

dashboard/db/migrate/20240307095343_create_lti_feedbacks.rb
carl-codeorg: Ok - don't think this is worth the risk of making the change to the inflector. Let's just leave it as is.

artem-vavilov:
Replaced with the feedback banner: [P20-793](https://codedotorg.atlassian.net/browse/P20-793)

PR #57271, opened 2024-03-14
skip ai eval ui tests in firefox and add more debug info

davidsbailey:
1. Sometimes ai_evaluate_student_code.feature fails in firefox due to a login-related problem that seems like a familiar issue for firefox: 
* https://cucumber-logs.s3.amazonaws.com/test/test/Firefox_teacher_tools_rubrics_ai_evaluate_student_code_output.html?versionId=r4qkIeC5KNtq_mmoPchyYAe4aqbkuM7x
* Screenshot 2024-03-14 at 10 50 30 AM

since we're trying hard to deflake this test right now, this PR proposes disabling the UI test in firefox until we've got the test into a more stable state.

2. in a previous PR, mark added a new ui test command to print the current user's info. this PR also adds that debug info to other scenarios. so far it looks like the user ids vary between test runs, however this logging could also be useful for matching up the RubricAiEvaluation records in the database to a particular test run.

## Testing story

Verified locally that the ui tests do not get run in firefox

verified on the test machine that user info gets printed:
Screenshot 2024-03-14 at 11 22 44 AM

### DIFF
--- a/dashboard/test/ui/features/teacher_tools/rubrics/ai_evaluate_student_code.feature
+++ b/dashboard/test/ui/features/teacher_tools/rubrics/ai_evaluate_student_code.feature
@@ -57,6 +57,7 @@ Feature: Evaluate student code against rubrics using AI
 
   Scenario: Student code is evaluated by AI when teacher requests individual evaluation
     Given I create a teacher-associated student named "Aiden"
+    And I get debug info for the current user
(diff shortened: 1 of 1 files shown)


davidsbailey:
going to merge this ahead of review to hopefully get more useful debugging info from upcoming DTTs.

PR #57284, opened 2024-03-14
Center level numbers in header row

etaderhold:
The title of [the ticket](https://codedotorg.atlassian.net/browse/TEACH-918) was to "Vertically center assessment icon," but to my eye the star icon was centered pretty well already, and the level numbers were a bit above center. This change nudges them down a bit.

Before:
image

After:
image

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/templates/sectionProgressV2/progress-table-v2.module.scss
+++ b/apps/src/templates/sectionProgressV2/progress-table-v2.module.scss
@@ -292,7 +292,6 @@ $level-cell-width: 52;
 }
 
 .assessmentLevelIcon {
-  //margin-top: -2px;
   font-size: 12px;
   color: $neutral_dark60;
 }
--- a/apps/src/templates/sectionProgressV2/progress-table-v2.module.scss
+++ b/apps/src/templates/sectionProgressV2/progress-table-v2.module.scss
@@ -216,7 +216,7 @@ $level-cell-width: 52;
       flex-wrap: nowrap;
 
       &LevelNumber {
-        padding-top: 2px;
+        margin-top: 0.0625em;
       }
 
       &Unexpanded {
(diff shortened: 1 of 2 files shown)


apps/src/templates/sectionProgressV2/progress-table-v2.module.scss
```diff
@@ -214,10 +214,14 @@ $level-cell-width: 52;
       justify-content: space-evenly;
       overflow-x: hidden;
       flex-wrap: nowrap;
-    }
 
-    .expandedHeaderLevelCellUnexpanded {
-      background-color: white;
+      &LevelNumber {
+        margin-top: 0.0625em;
```
lfryemason, line 219: Did you try using line height instead of margin? My only concern is that this might change based on zoom levels.

lfryemason (approved):
Did you try on multiple browsers? I know Bryan mostly saw the differences on Edge

etaderhold:
> Did you try on multiple browsers? I know Bryan mostly saw the differences on Edge

I did not try it on Edge. TIL that they have a Mac version.

I tried playing around with line height but I think I saw strange results (like maybe sometimes increasing the line height would move the text down and other times it would move up, depending on the value), but that might have been some other property.

Yes the alignment does look a bit different at different zoom levels, both then and now.

If you want to spend an hour and see if you can find a better solution I'm all in favor. CSS and I often don't get along very well.

apps/src/templates/sectionProgressV2/progress-table-v2.module.scss
dju90: If we end up using margin, I'm wondering if we can move the star up just a wee bit more, such that the bottom of the star lines up with the bottom of the text. Because the star is bottom heavy, at normal zoom levels it still looks a tiny bit lower in alignment right now compared to the text. (I pulled the branch and looked at in on Chrome and Firefox)
image

dju90: From office hours - probably a font issue with Metropolis. Comparing between Firefox, Chrome, and Edge on Windows at varying browser zoom levels and screen zoom levels, Eric's margin seemed to have the least variance, so we will keep it as the PR is currently written for now.

PR #57285, opened 2024-03-14
[Game Design] CT-357: Handle long variable bubble labels and values

ebeastlake:
The following PR:
- Moves the "magic" numbers (padding, textSizeValue, strokeRadius, etc.) into a "config" object for variable bubbles used throughout the bubble drawing code.
- Adds a `drawUtils.truncateText` function that accepts a piece of text, a maximum width, and text size and truncates the given text, if necessary, to fit within the provided width.
- Adds JSDocs for `drawVariableBubbles` in CoreLibrary.
- Adds logic to restrict the length of the label in the variable bubble to `config.maxLabelLength` (currently 30 characters).
- Adds logic to ensure the text for the value fits within the available remaining space by truncating it, if necessary.
- Note: `drawUtils.variableBubble` now accepts "displayText," which should be a string whose width is smaller than the playspace. 

The truncation behavior is demonstrated in the video below. This video features a **known edge case**: If the variable label is very long and there is a large numerical value that p5 converts to scientific notation, the scientific notation is cut off, making it look like the value is 1.2 instead of 1.2E9 or something. Do we care about this?

https://github.com/code-dot-org/code-dot-org/assets/26844240/65a7beff-9080-4280-8f09-1e1aec0b26a2

My thought is that fixing the scientific notation + large label is not a common enough case that it's worth handling separately (which would require something like using a different max character limit if we detect the value is a number in scientific notation). 

## Links

Jira ticket: https://codedotorg.atlassian.net/browse/CT-357

## Testing story

Tested manually on a Sprite Lab level, including numerical, text, array, and boolean values (see above).

## PR Checklist:

- [x] Code is well-commented

### DIFF
--- a/apps/src/p5lab/spritelab/CoreLibrary.js
+++ b/apps/src/p5lab/spritelab/CoreLibrary.js
@@ -117,11 +117,13 @@ export default class CoreLibrary {
   }
 
   /**
-   * Draws bubbles for each variable in `variableBubbles` array on the canvas. It truncates labels and values to fit within
-   * the allocated space, ensuring neither exceeds half of the available width, accounting for separators.
-   * Configuration for text size, padding, stroke weight, and stroke radius is specified within the method.
+   * Draws bubbles for each variable in the `variableBubbles` array. Labels are truncated with an ellipsis
+   * if they exceed the maximum character limit. Values are truncated as needed to fit the remaining space.
    *
-   * @param {Object[]} this.variableBubbles - Array of variable objects to be displayed, each containing `name`, `label`, and `location`.
+   * @param {Object[]} this.variableBubbles - An array of objects, each representing a variable to be displayed. Each object should include:
+   *  - `name`: A string identifier for the variable used to get the value from the JSInterpreter.
+   *  - `label`: A label for the variable to be displayed in the bubble.
+   *  - `location`: An object specifying the `x` and `y` coordinates where the bubble should be drawn.
    */
   drawVariableBubbles() {
     const config = {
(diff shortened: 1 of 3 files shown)


apps/src/p5lab/spritelab/CoreLibrary.js
```diff
@@ -116,18 +116,69 @@ export default class CoreLibrary {
...
+   * Draws bubbles for each variable in the `variableBubbles` array. Labels are truncated with an ellipsis
+   * if they exceed the maximum character limit. Values are truncated as needed to fit the remaining space.
+   *
+   * @param {Object[]} this.variableBubbles - An array of objects, each representing a variable to be displayed. Each object should include:
+   *  - `name`: A string identifier for the variable used to get the value from the JSInterpreter.
+   *  - `label`: A label for the variable to be displayed in the bubble.
+   *  - `location`: An object specifying the `x` and `y` coordinates where the bubble should be drawn.
+   */
   drawVariableBubbles() {
+    const config = {
```
ebeastlake, line 129: I was tired of having these magic numbers repeated in multiple places. Thoughts? Should this be stored elsewhere, like in a constants file as VARIABLE_BUBBLES_CONFIG?

molly-moen: I think either here or in a constants file is fine! As long as the numbers aren't repeated.

molly-moen (approved):
Looks good! Per the scientific notation question: I think ideally we would preserve that, but it's a pretty rare scenario. It should be reasonable to solve, we would just need to check the end of the string for `e` and preserve anything after that. That could be a follow-up.

PR #57286, opened 2024-03-15
Music: wait validation

breville:
This adds a new kind of validation to Music Lab which waits until playback has passed a given measure.  It allows us to show a "Let's listen..." validation message while the playback occurs, before revealing more detailed information about what's missing.  

Because of the ordering of the validations, a level success (playing three sounds simultaneously) still shows as soon as it's done.  

Because "Let's listen..." only makes sense while the program is running, a new optional `"showOnlyWhileRunning": true` parameter is used in the validation definition.

And because we don't want this particular validation to persist during a run once we pass the measure in question (that is, this validation will often start failing partway through a run), we avoid accumulating it, and rather check fresh for it each frame.

https://github.com/code-dot-org/code-dot-org/assets/2205926/904252b3-c95d-41df-a5ae-e226a6bfba0c

### DIFF
--- a/apps/src/lab2/lab2Redux.ts
+++ b/apps/src/lab2/lab2Redux.ts
@@ -45,6 +45,8 @@ export interface LabState {
   isLoadingProjectOrLevel: boolean;
   // If the lab is loading. Can be updated by lab-specific components.
   isLoading: boolean;
+  // Whether the lab is currently running.
+  isRunning: boolean;
   // Error currently on the page, if present.
   pageError: PageError | undefined;
   // channel for the current project, or undefined if there is no current project.
@@ -67,6 +69,7 @@ const initialState: LabState = {
   initialSources: undefined,
   validationState: getInitialValidationState(),
   levelProperties: undefined,
+  isRunning: false,
 };
 
 // Thunks
@@ -238,6 +241,9 @@ const labSlice = createSlice({
   reducers: {
     setIsLoading(state, action: PayloadAction<boolean>) {
       state.isLoading = action.payload;
+    },
+    setIsRunning(state, action: PayloadAction<boolean>) {
+      state.isRunning = action.payload;
     },
     setPageError(
       state,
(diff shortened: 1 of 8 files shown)


thomasoniii (approved):
lgtm 🚀

apps/src/lab2/lab2Redux.ts
```diff
@@ -45,6 +45,8 @@ export interface LabState {
   isLoadingProjectOrLevel: boolean;
   // If the lab is loading. Can be updated by lab-specific components.
   isLoading: boolean;
+  // Whether the lab is currently running.
+  isRunning: boolean;
```
sanchitmalhotra126, line 49: This makes me a bit nervous 🙂 probably being a little over-cautious because this feels a bit StudioApp-y, but I also don't think it feels right to have Lab2 know about the "running" state of a lab, since that can vary quite a bit from lab to lab (ex. AI Chat vs Music Lab). It feels like that's something the lab should manage internally.

sanchitmalhotra126 (commented):
Thanks for diving into this! This feels a bit tricky - my main issue is storing `isRunning` state in Lab2 as noted below, but also understand that this gets complicated because progress is managed by Lab2. 

That being said, the desired behavior feels pretty music lab-specific. Could we pull this behavior and logic down into the MusicValidator/MusicView layer more? One idea -
- We always have a base "Let's listen..." condition that shows while playback is in progress (this is internal to MusicValidator; not added by levelbuilders).
- If a passing condition is met while playback is in progress, display that immediately and pass the level.
- If not, show the most relevant failing condition as soon as playback stops.
- When playback starts again, reset validation to the "Let's listen..." condition

apps/src/lab2/lab2Redux.ts
breville: Good point.  I also felt uncomfortable about this, especially since it duplicated `MusicState`'s `isPlaying`.

breville:
> That being said, the desired behavior feels pretty music lab-specific. Could we pull this behavior and logic down into the MusicValidator/MusicView layer more?

Thanks for the thoughtful feedback!  I revisited the entire approach after looking at how earlier labs deal with timeouts (such as `timeout_failure_tick`/`timeout_after_when_run` in Play Lab, and `level_verification_timeout` in MineCraft) and am proposing a Music Lab-specific implementation in https://github.com/code-dot-org/code-dot-org/pull/57331.

PR #57287, opened 2024-03-15
Music: fix for library switching

breville:
Music Lab assumes that each project will always use the sound library that it's created with, and so we don't mind some erroneous behaviour when an existing project is loaded with a different library through the `?library=` URL parameter.  That said, the `play sound` block was throwing an exception when a sound wasn't recognised, which led to some "unknown" blocks being saved.

With this change, the exception is no longer thrown, and instead the block field is rendered blank.  We should still notice this if it happens unexpectedly (and we plan to never seen this happening for end-users), but it makes internal scenarios a bit less confusing.  In fact, it means that as a project changes back and forth between libraries, all blocks maintain their internal values, and those that can be played, will.

To illustrate, here's a project loaded with the `intro2024` library:
Screenshot 2024-03-15 at 4 00 10 PM

And here's the same project loaded with the `launch2024` library via URL parameter:
Screenshot 2024-03-15 at 3 59 43 PM

### DIFF
--- a/apps/src/music/blockly/FieldSounds.js
+++ b/apps/src/music/blockly/FieldSounds.js
@@ -198,18 +198,20 @@ class FieldSounds extends GoogleBlockly.Field {
     // Add an image for the sound type.
     const soundType = this.options
       .getLibrary()
-      .getSoundForId(this.getValue()).type;
+      .getSoundForId(this.getValue())?.type;
 
-    GoogleBlockly.utils.dom.createSvgElement(
-      'image',
-      {
-        x: 6,
-        y: 3,
-        width: 15,
-        href: `/blockly/media/music/icon-${soundType}.png`,
-      },
-      this.backgroundElement
-    );
+    if (soundType) {
+      GoogleBlockly.utils.dom.createSvgElement(
+        'image',
+        {
+          x: 6,
+          y: 3,
+          width: 15,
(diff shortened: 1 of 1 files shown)


thomasoniii (approved):
lgtm 🚀

apps/src/music/blockly/FieldSounds.js
```diff
@@ -198,18 +198,20 @@ class FieldSounds extends GoogleBlockly.Field {
...
-        x: 6,
-        y: 3,
-        width: 15,
-        href: `/blockly/media/music/icon-${soundType}.png`,
-      },
-      this.backgroundElement
-    );
+      .getSoundForId(this.getValue())?.type;
+
+    if (soundType) {
```
sanchitmalhotra126, line 203: Do we want to log when this happens, in case this is also somehow happening in the wild (or to make sure we're not affecting projects when we update libraries)?

breville: Good call.  That said, I worked on the code for it, and then observed that we have an existing warning that does the job:
https://github.com/code-dot-org/code-dot-org/blob/e4f50e71cecc856f1f43e8f1851e6830b87af88f/apps/src/music/player/sequencer/Simple2Sequencer.ts#L204-L207

sanchitmalhotra126 (approved):

PR #57288, opened 2024-03-15
Music: update header button behavior

breville:
This work began with the goal of just disabling the "Start Over" button for read-only workspaces, but ended up doing a couple things:

- The refresh, undo, and redo buttons are no longer shown for read-only workspaces.
- The undo and redo buttons are properly disabled when not available, taking them out of keyboard navigation.

### DIFF
--- a/apps/src/music/views/HeaderButtons.tsx
+++ b/apps/src/music/views/HeaderButtons.tsx
@@ -10,6 +10,7 @@ import {AnalyticsContext} from '../context';
 import {MusicState} from '../redux/musicRedux';
 import moduleStyles from './undo-redo-buttons.module.scss';
 import musicI18n from '../locale';
+import {isReadOnlyWorkspace} from '@cdo/apps/lab2/lab2Redux';
 
 interface HeaderButtonsProps {
   onClickUndo: () => void;
@@ -25,6 +26,7 @@ const HeaderButtons: React.FunctionComponent<HeaderButtonsProps> = ({
   onClickRedo,
   clearCode,
 }) => {
+  const canReload: boolean = !useSelector(isReadOnlyWorkspace);
   const {canUndo, canRedo} = useSelector(
     (state: {music: MusicState}) => state.music.undoStatus
   );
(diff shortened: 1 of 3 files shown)


apps/src/music/views/undo-redo-buttons.module.scss
```diff
@@ -22,5 +22,6 @@
   &Disabled {
     opacity: 50%;
     cursor: default;
+    pointer-events: none;
```
breville: This doesn't actually prevent selection by pressing Enter when using keyboard navigation, hence the check in code as well, but felt prudent.

apps/src/music/views/HeaderButtons.tsx
```diff
@@ -73,7 +79,10 @@ const HeaderButtons: React.FunctionComponent<HeaderButtonsProps> = ({
       <button
```
bencodeorg: Why not actually disable the button via the `disabled` attribute?

sanchitmalhotra126: Should we just not render the button at all in this case?

breville: Good call.  Will use this for the undo and redo buttons.

breville: Good call.  Will apply this to three of the button: refresh, undo, redo.

breville:
Thanks for the feedback, @bencodeorg and @sanchitmalhotra126.  I used both of your suggestions broadly.  I've ended up just hiding three of the buttons (refresh, undo, and redo) when the workspace is read-only.  And by applying the `disabled` attribute to the undo and redo buttons when they aren't enabled, they get taken out of keyboard navigation which seems the right approach.

apps/src/music/views/undo-redo-buttons.module.scss
breville: Ended up doing a slightly different approach.

sanchitmalhotra126 (approved):

PR #57101, opened 2024-03-07
Fixing bug in OneTrust feature flag

daynew:
* Fixing a bug where turning off OneTrust would show "return" at the top of the screen.
* Make the code.org and hourofcode.com OneTrust feature flag code behavior match.
* Add unit and UI tests

## Screenshots
Before and after changes when visiting http://localhost.code.org:3000/index?onetrust_cookie_scripts=off
before|after
---|---
before|after

## Links
* [JIRA](https://codedotorg.atlassian.net/browse/P20-781)

## Testing story
Automated testing:
* Unit tests
* UI tests

Manual testing: Visited the following URLs and verified the page rendered fine and onetrust libraries were not loaded.
* http://localhost.code.org:3000/index?onetrust_cookie_scripts=off
* http://localhost-studio.code.org:3000/courses?onetrust_cookie_scripts=off
* http://localhost.hourofcode.com:3000/us?onetrust_cookie_scripts=off

I also visited the above URLs and substituted `onetrust_cookie_scripts` with `test`, `prod`, and nothing to verify the expected OneTrust libraries were loaded.

## PR Checklist:

- [ ] Tests provide adequate coverage
- [ ] Privacy and Security impacts have been assessed
- [ ] Code is well-commented
- [ ] New features are translatable or updates will not break translations
- [ ] Relevant documentation has been added or updated
- [ ] User impact is well-understood and desirable
- [ ] Pull Request is labeled appropriately
- [ ] Follow-up work items (including potential tech debt) are tracked and linked

PR #57289, opened 2024-03-15
Support routing to localhost.hourofcode.com on CI

daynew:
In a [separate PR](https://github.com/code-dot-org/code-dot-org/pull/57101) I added a UI test which tries to visit http://localhost.hourofcode.com:3000/us. On my laptop it worked fine, but on CI it was failing. On CI, the UI test was actually trying to visit https://test.hourofcode.com/us. Here is a summary of the changes in this PR in order to get UI tests able to hit localhost.hourofcode.com successfully.
* Update the `bundle exec ./runner.rb...` command run by CI to target localhost.
* Update `canonical_url` to return localhost* URLs if the code is currently running on a CI webserver or :development.
* Update `config.yml.erb` to handle URLs which might have `:`'s in them because now it might have `localhost.code.org:3000` in it.
* Update `test_environment_aware_dynamic_config_helper.rb` to correctly distinguish between our managed test server and a CI webserver.

## Testing story
* Existing automated tests pass.

## PR Checklist:

- [ ] Code is well-commented


lib/cdo.rb
```diff
@@ -274,6 +275,11 @@ def running_web_application?
       %w(puma thin).include?(File.basename($0))
     end
 
+    # Is this code running in a webserver managed by Drone?
+    def drone_webserver?
```
Hamms: Re: your question in slack about "drone" vs "ci", I think it's probably fine to reference specific products like drone in *comments*, but for variable and method names I'd prefer to use generic terms when possible

config.yml.erb
```diff
@@ -323,7 +323,7 @@ dcdo_table_name:       dcdo_<%=env%>
 gatekeeper_table_name: gatekeeper_<%=env%>
 
 # Poste (internal transactional email processing framework)
-poste_host: <%= code_org_url.sub('//', '') %>
+poste_host: '<%= code_org_url.sub('//', '') %>'
```
Hamms, line 326: Are these quotes (and the ones below) necessary now in a way they weren't before, or is this just making them generally more resilient?

This is probably unnecessary for a URL that we don't expect to ever have unencoded `'` or `"` characters of its own, but I'm also curious if something like this would be more robust:
```suggestion
poste_host: <%= code_org_url.sub('//', '').inspect %>
```

Hamms (commented):
One question and one recommended change, but generally looks good to me!

config.yml.erb
daynew: The issue is that now this returns `localhost.code.org:3000` on CI servers. The YAML parser sees the `:` and says "It looks like you are trying to define a Hash... oops looks like your Hash syntax is bad... \<throws exception\>". By quoting it, we explicitly set the type for the value rather than relying on the YAML parser to divine it.

Before, the CI servers would incorrectly set this to `test.code.org`.

Hamms: Ahh, that makes sense. Oh, yaml :upside_down_face: 

Thanks!

lib/cdo.rb
daynew: I changed my usages of "drone" to "CI".

Hamms (approved):
LGTM, thanks for the improvements!

PR #57293, opened 2024-03-15
[Google Blockly] remove dark mode option from context menu

mikeharv:
## Background and Context

Last month, @breville documented an issue he described as an "endless toggle" problem related to themes in Google Blockly labs:

> Repro:
> 1. visit https://studio.code.org/projectbeats
>   * choose Turn off dark mode.
>   * choose Turn on dark mode.
> 2. visit https://studio.code.org/projects/spritelab
>   * observe that it's unexpectedly dark mode.
>   * choose Turn off dark mode.
> 3. visit visit https://studio.code.org/projectbeats again
>   * observe that it's unexpectedly light mode.

We want to move towards greater consistency between our Blockly labs, but having one lab with mostly dark components makes it less straight-forward. While this was sharing of theme preferences was intentional, he also pointed out that it created the potential for a "partial dark mode" experience that was potentially undesired:
imageimage

**Option 1: Separate storage entries**

As a potential fix, Brendan proposed that we create separate state to track theme preferences in labs with dark mode as the default:
* https://github.com/code-dot-org/code-dot-org/pull/57256

This solves the "endless toggle" problem, but doubles the work for a user who wants an accessible block theme in both types of labs. It also adds an additional thing to `localStorage` which isn't our desired long-term solution. (Ticket for that: https://codedotorg.atlassian.net/browse/CT-77)

**Option 2: "Lab default" theme option**
As a counter-proposal, I suggested that we could simply add a new option to the context menu to revert to the lab's default theme (effectively deleting the `localStorage` entry). This can potentially solve the "endless toggle" problem, but it puts the onus on the effected user to figure out how. This was not drafted in a pull request.

Neither of these two proposals address the second problem: It is still possible to get into a "partial dark mode" state.

**Option 3: Remove dark mode toggling**
Finally, @ebeastlake proposed that turn off dark and light mode options for all Google Blockly labs until we can fully support dark/light mode for the full platform (not just the Blockly workspace). This effectively hides a partially-built feature and solves both of the problems that Brendan raised. This pull request implements Emily's suggested solution.

Full Slack thread: https://codedotorg.slack.com/archives/C05DK21DAHK/p1707716345562319

### Context Menu updates
In this branch, we remove the completely remove option that toggles dark mode (ie. the dark and light versions of our 5 types of themes). 
When a user selects another theme via context menu, we only save the base name (excluding "dark), which describes the user's preferences for block styles.

### Inject Updates
Upon creating a workspace, we will always set the theme to one that matches the light/dark default of the lab. For example, if a user has a saved `'cdohighcontrastdark'` in `localStorage`, we will use the dark high contrast theme for Music Lab, but the standard (light) high contrast theme for other labs. If a user has saved `'cdomodern'` (our light default theme), we will still load Music Lab with the default dark theme. Needing to get the base name is effectively backwards-compatibility stuff for people who have a saved option. 

### Other cleanup
* `getUserTheme`'s `themeOption` parameter was incorrectly listed as a string. Updated to use the `Theme` type.
* A constant for the string `'dark'` was moved to a shared constants file.
* The helper function to get a theme's base name was moved out of the context menu file and into utils.
* The helper function `isDarkTheme` was refactored to take a theme argument rather than a workspace. This better matches the purpose of the function and we were already accessing the theme in the same way outside of this scope.

These changes make it so that users will always see a dark theme in Music Lab (our only dark-default Blockly lab) and a light theme elsewhere. Users can still select between our modern block styles, high contrast block styles, or any of our three block themes for users with limited color vision.

## Links

Slack thread: https://codedotorg.slack.com/archives/C05DK21DAHK/p1707716345562319

## Testing story

Before making any changes to my branch, I opened Music Lab and set the theme to dark/protanopia. After making the changes, I opened Music Lab in another tab and saw that it the theme had not change. I opened Sprite Lab and saw that it was using light/protanopia. I then manually updated my local storage entry to several other potential light/dark options and observed that for each case the workspace was loaded with a theme that matched the lab's other components.
imageimage

I confirmed that option to toggle dark mode was removed from the context menu. I confirmed that we are now only storing the theme's base name in `localStorage`.

## Follow-up work

* Move theme preferences out of `localStorage` ([Jira issue](https://codedotorg.atlassian.net/browse/CT-77))
* Improve discoverability of accessibility features ([Google Doc](https://docs.google.com/document/d/1KeyWlKz7ma-7Oc7ICAtIbreuNUFnJTN10c2pQi3pa0U/edit#heading=h.qi8xxyq72bct))
* (Long-term) Eventually build support for toggling dark mode on/off in our labs ([Figma board](https://www.figma.com/file/hExjmRsMv3jseWOIMjLbj0/Dark-Mode?type=design&node-id=1038-8033&mode=design&t=XyLwWNyi0tCESRRl-0))

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/blockly/addons/cdoUtils.ts
+++ b/apps/src/blockly/addons/cdoUtils.ts
@@ -7,6 +7,7 @@ import {blocks as procedureBlocks} from '../customBlocks/googleBlockly/procedure
 import {
   BLOCK_TYPES,
   CLAMPED_NUMBER_REGEX,
+  DARK_THEME_SUFFIX,
   DEFAULT_SOUND,
   stringIsXml,
   Themes,
@@ -26,9 +27,10 @@ import {
   getProjectXml,
   processIndividualBlock,
 } from '@cdo/apps/blockly/addons/cdoXml';
-import {Block, BlockSvg, Field, WorkspaceSvg} from 'blockly';
+import {Block, BlockSvg, Field, Theme, WorkspaceSvg} from 'blockly';
 import {BlockColor, JsonBlockConfig, WorkspaceSerialization} from '../types';
 import experiments from '@cdo/apps/util/experiments';
+import {getBaseName} from '../utils';
 
 /**
  * Loads blocks to a workspace.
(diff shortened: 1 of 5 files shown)


apps/src/blockly/addons/cdoUtils.ts
```diff
@@ -293,16 +295,30 @@ export function getField(type: string) {
 
 /**
  * Returns a theme object, based on the presence of an option in the browser's localStorage.
- * @param {string} type
+ * @param {?Theme} themeOption
```
molly-moen, line 298: is `themeOption` the default theme? Mind adding a comment to that effect?

molly-moen (approved):
Looks good to me!

thomasoniii (approved):
lgtm 🚀

breville (approved):
@mikeharv Thanks for such a thoughtful write-up and implementation.  

@ebeastlake Thanks for a strong call on building out the ideal user experience.  

In my initial proposal, I had wanted to preserve the flexibility of what had been built (and all the hard work behind it), but from this discussion it's clear that we have a long term vision of the ideal user experience and can get there in time.

PR #57298, opened 2024-03-15
skip ai eval ui test

Nokondi:
This will skip the ai_evaluate_student_code ui test pending an investigation into an error that is causing incorrect AI evaluations to be retrieved from the backend.


dashboard/test/ui/features/teacher_tools/rubrics/ai_evaluate_student_code.feature
```diff
@@ -1,3 +1,4 @@
+@skip
 # AI evaluation is stubbed out in UI tests via the /api/test/ai_proxy/assessment route.
 @no_firefox
```
bethanyaconnor, line 3: nit: do you also need to remove this now?

bethanyaconnor (approved):
Thanks! Feel free to merge ahead of Drone so we can get another DTT going!

PR #57302, opened 2024-03-15
Lab2: add validation messages to loc pipeline

sanchitmalhotra126:
Add Lab2 validation messages to the localization pipeline. This will be necessary for localizing Music Lab ahead of our May marketing launch.

Note that while I tested this with the music-intro-2024 script, the content in that script won't actually be added to the production loc pipeline until we manually add it to the list of translatable units. We'll do that once all localization support is in place and the strings are finalized (https://codedotorg.atlassian.net/browse/LABS-569).

## Links

https://codedotorg.atlassian.net/browse/LABS-548

## Testing story

Tested this end to end by running the i18n test sync locally. Details below -

**Sync In** produces a source file like this:

Screenshot 2024-03-15 at 10 17 52 AM

After running **Sync Up**, we can view and translate the content in Crowdin (these are pseudo-translations for testing purposes):

Screenshot 2024-03-15 at 10 20 26 AM

**Sync Down** and **Sync Out** produces output files like this:

Screenshot 2024-03-15 at 10 28 36 AM

And finally we see the translated content when running the lab in a different locale:

Screenshot 2024-03-15 at 10 32 09 AM

### DIFF
--- a/dashboard/app/models/levels/level.rb
+++ b/dashboard/app/models/levels/level.rb
@@ -727,6 +727,20 @@ class Level < ApplicationRecord
     end
   end
 
+  def localized_validations
+    if should_localize?
+      validations.each do |validation|
+        validation['message'] = I18n.t(
+          validation["key"],
+          scope: [:data, :validations, name],
+          default: validation["message"],
+          smart: true
+        )
+      end
+    end
+    validations
+  end
+
   # There's a bit of trickery here. We consider a level to be
   # hint_prompt_enabled for the sake of the level editing experience if any of
   # the scripts associated with the level are hint_prompt_enabled.
@@ -792,6 +806,8 @@ class Level < ApplicationRecord
     properties_camelized[:appName] = game&.app
     properties_camelized[:useRestrictedSongs] = game.use_restricted_songs?
     properties_camelized[:usesProjects] = try(:is_project_level) || channel_backed?
+    # Localized properties
+    properties_camelized["validations"] = localized_validations if properties_camelized["validations"]
     properties_camelized
   end
 
--- a/dashboard/app/models/levels/level.rb
+++ b/dashboard/app/models/levels/level.rb
@@ -729,7 +729,8 @@ class Level < ApplicationRecord
 
   def localized_validations
     if should_localize?
-      validations.each do |validation|
+      validations_clone = validations.map(&:clone)
+      validations_clone.each do |validation|
         validation['message'] = I18n.t(
           validation["key"],
           scope: [:data, :validations, name],
@@ -737,8 +738,10 @@ class Level < ApplicationRecord
           smart: true
         )
       end
+      validations_clone
+    else
+      validations
     end
-    validations
   end
 
   # There's a bit of trickery here. We consider a level to be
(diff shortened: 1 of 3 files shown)


breville (approved):
This is great!

Out of curiosity, what happens if we change one of the strings, while leaving the key intact?  Do we reset the existing translations to the new English, or do the old translations keep being used until the English is re-translated?

And related, how stable are these keys?  For example, if somebody changes the order of validations in levelbuilder, do their keys stay intact?  What if they change properties of a validation in levelbuilder, perhaps the English text, or perhaps something else?

sanchitmalhotra126:
> Out of curiosity, what happens if we change one of the strings, while leaving the key intact? Do we reset the existing translations to the new English, or do the old translations keep being used until the English is re-translated?

I believe we will use the old translations until the next i18n sync has completed. Is that right @mgc1194?
 
> And related, how stable are these keys? For example, if somebody changes the order of validations in levelbuilder, do their keys stay intact? What if they change properties of a validation in levelbuilder, perhaps the English text, or perhaps something else?

The keys should be stable throughout reordering, editing, etc I believe. We only generate keys if some existing validation doesn't have a key already, or when creating a new validation.

dashboard/app/models/levels/level.rb
```diff
@@ -727,6 +727,20 @@ def localized_rubric_property(property)
     end
   end
 
+  def localized_validations
+    if should_localize?
+      validations.each do |validation|
```
mgc1194: I suggest cloning the validations and editing the copy instead of the original content.
In general, we want to preserve the original data, specifically, it can avoid side effects if validations are used somewhere else in the code. 

Here's an example:
https://github.com/code-dot-org/code-dot-org/blob/ad90d0f431b63944da72a5ff3a2ad701d3b1aed3/dashboard/app/models/activity_section.rb#L65-L73

mgc1194 (changes requested):
These changes look great.
We don't need to change anything in the sync out because `course_content/sync_out.rb` handles localized level properties.
I just have one request regarding modifying the original data.

It also would be really neat if you could update the [unit tests](https://github.com/code-dot-org/code-dot-org/blob/40420e1eef56c895a71a190b1796713b0772b68f/bin/test/i18n/resources/dashboard/course_content/test_sync_in.rb#L334) 
Once that's updated it's good to go.


Regarding keys:

> Out of curiosity, what happens if we change one of the strings, while leaving the key intact? Do we reset the existing translations to the new English, or do the old translations keep being used until the English is re-translated?

>I believe we will use the old translations until the next i18n sync has been completed. Is that right @mgc1194?

Our approach has been to keep old translations until new ones are available.

dashboard/app/models/levels/level.rb
sanchitmalhotra126: Done!

sanchitmalhotra126:
> It also would be really neat if you could update the [unit tests](https://github.com/code-dot-org/code-dot-org/blob/40420e1eef56c895a71a190b1796713b0772b68f/bin/test/i18n/resources/dashboard/course_content/test_sync_in.rb#L334) Once that's updated it's good to go.

Added tests!

mgc1194 (approved):
Awesome!! Thanks for including the changes I requested and for adding tests!!

PR #57306, opened 2024-03-15
Add project saving to Web Lab 2

molly-moen:
This PR adds project saving to Web Lab 2 and unifies the data model for projects between Python Lab and Web Lab 2. We are using the data model created by Jim for CDO IDE, which is still subject to change. In order to support the new data model I needed to update `files_api.has_valid_encoding?` which validates source code is in UTF-8 format.

The changes otherwise are pretty straightforward, we are just hooking into the Lab2 Project Manager code. I added a second Web Lab 2 level to allthethings to verify we can switch between pages without a page reload and load code correctly. 

## Links
- jira ticket: [https://codedotorg.atlassian.net/browse/CT-389](https://codedotorg.atlassian.net/browse/CT-389)

## Testing story
Tested locally. These changes are isolated to Python Lab and Web Lab 2, which are not in use by users yet.

## PR Checklist:

- [ ] Code is well-commented


apps/src/lab2/responseValidators.ts
```diff
@@ -38,9 +38,9 @@ const PythonSourceResponseValidator: ResponseValidator<
     if (typeof responseToValidate.source === 'string') {
       throw new ValidationError('Python sources must be a JSON object');
     }
-    const source = responseToValidate.source as NestedSourceCode;
+    const source = responseToValidate.source as MultiFileSource;
     // TODO: support a nested main.py
-    if (!source['main.py']) {
+    if (!source?.files?.['main.py']) {
```
molly-moen: fyi @thomasoniii for Python Lab I kept the keys of the file map as the file names rather than file ids. This made it easier to look up the main file. We can discuss if we want to do this for all labs or do ids for all labs.

dashboard/legacy/middleware/files_api.rb
```diff
@@ -479,9 +479,18 @@ def has_valid_encoding?(source)
     if source.is_a?(Hash)
       # Iterate over each file
       source.each_key do |key|
-        # Multi-file source structure:
-        # {"source":{"MyClass.java":{"text":"“public class ClassName: {...<code here>...}”","isVisible":true}}
-        return false unless source[key]["text"]&.force_encoding("UTF-8")&.valid_encoding?
+        # Nested file source structure for lab2 labs such as Python Lab and Web Lab 2 is
+        # {files: {filename: {contents: "...<code here>...",...}}, folders: {id: {id: <id>, name: <folder_name>,...}}}
+        if key == "files"
+          source[key].each_key do |file|
```
molly-moen: I've realized there's a bug here if a Java Lab project for some reason has a single file named "files". I'll update this.

apps/src/lab2/responseValidators.ts
thomasoniii: I think I went with ids for two reasons (I think I originally had names as well). 

The main one was just renaming the file - My structure was keying off of name _and_ storing the name as an attribute, so a name change also meant changing the key pointing to the object. Not the end of the world keeping the two things in sync, but since the id should be immutable it doesn't pop up for that case since the user can't change it.

And the other was consistency with the folder approach. Since I'd set it up so that the file object points to the folder object (as opposed to the folder containing a list of files), I'd gone with `folderId` because otherwise if the folder was renamed you'd have to go update the `folderName` attributes in all of the related files.

Note that if the data was stored the other way - where a folder contains a list of fileIds, then we'd definitely want to key off of `fileId` instead of `fileName` for the same reason - if you renamed a file, you'd also need to go update that folder's list of files to include the new name.

Anyway, to reduce the redundant updates at least one of the objects really needed an ID value, so I finally just went with including it on both of them for consistency's sake.

All that said, it's certainly open to debate and we can adjust and re-tool as desired.

thomasoniii (commented):
lgtm 🚀

apps/src/lab2/responseValidators.ts
molly-moen: That makes sense. I'll update python to use ids for file keys.

PR #57315, opened 2024-03-15
[ProgressV2] Fix adding students throwing fatal error

lfryemason:
Crash was caused by trying to get data from `studentLevelProgressByUnit` for a student that was just added, but progress data was not loaded. 

This PR:
* Filters all students that don't have progress data loaded from the table
* Async loads progress data again when student progress data is not found
* Only show skeleton if data is not loaded, even if data is getting refreshed

## Links
https://codedotorg.atlassian.net/browse/TEACH-938

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/templates/sectionProgressV2/SectionProgressV2.jsx
+++ b/apps/src/templates/sectionProgressV2/SectionProgressV2.jsx
@@ -17,12 +17,13 @@ function SectionProgressV2({
   unitData,
   isLoadingProgress,
   isRefreshingProgress,
+  isLevelProgressLoaded,
 }) {
   const [expandedLessonIds, setExpandedLessons] = React.useState([]);
 
   const levelDataInitialized = React.useMemo(() => {
-    return unitData && !isLoadingProgress && !isRefreshingProgress;
-  }, [unitData, isLoadingProgress, isRefreshingProgress]);
+    return unitData && isLevelProgressLoaded;
+  }, [unitData, isLevelProgressLoaded]);
 
   React.useEffect(() => {
     if (!unitData && !isLoadingProgress && !isRefreshingProgress) {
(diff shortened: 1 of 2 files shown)


apps/src/templates/sectionProgressV2/SectionProgressV2.jsx
```diff
@@ -17,12 +17,13 @@ function SectionProgressV2({
...
   isRefreshingProgress,
+  isLevelProgressLoaded,
 }) {
   const [expandedLessonIds, setExpandedLessons] = React.useState([]);
 
   const levelDataInitialized = React.useMemo(() => {
-    return unitData && !isLoadingProgress && !isRefreshingProgress;
-  }, [unitData, isLoadingProgress, isRefreshingProgress]);
+    return unitData && isLevelProgressLoaded;
+  }, [unitData, isLevelProgressLoaded]);
```
etaderhold, line 26: Could the `useEffect` just below also be updated to use `isLevelProgressLoaded` or is there a subtle distinction there?

etaderhold (approved):
Looks nice! Reading over the code it _looks_ like after you add a student and switch back to the progress view the table will initially show up without the new student in it, but an API call will go out right away to retrieve updated progress, and the new student will show up as soon as this API call returns. Is that accurate?

apps/src/templates/sectionProgressV2/SectionProgressV2.jsx
lfryemason: No because we don't want to send off a request if we are already loading. We could add `isLevelProgressLoaded` in addition to the three already there, but it would be redundant

lfryemason:
> Looks nice! Reading over the code it _looks_ like after you add a student and switch back to the progress view the table will initially show up without the new student in it, but an API call will go out right away to retrieve updated progress, and the new student will show up as soon as this API call returns. Is that accurate?

Yep! That's accurate. I thought that would be better than showing a student mislabeled with no progress.

Another option for future UI if this gets confusing is to add a `refresh` button that only shows up if it is out of date or a small spinner at the top when some progress is getting reloaded.

PR #57318, opened 2024-03-15
Gen AI lab: manage state for model customizations in Redux

bencodeorg:
Moves state management for model customizations for generative AI lab into Redux. Note that this doesn't actually store user updates anywhere other than Redux -- the next step is to setting up project storage to store student model customizations.

## Testing story

Tested manually that initial model customization state was loaded from levelbuilder, and that Redux was updated appropriately after interacting with each control.


bencodeorg:
Excuse the extensive commit list here, the changes here really only reflect the last 6 commits or so.

apps/src/aichat/redux/aichatRedux.ts
```diff
@@ -124,6 +133,37 @@ const aichatSlice = createSlice({
         chatMessage.status = status;
       }
     },
+    setLevelAiCustomizations: (
+      state,
+      action: PayloadAction<LevelAiCustomizations>
+    ) => {
+      state.levelAiCustomizations = action.payload;
+    },
+    setLevelAiCustomizationProperty: (
```
bencodeorg: These reducers heavily borrowed from here:

https://github.com/code-dot-org/code-dot-org/blob/18a5815741461ba0c6984129db5eee68a5275829/apps/src/lab2/levelEditors/aiCustomizations/EditAiCustomizations.tsx#L69-L99

Feel a little torn on the best way to manage all of these...`levelAiCustomizations` has some top level properties that have simple values (temperature, systemPrompt, etc), some that have a complicated value (modelCardInfo), and one that doesn't have a visibility (`hidePresentationPanel`). So, sort of a mixed bag.

apps/src/aichat/redux/aichatRedux.ts
```diff
@@ -23,13 +30,15 @@ export interface AichatState {
   showWarningModal: boolean;
   // Denotes if there is an error with the chat completion response
   chatMessageError: boolean;
+  levelAiCustomizations: LevelAiCustomizations;
```
sanchitmalhotra126: I think these can just be `AiCustomizations`. The other properties in `LevelAiCustomizations` (visibility and `hidePresentationPanel`) are effectively read only values that we can just grab from the level properties in lab2 redux when required. That hopefully might make the reducer a little simpler?

Also minor, but maybe we can call this something like `currentAiCustomizations` to reflect that it's the current state of the data in the student's customizations tabs (as opposed to what was provided by the level)?

bencodeorg: Yeah I was torn on this...I thought it was cleaner to have all of the reading from redux within each component happen from `aichatRedux`, rather than have to read from lab2 redux to get visibility for each property, then from `aichatRedux` to get/set values. You think better to read from both places in redux in the components?

Good call re: currentAiCustomizations, will update.

sanchitmalhotra126: Yeah, I think it's fine either way - agree that it would be nicer to read it from one store, but also I do think keeping them separate enforces a separation between readonly, level-specific data and mutable properties that are being actively updated by the user. Also I do think there's value in decoupling these two pieces, since it gives us a little bit more flexibility in the structure of the data as it applies to the student-facing UI vs the structure of the data that is actually saved to the project/read from the level. For now they're the same (and will probably stay the same) but hypothetically if there was some text field that was only added by the student and not read from the level or the project - that would stay in aichat redux along with these properties and wouldn't have to be added to/reconciled with the `LevelAiCustomizations` model.

bencodeorg: Ok, that sounds reasonable. Let me take a pass that uses both redux stores rather than just aichat and see how it looks.

bencodeorg:
One thing I've been thinking while doing this is that it might be nice to group all of the value/visibility ai customizations under a single key in `LevelAiCustomizations` (`modelCustomizations`?) so that we can more easily loop through it without having to always make an exception for `hidePresentationPanel`:

https://github.com/code-dot-org/code-dot-org/blob/a7e75c29165b3d6cd39b1984c65291ac172fb0f0/apps/src/aichat/types.ts#L69-L77

sanchitmalhotra126:
> One thing I've been thinking while doing this is that it might be nice to group all of the value/visibility ai customizations under a single key in `LevelAiCustomizations` (`modelCustomizations`?) so that we can more easily loop through it without having to always make an exception for `hidePresentationPanel`:

Yeah that makes sense! I had a similar thought as well looking at this again 🙂 I was thinking of maybe splitting up the visibility and values into separate objects to make it easier to key into, but I also like your idea of putting everything under one key to make it easier to iterate through. Were you thinking something like -

```
interface AiCustomizations {
  botName: string;
  ... 
}

interface LevelAiSettings (or a different name) {
  initialCustomizations: {[key in keyof AiCustomization]}: {value: AiCustomizations[key]; visibility: Visibility};
  hidePresentationPanel?: boolean;
}
```

Just for visibility, my other initial idea was something like
```
interface LevelAiSettings {
  initialCustomizations: AiCustomizations;
  visibilities: {[key in keyof AiCustomizations]}: Visibility;
  hidePresentationPanel?: boolean;
}
```
Which could also make things easier to manage (ex. wouldn't need a separate EMPTY_MODEL_CUSTOMIZATIONS and EMPTY_MODEL_CUSTOMIZATIONS_STUDENT)

apps/src/aichat/redux/aichatRedux.ts
```diff
@@ -124,6 +130,38 @@ const aichatSlice = createSlice({
...
+        property: keyof AiCustomizations;
+        value: AiCustomizations[typeof property];
+      }>
+    ) => {
+      const {property, value} = action.payload;
+      const updatedAiCustomizations: AiCustomizations = {
+        ...state.currentAiCustomizations,
+        [property]: value,
+      };
+      state.currentAiCustomizations = updatedAiCustomizations;
```
sanchitmalhotra126, line 147: nit: I think you may be able to just do `state.currentAiCustomizations[property] = value` here?

apps/src/aichat/views/AichatView.tsx
```diff
@@ -20,6 +23,23 @@ const AichatView: React.FunctionComponent = () => {
     dispatch(sendSuccessReport('aichat'));
   }, [dispatch]);
 
+  const initialAiCustomization = useAppSelector(
+    state =>
+      (state.lab.levelProperties as AichatLevelProperties | undefined)
+        ?.initialAiCustomizations || EMPTY_AI_CUSTOMIZATIONS
+  );
+  const aiCustomizations: AiCustomizations = {
```
sanchitmalhotra126: Yeah, looking at this, curious if you think it would be helpful to have the values and visibilities split up so you could just call dispatch with the values directly instead of mapping to this object?

sanchitmalhotra126 (approved):
Generally LGTM - happy to keep chatting about interface structure, but also ok if you want to get this shipped and rework the interfaces as a follow-up

apps/src/aichat/redux/aichatRedux.ts
bencodeorg: Yeah I had that originally, TS doesn't like it?

```
ERROR in ./src/aichat/redux/aichatRedux.ts:143:7
TS2322: Type 'string | number | string[] | ModelCardInfo' is not assignable to type 'never'.
  Type 'string' is not assignable to type 'never'.
    141 |     ) => {
    142 |       const {property, value} = action.payload;
  > 143 |       state.currentAiCustomizations[property] = value;
        |       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    144 |     },
    145 |     setModelCardProperty: (
    146 |       state,

Found 1 error in 4952 ms.

bencodeorg:
> > One thing I've been thinking while doing this is that it might be nice to group all of the value/visibility ai customizations under a single key in `LevelAiCustomizations` (`modelCustomizations`?) so that we can more easily loop through it without having to always make an exception for `hidePresentationPanel`:
> 
> Yeah that makes sense! I had a similar thought as well looking at this again 🙂 I was thinking of maybe splitting up the visibility and values into separate objects to make it easier to key into, but I also like your idea of putting everything under one key to make it easier to iterate through. Were you thinking something like -
> 
> ```
> interface AiCustomizations {
>   botName: string;
>   ... 
> }
> 
> interface LevelAiSettings (or a different name) {
>   initialCustomizations: {[key in keyof AiCustomization]}: {value: AiCustomizations[key]; visibility: Visibility};
>   hidePresentationPanel?: boolean;
> }
> ```
> 
> Just for visibility, my other initial idea was something like
> 
> ```
> interface LevelAiSettings {
>   initialCustomizations: AiCustomizations;
>   visibilities: {[key in keyof AiCustomizations]}: Visibility;
>   hidePresentationPanel?: boolean;
> }
> ```
> 
> Which could also make things easier to manage (ex. wouldn't need a separate EMPTY_MODEL_CUSTOMIZATIONS and EMPTY_MODEL_CUSTOMIZATIONS_STUDENT)

Yep I was thinking something like the former, although I kinda dig what you've got! I kind of tried to refactor while I was doing this and it got messy, so I think doing it as a follow-up will be easier.

apps/src/aichat/redux/aichatRedux.ts
sanchitmalhotra126: Oh huh interesting. Not sure why it's considering that type `never`, maybe something with the `WriteableDraft` type that redux toolkit creates? But anyway what you have is fine in that case!

fisher-alice:
> 
> Just for visibility, my other initial idea was something like
> 
> ```
> interface LevelAiSettings {
>   initialCustomizations: AiCustomizations;
>   visibilities: {[key in keyof AiCustomizations]}: Visibility;
>   hidePresentationPanel?: boolean;
> }
> ```
> 
> Which could also make things easier to manage (ex. wouldn't need a separate EMPTY_MODEL_CUSTOMIZATIONS and EMPTY_MODEL_CUSTOMIZATIONS_STUDENT)


> Just for visibility, my other initial idea was something like
> 
> ```
> interface LevelAiSettings {
>   initialCustomizations: AiCustomizations;
>   visibilities: {[key in keyof AiCustomizations]}: Visibility;
>   hidePresentationPanel?: boolean;
> }
> ```
> 
> Which could also make things easier to manage (ex. wouldn't need a separate EMPTY_MODEL_CUSTOMIZATIONS and EMPTY_MODEL_CUSTOMIZATIONS_STUDENT)

Just wanted to +1 the proposed idea above. It was a bit confusing at first to understand the difference between `aiCustomizations` and `initialAiCustomizations` in `AiChatView`. I first assumed they were of they same type.

PR #57322, opened 2024-03-15
INF-1141 Report metrics on activejob queue

cat5inthecradle:
Monitor the queue length and age of the oldest job in the ActiveJob Delayed Job queue.

---

These metrics help us answer the question, "Is the queue being processed?"

* If the JobCount metric passes a threshold, then our queue might be slow or stopped - BUT, the threshold depends on the number of active users triggering job creation, as well as the number and processing speed of workers, so only when the latter two variables are stable is this metric sufficient for alarms. This metric is very useful for analyzing the service, however.
* If the OldestJobAge metric gets too high, then we either have a stuck unprocessable job, or more likely, we have become unable to keep up with the queue. A threshold for this metric can be defined independently of job creation or handling rate.

## Links

- Jira: [INF-1141](https://codedotorg.atlassian.net/browse/INF-1141)
- [Delayed Job docs](https://github.com/collectiveidea/delayed_job?tab=readme-ov-file#gory-details)

## Testing story

Testing this locally, it appears to work fine, however I have a weird timezone issue on my local workstation resulting in weird numbers for the Age metric. A smattering of data points can be [seen in CloudWatch](https://us-east-1.console.aws.amazon.com/cloudwatch/home?region=us-east-1#metricsV2?graph=~(metrics~(~(~'CDO*2fActiveJob~'JobCount~'Environment~'development)~(~'.~'OldestJobAge~'.~'.~(yAxis~'right)))~view~'timeSeries~stacked~false~region~'us-east-1~start~'2024-03-15T21*3a30*3a00.000Z~end~'2024-03-15T23*3a00*3a00.000Z~stat~'Maximum~period~300)&query=~'*7bCDO*2fActiveJob*2cEnvironment*7d).

## Follow-up work

Additional metrics help build the full picture, for example:
* WaitTime could be reported on every job start
* HandlingTime could be reported on every job completion

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/bin/cron/report_activejob_metrics
+++ b/bin/cron/report_activejob_metrics
@@ -1,0 +1,57 @@
+#!/usr/bin/env ruby
+require_relative 'only_one'
+abort 'Script already running' unless only_one_running?(__FILE__)
+
+require_relative '../../dashboard/config/environment'
+
+METRICS_NAMESPACE = 'CDO/ActiveJob'
+
+def push_queue_length_metric
+  job_count = Delayed::Job.count
+
+  Cdo::Metrics.push(METRICS_NAMESPACE,
+    [
+      {
+        metric_name: 'JobCount',
+        value: job_count,
+        unit: 'Count',
+        timestamp: Time.now,
+        dimensions: [
+          {
+            name: 'Environment',
+            value: CDO.rack_env # is this the right value?
+          }
+        ]
+      }
+    ]
+  )
+end
+
+def push_oldest_job_age_metric
+  oldest_job = Delayed::Job.order(:created_at).first
+  age_in_seconds = oldest_job ? Time.now.utc - oldest_job.created_at : 0
+
+  Cdo::Metrics.push(METRICS_NAMESPACE,
+    [
+      {
+        metric_name: 'OldestJobAge',
+        value: age_in_seconds,
+        unit: 'Seconds',
+        timestamp: Time.now,
+        dimensions: [
+          {
+            name: 'Environment',
+            value: CDO.rack_env # is this the right value?
+          }
+        ]
+      }
+    ]
+  )
+end
+
+def main
+  push_queue_length_metric
+  push_oldest_job_age_metric
+end
+
+main
--- a/bin/cron/report_activejob_metrics
+++ b/bin/cron/report_activejob_metrics
@@ -4,7 +4,7 @@ abort 'Script already running' unless only_one_running?(__FILE__)
 
 require_relative '../../dashboard/config/environment'
 
-METRICS_NAMESPACE = 'CDO/ActiveJob'
+METRICS_NAMESPACE = 'code-dot-org/ActiveJob'
 
 def push_queue_length_metric
   job_count = Delayed::Job.count
--- a/cookbooks/cdo-apps/templates/default/crontab.erb
+++ b/cookbooks/cdo-apps/templates/default/crontab.erb
@@ -3,7 +3,7 @@
   # non-persistent servers. All times are UTC.
   #
   # None of these schedules, unless specifically noted, needs to change for
-  # dayling savings time.
+  # daylight savings time.
   #
   # CRONJOB TIME SYNTAX: https://help.ubuntu.com/community/CronHowto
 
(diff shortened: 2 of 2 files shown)


bin/cron/report_activejob_metrics
```diff
@@ -0,0 +1,57 @@
+#!/usr/bin/env ruby
+require_relative 'only_one'
+abort 'Script already running' unless only_one_running?(__FILE__)
+
+require_relative '../../dashboard/config/environment'
+
+METRICS_NAMESPACE = 'CDO/ActiveJob'
```
cat5inthecradle: We are not currently prefixing our custom metric namespaces with `CDO/` but I'm proposing we start. Perhaps it's redundant. Unfortunately you cannot delete metrics, so the `CDO/ActiveJob` namespace will remain in Cloudwatch for at least 15 months, until the metrics I've already pushed expire.

sureshc: I support standardizing on a namespace for custom metrics! I’d prefer something like `code-dot-org`

bin/cron/report_activejob_metrics
```diff
@@ -0,0 +1,57 @@
...
+    [
+      {
+        metric_name: 'JobCount',
+        value: job_count,
+        unit: 'Count',
+        timestamp: Time.now,
+        dimensions: [
+          {
+            name: 'Environment',
+            value: CDO.rack_env # is this the right value?
```
sureshc, line 22: I’d prefer a value that uniquely identifies a given deployment. Specifically the value set in the CloudFormation Parameter `SubdomainName` https://github.com/code-dot-org/aiproxy/blob/76eb0b0b9f1ddef78963b078343ba9a00fb622e7/cicd/3-app/aiproxy/template.yml#L14-L16

sureshc: oops … we’re over in the main `code-dot-org` repository, so `DashboardSubdomainName` is what I’d want to use https://github.com/code-dot-org/code-dot-org/blob/77d377d367f650344c56f3257049415a294b5407/aws/cloudformation/cloud_formation_stack.yml.erb#L17-L20

We don’t have an easy way to pass that value into ` CDO.` configuration setting yet. I’ve always wanted to implement `!StackParameter` directive to do that, just like our `!Secret` and `!StackSecret` systems.

The hacky way to do that now is to provision a Stack-specific Secret in the template whose value is populated from a Parameter.

sureshc: Example of using a Secret to populate a non-secret configuration setting. I recognize that this is messy 😢

https://github.com/code-dot-org/code-dot-org/blob/77d377d367f650344c56f3257049415a294b5407/aws/cloudformation/components/database.yml.erb#L88-L93

bin/cron/report_activejob_metrics
cat5inthecradle: From a usability perspective, I'm leaning acronym, but could go either way. The place where we would most often deal with the prefix is when searching for metrics. These would all have the prefix appended.
image

sureshc: Any standardized name is better than none, but also every new engineer starts looking at our codebase and needs about 5 weeks before they realize what `CDO` stands for 😂

bin/cron/report_activejob_metrics
sureshc: `CDO.rack_env` does not uniquely identify a specific environment. There are currently 3 different types of environments that have `CDO.rack_env` set to the value `test` (managed test server, an engineer’s workstation when executing unit tests, and continuous integration builds). Would all 3 types of environments publish the same Metric?

sureshc (approved):
Would be nice to solve the conundrum that `CDO.rack_env` != a specific environment, but otherwise, this is pretty straightforward.

bin/cron/report_activejob_metrics
cat5inthecradle: We're [already](https://github.com/code-dot-org/code-dot-org/blob/2fa69d96a0f7e04970bfce9b0e47f0a94b08271e/dashboard/app/jobs/evaluate_rubric_job.rb#L87) reporting the environment using `rack_env` in relation to this job processing.

I think `rack_env` is what I want, with the exception of the case where we're running tests locally.

PR #57329, opened 2024-03-16
Revert "Revert "JS integration for Statsig""

hannahbergam:
This PR represents a second attempt at a client-side Statsig integration. Though we were not calling any Statsig events anywhere, I misunderstood that just importing the StatsigReporter file into the AnalyticsReporter file would construct an instance of the class on every page sending Amplitude events, resulting in the console errors. 

Updates/improvements/additional checks:
1. There was an error in the way I was reading configuration data from the DOM. 
       Specifically, managed_test_server will result in a string value (not an actual boolean), so I'm now checking against a string and using a default boolean of false.
       I've updated the way the api key is retrieved to deconstruct the element first, and can confirm that a string value for statsig_api_client_key in my locals.yml file overrides the 'fake-client-key' (see image below). I believe that in the first attempt, we were not actually accessing the secrets as overrides, instead trying to initialize with the placeholder fake key. 
2. I updated the always_send functionality to only look at this.local_mode instead of a combination of production and staging. I've since learned that unlike Amplitude, we are not sending any Statsig events from staging. 
3. I confirmed that running this locally, the initialize() is never called, so we are free of any console errors from an incorrect client key.
4. Somewhat as an aside, I added a test file to verify the userId's will be updated (padded or prepended) matching our approach in Amplitude. It is much harder to test the constructor itself in fake environments.
5. I was also able to verify that the user properties update happens correctly from current user redux (by updating a teacher dashboard amplitude event to send to statsig after a user has logged in which would theoretically mean their user info has already been sent to statsig). See third pic below to see that the event sent with a user id!

I unfortunately cannot be 100% sure that the production environment configuration will work before launching it. Because we are not sending events from Staging either, I am unable to test events sending there. 

(There is more info in a comment below about how I tested the production key locally). 

Additionally, I was able to check that local events log to the console (see second image)

Screenshot 2024-03-18 at 10 04 25 PM

Screenshot 2024-03-19 at 3 41 36 PM

Screenshot 2024-03-20 at 12 44 45 PM


bethanyaconnor:
> I unfortunately cannot be 100% sure that the production environment configuration will work before launching it. Because we are not sending events from Staging either, I am unable to test events sending there.

Can you use a key locally and check that the configuration works? I think it's okay if some data makes it to statsig when developing the integration

apps/test/unit/lib/util/StatsigReporterTest.js
```diff
@@ -0,0 +1,32 @@
+import {expect} from '../../../util/reconfiguredChai';
+import statsigReporter from '@cdo/apps/lib/util/StatsigReporter';
+import * as utils from '@cdo/apps/utils';
+import {stub} from 'sinon';
+
+describe('AnalyticsReporter', () => {
```
bethanyaconnor: nit: should this be `StatsigReporter`?

hannahbergam: Good catch! Fixed

hannahbergam:
> > I unfortunately cannot be 100% sure that the production environment configuration will work before launching it. Because we are not sending events from Staging either, I am unable to test events sending there.
> 
> Can you use a key locally and check that the configuration works? I think it's okay if some data makes it to statsig when developing the integration

Yes- this works! I took the secrets key for production and put it into my locals.yml file, switched line 24 (local_mode) to be a double bang (making local_mode false so the event would send), and added a third param (PLATFORMS.BOTH) to the teacherHomepage event for logging in. That resulted in this!
Screenshot 2024-03-19 at 9 35 41 AM

bethanyaconnor (approved):

apps/src/templates/currentUserRedux.js
```diff
@@ -185,6 +186,13 @@ export default function currentUser(state = initialState, action) {
       user_type,
       experiments.getEnabledExperiments()
     );
+    // Calling Statsig separately to emphasize different user integrations
+    // and because dual reporting is aspirationally temporary (March 2024)
+    statsigReporter.setUserProperties(
```
hannahbergam, line 191: This function is async- will this have any impact on currentUser redux running effectively?

bethanyaconnor: This is starting to push my JS knowledge a bit but it looks like async functions implicitly [return a Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function). So, I _think_ that this is fine for this redux store.

apps/src/lib/util/StatsigReporter.js
```diff
@@ -0,0 +1,112 @@
...
+    };
+    if (!this.shouldPutRecord(ALWAYS_SEND)) {
+      this.log(
+        `User properties: userId: ${formattedUserId}, userType: ${userType}, signInState: ${!!userId}`
+      );
+    }
+    await Statsig.updateUser(user);
+  }
+
+  sendEvent(eventName, payload) {
```
nicklathe, line 57: Statsig `sendEvent` expects an event name, event value, and a metadata object (optional) ([see here](https://docs.statsig.com/client/jsClientSDK#logging-an-event)). I believe by implementing this like you have it will result in event metrics not formatting correctly in the Statsig dashboard.

nicklathe:
I'm starting to add some client side events for LTI related things, so I'm looking forward to this getting merged!

I added an unsolicited comment, discovered the issue when digging into using this client side wrapper.

apps/src/lib/util/StatsigReporter.js
hannahbergam: I am not sure how to make this match our current analytics reporting, as we've formatted all our client side reporting to send an event name and an optional block of additional params. Can we chat through this for a few minutes?

hannahbergam: We were able to resolve offline- I've left a comment in the code explaining the approach we landed on!

apps/src/templates/currentUserRedux.js
hannahbergam: I was able to move this through to fix the console errors and didn't notice any lag in loadtime. Marking resolved

PR #57330, opened 2024-03-17
Datablock Storage: don't use `_url()` helper in tests

snickell:
In tests we currently reference our controller using the `def _url()` helper:
```
  def _url(action)
    return "/datablock_storage/#{@channel_id}/#{action}"
  end
```

Bethany pointed out in https://github.com/code-dot-org/code-dot-org/pull/56279#discussion_r1526498126 that we should be able to reference the controllers directly using Rails conventions, so this PR converts:
`post _url(:create_record)` to `post :create_record`


snickell:
The result of doing this naive conversion to remove `_url` that all tests error out with messages like:
```
ERROR["test_add_shared_table_cannot_overwrite_an_existing_table", "DatablockStorageControllerTest", 4.605899999995017]
 test_add_shared_table_cannot_overwrite_an_existing_table#DatablockStorageControllerTest (4.61s)
Minitest::UnexpectedError:         URI::InvalidURIError: bad URI(is not URI?): "https://test-studio.code.org:443create_record"
            test/controllers/datablock_storage_controller_test.rb:565:in `block in <class:DatablockStorageControllerTest>'
```

So we're getting urls like `https://test-studio.code.org:443create_record`, but they should look like `https://test-studio.code.org:443/datablock_storage/:channel_id/create_record`.

`routes.rb` looks like:
```
    resources :datablock_storage, path: '/datablock_storage/:channel_id/', only: [:index] do
      collection do
        post :create_record
```

What stands out to me is:
1. How would the Rails "conventional" version know the correct channel_id to insert in the URL? If we have to pass that in, I'll probably create a _url wrapper anyway, so we're just trading off how the (currently very very simple) URL wrapper is written, which doesn't seem like a very big win.
2. The generated URL is missing even the initial `/`, which suggests to me that the `path: '/datablock_storage/:channel_id/'` in routes.rb isn't being taken into account. I think I'm concerned mostly if that has other negative implications long-term for its "rails-ness"

snickell:
Compare to projects controller, which also uses :channel_id in the URL, and looks like this in tests:
```
  test 'playlab project level has sharing meta tags' do
    # snip
    get :show, params: {key: 'playlab', channel_id: channel, share: true}
  end
```

And like this in routes.rb:
```
    resources :projects, path: '/projects/', only: [:index] do
      collection do
        # SNIP and rewrite to simplify
        get "/playlab/:channel_id", to: 'projects#show', as: "playlab_project_share", share: true
      end
    end
```

Maybe the missing bit here is the `to:` bit in routes.rb?

That said, I'm not sure its worth adding the visual noise of adding a to: line to all the routes.rb entries just to avoid writing a very simple manual `def _url` helper.... since this controllers should never have references from another controller directly anyway (should all go through datablockStorage.js), not having datablock_storage_url and friends working seems like it might not be necessary.

snickell:
Hm, not sure of what to do here, tried modifying as follows....

Test: added channel_id in params
```
  test "just create a record"  do
    post :create_record, params: {
      channel_id: @channel_id,
      table_name: 'mytable',
      record_json: {"name" => 'bob', "age" => 8}.to_json,
    }
    assert_response :success
  end
```

routes.rb, set a to: and as: clause
```
        post :create_record, to: 'datablock_storage#create_record', as: 'datablock_storage_create_record'
```

But still getting:
```
➜  dashboard git:(datablock/dont-use-_url) ✗ bundle exec spring testunit ./test/controllers/datablock_storage_controller_test.rb --name "test_just_create_a_record"
Running via Spring preloader in process 43253
Started with run options --name test_just_create_a_record --seed 47562

ERROR["test_just_create_a_record", "DatablockStorageControllerTest", 5.237676999997348]
 test_just_create_a_record#DatablockStorageControllerTest (5.24s)
Minitest::UnexpectedError:         URI::InvalidURIError: bad URI(is not URI?): "https://test-studio.code.org:443create_record"
            test/controllers/datablock_storage_controller_test.rb:78:in `block in <class:DatablockStorageControllerTest>'

  1/1: [=======================================================================================================================================================================================] 100% Time: 00:00:05, Time: 00:00:05

Finished in 6.81385s
1 tests, 0 assertions, 0 failures, 1 errors, 0 skips
```

PR #57331, opened 2024-03-17
Music: delay showing non-success validation

breville:
With this change, a Music Lab level can provide a `validationTimeout` specifying the measure at which non-success validation messages can start appearing.  That said, if this timeout has been specified but the playhead has already completed the final scheduled measure, non-success validation messages can start appearing at that point.

If the level does not provide this value, non-success validation messages can start appearing after the first measure has completed.  This creates a moment of anticipation before seeing initial feedback.

Success validation messages will continue to be shown as soon as they are available.

#### musiclab_intro_repeat

https://github.com/code-dot-org/code-dot-org/assets/2205926/3018c48b-95c4-481e-aa39-f11f93bf0e01

The initial use case is the level in which we want to give the user the opportunity to repeat three sounds before non-success validation messages are shown.  Four scenarios are demonstrated:
- If the user plays three 2-measure sounds, then they will see the success validation messages before any non-success ones have the chance to be shown.  
- If the user plays three short sounds, they will see the success validation message as soon as they start playing the third sound.
- If the user plays two short sounds, they will see the success validation message as soon as the second measure completes.  Despite the playhead not yet reaching the specified timeout value, there are no more scheduled measures to play.  
- If the user plays three very long sounds, they will see non-success validation messages shown, but not until they reach measure 6.

By limiting the sounds for this level to only be those that are 2 measures long, we could always deliver the first option.

#### musiclab_intro_trigger

https://github.com/code-dot-org/code-dot-org/assets/2205926/dd57a022-fb82-413d-9568-8ee559e88b63

Another potential use case is the level in which we want the user to play a triggered sound.  In this implementation, they will see a non-success validation message after the first measure has played, which is the default when the level does not specify a timeout.  We could build support for the level to delay the display of this message until a certain timeout, even though there are no other scheduled sounds, but I've held off on that for now.

## Related

This is an alternative to the idea sketched out in https://github.com/code-dot-org/code-dot-org/pull/57286.  In this approach, the timeout mangement is Music Lab-specific, and is inspired by earlier labs' timeout systems.

### DIFF
--- a/apps/src/lab2/progress/ProgressManager.ts
+++ b/apps/src/lab2/progress/ProgressManager.ts
@@ -76,8 +76,8 @@ export default class ProgressManager {
 
     // Go through each validation to see if we have a match.
     for (const validation of this.currentValidations) {
-      // If it's not a next validation, then make sure the lab-specific validator
-      // is ready for it.
+      // If it's a non-successful validation (i.e. validation.next is false), then
+      // make sure the lab-specific validator is ready for it.
       if (this.validator.shouldCheckNextConditionsOnly() && !validation.next) {
         continue;
       }
--- a/apps/src/lab2/progress/ProgressManager.ts
+++ b/apps/src/lab2/progress/ProgressManager.ts
@@ -7,6 +7,7 @@ import {Condition, Validation} from '@cdo/apps/lab2/types';
 // the validation works is up to the implementor.
 export abstract class Validator {
   abstract shouldCheckConditions(): boolean;
+  abstract shouldCheckNextConditionsOnly(): boolean;
   abstract checkConditions(): void;
   abstract conditionsMet(conditions: Condition[]): boolean;
   abstract clear(): void;
@@ -75,6 +76,12 @@ export default class ProgressManager {
 
     // Go through each validation to see if we have a match.
     for (const validation of this.currentValidations) {
+      // If it's not a next validation, then make sure the lab-specific validator
+      // is ready for it.
+      if (this.validator.shouldCheckNextConditionsOnly() && !validation.next) {
+        continue;
+      }
+
       if (validation.conditions) {
         // Ask the lab-specific validator if this validation's
         // conditions are met.
--- a/apps/src/music/views/MusicView.jsx
+++ b/apps/src/music/views/MusicView.jsx
@@ -118,6 +118,7 @@ class UnconnectedMusicView extends React.Component {
     this.musicValidator = new MusicValidator(
       this.getIsPlaying,
       this.getPlaybackEvents,
+      this.getValidationTimeout,
       this.player
     );
 
@@ -319,6 +320,17 @@ class UnconnectedMusicView extends React.Component {
 
   getIsPlaying = () => {
     return this.props.isPlaying;
+  };
+
+  getValidationTimeout = () => {
+    // The level can specify a desired timeout, in measures, before it starts showing
+    // non-"next" validation messages.  If a timeout is not specified, default to 2,
+    // which allows for one measure of playback.  That said, if we're already past the
+    // last measure of music, then non-"next" validation messages can start showing.
+    return Math.min(
+      this.props.levelData?.validationTimeout || 2,
+      this.sequencer.getLastMeasure()
+    );
   };
 
   getPlaybackEvents = () => {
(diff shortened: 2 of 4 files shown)


apps/src/lab2/progress/ProgressManager.ts
```diff
@@ -75,6 +76,12 @@ export default class ProgressManager {
 
     // Go through each validation to see if we have a match.
     for (const validation of this.currentValidations) {
+      // If it's not a next validation, then make sure the lab-specific validator
```
sanchitmalhotra126: nit: might not be obvious what "next" means in this context, maybe could add something like `// If it's not a passing validation (i.e. validation.next is false), then...`

apps/src/music/views/MusicView.jsx
```diff
@@ -321,6 +322,21 @@ class UnconnectedMusicView extends React.Component {
...
   };
 
+  getValidationTimeout = () => {
+    // The level can specify a desired timeout, in measures, when we can start showing
+    // non-success validation messages.  That said, if we've just completed playing the
+    // last measure before reaching that specified value, we can start showing the
+    // messages.
+    // If no timeout is specified, then we can starting showing the non-success messages
+    // at measure 2.
+    return this.props.levelData?.validationTimeout
```
sanchitmalhotra126, line 332: Can we define this on `levelProperties.validations`? That way it's co-located with other validation data and could be edited with the UI?

Edit: oh but `validations` is an array so it would mean a lot of code changes to update its structure. Fine with it here for now then, maybe we can move it as a follow-up.

apps/src/music/views/MusicView.jsx
```diff
@@ -321,6 +322,21 @@ class UnconnectedMusicView extends React.Component {
...
+      ? Math.min(
+          this.props.levelData?.validationTimeout,
+          this.sequencer.getLastMeasure()
```
sanchitmalhotra126, line 335: One thought - why not flip this, and only check validation once we've completed all measures? That way we'll always wait until all sounds have played before evaluating? 

Thinking specifically about the case you showed in your video with the longer sound ("relaxed_beat") where the timeout is 6 measures - looks like we still show the failing message after 6 measures even though the student has queued up 3 sounds, which similarly confusing to what we currently have. Would it be better to instead just wait until the last measure in that case? And instead of being a validation "timeout" this could be more of a validation "delay", i.e. "wait at least this long before evaluating". This could be helpful for the trigger level where initially the total number of measures will be very short (or 0) but by providing this delay, we'd still give the student some time to trigger a sound before showing the failing message.

That being said, it might still be useful to have a timeout but I think it would work well to use it in conjunction with a delay as well. For example, if we had some validation around using a function, but the student used a repeat block to play a sound 100 times, it would still be good to have some timeout (ex. 16 measures).

sanchitmalhotra126 (commented):
Thanks for iterating on this! I like this approach much better. One thought around a possible refinement, but generally looks good.

apps/src/music/views/MusicView.jsx
breville: Good point, and agreed that it's a good follow-up.

apps/src/lab2/progress/ProgressManager.ts
breville: Good call.  Updated.  I do wonder whether we could rename `next` to `success` everywhere, in a bigger follow-up.

apps/src/music/views/MusicView.jsx
breville: I think we could make this more elaborate, but we'll always have the tension between waiting a while but also not waiting too long.  We have a tradition (for better or worse) of timeouts in other labs (usually with countdown bar or "clock" UI elements) and while they aren't perfect, I think they make sense because we want to give users timely feedback, even if their program runs for much longer.  You're very right that we would still need a timeout, even with a delay.  I do like that here, unlike in those other labs, the program keeps running, even after we've given that (hopefully) timely feedback.

So, I'm inclined to try just adding this single parameter for now, but we can see what we learn as we (and curriculum) use it.

sanchitmalhotra126: (Discussed offline) that sounds good to me. Happy to keep iterating on this as we add more levels and validation types as well!

sanchitmalhotra126 (approved):

PR #57333, opened 2024-03-18
Music: new share UI

breville:
Initial work on a new project share dialog UI for Music Lab.

It is a new implementation of the project share dialog, specifically for Lab2.  This initial version for Music Lab just has a "copy to clipboard" button, and a bigger QR code.  The clipboard icon gets a checkmark when the button is clicked. 

Screenshot 2024-03-18 at 6 38 51 PM

## Related

A previous attempt at this work can be found at https://github.com/code-dot-org/code-dot-org/pull/57224, but it attempted to mostly reuse existing code because the legacy UI covers a lot of variants via parameters.  Instead, we're going for a new implementation, but might need to add support for some of those variants in the future.


apps/src/lab2/views/dialogs/ShareDialog.tsx
```diff
@@ -0,0 +1,78 @@
...
+      type="button"
+      onClick={handleCopyToClipboard}
+      className={moduleStyles.copyToClipboard}
+    >
+      <FontAwesomeV6Icon
+        iconName={copiedToClipboard ? 'clipboard-check' : 'clipboard'}
+        iconStyle="thin"
+        className={moduleStyles.copyToClipboardIcon}
+      />
+      Copy link to project
```
sanchitmalhotra126: Loc for this?

apps/src/lab2/views/dialogs/ShareDialog.tsx
```diff
@@ -0,0 +1,78 @@
...
+import QRCode from 'qrcode.react';
+import copyToClipboard from '@cdo/apps/util/copyToClipboard';
+import {hideShareDialog} from '@cdo/apps/code-studio/components/shareDialogRedux';
+
+const CopyToClipboardButton: React.FunctionComponent<{shareUrl: string}> = ({
+  shareUrl,
+}) => {
+  const [copiedToClipboard, setCopiedToClipboard] = useState(false);
+
+  const handleCopyToClipboard = () => {
```
sanchitmalhotra126: nit: `useCallback`

apps/src/lab2/views/dialogs/ShareDialog.tsx
```diff
@@ -0,0 +1,78 @@
...
+        iconName={copiedToClipboard ? 'clipboard-check' : 'clipboard'}
+        iconStyle="thin"
+        className={moduleStyles.copyToClipboardIcon}
+      />
+      Copy link to project
+    </button>
+  );
+};
+
+const ShareDialog: React.FunctionComponent<{shareUrl: string}> = ({
```
sanchitmalhotra126, line 43: Mind adding some docs indicating that this is a new refreshed share dialog that supports a subset of share features currently?

apps/src/lab2/views/dialogs/ShareDialog.tsx
```diff
@@ -0,0 +1,78 @@
...
+    () => dispatch(hideShareDialog()),
+    [dispatch]
+  );
+
+  return (
+    <FocusLock>
+      <div className={moduleStyles.dialogContainer}>
+        <div id="share-dialog" className={moduleStyles.shareDialog}>
+          <Typography semanticTag="h1" visualAppearance="heading-lg">
+            Share Project
```
sanchitmalhotra126: loc? I'm guessing we have a localized string for this already?

sanchitmalhotra126 (approved):
Nice, thanks for iterating on this! Just a few minor comments

apps/src/lab2/views/dialogs/ShareDialog.tsx
breville: I was going to hold off on loc since we are expecting a design pass on this dialog soon, but both strings are already available, so will do so!

apps/src/lab2/views/dialogs/ShareDialog.tsx
breville: Good call.

apps/src/lab2/views/dialogs/ShareDialog.tsx
breville: Will do.

apps/src/lab2/views/dialogs/ShareDialog.tsx
breville: Done.

PR #57337, opened 2024-03-18
[Google Blockly] only use cache for dropdown fields on actual behavior blocks

mikeharv:
@onlinecsteacher reported on Slack that on level edit pages we were showing a bad preview for behavior dropdown fields like this green one:

Screenshot 2024-03-15 at 4 20 55 PM

This block looks just like the `behaviorPicker` which dynamically creates a menu of options based on behavior definition blocks found on the hidden workspace. While the field name is `'BEHAVIOR'`, this is not the same block. In fact, there are several custom blocks with fields named like this that are standard (not dynamic) dropdown fields. The options on these blocks look like `['wandering', 'new Behavior(wandering, [])']`
Because of some oversimplified logic, we were calling `getOptions` with `useCache = true` on these blocks, which had some additional logic only intended for behavior picker blocks. With these other blocks, because the hidden workspace wasn't being loaded on edit pages, we'd call `options.push([newValue, newValue])` which meant the block would get an option like `['new Behavior(wandering, [])', 'new Behavior(wandering, [])']`. This is why levelbuilders were seeing the actual field value instead of its user-friendly label. 

I updated the logic to only do any of this for blocks that have the correct type.
Now the field shows the expected label:

image

## Follow-up work

`behaviorPicker` is a custom input type that can theoretically be added to any block. Today it is only used on one block, but if that changes, we'd have to strengthen this revised logic. Because levelbuilders otherwise have the ability to make new blocks, we may want to address this sooner than later. For example, if a levelbuilder cloned the block into another pool it wouldn't work quite right.

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/blockly/addons/cdoFieldDropdown.ts
+++ b/apps/src/blockly/addons/cdoFieldDropdown.ts
@@ -37,12 +37,13 @@ export default class CdoFieldDropdown extends GoogleBlockly.FieldDropdown {
    * by quotes in xml.
    */
   doClassValidation_(newValue?: string) {
+    const sourceBlock = this.getSourceBlock();
     if (newValue === EMPTY_OPTION) {
       return newValue;
     } else {
       // For behavior picker blocks, we need to regenerate menu options each time,
       // in case a behavior has been renamed.
-      const useCache = this.name !== 'BEHAVIOR';
+      const useCache = sourceBlock?.type === 'gamelab_behaviorPicker';
       for (const option of this.getOptions(useCache, newValue)) {
         if (option[1] === newValue) {
           return newValue;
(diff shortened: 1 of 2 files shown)


molly-moen (approved):
General question:

> Today it is only used on one block, but if that changes, we'd have to strengthen this revised logic.

How would we strengthen this logic?

mikeharv:
> General question:
> 
> > Today it is only used on one block, but if that changes, we'd have to strengthen this revised logic.
> 
> How would we strengthen this logic?

I think there's a couple options...
1. Low lift - Add multiple block types to the check. The downside here is that levelbuilders would need eng support to use this custom input type, so it's not really a longterm solution.
2. Add some new state to the field if it's a behavior picker and look for that instead. We could probably pass it in as `config`, but it may require defining our interface for that object to satisfy TS.
3. Extend `CdoFieldDropdown` for behavior picker fields, eg `CdoFieldBehaviorPicker`. This class could have whatever unique logic we want and we'd only use it where we actually want it.

After writing these out, I'd probably vote for option 3.

PR #57343, opened 2024-03-18
Drop users_view SQL View to Unblock Aurora 3 / MySQL 8 Upgrade

sureshc:
https://codedotorg.atlassian.net/browse/INF-1201

In some existing environments, it isn’t possible to upgrade the database cluster from Aurora Engine 2.x / MySQL 5.7 to Aurora Engine 3.x / MySQL 8 due to issues with the `users_view` SQL View. During pre-upgrade checks on some environments (such as the staging database), Aurora is noting corruption in the View and blocking the MySQL upgrade.

It appears this View is no longer being utilized. It was created to assist with the transition from only supporting password authentication to supporting multiple forms of authentication (such as OAuth via Google, Microsoft, etc.) to make it easier to lookup the email address of a user, particularly from the the original Contact Rollups job.

Drop this View to unblock the Aurora / MySQL upgrade.

## Follow-up work

Remove the `scenic` gem.

## PR Checklist:

- [ ] Code is well-commented


sureshc:
Looking for feedback from @code-dot-org/platform since this is a change to how our User and UserAuthentication models are accessed.

Hamms (commented):
Probably worth also removing the Scenic config that this view was created from: https://github.com/code-dot-org/code-dot-org/blob/staging/dashboard/db/views/users_view_v01.sql

sureshc:
> Probably worth also removing the Scenic config that this view was created from: https://github.com/code-dot-org/code-dot-org/blob/staging/dashboard/db/views/users_view_v01.sql

Agreed. We need the gem to execute the `drop_view` method in this Pull Request, but as soon as the View is gone we should remove the gem. Creating a follow up Pull Request to drop the `scenic` gem.

Hamms:
I'm not referring to the gem, which I agree should be removed in a follow-up PR. I'm referring to the `users_view`-specific config file that the gem used to create this view. Does that still need to be present in order for `drop_view` to be executed successfully, or is it only necessary for the table to be created?

sureshc:
> I'm not referring to the gem, which I agree should be removed in a follow-up PR. I'm referring to the `users_view`-specific config file that the gem used to create this view. Does that still need to be present in order for `drop_view` to be executed successfully, or is it only necessary for the table to be created?

Oops. Yes. I was so ready for someone to suggest dropping the `scenic` gem that I did’t actually read the full text of your comment 😐

I’ve deleted that file!

Hamms (approved):
LGTM from an infrastructure perspective! But I'm likewise interested in also getting this looked at from a platform perspective.

carl-codeorg (approved):
LGTM

wilkie (commented):
Happy to see the deletions! Thank you! 🎉 

(Didn't know this existed)

PR #57344, opened 2024-03-18
[Google Blockly] extend dropdown field class for behavior picker input

mikeharv:
Earlier this week, we fixed a bug that stemmed from custom dropdown logic for behaviors. We have a handful of static dropdown fields that list some behaviors as well as one custom `behaviorPicker` input. Each of these use the name `BEHAVIOR` for their field, which we were incorrectly using to determine how we should generate the menu of options. That was fixed here:
* https://github.com/code-dot-org/code-dot-org/pull/57337

As of the merging of the branch above, we now look at the actual block type. This works for now because only one block type uses the `behaviorPicker` and that block only has one dropdown field. However, this means that levelbuilders would be unable to reuse this custom input type for other new blocks without an engineer. This isn't a precedent we want to set.

To handle this better, we can extend the `CdoFieldDropdown` class and be sure to use it only where needed. This way, behavior picker fields are technically different than dropdown fields and we don't need to make any attempt to guess how they should work based upon block configuration. This would allow a levelbuilder to create a new behavior picker block with any type and any field name they choose.

After extending the class, I was able to remove a significant amount custom logic from the base class, including the entire `getOptions` override.

There are no UI changes here.

## Testing story

I checked the original behavior picker block, several behavior dropdown field blocks, and a few other random blocks with dropdown fields and didn't notice any regressions.

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/blockly/addons/cdoFieldBehaviorPicker.ts
+++ b/apps/src/blockly/addons/cdoFieldBehaviorPicker.ts
@@ -1,0 +1,86 @@
+import CdoFieldDropdown from './cdoFieldDropdown';
+import {
+  FieldDropdownConfig,
+  FieldDropdownValidator,
+  MenuGeneratorFunction,
+} from 'blockly/core';
+import {EMPTY_OPTION} from '../constants';
+
+type CustomMenuGenerator = CustomMenuOption[] | MenuGeneratorFunction;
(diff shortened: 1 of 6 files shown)


molly-moen (approved):
Nice fix!

PR #57346, opened 2024-03-18
[Google Blockly] disable hidden variable blocks

mikeharv:
On Friday, @amb-b reported an issue with validation code in some levels where students use variables in their programs.

> Seems like there's something wrong with the validation for [this level](https://levelbuilder-studio.code.org/s/coursef-2023/lessons/10/levels/3). The correct code will not pass this level. 

For example, this code would not pass: 
image

The criterion command `anySpeechIncludesValues` was always returning false. This was due to a problem in generated code from the hidden definition workspace. Levels like this one included "invisible" `variables_set` block which were used to pre-populate the dropdown field of visible variable blocks. These were recently moved to the hidden workspace because Google Blockly doesn't support invisible blocks: https://github.com/code-dot-org/code-dot-org/pull/56851

Unfortunately, our logic to disable orphaned blocks like these was not running on the hidden definition workspace. We didn't initially think we'd need this because that workspace only held procedure definitions which should never be disabled. Because these stray blocks were suddenly enabled, their generated code was being included. This caused the JS Interpreted to re-initialized (and set to 0) the variables set on the invisible blocks. This meant that the validation code would always assume the values were 0.

Normally, you'd still be able to pass this level with variables set to 0. However, I recently introduced a small regression when refactoring this criterion command last week: https://github.com/code-dot-org/code-dot-org/pull/57232/files#r1529108384

Timeline and Student Experience:

1. **Before Sprite Lab migrated** - Things working as normal
2. **After migrating Sprite Lab** - Unexpected `variables_set` blocks shown on workspace. Validation working as expected as the blocks were disabled.
3. **After hiding blocks (~3 weeks ago)** - Generated code would have been re-initializing variables. Validation likely would have been passing students as `0` is still a valid value.
4. **After refactoring code (~4 days ago)** - Generated code still re-initializing variables, but `0` is now incorrectly rejected as a valid value. Levels cannot be passed.
5. **Quick Fix (Friday)** - All students will pass this criterion automatically

The good news is that this only became a bug that blocked student learning when the criterion command refactor was merged late Thursday. With this branch, the validation and generated code should both work as expected. 

Tested below with both `0` and `3` (green bar represents 'passing' state):
imageimage

Criterion once again fails students (purple bar) if they don't use the variable in their `say` block:
image

Relevant Code not shown in diff:
* `eventHandlers.ts` -> `updateBlockEnabled()`: Enables blocks if their parent block is enabled or if they don't have output/previous connections (e.g. events/definitions). Disabled blocks if they have no parent/a disabled parent and they have an output/previous connection.
* `eventHandlers.ts` -> `disableOrphans()`: Calls `updateBlockEnabled` on blocks that have been moved/created

In other words, this change should only disable blocks on the hidden workspace if they are not top blocks. I confirmed that we still generate code for behaviors and function, which also still work as expected.

## Links

Jira - https://codedotorg.atlassian.net/browse/CT-409
Slack thread: https://codedotorg.slack.com/archives/C03DBDN67B7/p1710521309576259
Zendesk: https://codeorg.zendesk.com/agent/tickets/488670

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/p5lab/spritelab/commands/criterionCommands.js
+++ b/apps/src/p5lab/spritelab/commands/criterionCommands.js
@@ -79,25 +79,21 @@ export const commands = {
   // like to give students the flexibility of using the value from either the
   // current or previous frame.
   anySpeechIncludesValues(currentVariables, previousVariables) {
-    // const spriteIds = this.getSpriteIdsInUse();
-    // const values = Object.values(currentVariables).concat(
-    //   previousVariables ? Object.values(previousVariables) : []
-    // );
+    const spriteIds = this.getSpriteIdsInUse();
+    const values = Object.values(currentVariables).concat(
+      previousVariables ? Object.values(previousVariables) : []
+    );
 
-    // for (const spriteId of spriteIds) {
-    //   const speechText = this.getLastSpeechBubbleForSpriteId(spriteId)?.text;
-    //   if (
-    //     speechText &&
-    //     values.some(value => `${speechText}`.includes(`${value}`))
-    //   ) {
-    //     return true;
-    //   }
-    // }
-    // return false;
-
-    // TODO: Reenable this once we get this validation working again
-    // https://codedotorg.atlassian.net/browse/CT-409
-    return true;
+    for (const spriteId of spriteIds) {
+      const speechText = this.getLastSpeechBubbleForSpriteId(spriteId)?.text;
+      if (
+        speechText &&
+        values.some(value => `${speechText}`.includes(`${value}`))
+      ) {
+        return true;
+      }
+    }
+    return false;
   },
 
   // Return true if exactly one sprite began speaking.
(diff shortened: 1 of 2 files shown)


apps/src/p5lab/spritelab/commands/criterionCommands.js
```diff
@@ -79,25 +79,21 @@ export const commands = {
...
-    return true;
+    const spriteIds = this.getSpriteIdsInUse();
+    const values = Object.values(currentVariables).concat(
+      previousVariables ? Object.values(previousVariables) : []
+    );
+
+    for (const spriteId of spriteIds) {
+      const speechText = this.getLastSpeechBubbleForSpriteId(spriteId)?.text;
+      if (
+        speechText !== undefined &&
```
mikeharv, line 90: This line was explicitly changed after the revert.

molly-moen (approved):
🎉

PR #57352, opened 2024-03-18
re-raise exceptions with added details

cat5inthecradle:
`exception.message` is a frozen string, and cannot be appended to. Fix all instances where we were doing this by building a new exception object.

## PR Checklist:

- [ ] Code is well-commented


sureshc (approved):

bin/i18n/utils/crowdin_client.rb
```diff
@@ -215,15 +215,17 @@ def download_translation(source_file_id, language_id, dest_path, etag: nil)
...
-          exception.message << "\nSourceID: #{source_file_id}"
-          exception.message << "\nLangID:   #{language_id}"
-          exception.message << "\nDestPath: #{dest_path}"
-          exception.message << "\nEtag:     #{etag.inspect}"
-          exception.message << "\nResponse: #{translation_build.inspect}"
-          exception.message << "\nAttempt:  #{attempt}"
-
-          raise exception
+          # re-raise the exception with added details
+          modified_message = exception.message
```
Hamms, line 219: Does this work as-is, or do we need to do something like 

```suggestion
          modified_message = exception.message.dup
```

cat5inthecradle: Interestingly, it seems like the frozen-ness of `exception.message`  varies depending on whether it was created with a frozen string or not. This PR may not be necessary in all cases (secrets.rb is the only one that I've manually exercised).

After an offline discussion, we both agreed that in general one shouldn't be manipulating the exception object, so I think this PR still makes sense, even if the other two files touched are necessarily experiencing the same problem.

Hamms (approved):

PR #57356, opened 2024-03-18
Create & Update ParameterGroups in Preparation for Aurora 3 / MySQL 8 Upgrade

sureshc:
https://codedotorg.atlassian.net/browse/INF-1201

In preparation for upgrading our database from Aurora Engine 2.x / MySQL 5.7 to Aurora Engine 3 / MySQL 8.x:

- create a new Aurora3 Cluster ParameterGroup
- move all instance settings to the cluster ParameterGroups
- stop specifying a ParameterGroup on each Database Instance. Existing clusters will continue using the Instance ParameterGroup they were assigned when initially provisioned, but new clusters will use the default DB Instance ParameterGroup, but that does not impact the configuration of the Instance because the settings specified in the Cluster ParameterGroup as the same.

Testing with setting the Instance ParameterGroup Name on the cluster to apply to all instances within the cluster (as [recommended here](https://github.com/aws-cloudformation/cloudformation-coverage-roadmap/issues/896#issuecomment-1514031406)) yielded some unexplained results, such as a new cluster where the instances used the default ParameterGroup. 

## Testing story

1. Provision an adhoc from a feature branch off of the staging branch:  `code-dot-org % bundle exec rake adhoc:start RAILS_ENV=adhoc DATABASE=yes`
2. Edit `lib/rake/stack.rake` to hardcode `database: true` directly on the adhoc so that subsequent builds of new commits on the adhoc preserve the Aurora cluster
3. Merge this Pull Request into the adhoc's feature branch, and ensure build completes successfully and monitor database Events

Result: The reader database instance will be restarted when the commits on this Pull Request are applied to any existing environment (such as production, levelbuilder, test, and staging), but not the writer.

Reader instance restarted:
DF5992D0-89D2-4F90-BD13-B7663D3F1C48_1_201_a

Writer Instance NOT restarted:

A67ACCF8-5EE9-43A8-A4BF-90D21C2B08AD_1_201_a

But the ParameterGroup modification is not applied yet to the writer:
image

## Deployment strategy

To minimize the impact of instance restarts, deploy this change in a 2nd production release during low usage time periods.

1. Monitor production performance and errors while the CloudFormation Stack is updated. It's likely that the production database reader instance will restart to apply the new changes to the cluster ParameterGroup
2. manually reboot the writer if needed to Sync the ParameterGroup settings
3. verify new web application server instances can launch after this change is deployed.

## Follow-up work

Merge #57815 to explicitly set `DBParameterGroupName` on the `AWS::RDS::DBInstance` to the default ParameterGroup  and manually configure the production database instances to use the default ParameterGroup

## PR Checklist:

- [ ] Code is well-commented


aws/cloudformation/components/database.yml.erb
```diff
@@ -280,7 +328,7 @@
     Type: AWS::RDS::DBCluster
     Properties:
       DBClusterIdentifier: !Sub "${AWS::StackName}-cluster"
-      DBClusterParameterGroupName: !Ref AuroraClusterDBParameters
+      DBClusterParameterGroupName: !Ref Aurora2ClusterDBParameters
```
cat5inthecradle: [update requires some interruptions](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-rds-dbinstance.html#cfn-rds-dbinstance-dbparametergroupname)

Is that okay/expected?

sureshc: oh, thanks for catching that. We’re hopefully not actually changing the Physical resource that’s being assigned to the cluster and instance (which would happen if we changed the [DBParameterGroupName](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-rds-dbparametergroup.html#cfn-rds-dbparametergroup-dbparametergroupname) attribute), but it’s not worth the risk. I’ll leave the existing ParameterGroup name unchanged. Having an `Aurora3…` name for the new Resources might be a little confusing, but will totally make sense when we upgrade to Aurora Engine 4!

sureshc: Note in the Database Instance Event log when it applied the fix for this issue (keeping the existing ParameterGroup Logical ID and Physical ID) that each instance in the cluster restarted, which is NOT good. Glad you caught this.

image

sureshc:
@cat5inthecradle I've got some new logic to support the Aurora 3 / MySQL 8 upgrade (declaring the Instance Parameter Group on the cluster). Can you review that?

aws/cloudformation/components/database.yml.erb
```diff
@@ -281,6 +329,8 @@
     Properties:
       DBClusterIdentifier: !Sub "${AWS::StackName}-cluster"
       DBClusterParameterGroupName: !Ref AuroraClusterDBParameters
+      # Use same ParameterGroup for writer and all readers so that any reader can be promoted to writer during a Failover.
+      DBInstanceParameterGroupName: !Ref AuroraWriterDBParameters
```
sureshc: Does NOT trigger restart https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-rds-dbcluster.html#cfn-rds-dbcluster-dbinstanceparametergroupname

aws/cloudformation/components/database.yml.erb
```diff
@@ -303,8 +353,6 @@
     Type: AWS::RDS::DBInstance
     Properties:
       DBInstanceIdentifier: !Sub "${AWS::StackName}-<%=i%>"
-      # Use same ParameterGroup for writer and all readers so that any reader can be promoted to writer during a Failover.
-      DBParameterGroupName: !Ref AuroraWriterDBParameters
```
cat5inthecradle, line 307: Looks like this could cause [interruptions](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-rds-dbinstance.html#cfn-rds-dbinstance-dbparametergroupname) too? Are we thinking it won't because the cluster will apply the same params?

sureshc: Yes. I ran this test. Provisioned an adhoc on a feature branch that started out from the `staging` branch and then merged this changed into that feature branch. The adhoc Stack applied the changes, but did NOT restart any of the database instances.

adhoc created from staging branch 15:30 PM PT
adhoc updated from this branch 20:20 PM PT

cluster events
image


No instance shutdown/restart events on either database instance when the ParameterGroup changes were applied ~20:30 PM PT

image

cat5inthecradle (approved):

aws/cloudformation/components/database.yml.erb
sureshc: hmmmm ... the instances were provisioned with the default ParameterGroup

image

aws/cloudformation/components/database.yml.erb
```diff
@@ -148,14 +157,48 @@
       Parameters:
         # BEGIN: Static configuration settings. Changing these settings will trigger a restart of each database instance.
         performance_schema: 1
+        # END: Static configuration settings.
+
+        # BEGIN: Dynamic configuration settings. Changing these settings does NOT require nor trigger restart of instances.
+
+        # Prevent a slow query from degrading instance or cluster performance.
+        # https://dev.mysql.com/doc/refman/8.0/en/server-system-variables.html#sysvar_max_execution_time
+        max_execution_time: 30000
```
unlox775-code-dot-org, line 166: It looks like we are removing the last place referencing this parameter group, is there a reason we are updating the max_execution_time on it?

cat5inthecradle (approved):
After discussion, we believe the risk here is that the instances might reboot when the parameter group is edited. We'll target a low usage time over the weekend to make the change, and will monitor the process.

aws/cloudformation/components/database.yml.erb
cat5inthecradle: Suresh noted in the description, but existing instances will continue to use the parameter group they were created with. These changes here will modify the parameter group being used.

This PR is doing three things things.
1. editing a some settings in the in-use Aurora 2 parameter group
2. creating a new Aurora 3 parameter group
3. removing the per-instance specification of a parameter group (a no-op on existing instances)

A follow up PR will
4. specify the default parameter group for new instances via the cluster

@sureshc This PR description talks a lot about item 3, but all the risk and service interruption is coming from item 1, right?

PR #57357, opened 2024-03-18
Instrument Cookie-based Session Storage

Hamms:
In preparation for moving our session storage out of the user's browser and into some centralized data store, we'd like to get an idea of just how many active sessions we actually have, and of how much data is being stored in them.

To that end, this PR adds some simple instrumentation to the read, write, and delete operations on the CookieStore which will publish metrics to AWS CloudWatch.

## Testing story

Did some manual testing locally; also did some testing on an adhoc, both regular manual testing and automated stress testing. For manual testing, I logged in, logged out, logged back in, and viewed a couple of levels with callouts to add data to the cookie. For automated testing, I simply wrapped `curl -s https://adhoc-instrumented-cookie-store-studio.cdn-code.org/users/sign_in > /dev/null &` in a couple of `for` loops to hammer the server with a bunch of anonymous requests, each of which created a new temp session. Verified with manual testing that we can track changes in cookie size over time. Verified with automated testing that the server can hold up to a reasonable load.

I decided not to add a unit test for this new `InstrumentedCookieStore` class; because it's intended to be a temporary addition while we collect a little bit of data, a test felt unnecessary. Let me know if you think one would be valuable, and I'd be happy to add it!

## Follow-up work

Plan to let this run for a week or so to collect data, then either this PR can be reverted or we can remove this functionality as part of our switch to whatever new data store we end up settling on.

## PR Checklist:

- [ ] Code is well-commented


snickell:
We might want to be a little careful even instrumenting this with the full firehose.... this is potentially (?) a LOT of metric data points compared to our current load. You'd know better than me if that's a concern here, but we potentially check the session a LOT. A trick I've used in the past is to take a big stream like this and get useful data, is to just sample it, e.g. flip a random number, and only store 1/100th of the time.

Hamms:
Oh, for sure; what I have here is not only wildly impractical, it doesn't even achieve my end goal of giving us a sense of how much data will be stored, because it doesnt distinguish between new session data and existing session data. Right now, this is just me messing around with some ideas

dashboard/config/custom_session_store/instrumented_cookie_store.rb
```diff
@@ -0,0 +1,45 @@
...
+  module Session
+    # Simple wrapper around the standard Rails decentralized cookie-based
+    # session store that we currently use in production. The primary goal is to
+    # collect some data about current usage patterns, to inform our decisions
+    # about a possible switch to a centralized session store.
+    #
+    # TODO infra: remove this helper after we've gathered enough data.
+    class InstrumentedCookieStore < CookieStore
+      METRICS_NAMESPACE = 'code-dot-org/ActionDispatch'.freeze
+      METRICS_LOG_WINDOW = 1.minute
```
Hamms, line 15: Super open to feedback on what this window should be; one data point per frontend per minute is a pretty conservative rate of requests being put into the buffer, and does give the `@session_info` in-memory object enough time to theoretically grow rather large. Speeding up to one data point per frontend per second might be a better balance, but it also might be unnecessarily fast. What do y'all think?

dashboard/config/custom_session_store/instrumented_cookie_store.rb
```diff
@@ -0,0 +1,45 @@
...
+      private def log_session_info(session_id, session_data)
+        @session_info ||= {}
+        @session_info_last_logged_at ||= Time.now
+
+        session_size = ObjectSpace.memsize_of(session_data)
+        @session_info[session_id.private_id] = [session_size, @session_info.fetch(session_id.private_id, 0)].max
+
+        if @session_info_last_logged_at < Time.now - METRICS_LOG_WINDOW
+          num_sessions = @session_info.size
+          Cdo::Metrics.put(METRICS_NAMESPACE, 'num_sessions_seen', num_sessions, {Environment: CDO.rack_env})
```
Hamms: Something I considered is actually adding the configured window length as a dimension, so if at some point in the future we want to change the frequency of logging, we won't be confused by the dramatic change in the number of sessions seen during that window.

```suggestion
          Cdo::Metrics.put(METRICS_NAMESPACE, 'num_sessions_seen', num_sessions, {Environment: CDO.rack_env, MetricsLogWindow: METRICS_LOG_WINDOW})
```

I ultimately didn't bother because it felt a bit unnecessary; let me know if you think it could be valuable!

dashboard/config/custom_session_store/instrumented_cookie_store.rb
Hamms: Some back-of-the napkin math; according to [our Requests/Server logs](https://us-east-1.console.aws.amazon.com/cloudwatch/home?region=us-east-1#metricsV2?graph=~(metrics~(~(~(expression~'requests*2fservers~label~'Requests*20*2f*20Server~id~'efficiency~yAxis~'left~period~60~region~'us-east-1))~(~'AWS*2fEC2~'CPUUtilization~'AutoScalingGroupName~'Frontends-autoscale-prod~(id~'cpu~label~'CPU*20Utilization~region~'us-east-1~yAxis~'right~visible~false))~(~'AWS*2fApplicationELB~'ActiveConnectionCount~'LoadBalancer~'app*2fautoscale-prod*2f587ee8f918e6fd28~(id~'active_connections~label~'ActiveConnections~yAxis~'right~visible~false~region~'us-east-1))~(~'.~'RequestCount~'.~'.~(stat~'Sum~id~'requests~label~'Requests~visible~false~region~'us-east-1))~(~'.~'TargetResponseTime~'.~'.~(label~'Latency~yAxis~'right~id~'latency~color~'*23ff7f0e~region~'us-east-1~stat~'p95))~(~'AWS*2fAutoScaling~'GroupInServiceInstances~'AutoScalingGroupName~'Frontends-autoscale-prod~(id~'servers~yAxis~'right~label~'Web*20Application*20Servers~visible~false~region~'us-east-1)))~period~60~stat~'Average~region~'us-east-1~view~'timeSeries~stacked~false~title~'Web*20Application*20Server*20Efficiency~yAxis~(left~(min~0)~right~(min~0))~start~'-PT336H~end~'P0D)), it looks like we're usually dealing with 15-20k requests per minute per server during normal usage, peaking at around 25k requests. If the average user session is responsible for 10-20 requests in a one-minute period, that's about 1-2k unique sessions being logged in each time window on a given server, which seems reasonable to me.

dashboard/config/custom_session_store/instrumented_cookie_store.rb
```diff
@@ -0,0 +1,45 @@
...
+          average_size = @session_info.values.sum / num_sessions
+          Cdo::Metrics.put(METRICS_NAMESPACE, 'average_max_session_size', average_size, {Environment: CDO.rack_env})
```
cat5inthecradle: Tiniest of nits: CloudWatch metrics typically use `CamelCase` and `CamelCase_Modifier` for metric names, with a few very rare examples of `First_word_caps_snake_case`. No practical problem here, and I can think of no scenarios where anyone programatically loads metric names to the point where you'd need consistency.

dashboard/config/custom_session_store/instrumented_cookie_store.rb
snickell: I don't hate a minute, at max-4kb per session (thanks cookies!), even if it were one-user-one-request, we're still only looking at 100MB session data in server RAM before its flushed, and probably more like 5-10MB based on your napkin math. Feels safe in both directions, so ideal.

dashboard/config/custom_session_store/instrumented_cookie_store.rb
```diff
@@ -0,0 +1,45 @@
...
+require 'objspace'
+
+module ActionDispatch
+  module Session
+    # Simple wrapper around the standard Rails decentralized cookie-based
+    # session store that we currently use in production. The primary goal is to
+    # collect some data about current usage patterns, to inform our decisions
+    # about a possible switch to a centralized session store.
+    #
+    # TODO infra: remove this helper after we've gathered enough data.
```
snickell: I'd lowkey prefer to see this TODO link back to a "remove this code" PR, even if its an unstarted PR for now. You're probably more rememberful than I am, but I discovered a few of these in my career that I'd forgotten when something urgent came up, and one of them was a regretful waste of resources for a couple years.

dashboard/config/custom_session_store/instrumented_cookie_store.rb
snickell: I'm betting on us getting enough info from one sampling, don't overbuilds it says I.

dashboard/config/custom_session_store/instrumented_cookie_store.rb
```diff
@@ -0,0 +1,45 @@
...
+    class InstrumentedCookieStore < CookieStore
+      METRICS_NAMESPACE = 'code-dot-org/ActionDispatch'.freeze
+      METRICS_LOG_WINDOW = 1.minute
+
+      # @see https://github.com/rails/rails/blob/ac87f58207cff18880593263be9d83456aa3a2ef/actionpack/lib/action_dispatch/middleware/session/cookie_store.rb#L104-L107
+      private def write_session(req, session_id, session_data, options)
+        log_session_info(session_id, session_data)
```
snickell, line 19: I don't understand enough about how the whole flow here works to evaluate this tradeoff.... but since `log_session_info` is coming before `super()`, what would happen if godsforbid some unknown unknowns type exceptions starts triggering in log_session_info? Obviously eating a silent failure to flush the logs (and lower the server RAM) would be not great too, but at least we can reason pretty clearly about those impacts (RAM gets full, OOM goes to town, new instance is maybe scaled up?), and we can watch the RAM.

On the flipside, if something interfered with durability of our session cookies, even in obscure conditions, it could be mildly catastrophic, and relatively hard to predict the behavior of.

My instinct would be to want log_session_info wrapped in a catch-all rescue block, so the real write_session is basically guaranteed to be called as usual.

snickell: See below for maybe a better variation on this idea: include a rescue that covers all of log_session_info, and clear @session_info in the rescue to avoid RAM filling up too.

dashboard/config/custom_session_store/instrumented_cookie_store.rb
```diff
@@ -0,0 +1,45 @@
...
+
+        if @session_info_last_logged_at < Time.now - METRICS_LOG_WINDOW
+          num_sessions = @session_info.size
+          Cdo::Metrics.put(METRICS_NAMESPACE, 'num_sessions_seen', num_sessions, {Environment: CDO.rack_env})
+          average_size = @session_info.values.sum / num_sessions
+          Cdo::Metrics.put(METRICS_NAMESPACE, 'average_max_session_size', average_size, {Environment: CDO.rack_env})
+          @session_info = {}
+          @session_info_last_logged_at = Time.now
+        end
+      end
```
snickell, line 53:
```suggestion
      rescue => e
        # Something went wrong logging. Lets fail in a way that lets write_session
        # continue uninterrupted, and doesn't fill our server's RAM up to failure.
        @session_info = {}
      end
```

snickell (approved):
Looks good! One idea for reducing possible "unknown unknowns" risk intercepting such a critical system.

snickell:
> I decided not to add a unit test for this new InstrumentedCookieStore class; because it's intended to be a temporary addition while we collect a little bit of data, a test felt unnecessary.

Totally agree, I'd much rather see creative/abusive interactive testing here vs a short-lived unit test.

dashboard/config/custom_session_store/instrumented_cookie_store.rb
cat5inthecradle: For comparison, unless you specifically set the `storage_resolution` to specify [high-resolution metrics](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/publishingMetrics.html#high-resolution-metrics), CloudWatch will default to standard resolution of 1 minute. If you send multiple instances of the same metric within the same minute, CloudWatch will take the _average_ of the submitted values. It's also worth noting that every `PutMetricData` call for a custom metric is charged ([pricing page](https://aws.amazon.com/cloudwatch/pricing/)).

If you change to less than a minute, you'll need to adjust the `storage_resolution` in the metric payload.

That said, I think a minute is fine too.

Hamms: Great feedback, thank you both! I feel much more confident in the one-minute window now

dashboard/config/custom_session_store/instrumented_cookie_store.rb
Hamms: Great idea! It is far too easy to forget about TODOs :upside_down_face:

dashboard/config/custom_session_store/instrumented_cookie_store.rb
Hamms: This is a great suggestion, thank you!

I also added some honeybadger logging to the error handler. This does technically introduce the possibility that an error within our honeybadger logging could bubble up to affect `write_session`, but honeybadger logging in an error handler is a pattern we use frequently in the codebase and it hasn't caused us problems yet so I *think* it should be fine.

Hamms:
Incorporated y'all's suggestions, and added some more notes to the "Testing story" section. PTAL!

cat5inthecradle (approved):
LGTM!

Regarding the additional metrics with different names or dimensions from development. They'll be cleared out after 14 days. ([source](https://repost.aws/knowledge-center/cloudwatch-find-metric-in-console))

I found the metrics in cloudwatch from your testing, if anybody is curious:
https://us-east-1.console.aws.amazon.com/cloudwatch/home?region=us-east-1#metricsV2?graph=~(metrics~(~(~'code-dot-org*2fActionDispatch~'NumSessionsSeen~'Environment~'adhoc~(yAxis~'right~color~'*2308aad2))~(~'...~'development~(yAxis~'right~color~'*2308aad2))~(~'.~'AverageMaxSessionSize~'.~'adhoc~(color~'*23b088f5))~(~'...~'development~(color~'*23b088f5)))~view~'timeSeries~stacked~false~region~'us-east-1~stat~'Average~period~300)&query=~'*7bcode-dot-org*2fActionDispatch*2cEnvironment*7d

snickell (approved):
LGTM, lets do this!

PR #57362, opened 2024-03-19
[Game Design] Follow-up: Handle variable bubble scientific notation

ebeastlake:
This PR is a follow-up to https://github.com/code-dot-org/code-dot-org/pull/57285.

The video in the original PR (linked above) shows a known edge case: If the variable label is very long and there is a large numerical value that p5 converts to scientific notation, the scientific notation is cut off, making it look like the value is 1.2 instead of 1.2E9 or something. See the screenshot below.

Screenshot 2024-03-19 at 1 47 35 PM

This PR reduces the severity of the bug described above in two ways:
- It shortens the `maxLabelLength` from 30 characters to 20. I did this because, at 30 characters, using wide characters will still overflow the playspace. At 20 label characters, even with a string of `w`s, it's clear that the value is in scientific notation even if the exponent isn't always on screen (see below). cc @samantha-code 
- It introduces a new `formatForPlayspace` util that will take large numbers and render them using only two units of precision before the exponent. 

Screenshot 2024-03-18 at 11 11 34 PM

Note: I say reduces, not solves, because if the label uses sufficiently wide characters and/or the exponent gets sufficiently large, there is still part of the number that may not be visible.

## Links

Jira ticket: None
Previous PR: https://github.com/code-dot-org/code-dot-org/pull/57285

## Testing story

Tested locally.

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/p5lab/spritelab/CoreLibrary.js
+++ b/apps/src/p5lab/spritelab/CoreLibrary.js
@@ -10,6 +10,7 @@ import {
   displayWorkspaceAlert,
 } from '../../code-studio/projectRedux';
 import msg from '@cdo/locale';
+import {formatForPlayspace} from '../utils';
 
 export default class CoreLibrary {
   constructor(p5, jsInterpreter) {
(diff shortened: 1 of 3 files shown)


molly-moen (approved):
Nice!

Issue #57368, opened 2024-03-19
ESLint import/order rule should not enforce grouping of imports


snickell:
I'm seeing some behavior I don't like with the import/order rule around it trying to enforce single-groups, rather than trusting the human to create import groups that make semantic sense to them. I think this is more important than alphabetization, and personally, if I had to chose one, I'd rather have the ability for humans to group imports (and encourage people to do this) than have them in one large block in alphabetical order.

I tried applying eslint to webpack.config.js, and it went from this:
```
const webpack = require('webpack');
const path = require('path');
const sass = require('sass');

// Webpack Plugins:
const {BundleAnalyzerPlugin} = require('webpack-bundle-analyzer');
const CopyPlugin = require('copy-webpack-plugin');
const LiveReloadPlugin = require('webpack-livereload-plugin');
const {StatsWriterPlugin} = require('webpack-stats-plugin');
const TerserPlugin = require('terser-webpack-plugin');
const UnminifiedWebpackPlugin = require('unminified-webpack-plugin');
const {WebpackManifestPlugin} = require('webpack-manifest-plugin');
const WebpackNotifierPlugin = require('webpack-notifier');
const ForkTsCheckerWebpackPlugin = require('fork-ts-checker-webpack-plugin');
const {PyodidePlugin} = require('@pyodide/webpack-plugin');
const CircularDependencyPlugin = require('circular-dependency-plugin');
```

To this (which it still errored out on due to the newline, but if it had succeeded, it would not have allowed webpack plugins to be a separate group o fimports from path, sass, and webpack, which I think lowers code quality):
```
const {PyodidePlugin} = require('@pyodide/webpack-plugin');
const CircularDependencyPlugin = require('circular-dependency-plugin');
const CopyPlugin = require('copy-webpack-plugin');
const ForkTsCheckerWebpackPlugin = require('fork-ts-checker-webpack-plugin');
const path = require('path');
const sass = require('sass');
const TerserPlugin = require('terser-webpack-plugin');
const UnminifiedWebpackPlugin = require('unminified-webpack-plugin');
const webpack = require('webpack');

// Webpack Plugins:
const {BundleAnalyzerPlugin} = require('webpack-bundle-analyzer');
const LiveReloadPlugin = require('webpack-livereload-plugin');
const {WebpackManifestPlugin} = require('webpack-manifest-plugin');
const WebpackNotifierPlugin = require('webpack-notifier');
[...]

PR #57369, opened 2024-03-19
`yarn lint` checks same file list as lint.rb eslint pre-commit hook

snickell:
Prior to this change, the list of files checked by `yarn lint` did not match the list of files eslinted using the /apps eslint config by our pre-deploy git hook. The result was that it was easy to accidentally add lint rules (like #57368) that did not fix all the files that could be flagged the next time they were modified.

After this PR: `yarn lint` will catch the same files/issues in /apps as the pre-deploy hook.

### Changes in this PR
1. Fixes `lint.rb`, which is invoked by our git pre-commit hook, so that `filter_eslint_apps()` only returns files that are, in fact, in `/apps`. Previously, our `/apps` eslint config was applied to lots of files outside apps, with a very adhoc list of exceptions added over the years. Many files currently in the repo would be flagged by this over-application, if they were ever changed. I think this was originally a very old bug in `filter_eslint_apps()` that never got fixed.
2. `yarn lint` now checks /apps/**. Previously it only checked `/apps/src/**` and `/apps/test/**`.
    - Fixes lint complaints generated by increasing the scope of yarn lint to run on all files in /apps

### How Tested

Temporarily modified lint.rb to return ALL JS/JSX/TS/TSX files in the repo, not just those currently modified in the branch. Then ran that output through `filter_eslint_apps()`, and finally `run_eslint_apps()`, which invokes `eslint` directly with a list of files. List of files was verified to only include files in /apps, and pre-deploy hook passed. Then added eslint errors to files in /apps, and outside apps to verify that it would flag those in apps, but not outside.

### DIFF
--- a/apps/.eslintrc.js
+++ b/apps/.eslintrc.js
@@ -177,6 +177,9 @@ module.exports = {
       },
     },
     {
+      rules: {
+        'import/order': 'off',
+      },
       // We are actively working to decrease the number of folders in this list.
       // To turn on the order rule for a folder, remove it from this list and run `yarn lint --fix`
       // Commit any changes made.
@@ -352,9 +355,6 @@ module.exports = {
         'test/util/tutorialExplorer/**',
         'test/util/weblab/**',
       ],
-      rules: {
-        'import/order': 'off',
-      },
     },
   ],
 };
--- a/apps/package.json
+++ b/apps/package.json
@@ -10,7 +10,7 @@
   "scripts": {
     "lint": "yarn lint:scss && yarn lint:js",
     "lint:fix": "yarn lint:scss --fix && yarn lint:js --fix",
-    "lint:js": "eslint --format ./.eslintCustomMessagesFormatter.js --ext .js,.jsx,.ts,.tsx src test --resolve-plugins-relative-to .",
+    "lint:js": "eslint --format ./.eslintCustomMessagesFormatter.js --ext=.js,.jsx,.ts,.tsx .",
     "lint:scss": "npx stylelint '**/*.scss'",
     "test": "./run-tests-in-parallel.sh",
     "test:unit": "grunt karma --testType=unit",
--- a/tools/hooks/lint.rb
+++ b/tools/hooks/lint.rb
@@ -8,16 +8,10 @@ SHARED_JS_DIR = "#{REPO_DIR}/shared/js".freeze
 SCSS_GLOB = "#{REPO_DIR}/#{YAML.load_file('.scss-lint.yml')['scss_files'] || '*'}".freeze
 
 def filter_eslint_apps(modified_files)
+  full_apps_dir = File.expand_path(APPS_DIR)
+
   modified_files.select do |f|
-    (f.end_with?(".js", ".jsx", ".ts", ".tsx")) &&
-      !(f.end_with?('.min.js') ||
-        f.match(/public\/.+package\//) ||
-        f.include?('apps/lib/') ||
-        f.include?('shared/') ||
-        f.include?('dashboard/config/') ||
-        f.include?('pegasus/sites/virtual') ||
-        f.include?('dashboard/public')
-      )
+    f.start_with?(full_apps_dir) && f.end_with?(".js", ".jsx", ".ts", ".tsx")
   end
 end
 
--- a/apps/.storybook/decorators.js
+++ b/apps/.storybook/decorators.js
@@ -1,7 +1,8 @@
 import {createStore, combineReducers, applyMiddleware} from 'redux';
+import reduxThunk from 'redux-thunk';
+
 import isRtl from '@cdo/apps/code-studio/isRtlRedux';
 import responsive from '@cdo/apps/code-studio/responsiveRedux';
-import reduxThunk from 'redux-thunk';
 
 export const reduxStore = (reducers = {}, state = {}) => {
   return createStore(
--- a/apps/.storybook/main.js
+++ b/apps/.storybook/main.js
@@ -1,7 +1,9 @@
+import {serverRequire} from '@storybook/core-common';
+import {loadCsf} from '@storybook/csf-tools';
+
 import envConstants from '../envConstants';
 import storybookWebpackConfig from '../webpackStorybook.config';
-import {loadCsf} from '@storybook/csf-tools';
-import {serverRequire} from '@storybook/core-common';
+
 import {compile} from './compile';
 
 const staticDirs = envConstants.STORYBOOK_STATIC_ASSETS
--- a/apps/.storybook/unplugin.js
+++ b/apps/.storybook/unplugin.js
@@ -1,5 +1,6 @@
-import {createUnplugin} from 'unplugin';
 import {serverRequire} from '@storybook/core-common';
+import {createUnplugin} from 'unplugin';
+
 import {compile} from './compile';
 
 export const STORIES_REGEX = /\.dynamicStory\.[tj]sx?/;
--- a/apps/Gruntfile.js
+++ b/apps/Gruntfile.js
@@ -1,15 +1,15 @@
 var chalk = require('chalk');
 var child_process = require('child_process');
-var path = require('path');
 var fs = require('fs');
 var _ = require('lodash');
+var path = require('path');
 var sass = require('sass');
-var envConstants = require('./envConstants');
-var checkEntryPoints = require('./script/checkEntryPoints');
 
-const {ALL_APPS, appsEntriesFor} = require('./webpackEntryPoints');
-const {createWebpackConfig} = require('./webpack.config');
+var envConstants = require('./envConstants');
 const {VALID_KARMA_CLI_FLAGS} = require('./karma.conf');
+var checkEntryPoints = require('./script/checkEntryPoints');
+const {createWebpackConfig} = require('./webpack.config');
+const {ALL_APPS, appsEntriesFor} = require('./webpackEntryPoints');
 
 // Review every couple of years to see if an increase improves test performance
 const MEM_PER_KARMA_PROCESS_MB = 4300;
--- a/apps/eslint/index.js
+++ b/apps/eslint/index.js
@@ -2,9 +2,9 @@ module.exports = {
   rules: {
     'style-blocks-below-class': {
       meta: {
-        fixable: 'code'
+        fixable: 'code',
       },
-      create: function(context) {
+      create: function (context) {
         function isStyleBlock(item) {
           return (
             item.type === 'VariableDeclaration' &&
@@ -43,11 +43,10 @@ module.exports = {
               context.report({
                 node: styleBlock,
                 message: 'Style block should be declared after class',
-                fix: function(fixer) {
+                fix: function (fixer) {
                   const sourceCode = context.getSourceCode();
-                  const styleComments = sourceCode.getCommentsBefore(
-                    styleBlock
-                  );
+                  const styleComments =
+                    sourceCode.getCommentsBefore(styleBlock);
 
                   // if there are style comments, don't automatically re-arrange block
                   if (styleComments.length) {
@@ -62,14 +61,14 @@ module.exports = {
                     fixer.insertTextAfter(
                       lastClassDeclaration,
                       '\n\n' + styleBlockText
-                    )
+                    ),
                   ];
-                }
+                },
               });
             }
-          }
+          },
         };
-      }
-    }
-  }
+      },
+    },
+  },
 };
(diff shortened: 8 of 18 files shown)


tools/hooks/lint.rb
```diff
@@ -8,16 +8,10 @@
...
-        f.match(/public\/.+package\//) ||
-        f.include?('apps/lib/') ||
-        f.include?('shared/') ||
-        f.include?('dashboard/config/') ||
-        f.include?('pegasus/sites/virtual') ||
-        f.include?('dashboard/public')
-      )
+    f.start_with?(full_apps_dir) && f.end_with?(".js", ".jsx", ".ts", ".tsx")
   end
 end
```
snickell, line 16: Unlike the other filter_* methods in lint.rb, `filter_eslint_apps()` is missing the most basic directory check, which causes it to enforce /apps eslint rules across the repo, with a list of exceptions added one by one (lines 14 - 19), but still actually doesn't cover all files it mis-includes, like pegasus/sites.v3.

Compare with the code for other `filter_*` methods like this:
```
def filter_scss_apps(modified_files)
  modified_files.select {|f| f.include?('apps/') && f.end_with?(".scss")}
end
```

Or:
```
def filter_eslint_shared(modified_files)
  modified_files.select do |f|
    f.match(%r{/shared/js/[^/]+.js})
  end
end
```

I think this is a very old bug that we never fixed.

As a result of this bug, there are a lot of JS files outside /apps that if they were ever modified would be failed by this pre-commit hook (many of them are pre-bundled libraries, and it def shouldn't apply). Some of the files that would be mis-failed by this hook due to its missing directory check are:
```
/aws/cloudformation/ami-manager.js
/aws/cloudformation/cfn-response.js
/aws/cloudformation/count_asg.js
/aws/cloudformation/honeybadgerNotify.js
/aws/cloudformation/lambdas/sql-user/index.js
/aws/cloudformation/lambdas/sql-user/index.test.js
/aws/cloudformation/s3Object.js
/aws/cloudformation/s3PartitionCloudFrontLog.js
/aws/cloudformation/slackCloudWatchEvent.js
/aws/cloudformation/test/index.js
/aws/cloudformation/test/testHoneybadgerNotify.js
/aws/cloudformation/test/testS3PartitionCloudFrontLog.js
/aws/cloudformation/vpcClassicLink.js
/aws/offsite/aurora-snapshots-fargate/restore-and-verify-backup.js
/aws/offsite/aurora-snapshots-fargate/test/test-restore-and-verify-backup.js
/aws/offsite/prune-aurora-backups/src/prune-aurora-backups.js
/aws/offsite/prune-aurora-backups/src/tests/test-prune-aurora-backups.js
/bin/generate-pdf/generatePdf.js
/bin/oneoff/pullOutBlocks.js
/dashboard/scripts/archive/migrateAnimationData.js
/dashboard/scripts/archive/zendesk_theme_js.js
/dashboard/test/ui/js/set-tab-status-icon.js
/dashboard/vendor/assets/javascripts/add2home.js
/lib/cdo/pegasus/screencap.js
/lib/pdf/pdf_rasterize.js
/tools/scripts/brokenLinkChecker/brokenLinkChecker.js
/tools/scripts/prune-rds-backups.js
```

apps/.eslintrc.js
```diff
@@ -352,9 +355,6 @@ module.exports = {
         'test/util/tutorialExplorer/**',
         'test/util/weblab/**',
       ],
-      rules: {
-        'import/order': 'off',
-      },
```
snickell, line 357: This is a readability change (ironically breaking alphabetical order 😆 ). The list of excluded dirs is very long, and you have to scan to the very end to find out WHICH rule has all these excluded dirs.

apps/package.json
```diff
@@ -10,7 +10,7 @@
   "scripts": {
     "lint": "yarn lint:scss && yarn lint:js",
     "lint:fix": "yarn lint:scss --fix && yarn lint:js --fix",
-    "lint:js": "eslint --format ./.eslintCustomMessagesFormatter.js --ext .js,.jsx,.ts,.tsx src test --resolve-plugins-relative-to .",
+    "lint:js": "eslint --format ./.eslintCustomMessagesFormatter.js --ext=.js,.jsx,.ts,.tsx .",
```
snickell, line 13: `yarn lint:js` now applies to all of /apps, relying on /apps/.eslintignore for skipping files. Previously eslint was just applied to /apps/src and /apps/test, which resulted in it not flagging some files on `yarn lint` that WERE flagged by the pre-commit hook.

tools/hooks/lint.rb
lfryemason: Do the files that are currently included in this eslint file but will be ignored on future lint runs need their own eslint config or some way to get linted similarly to apps?

lfryemason (approved):

Hamms (commented):
I'm torn on this one. It's obviously a big win to make things behave more consistently and predictably, but we do want to be able to lint some of the javascript we have that lives outside of `/apps`. Ideally, we'd accomplish that by moving everything to within `/apps` so that it would be the case that `yarn lint` can lint everything we want to be linted, but until that happens I feel like entirely losing the ability to lint these strays is a bit of a step backwards.

Put another way, I'd say the problem is less that `lint.rb` is looking at too many files and more that that it's hard to make `yarn lint` look at everything we want it to.

snickell:
@lfryemason thought of this, and we went through and looked at folders outside apps/ where:
- we have JS
- this lint rule is applied
- and actually passes more than a small percentage of files in the folder (i.e. the lint rule is actually "in useful use"

The list is very small. I think those places should do like /shared does, and add a lint.rb hook appropriate to their situation and usage.

Hamms:
Oh, fantastic! I'd be curious to see that list, just for my own edification. A standalone `lint.rb` hook just for the stray JS files makes sense to me; would it be reasonable to include that in this PR, or is it a big enough change to justify being on its own?

snickell:
Yup, I was just gonna regenerate the list, and good to have documented here too. My thinking was to add another lint.rb hook, but only for directories where it clearly made sense, and I'm comfortable fixing all the lint issues. One thing that's important to me here is that: if we pre-commit lint it when you change it, it should be correct in git already. I think the worst is when we have lint rules that if you touch a file at all, suddenly you have to reformat the whole file to fix lint errors, better to just have no linting for that file and add it once its ready to be linted.

snickell:
@Hamms I made us a list :) TL;DR: 3 JS files outside pegasus, and 10 files in JS pegasus (but not distinguishable from the 75% of pegasus that fails or linter or shouldn't be linted at all since its vendor code) pass our linter. Everything else oustide apps fails our linter (and would trip the pre-commit hook prior to this pr if ever changed).

🟢 JS files that pass our linter and would no longer be linted 😢 :
  /dashboard/test/ui/test_status.js
  /dashboard/vendor/assets/javascripts/emulate-print-media.js
  /tools/scripts/convertScssToJs.js
  
🟢 These files pass the linter, **but** only 25% of files in /pegasus/sites.v3 do (and 30% shouldn't ever be linted because its vendor code) 🤷 :
  /pegasus/sites.v3/code.org/public/js/dance_inspirational_videos.js
  /pegasus/sites.v3/code.org/public/js/font_awesome_icon_class.js
  /pegasus/sites.v3/code.org/public/js/hoc-overview.js
  /pegasus/sites.v3/code.org/public/js/page/educate_csd.js
  /pegasus/sites.v3/code.org/public/js/tabs-section.js
  /pegasus/sites.v3/code.org/public/js/top_skinny_banner.js
  /pegasus/sites.v3/hourofcode.com/public/js/how-to.js
  /pegasus/sites.v3/hourofcode.com/public/js/page/ai.js


🔴 JS files in the repo that do NOT pass our current linter 🤬:
  /aws/cloudformation/ami-manager.js
  /aws/cloudformation/cfn-response.js
  /aws/cloudformation/count_asg.js
  /aws/cloudformation/honeybadgerNotify.js
  /aws/cloudformation/lambdas/sql-user/index.js
  /aws/cloudformation/lambdas/sql-user/index.test.js
  /aws/cloudformation/s3Object.js
  /aws/cloudformation/s3PartitionCloudFrontLog.js
  /aws/cloudformation/slackCloudWatchEvent.js
  /aws/cloudformation/test/index.js
  /aws/cloudformation/test/testHoneybadgerNotify.js
  /aws/cloudformation/test/testS3PartitionCloudFrontLog.js
  /aws/cloudformation/vpcClassicLink.js
  /aws/offsite/aurora-snapshots-fargate/restore-and-verify-backup.js
  /aws/offsite/aurora-snapshots-fargate/test/test-restore-and-verify-backup.js
  /aws/offsite/prune-aurora-backups/src/prune-aurora-backups.js
  /aws/offsite/prune-aurora-backups/src/tests/test-prune-aurora-backups.js
  /bin/generate-pdf/generatePdf.js
  /bin/oneoff/pullOutBlocks.js
  /dashboard/scripts/archive/migrateAnimationData.js
  /dashboard/scripts/archive/zendesk_theme_js.js
  /dashboard/test/ui/js/set-tab-status-icon.js
  /dashboard/vendor/assets/javascripts/add2home.js
  /lib/cdo/pegasus/screencap.js
  /lib/pdf/pdf_rasterize.js
  /pegasus/sites.v3/code.org/public/js/admins_email_signup_form.js
  /pegasus/sites.v3/code.org/public/js/applab-docs.js
  /pegasus/sites.v3/code.org/public/js/csf-congrats.js
  /pegasus/sites.v3/code.org/public/js/curriculum-pdf.js
  /pegasus/sites.v3/code.org/public/js/extended_us_map.js
  /pegasus/sites.v3/code.org/public/js/mobile-footer.js
  /pegasus/sites.v3/code.org/public/js/petition.js
  /pegasus/sites.v3/code.org/public/js/promote.js
  /pegasus/sites.v3/code.org/public/js/terms_interstitial.js
  /pegasus/sites.v3/code.org/public/js/volunteer-local.js
  /pegasus/sites.v3/code.org/public/js/volunteer-unsubscribe.js
  /pegasus/sites.v3/hourofcode.com/public/js/test-data.js
  /tools/scripts/brokenLinkChecker/brokenLinkChecker.js
  /tools/scripts/prune-rds-backups.js

🔴 JS files that do NOT pass, and shouldn't even be tried (vendor code) 🙃:
  /pegasus/sites.v3/code.org/public/js/bootstrap-datepicker.js
  /pegasus/sites.v3/code.org/public/js/bootstrap.js
  /pegasus/sites.v3/code.org/public/js/codemirror.js
  /pegasus/sites.v3/code.org/public/js/html5shiv.js
  /pegasus/sites.v3/code.org/public/js/masonry.pkgd.js
  /pegasus/sites.v3/code.org/public/js/page/minecraft_page.js
  /pegasus/sites.v3/code.org/public/js/page/video_page.js
  /pegasus/sites.v3/code.org/public/js/raphael.js
  /pegasus/sites.v3/code.org/public/js/selectize.js
  /pegasus/sites.v3/code.org/public/js/video.js
  /pegasus/sites.v3/hourofcode.com/public/js/html5shiv.js
  /pegasus/sites.v3/hourofcode.com/public/js/markerclusterer_packed.js

snickell:
My feeling: files that are hooked to be linted but do not pass are worse than files that do pass but will not be linted. The only systemic casualty I really see here are the files in pegasus that Kelby has passing lint, BUT they're so in the noise of the files next to them that don't pass lint, and I don't think our full lint ruleset really makes sense in those files anyway.

Proposal:
1. Merge this PR as-is
2. Offer to @kelbyhawn to help setup an independent linter of her choosing (if she wants) operating on pegasus, and hooked into lint.rb. This also moves us further in decoupling pegasus from dashboard.
3. Remove /.eslintrc.js and/or merge it into /apps/.eslintrc.js: be explicit that the apps eslint is for apps, and there's no problem with adding another eslint setup, but its a separate module.

kelbyhawn:
Hey @snickell! @bethanyaconnor and I talked to @davidsbailey about linting on Pegasus, and we're going to move non-vendor code js files into `apps/src` so they can still get linted and use webpack entry points as seen [here](https://github.com/code-dot-org/code-dot-org/blob/79967d8e8fa3e42cf4634e72df97e275a2625e3e/apps/webpackEntryPoints.js#L175-L193). 

Jira ticket: [ACQ-1669](https://codedotorg.atlassian.net/browse/ACQ-1669)

davidsbailey:
> we're going to move non-vendor code js files into `apps/src` so they can still get linted and use webpack entry points as seen [here](https://github.com/code-dot-org/code-dot-org/blob/79967d8e8fa3e42cf4634e72df97e275a2625e3e/apps/webpackEntryPoints.js#L175-L193).

And so those files can benefit from being transpiled by babel! including (1) the ability to use more ES6 features, and (2) better cross-browser compatibility.

davidsbailey:
> 🔴 JS files in the repo that do NOT pass our current linter 🤬:

some of these files look like they are written to be run in node.js. I wouldn't expect to have this kind of file follow the exact set of linting rules as the rest of our site.

snickell:
> non-vendor code js files into apps/src

I like this approach a lot!

PR #57373, opened 2024-03-19
Update top_skinny_banner view to work with and without a url

kelbyhawn:
Refactors the `top_skinny_banner.haml` view so it works with or without a URL. Set `banner_url: false` where the view is used. The entire banner is no longer a link, and now only the text will link, or not link.

**Jira ticket:** [ACQ-1648](https://codedotorg.atlassian.net/browse/ACQ-1648)

----

## With url

https://github.com/code-dot-org/code-dot-org/assets/9256643/bee8442f-92c0-40b4-85fe-8e8a0da70c0c

## Without url

https://github.com/code-dot-org/code-dot-org/assets/9256643/9831bea4-ae45-4ba9-aa04-ed20fbd6c7e4

### DIFF
--- a/pegasus/sites.v3/code.org/styles_min/040-page.css
+++ b/pegasus/sites.v3/code.org/styles_min/040-page.css
@@ -586,34 +586,31 @@ aside.top-skinny-banner {
   &.light-theme {
     background: var(--neutral_dark10);
 
-    a > *,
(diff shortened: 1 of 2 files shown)


TurnerRiley (approved):
Looks great! Thank you for the speedy update!

PR #57376, opened 2024-03-19
Split Up Name Argument in `Cdo::Metrics.put`

Hamms:
Our `Cdo::Metrics` helper currently has three methods for logging metrics data: `push`, `put`, and `put_metric`. Both `push` and `put_metric` expect you to define the namespace and metric name as separate strings, but `put` expects you to put them both in a single string separated by a `/`. Unfortunately (and I think unintentionally), this precludes us from being able to use namspaces which themselves have a `/` in them. Not only do `push` and `put_metric` allow that kind of formatting, but we are indeed [taking advantage of that functionality](https://github.com/code-dot-org/code-dot-org/blob/b3dcd016d5cb17c696e0fc9e6436f87ad5a1e2a2/bin/cron/report_activejob_metrics#L7-L12)

Because I don't think we have a specific reason to disallow `/` in the namespace and because the specific implementation of the `put` method does not appear to be actually saving us from a significant amount of complexity, I propose that we refactor the method to be more in line with the other methods in the class.

Thoughts? Concerns? Any specific context as to why we originally implemented it this way that I'm missing?

## Testing story

Updated one unit test to reflect this change; relying on other unit tests to verify no unexpected side effects.

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/lib/cdo/app_server_hooks.rb
+++ b/lib/cdo/app_server_hooks.rb
@@ -28,7 +28,7 @@ module Cdo
 
     def self.after_fork(host:)
       require 'cdo/aws/metrics'
-      Cdo::Metrics.put('App Server/WorkerBoot', 1, Host: host)
+      Cdo::Metrics.put('App Server', 'WorkerBoot', 1, Host: host)
 
       # Statsig is initialized here for managed environments. For development, it is
       # intialized in config/initializers/statsig.rb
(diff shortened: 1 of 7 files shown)


cat5inthecradle:
LGTM, but, [missed a spot](https://github.com/code-dot-org/code-dot-org/blob/3bb0153946b7f7d3b7b54c8fcefceba6ada94dd6/lib/cdo/lighthouse.rb#L59)?

Hamms:
Oh, good catch, thanks! That's what I get for always trying to use explicit parens in my own Ruby code; I always forget that implicit parens are also a thing I have to look out for :smile:

snickell (approved):
✅, appreciate the arg name changing from name => metric_name too.

cat5inthecradle (approved):

PR #57382, opened 2024-03-19
[Sprite Lab] refactor getVariableValue

mikeharv:
This a quick refactor of `getVariableValue`. It adds an optional parameter, which if provided, overrides any returned `undefined` values. This function was created for variable bubbles which use an empty string instead of "undefined", but there are other cases where we want the unmodified values such as validation.

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/p5lab/spritelab/CoreLibrary.js
+++ b/apps/src/p5lab/spritelab/CoreLibrary.js
@@ -148,7 +148,8 @@ export default class CoreLibrary {
         return;
       }
 
-      const value = this.getVariableValue(name);
+      // Use variable value or empty string (if undefined).
+      const value = this.getVariableValue(name, '');
 
       // Determine if the label needs truncation and append an ellipsis if so
       const displayLabel =
@@ -182,7 +183,7 @@ export default class CoreLibrary {
     });
   }
 
-  getVariableValue(variableName) {
+  getVariableValue(variableName, defaultValue) {
     if (!this.jsInterpreter) {
       console.error('JS Interpreter not set in CoreLibrary');
       return;
@@ -192,7 +193,7 @@ export default class CoreLibrary {
       // Blockly does not execute code or track the runtime values of variables, so we need to
       // evaluate the variable's value using the JSInterpreter.
       const result = this.jsInterpreter.evaluateWatchExpression(variableName);
-      return typeof result === 'undefined' ? '' : result;
+      return typeof result === 'undefined' ? defaultValue : result;
     } catch (e) {
       console.error(`Error evaluating variable '${variableName}': ${e}`);
       return '';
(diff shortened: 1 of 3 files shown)


apps/src/p5lab/spritelab/CoreLibrary.js
```diff
@@ -148,7 +148,8 @@ export default class CoreLibrary {
         return;
       }
 
-      const value = this.getVariableValue(name);
+      // Use variable value or empty string (if undefined).
+      const value = this.getVariableValue(name, '');
```
mikeharv, line 152: This is the original use case for variable bubbles. We want to us an empty string for undefined values.

apps/src/p5lab/spritelab/commands/criterionCommands.js
```diff
@@ -548,11 +548,12 @@ export const commands = {
...
 
   // Returns true if a minimum number of watched variables have a valid value.
-  // If unset, a variable's value is an empty string.
+  // If unset, a variable's value is undefined.
   variableValueSet(min = 1) {
     const studentVars = this.getVariableBubbles();
     const filteredVars = studentVars.filter(
-      studentVar => this.getVariableValue(studentVar.name) !== ''
+      studentVar =>
+        typeof this.getVariableValue(studentVar.name) !== 'undefined'
```
mikeharv, line 556: Here it's actually preferable that we can potentially have an undefined value. Now, if a student's code says `set [name] to ''`, it will validate differently than if they only show the variable without setting it.

apps/src/p5lab/spritelab/commands/validationCommands.js
```diff
@@ -246,9 +246,10 @@ export const commands = {
     variableBubbles.forEach(variableBubble => {
       this.previous.variableBubbles.push({
         ...variableBubble,
-        // Run the variable name through the JS interpreter to get its value.
+        // Run the variable name through the JS interpreter to get its value,
+        // using an empty string for undefined variables.
         // We need to store the previous value to validate changes between frames.
-        value: this.getVariableValue(variableBubble.name),
+        value: this.getVariableValue(variableBubble.name, ''),
```
mikeharv, line 252: This is where we we store the current bubbles into the `previous` object at the end of a frame, so that we can make comparisons against the next frame.

apps/src/p5lab/spritelab/commands/criterionCommands.js
```diff
@@ -60,7 +60,7 @@ export const commands = {
     const spriteIds = this.getSpriteIdsInUse();
     return (
       spriteIds.filter(id => this.getLastSpeechBubbleForSpriteId(id)).length >=
-      1
+      min
```
mikeharv, line 63: Quick bug fix, unrelated to PR. `min` parameter was recently added but unused.

molly-moen (approved):

ebeastlake (approved):

PR #57384, opened 2024-03-19
Add a .white utility class for use on text elements

kelbyhawn:
Adds a `.white` class for use on text elements and links, except for links styled as buttons using the `.link-button` class. This cuts down on the need for custom styles or inline styles on text elements that need to be white on a dark background. 

I updated the places that were doing this with inline styles in this PR too.

**Jira ticket:** [ACQ-1647](https://codedotorg.atlassian.net/browse/ACQ-1647)


TurnerRiley (approved):
Looks great!

PR #57388, opened 2024-03-19
INF-1141 Publish additional ActiveJob metrics

cat5inthecradle:
Adding instrumentation to the ApplicationJob class, which is extended by other jobs.

Note that the event callback methods stack with any callbacks implemented by child classes, so you can't accidentally overwrite this implementation.

No tests for this, though I'm happy to explore adding that.

Here is an example of how these metrics present in CloudWatch during which I simulated the queue worker not being able to keep up with job creation.

image

## Links

Jira: [INF-1141](https://codedotorg.atlassian.net/browse/INF-1141)

## PR Checklist:

- [ ] Code is well-commented


dashboard/app/jobs/application_job.rb
```diff
@@ -1,7 +1,72 @@
...
+
+    pp metric
+    Cdo::Metrics.put_metric(METRICS_NAMESPACE, metric)
+  end
+
+  after_perform do |job|
+    perform_complete_at = Time.now
+    execution_time = perform_complete_at - @perform_started_at
+    total_time = perform_complete_at - Time.parse(job.enqueued_at)
+    jobs_in_queue = Delayed::Job.where(queue: job.queue_name).where("handler LIKE ?", "%job_class: #{job.class.name}%").count
```
cat5inthecradle: This is pretty ugly, and perhaps not very performant. Very interested in other ideas here. I would love to keep this as as per-job-class metric, but we could simplify by just reporting the total queue size with `Delayed::Job.count`

Jobs look like this in the database (copied as JSON)
```json
{
"delayed_jobs": [
	{
		"id" : 497,
		"priority" : 0,
		"attempts" : 0,
		"handler" : "--- !ruby\/object:ActiveJob::QueueAdapters::DelayedJobAdapter::JobWrapper\njob_data:\n  job_class: SampleJob\n  job_id: 15405e40-6b73-4d0c-ba4c-95c50105b530\n  provider_job_id:\n  queue_name: default\n  priority:\n  arguments:\n  - world\n  executions: 0\n  exception_executions: {}\n  locale: en-US\n  timezone: UTC\n  enqueued_at: '2024-03-19T23:55:55Z'\n",
		"last_error" : null,
		"run_at" : "2024-03-19 23:55:55",
		"locked_at" : null,
		"failed_at" : null,
		"locked_by" : null,
		"queue" : "default",
		"created_at" : "2024-03-19 23:55:55.800409",
		"updated_at" : "2024-03-19 23:55:55.800409"
	}
]}

```

dashboard/app/jobs/application_job.rb
```diff
@@ -1,7 +1,71 @@
...
+      {
+        name: 'Environment',
+        value: CDO.rack_env
+      },
+      {
+        name: 'JobName',
+        value: name
+      }
+    ]
+  end
```
cat5inthecradle, line 24: This and the code that calls it, `job.class.common_dimensions` is some Ruby inheritance logic that I don't fully understand. If there's a better way to deduplicate this code while still referencing the name of the subclass, I'm all for it.

dashboard/app/jobs/application_job.rb
cat5inthecradle: This comment is marked as outdated because I moved this line.

dashboard/app/jobs/application_job.rb
davidsbailey: not sure I understand exactly what you don't like about your current implementation, but if you'd prefer you could convert this from a class method to an instance method.

dashboard/app/jobs/application_job.rb
davidsbailey: if I understand correctly, you are trying to monitor the activejob infrastructure, and not the implementation details of any particular job. based on that, monitoring/alerting on total # jobs seems best to me.

in the future if we end up wanting to monitor queues for individual jobs (for example if we define more than one queue), then perhaps we would shard the monitoring based on the `queue` field rather than inspecting the handler.

davidsbailey (commented):
a few initial comments

dashboard/app/jobs/application_job.rb
cat5inthecradle: Sounds good to me. We can always query the database if we have a non-zero number we want to learn more about.

dashboard/app/jobs/application_job.rb
```diff
@@ -1,7 +1,82 @@
...
+
+  # Parent class callbacks are called in addition to any callbacks defined in the job subclass
+  # https://guides.rubyonrails.org/v6.0/active_job_basics.html#available-callbacks
```
davidsbailey: are you able to determine the ordering of these callbacks, i.e. will the execution time include this work done in the callbacks of the child job? https://github.com/code-dot-org/code-dot-org/blob/b4a4d48aa2a00fa66b574187638f423bdf874115/dashboard/app/jobs/evaluate_rubric_job.rb#L131-L135

I would argue that it should, so, if it doesn't, a comment highlighting that would be nice.

davidsbailey (approved):
Nice work Darin! LGTM after considering comments.

dashboard/app/jobs/application_job.rb
cat5inthecradle: I was looking for the best way to deduplicate this array out of each of the metrics arrays below. If this is a suitable way, that's fine.

cat5inthecradle: I merged in some changes to the cron job and added a note that they publish to the same metric.

dashboard/app/jobs/application_job.rb
cat5inthecradle: Ran a test, I'll review this and update the comment to provide some insight into what gets run first. Looks like it's parent-then-child before the `perform` method, then it's 'child-then-parent` after.

```
[development] dashboard > SampleJob.perform_later("world")
In ApplicationJob - before_enqueue
In ApplicationJob - around_enqueue (before)
In SampleJob - before_enqueue
In SampleJob - around_enqueue (before)
In SampleJob - after_enqueue
In SampleJob - around_enqueue (after)
In ApplicationJob - around_enqueue (after)
In ApplicationJob - after_enqueue
...
Enqueued SampleJob (Job ID: d0e2cfca-3e2d-4312-a573-4caa6a2158b5) to Async(default) with arguments: "world"
Performing SampleJob (Job ID: d0e2cfca-3e2d-4312-a573-4caa6a2158b5) from Async(default) enqueued at 2024-03-26T19:45:25Z with arguments: "world"
...
In ApplicationJob - before_perform
In ApplicationJob - around_perform (before)
In SampleJob - before_perform
In SampleJob - around_perform (before)
In SampleJob - perform
Goodbye, ["world"]
In SampleJob - after_perform
In SampleJob - around_perform (after)
In ApplicationJob - around_perform (after)
In ApplicationJob - after_perform
Performed SampleJob (Job ID: d0e2cfca-3e2d-4312-a573-4caa6a2158b5) from Async(default) in 1028.48ms
```

davidsbailey: thank you for doing that, looks good to me!

cat5inthecradle: Actually reading that list, it looks asynchronous and nondeterminstic. Things happen when they're supposed to in relation to the job perform, but when it comes to Parent vs Child, or similar events, there's inconsistency.

1. Parent and Child events of the same type may happen in any order
2. Otherwise events happen as follows:

|   |
|---|
|`before_enqueue` and `around_enqueue(before)` in any order|
|_Job is added to database or other queue store_|
|`after_enqueue` and `around_enqueue(after)` in any order|
|`before_perform` and `around_perform(before)` in any order|
|_job is performed_|
|`after_perform` and `around_perform(after)` in any order|

dashboard/app/jobs/application_job.rb
```diff
@@ -1,7 +1,97 @@
...
+  # Parent class callbacks are called in addition to any callbacks defined in the job subclass
+  # Callback functions are executed in the following order:
+  #   before_enqueue and around_enqueue(before) in any order
+  #   Job is added to database or other queue store
+  #   after_enqueue and around_enqueue(after) in any order
+  #   before_perform and around_perform(before) in any order
+  #   job is performed
+  #   after_perform and around_perform(after) in any order
+  # When both the Parent and Child class define the same callback, they may be called in any order.
+  # https://guides.rubyonrails.org/v6.0/active_job_basics.html#available-callbacks
```
davidsbailey: ❤️

davidsbailey (approved):

cat5inthecradle:
drone failures are legit, investigating.

PR #57390, opened 2024-03-20
Music: optional vocals filter

breville:
A library can now specify whether it has vocals, with `"hasVocals": true`, and then the sound selection UI will offer "Vocals" as an option.

### DIFF
--- a/apps/src/music/player/MusicLibrary.ts
+++ b/apps/src/music/player/MusicLibrary.ts
@@ -23,6 +23,9 @@ export default class MusicLibrary {
   private bpm: number | undefined;
   private key: Key | undefined;
 
+  // Whether this library contains vocals.
+  private hasVocals: boolean;
+
   constructor(name: string, libraryJson: LibraryJson) {
     this.name = name;
     this.libraryJson = libraryJson;
@@ -42,6 +45,8 @@ export default class MusicLibrary {
     if (libraryJson.key) {
       this.key = Key[libraryJson.key.toUpperCase() as keyof typeof Key];
     }
+
+    this.hasVocals = !!libraryJson.hasVocals;
   }
 
   getDefaultSound(): string | undefined {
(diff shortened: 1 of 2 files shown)


sanchitmalhotra126 (commented):
Could we infer this from scanning the library JSON? Hopefully shouldn't be too slow if the library isn't huge, but that way we can omit any categories that aren't represented, not just vocals?

breville:
> Could we infer this from scanning the library JSON? Hopefully shouldn't be too slow if the library isn't huge, but that way we can omit any categories that aren't represented, not just vocals?

Fair call, and a fun exercise.  Implemented in https://github.com/code-dot-org/code-dot-org/pull/57437.

If the `SegmentedButtons` component supported disabled components, it would be neat to disable a value if the currently-selected panel didn't have any such entries.  cc @levadadenys in case it's on the roadmap!

levadadenys:
> > Could we infer this from scanning the library JSON? Hopefully shouldn't be too slow if the library isn't huge, but that way we can omit any categories that aren't represented, not just vocals?
> 
> Fair call, and a fun exercise. Implemented in #57437.
> 
> If the `SegmentedButtons` component supported disabled components, it would be neat to disable a value if the currently-selected panel didn't have any such entries. cc @levadadenys in case it's on the roadmap!

Not sure I understood correctly, but `SegmentedButtons` already has a support of  disabled state/prop for every segmentedButton in the `SegmentedButtons` component you render

```

export interface SegmentedButtonModel {
  /** Button Label */
  label?: string;
  /** Is button disabled */
  disabled?: boolean;
  /** Is button selected */
  selected?: boolean;
  /** Button unique value. Used for selected/not selected logic */
  value: string;
  /** Segmented Button Type
   * Please note (!) 'buttonType' affects which icons(props) can be rendered.
   * 'withLabel' - we can render 'iconLeft' and/or 'iconRight' only (both of them are optional)
   * 'iconOnly' - we have to render 'icon'.
   * 'number' - we can not render any icons at all.
   * */
  buttonType?: SegmentButtonType;
  /** Icon left from label*/
  iconLeft?: FontAwesomeV6IconProps;
  /** Icon right from label */
  iconRight?: FontAwesomeV6IconProps;
  /** Icon for IconOnly button type */
  icon?: FontAwesomeV6IconProps;
}

```

PR #57400, opened 2024-03-20
feat(lms): add support for teacher account creation and co-teacher support on sync

stephenliang:
This PR adds the following features:

1. Upon section sync, create teacher accounts if the LMS provides the e-mail address and the LMS indicates the user is a teacher. If not, a student account is created. If the user is a teacher on the LMS but a student account is created in this process, the teacher will be upgraded to a teacher account when the user first logs in via the upgrade teacher pathway.
2. Upon section sync, add co-teachers into the section.
3. Downgrade co-teacher to student in a section based on LMS response
4. Upgrade students to co-teacher in a section based on LMS response

## Links

- [jira ticket](https://codedotorg.atlassian.net/browse/P20-603)

### Scenario 1: Create new Code.org teacher account via section sync

"Stephen Student9" is a brand new account on Canvas. Code.org will create a brand new account on Code.org as a teacher and add the user as co-teacher.

[Create teacher account via sync.webm](https://github.com/code-dot-org/code-dot-org/assets/538214/91d5ac4d-fa58-4fec-b287-89c9105119d2)

### Scenario 2: Downgrade co-teachers to students

On canvas, change the role of students 3, 4, and 5 to student.

[Downgrade co-teachers to student.webm](https://github.com/code-dot-org/code-dot-org/assets/538214/826dd34e-2d14-40ba-9cbf-a2d46f54d4e1)

### Scenario 3: Upgrade students to co-teachers

On canvas, change the role of student 3 and 4 to teacher.

[Upgrade student to co-teacher.webm](https://github.com/code-dot-org/code-dot-org/assets/538214/dd98c7f3-6637-459f-ae72-8239e659502d)

### Scenario 4: Transfer a student from one Canvas section to becoming a teacher in a different Canvas section

[transfer student.webm](https://github.com/code-dot-org/code-dot-org/assets/538214/48fa7b6d-5e86-4c62-802f-a262d6c33baa)

### Scenario 5: Transfer a teacher from one Canvas section to becoming a student in a different Canvas section

[transfer teacher.webm](https://github.com/code-dot-org/code-dot-org/assets/538214/c3e4ad25-c71e-4732-8e79-2bd992546f06)

## Follow-up work

- https://codedotorg.atlassian.net/browse/P20-804

## PR Checklist:

- [x] Code is well-commented

### DIFF
--- a/dashboard/app/models/sections/section.rb
+++ b/dashboard/app/models/sections/section.rb
@@ -570,9 +570,9 @@ class Section < ApplicationRecord
 
   # Adds an instructor to the section
   # If the instructor was previously deleted, restore the instructor
-  # If the instructor was invited or inactive, reactive the instructor
+  # Make the instructor active if they had a different status
   # If the instructor did not previously exist, create the section instructor relationship
-  # Returns true if the instructor could be added, false otherwise
+  # Returns true if successful
   def add_instructor(user)
     transaction do
       Follower.find_by(section: self, student_user: user)&.destroy
--- a/dashboard/app/models/sections/section.rb
+++ b/dashboard/app/models/sections/section.rb
@@ -248,17 +248,7 @@ class Section < ApplicationRecord
   def ensure_owner_is_active_instructor
     return if user.blank?
 
-    si = SectionInstructor.with_deleted.find_by(instructor: user, section_id: id)
-    if si.blank?
-      SectionInstructor.create!(section_id: id, instructor: user, status: :active)
-    elsif si.deleted?
-      si.restore
-      si.status = :active
-      si.save!
-    elsif si.status != 'active'
-      si.status = :active
-      si.save!
-    end
+    add_instructor(user)
   end
 
   # return a version of self.students in which all students' names are
@@ -578,7 +568,36 @@ class Section < ApplicationRecord
   end
   before_validation :strip_emoji_from_name
 
-  public def add_instructor(email, current_user)
+  # Adds an instructor to the section
+  # If the instructor was previously deleted, restore the instructor
+  # If the instructor was invited or inactive, reactive the instructor
+  # If the instructor did not previously exist, create the section instructor relationship
+  # Returns true if the instructor could be added, false otherwise
+  public def add_instructor(user)
+    return false unless user.teacher?
+
+    Follower.find_by(section: self, student_user: user)&.destroy
+
+    si = SectionInstructor.with_deleted.find_by(instructor: user, section_id: id)
+    if si.blank?
+      SectionInstructor.create!(section_id: id, instructor: user, status: :active)
+    elsif si.deleted?
+      si.restore
+      si.status = :active
+      si.save!
+    elsif si.status != 'active'
+      si.status = :active
+      si.save!
+    end
+
+    true
+  end
+
+  public def remove_instructor(user)
+    SectionInstructor.find_by(instructor: user, section_id: id)&.destroy
+  end
+
+  public def invite_instructor(email, current_user)
     instructor = User.find_by!(email: email, user_type: :teacher)
     raise ArgumentError.new('inviting self') if instructor == current_user
 
--- a/dashboard/app/models/sections/section.rb
+++ b/dashboard/app/models/sections/section.rb
@@ -573,31 +573,24 @@ class Section < ApplicationRecord
   # If the instructor was invited or inactive, reactive the instructor
   # If the instructor did not previously exist, create the section instructor relationship
   # Returns true if the instructor could be added, false otherwise
-  public def add_instructor(user)
-    return false unless user.teacher?
-
-    Follower.find_by(section: self, student_user: user)&.destroy
-
-    si = SectionInstructor.with_deleted.find_by(instructor: user, section_id: id)
-    if si.blank?
-      SectionInstructor.create!(section_id: id, instructor: user, status: :active)
-    elsif si.deleted?
-      si.restore
-      si.status = :active
-      si.save!
-    elsif si.status != 'active'
-      si.status = :active
-      si.save!
+  def add_instructor(user)
+    transaction do
+      Follower.find_by(section: self, student_user: user)&.destroy
+      si = SectionInstructor.with_deleted.find_or_initialize_by(instructor: user, section_id: id)
+      si.restore if si.deleted?
+      si.active!
     end
 
     true
   end
 
-  public def remove_instructor(user)
-    SectionInstructor.find_by(instructor: user, section_id: id)&.destroy
+  # Removes an instructor
+  # Note: Will not remove the primary instructor to prevent orphaned sections
+  def remove_instructor(user)
+    SectionInstructor.find_by(instructor: user, section_id: id)&.destroy unless self.user == user
   end
 
-  public def invite_instructor(email, current_user)
+  def invite_instructor(email, current_user)
     instructor = User.find_by!(email: email, user_type: :teacher)
     raise ArgumentError.new('inviting self') if instructor == current_user
 
--- a/dashboard/lib/services/lti.rb
+++ b/dashboard/lib/services/lti.rb
@@ -190,14 +190,14 @@ module Services
         )
         had_changes ||= (user.new_record? || user.changed?)
         user.save!
-        if account_type == ::User::TYPE_STUDENT
-          add_student_result = section.add_student(user)
-          had_changes ||= add_student_result == Section::ADD_STUDENT_SUCCESS
-          current_students.add(user.id)
-        else
+        if account_type == ::User::TYPE_TEACHER
           add_instructor_result = section.add_instructor(user)
           had_changes ||= add_instructor_result
           current_teachers.add(user.id)
+        else
+          add_student_result = section.add_student(user)
+          had_changes ||= add_student_result == Section::ADD_STUDENT_SUCCESS
+          current_students.add(user.id)
         end
       end
 
--- a/dashboard/lib/services/lti.rb
+++ b/dashboard/lib/services/lti.rb
@@ -3,6 +3,7 @@ require 'queries/lti'
 require 'user'
 require 'authentication_option'
 require 'sections/section'
+require 'set'
 
 module Services
   module Lti
@@ -171,8 +172,10 @@ module Services
       client_id = lti_integration.client_id
       issuer = lti_integration.issuer
       section = lti_section.section
+      current_students = Set.new
+      current_teachers = Set.new
 
-      current_students = nrps_members.map do |nrps_member|
+      nrps_members.each do |nrps_member|
         account_type = Policies::Lti.get_account_type(nrps_member[:roles])
 
         user = Queries::Lti.get_user_from_nrps(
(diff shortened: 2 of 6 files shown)


dashboard/app/models/sections/section.rb
```diff
@@ -578,7 +568,36 @@ def update_code_review_expiration(enable_code_review)
   end
   before_validation :strip_emoji_from_name
 
-  public def add_instructor(email, current_user)
+  # Adds an instructor to the section
+  # If the instructor was previously deleted, restore the instructor
+  # If the instructor was invited or inactive, reactive the instructor
+  # If the instructor did not previously exist, create the section instructor relationship
+  # Returns true if the instructor could be added, false otherwise
+  public def add_instructor(user)
```
artem-vavilov: In Ruby, if a method is not under `private` or `protected` blocks, it's `public` by default.
```suggestion
  def add_instructor(user)
```

dashboard/app/models/sections/section.rb
```diff
@@ -578,7 +568,36 @@ def update_code_review_expiration(enable_code_review)
...
+    if si.blank?
+      SectionInstructor.create!(section_id: id, instructor: user, status: :active)
+    elsif si.deleted?
+      si.restore
+      si.status = :active
+      si.save!
+    elsif si.status != 'active'
+      si.status = :active
+      si.save!
+    end
```
artem-vavilov:
```suggestion
    transaction do
      followers.find_by(student_user: user)&.destroy

      si = section_instructors.with_deleted.find_or_initialize_by(instructor: user)
      si.restore if si.deleted?
      si.active!
    end 
```

dashboard/app/models/sections/section.rb
```diff
@@ -578,7 +568,36 @@ def update_code_review_expiration(enable_code_review)
...
+
+    true
+  end
+
+  public def remove_instructor(user)
```
artem-vavilov, line 596:
```suggestion
  def remove_instructor(user)
```

dashboard/app/models/sections/section.rb
```diff
@@ -578,7 +568,36 @@ def update_code_review_expiration(enable_code_review)
...
+    SectionInstructor.find_by(instructor: user, section_id: id)&.destroy
+  end
+
+  public def invite_instructor(email, current_user)
```
artem-vavilov, line 600:
```suggestion
  def invite_instructor(email, current_user)
```

dashboard/lib/services/lti.rb
```diff
@@ -93,13 +93,23 @@ def self.lti_user_type(current_user, lti_integration, nrps_section)
...
+      email_address = get_claim(nrps_member_message, :email)
+      account_type = Policies::Lti.get_account_type(nrps_member[:roles])
+
       user = ::User.new
       user.provider = ::User::PROVIDER_MIGRATED
-      user.user_type = ::User::TYPE_STUDENT
       user.name = get_claim_from_list(nrps_member_message, Policies::Lti::STUDENT_NAME_KEYS)
-      user.family_name = get_claim(nrps_member_message, :family_name)
+
+      if account_type == ::User::TYPE_TEACHER && email_address.present?
```
carl-codeorg, line 106: Why do we check for an email address here? I know we get email addresses from Canvas, but it isn't technically required for platforms to provide it, so I just wanted to clarify the reasoning for requiring it here.

dashboard/app/models/sections/section.rb
artem-vavilov: The code change is just a suggestion, but please perform multiple mutation requests in a transaction.

dashboard/app/models/sections/section.rb
```diff
@@ -248,17 +248,7 @@ def update_student_sharing(sharing_disabled)
...
-      SectionInstructor.create!(section_id: id, instructor: user, status: :active)
-    elsif si.deleted?
-      si.restore
-      si.status = :active
-      si.save!
-    elsif si.status != 'active'
-      si.status = :active
-      si.save!
-    end
+    add_instructor(user)
```
etaderhold, line 251: This hook was meant to ensure that we never get into a state where a section would be left without an instructor, as the section owner will always be an active instructor.

`add_instructor` does not provide a strong guarantee that it will actually add an instructor (if the user is a student), so I worry this will introduce the possibility of introducing orphaned sections. I suggest not relaxing this guarantee here.

dashboard/app/models/sections/section.rb
```diff
@@ -578,7 +568,36 @@ def update_code_review_expiration(enable_code_review)
...
+    elsif si.status != 'active'
+      si.status = :active
+      si.save!
+    end
+
+    true
+  end
+
+  public def remove_instructor(user)
+    SectionInstructor.find_by(instructor: user, section_id: id)&.destroy
```
etaderhold: Please first check whether the user is the section owner. If you really want to remove them as an instructor anyway, you should reassign the section ownership to one of the remaining instructors. Handle the "no remaining instructors" case in a sensible way (possibly delete the section? but make sure not to do that if you're about to add a new instructor I guess)

dashboard/app/models/sections/section.rb
stephenliang: Thanks! I was following the convention as others were also using `public` in the file but I updated it.

dashboard/lib/services/lti.rb
stephenliang: Since the LTI email address is optional and Code.org requires teachers to have e-mail addresses, I made it so that teacher accounts can only be created if an email address is provided and the user's role is an instructor. In the case the LTI email is not provided, the user is created as a student. When the user first SSO Logins via the LMS the user will encounter the upgrade account pathway and we'll ask for their e-mail address at that time.

dashboard/app/models/sections/section.rb
stephenliang: I went ahead and removed the student check to be consistent with the prior logic.

dashboard/app/models/sections/section.rb
stephenliang: Added the check!

dashboard/app/models/sections/section.rb
```diff
@@ -578,7 +568,29 @@ def update_code_review_expiration(enable_code_review)
   end
   before_validation :strip_emoji_from_name
 
-  public def add_instructor(email, current_user)
+  # Adds an instructor to the section
+  # If the instructor was previously deleted, restore the instructor
+  # If the instructor was invited or inactive, reactive the instructor
+  # If the instructor did not previously exist, create the section instructor relationship
+  # Returns true if the instructor could be added, false otherwise
```
etaderhold: Not returning false anymore...

```suggestion
  # Make the instructor active if they had a different status
  # If the instructor did not previously exist, create the section instructor relationship
  # Returns true if successful
```

dashboard/test/models/sections/section_test.rb
```diff
@@ -905,6 +905,68 @@ def verify(actual, expected)
...
+
+  test 'remove_instructor does not remove the primary teacher' do
+    user = create :teacher
+    section = create(:section, user: user)
+
+    si = SectionInstructor.find_by(instructor: user, section_id: section.id)
+    section.remove_instructor(user)
+    si.reload
+
+    assert_equal false, si.deleted?
```
etaderhold:
```suggestion
    refute si.deleted?
```

dashboard/test/models/sections/section_test.rb
```diff
@@ -905,6 +905,68 @@ def verify(actual, expected)
     assert_equal 1, section.instructors.length
   end
 
+  test 'add_instructor returns true and adds the teacher if not previously a co-teacher' do
+    section = create(:section)
+    user = create :teacher
+
+    result = section.add_instructor(user)
+
+    assert_equal true, result
```
etaderhold:
```suggestion
    result = section.add_instructor(user)

    assert result
```

...or
```suggestion
    assert section.add_instructor(user)
```

dashboard/lib/services/lti.rb
```diff
@@ -160,31 +171,53 @@ def self.sync_section_roster(lti_integration, lti_section, nrps_members)
...
           follower.destroy
           had_changes = true
         end
       end
+
+      # Prune teachers who have been removed from the section in the LMS
+      section.section_instructors.each do |section_instructor|
+        instructor = section_instructor.instructor
+        unless current_teachers.include?(instructor.id)
+          section.remove_instructor(instructor)
```
etaderhold, line 216: If `remove_instructor` no longer touches the section owner this is going to silently leave the teacher who was removed from the LMS as a teacher on our platform a rather large percentage of the time. Reassigning the section ownership to one of the remaining instructors and then removing this instructor would probably be wise.

stephenliang: We just had a [discussion on this scenario](https://docs.google.com/document/d/1P4FMaH_bOgOgq0KFu_1DX2Wq-UZsr9WPA-wwWhipfnE/edit#heading=h.5v0tvhpmqrdt), there will be a popup that will resolve this exact scenario where the teacher was removed (or became a student) on the LMS and a new primary teacher needs to be assigned. There will be a popup that enables the next teacher to become promoted to the primary teacher before the sync could occur.  This followup work is being refined at the moment but we'll circle back and resolve this piece in the follow up if that works for you.

etaderhold: Sure, sounds good. I also think the idea of just auto-promoting the one triggering the sync instead of making a special dialog to ask the teacher would be fine as well.

stephenliang: Autopromoting makes sense if there are no other teachers, but it gets fairly complicated as there is a concept of "sections" on the LMS as well - each having their own teacher and we don't necessarily want one section's teacher to become the owner of others too. We created the [followup](https://codedotorg.atlassian.net/browse/P20-804) task and I've updated the PR to reflect that.

carl-codeorg (approved):
This LGTM. We probably also want a follow-up story to update the co-teacher section of the UI to disable the manual addition/deletion of co-teachers when syncing is enabled.

dashboard/lib/services/lti.rb
```diff
@@ -160,31 +171,53 @@ def self.sync_section_roster(lti_integration, lti_section, nrps_members)
...
           nrps_member: nrps_member
         )
-        had_changes ||= (student.new_record? || student.changed?)
-        student.save!
-        add_student_result = lti_section.section.add_student(student)
-        had_changes ||= add_student_result == Section::ADD_STUDENT_SUCCESS
-        student
+        had_changes ||= (user.new_record? || user.changed?)
+        user.save!
+        if account_type == ::User::TYPE_STUDENT
```
nicklathe: Nit: elsewhere in the code the condition we're checking for is `User::TYPE_TEACHER` and letting everything that doesn't fit that criteria fall to the `else` block which assumes the user is a student. Would it make sense to follow that pattern here to consistency?

dashboard/test/lib/services/lti_test.rb
```diff
@@ -277,9 +312,27 @@ class Services::LtiTest < ActiveSupport::TestCase
...
   test 'should create a student user given an LTI NRPS member object' do
-    student_user = Services::Lti.initialize_lti_student_from_nrps(client_id: @id_token[:aud], issuer: @id_token[:iss], nrps_member: @nrps_student)
-    assert student_user
-    assert_equal student_user.user_type, User::TYPE_STUDENT
+    user = Services::Lti.initialize_lti_user_from_nrps(client_id: @id_token[:aud], issuer: @id_token[:iss], nrps_member: @nrps_student)
+    assert user
+    assert_equal user.user_type, User::TYPE_STUDENT
+  end
+
+  test 'should create a teacher user if LTI does not provide email despite instructor role' do
```
nicklathe:
```suggestion
  test 'should create a student user if LTI does not provide email despite instructor role' do
```

etaderhold (approved):

dashboard/lib/services/lti.rb
stephenliang: I went ahead and inverted the condition, it does make sense as student is a safer assumption.

nicklathe (approved):

PR #57401, opened 2024-03-20
Migrate Maker tests to React Testing Library

fisher-alice:
This PR migrates `SetupGuideTest.js` and `SetupChecklistTest.js` to React Testing Library.

`SetupGuideTest.js` was a straightforward conversion. `SetupChecklistTest` was more involved because the component is re-rendered after time elapses. At first, the setup checker displays a 'waiting' icon, and after the checker has completed the checklist, 4 success icons are displayed. I referred to [this documentation](https://testing-library.com/docs/example-update-props/) on how to update the component props.
I added a `title` property to the two `FontAwesome` icons to be able to able to query by title.

## Links
[jira](https://codedotorg.atlassian.net/browse/LABS-539)

## Testing story
Tested that the new version of the tests passed.

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/test/unit/lib/kits/maker/ui/SetupChecklistTest.js
+++ b/apps/test/unit/lib/kits/maker/ui/SetupChecklistTest.js
@@ -19,11 +19,10 @@ import microBitReducer, {
 } from '@cdo/apps/lib/kits/maker/microBitRedux';
 import analyticsReporter from '@cdo/apps/lib/util/AnalyticsReporter';
 
+// Speed up the tests by reducing the artificial delay between steps
 const STEP_DELAY_MS = 1;
 
 describe('SetupChecklist', () => {
-  // Speed up the tests by reducing the artificial delay between steps
-
   beforeEach(() => {
     sinon.stub(utils, 'reload');
     sinon.stub(window.console, 'error');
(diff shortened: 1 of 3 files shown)


cnbrenci (approved):
lgtm 🚀

apps/test/unit/lib/kits/maker/ui/SetupChecklistTest.js
```diff
@@ -135,16 +124,27 @@ describe('SetupChecklist', () => {
  * @param {number} intervalMs - time to wait between steps
  * @return {Promise}
  */
-function yieldUntil(wrapper, predicate, timeoutMs = 3500, intervalMs = 5) {
+function yieldUntil(
```
levadadenys, line 127: nit: not a part of this pr probably, but rather a general comment. I wonder why this function here is declared in the bottom of the file instead of the top section since it's used in tests. I mean, of course, it work even now, but it's harder to read it this way I think

levadadenys (approved):
LGTM!

apps/test/unit/lib/kits/maker/ui/SetupChecklistTest.js
fisher-alice: Yeah - I see this also in `SetupGuideTest.js`. I'm not sure what the standard convention is. I actually prefer the function definitions at the bottom, but I can ask the group if there's a general preference during check-in. Thanks!

PR #57402, opened 2024-03-20
[Sprite Lab] Replace `buildVariableLog` with native function

mikeharv:
Some Sprite Lab levels rely upon a `variableLog` helper library in order to validate student variables. The functions in this library work by scraping student variable values from the `window` of the JS Interpreter's scope. This hack has been sort of working but has some weaknesses. For example, it incorrectly lists other globally defined variables (e.g. from helper libraries and block helper code) as student variables. This means that students automatically meet criteria for a level that count the number of defined variables.

By moving this function to native, we can access `Blockly` directly and use it to get the real student variables. When a variable is defined on a Blockly workspace, it persists in the map even if it's not selected in any block fields. If it's selected only in a disabled block, it still gets declared (as `undefined`) in generated code. Therefore, we need to filter the full list of Blockly variables to once that have a valid value based on running them through the interpreter. 

The blocks available in Sprite Lab allow students to set number, string, and boolean values to variables. A recently created bitmap block returns an array, so that type is technically supported as well. (Core Blockly provides other blocks for defining and manipulating lists, but to date we haven't included them in student levels. If you're curious, they are usable here: https://studio.code.org/s/2021drafting/lessons/5/levels/3)

The new `buildVariableLog` returns an object of the same shape as the original but without unwanted variables from helper code. By removing the original from the helper library, levels will use the new version automatically. This means we shouldn't need to update any levels.

Once confirming that this works as expected, I then confirmed that we could remove the `gamelab_changeVarBy` block. This block's helper code was actually the source of the original global `varLog` declaration (now moved into the `variableLog` library). The other block helper code doesn't seem to be used anywhere and the block is not listed in _any_ levels (rare even for experimental blocks).

## Links

Jira [CT-418: Provide an alternative to buildVariableLog that works with Google Blockly methods](https://codedotorg.atlassian.net/browse/CT-418)
Jira [CT-419: Deprecate changeVarBy block without breaking validation code that uses varLog](https://codedotorg.atlassian.net/browse/CT-419)

## Testing story

I tested this using the level at `/s/coursef-2023/lessons/10/levels/5`. I chose this level for its relatively complex validation code that includes multiple conditions that rely upon `varLog` and other functions in the helper library:

_(Irrelevant criteria omitted below)_
```
// This level requires the variableLog Helper Library

if (World.frameCount == 1) {
  varLog = buildVariableLog();

  addCriteria(function(){
    return Object.keys(varLog).length >0; // Accesses the global `varLog`.
  }, 'createVariable');
  addCriteria(function(){
    return !varLabelStartsWithUnderscore(); // Helper library function. Accesses the global `varLog`.
  }, 'renameVariable');

  addCriteria(function(){
    return detectVariableLogChange(); // Helper library function. Accesses the global `varLog`.
  }, 'changeVariableBlock');

  addCriteria(function(){
    return (
      // Passes the global `varLog` as well as `previousVarLog` which is deep copy of it..
      anySpeechIncludesValues(varLog, previousVarLog) && detectVariableLogChange()
    );
  }, 'sayVariable');
}

delayEndInActiveApp();
check();
```

## PR Checklist:

- [ ] Code is well-commented


apps/src/p5lab/spritelab/commands/validationCommands.js
```diff
@@ -320,6 +320,32 @@ export const commands = {
...
+      .getAllVariables();
+    const variableNames = blocklyVariables.map(blocklyVariable =>
+      Blockly.JavaScript.getName(blocklyVariable.name)
+    );
+
+    for (const name of variableNames) {
+      const value = this.getVariableValue(name);
+      if (
+        ['number', 'string', 'boolean'].includes(typeof value) |
+        Array.isArray(value)
```
mikeharv, line 341: If `value` is an array, `typeof value` is 'object`.

molly-moen (approved):
Nice!

PR #57404, opened 2024-03-20
Add Issuer column to the lti_user_identities table

mgc1194:
## Story
Currently, our `lti_user_identities` table has the following columns: `id`, `subject`,  `lti_integration_id`,  `user_id`,  `created_at`, `updated_at`, and `deleted_at`. 

When searching for a user through the `lti_user_identities` table using the content of an LTI's JWT, only the `subject` is present. 
The `subject` is only guaranteed to be unique within an LMS and we could have a situation where two users, from different LMSs, have the same `subject`.

To avoid collision we need to add another column to the `lti_user_identities` table to store the issuer of an LTI Identity.

## Changes
This PR includes a migration to the `lti_user_identities` where:
-  A new column `issuer` is added.
- The new column can not be null
-  It is indexed since it will be primarily used for querying the `lti_user_identities` table.

## Links
- Jira ticket: [P20-789](https://codedotorg.atlassian.net/browse/P20-789)

## Follow-up work
- Modifying the existing LTI code to store the issuer each time a user identity is created.

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/dashboard/app/models/lti_user_identity.rb
+++ b/dashboard/app/models/lti_user_identity.rb
@@ -9,10 +9,12 @@
 #  created_at         :datetime         not null
 #  updated_at         :datetime         not null
 #  deleted_at         :datetime
+#  issuer             :string(255)      not null
 #
 # Indexes
 #
 #  index_lti_user_identities_on_deleted_at          (deleted_at)
+#  index_lti_user_identities_on_issuer              (issuer)
 #  index_lti_user_identities_on_lti_integration_id  (lti_integration_id)
 #  index_lti_user_identities_on_subject             (subject)
 #  index_lti_user_identities_on_user_id             (user_id)
--- a/dashboard/db/migrate/20240320051410_add_issuer_to_lti_user_identities.rb
+++ b/dashboard/db/migrate/20240320051410_add_issuer_to_lti_user_identities.rb
@@ -1,0 +1,7 @@
+class AddIssuerToLtiUserIdentities < ActiveRecord::Migration[6.1]
+  def change
+    add_column :lti_user_identities, :issuer, :string
+    change_column_null :lti_user_identities, :issuer, false
+    add_index :lti_user_identities, :issuer
+  end
+end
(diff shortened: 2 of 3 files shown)


dashboard/db/migrate/20240320051410_add_issuer_to_lti_user_identities.rb
```diff
@@ -0,0 +1,7 @@
+class AddIssuerToLtiUserIdentities < ActiveRecord::Migration[6.1]
+  def change
+    add_column :lti_user_identities, :issuer, :string
+    change_column_null :lti_user_identities, :issuer, false
+    add_index :lti_user_identities, :issuer
```
artem-vavilov, line 5: Just a demonstration of how it can be accomplished in a single line
```suggestion
    add_column :lti_user_identities, :issuer, :string, null: false, index: true
```

mgc1194: Yeah, this is the first time I am doing migrations, at first, I was experimenting and just created the column.
Then I realized I could not be null and we needed to index it, but the column was already created in my local DB

artem-vavilov: You can always rollback your migration: `cd dashboard && rails db:rollback`

dashboard/app/models/lti_user_identity.rb
```diff
@@ -9,10 +9,12 @@
 #  created_at         :datetime         not null
 #  updated_at         :datetime         not null
 #  deleted_at         :datetime
+#  issuer             :string(255)      not null
```
daynew, line 12: This is already stored in the LtiIntegration associated with the LtiUserIdentity. Can we use that instead?

daynew (changes requested):

dashboard/app/models/lti_user_identity.rb
mgc1194: There is no direct association between `User` and `LtiIntegration`.
Given an `id_token` received by an LTI exchange, we would have to use the `authentication_options` model.

The `lti_user_identity` is the only model that fully maps `Users` to `LTI Integrations`

This model was built early in the project and we did not consider the possibility of collisions.
It is also logical to include the issuer of an Identity in the model.

PR #57405, opened 2024-03-20
Added components for empty section and strings for no curriculum

kobryan0619:
Figma screenshots:
image

image

UPDATED Video of it working:
Note, it goes through 4 different sections with different configurations of curriculum assigned or students in the section along with one "control" section that has students and a curriculum assigned.  The image and the css can be slow the first time loading, but I think it might just be my machine... open to ideas as to how to test that before production.

https://github.com/code-dot-org/code-dot-org/assets/24883357/695ea294-de02-4f5f-935a-38ca4a001a3f

## Links

[Ticket](https://codedotorg.atlassian.net/browse/TEACH-876)
[Jira](https://www.figma.com/file/gCktlnafGYQSNJafD3FUgA/Progress-Tracking?node-id=3618%3A9312&mode=dev)


dju90:
Could we have a little more space between the graphic and the "it's a little empty here" text?

dju90:
Also, let's change the last exclamation point for the No Students scenario to a period to match the No Curriculums scenario. (Generally I like to reserve exclamation points for marketing content)

apps/test/unit/templates/teacherDashboard/EmptySectionsTest.jsx
```diff
@@ -0,0 +1,63 @@
...
+      </Router>
+    );
+
+    expect(screen.getByAltText('empty desk')).to.be.visible;
+    expect(screen.getByText(i18n.addStudents())).to.be.visible;
+    expect(screen.queryByAltText('blank screen')).not.be.visible;
+    expect(screen.queryByText(i18n.browseCurriculum())).not.be.visible;
+  });
+
+  it('only displays headline and description when there are students and a curriculum is assigned', () => {
```
kobryan0619, line 50: Notably this component should never be rendered if there are students in the section and the curriculum is assigned... I don't know if this means I should delete this test, modify `EmptySection` so it returns null in this case, or... do something else.  Thoughts?

dju90:
LGTM! 🎉

apps/src/templates/teacherDashboard/TeacherDashboard.jsx
```diff
@@ -128,14 +128,28 @@ function TeacherDashboard({
           path={TeacherDashboardPath.standardsReport}
           component={props => applyV1TeacherDashboardWidth(<StandardsReport />)}
         />
-        {/* Break out of Switch if we have 0 students. Display EmptySection component instead. */}
         {studentCount === 0 && (
           <Route
-            component={props =>
-              applyV1TeacherDashboardWidth(
```
lfryemason, line 135: Great that we can get rid of this v1 width call!

apps/src/templates/teacherDashboard/TeacherDashboard.jsx
```diff
@@ -154,13 +168,28 @@ function TeacherDashboard({
...
-            component={() =>
-              applyV1TeacherDashboardWidth(
-                <div className={dashboardStyles.text}>
-                  <SafeMarkdown markdown={i18n.noProgressSection()} />
-                </div>
-              )
-            }
+            component={props => (
+              <div className={dashboardStyles.emptyClassroomDiv}>
+                {location.pathname === TeacherDashboardPath.progress && (
```
lfryemason: Could you explain why we need the different className for progress and non-progress pages? Shouldn't it looks the same in all tabs?

lfryemason: If we do need this change, could we move it to a helper function in this file to deduplicate with the above OR put the logic into EmptySection

apps/src/templates/teacherDashboard/TeacherDashboard.jsx
```diff
@@ -128,14 +128,28 @@ function TeacherDashboard({
...
+                      className={dashboardStyles.emptyClassroomProgress}
+                      hasStudents={false}
+                    />
+                  </div>
+                )}
+                {location.pathname !== TeacherDashboardPath.progress && (
+                  <EmptySection
+                    className={dashboardStyles.emptyClassroom}
+                    sectionId={sectionId}
+                    hasStudents={false}
```
lfryemason:
```suggestion
                    hasStudents={false}
                    hasCurriculumAssigned={false}
```
I think I'd prefer to explicitly set both of these each time. Alternatively, we could use an enum object in the EmptySection. e.g. emptyType.NO_STUDENTS vs emptyType.NO_ASSIGNED_CURRICULUM to send in.

apps/src/templates/teacherDashboard/TeacherDashboard.jsx
kobryan0619: Great question... they are different in order to take into account the `Progress` page has a `Progress` header at the top and the others don't.  I suppose this is a design choice I made.  Without additional margin on the non-Progress pages the image moves up the page when switching between tabs.  From a design perspective, I liked the consistency of the location of the image on the page.  I also recognize there might be a more elegant solution to this problem using CSS...

lfryemason (approved):

PR #57406, opened 2024-03-20
[Progress] Improve dashboard page load by removing `coursesWithProgress` from page data

lfryemason:
coursesWithProgress is only used for two things:
* determining whether to show the `No assigned course` page
* Populating the unit selection dropdown

However, coursesWithProgress is put into every teacher dashboard page.
This PR seeks to improve page performance for all teacher dashboard pages by removing coursesWithProgress from the initial load. Instead:
* we will use information in the `section` object to determine whether to show the `No assigned course` page
* We will call a new endpoint to retrieve `coursesWithProgress` for unit selectors.

Also:
* Adds tests for UnitSelectorV2
* Converts UnitSelector v1 to a function component and uses scss instead of styles
* Converts UnitSelector v1 tests to React Testing Library

## Links
[TEACH-841](https://codedotorg.atlassian.net/browse/TEACH-841)

## PR Checklist:

- [ ] Code is well-commented


apps/src/redux/unitSelectionRedux.js
```diff
@@ -48,10 +61,58 @@ export const doesCurrentCourseUseFeedback = state => {
...
+      dispatch(setCoursesWithProgress(reorderedCourses));
+      dispatch(finishedLoadingCoursesWithProgress());
+    })
+    .catch(err => {
+      console.error(err.message);
+      dispatch(finishedLoadingCoursesWithProgress());
+    });
+};
+
+function fetchJSON(url, params) {
```
etaderhold: Don't we have a function in a central library that could be used for this purpose?

apps/test/unit/templates/teacherDashboard/teacherSectionsReduxTest.js
```diff
@@ -425,10 +431,12 @@ describe('teacherSectionsRedux', () => {
       const stateWithSections = reducer(initialState, setSections(sections));
       assert.isNull(stateWithSections.sectionBeingEdited);
       const state = reducer(stateWithSections, beginEditingSection(12));
+      console.log('lfm', state.sectionBeingEdited, state.sections[12]);
```
etaderhold:
```suggestion
```

etaderhold (approved):

dashboard/app/models/sections/section.rb
```diff
@@ -523,6 +523,12 @@ def participant_unit_ids
     Unit.joins(:user_scripts).where(user_scripts: {user_id: students.pluck(:id)}).distinct.select {|s| s.course_assignable?(user)}.pluck(:id)
   end
 
+  # Returns true if any student in the section has ever made progress on a unit
+  # that the instructor of the section can be an instructor for.
+  def any_student_has_progress?
```
etaderhold, line 594: This method (and the one above) seems to look at whether the section owner can assign the course. They should probably use the current user to better handle co-teachers who may have different course-assigning abilities.

apps/src/templates/teacherDashboard/TeacherDashboard.jsx
```diff
@@ -202,7 +202,7 @@ TeacherDashboard.propTypes = {
   isSectionAssignedCurriculum: PropTypes.bool.isRequired,
   showAITutorTab: PropTypes.bool,
   sectionProviderName: PropTypes.string,
-
+  anyStudentHasProgress: PropTypes.bool,
```
etaderhold, line 217: Remove `isSectionAssignedCurriculum` above.

dashboard/app/models/sections/section.rb
lfryemason: I agree that this is probably a fix we want to make. I want to minimize the amount of changes that this PR makes, especially to existing logic. I think this fits better in a different PR.

etaderhold: Ok. Tracked in https://codedotorg.atlassian.net/browse/TEACH-981

etaderhold (approved):

apps/src/templates/sectionProgress/UnitSelector.js
```diff
@@ -1,50 +1,82 @@
...
-      </div>
-    );
-  }
+  return isLoadingCourses ||
+    !coursesWithProgress ||
+    coursesWithProgress.length === 0 ? (
+    loadingSkeleton()
+  ) : (
+    <div>
+      <select
```
levadadenys, line 42: nit: I see that changing this might be out of scope of this task, but I wonder if we can use SimpleDropdown here to slightly improve ui? Wasn't able to find this dropdown locally so not sure if DSCO SimpleDropdown is applicable here in terms of styles/visual appearance

levadadenys (commented):
LGTM Overall!
1 non-blocking nit/suggestion

PR #57410, opened 2024-03-20
Split up PanelsView for reuse

sanchitmalhotra126:
Splits up the existing `PanelsView` into a `PanelsView` component that is Lab2-agnostic and just renders a set panels, and a `PanelsLabView` container that pulls the data from Lab2 and renders the view. This is a prefactor for reusing this view on the levelbuilder edit page for live previewing panels content.

## Testing story

Tested locally. No change to existing panels levels.

### DIFF
--- a/apps/src/panels/PanelsLabView.tsx
+++ b/apps/src/panels/PanelsLabView.tsx
@@ -1,0 +1,60 @@
+// Panels
+//
+// This is a React client for a panels level.  Note that this is
+// only used for levels that use Lab2.
+
+import React, {useCallback} from 'react';
+import {useAppDispatch, useAppSelector} from '@cdo/apps/util/reduxHooks';
+import {
+  sendSuccessReport,
+  navigateToNextLevel,
+} from '@cdo/apps/code-studio/progressRedux';
+import {PanelsLevelData} from './types';
+import PanelsView from './PanelsView';
+import useWindowSize from '../util/hooks/useWindowSize';
+
+const appName = 'panels';
+
+const PanelsLabView: React.FunctionComponent = () => {
+  const dispatch = useAppDispatch();
+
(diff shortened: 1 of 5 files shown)


breville (approved):

apps/src/panels/types.ts
```diff
@@ -1,12 +1,15 @@
...
+
+export interface Panel {
+  imageUrl: string;
+  text: string;
+  nextUrl?: string;
+  layout?: PanelLayout;
+  key: string;
+}
+
+export type PanelLayout = 'text-bottom-left' | 'text-top-right';
```
levadadenys: nit: declare this type before declaring `Panel` mybe? So that it'll be easier to read types

apps/src/panels/PanelsLabView.tsx
```diff
@@ -0,0 +1,60 @@
...
+        window.location.href = nextUrl;
+      } else {
+        dispatch(sendSuccessReport(appName));
+        dispatch(navigateToNextLevel());
+      }
+    },
+    [dispatch]
+  );
+
+  const panels = (levelData as PanelsLevelData | undefined)?.panels;
```
levadadenys, line 42: nit: Just for the context, is there a way to don't use `as` here? (What makes us use as here?)

levadadenys (approved):
LGTM!

apps/src/panels/PanelsLabView.tsx
sanchitmalhotra126: Yeah, this is because the `levelData` attribute that the lab2 framework provides is just a base type that specific labs extend (like here with `PanelsLevelData`). That being said, it doesn't feel great to have to cast it everywhere it's used, so might be better to move these types up into Lab2 and create union types out of all of them.

PR #57421, opened 2024-03-20
Make updated pages work with RTL languages

kelbyhawn:
Adds logic to set a `dir` attribute on the html tag of updated pages that use the `responsive_full_width` theme (this is all of the pages that have been redesigned/refactored). This adds Right To Left language support on these pages automatically so the layout is mirrored correctly for RTL languages.

Eventually we'll want to add this site-wide, but for now this a good compromise since it may break the RTL layout on some legacy pages.

**Note:** some elements may break for RTL, like the old video carousels, but the majority of things should be ok. Jorge said it's ok for the video carousels to break for now. 

**Jira ticket:** [ACQ-1651](https://codedotorg.atlassian.net/browse/ACQ-1651)

----

## Example on https://code.org/csd
| LTR | RTL |
| ----- | ----- |
| LTR | RTL |

### DIFF
--- a/pegasus/sites.v3/code.org/public/about/index.haml
+++ b/pegasus/sites.v3/code.org/public/about/index.haml
@@ -4,6 +4,7 @@ nav: about_nav
 video_player: true
 theme: responsive
 style_min: true
+set_dir: true
 ---
 
 - facebook = {:u=>'https://youtu.be/mTGSiB4kB18'}
--- a/pegasus/sites.v3/code.org/themes/responsive_full_width.haml
+++ b/pegasus/sites.v3/code.org/themes/responsive_full_width.haml
@@ -1,6 +1,6 @@
 !!! 5
 
-- if @header['set_dir']
+- if ! @header['set_dir']
   - dir = language_dir_class
 
 %html{dir: dir}
--- a/pegasus/sites.v3/code.org/public/professional-learning/international-partners.haml
+++ b/pegasus/sites.v3/code.org/public/professional-learning/international-partners.haml
@@ -1,6 +1,5 @@
 ---
 theme: responsive_full_width
-set_dir: true
 ---
 
 :ruby
--- a/pegasus/sites.v3/code.org/public/professional-learning/international.haml
+++ b/pegasus/sites.v3/code.org/public/professional-learning/international.haml
@@ -1,6 +1,5 @@
 ---
 theme: responsive_full_width
-set_dir: true
 ---
 
 :ruby


pegasus/sites.v3/code.org/public/about/index.haml
```diff
@@ -4,6 +4,7 @@ nav: about_nav
 video_player: true
 theme: responsive
 style_min: true
+set_dir: true
```
kelbyhawn, line 7: Jorge requested the https://code.org/about page to also work for RTL languages, so I added this here even though this page has not been updated.

pegasus/sites.v3/code.org/themes/responsive_full_width.haml
```diff
@@ -1,6 +1,6 @@
 !!! 5
 
-- if @header['set_dir']
+- if ! @header['set_dir']
```
TurnerRiley, line 3: Do we need the `if` statement anymore or can we just have `- dir = language_dir_class`? I'm a bit confused on how adding the `!` adds this functionality

TurnerRiley (commented):
Just for my own understanding, what does it mean that the carousels "break" in some right-to-left languages? Like do the styles look a bit off or is it unusable?

pegasus/sites.v3/code.org/themes/responsive_full_width.haml
kelbyhawn: I think it might have to do with the ability to set this functionality on individual pages with the `set_dir: true` seen in the About page file in this PR. 

If there comes a day when this can be added sitewide, all this logic can go away 😅

kelbyhawn:
> Just for my own understanding, what does it mean that the carousels "break" in some right-to-left languages? Like do the styles look a bit off or is it unusable?

Unfortunately it disappears, which is obvs not good, but Jorge said it was ok for now. Once I do [this ticket](https://codedotorg.atlassian.net/browse/ACQ-1359) it'll fix the issue and they'll work w/ RTL languages.

pegasus/sites.v3/code.org/themes/responsive_full_width.haml
TurnerRiley: Will this break any pages with `set_dir` as `false` or `nil` since that will then pass this condition and set `dir` to `language_dir_class`? Or is that the intention? Sorry just want to understand!

TurnerRiley (approved):
Looks good, just trying to understand the change but my understanding is not blocking since it's working!

pegasus/sites.v3/code.org/themes/responsive_full_width.haml
kelbyhawn: Oh great catch! It will break on pages where `set_dir: true` (it still works when set_dir: false or nil), so I removed that setting from two recently updated pages [in this commit](https://github.com/code-dot-org/code-dot-org/pull/57421/commits/7c3a3205b67d98ba3697ab5e084e9b29dbaaa528). None of the other pages where this is applied uses that `set_dir` setting so they should be good.

PR #57422, opened 2024-03-20
Fix to load behavior picker blocks from xml correctly

molly-moen:
@onlinecsteacher found an [issue](https://codedotorg.slack.com/archives/C05DMS18MEX/p1710952615656619) where behavior picker blocks would not persist their choice in start blocks, ex if you select `wandering` it will reload as `driving with arrow keys` (this first choice in the list). This is because of some intricacies of how the behavior picker works. The behavior picker needs all the behaviors to be loaded into the hidden workspace to get the appropriate list of options. If it sees an option it doesn't know about, we usually will reject it and end up with the first option in the list. We did have an exception to this to allow for any option while loading, to enable behavior picker blocks in an uncategorized toolbox. However, that check was only that the hidden workspace exists, not that blocks have been loaded to it. What is happening in this situation is the hidden definition workspace exists, and we've started loading blocks. Since we have xml blocks (all start blocks are xml), we convert all the blocks to json first via an unheaded workspace. When we try to assign the behavior picker value, we think that we should be loaded and therefore do not allow an unknown option.

The fix for this was to add a `hasLoadedBlocks` flag to the google blockly wrapper and use that to decide if we should allow unknown behavior options in the behavior picker.

Most of the time behavior pickers seem to be set to `driving with arrow keys` in the start blocks, which is why we are only just discovering this issue. This block also isn't widely used.

## Links
- [slack thread](https://codedotorg.slack.com/archives/C05DMS18MEX/p1710952615656619)

## Testing story
Tested locally. I tested that behavior pickers set via start blocks now load correctly both in start mode and for users in normal mode.

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/blockly/googleBlocklyWrapper.ts
+++ b/apps/src/blockly/googleBlocklyWrapper.ts
@@ -669,6 +669,7 @@ function initializeBlocklyWrapper(blocklyInstance: GoogleBlocklyInstance) {
   };
 
   blocklyWrapper.inject = function (container, opt_options) {
+    blocklyWrapper.hasLoadedBlocks = false;
     if (!opt_options) {
       opt_options = {};
     }
--- a/apps/src/blockly/googleBlocklyWrapper.ts
+++ b/apps/src/blockly/googleBlocklyWrapper.ts
@@ -669,6 +669,7 @@ function initializeBlocklyWrapper(blocklyInstance: GoogleBlocklyInstance) {
   };
 
   blocklyWrapper.inject = function (container, opt_options) {
+    // Set the default value for hasLoadedBlocks to false.
     blocklyWrapper.hasLoadedBlocks = false;
     if (!opt_options) {
       opt_options = {};
--- a/apps/src/blockly/addons/cdoFieldBehaviorPicker.ts
+++ b/apps/src/blockly/addons/cdoFieldBehaviorPicker.ts
@@ -75,7 +75,7 @@ export default class CdoFieldBehaviorPicker extends CdoFieldDropdown {
       // Check whether the initial newValue option already exists
       const optionExists = options.some(option => option[0] === newValue);
       // The hidden workspace is created after the main workspace flyout is populated.
-      const loadingFinished = Blockly.getHiddenDefinitionWorkspace();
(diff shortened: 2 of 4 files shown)


apps/src/blockly/googleBlocklyWrapper.ts
```diff
@@ -669,6 +669,7 @@ function initializeBlocklyWrapper(blocklyInstance: GoogleBlocklyInstance) {
   };
 
   blocklyWrapper.inject = function (container, opt_options) {
+    blocklyWrapper.hasLoadedBlocks = false;
```
ebeastlake, line 673: Why do we need to set both `Blockly.hasLoadedBlocks` in `cdoUtils` and `blocklyWrapper.hasLoadedBlocks` in this wrapper to false to start?

mikeharv: My read is that this line sets it as the default state. We toggle it to true after loading blocks. But we also need to set it back to false whenever we might load more blocks onto the workspace (e.g. starting over).

It could be that this line (added early in the commit history) is now redundant, or it could be that we need here because we populate a flyout before actually loading blocks onto the workspace.

mikeharv (approved):
Looks good to me, after addressing Emily's question. Thanks for taking care of this!

apps/src/blockly/googleBlocklyWrapper.ts
molly-moen: @mikeharv that is all correct :) I did add the second one later when I realized start over wasn't working. I think we need it in both places to make sure toolbox blocks load correctly (originally that check was on the hidden definition workspace existing, now it is this flag).

ebeastlake: It could be worth commenting to indicate the two places where it is set serve slightly different purposes, but no strong feelings either way!

ebeastlake (approved):
Nice work, Molly! 🎉

apps/src/blockly/googleBlocklyWrapper.ts
molly-moen: added clarifying comments!

PR #57425, opened 2024-03-20
re-raise exceptions with added details

cat5inthecradle:
Reverts code-dot-org/code-dot-org#57424 to reimplement https://github.com/code-dot-org/code-dot-org/pull/57352

`exception.message` is a frozen string, and cannot be appended to, so I'm building a new message string.

### DIFF
--- a/lib/cdo/secrets.rb
+++ b/lib/cdo/secrets.rb
@@ -66,7 +66,7 @@ module Cdo
         parse_json(get_secret_value(client, key))
       rescue => exception
         modified_message = "#{exception.class}: #{exception.message} Key: #{key}"
-        raise modified_message
+        raise exception.class.new(modified_message)
       end
     end
 


nicklathe (approved):
LGTM, thanks!

Hamms (commented):
Looks like we've got some tests [which do indeed want us to preserve the class of the exception](https://github.com/code-dot-org/code-dot-org/blob/7c97b3f391c1798a705c655023eccbd05dcc41a6/lib/test/cdo/test_secrets.rb#L58). Did you have a specific reason why you wanted to avoid initializing another instance of the original exception class, or was it just a generic optimization?

cat5inthecradle:
> Looks like we've got some tests [which do indeed want us to preserve the class of the exception](https://github.com/code-dot-org/code-dot-org/blob/7c97b3f391c1798a705c655023eccbd05dcc41a6/lib/test/cdo/test_secrets.rb#L58). Did you have a specific reason why you wanted to avoid initializing another instance of the original exception class, or was it just a generic optimization?

Shoot, I missed that. It makes sense to re-raise the same class here.

lib/cdo/secrets.rb
```diff
@@ -65,8 +65,8 @@ def get(key)
       @values[key] ||= client_promise.then do |client|
         parse_json(get_secret_value(client, key))
       rescue => exception
-        exception.message << " Key: #{key}"
-        raise
+        modified_message = "#{exception.class}: #{exception.message} Key: #{key}"
+        raise exception.class.new(modified_message)
```
davidsbailey, line 69: this won't always work:
```
[development] dashboard > Aws::S3::Errors::NotFound.new('message')
/Users/dsb/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/aws-sdk-core-3.180.0/lib/aws-sdk-core/errors.rb:17:in `initialize': wrong number of arguments (given 1, expected 2..3) (ArgumentError)
```
(apologies if you received this twice)

cat5inthecradle: Ugh, that's right. I'll noodle on this for a bit

PR #57426, opened 2024-03-20
Remove null ID elements from Applab design html

sanchitmalhotra126:
We received a few tickets ([1](https://codeorg.zendesk.com/agent/tickets/491450), [2](https://codeorg.zendesk.com/agent/tickets/491363)) where students were seeing extraneous elements in their design mode HTML, all with an ID of "null". In one of these cases there were thousands of these extra elements, causing the page to run out of memory and crash. It's still unclear why/how these elements are getting added, but this should at least filter them out and log if we see this again.

Before:
Screenshot 2024-03-20 at 3 04 06 PM

After:
Screenshot 2024-03-20 at 3 23 38 PM

## Testing story

Tested with affected student projects code (copied to S3 bucket of a local test project).

### DIFF
--- a/apps/src/applab/designMode.js
+++ b/apps/src/applab/designMode.js
@@ -1227,6 +1227,7 @@ function sanitizeLevelDom(levelDom) {
     '#codeApp',
     '#visualizationColumn',
     '#visualizationResizeBar',
+    '#null',
     '.editor-column',
   ];
 


breville (approved):

cnbrenci:
> but this should at least filter them out and log if we see this again.

is it supposed to log too?

sanchitmalhotra126:
> is it supposed to log too?

Yep! We'll log down below if any of the disallowed selectors are found: https://github.com/code-dot-org/code-dot-org/blob/5328f9a815bb4d7e7f74fef6e2ccdc8c540fe0f1/apps/src/applab/designMode.js#L1242-L1248

PR #57432, opened 2024-03-20
Webpack HMR follow-up: allow localhost.code.org, remove one error overlay layer

bencodeorg:
A couple of follow-ups to https://github.com/code-dot-org/code-dot-org/pull/56882:
- silences errors observed when accessing pegasus (ie, localhost.code.org) pages on port 3000
- removes one of two error overlays, which were introduced by this PR

I'm not (at a glance) sure where the other error overlay comes from, but we could dig into it further.

I'm also admittedly getting a little bit turned around about why we need to add localhost.code.org as an allowedHost though -- if you're using port 3000, why is pegasus hitting port 9000/webpack dev server at all?

image

## Links

- Slack thread: https://codedotorg.slack.com/archives/C046G4TRLEN/p1710959415427739

## Testing story

Tested manually that I only saw 1 error overlay instead of 2 after throwing an error in some pegasus JS code. Also saw the error that Kelby described no longer appear.

### DIFF
--- a/apps/webpack.config.js
+++ b/apps/webpack.config.js
@@ -693,7 +693,8 @@ function createWebpackConfig({
     ],
     devServer: envConstants.HOT
       ? {
-          allowedHosts: ['localhost-studio.code.org'],
+          allowedHosts: ['localhost-studio.code.org', 'localhost.code.org'],
+          client: {overlay: false},
           port: WEBPACK_DEV_SERVER_PORT,
           proxy: [
             {


snickell (approved):
Lets merge! We can figure out the 2nd overlay just so we understand later, and then see how folks feel about keeping/removing it.

I kind of prefer the overlays in the past when using HMR, because you don't have to look anywhere to see if/when there's an error, but I could see it being annoying since we didn't have it on before and we might have a backlog of "uncaught exceptions"

bencodeorg:
Also, breadcrumb for disabling other overlay if we want to -- `webpack-dev-server` still says overlay is enabled:

image

bencodeorg:
Also for posterity, the theorized reason pegasus is trying to open a websockets connection at all is that it is consuming HMR-ready JavaScript from disk, which includes wrapper code to set up a websocket connection.

PR #57433, opened 2024-03-20
Revert change in get_user LTI query back to use authentication options

mgc1194:
In PR #57190 we switch the `get_user` LTI query from using User's Authentication Options to using the `lti_user_identities` table.

After a second thought, we realized we would have to add the `issuer` to the `lti_user_identities` table duplicating data stored in other tables, making the introduced changes less advantageous than we first thought.

In addition, the previous implementation follows our Authentication Options pattern for other methods such as Google, Clever, etc. where given an authentication credential we always try to find the Authentication Options that match the credentials.

This PR rolls back to the previous implementation and reverts the changes to the unit tests that used the `get_user` LTI query.

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/dashboard/lib/services/lti.rb
+++ b/dashboard/lib/services/lti.rb
@@ -33,8 +33,7 @@ module Services
       auth_option = user.authentication_options.find(&:lti?)
       issuer, client_id, subject = auth_option.authentication_id.split('|')
       lti_integration = Queries::Lti.get_lti_integration(issuer, client_id)
-      identity = LtiUserIdentity.new(user: user, subject: subject, lti_integration: lti_integration)
-      identity.save!
+      LtiUserIdentity.create(user: user, subject: subject, lti_integration: lti_integration)
     end
 
     def self.create_lti_integration(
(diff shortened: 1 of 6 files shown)


dashboard/lib/services/lti.rb
```diff
@@ -33,8 +33,7 @@ def self.create_lti_user_identity(user)
       auth_option = user.authentication_options.find(&:lti?)
       issuer, client_id, subject = auth_option.authentication_id.split('|')
       lti_integration = Queries::Lti.get_lti_integration(issuer, client_id)
-      identity = LtiUserIdentity.new(user: user, subject: subject, lti_integration: lti_integration)
-      identity.save!
+      LtiUserIdentity.create(user: user, subject: subject, lti_integration: lti_integration)
```
carl-codeorg, line 36: What made it so that you have to do do a new + save! here instead of a create?

nicklathe (approved):
Yeah, agree with the reasoning behind this revert. I think I probably rushed pushing for this change as it was related to the other bug we discovered, but didn't think about how it more broadly fits with other authentication option functionality. :shipit:

carl-codeorg (approved):

dashboard/lib/services/lti.rb
mgc1194: I probably introduced those changes by mistake while trying to create the LTI User Identity at the same time we finish creating the LTI User. Hence the revert.

PR #57437, opened 2024-03-21
Music: show filters for available sound types

breville:
The sound type filter values shown in the sound selector UI are now generated dynamically, only including sound types that are allowed in the current library.

### A library with vocals

Screenshot 2024-03-21 at 1 24 56 PM

### A library without vocals

Screenshot 2024-03-21 at 1 25 20 PM


apps/src/music/views/SoundsPanel.tsx
```diff
@@ -286,6 +289,27 @@ const SoundsPanel: React.FunctionComponent<SoundsPanelProps> = ({
     );
   }
 
+  const showFilterTypes: {[key: string]: boolean} = {
+    all: true,
+  };
+
+  allSoundEntries.forEach(soundEntry => {
```
sanchitmalhotra126: Could we do this calculation up front as soon as we have the library loaded? That would be less logic in this component and might be helpful to have this information accessible elsewhere too?

breville: Good idea; done.

breville:
A potential second round of improvements here could dynamically enable and disable the filter values for the currently-selected pack, and might even deselect a filter if it wasn't showing any results, but we can wait to see if that work seems necessary.

sanchitmalhotra126 (approved):
LGTM!

PR #57438, opened 2024-03-21
Music: update libraries

breville:
Some small updates to the music libraries:

- reorder drumkits
- alphabetize the sounds in packs again
- only one instance of `5nare` drum, in the Glitch kit

And a tiny code change to default to the regular drum kit.

The libraries have already been updated on S3.  The copies here serve as historic snapshots.


bencodeorg (approved):
Any value in taking these out of LFS so the diffs are easier to track?

sanchitmalhotra126 (approved):

breville:
> Any value in taking these out of LFS so the diffs are easier to track?

Oh that could be useful!  There's probably a bit more flux around these files, but if it settles down, I might do that!

PR #57445, opened 2024-03-21
[Sprite Lab] wall sprites don't push each other

mikeharv:
Sprites in the new Game Design lessons have a explicitly set `group` property in order simplify handling of sprite collisions, gravity, and jumping. It is desired that sprites in the `walls` group block all other sprites from moving and prevent them from overlapping. It's important that wall sprites not block other wall sprites or they will be pushed away. This adds the ability to get the list of sprites with no group so that the helper library can automatically have wall sprites block them. 

Before: 

After:

As the levels have been iterated upon, a couple other groups ("enemies", "objects") have been removed so I've cleaned up references to them in the helper library.

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/dashboard/config/libraries/zGameDev.interpreted.js
+++ b/dashboard/config/libraries/zGameDev.interpreted.js
@@ -1,32 +1,16 @@
 var GRID_SIZE = 8;
 var CELL_SIZE = 400 / GRID_SIZE;
-var GRAVITY = -0.5;
+var GRAVITY = -0.25;
 
 setDefaultSpriteSize(CELL_SIZE);
 
 // Handles sprite interactions every frame.
 function gameDevLoop() {
   collide('collide', {group: 'players'}, {group: 'walls'});
-  collide('collide', {group: 'objects'}, {group: 'walls'});
-  collide('collide', {group: 'enemies'}, {group: 'walls'});
-  collide("collide", ({costume: "all"}), ({group: "walls"}));
- 
-  edgesCollide({group: 'players'});
-  edgesCollide({group: 'enemies'});
-  edgesCollide({group: 'objects'});
-  gravity({group: 'players'}, GRAVITY);
-  gravity({group: 'enemies'}, GRAVITY);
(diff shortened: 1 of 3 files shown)


dashboard/config/libraries/zGameDev.interpreted.js
```diff
@@ -1,32 +1,16 @@
...
 setDefaultSpriteSize(CELL_SIZE);
 
 // Handles sprite interactions every frame.
 function gameDevLoop() {
   collide('collide', {group: 'players'}, {group: 'walls'});
-  collide('collide', {group: 'objects'}, {group: 'walls'});
-  collide('collide', {group: 'enemies'}, {group: 'walls'});
-  collide("collide", ({costume: "all"}), ({group: "walls"}));
- 
+  collide('collide', {group: ''}, {group: 'walls'});
```
molly-moen, line 10: for my understanding: a group of `''` means the sprite has no group? How does making this change mean wall sprites don't collide with wall sprites (what was making that happen before)?

mikeharv: The `group` property is undefined for sprites unless explicitly set. We're using `''` as a way to signal that we want to return the ones without a group. What was doing it before was this removed line:

`collide("collide", {costume: "all"}, {group: "walls"});`

Which can be read "all sprites collide with walls". Now, it's just "all sprites without a group collide with walls".

mikeharv: It's currently only possible to set a value to `sprite.group` using a dropdown, and none of the options use an empty string.

molly-moen: thanks! It might be worth adding a comment explaining this.

molly-moen (approved):

dashboard/config/libraries/zGameDev.interpreted.js
mikeharv: Will add!

PR #57443, opened 2024-03-21
Part 5: Auto-freeze/unfreeze project when featuring in public gallery

fisher-alice:
This PR automatically freezes a featured project when it is 'active', (i.e. featured in the updated public gallery) and unfreezes a featured project when it is 'archived', (i.e., unfeatured and removed from the public gallery).

Currently, projects can only be frozen/unfrozen by the project owner if they have project validator permission. They can do this via the 'Freeze for use as an exemplar' button in the gray admin box:

Screenshot 2024-03-21 at 10 34 39 AM

Featured projects have 3 statuses: bookmarked, active, and archived. Active featured projects are currently displayed in public gallery.

@samantha-code requested that when a featured project's status is updated to 'active' that it is automatically frozen, i.e., it cannot be edited by the owner. 
https://github.com/code-dot-org/code-dot-org/pull/56932 ensures that the owner will be notified that their project is being featured in the public gallery and that they can remix the project or request to have it be removed/unfrozen. Their project will remain visible in their 'My Projects' list.

This PR includes updates so that a featured project is automatically frozen if its status is 'active' and automatically unfrozen if its status is 'archived'.

Note that in `project.js` in the `freeze` function, when a project is frozen it is hidden. This will remain the same since it is the expected behavior of exemplar projects that are frozen by curriculum writers. This is in contrast to the freezing of active featured projects - users will still be able to see the frozen project under 'My Projects' with notification info.

Because featured projects are frozen, I updated `increment_abuse` in `projets.rb` so that when a featured project is set to 'active', its abuse score is first reset to 0, then 'incremented by' -50. I also updated the query to fetch active published featured projects to include projects with `abuse_score` value equal to 0 or less. Currently projects with `abuse_score` less than 0 are fetched. The data for featured project cards now includes the `abluse_score` value.

## Links
[jira](https://codedotorg.atlassian.net/browse/LABS-585)

## Testing story

1. SETUP:
- I tested with two different accounts locally. First account is a student account and second account is a teacher account with project validator permission.

2. FEATURING A PROJECT (freeze)
- I created a project in the student account, then opened the project URL with the project validator account and bookmarked it as a featured project.
- Then as a project validator, I went to 'projects/featured' and featured the project.
- I confirmed that the student project was listed under 'My Projects' in the student account with the notification that it is being featured and frozen.  

Screenshot 2024-03-21 at 12 17 52 PM

-I also confirmed that the student project was displayed in the public featured projects gallery in both the student AND project validator accounts.

[...]

PR #57451, opened 2024-03-21
Revert "Update public projects gallery / Remove ability to publish projects"

fisher-alice:
Reverts code-dot-org/code-dot-org#56666

I'd like to do more thorough testing including testing with an upcoming change https://github.com/code-dot-org/code-dot-org/pull/57443. Noticed a bug when a project is repeatedly featured/unfeatured and would like to investigate more throughly.

Also, the `test` build has been failing, so we will see if reverting this PR will resolve the issue.


tshaffercodeorg (approved):
Signing off on reversion to unblock Test Build

PR #57457, opened 2024-03-21
Add Stretch Curriculum Recommender framework

TurnerRiley:
The Curriculum Catalog page already has one curriculum recommender live: the similar curriculum recommender which finds a (no surprise here) similar curriculum to the one the user is currently viewing. The next recommender is designed to find a Stretch Curriculum that shares some fundamental similarities (such as which grade levels are taught) with the given curriculum but also tries to prioritize differences in the other fields (such as different school subjects) to find more of a "stretch."

This PR adds the Stretch Curriculum Recommender function, its scoring framework, and updates tests for it. As part of this work, I also fixed a quick scoring bug and pulled out some reused code into helper functions.

These changes won't be live anywhere yet, this simply adds the functionality for it. My next PR will be to actually implement this and add end-to-end testing.

## Links
Jira ticket: [here](https://codedotorg.atlassian.net/jira/software/c/projects/ACQ/boards/64?assignee=60d62161f65054006980bd71&selectedIssue=ACQ-1383)
Spec: [here](https://docs.google.com/document/d/1Ya0FaDTugMVLxAV_2OxdYgP4kkch-BGeq4prH4wzd7I/edit)
Scoring flowchart on Figma: [here](https://www.figma.com/file/JIzLzUAVjLCBKX9Jit1FRD/Curriculum-Recommendation-System?type=whiteboard&node-id=0-1&t=bGu8xk5y7YfFhZ9l-0)

## Testing story
Add unit tests.

## Follow-up work
Implement on the Curriculum Catalog page to show the recommended stretch curriculum in the expanded card below the similar curriculum.

PR #57460, opened 2024-03-21
Implement Stretch Curriculum Recommender on the Curriculum Catalog page

TurnerRiley:
This PR shows the recommended Stretch Curriculum on the expanded curriculum catalog cards. This builds off of the Stretch Curriculum Recommender function added in [this PR](https://github.com/code-dot-org/code-dot-org/pull/57457/files) and [the PR](https://github.com/code-dot-org/code-dot-org/pull/56690) that implemented the recommended Similar Curriculum into the expanded card.

## Demo
https://github.com/code-dot-org/code-dot-org/assets/56283563/3984be40-3040-4de7-aed8-ce8c5ee73d57

## Links
Jira ticket to show Stretch Curriculum recommendation: [here](https://codedotorg.atlassian.net/jira/software/c/projects/ACQ/boards/64?assignee=60d62161f65054006980bd71&selectedIssue=ACQ-1384)
Jira ticket to add end-to-end testing: [here](https://codedotorg.atlassian.net/jira/software/c/projects/ACQ/boards/64?assignee=60d62161f65054006980bd71&selectedIssue=ACQ-1385)

## Testing story
Update unit tests and add end-to-end testing.

## Follow-up work
Update eyes tests to check for the recommended Stretch Curriculum and add Amplitude events for the Stretch Curriculum.


apps/test/unit/templates/curriculumCatalog/CurriculumCatalogTest.jsx
```diff
@@ -715,31 +739,33 @@ describe('CurriculumCatalog', () => {
...
-        // next-most-recommended similar curriculum.
-        let recommendedSimilarCurriculum = similarCurriculumRecommendations[0];
-        if (
-          curriculaTaughtBefore.includes(
-            recommendedSimilarCurriculum.course_offering_id
-          )
-        ) {
-          recommendedSimilarCurriculum = similarCurriculumRecommendations[1];
-        }
-
```
TurnerRiley, line 742: Now that [filtering has been moved into the recommender](https://github.com/code-dot-org/code-dot-org/pull/56690#discussion_r1513589459) (previously, a given page was responsible for filtering out curricula the recommender shouldn't consider before claling the recommender but now its all-inclusive), there's no need to check here if the result has been taught before as it would have already been filtered out. The check below that ensures the recommendations aren't in the array of curricula taught before still remains.

apps/test/unit/templates/curriculumCatalog/CurriculumCatalogTest.jsx
```diff
@@ -681,25 +684,46 @@ describe('CurriculumCatalog', () => {
...
         )[0];
 
+        // Get the Stretch Recommended Curriculum for the current test curriculum. If the top stretch curriculum result is the same
+        // as the similar curriculum, then get the second stretch curriculum to match the logic in CurriculumCatalog.jsx.
+        const recommendedStretchCurriculumResults = getStretchRecommendations(
+          FULL_TEST_COURSES,
+          currCurriculum.key,
+          null
+        );
+        const recommendedStretchCurriculum =
```
bethanyaconnor: Is there a way we can make a test that forces the stretch curriculum to be the same as the recommended curriculum? It would be nice to test that that filtering works in the component itself.

TurnerRiley: Sounds good! Is this what you were thinking? https://github.com/code-dot-org/code-dot-org/pull/57460/commits/51f79df0d8376ab76127f3c364f3e910c2b4b408

bethanyaconnor: not quite. I was more thinking of something that tested [this line](https://github.com/code-dot-org/code-dot-org/blob/f8df927d2751ed810efcd88bb426505dbff1fde2/apps/src/templates/curriculumCatalog/CurriculumCatalog.jsx#L128). Unless I'm misunderstanding the test?

TurnerRiley: I think the test I wrote checks that since it's figuring out what the Expected next result should be (either if the similar curriculum is the same as the stretch curriculum or not, and adjusts accordingly). `CurriculumCatalog` will then run the logic you linked to on [this line](https://github.com/code-dot-org/code-dot-org/blob/f8df927d2751ed810efcd88bb426505dbff1fde2/apps/src/templates/curriculumCatalog/CurriculumCatalog.jsx#L128) and if the tests pass then the CurriculumCatalog is getting the curriculum we're expecting it to. I also check at the end to ensure that there is the right number of expected overlaps since when comparing against the 3rd curriculum, both Similar and Stretch return the same top curriculum. Please let me know if I'm misunderstanding this or what you're saying though!

bethanyaconnor: chatted offline and [these lines](https://github.com/code-dot-org/code-dot-org/blob/f8df927d2751ed810efcd88bb426505dbff1fde2/apps/test/unit/templates/curriculumCatalog/CurriculumCatalogTest.jsx#L713-L728) test that filtering!

apps/src/templates/curriculumCatalog/ExpandedCurriculumCatalogCard.jsx
```diff
@@ -308,6 +314,23 @@ const ExpandedCurriculumCatalogCard = ({
                   text={recommendedSimilarCurriculum.display_name}
                   onClick={handleClickRecommendedSimilarCurriculum}
                 />
+                <img
+                  id="stretchCurriculumImage"
+                  src={recommendedStretchCurriculum.image || defaultImageSrc}
+                  style={{height: '100%'}}
```
bethanyaconnor: nit: could this be moved to scss?

bethanyaconnor (approved):
one nit but otherwise looks good! Thanks for your patience :)

apps/src/templates/curriculumCatalog/ExpandedCurriculumCatalogCard.jsx
TurnerRiley: Yep! I actually tested it and adding `height: 100%` to the `img` doesn't do anything here! So I'll just remove it

PR #57462, opened 2024-03-21
Do not update statsig user data on staging

hannahbergam:
Staging is getting a race condition failure because we're not using a staging server for Statsig but the userproperties is getting called to update. This pulls the userUpdate call into the same logic as initialize, so we will never try to update a user if we haven't called initialize first. 

## PR Checklist:

- [ ] Code is well-commented


bethanyaconnor (approved):
Seems reasonable!

PR #57464, opened 2024-03-21
Create WYSIWYG editor for Panels levels

sanchitmalhotra126:
This adds a live previewing, WYSIWYG editor for the "panels" level type in levelbuilder. Because this editor is in React, we can use the `PanelsView` component directly and render the experience as the user would see it in the level.

This also silently saves `panels` data to level properties instead of `levelData` for the purposes of localization, similar to https://github.com/code-dot-org/code-dot-org/pull/57211. The motivation here again is to make the data more easily accessible and translatable.

Editing an existing Panels level:

https://github.com/code-dot-org/code-dot-org/assets/85528507/c53bb985-77bf-4702-9327-772023897fe3

Creating Panels from scratch:

https://github.com/code-dot-org/code-dot-org/assets/85528507/e050e41c-ad48-4210-82d4-61454e000a1d

## Testing story

Tested editing existing panels levels in the music lab intro progression, and creating new panels levels from scratch.

### DIFF
--- a/apps/src/lab2/levelEditors/panels/EditPanels.tsx
+++ b/apps/src/lab2/levelEditors/panels/EditPanels.tsx
@@ -134,7 +134,7 @@ const EditPanels: React.FunctionComponent<EditPanelsProps> = ({
           type="button"
           onClick={addPanel}
           text="Add Panel"
-          color="grey"
+          color="gray"
           icon="plus"
         />
       </div>
--- a/apps/src/lab2/levelEditors/panels/EditPanels.tsx
+++ b/apps/src/lab2/levelEditors/panels/EditPanels.tsx
@@ -1,0 +1,280 @@
+import {Panel, PanelLayout} from '@cdo/apps/panels/types';
+import React, {useCallback, useEffect, useRef, useState} from 'react';
+import moduleStyles from './edit-panels.module.scss';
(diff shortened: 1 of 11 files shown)


apps/src/lab2/levelEditors/panels/EditPanels.tsx
```diff
@@ -0,0 +1,280 @@
...
+}) => {
+  const [panels, setPanels] = useState<Panel[]>(
+    sanitizePanels(initialPanels, levelName)
+  );
+  const [toastMessage, setToastMessage] = useState('');
+  const [toastIndex, setToastIndex] = useState(0);
+
+  const updatePanel = useCallback(
+    (panel: Panel, index: number) => {
+      setPanels([...panels.slice(0, index), panel, ...panels.slice(index + 1)]);
```
breville: Maybe a comment explaining what this is doing? :)

breville (approved):
This is absolutely fantastic!  Can't wait to see it being used.

Thanks for adding support for the custom image uploader.  We need it in more places than the single editor it's currently been in!

apps/src/lab2/levelEditors/panels/EditPanels.tsx
sanchitmalhotra126: Actually looking at this again, realized I should be using the panel `key` instead of `index`!

PR #57465, opened 2024-03-22
Handle no toolbox levels created via Google Blockly

molly-moen:
We found an issue where levels edited with Google Blockly that were supposed to have no toolbox ended up with a tiny blank toolbox. The underlying issue is if you create an empty toolbox via Google Blockly instead of saving `<xml />` as the toolbox blocks, `<xml xmlns="https://developers.google.com/blockly/xml"/>` is saved. This is an issue because we manually turned `<xml />` into an `undefined` toolbox to avoid this tiny empty toolbox. The quick fix for this was to also turn the version with the xmlns attribute into an `undefined` toolbox.

This is still a bit brittle, because if this attribute changes we could end up in the same situation again. We may want to use a regular expression instead.

## Links

- jira ticket: [CT-402](https://codedotorg.atlassian.net/browse/CT-402)

## Testing story
Tested locally that the new version of no toolbox works.

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/StudioApp.js
+++ b/apps/src/StudioApp.js
@@ -2841,7 +2841,8 @@ StudioApp.prototype.handleUsingBlockly_ = function (config) {
 
   // If levelbuilder provides an empty toolbox, some apps (like artist)
   // replace it with a full toolbox. I think some levels may depend on this
-  // behavior. We want a way to specify no toolbox, which is <xml></xml>
+  // behavior. We want a way to specify no toolbox, which is <xml></xml>.
+  // Google Blockly may also add a xmlns attribute to this xml.
   if (config.level.toolbox) {
     // Update CDO Blockly XML so it is compatible with mainline Google Blockly
     // (Nothing is changed if we are using CDO Blockly.)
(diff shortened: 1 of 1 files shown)


apps/src/StudioApp.js
```diff
@@ -2850,9 +2850,13 @@ StudioApp.prototype.handleUsingBlockly_ = function (config) {
...
 
     const toolboxWithoutWhitespace = config.level.toolbox.replace(/\s/g, '');
+    // An empty toolbox should be treated as no toolbox.
     if (
       toolboxWithoutWhitespace === '<xml></xml>' ||
-      toolboxWithoutWhitespace === '<xml/>'
+      toolboxWithoutWhitespace === '<xml/>' ||
+      // Google Blockly always adds the xmlns attribute.
+      toolboxWithoutWhitespace ===
+        '<xmlxmlns="https://developers.google.com/blockly/xml"/>'
```
mikeharv: Is there a missing space between xml and xmlns?

molly-moen: There is because `toolboxWithoutWhitespace` strips all whitespace from the string (probably to avoid diffferences in spacing around `\>`)

mikeharv: Got it. And it doesn't look like we use `toolboxWithoutWhitespace` anywhere else, so that's good. Is it possible Google Blockly would ever give separate open/close tags like `<xml xmlns="..."></xml>`? I'm guessing we haven't seen it.

molly-moen: They don't currently, but that doesn't mean they won't. I can add it as an option

mikeharv (approved):

PR #57470, opened 2024-03-22
Speed up the Dashboard test env loading

artem-vavilov:
The Dashboard test env loads 5.6 times faster with English i18n files only:
| Before  | After |
| ------- | ----- |
| before | after |


artem-vavilov:
https://github.com/code-dot-org/code-dot-org/pull/57661

PR #57473, opened 2024-03-22
Remove user-publish, update public gallery and display notification banner (parts 1,2,3)

fisher-alice:
This PR consists of 3 PRS that have been approved:
Part 1: Remove ability to publish projects (description and write-up included in https://github.com/code-dot-org/code-dot-org/pull/56666)
Part 2: Update projects public gallery to display only featured projects https://github.com/code-dot-org/code-dot-org/pull/56886
Part 3: Add information banner about updated public gallery for signed-in users https://github.com/code-dot-org/code-dot-org/pull/56918

There are two new commits:
- 554ff8259af67140a9dfb4dd48c05568389e2f7a - After consulting with @samantha-code , we are removing the report abuse icon on the project card in the public gallery 
- d7cec207b56bc2f421b1f663a4e81925a7736d75 - We are allowing featured projects with `abuse_score` equal to 0 or less to be featured. Prior to this commit, only featured projects with `abuse_score` less than 0 were displayed.

I merged https://github.com/code-dot-org/code-dot-org/pull/56666, but then reverted suspecting that the PR was causing a`test` issue [Slack thread](https://codedotorg.slack.com/archives/C0T0PNTM3/p1711040741216309). After the revert, the `test` build passed so most likely, the bug was caused by my PR. Neither I nor @Hamms can replicate the bug locally though on this branch and as you can see, the Drone build has been passing.

After consulting with @Hamms about the mysterious bug, I will: 
- open an adhoc on this branch to see if there are any build issues,
- try merging again after a day's DTP and to see if we see the same `test` build failure. 

 
## Follow-up work
https://github.com/code-dot-org/code-dot-org/pull/57476 - part 4 (adding notification when a user's project is frozen) is now in a separate PR along with related changes (auto-freezing a featured project when 'active').

## PR Checklist:

- [ ] Code is well-commented


apps/src/templates/projects/PublicGallery.jsx
```diff
@@ -42,18 +33,22 @@ class PublicGallery extends Component {
    * See the PropTypes of each component for a definition of each format.
    */
   mapProjectData(projectLists) {
-    return _.mapValues(projectLists, projectList => {
-      return projectList.map(projectData => {
-        return {
+    let allFeaturedProjects = [];
+    for (const projectListName in projectLists) {
+      projectLists[projectListName].forEach(projectData => {
+        allFeaturedProjects.push({
```
thomasoniii: You don't need to change this, but as a suggestion - 

two things jump out - one is that I just don't care for mixing explicit loops and methods. So either it should be two nested for loops or two nested `forEach` loops.

But moreso, it's good to get rid of using mutable variables defined externally outside of loops/callbacks, where possible. I think this'll do the same thing for you:

```
const allFeaturedProjects = projectLists.flatMap(
  projectListName => projectListName.flatMap(
    projectData => (
    {
      projectData: {
        ...projectData,
        publishedToPublic: true,
        publishedToClass: false,
      },
      currentGallery: 'public',
    }
  )
);
```

thomasoniii (approved):
minor non-blocking suggestion about getting rid of a loop external mutable variable (and note that my code suggestion may not be quite right and would definitely need to be tested), but otherwise lgtm 🚀

apps/src/templates/projects/PublicGallery.jsx
fisher-alice: Thanks Jim! `projectLists` is not an array but an object so I first create an array or project data objects and then transform the data. I also updated the comment to hopefully better clarify this function.

PR #55912, opened 2024-01-23
[Google Blockly] Fix: Costumes added from inside function editor are not usable

molly-moen:
We found in our bug bash that already created function blocks cannot use new costumes. The fix for this is to override the `doClassValidation` method in `cdoFieldImageDropdown` to not cache the options list. Since this only occurs on changing the dropdown value, it felt safe to not use the cache here. In Sprite Lab, `FieldImageDropdown` is used for the background and costume pickers, which both can update dynamically. It looks like it may also be used for a couple blocks in flappy and bounce, but those blocks do not seem to be used in the "new project" for flappy and bounce.
 
## Links
- jira ticket: [CT-275](https://codedotorg.atlassian.net/browse/CT-275)

## Testing story
Tested locally that we can now use a new costume in a pre-existing costume picker in the function editor workspace.

## PR Checklist:

- [ ] Tests provide adequate coverage
- [ ] Privacy and Security impacts have been assessed
- [ ] Code is well-commented
- [ ] New features are translatable or updates will not break translations
- [ ] Relevant documentation has been added or updated
- [ ] User impact is well-understood and desirable
- [ ] Pull Request is labeled appropriately
- [ ] Follow-up work items (including potential tech debt) are tracked and linked

PR #57475, opened 2024-03-22
[Google Blockly] extend FieldImageDropdown for costume and background pickers

mikeharv:
Similar to:
* https://github.com/code-dot-org/code-dot-org/pull/57344

A common levelbuilder practice is to edit the Animation JSON for a level after the blocks are set. This causes a flood of warnings in the console about an unavailable option in the image dropdown field. Since we already handle this gracefully and the curriculum team understands how it works, it feels reasonable to skip this warning. 

Removing this warning from all image dropdowns didn't feel ideal. This field type is used in Flappy, Bounce, Artist, and Play Lab, but it's only in Sprite Lab where the animations can be changed. Some of these labs aren't even migrated yet, and it could be helpful to have those warnings in place once that work is begun. For this reason I decided to extend the class specifically for the blocks that deal with dynamic animations, ie. `behaviorPicker` and `costumePicker` custom input type.

The last time we added an override to `CdoFieldImageDropdown` was also related to these blocks. https://github.com/code-dot-org/code-dot-org/pull/55912
We needed to add an override for `doClassValidation` to not cache the options list so that recently added animations would be usable in the function editor workspace. At the time, we decided it should be fine to do this for all image dropdowns, but acknowledged it was potentially overkill for cases where the list of images is static: https://github.com/code-dot-org/code-dot-org/pull/55912#discussion_r1465049544

The new class preserves our previous override, but without the warning. The warning _is_ part of the original `FieldGridDropdown` from the Blockly plugin, so it seems like we can also remove the `doClassValidation` override from `CdoFieldImageDropdown`. 

The same warning also exists in our custom classes `FieldBehaviorPicker` and `FieldDropdown`. I think it’s probably fine to leave those for now since we’d be less likely to find that the options have swapped out from under us in those cases.

## Links

Slack thread: https://codedotorg.slack.com/archives/C05DK21DAHK/p1711047168603369

## Testing story

I tested this by first confirming that sprite/background blocks were indeed using the new field type as expected. I tested general use, including adding animations from within the function editor to ensure that our previous fix was still working. I also observed that the warnings were gone in a level that previously had them.
imageimage

I also tested a Bounce level that uses our custom image dropdown and saw no regressions. By manually setting an invalid value into the start blocks of the level, I was able to see the console warning as before, now being produced by the super/plugin class:
imageimage

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/blockly/addons/cdoFieldImageDropdown.ts
+++ b/apps/src/blockly/addons/cdoFieldImageDropdown.ts
@@ -93,7 +93,8 @@ export class CdoFieldImageDropdown extends FieldGridDropdown {
         buttonElement.innerHTML = button.text;
         buttonElement.addEventListener('click', button.action);
         buttonElement.addEventListener('click', () =>
-          Blockly.DropDownDiv.hideIfOwner(this, true)
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
+          Blockly.DropDownDiv.hideIfOwner(this as any, true)
         );
         const menuItem = new Blockly.MenuItem(buttonElement, '');
         menuItem.setRole(Blockly.utils.aria.Role.OPTION);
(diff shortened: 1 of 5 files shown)


apps/src/blockly/addons/cdoFieldImageDropdown.ts
```diff
@@ -141,43 +142,11 @@ export class CdoFieldImageDropdown extends FieldGridDropdown {
     (this as any).updateColumnsStyling_();
 
     Blockly.DropDownDiv.showPositionedByField(
-      this,
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      this as any,
```
mikeharv, line 146: @molly-moen When I removed the `doClassValidation_` override (below), I started to get the following error:

```
Argument of type 'this' is not assignable to parameter of type 'Field<string | undefined>'.
```

For now, I've listed `any` as the type, which we were already doing in four other places within `showEditor_`. I'm not actually sure why it wasn't complaining before I removed the other method, but the change here and on L96-97 seem to resolve it.

molly-moen: Here is the full error message:
```
Argument of type 'this' is not assignable to parameter of type 'Field<string | undefined>'.
  Type 'CdoFieldImageDropdown' is not assignable to type 'Field<string | undefined>'.
    Types of property 'validator_' are incompatible.
      Type 'FieldValidator<string> | null' is not assignable to type 'FieldValidator<string | undefined> | null'.
        Type 'FieldValidator<string>' is not assignable to type 'FieldValidator<string | undefined>'.
          Types of parameters 'newValue' and 'newValue' are incompatible.
            Type 'string | undefined' is not assignable to type 'string'.
              Type 'undefined' is not assignable to type 'string'.ts(2345)
```

`Type 'FieldValidator<string> | null' is not assignable to type 'FieldValidator<string | undefined> | null'.` is the key part here I think. I _think_ this is a Blockly issue of invalid typing. `FieldGridDropdown` assigns `export type FieldDropdownValidator = FieldValidator<string>;`. I don't know why it works for them to call `dropdownDiv` with `this`, maybe they have looser typing rules than we do? Either way, I think it's fine to use any here, maybe just add a comment explaining that the validator type on the parent is invalid.

molly-moen (approved):
Looks good!

PR #57480, opened 2024-03-22
Revert "Revert "Datablock Storage: replace firebase for applab data features (#56279)"

snickell:
This PR reverts #57471 , which reverted the original datablock storage PR #56279 after it broke the `getColumn` block.

Fix `getColumn()`, which was producing debug console errors like "You tried to get a column called _ from a table called _ but that column doesn't exist.":
1. [71bde1b](https://github.com/code-dot-org/code-dot-org/pull/57480/commits/71bde1b3bd9ac6aadaa1f142a7119b9d9abd397a): `applabCommands.getColumn()`: now calls `Applab.storage.getColumn()` rather than `Applab.storage.readRecords()` like it did before datablockStorage.js:
    image
2. [d834f5b](https://github.com/code-dot-org/code-dot-org/pull/57480/commits/d834f5b4005589076eab7feb39b126d995b7510e): ec_datablocks.js: Adds an apps integration test for the getColumn block, esp since this block is what our curriculum actually uses.

Fix "Daily Weather" and "Spotify" current tables shows up with no records in the data browser:
1. [116e480](https://github.com/code-dot-org/code-dot-org/pull/57480/commits/116e4806ed4883881fdd3a8d56960415ba8201db): pass tableType="shared" from firebaseStorage.js back to addTableName() when registering "current tables" with redux

For more details on datablock storage, see: #56279.

### DIFF
--- a/apps/src/applab/Exporter.js
+++ b/apps/src/applab/Exporter.js
@@ -89,9 +89,9 @@ const APP_OPTIONS_ALLOWLIST = {
   hideSource: true,
   share: true,
   labUserId: false,
-  firebaseName: false,
-  firebaseAuthToken: false,
-  firebaseChannelIdSuffix: false,
+  firebaseName: false, // TODO: post-firebase-cleanup: #56994
+  firebaseAuthToken: false, // TODO: post-firebase-cleanup: #56994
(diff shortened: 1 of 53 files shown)


cnbrenci (approved):
Great to have test coverage for this now 🥳

PR #57481, opened 2024-03-22
Updates User Configuration and Turns off Statsig auto js error logging

hannahbergam:
When Statsig is initialized, it defaults to logging all js errors. This means that now our integration is working and we're logging lots of errors! The options were being logged to the user object. This pointed to a bigger issue with the way we were planning to keep track of user data. On Amplitude, we send it through Current User Redux and that persists across a user's session. For Statsig, they strongly recommend sending the user in during the initialize() call and 'do not maintain or update user objects'. That led to this new solution:

1. Load the user id and user type into the dom to be queried by the same query selector element
2. If an id exists, populate a user object with the id and type, and send through to initialize
3. If no id exists, send an empty user object 

This way, the options parameter will always be the third parameter so flags like disableErrorLogging will work properly. A bonus: this now more closely matches our Statsig ruby implementation, where the user is sent in on initialize.

Additionally, I am updating the managed_test_server param to its own variable before sending it through, which I believe will enable our test server to send events.

Screenshot 2024-03-22 at 4 08 58 PM

These errors do not count toward our billable events (see images from slack conversation with Statsig directly below).
Screenshot 2024-03-22 at 12 29 23 PM
Screenshot 2024-03-22 at 12 35 51 PM

## Links

- spec: []()
- jira ticket: https://codedotorg.atlassian.net/browse/ACQ-1660

## Testing story

I was able to show that initializing with user data makes it so every page refresh has a user_id, not just when we have a login (and the single send from current user redux). See image below. 

I was also able to verify that pegasus reporting still works as anticipated (I updated the amplitude event for 'page viewed' to send to statsig and it correctly sent with an empty user object). See second image below!

Screenshot 2024-03-22 at 7 40 35 PM
Screenshot 2024-03-22 at 7 50 15 PM

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/lib/util/StatsigReporter.js
+++ b/apps/src/lib/util/StatsigReporter.js
@@ -44,7 +44,8 @@ class StatsigReporter {
     this.initialize(api_key, user, options);
   }
 
-  // Initialize with a null user object- current user redux will update on sign in
+  // This user object will potentially update via a setUserProperties call
+  // (below) from current user redux
   async initialize(api_key, user, options) {
     if (this.shouldPutRecord(ALWAYS_SEND)) {
       await Statsig.initialize(api_key, user, options);
--- a/apps/src/lib/util/StatsigReporter.js
+++ b/apps/src/lib/util/StatsigReporter.js
@@ -31,9 +31,10 @@ class StatsigReporter {
     this.initialize(api_key, options);
   }
 
+  // Initialize with a null user object- current user redux will update on sign in
   async initialize(api_key, options) {
     if (this.shouldPutRecord(ALWAYS_SEND)) {
-      await Statsig.initialize(api_key, options);
+      await Statsig.initialize(api_key, {}, options);
     }
   }
 
--- a/apps/src/lib/util/StatsigReporter.js
+++ b/apps/src/lib/util/StatsigReporter.js
@@ -26,6 +26,7 @@ class StatsigReporter {
     const options = {
       environment: {tier: getEnvironment()},
       localMode: this.local_mode,
+      disableErrorLogging: true,
     };
     this.initialize(api_key, options);
   }
--- a/apps/src/lib/util/StatsigReporter.js
+++ b/apps/src/lib/util/StatsigReporter.js
@@ -13,10 +13,12 @@ const NO_EVENT_NAME = 'NO_VALID_EVENT_NAME_LOG_ERROR';
 class StatsigReporter {
   constructor() {
     let user = {};
-    user_id_element = document.querySelector('script[data-user-id]');
-    user_id = user_id_element ? user_id_element.dataset.userId : null;
-    user_type_element = document.querySelector('script[data-user-type');
-    user_type = user_type_element ? user_type_element.dataset.userType : null;
+    const user_id_element = document.querySelector('script[data-user-id]');
+    const user_id = user_id_element ? user_id_element.dataset.userId : null;
+    const user_type_element = document.querySelector('script[data-user-type');
+    const user_type = user_type_element
+      ? user_type_element.dataset.userType
+      : null;
     if (user_id) {
       user = {
         userID: this.formatUserId(user_id),
--- a/apps/src/lib/util/StatsigReporter.js
+++ b/apps/src/lib/util/StatsigReporter.js
@@ -12,6 +12,17 @@ const NO_EVENT_NAME = 'NO_VALID_EVENT_NAME_LOG_ERROR';
 
 class StatsigReporter {
   constructor() {
+    let user = {};
+    user_id_element = document.querySelector('script[data-user-id]');
+    user_id = user_id_element ? user_id_element.dataset.userId : null;
+    user_type_element = document.querySelector('script[data-user-type');
+    user_type = user_type_element ? user_type_element.dataset.userType : null;
+    if (user_id) {
+      user = {
+        userID: this.formatUserId(user_id),
+        custom: {type: user_type},
+      };
+    }
     const api_element = document.querySelector(
       'script[data-statsig-api-client-key]'
     );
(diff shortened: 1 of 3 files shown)


bethanyaconnor (approved):
Looks good based on [the documentation](https://docs.statsig.com/metrics/autogenerated#statsigapp_error)!

apps/src/lib/util/StatsigReporter.js
```diff
@@ -12,6 +12,17 @@ const NO_EVENT_NAME = 'NO_VALID_EVENT_NAME_LOG_ERROR';
 
 class StatsigReporter {
   constructor() {
+    let user = {};
+    user_id_element = document.querySelector('script[data-user-id]');
```
bethanyaconnor: We'll want to be careful about relying on this on cached pages, where it'll become out of date. The nice thing about using userRedux is that it's updated for every page load so it was always (eventually) up to date. 

I'm curious if we can still use the user passed through from userRedux, but save that user id in this class so that we can send it with events? Or is there a race condition there that would make that difficult?

bethanyaconnor:
Also can you update the title of this PR? It seems like this is a bit more than turning off JS error logging now :)

apps/src/lib/util/StatsigReporter.js
hannahbergam: Can we pair on this? It seems that the current user redux has been updating Amplitude on init(), which doesn't seem to get called for each page load.

I'm also surprised that a page would be cached through a change in user log out/sign in. This would make me think that some users could access data from a previously logged in user's cache. Is that the case or am I misunderstanding?

hannahbergam:
> Also can you update the title of this PR? It seems like this is a bit more than turning off JS error logging now :)

Ah yes- good call! Done

apps/src/lib/util/StatsigReporter.js
bethanyaconnor: oh weird -- it's probably not great if it doesn't get called on each page load 😬 

On cached pages (and also I think on uncached pages as a result), we fetch the user data asyncronously, which then is fed into the redux store [here](https://github.com/code-dot-org/code-dot-org/blob/159aa5956c4a92afa0df420228453f60c7bb4109/apps/src/code-studio/header.js#L219). Most of our pages aren't cached, so in most cases, this solution is fine, but it's always possible we accidentally try to use this on one of those cached page and become confused.

apps/src/lib/util/StatsigReporter.js
```diff
@@ -26,13 +39,15 @@ class StatsigReporter {
...
       environment: {tier: getEnvironment()},
       localMode: this.local_mode,
+      disableErrorLogging: true,
     };
-    this.initialize(api_key, options);
+    this.initialize(api_key, user, options);
   }
 
-  async initialize(api_key, options) {
+  // Initialize with a null user object- current user redux will update on sign in
```
bethanyaconnor: what does this comment mean sorry?

Asking because I do like the idea of comment here that this will potentially update the statsig user :)

apps/src/templates/currentUserRedux.js
```diff
@@ -188,11 +188,7 @@ export default function currentUser(state = initialState, action) {
     );
     // Calling Statsig separately to emphasize different user integrations
     // and because dual reporting is aspirationally temporary (March 2024)
-    statsigReporter.setUserProperties(
-      id,
-      user_type,
-      experiments.getEnabledExperiments()
```
bethanyaconnor, line 194: not a big deal but curious why the enabled experiments was removed?

dashboard/app/views/layouts/application.html.haml
```diff
@@ -6,6 +6,7 @@
     = render partial: 'i18n/crowdin_in_context_tool'
     = render inline: File.read(Rails.root.join('..', 'shared', 'haml', 'onetrust_cookie_scripts.haml')), type: :haml, locals: {dashboard: true, domain: 'code.org'}
     - appname = Rails.env.production? ? t(:appname) : "#{t(:appname)} [#{Rails.env}]"
+    - managed_test_server = "#{CDO.running_web_application? && CDO.test_system?}"
```
bethanyaconnor, line 9: did you ever figure out why this was set wrong? (can totally be done in a separate PR if not)

dashboard/app/views/layouts/application.html.haml
```diff
@@ -29,7 +30,10 @@
     -# auth key needed for statsig analytics
     %script{data: {'statsig-api-client-key': CDO.safe_statsig_api_client_key}}
     -# for statsig to recognized managed test server
-    %script{data: {'managed-test-server': "#{CDO.running_web_application? && CDO.test_system?}"}}
+    %script{data: {'managed-test-server': managed_test_server}}
+    -# user id and type for analytics logging
```
bethanyaconnor: nit: add a comment that this may not work for cached pages

bethanyaconnor (approved):
Okay I'm convinced! Left a couple questions/nits but mostly for more comments. Thanks for sticking with this!

dashboard/app/views/layouts/application.html.haml
hannahbergam: I don't think I was injecting ruby in the correct way before so am abstracting it out into its own variable to match what else is happening in this file!

apps/src/templates/currentUserRedux.js
hannahbergam: To be honest my initial thought was that we wouldn't be using the experiments framework alongside statsig's AB testing but I realize I'm not clear yet on how we will approach that

apps/src/lib/util/StatsigReporter.js
hannahbergam: Ah I hadn't updated it yet- it's updated and accurate now!

dashboard/app/views/layouts/application.html.haml
hannahbergam: Added!

apps/src/templates/currentUserRedux.js
bethanyaconnor: That's valid! And we can always add them back

PR #57489, opened 2024-03-22
Fixes issues with malformed evidence strings.

wilkie:
Since evidence is its own column, we can get away with not having some of the information and falling back to what is there. If the code isn't there, we still annotate with the message for those lines with the observations column instead. Here, we only use the line numbers to mark the highlights. The evidence listing is just a bulleted list of sentences provided in the observations column.

We also allow for partial matches of code. The AI sometimes truncates the code down (e.g. 'if (something) { ... }') and in those cases we want to gracefully fall back to highlighting the parts we match. It's not ideal.

Here is it showing the effect of truncated code regions. It would have been something like: `Lines 188-230: The student has... blah. `function movePlayer() { ... }` `function playerAtEdge() { ... }` etc:

image

Testing exposed the bug where the `LearningGoals` component does not render the AI information on the first load... only when the goal is switched will it do that. This is because the function to load the evaluations is not called outside of that carousel change method. I have updated it such that the aiEvaluations is a memo that is always up-to-date. It made testing a bit easier as well. (Note, this isn't true for any of our rubrics since those all start with the program guide learning goal, which is not AI assessed... so we never caught this in practice)

Updated:
* Adds ability to parse the partial code snippets when the AI 'helpfully' truncates the code it gives as a reference.
* Adds ability to fall back to the observations text when the evidence column does not contain messages alongside code references.
* Fixes issue where the `LearningGoals` component does not render the AI assessment for the first learning goal if it happens to be AI enabled.

## Links

- jira ticket: [AITT-541](https://codedotorg.atlassian.net/browse/AITT-541)

## Testing story

New tests to capture the changes to the annotator and the `annotateLines` method in `LearningGoals.jsx`.

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/EditorAnnotator.js
+++ b/apps/src/EditorAnnotator.js
@@ -747,7 +747,7 @@ export default class EditorAnnotator {
    * @returns {bool} Returns true when the editor is line based.
    */
   static isLineBased() {
-    return EditorAnnotator.annotator().isLineBased();
+    return EditorAnnotator.annotator()?.isLineBased() || false;
   }
 
   /**
@@ -756,7 +756,7 @@ export default class EditorAnnotator {
    * It may or not be still line-based.
    */
   static hasBlocks() {
-    return EditorAnnotator.annotator().hasBlocks();
+    return EditorAnnotator.annotator()?.hasBlocks() || false;
   }
 
   /**
@@ -766,7 +766,7 @@ export default class EditorAnnotator {
    * @returns {bool} Returns true when the editor is block based.
    */
   static isBlockBased() {
-    return EditorAnnotator.annotator().isBlockBased();
+    return EditorAnnotator.annotator()?.isBlockBased() || false;
   }
 
   static anonymizeCode_(code) {
@@ -1095,7 +1095,7 @@ export default class EditorAnnotator {
     icon = null,
     tipStyle = {}
   ) {
-    EditorAnnotator.annotator().annotateLine(
+    EditorAnnotator.annotator()?.annotateLine(
       lineNumber,
       message,
       logLevel,
@@ -1109,7 +1109,7 @@ export default class EditorAnnotator {
    * Removes annotations of a particular type.
    */
   static clearAnnotations(logLevel = 'INFO') {
-    EditorAnnotator.annotator().clearAnnotations(logLevel);
+    EditorAnnotator.annotator()?.clearAnnotations(logLevel);
   }
 }
 
--- a/apps/src/templates/rubrics/AiAssessmentBox.jsx
+++ b/apps/src/templates/rubrics/AiAssessmentBox.jsx
@@ -73,26 +73,50 @@ export default function AiAssessmentBox({
             <StrongText>{i18n.aiAssessmentEvidence()}</StrongText>
           </BodyFourText>
           <ul>
-            {aiEvidence.map(
-              (info, i) =>
-                info &&
-                info.firstLine && (
-                  <li key={i}>
-                    {/* Lines [firstLine]-[lastLine]: [message] */}
-                    {info.firstLine === info.lastLine &&
-                      i18n.aiAssessmentEvidenceLine({
-                        lineNumber: info.firstLine,
-                        feedbackForLine: info.message,
-                      })}
-                    {info.firstLine !== info.lastLine &&
-                      i18n.aiAssessmentEvidenceLines({
-                        firstLineNumber: info.firstLine,
-                        lastLineNumber: info.lastLine,
-                        feedbackForLines: info.message,
-                      })}
-                  </li>
+            {
+              /* When the message is the same as the whole observations, this
+               * was evidence that did not have a dedicated message. In this
+               * case, this is where we fall back to just showing a list of
+               * observations instead of the evidence. This won't have line
+               * numbers and is certainly worse but better than nothing. */
+              (
+                aiEvidence
+                  .filter(info => info && info.observations === info.message)
+                  .map(info => info.observations)[0] || ''
+              )
+                .split('. ')
+                .map(
+                  (line, i) =>
+                    line.trim() && (
+                      <li key={i}>
+                        {/* Just the observation as a whole */}
+                        {line.endsWith('.') ? line : line + '.'}
+                      </li>
+                    )
                 )
-            )}
+            }
+            {aiEvidence
+              .filter(info => info && info.observations !== info.message)
+              .map(
+                (info, i) =>
+                  info &&
+                  info.firstLine && (
+                    <li key={i}>
+                      {/* Lines [firstLine]-[lastLine]: [message] */}
+                      {info.firstLine === info.lastLine &&
+                        i18n.aiAssessmentEvidenceLine({
+                          lineNumber: info.firstLine,
+                          feedbackForLine: info.message,
+                        })}
+                      {info.firstLine !== info.lastLine &&
+                        i18n.aiAssessmentEvidenceLines({
+                          firstLineNumber: info.firstLine,
+                          lastLineNumber: info.lastLine,
+                          feedbackForLines: info.message,
+                        })}
+                    </li>
+                  )
+              )}
           </ul>
         </div>
       )}
--- a/apps/src/templates/rubrics/LearningGoals.jsx
+++ b/apps/src/templates/rubrics/LearningGoals.jsx
@@ -1,4 +1,4 @@
-import React, {useEffect, useState, useRef} from 'react';
+import React, {useEffect, useState, useRef, useMemo} from 'react';
 import PropTypes from 'prop-types';
 import i18n from '@cdo/locale';
 import classnames from 'classnames';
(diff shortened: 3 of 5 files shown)


apps/src/EditorAnnotator.js
```diff
@@ -1109,7 +1109,7 @@ export default class EditorAnnotator {
    * Removes annotations of a particular type.
    */
   static clearAnnotations(logLevel = 'INFO') {
-    EditorAnnotator.annotator().clearAnnotations(logLevel);
+    EditorAnnotator.annotator()?.clearAnnotations(logLevel);
```
wilkie, line 1112: Adding these null checks to get around a console error we were seeing when the annotations race the loading of the editor.

apps/src/templates/rubrics/LearningGoals.jsx
```diff
@@ -130,14 +148,25 @@ export function annotateLines(evidence) {
         continue;
       }
 
-      // We have some kind of code reference
-      hasSnippet = true;
```
wilkie, line 135: We can now trust that when no code references are found, we still have something to annotate because it comes from the dedicated `evidence` column. Before, when it was just `observations`, we didn't want to just assume it was relevant.

apps/src/templates/rubrics/LearningGoals.jsx
```diff
@@ -211,7 +242,6 @@ export default function LearningGoals({
     INVALID_UNDERSTANDING
   );
   const [aiFeedback, setAiFeedback] = useState(-1);
-  const [aiEvidence, setAiEvidence] = useState([]);
```
wilkie, line 215: This isn't really state. It is something calculated from state. So it got converted to a memo.

tess323:
This is super cool - thank you for diving into this.

apps/src/templates/rubrics/LearningGoals.jsx
davidsbailey: I think I am missing something here. are you assuming that the evidence column will always contain line numbers? If not, where would you show this information without a line number to attach it to or a code snippet which you can locate?

apps/src/templates/rubrics/LearningGoals.jsx
```diff
@@ -130,14 +148,25 @@ export function annotateLines(evidence) {
...
       let position = EditorAnnotator.findCodeRegion(snippet, {
         stripComments: true,
       });
 
+      // If it could not find the region, try a less aggressive approach
+      // by looking for common patterns.
+      if (!(position.firstLine || position.lastLine)) {
+        // We often see the AI truncate code blocks via something like:
+        // 'if (something) { ... }' ... This will at least find the beginning of it.
+        const truncationRegex = /{\s*[.]+\s*}/;
```
davidsbailey, line 162: Finding just the beginning seems like a good heuristic to me for now. if we get a lot of false positives, we could try harder to include what comes after the `...` in the match.

apps/src/templates/rubrics/LearningGoals.jsx
```diff
@@ -426,6 +454,15 @@ export default function LearningGoals({
     );
   };
 
+  const aiEvidence = useMemo(() => {
+    // Annotate the lines based on the AI observation
+    clearAnnotations();
+    if (!!aiEvalInfo && aiEvalInfo.evidence) {
```
davidsbailey: totally optional to address in this PR, but js now has safe navigation:
```suggestion
    if (!!aiEvalInfo?.evidence) {
```

apps/test/unit/templates/rubrics/LearningGoalsTest.jsx
```diff
@@ -80,6 +90,35 @@ const studentLevelInfo = {
...
+    );
+  });
+  afterEach(() => {
+    annotatorStub.restore();
+    annotateLineStub.restore();
+    clearAnnotationsStub.restore();
+    highlightLineStub.restore();
+    clearHighlightedLinesStub.restore();
+  });
+
```
davidsbailey, line 124: nit: unless they are needed to keep other tests from breaking, I think these better belong within the `  describe('annotateLines', () => {` block currently starting on line 266

apps/src/templates/rubrics/LearningGoals.jsx
wilkie: Yep, tl;dr: I am assuming the evidence column always has line numbers. I think it always did. There's no good way of knowing how to reference things. If the evidence column is not parseable, then it will display the observations column as evidence instead. If the evidence column has line information, it will highlight those lines (perhaps in the wrong place) and annotate with the observations column. Nothing is at all ideal besides the intended behavior of having evidence with line numbers and an identifiable code snippet.

This logic that was removed here was checking whether or not it has a code snippet. Before, when it was parsing the observations column, it had to distinguish between the content added as evidence in the right parseable form and whatever other observations exist. It is no longer ambiguous. The dedicated column means we can just start assuming the content should be in the parseable form and ALL content we see is 'evidence' and use a fallback instead of ignoring it. No code snippet? OK, we still assume the content is evidence related to the line numbers... annotate it and move on.

apps/src/templates/rubrics/LearningGoals.jsx
wilkie: I used them elsewhere in the commit! But my brain isn't always operating at the latest version. I do prefer having them and will add them! thank you for noticing this.

apps/test/unit/templates/rubrics/LearningGoalsTest.jsx
wilkie: I DRY'd them a little by putting those stubs/restores in functions so they could be then used in before/after in the `annotateLines` and `clearAnnotations` tests in that section.

davidsbailey:
I'm following along... but please do re-request review when you're ready for another look!

apps/src/templates/rubrics/AiAssessmentBox.jsx
```diff
@@ -73,26 +73,50 @@ export default function AiAssessmentBox({
...
-                        lastLineNumber: info.lastLine,
-                        feedbackForLines: info.message,
-                      })}
-                  </li>
+            {
+              /* When the message is the same as the whole observations, this
+               * was evidence that did not have a dedicated message. In this
+               * case, this is where we fall back to just showing a list of
+               * observations instead of the evidence. This won't have line
+               * numbers and is certainly worse but better than nothing. */
```
davidsbailey, line 81: pardon my slowness but I'm having trouble reconciling this with:

> Yep, tl;dr: I am assuming the evidence column always has line numbers.

davidsbailey: per offline conversation, if there are no line numbers then we will not be annotating anything. however, we still show the evidence in the TA popup window, and that appears to be what this comment is about.

davidsbailey: screenshot of this from wilkie: 
image

apps/src/templates/rubrics/LearningGoals.jsx
```diff
@@ -91,13 +91,23 @@ export function clearAnnotations() {
  * In this case, we might find this code on lines 8 through 10. So, we would
  * annotate line 8 and highlight lines 8 through 10.
  *
+ * Another case requiring a fallback is when the evidence feedback does not
+ * contain a message. Something like: `Lines 5-6: `if (something) {...}`.
+ * In this case, we can highlight the line, but we want to annotate the line
+ * with the full observations. This is why we also pass those along.
```
davidsbailey, line 98: I think it's worth noting somewhere (here?) that, while observations are separated by learning goal, they are shared across annotations within a specific learning goal. so, if you have two annotations in the Multiple Sprites learning goal, and you are using the observations text, you might end up with something like:
* Line 1: `createSprite('foo')`, [observations text describing creating sprites foo and bar]
* Line 3: `createSprite('bar')`, [observations text describing creating sprites foo and bar]

don't get me wrong, this seems like a good stopgap for now and better than showing nothing... but i think it is helpful to document the reality here.

apps/src/templates/rubrics/LearningGoals.jsx
davidsbailey: ok, maybe there is a simpler way to represent this but here is my understanding of the different scenarios:

| # | has line number? | `hasSnippet` | has evidence text? | annotation text | anootated line number determined by |
|--------|--------|--------|--------|--------|--------|
| 1 | yes | yes | yes | evidence text | code snippet |
| 2 | yes | no | yes | evidence text | AI line number |
| 3 | yes | yes | no | observations text | code snippet |
| 4 | yes | no | no | observations text | AI line number |
| 5 | no | * | * | none | none |

where `hasSnippet` (your variable name) means whether a code snippet exists for this annotation.

please let me know if this sounds right and I will open a ticket to get some UI test coverage for the major scenarios.

davidsbailey (approved):

Nokondi (approved):
I have nothing to add that Dave hasn't already posted. Great work as always!

PR #57476, opened 2024-03-22
Freeze notification and auto-freeze when project featured in public gallery (parts 4,5)

fisher-alice:
UPDATE Mar 25, 2024: This PR is closed due to testing issues. I'm going to break apart this PR into two PRS (part 4 and 5) for easier debugging of tests.

This PR:
- adds notification when a user's project is frozen
- auto-freezes a featured project when it is featured ('active')
- auto-unfreezes a featured project when it is unfeatured ('archived')

This PR is composed of two prior approved PRs:
Part 4 description: https://github.com/code-dot-org/code-dot-org/pull/56932
Part 5 description: https://github.com/code-dot-org/code-dot-org/pull/57443

Part 4 was initially part of https://github.com/code-dot-org/code-dot-org/pull/56666 which was reverted due to a not well-understood testing bug on `test` [Slack thread](https://codedotorg.slack.com/archives/C0T0PNTM3/p1711040741216309). I created this PR with part 4 and new updates that are related to the freezing of active featured projects.

## PR Checklist:

- [ ] Tests provide adequate coverage
- [ ] Privacy and Security impacts have been assessed
- [ ] Code is well-commented
- [ ] New features are translatable or updates will not break translations
- [ ] Relevant documentation has been added or updated
- [ ] User impact is well-understood and desirable
- [ ] Pull Request is labeled appropriately
- [ ] Follow-up work items (including potential tech debt) are tracked and linked

PR #57495, opened 2024-03-23
Part 5: Auto-freeze/unfreeze when featuring/archiving featured projects

fisher-alice:
This PR includes updates to:
- auto-freeze a featured project when it is featured ('active')
- auto-unfreeze a featured project when it is unfeatured ('archived')
The PR description and testing story is available in a prior-approved PR: https://github.com/code-dot-org/code-dot-org/pull/57443 that was merged into https://github.com/code-dot-org/code-dot-org/pull/57476 - this PR was not merged to `staging` due to testing issues.

Thus, I am testing and planning to merge part 4 and 5 separately.

This PR is different from the original part 5 PR . 
- I created helped methods to freeze and unfreeze featured projects in `featured_projects_controller.rb` which an early return if a project with `project_id` is not found.
- I changed the param name for `/featured_projects/` to `channel_id` to distinguish from the derived `project_id`.

## Testing story
1. SETUP:
- I tested with two different accounts locally. First account is a student account and second account is a teacher account with project validator permission.

2. FEATURING A PROJECT (freeze)
- I created a project in the student account, then opened the project URL with the project validator account and bookmarked it as a featured project.
- Then as a project validator, I went to 'projects/featured' and featured the project.
- I confirmed that the student project was still listed under 'My Projects' in the student account, however, when I opened the project as a student, it was view-only.

3. UNFEATURING A PROJECT (unfreeze)
- I then went to 'projects/featured' admin page as a project validator and unfeatured the student project. 
- I confirmed logged into the student account that the project was no longer frozen.

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/dashboard/app/controllers/featured_projects_controller.rb
+++ b/dashboard/app/controllers/featured_projects_controller.rb
@@ -1,6 +1,9 @@
 class FeaturedProjectsController < ApplicationController
   authorize_resource
 
+  # Set a project as a featured project whose status is 'bookmarked'. It will be included
+  # in the list of featured projects at 'projects/featured', but will not be displayed in the
+  # public gallery until its status is 'active'.
   def bookmark
     _, project_id = storage_decrypt_channel_id(params[:project_id])
     return render_404 unless project_id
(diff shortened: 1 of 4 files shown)


dashboard/app/controllers/featured_projects_controller.rb
```diff
@@ -68,4 +76,24 @@ def buffer_abuse_score(score = -50)
       'rack.input' => StringIO.new
     )
   end
+
+  def freeze_featured_project(project_id)
+    project = Project.find_by(id: project_id)
+    return if project.nil?
```
bencodeorg, line 83: I think it's worth a comment here that this early return is here to avoid issues we have with creating projects in unit tests. A little torn on this approach because if we actually were unable to find the project, we'd return from the controller action a success status, without actually freezing the project.

But, if you tried the other path we considered (creating a project in a `setup_all` block, which we've observed working in other places) and that's not working, I get that you have much of a choice.

dashboard/app/controllers/featured_projects_controller.rb
```diff
@@ -68,4 +76,24 @@ def buffer_abuse_score(score = -50)
...
+    project_value["updatedAt"] = DateTime.now.to_s
+    project.update! value: project_value.to_json
+  end
+
+  def unfreeze_featured_project(project_id)
+    project = Project.find_by(id: project_id)
+    return if project.nil?
+    project_value = JSON.parse(project.value)
+    project_value["frozen"] = false
+    # Unhide in case this project was frozen manually.
```
bencodeorg: Also think this comment might be worth expanding on for future readers.

bencodeorg (approved):
🎉

dashboard/app/controllers/featured_projects_controller.rb
fisher-alice: Updated!

dashboard/app/controllers/featured_projects_controller.rb
fisher-alice: Thanks Ben! I added a comment here as you suggested.
I'll merge as-is with the early return since we'd like the ability to freeze/unfreeze with the already merged updated public gallery work. But I'm going to continue investigating tests in this space and I think I can do a quick follow-up here to address your concern. [Draft PR](https://github.com/code-dot-org/code-dot-org/pull/57537)

PR #57499, opened 2024-03-25
Update .button-wrapper row gap

kelbyhawn:
Updates the `row-gap` for the `.button-wrapper` class that decreases vertical space when buttons wrap. This will fix this everywhere `.button-wrapper` is used (which is a lot of places!).

**Jira ticket:** [ACQ-1646](https://codedotorg.atlassian.net/browse/ACQ-1646)

----

| Before | After |
| ----- | ----- |
| Before | After |

### DIFF
--- a/pegasus/sites.v3/code.org/public/css/design-system-pegasus.scss
+++ b/pegasus/sites.v3/code.org/public/css/design-system-pegasus.scss
@@ -217,7 +217,8 @@ hr {
 .button-wrapper {
   display: flex;
   flex-wrap: wrap;
-  gap: 1.5rem;
+  column-gap: 1.5rem;
+  row-gap: 0.5rem;
 
   &.centered {
     justify-content: center;


hannahbergam (commented):
This seems good to me generally. Do you know if we overwrite this anywhere (and could therefore revert to relying on this styling)? And do you have a good sense of what eyes diffs you're expecting?

kelbyhawn:
> This seems good to me generally. Do you know if we overwrite this anywhere (and could therefore revert to relying on this styling)? And do you have a good sense of what eyes diffs you're expecting?

The styles for this class are only used on updated pages that use the `design-system-pegasus.scss` stylesheet so this shouldn't affect anything outside of updated pages. And this will only change the UI on mobile. I don't think we have any eyes tests for mobile besides the hamburger menu, but let me know if I'm wrong!

hannahbergam:
> > This seems good to me generally. Do you know if we overwrite this anywhere (and could therefore revert to relying on this styling)? And do you have a good sense of what eyes diffs you're expecting?
> 
> The styles for this class are only used on updated pages that use the `design-system-pegasus.scss` stylesheet so this shouldn't affect anything outside of updated pages. And this will only change the UI on mobile. I don't think we have any eyes tests for mobile besides the hamburger menu, but let me know if I'm wrong!

Sounds good to me! And I do think we have *some eyes tests for mobile (like the tutorial landing pages? I don't see a @no_mobile tag there) but I agree we don't have many.

hannahbergam (approved):
LGTM

PR #57500, opened 2024-03-25
Remove Derniere Danse

bencodeorg:
Removes a song (Derniere Danse by Indila) from Dance Party that isn't properly licensed. Associated new song manifest (`songManifest2024_v2`) has been uploaded to S3.

## Testing story

- Tested manually in production via URL param that the song no longer appears in the song dropdown.
- Tested manually in development that changing the song manifest version in code produced the same result.
- I did not test that the right deprecation warning is shown if you have a project with this song.


jmkulwik (approved):
Thanks for the quick turnaround!

PR #57502, opened 2024-03-25
Fix external link icon for RTL languages

kelbyhawn:
Applies RTL language specific styles to the `.has-external-link` class so it the icon is mirrored and shows in the right order.

**Jira ticket:** [ACQ-1659](https://codedotorg.atlassian.net/browse/ACQ-1659)

----

## Buttons
| Before | After | LTR |
| ----- | ----- | ----- |
| Before | After | LTR |

## Text links
| Before | After | LTR |
| ----- | ----- | ----- |
| Text_Before | Text_After | Text_LTR |


TurnerRiley (approved):
Thank you for doing this, looks great!

PR #57503, opened 2024-03-25
skip dropdown test on mobile

mikeharv:
Since Friday, this test has been failing due to a stale element reference. It doesn't look like we've changed anything related to this feature recently. I've confirmed that dropdown in the test level works as expected on iPhone and iPad Safari. The proposal here is to skip this test on mobile for now to unblock the pipeline.

## PR Checklist:

- [ ] Code is well-commented


ebeastlake:
Linking an issue to capture reenabling so it's tracked: https://codedotorg.atlassian.net/browse/CT-424

snickell (approved):
We did some work to repro, but were not able to due to weird sauce connect behavior. At this point, I think we should disable on mobile (since there's a ticket for following up later).

PR #57504, opened 2024-03-25
[Progress] Improve teacher dashboard load time by returning only used fields

lfryemason:
The redux store only stores some of the things that are returned from the section.summarize method on the server. Therefore, we can easily remove all fields from the server response to the teacher dashboard page that do not go into the redux store.

The redux stored objects. Note: isAssigned is not from the server object.
image

The selected section is only used for a few select other fields, otherwise the section list with the above `concise_summarize` will be used. Those fields can be found in [this](https://github.com/code-dot-org/code-dot-org/blob/staging/apps/src/sites/studio/pages/teacher_dashboard/show.js#L38) file. Therefore, for the selected section, we are only going to return the necessary fields with `selected_section_summarize`.

This is part 1 of a series of PRs to improve load time by optimizing and removing unnecessary code from sections and section object for the `teacher_dashboard` page.

## Links
[TEACH-960](https://codedotorg.atlassian.net/browse/TEACH-960)

## PR Checklist:

- [ ] Code is well-commented


apps/src/templates/teacherDashboard/teacherSectionsRedux.js
```diff
@@ -1343,7 +1343,6 @@ export const sectionFromServerSection = serverSection => ({
   unitId: serverSection.unit_id,
   courseId: serverSection.course_id,
   hidden: serverSection.hidden,
-  isAssigned: serverSection.isAssigned,
```
lfryemason, line 1348: This is not set by the server response and is instead set by https://github.com/code-dot-org/code-dot-org/blob/staging/apps/src/templates/teacherDashboard/teacherSectionsRedux.js#L1470

etaderhold (approved):
A definite step in the right direction! Have you measured any before/after differences in local load time?

I will say I'm not super excited about introducing multiple permutations of `summarize` here. This seems like it might be a good application for GraphQL, where the client is expected to specify only the fields it wants and the server assembles them to order. Searching Slack archives it looks like @davidsbailey did some investigation around introducing GraphQL about a year ago. Perhaps time to revisit?

In the meantime if either of these new summarize methods are strict subsets of the main summarize method, perhaps it would be better to call them from `summarize` to reduce duplication both now and in the future if/when new fields need to be added to the summary.

PR #57507, opened 2024-03-25
Conditionally Disable MySQL Client Reconnect

sureshc:
Client Reconnect is disabled in MySQL 8 server. Conditionally disable it in our Ruby MySQL Dashboard and Pegasus clients. Using a DCDO setting allows us to run a canary test on one of the production web application servers. Default to *ENABLE* client reconnect because we have a handful of unit tests that are encountering issues when client reconnect is disabled that we haven't debugged yet. Those issues appear to be related to conditions that only arise during unit testing (for example, see #58628).

ProxySQL [does not appear to utilize the deprecated MySQL client reconnect feature](https://proxysql.com/Documentation/global-variables/mysql-monitor-variables/#mysql-monitor_ping_interval), but we are using a fork of an outdated version of ProxySQL so this documentation may not accurately reflect our usage.

We are not explicitly configuring any client reconnect settings if those exist in the version of ProxySQL we’re using:

https://github.com/code-dot-org/code-dot-org/blob/79967d8e8fa3e42cf4634e72df97e275a2625e3e/cookbooks/cdo-mysql/templates/default/proxysql.cnf.erb#L51-L84

## Testing story

`bundle exec rake adhoc:start RAILS_ENV=adhoc DATABASE=yes STACK_NAME=adhoc-disable-mysql-client-reconnect`

## PR Checklist:

- [ ] Code is well-commented


sureshc:
Re-opened post git-lfs migration in #57507

cat5inthecradle:
> we are using a fork of an outdated version of ProxySQL

Not anymore!

dashboard/config/database.yml
```diff
@@ -29,10 +31,10 @@ mysql_defaults: &mysql_defaults
   # `MYSQL_OPT_CONNECT_TIMEOUT`: The connect timeout in seconds.
   connect_timeout: 2
 
-  # `MYSQL_OPT_RECONNECT`: Enable or disable automatic reconnection to the server if the connection is found to have
-  # been lost. Reconnect is off by default; this option provides a way to set reconnection behavior explicitly.
-  # See https://dev.mysql.com/doc/refman/5.7/en/c-api-auto-reconnect.html
+<% if DCDO.get('enable_mysql_client_reconnect', true) %>
+  # `MYSQL_OPT_RECONNECT` is deprecated https://dev.mysql.com/doc/c-api/8.0/en/c-api-auto-reconnect.html
   reconnect: true
+<% end %>
```
cat5inthecradle, line 37: This only gets executed at service start, correct? So it should only apply to new EC2 instances launched after the DCDO change - unless any running servers happen to restart their dashboard/pegasus services, right?

cat5inthecradle (approved):

dashboard/config/database.yml
sureshc: Correct.

PR #57515, opened 2024-03-25
Temporarily Lock Teacher Accounts on Repeated Failed Authentication

Hamms:
Add functionality which will temporarily lock out teacher accounts after repeated failed authentication attempts. Specifically, we enable Devise's [built-in "Lockable" module](https://www.rubydoc.info/github/plataformatec/devise/Devise/Models/Lockable), with a few tweaks in place to scope it to apply only to teachers and to use serialized properties rather than columns for some new fields.

## User Experience Overview

When a user fails to authenticate 10 times in a row*, they will be warned on the penultimate attempt that their account is about to be locked:

image

If they fail again, they will see a warning explaining the situation:

image

Further attempts to log in at this point will result in the same warning, even if the correct authentication information is provided. To unlock the account, the teacher can either wait 30 minutes* - at which time their account will be automatically unlocked - or check their email, where they should find a link they can click on to immediately unlock their account:

image

If they click the link within the unlock interval, they will land on the sign-in page with a message informing them that their account was unlocked, at which point they can again attempt to authenticate:

image

With that, the expected user experience is complete!

A couple of unexpected-but-possible user experiences to also note:

If they click the link either after they've already clicked it or after the unlock interval has passed, they will see this message informing them that their account has already been unlocked:

image

The user can also access this form by manually navigating to `/users/unlock`, which is the URL in the email without the included parameter. If they do so, they will gain access to a form which they can use to re-send the unlock instructions email, but which also complains about not having the expected parameter:

image

We could invest effort into cleaning up this form to make it look nicer to users who do end up here, but I don't expect any users to actually end up here. I just wanted to call it out as a possibility, and something to consider for potential follow-up work; particularly if we hear that our users are having trouble receiving the unlock instructions email and would like the option to re-send it.

\* Both the maximum number of attempts and the unlock interval are configurable; 10 attempts and 30 minutes are the values I landed on after [discussing with @alivira](https://codedotorg.slack.com/archives/CFTFD6BPV/p1718656117610699). Happy to tweak them if folks have a strong preference, but I think what we have makes sense.

## Links

**Devise**

- https://www.rubydoc.info/github/plataformatec/devise/Devise/Models/Lockable
- https://github.com/heartcombo/devise/blob/v4.9.3/lib/devise/models/lockable.rb

**Previous Work**

- https://github.com/code-dot-org/code-dot-org/pull/59246
- https://github.com/code-dot-org/code-dot-org/pull/59668

## Testing story

Added some relevant unit tests. I explored also adding some UI tests, but found that the nature of the user interaction (checking email or waiting a period of time) made the process difficult, and I'm not sure how much additional value the UI tests would add given that we have unit test coverage.

Does anyone think some UI and possibly also Eyes tests would be useful here? Please let me know if so.

## PR Checklist:

- [ ] Code is well-commented


dashboard/config/initializers/devise.rb
```diff
@@ -199,24 +199,24 @@
...
+  config.unlock_strategy = :both
 
   # Number of authentication tries before locking an account if lock_strategy
   # is failed attempts.
-  # config.maximum_attempts = 20
+  config.maximum_attempts = 10
 
   # Time interval to unlock the account if :time is enabled as unlock_strategy.
-  # config.unlock_in = 1.hour
+  config.unlock_in = 1.hour
```
Hamms: Super open to feedback about what people think would be ideal values for both the maximum number of allowed attempts before the user gets locked out and the duration of time the account will remain locked before being automatically unlocked.

cat5inthecradle:
What happens if you trigger lockout and then immediately start the password reset process?

Hamms:
Testing in my local environment, it looks like the password reset process can be fully completed while the account is locked, and that doing so will also unlock the account.

See in particular [this line of code](https://github.com/heartcombo/devise/blob/1d6658097e364d45b5e059976f1e822eee7d67da/app/controllers/devise/passwords_controller.rb#L38) in Devise's passwords controller

cat5inthecradle:
> the password reset process can be fully completed while the account is locked, and that doing so will also unlock the account.

Awesome, that sounds like the desired behavior to me.

dashboard/lib/devise/models/custom_lockable.rb
```diff
@@ -0,0 +1,44 @@
+module Devise
+  module Models
+    # Extend Devise's built-in Lockable functionality for compatibility with
+    # our particular implementation needs. In particular, we scope it to apply
+    # only to teachers and add SerializedProperties support.
+    #
+    # See https://www.rubydoc.info/github/plataformatec/devise/Devise/Models/Lockable
+    module CustomLockable
```
carl-codeorg, line 8: Could we also override the `lock_access!` method to publish an event to Statsig? Product will likely want this metric available for dashboards.

Hamms: Darin and I were just talking about this!

We would indeed like to add more metrics to our general user authentication flow. We were thinking CloudWatch would be a good place, but are we already tracking metrics for some other related things in Statsig?

carl-codeorg: Yeah recently we (platform) have been publishing all of our new backend metrics to Statsig. Product has access to it and generally expects us to put all our metrics there.

It'd be nice to do something like use Cloudwatch as our generic store for logs and metrics, and then pipe them into systems like Statsig/Amplitude (or whatever the next flavor of the month is) using lambdas, but in the meantime it's probably safest to publish backend metrics directly to Statsig. Could also send to both Cloudwatch and Statsig, although that does double up on network calls.

cat5inthecradle: I suggest we log these low level system metrics to cloudwatch using our existing [Metrics buffer](https://github.com/code-dot-org/code-dot-org/blob/1fc15fc5d5ed7a0505d838ed694cbdd7e2d93815/lib/cdo/aws/metrics.rb) implementation.

If there is a particular product need to start tracking particular user login events, then we can add in the logging to that system (doubling up), but I would like to avoid storing low level metrics like this in a system that we may need to extract them from down the road.

We have existing CloudWatch metrics in this vein - [User SoftDelete](https://us-east-1.console.aws.amazon.com/cloudwatch/home?region=us-east-1#metricsV2?graph=~(metrics~(~(~'User~'SoftDelete~'Environment~'production~'UserType~'student)~(~'...~'teacher))~view~'timeSeries~stacked~false~region~'us-east-1~start~'-PT1008H~end~'P0D~period~86400~stat~'Sum)&query=~'*7bUser*2cEnvironment*2cUserType*7d). Given that, I would suggest putting these in the `User` namespace, and then metric names like `Locked`, `Unlocked`, `LogIn`, `Create`, etc. The `SoftDelete` metric includes dimensions `environment`, and `UserType`, which I think are appropriate here.

carl-codeorg: @cat5inthecradle that's a great example; didn't realize we had some of these metrics in place already in Cloudwatch.

At the same time, I don't see this as a low-level system metric. If we have a new way to potentially lock out users, I assume product will want to know about it. If we must keep this metric only in Cloudwatch, can we provide read access for platform PM? If we're looking to rework metrics over the next planning year, I suspect PM will need read access anyway, since Cloudwatch makes a ton of sense as the primary store for metrics and logs.

Hamms: I added some overrides to log to both CloudWatch and Statsig, at least for now; we can eventually remove one if we decide to standardize around the other. Please take another look!

dashboard/lib/devise/models/custom_lockable.rb
```diff
@@ -0,0 +1,85 @@
...
+      # in CloudWatch in order to preverse metrics for the long term. And as a
+      # newly-enabled feature, we want to also log them in Statsig to make the
+      # metric available on the product team's dashboards in the short term.
+
+      # @override https://github.com/heartcombo/devise/blob/v4.9.3/lib/devise/models/lockable.rb#L42-L50
+      def lock_access!
+        # CloudWatch
+        Cdo::Metrics.put(
+          'User', 'DeviseLockableAccessLocked', 1, {
+            Environment: CDO.rack_env
```
cat5inthecradle: Would it be useful to add UserType?

```suggestion
            Environment: CDO.rack_env,
            UserType: self.user_type
```

based on https://github.com/code-dot-org/code-dot-org/blob/f5a73818b52ae1d76a47fdbcf51150daeefec08e/dashboard/app/models/user.rb#L2592-L2597

cat5inthecradle (approved):

dashboard/lib/devise/models/custom_lockable.rb
Hamms: It's not particularly useful for this feature as it's currently implemented, since we explicitly scope it to apply only when `user_type == User::TYPE_TEACHER`. We could add it for either parallelism with our other metrics or future-proofing against possible refactoring; do you think that's worth doing?

cat5inthecradle: Yeah, those are the two reasons. I hypothesize that students would be more negatively impacted than teachers if we were to change password requirements in the future, so that's leading me to want to start collecting some baseline data now. I [see that](https://github.com/code-dot-org/code-dot-org/blob/28f9936d4275e07703709a37b4438ca0b7ccf076/dashboard/app/controllers/registrations_controller.rb#L165-L170) for "lti_user_created" we track `user_type` but for  "Sign Up Finished Backend" we're not, so there's not a consistent pattern yet.

Hamms: Added!

carl-codeorg (approved):
LGTM! Thanks for adding the metrics. Tested locally and the lockout seems to work for oauth and LTI logins, if those methods are also attached to the user.

PR #57517, opened 2024-03-25
P20-793: Add LMS Feedback banner

artem-vavilov:
## Summary
The banner will be displayed to LMS-integrated teachers 2 days after the account is created.
Once the banner is closed, it will not be shown to the teacher again.

## Screenshots
| Unanswered  | Answered |
| ------------- | -----------|
| before answering | after answering |
| unasnwered | answered |

## Overview
lti-feedback-banner

## Animations
animated-lti-feedback-banner

## Links
- JIRA ticket: [P20-793](https://codedotorg.atlassian.net/browse/P20-793)
- https://github.com/code-dot-org/code-dot-org/pull/57538


dashboard/db/migrate/20240321184256_create_lti_feedbacks.rb
```diff
@@ -0,0 +1,13 @@
+class CreateLtiFeedbacks < ActiveRecord::Migration[6.1]
```
artem-vavilov: The migration will be moved to a separate pull request after review

apps/src/templates/currentUserRedux.js
```diff
@@ -80,13 +81,16 @@ export const setProgressTableV2ClosedBeta = progressTableV2ClosedBeta => ({
...
+  uuid: null,
   userName: null,
   userType: 'unknown',
   userRoleInCourse: CourseRoles.Unknown,
   signInState: SignInState.Unknown,
   hasSeenStandardsReportInfo: false,
   isBackgroundMusicMuted: false,
   isSortedByFamilyName: false,
+  isLti: undefined,
+  isTeacher: undefined,
```
carl-codeorg, line 93: If we already have `userType`, do we also need isTeacher?

apps/src/redux/lti/ltiFeedbackReducer.ts
```diff
@@ -0,0 +1,113 @@
...
+    case LTI_FEEDBACK_CREATE_SUCCESS:
+      return {...state, isLoading: false, ltiFeedback: action.ltiFeedback};
+    case LTI_FEEDBACK_CREATE_FAILURE:
+      return {...state, isLoading: false, error: action.error};
+    default:
+      return state;
+  }
+}
+
+export const fetchLtiFeedback = (dispatch: Dispatch<LtiFeedbackAction>) => {
```
carl-codeorg, line 56: Is it possible to use an async function and try/catch with `await` instead of chaining `.then`s?

apps/src/redux/lti/ltiFeedbackReducer.ts
```diff
@@ -0,0 +1,113 @@
...
+    })
+    .catch(error => {
+      dispatch({
+        type: LTI_FEEDBACK_FETCH_FAILURE,
+        error: error instanceof Error ? error.message : 'An error occurred',
+      });
+    });
+};
+
+export const createLtiFeedback = async (
```
carl-codeorg, line 83: Same as above; if we are in an async function why not use `await`? Is it just React convention?

apps/src/redux/lti/ltiFeedbackReducer.ts
artem-vavilov: It's a reducer that returns data via a dispatcher and not directly via Promise:
https://github.com/code-dot-org/code-dot-org/blob/f9fb91cc6df63eb7d0c780d96b1555f389214722/apps/src/lib/ui/lti/feedback/LtiFeedbackBanner.tsx#L37-L42

apps/src/redux/lti/ltiFeedbackReducer.ts
artem-vavilov: We need async here only because of this function:
 https://github.com/code-dot-org/code-dot-org/blob/f9fb91cc6df63eb7d0c780d96b1555f389214722/apps/src/redux/lti/ltiFeedbackReducer.ts#L93

apps/src/templates/currentUserRedux.js
artem-vavilov: Since we use this frequently, I thought it would be helpful to have it in a separate property for reusability.
But I can use `userType` instead as we usually do.

apps/src/lib/ui/lti/feedback/LtiFeedbackBanner.tsx
```diff
@@ -0,0 +1,171 @@
...
+  fetchLtiFeedback,
+  createLtiFeedback,
+} from '@cdo/apps/redux/lti/ltiFeedbackReducer';
+import i18n from '@cdo/locale';
+
+import './LtiFeedbackBanner.scss';
+
+// LtiFeedbackBanner statuses
+const UNAVAILABLE = 'unavailable';
+const AVAILABLE = 'available';
```
daynew: Leave a short comment explaining what "unavailable" or "available" mean for this banner.

...after reading through the code a bit more is this their meaning?
* unavailable - The user does not qualify to see the banner.
* available - The user has not answered the feedback banner yet.

If I'm right about "available", maybe it could be "unanswered" because it is more closely related to the "answered" status?

apps/src/lib/ui/lti/feedback/LtiFeedbackBanner.tsx
```diff
@@ -0,0 +1,171 @@
...
+   */
+  const [status, setStatus] = useState<string>(() => {
+    if (!currentUser.isLti || !currentUser.isTeacher) return UNAVAILABLE;
+
+    let status = tryGetLocalStorage(key, '');
+    if (status === UNAVAILABLE) status = '';
+
+    !status && fetchLtiFeedback(ltiFeedbackAction);
+
+    return status;
```
daynew: Why is UNAVAILABLE sometimes `''` and sometimes `'unavailable'`?

There are 4 constants for the "status" defined at the top of the file, but this function defines a 5th value `''` which could be returned. There should be documentation for this 5th value.

apps/src/lib/ui/lti/feedback/LtiFeedbackBanner.tsx
```diff
@@ -0,0 +1,171 @@
...
+    } else if (ltiFeedback) {
+      setStatus(ANSWERED);
+    }
+  }, [ltiFeedback]);
+
+  /**
+   * Effect for handling errors.
+   */
+  useEffect(() => {
+    error && setStatus('');
```
daynew: Another instance of `''` being used. Maybe this should be a constant as well?

carl-codeorg (commented):
This LGTM. Honestly the banner is prettier than the rest of the teacher dashboard! Probably want to get signoff from @daynew or @stephenliang who have more React experience though.

apps/src/lib/ui/lti/feedback/LtiFeedbackBanner.tsx
```diff
@@ -0,0 +1,171 @@
...
+                    aria-hidden="true"
+                  />
+                </button>
+              </span>
+            </span>
+          ) : (
+            <span>
+              <span id="lti-feedback-banner-title" aria-hidden="true">
+                {i18n.lti_feedbackBanner_shareMore_text()}
+              </span>
```
daynew, line 158: Why is this text being hidden from A11y tools (`aria-hidden="true"`)?

apps/src/templates/currentUserRedux.js
carl-codeorg: Not a huge deal, my 2 cents is to leverage what's already there rather than introducing another prop though.

apps/src/lib/ui/lti/feedback/LtiFeedbackBanner.tsx
```diff
@@ -0,0 +1,171 @@
...
+          title=""
+          aria-hidden="true"
+        />
+
+        <Fade in={!isLoading}>
+          {status === AVAILABLE ? (
+            <span>
+              <span id="lti-feedback-banner-title" aria-hidden="true">
+                {i18n.lti_feedbackBanner_question()}
+              </span>
```
daynew, line 124: Why is this text being hidden from A11y tools (`aria-hidden="true"`)?

apps/src/lib/ui/lti/feedback/LtiFeedbackBanner.tsx
```diff
@@ -0,0 +1,171 @@
...
+ * LtiFeedbackBanner component
+ * This component is responsible for displaying the LMS Feedback banner for LTI teachers.
+ * The banner asks the teacher for feedback and provides options to respond.
+ * The banner's visibility and state are managed through local storage and state variables.
+ *
+ * @component
+ */
+const LtiFeedbackBanner: React.FC = () => {
+  const currentUser = getStore().getState().currentUser;
+  const key = `lti-fb-${currentUser.uuid}`;
```
daynew, line 37: Why the uuid? So we aren't exposing the user.id in the local storage key?

daynew (approved):
Look good to me.  Just a few minor comments and requests for clarification.

apps/src/lib/ui/lti/feedback/LtiFeedbackBanner.tsx
artem-vavilov: Yes, it makes much more sense. Thank you, @daynew!

apps/src/lib/ui/lti/feedback/LtiFeedbackBanner.tsx
artem-vavilov: `''` means no status yet, so we need to trigger `fetchLtiFeedback(ltiFeedbackAction)` to update the status based on the `ltiFeedback` presence:
https://github.com/code-dot-org/code-dot-org/blob/f9fb91cc6df63eb7d0c780d96b1555f389214722/apps/src/lib/ui/lti/feedback/LtiFeedbackBanner.tsx#L69-L75

and `if (status === UNAVAILABLE) status = '';` simply clear the `UNAVAILABLE` status if it's no longer unavailable

apps/src/lib/ui/lti/feedback/LtiFeedbackBanner.tsx
artem-vavilov: I believe the `userId` is already exposed; it's just because an integer value much easier to abuse.

apps/src/lib/ui/lti/feedback/LtiFeedbackBanner.tsx
artem-vavilov: It's because we already use it as the alert title:

https://github.com/code-dot-org/code-dot-org/blob/f9fb91cc6df63eb7d0c780d96b1555f389214722/apps/src/lib/ui/lti/feedback/LtiFeedbackBanner.tsx#L101

unasnwered

apps/src/lib/ui/lti/feedback/LtiFeedbackBanner.tsx
artem-vavilov: It's because we already use it as the alert title:

https://github.com/code-dot-org/code-dot-org/blob/f9fb91cc6df63eb7d0c780d96b1555f389214722/apps/src/lib/ui/lti/feedback/LtiFeedbackBanner.tsx#L101

answered

dashboard/db/migrate/20240321184256_create_lti_feedbacks.rb
artem-vavilov: https://github.com/code-dot-org/code-dot-org/pull/57538

carl-codeorg (approved):
Looking great with the animations. :shipit: as soon as the migration goes through.

PR #57524, opened 2024-03-25
Gen AI lab: save student customizations to S3, output changes to chat window

bencodeorg:
Two major changes:
- stores (and loads) student model customizations from S3. If no student-specific customizations exist, we use levelbuilder customizations.
- on save, compares to the most recently saved version, and outputs the changed fields to the chat window.

**Example message logged on save (UI obviously needs work :))**

image

## Links

- Log to chat window: https://codedotorg.atlassian.net/browse/LABS-517
- Project saving: https://codedotorg.atlassian.net/browse/LABS-537

## Testing story

When testing manually, I was able to:
- load initial model configuration from levelbuilder.
- update that configuration, and hit the "Update"button. I see writes to S3 in the network tab.
- see a message written to the chat window with the appropriate fields marked as having been updated.
- reload the page, and see the newly saved model configuration loaded correctly.

## Follow-up work

- Styling improvements on messages sent to chat window.


apps/src/aichat/views/ModelCustomizationWorkspace.tsx
```diff
@@ -29,14 +29,13 @@ const ModelCustomizationWorkspace: React.FunctionComponent = () => {
...
             showPromptCustomization && {
-              title: 'Prompt',
+              title: 'Setup',
               content: <PromptCustomization />,
             },
             isVisible(retrievalContexts.visibility) && {
               title: 'Retrieval',
               content: <RetrievalCustomization />,
             },
-            {title: 'Fine Tuning', content: 'fine tuning content TBD'},
```
bencodeorg, line 39: Fine tuning tab has been removed in latest mocks, and "Prompt" is now "Setup"

bencodeorg:
One decision point here was whether to store the most recently saved model customizations (ie, "sources") in the AI Chat redux store, or to access them from the existing project manager. The project manager keeps track of the most recently saved sources, and should be identical to what is being stored in `previouslySavedAiCustomizations` in this PR.

I ended up deciding to do the former (use the AI Chat redux store) because it felt clearer than comparing the current AI customizations (typed as AiCustomizations) with the most recently saved sources (just typed as a string, and in a different redux store). Open to other opinions though.

apps/src/aichat/redux/aichatRedux.ts
```diff
@@ -15,7 +16,58 @@ import {
...
 } from '../types';
 
+const haveDifferentValuesStringArray = (
+  value1: string[],
+  value2: string[]
+): boolean =>
+  value1.length !== value2.length ||
+  value1.some((item, index) => item !== value2[index]);
+
+const haveDifferentValues = (
```
sanchitmalhotra126, line 21: One potential simplification - can we just `JSON.stringify` and compare for obj/array values?

apps/src/aichat/views/AichatView.tsx
```diff
@@ -23,23 +26,36 @@ const AichatView: React.FunctionComponent = () => {
...
+      temperature: levelAiCustomizationsWithVisibility.temperature.value,
+      systemPrompt: levelAiCustomizationsWithVisibility.systemPrompt.value,
+      retrievalContexts:
+        levelAiCustomizationsWithVisibility.retrievalContexts.value,
+      modelCardInfo: levelAiCustomizationsWithVisibility.modelCardInfo.value,
     };
 
-    dispatch(setAiCustomizations(aiCustomizations));
-  }, [dispatch, initialAiCustomizations]);
+    const reconciledAiCustomizations = Object.keys(studentAiCustomizations)
```
sanchitmalhotra126: I think we'll need to reconcile the values here case by case, so it might be good to shift this logic to the reducer. Specifically - if the property's visibility is "readonly" or "hidden", we should use the level's customization (i.e. whatever was set by the levelbuilder) over the student's. Otherwise, if the visibility is "editable", we should use the student's customization value if present, or fall back to the level's customization value (or to the DEFAULT/EMPTY setting if neither were provided).

My thought was to just have a single redux action like `dispatch(setStartingAiCustomizations({levelAiCustomizations, studentAiCustomizations})` (can definitely have a better name 🙂) and then do all the reconciliation logic in the reducer function?

apps/src/aichat/views/modelCustomization/PublishNotes.tsx
```diff
@@ -23,6 +26,8 @@ const PublishNotes: React.FunctionComponent = () => {
     state => state.aichat.currentAiCustomizations
   );
 
+  const onUpdate = () => dispatch(updateAiCustomization());
```
sanchitmalhotra126: nit: can we wrap these with `useCallback`?

apps/src/aichat/views/modelCustomization/PublishNotes.tsx
```diff
@@ -52,7 +57,7 @@ const PublishNotes: React.FunctionComponent = () => {
         })}
       </div>
       <div className={styles.footerButtonContainer}>
-        <button type="button" disabled={isDisabled(visibility)}>
+        <button type="button" disabled={false} onClick={onUpdate}>
```
sanchitmalhotra126: Was the `disabled` change intentional?

sanchitmalhotra126:
> One decision point here was whether to store the most recently saved model customizations (ie, "sources") in the AI Chat redux store, or to access them from the existing project manager. The project manager keeps track of the most recently saved sources, and should be identical to what is being stored in `previouslySavedAiCustomizations` in this PR.
> 
> I ended up deciding to do the former (use the AI Chat redux store) because it felt clearer than comparing the current AI customizations (typed as AiCustomizations) with the most recently saved sources (just typed as a string, and in a different redux store). Open to other opinions though.

This makes sense to me! I also prefer to keep this within the AI Chat layer in case we want to tweak the data format at all

apps/src/aichat/views/modelCustomization/PublishNotes.tsx
bencodeorg: Nope, thanks!

apps/src/aichat/views/modelCustomization/PublishNotes.tsx
bencodeorg: Yep, done!

apps/src/aichat/views/AichatView.tsx
bencodeorg: hmm ok that makes sense. Is there an actual case where the levelbuilder and student customizations could diverge if the field is hidden or readonly, or just feels like a better practice? I thought the student's customizations would always just reflect the levelbuilder values, since they cannot change them.

apps/src/aichat/redux/aichatRedux.ts
bencodeorg: Oh good thought, I'll do that!

apps/src/aichat/views/AichatView.tsx
sanchitmalhotra126: Yeah, one specific case would be if a readonly/hidden value is changed after the student runs the project for the first time. The student would have the old value saved to their main.json but we'd want to use the updated levelbuilder value

bencodeorg: Yeah that makes sense -- I think my question is whether that a thing that actually happens (ie, levelbuilders updating level config after a level is in use)? It feels potentially more confusing to me that a student might have written some customization on a level, then it partially (? if we're doing this per field) is overwritten by a levelbuilder?

sanchitmalhotra126: Yeah, I do think while it's uncommon it could definitely happen as things change in the pilot (ex. we learn that a certain system prompt isn't working well or providing bad responses). But to clarify, this would only apply to readonly/hidden fields, so the student would never be able to customize those fields in the first place. For editable fields, we'll always go with the latest student-edited value, even if the levelbuilder value has been updated. Here's a thread with Dan & Sam discussing this: https://codedotorg.slack.com/archives/C06FELVTV0Q/p1710190631660709

bencodeorg: Super helpful context, thank you!

apps/src/aichat/redux/aichatRedux.ts
```diff
@@ -129,8 +204,48 @@ const aichatSlice = createSlice({
...
+      for (const customizationUntyped in reconciledAiCustomizations) {
+        const customization = customizationUntyped as keyof AiCustomizations;
+
+        if (
+          levelAiCustomizationsWithVisibility[customization].visibility ===
+            Visibility.EDITABLE &&
+          studentAiCustomizations[customization]
+        ) {
+          reconciledAiCustomizations = {
+            ...reconciledAiCustomizations,
```
bencodeorg, line 238: This is me battling Typescript again:

https://github.com/code-dot-org/code-dot-org/pull/57318#discussion_r1532527865

apps/src/aichat/redux/aichatRedux.ts
```diff
@@ -129,8 +204,48 @@ const aichatSlice = createSlice({
...
+        botName: levelAiCustomizationsWithVisibility.botName.value,
+        temperature: levelAiCustomizationsWithVisibility.temperature.value,
+        systemPrompt: levelAiCustomizationsWithVisibility.systemPrompt.value,
+        retrievalContexts:
+          levelAiCustomizationsWithVisibility.retrievalContexts.value,
+        modelCardInfo: levelAiCustomizationsWithVisibility.modelCardInfo.value,
+      };
+
+      for (const customizationUntyped in reconciledAiCustomizations) {
+        const customization = customizationUntyped as keyof AiCustomizations;
```
bencodeorg, line 230: This is also me battling Typescript :) There's gotta be a better way to do this??

apps/src/aichat/views/tabs/tabs.module.scss
```diff
@@ -8,6 +8,7 @@
   padding-left: 5px;
   overflow-x: auto;
   overflow-y: hidden;
+  min-height: 30px;
```
bencodeorg, line 11: This semi-resolves (at least improves, there's still a small visible change in the container height when making the viewport smaller, but at least tabs don't disappear) an issue @fisher-alice noticed with tabs disappearing on smaller screens

apps/src/aichat/views/AichatView.tsx
bencodeorg: Updated, please take another look when you have a chance!

apps/src/aichat/redux/aichatRedux.ts
```diff
@@ -39,6 +78,36 @@ const initialState: AichatState = {
 
 // THUNKS
 
+export const updateAiCustomization = createAsyncThunk(
```
sanchitmalhotra126, line 84: nit: mind adding a comment explaining what this is doing?

apps/src/aichat/redux/aichatRedux.ts
sanchitmalhotra126: Yeah unfortunately I don't know if there is (at least not that I know of) :/ I mean I think you could do something like

```
for (const customization of Object.keys(reconciledAiCustomizations) as (keyof AiCustomizations)[]) {
  // customization is keyof AiCustomization
  ... 
}
```
But don't know if that's much better than just casting it here

sanchitmalhotra126 (approved):
LGTM! 

Again we can probably revisit the types to make them easier to work with if necessary, but looks good!

PR #57528, opened 2024-03-26
Simplify SETUP.md for MacOS

snickell:
Its probably easiest to read this PR with the markdown rendered, possibly as side-by-side before/after blocks: 
[Before](https://github.com/code-dot-org/code-dot-org/blob/1c8330e8668ba127aefb22368c9454941fa408cd/SETUP.md#macos) | [**After**](https://github.com/code-dot-org/code-dot-org/blob/seth/no-rosetta/SETUP.md#macos)

### Changes:
1. Don't suggest using Rosetta for Apple Silicon (not needed anymore, much faster)
2. Shorter + less complex + visually clearer = easier to follow
3. Use multi-line code blocks for macOS setup commands, as these are easier to single-click copy-paste, and perhaps less visually overwhelming:
image
4. Suggest the repo be checked out BEFORE doing platform-specific setup. Why? While this complicates `git lfs` slightly (requires an explicit `git lfs pull`), it means the instructions are actually valid / they work... since `nvm install` and `rbenv install` don't actually work until the repo is checked out (as they use .nvmrc and .ruby-version).
5. `brew install` packages in one big block at the start (just like we do for apt-get install on ubuntu), rather than having to tediously paste in 20 brew install commands one by one as you come to each.
6. Remove now unnecessary build hacks from "Apple Silicon (M1) or Intel Mac bundle install steps" section from the end.
7. Move `bundle exec rake install` tips (equally relevant to all OSes) out of the linux-only section
8. Install `brew pdftk-java` (which automatically pulls in a cpu-correct jdk) rather than complex instructions to install intel-only jdk8
9. More consistently use the very nice `<summary></summary>` formatting for troubleshooting tips

### Testing:

- [x] Instructions work on a fresh user account on an apple silicon macbook running macOS 14
~- [ ] Instructions work on a fresh user account on an x86 macbook (preferably running macOS 13)~ https://github.com/code-dot-org/code-dot-org/pull/58321 removed x86 as supported
- [x] pdftk can still generated curriculum pdfs (how to test this?)


SETUP.md
```diff
@@ -3,45 +3,36 @@
...
 
      ```sh
      ruby --version  # --> ruby 3.0.5
      node --version  # --> v18.16.0
      ```
 
-1. Clone the repo, which also may take a while.
-    - Note you should have `git lfs --version` >= 3.0 installed prior to cloning, see OS-specific install steps referenced above.
-    - The simplest option is to clone via SSH with: `git clone git@github.com:code-dot-org/code-dot-org.git`
-    - The fastest option is to clone via HTTP with: `git clone https://github.com/code-dot-org/code-dot-org.git`. Although faster than SSH, this option requires you to reauthenticate every time you want to update. You will therefore probably want to switch to SSH after the initial clone with `git remote set-url origin git@github.com:code-dot-org/code-dot-org.git`
```
snickell, line 28: We now suggest cloning the repo before doing os-specific steps. While this requires adding a `git lfs pull` below, it means that `nvm install` and `rbenv install` actually work in the OS specific sections. i.e.: this ordering is actually correct!

SETUP.md
```diff
@@ -3,45 +3,36 @@
...
+1. `git lfs pull`
 
-1. `cd code-dot-org`
-
-1. `gem install bundler -v 2.3.22`
-
-1. `rbenv rehash`
+1. `gem install bundler -v 2.3.22 && rbenv rehash`
```
snickell, line 31: Collapse gem install bundler and rbenv rehash together: they're really one command, as the `rbenv rehash` is there to make the new bundler command "take".

SETUP.md
```diff
@@ -3,45 +3,36 @@
...
 
 1. `bundle install`
     - This step often fails to due environment-specific issues. Look in the [Bundle Install Tips](#bundle-install-tips) section below for steps to resolve many common issues.
 
 1. `bundle exec rake install:hooks`
-    <details>
-      <summary>Troubleshoot: `rake aborted! Gem::LoadError: You have already activated...` </summary>
-
-      - If you have issue `"rake aborted! Gem::LoadError: You have already activated rake 12.3.0, but your Gemfile requires rake 11.3.0."`, make sure you add `bundle exec` in front of the `rake install:hooks` command
-    </details>
```
snickell, line 44: The `bundle exec` is now explicitly present in the command (in an older version SETUPers were left to add the `bundle exec`, so this troubleshooting tip was relevant if you forgot)

SETUP.md
```diff
@@ -66,24 +57,30 @@ You can do Code.org development using OSX, Ubuntu, or Windows (running Ubuntu in
...
+    <details>
+        <summary>This will take 30 minutes, or more</summary>
+        The most expensive are the "seeding" tasks, where your local DB is populated from data in the repository. Some of the seeding rake tasks can take several minutes. The longest one, `seed:scripts`, can take > 10 minutes, but it should at least print out progress as it goes.
+    </details>
+    <details>
+        <summary>If `bundle exec rake install` is interrupted before finishing...</summary>
+        If, for any reason, you are forced to interrupt the `bundle exec rake install` command before it completes,
+        cd into dashboard and run `bundle exec rake db:drop` before trying `bundle exec rake install` again.
+        `bundle exec rake install` must always be called from the local project's root directory, or it won't work.
+    </details>
```
snickell, line 70: This section comes from the linux-specific instructions. Its a good tip, but its not just linux-specific, so here we make it available to all platforms.

SETUP.md
```diff
@@ -66,24 +57,30 @@ You can do Code.org development using OSX, Ubuntu, or Windows (running Ubuntu in
...
 
 1. Run the website `bin/dashboard-server`
 
-1. Visit <http://localhost-studio.code.org:3000/> to verify it is running.
+1. **Open <http://localhost-studio.code.org:3000/>** to verify its running.
 
-1. Install necessary plugins described in the [Editor configuration](#editor-configuration) section below.
-
-After setup, read about our [code styleguide](./STYLEGUIDE.md), our [test suites](./TESTING.md), or find more docs on [the wiki](https://github.com/code-dot-org/code-dot-org/wiki/For-Developers).
+After setup, [configure your editor](#editor-configuration), read about our [code styleguide](./STYLEGUIDE.md), our [test suites](./TESTING.md), or find more docs on [the wiki](https://github.com/code-dot-org/code-dot-org/wiki/For-Developers).
```
snickell, line 84: I think its kind of ugly to have a not-very-clear "setup your editor" section included as a step: this step is unlike the other explicit commands, and it detracts from the "SETUP.md got dashboard running!" nice ending to the steps. I think its more of a follow-on, and its not really required anyway.

SETUP.md
```diff
@@ -104,102 +101,86 @@ External contributors can supply alternate placeholder values for secrets normal
 
 ### macOS
 
-These steps are for Apple devices running **macOS Monterey and Ventura**, including those running on [Apple Silicon (M1|M2)](https://en.wikipedia.org/wiki/Apple_silicon#M_series). 
+Setup steps for Apple devices running **macOS Ventura and Sonoma**, including those running on [Apple Silicon](https://en.wikipedia.org/wiki/Apple_silicon#M_series) and Intel x86:
```
snickell: Include last two current macOS versions, explicitly mention that these instructions should (still) work on an x86 mac.

SETUP.md
```diff
@@ -258,11 +239,7 @@ Note: Virtual Machine Users should check the [Alternative note](#alternative-use
         1. `export CHROME_BIN=$(which chromium-browser)`
 1. Finally, configure your mysql to allow for a proper installation. You may run into errors if you did not leave mysql passwords blank
     1. `echo "ALTER USER 'root'@'localhost' IDENTIFIED WITH mysql_native_password BY '';" | sudo mysql`
-1. **IMPORTANT:** Read the following notes, then go back up to the [overview](#overview) and run the commands there.
-    1. If, for any reason, you are forced to interrupt the `bundle exec rake install` command before it completes,
-       cd into dashboard and run `bundle exec rake db:drop` before trying `bundle exec rake install` again
-    1. `bundle exec rake install` must always be called from the local project's root directory, or it won't work.
-    1. Finally, don't worry if your versions don't match the versions in the overview if you're following this method; the installation should still work properly regardless
+1. Return to the [overview](#overview) and continue setup.
```
snickell, line 247: This note was not linux specific, but its a great tip: its been moved to the shared setup instructions as a troubleshooting tip for `bundle exec rake install`

SETUP.md
```diff
@@ -414,27 +391,7 @@ Wondering where to start?  See our [contribution guidelines](CONTRIBUTING.md) fo
...
-```
-
-After `bundle install` completes successfully, before attempting the `bundle exec rake ...` commands, execute the following command to avoid errors with **ffi**
-
-```sh
-bundle update ffi
-```
-
-
-### ImageMagick with Pango
```
snickell, line 440: I started fresh, cleared my ~/.rbenv, and I'm pretty sure this is no longer necessary. Brew was messing up the openssl include and library path before (couple years past), but its straightened out now. Explicitly updating ffi is no longer required, our Gemfile.lock version is plenty fresh.

SETUP.md
```diff
@@ -414,27 +391,7 @@ Wondering where to start?  See our [contribution guidelines](CONTRIBUTING.md) fo
...
+#### ImageMagick with Pango
```
snickell, line 399: Use the same `####` column indent as other sections, so they don't become subsections of `imagemagick with pango` by accident

SETUP.md
```diff
@@ -104,102 +101,86 @@ External contributors can supply alternate placeholder values for secrets normal
 
 ### macOS
```
snickell: ### macOS section

For the macOS specific section, the changes are extensive, and I think its much easier to look at the markdown rendered versions before and after side by side, rather than trying to stare through the diff noise:
[Before](https://github.com/code-dot-org/code-dot-org/blob/1c8330e8668ba127aefb22368c9454941fa408cd/SETUP.md#macos) | [**After**](https://github.com/code-dot-org/code-dot-org/blob/seth/no-rosetta/SETUP.md#macos)

SETUP.md
cat5inthecradle: should we list `git lfs` in the dependency check instructions above?

snickell: Added!

cat5inthecradle (approved):

PR #57534, opened 2024-03-26
Music: one pack

breville:
Music Lab will now allow a workspace to use one restricted sound pack from the current library, along with all of the unrestricted packs.  When in a standalone project level, unless the level already specifies a pack, the user will see a modal dialog UI which allows them to select the pack.  The only way to change pack is to "start over", after which the modal will be shown again.  A non-standalone project level may specify a pack, but the dialog will not be available.

Because a music library can have multiple restricted packs, it's possible to switch restricted pack without a page reload.  For the same reason, it's possible for a progression to use a different restricted sound pack for each level in a lesson.

The modal UI allows the user to preview the sound for a pack, which plays the pack's first sound marked with `"type": "preview"`, if one exists.

The sound selection UI will show the restricted sound pack first, before the others.  It will not show any sounds marked as previews.  

The default `play sound` block value now picks the first non-preview sound from the first available pack in the library.

Because the pack can only be changed when "start over" is pressed, a small preview of the pack image, name, and artist is shown as part of the "start over" button, when in a standalone project level.  Because a pack can be specified by a non-standalone project level, but can not be changed by the user, the small preview is shown separately in the header, beside the "start over" button, in that case.

The source saved to the server gains a new `packId` value:
```
{ labConfig: { music: { library: "launch2024", packId: "pop"} }, source: "..." }
```

### Follow-up

Some follow-up work will probably want to take the chosen pack's bpm and key and use that for remaining sounds so that they match, when using Tone.js.

### Media

Here is a standalone project level using a music library with four restricted packs and two non-restricted packs.  The dialog is shown immediately, the "start over" button expands to contain the selected pack, and starting over causes the dialog to be shown again.  

https://github.com/code-dot-org/code-dot-org/assets/2205926/a709a4b9-578a-4ab3-b2de-2a2d6c990dcb

Here is a level in a progression that has specified a `packId`.  In this case, the pack will be available for use, and the selected pack is shown in the header, but separate to the "start over" button.

Screenshot 2024-03-30 at 11 37 22 AM


apps/src/music/views/HeaderButtons.tsx
```diff
@@ -25,14 +72,24 @@ const HeaderButtons: React.FunctionComponent<HeaderButtonsProps> = ({
...
+  const currentPackId = useAppSelector(state => state.music.packId);
   const analyticsReporter = useContext(AnalyticsContext);
   const dialogControl = useContext(DialogContext);
 
+  const library = MusicLibrary.getInstance();
+
+  let packFolder = null;
+
+  if (library && currentPackId) {
+    packFolder = library.getAllowedFolderForFolderId(undefined, currentPackId);
```
sanchitmalhotra126, line 90: For later - now that instrument, kit, and pack sounds are separated in the library, it would be great to rework this and similar functions so that we don't need to know the implicit assumption that a folder type of `undefined` means non-instrument/non-kit sounds.

apps/src/music/views/PackDialog.tsx
```diff
@@ -0,0 +1,174 @@
...
+    <FocusLock className={styles.focusLock}>
+      <div className={styles.dialogContainer}>
+        <div id="pack-dialog" className={styles.packDialog}>
+          <div id="hidden-item" tabIndex={0} role="button" />
+          <Typography
+            semanticTag="h1"
+            visualAppearance="heading-lg"
+            className={styles.heading}
+          >
+            Choose Pack
```
sanchitmalhotra126, line 151: Localization?

apps/src/music/views/MusicView.jsx
```diff
@@ -634,6 +655,11 @@ class UnconnectedMusicView extends React.Component {
           clearCode={this.clearCode}
           validator={this.musicValidator}
           player={this.player}
+          allowPackSelection={
+            this.library?.getHasRestrictedPacks() &&
+            !this.props.levelData?.packId &&
```
sanchitmalhotra126: Should we check `packId` in redux here so we don't show the dialog once the pack has been selected?

apps/src/music/views/PackDialog.module.scss
```diff
@@ -0,0 +1,102 @@
...
+    border: solid 1px $neutral_dark40;
+
+    .heading {
+      color: $neutral_light;
+    }
+
+    .packs {
+      width: 100%;
+      height: 100%;
+      overflow-y: scroll;
```
sanchitmalhotra126: Minor: can we change this to `auto`? I noticed that on my machine, I was getting the scrollbar even if there was only one row of items

Screenshot 2024-04-02 at 11 05 29 AM

sanchitmalhotra126 (approved):
This is awesome! Played around with this a bit on standalone project levels and level progressions and it works great. Nice work!

apps/src/music/views/PackDialog.tsx
breville: Expecting a design pass soon, so I think it makes sense to wait and see what the UI looks like after that.

apps/src/music/views/PackDialog.module.scss
breville: Good catch.  Done.

apps/src/music/views/HeaderButtons.tsx
breville: Good call.

apps/src/music/views/MusicView.jsx
breville: Here it actually means "is the user allowed to ever select their own pack", because `MusicLabView` also passes this value through to `HeaderButtons` which uses it to decide which of the two UI variants to show: a "combined artist + start over button" if the user is allowed to select their own pack, versus "separate artist + start over buttons" if the user is not allowed to select their own pack.

apps/src/music/views/HeaderButtons.tsx
breville: Hopefully addressed this in https://github.com/code-dot-org/code-dot-org/pull/57794.

PR #57535, opened 2024-03-26
Add noindex to any page with allthe in the path

bethanyaconnor:
Adds a noindex meta tag to any Rails path with "allthe" in the URL.

We have a few `allthe*` curriculum these days, so I didn't want to constrain this to just `allthethings`. I'm happy to explore a tighter regex if folks want me to, though I don't think we have any page on the site with "allthe" in the URL, except for these curriculum.

Searching "allthe" in our script_json/ directory gave me these results, which I believe are all test curricula.
```
bethanys-mbp:code-dot-org bethanyconnor$ git grep -l 'allthe' dashboard/config/scripts_json/
dashboard/config/scripts_json/alltheblocklythings.script_json
dashboard/config/scripts_json/allthehiddenthings.script_json
dashboard/config/scripts_json/allthemigratedthings.script_json
dashboard/config/scripts_json/alltheplcthings.script_json
dashboard/config/scripts_json/alltheselfpacedplthings.script_json
dashboard/config/scripts_json/allthesurveys.script_json
dashboard/config/scripts_json/allthethings.script_json
dashboard/config/scripts_json/allthettsthings.script_json
dashboard/config/scripts_json/allthevalidation.script_json
```

There is `dashboard/config/courses/allthethingscourse.course`, which creates a course names `allthethingscourse`, but that is also a test curriculum. The string "allthe" does not appear in routes.rb. 

## Testing story

spot checked a few pages, including the home page, the allthethings unit overview, some allthethings levels, and coursee-2022. They all did or did not have this meta tag as expected.

## PR Checklist:

- [ ] Code is well-commented


Hamms (commented):
This does feel like the kind of thing it would be nice to make discoverable from the perspective of a content editor creating new scripts. We certainly have a convention of only ever using "allthe*" for test scripts, but I worry about leaning too heavily on undocumented and unenforced conventions.

How hard would it be to add a notification about this convention and its effects on searchability that pops up here if it detects a relevant name?

image

Hamms:
Almost feels like we should have a "this is intended for internal testing" flag that can be set on a unit. Be a bit more explicit and intentional about all of this.

bethanyaconnor:
> Almost feels like we should have a "this is intended for internal testing" flag that can be set on a unit. Be a bit more explicit and intentional about all of this.

ooo I like that! Though, I think most curriculum writers keep their test scripts in `in_development`, which wouldn't show up in search results. Relatedly, part of me is wondering if we should just add this for all curriculum that isn't `preview` or `stable` but that feels like more of a product change than I'd like to make here.

Hamms:
Oooh, yeah, tying this to our existing unit lifecycle property makes a ton of sense to me. Definitely seems like anything that hasn't been officially released shouldn't be getting indexed.

PR #54597, opened 2023-10-31
HoC 2023 - Code.org - Update code.org/hourofcode

dmantonyuk:
_(This PR description is authored by @kelbyhawn)_

Update https://code.org/hourofcode
- Page is fully responsive
- Updated OpenGraph tags and image
- Redirected and deprecated https://code.org/hourofcode/overview to https://code.org/hourofcode
- Updated outdated `/hourofcode/overview` links

**Jira ticket:** https://codedotorg.atlassian.net/browse/ACQ-995

----

localhost code org_3000_hourofcode

PR #57536, opened 2024-03-26
Delete unused pegasus files

bethanyaconnor:
Removes some dead code from pegasus.

- The reference to `hoc-overview.js` was removed in https://github.com/code-dot-org/code-dot-org/pull/54597
- The reference to `csf-congrats.js` was removed in https://github.com/code-dot-org/code-dot-org/pull/48211
- `middle-high-alt.js` and `school_info.haml` were listed in https://codedotorg.atlassian.net/browse/XTEAM-415 as dead code. I couldn't figure out when the references were removed, but I confirmed there are no references to them.


kelbyhawn (approved):
🔥

hannahbergam (approved):
FUN! Confirmed locally all but school_info (just trickier search-wise)

bethanyaconnor:
> Confirmed locally all but school_info (just trickier search-wise)

I only found two results for "school_info" in pegasus and neither loaded the view. If we're somehow using the view across pegasus/dashboard, I'm excited to see what breaks :)

PR #57537, opened 2024-03-26
Update featured_project_controller and tests

fisher-alice:
This PR updates the functions that freeze and unfreeze featured projects in the `featured_projects_controller`. It removes the early return if `project_id` is not defined which was added to avoid unit test errors having to do with created projects added in https://github.com/code-dot-org/code-dot-org/pull/57495.

Thanks to @bencodeorg for his help with this PR!

I closed [this past PR](https://github.com/code-dot-org/code-dot-org/pull/57476) due to these unit test errors in test files un-related to my changes. See [failed Drone build](https://drone.cdn-code.org/code-dot-org/code-dot-org/3171/1/5)

Screenshot 2024-03-26 at 9 19 30 AM

Screenshot 2024-03-26 at 9 19 51 AM

Dashboard unit tests that include created projects have historically caused issues in testing. See [Slack thread](https://codedotorg.slack.com/archives/C046G4TRLEN/p1697038404300649) with another example of similar issue reported by @bencodeorg .

After consulting with Ben, I modified the tests in `featured_projects_controller_test.rb` to use `setup_all` instead of `setup`. This resolved the mysql timeout errors! I tried setting `self.use_transactional_test_case` to `false`, but this did not seem to make a difference so removed dc267143ecfc75b3e0657b64290065c17cd14bc5.

I updated the `featured_project` factory by adding 2 additional types of `featured_project`: `active_featured_project` and `archived_featured_project` to help with testing.

WIth these changes, I've re-enabled two skipped tests that were flaky in the past -  'project validators can feature projects' and 'featuring a currently unfeatured project should update the correct featured project'. Hopefully, these tests will be more robust.

There is a comment in `code_review_controller_test` that may be related to the mysql timeout errors I was seeing prior to using `setup_all`.
```
  # Setting this to true causes some weird db locking issue, possibly due to
  # some writes to the projects table coming from a different connection via
  # sequel.
  self.use_transactional_test_case = false
```
Although I didn't need to set this value to `false`, it does seem like the `projects` table has some complexities which have  caused testing issues to occur.

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/dashboard/app/controllers/featured_projects_controller.rb
+++ b/dashboard/app/controllers/featured_projects_controller.rb
@@ -17,14 +17,7 @@ class FeaturedProjectsController < ApplicationController
     _, project_id = storage_decrypt_channel_id(params[:channel_id])
     return render_404 unless project_id
     @featured_project = FeaturedProject.find_or_create_by!(project_id: project_id)
-    puts "@featured_project in controller - feature #{@featured_project}"
-    puts "@featured_project.featured_at #{@featured_project.featured_at}"
-    puts "@featured_project.unfeatured_at #{@featured_project.unfeatured_at}"
     @featured_project.update! unfeatured_at: nil, featured_at: DateTime.now
-    puts "@featured_project in feature#{@featured_project}"
-    puts "@featured_project after UPDATE #{@featured_project}"
-    puts "@featured_project.featured_at #{@featured_project.featured_at}"
-    puts "@featured_project.unfeatured_at #{@featured_project.unfeatured_at}"
     # Set the featured project's abuse score to -50.
     buffer_abuse_score
     freeze_featured_project(project_id)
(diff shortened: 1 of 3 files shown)


bencodeorg:
Thank you for digging into this!

I haven't worked with factories in a while, so I pulled down your branch to see if what I was thinking might simplify some of this actually made sense. Tests appeared to pass -- I have two commits (branched off your branch) here:

https://github.com/code-dot-org/code-dot-org/pull/57548

fisher-alice:
> Thank you for digging into this!
> 
> I haven't worked with factories in a while, so I pulled down your branch to see if what I was thinking might simplify some of this actually made sense. Tests appeared to pass -- I have two commits (branched off your branch) here:
> 
> #57548

This is great! I appreciate the simplification. Just merged your changes.

bencodeorg (approved):
Nice!

PR #57539, opened 2024-03-26
Run drone builds on all branches

lfryemason:
From a recent slack conversation:

Currently, we run drone builds only on branches that are merging directly to `staging` or `staging-next` branches. This was originally to prevent waste of resources. However, this does not prevent much since most branches go directly to staging directly and there are cases where running drone on other branches would be helpful. E.g. a feature branch that merges smaller branches into a main feature branch before merging the feature in one PR. In this feature branch case, it would be helpful to run drone builds on all PRs to prevent bugs or large debugging before merging the larger PR.

This changes:
* Drone only runs on PRs merging into `staging` or `staging-next` -> Drone runs on all PRs not merging into `production` or `test` 

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -114,6 +114,7 @@ trigger:
     exclude:
       - "production"
       - "test"
+      - "levelbuilder"
   event:
     - pull_request
 
@@ -283,6 +284,7 @@ trigger:
     exclude:
       - "production"
       - "test"
+      - "levelbuilder"
   event:
     - push
 ---
--- a/.drone.yml
+++ b/.drone.yml
@@ -18,8 +18,8 @@ steps:
     image: curlimages/curl
     commands:
       # If running on EC2, get hostname from ec2 metadata
-      - "IMDS_TOKEN=$(curl -s -X PUT http://169.254.169.254/latest/api/token -H "X-aws-ec2-metadata-token-ttl-seconds: 21600")"
-      - "hostname=$(curl -s -H "X-aws-ec2-metadata-token: $IMDS_TOKEN" http://169.254.169.254/latest/meta-data/public-hostname)"
+      - "IMDS_TOKEN=$(curl -s -X PUT http://169.254.169.254/latest/api/token -H \"X-aws-ec2-metadata-token-ttl-seconds: 21600\")"
+      - "hostname=$(curl -s -H \"X-aws-ec2-metadata-token: $IMDS_TOKEN\" http://169.254.169.254/latest/meta-data/public-hostname)"
       - echo "Running on $hostname"
 
   # We cache a clone of staging in the `cache-staging-clone` pipeline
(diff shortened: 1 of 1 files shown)


.drone.yml
```diff
@@ -242,8 +242,9 @@ volumes:
 
 trigger:
   branch:
-    - "staging"
-    - "staging-next"
+    exclude:
+      - "production"
+      - "test"
```
cat5inthecradle, line 249: Our automated Deploy To Levelbuilder (DTL) process [creates a PR every day](https://github.com/code-dot-org/code-dot-org/pulls?q=is%3Apr+base%3Alevelbuilder+is%3Aclosed). We should add it to the list too.

```suggestion
    exclude:
      - "production"
      - "test"
      - "levelbuilder"
```

Hamms (approved):
Looks good to me!

I am curious if we have any metrics on number of drone runs per day, or even number of servers that get spun up every day. Would be interesting to see how much of an impact this change has on our usage, even if just to verify that it's not very much.

.drone.yml
```diff
@@ -279,12 +281,12 @@ steps:
...
+      - "levelbuilder"
```
wilkie: Aaaactually... should we only do this on a push to `staging` / `staging-next` ? Is that the intended behavior... that it is a cache of what `staging` looks like?

wilkie: I'm not confident in what this does to be sure.

lfryemason:
@Hamms I realized we could get a quick check of how many branches will cause an extra drone build with github.

Since February 1 (~7 weeks), there were 60 PRs that would have triggered
image

search query: `is:pr updated:>=2024-02-01 -base:levelbuilder -base:staging -base:staging-next -base:production -base:test is:closed `

All of them were human made (and most from Seth)

.drone.yml
Hamms: I *think* it makes sense to still use the cache of staging on branches off branches; the idea is that rather than cloning the entire repro from scratch, we restore a cache of what's on staging and then just fetch whatever deltas are necessary for the specific branch we're looking at. Even branches off branches off staging should benefit from that.

Hamms (commented):
Oh, one other thing I forgot earlier:

Whenever you make a change to `.drone.yml`, you also need to regenerate the hmac signature. See ["making changes" in our Drone documentation](https://docs.google.com/document/d/1Qls20xfNN6T_DErOMwVQOFJsxAEAdMWFHzBDIbxyUQQ/edit#heading=h.blmise7gzfvg) for instructions on how to do that.

.drone.yml
cat5inthecradle: Changing this trigger might be causing us to cache staging a lot more often. Notice the missing `event`

Hamms: Oh, lol, no; @wilkie you were absolutely right! This isn't the step that *restores* the cache, it's the step that *creates* the cache

Hamms (approved):
Might be nice to remove `staging-next` from the `cache-staging-clone` trigger while we're in here, but even without that this LGTM! Thanks for the improvement

lfryemason:
Drone now running on branches off this branch. Going to merge
https://github.com/code-dot-org/code-dot-org/pull/57598

PR #57541, opened 2024-03-26
Diversity page updates

dmcavoy:
Update code.org/diversity. Redirected code.org/promote/diversitydata to /diversity.

Note there are a lot of file changes because I did some clean up. I removed old images we didn't need any more and moved all the still needed images to one folder. Also FYI there is a tableau dashboard that doesn't show in the screenshot. Also for some reason the full screen screenshot didn't get the whole page...

localhost code org_3000_diversity (1)


kelbyhawn (approved):
Looks great! 🎉 Thanks for doing this one.

PR #57542, opened 2024-03-26
Add templates folder unit tests to import order linting

lfryemason:

**For PR reviewers**: please double check that all comments in imports are appropriately moved with the line they are commenting on.


Hamms (approved):
LGTM!

Doesn't look like there were very many comments; all I saw were two instances of a `react testing library import` comment, and both look to me to have been moved correctly.

PR #57546, opened 2024-03-26
Remove "Partner with us" from the hamburger menu

kelbyhawn:
Removes the "Partner with us" link from the hamburger menu on code.org and studio.code.org.

 **Jira ticket:** [ACQ-1652](https://codedotorg.atlassian.net/browse/ACQ-1652)

----

## Signed out
| Before | After | 
| ----- | ------ |
| Before_signed_out | After_signed_out |

## Signed in
| Before | After | 
| ----- | ------ |
| Before_signed_in | After_signed_in |

### DIFF
--- a/lib/cdo/hamburger.rb
+++ b/lib/cdo/hamburger.rb
@@ -121,7 +121,6 @@ class Hamburger
       {title: "educate_middle", url: CDO.code_org_url("/educate/curriculum/middle-school")},
       {title: "educate_high", url: CDO.code_org_url("/educate/curriculum/high-school")},
       {title: "educate_hoc", url: "https://hourofcode.com"},
-      {title: "educate_partner", url: CDO.code_org_url("/educate/partner")},
       {title: "educate_beyond", url: CDO.code_org_url("/educate/curriculum/3rd-party")},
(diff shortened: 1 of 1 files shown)


lib/cdo/hamburger.rb
```diff
@@ -121,7 +121,6 @@ def self.get_hamburger_contents(options)
       {title: "educate_middle", url: CDO.code_org_url("/educate/curriculum/middle-school")},
       {title: "educate_high", url: CDO.code_org_url("/educate/curriculum/high-school")},
       {title: "educate_hoc", url: "https://hourofcode.com"},
-      {title: "educate_partner", url: CDO.code_org_url("/educate/partner")},
```
bethanyaconnor, line 124: Can we remove the string associated with this? (not worth it if we think it might be used elsewhere)

bethanyaconnor (approved):
🚮

lib/cdo/hamburger.rb
kelbyhawn: It doesn't look like it used anywhere else, but I'm hesitant to remove it just in case we want to use it again.

PR #57547, opened 2024-03-26
Lab2: use level paths instead of full URLs

sanchitmalhotra126:
After enabling [Webpack hot module replacement](https://github.com/code-dot-org/code-dot-org/pull/56882), we noticed an issue with Lab2 progressions; Lab2 progressions don't perform a full page reload between levels, but do update the browser stack so that back/forward buttons still act as intended. The issue here is that we were previously updating the stack using level's fully qualified URL which is constructed on the rails side, with a hardcoded port of 3000. Because HMR requires us to run on port 9000, pushing a URL with port 3000 onto the stack was causing a crash. The solution here is to just pass through and use the relative level path instead of full URL, which is a more ideal solution anyway.

https://github.com/code-dot-org/code-dot-org/assets/85528507/638bf04a-6af5-4d24-8199-26194f773723

## Links

https://codedotorg.slack.com/archives/C05DMS18MEX/p1711039238242839

## Testing story

Tested with the Music Lab intro script + a few allthethings lab2 progressions.

### DIFF
--- a/apps/src/code-studio/browserNavigation.js
+++ b/apps/src/code-studio/browserNavigation.js
@@ -49,10 +49,10 @@ export function setupNavigationHandler(lessonData) {
 // onto the browser session history stack, and updating the window title.
 export function updateBrowserForLevelNavigation(
   progressStoreState,
-  levelUrl,
+  levelPath,
   levelId
 ) {
-  window.history.pushState({}, '', levelUrl + window.location.search);
+  window.history.pushState({}, '', levelPath + window.location.search);
   setWindowTitle(progressStoreState, levelId);
 }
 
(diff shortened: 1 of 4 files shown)


bencodeorg (approved):
Looks great, thanks for doing this!

Was about to ask about what is allows to be pushed to history, but [the docs](https://developer.mozilla.org/en-US/docs/Web/API/History/pushState) answered my question in case anyone else is curious:

> The new history entry's URL. Note that the browser won't attempt to load this URL after a call to pushState(), but it may attempt to load the URL later, for instance, after the user restarts the browser. **The new URL does not need to be absolute; if it's relative, it's resolved relative to the current URL. The new URL must be of the same [origin](https://developer.mozilla.org/en-US/docs/Glossary/Origin) as the current URL; otherwise, pushState() will throw an exception. I**f this parameter isn't specified, it's set to the document's current URL.

PR #57556, opened 2024-03-26
Revert "Update Redis gem to 4.8.1"

Hamms:
Reverts code-dot-org/code-dot-org#57204

Seems like this caused a production build failure. See thread at https://codedotorg.slack.com/archives/C0T0PNTM3/p1711487652183599 for more

PR #57557, opened 2024-03-26
Revert "Revert "Update Redis gem to 4.8.1""

Hamms:
Reverts code-dot-org/code-dot-org#57556, restoring https://github.com/code-dot-org/code-dot-org/pull/57204

The original attempt failed in production because we had an invocation of the deprecated method `Redis.connect` gated behind an `if CDO.geocoder_redis_url` clause, and only the production environment had a setting for that configuration value.

In order to prevent future similar occurrences, we add a functional default value for that property. We currently install Redis in developer environments as per SETUP.md and in our Chef-managed environments [via the `cdo-redis` gem](https://github.com/code-dot-org/code-dot-org/blob/df21b3aac9042f063e81286fc1a26d7a393f7f34/cookbooks/cdo-apps/recipes/default.rb#L141-L142). That leaves Drone as the only environment to not currently have access to Redis, which this PR corrects.

## Testing story

I updated the test logic in one commit and verified that with that change in place, the test caught the invalid invocation of `Redis.connect`. I then fixed the invalid invocation with another commit, and verified that the test now passes.

## PR Checklist:

- [ ] Code is well-commented


breville (approved):
Thanks for working through this.  Not familiar with these configurations, but just like a contestant on **The Bachelor** franchise, I trust the process :)

PR #57558, opened 2024-03-26
Provision new Redis-based Session Store

Hamms:
The first step in our efforts to transfer user session data from cookies to a centralized data store is to provision the data store in Cloudformation, so we can begin writing data to it.

We have quite a few options for how exactly we want to provision this, but broadly speaking we can choose either Serverless or On-Demand nodes, and if we choose On-Demand we also must choose both the size and quantity of nodes we want.

According to Google Analytics, we've had 28M users over the last 90 days, and 84M over the last year, which works out to between 45 and 60 million users on average over an arbitrary 200-day period (200 days being our current session duration configuration). At ~300 bytes per session on average, that's 13.5-18 GB of data we'd ideally like to store.

Storing 18GB of data in Severless ElastiCache would cost ~$1620/month; even 13GB would still be ~$1170/month.  Alternatively, we could use three of the smallest memory-optimized cache node (cache.r7g.large) and get 39.21GB of storage for ~$480/month, or even just use two for 26.14GB of storage at ~$320/month. An even more targeted solution could be to use a larger number of smaller nodes; 6 cache.t4g.medium nodes would allow us to store 18.5GB for ~$280/month.

Throughput is less of an issue for us than storage. According to CloudWatch, we average ~800M requests per week, 15M/hour, and under 4,200/second. At an average of 300 bytes stored per session and one GET plus one PUT per request, that's 2.5 Gigabits of needed network performance for on-demand nodes (well under the 12.5 promised by cache.r7g.large nodes), or ~$25/month in ECPUs for a Serverless model.

This PR proposes we provision a single cluster consisting of three cache.r7g.large nodes, with encryption, automatic failover, and multi-zone availability all enabled. That should give us plenty of room for growth without too much unnecessary expense. The next step in our broader efforts to enable this feature is to [begin writing data to redis so we can get a more accurate idea of actual usage](https://github.com/code-dot-org/code-dot-org/pull/57562), at which point we can revisit these details.

## Links

- [Average Maximum Session Size and Number of Sessions Seen](https://us-east-1.console.aws.amazon.com/cloudwatch/home?region=us-east-1#metricsV2?graph=~(metrics~(~(~'code-dot-org*2fActionDispatch~'AverageMaxSessionSize~'Environment~'production~(yAxis~'right))~(~'.~'NumSessionsSeen~'.~'.))~view~'timeSeries~stacked~false~region~'us-east-1~period~60~stat~'Average~start~'-PT72H~end~'P0D)&query=~'*7bcode-dot-org*2fActionDispatch*2cEnvironment*7d)
- [Total requests per week](https://us-east-1.console.aws.amazon.com/cloudwatch/home?region=us-east-1#metricsV2?graph=~(metrics~(~(~'AWS*2fApplicationELB~'RequestCount~'LoadBalancer~'app*2fautoscale-prod*2f587ee8f918e6fd28~(id~'requests~label~'Requests~region~'us-east-1)))~period~3600~stat~'Sum~region~'us-east-1~view~'timeSeries~stacked~false~title~'Web*20Application*20Server*20Efficiency~yAxis~(left~(min~0)~right~(min~0))~start~'-PT336H~end~'P0D))
- [90-day active users](https://analytics.google.com/analytics/web/#/p367744764/reports/reportinghub?params=_u..nav%3Dmaui%26_u.dateOption%3Dlast90Days%26_u.comparisonOption%3Ddisabled)

## Testing story

To test, I temporarily moved the newly-added components outside of the `if frontends` block and started an adhoc of this branch. I verified that [the expected resource](https://us-east-1.console.aws.amazon.com/elasticache/home?region=us-east-1#/redis/adsy4mzt3n3c8qg) was created successfully.

I was unfortunately unable to actually connect to the provisioned redis cluster from the adhoc, because [the security group](https://github.com/code-dot-org/code-dot-org/blob/32c87af11266a304008bb50b3adcaa6f83ab09dd/aws/cloudformation/vpc.yml.erb#L182-L191) we apply to the resource only allows access from frontends. Once this change has made it to production, I plan to verify that the cluster is accessible from a production frontend server.

## Deployment strategy

Plan to monitor the DTP that includes this change closely, and have a revert prepped.

## Follow-up work

- https://github.com/code-dot-org/code-dot-org/pull/57562
- Eventually, migrate session data to redis and stop using the cookie for storing anything more than just an identifier.

## PR Checklist:

- [ ] Code is well-commented


cat5inthecradle:
RE: Testing Cloudformation templates, there's no test that comes close to reality, so deploying to a non-prod environment - either an adhoc, or using it in Test - is generally the best way. There is some linting and validating we can do though.

**cfn-lint** and `cloudformation validate` are great... but they're not going to work on an erb file. I think the following process works:

```
# (edit your locals.yml to populate slack_bot_token, slack_token, and any other secrets the following complains about)
VERBOSE=1 bundle exec rake stack:validate RAILS_ENV=production > stackdump.yml
# (edit the stackdump.yml file to make it proper yml
cfn-lint stackdump.yml
```

~Doing that myself on your branch, it looks like we actually have a number of linting warnings, likely because we're covering lots of conditional cases and need some just-in-case redundancy - but there's surely some opportunities here.. Your new resource is between lines 832-855 and has a couple of errors which I'll extract to inline comments in the PR.~

EDIT: I was running an old version of cfn-lint - your new template code passes the basic linter.

cat5inthecradle:
For reference, here's the template that I rendered with `RAILS_ENV=production`

```
  SessionStoreSubnetGroup:
    Type: AWS::ElastiCache::SubnetGroup
    Properties:
      Description: Session Store Subnet Group
      SubnetIds: [{"Fn::ImportValue":"VPC-SubnetB"},{"Fn::ImportValue":"VPC-SubnetC"},{"Fn::ImportValue":"VPC-SubnetD"},{"Fn::ImportValue":"VPC-SubnetE"}]
  SessionStoreGroup:
    Type: AWS::ElastiCache::ReplicationGroup
    Properties:
      ReplicationGroupDescription: Session Store Replication Group
      # 3 nodes of cache.r7g.large will cost ($0.219 * 730 * 3) = ~$480/month
      # and provide just under 40GB of storage.
      CacheNodeType: cache.r7g.large
      Engine: Redis
      EngineVersion: 7.1

      ClusterMode: Disabled
      NumCacheClusters: 3 # 1 primary and 2 replicas
      AutomaticFailoverEnabled: true
      MultiAZEnabled: true
      TransitEncryptionEnabled: true
      AtRestEncryptionEnabled: true

      SecurityGroupIds: [!ImportValue VPC-RedisSecurityGroup]
      CacheSubnetGroupName: !Ref SessionStoreSubnetGroup
  ```

cat5inthecradle:
Went down a rabbit hole into #57577 to get some better local linting setup.

cat5inthecradle (approved):
This looks good if I accept the way our sessions currently work.

The idea of enabling ClusterMode, setting up AutoScaling and more, smaller, nodes interests me. Does our 200 day (:scream:) session need to be that long-lived? This spans summer and effectively smooths out our storage usage over time, right?

During the summer (and holidays, weekends, and nights) we will see less throughput demand. We would also see fewer _new_ sessions, which might translate to lower storage requirements.

$480/month is small, and we can scale up manually before Hour of Code - though research is suggesting that this would incur downtime and possibly some failed writes on the edges of that downtime window. It might make sense to invest in some error handling and resiliency in the implementation of the cache, if that's not already a feature, so users can weather a scale-up.

Hamms:
Fantastic, thank you for the thorough review!

Definitely agree that some kind of auto-scaling would be nice, but I think our expected usage is going to be low enough that it'll be unnecessary. I am curious, though, where did you see that scaling up might incur downtime? In particular, note that this is a "cluster mode disabled" cluster, for which scaling just involves [increasing the number of replicas](https://docs.aws.amazon.com/AmazonElastiCache/latest/APIReference/API_IncreaseReplicaCount.html); my understanding is that should be seamless. Eager to learn if I'm wrong about that!

I think our session certainly doesn't *need* to be that long-lived, and if we wanted to spend even less on this cluster then shortening the expiration would probably be the easiest way to reduce our needs here. But as you said, the proposed cost is pretty reasonable and I think gets us enough that we shouldn't need to worry about scaling up. I'm not even sure if we will need to scale up for Hour of Code; from a data storage perspective, increased usage just means we'll be expiring old sessions (following the least-recently-used strategy) slightly sooner than we otherwise would, which should be fine. And from a throughput perspective, I think we've got more than enough leeway for the ~5x increase that I expect.

Hamms:
I added some notes to "Testing story"; PTAL!

cat5inthecradle:
>scaling just involves [increasing the number of replicas](https://docs.aws.amazon.com/AmazonElastiCache/latest/APIReference/API_IncreaseReplicaCount.html)

Perfect. I was thinking scaling up, but scaling out makes far more sense here.

PR #57561, opened 2024-03-26
App Lab: add api.sportsdata.io

breville:
Add support for `api.sportsdata.io` to be accessed using `startWebRequest`.

**Tip:** Use `bundle exec rake seed:code_docs` in `dashboard/` to see the updated documentation locally at `docs/ide/applab/expressions/startWebRequest`.


bencodeorg:
Were you able to test this? It looks paid (or at least requires an API key)?

bencodeorg:
TIL re: `seed:code_docs`!

breville:
> Were you able to test this? It looks paid (or at least requires an API key)?

I did not test the service.

bencodeorg (approved):
Presumably this will unblock some student who wrote in who wants to use it? Seems like we allow other APIs that require keys (eg, https://coinlayer.com/documentation), so that's not anything new here.

breville:
> Presumably this will unblock some student who wrote in who wants to use it? Seems like we allow other APIs that require keys (eg, https://coinlayer.com/documentation), so that's not anything new here.

Yes, this came from a user request.  And we do permit APIs so long as the key can be passed by query parameter, which is possible here.

PR #57562, opened 2024-03-26
Write Session Data to Redis

Hamms:
Notably, however, do not yet begin relying on Redis for *reading* session data.

We want to get an accurate evaluation of what real-world usage of this new Redis cluster will look like, and also benefit from some load testing in advance of an actual switchover. To that end, this PR proposes that we begin writing all session data to Redis as a side effect of using the existing cookie-based store. This new functionality will be gated behind a DCDO flag and wrapped in error handling logic to minimize the chances of this experiment affecting the user experience.

Also remove the instrumented cookie store; we don't need to manually gather data for estimates now that we'll have an actual Redis cluster to monitor.

## Links

Follow-up to https://github.com/code-dot-org/code-dot-org/pull/57558

## Testing story

Tested the Redis cluster provisioned in production by SSHing into a frontend instance and executing some manual queries from the dashboard console:

```ruby
[production] dashboard > client = Redis.new(url: "rediss://[redacted].cache.amazonaws.com:6379")
=> #<Redis client v3.3.3 for redis://[redacted].cache.amazonaws.com:6379/0>
[production] dashboard > client.get('foo')
=> nil
[production] dashboard > client.keys
=> []
```

Tested the functionality added in this PR by temporarily modifying `cloud_formation_stack.yml.erb` to provision a redis session storage cluster even in an adhoc environment, then verified on [an adhoc](https://adhoc-write-to-redis-based-session-store-studio.cdn-code.org/users/sign_in) that registration, authentication, and session destruction all work as expected, and that I can successfully find session data in the redis store.

## Deployment strategy

We should plan to monitor Honeybadger and Zendesk after this DTPs, and have a revert prepped just in case.

After deployment, we should plan to set the DCDO flag `mirror_session_in_redis_enabled` to `true` so we can begin collecting data.

## Follow-up work

Assuming that our new Redis cluster is able to handle the load generated by this mirroring, we can follow this up with [an implementation that will begin using the Redis store rather than cookies as the source of truth](https://github.com/code-dot-org/code-dot-org/blob/8dc80dd991ebccb34c915816cfc85d9ae8dd2973/dashboard/config/custom_session_store/migrate_cookies_to_redis_store.rb).

## PR Checklist:

- [ ] Code is well-commented


aws/cloudformation/bootstrap_chef_stack.sh.erb
```diff
@@ -113,7 +113,7 @@ cat <<JSON > $FIRST_BOOT
   },
   "cdo-secrets": {
 <% if frontends -%>
-    "session_store_server": "redis://${SessionStoreGroup.PrimaryEndPoint.Address}:${SessionStoreGroup.PrimaryEndPoint.Port}",
+    "session_store_server": "rediss://${SessionStoreGroup.PrimaryEndPoint.Address}:${SessionStoreGroup.PrimaryEndPoint.Port}",
```
Hamms, line 116: Note that [because we have enabled encryption for the underlying Redis cluster](https://github.com/code-dot-org/code-dot-org/blob/cf805a0e720f0c35f4d0a8b00c3633a59f71b3c3/aws/cloudformation/components/session_store.yml.erb#L21), we have to specifically use a "redis secure" connection here, hence `rediss` rather than `redis`.

I very briefly tried to find a way to automatically determine whether or not this is necessary based on the CloudFormation output, but couldn't. I decided to just quickly hardcode it instead; please let me know if you think it would be worth me spending a bit more time on this!

sureshc:
We don't yet have a standard way to take the Physical ID of a Resource provisioned in CloudFormation and pass that configuration value to our Ruby web application servers.

Specifically, we don't have an easy way to populate `CDO.session_store_server` from the `SessionStore` CloudFormation Resource. Our usage of `SessionStoreGroup.PrimaryEndPoint.Address` and `SessionStoreGroup.PrimaryEndPoint.Port` in this feature branch only passes those values to the Chef cookbooks that execute when the daemon instance is provisioned. Those values don't get persisted to that server nor are they ever passed to web application servers or the console.

The very janky workaround I've come up with is to publish an AWS Stack-specific Secret in the template and then use `!StackSecret` to populate `CDO.session_store_server`

cat5inthecradle (approved):
LGTM

Would there be value in reading the session from Redis in order to evaluate that half of usage too? Or is it pretty much 1:1 with writes?

Hamms:
> Specifically, we don't have an easy way to populate CDO.session_store_server from CloudFormation Logical IDs and Attributes SessionStoreGroup.PrimaryEndPoint.Address and SessionStoreGroup.PrimaryEndPoint.Port

Suresh, can you clarify how the changes to `aws/cloudformation/bootstrap_chef_stack.sh.erb` are insufficient? Note that the old version of this value has already been populated on frontends: 

```
[production] dashboard > CDO.session_store_server
=> "redis://[redacted].cache.amazonaws.com:6379"
```

Is it just a problem with updating the value once it's been initially set? My expectation is that whenever we spin up a new frontend, it will pick up these changes.

> Would there be value in reading the session from Redis in order to evaluate that half of usage too? Or is it pretty much 1:1 with writes?

In my local testing, it seems like session reads and writes are in fact *exactly* 1:1; every single request incurs exactly one write and one read. I'm certainly not opposed to adding some logic to also read just for thoroughness, but I think just doubling the number of operations generated as a result of this change should accurately capture the number of operations we'll need for full usage.

cat5inthecradle:
> We don't yet have a standard way to take the Physical ID of a Resource provisioned in CloudFormation and pass that configuration value to our Ruby web application servers.
> 
> Specifically, we don't have an easy way to populate `CDO.session_store_server` from the `SessionStore` CloudFormation Resource. Our usage of `SessionStoreGroup.PrimaryEndPoint.Address` and `SessionStoreGroup.PrimaryEndPoint.Port` in this feature branch only passes those values to the Chef cookbooks that execute when the daemon instance is provisioned. Those values don't get persisted to that server nor are they ever passed to web application servers or the console.
> 
> The very janky workaround I've come up with is to publish an AWS Stack-specific Secret in the template and then use `!StackSecret` to populate `CDO.session_store_server`

`session_store_server` should get set correctly on frontends with this change - the bootstrap script runs whenever we wake up AMI Builder.

Hamms:
I'll also plan to resolve https://app.honeybadger.io/projects/3240/faults/106697134 once this PR is merged, since it removes the relevant code.

unlox775-code-dot-org (approved):
Looks good!

PR #57563, opened 2024-03-26
Display only sections being actively synced in the sync confirmation

carl-codeorg:
Recently, the LTI roster sync logic was changed to include only sections where the current user is listed as a teacher. Even if the user owns the course in the LMS, if they aren't listed as a teacher for a given section, that section will not be synced. However, the sync confirmation modal wasn't updated to match this new functionality, so it was still displaying student counts for all sections in the course.

This PR:
- Changes `Services::Lti.sync_course_roster`, which used to just return a boolean for whether or not changes occurred, to returning the whole roster result object that used to only be returned from the `sync_course` route in the controller. Instead of creating this object directly from the NRPS response JSON, the object is now constructed during the actual sync process, and accurately reflects which courses were actually updated. The object used to have both an :all and a :changed property, but the :changed property wasn't being assigned any values. The sync result object is then returned, the same as before, from the `sync_course` route.
- Now that the object returned by the route accurately reflects which sections were actually synced, the `LtiSectionSyncDialog` has been updated to list summaries for only the sections present in the `changed` object, instead of all the sections in the `all` object.
- Added handling in the `LtiSectionSyncDialog` for a response that includes no changes. This is specifically for the sync button in the Teacher Dashboard. If you click the button, the dialog still pops up to provide feedback that the operation was successful despite being a no-op, with updated messaging. See screenshot below for details. Open to suggestions on the new wording.

Example Canvas course where the teacher is listed for Section B, but not Section A:
Screenshot 2024-03-26 at 9 36 29 PM

After syncing, the dialog now only lists section B:
Screenshot 2024-03-26 at 9 37 10 PM

After clicking the sync button on a section that is already up-to-date:
Screenshot 2024-03-26 at 10 41 48 PM

## Links

[JIRA](https://codedotorg.atlassian.net/browse/P20-799)

## Testing story

LTI Controller tests have been updated with new mock return values for the stubbed sync_course_roster service method, which now returns an object instead of a boolean.

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/i18n/common/en_us.json
+++ b/apps/i18n/common/en_us.json
@@ -1602,6 +1602,7 @@
   "low": "low",
   "ltiSectionSyncButtonDisabledAltText": "Roster sync is disabled. To re-enable, go to your account settings.",
   "ltiSectionSyncDialogDescription": "New to Code.org? A [section]({aboutSectionsUrl}) is used to manage your classroom and assign curricula. We've identified the following sections & students associated with your course.\n\nYou can re-sync your sections & students at any time! Learn more [about syncing here]({aboutSyncingUrl}).",
+  "ltiSectionSyncDialogDescriptionNoChange": "You're all set! Your sections & students are already up to date.\n\nYou can re-sync your sections at any time. Learn more [about syncing here]({aboutSyncingUrl})",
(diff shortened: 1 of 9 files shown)


nicklathe (approved):
LGTM

PR #57567, opened 2024-03-26
Fix Violations of and Reenable `Style/GlobalStdStream`

Hamms:
Quoting from [the official Ruby style guide](https://rubystyle.guide/#global-stdout):

> Use `$stdout/$stderr/$stdin` instead of `STDOUT/STDERR/STDIN`. `STDOUT/STDERR/STDIN` are constants, and while you can actually reassign (possibly to redirect some stream) constants in Ruby, you’ll get an interpreter warning if you do so.

Most changes made automatically with `bundle exec rubocop --autocorrect-all --only Style/GlobalStdStream`. I then manually updated a few references in comments and unlinted files, just for consistency.

## Links

- https://rubystyle.guide/#global-stdout
- https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/GlobalStdStream
- https://docs.rubocop.org/rubocop/cops_style.html#styleglobalstdstream

## Testing story

Relying on existing unit tests to verify that this does not result in any change to functionality.

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/.chef/knife.rb
+++ b/.chef/knife.rb
@@ -2,7 +2,7 @@
 
 current_dir = File.dirname(__FILE__)
 log_level                :info
-log_location             $stdout
+log_location             STDOUT
 node_name                @node_name = ENV['CDO_CHEF_NODE_NAME'] || 'cdo-ci'
                          @org_name = ENV['CDO_CHEF_ORG_NAME'] || 'code-dot-org'
 client_key               @client_key = ENV['CDO_CHEF_CLIENT_KEY'] || "#{ENV['HOME']}/.chef/#{@node_name}.pem"
--- a/.chef/knife.rb
+++ b/.chef/knife.rb
@@ -2,7 +2,7 @@
 
 current_dir = File.dirname(__FILE__)
 log_level                :info
-log_location             STDOUT
+log_location             $stdout
 node_name                @node_name = ENV['CDO_CHEF_NODE_NAME'] || 'cdo-ci'
                          @org_name = ENV['CDO_CHEF_ORG_NAME'] || 'code-dot-org'
 client_key               @client_key = ENV['CDO_CHEF_CLIENT_KEY'] || "#{ENV['HOME']}/.chef/#{@node_name}.pem"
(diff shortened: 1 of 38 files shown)


wilkie (approved):
LGTM

wilkie:
https://github.com/code-dot-org/code-dot-org/blob/18d7ebf79e654361b24c63af43dfcf2e13ad5256/.chef/knife.rb#L5 has an instance of `STDOUT`

Not sure if that's important to update as well.

Hamms:
Huh, I wonder why that code path isn't linted by Rubocop. Looks like there are a handful of comments that could probably stand to get updated, too.

```
[~/code-dot-org (fix-Style/GlobalStdStream)]$ git grep "STD\(OUT\|ERR\)" -- ':!apps/.yarn'
.chef/knife.rb:log_location             STDOUT
cookbooks/cdo-users/templates/default/chef_knife.rb.erb:log_location             STDOUT
dashboard/test/lib/image_lib_test.rb:    $stderr = STDERR
lib/cdo/google/sheet.rb:# any unexpected output to STDERR occurs.
lib/cdo/honeybadger.rb:      config.logging.path = "STDOUT"
lib/cdo/honeybadger.rb:    # Unhandled Ruby exceptions are dumped to STDERR in the following format:
lib/cdo/rake_utils.rb:  # Alternate version of RakeUtils.rake which always streams STDOUT to the shell
lib/cdo/rake_utils.rb:  # Alternate version of RakeUtils.system which always streams STDOUT to the
lib/cdo/rake_utils.rb:  # - redirect STDOUT and STDERR streams to a pipe
```

I think it's probably not *important* to update these per se, but seems worth doing anyway, just for consistency.

Hamms:
Frustratingly, because `.chef/knife.rb` isn't regularly linted, it has a number of other lint failures that block the build when I attempt to update it. It also looks like a file that expects to be generated; see https://github.com/code-dot-org/code-dot-org/blob/staging/cookbooks/cdo-users/templates/default/chef_knife.rb.erb. To that end, I think I'm going to leave it well enough alone.

PR #57576, opened 2024-03-27
[Levelbuilder] strip block ids when copying block xml

mikeharv:
The levelbuilder Extra Links interface includes buttons for working directly with Block XML. Curriculum writers use these for multiple purposes, including getting the XML to include in feedback strings used with Sprite Lab validation. 

image

Recently, we had an issue where this block XML included a special character `}` which was breaking translation files:
https://codedotorg.slack.com/archives/C03DBDN67B7/p1711145896149469?thread_ts=1711139550.841369&cid=C03DBDN67B7
```
Error: key 'gameDesignMissingType' in localization file i18n/spritelab/ar_sa.json has a format issue: SyntaxError: Expected "#", "{", doubled apostrophe, end of input, escaped string, or plain char but "}" found.
```
This was the string in question:
```
"gameDesignMissingType": "You need to add another block so the monster can land on the green platforms:\n\n<xml><block xmlns=\"https://developers.google.com/blockly/xml\" type=\"GameDev_setGroup\" id=\"s0~~e0!}K[fm;ShKs2sG\">\n <field name=\"GROUP\">\"walls\"</field>\n <value name=\"SPRITE\">\n <block type=\"gamelab_allSpritesWithAnimation\" id=\"x,sMj}}CfvGtzA+U2wfm\">\n <field name=\"ANIMATION\">\"ground_grass\"</field>\n </block>\n </value>\n</block></xml>"
```

This branch aims to prevent this sort of issue by removing the unneeded block ids as soon as the levelbuilder user copies the block. As @dancodedotorg said in the thread linked above:
> the id tag itself isn't the problem... It's that IDs can contain reserved characters. The one Molly originally highlighted had `}` characters in the ID, which is what the error complained about. If an ID happens to not have any of those characters, we suspect it would still work. But this removes that as a possibility

@dancodedotorg has also created a Chrome extension to help other curriculum writers perform the JSON escaping that is required to get the strings into a publishable state. He has already modified this so that it will now remove block IDs.

Despite this improvement, it makes sense for us to implement a "real" fix that doesn't rely on internal users adding a Chrome extension. 

To fix this, I was able to leverage a recently created util `removeIdsFromBlocks` which was originally created to prevent writing duplicate entries to the level_sources table. Previously, the function assumed we wanted to preserve ids if they were explicitly listed in the level's start blocks or toolbox XML. In this new context, that's not the case so I moved those exempted ids to an optional parameter.

In addition to the change described above, I also made the following improvements while I was in this space:

* Changed the type of `levelBlockIds` from array to set. When levelbuilders save start blocks for a Google Blockly we save all of their randomly generated block ids into the level configuration. This means that a level could have hundreds of (supposedly) "explicitly set" block ids. This doesn't harm anything from the perspective of the original `level_sources` problem, but the performance can be improved this way.
* Updated the **"Copy all blocks"** button to use `getProjectXml` instead of `blockSpaceToDom`. This custom method better guarantees that the levelbuilder gets the XML for _all_ blocks, including those on the new hidden workspace. I also tested the copying/pasting a full workspace worth of blocks succeeds in copying procedure definition blocks as expected.
* Switched to query selector to get all blocks in `removeIdsFromBlocks`. This has the advantage of no longer requiring the function to be recursive.
* Updated the  **"Copy selected block"** to wrap the block element in an `<xml></xml>` container element. This makes it so that the container is consistently present for `removeIdsFromBlocks`'s `element` argument. This allows the single block to be selected by the query selector mentioned above. It should also help curriculum writers who usually have to manually add the container when adding block XML to a markdown field or feedback string. It also allows us to simplify the logic in `pasteBlocksToWorkspace`.

## PR Checklist:

- [ ] Code is well-commented


molly-moen (approved):
Nice!

PR #57577, opened 2024-03-27
Add a rake task to perform local cloudformation linting

cat5inthecradle:
Our `rake stack:validate` task works by creating a change set in cloudformation, which is very effective, but it also takes longer, requires AWS access, and is one `dry_run` flag away from actually executing changes, especially dangerous if you want to validate a stack for production.

This PR adds `rake stack:lint` which does require a local installation of the [cfn-lint](https://github.com/aws-cloudformation/cfn-lint) python module. I created this task because it's very tedious to do this manually by linting some heavily edited output from `VERBOSE=1 rake stack:validate`.

I've also fixed the linter warnings present in our main cloudformation template.

## Testing story

First, revert any of the fixed linting issues that this PR resolves.
Then, execute the following, as I did to test this manually.

```
> bundle exec rake stack:lint
Finished stack:environment (less than a minute)
W3011 Both UpdateReplacePolicy and DeletionPolicy are needed to protect Resources/Daemon from deletion
/tmp/template20240327-2413467-xcyjvf.yml:453:3

WARNING: cfn-lint detected issues in the template.
Finished stack:lint (less than a minute)
```

Fix the issue and run the command again. Lint numbers are 2-3 off of the actual template for some reason, but you can also hunt down the issue by looking for the resource name, in the above example, we're looking for the `Daemon` resource, which is tricky to find because of ERB, so the `VERBOSE` flag might help (see below)

```
> bundle exec rake stack:lint
Finished stack:environment (less than a minute)
SUCCESS: cfn-lint passed successfully. Run `validate` task to further check for CloudFormation syntax errors.
Finished stack:lint (less than a minute)
```

Like `validate` this accepts the `VERBOSE=1` flag, which will dump the full template to the console. Because it's 1600 lines long, you'll have a better time if you send the output to a file or pipe it to `less`

```
> VERBOSE=1 bundle exec rake stack:lint RAILS_ENV=production
Finished stack:environment (less than a minute)
[   1] ---
[   2] AWSTemplateFormatVersion: 2010-09-09
[   3] Description: AWS CloudFormation Template for Code.org application
...
[ 771] Outputs:
[ 772]   DashboardURL:
[ 773]     Value: "https://development-studio.code.org"
[ 774]     Description: Dashboard URL
[ 775]   PegasusURL:
[ 776]     Value: "https://development.code.org"
[ 777]     Description: Pegasus URL
[ 778] # display information about how to ssh to console if this is a single instance adhoc environment

SUCCESS: cfn-lint passed successfully. Run `validate` task to further check for CloudFormation syntax errors.
Finished stack:lint (less than a minute)
```

You can run the linter against different environment configurations

```
bundle exec rake stack:lint RAILS_ENV=production
```

Finally, this is also available for other named stacks.

```
bundle exec rake stack:iam:lint
```

## Security

Running the linter will attempt to load secrets from AWS Secrets Manager. Sharing the rendered template may reveal secrets (this was already the case with verbose validate).

Production secrets are not accessible to most users, so running the linter requires them to supply a placeholder in locals.yml

## PR Checklist:

- [ ] Code is well-commented


cat5inthecradle:
Some of my own nits:

* Not unit tested - this feels pretty safe though, as it's just local static-analysis.
* Requires local software that's not listed in SETUP.md - but perhaps the just-in-time instruction to install it is better for a rarely used task.
* I ran into an issue where I had an old version of cfn-lint. This doesn't protect you from that, and you might get warnings/errors if you use newer cloudformation features.
* The tempfile creation feels clunky, but I think that's the best way to interact with `cfn-lint`

lib/cdo/aws/cloud_formation.rb
```diff
@@ -48,14 +49,43 @@ def initialize(stack:, log: Logger.new(STDOUT), **options)
...
+      if options[:verbose]
+        log.info template.lines.map.with_index(1) {|line, i| format("[%4d] %s", i, line)}.join
+      end
+      Tempfile.create(['template', '.yml']) do |tempfile|
+        tempfile.write(template)
+        tempfile.close
+
+        cfn_lint_installed = system('which cfn-lint > /dev/null 2>&1' || 'where cfn-lint > NUL 2>&1')
+        unless cfn_lint_installed
+          log.info "WARNING: cfn-lint is not installed, skipping lint task."
```
Hamms: Could we add a link to something like https://github.com/aws-cloudformation/cfn-lint?tab=readme-ov-file#install here, just to make this super easy

lib/cdo/aws/cloud_formation.rb
```diff
@@ -48,14 +49,43 @@ def initialize(stack:, log: Logger.new(STDOUT), **options)
...
+      stack.dry_run = true
+      template = stack.render
+      if options[:verbose]
+        log.info template.lines.map.with_index(1) {|line, i| format("[%4d] %s", i, line)}.join
+      end
+      Tempfile.create(['template', '.yml']) do |tempfile|
+        tempfile.write(template)
+        tempfile.close
+
+        cfn_lint_installed = system('which cfn-lint > /dev/null 2>&1' || 'where cfn-lint > NUL 2>&1')
```
Hamms: I'm not familiar with `where` or `NUL`; what development environment is this meant to support?

aws/cloudformation/components/ami.yml.erb
```diff
@@ -112,18 +112,17 @@ frontend_device_name ||= '/dev/sda1'
     LaunchConfiguration: {Ref: 'FrontendLaunchConfig'}
   %>
   Frontends:
-    DependsOn: [ASGCount]
```
Hamms, line 115: Did the linter say why this is not necessary?

Hamms (approved):
Couple of questions, but nothing blocking; this looks great! So happy to see some better dev tooling for this part of the codebase; thank you!

lib/cdo/aws/cloud_formation.rb
cat5inthecradle: Windows, but as I'm looking at this, I'm realizing that it's probably much simpler to just do something like:

`cfn_lint_installed = system('cfn-lint --version > /dev/null 2>&1')`

aws/cloudformation/components/ami.yml.erb
cat5inthecradle: Yes, though I could be convinced to leave it for extra clarity and add an exception to the linter.

The presence of `!GetAtt ASGCount.<property>` later in the resource is enough to define the dependency. Cloudformation builds the dependency tree by looking at `!GetAtt` and `!Ref`, so you only really need `DependsOn` if for some reason that's not sufficient.

lib/cdo/aws/cloud_formation.rb
cat5inthecradle: And if they're using WSL, then maybe we don't need to worry about windows quirks...

cat5inthecradle:
```suggestion
        cfn_lint_installed = system('which cfn-lint > /dev/null 2>&1')
```

Hamms: LGTM!

aws/cloudformation/components/ami.yml.erb
Hamms: Cool, thanks! I think that makes enough sense that an exception isn't necessary

PR #57579, opened 2024-03-27
Add DCDO flag for AI TA launch

kelbyhawn:
Adds the DCDO flag `ai-teaching-assistant-launch` to be used on banners for the AI TA launch in early April.

**Jira ticket:** [ACQ-1671](https://codedotorg.atlassian.net/browse/ACQ-1671)

### DIFF
--- a/lib/dynamic_config/dcdo.rb
+++ b/lib/dynamic_config/dcdo.rb
@@ -45,6 +45,8 @@ class DCDOBase < DynamicConfigBase
       # Whether the scholarship dropdown is locked on the application dashboard.
       'scholarship-dropdown-locked': DCDO.get('scholarship-dropdown-locked', true),
       hoc_mode: DCDO.get('hoc_mode', false),
+      # Whether to show the marketing banners for the AI Teacher Assistant launch. Can be removed later.
+      'ai-teaching-assistant-launch': DCDO.get('ai-teaching-assistant-launch', false),
     }
   end
 end


lib/dynamic_config/dcdo.rb
```diff
@@ -45,6 +45,8 @@ def frontend_config
       # Whether the scholarship dropdown is locked on the application dashboard.
       'scholarship-dropdown-locked': DCDO.get('scholarship-dropdown-locked', true),
       hoc_mode: DCDO.get('hoc_mode', false),
+      # Whether to show the marketing banners for the AI Teacher Assistant launch. Can be removed later.
+      'ai-teaching-assistant-launch': DCDO.get('ai-teaching-assistant-launch', false),
```
bethanyaconnor, line 49: Just checking -- do we need to reference this in JS?

kelbyhawn: I don't think so 🤔 Where were you thinking this would go?

bethanyaconnor: You only need to add DCDO flags to this file if you want to access them in javascript. If you're only accessing them in haml/ruby code, you can just include it in the PR.

bethanyaconnor: (I'm like 99% sure but now I'm wondering if I missed a memo lol)

kelbyhawn: Oh! I've always added them here for use across all sites. I wanted to get this in now since I'll need to use it on a couple separate PRs.

bethanyaconnor: I don't think it's needed if you're only using it from the server-side, but it's safe to add here either way. Up to you :)

bethanyaconnor (approved):

PR #57580, opened 2024-03-27
Create a wrapper around the Mailjet API

bethanyaconnor:
This PR creates a wrapper around the `mailjet` SDK. It also creates the method that will send the welcome email and create a contact.

I decided to include a constants file that lists the template ids (ids from the mailjet platform), as well as some general metadata about each email (such as who it is from). I figured this would be extensible if we ever moved workshop emails to this platform as well, which don't come from Hadi. In all likelihood, I'll update these constants again before full launch.

See the [tech plan](https://docs.google.com/document/d/1gLY6PezbIAV6Rm4WwRb_4Ig6I762iq8liM0sUKTE1ZI/edit) for details.


lib/cdo/mailjet.rb
```diff
@@ -0,0 +1,110 @@
...
+require_relative './shared_constants/mailjet_constants'
+
+module MailJet
+  include MailJetConstants
+
+  API_KEY = CDO.try(:mailjet_api_key).freeze
+  SECRET_KEY = CDO.try(:mailjet_secret_key).freeze
+
+  def self.enabled?
+    return false unless DCDO.get('use_mailjet', false)
```
hannahbergam: I always have a hard time reading 'unless', but I'm surprised ust_mailjet would be false here

lib/cdo/mailjet.rb
```diff
@@ -0,0 +1,110 @@
...
+      'staging'
+    else
+      'development'
+    end
+  end
+
+  def self.create_contact_and_send_welcome_email(user)
+    return unless enabled?
+
+    return if user&.id.nil?
```
hannahbergam: Can we switch this one to an 'unless' to match the other two for readability?

lib/cdo/mailjet.rb
```diff
@@ -0,0 +1,110 @@
...
+    return unless user.teacher?
+
+    create_contact(user.email, user.name, user.created_at.to_datetime.rfc3339)
```
hannahbergam: Can there be a comment describing rfc3339?

lib/cdo/mailjet.rb
```diff
@@ -0,0 +1,110 @@
...
+
+    Mailjet.configure do |config|
+      config.api_key = API_KEY
+      config.secret_key = SECRET_KEY
+      config.api_version = "v3"
+    end
+
+    # If a contact already exists, Mailjet will raise an error.
+    # This shouldn't happen as user emails should be unique, but
+    # we don't want to block sign up if it does.
```
hannahbergam: 🎉

lib/cdo/mailjet.rb
```diff
@@ -0,0 +1,110 @@
...
+        is_excluded_from_campaigns: true,
+        email: email,
+        name: name
+      )
+    rescue => exception
+      Honeybadger.notify(exception)
+    end
+
+    # Most likely, the above would fail if a contact already exists.
+    # In that case, we want to update the contact with the sign up date.
```
hannahbergam: Is this in the case that (for example), someone would opt in to receive the 'host an hour of code' email but not yet have an account?

hannahbergam (commented):
I left a few questions- thanks for adding the tests!

lib/cdo/mailjet.rb
cat5inthecradle: This feels super unintuitive and requires a few mental jumps. Why can't we just use the much simpler form?

```suggestion
    return DCDO.get('use_mailjet', false)
```

lib/cdo/mailjet.rb
cat5inthecradle: I'm guessing we're doing this because [Mailjet supports rfc3339](https://documentation.mailjet.com/hc/en-us/articles/360042706774-Which-date-formats-are-supported-by-Mailjet) formatted dates (iso-8601 would probably be fine too, but they specifically name rfc3339)

I'd prefer we shove the mailjet-specific formatting down into the `create_contact` function, then it's more clear that we're doing this explicitly for Mailjet.

lib/cdo/mailjet.rb
```diff
@@ -0,0 +1,110 @@
...
+          {
+            name: 'sign_up_date',
+            value: sign_up_date
+          }
+        ]
+      )
+    rescue => exception
+      Honeybadger.notify(exception)
+    end
+  end
```
cat5inthecradle: Right now this function always makes three API calls. Can we reduce that to one or two calls by rearranging it?


* FIND contact by email
* if found, UPDATE the sign_up_date _if necessary_
* else CREATE with sign_up_date attributes

I think that would also make it more straightforward to add test cases for the various code paths.

lib/cdo/mailjet.rb
```diff
@@ -0,0 +1,110 @@
...
+          Email: from_email,
+          Name: from_name
+        },
+        To: [
+          {
+            Email: to_email,
+            Name: to_name
+          }
+        ],
+        TemplateID: template_id,
```
cat5inthecradle, line 106: Would there be value in enforcing that the template_id be found in `MailJetConstants`?

lib/cdo/mailjet.rb
cat5inthecradle: Additionally, this function is squashing _all_ errors. That seems undesireable and I can think of a few scenarios where you definitely don't want to be calling `send_template_email` if `create_contact` fails.

cat5inthecradle (changes requested):
Got some nits about the implementation and always like to see more exhaustive tests around non-happy paths.

At a high level I think that the API you're exposing from Cdo::MailJet makes sense, and gives us a good place to wire in error handling, metrics, etc

lib/cdo/mailjet.rb
bethanyaconnor: Combined this line with the one below

bethanyaconnor: And used `present?` over negating `nil?`

lib/cdo/mailjet.rb
bethanyaconnor: oo true, I always forget about how many ways we get user emails. That seems like a likely pathway!

lib/cdo/mailjet.rb
bethanyaconnor: > CREATE with sign_up_date attributes

This always requires two API calls because properties have to be updated through a [separate API](https://dev.mailjet.com/email/reference/contacts/contact-properties/).

I think you've got a good point about finding the contact first then branching though -- I'll update.

cat5inthecradle: Ah, that's frustrating. In that case perhaps the create doing double duty as an existence check is fine.

lib/cdo/mailjet.rb
```diff
@@ -0,0 +1,110 @@
...
+  def self.subaccount
+    case Rails.env
+    when 'production'
+      'production'
+    when 'staging'
+      'staging'
+    else
+      'development'
+    end
+  end
```
cat5inthecradle: I was re-reading the tech plan and was reminded that we're explicitly planning on NOT connecting to mailjet for Test and Drone. Is there a good place to make that intention clear in the code?

lib/cdo/mailjet.rb
bethanyaconnor: Refactored and it's much cleaner! I ended up removing the error squashing in general, but I'm very open to discussion there. In particular, moving the call to find the contact up means that the primary error case disappeared, so I'm less attached to squashing the errors. There's a part of me that's considering adding them back, however, in case MailJet is down or something. Thoughts?

lib/cdo/mailjet.rb
bethanyaconnor: @cat5inthecradle, is the change in ade564409238746dbe46c9a771b946eae818c0cc what you were thinking?

lib/cdo/mailjet.rb
bethanyaconnor: Yeah -- I could add a check in `enabled?`. Let me do that.

lib/cdo/mailjet.rb
bethanyaconnor: Do you mean to push the fetching of the template ID down to this function?

cat5inthecradle: I was just noticing that there's nothing to force future developers to use to use the template listing instead of hardcoding template ID's elsewhere in the codebase. Once we have examples of using the wrapper, the convention might be more clear though.

cat5inthecradle (approved):

lib/cdo/mailjet.rb
cat5inthecradle: oh, sorry, I meant clear to future developers wondering why production and staging are getting values but not anything else (because we don't have a standard convention or clear guidance about the different environments)

lib/cdo/mailjet.rb
cat5inthecradle: Yep!

lib/cdo/mailjet.rb
cat5inthecradle: Depends if we're going to do anything about it. For a general wrapper, I think it makes sense to raise an exception if `create_contact` fails to create a contact. Catching that exception is the responsibility of the caller, and it can decide whether to retry, fail, or move on. Unless there's a specific exception category that this wrapper should be responsible for resolving (for example, I've worked with API's that have rate limits that were resolved by trying again one more time right now)

lib/cdo/mailjet.rb
bethanyaconnor: oh gotcha! Yeah I can add something here as well

lib/cdo/mailjet.rb
bethanyaconnor: Makes sense! And thinking through it, I don't want to add any exception handling here. When I hook this up into the actual user sign up flow, I might, but that is coming in a future PR.

PR #57586, opened 2024-03-27
Improve Examples for Publishing Secrets

sureshc:
We've had two recent cases where engineers published Secrets that were inadvertently populated with JSON/complex values. The engineer intended to publish AWS Secrets Manager Secret `production/cdo/foo` with value `bar` and instead populated it with the value `{"foo": "bar"}` due to confusion about how to format the `config/secrets.yml` file.

Update the example with a better illustration of how to publish Secrets with values that are either simple or complex .

## Testing story

`bin/update_secrets` using a `secrets.yml` populated with the examples from `secrets.yml.template`

image

image

## Follow-up work

Delete all the example Secrets that were published during the manual test of this new template.

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/config/secrets.yml.template
+++ b/config/secrets.yml.template
@@ -1,18 +1,20 @@
 # Sample secrets.yml file.
-# Top-level key is secret name, second-level key is environment name.
-# (`base` is an alias for staging, test, levelbuilder, and production.)
-# The value is the secret string.
-# (Complex objects will be converted to JSON.)
 ---
-foo:
-  base: bar # staging, test, levelbuilder and production
-  production: baz
-qux:
-  development:
-    qux: quux
-  staging:
-    qux: quuz
-  test:
-    qux: quuz
-  production:
-    qux: quuz
+# Publish AWS Secrets Manager Secrets named `[environment type]/cdo/acme_api_key` that can be used to populate `CDO.acme_api_key`
+acme_api_key:
+  development: 123ABC      # Development API key
+  adhoc: 123ABC            # Development API key
+  staging: 123ABC          # Development API key
+  test: ZYX987             # Test API key
+  levelbuilder: ZYX987     # Test API key
+  production: a1b2c3d4e5f6 # Production API key
+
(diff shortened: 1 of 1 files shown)


config/secrets.yml.template
```diff
@@ -1,18 +1,20 @@
 # Sample secrets.yml file.
```
bethanyaconnor, line 1: I'm wondering if it's worth including a note that, if an environment doesn't have a secret value, to just delete that line. But that's definitely a nit not a blocker!

bethanyaconnor (approved):
Thank you for doing this!

Hamms (approved):
I'm not sufficiently familiar with this functionality to review the examples for correctness, but from a clarity perspective they look good to me!

config/secrets.yml.template
```diff
@@ -1,18 +1,20 @@
...
+acme_api_key:
+  development: 123ABC      # Development API key
+  adhoc: 123ABC            # Development API key
+  staging: 123ABC          # Development API key
+  test: ZYX987             # Test API key
+  levelbuilder: ZYX987     # Test API key
+  production: a1b2c3d4e5f6 # Production API key
+
+# Publish AWS Secrets Manager Secrets named `[environment type]/cdo/wile_e_coyote_credentials` that can be used to
+# populate `CDO.wile_e_coyote_credentials` with a value whose data type is complex.
```
hannahbergam, line 13: I know this is already merged, but in case there is a chance for a followup: The issue we ran into with Statsig seemed to be that we weren't sure how to verify the AWS secrets looked 'right'. Could that be in a comment somewhere here? i.e. something like 'Once these secrets appear on AWS secret manager, you can verify the formatting of the secret 'value' is the plaintext string, not a nested object'

cat5inthecradle: I think that belongs in secrets.md, but perhaps something like this?

```
Cdo::Secrets.new.get!('development/cdo/your_new_secret')
```

PR #57599, opened 2024-03-27
Lab2: Allow lab2 levels to navigate to non-lab2 levels

sanchitmalhotra126:
This mostly worked already; the main change we need here is that the `navigateToNextLevel` redux action assumes that we are navigating to a lab2 level and therefore tries to update the browser stack. Instead, we just need to check if the current and next level are lab2, and either update the browser stack, or do a full page load accordingly.

https://github.com/code-dot-org/code-dot-org/assets/85528507/92b8dfb6-e1c3-4607-83f1-5e1fc7cb9b28

## Links

Motivated by wanting to use the Panels level in non-lab2 progressions: https://codedotorg.slack.com/archives/CFGAVL2CA/p1711569834707299

## Testing story

Tested locally on allthethings with a progression containing both lab2 and non-lab2 levels.

### DIFF
--- a/apps/src/code-studio/progressRedux.ts
+++ b/apps/src/code-studio/progressRedux.ts
@@ -22,14 +22,20 @@ import {
 import {
   processServerStudentProgress,
   getLevelResult,
+  processedLevel,
 } from '@cdo/apps/templates/progress/progressHelpers';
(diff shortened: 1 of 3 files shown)


breville (approved):
Nice!

PR #57602, opened 2024-03-27
Support: `HOT=0 yarn start` to disable HMR 

bencodeorg:
An explicit HOT=0 env variable will disable hot module reloading.

- `webpack-dev-server` is still enabled, and used to serve files via proxy AND writeToDisk: to update apps/build
- but the compiled JS will not contain hot reload hooks, so it shouldn't attempt to access port 9000

### DIFF
--- a/apps/envConstants.js
+++ b/apps/envConstants.js
@@ -38,7 +38,7 @@ module.exports = {
   DRONE: process.env.DRONE,
   CIRCLE_TEST_REPORTS: process.env.CIRCLE_TEST_REPORTS,
   // If set, run webpack dev server (with hot reloading)
-  HOT: false,
+  HOT: !(process.env.HOT === '0'),
   // Include static assets when serving storybook locally
   STORYBOOK_STATIC_ASSETS: process.env.STORYBOOK_STATIC_ASSETS,
 };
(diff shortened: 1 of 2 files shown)


snickell (approved):
LGTM

apps/envConstants.js
```diff
@@ -38,7 +38,7 @@ module.exports = {
   DRONE: process.env.DRONE,
   CIRCLE_TEST_REPORTS: process.env.CIRCLE_TEST_REPORTS,
   // If set, run webpack dev server (with hot reloading)
-  HOT: !!process.env.HOT,
+  HOT: process.env.HOT === '1',
```
bencodeorg, line 41: Gotta run home right now, but will this be false in our Gruntfile dev setup because we set process.env.HOT to 1 (number)? 🤔 Can check a bit later

snickell: process.env.* is always a string, just verified:
```
> process.env.boo
undefined
> process.env.boo = true
true
> process.env.boo
'true'
> 
```

bencodeorg:
I can't approve as author, but LGTM if you want to merge before tomorrow :)

PR #57604, opened 2024-03-28
Zip Search Populating School Dropdown List

hannahbergam:
This PR utilizes the existing zip search to pull the schools by zip code into the dropdown with the other three 'default' options: Select a School, I don't teach at a school, and enter manually.

I elected to reuse the existing search but was able to pare down the search *call significantly because we are removing location search functionality. This way we are able to minimize the risk (and time) involved with writing a new search function, while streamlining performance. 

This exists currently in debouncedSearch from [SchoolAutocompleteDropdown here.](https://github.com/code-dot-org/code-dot-org/blob/staging/apps/src/templates/SchoolAutocompleteDropdown.jsx#L49)

The back end search is fairly involved and complex, and I am assuming most of these decisions were made for a clear (though perhaps not quite well documented) reason, and have elected to advise a full rewrite as a future performance enhancement rather than a P0 requirement here. [Here is the Jira ticket](https://codedotorg.atlassian.net/browse/ACQ-1717) for that followup work. 

I also added a sort functionality, an improvement over the existing search. This unfortunately highlights our duplicate schools issue, which should have a resolution soon from our RED team.

Screenshot 2024-04-03 at 1 47 58 PM

## Links

- spec: []()
- jira ticket: https://codedotorg.atlassian.net/browse/ACQ-1611

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/templates/SchoolZipSearch.jsx
+++ b/apps/src/templates/SchoolZipSearch.jsx
@@ -1,4 +1,4 @@
-import React, {useState} from 'react';
+import React, {useState, useEffect} from 'react';
 import PropTypes from 'prop-types';
 import i18n from '@cdo/locale';
 import {BodyTwoText} from '@cdo/apps/componentLibrary/typography';
@@ -18,6 +18,16 @@ export default function SchoolZipSearch({fieldNames, zip}) {
   const [inputManually, setInputManually] = useState(false);
   const [dropdownSchools, setDropdownSchools] = useState([]);
 
+  useEffect(() => {
+    const searchUrl = `/dashboardapi/v1/schoolsearch/${zip}/40`;
+    fetch(searchUrl, {headers: {'X-Requested-With': 'XMLHttpRequest'}})
+      .then(response => (response.ok ? response.json() : []))
+      .then(json => {
+        const schools = json.map(school => constructSchoolOption(school));
+        setDropdownSchools(schools);
+      });
+  }, [zip]);
+
   const onSchoolChange = e => {
     const schoolId = e.target.value;
     setSelectedSchoolNcesId(schoolId);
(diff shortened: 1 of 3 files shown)


TurnerRiley (approved):
Looks great! Out of curiosity, do you think it would be worth making a ticket to add a Storybook or test file for SchoolZipSearch just to have in general to check that its returning the right value and that it handles having a combo of hardset values (e.g. "clickToAdd") and queried school values well?

apps/src/templates/SchoolZipSearch.jsx
```diff
@@ -53,6 +69,7 @@ export default function SchoolZipSearch({fieldNames}) {
             styleAsText={true}
             onClick={() => {
               setInputManually(false);
+              setSelectedSchoolNcesId('selectASchool');
```
bethanyaconnor: nit: could this be moved into its own constant, used both here and in `SEARCH_DEFAULTS`?

bethanyaconnor (commented):
Are there any tests to add here?

hannahbergam:
> Are there any tests to add here?

@bethanyaconnor and @TurnerRiley I added a sort functionality, updated the description, and added a test with a fetch stub that verifies the sort is working! (Also there's a new screenshot to show that off)

apps/test/unit/templates/SchoolZipSearchTest.jsx
```diff
@@ -0,0 +1,79 @@
...
+  if (a === null || b === null) return false;
+  if (a.length !== b.length) return false;
+  for (var i = 0; i < a.length; ++i) {
+    if (a[i] !== b[i]) return false;
+  }
+  return true;
+};
+
+describe('SchoolDataInputs', () => {
+  allowConsoleWarnings();
```
bethanyaconnor: curious why this is needed?

bethanyaconnor (approved):
Test looks good! One question but otherwise lgtm!

PR #57621, opened 2024-03-28
Fix Violations of and Reenable `Style/AccessModifierDeclarations`

Hamms:
> Access modifiers should be declared to apply to a group of methods or inline before each method, depending on configuration. EnforcedStyle config covers only method definitions. Applications of visibility methods to symbols can be controlled using AllowModifiersOnSymbols config.

Fix applied automatically with `bundle exec rubocop --autocorrect-all --only Style/AccessModifierDeclarations`. I then automatically fixed up a few resulting excess newlines with `bundle exec rubocop --autocorrect-all --only Layout/EmptyLines`

This is sort of a weird one. There are two ways to declare methods as `private` in Ruby; either in a group like:

```ruby
class Foo

  private

  def bar; end
  def baz; end

end
```

or inline like:

```ruby
class Foo

  private def bar; end
  private def baz; end

end
```

This Cop attempts to enforce consistency; in our case, consistently using the inline rather than the group approach. I personally prefer inline; it's slightly more verbose, but eliminates the potential confusion of looking at methods declared at the bottom of a file and not realizing that they're all private because the single declaration of that is offscreen.

What do folks think? Would anyone prefer to enforce the group style, or not enforce anything at all?

## Links

- https://docs.rubocop.org/rubocop/cops_style.html#styleaccessmodifierdeclarations
- https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/AccessModifierDeclarations

## Testing story

## PR Checklist:

- [ ] Code is well-commented


cnbrenci (approved):
I agree on preference for inline declaration and prefer to enforce it for consistency.

PR #57622, opened 2024-03-28
Removed Not Started Icon

kobryan0619:
Removed the "Not Started" icon and reordered the icons as per Bryan's feedback.

image

NOTE: We do have a `NOT_STARTED` type in the `ItemType` [file](https://github.com/code-dot-org/code-dot-org/blob/32c87af11266a304008bb50b3adcaa6f83ab09dd/apps/src/templates/sectionProgressV2/ItemType.jsx#L6) as well as inside the `ProgressIcon.jsx` file.  I am reluctant to remove anything from those files now just in case this gets added back in later.  Open to thoughts though...

## Links

[Ticket](https://codedotorg.atlassian.net/jira/software/projects/TEACH/boards/65?selectedIssue=TEACH-969)

## Testing story

Updated existing tests.

### DIFF
--- a/apps/src/templates/sectionProgressV2/AssignmentCompletionStatesBox.jsx
+++ b/apps/src/templates/sectionProgressV2/AssignmentCompletionStatesBox.jsx
@@ -16,16 +16,6 @@ export default function AssignmentCompletionStatesBox({hasValidatedLevels}) {
       <div className={styles.icons}>
         <div className={styles.legendColumn}>
           <LegendItem
-            itemType={ITEM_TYPE.NOT_STARTED}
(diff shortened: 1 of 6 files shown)


lfryemason (commented):
I would prefer to remove all references and the object itself. It's easy enough to add back in (we could even revert this PR), but if we never add it back, then this is just unused code sitting in the repo forever.

Otherwise, looks great!

lfryemason:
You could add a comment saying 'Not Started removed in PR#XXXXX'

dju90:
LGTM! 🎉

lfryemason (approved):
Thanks Kaitie!

PR #57625, opened 2024-03-28
[Google Blockly] Remove ids from all variable fields before adding to flyout

mikeharv:
Report from @onlinecsteacher:
>I’m noticing a bug with trying to save a variable name into a conditional in a Sprite Lab [level](https://levelbuilder-studio.code.org/s/pilot-k5-game-design/lessons/4/levels/3). I want the conditional to have the variable score and in fact it shows up correctly in the edit page but when I click Save, the name reverts to the default i . I’m including the console message if helpful.

<table><tr><td>

Screenshot 2024-03-27 at 4 30 01 PM

</td><td>

Screenshot 2024-03-27 at 3 28 37 PM
</td></tr></table>

Relates to this Sprite Lab migration task:
- https://github.com/code-dot-org/code-dot-org/pull/54302

If a level uses an auto-populated Variables toolbox category, we need to merge the dynamic blocks with any static blocks coming from the XML.  Blockly supports deserializing variables from block XML that give either a name or an id, both of which are found in the workspace's variable map. However, if the flyout block's variable id doesn't match the one in the variable map, Blockly assumes it's a mistake an assigns it a new name (e.g. `i`).  

Today we account for this by attempting to replace serialized ids with names, which is safer for our purposes. I mistakenly had assumed that our convention of using `"VAR"` and `"VARIABLE"` for the names of the fields was more consistent. Unfortunately, some blocks, like `ifVarEquals` use completely different names. In fact, field names have proven to be an unreliable source of truth for this sort of thing before.... See: https://github.com/code-dot-org/code-dot-org/pull/57337

This branch proposes a refactor and change to that logic that fixes the problem. We'll use the variables serialization to create a map of ids to names. For any block in the Variables category, we'll look for fields that have an id value that is represented in the map. When we find out, we'll replace the id with the name. Now when the flyout is populated, the blocks all have the expected variable names:

image

## Links

Jira: https://codedotorg.atlassian.net/browse/CT-438
Slack Thread: https://codedotorg.slack.com/archives/C05DK21DAHK/p1711578681366349

## PR Checklist:

- [ ] Code is well-commented


apps/src/blockly/addons/cdoUtils.ts
```diff
@@ -503,63 +508,60 @@ export function getLevelToolboxBlocks(customCategory: string) {
...
+function updateVariableFields(
+  block: {fields: SerializedFields},
+  serializedVariableMap: Map<string, string>
+): void {
+  const fields = block.fields;
+  for (const key in fields) {
+    const field = fields[key];
+    if (field.id && serializedVariableMap.has(field.id)) {
+      field.name = serializedVariableMap.get(field.id)!;
+      delete field.id;
```
ebeastlake, line 543: Do we need to delete them? Can you think of any scenarios in which we'd need this on a subsequent run or something?

ebeastlake (approved):
General approach looks good to me pending a clean Drone run!

apps/src/blockly/addons/cdoUtils.ts
mikeharv: We need to delete them because they if Blockly can't find a variable in the workspace with the same id, it will create a new one with an undesired name (`i`). It should be safe since we don't reuse the serialization AFAIK. To be on the safe side I just pushed a change that will create a copy of the blocks object so the original doesn't get modified at all, just in case.

PR #57627, opened 2024-03-28
Fix `test_secrets_config` Test

Hamms:
Follow-up to https://github.com/code-dot-org/code-dot-org/pull/57586, which significantly improved the quality of the examples in that file. Unfortunately and unexpectedly, it turns out that file is also consumed by a test:

https://github.com/code-dot-org/code-dot-org/blob/660f150dc904cd2f77a024cde556a106140e37ab/lib/test/cdo/test_secrets_config.rb#L20-L24

Even more unfortunately, our "test only changed files" logic does not recognize that changes to the config files should result in `lib/` tests getting rerun. In addition to updating the test to reflect the updates to that file, this PR also updates our "test changed" logic so that we run lib tests when config files are changed.

## Testing story

First I pushed a commit that updated the test config, so we can verify that Drone now catches this test failure. Then I pushed a commit which updates the test so we can verify that the test now passes.

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/config/secrets.yml.template
+++ b/config/secrets.yml.template
@@ -1,4 +1,4 @@
-# Sample secrets.yml file.
+# Sample secrets.yml file; consumed by test_secrets_config.rb
 ---
 # Publish AWS Secrets Manager Secrets named `[environment type]/cdo/acme_api_key` that can be used to populate `CDO.acme_api_key`
 acme_api_key:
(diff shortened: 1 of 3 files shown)


bencodeorg (approved):

sureshc (approved):
Thank you for cleaning up the mess I made!

PR #57629, opened 2024-03-28
[ProgressV2] change width of lesson cells to squares

lfryemason:
image

## Links
[TEACH-967](https://codedotorg.atlassian.net/browse/TEACH-967)

## PR Checklist:

- [ ] Code is well-commented


etaderhold (approved):
I do think the margins on expandable two-digit lessons can look a bit tight (especially on "20" with two wider digits), and so does the gap between the small lock and the dot on locked lessons with progress, but if that's the trade-off we want to make so we squeeze more info on the screen so be it.

image

PR #57631, opened 2024-03-28
P20-722: Validate NRPS Full response on course sync

artem-vavilov:
## Summary
Validates the NRPS Full response and displays errors if missing custom keys `section_id` and `section_names` within LTI key configurations are detected during course sync.

## Screenshot
Screenshot 2024-03-28 at 22 26 35

## Overview
lti-sync-errors

## Links
- JIRA task: [P20-722](https://codedotorg.atlassian.net/browse/P20-722)

### DIFF
--- a/dashboard/config/locales/en.yml
+++ b/dashboard/config/locales/en.yml
@@ -1422,7 +1422,7 @@ en:
       missing_params: "Missing required param(s): School/District name, Client ID, LMS, Email"
       unsupported_lms_type: "Unsupported LMS platform type"
       wrong_resource_type: "Only LtiResourceLink is supported right now"
-      missing_key_config: 'Your LTI key configuration is missing the required "%{field}" field'
+      missing_tool_config: 'Your LTI Tool configuration is missing the required "%{field}" field'
     iframe_button: "Open in New Tab"
     iframe_message: "Code.org cannot be run in an embedded window. Please open it in a new tab."
     iframe_title: "Opening Code.org in New Tab"
--- a/dashboard/config/locales/en.yml
+++ b/dashboard/config/locales/en.yml
@@ -1422,6 +1422,7 @@ en:
       missing_params: "Missing required param(s): School/District name, Client ID, LMS, Email"
       unsupported_lms_type: "Unsupported LMS platform type"
       wrong_resource_type: "Only LtiResourceLink is supported right now"
+      missing_key_config: 'Your LTI key configuration is missing the required "%{field}" field'
     iframe_button: "Open in New Tab"
     iframe_message: "Code.org cannot be run in an embedded window. Please open it in a new tab."
     iframe_title: "Opening Code.org in New Tab"
--- a/dashboard/lib/services/base.rb
+++ b/dashboard/lib/services/base.rb
@@ -1,0 +1,11 @@
+module Services
+  class Base
+    def self.call(...)
+      new(...).call
+    end
+
+    def call
+      raise NotImplementedError
+    end
+  end
+end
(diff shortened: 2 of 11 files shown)


dashboard/lib/services/base.rb
mgc1194, line 1: YES!! THanks for implementing this!

dashboard/lib/services/lti/nrps_response_validator.rb
```diff
@@ -0,0 +1,52 @@
+module Services
+  module Lti
+    class NRPSResponseValidator < Services::Base
+      REQUIRED_CUSTOM_FIELDS = %i[section_ids section_names].freeze
```
mgc1194, line 4: I wonder if we should also a¡validate the `context_id` the same way you have done it here.

artem-vavilov: The task states that we need to handle only these two fields

dashboard/config/locales/en.yml
```diff
@@ -1422,6 +1422,7 @@ en:
       missing_params: "Missing required param(s): School/District name, Client ID, LMS, Email"
       unsupported_lms_type: "Unsupported LMS platform type"
       wrong_resource_type: "Only LtiResourceLink is supported right now"
+      missing_key_config: 'Your LTI key configuration is missing the required "%{field}" field'
```
nicklathe: I think we need a more specific error message here, as well as a link to the documentation that specifies what custom variables they need to define in their LTI Tool configuration. Something like:

```suggestion
      missing_key_config: 'Your LTI Tool configuration is missing the required custom field: "%{field}"'
```

Then in a paragraph below, we could add something like `Please reference the required custom fields in the LTI Onboarding Docs {url}`

nicklathe: The wording `custom fields` is what Canvas calls these. In Schoology, they are called `custom paramters`. However, since these variables are currently only required in Canvas, I think it's safe to use the Canvas name, i.e. `custom fields`

carl-codeorg: Yeah I think I like Nick's wording here. Since the new validator specifically validates custom fields within the NRPS response, we can be specific in our language. Later if we expand to do broader validation on standard fields in NRPS, or other params like what Mario suggested, it would make sense to also have generic language.

dashboard/lib/services/lti/nrps_response_validator.rb
carl-codeorg: This PR is specifically for NRPS, but that doesn't mean we can't open a ticket later for an ID token validator.

dashboard/config/locales/en.yml
artem-vavilov: How about this approach? In this case, `custom_fields[section_ids]` is an i18n variable, allowing us to reuse the same message for every possible field:
`Your LTI Tool configuration is missing the required "custom_fields[section_ids]" field`

carl-codeorg: That works too. IMO we just need to be informative enough to point them in the direction of LTI custom fields

nicklathe: Yeah, to Carl's point, we need them to understand that they need to go into their Canvas LTI Tool integration section, and specifically add the custom fields we are requiring. The way its currently worded even took me a minute to figure out what `LTI Key` we were talking about. So as long as we are specifically saying "custom fields" I'm good with it.

carl-codeorg (approved):
:shipit:

nicklathe (approved):

PR #57634, opened 2024-03-28
Fix UI status for pending requests to evaluate entire class

davidsbailey:
Finishes https://codedotorg.atlassian.net/browse/AITT-539. There were problems with how we calculated attempted_unevaluated_count, and also pending_count needed to be added in order to know when to show the pending state. The main changes in this PR are in these commits:
* 09f8aa6ae92d71cb26e3c0a8adbf1566763a97b9 fix attempted_unevaluated_count and add pending_count (on server)
* f719442ac42ff05d08f9b667d07d4ba92f169c4a fix logic for pending evals in status-all response handler (on client)
* re-enable UI test

the other commits are cleanup of various kinds.

## Testing story
* re-enabled UI test
* also ran the UI test locally with this line commented out, to make timing more closely resemble DTT: https://github.com/code-dot-org/code-dot-org/blob/e55a85df86803935cdb83a90969bdf8f085bf938/dashboard/config/environments/development.rb#L81-L85
* updated + added unit tests in js and ruby

### DIFF
--- a/apps/src/templates/rubrics/RubricSettings.jsx
+++ b/apps/src/templates/rubrics/RubricSettings.jsx
@@ -110,7 +110,6 @@ export default function RubricSettings({
   // load initial ai evaluation status
   useEffect(() => {
     if (!!rubricId && !!sectionId) {
-      console.log('initial fetchAiEvaluationStatusAll');
       fetchAiEvaluationStatusAll(rubricId, sectionId).then(response => {
         if (!response.ok) {
           setStatusAll(STATUS_ALL.ERROR);
@@ -180,7 +179,6 @@ export default function RubricSettings({
     if (polling && !!rubricId && !!sectionId) {
       const intervalId = setInterval(() => {
         refreshAiEvaluations();
-        console.log('updated fetchAiEvaluationStatusAll');
         fetchAiEvaluationStatusAll(rubricId, sectionId).then(response => {
           if (!response.ok) {
             setStatusAll(STATUS_ALL.ERROR);
--- a/apps/src/templates/rubrics/RubricSettings.jsx
+++ b/apps/src/templates/rubrics/RubricSettings.jsx
@@ -130,6 +130,7 @@ export default function RubricSettings({
     setDisplayDetails(!displayDetails);
   };
 
+  // load initial ai evaluation status
   useEffect(() => {
     if (!!rubricId && !!sectionId) {
       fetchAiEvaluationStatusAll(rubricId, sectionId).then(response => {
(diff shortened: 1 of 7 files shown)


dashboard/app/models/rubric_ai_evaluation.rb
```diff
@@ -25,7 +25,7 @@ class RubricAiEvaluation < ApplicationRecord
   belongs_to :requester, class_name: 'User'
   belongs_to :rubric
 
-  has_many :learning_goal_ai_evaluations, inverse_of: :rubric_ai_evaluation
+  has_many :learning_goal_ai_evaluations, inverse_of: :rubric_ai_evaluation, dependent: :destroy
```
cearachew, line 28: What does this do here?

apps/src/templates/rubrics/RubricSettings.jsx
```diff
@@ -208,13 +187,13 @@ export default function RubricSettings({
               // we can't fetch the csrf token from the DOM because CSRF protection
               // is disabled on script level pages.
               setCsrfToken(data.csrfToken);
-              setUnevaluatedCount(data.attemptedUnevaluatedCount);
+              setEvaluatedCount(data.lastAttemptEvaluatedCount);
```
cearachew, line 190: Why is this changed here, but not in the initial useEffect? Is EvaluatedCount used anywhere? UnevaluatedCount is used for the ready status string, is it fine for that to not be updated?

cearachew: Ok, I see where evaluatedCount is used (for the summary string), but I think that unevaluatedCount also needs to be updated here?

dashboard/app/models/rubric_ai_evaluation.rb
davidsbailey: oops, I should definitely have explained this! this makes it so that, during local development, when you do `RubricAiEvaluation.find(...).destroy`, instead of getting a foreign key error and having to go delete the corresponding `learning_goal_ai_evaluations` manually, it will delete them for you. I ended up doing this a fair amount during development and thought it would be worth keeping this change in for future use. FYI @Nokondi 

we do not appear to have any codepaths for deleting RubricAiEvaluation objects in our product, so this should have no effect for end users.

apps/src/templates/rubrics/RubricSettings.jsx
davidsbailey: what I am seeing is that unevaluatedCount is shown initially, but that once the Run button is clicked, unevaluatedCount is not displayed again (the `1` in this video):

https://github.com/code-dot-org/code-dot-org/assets/8001765/a2be3af6-ad48-4cb9-b5ef-57986dc694f9

so my thinking was that if it did not affect what we show to the user (or send to the server), then it wasn't worth updating. curious how this reasoning sits with you?

cearachew: Ok, I thought there were cases where it could be updated on polling, but it looks like polling is done only after the click. Looks good to me!

cearachew (approved):

PR #57643, opened 2024-03-29
Fix URL redirects from cached levels not working (Via Liam)

wilkie:
**Problem**: This addresses the attached tickets which were reported from some teachers where the shared links to projects and such were not working as expected. What was happening is that certain levels cause an authentication check. This has them sign in, but the post-sign-in redirect did not respect the desired redirect for that level.

**Solution**: Thanks to Liam, we have a simple solution that 'remembers' the redirect via a session parameter. This is the conventional way of having the 'post signin' redirect happen, but it does not happen on `authenticate_user!` is some situations, apparently. Here, it carefully redirects to a route that will sign in the user with this session parameter being used.

Sneaking in a refactor to a test step and a fix for a potentially flaky test that was causing me grief.

## Links

- jira ticket: [TEACH-758](https://codedotorg.atlassian.net/browse/TEACH-758)

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/dashboard/app/controllers/api/v1/users_controller.rb
+++ b/dashboard/app/controllers/api/v1/users_controller.rb
@@ -83,11 +83,14 @@ class Api::V1::UsersController < Api::V1::JSONApiController
   # cached pages do not have a valid user auth token and therefore will attempt to
   # redirect to the sign in page instead of the correct page if the user is signed in.
   # See https://codedotorg.atlassian.net/browse/TEACH-758 for more details.
-  # NOTE: the `user_return_to` path must not include a redirect back to this path
-  # or there will be an infinite redirect loop. e.g. strip out `login_required` URL
-  # parameters before calling this method.
   # GET /api/v1/users/cached_page_auth_redirect
   def cached_page_auth_redirect
+    # We must ensure that we remove any redirections to this page or we would enter
+    # an infinite loop.
+    [params, session].each do |context|
+      context[:user_return_to] = nil if context[:user_return_to]&.include?('cached_page_auth_redirect')
+    end
+
     if user_signed_in?
       redirect_to params[:user_return_to] || home_url
     else
--- a/dashboard/app/controllers/api/v1/users_controller.rb
+++ b/dashboard/app/controllers/api/v1/users_controller.rb
@@ -3,7 +3,7 @@ require 'cdo/firehose'
 class Api::V1::UsersController < Api::V1::JSONApiController
   before_action :load_user
   skip_before_action :verify_authenticity_token
-  skip_before_action :load_user, only: [:current, :netsim_signed_in, :post_sort_by_family_name, :post_show_progress_table_v2, :get_current_permissions, :post_disable_lti_roster_sync, :update_ai_tutor_access]
+  skip_before_action :load_user, only: [:current, :netsim_signed_in, :post_sort_by_family_name, :cached_page_auth_redirect, :post_show_progress_table_v2, :get_current_permissions, :post_disable_lti_roster_sync, :update_ai_tutor_access]
   skip_before_action :clear_sign_up_session_vars, only: [:current]
 
   private def to_bool(val)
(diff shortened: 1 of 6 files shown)


dashboard/test/ui/features/platform/signing_in.feature
```diff
@@ -16,6 +16,7 @@ Scenario: Student sign in from code.org
   Then I wait to see "#header_user_menu"
   And I wait until element ".display_name" is visible
   And element ".display_name" contains text "Bob"
+  And I delete the cookie named "_loc_notice"
```
wilkie, line 19: This test was flaky and driving me nuts. It would occasionally not be able to press the `signin` button because it would race the language selection modal. The reload was to ensure that the cookie was set that would disable that modal from appearing a second time. Yet, if it reloaded _before_ the modal and cookie happened, it would just do it again.

I just set the cookie instead of reloading.

dashboard/test/ui/features/step_definitions/steps.rb
```diff
@@ -1020,18 +1026,16 @@ def wait_for_jquery
...
-    params[:domain] = '.code.org' # top level domain cookie
-  end
+And(/^I set the pagemode cookie to "([^"]*)"$/) do |cookie_value|
+  set_cookie 'pm', cookie_value
+end
 
-  @browser.manage.add_cookie params
+And(/^I set the cookie named "([^"]*)" to "([^"]*)"$/) do |key, value|
+  set_cookie key, value
 end
```
wilkie, line 1039: DRYing this up and adding a generic rule to manipulate cookies.

dashboard/test/ui/features/platform/signing_in.feature
```diff
@@ -69,3 +70,18 @@ Scenario: Join existing section from sign in page goes to section join page
...
+  Then I wait until I am on "http://studio.code.org/users/sign_in"
+  And I wait to see "#signin"
+  And I fill in username and password for "Carah Student"
+  And I click "#signin-button" to load a new page
+  Then I wait until I am on "http://studio.code.org/s/poem-art-2021/lessons/1/levels/1"
+
+Scenario: Student already logged in navigates to provided cached level link with a login_required parameter
+  Given I create a student who has never signed in named "François Student" and go home
+  And I am on "http://studio.code.org/s/poem-art-2021/lessons/1/levels/1?login_required=true"
+  Then I wait until I am on "http://studio.code.org/s/poem-art-2021/lessons/1/levels/1"
```
wilkie: These tests both fail with the original code before this PR patch.

wilkie: image

Thankfully, I added some context around a timeout to show me what page it is actually on... and it doesn't redirect.

wilkie: Same for when the student is signed in:

image

wilkie: I've used the star wars level since that is a cached level (affected by the bug) and seeded in the test environment.

dashboard/app/controllers/api/v1/users_controller.rb
```diff
@@ -78,6 +78,24 @@ def get_contact_details
...
 
+  # This is used as a way to redirect correctly to either the sign in page or the
+  # URL specified in the user_return_to parameter. This is necessary because
+  # cached pages do not have a valid user auth token and therefore will attempt to
+  # redirect to the sign in page instead of the correct page if the user is signed in.
+  # See https://codedotorg.atlassian.net/browse/TEACH-758 for more details.
+  # NOTE: the `user_return_to` path must not include a redirect back to this path
+  # or there will be an infinite redirect loop. e.g. strip out `login_required` URL
+  # parameters before calling this method.
+  # GET /api/v1/users/cached_page_auth_redirect
```
daynew: Could we avoid this by checking for it and stripping it?
Something like:
```
if params[:user_return_to]&.include?('cached_page_auth_redirect') || session[:user_return_to]&.include?('cached_page_auth_redirect')
  params[:user_return_to] = nil
  session[:user_return_to] = nil
```

daynew (approved):
I read through the JIRA investigation and this code. These changes make sense to me.

PR #57650, opened 2024-03-29
Enable eslint `import-order` for Netsim

Hamms:
Follow-up to https://github.com/code-dot-org/code-dot-org/pull/57247

This PR enables the new `import-order` rule for all `netsim/` directories. Fixes were mostly applied automatically with `yarn lint:fix`, but I had to follow it up with a few manual fixes that the autocorrection couldn't handle. I'm not sure *why* exactly the autocorrection struggled with some of these, but I split the automatic and manual fixes out into two separate PRs in case so they can be inspected individually.

Please confirm that any comments or intentional organizational decisions are preserved!

## Testing story

## PR Checklist:

- [ ] Code is well-commented


apps/src/netsim/netsim.js
```diff
@@ -3,25 +3,28 @@
...
 import $ from 'jquery';
 import React from 'react';
 import ReactDOM from 'react-dom';
-import {openDialog as openInstructionsDialog} from '../redux/instructionsDialog';
+
 import {getStore} from '../redux';
+import {openDialog as openInstructionsDialog} from '../redux/instructionsDialog';
+
 var _ = require('lodash');
 /** @type {Object<string, function>} */
```
lfryemason: Not entirely sure what this comment is for. Could we just remove?

lfryemason (approved):
One weird comment, otherwise looks good

apps/src/netsim/netsim.js
Hamms: Good catch!

Looks like it was originally added in [this commit](https://github.com/code-dot-org/code-dot-org/commit/bb6e2562fb63#diff-5999b620ac14e29b4000442a7a132c62dbb12c12079381f3c43cfeeff22cfda5), alongside several other type annotations. I feel pretty comfortable saying we don't need a type annotation for this import, particularly given how ubiquitous it is in our codebase; I'll remove.

PR #57656, opened 2024-03-29
Remove responsive Icon Key

kobryan0619:
Shows all options in the Icon Key and Dialog regardless of what is open on the screen.

Below is what will always show on the screen for the key and dialog. 
image

image

And, here's a video which shows nothing changes on the screen.

https://github.com/code-dot-org/code-dot-org/assets/24883357/2e33be4a-0df1-4db3-8089-0f24d9146935

## Links

[Ticket](https://codedotorg.atlassian.net/jira/software/projects/TEACH/boards/65?selectedIssue=TEACH-970)

## Testing story

Modified existing tests

### DIFF
--- a/apps/src/templates/sectionProgressV2/IconKey.jsx
+++ b/apps/src/templates/sectionProgressV2/IconKey.jsx
@@ -17,11 +17,7 @@ import TeacherActionsBox from './TeacherActionsBox';
 
 import styles from './progress-table-legend.module.scss';
 
-export default function IconKey({
-  isViewingValidatedLevel,
-  expandedLessonIds,
-  sectionId,
-}) {
+export default function IconKey({sectionId}) {
   const [isOpen, setIsOpen] = useState(
     tryGetLocalStorage('iconKeyIsOpen', 'true') !== 'false'
   );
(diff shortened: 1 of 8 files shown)


apps/src/templates/sectionProgressV2/IconKey.jsx
```diff
@@ -17,11 +17,7 @@ import TeacherActionsBox from './TeacherActionsBox';
 
 import styles from './progress-table-legend.module.scss';
 
-export default function IconKey({
-  isViewingValidatedLevel,
-  expandedLessonIds,
-  sectionId,
-}) {
+export default function IconKey({sectionId}) {
```
lfryemason, line 20: I think we should remove the extra properties wherever we make an IconKey, otherwise, everything looks great!

lfryemason (approved):

apps/src/templates/sectionProgressV2/IconKey.jsx
kobryan0619: Clarifying question: is the `sectionId` the extra prop you are referring to?  It looks like this prop is used to track user interaction with `AnalyticsReporter`.  I left that in because I assumed we still wanted to track that data here.

PR #57657, opened 2024-03-29
Music: ensure music levels update when level properties change

sanchitmalhotra126:
We had an issue where certain levels in the Music Lab Elementary onboarding progression where showing a blanked out blockly workspace. This was because if two levels had identical level data and initial sources, we would never call the `onLevelLoad` hook in `MusicView` to initialize the blockly workspace. Since the Blockly workspace always gets disposed when we change levels, this would result in an empty workspace when loading the new level.

The fix here is to compare the entire level properties object when determining whether to call the `onLevelLoad` hook, so that if any property changes, we'll load all our components for the level as expected. Furthermore, I added an `id` field to the level properties object with the current level's ID, to guarantee that each level's properties are unique. That way if labs want to be sure if a level has changed, even if the other properties are identical, they can check the ID to know for sure.

Some additional context here - we do already have a current level ID stored in progress redux that some labs, including Music Lab, listen to for changes. There's a slight nuance here though, because that current level ID in progress redux technically updates before all the other lab2 fields have updated. This is because the change in the current level ID in progress redux is what triggers Lab2's ProgressContainer to start fetching data for the new level. So by adding the ID to the level properties, we always have a way to listen for level changes in a way that guarantees that all the level information has been loaded, rather than listening to the level ID in progress redux which will change before Lab2 has loaded level information. That being said, there are still cases where we may want to listen for the level ID in progress redux; specifically if we want to take any actions as soon as the level has changed, but before Lab2 has loaded the data for the next level. For example in Music Lab, we want to stop playback as soon as we start transitioning away from the level, so that music is not playing while we're loading the next level.

## Links

https://codedotorg.slack.com/archives/C043WM7TCH3/p1711726843187269

## Testing story

Tested this on the music lab intro progression, elementary intro progression, standalone levels, and project beats.

### DIFF
--- a/apps/src/lab2/lab2Redux.ts
+++ b/apps/src/lab2/lab2Redux.ts
@@ -198,7 +198,7 @@ export const setUpWithoutLevel = createAsyncThunk(
         {
           initialSources: sources,
           channel,
-          levelProperties: {appName: payload.appName},
+          levelProperties: {id: 0, appName: payload.appName},
         },
         thunkAPI.signal.aborted,
         thunkAPI.dispatch
--- a/apps/src/lab2/types.ts
+++ b/apps/src/lab2/types.ts
@@ -116,6 +116,7 @@ export interface Level {
  */
 export interface LevelProperties {
   // Not a complete list; add properties as needed.
+  id: number;
   isProjectLevel?: boolean;
   hideShareAndRemix?: boolean;
   usesProjects?: boolean;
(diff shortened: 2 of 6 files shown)


breville (approved):
Thanks for the quick fix and the comprehensive test pass.  

Curious if there might be a way to make some of the major state transitions more explicit for Lab2 levels in the future?

sanchitmalhotra126:
> Curious if there might be a way to make some of the major state transitions more explicit for Lab2 levels in the future?

Yeah, I think the level properties `id` can hopefully serve that purpose since it should be a reliable indication that the level has changed? But i'm sure there's more we can add as we need.

molly-moen (approved):
Nice fix. As we've been adding more lab2 lab types I think it may be useful for us to try to figure out how to hoist some of this logic out of individual lab types, because it tends to be similar for each lab. Not actionable here, just something I've been pondering and may try to address.

PR #57658, opened 2024-03-29
Refactor CollapsibleSection to be more generic

fisher-alice:
Currently `CollapsibleSection` is being used in the levelbuilder edit page for AI customizations. I also need collapsible sections for the presentation view of the model card so this PR refactors the component so that it is more generic.

Updates include:
- option to place an icon to the left of the title (this is required by the mock design of the presentation view of the model card).  I have not yet added the option of adding an icon to the right, but that could be in a follow-up if there is interest in this.
- Title text and content are now clickable so they can be used to expand/collapse content.
- Title text changes to teal color hover.

Note that I set the default values for the title `Typography` and the icons used for expanding/collapsing as the values currently used in `EditAiCustomizations` and `FieldSection` so that changes are not needed other than the location of the import.

There is no change in functionality on the `aichat` levelbuilder edit page other than the updates to the text (teal on hover and clickable to expand/collapse section).

AFTER UPDATE:

Below is a screenshot of collapsible sections with icons from the Presentation view in `aichat` lab.
Screenshot 2024-03-29 at 8 49 59 PM

As a comparison, here are collapsible sections in the levelbuilder edit page:
Screenshot 2024-03-29 at 8 53 11 PM

## PR Checklist:

- [ ] Code is well-commented


apps/src/templates/CollapsibleSection.tsx
```diff
@@ -0,0 +1,55 @@
...
+  titleSemanticTag?: SemanticTag;
+  titleVisualAppearance?: VisualAppearance;
+  children: React.ReactNode;
+  initiallyCollapsed?: boolean;
+  collapsedIcon?: string;
+  expandedIcon?: string;
+}> = ({
+  title,
+  titleSemanticTag = 'p',
+  titleVisualAppearance = 'body-one',
```
sanchitmalhotra126, line 27: What do you think of instead passing in the typography class name (`BodyOneText`) in place of these? Curious what @levadadenys thinks is the best way to pass this information in here

apps/src/templates/CollapsibleSection.tsx
```diff
@@ -0,0 +1,55 @@
...
+        >
+          <FontAwesomeV6Icon
+            iconName={collapsed ? collapsedIcon : expandedIcon}
+            iconStyle="solid"
+          />
+        </button>
+        <Typography
+          semanticTag={titleSemanticTag}
+          visualAppearance={titleVisualAppearance}
+          className={moduleStyles.fieldTitle}
```
sanchitmalhotra126: I think we'll probably want to be able to pass in a custom class name as a prop as well?

apps/src/templates/CollapsibleSection.tsx
```diff
@@ -0,0 +1,55 @@
+import FontAwesomeV6Icon from '@cdo/apps/componentLibrary/fontAwesomeV6Icon/FontAwesomeV6Icon';
+import Typography from '@cdo/apps/componentLibrary/typography/Typography';
+import React, {useState} from 'react';
+import moduleStyles from './collapsible-section.module.scss';
+import {
+  SemanticTag,
+  VisualAppearance,
+} from '@cdo/apps/componentLibrary/typography/types';
+
+const CollapsibleSection: React.FunctionComponent<{
```
sanchitmalhotra126: nit: I know I had put the prop type definition inline (due to hastiness 🙂), but would you mind moving this out into a `CollapsibleSectionProps` definition?

apps/src/templates/collapsible-section.module.scss
```diff
@@ -0,0 +1,17 @@
+@import '@cdo/apps/mixins.scss';
+
+.titleRow {
```
sanchitmalhotra126, line 4: nit: might need to format this file (in case the linter complains)

sanchitmalhotra126 (commented):
Thanks for doing this! A few minor questions/comments

apps/src/templates/CollapsibleSection.tsx
fisher-alice: I'll defer to @levadadenys . I initially thought of this approach, but it seems like maybe I'd need to add a type in `typography/types.ts` for the Typography types defined in `TypographyElements.
Something like:
```
type TypographyElement = Heading1 | Heading2 | ... | BodyOneText | ...;
```
Unsure about this. Thanks!

levadadenys: I think @fisher-alice used the best possible approach here from what I see in the code. If we assume title can be rendered with different sematicTag and VisualAppearance - that's how it's intended to be done.

Passing class name instead won't solve different semantic tags and also will only work if we'll be passing typographyModuleStyles[visualAppearance] which might be harder to read and understand, plus className of typography elements are in moduleStyles, meaning we can't just pass a 'body-one' className string since moduleStyles classNames are parsed into some 'coded' classNames.

Two things we might try here I think:
1) 
```
 import {
   SemanticTag as TypographyElementSematicTag,
   VisualAppearance as TypographyElementVisualAppearance,
 } from '@cdo/apps/componentLibrary/typography/types';
 ```
  for better readability
2) If we're sure that semanticTag will always be 'p' - don't pass it as a prop

Comment (fisher-alice) [reply apps/src/templates/CollapsibleSection.tsx:27]: Thanks Denys!
I will update to your first suggestion above.
And `semanticTag` will not always be 'p' so will keep it as a prop. For the presentation view of model card, the `semanticTag` is assigned 'h6'.

**Comment by fisher-alice** [comment]
> Thanks for doing this! A few minor questions/comments

Hey Sanchit, I addressed your comments but wanted to note that I did make a few updates since you last reviewed since I'm concurrently working on presentation view model card.

I added:
- option to place an icon to the left of the title,
- title text and content are now clickable so they can be used to expand/collapse content,
- title text changes to teal color hover.

I updated the PR description accordingly. Thanks!

File: apps/src/templates/CollapsibleSection.tsx [hunk 5]
```diff
@@ -0,0 +1,84 @@
...
+} from '@cdo/apps/componentLibrary/typography/types';
+
+interface CollapsibleSectionProps {
+  title: string;
+  titleSemanticTag?: TypographyElementSemanticTag;
+  titleVisualAppearance?: TypographyElementVisualAppearance;
+  titleStyle?: string;
+  titleIcon?: string;
+  titleIconStyle?: string;
+  children: React.ReactNode;
```
sanchitmalhotra126: minor readability nit: since there are so many optional props in this, could we move up `children` under `title` so the required props are at the top?

apps/src/templates/CollapsibleSection.tsx
```diff
@@ -0,0 +1,84 @@
...
+  const [collapsed, setCollapsed] = useState(initiallyCollapsed);
+
+  const hasTitleIcon = titleIcon !== undefined;
+
+  return (
+    <>
+      <div className={moduleStyles.titleRow}>
+        <button
+          type="button"
+          onClick={() => setCollapsed(!collapsed)}
```
sanchitmalhotra126: nit: pull this out into a `useCallback` and reuse (since it looks like it's used in a few different places)

sanchitmalhotra126 (approved):
LGTM! Just a few minor comments

PR #51416, opened 2023-04-19
New school and school district data ingestion from NCES for 2021-2022

vijayamanohararaj:
Seeding script updates to ingestion new school and school district information (through CSV files uploaded to S3) that contain latest NCES data for schools and school districts.

## Links
- jira ticket: https://codedotorg.atlassian.net/browse/ACQ-460?atlOrigin=eyJpIjoiNWU0MWIxMzM5Nzk3NDE5YmFjYzdlZDI5ZTMzY2IyNTciLCJwIjoiaiJ9 

## Testing story
Validated the changes locally by running the import scripts before making changes and bringing the environment to a similar state to production. Then ran the changes in dry run mode and then actually running the script for the seeding. 

## Deployment strategy
The seeding scripts are run as part of all the DTPs, so these changes should apply as part of the next DTP. These also have logs that would go to cloud watch, which has been confirmed for earlier runs and will use that for validation.

## Follow-up work

Once this is done, there will be an additional one off script to use this ingested data and populate the school stats by year table. That will be a separate PR.

## PR Checklist:

- [ ] Tests provide adequate coverage
- [ ] Privacy and Security impacts have been assessed
- [ ] Code is well-commented
- [ ] New features are translatable or updates will not break translations
- [ ] Relevant documentation has been added or updated
- [x] User impact is well-understood and desirable
- [x] Pull Request is labeled appropriately
- [ ] Follow-up work items (including potential tech debt) are tracked and linked

PR #57659, opened 2024-03-29
New school and school district data ingestion from NCES for 2022-2023

TurnerRiley:
Updates seeding to ingest 2022-2023 school and school district NCES data. The 2022-2023 data has already been uploaded in CSV format to our [Google Drive](https://drive.google.com/drive/u/1/folders/14CxZes5y5SsuviqnOg45WeWg027Xfcit) and to our [S3 bucket](https://s3.console.aws.amazon.com/s3/buckets/cdo-nces?region=us-east-1&bucketType=general&prefix=2022-2023/ccd/&showversions=false).

These changes follow [last year's changes](https://github.com/code-dot-org/code-dot-org/pull/51416) and are virtually identical other than the years being updated.

### `seed:school_districts` output
test_run_local

### `seed:schools` output
school_real_local

## Links
Jira ticket: [here](https://codedotorg.atlassian.net/jira/software/c/projects/ACQ/boards/64?assignee=60d62161f65054006980bd71&selectedIssue=ACQ-1617)
Last year's PR for this equivalent change: [here](https://github.com/code-dot-org/code-dot-org/pull/51416)

## Testing story
First I updated my local db to match that of production in the `school_districts` and `schools` tables. Then, I did a dry run of both seeding tasks and checked that the output made sense and roughly paralleled last year's numbers of added/updated/unchanged schools. Lastly, I ran them for real locally, saw the same numbers in the output, and checked that the number of entries in `school_districts` and `schools` went up by the number of new school districts and schools.

## Deployment strategy
Since seeding occurs during the DTP, we should see these changes in the DTP after this is merged.

## Follow-up work
Following @vijayamanohararaj 's strategy from last year, there will be a followup PR with a one-off script to use this ingested data (as well as the `locale_public.csv` file) to populate the `school_stats_by_year` table.

### DIFF
--- a/dashboard/app/models/school.rb
+++ b/dashboard/app/models/school.rb
@@ -421,6 +421,29 @@ class School < ApplicationRecord
           }
         end
       end
+
+      CDO.log.info "Seeding 2022-2023 public school data."
+      AWS::S3.seed_from_file('cdo-nces', "2022-2023/ccd/schools_public.csv") do |filename|
+        merge_from_csv(filename, {headers: true, quote_char: "\x00", encoding: 'bom|utf-8'}, true, is_dry_run: false, ignore_attributes: ['last_known_school_year_open']) do |row|
+          row = row.to_h.map {|k, v| [k, sanitize_string_for_db(v)]}.to_h
+          {
+            id:                           row['School ID - NCES Assigned [Public School] Latest available year'].to_i.to_s,
+            name:                         row['School Name'].upcase,
+            address_line1:                row['Location Address 1 [Public School] 2022-23'].to_s.upcase.truncate(50).presence,
+            address_line2:                row['Location Address 2 [Public School] 2022-23'].to_s.upcase.truncate(30).presence,
+            address_line3:                row['Location Address 3 [Public School] 2022-23'].to_s.upcase.presence,
+            city:                         row['Location City [Public School] 2022-23'].to_s.upcase.presence,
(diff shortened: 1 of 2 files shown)


dashboard/app/models/school.rb
```diff
@@ -421,6 +421,29 @@ def self.seed_from_s3
           }
         end
       end
+
```
TurnerRiley, line 424: [This comment](https://github.com/code-dot-org/code-dot-org/pull/57659/files#diff-063228a75a005dea95d3a994b92595419c33296fb1ae270f8e2533d20c7e5e80R401) appears before the previous school imports but I couldn't find any matches in our code base to `.to_s.slice` so I removed it since it doesn't seem relevant anymore. I'm totally open to pushback if this comment is helpful!

dashboard/app/models/school.rb
```diff
@@ -421,6 +421,30 @@ def self.seed_from_s3
...
+      # Some of this data has #- appended to the front, so we strip that off with .to_s.slice(2) (it's always a single digit)
```
TurnerRiley, line 425: This comment was at the start of each of the last few years' sections, but I'm not convinced it's still being used since [in the previous one-off scripts](https://github.com/code-dot-org/code-dot-org/blob/staging/bin/oneoff/nces_data/import_school_stats_by_years_2020_2021_ccd#L87) we still check for `#-[value]` (such as `1-Yes`). Curious on y'all's thoughts of me removing it?

bethanyaconnor (approved):
🎉

PR #57661, opened 2024-03-29
P20-828: Lazy load i18n files on the Dashboard test env

artem-vavilov:
The dashboard test env loads 6 times faster by lazily loading i18n files:
| Before  | After |
| ------- | ----- |
| without-lazy-loading | with-lazy-loading |

## Links

- [P20-806](https://codedotorg.atlassian.net/browse/P20-806)
- [P20-828](https://codedotorg.atlassian.net/browse/P20-828)

### DIFF
--- a/dashboard/config/environments/test.rb
+++ b/dashboard/config/environments/test.rb
@@ -13,6 +13,9 @@ Dashboard::Application.configure do
   # preloads Rails for running tests, you may have to set it to true.
   config.eager_load = true
 
+  # Configure i18n
+  config.i18n.backend = Cdo::I18n::LazyLoadableBackend.new(lazy_load: true)
+
   # Configure static asset server for tests with Cache-Control for performance.
   config.public_file_server.enabled = true
   config.public_file_server.headers = {'Cache-Control' => "public, max-age=3600, s-maxage=1800"}
--- a/dashboard/config/environments/test.rb
+++ b/dashboard/config/environments/test.rb
@@ -30,7 +30,7 @@ Dashboard::Application.configure do
     # Version of your assets, change this if you want to expire all your assets.
     config.assets.version = '1.0'
 
-    # Speeds up the env loading by avoid loading all i18n files up front.
+    # Avoid loading all i18n files up front, which can significantly slow down initialization.
     # Instead, it only loads i18n files that belong to the current locale.
     config.i18n.backend = Cdo::I18n::LazyLoadableBackend.new(lazy_load: true)
   end
--- a/dashboard/config/environments/test.rb
+++ b/dashboard/config/environments/test.rb
@@ -13,9 +13,6 @@ Dashboard::Application.configure do
   # preloads Rails for running tests, you may have to set it to true.
   config.eager_load = true
 
-  # Configure i18n
-  config.i18n.backend = Cdo::I18n::LazyLoadableBackend.new(lazy_load: true)
-
   # Configure static asset server for tests with Cache-Control for performance.
   config.public_file_server.enabled = true
   config.public_file_server.headers = {'Cache-Control' => "public, max-age=3600, s-maxage=1800"}
@@ -32,6 +29,10 @@ Dashboard::Application.configure do
 
     # Version of your assets, change this if you want to expire all your assets.
     config.assets.version = '1.0'
+
+    # Speeds up the env loading by avoid loading all i18n files up front.
+    # Instead, it only loads i18n files that belong to the current locale.
+    config.i18n.backend = Cdo::I18n::LazyLoadableBackend.new(lazy_load: true)
   end
 
   config.assets.quiet = true
--- a/lib/cdo/i18n_backend.rb
+++ b/lib/cdo/i18n_backend.rb
@@ -220,15 +220,20 @@ module Cdo
         @loaded_files ||= Set.new
       end
 
+      # Get the set of valid locales for a given base locale code. For example,
+      # for a base of `:pt`, we expect valid locales `:pt`, `:"pt-BR"`, `:"en-US"`, and `en`
+      def valid_locales_for(base_locale)
+        valid_locales = Set.new(::I18n.fallbacks[base_locale])
+        valid_locales << base_locale
+        valid_locales << LOCALES_MAPPING[base_locale] # en: :'en-US'
+        valid_locales.compact_blank
+      end
+
       # The original method has been modified to also load files for the current locale's fallbacks,
       # like loading "en-US" files for the "en" locale and vice versa.
       # https://github.com/ruby-i18n/i18n/blob/v1.12.0/lib/i18n/backend/lazy_loadable.rb#L164-L171
       def filenames_for_current_locale
-        # Enables loading both the current locale and its fallbacks, such as "en" for "en-US" and vice versa.
-        valid_locales = Set.new(::I18n.fallbacks[::I18n.locale])
-        valid_locales << LOCALES_MAPPING[::I18n.locale] # en: :'en-US'
-        valid_locales.compact_blank
-
+        valid_locales = valid_locales_for(::I18n.locale)
         # Excludes already loaded i18n files to prevent them from being reloaded during locale switching.
         # For example, for the "de-DE" locale, the i18n fallbacks are "de", "en", and "en-US".
         # Since "en-US" (and "en") is the default locale, its i18n files are already loaded.
@@ -238,12 +243,23 @@ module Cdo
         end
       end
 
-      # Prevents from raising an error when the locale in the filename and
-      # the root locale key in the file content mismatch,
-      # like when `en.yml` has a root locale key of `en-US:` instead of `en:`.
-      # This error is no longer relevant because the functionality has been changed to load such files regardless.
+      # Checks if a filename is named in correspondence to the translations it loaded.
+      # The locale extracted from the path must be either the single locale loaded in
+      # the translations or one of the expected variations of the locale, taking into
+      # account both long locale codes and fallbacks.
       # https://github.com/ruby-i18n/i18n/blob/v1.12.0/lib/i18n/backend/lazy_loadable.rb#L173-L181
-      def assert_file_named_correctly!(...)
+      def assert_file_named_correctly!(file, translations)
+        expected_locale = ::I18n::Backend::LocaleExtractor.locale_from_path(file)
+        return if expected_locale.nil?
+
+        valid_locales = valid_locales_for(expected_locale)
+        unexpected_locales = translations.each_key.reject {|locale| valid_locales.include?(locale.to_sym)}
+        return if unexpected_locales.empty?
+
+        warn "INVALID I18N: Incorrect filename of #{file}".yellow
+        warn "       valid: #{valid_locales.inspect}".blue
+        warn "    expected: #{expected_locale.inspect}".green
+        warn "  unexpected: #{unexpected_locales.inspect}".red
       end
     end
   end
--- a/lib/cdo/i18n_backend.rb
+++ b/lib/cdo/i18n_backend.rb
@@ -175,6 +175,13 @@ module Cdo
       include Plugins
     end
 
+    # The class extends the `LazyLoadable` backend from the i18n gem to support CDO i18n files:
+    # - Modified the `LocaleExtractor` class to support filenames that include both a prefix and the locale (e.g., "courses.en-US.yml").
+    # - Modified the `filenames_for_current_locale` method to load files for the current locale and its fallbacks, (e.g, "en" for "en-US" and vice versa)
+    # - Modified the `eager_load!` method to ignore eager loading when lazy loading is enabled.
+    # - Prevented the `assert_file_named_correctly!` method from raising an irrelevant error regarding file locale mismatches.
+    # - Implemented a mechanism to avoid reloading already loaded translation files.
+    # https://github.com/ruby-i18n/i18n/blob/v1.12.0/lib/i18n/backend/lazy_loadable.rb
     class LazyLoadableBackend < ::I18n::Backend::LazyLoadable
       include Plugins
 
@@ -182,9 +189,12 @@ module Cdo
         locales[k.to_sym] = v.to_sym if v.is_a?(String)
       end.freeze
 
+      # The original method has been modified to works on i18n files named with a prefix and the locale,
+      # like "common.en.yml" or "common.en-US.json" and not only on files named with the locale, like "en-US.json".
+      # https://github.com/ruby-i18n/i18n/blob/v1.12.0/lib/i18n/backend/lazy_loadable.rb#L55-L63
       class ::I18n::Backend::LocaleExtractor
         def self.locale_from_path(path)
-          # Extracts the locale name from the path, like "en", "en-US", "haw", "haw-HI", etc.
+          # Extracts the locale from the path, like "en", "en-US", "haw", "haw-HI", etc.
           path[/\b([a-z]{2,3}(?:-[A-Z]{2})?)\b\.\w+$/, 1]&.to_sym
         end
       end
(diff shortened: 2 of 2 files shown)


lib/cdo/i18n_backend.rb
```diff
@@ -164,12 +164,69 @@ def translate(locale, key, options = ::I18n::EMPTY_HASH)
...
+    end
+
+    class LazyLoadableBackend < ::I18n::Backend::LazyLoadable
+      include Plugins
+
+      LOCALES_MAPPING = YAML.load_file(CDO.dir('dashboard/config/locales.yml')).each_with_object({}) do |(k, v), locales|
+        locales[k.to_sym] = v.to_sym if v.is_a?(String)
+      end.freeze
+
+      class ::I18n::Backend::LocaleExtractor
```
Hamms, line 195: Please add some comments explaining that here we are intentionally reopening this existing class object from the underlying repository in order to overwrite its functionality. Just from looking at this code, it's not clear why it's important to do it this way rather than just making `locale_from_path` a local method on LazyLoadableBackend itself. Some links to [the original implementation](https://github.com/ruby-i18n/i18n/blob/7842cab498a9b1da0ddd27a986caf61c38ad776b/lib/i18n/backend/lazy_loadable.rb#L55-L63) would also be useful!

lib/cdo/i18n_backend.rb
```diff
@@ -164,12 +164,69 @@ def translate(locale, key, options = ::I18n::EMPTY_HASH)
...
       include MarkdownTranslate
       include SafeInterpolation
       include I18nStringUrlTrackerPlugin
     end
+
+    class SimpleBackend < ::I18n::Backend::Simple
+      include Plugins
+    end
+
+    class LazyLoadableBackend < ::I18n::Backend::LazyLoadable
```
Hamms, line 185: I'd love to see some comments explaining why we added this functionality!

In particular, I'd like to see a high-level summary of the ways in which the standard `LazyLoadable` backend was not quite sufficient for our purposes, and why we modified it in the way we did.

lib/cdo/i18n_backend.rb
```diff
@@ -164,12 +164,69 @@ def translate(locale, key, options = ::I18n::EMPTY_HASH)
...
+        (::I18n.load_path.flatten - loaded_files.to_a).select do |path|
+          path_locale = ::I18n::Backend::LocaleExtractor.locale_from_path(path)
+          path_locale.nil? || valid_locales.include?(path_locale)
+        end
+      end
+
+      # Prevents from raising an error when the locale in the filename and
+      # the root locale key in the file content mismatch,
+      # like when `en.yml` has a root locale key of `en-US:` instead of `en:`
+      def assert_file_named_correctly!(...)
```
Hamms: Is there a version of this error checking that would work for us? I don't love the idea of just preventing errors entirely. Can we still check to see that the locale key *starts with* the expected locale, for example, even if we expect it to sometimes have extra region data? Or, even better, use `LOCALES_MAPPING`?

dashboard/config/environments/test.rb
```diff
@@ -13,6 +13,9 @@
   # preloads Rails for running tests, you may have to set it to true.
   config.eager_load = true
 
+  # Configure i18n
```
Hamms: Please explain what specifically we're configuring i18n to do, and most importantly *why* we're doing this configuration. Comments are most useful when they're used to capture context that the person reading this code in the future might not otherwise have.

lib/cdo/i18n_backend.rb
```diff
@@ -164,12 +164,69 @@ def translate(locale, key, options = ::I18n::EMPTY_HASH)
...
+        super unless lazy_load?
+      end
+
+      protected
+
+      def loaded_files
+        @loaded_files ||= Set.new
+      end
+
+      def filenames_for_current_locale
```
Hamms: Can you clarify what we're trying to achieve by overwriting this method? In particular, it's not clear to me why we're looking at values in the `load_path` *except* for those that have been loaded

Hamms (commented):
So excited to see this work getting done, thank you! I have a handful of questions and there are a few places I'd like to see some comments in the code, but for the most part this looks great!

One other thing to consider is that this does introduce a new difference between our test and production environments, which weakens the accuracy of our tests. This functionality seems super useful when running one-off tests locally in a development environment, but in the context of a full test run as part of the CI/CD pipeline, I think the time savings are negligible and the divergence from production has more potential to obscure legitimate problems that might manifest in production. Is there a way of configuring this that allows us to distinguish between local and automated testing?

lib/cdo/i18n_backend.rb
artem-vavilov: This error highlights files that won't load due to mismatched locales in their names and content.
But since we changed the class to load such files, the error isn't relevant to us anymore.

lib/cdo/i18n_backend.rb
Hamms: Comments look good to me, thanks!

lib/cdo/i18n_backend.rb
```diff
@@ -164,12 +164,88 @@ def translate(locale, key, options = ::I18n::EMPTY_HASH)
...
+    class LazyLoadableBackend < ::I18n::Backend::LazyLoadable
+      include Plugins
+
+      LOCALES_MAPPING = YAML.load_file(CDO.dir('dashboard/config/locales.yml')).each_with_object({}) do |(k, v), locales|
+        locales[k.to_sym] = v.to_sym if v.is_a?(String)
+      end.freeze
+
+      # The original method has been modified to works on i18n files named with a prefix and the locale,
+      # like "common.en.yml" or "common.en-US.json" and not only on files named with the locale, like "en-US.json".
+      # https://github.com/ruby-i18n/i18n/blob/v1.12.0/lib/i18n/backend/lazy_loadable.rb#L55-L63
```
Hamms, line 194: Very nice touch linking to a tagged version rather than just an arbitrary commit!

lib/cdo/i18n_backend.rb
Hamms: Just to make sure I understand correctly: is this one also a problem specific to our fallback functionality? As in, it's a problem because when the locale is `de-DE`, some of the files this method will be looking at will be named `en-US`? It makes sense to me that it would be too difficult to update the existing logic to take that into account; or at least, more difficult than it's really worth.

However, the comment (and my own initial reading when I wasn't thinking about the possibility of fallbacks) makes it seem like it's just a problem with us sometimes using `en` and sometimes using `en-US`, and that _does_ seem to me like something that would be easy enough to update the existing logic to accommodate. If that's not the case, could you update the comment to specify? Thanks!

Hamms (approved):
The added comments are super helpful, thank you! Particularly the explanation about how some of this logic interacts with our configured fallbacks; the detail about that causing issues with files already being loaded is a subtlety I missed in my first read.

Glad to see it was easier than I expected it to be to make this apply only to local testing.

I do still have one question about the `assert_file_named_correctly` method, but it's nonblocking. This LGTM!

lib/cdo/i18n_backend.rb
artem-vavilov: The issue with the COD i18n files is that many of them include a full locale like `de-DE` in the filename (e.g. [blocks.de-DE.yml](https://github.com/code-dot-org/code-dot-org/blob/staging/dashboard/config/locales/blocks.de-DE.yml)) but the root locale key is just `de` and this 
triggers the error.

artem-vavilov: The method ensures that for current locales like `pt-PT`, only `pt-PT` translations are loaded. However, the i18n backend has been changed to support fallback locales (`pt`, `pt-BR`, `en-US`, etc.). Also, many `pt-PT` files contain `pt` translations and modifications have been made to the code to allow these translations to be loaded regardless. 
Therefore, the original check for loading exclusively the current locale translations is no longer necessary.

Hamms: Hmm, I'm still not understanding why the check is no longer necessary. I get that it's now a lot more complicated to determine whether or not a given locale makes sense for a given file, but I don't see how our new loading scheme makes it _unnecessary_. I think we'd still like to know if a file with `pt-PT` in the filename is trying to load translations for the locale key `de`, for example.

Why not extract the `valid_locales` logic in `filenames_for_current_locale` into a helper method, then reimplement `assert_file_named_correctly` to check that the loaded locale is one of the multiple valid locales? Something along these lines (but probably optimized a bit better for performance):

```ruby
# Get the set of valid locales for a given base locale code. For example,
# for a base of `:pt`, we expect valid locales `:pt`, `:"pt-BR"`,
# `:"en-US"`, and `en`
def valid_locales_for(base_locale)
  valid_locales = Set.new(::I18n.fallbacks[base_locale])
  valid_locales << LOCALES_MAPPING[base_locale] # en: :'en-US'
  return valid_locales.compact_blank
end

def filenames_for_current_locale
  # Excludes already loaded i18n files to prevent them from being reloaded during locale switching.
  # For example, for the "de-DE" locale, the i18n fallbacks are "de", "en", and "en-US".
  # Since "en-US" (and "en") is the default locale, its i18n files are already loaded.
  (::I18n.load_path.flatten - loaded_files.to_a).select do |path|
    path_locale = ::I18n::Backend::LocaleExtractor.locale_from_path(path)
    path_locale.nil? || valid_locales_for(::I18n.locale).include?(path_locale)
  end
end

# Checks if a filename is named in correspondence to the translations it loaded.
# The locale extracted from the path must be either the single locale loaded in
# the translations or one of the expected variations of the locale, taking into
# account both long locale codes and fallbacks
def assert_file_named_correctly!(file, translations)
  loaded_locales = translations.keys.map(&:to_sym)
  expected_locale = ::I18n::Backend::LocaleExtractor.locale_from_path(file)
  unexpected_locales = loaded_locales.reject do |locale|
    expected_locale.nil? ||
      locale == expected_locale ||
      expected_locale.start_with?(locale.to_s) ||
      valid_locales_for(locale).include?(expected_locale)
  end

  raise ::I18n::Backend::LazyLoadable::FilenameIncorrect.new(file, expected_locale, unexpected_locales) unless unexpected_locales.empty?
end
```

lib/cdo/i18n_backend.rb
```diff
@@ -164,12 +164,104 @@ def translate(locale, key, options = ::I18n::EMPTY_HASH)
...
+        (::I18n.load_path.flatten - loaded_files.to_a).select do |path|
+          path_locale = ::I18n::Backend::LocaleExtractor.locale_from_path(path)
+          path_locale.nil? || valid_locales.include?(path_locale)
+        end
+      end
+
+      # Checks if a filename corresponds to the locale it has loaded.
+      # The locale extracted from the path must be either the single locale loaded in
+      # the translations or one of the expected variations of the locale, taking into
+      # account both long locale codes and fallbacks.
```
Hamms, line 259: Please add a comment calling out that the original implementation raised an error but we're changing that to simply log the details to the console, and explaining that we're doing so because the error isn't all that critical to catch and our new loading logic includes more weird edge cases than we really want to deal with.

dashboard/config/environments/test.rb
```diff
@@ -29,6 +29,10 @@
 
     # Version of your assets, change this if you want to expire all your assets.
     config.assets.version = '1.0'
+
+    # Speeds up the env loading by avoid loading all i18n files up front.
```
Hamms: Grammar nit: technically, it's more correct to use "avoiding" here:

```suggestion
    # Speeds up the env loading by avoiding loading all i18n files up front.
```

But "loading by avoiding loading" is a bit of a mouthful. What about something like,

```suggestion
    # Avoid loading all i18n files up front, which can significantly slow down initialization.
```

lib/cdo/i18n_backend.rb
```diff
@@ -164,12 +164,104 @@ def translate(locale, key, options = ::I18n::EMPTY_HASH)
...
+    # - Implemented a mechanism to avoid reloading already loaded translation files.
+    # https://github.com/ruby-i18n/i18n/blob/v1.12.0/lib/i18n/backend/lazy_loadable.rb
+    class LazyLoadableBackend < ::I18n::Backend::LazyLoadable
+      include Plugins
+
+      LOCALES_MAPPING = YAML.load_file(CDO.dir('dashboard/config/locales.yml')).each_with_object({}) do |(k, v), locales|
+        locales[k.to_sym] = v.to_sym if v.is_a?(String)
+      end.freeze
+
+      # The original method has been modified to works on i18n files named with a prefix and the locale,
```
Hamms: Grammar nit:
```suggestion
      # The original method has been modified to work on i18n files named with a prefix and the locale,
```

lib/cdo/i18n_backend.rb
```diff
@@ -164,12 +164,104 @@ def translate(locale, key, options = ::I18n::EMPTY_HASH)
...
+      # https://github.com/ruby-i18n/i18n/blob/v1.12.0/lib/i18n/backend/lazy_loadable.rb#L173-L181
+      def assert_file_named_correctly!(file, translations)
+        expected_locale = ::I18n::Backend::LocaleExtractor.locale_from_path(file)
+        return if expected_locale.nil?
+
+        valid_locales = valid_locales_for(expected_locale)
+        unexpected_locales = translations.each_key.reject {|locale| valid_locales.include?(locale.to_sym)}
+        return if unexpected_locales.empty?
+
+        warn "INVALID I18N: Incorrect filename of #{file}".yellow
```
Hamms: Some more specificity here would be good. What about something like, `INVALID I18N: File #{file.inspect} contains translations for unexpected locale`

lib/cdo/i18n_backend.rb
```diff
@@ -164,12 +164,104 @@ def translate(locale, key, options = ::I18n::EMPTY_HASH)
...
+      end
+
+      # Get the set of valid locales for a given base locale code. For example,
+      # for a base of `:pt`, we expect valid locales `:pt`, `:"pt-BR"`, `:"en-US"`, and `en`
+      def valid_locales_for(base_locale)
+        valid_locales = Set.new(::I18n.fallbacks[base_locale])
+        valid_locales << base_locale
+        valid_locales << LOCALES_MAPPING[base_locale] # en: :'en-US'
+        valid_locales.compact_blank
+      end
```
Hamms: Is it worth memoizing this method? I don't think we expect the results to ever change for a given locale, and with the changes to `assert_file_named_correctly` it seems like it's going to get called pretty frequently.

Hamms (commented):
Couple of questions and nitpicks, but overall this is looking great. Very nearly there!

lib/cdo/i18n_backend.rb
artem-vavilov: Do you mean memoizing in a hash per locale?

Hamms: Yep! Just something simple. Not super important, but could be a nice optimization

lib/cdo/i18n_backend.rb
```diff
@@ -164,12 +164,118 @@ def translate(locale, key, options = ::I18n::EMPTY_HASH)
...
+
+      # Checks if a filename corresponds to the locale it has loaded.
+      # The locale extracted from the path must be either the single locale loaded in
+      # the translations or one of the expected variations of the locale, taking into
+      # account both long locale codes and fallbacks.
+      #
+      # The original implementation was changed from raising an error to log the details to the console,
+      # because the error isn't all that critical to catch and
+      # the new loading logic includes more weird edge cases.
```
Hamms: nit: grammar and line wrapping
```suggestion
      # The original implementation was changed from raising an error to logging
      # the details to the console, because the error isn't all that critical
      # to catch and the new loading logic includes more weird edge cases.
```

Hamms (approved):
Super tiny nit, then this looks good to me! :shipit:

dashboard/config/environments/test.rb
```diff
@@ -29,6 +29,10 @@
 
     # Version of your assets, change this if you want to expire all your assets.
     config.assets.version = '1.0'
+
+    # Avoid loading all i18n files up front, which can significantly slow down initialization.
+    # Instead, it only loads i18n files that belong to the current locale.
+    config.i18n.backend = Cdo::I18n::LazyLoadableBackend.new(lazy_load: true)
```
stephenliang, line 35: Out of curiosity, does this mean that we will use `LazyLoadableBackend` in the `test` environment? Would it be better to only lazy load in the unit test execution environment only (I believe this is with the `ENV['UNIT_TEST']` flag.

artem-vavilov: We use it only in the local test environment, so why not use it for integration tests too?

stephenliang: Isn't the `test` environment this [website](https://test-studio.code.org)? I'm just wondering if that would cause that site to use the `LazyLoadableBackend` instead of the regular one which would cause a discrepancy against production.

artem-vavilov: In Ruby on Rails, the test environment is specifically designed for running tests.
I don't know if we're using it for that website, but it would be incorrect to do so.

PR #51574, opened 2023-04-27
Fixing NCES data ingestion script which was using the wrong index for most of the headers

vijayamanohararaj:
The indexing into the CSV files were incorrect and hence parsed all the values as nulls. This resulted in incorrect data import for the latest year. The change is to fix the headers to match the CSV.

## Testing story
I tested this locally again with the new change and this time tried to sample some of the data for various different columns to ensure that they are not all either nil or false, which is what happened in production.

Also, for more context, previously, I was expecting the script to fail with errors (like the ones for school did) when a header is incorrect and hence didn't explicitly check for data correctness. This lead to the bug seeping into production which was the first place where an E2E AFE scenario can be tested. In this change, also did backend data validations as mentioned above. As this gets rolled out, I'm planning to do similar data validations in staging and in production after running the script.

PR #51491, opened 2023-04-24
NCES school data import for 2021 to 2022

vijayamanohararaj:
One off scrip to ingest the stats for schools for SY 2021-2022. This would populate the SchoolStatsByYear table using the csv uploaded for the corresponding school year.

## Links
JIRA https://codedotorg.atlassian.net/browse/ACQ-460?atlOrigin=eyJpIjoiZmRmZGRjODIyMzQ1NDIyMzg5YWVhZWZiYjU2NjdjNDQiLCJwIjoiaiJ9

## Testing story

When testing locally, the script failed for 123 schools as they were missing corresponding entries in Schools table. Those schools were skipped as duplicates because there was an existing school entry with the same state school ID. This issue is still unresolved, but it looks like ingestions in the past also hit this issue from cloud watch logs and DB queries. Hence, moving forward with this to unblock ingestion for majority of the schools, while I continue following up on the issue with duplicates due to state school ID.

https://codedotorg.slack.com/archives/C04540KNGDQ/p1682359022125929

## Deployment strategy
Will execute the script on production once rolled out. The script has an override to specify dry run, so will start with dry run and then execute if that looks good.

## Follow-up work
Resolution of the duplicate issue for the 123 schools hit during local run.

## PR Checklist:

- [ ] Tests provide adequate coverage
- [ ] Privacy and Security impacts have been assessed
- [x] Code is well-commented
- [ ] New features are translatable or updates will not break translations
- [x] Relevant documentation has been added or updated
- [x] User impact is well-understood and desirable
- [x] Pull Request is labeled appropriately
- [x] Follow-up work items (including potential tech debt) are tracked and linked

PR #57664, opened 2024-03-29
NCES school data import for 2022 to 2023

TurnerRiley:
Adds one-off script to ingest the stats for schools for the 2022-2023 school year. Once run, it will populate the `SchoolStatsByYear` table using the [schools_public.csv](https://s3.console.aws.amazon.com/s3/object/cdo-nces?region=us-east-1&bucketType=general&prefix=2022-2023/ccd/schools_public.csv) and [locale_public.csv](https://s3.console.aws.amazon.com/s3/object/cdo-nces?region=us-east-1&bucketType=general&prefix=2022-2023/ccd/locale_public.csv) files.

This PR mirrors @vijayamanohararaj 's [mostly equivalent PR](https://github.com/code-dot-org/code-dot-org/pull/51491) from last year. The only differences are:
- Title 1 Status is not included in this year's data so we're going to populate it with last year's data (if available for the given school) and go back to update if/when we can access data that maps Title 1 status to NCES school ids
- the school year has been updated
- this PR consolidates the changes in [the equivalent PR](https://github.com/code-dot-org/code-dot-org/pull/51491) and [its related fix-forward](https://github.com/code-dot-org/code-dot-org/pull/51574).

## Links
Jira ticket: [here](https://codedotorg.atlassian.net/jira/software/c/projects/ACQ/boards/64?assignee=60d62161f65054006980bd71&selectedIssue=ACQ-1617)
PR this builds on for ingesting the school and school_districts data: [here](https://github.com/code-dot-org/code-dot-org/pull/57659)
Last year's equivalent PR this mirrors: [here](https://github.com/code-dot-org/code-dot-org/pull/51491) (with a fix-forward [here](https://github.com/code-dot-org/code-dot-org/pull/51574))

### Testing the script as a whole
Ran the script locally and saw the number of entries added equaled the number of total school entries (1,040 new + 8,655 updated + 91,695 unchanged = 101,390).

Results from running this PR's script:
schooL_stat_import
Results from when school data was imported:
school_import

### Testing Title 1 Statuses
Since this script also pulls from last year's Title 1 Status entries, I wanted to check that it was properly acquiring them for non-null values from last year. I ran the script first setting all Title 1 Statuses this year to `NULL` and this was the distribution of Title 1 Status values (including the 2021-2022 data):
before

Then, I ran the script setting the Title 1 Status to the previous year's value. The log showed that 88,411 entries were updated, and when I re-queried the distribution of Title 1 Status values the `NULL` count went down by 88,411 (while others went up by the same total amount) so I have confidence this is able to properly set this year's Title 1 Status to last years (when available).
updated
after

## Deployment strategy
Once merged, I'll run the script on production. Just like last year, this script has an override to specify a dry run, so I'll start with a dry run and then execute if that looks good.


dashboard/app/models/school_stats_by_year.rb
```diff
@@ -58,7 +58,7 @@ def self.merge_from_csv(filename, options = {col_sep: "\t", headers: true, quote
     errored_schools = []
 
     ActiveRecord::Base.transaction do
-      CSV.read(filename, options).each do |row|
+      CSV.read(filename, **options).each do |row|
```
TurnerRiley, line 61: I wasn't able to get any of the oneoff school_stats_by_year scripts to run without making this update (it was complaining about having 2 parameters when it only expected 1). I'm not quite sure why it [worked for Vijaya](https://github.com/code-dot-org/code-dot-org/pull/51491) without this change here but didn't work for me. Does this seem like a safe change?

bethanyaconnor: I think so. I _think_ this might be related: https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/, which would make this a consequence of the Ruby upgrade

bethanyaconnor (commented):
In our screenshot, it looks like stats only got added, not updated. Is that expected? Or just a consequence of local testing?

TurnerRiley:
> In our screenshot, it looks like stats only got added, not updated. Is that expected? Or just a consequence of local testing?

Yep the stats are by year so we should be seeing them all added for the new school year!

dashboard/app/models/school_stats_by_year.rb
TurnerRiley: Ahhh gotcha!

bethanyaconnor (approved):

PR #57668, opened 2024-03-29
Gen AI: use new `Button` component

bencodeorg:
Doing a little Friday afternoon styling 💅 -- two changes:
- switch over to using @levadadenys new `Button` component (works great!)
- move the new presentation panel toggle into a top-level container. The buttons at the bottom of the window ("Update", etc were being pushed off the page)

| Before | After (changes just in button styling in a few places) |
| - | - |
| image | image |

## Testing story

Just testing manually that things look as expected on both the presentation and edit views.


sanchitmalhotra126 (approved):
Looks great!

apps/src/aichat/views/modelCustomization/PromptCustomization.tsx
```diff
@@ -109,9 +110,12 @@ const PromptCustomization: React.FunctionComponent = () => {
         )}
       </div>
       <div className={styles.footerButtonContainer}>
-        <button type="button" disabled={allFieldsDisabled} onClick={onUpdate}>
-          Update
-        </button>
+        <Button
+          text="Update"
```
fisher-alice, line 107: Add 'edit' icon here as well? I see it in mock up.

apps/src/aichat/views/aichatView.module.scss
```diff
@@ -13,8 +13,6 @@
   overflow: hidden;
   user-select: none;
   display: flex;
-  flex-direction: row;
-  gap: 2px;
   margin: 2px;
 
   .instructionsArea {
```
fisher-alice, line 18: Looking at mockup, it seems like instructions and customization panels are about the same width. https://codedotorg.slack.com/archives/C06FELVTV0Q/p1711150032784399
What do you think about assigning the width of both areas to something like `30%`?

fisher-alice: Also - this is outside of scope of this PR, but observing that the button is currently not in view in the 'Publish' tab. I tried adding `overflow-y` to `cutomizationArea` but unsure why it didn't work.

fisher-alice (approved):
Nice refactoing! Just a couple minor questions.

apps/src/aichat/views/modelCustomization/PromptCustomization.tsx
bencodeorg: Done!

apps/src/aichat/views/aichatView.module.scss
bencodeorg: ugh, getting headers and footers to stick on smaller screens will be the death of me! yeah I'd like to get this merged to avoid more merge conflicts, can follow up with changes dealing with buttons disappearing and width of the different panels, if that's ok with you.

fisher-alice: That's totally fine with me - thanks!

PR #57671, opened 2024-03-30
Music: sound icons

breville:
This updates the sound icons in Music Lab to use Font Awesome rather than PNGs, and centralizes sound styling.  

The new icons can be seen in the sound selection UI (`SoundsPanel`) which renders using React, and the `play sound` block (`FieldSounds`) which renders using SVG.

While making this change, the relevant styling information has been centralized in a `SoundStyle` helper with accompanying SCSS.  

The styling of timeline elements (`TimelineElement`) now uses this shared helper as well.

### Media

The new icons can be seen in the sound selection UI:

Screenshot 2024-03-30 at 9 45 17 PM

The new icons can also be seen in the `play sound` blocks:

Screenshot 2024-03-30 at 9 57 05 PM

### Colors

I took advantage of the new centralized CSS to tweak the timeline element colors a bit, too.


apps/src/music/utils/SoundStyle.ts
```diff
@@ -0,0 +1,93 @@
...
+    classNameBackground: string;
+    // Class name for the border-color style attribute.
+    classNameBorder: string;
+  };
+};
+
+// A centralized collection of styling information for sounds, used by the SoundsPanel,
+// FieldSounds, and TimelineElement.
+
+const soundStyle: SoundStyle = {
```
sanchitmalhotra126, line 27: Neat!

apps/src/music/views/SoundStyle.module.scss
```diff
@@ -0,0 +1,108 @@
+@import 'color.scss';
+
+.sound {
+  &Beat {
```
sanchitmalhotra126: optional - I think you could use a mixin here to avoid duplicating the structure if you'd like? Something like
```
@mixin sound-type-style($type, $color, $fill, $background, $border) {
  &#{type} {
    &Color { color: $color }
    &Fill { fill: $fill }
    ...
  }

...

.sound {
  @include sound-type-style(Beat, #FC55FF, #FC55FF, ... );
  @include sound-type-style(Bass, ... );
  ...
}
```

apps/src/music/views/SoundStyle.module.scss
```diff
@@ -0,0 +1,108 @@
...
+    }
+    &Background {
+      background-color: #a35a02;
+    }
+    &Border {
+      border-color: #ff8c00;
+    }
+  }
+
+  &Pattern {
```
sanchitmalhotra126: While we're at it, thoughts on updating colors for Pattern and Chord? I remember these were chosen kind of arbitrarily, maybe worth asking @markabarrett for some colors that would work with this palette?

sanchitmalhotra126 (approved):
Slick code refactor, LGTM!

apps/src/music/views/SoundStyle.module.scss
breville: Good question.  They were designed to match the colors seen in the Blockly field, but open to other ideas!

apps/src/music/views/SoundStyle.module.scss
breville: Love it.  Done.

PR #57672, opened 2024-03-31
Raise `StudentFacingError` when importing an invalid CSV file

### LINKED ISSUE #57676: Importing an invalid CSV produces no error
When importing an invalid CSV in Firebase we don't get a user-facing error, but we do get a JS error in the browser JS console:
Image

With a fix PR: https://github.com/code-dot-org/code-dot-org/pull/57672 we now get a 400 message not a 500, but it still doesn't show up anywhere (except network errors).

https://app.honeybadger.io/projects/3240/faults?q=request.url%3A%22%2Adatablock%2A%22+-is%3Aresolved+-is%3Aignored#:~:text=CSV%3A%3AMalformedCSVError%20in%20datablock_storage%23import_csv

An obvious followup to this PR would be one that makes this a user visible error. In the best case, it might even include a hint on how the student can modify the CSV they are importing to be valid according to our expected CSV format.

snickell:
This does NOT cause an error to show up for the student (which would be nice, but silent failure matches Firebase's behavior, which is our target bug for bug), but it also does not cause a 500 error that shows up in Honey Badger.

Fixes: #57676

### DIFF
--- a/dashboard/app/models/datablock_storage_table.rb
+++ b/dashboard/app/models/datablock_storage_table.rb
@@ -231,6 +231,8 @@ class DatablockStorageTable < ApplicationRecord
     end
 
     create_records(records)
+  rescue CSV::MalformedCSVError => exception
+    raise StudentFacingError.new(:INVALID_CSV), "Could not import CSV as it was not in the format we expected: #{exception.message}"
   end
 
   def add_column(column_name)


cnbrenci (approved):
I'll log a follow-up in Student Labs for a better UI experience for failures to upload CSVs.

PR #57673, opened 2024-03-31
Music: embeddable mini-player

breville:
This explores an embeddable version of the Music Lab mini-player that can be used on third-party sites.  

Designed for internal use for now, it will let us test whether the cross-domain requests work.

It can be embedded in an `iframe` with a comma-separated list of channel IDs, such as https://studio.code.org/musiclab/embed?channels=channelID1,channelID2,channelID3


dashboard/app/controllers/musiclab_controller.rb
```diff
@@ -55,6 +55,26 @@ def gallery
       to_json
   end
 
+  def embed
+    response.headers['X-Frame-Options'] = 'ALLOWALL'
+    response.headers['Content-Security-Policy'] = ''
```
breville, line 60: Borrowed from: https://github.com/code-dot-org/code-dot-org/blob/db70996804531e81bb83452572b9f1aaca72ac98/dashboard/app/controllers/projects_controller.rb#L372-L375

dashboard/config/routes.rb
```diff
@@ -45,6 +45,7 @@
     get "/projectbeats", to: "musiclab#index"
     get "/musiclab/menu", to: "musiclab#menu"
     get "/musiclab/gallery", to: "musiclab#gallery"
+    get "/musiclab/embed", to: "musiclab#embed"
```
sanchitmalhotra126, line 48: minor: what do you think about calling this `musiclab/playlist_embed` / `embed_playlist` or similar?

edit: or alternatively `/playlist/embed` if we think that we might use `/playlist` to display the same content just on our site (without an iframe)?

sanchitmalhotra126 (approved):
LGTM 👍

dashboard/config/routes.rb
breville: Good question.  Given that we are pausing this feature work right now, beyond this cross-site testing, I think I'll hold off on a rename until we learn more about what happens next.  But agreed that differentiating this from other types of embeds is probably a good next move.

PR #57683, opened 2024-04-01
Retry table_or_create on failure

### LINKED ISSUE #57681: ActiveRecord::RecordNotUnique in datablock_storage#create_record Mysql2::Error: Duplicate entry 'NUMBER-TableName' for key 'PRI...
https://app.honeybadger.io/projects/3240/faults?q=request.url%3A%22%2Adatablock%2A%22+-is%3Aresolved+-is%3Aignored#:~:text=ActiveRecord%3A%3ARecordNotUnique%20in%20datablock_storage%23create_record

snickell:
Fixes race condition caused by first_or_create() not being atomic, found using honeybadger. If two first_or_create() run in parallel, they might both try to create, but the second create will fail due to duplicate primary key. The fix is to retry on the first failure.

Fixes #57681

### DIFF
--- a/dashboard/app/controllers/datablock_storage_controller.rb
+++ b/dashboard/app/controllers/datablock_storage_controller.rb
@@ -335,6 +335,9 @@ class DatablockStorageController < ApplicationController
 
   private def table_or_create
     where_table.first_or_create
+  rescue ActiveRecord::RecordNotUnique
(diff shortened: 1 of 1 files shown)


cnbrenci (approved):
Nice! 🚀

PR #57692, opened 2024-04-01
feat(lms): disable adding or removing co-teachers if the section is managed by lms

stephenliang:
When the roster is managed externally by an LMS, adding or removing co-teachers will result in a desynchronized state which will lead to some confusion when a roster sync occurs. Instead, inform the teacher to make updates on the LMS directly.

## Links

- [jira ticket](https://codedotorg.atlassian.net/browse/P20-801)

### Scenario 1: Go to co-teacher dashboard for LMS managed roster

Screenshot from 2024-04-01 12-03-52

Expected:
1. No co-teachers can be added
2. No co-teachers can be removed
3. i18n text for co-teacher replaced with LMS specific i18n text

### Scenario 2: Go to co-teacher dashboard for native roster (not LMS)

Expected:
1. Co-teachers can be added
2. Co-teachers can be removed
3. i18n text displays normal text

### Scenario 3: LTI Roster Sync Disabled (but is a LMS managed section)

Screenshot from 2024-04-02 14-19-45

Expected:
1. Co-teachers can be added
2. Co-teachers can be removed
3. i18n text displays normal text

## Follow-up work

N/A

## Privacy

N/A

## Security

N/A

## Caching

N/A

## PR Checklist:

- [x] Code is well-commented

### DIFF
--- a/apps/src/templates/sectionsRefresh/SectionsSetUpContainer.jsx
+++ b/apps/src/templates/sectionsRefresh/SectionsSetUpContainer.jsx
@@ -307,6 +307,10 @@ export default function SectionsSetUpContainer({
   };
 
   const renderCoteacherSection = () => {
+    const isCoTeacherManagementDisabled =
+      sections[0].primaryInstructor?.ltiRosterSyncEnabled === true &&
+      sections[0].loginType === 'ltiV1';
+
     return renderExpandableSection(
       'uitest-expandable-coteacher',
       () => (
@@ -328,7 +332,7 @@ export default function SectionsSetUpContainer({
           sectionMetricInformation={getCoteacherMetricInfoFromSection(
             sections[0]
           )}
-          loginType={sections[0].loginType}
+          disabled={isCoTeacherManagementDisabled}
         />
       ),
       isCoteacherOpen,
--- a/apps/src/templates/sectionsRefresh/SectionsSetUpContainer.jsx
+++ b/apps/src/templates/sectionsRefresh/SectionsSetUpContainer.jsx
@@ -328,6 +328,7 @@ export default function SectionsSetUpContainer({
           sectionMetricInformation={getCoteacherMetricInfoFromSection(
             sections[0]
           )}
+          loginType={sections[0].loginType}
         />
       ),
       isCoteacherOpen,
--- a/apps/src/templates/sectionsRefresh/coteacherSettings/AddCoteacher.jsx
+++ b/apps/src/templates/sectionsRefresh/coteacherSettings/AddCoteacher.jsx
@@ -93,7 +93,7 @@ export default function AddCoteacher({
   addError,
   setAddError,
   sectionMetricInformation,
-  loginType,
+  disabled,
 }) {
   const [inputValue, setInputValue] = useState('');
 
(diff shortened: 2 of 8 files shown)


apps/src/templates/sectionsRefresh/coteacherSettings/AddCoteacher.jsx
```diff
@@ -195,6 +196,10 @@ export default function AddCoteacher({
   const isMaxCoteachers = numCoteachers >= 5;
 
   const getErrorOrCount = () => {
+    if (loginType === 'ltiV1') {
```
carl-codeorg: Do you know why the value here is "ltiV1"? I was going to suggest using the value from `SectionLoginType` in `SharedConstants`, but that value is "lti_v1". Everywhere else I've seen it, it's snake case rather than camel case.

stephenliang: The controller [camelizes](https://github.com/code-dot-org/code-dot-org/blob/b1f8f0949fb08b9ba65baf7ac09d247352fa1e94/dashboard/app/controllers/sections_controller.rb#L33) the json

carl-codeorg: Oh dang, does that actually camelize the _values_? Makes me wonder if that's a bug and the intended behavior was to camelize the keys of the object before stringifying it.

Anyway, I guess this seems to fit with the way the page works. It seems to use the old sections controller instead of the "v1" controller, so we get the camelized section instead of the result from the `summarize` method that is used in other parts of the app.

apps/src/templates/sectionsRefresh/SectionsSetUpContainer.jsx
```diff
@@ -307,6 +307,10 @@ export default function SectionsSetUpContainer({
   };
 
   const renderCoteacherSection = () => {
+    const isCoTeacherManagementDisabled =
```
carl-codeorg, line 310: Nice, this also helps contain the camelized `ltiV1` string to one spot.

carl-codeorg (approved):
:shipit:

apps/src/templates/sectionsRefresh/SectionsSetUpContainer.jsx
nicklathe: I was just about to comment on trying to at least move the `ltiV1` string to a variable, but this fixes it! Nice 👍

nicklathe (approved):

apps/src/templates/sectionsRefresh/coteacherSettings/CoteacherSettings.jsx
```diff
@@ -77,7 +78,7 @@ export default function CoteacherSettings({
 
   return (
     <div className={styles.expandedSection}>
-      {i18n.coteacherAddInfo()}
+      {disabled ? i18n.coteacherLtiAddInfo() : i18n.coteacherAddInfo()}
```
lfryemason, line 81:
```suggestion
      {isLtiManagedSection ? i18n.coteacherLtiAddInfo() : i18n.coteacherAddInfo()}
```
Because `disabled` in this case changes the text to be lti specific, let's rename.

lfryemason (approved):
Looks good, one nit

cearachew (approved):

PR #57695, opened 2024-04-01
Update LTI sync summary window

carl-codeorg:
Updates the sync summary modal to include more information. Information is now in a table that shows the section owner and a teacher count. Since the course name is now shown in a separate area, the section names shown in the modal are now just the "short" section name without the course (Canvas only).

This is the first part of the task to allow users to edit the section owner during a sync. The next part will change the primary teacher column in the table to include dropdowns for changing the owner of each section.

Canvas section:
Screenshot 2024-04-01 at 12 59 07 PM

New sync confirmation window:
Screenshot 2024-04-01 at 12 59 37 PM

## Links

[JIRA](https://codedotorg.atlassian.net/browse/P20-804)
Eventual modal mockup including dropdown:
image-20240321-184515

## PR Checklist:

- [ ] Code is well-commented


apps/src/lib/ui/lti/sync/LtiSectionSyncDialog.tsx
```diff
@@ -191,6 +201,48 @@ export default function LtiSectionSyncDialog({
 
   const hideCloseButton = currentView === SubView.SPINNER;
 
+  const SectionRow = (id: string, section: LtiSection) => {
+    const sectionOwnerName = Object.entries(section.instructors).find(
+      ([id, instructor]) => instructor.isOwner
+    )?.[1]?.name;
```
stephenliang: Curious why we are looking at index `1` instead of `0`?

carl-codeorg: I think this was leftover accidentally from when I had `instructors` as an object instead of an array. I can get rid of the `Object.entries` and just do a `find`

apps/src/lib/ui/lti/sync/LtiSectionSyncDialog.tsx
```diff
@@ -202,9 +202,9 @@ export default function LtiSectionSyncDialog({
   const hideCloseButton = currentView === SubView.SPINNER;
 
   const SectionRow = (id: string, section: LtiSection) => {
-    const sectionOwnerName = Object.entries(section.instructors).find(
-      ([id, instructor]) => instructor.isOwner
-    )?.[1]?.name;
+    const sectionOwnerName = section.instructors.find(
```
stephenliang, line 201: nit/optional: if no instructor can be found it'll return the owner name as `undefined`. It'd be nice to include a default for this case, however I don't think this is a possible scenario (today at least)

apps/src/lib/ui/lti/sync/LtiSectionSyncDialog.tsx
```diff
@@ -127,28 +129,23 @@ export default function LtiSectionSyncDialog({
...
-            <li key={section_id}>
-              <b>{section.name}</b> &mdash; {studentCount}
-            </li>
-          );
-        }
-      );
+    let dialogDescription, sectionsTable;
+    if (
+      syncResult &&
+      syncResult.changed &&
```
stephenliang:
```suggestion
      syncResult?.changed &&
```

stephenliang (approved):

apps/src/lib/ui/lti/sync/LtiSectionSyncDialog.tsx
carl-codeorg: Good catch; I'll fix this in the next PR when the dropdown to select an owner is added.

nicklathe (approved):

PR #57698, opened 2024-04-01
Rework AI Chat types (part 1)

sanchitmalhotra126:
First part of reworking some of the AI Chat types to make them a bit easier to work with. Notably, this turns the `LevelAiCustomizations` type into a `LevelAiSettings` type that lists initial level customizations and visibilities separately, for easier comparison and lookup. 

This first part only updates the user-facing code (effectively anything in apps/src/aichat). The second part will complete the updates in the levelbuilder edit code (apps/lab2/levelEditors). Just splitting these apart for easier reviewing/readability, but I'll plan on merging them together to avoid breaking functionality.

## Links

https://codedotorg.atlassian.net/browse/LABS-645

## Testing story

Tested with allthethings AI Chat levels. Tested basic functionality and project saving/loading.

### DIFF
--- a/apps/src/aichat/redux/aichatRedux.ts
+++ b/apps/src/aichat/redux/aichatRedux.ts
@@ -4,7 +4,10 @@ import {LabState} from '@cdo/apps/lab2/lab2Redux';
 import Lab2Registry from '@cdo/apps/lab2/Lab2Registry';
 const registerReducers = require('@cdo/apps/redux').registerReducers;
 
-import {EMPTY_AI_CUSTOMIZATIONS_STUDENT} from '../views/modelCustomization/constants';
+import {
+  DEFAULT_VISIBILITIES,
+  EMPTY_AI_CUSTOMIZATIONS,
+} from '../views/modelCustomization/constants';
 import {initialChatMessages} from '../constants';
 import {getChatCompletionMessage} from '../chatApi';
 import {
@@ -14,13 +17,14 @@ import {
   Status,
   AiCustomizations,
   ModelCardInfo,
-  LevelAiCustomizations,
   Visibility,
+  LevelAiChatSettings,
 } from '../types';
+import {RootState} from '@cdo/apps/types/redux';
 
 const haveDifferentValues = (
-  value1: string[] | string | number | ModelCardInfo,
-  value2: string[] | string | number | ModelCardInfo
+  value1: AiCustomizations[keyof AiCustomizations],
+  value2: AiCustomizations[keyof AiCustomizations]
 ): boolean => {
   if (typeof value1 === 'object' && typeof value2 === 'object') {
     return JSON.stringify(value1) !== JSON.stringify(value2);
@@ -39,12 +43,8 @@ const findChangedProperties = (
 
   const changedProperties: string[] = [];
   Object.keys(next).forEach(key => {
-    if (
-      haveDifferentValues(
-        previous[key as keyof AiCustomizations],
-        next[key as keyof AiCustomizations]
-      )
-    ) {
+    const typedKey = key as keyof AiCustomizations;
+    if (haveDifferentValues(previous[typedKey], next[typedKey])) {
       changedProperties.push(key);
     }
   });
@@ -66,6 +66,7 @@ export interface AichatState {
   chatMessageError: boolean;
   currentAiCustomizations: AiCustomizations;
   previouslySavedAiCustomizations?: AiCustomizations;
+  fieldVisibilities: {[key in keyof AiCustomizations]: Visibility};
 }
 
 const initialState: AichatState = {
@@ -73,7 +74,8 @@ const initialState: AichatState = {
   isWaitingForChatResponse: false,
   showWarningModal: true,
   chatMessageError: false,
-  currentAiCustomizations: EMPTY_AI_CUSTOMIZATIONS_STUDENT,
+  currentAiCustomizations: EMPTY_AI_CUSTOMIZATIONS,
+  fieldVisibilities: DEFAULT_VISIBILITIES,
 };
 
 // THUNKS
@@ -84,7 +86,7 @@ const initialState: AichatState = {
 export const updateAiCustomization = createAsyncThunk(
   'aichat/updateAiCustomization',
   async (_, thunkAPI) => {
-    const state = thunkAPI.getState() as {lab: LabState; aichat: AichatState};
+    const state = thunkAPI.getState() as RootState;
     const {currentAiCustomizations, previouslySavedAiCustomizations} =
       state.aichat;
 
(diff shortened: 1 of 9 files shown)


bencodeorg (approved):
So much better, thanks for doing this!

apps/src/aichat/types.ts
```diff
@@ -67,6 +68,13 @@ export enum Visibility {
   EDITABLE = 'editable',
 }
 
+export interface LevelAiChatSettings {
```
fisher-alice: Could you add a comment here explaining this as compared to `LevelAiCustomizations` and why we have both?

apps/src/lab2/levelEditors/aiCustomizations/EditAiCustomizations.tsx
```diff
@@ -51,7 +51,7 @@ const EditAiCustomizations: React.FunctionComponent<{
 }> = ({initialCustomizations}) => {
   const [aiCustomizations, setAiCustomizations] =
     useState<LevelAiCustomizations>(
-      initialCustomizations || EMPTY_AI_CUSTOMIZATIONS
+      initialCustomizations || EMPTY_AI_LEVEL_CUSTOMIZATIONS
     );
 
   const setPropertyVisibility = useCallback(
```
fisher-alice, line 57: This is not within scope of PR but just curious about  `setPropertyValue` on line 71 below. I see that the type for `value` is `string | number | string[]`. But in `UpdateContext.ts`, the type for `setPropertyValue` is `string | number`. Is there a reason why the types are different?
Also, would it be okay to use `value: AiCustomizations[keyof AiCustomizations]` instead or did you not do this since the model card is updated by a different function?

apps/src/aichat/redux/aichatRedux.ts
```diff
@@ -84,7 +86,7 @@ const initialState: AichatState = {
 export const updateAiCustomization = createAsyncThunk(
   'aichat/updateAiCustomization',
   async (_, thunkAPI) => {
-    const state = thunkAPI.getState() as {lab: LabState; aichat: AichatState};
+    const state = thunkAPI.getState() as RootState;
```
fisher-alice, line 89: Do you mind explaining how `RootState` is assigned correctly here?

apps/src/aichat/views/modelCustomization/constants.ts
```diff
@@ -38,10 +39,25 @@ export const EMPTY_AI_CUSTOMIZATIONS: LevelAiCustomizations = {
...
+  [key in keyof AiCustomizations]: Visibility;
+} = {
+  botName: Visibility.EDITABLE,
+  temperature: Visibility.EDITABLE,
+  systemPrompt: Visibility.EDITABLE,
+  retrievalContexts: Visibility.EDITABLE,
+  modelCardInfo: Visibility.EDITABLE,
+};
+
+export const EMPTY_LEVEL_AI_CHAT_SETTINGS: LevelAiChatSettings = {
```
fisher-alice: Same question as above - `DEFAULT_LEVEL_AI_CHAT_SETTINGS`?

apps/src/aichat/types.ts
fisher-alice: We've been using lowercase in `aichat` for other names like `AichatLevelProperties`. Should we be consistent and use `LevelAichatSettings`?

apps/src/aichat/types.ts
```diff
@@ -67,6 +68,13 @@ export enum Visibility {
...
+  initialCustomizations: AiCustomizations;
+  visibilities: {[key in keyof AiCustomizations]: Visibility};
+  /** If the presentation panel is hidden from the student. */
+  hidePresentationPanel?: boolean;
+}
+
 /**
  * Level-defined AI customizations for student chat bots set by levelbuilders on the level's properties.
  * Each field is the same as AiCustomizations, but with an additional visibility property.
```
fisher-alice, line 84: Nit question: why was `type` used vs `interface` below on line 82?

apps/src/aichat/views/modelCustomization/constants.ts
```diff
@@ -26,7 +27,7 @@ export const EMPTY_MODEL_CARD_INFO: ModelCardInfo = {
   exampleTopics: [],
 };
 
-export const EMPTY_AI_CUSTOMIZATIONS: LevelAiCustomizations = {
+export const EMPTY_AI_LEVEL_CUSTOMIZATIONS: LevelAiCustomizations = {
```
fisher-alice, line 29: What do you think about naming this `DEFAULT_AI_LEVEL_CUSTOMIZATIONS since it includes default values for visibility?

fisher-alice (approved):
Awesome refactor! Just left a couple questions for my understanding and a couple renaming suggestions. Thanks!

apps/src/aichat/types.ts
sanchitmalhotra126: Will add a comment - `LevelAiCustomizations` will go away in the following change (it's here for now since the level editor files still reference it)

apps/src/lab2/levelEditors/aiCustomizations/EditAiCustomizations.tsx
sanchitmalhotra126: Oh yep, just a miss on my part 🙂 when I update the editor files in the next PR, I'll use `AiCustomizations[keyof AiCustomizations]` as well

apps/src/aichat/redux/aichatRedux.ts
sanchitmalhotra126: Yep, `RootState` comes from here: https://github.com/code-dot-org/code-dot-org/blob/95f488ffef1fe252a86552c4b3586d7c8fa3f0ab/apps/src/types/redux.ts#L23

This describes the type of the global state object, for all redux slices that have been migrated to typescript. In theory if all of our redux code was in TypeScript, this RootState type would completely describe the global state, but for now we can use this in TypeScript redux files at least. Also the previous approach was fine, I was just experimenting with a way to type the `thunkAPI` implicitly, but found that the redux-toolkit docs recommend this approach: https://redux-toolkit.js.org/usage/usage-with-typescript#typing-async-thunks-inside-createslice

apps/src/aichat/types.ts
sanchitmalhotra126: I used `type` to do the inline union (`&`) between the part of the definition that describes the keys of `AiCustomizations` mapped to `value` and `visibility`, and the `hidePresentationPanel` field. I'm actually not sure how to do the same thing with an `interface` so just went with `type`.

Edit: also not sure if the `[key in keyof AiCustomizations]: ...` syntax is valid in interfaces? Or at least I wasn't able to get it to work 🙂

apps/src/aichat/views/modelCustomization/constants.ts
sanchitmalhotra126: This will be deleted in the following PR, but will rename the other constant

apps/src/aichat/types.ts
fisher-alice: > Will add a comment - LevelAiCustomizations will go away in the following change (it's here for now since the level editor files still reference it)

Oh cool! I was wondering about that. Thanks so much for the upcoming refactor on the levelbuilder end.

apps/src/aichat/types.ts
fisher-alice: Oh interesting - thanks for the explanation!

apps/src/aichat/redux/aichatRedux.ts
fisher-alice: Thanks so much for the explanation! Still wrapping my head around redux-toolkit.

PR #57700, opened 2024-04-01
Fix bug where site crashes on clever login

lfryemason:
On clever login, there was a race condition where `window.innerWidth` was returning 0, causing a null pointer exception in `responsiveRedux`. 

This PR returns `xs` ResponsiveSize if we cannot determine the window width

## Links
[Slack thread](https://codedotorg.slack.com/archives/C045UAX4WKH/p1712002341313539)

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/code-studio/responsiveRedux.js
+++ b/apps/src/code-studio/responsiveRedux.js
@@ -17,7 +17,14 @@ const Breakpoints = [
 ];
 
 export function getResponsiveBreakpoint(width) {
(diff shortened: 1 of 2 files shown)


stephenliang (approved):
A future improvement could be to tie this to a resize event listener so we update breakpoints based on the `innerWidth` changing. Thanks for the quick turnaround on this!

PR #57705, opened 2024-04-01
Create ActionBlocksWrapper component

kelbyhawn:
Creates an `ActionBlocksWrapper` component that uses a `OneColumnActionBlock` component to create two and three column action block sections for use on pages on studio.code.org. This is the React implementation of the `.action-block__wrapper` classes on code.org ([ref](https://docs.google.com/document/d/1wev8I2BQVViNXPuhLEXMpIAYSoAJb0kVoSRKaI3zEAk/edit#heading=h.ncdwn07vxw41)).
- Columns are dynamically set between two and three columns with CSS based on if there are three or more tile objects
- Fully responsive
- Works with LTR and RTL languages 

**Jira ticket:** [ACQ-1704](https://codedotorg.atlassian.net/browse/ACQ-1704)

----

## Storybook

Screenshot 2024-04-02 at 8 14 50 AM

Screenshot 2024-04-02 at 8 15 01 AM

## RTL

RTL

## Responsive

https://github.com/code-dot-org/code-dot-org/assets/9256643/d1710c29-9635-4d4f-b4a7-99d585dce401

### DIFF
--- a/apps/src/templates/studioHomepages/ActionBlocksWrapper.jsx
+++ b/apps/src/templates/studioHomepages/ActionBlocksWrapper.jsx
@@ -69,6 +69,8 @@ export default function ActionBlocksWrapper({actionBlocks}) {
       className={
         // This automatically adjusts the column count CSS
         // based on the number of blocks in an array.
+        // If there are 3 or more, it'll use a 3-column layout.
+        // For example: 6 blocks would result in a 3x2 grid.
         actionBlocks?.length >= 3
           ? styles.wrapperThreeCol
           : styles.wrapperTwoCol
--- a/apps/src/templates/studioHomepages/ActionBlocksWrapper.jsx
+++ b/apps/src/templates/studioHomepages/ActionBlocksWrapper.jsx
@@ -26,7 +26,7 @@ const OneColumnActionBlock = ({
         {heading && (
           <Heading2 visualAppearance="heading-md">{heading}</Heading2>
         )}
-        {imageUrl && <img src={imageUrl} alt="" className={styles.image} />}
+        {imageUrl && <img src={imageUrl} alt="" />}
         {description && <BodyThreeText>{description}</BodyThreeText>}
       </div>
       <div className={styles.buttonWrapper}>
--- a/apps/src/templates/studioHomepages/ActionBlocksWrapper.jsx
+++ b/apps/src/templates/studioHomepages/ActionBlocksWrapper.jsx
@@ -1,0 +1,93 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+import {LinkButton} from '@cdo/apps/componentLibrary/button';
+import {
+  Heading2,
+  OverlineTwoText,
+  BodyThreeText,
+} from '@cdo/apps/componentLibrary/typography';
+import styles from './actionBlocksWrapper.module.scss';
+
+const OneColumnActionBlock = ({
+  overline,
+  imageUrl,
+  heading,
+  description,
+  buttons,
+}) => {
+  return (
+    <div className={styles.oneColumnActionBlock}>
+      <div className={styles.contentWrapper}>
+        {overline && (
+          <OverlineTwoText className={styles.overline}>
+            {overline}
+          </OverlineTwoText>
+        )}
+        {heading && (
+          <Heading2 visualAppearance="heading-md">{heading}</Heading2>
+        )}
+        {imageUrl && <img src={imageUrl} alt="" className={styles.image} />}
+        {description && <BodyThreeText>{description}</BodyThreeText>}
+      </div>
+      <div className={styles.buttonWrapper}>
+        {buttons &&
+          buttons.map((button, index) => (
+            <LinkButton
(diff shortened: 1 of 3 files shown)


apps/src/templates/studioHomepages/ActionBlocksWrapper.jsx
```diff
@@ -0,0 +1,93 @@
...
+import React from 'react';
+import {LinkButton} from '@cdo/apps/componentLibrary/button';
+import {
+  Heading2,
+  OverlineTwoText,
+  BodyThreeText,
+} from '@cdo/apps/componentLibrary/typography';
+import styles from './actionBlocksWrapper.module.scss';
+
+const OneColumnActionBlock = ({
```
kelbyhawn, line 11: I included this in the same file since I don't think these will be used outside of the wrapper anywhere. If we decide to use Swiper that may change, but I can refactor if needed.

apps/src/templates/studioHomepages/ActionBlocksWrapper.story.jsx
```diff
@@ -0,0 +1,44 @@
...
+
+export default {
+  component: ActionBlocksWrapper,
+};
+
+const generateActionBlocks = tileCount => {
+  const actionBlocks = [];
+  for (let i = 0; i < tileCount; i++) {
+    actionBlocks.push({
+      overline: `Overline ${i + 1}`,
```
bethanyaconnor, line 12: Thank you for using the prop name in the text here! I wasn't sure what an overline was and it was so useful to see the screenshot and immediately understand!

apps/src/templates/studioHomepages/ActionBlocksWrapper.jsx
```diff
@@ -0,0 +1,93 @@
...
+              type={button.type}
+            />
+          ))}
+      </div>
+    </div>
+  );
+};
+
+OneColumnActionBlock.propTypes = {
+  overline: PropTypes.string,
```
bethanyaconnor, line 50: what, if any, of these are required out of curiousity? ie could this component be used without an image (for example)?

apps/src/templates/studioHomepages/ActionBlocksWrapper.jsx
```diff
@@ -0,0 +1,93 @@
...
+        {heading && (
+          <Heading2 visualAppearance="heading-md">{heading}</Heading2>
+        )}
+        {imageUrl && <img src={imageUrl} alt="" />}
+        {description && <BodyThreeText>{description}</BodyThreeText>}
+      </div>
+      <div className={styles.buttonWrapper}>
+        {buttons &&
+          buttons.map((button, index) => (
+            <LinkButton
```
bethanyaconnor, line 35: 😍 for using the new `LinkButton` component!

apps/src/templates/studioHomepages/ActionBlocksWrapper.jsx
```diff
@@ -0,0 +1,93 @@
...
+  ),
+};
+
+export default function ActionBlocksWrapper({actionBlocks}) {
+  return (
+    <div
+      className={
+        // This automatically adjusts the column count CSS
+        // based on the number of blocks in an array.
+        actionBlocks?.length >= 3
```
bethanyaconnor, line 74: so we only support two or three action blocks? (that's fine, just making sure I understand)

If so, can you add a comment above this component stating that?

bethanyaconnor (approved):
Looks great! Just a couple quick questions

apps/src/templates/studioHomepages/ActionBlocksWrapper.jsx
kelbyhawn: I was so pumped about this too! (Thank you @levadadenys! ❤️ )

apps/src/templates/studioHomepages/ActionBlocksWrapper.jsx
kelbyhawn: I'll update the comment, but we can hypothetically have any number we want 🙌  That logic just breaks up the grid into three columns once there's three or more. For example, it'll look like the screenshot below if you have 6 — I'll add this example to the Storybook too!
Screenshot 2024-04-02 at 12 36 19 PM

apps/src/templates/studioHomepages/ActionBlocksWrapper.jsx
kelbyhawn: Correct! Most will have at least a heading, image, description, and button, but I left off requirements to increase the flexibility. For example, we have some on Pegasus that just have a heading, description, and button.

apps/src/templates/studioHomepages/ActionBlocksWrapper.jsx
bethanyaconnor: right `>= 3` does not mean `== 3` 🤦‍♀️ thanks for updating the comment -- this looks great!

bethanyaconnor (approved):

PR #57714, opened 2024-04-02
Validate emails before adding MailJet contact

bethanyaconnor:
MailGun, a related product to MailJet, offers an API to validate email addresses and out technical account manager suggested we use this functionality. This checks to make sure the emails we add to MailJet are at least real addresses and that they're not known spammy addresses.

This PR adds the MailGun gem and utilizes this API. More info on this API is [here](https://documentation.mailgun.com/docs/inboxready/mailgun-validate/single-valid-ir/).

This also turned into a bit of a refactor to ensure the email validation is only checked once; when I first wrote it, I was checking email validation on both contact creation and email send. I decided to do this by making it so emails must be sent to MailJet contacts, even though that's not strictly necessary. Definitely open to feedback here!


lib/cdo/mailjet.rb
```diff
@@ -16,7 +18,8 @@ def self.enabled?
     DCDO.get('use_mailjet', false) &&
       !Rails.env.test? &&
       API_KEY.present? &&
-      SECRET_KEY.present?
+      SECRET_KEY.present? &&
+      MAILGUN_API_KEY.present?
```
hannahbergam, line 22: Now when I see api keys I'm called to ask: have we tested that the key is correctly configured? I'm sorry if this has already been tested elsewhere and I'm late to the game!

lib/cdo/mailjet.rb
```diff
@@ -57,29 +61,33 @@ def self.create_contact(email, name, sign_up_date)
...
-          value: sign_up_date_rfc3339
+          name: field_name,
+          value: field_value
         }
       ]
     )
   end
 
-  def self.send_template_email(to_email, to_name, email_config)
+  def self.send_template_email(contactdata, email_config)
```
hannahbergam: Is there any world in which this returns an error?

lib/test/cdo/test_mailjet.rb
```diff
@@ -6,23 +6,22 @@ def setup
     MailJet.stubs(:enabled?).returns(true)
   end
 
-  def test_create_contact
-    email = 'fake.email@email.com'
+  def test_create_contactdata
+    email = 'fake.email@test.xx'
```
hannahbergam: 🎉

lib/test/cdo/test_mailjet.rb
```diff
@@ -65,6 +84,42 @@ def test_send_template_email
...
+    Mailgun::Address.any_instance.expects(:validate).returns({'result' => 'deliverable'})
+    assert MailJet.valid_email?('test@email.com')
+  end
+
+  def test_valid_email_undeliverable
+    Mailgun::Address.any_instance.expects(:validate).returns({'result' => 'undeliverable'})
+    refute MailJet.valid_email?('test@email.com')
+  end
+
+  def test_valid_email_do_not_send
```
hannahbergam, line 126: Thanks for all these tests!

hannahbergam (commented):
This seems good to me- I left a few small questions!

lib/cdo/mailjet.rb
bethanyaconnor: no it's a good question! I've actually been checking in Secrets Manager for each key every time I update one, particularly for the exact issue that hit us with Statsig 😅 Obviously, testing that the key is definitely correct (ie I didn't copy all but the last character or something) is harder to test before this hits products, but, as this is behind a DCDO flag, hopefully it would be a short outage if there is one.

lib/cdo/mailjet.rb
bethanyaconnor: I'm trying to remember -- I actually had more issues with this failing silently instead of throwing errors. Do you think this should be wrapped in exception handling in some way? (Might also be feedback for me to incorporate into https://github.com/code-dot-org/code-dot-org/pull/58099)

bethanyaconnor:
@hannahbergam I found a bug when testing on the adhoc and updated this PR with that fix in [69c2e61](https://github.com/code-dot-org/code-dot-org/pull/57714/commits/69c2e61adab978b58f6cb68536c6db5526a10e51)

hannahbergam (approved):
Good find! So I'm following- it sounds like MailJet has two different concepts for a 'Contact' and that contact's 'ContactData' (and prior to this commit, those were combined in ContactData).  Now I'm curious about all the data that could live in 'data' that wouldn't be represented in 'Contact', but that's an aside! LGTM

bethanyaconnor:
Yeahhh, it seems like `Contact` is a really simple data structure with name and email and `Contactdata` is all the custom stuff. Some more info in [their documentation](https://dev.mailjet.com/email/guides/contact-management/).

PR #55993, opened 2024-01-26
[Minecraft] Prepare for Google Blockly Migration

mikeharv:
This branch offers a reasonably small number of changes that should effectively make Minecraft ready to test for migration to Google Blockly! 🎉 

There are quite a few files in this branch, but the changes are (hopefully) each fairly small and reasonable.

### Translations
The `repeat n times` block relies on translations that currently live in the CDO Blockly repo. I added new translations to our common locales. Since we might need to use either version of Blockly for now, I opted for setting the label like `.appendField(blocklyTranslation || commonTranslation)` since this would prioritize the Blockly repo's translations for CDO Blockly and give time for CrowdIn to catch up so we have fewer translation regressions on this block after flipping.

### Styles
In order for blocks to be compatible with our various Google Blockly themes, they need to have a set style property. CDO Blockly however uses a static HSV color value. Fortunately, Google Blockly prioritizes the style if it's present, so there's no harm in assigning both.
`Block.setStyle` doesn't normally exist on CDO Blockly, so I patched in the wrapper so that it doesn't cause any errors when called. I had previously added some pretty clunky logic where we were checking for the existence of this function to determine which Blockly version was in use. As a result of it now existing in both places, I had to clean up a couple of these places, such as in [`block_utils`](https://github.com/code-dot-org/code-dot-org/pull/55993/commits/7ee9d5ffdb738c0e2a52fa5378b74847c76e7112#diff-4204ee82808ca5d21acf4aef2f991c9674c29b599070f1c7aedc237c3ec3b461).
I needed to add styles to all the craft blocks (except for Aquatic, which were already completed year https://github.com/code-dot-org/code-dot-org/pull/50872), and some "common" blocks (e.g. repeat loops) shared across labs. Since I was already in the [`blocksCommon.js`](https://github.com/code-dot-org/code-dot-org/pull/55993/commits/7ee9d5ffdb738c0e2a52fa5378b74847c76e7112#diff-4ea1f53fc8a9f11e3683471951371663e4649e501aa020c9c06051be88db3ba8) file, I added styles for each block here, some of which are used more in Maze/Artist etc.

### Generators
For the v10 upgrade, we had to address a change that moves generator functions to a new `generator.forBlock` object. For our common blocks, I added some helper functions to the `customBlocks/*/commonBlocks.js` files that allow us to handle cases where we would set one block code generator to another, or assign a new function as a block code generator. 

[...]

PR #57719, opened 2024-04-02
[Google Blockly] Generate code for block array

mikeharv:
Some CDO Blockly labs (Maze, Artist, Play Lab) need to be able to generate code for an arbitrary list of blocks via `blocksToCode`. This could have a range of purposes beyond just running the student's code, such as running the solution blocks in Artist or initialization blocks in Artist or Play Lab.
Note: `blocksToCode` _used to be_ part of the Minecraft code, but because we only ever needed to generate code from the main workspace, we were able to safely replace it with our custom `Blockly.getWorkspaceCode()`. https://github.com/code-dot-org/code-dot-org/pull/55993/commits/3c8902bd93dd86fc183976d7abf178ed19987f91

Because `blocksToCode` is used in other labs that we intend to migrate, and it won't always just be looking at workspace blocks, it makes to actually just support it. Like the CDO Blockly version, the new Google Blockly generator function will iterate through the list using `blockToCode`. 
CDO Blockly code: https://github.com/code-dot-org/blockly/blob/f012d8262f21bae3e54fb11dd8bc29cf0d29f3cd/core/code_generation/generator.js#L60-L99

This change makes it possible to run the student code in the first level of our Maze HOC progression:

**Before:**
image

**After:**
image

While working on this, I also made some changes to `blockSpaceToCode`. This function works almost the exact same way except the block array is always the workspace's top blocks. I refactored `blockSpaceToCode` to call the new `blocksToCode` using that list. I was also curious about its optional `opt_typeFilter` parameter. This was added to CDO Blockly for the deprecated Eval lab (a Play Lab descendent). Since we are not going to migrate Eval, it save to remove the parameter now.

## Links

Jira: https://codedotorg.atlassian.net/browse/CT-465

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/blockly/addons/cdoGenerator.ts
+++ b/apps/src/blockly/addons/cdoGenerator.ts
@@ -22,7 +22,7 @@ export default function initializeGenerator(
   // This function was a custom addition in CDO Blockly, so we need to add it here
   // so that our code generation logic still works with Google Blockly
   blocklyWrapper.Generator.blockSpaceToCode = function (name) {
-    let blocksToGenerate = blocklyWrapper.mainBlockSpace.getTopBlocks(
+    const blocksToGenerate = blocklyWrapper.mainBlockSpace.getTopBlocks(
       true /* ordered */
     );
     return blocklyWrapper.Generator.blocksToCode(name, blocksToGenerate);
--- a/apps/src/blockly/addons/cdoGenerator.ts
+++ b/apps/src/blockly/addons/cdoGenerator.ts
@@ -21,20 +21,25 @@ export default function initializeGenerator(
 
   // This function was a custom addition in CDO Blockly, so we need to add it here
   // so that our code generation logic still works with Google Blockly
-  blocklyWrapper.Generator.blockSpaceToCode = function (name, opt_typeFilter) {
-    const generator = blocklyWrapper.getGenerator();
-    generator.init(blocklyWrapper.mainBlockSpace);
+  blocklyWrapper.Generator.blockSpaceToCode = function (name) {
     let blocksToGenerate = blocklyWrapper.mainBlockSpace.getTopBlocks(
       true /* ordered */
     );
-    if (opt_typeFilter) {
-      if (typeof opt_typeFilter === 'string') {
-        opt_typeFilter = [opt_typeFilter];
-      }
-      blocksToGenerate = blocksToGenerate.filter(block =>
-        (opt_typeFilter as string[]).includes(block.type)
+    return blocklyWrapper.Generator.blocksToCode(name, blocksToGenerate);
+  };
+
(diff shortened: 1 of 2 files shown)


apps/src/blockly/addons/cdoGenerator.ts
```diff
@@ -21,20 +21,25 @@ export default function initializeGenerator(
...
-        (opt_typeFilter as string[]).includes(block.type)
+    return blocklyWrapper.Generator.blocksToCode(name, blocksToGenerate);
+  };
+
+  // Used to generate code for an array of top blocks.
+  blocklyWrapper.Generator.blocksToCode = function (
+    name: string,
+    blocksToGenerate: Block[]
+  ) {
+    if (name !== 'JavaScript') {
```
molly-moen, line 44: is `name` a cdo blockly parameter that we need to support still? It doesn't seem like it's used beyond this check.

molly-moen (approved):
Nice!

apps/src/blockly/addons/cdoGenerator.ts
mikeharv: Every call to `blocksToCode` in the main repo does pass `'JavaScript'` in as the first argument.

In theory, we'd use it to specify which generator use (JavaScript, Python, etc.). We've only ever supported JavaScript with Blockly, but presumably that could change?

I wasn't planning to change the function's signature here though since it needs to keep working as is for CDO Blockly.

mikeharv:
@molly-moen I had to re-add the optional parameter for filtering because we were actually using it for Flappy as well.
This is the case I missed when searching for uses of `blockSpaceToCode`: https://github.com/code-dot-org/code-dot-org/blob/mike/blocksToCode/apps/src/flappy/flappy.js#L812-L823

```
  // Map event handler hooks (e.g. Flappy.whenClick) to the generated code.
  const generator = Blockly.Generator.blockSpaceToCode.bind(
    Blockly.Generator,
    'JavaScript'
  );
  const events = {
    whenClick: {code: generator('flappy_whenClick')},
    whenCollideGround: {code: generator('flappy_whenCollideGround')},
    whenEnterObstacle: {code: generator('flappy_whenEnterObstacle')},
    whenCollideObstacle: {code: generator('flappy_whenCollideObstacle')},
    whenRunButton: {code: generator('when_run')},
  };
```

I missed that weren't simply calling it but using `.bind` to set the function as `generator`. Then when that "new" function was called it did have block type strings passed.

Here's the revert commit: https://github.com/code-dot-org/code-dot-org/pull/57719/commits/bd590e7fb673d16bf39b03f8b4f64b6c40252ddd

Edit: For additional context, it looks like Flappy needs to do this so we don't run any code until the student clicks the app after pressing run. Here's what happens if we don't filter and end up running everything immediately:

image

PR #57732, opened 2024-04-02
Revert "Provision new Redis-based Session Store"

Hamms:
Reverts code-dot-org/code-dot-org#57558; prepping this PR just in case something goes wrong with this change that currently only applies to production.


Hamms:
The change was deployed without any issue; closing this PR as it's no longer necessary.

PR #57734, opened 2024-04-02
Adds event logging to existing school association flow

hannahbergam:
This PR adds amplitude and statsig events to the existing school association flow. These events won't perfectly map to the new flow because the component works differently, but we will still be able to log a few of those fields the same way. 

I'm adding the constants for the fields I'm about to set up (because I was reading from the product spec and figured why not). The remainder of them will be utilized in soon-to-come prs!

I also updated the label on the user type field to be more consistent after looking through the metrics dashboards with @dmcavoy . 

One last thing- I removed an unused enabledExperiments param that I already removed in [this pr](https://github.com/code-dot-org/code-dot-org/pull/57481) and just hadn't implemented in this updateUser call yet. 

Here are the three events:
Screenshot 2024-04-03 at 12 47 29 PM
Screenshot 2024-04-03 at 12 47 58 PM
Screenshot 2024-04-03 at 12 52 57 PM

## Links

- spec: []()
- jira ticket: https://codedotorg.atlassian.net/browse/ACQ-1709
- https://codedotorg.atlassian.net/browse/ACQ-1712

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/sites/studio/pages/devise/registrations/_finish_sign_up.js
+++ b/apps/src/sites/studio/pages/devise/registrations/_finish_sign_up.js
@@ -275,11 +275,15 @@ $(document).ready(() => {
 
   function onSchoolChange(_, event) {
     schoolData.ncesSchoolId = event ? event.value : '';
-    analyticsReporter.sendEvent(
-      EVENTS.SCHOOL_SELECTED_FROM_LIST,
-      {ncesId: schoolData.ncesSchoolId},
-      PLATFORMS.BOTH
-    );
+    // ID is set to -1 and '' respectively as user toggles 'I cannot find my
+    // school above': exlude these from school selection events
+    if (!['-1', ''].includes(schoolData.ncesSchoolId)) {
+      analyticsReporter.sendEvent(
+        EVENTS.SCHOOL_SELECTED_FROM_LIST,
+        {ncesId: schoolData.ncesSchoolId},
+        PLATFORMS.BOTH
+      );
+    }
     renderSchoolInfo();
   }
 
--- a/apps/src/sites/studio/pages/devise/registrations/_finish_sign_up.js
+++ b/apps/src/sites/studio/pages/devise/registrations/_finish_sign_up.js
@@ -6,7 +6,7 @@ import getScriptData from '@cdo/apps/util/getScriptData';
 import firehoseClient from '@cdo/apps/lib/util/firehose';
 import experiments from '@cdo/apps/util/experiments';
 import analyticsReporter from '@cdo/apps/lib/util/AnalyticsReporter';
-import {EVENTS} from '@cdo/apps/lib/util/AnalyticsConstants';
+import {EVENTS, PLATFORMS} from '@cdo/apps/lib/util/AnalyticsConstants';
 
 const TEACHER_ONLY_FIELDS = [
   '#teacher-name-label',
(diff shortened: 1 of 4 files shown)


apps/src/sites/studio/pages/devise/registrations/_finish_sign_up.js
```diff
@@ -270,6 +275,15 @@ $(document).ready(() => {
 
   function onSchoolChange(_, event) {
     schoolData.ncesSchoolId = event ? event.value : '';
+    // ID is set to -1 and '' respectively as user toggles 'I cannot find my
```
TurnerRiley, line 278: What does it mean here that it's set to -1 and '' respectively for the singular 'I cannot find my school above'?

apps/src/sites/studio/pages/devise/registrations/_finish_sign_up.js
```diff
@@ -270,6 +275,15 @@ $(document).ready(() => {
...
+    // school above': exlude these from school selection events
+    if (!['-1', ''].includes(schoolData.ncesSchoolId)) {
+      analyticsReporter.sendEvent(
+        EVENTS.SCHOOL_SELECTED_FROM_LIST,
+        {ncesId: schoolData.ncesSchoolId},
+        PLATFORMS.BOTH
+      );
+    }
```
TurnerRiley, line 286: Would it work to have an else case here that logs the `ADD_MANUALLY_CLICKED` instead of in `SchoolAutocompleteDropdownWithLabel.jsx`? (not sure if that's possible/makes more sense or not)

TurnerRiley (approved):
Overall, looks great! Just had 2 quick questions

apps/src/sites/studio/pages/devise/registrations/_finish_sign_up.js
hannahbergam: I could see how that might make sense! However, I think the school ID getting updated is kind of a byproduct of clicking that checkbox, and I like having the event tied more directly to that checkbox in the codebase so someone reading through would be able to follow what the event is for. I think putting it here might be confusing

apps/src/sites/studio/pages/devise/registrations/_finish_sign_up.js
hannahbergam: I wasn't sure how better to say this so maybe you can help! I am looking for any nces id update that isn't -1 or an empty string (hence the !includes below) because that indicates a *real school selection as opposed to a byproduct of clicking 'I cannot find my school above'.

TurnerRiley: Oh I see gotcha! I think I read it as selecting 'I cannot find my school' can either set the ID to -1 or ''. Maybe if the comment could specify that 'I cannot find my school above' maps to -1 and setting it to the default empty maps to '' (if I'm understanding this right)?

apps/src/sites/studio/pages/devise/registrations/_finish_sign_up.js
TurnerRiley: That makes sense! Sounds good!

apps/src/sites/studio/pages/devise/registrations/_finish_sign_up.js
hannahbergam: I'm going to update this comment in a followup. Hoping to merge this and get the new events logging before EOD:)

PR #57743, opened 2024-04-02
P20-829: Prepare Units only when needed on the Dashboard test run

artem-vavilov:
| Before  | After |
| ------- | ----- |
| before | after |

## Links
- JIRA ticket [P20-829](https://codedotorg.atlassian.net/browse/P20-829)
- Related PR: https://github.com/code-dot-org/code-dot-org/pull/57661


Hamms (approved):
This is great! Constrain some deprecated functionality to only the parts of the codebase that actually still rely on it, and make sure that future devs know not to go spreading it around.

And all in the service of such a significant time savings, too! I'm genuinely looking forward to running some tests locally once all your improvements make it in. Thank you!

PR #57744, opened 2024-04-02
Rework AI Chat types (part 2)

sanchitmalhotra126:
Follow up to https://github.com/code-dot-org/code-dot-org/pull/57698. This refactors code in the levelbuilder edit files to use the new `LevelAiSettings` type and deletes the now unused `LevelAiCustomizations` type. Also updates naming from `ai_customizations` to `aichat_settings` for consistency.

## Links

https://codedotorg.atlassian.net/browse/LABS-645

## Testing story

Tested by locally editing existing allthethings levels and creating a new AI Chat level, and confirmed that behavior remains the same.


apps/src/lab2/levelEditors/aichatSettings/EditAichatSettings.tsx
```diff
@@ -214,10 +197,10 @@ const EditAiCustomizations: React.FunctionComponent<{
               </label>
               <Checkbox
                 name="hidePresentationPanel"
-                checked={aiCustomizations.hidePresentationPanel || false}
+                checked={aichatSettings.hidePresentationPanel || false}
```
fisher-alice: Nit: just curious - is there ever a case when this optional property is undefined since it's defined in `DEFAULT_LEVEL_AICHAT_SETTINGS`?

apps/src/lab2/levelEditors/aichatSettings/ModelCardFields.tsx
```diff
@@ -1,16 +1,17 @@
...
+import moduleStyles from './edit-aichat-settings.module.scss';
 import {
   MAX_ASK_ABOUT_TOPICS,
   MODEL_CARD_FIELDS_AND_LABELS,
 } from '@cdo/apps/aichat/views/modelCustomization/constants';
 import {UpdateContext} from './UpdateContext';
 
 const ModelCardFields: React.FunctionComponent = () => {
-  const {setModelCardPropertyValue, aiCustomizations} =
+  const {setModelCardPropertyValue, aichatSettings: aiCustomizations} =
```
fisher-alice: Does this create an alias for `aichatSettings`? If so, could you explain why you did this instead of just using `aichatSettings`? If not, could you explain what this does?

fisher-alice (approved):
🎉 Just a couple questions for my own understanding. Such an improvement - thanks!

apps/src/lab2/levelEditors/aichatSettings/EditAichatSettings.tsx
sanchitmalhotra126: Oh nope, there shouldn't be, I can remove this. Edit: actually I forgot - we do need this because `hidePresentationPanel` is an optional property, and `checked` must be a boolean (cannot be undefined)

apps/src/lab2/levelEditors/aichatSettings/ModelCardFields.tsx
sanchitmalhotra126: Oh good catch, this is just something I forgot to clean up after renaming variables. Will fix!

apps/src/lab2/levelEditors/aichatSettings/EditAichatSettings.tsx
fisher-alice: Thanks! Yes, I was curious why `hidePresentationPanel` is optional.

sanchitmalhotra126: Oh gotcha, yeah good question I suppose it doesn't have to be. Will update!

PR #57745, opened 2024-04-02
Fix Violations of and Reenable `Style/ExplicitBlockArgument`

Hamms:
> Enforces the use of explicit block argument to avoid writing block literal that just passes its arguments to another block.

Fix applied automatically with `bundle exec rubocop --autocorrect-all --only Style/ExplicitBlockArgument`

I don't have a strong feeling about this one; I generally agree that explicit is better than implicit, but on the other hand I think recognizing what `&block` is doing takes more Ruby-specific knowledge than `yield`. I'm inclined to say `Style/` Cops should be enabled unless we have a good reason not to, though, so I'm open to hearing good reasons if anyone has one!

## Links

- https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/ExplicitBlockArgument
- https://docs.rubocop.org/rubocop/cops_style.html#styleexplicitblockargument

## Testing story

## PR Checklist:

- [ ] Code is well-commented


bencodeorg (approved):
As someone whose Ruby is both rusty / was never that great, `yield` and `&block` are both a bit tricky to understand :) But I agree deferring to Rubocop without good reason to do otherwise makes sense to me!

bethanyaconnor (approved):
I like this change -- `yield` has always felt like magic to me and I like making it more explicit when a block is expected.

Hamms:
Fantastic, thank you both! I feel much better about this change now

PR #57750, opened 2024-04-03
Panels: two more layouts

breville:
This simple change adds two more layouts to the Panels level type.  "Top left" and "bottom right" are now available, augmenting the existing "top right" and "bottom left".  

Screenshot 2024-04-03 at 6 14 13 PM


sanchitmalhotra126 (approved):
🎉

PR #57755, opened 2024-04-03
[Google Blockly] support gray undeletable blocks

mikeharv:
Certain labs, like Maze and Artist change the color of deletable blocks to gray. This can be seen in places like levels 14 and 20 of the original Hour of Code progression:

### CDO Blockly (Baseline)
image

As of staging, we don't treat these blocks differently than other undeletable blocks:

### Google Blockly (Before/staging)
image

This branch establishes parity for this feature with CDO Blockly:

### Google Blockly (After/feature branch)
image

Whether we do this or not is determined by an `inject` option `grayOutUndeletableBlocks` which is already present in those lab files. We'll now add this option to the wrapper. If it's true, and the block is undeletable, we set a different block color. This can be handled by a new override for `setDeletable` which called during deserialization, or when toggling the deletable state (via levelbuilder context menu or console commands). Comments have been left with links to the comparable CDO Blockly logic, where applicable.

As you can see in the screenshots above, the `when run` block is not gray even though it is always undeletable. This is due to a `shouldBeGrayedOut()` override in the block definition here: 
https://github.com/code-dot-org/code-dot-org/blob/mike/undeletable-gray-blocks/apps/src/blocksCommon.js#L260-L262
We also want to exempt procedure definition blocks from being "grayed out" so I've created a new mixin for that.

## Links

Jira: https://codedotorg.atlassian.net/browse/CT-468

## Testing story

I tested that we do not show gray blocks in a read only workspace by comparing a project to the signed out view:
image
image

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/blockly/customBlocks/googleBlockly/proceduresBlocks.ts
+++ b/apps/src/blockly/customBlocks/googleBlockly/proceduresBlocks.ts
@@ -64,6 +64,7 @@ export const blocks = GoogleBlockly.common.createBlockDefinitionsFromJsonArray([
       'procedures_block_frame',
       'procedure_def_mini_toolbox',
       'modal_procedures_no_destroy',
+      'procedure_def_no_gray_out',
     ],
     mutator: 'procedure_def_mutator',
   },
@@ -280,6 +281,13 @@ GoogleBlockly.Extensions.registerMixin(
   procedureCallerOnChangeMixin
 );
 
+// Labs like Maze and Artist turn undeletable blocks gray. This is not
+// done for special blocks like "when run" or procedure definitions.
+GoogleBlockly.Extensions.registerMixin('procedure_def_no_gray_out', {
+  shouldBeGrayedOut: function () {
+    return false;
+  },
+});
 /**
  * Constructs the blocks required by the flyout for the procedure category.
  * Modeled after core Blockly procedures flyout category, but excludes unwanted blocks.
--- a/apps/src/blockly/googleBlocklyWrapper.ts
+++ b/apps/src/blockly/googleBlocklyWrapper.ts
@@ -8,7 +8,11 @@ import {
 } from '@blockly/plugin-scroll-options';
 import {LineCursor, NavigationController} from '@blockly/keyboard-navigation';
 import {CrossTabCopyPaste} from '@blockly/plugin-cross-tab-copy-paste';
-import {BlocklyVersion, WORKSPACE_EVENTS} from '@cdo/apps/blockly/constants';
+import {
+  BlockColors,
+  BlocklyVersion,
+  WORKSPACE_EVENTS,
+} from '@cdo/apps/blockly/constants';
 import styleConstants from '@cdo/apps/styleConstants';
 import * as utils from '@cdo/apps/utils';
 import initializeCdoConstants from './addons/cdoConstants';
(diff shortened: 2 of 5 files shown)


apps/src/blockly/customBlocks/googleBlockly/proceduresBlocks.ts
```diff
@@ -280,6 +281,13 @@ GoogleBlockly.Extensions.registerMixin(
   procedureCallerOnChangeMixin
 );
 
+// Labs like Maze and Artist turn undeletable blocks gray. This is not
+// done for special blocks like "when run" or procedure definitions.
+GoogleBlockly.Extensions.registerMixin('procedure_def_no_gray_out', {
+  shouldBeGrayedOut: function () {
```
mikeharv, line 287: CDO Blockly version: https://github.com/code-dot-org/blockly/blob/v4.1.0/blocks/procedures.js#L309-L311

apps/src/blockly/googleBlocklyWrapper.ts
```diff
@@ -499,6 +503,25 @@ function initializeBlocklyWrapper(blocklyInstance: GoogleBlocklyInstance) {
     return false;
   };
 
+  // Labs like Maze and Artist turn undeletable blocks gray.
+  extendedBlockSvg.shouldBeGrayedOut = function () {
```
mikeharv, line 507: CDO Blockly version: https://github.com/code-dot-org/blockly/blob/v4.1.0/core/ui/block.js#L1861-L1867

apps/src/blockly/googleBlocklyWrapper.ts
```diff
@@ -499,6 +503,25 @@ function initializeBlocklyWrapper(blocklyInstance: GoogleBlocklyInstance) {
...
+      blocklyWrapper.grayOutUndeletableBlocks &&
+      !this.workspace.isReadOnly() &&
+      !this.isDeletable()
+    );
+  };
+
+  const originalSetDeletable = blocklyWrapper.Block.prototype.setDeletable;
+  // Replace the original setDeletable with a version that will also re-color
+  // blocks if they are meant to be gray.
+  extendedBlockSvg.setDeletable = function (deletable) {
```
mikeharv, line 518: CDO Blockly version: https://github.com/code-dot-org/blockly/blob/v4.1.0/core/ui/block.js#L1851-L1856

We skip the `this.svg_` check here because we're only extending this for (rendered) Block SVGs. In the case of a headless workspace, we can skip attempting to set a color.

apps/src/blockly/googleBlocklyWrapper.ts
```diff
@@ -542,6 +565,10 @@ function initializeBlocklyWrapper(blocklyInstance: GoogleBlocklyInstance) {
     return this.getAllBlocks().filter(block => block.isEnabled());
   };
 
+  extendedWorkspaceSvg.isReadOnly = function () {
```
mikeharv, line 568: CDO Blockly version: https://github.com/code-dot-org/blockly/blob/v4.1.0/core/ui/block_space/block_space_editor.js#L1181-L1183

Note that in CDO Blockly, we set `this.readOnly_` in the Block Space Editor's constructor logic. We could achieve similar by modifying the workspace after the `inject` call, but that didn't seem warranted for such a simple check. Core Google Blockly frequently checks `workspace.options.readOnly`, so I've chosen to align to that instead.

molly-moen (approved):

PR #57756, opened 2024-04-03
[AI Tutor] CT-440: Migration for ai_tutor_iteraction_feedbacks table

ebeastlake:
The following PR creates the migration for a new `ai_tutor_iteraction_feedbacks` table. 

Long term, I expect the UI for this may be more like the Teacher Tools' AI TA feedback (see below), and we would need to add 1) column(s) to store boolean values associated with checkboxes for detailed feedback included in the UI and/or 2) a column to store the text of free-form feedback.

Screenshot 2024-04-03 at 12 22 05 AM

Since I don't know what these checkboxes/more detailed feedback paths might look like yet, I left them out of this version. This table should be pretty easy to alter after the fact, especially while I'm just testing with a few rows. 

## Links

Thread in #ai-tutor about how this should look: https://codedotorg.slack.com/archives/C06KYAJDB24/p1711573169852269

Thread in #ai-tutor with old Figma files from Edu bot: https://codedotorg.slack.com/archives/C06KYAJDB24/p1712000586140329?thread_ts=1711990444.119789&cid=C06KYAJDB24

Jira ticket: https://codedotorg.atlassian.net/browse/CT-440

## Testing story

Ran the migration locally using `rails db:migrate` and inspected.

Screenshot 2024-04-03 at 8 45 27 AM

## Follow-up work

Potentially: 
- Decide which checkbox-based feedback buttons we want, if any, and write alter to add columns
- Add an "other_context" column as a catch-all for text-based feedback at the interaction level
- Add this to the cron job that deletes student responses after 30 days: https://codedotorg.atlassian.net/browse/CT-472

## PR Checklist:

- [x] Code is well-commented

### DIFF
--- a/dashboard/app/models/ai_tutor_interaction.rb
+++ b/dashboard/app/models/ai_tutor_interaction.rb
@@ -10,9 +10,9 @@
 #  type               :string(255)
 #  project_id         :string(255)
 #  project_version_id :string(255)
-#  prompt             :text(16777215)
+#  prompt             :text(65535)
 #  status             :string(255)
-#  ai_response        :text(16777215)
+#  ai_response        :text(65535)
 #  created_at         :datetime         not null
 #  updated_at         :datetime         not null
 #
--- a/dashboard/app/models/ai_tutor_interaction.rb
+++ b/dashboard/app/models/ai_tutor_interaction.rb
@@ -10,9 +10,9 @@
 #  type               :string(255)
 #  project_id         :string(255)
 #  project_version_id :string(255)
-#  prompt             :text(65535)
+#  prompt             :text(16777215)
 #  status             :string(255)
-#  ai_response        :text(65535)
+#  ai_response        :text(16777215)
 #  created_at         :datetime         not null
 #  updated_at         :datetime         not null
 #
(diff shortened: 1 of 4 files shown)


dashboard/app/models/ai_tutor_interaction.rb
```diff
@@ -10,9 +10,9 @@
 #  type               :string(255)
 #  project_id         :string(255)
 #  project_version_id :string(255)
-#  prompt             :text(16777215)
```
ebeastlake: Does anyone know why this got updated? 🤔

dashboard/db/schema.rb
```diff
@@ -2373,12 +2383,8 @@
     t.index ["word", "definition"], name: "index_vocabularies_on_word_and_definition", type: :fulltext
   end
 
-  add_foreign_key "census_submission_form_maps", "census_submissions"
```
ebeastlake: There's a lot of unnecessary diff here, and I'm not sure where it came from. I don't have time to troubleshoot because I'm OOO the rest of the week, but I'll look when I'm back!

dashboard/app/models/ai_tutor_interaction.rb
jordan-springer: @ebeastlake you mentioned adding some boolean columns here, what did you have in mind?

jordan-springer (approved):
cool stuff!

PR #57757, opened 2024-04-03
AI rubric product tour with intro js

Nokondi:
- Installed intro.js and intro.js react wrapper
- Setup steps and placement of product tour
- Created new scss theme to style product tour.
- [Bonus: this PR also resolves the issue in [AITT-515](https://codedotorg.atlassian.net/browse/AITT-515?atlOrigin=eyJpIjoiYWE4MDhjNjYxN2YzNGY4ZTk1ZTZmM2ZmOGJiYjg0ODciLCJwIjoiaiJ9)

https://github.com/code-dot-org/code-dot-org/assets/5552007/d81995ae-00a4-486b-be3f-b019b2773639

## Links
[Figma Mock](https://www.figma.com/file/WBcs7BY9IC6Nq5rG4V1SWF/AI-TA%3A-Rubrics?type=design&node-id=4216-1276&mode=design&t=4DPeGrX6tQv1yfxK-0)
[Jira Ticket](https://codedotorg.atlassian.net/browse/AITT-533?atlOrigin=eyJpIjoiZjlhNWE2MzAyMDRkNDNlOWE5Y2ZmZWE0MDAzYjI4MTMiLCJwIjoiaiJ9)


apps/src/templates/rubrics/RubricContainer.jsx
```diff
@@ -52,12 +57,15 @@ export default function RubricContainer({
 
   const [feedbackAdded, setFeedbackAdded] = useState(false);
 
+  const [stepsEnabled, setStepsEnabled] = useState(false);
```
davidsbailey: rename to `productTour`? If these mean the same thing, I think it would be helpful for readability to name them consistently

dashboard/config/routes.rb
```diff
@@ -1082,6 +1082,8 @@
         get 'get_teacher_evaluations_for_all'
         get 'ai_evaluation_status_for_user'
         get 'ai_evaluation_status_for_all'
+        get 'get_ai_rubrics_tour_seen'
+        get 'update_ai_rubrics_tour_seen'
```
davidsbailey: this isn't safe -- please use PUT or POST for anything that can cause a state change.

apps/src/templates/rubrics/RubricContainer.jsx
```diff
@@ -102,92 +110,349 @@ export default function RubricContainer({
...
+  });
+
   // Currently the settings tab only provides a way to manually run AI.
   // In the future, we should update or remove this conditional when we
   // add more functionality to the settings tab.
-  const showSettings = onLevelForEvaluation && teacherHasEnabledAi;
+  const showSettings =
+    onLevelForEvaluation && teacherHasEnabledAi && !stepsEnabled;
+
+  // Steps for product tour
```
davidsbailey: please extract product tour config (here through line 231) to a separate JS file to cut down on file size

apps/src/templates/rubrics/RubricContainer.jsx
```diff
@@ -102,92 +110,349 @@ export default function RubricContainer({
...
+            {/* {showSettings && (
+              <RubricSettings
+                productTour={stepsEnabled}
+                visible={selectedTab === TAB_NAMES.SETTINGS}
+                refreshAiEvaluations={fetchAiEvaluations}
+                rubric={rubric}
+                sectionId={sectionId}
+                tabSelectCallback={tabSelectCallback}
+              />
+            )} */}
```
davidsbailey: please re-enable or remove

apps/src/templates/rubrics/introjs.scss
```diff
@@ -0,0 +1,448 @@
+@import 'color';
+@import 'font';
+
+.introjs-overlay {
+  position: absolute;
+  box-sizing: content-box;
+  z-index: 999999;
```
davidsbailey, line 7: this seems like a very high starting point, and the size of the numbers makes it difficult to see how they relate to each other. If possible I would suggest using numbers at most in the 100-200 range for improved readability. if another high z-index prevents this... let's take a follow-up item to reduce the arms race here.

davidsbailey (commented):
Woohoo, so exciting to see this come together! Some initial comments are inline.

Top-level comment: the spec says "Tour should display the first time a user lands on a page with an editable rubric", but when I go to http://localhost-studio.code.org:3000/s/csd3-2023/lessons/11/levels/4 (as opposed to /5), I still see the tour. I think this looks like a bug -- it is confusing because while the tour works, the highlighted features then disappear since evaluation is not possible on that level.

apps/src/templates/rubrics/RubricContainer.jsx
```diff
@@ -102,92 +110,349 @@ export default function RubricContainer({
...
-      >
-        <div className={style.rubricHeaderRedesign}>
-          <div className={style.rubricHeaderLeftSide}>
-            <img
-              src={aiBotOutlineIcon}
-              className={style.aiBotOutlineIcon}
-              alt={i18n.rubricAiHeaderText()}
-            />
-            <span>{i18n.rubricAiHeaderText()}</span>
+      {stepsEnabled ? (
```
davidsbailey: I am seeing a lot more duplicated code between the two branches of this conditional compared to the number of differences. my recommendation would be to DRY this up and use this conditional inline to change the behavior each place a difference is needed. otherwise it seems like we are setting ourselves up for the product tour to lag or break as other updates are made to the UI.

apps/src/templates/rubrics/RubricContainer.jsx
davidsbailey: this may be moot however if you are getting rid of the `stepsEnabled` ternary

davidsbailey (commented):
sorry for splitting my comments up -- I've finished reviewing now. nice job pulling this together!

my other top-level feedback is that this should really have a UI test: https://codedotorg.atlassian.net/browse/AITT-575

apps/src/templates/rubrics/RubricContainer.jsx
davidsbailey: Screenshot 2024-04-04 at 2 31 57 PM


Screenshot 2024-04-04 at 2 32 10 PM

apps/test/unit/templates/rubrics/RubricContainerTest.jsx
```diff
@@ -686,4 +692,61 @@ describe('RubricContainer', () => {
...
+      </Provider>
+    );
+
+    await act(async () => {
+      await Promise.resolve();
+    });
+
+    expect(screen.findAllByText('Getting Started with AI Teaching Assistant'))
+      .to.be.empty;
+  });
```
davidsbailey, line 764: I'm glad to see you're leaning in to React Testing Library! Since it seems like things part way through the product tour could break, my recommendation would be to have some coverage of clicking through the tour. this seems important because people will inevitably be making changes without necessarily thinking about the tour, which risks breaking it.

davidsbailey (commented):
juuust one more comment 😅

davidsbailey:
hey Mark, since the launch deadline is coming up, I want to say the min bar for shipping this is (1) turn the GET into a PUT/POST, and (2) make it not show up on non-eval levels, and (3) if possible a tiny bit more unit test coverage. I think the rest can be deferred to a follow-up PR, as long as it's done before moving onto other tasks.

tess323:
@davidsbailey  is the final call here, but if were still banging our head on tests I am okay with calling the following a fast follow.  We have the ability to slow roll folks into the pilot and think getting a few more hands on this is worth it.  

 (2) make it not show up on non-eval levels
 
Fully aligned on 1+3 as a bar. 

If the ad-hoc works tomorrow then I am super happy holding for all of this.

davidsbailey (approved):
looks good to merge once tests are passing. once you're back next week, it would be great to make sure we have a list of any PR comments not addressed in order to hit the ship deadline (if any) so that we don't lose track of them.

dashboard/test/ui/features/teacher_tools/rubrics/ai_evaluate_student_code.feature
```diff
@@ -39,6 +39,7 @@ Feature: Evaluate student code against rubrics using AI
     And element ".teacher-panel td:eq(1)" contains text "Aiden"
     And I click selector ".teacher-panel td:eq(1)" to load a new page
     And I wait for the page to fully load
+    And I click selector ".introjs-skipbutton" if it exists
```
davidsbailey, line 42: from the saucelabs video, it looks like the skipbutton is not getting clicked.

looking at this test step, it looks unreliable because when the "I wait for the page to fully load" step completes, the skipbutton may not have rendered yet. this kind of timing issue can often explain the frustrating issue of not being able to repro locally.

to prove/disprove, you could add a wait here (10 sec) to see if that fixes. 

for a shippable solution, you could do something like:
1. choose an element that will be in the DOM regardless of whether introjs is active, and render it differently when introjs is active
2. wait for that element to load
3. write a custom step that looks at that element and decides whether to wait for the skipbutton, and then click it

davidsbailey: ...at this point if the sleep statement fixes it, I think it could be ok to use that approach to get this PR unblocked, and do the "shippable" fix as a fast follow

dashboard/test/ui/features/teacher_tools/rubrics/teacher_view_of_rubric.feature
```diff
@@ -51,36 +53,69 @@ Scenario: Teachers can give and send feedback on the rubric to students.
...
+  And I click selector ".introjs-button:contains(Next Tip)" once I see it
+  And I wait until element "h1:contains(Understanding AI Confidence)" is visible
+  And I wait until element ".introjs-tooltiptext" is visible
+  And I click selector ".introjs-button:contains(Next Tip)" once I see it
+  And I wait until element "h1:contains(Assigning a Rubric Score)" is visible
+  And I wait until element ".introjs-tooltiptext" is visible
+  And I click selector ".introjs-button:contains(Next Tip)" once I see it
+  And I wait until element "h1:contains(How did we do?)" is visible
+  And I wait until element ".introjs-tooltiptext" is visible
+  And I click selector ".introjs-button:contains(Done)" once I see it
```
davidsbailey: to complete the scenario, I would suggest additional steps which verifies that we're now back into a mode where the teacher can interact with the rubric UI. For example:
* verify contents of rubrics UI matches what it should look like for the current level (as opposed to the tour)
* switch tabs or learning goals, and confirm that the UI updates correspondingly

since this has been tested manually, this seems fine to do as a fast-follow.

apps/src/templates/rubrics/RubricContainer.jsx
```diff
@@ -107,6 +115,69 @@ export default function RubricContainer({
...
+  const updateTourStatus = useCallback(() => {
+    const bodyData = JSON.stringify({seen: productTour});
+    const rubricId = rubric.id;
+    const url = `/rubrics/${rubricId}/update_ai_rubrics_tour_seen`;
+    HttpClient.post(url, bodyData, true, {
+      'Content-Type': 'application/json',
+    })
+      .then(response => {
+        return response.json();
+      })
```
davidsbailey, line 127: I like this as a way to make sure we do not read the response body/json twice

apps/src/templates/rubrics/introjs.scss
davidsbailey: this issue looks like it could still use follow-up (in a later PR)

apps/src/templates/rubrics/productTourHelpers.js
davidsbailey, line 1: 🤩

davidsbailey (commented):
@Nokondi I think this is a great workaround for the unit test issues. once again nice job pulling this all together! Looks good to merge. I am seeing one follow-up remaining, for the high z-indexes.

dashboard/test/ui/features/teacher_tools/rubrics/teacher_view_of_rubric.feature
Nokondi: Ah, the high z-index is actually part of the introjs module. I had to override their built-in css for some of our styling, so some elements (like the ones with the crazy z-index) are copied straight from their css.

davidsbailey:
> Ah, the high z-index is actually part of the introjs module. I had to override their built-in css for some of our styling, so some elements (like the ones with the crazy z-index) are copied straight from their css.

ugh! I see these now. boo z-index arms race. I agree in that case there is nothing to do about it then.

PR #57759, opened 2024-04-03
Nicklathe/debug statsig

nicklathe:
This fixes the bug where Statsig was not being initialized in Production. This was due to misunderstanding the (slightly unclear) [Statsig initialization instructions](https://docs.statsig.com/server/rubySDK#initializing-statsig-in-a-rails-application). In addtion to initializing in Puma, it's also necessary to initialize in the Rails initializer. This is true for all environments, not just development.

Statsig is idempotent in it's initialization, and is safe to re-init even if it's already running.

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/dashboard/config/initializers/statsig.rb
+++ b/dashboard/config/initializers/statsig.rb
@@ -1,6 +1,6 @@
 require "cdo/statsig"
-# Statsig is initialized here for the development environment. In managed
-# environments, it is initialized in lib/cdo/app_server_hooks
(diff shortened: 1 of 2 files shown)


carl-codeorg (approved):

wilkie (approved):
I trust your understanding of their documentation. And if this is going to fail a build, it will fail in staging first. Seems good to go!

Thanks for the patience and the probable headaches digging into this for all of us!

PR #57221, opened 2024-03-12
validate learning goals when validating ai config

davidsbailey:
Finishes https://codedotorg.atlassian.net/browse/AITT-242.

This PR adds validations to ensure that existing ai-enabled lessons are properly configured, and defines a process for enabling new lessons. note that although we do not plan to launch any "new" lessons in april, this PR will help make sure that lessons in units like `interactive-games-animations-2023` are properly configured. This PR will also help prevent regressions in our AI config.

## Validations

Previous work validates the following:
* when UI tests run, for every lesson that has ai evaluation enabled in our ruby code:
  * make sure AI config exists in S3 containing all the files needed for AI to evaluate it (system_prompt.txt, standard_rubric.csv, etc).

This PR adds the following validations:
* when UI tests run, for every lesson for which ai evaluation is configured in our ruby code:
  * validate that there is a corresponding rubric in the database
  * validate that for every learning goal in the rubric in the database, there is a corresponding row in system_prompt.csv file in S3 describing that learning goal.
* when a levelbuilder creates/updates a learning goal, if the learning goal is ai_enabled, make sure there is a corresponding entry in standard_rubric.csv in S3 describing that learning goal.

## lesson launch process

Based on the old and new rules, the process for adding ai evaluation for a new lesson becomes:
1. **create rubric**: create rubric and learning goals via levelbuilder, with ai_enabled flag initially disabled on each learning goal
2. **create S3 config**: based on those rubrics / learning goals, create config in S3 telling AI how to evaluate that lesson, and tune it for accuracy
3. **launch**:
    a. add the lesson to this map: https://github.com/code-dot-org/code-dot-org/blob/2fa69d96a0f7e04970bfce9b0e47f0a94b08271e/dashboard/app/jobs/evaluate_rubric_job.rb#L18-L23
    b. via levelbuilder ui, set ai_enabled flag for each learning goal that can be ai-evaluated 

note that step 3a and 3b can be done separately, or in a single PR by running the levelbuilder ui locally. 

## Screenshots

When `validate_learning_goals` fails during UI tests, useful error details are shown: 
Screenshot 2024-03-13 at 9 58 02 AM

## Testing story

* UI test which runs `validate_ai_config` caught a number of problems (see inline comments)
* new unit tests cover expected behavior for create/update of learning goals via levelbuilder
* I'm sending this out ahead of drone... but do have one passing unit test run and one passing UI test run between the two most recent drone runs 😅 

## Future work

[...]

PR #57762, opened 2024-04-03
enable ai evaluation in other copies of csd3

davidsbailey:
Finishes https://codedotorg.atlassian.net/browse/AITT-490. basically there are 3 units which are copies of csd3 in other courses, and this PR enables AI for them. The whole process looked something like this:

already in staging:
* Tess manually copied rubric contents from csd3 to other 3 units using levelbuilder UI
* I made some additional edits via levelbuilder to fix one missing / broken rubric

in this PR:
1. updated `UNIT_AND_LEVEL_TO_LESSON_S3_NAME` in ruby code to provide pointers to ai config for the 3 new units
2. using levelbuilder_mode locally, I checked the "AI Enabled" checkbox, resulting in changes to script_json files which are visible in this PR

this is all following the new "lesson launch process" outlined previously in https://github.com/code-dot-org/code-dot-org/pull/57221#:~:text=lesson%20launch%20process

fun fact, the new validation actually caught some errors during this process which would have led to one lesson being broken. specifically, in step 2 above, i initially got an error message in the levelbuilder UI because the rubric for lesson 14 was missing from focus-on-coding3-2023. after I entered it manually, the validation failed again because I had copied the wrong rubric contents in. hooray for validation 😁 

## Testing story

* validations were run on each levelbuilder save (catching one issue as noted above)
* `EvaluateRubricJob.new.validate_ai_config` is passing, and also gets run by drone
* manually verified that AI bot + rubrics are showing up on at least 1 lesson in each newly-ai-enabled unit

## Deployment strategy

just noting that this is enough to launch ai for the 3 other units.  no update to S3 releases is necessary, because this piggy-backs off the existing lesson config in S3, which is possible because the lessons are identical.


wilkie (approved):
Did a manual check for a good portion of the levels and whether or not their ai_enabled matches what we are prompting. Seems good to go!

PR #57764, opened 2024-04-03
LinkButton component style updates

kelbyhawn:
Adds link tag (`a.button`) selectors to button style groups on `_baseButton.module.scss` to override link styles on `application.scss`: https://github.com/code-dot-org/code-dot-org/blob/21c14b9c5c9c2edc65111afb229ab773fab58670/dashboard/app/assets/stylesheets/application.scss#L491-L518

**Also includes the following enhancements:** 
  - Adds a border to all buttons (primary and disabled) so they're the same height as the secondary buttons
  - Adds a subtle hover transition animation

💬 [Slack convo](https://codedotorg.slack.com/archives/C0532LCK125/p1712162369463069) - see more context on how styles are being overridden. 

----

## Border
| Before | After |
| ---- | ---- |
| Before | After |

## Hover transition

https://github.com/code-dot-org/code-dot-org/assets/9256643/31584260-001d-495f-b510-ea048a44d20a

### DIFF
--- a/apps/src/componentLibrary/button/CHANGELOG.md
+++ b/apps/src/componentLibrary/button/CHANGELOG.md
@@ -2,9 +2,9 @@
 
 All notable changes to this project will be documented in this file.
 
-## [0.2.1](https://github.com/code-dot-org/code-dot-org/pull/56925)
+## [0.2.1](https://github.com/code-dot-org/code-dot-org/pull/57764)
 * updated link button styles to prevent overrides by application.scss
-* added a border to all buttons so they are the same height
+* added a transparent border to all buttons, so they are the same height
 * added a subtle hover transition animation
 
 ## [0.2.0](https://github.com/code-dot-org/code-dot-org/pull/56925)
--- a/apps/src/componentLibrary/button/CHANGELOG.md
+++ b/apps/src/componentLibrary/button/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 All notable changes to this project will be documented in this file.
 
+## [0.2.1](https://github.com/code-dot-org/code-dot-org/pull/56925)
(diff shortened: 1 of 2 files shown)


TurnerRiley (approved):

apps/src/componentLibrary/button/CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 All notable changes to this project will be documented in this file.
 
+## [0.2.1](https://github.com/code-dot-org/code-dot-org/pull/56925)
```
kelbyhawn: @levadadenys what's the version pattern here? Let me know if I need to change it!

levadadenys: I feel like this is ok to make it 0.2.1, just update the link to be a link to this pr

levadadenys (commented):
Animation looks great, thank you!
Re adding border to all buttons - there's now an issue that buttons are 'jumpy' when clicked.

https://github.com/code-dot-org/code-dot-org/assets/22244040/6a4357f3-2f58-4d73-8994-8936b9b3a8e9

We need to remove border again, or adjust the styling of the active button border, since there's !important styles in `./apps/style/common.scss line 655 styles` setting border to   border: 1px solid #888 !important; (that's why in our module.scss we have `border: none !important` which makes button jumpy now)

levadadenys:
I've just pushed a commit, updating changelog and fixing the 'jumpy' buttons issue. Now buttons have transparent border by default always (including active state)

@kelbyhawn if that follows your initial idea, from my side - I'm good/+1 to merging this

levadadenys (approved):

PR #57775, opened 2024-04-03
Gen AI: initial "compare models" dialog

bencodeorg:
Adds an initial version of the new "compare model" dialog to our new generative AI lab. There's still work do to here, but felt like a good checkpoint. See "Follow-up work" for what's next.

On request from design, this includes some logic to make the content unscrollable initially, but allows the student to scroll upon request.

https://github.com/code-dot-org/code-dot-org/assets/25372625/16fd0879-9b4e-4230-a6e9-826d4f299ebe

## Testing story

Just testing manually for now.

## Follow-up work

- I think the desire from design is to have a single scroll bar across both columns, but we can work that out in the future.
- Styling improvements.
- Store a model choice as part of a student's project (and allow levelbuilders to configure this choice). At that point, we'll probably want to set up some logic to default the left panel to the currently chosen model.
- Get actual model description data.


bencodeorg:
@levadadenys I'm having issues getting a full width dropdown using `SimpleDropdown` -- we don't have any way of directly applying styling to the `select` element, but that's the main way I was able to get a dropdown to appear at full width (ie, something like the change I [reverted here](https://github.com/code-dot-org/code-dot-org/pull/57775/commits/82acb492777eae8dd6b3b7720c3eba1d7d5b7158) (I took the video recording with `width: 100%` set). Without that change, the dropdown looks like the following:

image

Do you have thoughts on the best way to do a full width dropdown?

apps/src/aichat/views/modelCustomization/ModelDescriptionPanel.tsx
```diff
@@ -0,0 +1,106 @@
+import React, {useEffect, useState, useRef} from 'react';
+
+import SimpleDropdown from '@cdo/apps/componentLibrary/dropdown/simpleDropdown';
+import {BodyThreeText, StrongText} from '@cdo/apps/componentLibrary/typography';
+import Button from '@cdo/apps/componentLibrary/button/Button';
+
+import styles from './compare-models-dialog.module.scss';
+
+type ModelDescription = {
```
fisher-alice, line 9: Should we move this to `types.ts`?

apps/src/aichat/views/modelCustomization/PromptCustomization.tsx
```diff
@@ -35,6 +43,33 @@ const PromptCustomization: React.FunctionComponent = () => {
   return (
     <div className={styles.verticalFlexContainer}>
       <div>
+        <div className={classNames(styles.inputContainer)}>
```
fisher-alice: Would a subrender function make sense here for easier readability?

apps/src/aichat/views/modelCustomization/PromptCustomization.tsx
```diff
@@ -35,6 +43,33 @@ const PromptCustomization: React.FunctionComponent = () => {
...
+          <SimpleDropdown
+            labelText="Pick a model:"
```
fisher-alice: If this field is read-only, would we want the user still see 'Pick a model' label above?
Generally speaking, what is the plan for read-only vs editable fields?

apps/src/aichat/views/modelCustomization/PromptCustomization.tsx
```diff
@@ -35,6 +43,33 @@ const PromptCustomization: React.FunctionComponent = () => {
...
+            items={[
+              {value: 'llama2', text: 'LLama 2'},
+              {value: 'gpt', text: 'ChatGPT'},
+            ]}
+            selectedValue={chosenModel}
+            name="model"
+            size="s"
+            className={styles.updateButton}
+          />
+          <div>
```
fisher-alice: Do we need this div?

apps/src/aichat/views/modelCustomization/ModelDescriptionPanel.tsx
```diff
@@ -0,0 +1,106 @@
...
+        <div className={styles.textContainer}>
+          <BodyThreeText>{selectedModel.overview}</BodyThreeText>
+        </div>
+        <br />
+        <StrongText>Training Data</StrongText>
+        <div className={styles.textContainer}>
+          <BodyThreeText>{selectedModel.trainingData}</BodyThreeText>
+        </div>
+      </div>
+      {showViewMoreButton && (
```
fisher-alice, line 90: Wondering if we should have a 'View less' button as well.

bencodeorg: Maybe? I do think we might need another round of revision here anyway with design guidance, as Mark mentioned that ideally we'd just have one scroll bar across the entire dialog (vs. one per column). I felt like this PR got the bit of "how do you include/exclude some UI based on whether a content is overflowing or not", and we could iron out the finer points once we cross the milestone at the end of next week of having a "lo-fi" version of the UI in place.

apps/src/aichat/views/modelCustomization/PromptCustomization.tsx
bencodeorg: mmm good q...I basically deferred on all of the levelbuilder-related stuff in this PR, but I see [this comment from Dan](https://codedotorg.slack.com/archives/C06FELVTV0Q/p1711478711932649?thread_ts=1711473774.746449&cid=C06FELVTV0Q) -- maybe we show the dropdown (disabled) but no label (we'd want to be careful with for a11y though)

apps/src/aichat/views/modelCustomization/PromptCustomization.tsx
bencodeorg: I don't believe so! Removing. Wish there was an "extra div" linter 🤦

apps/src/aichat/views/modelCustomization/PromptCustomization.tsx
bencodeorg: Do you mean for all of the new stuff (dropdown/button/dialog)? Just tried something along those lines, LMK what you think. Could probably break up / refactor the rest of this as well, Sanchit and I chatted about reusing some of the components he added on the levelbuilder side (eg, `FieldSection`)

apps/src/aichat/views/modelCustomization/CompareModelsDialog.tsx
```diff
@@ -0,0 +1,47 @@
...
+import styles from './compare-models-dialog.module.scss';
+
+const CompareModelsDialog: React.FunctionComponent<{onClose: () => void}> = ({
+  onClose,
+}) => {
+  const [chosenModelLeft, setChosenModelLeft] = useState<string>('llama2');
+  const [chosenModelRight, setChosenModelRight] = useState<string>('mistral');
+
+  return (
+    <AccessibleDialog onClose={onClose} modalId={styles.modelComparisonDialog}>
```
bencodeorg: Getting an ID from a .module.scss file isn't a pattern I see elsewhere, which usually isn't a good sign 😛 .

I wanted a CSS selector that is more specific than the class names generated within AccessibleDialog's own styling, so ended up with this. It seemed like passing in another class name wouldn't guarantee that it would be more specific than what is applied by AccessibleDialog. Is there a "right" way to do this?

apps/src/aichat/views/modelCustomization/PromptCustomization.tsx
fisher-alice: > Do you mean for all of the new stuff (dropdown/button/dialog)? Just tried something along those lines, LMK what you think.

Yes - thanks! Look forward to seeing the commit.

> Could probably break up / refactor the rest of this as well

Nice!

> Sanchit and I chatted about reusing some of the components he added on the levelbuilder side (eg, FieldSection)

Looking at Sanchit's`ModelCardFields` and `FieldSection`, they seem different enough from the user view so that making them generic seems like it may take a bit of work. But think definitely work a refactor/using same components after the P0 is reached!

apps/src/aichat/views/modelCustomization/ModelDescriptionPanel.tsx
fisher-alice: SGTM!

apps/src/aichat/views/modelCustomization/PromptCustomization.tsx
bencodeorg: hah sorry forgot to push, done!

apps/src/aichat/views/modelCustomization/PromptCustomization.tsx
fisher-alice: Yeah - that is a useful thread to reference - thanks!
@dancodedotorg said:
> If read-only, I'd expect the dropdown to be visible but "Compare Models" button to be hidden or disabled.

So it seems like if it's read-only, we'd want the label above the dropdown to perhaps be: 'Selected model' or 'Model'? @dancodedotorg let us know what you think.

This question is not for this PR but follow-up work.

apps/src/aichat/views/modelCustomization/CompareModelsDialog.tsx
levadadenys: Not sure if there's a problem with using ID. Though, one more thing you can try doing is to write a htmlElement.Classname style (eg. `div.modelComparisonDialog`) in module styles. This way the priority of the styles will be higher than just `.modelComparisonDialog`

levadadenys:
> @levadadenys I'm having issues getting a full width dropdown using `SimpleDropdown` -- we don't have any way of directly applying styling to the `select` element, but that's the main way I was able to get a dropdown to appear at full width (ie, something like the change I [reverted here](https://github.com/code-dot-org/code-dot-org/pull/57775/commits/82acb492777eae8dd6b3b7720c3eba1d7d5b7158) (I took the video recording with `width: 100%` set). Without that change, the dropdown looks like the following:
> 
> image
> Do you have thoughts on the best way to do a full width dropdown?

Oh yeah, I see now. Thanks for pinging me on this one!
Looks like setting `width: 100%` doesn't brake anything, so I think we should be ok with adding this. (might worth doing in a separate PR with updating corresponding Changelog though).

And also, it brings to me another point here. It looks like we don't really need to set a classname to select element since there's only one select in this component. So might also worth changing `.dropdown` -> `select`, this way if you'll wrap the className of simple dropdown container into another classname in higher level module styles - we'll be able to set more custom styles to selectElement (if it's needed of course)

apps/src/aichat/views/modelCustomization/CompareModelsDialog.tsx
fisher-alice: As an alternative, you could use the `styles` prop for `AccessibleDialog` and add the modal styling from `accessible-dialogue` to class `modal` in `compare-models-dialog`. Since that styling is duplicated, maybe make a mix-in?

apps/src/aichat/views/modelCustomization/PromptCustomization.tsx
fisher-alice: Looks good!

apps/src/aichat/views/modelCustomization/PromptCustomization.tsx
dancodedotorg: Asking about this in Slack and tagged Alice!

apps/src/aichat/views/modelCustomization/ModelDescriptionPanel.tsx
bencodeorg: I think when we've chatted about this before we've been ok leaving types used in one place in the file that they're used?

apps/src/aichat/views/modelCustomization/PromptCustomization.tsx
fisher-alice: Just documenting that Sam and Dan prefer `Selected model' - see thread above.

apps/src/aichat/views/modelCustomization/compare-models-dialog.module.scss
```diff
@@ -0,0 +1,100 @@
...
+
+.modelDescriptionContainer {
+  flex: 1;
+  padding: 5px 10px;
+  display: flex;
+  flex-direction: column;
+}
+
+#modelComparisonDialog {
+  height: 600px;
```
fisher-alice: The 'Finish' button is getting cut off one of my screens. 

Screenshot 2024-04-04 at 10 45 58 AM

What do you think about using something like `calc(vh - 100px)`?

bencodeorg: TIL, thanks! Looks great, [updated here](https://github.com/code-dot-org/code-dot-org/pull/57775/commits/c1e2c52c293acaf04c459db359461b815baec4e8)

apps/src/aichat/views/modelCustomization/CompareModelsDialog.tsx
bencodeorg: Denys's approach seemed simpler, so rolled with that. LMK if it looks ok!

apps/src/aichat/views/modelCustomization/PromptCustomization.tsx
bencodeorg: Updated!

bencodeorg:
> Oh yeah, I see now. Thanks for pinging me on this one! Looks like setting `width: 100%` doesn't brake anything, so I think we should be ok with adding this. (might worth doing in a separate PR with updating corresponding Changelog though).
> 
> And also, it brings to me another point here. It looks like we don't really need to set a classname to select element since there's only one select in this component. So might also worth changing `.dropdown` -> `select`, this way if you'll wrap the className of simple dropdown container into another classname in higher level module styles - we'll be able to set more custom styles to selectElement (if it's needed of course)

@levadadenys either approach seems fine to me! I can merge this with the dropdowns looking a little funky since this isn't in use yet, but would you be up to implement whichever approach feels better to you? I'm a little nervous to mess with the internals of this component since it's already in use by other teams / my CSS skills are not the best.

fisher-alice (approved):
Looks good - thanks for the fast iteration!

bencodeorg:
> @levadadenys either approach seems fine to me! I can merge this with the dropdowns looking a little funky since this isn't in use yet, but would you be up to implement whichever approach feels better to you? I'm a little nervous to mess with the internals of this component since it's already in use by other teams / my CSS skills are not the best.

@levadadenys -- I ended up putting something together for this here, LMK what you think!

https://github.com/code-dot-org/code-dot-org/pull/57808

PR #57776, opened 2024-04-03
Hide deprecated projects from personal project table

mikeharv:
Calc and Eval were deprecated in January 2023, but projects of these types still show up in the personal projects table. 

This makes it harder for us to actually end support for (ie. break) these labs. Having the freedom to do so could expedite the process of migrating other labs to mainline Blockly, specifically Play Lab ("Studio") and Star Wars.

Slack threads:
* https://codedotorg.slack.com/archives/C05DK21DAHK/p1712166286197639
* https://codedotorg.slack.com/archives/C03DBDN67B7/p1712170539321259

We can filter the projects list by type to remove them from the personal projects list.

Before:
image

After:
image

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/templates/projects/PersonalProjectsTable.jsx
+++ b/apps/src/templates/projects/PersonalProjectsTable.jsx
@@ -17,6 +17,8 @@ import DeleteProjectDialog from '@cdo/apps/templates/projects/deleteDialog/Delet
 import FrozenProjectInfoDialog from '@cdo/apps/templates/projects/frozenProjectInfoDialog/FrozenProjectInfoDialog';
 import {isSignedIn} from '@cdo/apps/templates/currentUserRedux';
 import SafeMarkdown from '@cdo/apps/templates/SafeMarkdown';
+import {filter} from 'lodash';
+import {DEPRECATED_LABS} from '@cdo/apps/constants';
 
 const PROJECT_DEFAULT_IMAGE = '/blockly/media/projects/project_default.png';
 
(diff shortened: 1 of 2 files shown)


molly-moen (approved):

apps/src/templates/projects/PersonalProjectsTable.jsx
```diff
@@ -17,6 +17,8 @@ import DeleteProjectDialog from '@cdo/apps/templates/projects/deleteDialog/Delet
 import FrozenProjectInfoDialog from '@cdo/apps/templates/projects/frozenProjectInfoDialog/FrozenProjectInfoDialog';
 import {isSignedIn} from '@cdo/apps/templates/currentUserRedux';
 import SafeMarkdown from '@cdo/apps/templates/SafeMarkdown';
+import {filter} from 'lodash';
```
fisher-alice: Just curious why you're using `filter` from 'lodash' vs JS native filter.

fisher-alice (approved):
Nice! Just left a small question.

apps/src/templates/projects/PersonalProjectsTable.jsx
mikeharv: Hmm, that's odd. I wasn't actually using it. :) 

I think it got added by VSCode automatically at some point as I was using auto-fill suggestions and I just missed that it happened. It was probably about to be caught by drone as a linting error. Good catch!

PR #57782, opened 2024-04-03
[Python Lab] add support for matplotlib and small updates

molly-moen:
## Warning!!

## Summary

The work in this PR came out of investigating if we could get matplotlib working in Python Lab. matplotlib is a supported pyodide package, but there are some issues with getting it working in a web worker (see [this issue](https://github.com/pyodide/matplotlib-pyodide/issues/6) for details). As a simple workaround we can patch the matplotlib `show` function to print a base64 string, then show said string in an image element. This may prove to be a little fragile; I filed a [follow up task](https://codedotorg.atlassian.net/browse/CT-475) to productionize this.

This work required a couple things that caused me to do a few refactors/improvements more generally to Python Lab. These are:
- update the console log storage in redux to have different log types, image, system out, system in (not used yet), and system message (things we send, such as start/end messages). 
- Clean up and update how we wrap the user's code with our custom patches. Previously this was all done in `pyodideWorkerManager`. Now I have moved the existing flush system out patch (which is appended to all student code) and the new matplotlib patch into a separate file, and created a utils file to handle wrapping the user's code.
- Always load pyodide on page load. Now that we are patching matplotlib, we would be importing numpy and matplotlib on first run if we continued to lazy load. matplotlib takes a few seconds to load, so starting that process early makes the first run faster. We can do this now because we only load the component if we are on a Python Lab level, as opposed to previously when it was loaded for all lab2 levels.

This is what it looks like to run a program that outputs two graphs
Screenshot 2024-04-03 at 2 34 41 PM

## Links
- jira tickets: [CT-448](https://codedotorg.atlassian.net/browse/CT-448), [CT-458](https://codedotorg.atlassian.net/browse/CT-458)

## Testing story
Tested locally. Python Lab is isolated code and not in use, so this is low risk.

## Follow-up work
- [CT-475](https://codedotorg.atlassian.net/browse/CT-475) is a placeholder for productionizing this work.

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/pythonlab/PythonEditor.tsx
+++ b/apps/src/pythonlab/PythonEditor.tsx
@@ -36,7 +36,7 @@ const PythonEditor: React.FunctionComponent = () => {
     const parsedData = data ? (data as PermissionResponse) : {permissions: []};
     // For now, restrict running python code to levelbuilders.
     if (parsedData.permissions.includes('levelbuilder')) {
-      dispatch(appendSystemMessage('Running code...'));
+      dispatch(appendSystemMessage('Running program...'));
       if (source) {
         const code = getFileByName(source.files, 'main.py')?.contents;
         if (code) {
--- a/apps/src/pythonlab/PythonEditor.tsx
+++ b/apps/src/pythonlab/PythonEditor.tsx
@@ -3,7 +3,7 @@ import {darkMode} from '@cdo/apps/lab2/views/components/editor/editorThemes';
 import {python} from '@codemirror/lang-python';
 import moduleStyles from './python-editor.module.scss';
 import {useDispatch} from 'react-redux';
-import {appendOutput, resetOutput, setSource} from './pythonlabRedux';
+import {appendSystemMessage, resetOutput, setSource} from './pythonlabRedux';
 import Button from '@cdo/apps/templates/Button';
 import {runPythonCode} from './pyodideRunner';
 import {useFetch} from '@cdo/apps/util/useFetch';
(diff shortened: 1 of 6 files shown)


apps/src/pythonlab/PythonEditor.tsx
```diff
@@ -102,7 +102,22 @@ const PythonEditor: React.FunctionComponent = () => {
       <div>
         Output:
         {codeOutput.map((outputLine, index) => {
-          return <div key={index}>{outputLine}</div>;
+          if (outputLine.type === 'img') {
```
ebeastlake, line 105: non-blocking nit: Should we eventually move these types to a constants file somewhere?

apps/src/pythonlab/PythonEditor.tsx
```diff
@@ -102,7 +102,22 @@ const PythonEditor: React.FunctionComponent = () => {
...
+            return (
+              <img
+                key={index}
+                src={`data:image/png;base64,${outputLine.contents}`}
```
ebeastlake, line 109: Nice solution!

apps/src/pythonlab/patches/patches.ts
```diff
@@ -0,0 +1,51 @@
...
+// Ensure stdout is flushed at the end of the user's program
+// so all of their prints show up. This patch is always appended to the end of the user's code.
+const FLUSH_STDOUT = `import sys
+import os
+
+sys.stdout.flush()
+os.fsync(sys.stdout.fileno())
+`;
+
+// TODO: Should we always patch matplotlib? Or can we be smarter about when to patch it?
```
ebeastlake, line 46: Is there some comment about this included in the task to productionize the patch? If not, can we add a comment?

ebeastlake (approved):
Nice work! 🎉

apps/src/pythonlab/patches/patches.ts
molly-moen: I'll add a link to the ticket here!

apps/src/pythonlab/PythonEditor.tsx
molly-moen: This is going to get refactored completely so I think for now I'll leave as-is

molly-moen: but yes, they should eventually be constants :)

ebeastlake: Haha, I figured! As I typed it I knew I was just making the "obligatory seeing plaintext in a condition" comment 😂 Sounds good to me!

PR #57786, opened 2024-04-03
Ceara/aitt 543 565 more rubric amplitude

cearachew:
## Warning!!

## Description

Adds amplitude/statsig logging to various events to track the time between student submitting/ teacher pressing evaluate and teachers looking at ai evaluations. Adds the following events:
```  
  TA_RUBRIC_INDIVIDUAL_AI_EVAL: 'TA Rubric Individual AI Eval Requested',
  TA_RUBRIC_SECTION_AI_EVAL: 'TA Rubric Section AI Eval Requested',
  TA_RUBRIC_AI_PAGE_VISITED: 'TA Rubric AI Level Page Visited',
  TA_RUBRIC_STUDENT_AI_SUBMITTED: 'TA Rubric Student AI Level Submitted',
```

## Links

https://codedotorg.atlassian.net/browse/AITT-565
https://codedotorg.atlassian.net/browse/AITT-543

## Testing story

Adds unit tests for the UI amplitude/ statsig logging, but I couldn't figure out how to add a test for the logging in the ruby tests/ couldn't find any other examples of testing the statsig logging in ruby.

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/sites/studio/pages/levels/show.js
+++ b/apps/src/sites/studio/pages/levels/show.js
@@ -16,6 +16,8 @@ import {setLevel, setScriptId} from '@cdo/apps/aiTutor/redux/aiTutorRedux';
 import experiments from '@cdo/apps/util/experiments';
 import RubricFloatingActionButton from '@cdo/apps/templates/rubrics/RubricFloatingActionButton';
 import AITutorFloatingActionButton from '@cdo/apps/code-studio/components/aiTutor/aiTutorFloatingActionButton';
+import analyticsReporter from '@cdo/apps/lib/util/AnalyticsReporter';
+import {EVENTS, PLATFORMS} from '@cdo/apps/lib/util/AnalyticsConstants';
 
 $(document).ready(initPage);
 
@@ -102,6 +104,21 @@ function initPage() {
       'rubric-fab-mount-point'
     );
     if (rubricFabMountPoint) {
+      //rubric fab mount point is only true for teachers
+      if (
+        experiments.isEnabled('ai-rubrics') &&
+        !!rubric &&
+        rubric.learningGoals.some(lg => lg.aiEnabled)
+      ) {
+        analyticsReporter.sendEvent(
+          EVENTS.TA_RUBRIC_AI_PAGE_VISITED,
+          {
+            ...reportingData,
+            studentId: !!studentLevelInfo ? studentLevelInfo.user_id : '',
+          },
+          PLATFORMS.BOTH
+        );
+      }
       ReactDOM.render(
         <Provider store={getStore()}>
           <RubricFloatingActionButton
--- a/apps/src/sites/studio/pages/levels/show.js
+++ b/apps/src/sites/studio/pages/levels/show.js
@@ -100,6 +100,8 @@ function initPage() {
     };
     getStore().dispatch(setTaRubric(rubric));
 
+    console.log(config);
+    console.log(rubric);
     const rubricFabMountPoint = document.getElementById(
       'rubric-fab-mount-point'
     );
@@ -108,7 +110,8 @@ function initPage() {
       if (
         experiments.isEnabled('ai-rubrics') &&
         !!rubric &&
-        rubric.learningGoals.some(lg => lg.aiEnabled)
+        rubric.learningGoals.some(lg => lg.aiEnabled) &&
+        config.level_name === rubric.level.name
       ) {
         analyticsReporter.sendEvent(
           EVENTS.TA_RUBRIC_AI_PAGE_VISITED,
--- a/apps/src/sites/studio/pages/levels/show.js
+++ b/apps/src/sites/studio/pages/levels/show.js
@@ -100,8 +100,6 @@ function initPage() {
     };
     getStore().dispatch(setTaRubric(rubric));
 
-    console.log(config);
-    console.log(rubric);
     const rubricFabMountPoint = document.getElementById(
       'rubric-fab-mount-point'
     );
--- a/apps/src/templates/rubrics/RubricSettings.jsx
+++ b/apps/src/templates/rubrics/RubricSettings.jsx
@@ -10,7 +10,7 @@ import {
   StrongText,
 } from '@cdo/apps/componentLibrary/typography';
 import analyticsReporter from '@cdo/apps/lib/util/AnalyticsReporter';
-import {EVENTS} from '@cdo/apps/lib/util/AnalyticsConstants';
+import {EVENTS, PLATFORMS} from '@cdo/apps/lib/util/AnalyticsConstants';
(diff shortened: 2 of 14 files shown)


apps/src/templates/rubrics/RubricSettings.jsx
```diff
@@ -231,6 +231,16 @@ export default function RubricSettings({
...
+    const eventName = EVENTS.TA_RUBRIC_SECTION_AI_EVAL;
+    analyticsReporter.sendEvent(
+      eventName,
+      {
+        ...(reportingData || {}),
+        rubricId: rubricId,
+        sectionId: sectionId,
+      },
+      PLATFORMS.BOTH
+    );
```
davidsbailey, line 243: unfortunately I think what we really want here is to log an individual event for each student that is getting processed as a result of this request, which would need to be done on the backend. otherwise I'm worried it will be too hard to reconstruct within the statsig tool after the fact.

dashboard/app/controllers/activities_controller.rb
```diff
@@ -124,6 +125,16 @@ def milestone
       is_ai_experiment_enabled = current_user && Experiment.enabled?(user: current_user, script: @script_level&.script, experiment_name: 'ai-rubrics')
       is_level_ai_enabled = EvaluateRubricJob.ai_enabled?(@script_level)
       if is_ai_experiment_enabled && is_level_ai_enabled && params[:submitted] == 'true'
+        metadata = {
+          'userId' => current_user.id,
+          'levelId' => @level.id,
+          'levelName' => @level.name,
```
davidsbailey: this metadata will need to match up across event types, and will need to uniquely identify the level/lesson, keeping mind mind that levels are shared across ai-enabled units. I would recommend logging script name and level name. 

on the client, script name and level name appear to already be available inside `reportingData`.

I'm not sure exactly how this will work in statsig yet, but to minimize the chances of us running into problems, please use the exact same names for these fields across event types, e.g. "studentId" here if that's what we're using in other places.

apps/src/sites/studio/pages/levels/show.js
```diff
@@ -102,6 +104,21 @@ function initPage() {
       'rubric-fab-mount-point'
     );
     if (rubricFabMountPoint) {
+      //rubric fab mount point is only true for teachers
+      if (
+        experiments.isEnabled('ai-rubrics') &&
+        !!rubric &&
+        rubric.learningGoals.some(lg => lg.aiEnabled)
+      ) {
```
davidsbailey: this looks like a good start, but can you please further limit this so that we only log when the teacher is looking at the final level in the progression, where ai evaluation results are displayed. for example:
    * should trigger on http://localhost-studio.code.org:3000/s/csd3-2023/lessons/11/levels/5?section_id=X&user_id=X
    * should NOT trigger on http://localhost-studio.code.org:3000/s/csd3-2023/lessons/11/levels/4?section_id=X&user_id=X

davidsbailey (commented):
nice progress Ceara! some comments inline.

For validation, I think it would make sense to verify that each event can make it into statsig, possibly with help from Nick. I believe statsig has support for local development, though you may have to ask Nick exactly how.

for testing on the ruby side, I think verifying the right calls are being made to `Metrics::Events.log_event` would be sufficient. I believe this would be something like 
```
Metrics::Events.stubs(:log_event).once
```
or possibly something like this if you want to verify params: https://github.com/code-dot-org/code-dot-org/blob/e4741992140657d6b4f90961913f9a6b4fdee5bb/dashboard/test/controllers/rubrics_controller_test.rb#L51-L53

apps/src/templates/rubrics/RubricSettings.jsx
cearachew: Just to double check, you mean that we want an event for each student in this loop right before the evaluate rubric job is called?: https://github.com/code-dot-org/code-dot-org/blob/d5ed4e18f86377b9d0ad83fa12ae94d60784dd96/dashboard/app/controllers/rubrics_controller.rb#L196-L200
Is this event in rubric settings still useful, or should it be removed? Does the event for the button click for an individual student need to be moved to the backend as well?

apps/src/sites/studio/pages/levels/show.js
cearachew: Updated this

dashboard/app/controllers/activities_controller.rb
cearachew: Updated this

apps/src/templates/rubrics/RubricSettings.jsx
davidsbailey: > Just to double check, you mean that we want an event for each student in this loop right before the evaluate rubric job is called?:

correct!

> Is this event in rubric settings still useful, or should it be removed? 

it could be useful, so since it is already implemented let's keep it.

> Does the event for the button click for an individual student need to be moved to the backend as well?

let's keep this one in the frontend, since it could be useful to have this in amplitude too.

dashboard/app/controllers/activities_controller.rb
```diff
@@ -124,6 +125,16 @@ def milestone
       is_ai_experiment_enabled = current_user && Experiment.enabled?(user: current_user, script: @script_level&.script, experiment_name: 'ai-rubrics')
       is_level_ai_enabled = EvaluateRubricJob.ai_enabled?(@script_level)
       if is_ai_experiment_enabled && is_level_ai_enabled && params[:submitted] == 'true'
+        metadata = {
+          'studentId' => current_user.id,
+          'unitName' => @script_level.script.name,
+          'levelName' => @level.name,
+        }
```
davidsbailey, line 132: 💯

apps/src/templates/rubrics/RubricSettings.jsx
davidsbailey: > > Is this event in rubric settings still useful, or should it be removed?
> 
> it could be useful, so since it is already implemented let's keep it.

if you keep it, it would be handy to have a different event name, so that we can easily distinguish between the teacher doing a thing and the individual student project being evaluated.

apps/src/sites/studio/pages/levels/show.js
```diff
@@ -98,10 +100,28 @@ function initPage() {
     };
     getStore().dispatch(setTaRubric(rubric));
 
+    console.log(config);
+    console.log(rubric);
```
davidsbailey: remove debug code?

davidsbailey (approved):
looks great!

apps/src/sites/studio/pages/levels/show.js
cearachew: oof you're right

PR #57787, opened 2024-04-03
Prevent levelbuilders from changing template levels in published units

etaderhold:
This change is in response to a past incident where a change to the template level in a published unit caused students to lose access to their work. The suggestion was to prevent changes in template levels once a unit with that level has been published, similar to how level ordering can't be changed in published units. This uses the same criteria used for the existing restriction on level ordering and deletion from published scripts.

Now the template level section will look like this when editing a level that's part of a published script, giving information about what the template level is, and informing the user this cannot be modified in published content.
image

## Links

[JIRA](https://codedotorg.atlassian.net/browse/TEACH-818)
[Slack thread](https://codedotorg.slack.com/archives/C045UAX4WKH/p1704749315263649)

## PR Checklist:

- [ ] Code is well-commented


kakiha11 (approved):
LGTM! Thanks Eric.

kobryan0619 (approved):
This change does look good to me!  

This might be more of a sidebar, but I am wondering if the content of that text change has already been run through the T&L team.   The "What this does" section feels confusing to me as someone who has used Levelbuilder a bit, I imagine for a newer-to-levelbuilder-user, this would be very confusing.  That might be beyond the scope of this task, but I thought I would throw it out there.

PR #57794, opened 2024-04-04
Music: library improvements

breville:
Some improvements to the music library code, including simpler enumeration functions for use by `SoundsPanel` and `PackDialog`.


thomasoniii (approved):
lgtm 🚀

apps/src/music/player/MusicPlayer.ts
```diff
@@ -539,7 +539,7 @@ export default class MusicPlayer {
       return;
     }
 
-    for (const instrument of library.folders.filter(
+    for (const instrument of library.instruments.filter(
```
sanchitmalhotra126: I think this will need to be both `[...library.instruments, ...library.kits]`

sanchitmalhotra126 (approved):
Much cleaner! Thanks!

apps/src/music/player/MusicPlayer.ts
breville: Great catch.  Fixed.

thomasoniii (approved):

apps/src/music/player/MusicPlayer.ts
```diff
@@ -539,7 +539,7 @@ export default class MusicPlayer {
       return;
     }
 
-    for (const instrument of library.instruments.filter(
+    for (const instrument of [...library.instruments, ...library.kits].filter(
       folder => folder.type === 'instrument' || folder.type === 'kit'
     )) {
```
thomasoniii: Keeping in mind I know nothing about the data, but why is this filter necessary?

Just looking at it, it looks like it's iterating over a list of instruments and a list of kits, but then it's filtering out any folder type that isn't an instrument or a kit.

Which implies that we could have something that's not an instrument or a kit in the list of instruments and kits.

How could that happen? Is it something that should/could be prevented elsewhere?

sanchitmalhotra126: Good catch! The `filter` is old code from when we didn't separate `instrument`s and `kit`s in the library. The filter can be removed now.

breville: Thanks for the catch.  It looks like this code was removed entirely in https://github.com/code-dot-org/code-dot-org/pull/57849?

sanchitmalhotra126: Oh true, this function has been removed now since we don't need to pre-load all instrument sounds at once.

sanchitmalhotra126 (approved):

PR #57796, opened 2024-04-04
[Google Blockly] Move properties to list in constants

mikeharv:
This is a quick clean-up for the read-only and settable properties found in the Google Blockly wrapper (and its test). Since we have so many wrapped properties, they were taking up a significant portion of the increasingly long wrapper file. There were also 32 read-only properties that had been added to the wrapper but not its test. This will keep the list in sync automatically. While I was at it, I resorted them alphabetically. No functional differences. I checked that a Sprite Lab project could be loaded without errors and that the wrapper test was still passing locally before pushing.

Note: I expect a merge conflict between this branch and https://github.com/code-dot-org/code-dot-org/pull/57778. Depending on which branch merges first, I will plan to update the other branch.


apps/src/blockly/googleBlocklyWrapper.ts
```diff
@@ -236,95 +238,9 @@ function initializeBlocklyWrapper(blocklyInstance: GoogleBlocklyInstance) {
...
-  blocklyWrapper.wrapReadOnlyProperty('VARIABLE_CATEGORY_NAME');
-  blocklyWrapper.wrapReadOnlyProperty('Variables');
-  blocklyWrapper.wrapReadOnlyProperty('VariableMap');
-  blocklyWrapper.wrapReadOnlyProperty('VariableModel');
-  blocklyWrapper.wrapReadOnlyProperty('weblab_locale');
-  blocklyWrapper.wrapReadOnlyProperty('WidgetDiv');
-  blocklyWrapper.wrapReadOnlyProperty('Workspace');
-  blocklyWrapper.wrapReadOnlyProperty('WorkspaceSvg');
-  blocklyWrapper.wrapReadOnlyProperty('Xml');
+  READ_ONLY_PROPERTIES.forEach(prop => {
```
molly-moen, line 242: 🎉

molly-moen (approved):
so much neater!

PR #57805, opened 2024-04-04
Connect presentation view model card to redux

fisher-alice:
This PR connects the presentation view model card to values from the redux store.

The screencast video below shows the presentation view of the model card values are set to read-only. And then when the values are set to be editable, you can see that if the user updates a model card field, the change is reflected in the presentation view of the model card.

https://github.com/code-dot-org/code-dot-org/assets/107423305/42502623-04cf-4601-a32e-0bce49781c4b

Example topics and technical info are displayed as bulleted lists.
Screenshot 2024-04-04 at 3 38 07 PM

If there are no examples, then the phrase 'No examples available' is shown:

Screenshot 2024-04-04 at 12 53 20 PM

## Links
[jira](https://codedotorg.atlassian.net/browse/LABS-550)

## Testing story
I tested changes locally using a student account, levelbuilder account, and signed out. I checked with both editable and read-only settings.

## PR Checklist:

- [ ] Code is well-commented


apps/src/aichat/views/presentation/PresentationView.tsx
```diff
@@ -1,40 +1,65 @@
...
+  const currentAiCustomizations = useAppSelector(
+    state => state.aichat.currentAiCustomizations
+  );
+  const modelCardInfo = currentAiCustomizations.modelCardInfo;
+
+  // These are temporary constants. They will be retrieved from s3.
+  const EXAMPLE_MODEL_NAME = 'Model A';
+  const EXAMPLE_MODEL_TRAINING_DATA = 'Model A Training Data';
+
+  const TECHNICAL_INFO_VALUES: (string | number | boolean)[] = [
```
sanchitmalhotra126: might be good to store this as a map of field name -> value, instead of relying on ordering in case that changes?

apps/src/aichat/views/presentation/ModelCardRow.tsx
```diff
@@ -16,6 +16,26 @@ const ModelCardRow: React.FunctionComponent<ModelCardRowProps> = ({
...
   expandedContent,
 }) => {
+  const getExpandedContentToDisplay = () => {
+    if (Array.isArray(expandedContent)) {
+      // Remove empty strings from the array.
+      const checkedExpandedContent = expandedContent.filter(
+        content => content.length !== 0
+      );
+      if (checkedExpandedContent.length === 0) {
+        return <p>No examples available</p>;
```
sanchitmalhotra126: minor: is this only used for the "Example Topics" entry? If so, should we just explicitly check that `keyName === 'exampleTopics'`, instead of checking if the items are an array?

Alternatively; if it's meant to be reusable (if we hypothetically have other multiline fields), then this is fine, but we'd probably want to abstract out the `No examples available` default text

apps/src/aichat/views/modelCustomization/constants.ts
```diff
@@ -11,7 +11,14 @@ export const SET_TEMPERATURE_STEP = 0.1;
...
 
-export const MODEL_CARD_FIELDS_AND_LABELS: [
+export const MODEL_CARD_FIELDS_AND_LABELS: [keyof ModelCardInfo, string][] = [
+  ['description', 'Description'],
+  ['intendedUse', 'Intended Use'],
+  ['limitationsAndWarnings', 'Limitations and Warnings'],
+  ['testingAndEvaluation', 'Testing and Evaluation'],
+];
+
+export const ALL_MODEL_CARD_FIELDS_LABELS_ICONS: [
```
sanchitmalhotra126: Hmm, it would be ideal to not have to create a second array here. The 'ALL' version is technically correct; the previous version skipped `'exampleTopics'` since it wasn't a single string field and at the time we were just handling `'exampleTopics'` separately (see `ModelCardFields.tsx` for example). I wonder if we should just keep this full list and update the affected components to specifically skip the `'exampleTopics'` if it's already being rendered a different way?

apps/src/aichat/views/presentation/PresentationView.tsx
```diff
@@ -1,40 +1,65 @@
...
+    EXAMPLE_MODEL_NAME,
+    EXAMPLE_MODEL_TRAINING_DATA,
+    currentAiCustomizations.systemPrompt,
+    currentAiCustomizations.temperature,
+    currentAiCustomizations.retrievalContexts.length > 0,
+  ];
+
+  const getTechnicalInfo = () => {
```
sanchitmalhotra126: Seems like a good candidate for a `useMemo` - that way it should only update when `currentAiCustomizations` changes

apps/src/aichat/views/presentation/ModelCardRow.tsx
```diff
@@ -7,7 +7,7 @@ interface ModelCardRowProps {
   keyName: string;
   title: string;
   titleIcon: string;
-  expandedContent: string;
+  expandedContent: React.ReactNode;
```
sanchitmalhotra126: This is just the value of a model card field right? Should be either `string | string[]` or `ModelCardInfo[keyof ModelCardInfo]`?

sanchitmalhotra126 (commented):
A few comments, but nothing major

apps/src/aichat/views/presentation/ModelCardRow.tsx
fisher-alice: Yes, we have one other multiline field (Technical Info) which should not ever be blank. So still want to filter out empty strings. But I'll update to more generic text. 'Not available'. Thanks!

apps/src/aichat/views/modelCustomization/constants.ts
fisher-alice: Sure - using one array and skipping 'exampleTopics' when needed sounds good. It's currently being used in the levelbuilder edit page. We may also add 'botName' to this list as well in a follow-up.

apps/src/aichat/views/presentation/PresentationView.tsx
fisher-alice: Thanks for helping me with the typing of this!

apps/src/aichat/views/presentation/ModelCardRow.tsx
```diff
@@ -16,6 +16,26 @@ const ModelCardRow: React.FunctionComponent<ModelCardRowProps> = ({
   titleIcon,
   expandedContent,
 }) => {
+  const getExpandedContentToDisplay = () => {
```
sanchitmalhotra126: (sorry forgot to catch this on the last rev) Oh I think it would be good to use a `useMemo` here too

apps/src/aichat/views/presentation/PresentationView.tsx
```diff
@@ -1,40 +1,67 @@
...
+  const {botName, systemPrompt, temperature, retrievalContexts} =
+    currentAiCustomizations;
+  const modelCardInfo = currentAiCustomizations.modelCardInfo;
+
+  // These are temporary constants. They will be retrieved from s3.
+  const EXAMPLE_MODEL_NAME = 'Model A';
+  const EXAMPLE_MODEL_TRAINING_DATA = 'Model A Training Data';
+
+  const technicalInfo = useMemo(() => {
+    const TECHNICAL_INFO: {
```
sanchitmalhotra126: nit: lowercase since this is a local var? (maybe `technicalInfoData` or something)

sanchitmalhotra126 (approved):
LGTM! Just a couple minor comments

apps/src/aichat/views/presentation/ModelCardRow.tsx
fisher-alice: No worries! Thanks for the reminder.

PR #57806, opened 2024-04-04
P20-788: Prevent user creation if the associated LTI User Identity creation fails

artem-vavilov:
## Summary
Prevents user creation if the associated LTI User Identity creation fails by raising an error in the `after_create` callback

## Links
- JIRA ticket [P20-788](https://codedotorg.atlassian.net/browse/P20-788)


carl-codeorg (approved):
LGTM - much simpler solution that moving around the `lti_user_identity` creation!

PR #57419, opened 2024-03-20
[JIMT][CDOIDE]

thomasoniii:
brings in the code from cdo-ide-poc. You don't need to look at this yet unless you want to, I'm just curious to see if the tests pass.

Known things that have been lost upon bringing it in:
- Editing _frequently_ doesn't work - it'll move the cursor back to the start, especially if it's inside an element which is wrapped (e.g., parentheses or html tags). This is due to...something...with our code editor or how the wrapper component is invoked. Looks like it's improperly re-rendering a parent and causing it to lose focus.
- @uiw/react-codemirror cannot seem to be installed in our environment. It seems janky. :-(
- creating unexpected file types (e.g., *.txt) tends to crash it. We just need to hand in a default language module to code mirror. I assume there's a `lang-text` or the like, I just didn't look yet.

Anyway, we can iterate away now.

PR #57807, opened 2024-04-04
[JIMT][CT-479] remove unused parts of Code Bridge

thomasoniii:
Follow on to #57419 , this just tosses out a bunch of code that we're not using or planning on using any time soon.

So this'll be the PR to revert (or at least refer to) when we want to add this functionality back in.


molly-moen (approved):
🧹

PR #57808, opened 2024-04-04
Full width by default in `SimpleDropdown`

bencodeorg:
Follow-up to https://github.com/code-dot-org/code-dot-org/pull/57775, based on discussion [here](https://github.com/code-dot-org/code-dot-org/pull/57775#issuecomment-2037636273) -- this defaults the `SimpleDropdown` component to take up the full width of its parent, and allows customization of this property (eg, setting `width: auto`) by passing in a classname to `SimpleDropdown`.

## Testing story

Tested a few things:
- I looked at gen AI lab usages of this component and saw the desired behavior (full width).
- I tested passing in an override width via the `className` prop (`width: auto`), and saw width revert to the width of the content.
- I looked at one use of this component outside of gen AI lab (in the teacher panel on the right side of the page when you're looking at a level as a teacher) and saw no visible change.
- I also looked at Storybook components, which looked good as well.

### DIFF
--- a/apps/src/componentLibrary/dropdown/simpleDropdown/CHANGELOG.md
+++ b/apps/src/componentLibrary/dropdown/simpleDropdown/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 All notable changes to this project will be documented in this file.
 
+## [0.4.0](https://github.com/code-dot-org/code-dot-org/pull/57105)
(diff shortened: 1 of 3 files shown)


levadadenys (approved):
Thanks for this update! LGTM!

I'll need to add a classname to select element in the future update thoguh 😄 , but that would be only for thin vs thick styles, meaning that most of the styles of select element will remain under `select`, not a classname css selector

PR #57772, opened 2024-04-03
P20-829: Undo optimization of deprecated unit fixtures seeding

artem-vavilov:
Only relevant on Apr 3, to be merged if test build doesn't pass (apply this as an alternative to reverting https://github.com/code-dot-org/code-dot-org/pull/57743). Can delete after apr 3 if not already applied.

## Links
- JIRA ticket [P20-829](https://codedotorg.atlassian.net/browse/P20-829)
- Related PR: https://github.com/code-dot-org/code-dot-org/pull/57743

PR #57812, opened 2024-04-04
P20-829: Fix `Unit.clear_cache`

artem-vavilov:
1. Fixed `Unit.clear_cache`: since `@@script_level_cache` wasn't being cleared, the data set during the first test remained unchanged in subsequent tests.
https://github.com/code-dot-org/code-dot-org/blob/751e8529137d7b72ab4dee2dbc3321b4fd229cef/dashboard/app/models/unit.rb#L452-L459
2. Re-introduced the `seed_deprecated_unit_fixtures` optimization that was reverted here: https://github.com/code-dot-org/code-dot-org/pull/57772

## Links
- JIRA ticket [P20-829](https://codedotorg.atlassian.net/browse/P20-829)
- Related PR: https://github.com/code-dot-org/code-dot-org/pull/57743

### DIFF
--- a/dashboard/app/models/unit.rb
+++ b/dashboard/app/models/unit.rb
@@ -1723,6 +1723,7 @@ class Unit < ApplicationRecord
     raise "only call this in a test!" unless Rails.env.test?
     @@unit_cache = nil
     @@unit_family_cache = nil
+    @@script_level_cache = nil
     @@level_cache = nil
     @@all_scripts = nil
     @@visible_units = nil
--- a/dashboard/test/test_helper.rb
+++ b/dashboard/test/test_helper.rb
@@ -178,10 +178,6 @@ class ActiveSupport::TestCase
     assert_difference(args.collect(&:to_s).collect {|class_name| "#{class_name}.count"}, &block)
   end
 
-  setup_all do
-    seed_deprecated_unit_fixtures
-  end
-
   def assert_does_not_create(*args, &block)
     assert_no_difference(args.collect(&:to_s).collect {|class_name| "#{class_name}.count"}, &block)
   end


dashboard/app/models/unit.rb
```diff
@@ -1723,6 +1723,7 @@ def self.clear_cache
     raise "only call this in a test!" unless Rails.env.test?
     @@unit_cache = nil
     @@unit_family_cache = nil
+    @@script_level_cache = nil
```
artem-vavilov, line 1726: Since `@@script_level_cache` wasn't being cleared, the data set during the first test remained unchanged in subsequent tests:
https://github.com/code-dot-org/code-dot-org/blob/751e8529137d7b72ab4dee2dbc3321b4fd229cef/dashboard/app/models/unit.rb#L452-L459

artem-vavilov: It worked previously because we seeded deprecated unit fixtures before enabling Unit caching with `Unit.stubs(:should_cache?).returns(true)`:
https://github.com/code-dot-org/code-dot-org/blob/751e8529137d7b72ab4dee2dbc3321b4fd229cef/dashboard/test/integration/callouts_test.rb#L3-L18

artem-vavilov: [The error](https://github.com/code-dot-org/code-dot-org/pull/57772#pullrequestreview-1978375772) can be reproduced by running the command: `cd dashboard && RAILS_ENV=test RACK_ENV=test bundle exec rails test --seed 37263 test/integration/callouts_test.rb test/models/unit_test.rb`

Hamms (approved):
Fantastic find! Obscure interactions between two disparate unit tests can be real tough to track down; I'm so thrilled that we now know what the underlying issue was. Thank you for the investigation!

This does raise the question of why `clear_cache` was previously not clearing the ScriptLevel cache; glancing through the history of the method, however, it does seem like this was not an intentional exclusion. Looks like the method originally only cleared `@@unit_cache` (technically, `@@script_cache`), and the other cache values were added individually over time on an as-needed basis. That then raises a second question for me: **are there any other locally cached values that should also be added to `clear_cache`?**

I think that question can be addressed in a follow-up PR, though; this one looks great to me! Thank you for the diligent debugging.

PR #57814, opened 2024-04-04
Gen AI lab: update styling / content of model update messages

bencodeorg:
This PR updates the styling and content of the messages put in the chat window when a user updates their model customizations, and adds a "delete" button to each message to remove them from the window if desired.

**New model update message UI**

image

## Testing story

Tested manually as I changed different customizations.

## Follow-up work

[I started a thread in Slack](https://codedotorg.slack.com/archives/C06FELVTV0Q/p1712267186733659) about the level of detail we want to include in the messages that are sent here (eg, do we want separate messages for updates to, say, temperature and system prompt)? It sounds like we want to include one message per update, not per tab, which is what is implemented here. There's a slight difference in Figma in the language for updates to the publish tab ("has been published" rather than "has been updated"), but [I'm seeking guidance on some specifics there](https://codedotorg.slack.com/archives/C06FELVTV0Q/p1712271704257019) about how we want saves to work for that tab.

### DIFF
--- a/apps/src/aiTutor/types.ts
+++ b/apps/src/aiTutor/types.ts
@@ -64,6 +64,8 @@ export enum Role {
   ASSISTANT = 'assistant',
   USER = 'user',
   SYSTEM = 'system',
+  // only used in Aichat, but our types are currently tangled up :)
+  MODEL_UPDATE = 'update',
 }
 export type Status =
   (typeof AiTutorInteractionSaveStatus)[keyof typeof AiTutorInteractionSaveStatus];
--- a/apps/src/aichat/redux/aichatRedux.ts
+++ b/apps/src/aichat/redux/aichatRedux.ts
@@ -7,6 +7,7 @@ const registerReducers = require('@cdo/apps/redux').registerReducers;
 import {
   DEFAULT_VISIBILITIES,
   EMPTY_AI_CUSTOMIZATIONS,
+  READABLE_AI_CUSTOMIZATIONS,
 } from '../views/modelCustomization/constants';
 import {initialChatMessages} from '../constants';
 import {getChatCompletionMessage} from '../chatApi';
@@ -108,7 +109,8 @@ export const updateAiCustomization = createAsyncThunk(
         addChatMessage({
           id: 0,
           role: Role.SYSTEM,
-          chatMessageText: property,
+          chatMessageText:
+            READABLE_AI_CUSTOMIZATIONS[property as keyof AiCustomizations],
           status: Status.OK,
           timestamp: getCurrentTime(),
         })
--- a/apps/src/aichat/redux/aichatRedux.ts
+++ b/apps/src/aichat/redux/aichatRedux.ts
@@ -7,7 +7,7 @@ const registerReducers = require('@cdo/apps/redux').registerReducers;
 import {
   DEFAULT_VISIBILITIES,
   EMPTY_AI_CUSTOMIZATIONS,
-  READABLE_AI_CUSTOMIZATIONS,
+  AI_CUSTOMIZATIONS_LABELS,
 } from '../views/modelCustomization/constants';
 import {initialChatMessages} from '../constants';
 import {getChatCompletionMessage} from '../chatApi';
@@ -110,7 +110,7 @@ export const updateAiCustomization = createAsyncThunk(
           id: 0,
           role: Role.MODEL_UPDATE,
           chatMessageText:
-            READABLE_AI_CUSTOMIZATIONS[property as keyof AiCustomizations],
+            AI_CUSTOMIZATIONS_LABELS[property as keyof AiCustomizations],
           status: Status.OK,
           timestamp: getCurrentTime(),
         })
@@ -197,16 +197,12 @@ const aichatSlice = createSlice({
       state.chatMessages.push(newMessage);
     },
     removeChatMessage: (state, action: PayloadAction<number>) => {
-      const updatedMessages = [...state.chatMessages];
-      const messageToRemovePosition = updatedMessages.findIndex(
-        message => message.id === action.payload
+      const updatedMessages = state.chatMessages.filter(
+        message => message.id !== action.payload
       );
-      if (!messageToRemovePosition) {
-        return;
+      if (updatedMessages.length !== state.chatMessages.length) {
+        state.chatMessages = updatedMessages;
       }
-      updatedMessages.splice(messageToRemovePosition, 1);
-
-      state.chatMessages = updatedMessages;
     },
     clearChatMessages: state => {
       state.chatMessages = initialChatMessages;
--- a/apps/src/aichat/redux/aichatRedux.ts
+++ b/apps/src/aichat/redux/aichatRedux.ts
@@ -53,6 +53,7 @@ const findChangedProperties = (
 };
 
 const getCurrentTimestamp = () => moment(Date.now()).format('YYYY-MM-DD HH:mm');
+const getCurrentTime = () => moment(Date.now()).format('LT');
 
 export interface AichatState {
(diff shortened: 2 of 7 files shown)


apps/src/aichat/views/chatMessage.module.scss
```diff
@@ -1,35 +1,64 @@
...
+}
+
+.check {
+  color: $light_affirmative_500;
+  font-size: 16px;
 }
 
 .userMessage {
-    background-color: $lightest_cyan;
+  background-color: $lightest_cyan;
```
bencodeorg, line 51: Just some spacing fixes :)

apps/src/aichat/views/ChatMessage.tsx
```diff
@@ -1,16 +1,23 @@
 import React from 'react';
-import moduleStyles from './chatMessage.module.scss';
+import {useSelector} from 'react-redux';
 import classNames from 'classnames';
-import aichatI18n from '../locale';
+
+import {useAppDispatch} from '@cdo/apps/util/reduxHooks';
```
bencodeorg, line 5: Just some reordering of imports into global -> absolute -> relative

apps/src/aichat/types.ts
```diff
@@ -13,6 +13,7 @@ export enum Role {
   ASSISTANT = 'assistant',
   USER = 'user',
   SYSTEM = 'system',
+  MODEL_UPDATE = 'update',
```
bencodeorg, line 16: I think we'll need to filter out this new role / message type in what we send to our back end...LMK if anyone has alternative ideas on where/how we should store these model update messages.

apps/src/aichat/views/modelCustomization/compare-models-dialog.module.scss
```diff
@@ -14,7 +14,7 @@
 }
 
 div.modelComparisonDialog {
-  height: calc(100vh - 100px);
+  height: calc(90vh - 100px);
```
bencodeorg, line 17: Unrelated follow-up to https://github.com/code-dot-org/code-dot-org/pull/57775, turns out that at full screen this was overflowing on the bottom. Looks great at smaller heights (and now at full height!)

apps/src/aichat/redux/aichatRedux.ts
```diff
@@ -190,6 +196,18 @@ const aichatSlice = createSlice({
...
+      const messageToRemovePosition = updatedMessages.findIndex(
+        message => message.id === action.payload
+      );
+      if (!messageToRemovePosition) {
+        return;
+      }
+      updatedMessages.splice(messageToRemovePosition, 1);
+
+      state.chatMessages = updatedMessages;
+    },
```
thomasoniii: this is fine as is, but I think it could also be simplified to:
```
const updatedMessages = state.chatMessages.filter(m => m.id !== action.payload)
if (updatedMessages.length !== state.chatMessages.length) {
  state.chatMessages = updatedMessages;
}
```

(or some other wordier variation with temporary variables in the middle)

Instead of copying the array and finding something to delete, we just create a new array that's filtered to keep everything _except_ the single message we're looking for.

The length check between the arrays just ensures that we only overwrite in redux if we actually removed something. If we'd be guaranteed that we're only removing a chat message that actually exists, we could just assign the filtered value straight to `state.chatMessages`.

Dealer's choice as to whether you actually think this is simpler. 😉

thomasoniii (approved):
minor non-blocking filtering suggestion, but otherwise lgtm 🚀

apps/src/aiTutor/types.ts
```diff
@@ -64,6 +64,8 @@ export enum Role {
   ASSISTANT = 'assistant',
   USER = 'user',
   SYSTEM = 'system',
+  // only used in Aichat, but our types are currently tangled up :)
+  MODEL_UPDATE = 'update',
```
sanchitmalhotra126, line 68: Can we just use the `Role` type in `aichat/types`, or do we need to update this one too?

apps/src/aichat/views/modelCustomization/constants.ts
```diff
@@ -53,3 +53,13 @@ export const DEFAULT_LEVEL_AICHAT_SETTINGS: LevelAichatSettings = {
   visibilities: DEFAULT_VISIBILITIES,
   hidePresentationPanel: false,
 };
+
+export const READABLE_AI_CUSTOMIZATIONS: {
```
sanchitmalhotra126: super nit: what do you think about naming this something like `AI_CUSTOMIZATIONS_LABELS` or `TRANSLATED_AI_LABELS` since these will be eventually coming from translated strings?

apps/src/aichat/types.ts
sanchitmalhotra126: Yeah, I think when we invoke the chat API we'll just filter out messages of this type.

apps/src/aichat/views/ChatMessage.tsx
```diff
@@ -91,13 +94,42 @@ const displayAssistantMessage = (status: string, chatMessageText: string) => {
   }
 };
 
+const displayModelUpdateMessage = (
```
sanchitmalhotra126, line 97: minor: thoughts on making this a separate component (maybe still lives in this file) as opposed to a render function?

sanchitmalhotra126 (approved):
Looks good! A few minor comments

apps/src/aiTutor/types.ts
bencodeorg: I get this TS error unless I add it:

```
ERROR in ./src/aichat/redux/aichatRedux.ts:159:7
TS2345: Argument of type 'import("/Users/benjaminbrooks/code-dot-org/apps/src/aichat/types").ChatCompletionMessage[]' is not assignable to parameter of type 'import("/Users/benjaminbrooks/code-dot-org/apps/src/aiTutor/types").ChatCompletionMessage[]'.
  Type 'import("/Users/benjaminbrooks/code-dot-org/apps/src/aichat/types").ChatCompletionMessage' is not assignable to type 'import("/Users/benjaminbrooks/code-dot-org/apps/src/aiTutor/types").ChatCompletionMessage'.
    Types of property 'role' are incompatible.
      Type 'import("/Users/benjaminbrooks/code-dot-org/apps/src/aichat/types").Role' is not assignable to type 'import("/Users/benjaminbrooks/code-dot-org/apps/src/aiTutor/types").Role'.
        Property 'MODEL_UPDATE' is missing in type 'import("/Users/benjaminbrooks/code-dot-org/apps/src/aiTutor/types").Role'.
    157 |       newMessageId,
    158 |       message,
  > 159 |       appropriateChatMessages
        |       ^^^^^^^^^^^^^^^^^^^^^^^
    160 |     );
    161 |
    162 |     // Find message in chatMessages and update status.
```

I think because the chatApi is used in aichatRedux, and the type referenced in chatApi is from aiTutor, not aichat. I was kind of hoping that some of the type refactor work @ebeastlake is doing [here](https://github.com/code-dot-org/code-dot-org/pull/57669) might help, but not actually sure it will. In any case, feels like something we can disentangle down the line.

apps/src/aichat/views/ChatMessage.tsx
bencodeorg: I think for consistency with other very similar render functions (displayAssistantMessage, etc) I'll leave here for now, but definitely up to refactor later if preferred.

apps/src/aichat/views/modelCustomization/constants.ts
bencodeorg: Much better, thanks!

apps/src/aichat/redux/aichatRedux.ts
bencodeorg: I like it, updated!

PR #57819, opened 2024-04-04
Switch Database GTID mode to ON

sureshc:
To facilitate upgrade of the production database to Aurora 3 with minimal downtime we will create a new production cluster that replicates via the MySQL binary log from the existing cluster. Switch the Global Transaction Identifier (GTID) Mode on the database from `ON_PERMISSIVE` to `ON` to allow us to use the `mysql.rds_set_external_master_with_auto_position` [stored procedure](https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/mysql-stored-proc-replicating.html#mysql_rds_set_external_master_with_auto_position) to configure the new production cluster to start replicating from the existing production cluster from the correct position in the binary log. This will make it easier to automate the provisioning of the new production cluster and, specifically, to automate configuration of the new cluster to replicate from the existing cluster.

Without GTID Mode set to `ON` we have to use `mysql.rds_set_external_master` [stored procedure](https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/mysql-stored-proc-replicating.html#mysql_rds_set_external_master) to configure replication and that requires reading through the logs on the new cluster and determining exactly which transaction in the binary log we need to start replicating from.

More on Global Transaction Identifiers (GTID) here https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/mysql-replication-gtid.html

## Testing story

Successfully tested by provisioning an adhoc off of this feature branch that also enables binary logging so that the two GTID settings are actually utilized: https://github.com/code-dot-org/code-dot-org/pull/58005

## Deployment strategy

This is a static ParameterGroup setting will trigger restart of all database instances in the cluster on all environments except for production. Because the production database cluster is not (yet) provisioned via CloudFormation, this change will NOT automatically restart the production database instances, but we will need to manually restart the reader and writer database instances for the change to take effect. We’ll experience about 30 seconds of down time when we restart the writer database instance and if we carry out that action on a Saturday evening (Pacific Time) that will cause about 100 HTTP requests to fail, which is relatively low.

## PR Checklist:

- [ ] Code is well-commented


aws/cloudformation/components/database.yml.erb
```diff
@@ -131,7 +131,8 @@
         # https://github.com/code-dot-org/code-dot-org/blob/654d0cf90dd65848fdd045df3b9e4f85903d9f68/pegasus/test/fixtures/fake_dashboard.rb#L257-L271
         # Disable binary logging to improve write performance.
         binlog_format: 'OFF'
-        gtid-mode: ON_PERMISSIVE
+        gtid_mode: 'ON'
```
sureshc: How long have we had this bug due to the web console [displaying the wrong key name](https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/mysql-replication-gtid.html) for this configuration setting? 
image

sureshc: OK. Not a defect. The ParameterGroup setting name is `gtid-mode`

PR #57826, opened 2024-04-05
[Google Blockly] Field Angle Plugin

mikeharv:
CDO Blockly includes several angle fields that are used with our Artist lab:
image

Mainline Blockly's plugin library also includes an angle field:
image

While these feature obviously different UI, they are reasonably close. Currently, any artist level that includes turn blocks (ie. all of them) cannot be loaded due to missing fields. This branch adds Blockly's plugin for angle fields and extends each in the barest form possible just to get them working. The result is a fully usable field, albeit one that doesn't quite look or work how our users would expect.

https://github.com/code-dot-org/code-dot-org/assets/43474485/79e6bf5e-0d87-4f4a-9455-045315af5ac6

## Links

Jira - https://codedotorg.atlassian.net/browse/CT-477

## Follow-up work

I have a follow-up branch that is able to make the following customizations:
* If the turn direction dropdown is set to "Right", the angle will increase in a clockwise direction
* Differentiate between "text input" and "dropdown" fields
* If the block XML includes config for valid values, they are respected.

Additional follow-ups:
* Override `dropdownCreate()` to customize the editor UI and/or CSS as desired.
* Ensure field serializes correctly to JSON (for projects)
* Ensure field works as expected in all supported contexts (?)
* Other? (Not totally sure what else yet since this is just the first baby step.)

## PR Checklist:

- [ ] Code is well-commented


molly-moen (approved):
Nice that this already semi-exists!

PR #57836, opened 2024-04-05
allow teacher to rerun ai eval when previous attempt resulted in error

davidsbailey:
FInishes https://codedotorg.atlassian.net/browse/AITT-450

before, run button is disabled in error case:

Screenshot 2024-04-05 at 11 45 06 AM

after, run button is enabled in error case, and rerunning succeeds:

https://github.com/code-dot-org/code-dot-org/assets/8001765/26def329-e0b5-4f4a-87c4-07c41b7b451e

## Testing story

* updated unit test
* manual verification to produce screenshots

### DIFF
--- a/apps/src/templates/rubrics/RunAIAssessmentButton.jsx
+++ b/apps/src/templates/rubrics/RunAIAssessmentButton.jsx
@@ -59,6 +59,7 @@ export default function RunAIAssessmentButton({
     return i18n.runAiAssessment();
   };
 
+  // fetch initial status
   useEffect(() => {
     if (!!rubricId && !!studentUserId) {
       fetchAiEvaluationStatus(rubricId, studentUserId).then(response => {
@@ -94,6 +95,7 @@ export default function RunAIAssessmentButton({
     }
   }, [rubricId, studentUserId, setStatus]);
 
+  // poll for status updates
   useEffect(() => {
     if (polling && !!rubricId && !!studentUserId) {
       const intervalId = setInterval(() => {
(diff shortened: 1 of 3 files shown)


apps/src/templates/rubrics/RunAIAssessmentButton.jsx
```diff
@@ -169,7 +171,7 @@ export default function RunAIAssessmentButton({
             color={Button.ButtonColor.neutralDark}
             onClick={handleRunAiAssessment}
             style={{margin: 0}}
-            disabled={status !== STATUS.READY}
+            disabled={status !== STATUS.READY && status !== STATUS.ERROR}
```
cearachew, line 174: I couldn't see in the video, but according to this it's still disabled after a successful run, right?

cearachew (commented):
Is there any concern about openAI cost if there are many errors/ reruns on a project? Afaik not being able to parse the format of the returned data from openai is one source of general errors and the same input rerun could be more likely to have the same formatting errors in the output, if that makes sense

cearachew (approved):
Anyway, the prev question about cost is more "just a thing that occurred to me" than a reason to not merge this. This looks good! :shipit:

apps/src/templates/rubrics/RunAIAssessmentButton.jsx
davidsbailey: correct. that is the `SUCCESS` state

Screenshot 2024-04-05 at 1 35 54 PM

davidsbailey:
> Is there any concern about openAI cost if there are many errors/ reruns on a project? Afaik not being able to parse the format of the returned data from openai is one source of general errors and the same input rerun could be more likely to have the same formatting errors in the output, if that makes sense

Good point... we have been kicking the can down the road on this one. one way to address this is tracked here: https://codedotorg.atlassian.net/browse/AITT-257

apps/src/templates/rubrics/RunAIAssessmentButton.jsx
davidsbailey: sorry, the screenshot I showed is for the `ALREADY_EVALUATED` state, but I did previously confirm this is also disabled for the `SUCCESS` state which we show after evaluation succeeds.

PR #57838, opened 2024-04-05
Integrate Tabs into My PL page

TurnerRiley:
Sets up the framework for the tabs (using the new `Tabs` component) that will be used on the My PL page. It currently logs in the console which tab is being selected but does not currently change what content is being shown (this is outside the scope of this ticket). I tried to proactively set up the logic framework with TODOs of what work remains for filtering which content is shown based on the user's permissions and which tab they have selected.

### Note about design
One note about how I set this up is that I'm not using the `tabContent` tag which would normally be passed in with each tab's `value` and `text`. This is because in the Figma it shows the tabs being essentially part of the banner with the content below, so I figured it might be easier to have the tabs be at the bottom of the banner, and then just have a `useState` variable that is tracking the `currentTab` to know which content to show.

The alternative would be to use the `tabContent` tag and dynamically pass in which content should be shown based on the user permissions, but then we would need to style the top of the `Tabs` component to look like it's part of the banner and style the content to look like its not (which would be inconsistent with how it actually looks in our code, making it harder to read/understand how to make updates). After discussing with Hannah, I think the better approach will be what I have now to have the structure of it match our mental models (and Figma models) of it having the tabs in the header and the content separate and below.

### Demo that it looks the same for now (only showing one tab for now, and we only display the tabs if there's more than one)
Capture

## Links
Jira ticket: [here](https://codedotorg.atlassian.net/jira/software/c/projects/ACQ/boards/64?assignee=60d62161f65054006980bd71&selectedIssue=ACQ-1641)
Product spec: [here](https://docs.google.com/document/d/1Ee02kw_ei2x6wOS06K768dC-kUhGDlydjoE6BZipbPE/edit#bookmark=kix.njp3v1ls2g5r)
Figma: [here](https://www.figma.com/file/sZ6d1uhJWFksJ6j3Fnmbnc/PL-Flow-Improvements?type=design&node-id=1958-8045&mode=design&t=48PPodQzYznD4Vwo-0)

## Testing story
Local testing.

## Follow-up work
The commented TODO items to have user permissions and the selected tab affect what content is being shown.


apps/src/code-studio/pd/professional_learning_landing/LandingPage.jsx
```diff
@@ -13,6 +13,35 @@ import HeaderBannerNoImage from '@cdo/apps/templates/HeaderBannerNoImage';
 import TwoColumnActionBlock from '@cdo/apps/templates/studioHomepages/TwoColumnActionBlock';
 import ActionBlocksWrapper from '@cdo/apps/templates/studioHomepages/ActionBlocksWrapper';
 import style from './landingPage.module.scss';
+import Tabs from '@cdo/apps/componentLibrary/tabs';
+
+const DEFAULT_SELECTED_TAB = 'myPL';
+
+const getAvailableTabs = () => {
+  // [TODO]: return a subset of the tabs below based on the user's permission level
```
bethanyaconnor, line 18: I love this approach, but I don't think we want to launch with showing them all. I know this was a bit of a weird task in that sense, but I think, for now, we should only show the `myPL` tab and we can add the rest as we implement them?

I think all I'm asking for here is to remove/comment out all but the first option. The strings and everything can stay!

apps/src/code-studio/pd/professional_learning_landing/LandingPage.jsx
```diff
@@ -98,7 +131,24 @@ export default function LandingPage({
       <HeaderBannerNoImage
         headingText={i18n.professionalLearning()}
         backgroundColor={color.light_gray_50}
-      />
+      >
+        {availableTabs && (
```
bethanyaconnor: I think if we only have one tab, we shouldn't show any tabs, so we might want this to be `availableTabs > 1`. (`currentTab` in that case could default to the first of `availableTabs` and never be changeable.)

bethanyaconnor (commented):
Thank you for hopping on this! I'm sorry I wasn't super clear on this task but this looks really good in the screenshot. Have one comment about rollout and one about a quick design piece.

Love that this should make each of the following tasks to add these tabs much easier 😍

apps/src/code-studio/pd/professional_learning_landing/landingPage.module.scss
```diff
@@ -8,6 +9,22 @@ main.wrapper {
   margin-inline: auto;
 }
 
+section {
+  padding: 4rem 1rem 0 1rem !important;
```
bethanyaconnor, line 13: why does this need to use the `!important` tag?

bethanyaconnor (approved):
One question about the `!important` tag but otherwise looks good to me! Thanks!

apps/src/code-studio/pd/professional_learning_landing/LandingPage.jsx
```diff
@@ -95,10 +127,26 @@ export default function LandingPage({
...
   return (
     <>
-      <HeaderBannerNoImage
-        headingText={i18n.professionalLearning()}
-        backgroundColor={color.light_gray_50}
-      />
+      <div className={`${headerContainerStyles} ${style.headerContainer}`}>
+        <HeaderBannerNoImage headingText={i18n.professionalLearning()}>
+          {availableTabs.length > 1 && (
+            <div className={style.myPlTabsContainer}>
```
kelbyhawn: This could be changed to a `<nav>` element for better accessibility.

apps/src/code-studio/pd/professional_learning_landing/landingPage.module.scss
kelbyhawn: (Unrelated to Bethany's comment above) Thanks for adding this! Can you remove the inline style on line 162 now that this is here?

kelbyhawn (approved):
Exciting, thanks for doing this! 🎉 Left two non-blocking comments.

apps/src/code-studio/pd/professional_learning_landing/landingPage.module.scss
TurnerRiley: I believe it's because I wasn't able to apply styles to the `HeaderBannerNoImage`'s intrinsic `section` tag so I had to override it in here. If there's a better way to do this I'm definitely open to pushback!

apps/src/code-studio/pd/professional_learning_landing/landingPage.module.scss
```diff
@@ -8,6 +9,26 @@ main.wrapper {
...
+
+.headerWithoutTabsContainer {
+  padding-bottom: 4rem;
+}
+
+.myPlTabsContainer {
+  margin-top: 1.5rem;
+}
+
+.recommendedPlContainer {
```
kelbyhawn: This can be removed since that `<section>` element will adopt the `section` styles you added on line 12. Sorry, I should've been more clear in my original comment!

kelbyhawn (approved):

PR #57844, opened 2024-04-05
Music: skip progression

breville:
If a music or panels level specifies the `skipUrl` property, the level will now offer a "Skip to project" button.  When pressed, a confirmation dialog will be shown, and if the user confirms, they will be taken to the specified URL.

https://github.com/code-dot-org/code-dot-org/assets/2205926/541575ae-2290-4c42-bed0-1abfd1334e0e

The button is identical in both level types.


apps/src/panels/PanelsLabView.tsx
```diff
@@ -27,6 +31,9 @@ const PanelsLabView: React.FunctionComponent = () => {
   const currentAppName = useAppSelector(
     state => state.lab.levelProperties?.appName
   );
+  const skipUrl = useAppSelector(state => state.lab.levelProperties?.skipUrl);
```
sanchitmalhotra126, line 34: Should we provide authoring support for this in levelbuilder? Or maybe we want to hold off since it's a very specific use case?

sanchitmalhotra126 (approved):
Nice solution! I do think we might eventually want to make the SkipDialog a little more generic if it's meant to be reusable by other labs, but that can happen once other labs actually start using it.

apps/src/panels/PanelsLabView.tsx
breville: I was thinking about it, and certainly open to it, but happy to do as a follow-up if/when we find a need for it.

PR #57847, opened 2024-04-05
Fix Violations of and Reenable `Style/RescueModifier`

Hamms:
[The official Ruby Style guide](https://rubystyle.guide/#no-rescue-modifiers) recommends avoiding the use of `rescue` in its modifier form. It does somewhat beg the question of why include a modifier form in the first place, but that's just by the by.

Quoting from Rubocop documentation, this rule:

> Checks for uses of `rescue` in its modifier form is added for following reasons:
> - The syntax of modifier form `rescue` can be misleading because it might lead us to believe that `rescue` handles the given exception but it actually rescue all exceptions to return the given rescue block. In this case, value returned by handle_error or SomeException.
> - Modifier form `rescue` would rescue all the exceptions. It would silently skip all exception or errors and handle the error.  Example: If `NoMethodError` is raised, modifier form rescue would handle the exception.

Initial fix applied automatically with `bundle exec rubocop --autocorrect-all --only Style/RescueModifier`; I then manually ran a few more autocorrect commands to fix up some of the indentation and layout errors produced by the initial autocorrection.

In general, I think `Style/` Cops are always optional, so I'm quite open to pushback on enabling this rule if folks want to continue to be able to use `rescue` in its modifier form. I lean towards being in favor of enabling it, myself; although I appreciate the aesthetics of inlining a rescue clause, I put a lot of weight on the official Ruby style guide's recommendations, and the reasons laid out in Rubocop's documentation seem compelling.

What do y'all think?

## Warning!!

## Links

- https://rubystyle.guide/#no-rescue-modifiers
- https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/RescueModifier
- https://docs.rubocop.org/rubocop/cops_style.html#stylerescuemodifier

## Testing story

## PR Checklist:

- [ ] Code is well-commented


dashboard/legacy/middleware/net_sim_api.rb
```diff
@@ -549,7 +549,9 @@ def parse_table_map_from_query_string(query_string)
 def parse_ids_from_query_string(query_string)
   [].tap do |ids|
     CGI.parse(query_string)['id[]'].each do |id|
-      ids << Integer(id, 10) rescue ArgumentError
```
davidsbailey, line 552: I have to say, this looks like a potential misuse of the "modifier" syntax -- it looks like the caller wanted to rescue this type of error, rather than insert this error class into the list of returned ids. in fact it looks like this error was introduced specifically when switching to the modifier syntax: https://github.com/code-dot-org/code-dot-org/commit/d04d4120d30b6220332a616ccec430b68dbdd72f

davidsbailey:
While I do like the original syntax because it allows you to write more concisely, the somewhat unexpected behavior of the syntax does make it a bit of a land mine if you are new to it and do not take the time to look up what it does (see inline comment). 

since the obscurity of the syntax makes it harder to read and somewhat error prone, I'm inclined to vote yes on this PR. however I am also interested to hear what others think :-)

davidsbailey (approved):

dashboard/legacy/middleware/net_sim_api.rb
Hamms: Oh, great catch! Definitely agree with that analysis; I'm tempted to manually update this fix to reflect that presumed intent rather than the current functionality.

cat5inthecradle (approved):

PR #57849, opened 2024-04-06
ToneJS: setup sampler instruments on demand

sanchitmalhotra126:
## Warning!!

## Description

This adds the ability to load/set up sampler instruments for the ToneJS player on demand (instead of all at once, like we do currently). There are a couple of nuances to this to make sure the user experience remains consistent, and this does add some extra loading time whenever a user opens up a new instrument for the first time, because we need to cache all the samples for the instrument at once. Some details:
- whenever an instrument block is added to the project (or we open a project with an instrument block) we will auto-load that instrument as part of the precaching process. We compile all the samples needed for the instruments and batch load them with the other samples to have a consistent loading bar.
- if an instrument block is dragged out but not connected to the when run block, we'll start loading the instrument as soon as the field is opened up.
- The loads do take a bit of time, but once we are using ToneJS, we could definitely strip down the number of samples. I experimented with reducing the number of piano samples from 36 to 10 (one sample every major third) and the playback still sounded quite good with a much faster load time. See video below for an example.

Note: I'm also realizing the name 'instrument' is very overloaded here, sometimes meaning specifically instruments in the "play chord" block or sometimes referring to all non-'pack' sampler-based sounds (both 'kits' and 'instruments'). I tried to use 'sampler' to be more generic but it's definitely not consistent, so that's something we can clean up later as well. Let me know if anything reads confusingly though.

**Loading new instruments** (caching is off for this example)

https://github.com/code-dot-org/code-dot-org/assets/85528507/c4a887df-6f93-46dd-8b90-e3b71b71e853

**Loading instruments with reduced samples**: compare the load time of Piano/Pluck to Rhodes (caching also off here)

https://github.com/code-dot-org/code-dot-org/assets/85528507/e42f03fd-1b1d-48ad-9ca1-9bc3fb8886c2

 
## Links

https://codedotorg.atlassian.net/browse/LABS-531

## Testing story

Tested this on a standalone project and intro progression with & without ToneJS. 

## Follow-up work

I'd love to go back in and consolidate/clean up some of the types (between fields.js, FieldChord/Pattern, and Chord/PatternPanel) since there's a lot of repetition, but that'll likely require some JS-TS conversions so I held off for now.

### DIFF
--- a/apps/src/music/blockly/FieldChord.ts
+++ b/apps/src/music/blockly/FieldChord.ts
@@ -6,6 +6,7 @@ import {ChordEventValue} from '../player/interfaces/ChordEvent';
 import MusicLibrary from '../player/MusicLibrary';
 import {getNoteName} from '../utils/Notes';
 import {generateGraphDataFromChord, ChordGraphNote} from '../utils/Chords';
+import {LoadFinishedCallback} from '../types';
 const experiments = require('@cdo/apps/util/experiments');
 const color = require('@cdo/apps/util/color');
 
@@ -20,6 +21,15 @@ interface FieldChordOptions {
   previewNote: (note: number, instrument: string, onStop?: () => void) => void;
   cancelPreviews: () => void;
   currentValue: ChordEventValue;
+  setupSampler: (
+    instrument: string,
+    onLoadFinished?: LoadFinishedCallback
+  ) => void;
+  isInstrumentLoading: (instrument: string) => boolean;
+  isInstrumentLoaded: (instrument: string) => boolean;
+  registerInstrumentLoadCallback: (
+    callback: (instrumentName: string) => void
+  ) => void;
 }
 
 /**
@@ -185,10 +195,8 @@ export default class FieldChord extends Field {
       React.createElement<ChordPanelProps>(ChordPanel, {
         library: this.options.getLibrary(),
         initValue: this.getValue(),
-        previewChord: this.options.previewChord,
-        previewNote: this.options.previewNote,
-        cancelPreviews: this.options.cancelPreviews,
         onChange: this.onValueChange,
+        ...this.options,
       }),
       this.newDiv
     );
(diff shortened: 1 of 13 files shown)


breville (approved):
This is great.

I played with it and tried to break things but couldn't :).  Even changing patterns and chords while playback was underway worked just as expected.

I do agree that the nomenclature is confusing.  We currently have "instruments" (for chords) and "kits" (for patterns) so I think it would be good to find a new term (maybe "samplers" as you suggest) for this new concept.

The loading overlay is nice.  The only thing I noticed is that I could tab-navigate the panel's UI behind it.  I did try pressing enter on a couple elements, but wasn't sure whether they were disabled or just didn't do anything meaningful while in that state.

Finally, I think it will be worth reducing load times.  With caching disabled, even on a decent internet connection, it takes a while to load all these sounds.  Come to think of it, do we give the browser the opportunity to parallelize those loads, or are they strictly serialized?

sanchitmalhotra126:
> I do agree that the nomenclature is confusing. We currently have "instruments" (for chords) and "kits" (for patterns) so I think it would be good to find a new term (maybe "samplers" as you suggest) for this new concept.

Agreed. I'd like to do a follow-up to clean up some of the repeated types, so can try to address this then too.
 
> The loading overlay is nice. The only thing I noticed is that I could tab-navigate the panel's UI behind it. I did try pressing enter on a couple elements, but wasn't sure whether they were disabled or just didn't do anything meaningful while in that state.

Oh good catch! Let me see if I can block that as well (in follow-up).

> Finally, I think it will be worth reducing load times. With caching disabled, even on a decent internet connection, it takes a while to load all these sounds. Come to think of it, do we give the browser the opportunity to parallelize those loads, or are they strictly serialized?

Great question - I actually don't know 🙂 let me double check though because if they are serialized, we could potentially optimize this. Regardless, I do think we can cut the number of instrument samples by ~3/4 with this.

PR #57850, opened 2024-04-06
Music: more UI testing

breville:
The Music Lab UI test now does a little more: it opens the sound selection UI, chooses a pack, and chooses a sound.

Follow-up to https://github.com/code-dot-org/code-dot-org/pull/57783.

### DIFF
--- a/dashboard/test/ui/features/star_labs/musiclab/musiclab_drag_block.feature
+++ b/dashboard/test/ui/features/star_labs/musiclab/musiclab_drag_block.feature
@@ -27,7 +27,22 @@ Scenario Outline: Dragging play sound block
   # There should now be a music timeline entry.
   And element ".timeline-element" is visible
 
+  # Click the field inside the attached "play sound" block.
+  And I click block field "[data-id='when-run-block'] > [data-id='play_sound_at_current_location_simple2'] > .blocklyEditableText"
+
+  # Click on the second pack inside the sounds panel.
+  And I click selector "#sounds-panel .sounds-panel-folder-row:nth-of-type(2)"
+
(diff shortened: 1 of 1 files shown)


thomasoniii (approved):
lgtm 🚀

dashboard/test/ui/features/star_labs/musiclab/musiclab_drag_block.feature
```diff
@@ -27,7 +27,22 @@ Scenario Outline: Dragging play sound block
...
+  And I click selector "#sounds-panel .sounds-panel-sound-row:nth-of-type(2)"
+
+  # The sounds panel should be dismissed.
+  And I wait until element "#sounds-panel" is not visible
+
+  # There should still be a music timeline entry.
+  And element ".timeline-element" is visible
+
 Examples:
   | url                                                       | test_name               |
```
sanchitmalhotra126, line 46: Should we add a standalone project link here too? (not exactly sure how those work in UI tests, since presumably one would have to be created)

sanchitmalhotra126 (approved):
Thanks for adding this!

dashboard/test/ui/features/star_labs/musiclab/musiclab_drag_block.feature
breville: Absolutely.  This will also help test the  `launch2024` music library in addition to `intro2024`.  I tried it out in this PR, but came across the issue that the `Then I wait until I am on`  step doesn't work because of the dynamic URL. 
 Would love to figure this out in a small follow-up.

PR #57862, opened 2024-04-08
Move botName from AiCustomizations to ModelCardInfo

fisher-alice:
This PR moves the property `botName` from `AiCustomizations` to `ModelCardInfo` so that the bot name is updated in the 'Publish' tab instead of the 'Setup' tab. I also updated the name of `PromptCustomization` to `SetupCustomization` to match the user view of component.

## Before update:

SETUP:
Screenshot 2024-04-08 at 10 57 08 AM

PUBLISH:
Screenshot 2024-04-08 at 10 57 18 AM

LEVELBUILDER EDIT:
Screenshot 2024-04-08 at 10 57 53 AM

## After update

SETUP:
Screenshot 2024-04-08 at 10 54 09 AM

PUBLISH:
Screenshot 2024-04-08 at 12 05 15 PM

LEVELBUILDER EDIT:
Screenshot 2024-04-08 at 10 53 39 AM

## Links

[jira](https://codedotorg.atlassian.net/browse/LABS-674)

## Follow-up work

Currently, the 'Update' button is not viewable unless a user zooms out. This is true for the 'Retrieval' panel as well when there are several retrieval statements. Update so text content is scrollable while button sticks to bottom of panel.

Screenshot 2024-04-08 at 12 06 24 PM

Screenshot 2024-04-08 at 11 01 58 AM

## PR Checklist:

- [ ] Code is well-commented


apps/src/aichat/views/ModelCustomizationWorkspace.tsx
```diff
@@ -1,28 +1,28 @@
 import React from 'react';
 
 import Tabs, {Tab} from './tabs/Tabs';
-import PromptCustomization from './modelCustomization/PromptCustomization';
```
bencodeorg, line 4: Thanks for renaming this :)

dashboard/config/levels/custom/aichat/AI Chat Test Level.level
```diff
@@ -48,29 +47,25 @@
     },
     "aichat_settings": {
       "initialCustomizations": {
-        "botName": "Recycler Chatbot",
```
bencodeorg, line 51: Should we delete these to avoid future confusion?

dashboard/config/levels/custom/aichat/AI Chat Test Level.level
```diff
@@ -48,29 +47,25 @@
...
-            "What are everyday items I use that I can recycle?",
-            "Communities in Tucson focused on recycling"
-          ]
+
+          ],
+          "botName": "Recycling Bot"
         }
       },
       "visibilities": {
-        "botName": "readonly",
```
bencodeorg, line 73: Delete this too?

bencodeorg (commented):
I saw you have this in follow-up work, but I think we probably want the bot name to be an `input` rather than a `textarea`. Might be worth just doing that now (I think you could just copy over the HTML you're deleting from `SetupCustomization`?

fisher-alice:
> I saw you have this in follow-up work, but I think we probably want the bot name to be an `input` rather than a `textarea`. Might be worth just doing that now (I think you could just copy over the HTML you're deleting from `SetupCustomization`?

Sure! Just updated.

apps/src/aichat/views/modelCustomization/PublishNotes.tsx
```diff
@@ -26,17 +27,22 @@ const PublishNotes: React.FunctionComponent = () => {
...
+    return property === 'botName' ? 'input' : 'textarea';
+  };
+
   return (
     <div className={styles.verticalFlexContainer}>
-      <div>
-        {MODEL_CARD_FIELDS_LABELS_ICONS.map(([property, label]) => {
-          return (
-            isVisible(visibility) && (
+      {isVisible(visibility) && (
```
sanchitmalhotra126, line 36: nit: might be cleaner to invert and just return an empty div if `isVisible(visibility)` is false?

sanchitmalhotra126 (approved):
Thanks for doing this!

bencodeorg (approved):
LGTM! If not too much trouble, could you update the second AIChat level as well?

apps/src/aichat/views/modelCustomization/PublishNotes.tsx
fisher-alice: Talked offline with @sanchitmalhotra126 , and we agreed that leaving as-is makes sense. Thanks!

PR #57863, opened 2024-04-08
Handle HTTP exceptions from weather API's

cat5inthecradle:
## Warning!!

---

We're receiving the following error in Honeybadger:

```
Failed to open TCP connection to api.openweathermap.org:80 (execution expired) (Net::OpenTimeout)
```

We're not catching that error, which is in the middle of a loop making several HTTP calls. This PR captures failed HTTP connections and handles the same way we're handling non-200 responses - by raising a more explicit Honeybadger error continuing on with the script. Currently, the script is exiting and we are collecting no data when the API call fails.

Counterpoint: All of these calls are just hitting the same API with different parameters. This code change might instead result in _more_ Honeybadger errors because this timout likely indicates a problem with the service which will only continue failing. This PR may result in partial data collection with several cites missing from the dataset. The current behavior might be preferred, as no data might be better than partial data.

I'm interested in thoughts on how to handle this better. Perhaps this is a good candidate for an async job that tries again later, or have it run more often but exit early if we already have data for today.

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/bin/cron/applab_datasets/daily_weather
+++ b/bin/cron/applab_datasets/daily_weather
@@ -163,7 +163,8 @@ def get_weather_data
 
     url = "http://api.openweathermap.org/data/2.5/forecast/daily?zip=#{office.zip_code},us&cnt=#{NUM_DAYS}&units=imperial&appid=#{API_KEY}"
     response = Net::HTTP.get_response(URI(url))
-    if response.code != '200'
+    raise "Response #{response} from #{office.city}" if response.code != '200'
+    rescue
       Honeybadger.notify(
         error: response.message,
         error_message: "Failed to retrieve data from Open Weather API for use with App Labs datasets.",


bin/cron/applab_datasets/daily_weather
```diff
@@ -163,7 +163,8 @@ def get_weather_data
 
     url = "http://api.openweathermap.org/data/2.5/forecast/daily?zip=#{office.zip_code},us&cnt=#{NUM_DAYS}&units=imperial&appid=#{API_KEY}"
     response = Net::HTTP.get_response(URI(url))
-    if response.code != '200'
+    raise "Response #{response} from #{office.city}" if response.code != '200'
```
Hamms, line 166: I think you'll need to wrap this in an explicit `begin`, rather than trying to rescue as part of the `each` block.

```ruby
    begin
      response = Net::HTTP.get_response(URI(url))
      raise "Response #{response} from #{office.city}" if response.code != '200'
    rescue
      Honeybadger.notify(
```

Hamms (commented):
> The current behavior might be preferred, as no data might be better than partial data.

I think from a perspective of making this change an an experiment to see what kind of data we do get, there's still plenty of value in only getting partial data. This change seems reasonable to me (with one tweak; see below), at least for the short term!

PR #57869, opened 2024-04-08
Add SelfPacedProgressTable to My PL LandingPage component

kelbyhawn:
Adds the `SelfPacedProgressTable` component to https://studio.code.org/my-professional-learning
- This section only shows up when a teacher has Self-paced PL course activity
  - Updated the unit tests for this page and tested locally to make sure this works
- Created a `constants.js` file to share test data between Storybook and `landingPageTest.js` unit tests (thanks for the tip @hannahbergam!)
- Also added the `tableStyles.scss`  stylesheet to this page so the _My Workshops_ table matches the updated UI
  - This will get an overhaul soon too

**Followup:** the existing self-paced courses section on the Teacher homepage will be removed and replaced with a link to check this page to see self-paced course progress ([Jira ticket](https://codedotorg.atlassian.net/browse/ACQ-1716))

**Jira ticket:** [ACQ-1715](https://codedotorg.atlassian.net/browse/ACQ-1715)

----

## Has Self-Paced PL Courses
localhost-studio code org_3000_my-professional-learning

## New Teacher (w/o Self-Paced courses)
localhost-studio code org_3000_my-professional-learning


apps/src/code-studio/pd/professional_learning_landing/LandingPage.jsx
```diff
@@ -73,6 +75,7 @@ export default function LandingPage({
           text: i18n.plLandingGettingStartedButton(),
         },
       ]}
+      marginBottom={'0'}
```
kelbyhawn, line 78: Added this to remove the bottom margin from the Getting Started `TwoColumnActionBlock` component used on this page for new teachers.

apps/src/code-studio/pd/professional_learning_landing/landingPage.module.scss
```diff
@@ -10,11 +10,12 @@ main.wrapper {
 }
 
 section {
-  padding: 4rem 1rem 0 1rem !important;
+  padding: 4rem 0 0 !important;
 }
 
 .headerContainer {
   background-color: color.$light_gray_50;
+  padding: 0 1rem;
```
kelbyhawn, line 18: Did a little bit of refactoring here related to [this PR](https://github.com/code-dot-org/code-dot-org/pull/57838) since the inline padding was being applied to all the `<section>`s on the page. Here's a screenshot of the tabs showing to confirm this didn't break, but these will not show up on the page yet!

Screenshot 2024-04-09 at 11 03 02 AM

apps/src/code-studio/pd/professional_learning_landing/selfPacedProgressTable.module.scss
```diff
@@ -17,6 +17,7 @@ table.selfPacedProgressTable {
   height: 0.5rem;
   border-radius: 1rem;
   margin-block-start: 0.25rem;
+  overflow: hidden;
```
kelbyhawn, line 20: This fixes a little UI bug on the progress bar if it gets close to the end:

| Before | After |
| ---- | ---- |
| Before | After |

apps/src/code-studio/pd/professional_learning_landing/SelfPacedProgressTable.jsx
```diff
@@ -39,7 +39,7 @@ const CourseRow = ({
             {percent_completed}% {i18n.selfPacedPlCompleted()}
           </BodyThreeText>
           {/* Progress bar */}
-          <div className={styles.progressBar}>
+          <div className={styles.progressBar} data-testid="progress-bar">
```
hannahbergam, line 42: Just a thought for future/followup work- it would be super cool if this had some sore of accessible readout (i.e. label) of progress! I know that isn't a focus for this page but always something to keep in mind in case it's easy to add.

apps/src/code-studio/pd/professional_learning_landing/selfPacedProgressTable.module.scss
hannahbergam: That attention to detail🔥

hannahbergam (approved):
Yay! Looks great! One small comment but non-blocking

apps/src/code-studio/pd/professional_learning_landing/SelfPacedProgressTable.jsx
kelbyhawn: I thought about adding something like that! But it says "(percent)% Completed" right above it so I thought it would be redundant 😅

hannahbergam: oo does a screen reader read that out loud? I'm not sure if this is what you're referring to, but there is a way to add an aria-label that doesn't show, but a screen reader would still read

kelbyhawn: Oh it doesn't! VoiceOver on Mac only reads links and buttons. I checked this on the Teacher homepage, and it does the same thing there too (only reads links and buttons). I'll add a ticket for this in the next AccHack.

**Edit:** made a ticket: https://codedotorg.atlassian.net/browse/A11Y-281

hannahbergam: Okay- it should be as simple as wrapping the text in a <label>. But yes- thanks for making a ticket!

kelbyhawn: ~Neither the paragraph or progress bar is a link, so it'd need to be something else.~ Misread that to say "wrap in an a (link tag)" 🤦‍♀️  I'm too scattered today to look into it, but will look tomorrow! This makes me excited for the Deque training, since I assumed screen readers read all the text on a page 🥲

hannahbergam: Ahh yes okay I'm following! To be clear, folks using screen readers use the tab key to scan through interactive elements (like you mentioned- buttons, links, input fields, etc), but they can also use keys to navigate by header size (and drill down into reading the text directly if they choose). This is a super interesting case because A11y-wise, really we should be treating this div as an image with an alt text. Happy to brainstorm together at some point!

kelbyhawn: > use keys to navigate by header size (and drill down into reading the text directly if they choose

A ha! That makes sense, thanks for explaining. This is an interesting case…maybe the `<figure>` and `<figcaption>` elements would work? ([MDN docs](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/figure_role)) 

Let's def brainstorm on this, and thanks for bringing it up!

PR #57871, opened 2024-04-08
LTI sync summary updates v2

carl-codeorg:
Pt. 2 of the teacher-related updates for the sync confirmation modal. The [previous PR](https://github.com/code-dot-org/code-dot-org/pull/57695) updated the format to be a table instead of a list of sections. This one adds the following:

- Changes the Primary Teacher column from just the name of the teacher to a dropdown that includes all instructors for that section. This allows the user to solve situations where the section owner is ambiguous. Canvas sections can have multiple instructors, but none are considered "primary" in the LMS.
- Adds a new route to the LTI controller to update sections. I wasn't sure where to put this route, but I ended up putting into an Lti::V1::LtiSectionsController. This avoids adding to one of the two existing Sections controllers and contains this LTI-specific route to an LTI controller. Let me know if you have ideas for a better location for this route.
- Changes the `Continue` button in the modal to post to the new route, which will update the section owners.

Example screenshots:

Section B here has two teachers:
(The width on the dropdowns is uneven. A [PR](https://github.com/code-dot-org/code-dot-org/pull/57808) was merged to make them 100% width, but it had to be reverted. Assuming a fixed PR will eventually be merged, so I'm leaving the dropdowns as-is for now)
Screenshot 2024-04-08 at 3 49 12 PM

Here I change the teacher and submit:
Screenshot 2024-04-08 at 3 49 28 PM

Now `Carl Teacher2` is the primary teacher, and the old section owner has been "demoted" to a co-teacher:
Screenshot 2024-04-08 at 3 50 45 PM

## Links

[JIRA](https://codedotorg.atlassian.net/browse/P20-804)

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/lib/ui/lti/sync/LtiSectionSyncDialog.tsx
+++ b/apps/src/lib/ui/lti/sync/LtiSectionSyncDialog.tsx
@@ -138,6 +138,9 @@ export default function LtiSectionSyncDialog({
   };
 
   const handleUpdateSectionOwners = () => {
+    if (!syncResult.changed || Object.keys(syncResult.changed).length === 0) {
+      return handleClose();
+    }
     return $.ajax({
       url: '/lti/v1/sections/bulk_update_owners',
       type: 'PATCH',
--- a/apps/src/lib/ui/lti/sync/LtiSectionSyncDialog.tsx
+++ b/apps/src/lib/ui/lti/sync/LtiSectionSyncDialog.tsx
@@ -247,12 +247,16 @@ export default function LtiSectionSyncDialog({
             labelText={i18n.ltiSectionSyncDialogHeaderPrimaryInstructor()}
             isLabelVisible={false}
             size={'xs'}
-            onChange={e =>
+            onChange={e => {
+              e.persist();
               setSectionOwners(owners => ({
                 ...owners,
-                [section.lti_section_id]: parseInt(e.target.value, 10),
-              }))
-            }
+                [section.lti_section_id.toString()]: parseInt(
+                  e.target.value,
+                  10
+                ),
+              }));
+            }}
           />
         </td>
         <td style={styles.tableCellNumber}>{section.size}</td>
--- a/apps/src/lib/ui/lti/sync/LtiSectionSyncDialog.tsx
+++ b/apps/src/lib/ui/lti/sync/LtiSectionSyncDialog.tsx
@@ -6,11 +6,13 @@ import DialogFooter from '@cdo/apps/templates/teacherDashboard/DialogFooter';
 import SafeMarkdown from '@cdo/apps/templates/SafeMarkdown';
 import {LmsLinks} from '@cdo/apps/util/sharedConstants';
 import Spinner from '@cdo/apps/code-studio/pd/components/spinner';
+import {SimpleDropdown} from '@cdo/apps/componentLibrary/dropdown';
 import {
   LtiSection,
   LtiSectionMap,
   LtiSectionSyncDialogProps,
   LtiSectionSyncResult,
+  LtiSectionOwnerMap,
   SubView,
 } from './types';
 import PropTypes from 'prop-types';
@@ -29,6 +31,20 @@ export default function LtiSectionSyncDialog({
 }: LtiSectionSyncDialogProps) {
   const initialView = syncResult.error ? SubView.ERROR : SubView.SYNC_RESULT;
   const [currentView, setCurrentView] = useState<SubView>(initialView);
+  const [sectionOwners, setSectionOwners] = useState<LtiSectionOwnerMap>(() => {
+    const sectionOwners: LtiSectionOwnerMap = {};
+    if (syncResult.changed) {
+      Object.values(syncResult.changed).forEach(section => {
+        const sectionOwner = section.instructors.find(
+          instructor => instructor.isOwner
+        );
+        if (sectionOwner) {
+          sectionOwners[section.lti_section_id] = sectionOwner.id;
+        }
+      });
+    }
+    return sectionOwners;
+  });
 
   const handleClose = () => {
     setCurrentView(SubView.SPINNER);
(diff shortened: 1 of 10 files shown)


dashboard/app/controllers/lti/v1/lti_sections_controller.rb
```diff
@@ -0,0 +1,23 @@
+module Lti
+  module V1
+    class LtiSectionsController < ApplicationController
```
artem-vavilov: It's already organized within the `Lti` namespace
```suggestion
    class SectionsController < ApplicationController
```

dashboard/config/routes.rb
```diff
@@ -612,6 +612,7 @@
     namespace :lti do
       namespace :v1 do
         resource :feedback, controller: :feedback, only: %i[create show]
+        post '/sections/owners', to: 'lti_sections#bulk_update_owners'
```
artem-vavilov: Please use the [PATCH Rest API method](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/PATCH) for partial updating of a resource:
```suggestion
        resources :sections, only: [] do
          collection do
            patch :bulk_update_owners
          end
        end
```

apps/src/lib/ui/lti/sync/LtiSectionSyncDialog.tsx
```diff
@@ -120,6 +136,18 @@ export default function LtiSectionSyncDialog({
       PLATFORMS.STATSIG
     );
   };
+
+  const handleUpdateSectionOwners = async () => {
+    return $.post({
+      url: '/lti/v1/sections/owners',
```
artem-vavilov:
```suggestion
      url: '/lti/v1/sections/bulk_update_owners',
```

apps/src/lib/ui/lti/sync/LtiSectionSyncDialog.tsx
```diff
@@ -120,6 +136,19 @@ export default function LtiSectionSyncDialog({
...
   };
+
+  const handleUpdateSectionOwners = () => {
+    return $.ajax({
+      url: '/lti/v1/sections/bulk_update_owners',
+      type: 'PATCH',
+      data: {
+        section_owners: sectionOwners,
+      },
+      success: () => {
```
stephenliang, line 150: Optional: Just for my own education, do we have a way to report failures to Honeybadger or some other system? (To help with debugging if needed)

artem-vavilov (approved):

apps/src/lib/ui/lti/sync/LtiSectionSyncDialog.tsx
```diff
@@ -198,13 +227,30 @@ export default function LtiSectionSyncDialog({
...
+    }));
     return (
       <tr key={id} role={'gridcell'}>
         <td style={styles.tableCellText}>{section.short_name}</td>
-        <td style={styles.tableCellText}>{sectionOwnerName}</td>
+        <td style={styles.tableCellText}>
+          <SimpleDropdown
+            name={`instructor-dropdown-${id}`}
+            items={instructorOptions}
+            selectedValue={sectionOwners[section.lti_section_id].toString()}
```
stephenliang:
```suggestion
            selectedValue={
              sectionOwners[section.lti_section_id]?.toString() ?? undefined
            }
```

Theoretically this would throw if `sectionOwners[section.lti_section_id]` returns falsy.

apps/src/lib/ui/lti/sync/LtiSectionSyncDialog.tsx
```diff
@@ -198,13 +227,30 @@ export default function LtiSectionSyncDialog({
...
+            labelText={i18n.ltiSectionSyncDialogHeaderPrimaryInstructor()}
+            isLabelVisible={false}
+            size={'xs'}
+            onChange={e =>
+              setSectionOwners(owners => {
+                return Object.assign({}, owners, {
+                  [section.lti_section_id]: parseInt(e.target.value, 10),
+                });
```
stephenliang: nit: can be shortened

```suggestion
              setSectionOwners(owners => ({
                ...owners,
                [section.lti_section_id]: parseInt(e.target.value, 10),
              }))
```

apps/src/lib/ui/lti/sync/LtiSectionSyncDialog.tsx
carl-codeorg: One thing we could do is instrument the route with a honeybadger call; otherwise any unhandled errors will still get reported to honeybadger, but possibly with less context.

stephenliang: Do you mean on the rails side to capture errors? This should work for backend originated errors but not for things like browser side network issues. I'm OK with pushing it as, it's just that if it fails I think the user would be stuck on the sync screen which they may interpret as frozen (no feedback an error happened or what to do).

dashboard/app/controllers/lti/v1/sections_controller.rb
```diff
@@ -0,0 +1,36 @@
+require 'metrics/events'
+
+module Lti
+  module V1
+    class SectionsController < ApplicationController
+      before_action :authenticate_user!
```
stephenliang, line 6: Do we also need to verify that the current user is one of the teachers in the section? Otherwise it could possibly allow any arbitrary teacher to update any section.

carl-codeorg: Within the route code we do `authorize! :manage, section` for each section we're trying to update. It saves us having to do the lookup in the before_action hook.

carl-codeorg: The permissions for :manage are in https://github.com/code-dot-org/code-dot-org/blob/staging/dashboard/app/models/ability.rb

apps/src/lib/ui/lti/sync/LtiSectionSyncDialog.tsx
carl-codeorg: Oh I see what you mean, yeah you're right. We could potentially piggyback on the functionality we use to pass specific errors back to the front-end during sync.

nicklathe (approved):

stephenliang (approved):

PR #57872, opened 2024-04-08
Music: load sounds in parallel

sanchitmalhotra126:
## Warning!!

## Description

Parallelize loading of sounds in our sound cache layer. @breville had a great observation on https://github.com/code-dot-org/code-dot-org/pull/57849, asking if we load sounds in parallel. Turns out we were loading sounds strictly serially, and by instead parallelizing loads, naturally we can speed up load times quite a bit. This is especially relevant given the previous PR, since when using instruments in ToneJS, we need to pre-cache all samples for the instrument before it can be used. Here are a few A/B comparisons:

### Loading multiple sounds (non-instrument/kit)

**Before** (production, caching disabled):

https://github.com/code-dot-org/code-dot-org/assets/85528507/598e0341-edb9-49f4-b27c-de53f32373b8

**After** (localhost, caching disabled):

https://github.com/code-dot-org/code-dot-org/assets/85528507/553493a2-9208-4b59-877f-04cf2d8cbd46

### Loading samplers (instruments/kits) with ToneJS

**Before** (localhost on staging, caching disabled):

https://github.com/code-dot-org/code-dot-org/assets/85528507/d32d7305-33b5-43ca-afbb-2354015baa61

**After** (localhost, caching disabled):

https://github.com/code-dot-org/code-dot-org/assets/85528507/efdf6aba-9f89-4071-b4f6-8bbf8ae8a51d

Note that we could further reduce instrument load time by reducing the number of samples, since ToneJS's sampler allows us to only use a subset.

## Testing story

Tested locally with different project types

### DIFF
--- a/apps/src/music/player/SoundCache.ts
+++ b/apps/src/music/player/SoundCache.ts
@@ -38,7 +38,7 @@ class SoundCache {
       updateLoadProgress?: (progress: number) => void;
     } = {}
   ): Promise<void> {
-    const failedSounds = [];
+    const failedSounds: string[] = [];
     const {onLoadFinished, updateLoadProgress} = callbacks;
     const startTime = Date.now();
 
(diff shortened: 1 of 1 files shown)


breville (approved):
This is great, and a nice variation on the existing code.  Thanks for investigating!

Presumably all modern browsers will do the right thing about how many requests actually run simultaneously.

PR #57875, opened 2024-04-08
Serve pyodide from code.org

molly-moen:
## Warning!!

## Summary

By default, Pyodide packages are downloaded from the Pyodide CDN. This could potentially cause issues for students behind firewalls. The Pyodide npm package already has the main pyodide files we need, we just needed to do a couple tweaks to make it accessible:
- register a `wasm` mime type
- update the url we give to `loadPyodide`

After those two changes, the main pyodide files (which include a `wasm` file as well as a few others) are correctly fetched from a `studio.code.org` url. However, any non-standard libraries (which previously would be fetched from the pyodide cdn) now will not work until we download the appropriate `.whl` file and serve it from the same url as the main pyodide files. To prove out this path, I added the numpy `.whl` file (downloaded from the pyodide GitHub release) and added it to `apps/lib/pyodide`. I then copied it to the same output file as the main pyodide files via the Gruntfile. This is similar to how we serve files for droplet and blockly.

## Links

- jira ticket: [CT-420](https://codedotorg.atlassian.net/browse/CT-420)

## Testing story
Tested locally that you can use pyodide and numpy, and the files are fetched from a code.org url.

## Follow-up work
- Enable more packages. Our current list is numpy, pandas, scipy, and matplotlib, which are all available in the pyodide github release. [ticket](https://codedotorg.atlassian.net/browse/CT-447)
- This will break matplotlib, added in #57782. I will plan to fast-follow to re-enable matplotlib by adding the appropriate `.whl` files. [ticket](https://codedotorg.atlassian.net/browse/CT-488)

## PR Checklist:

- [ ] Code is well-commented


apps/lib/pyodide/README.md
```diff
@@ -0,0 +1,9 @@
+# Python Packages
+We use [Pyodide](https://pyodide.org) to run Python on our site. In order to serve up python packages from a code.org endpoint
+(to avoid firewall issues), any package outside of the standard library needs to be copied to this folder as a `.whl` file.
+Many packages are already built for Pyodide and their `.whl` file can be retrieved from the
+[Pyodide release](https://github.com/pyodide/pyodide/releases) that matches our current version. Otherwise you will need to download
+or create a wheel and add it here (this path has not been tested yet).
```
cat5inthecradle:
```suggestion
or create a _wheel_ and add it here (this path has not been tested yet, please document that process here if you do).
```
:stuck_out_tongue_winking_eye:

apps/src/pythonlab/pyodideWebWorker.js
```diff
@@ -2,7 +2,7 @@ import {loadPyodide} from 'pyodide';
 
 async function loadPyodideAndPackages() {
   self.pyodide = await loadPyodide({
-    indexURL: 'https://cdn.jsdelivr.net/pyodide/v0.24.1/full',
+    indexURL: '/assets/js/pyodide/',
```
cat5inthecradle: Would there be value versioning this on our end as well?

cat5inthecradle (approved):

apps/src/pythonlab/pyodideWebWorker.js
molly-moen: The source of truth for the version is in [package.json](https://github.com/code-dot-org/code-dot-org/blob/9ca8eb33defacfb151e3d9fd0523a09badecd991/apps/package.json#L262). We don't need to put a version here because it will always be copied to the same folder.

ebeastlake (approved):

apps/src/pythonlab/pyodideWebWorker.js
Hamms: What about cache-busting? I believe we heavily cache `/assets/js` paths in cloudformation, and rely on sprockets adding md5 hashes to all the the filenames. Will that work here?

molly-moen: Good call out. We won't be updating this frequently (maybe every few months to six months). I don't think we have control of the end of the url here (this is the prefix that the filenames will get appended to). So I think that means we can't cache-bust? Is there a way when we upgrade to force the cache to clear?

Hamms: It is possible to force-clear the cache on a deploy, but it's not ideal. Would it be feasible to update Gruntfile.js to put the files in a version-specific directory? Then we could do something like `indexURL: '/assets/js/pyodide/v0.24.1/'` here, similar to what we were doing before.

Hamms: Looks like we can even fetch the version programmatically! https://github.com/pyodide/pyodide/blob/fd80c7b1ed3ebfba4b136574b1e787984b2ef9fb/src/js/pyodide.ts#L20

molly-moen: Ooh thanks! Got it working so it's versioned now!

PR #53625, opened 2023-09-05
Add sorting by family name to progress table (TEACH-564)

etaderhold:
This change adds a "Sort by:" dropdown to the progress table. It uses the same sorting logic as on the teacher panel student table.

Screenshot (sorted by family name)
image

Screenshot (sorted by display name)
image

## Links
- jira ticket: [TEACH-564](https://codedotorg.atlassian.net/browse/TEACH-564)

## PR Checklist:

- [ ] Tests provide adequate coverage
- [ ] Privacy and Security impacts have been assessed
- [ ] Code is well-commented
- [ ] New features are translatable or updates will not break translations
- [ ] Relevant documentation has been added or updated
- [ ] User impact is well-understood and desirable
- [ ] Pull Request is labeled appropriately
- [ ] Follow-up work items (including potential tech debt) are tracked and linked

PR #57877, opened 2024-04-08
Revert "Revert "Full width by default in `SimpleDropdown`""

bencodeorg:
Resuscitates https://github.com/code-dot-org/code-dot-org/pull/57808, which was reverted [here](https://github.com/code-dot-org/code-dot-org/pull/57808) due to an eyes failure -- details in [this Slack thread](https://codedotorg.slack.com/archives/C0T0PNTM3/p1712600463000159). The text in a dropdown (family name vs. display name) on the Progress tab of the Teacher Dashboard was being slightly cut off :

image

New change in [ea7dd64](https://github.com/code-dot-org/code-dot-org/pull/57877/commits/ea7dd6498a93054ca835b4203636984a4b18f396) removes a hard coded width on the container div.

~~@etaderhold looks like you added this logic originally, any recollections on why the hard coded width is set? The dropdown will now expand to hold a long bit of text as one of the options, whereas it would have been cut off previously, not sure if that's a bad idea:~~

image

Alternatively, I could just make the hard coded width a bit larger, so LMK if that's preferable.

**Update:** I ended up going with this approach ^^^ -- looks like it still aligns properly (ie, takes 15px for the margin, then the rest of the space for the dropdown) 🤷 

image

### DIFF
--- a/apps/src/componentLibrary/dropdown/simpleDropdown/CHANGELOG.md
+++ b/apps/src/componentLibrary/dropdown/simpleDropdown/CHANGELOG.md
@@ -3,7 +3,8 @@
 All notable changes to this project will be documented in this file.
 
 ## [0.4.0](https://github.com/code-dot-org/code-dot-org/pull/57105)
-* use `width: 100%` instead of `width: auto` as default when styling `select` element. 
+* use `width: 100%` instead of `width: auto` as default when styling `select` element.
+* style select element with the use of `select` css selector instead of `.dropdown`
 
 ## [0.3.0](https://github.com/code-dot-org/code-dot-org/pull/57105)
 * moved `SimpleDropdown` to dropdown folder
--- a/apps/src/componentLibrary/dropdown/simpleDropdown/CHANGELOG.md
+++ b/apps/src/componentLibrary/dropdown/simpleDropdown/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 All notable changes to this project will be documented in this file.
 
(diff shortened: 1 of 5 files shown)


bencodeorg:
Ooh I think I found a relevant discussion in this PR:

https://github.com/code-dot-org/code-dot-org/pull/53625#discussion_r1319011129

etaderhold (commented):
I don't 100% recall the reason for the fixed width, but I'd guess it's so that the display name selector would line up with the column for student names and so that the "Lessons attempted in..." header would be aligned with the main table content. I'd suggest consulting with Molly Peredo and/or @dju90 before pushing this header over to the right.

bencodeorg:
Thanks Eric! Just went for a simpler approach that keeps the header lined up, LMK what you think.

etaderhold (approved):
Making the box wide enough for the text while maintaining alignment LGTM. Thanks!

levadadenys (approved):
LGTM!
Had a couple more `.dropdown` occurences that I've replaced and added a message into changelog re replacing `.dropdown` with `select` in scss

levadadenys:
Also re usage of wide dropdowns and truncating the text - will discuss it with design team, cause there might be places where we want to keep dropdown not too wide, but just truncating the text. The best solution is probably to control the width of the dropdown by setting the parent width (as we do it right now) and just add  `text-overflow: ellipsis;`.

Meaning that we should be good to go with SimpleDropdown like this now, and after discussion with design team I'll updating it if needed in a separate PR

bencodeorg:
Nice, thanks @levadadenys!

PR #57879, opened 2024-04-08
Gen AI Lab: add bot icon and clear button to chat window

bencodeorg:
Adds a bot icon from our design team to the chat window, as well as a clear button in the header.

**After**

image

**Before**

image


apps/src/aichat/views/AichatView.tsx
```diff
@@ -8,17 +8,37 @@ import {useAppDispatch, useAppSelector} from '@cdo/apps/util/reduxHooks';
...
+    <div>
+      <Button
+        onClick={onClear}
+        text="Clear"
+        iconLeft={{iconName: 'paintbrush'}}
+        size="xs"
+        color="white"
+        type="secondary"
+      />
+      <CopyButton />
```
levadadenys, line 37: I see that `CopyButton` is using an old version of `Button` component, meaning we have here a `<button>` inside of a `<button>`. What do you think about refactoring a `CopyButton` to use our DesignSystem new `Button` component, which will make it a little bit more clear?

levadadenys (commented):
~~One question re CopyButton, other than that LGTM!~~
Sorry, I thought `CopyButton` is rendered as a child of `Button`, must be me being not completely awaken 😆

levadadenys (approved):
LGTM!

nit for the future: (Maybe we already have this in jira) Maybe update the CopyButton to use new DesignSystem Button (for better ui)?

bencodeorg:
@levadadenys added to Jira here: https://codedotorg.atlassian.net/browse/LABS-680

fisher-alice (approved):
Nice!

sanchitmalhotra126 (approved):
Nice! Just to double check are we / @markabarrett okay with the two different button styles (for copy and clear)?

bencodeorg:
> Nice! Just to double check are we / @markabarrett okay with the two different button styles (for copy and clear)?

@sanchitmalhotra126 oh the styling is definitely not perfect, I expect we'll continue to iterate/improve here (eg, the header background itself is black here, but gray in the mocks, so we'd ideally restyle the header and also change the color of the clear button as well). The copy button was actually not in the mocks, but [Sam suggested](https://codedotorg.slack.com/archives/C06FELVTV0Q/p1712598076220549) we keep it around for the pilot since it already exists.

PR #57902, opened 2024-04-09
Update Curriculum Recommender fake curricula time stamps

TurnerRiley:
Yesterday, there were [flaky Curriculum Recommender unit tests](https://codedotorg.slack.com/archives/C0T0PNTM3/p1712624576741659). Nearly all the values are hard-coded to test the recommenders so my guess is that the flakiness came from setting the published dates and comparing them. If a curriculum was published within 1 or 2 years ago, [it gains a certain number of points when being scored](https://github.com/code-dot-org/code-dot-org/blob/staging/apps/src/util/curriculumRecommender/curriculumRecommender.js#L252-L258). Since the test curricula had hard-coded published dates that were right on the cusps (e.g. Today minus one year exactly), I think sometimes the test would run and a given curriculum would have been considered published within one year ago, and sometimes not (thus, why the failing test showed the curricula scores in the wrong order).

This PR increases each published date by one month so that there are no dates right on the cusp of scoring points or not.

## Links
Slack discussion: [here](https://codedotorg.slack.com/archives/C0T0PNTM3/p1712624576741659)

## Testing story
Tested locally and it's still passing.

### DIFF
--- a/apps/test/unit/util/curriculumRecommenderTestCurricula.js
+++ b/apps/test/unit/util/curriculumRecommenderTestCurricula.js
@@ -229,7 +229,7 @@ const FULL_TEST_COURSES = [
     school_subject: 'math',
     published_date: moment
       .utc()
-      .subtract(6, 'months')
+      .subtract(7, 'months')
       .format(UTC_PUBLISHED_DATE_FORMAT),
     device_compatibility:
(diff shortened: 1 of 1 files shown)


hannahbergam (approved):
LGTM, though admittedly I didn't check to make sure they're a month in the correct direction (trusting drone here!)

PR #57903, opened 2024-04-09
[Google Blockly] strip user code

mikeharv:
In older Blockly labs (Maze, Artist, Play Lab), there are some differences between the raw generated code that we feed into the JS interpreter and that which we show the student. For example, in these labs, additional code is used to prevent us from attempting to execute infinite loops. The generated code can also include block ids and lab-specific namespaces. Here is an example program on CDO Blockly and the associated "Show Code" modal:
|image|image|
|-|-|

Here is the same program on Google Blockly:
|image|image|
|-|-|

Note that the more complicated-looking code is still what we actually execute. CDO Blockly was accomplishing this cleanup with a call to a custom `strip` function which removes or modifies some stuff through regular expressions. We can actually use the same `strip` function to process code that Google Blockly generates, so I moved it (and some related constants) into our utils file. Some of the constants were actually already defined in both wrappers which felt a bit redundant.

The only functional change relates to differences in how block ids are generated.

Previously, it was sufficient to remove block ids by looking for a series of numbers:

`.replace(/(,\s*)?'block_id_\d+'\)/g, ')')`

The `'block_id_\d+'` part above matches the literal string `'block_id_'` followed by `\d+`, which matches one or more digits. To make it match any characters (except quotes), I replaced the latter with `[^']+`. 

With the changes in this branch, the code we show matches the CDO Blockly baseline. (Nit: there's a slight indentation difference inside the loop, but I think that's fine.)
image

## Warning!!

## Testing story

In addition to testing the indented behavior with the change, I also made sure I was still able to execute my code in this Maze level (with both versions of Blockly), as well as a couple random projects of other labs. No regressions found. The changes to the code is only for some very specific matched expressions, so this doesn't feel overly risky.

Making this change also required reverting a migration-related Minecraft change: https://github.com/code-dot-org/code-dot-org/pull/55993/commits/3c8902bd93dd86fc183976d7abf178ed19987f91

In order to get Minecraft migrated, I switched from using `Generator.blocksToCode` to `getWorkspaceCode()`. This change worked because there was no difference in the two, and was "needed" because `blocksToCode` wasn't defined yet. However, the generated code here is passed to the interpreter, so it needs to be the raw, un-stripped code. If we remove the infinite loop logic from the generated code given to the interpreter, then it's easy for students to get into frozen state (see pics). For this reason, it makes sense to go back to using the generator function, now that it has been defined.
image

image

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/craft/agent/craft.js
+++ b/apps/src/craft/agent/craft.js
@@ -712,7 +712,13 @@ export default class Craft {
     });
 
     // Run user generated code, calling appCodeOrgAPI
-    const code = Blockly.getWorkspaceCode();
+    let code = '';
+    let codeBlocks = Blockly.mainBlockSpace.getTopBlocks(true);
+    if (studioApp().initializationBlocks) {
+      codeBlocks = studioApp().initializationBlocks.concat(codeBlocks);
+    }
+
+    code = Blockly.Generator.blocksToCode('JavaScript', codeBlocks);
     CustomMarshalingInterpreter.evalWith(code, {
       moveForward: function (blockID) {
         appCodeOrgAPI.moveForward(
--- a/apps/src/blockly/cdoBlocklyWrapper.js
+++ b/apps/src/blockly/cdoBlocklyWrapper.js
@@ -12,17 +12,13 @@ import {
   setHasIncompatibleSources,
 } from '@cdo/apps/redux/blockly';
 import * as blockUtils from '../block_utils';
-import {handleCodeGenerationFailure} from './utils';
+import {
+  INFINITE_LOOP_TRAP,
+  LOOP_HIGHLIGHT,
+  handleCodeGenerationFailure,
+  strip,
+} from './utils';
 import {MetricEvent} from '@cdo/apps/lib/metrics/events';
-
-const INFINITE_LOOP_TRAP =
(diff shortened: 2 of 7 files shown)


apps/src/craft/agent/craft.js
```diff
@@ -712,7 +712,13 @@ export default class Craft {
     });
 
     // Run user generated code, calling appCodeOrgAPI
-    const code = Blockly.getWorkspaceCode();
+    let code = '';
+    let codeBlocks = Blockly.mainBlockSpace.getTopBlocks(true);
```
molly-moen, line 716: would it be worthwhile to move this code into a wrapper function so we don't have to repeat it? Does the reference to `studioApp().initializationBlocks` mean we can't move this into the wrapper?

mikeharv: Potentially... although the changes in these craft files are really just a straight revert of https://github.com/code-dot-org/code-dot-org/commit/3c8902bd93dd86fc183976d7abf178ed19987f91

There's quite a few variations of the logic here in the various lab files, so I think it would take some thought to figure out exactly what we want to abstract and move to the wrapper.

molly-moen: 👍 makes sense

molly-moen (approved):

ebeastlake (approved):

PR #57909, opened 2024-04-09
Fix Facebook icon not loading in Eyes tests - V2

kelbyhawn:
This PR is a continuation of https://github.com/code-dot-org/code-dot-org/pull/56321 to try and fix a flaky eyes test on the certificates page Facebook icon. Increases the wait time from 2 seconds to 5 seconds.

💬 [Slack convo](https://codedotorg.slack.com/archives/C04540KNGDQ/p1712694021852239)


bethanyaconnor (approved):

breville:
Interestingly, this icon appears to load slowly because we are waiting for the reachability test for Facebook to complete, [here](https://github.com/code-dot-org/code-dot-org/blob/7f48357216759fa65f0233bed97b50a70e8fbf9f/apps/src/templates/certificates/SocialShare.jsx).  If we are happy for the Eyes test to test our UI, and not the reachability of Facebook, we could add support for an optional, internal URL parameter so that Eyes tests always show the icon, which will then render immediately.

PR #57913, opened 2024-04-09
Music: use pack bpm/key after choosing

sanchitmalhotra126:
## Warning!!

## Description

Use the BPM and key from the selected pack once chosen. We'll read the bpm/key from the folder directly, but if it's not specified, we'll use the bpm/key from the first sound in the pack that has the relevant information.

## Links

https://codedotorg.atlassian.net/browse/LABS-675

## Testing story

Tested with standalone projects, with and without restricted packs, and on the onboarding flow.

### DIFF
--- a/apps/src/music/player/MusicLibrary.ts
+++ b/apps/src/music/player/MusicLibrary.ts
@@ -214,7 +214,7 @@ export default class MusicLibrary {
       return this.bpm;
     }
     const folder = this.getFolderForFolderId(this.currentPackId);
-    // Read BPM from the folder, or the first sound that has a BPM if not present.
+    // Read BPM from the folder, or the first sound that has a BPM if not present on the folder.
     return (
       folder?.bpm || folder?.sounds.find(sound => sound.bpm !== undefined)?.bpm
     );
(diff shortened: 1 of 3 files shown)


apps/src/music/player/MusicLibrary.ts
```diff
@@ -208,12 +208,28 @@ export default class MusicLibrary {
...
   }
 
+  // Returns the library BPM, or the BPM of the currently selected pack if present.
   getBPM(): number | undefined {
-    return this.bpm;
+    if (!this.currentPackId) {
+      return this.bpm;
+    }
+    const folder = this.getFolderForFolderId(this.currentPackId);
+    // Read BPM from the folder, or the first sound that has a BPM if not present.
```
breville: super nit: "if not present" refers to "the folder"?

breville:
> We'll read the bpm/key from the folder directly, but if it's not specified, we'll use the bpm/key from the first sound in the pack that has the relevant information.

Do we need bpm/key in individual sounds for any other reason?  Curious whether it would be sufficient to just use the bpm/key from the folder, and to otherwise use the default.  This could keep the library format simple.

sanchitmalhotra126:
> Do we need bpm/key in individual sounds for any other reason? Curious whether it would be sufficient to just use the bpm/key from the folder, and to otherwise use the default. This could keep the library format simple.

Currently we do read bpm/key for each individual sound when queueing samples, to determine what pitch/bpm offset to apply for each sound. Though, we could definitely instead just read the bpm/key from the sound's folder instead if that's easier to author. I guess the only question is, would we ever have a sound pack with a mix of keys/bpms?

PR #57916, opened 2024-04-09
Gen AI: add example topics and prompts

bencodeorg:
## Warning!!

## Description

Adds support for the "Example Prompts and Topics" list that can already be customized in levelbuilder. Reuses the component built on the levelbuilder side for adding/removing example topics and prompts.

| Edit View | User View |
| - | - |
| image | image |

## Testing story

Tested manually that a levelbuilder configured change appeared in the level, and that I could add/remove entries and see them on page reload. I also checked that blank entries disappeared on save.

### DIFF
--- a/apps/src/aichat/redux/aichatRedux.ts
+++ b/apps/src/aichat/redux/aichatRedux.ts
@@ -93,17 +93,37 @@ export const updateAiCustomization = createAsyncThunk(
     const {currentAiCustomizations, previouslySavedAiCustomizations} =
       state.aichat;
 
+    // Remove any empty example topics on save
+    const trimmedExampleTopics =
+      currentAiCustomizations.modelCardInfo.exampleTopics.filter(
+        topic => topic.length
+      );
+    thunkAPI.dispatch(
+      setModelCardProperty({
+        property: 'exampleTopics',
+        value: trimmedExampleTopics,
+      })
+    );
+
+    const trimmedCurrentAiCustomizations = {
+      ...currentAiCustomizations,
+      modelCardInfo: {
+        ...currentAiCustomizations.modelCardInfo,
+        exampleTopics: trimmedExampleTopics,
+      },
+    };
+
     await Lab2Registry.getInstance()
       .getProjectManager()
-      ?.save({source: JSON.stringify(currentAiCustomizations)}, true);
+      ?.save({source: JSON.stringify(trimmedCurrentAiCustomizations)}, true);
 
     thunkAPI.dispatch(
-      setPreviouslySavedAiCustomizations(currentAiCustomizations)
+      setPreviouslySavedAiCustomizations(trimmedCurrentAiCustomizations)
     );
 
     const changedProperties = findChangedProperties(
       previouslySavedAiCustomizations,
-      currentAiCustomizations
+      trimmedCurrentAiCustomizations
     );
     changedProperties.forEach(property => {
       thunkAPI.dispatch(
--- a/apps/src/aichat/views/modelCustomization/PublishNotes.tsx
+++ b/apps/src/aichat/views/modelCustomization/PublishNotes.tsx
@@ -1,5 +1,4 @@
 import React, {useCallback} from 'react';
-import {AnyAction, ThunkDispatch} from '@reduxjs/toolkit';
 
 import {
   setModelCardProperty,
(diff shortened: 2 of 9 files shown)


apps/src/templates/MultiItemInput.tsx
```diff
@@ -31,28 +41,31 @@ export const MultiItemInput: React.FunctionComponent<{
...
               )}
               onChange={e => onChange(index, e.target.value)}
+              disabled={readOnly}
             />
           );
         })}
       </div>
-      <div className={moduleStyles.buttonsRow}>
-        {showPlus && (
+      {!readOnly && (
```
bencodeorg, line 50: This is a weird looking diff, but I put it thru a diffchecker and the only difference is adding the readonly wrapper boolean :)

apps/src/lab2/levelEditors/aichatSettings/ModelCardFields.tsx
```diff
@@ -13,7 +15,7 @@ const ModelCardFields: React.FunctionComponent = () => {
   const exampleTopics = modelCardInfo.exampleTopics;
   return (
     <div className={moduleStyles['model-card-fields']}>
-      {MODEL_CARD_FIELDS_LABELS_ICONS.map(([property, label]) => {
+      {MODEL_CARD_FIELDS_LABELS_ICONS.map(([property, label, _]) => {
```
bencodeorg, line 18: My editor was giving me big red error labels since we didn't include this third element when destructuring this array, so I added this :)

apps/src/aichat/redux/aichatRedux.ts
```diff
@@ -93,17 +93,37 @@ export const updateAiCustomization = createAsyncThunk(
     const {currentAiCustomizations, previouslySavedAiCustomizations} =
       state.aichat;
 
+    // Remove any empty example topics on save
```
bencodeorg, line 96: Not sure if there's a better place to handle this -- basically wanted to cover the case where you hit plus a bunch of times to generate new entries in the example topics list. This approach removes those blank entries when you save.

apps/src/aichat/views/modelCustomization/PublishNotes.tsx
```diff
@@ -35,26 +37,35 @@ const PublishNotes: React.FunctionComponent = () => {
...
-                  onChange={event =>
-                    dispatch(
-                      setModelCardProperty({
-                        property: property,
-                        value: event.target.value,
-                      })
-                    )
-                  }
-                />
+                {property === 'exampleTopics' && (
```
bencodeorg, line 48: We've got a lot of these key strings floating around, I think we should figure out a way to pull them from a constant.

apps/src/templates/MultiItemInput.tsx
```diff
@@ -31,28 +41,31 @@ export const MultiItemInput: React.FunctionComponent<{
...
+        <div className={moduleStyles.buttonsRow}>
+          {showPlus && (
+            <button
```
sanchitmalhotra126: What do you think about using design system buttons for these now? When I was doing the levelbuilder edit page I kept the UI pretty minimal but now that this is student facing, might be worth tweaking the presentation on these a bit?

apps/src/aichat/redux/aichatRedux.ts
sanchitmalhotra126: I'd be curious to ask @dancodedotorg / @samantha-code - would students be confused if they had blank entries that suddenly disappeared upon clicking "Update"?

apps/src/aichat/views/modelCustomization/PublishNotes.tsx
sanchitmalhotra126: Yep that could be good, though we do already get automatic checking so long as `property` is typed correctly (which I think here is `keyof ModelCardInfo`). If the key is misspelled/not a key of `ModelCardInfo`, then TS will complain, effectively enforcing that we only compare against strings that are actually keys of `ModelCardInfo` (ex: `"This comparison appears to be unintentional because the types 'keyof ModelCardInfo' and '"hello"' have no overlap.ts(2367)"`)

apps/src/aichat/views/modelCustomization/PublishNotes.tsx
```diff
@@ -35,26 +37,35 @@ const PublishNotes: React.FunctionComponent = () => {
...
+                  <ExampleTopicsInputs
+                    topics={modelCardInfo.exampleTopics}
+                    readOnly={isDisabled(visibility)}
+                  />
+                )}
+                {property !== 'exampleTopics' && (
```
sanchitmalhotra126, line 54: minor: instead of checking if property is `exampleTopics` twice, you could just return null for exampleTopics in this `map` function, and stick an `ExampleTopicsInput` at the end? That's similar to what we do in `ModelCardFields`

apps/src/aichat/views/modelCustomization/PublishNotes.tsx
bencodeorg: Oh nice, good to know!

apps/src/aichat/views/modelCustomization/PublishNotes.tsx
bencodeorg: hmm that sounds reasonable...I just took a stab at this, problem is that the header text (`label`) comes from the constant we're mapping over. lmk if I'm missing something

apps/src/templates/MultiItemInput.tsx
bencodeorg: I actually don't see any mocks for this at all :) Updated to use the new `Button` component, and used the "secondary" styling that we use for the "Add" button when adding retrieval items. I've generally been trying to follow the guidance Jessie shared on Monday at sprint planning to focus efforts on getting things working end to end (ie, avoiding sinking too much time into styling perfection just yet)

image

apps/src/aichat/views/modelCustomization/PublishNotes.tsx
sanchitmalhotra126: Oh yeah, my recommendation was to early return in this map function so you don't render the header text for example topics at all, and instead just add it after the map. something like 
```
<div className={styles.customizationContainer}>
  {MODEL_CARD_FIELDS_LABELS_ICONS.map(([property, label, _]) => {
    if (property === 'exampleTopics') {
      return null;
    }
    return ( 
      <div className={styles.inputContainer} key={property}>
        ...
      </div>);
  }
  <div className={styles.inputContainer} key={'exampleTopics'}>
    <label ... />
    <ExampleTopicsInput ... />
  </div>
</div>
```

bencodeorg: Yeah I gotcha, the label content (ie, `label`) for example topics comes from MODEL_CARD_FIELDS_LABELS_ICONS though, so I'd have to hard code that as well, which I viewed as a vote for leaving as-is

sanchitmalhotra126: Oh fair, yep that's fine then!

sanchitmalhotra126 (approved):

PR #50709, opened 2023-03-10
Set `EnforcedStyle` to `inline` for `Style/AccessModifierDeclarations`

Hamms:
Alternative to https://github.com/code-dot-org/code-dot-org/pull/50457; rather than defaulting to `group` style, explicitly set `inline` style.

## Links

- https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/AccessModifierDeclarations
- https://docs.rubocop.org/rubocop/cops_style.html#styleaccessmodifierdeclarations

## PR Checklist:

- [ ] Tests provide adequate coverage
- [ ] Privacy and Security impacts have been assessed
- [ ] Code is well-commented
- [ ] New features are translatable or updates will not break translations
- [ ] Relevant documentation has been added or updated
- [ ] User impact is well-understood and desirable
- [ ] Pull Request is labeled appropriately
- [ ] Follow-up work items (including potential tech debt) are tracked and linked

PR #57918, opened 2024-04-10
Custom Rubocop Rule for Grouped Inline Private Methods

Hamms:
Follow-up to https://github.com/code-dot-org/code-dot-org/pull/57621, which enforced the use of inline private methods. Also see https://github.com/code-dot-org/code-dot-org/pull/50709 for more discussion on the topic.

Although it is nice to see explicit declarations for each method and to eliminate the possibility of reading code without realizing that the methods you're looking at are actually private because of an offscreen access modifier declaration, it is unfortunate that with the change we also lost enforcement for grouping all private methods together. Separation of interface methods from implementation details is a valuable tool for maintaining Ruby code, and by enforcing a strict grouping we can encourage developers to extract logically-related components into explicit modules or concerns, rather than relying on manual method ordering to attempt to make oversized class definitions more readable.

To that end, this PR proposes a compromise solution: continue to use the `inline` style for the `AccessModifierDeclarations` rule, but also add a rule of our own that forces all such inlined methods to be declared at the end of their respective code block. This is a more opinionated rule than we had previously enforced, and may or may not be more intrusive than we'd like. To help us evaluate the effects of this rule, this PR contains two kinds of changes: first, [the new Rubocop rule itself](https://github.com/code-dot-org/code-dot-org/pull/57918/files#diff-e9c78e03a0f249403808a1bd918e4a28e6c7486e63ac0ca6e2fd949f4131e7ce), which is actually fairly straightforward. The second and much larger batch of changes represents fixes for all current violations of the rule in the codebase. Please take a careful look at the kind of changes this rule will expect us to make.

I'm very interested to hear what folks think about this approach; **do the downsides outweigh the upsides?** What are the specific upsides and downsides that stand out most to you? Please discuss, and feel free to propose changes or to push back entirely on this direction.

## Links

- [Implementation of the related `AccessModifierDeclarations` rule](https://github.com/rubocop/rubocop/blob/v1.52.1/lib/rubocop/cop/style/access_modifier_declarations.rb)


dashboard/app/controllers/api/v1/pd/applications_controller.rb
```diff
@@ -13,6 +13,14 @@ class ApplicationsController < ::ApplicationController
     REGIONAL_PARTNERS_ALL = "all"
     REGIONAL_PARTNERS_NONE = "none"
 
+    TYPES_BY_ROLE = {
+      csd_teachers: TEACHER_APPLICATION_CLASS,
+      csp_teachers: TEACHER_APPLICATION_CLASS,
+      csa_teachers: TEACHER_APPLICATION_CLASS
+    }
+    ROLES = TYPES_BY_ROLE.keys
+    TEACHER_ROLES = ROLES.select {|role| role.to_s.include?('teachers')}
```
Hamms, line 22: This new rule does reveal a few places in the codebase where we are declaring constants or in some cases invoking setup methods like `prepend_before_action` down toward the bottom of the file, in amongst the private methods. I was not intending to catch such cases with this rule (going in, I was only thinking about the relative ordering of method definitions), but after seeing it in practice I think I like it.

In general, I think the best place for constants to be declared and for modifier methods to be invoked is right at the top of the file. This may lead to some situations in which some constants are far removed from the methods in which they are relevant, but the Ruby way to deal with that is to extract functionality into modules and/or concerns whenever possible, rather than to rely on manual ordering within a single large class definition.

artem-vavilov (approved):
Great job! 🙌🏻

cat5inthecradle (approved):

davidsbailey:
Thank you for soliciting feedback. I would vote against this one personally. The first post I found on this topic happens to articulate my preferred way of organizing methods within a class, which is to keep helper methods closer to the place they are called from: https://softwareengineering.stackexchange.com/questions/199311/most-human-friendly-way-to-order-class-method-definitions/199317#199317

When a class has too many methods, I'm not sure one organization method will keep it more readable than the other, and my preferred solution is to reduce the file size by extracting part of it into a new file.

I'm happy to be outvoted though! Hopefully with one vote for and against, others are more comfortable sharing their opinions 🙂

Hamms:
I would argue that the need to maintain a particular ordering of methods within a definition in order for code to be readable is a code smell. The right way to ensure related code is grouped together is to extract the related code into a single entity. In the specific case of Ruby, that entity doesn't even have to be another file; it can be a simple named module within the file which is included by the relevant class.

My hope is that this rule will help us to maintain the best of both worlds. By organizing public and private methods separately we enforce the distinction between implementation details and interfaces; we can still manually organize our code to maximize readability, but now we are encouraged to do so in "the Ruby way", by breaking our oversized classes up into more manageably-sized chunks.

dashboard/app/jobs/concerns/ai_rubric_config.rb
```diff
@@ -139,11 +137,14 @@ def self.validate_learning_goals_for_rubric(rubric)
...
 
   def self.get_s3_learning_goals(lesson_s3_name)
     rubric_csv = read_file_from_s3(lesson_s3_name, 'standard_rubric.csv')
     rubric_rows = CSV.parse(rubric_csv, headers: true).map(&:to_h)
     rubric_rows.map {|row| row['Key Concept']}
   end
+
+  private_class_method :validate_ai_config_for_lesson
+  private_class_method :validate_learning_goals
+  private_class_method :validate_learning_goals_for_rubric
```
davidsbailey: it seems very strange to separate the public/private status from the method definition. perhaps these should be inline instead? that's probably easier than updating the new lint rule to recognize this situation.
```
private_class_method def self.get_s3_learning_goals(lesson_s3_name)
```

davidsbailey (approved):
Thanks for the update!

Hamms:
Thank you for the thorough consideration!

PR #57927, opened 2024-04-10
Fix uncaught nil error during LTI auth

carl-codeorg:
During login, users have hit a `NoMethodError: undefined method '[]' for nil:NilClass` error trying to authenticate via LTI. This handles a nil value for the cached state and nonce with a 401 instead of throwing an unhandled error resulting in a 500/"sad bee".

## Links

[JIRA](https://codedotorg.atlassian.net/browse/P20-866)

## PR Checklist:

- [ ] Code is well-commented


wilkie (approved):
Looks right to me!

PR #57928, opened 2024-04-10
[JIMT]CT-492] editor-overwrites - fixes editor from overwriting itself

thomasoniii:
If you changed levels, your editor would overwrite the contents of the proper file with the default values provided in Weblab2View.tsx.

There were two issues going on -
- CDOIDE was setting the `shouldNotifyProjectUpdate` flag too late in the process, which would noisily cause an update of the last project value _before_ `replaceProject` was called for the new one. This would end de-syncing and overwriting with the prior value of the project. Fix here was to swap the effect order and bubble up where the `shouldNotifyProjectUpdate` flag was set.
- `saveProject` would fire a save even if the project itself was the same. This could cause de-sync issues when switching rapidly between levels and could end up overwriting or even swapping content. Fix here was to only save if the value actually changed.

## Links

- jira ticket: [CT-492](https://codedotorg.atlassian.net/browse/CT-492)


apps/src/weblab2/CDOIDE/CDOIDE.tsx
```diff
@@ -55,6 +55,19 @@ export const CDOIDE = React.memo(
...
+    // to store whether we should fire off the next update. Internally, the reducer will set the flag
+    // to true if we're ACTUALLY replacing the project, which will then make it all work and sync up.
+    //
+    // Yes, this is a little crazy. Yes, I think this should be refactored out into a wrapper that
+    // explicitly calls setProject only on appropriate actions. Yes, I'll look into it.
+    useEffect(() => {
+      shouldNotifyProjectUpdate.current = false;
+      projectUtilities.replaceProject(project);
+    }, [project, projectUtilities]);
+
```
thomasoniii, line 70: Yes, I _know_ that effects shouldn't depend upon each other, but because we're using a ref here it kinda has to until this is swapped out with something a little less clunky.

apps/src/weblab2/Weblab2View.tsx
```diff
@@ -83,7 +83,7 @@ const defaultProject: ProjectType = {
   <link rel="stylesheet" href="styles.css"/>
   <body>
     Content goes here!
-    <div class="foo">Foo class!</div>
+    <div class="foo">[DEFAULT] Foo class!</div>
```
thomasoniii, line 86: Just to hopefully help us see when we're loading in default data. I'd encourage everyone to change your data when you navigate in and save, so that if this ever shows up it's more clear.

apps/src/weblab2/Weblab2View.tsx
```diff
@@ -148,7 +148,10 @@ const Weblab2View = () => {
 
   const setProject = (newProject: MultiFileSource) => {
     setCurrentProject(newProject);
-    if (Lab2Registry.getInstance().getProjectManager()) {
+    if (
+      Lab2Registry.getInstance().getProjectManager() &&
+      newProject !== currentProject
```
thomasoniii: if we blindly save even if there's no change, it can overwrite if you navigate too fast. 

This is kinda a hack solution - it'd be better for CDOIDE not to fire off the additional noisy note, and it'd also be better if the project manager's save could somehow take into account if nothing's changing (but that's probably asking too much since it doesn't maintain state otherwise).

molly-moen: The project manager actually does check if nothing has changed, and does not save in that case ([link](https://github.com/code-dot-org/code-dot-org/blob/457197fcbc2d64dc1bb2c3973ef89d7dc0fd5ed1/apps/src/lab2/projects/ProjectManager.ts#L287)). Is the problem here that the channel id has changed but we are sending a late update with the code from the previous channel id?

molly-moen: Why does this get called if you navigate between levels? Shouldn't we only call `setProject` when the project has been changed, not when it is initially loaded?

thomasoniii: The problem was that the `setProject` function wasn't memoized.

In turn, when you navigated it'd create a new `setProject` function which'd fall into the IDE and trigger the effect to claim that the `project` had changed, even though it hadn't in this case. With a memoized function, it behaves as expected.

Further, though, this is a bigger issue beyond this code - because the project manager's `save` method only takes _data_ to save and not the channelID to save it to (I think I'm using the right terms), it's susceptible to this type of issue -

basically, if the user navigates old data from the prior level may still be loaded into components. If something is triggered which would cause a save, it might push forward the stale data to the project manager, which would then overwrite the _new_ channelID with the prior channel's data. It's assuming that all data that currently exists on the page is new, but if it hasn't been refreshed yet it's stale, and if that stale data triggers a save, it'd overwrite.

As it did in this case since the function wasn't memoized. This should be flagged as a future fix to circle back to.

molly-moen: I made a follow up task to look into the caching idea I came up with: https://codedotorg.atlassian.net/browse/CT-500

molly-moen (approved):

PR #57931, opened 2024-04-10
Add amplitude event for moving rubric window

Nokondi:
Added Amplitude analytics event for reporting when the rubric window is moved. Reports movement start and end positions in two separate events.

Jira Ticket: [AITT-554](https://codedotorg.atlassian.net/browse/AITT-554?atlOrigin=eyJpIjoiYjFmMmI5MmFmMDA0NGIzNWIyMTIyMDlkYTdkMTM4NWQiLCJwIjoiaiJ9)

### DIFF
--- a/apps/test/unit/templates/rubrics/RubricContainerTest.jsx
+++ b/apps/test/unit/templates/rubrics/RubricContainerTest.jsx
@@ -771,7 +771,7 @@ describe('RubricContainer', () => {
     stubFetchTeacherEvaluations(noEvals);
     stubFetchTourStatus({seen: true});
 
-    const {queryByText, getByTestId} = render(
+    const {getByTestId} = render(
       <Provider store={store}>
         <RubricContainer
           rubric={defaultRubric}
--- a/apps/test/unit/templates/rubrics/RubricContainerTest.jsx
+++ b/apps/test/unit/templates/rubrics/RubricContainerTest.jsx
@@ -762,4 +762,47 @@ describe('RubricContainer', () => {
     expect(queryByText('Getting Started with AI Teaching Assistant')).to.not
       .exist;
   });
+
+  it('sends event when window is dragged', async function () {
+    const sendEventSpy = sinon.spy(analyticsReporter, 'sendEvent');
+    stubFetchEvalStatusForUser(readyJson);
+    stubFetchEvalStatusForAll(readyJsonAll);
+    stubFetchAiEvaluations(mockAiEvaluations);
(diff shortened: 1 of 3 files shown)


apps/test/unit/templates/rubrics/RubricContainerTest.jsx
```diff
@@ -762,4 +762,47 @@ describe('RubricContainer', () => {
...
+    expect(sendEventSpy).to.have.been.calledWith(
+      EVENTS.TA_RUBRIC_WINDOW_MOVE_START,
+      {window_x_start: 0, window_y_start: 0}
+    );
+
+    fireEvent.mouseUp(element);
+
+    expect(sendEventSpy).to.have.been.calledWith(
+      EVENTS.TA_RUBRIC_WINDOW_MOVE_END,
+      {window_x_end: 0, window_y_end: 0}
```
davidsbailey, line 804: this is confusing... shouldn't these be 100/100? are you able to validate that the right values come through in manual testing?

Nokondi: Yes, it works locally. This is something I read about with the way that react testing library renders components. I had a similar issue when writing the unit test for moving the window (i.e., it didn't return the correct values, but it did register that the values had changed). I'll see if I can find a link to what I read before.

Nokondi: Sorry, meant to say that I've tested manually and gotten the correct results:
image

davidsbailey: that would be great. I'm not sure how common this will be, but until it becomes common knowledge I would suggest including a short explanation with a link to the issue as a code comment

davidsbailey (approved):
lgtm after considering test code comments

apps/test/unit/templates/rubrics/RubricContainerTest.jsx
Nokondi: Here's the explanation that I pieced together: React Testing Library uses JSDom as the basis for its render function. JSDom doesn't actually _render_ elements, so everything will show up with a position of (0, 0). For the draggable unit test, I was able to get around this because the draggable component applies a transformation to the component, so if you look at the translation, you'll see the change, but the raw x and y position values will always show as (0, 0). It's _possible_ you can get more accurate position readings by using RTL with Jest, but I haven't confirmed that.
image
https://github.com/jsdom/jsdom/issues/1590

PR #57940, opened 2024-04-10
feat(jest): migrate unit tests from karma to jest

stephenliang:
# Migration of unit tests from karma to jest

This pull request migrates the Code.org apps unit tests from karma to [jest](https://jestjs.io/). Karma was [deprecated](https://github.com/karma-runner/karma?tab=readme-ov-file#karma-is-deprecated-and-is-not-accepting-new-features-or-general-bug-fixes) in April 2023. 

In the years since Karma was introduced, the [Javascript test ecosystem](https://docs.google.com/document/d/1-U41r5O8Y404HeWHOAn2vxUI6G1lSJbtK7xi4j2eZsA/edit#heading=h.apat1dfk2ipw) has matured introducing new testing frameworks such as jest, vitest, mocha, and jasmine. For react based applications, the industry has consolidated toward the using of jest for unit test execution. Due to the industry shift to using other test frameworks, Karma has since become [deprecated](https://github.com/karma-runner/karma?tab=readme-ov-file#karma-is-deprecated-and-is-not-accepting-new-features-or-general-bug-fixes) in April 2023. 

Additionally, Karma has historically caused a number of pain points to engineers developing the Code.org frontend:

* **Unit test feedback time:** When an engineer is developing a feature, an engineer will typically run unit tests multiple times. However, Karma requires webpack (compiling) each time a unit test executes regardless of the size of the test suite resulting in a cycle time of 1-2 minutes per test execution. If an engineer runs unit tests 10-20 times per feature, this can easily add 30 or more minutes of wait time per feature.
* **Client State pollution**: Since Karma executes within a browser context, if a unit test manipulates the browser state (DOM or window object), subsequent tests will inherit the same manipulated state. This results in differing behavior where a test passes if run alone but fails if run in a group (or vice versa). The student learning team recently encountered such an error resulting in a week of time spent debugging this test. Additionally, react testing library has had issues with DOM pollution requiring special cleanup code.
* **Dependency on Browsers**: Typically, unit tests are meant to test the behavior of code with as many mocked out dependencies as possible. This includes interaction with browsers, which may differ between which browser is actually used. By relying on an actual browser, the unit test could fail if the browser changes or updates. Additionally, the CI/CD infrastructure needs to install and be able to execute the given browser. 
* **Karma Deprecation:** Karma has been [deprecated](https://github.com/karma-runner/karma?tab=readme-ov-file#karma-is-deprecated-and-is-not-accepting-new-features-or-general-bug-fixes) and will soon stop providing critical security fixes.

Additional details and context are available on the [internal jest migration document](https://docs.google.com/document/d/1-U41r5O8Y404HeWHOAn2vxUI6G1lSJbtK7xi4j2eZsA).

## Performance Improvement

|                                     | jest | karma | % Improvement |
|-------------------------------------|------|-------|---------------|
| Single Test (Drone)                 | 6s   | 75s   | 92%           |
| Single Test (Stephen's workstation) | 3.7s | 28s   | 86%           |
| Full Suite (Drone)                  | 228s | 252s  | 9.5%          |
| Full Suite (Stephen's workstation)  | 42s  | 86s   | 51%           |

## Demonstration of Karma Single Test Run

Karma takes 75 seconds to execute a single test on an adhoc

[karma.webm](https://github.com/code-dot-org/code-dot-org/assets/538214/dc134ae2-a84c-4a0c-8b50-4ffb7d8fc3d5)

## Demonstration of Jest Single Test Run

Jest takes 6 seconds to execute a single test on an adhoc

[jest.webm](https://github.com/code-dot-org/code-dot-org/assets/538214/5e992448-989b-4663-9247-362e5e6db8a0)

## Links

- [jest migration document](https://docs.google.com/document/d/1-U41r5O8Y404HeWHOAn2vxUI6G1lSJbtK7xi4j2eZsA)

## Testing story

1. Full regression test of all unit tests
2. Full regression test of selenium tests
3. Spot checking all student labs

## Deployment strategy

This PR is to be merged after the AP CSP peak period. This PR will be deployed in multiple phases:

1. A notification to be sent out to Code.org engineering of jest migration
2. Training session on jest to be presented to Code.org engineering
3. This PR to be merged on a Friday evening to minimize impact to engineers
4. E-mail notification to be sent for engineers to rebase and start using jest the following Monday
5. Monitor engineering channels to ensure the successful deployment of jest
6. Monitor deployment processes to ensure no interference from jest migration

## Follow-up work

Out of scope is the migration of karma for integration tests.

## Security

This PR will remove the deprecated karma dependency and its vulnerable transitives.

## PR Checklist:

- [x] Code is well-commented


apps/src/templates/instructions/codeReviewV2/Comment.jsx
```diff
@@ -25,7 +25,7 @@ function Comment({
   currentUserId,
   viewingAsOwner,
 }) {
-  const isMounted = useRef(false);
+  const isMounted = React.useRef(false);
```
davidsbailey, line 28: why is this change necessary?

stephenliang: The following test was failing:

```
displays hide option for visible resolved comment

Expected value   1
Received:
  0

Message:
  expected {} to have a length of 1 but got 0
    at lengthOf (/home/sliang/git-workspaces/code-dot-org/apps/test/unit/templates/instructions/codeReviewV2/CommentTest.jsx:78:53)
    at processTicksAndRejections (node:internal/process/task_queues:95:5)
```

The reason was that Comment.jsx line 136 requires `isMounted` to be set which in Karma was done via the browser. In enzyme/jsdom `useRef` doesn't ever get set and so I needed to mock out `useRef` to return true which was done in the test:

```
jest.spyOn(React, 'useRef').mockReturnValue({current: true});
```

In order for this mock to use, I needed to use the `React.useRef` version to be able to spy on this function. The exported named function did not work with the spy so this change was made in the code.

If this test was ported to RTL this won't need to be done.

davidsbailey: Makes sense, thanks for explaining!

apps/src/applab/applab.js
```diff
@@ -692,7 +692,7 @@ Applab.init = function (config) {
     isCurriculumLevel: isCurriculumLevel(config.level.validationEnabled),
   });
 
-  config.dropletConfig = dropletConfig;
+  config.dropletConfig = {...dropletConfig};
```
wilkie, line 695: Why do this? Does it need to be copied in order to support being tested?

apps/src/templates/instructions/codeReviewV2/Comment.jsx
wilkie: Would the conventional using `* as blah` work to spy on exported functions like this in Jest?

```
// CommentTest.jsx
import * as react from 'react'

...

Jest.spyOn(react, 'useRef').mockReturnValue({current: true});
```

Just curious as this is the approach I would traditionally use.

apps/src/applab/applab.js
stephenliang: With the jest migration, it enabled certain modules to be imported as ESM instead of CommonJS. The Applab init code imports `dropletConfig` which is then passed to `Immutable.fromJS`. This function will return an Immutable for a CommonJS import, but for ESM it will return the same input which causes the subsequent call `mergeDeepWith` to fail (see below stack).

To workaround this while being compatible with CommonJS and ESM, I do a shallow clone by spreading the object which converts it from an ES Module to a plain old JS object as [recommended here](https://stackoverflow.com/a/48510588). Overall this change is useful to help us enable a later migration to native ESM in the future.

 Without this change, the unit test fails with:

```
  ● Applab › Applab.init() › configuration options › hasDataMode page constant › is false in widget mode

    TypeError: baseImmutable.mergeWith is not a function

      538 |
      539 |   var baseImmutable = Immutable.fromJS(baseObject);
    > 540 |   return baseImmutable.mergeWith(deepConcatMerger, overrides).toJS();
          |                        ^
      541 | }
      542 |
      543 | /**

      at Object.mergeWith [as deepMergeConcatArrays] (src/utils.js:540:24)
      at Object.deepMergeConcatArrays [as init] (src/applab/applab.js:738:34)
      at Object.init (test/unit/applab/applabTest.js:691:18)
```

apps/src/templates/instructions/codeReviewV2/Comment.jsx
stephenliang: You should be able to do the same style mocking in Jest.

PR #57944, opened 2024-04-10
Add backend controller and client for Gen AI

fisher-alice:
This PR adds the `aichat_controller` which will make requests to Sagemaker when the API is available and the corresponding client file `aichatCompletionApi` which makes a call to the controller. This will be used when a user submits a chat message in the Aichat workspace.

The controller receives:
- messages including the system prompt, retrieval contexts, past user messages and assistant responses
- the temperature from `savedAiCustomizations`
- `chatContext` that includes the newest user message text, `userId`, `currentLevelId`, `scriptId`, and `channelId`.

The controller sends back a dummy assistant response for now. Screenshot from dev console after adding print statements in `aichatCompletionApi.ts` and `aichatRedux` of `response`.

Screenshot 2024-04-11 at 3 27 13 PM

Response displayed in workspace:
Screenshot 2024-04-11 at 4 00 38 PM

## Links
[jira](https://codedotorg.atlassian.net/browse/LABS-615)

## Testing story
Tested locally using /allthethings levels.

### Screenshots from Network Tab:
Screenshot 2024-04-11 at 3 22 53 PM

Screenshot 2024-04-11 at 3 23 21 PM

Screenshot 2024-04-11 at 3 23 28 PM

## PR Checklist:

- [ ] Code is well-commented


apps/src/aichat/views/presentation/ModelCardRow.tsx
```diff
@@ -27,8 +27,8 @@ const ModelCardRow: React.FunctionComponent<ModelCardRowProps> = ({
       }
       return (
         <ul>
-          {checkedExpandedContent.map(content => (
-            <li>{content}</li>
+          {checkedExpandedContent.map((content, index) => (
+            <li key={index}>{content}</li>
```
fisher-alice, line 31: Unrelated update - adding key here and in `PresentationView`.

apps/src/aichat/views/chatWorkspace.module.scss
```diff
@@ -12,6 +12,7 @@ $default-spacing: 2px;
 
 .conversationArea {
   margin: 5px 2%;
+  height: 65vh;
```
fisher-alice: Added so that chat workspace is scrollable.

dashboard/app/controllers/aichat_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+    locale = params[:locale] || "en"
+    # Check only the newest message from the user for inappropriate content.
+    chatContext = params[:chatContext]
+    newMessageText = chatContext["userMessage"]   
+    filter_result = ShareFiltering.find_failure(newMessageText, locale) if newMessageText
+    # If the content is inappropriate, we skip sending to endpoint and instead hardcode a warning response on the front-end.   
+    return render(status: :ok, json: {status: filter_result.type, flagged_content: filter_result.content}) if filter_result
+    inputs = params[:inputs]
+    temperature = params[:temperature]
+    # TODO: Format input to send to Sagemaker.
```
fisher-alice, line 19: Still waiting clarification on how to format. Using [this ref](https://aws.amazon.com/blogs/machine-learning/best-prompting-practices-for-using-the-llama-2-chat-llm-through-amazon-sagemaker-jumpstart/) as a reference for now.

apps/src/aichat/views/presentation/ModelCardRow.tsx
fisher-alice: Unrelated update to add key to list items here and below.

apps/src/aichat/types.ts
```diff
@@ -1,22 +1,35 @@
...
+  status: AichatInteractionStatus;
   timestamp?: string;
 };
 
+export type ChatContext = {
+  userMessage: string;
+  userId: number;
+  currentLevelId: string | null;
+  scriptId: number | null;
+  channelId: string | undefined;
```
fisher-alice, line 18: Not sure how to access a user's `sectionId`. Maybe access on backend?

dashboard/app/controllers/aichat_controller.rb
```diff
@@ -0,0 +1,34 @@
...
+    response = request_chat_completion(payload)
+    render(status: response[:status], json: response[:json])
+  end
+
+  def has_required_params?
+    params[:inputs].present? && params[:temperature].present? && params[:chatContext].present?
+  end
+
+  def request_chat_completion(payload)
+    response_body = {role: "assistant", content: "This is an assistant response from Sagemaker"}
```
fisher-alice, line 28: When I print the `payload`, I see the following:
Screenshot 2024-04-11 at 3 30 26 PM

Not sure if `#<ActionController::Parameters` is a problem when assigned from `params`. If it is, then would love input on how to avoid.

dashboard/app/controllers/aichat_controller.rb
```diff
@@ -0,0 +1,34 @@
+class AichatController < ApplicationController
+  # POST /aichat/chat_completion
```
bencodeorg, line 7: ~~Did we decide on how we're going to do authorization for this action? eg, you need to be in a section of a pilot teacher?~~ chatted offline, this will be a follow-up

dashboard/app/controllers/aichat_controller.rb
```diff
@@ -0,0 +1,34 @@
+class AichatController < ApplicationController
```
bencodeorg, line 1: Thoughts on using the name `AichatController` instead of `SagemakerController` or something, since there's no associated aichat rails model or anything? I remember the old one was `OpenaiChatController`, yeah?

apps/src/aichat/views/presentation/PresentationView.tsx
```diff
@@ -59,6 +59,7 @@ const PresentationView: React.FunctionComponent = () => {
               title={label}
               titleIcon={iconName}
               expandedContent={modelCardInfo[property]}
+              key={property}
```
bencodeorg: Was this because of a warning in the console (I think I saw this as well)? Do you need `keyName`?

dashboard/app/controllers/aichat_controller.rb
```diff
@@ -0,0 +1,34 @@
...
+  # POST /aichat/chat_completion
+  def chat_completion
```
bencodeorg, line 8: Could you put an example expected shape of the params coming in? I'm finding myself trying to reverse engineer it from the code

dashboard/app/controllers/aichat_controller.rb
```diff
@@ -0,0 +1,34 @@
...
+    payload = {
+      inputs: [params[:inputs]],
+      parameters: {temperature: params[:temperature]},
+    }
+    response = request_chat_completion(payload)
+    render(status: response[:status], json: response[:json])
+  end
+
+  def has_required_params?
+    params[:inputs].present? && params[:temperature].present? && params[:chatContext].present?
```
bencodeorg: I think the rails-y way to do this is [params.require([:inputs, ...])](https://api.rubyonrails.org/classes/ActionController/Parameters.html#method-i-require). It might even generate the right response for you if it raises an error, not sure

dashboard/app/controllers/aichat_controller.rb
```diff
@@ -0,0 +1,34 @@
...
+    unless has_required_params?
+      return render(status: :bad_request, json: {})
+    end
+    # Check for PII / Profanity
+    # Copied from ai_tutor_interactions_controller.rb - not sure if filtering is working.
+    locale = params[:locale] || "en"
+    # Check only the newest message from the user for inappropriate content.
+    new_message_text = params[:chatContext]["userMessage"]
+    puts "new_message_text = #{new_message_text}"
+    filter_result = ShareFiltering.find_failure(new_message_text, locale) if new_message_text
```
bencodeorg, line 16: Is it not possible to test if this works?

dashboard/app/controllers/aichat_controller.rb
bencodeorg: FWIW I think aligning the controller with where it's used makes sense, but maybe eventually putting the stuff that is specific to Sagemaker somewhere else? [This Pardot helper](https://github.com/code-dot-org/code-dot-org/blob/staging/lib/cdo/contact_rollups/v2/pardot_helpers.rb) we use to aggregate information for marketing emails (and talk to a third party Pardot API) comes to mind...just thinking out loud, don't need to overoptimize for now.

dashboard/app/controllers/aichat_controller.rb
bencodeorg: I think this is fine, [per docs here](https://api.rubyonrails.org/classes/ActionController/Parameters.html):

> You can fetch values of ActionController::Parameters using either :key or "key".
> 
> params = ActionController::Parameters.new(key: "value")
> params[:key]  # => "value"
> params["key"] # => "value"

dashboard/app/controllers/aichat_controller.rb
bencodeorg, line 16: Do we want to return early or something if no `new_message_text`?

apps/src/aichat/types.ts
```diff
@@ -1,22 +1,35 @@
...
+
+export enum AichatInteractionStatus {
+  ERROR = 'error',
+  PII_VIOLATION = 'pii_violation',
+  PROFANITY_VIOLATION = 'profanity_violation',
+  OK = 'ok',
+  UNKNOWN = 'unknown',
+}
+
+export enum PiiTypes {
```
bencodeorg: Didn't Emily just move this to a shared constant? / can we remove it there if not being used?

apps/src/aichat/types.ts
bencodeorg: Yeah I think that will be fine

apps/src/aichat/views/UserChatMessageEditor.tsx
```diff
@@ -16,14 +19,39 @@ const UserChatMessageEditor: React.FunctionComponent = () => {
...
+
+  const channelId = getStandaloneProjectId();
+
   const dispatch = useAppDispatch();
+
   const handleSubmit = useCallback(() => {
     if (!isWaitingForChatResponse) {
-      dispatch(submitChatMessage(userMessage));
+      const chatContents: ChatContext = {
+        userMessage,
```
bencodeorg: Is it unholy to access Redux outside of React in our code? Could we just get all of this info from redux in the API?

apps/src/aichat/aichatCompletionApi.ts
```diff
@@ -0,0 +1,94 @@
...
+ * if successful.
+ * TODO: Awaiting details on how to format input for endpoint.
+ */
+export async function getAichatCompletionMessage(
+  aiCustomizations: AiCustomizations,
+  storedMessages: ChatCompletionMessage[],
+  chatContext: ChatContext
+) {
+  const {systemPrompt, temperature, retrievalContexts} = aiCustomizations;
+  const messagesToSend = [
```
bencodeorg: Do you think this structure might change depending on how Sagemaker wants things formatted, or will we just handle whatever reshaping we need to do in Rails?

apps/src/aichat/aichatCompletionApi.ts
```diff
@@ -0,0 +1,94 @@
...
+      .logError('Error in aichat completion request', error as Error);
+  }
+  return response;
+}
+
+const formatRetrievalContextsForAichatCompletion = (
+  retrievalContexts: string[]
+): AichatCompletionMessage[] => {
+  return retrievalContexts.map(context => {
+    return {role: Role.USER, content: context};
```
bencodeorg: Interesting -- total noob here, is this the way retrieval context is typically treated by a model (the same as a message from the user)?

apps/src/aichat/redux/aichatRedux.ts
```diff
@@ -140,66 +137,50 @@ export const updateAiCustomization = createAsyncThunk(
...
+    if (chatApiResponse?.role === Role.ASSISTANT) {
       const assistantChatMessage: ChatCompletionMessage = {
-        id: chatApiResponse.id + 1,
+        id: newMessageId + 1,
         role: Role.ASSISTANT,
         status: Status.OK,
-        chatMessageText: chatApiResponse.assistantResponse,
-        // The accuracy of this timestamp is debatable since it's not when our backend
-        // issued the message, but it's good enough for user testing.
-        timestamp: getCurrentTimestamp(),
```
bencodeorg, line 200: No more timestamp?

apps/src/aichat/redux/aichatRedux.ts
```diff
@@ -68,7 +67,7 @@ export interface AichatState {
   // Denotes if there is an error with the chat completion response
   chatMessageError: boolean;
   currentAiCustomizations: AiCustomizations;
-  previouslySavedAiCustomizations?: AiCustomizations;
+  savedAiCustomizations: AiCustomizations;
```
bencodeorg, line 69: Nice, thanks for doing this!

apps/src/aichat/redux/aichatRedux.ts
```diff
@@ -140,66 +137,50 @@ export const updateAiCustomization = createAsyncThunk(
...
-    thunkAPI.dispatch(
-      updateChatMessageStatus({
-        id: chatApiResponse.id,
-        status: chatApiResponse.status,
-      })
-    );
-
-    // Add assistant chat messages to chatMessages.
-    if (chatApiResponse.assistantResponse) {
+    console.log('chatApiResponse', chatApiResponse);
```
bencodeorg, line 176: Keep this log or no?

apps/src/aichat/redux/aichatRedux.ts
```diff
@@ -140,66 +137,50 @@ export const updateAiCustomization = createAsyncThunk(
...
-    const systemPrompt = (state.lab.levelProperties as AichatLevelProperties)
-      ?.systemPrompt;
-    // TODO: move a check for undefined systemPrompt to AIchatView and throw an error dialog
-    // there if systemPrompt is undefined.
-    if (systemPrompt === undefined) {
-      throw new Error('systemPrompt is undefined');
-    }
+    const aiCustomizations = state.aichat.savedAiCustomizations;
     const storedMessages = state.aichat.chatMessages;
+    const newMessageText = chatContext['userMessage'];
```
bencodeorg: Nit: is there a reason this variable is here? Just access directly as `chatMessageText: chatContext.userMessage`?

bencodeorg (commented):
Sorry, kind of got comment happy :) Mostly small things / me trying to understand this stuff better!

One overall q -- is there any value in maintaining the existing API that can actually talk to an AI (ie, openai) in the short term, rather than cutting over to this new version that isn't actually hooked up? I was wondering if Dan and co. might find that useful for taking screenshots/etc.

apps/src/aichat/aichatCompletionApi.ts
sanchitmalhotra126: The goal is to handle whatever reshaping we need to do in Rails. That being said @fisher-alice I think it would be good to send the system prompt and retrieval separately from the chat messages (I think it can be as simple as just sending the whole of `aiCustomizations` as-is). You're right in that system prompt and retrieval will actually just be prepended to the list of messages, but there will probably be more formatting required (like wrapping the prompt in `[INST]` tags) so it would be helpful to have them separated out

apps/src/aichat/aichatCompletionApi.ts
sanchitmalhotra126: Yep! for our purposes, retrieval is just enhancing the chat context/prompt. We're technically really just doing the "augmented generation" part of "retrieval augmented generation" because students are sourcing content directly (rather us "retrieving" from a large dataset like Wikipedia)

apps/src/aichat/views/UserChatMessageEditor.tsx
```diff
@@ -16,14 +19,39 @@ const UserChatMessageEditor: React.FunctionComponent = () => {
...
   );
 
-  // TODO: If systemPrompt is undefined, handle this error case.
+  const userId: number = useSelector(
+    (state: {currentUser: CurrentUserState}) => state.currentUser.userId
+  );
+
+  const {currentLevelId, scriptId} = useSelector(
+    (state: {progress: ProgressState}) => state.progress
+  );
```
sanchitmalhotra126: nit: use `useAppSelector` here so you don't have to manually cast the `state` type

apps/src/aichat/views/UserChatMessageEditor.tsx
```diff
@@ -16,14 +19,39 @@ const UserChatMessageEditor: React.FunctionComponent = () => {
...
+
+  const channelId = getStandaloneProjectId();
```
sanchitmalhotra126: This should come from lab2Redux (`state.lab.channel.id`). the `standaloneProjectId` only exists for standalone projects (/projects/aichat) and we only fetch it when first loading the page because we need to grab it out of app options.

apps/src/aichat/views/UserChatMessageEditor.tsx
sanchitmalhotra126: Yeah, so technically we shouldn't be accessing redux state outside of React/redux code. We do have a lot of instances of that in older code, but in newer code we're generally trying to avoid it to enforce the separation between react/redux and non-react code.

apps/src/aichat/views/UserChatMessageEditor.tsx
```diff
@@ -16,14 +19,39 @@ const UserChatMessageEditor: React.FunctionComponent = () => {
...
     if (!isWaitingForChatResponse) {
-      dispatch(submitChatMessage(userMessage));
+      const chatContents: ChatContext = {
+        userMessage,
+        userId,
+        currentLevelId,
+        scriptId,
+        channelId,
+      };
+      dispatch(submitChatContents(chatContents));
```
sanchitmalhotra126: Since you're using a thunk action to send this message to the API, it might be easier to grab all the other redux-driven values in the thunk directly (userId, currentLevelId, etc), since you have access to the entire state object. That way this `useCallback` also has way fewer dependencies

dashboard/app/controllers/aichat_controller.rb
sanchitmalhotra126: Yeah, my thought was it would be good to decouple the name of the controller with the specific endpoint it's hitting (in case we for some reason switch from SageMaker back to OpenAI or something else). I do think moving SageMaker-specific code to its own unit will be good though.

sanchitmalhotra126 (commented):
Nice, this is coming together well! A few comments but looks solid overall

apps/src/aichat/views/presentation/PresentationView.tsx
fisher-alice: Yes - added bc of a warning. Oh yeah - I'll remove `keyName` then.

dashboard/app/controllers/aichat_controller.rb
fisher-alice: Oh yeah! I see Erin B added tests for this and for no new user text message at https://github.com/code-dot-org/code-dot-org/blob/0fae0afe87176231355469c3d8d6cd7ee5eb68d4/dashboard/test/controllers/openai_chat_controller_test.rb#L51-L77. Will add in a follow-up.

dashboard/app/controllers/aichat_controller.rb
fisher-alice: Nice - thanks! Will do.

dashboard/app/controllers/aichat_controller.rb
fisher-alice: Sure - what about
```
 return render(status: :ok, json: {role: "assistant", content: "No new user message was received."}) if !new_message_text
```

apps/src/aichat/redux/aichatRedux.ts
fisher-alice: Going to keep it for now. Thanks!

fisher-alice:
> Sorry, kind of got comment happy :) Mostly small things / me trying to understand this stuff better!
> 
> One overall q -- is there any value in maintaining the existing API that can actually talk to an AI (ie, openai) in the short term, rather than cutting over to this new version that isn't actually hooked up? I was wondering if Dan and co. might find that useful for taking screenshots/etc.

Thanks Ben - this is all great feedback! Posted your overall question in [this Slack thread](https://codedotorg.slack.com/archives/C06FELVTV0Q/p1712952108657529) and the response from Dan and Sanchit is that we should go ahead and proceed with getting `aichat` connect to new endpoint. Thanks!

apps/src/aichat/types.ts
fisher-alice: yes, this is available as a shared constant, but from [this discussion](https://github.com/code-dot-org/code-dot-org/pull/57669#discussion_r1548594055) there seems to be typing issue when working with a shared constant. I can import from shared constants for now and see how it goes when we get to using this constant.

apps/src/aichat/views/presentation/PresentationView.tsx
fisher-alice: Ended up being part of separate PR https://github.com/code-dot-org/code-dot-org/pull/57996

apps/src/aichat/aichatCompletionApi.ts
fisher-alice: Gotcha - I wasn't sure if we wanted to format in the front end or back end but will send `aiCustomizations` as is to the backend.

apps/src/aichat/redux/aichatRedux.ts
fisher-alice: Oh yeah - I see it's added for other chat messages. I'll add back.

apps/src/aichat/views/UserChatMessageEditor.tsx
fisher-alice: Only for `aichat` state which is the `RootState` though - correct?

apps/src/aichat/views/UserChatMessageEditor.tsx
fisher-alice: Will assign with `useSelector` but okay to leave in this component?

apps/src/aichat/views/UserChatMessageEditor.tsx
sanchitmalhotra126: `RootState` is the type of the global application state (for whatever slices have been ported to TypeScript already), so it will include `progress` as well

fisher-alice: Based on comment below, moving to `aichatRedux`. Thanks!

apps/src/aichat/views/UserChatMessageEditor.tsx
fisher-alice: Yes - that's much simpler - thanks!

dashboard/app/controllers/aichat_controller.rb
fisher-alice: Discussed offline and sending message text separate so will be checked by `require` above.

apps/src/aichat/types.ts
```diff
@@ -54,6 +63,12 @@ export interface AiCustomizations {
   modelCardInfo: ModelCardInfo;
 }
 
+export interface AiCustomizationsForBackend {
```
fisher-alice: Adding bc there is no need to send `modelCardInfo` to backend.

apps/src/util/sharedVoices.js
```diff
@@ -0,0 +1,33 @@
+/* eslint-disable */
```
bencodeorg: Accidentally added I think

apps/src/aichat/aichatCompletionApi.ts
```diff
@@ -0,0 +1,89 @@
...
+} from './types';
+import HttpClient from '@cdo/apps/util/HttpClient';
+import Lab2Registry from '@cdo/apps/lab2/Lab2Registry';
+
+const CHAT_COMPLETION_URL = '/aichat/chat_completion';
+
+/**
+ * This function sends a POST request to the aichat completion backend controller.
+ */
+export async function postAichatCompletionMessage(
```
bencodeorg, line 18: This set of arguments makes sense to me!

apps/src/aichat/aichatCompletionApi.ts
```diff
@@ -0,0 +1,89 @@
...
+  }
+}
+
+/**
+ * This function formats chat completion messages and aiCustomizations, passes data
+ * to `postAichatCompletion`, then returns the status of the response and assistant message
+ * if successful.
+ * TODO: Awaiting details on how to format input for endpoint.
+ */
+export async function getAichatCompletionMessage(
```
bencodeorg: Nit: the `get` and `post` naming imply to me two different HTTP request types...could the `post` one be `requestAiChatCompletionMessage` (or something else you come up with 😆 )

apps/src/aichat/types.ts
bencodeorg: Nice! Can AiCustomizations extend this then?

bencodeorg (approved):
Looks like a good checkpoint to me!

apps/src/aichat/aichatCompletionApi.ts
sanchitmalhotra126: +1 - could we even just merge this into `postAichatCompletionMessage` since it's mainly just passing on parameters?

apps/src/aichat/types.ts
sanchitmalhotra126: We could also use TS's `Omit` here and just simplify this to 
```
export type AiCustomizationsForBackend = Omit<AiCustomizations, 'modelCardInfo'>
```
Also, thoughts on naming this something like `AichatParameters/Settings` or something?

apps/src/aichat/redux/aichatRedux.ts
```diff
@@ -140,66 +139,64 @@ export const updateAiCustomization = createAsyncThunk(
...
-    if (systemPrompt === undefined) {
-      throw new Error('systemPrompt is undefined');
-    }
+// This thunk's callback function submits a user's chat content and AI customizations to
+// the chat completion endpoint, then waits for a chat completion response, and updates
+// the user messages.
+export const submitChatContents = createAsyncThunk(
+  'aichat/submitChatContents',
+  async (newUserMessageText: string, thunkAPI) => {
+    const state = thunkAPI.getState() as {
```
sanchitmalhotra126: Optional: can we go ahead and add `CurrentUserState` to `RootState` in `types/redux` and just cast this to `RootState`?

sanchitmalhotra126 (approved):
Looks good!

apps/src/aichat/types.ts
fisher-alice: Cool! I'll use `Omit` and rename as `AichatParameters`.

apps/src/aichat/aichatCompletionApi.ts
fisher-alice: I was wondering about combining this since we're formatting data on the backend. Will do - thanks to you both!

apps/src/aichat/views/chatWorkspace.module.scss
fisher-alice: Style update moved to separate PR https://github.com/code-dot-org/code-dot-org/pull/57996

apps/src/aichat/views/presentation/ModelCardRow.tsx
fisher-alice: Style update moved to separate PR https://github.com/code-dot-org/code-dot-org/pull/57996

apps/src/aichat/types.ts
fisher-alice: Chatted offline with Ben to get more context on this - thanks for sharing prev discussion at https://github.com/code-dot-org/code-dot-org/pull/57669#discussion_r1548185532.
Since @ebeastlake is still unsure whether she will use this on the backend, will leave in shared constants since she will move if she doesn't end up using.

apps/src/aichat/aichatCompletionApi.ts
fisher-alice: Update: sending `aiCustomizations` without `modeCardInfo` as `aiParameters`. Thanks!

PR #57945, opened 2024-04-10
Fix Violations of and Enable Optional `Layout/ClassStructure` Rubocop Rule

Hamms:
This is a big one!

This PR proposes that we enable the optional `Layout/ClassStructure` Rubocop rule, which will enforce strict ordering of different types of code blocks within a class definition. Specifically, [the default ordering it enforces](https://github.com/rubocop/rubocop/blob/276a538014d88cee5997a4204d93bc38f4ff4699/config/default.yml#L466-L483) is:

```
Layout/ClassStructure:
  ExpectedOrder:
    - module_inclusion
    - constants
    - public_class_methods
    - initializer
    - public_methods
    - protected_methods
    - private_methods
```

This by and large matches how we tend to organize our classes anyway, although of course not all of our classes conform to this all of the time. The vast majority of our classes required only a few fixes, but ~20 of them required over a hundred lines worth of changes, and three of them required over a thousand lines. See [this gist](https://gist.github.com/Hamms/08005634fd5e5d8a5d37c5997c414e2d) for an ordered summary of changes.

I decided to save the big three (`teacher_application.rb`, `unit.rb`, and `user.rb`) as future work items, and added `rubocop:ignore` directives to them. All the rest were automatically updated with `bundle exec rubocop --autocorrect-all --only Layout/ClassStructure`.

This rule is not only optional, but also fairly onerous. Automatically enforcing a strict structure makes it harder for developers to manually organize related functionality within a class together; this can arguably be a good thing, as it incentivizes us to do that organization in modules or concerns instead, thereby keeping our classes maintainably slim. But it could conversely be argued that the increased adherence to Ruby standards we gain as a result is not worth the cost in developer flexibility.

What do y'all think?

## Links

- https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Layout/ClassStructure
- https://docs.rubocop.org/rubocop/cops_layout.html#layoutclassstructure

## Testing story

Relying on existing tests to verify that this reorganization does not result in any change to functionality.

## Deployment strategy

Because of the significant number of changes in this PR and the high potential for errors via merge conflicts or new violations being introduced into the deployment pipeline by other PRs being merged at inopportune times, we do not actually enable the rule in this PR. We will plan to enable the rule in a follow-up PR once all existing violations have been fixed; hopefully that PR can be created, approved, and merged relatively quickly and therefore avoid potential conflicts.

## Follow-up work

We should plan to also fix up the three biggest offenders, whether by automatically applying the thousand or so lines worth of changes to bring them into compliance or by manually slimming them down and extracting some of their functionality into modules or concerns.


dashboard/app/controllers/api/v1/users_controller.rb
```diff
@@ -297,4 +293,7 @@ def verify_captcha
       return head :bad_request
     end
   end
```
wilkie: Should there be a newline between the `end` and the `private def`

cat5inthecradle:
Would applying #57918 fix a number of violations here?

PR #57949, opened 2024-04-11
Gen AI: separate "publish" and "save" buttons when filling out model card

bencodeorg:
## Warning!!

## Description

A few related updates related to splitting the "Publish" button in the "Publish" tab into separate "Publish" and "Save" buttons:
- "Save" will save your model card info to S3 (and allows saving incomplete information).
- A status bar tells you whether you need to fill out the card, or if you're ready to publish.
- We hide the presentation/edit mode toggle until...
- You've clicked the "publish" button, which is disabled until you've filled out the card. We set your model to published (via the new `hasPublished` state), and show the toggle at that point.
- If you go back to the edit view and save your model card into an incomplete state, we hide the presentation toggle until you've filled out a valid model card again. I chatted about this flow with 
- Hide the "Publish" tab if a levelbuilder has configured hiding presentation toggle, since it doesn't really make sense to allow publishing if you can't view the presentation view.

https://github.com/code-dot-org/code-dot-org/assets/25372625/db83a20b-6aff-42f1-bf29-7362da5615bc

## Testing story

Tested manually -- see video above. Also tested that changing the hide presentation view setting in levelbuilder resulted in the "Publish" tab being hidden.

## Follow-up work

A couple follow-ups I can think of:
- I currently have `hasPublished` not being saved to S3 (ie, state does not persist between page loads). We need to update this, but I couldn't quite figure out whether to add it as part of the `AiCustomizations` type, or store it separately. In any case, it felt like we could do it as a fast follow.
- I also think we probably want to distinguish in the notifications that go to the chat window between "publishing" and a saved update to some field in the model card. I can implement that in conjunction with the other follow-up above.


apps/src/aichat/views/modelCustomization/PublishNotes.tsx
```diff
@@ -24,24 +29,34 @@ const PublishNotes: React.FunctionComponent = () => {
     state => state.aichat.currentAiCustomizations
   );
 
-  const onUpdate = useCallback(
-    () => dispatch(updateAiCustomization()),
-    [dispatch]
-  );
+  const onSave = useCallback(() => {
```
bencodeorg, line 33: These multiple dispatches I believe are not the right way to do this, so looking for input :)

I see we have a couple instances where we put some redux updates that we want to happen after some async operation in this structure:

https://github.com/code-dot-org/code-dot-org/blob/194f7f04771c9cc1fab64fbcc598585ee8406bcd/apps/src/aichat/redux/aichatRedux.ts#L318-L331

We also dispatch some redux actions inside of the thunk here:

https://github.com/code-dot-org/code-dot-org/blob/194f7f04771c9cc1fab64fbcc598585ee8406bcd/apps/src/aichat/redux/aichatRedux.ts#L120-L139

Is one of these preferable? Is there a difference? Skimming through docs here, it seems like the first approach might be more standard?

https://redux-toolkit.js.org/api/createAsyncThunk

apps/src/aichat/redux/aichatRedux.ts
```diff
@@ -78,7 +82,10 @@ const initialState: AichatState = {
   showWarningModal: true,
   chatMessageError: false,
   currentAiCustomizations: EMPTY_AI_CUSTOMIZATIONS,
+  previouslySavedAiCustomizations: EMPTY_AI_CUSTOMIZATIONS,
   fieldVisibilities: DEFAULT_VISIBILITIES,
+  viewMode: ViewMode.EDIT,
+  hasPublished: false,
```
bencodeorg, line 93: As mentioned in the follow-ups section of the description, this needs to be persisted along with the other model customizations. Haven't decided if we should update the `AiCustomizations` type to include it, or keep it as a separate top-level property. LMK if you have opinions!

fisher-alice: Don't have a strong opinion on this either - initially, I thought it should be part of AiCustomizations as it pertains to the model card. However, its value determines whether the presentation view is possible or not, so I can see that as a reason for keeping as a top-level property. Maybe slightly leaning towards top-level.

apps/src/aichat/views/modelCustomization/PublishNotes.tsx
```diff
@@ -1,18 +1,23 @@
...
 
 import {MODEL_CARD_FIELDS_LABELS_ICONS} from './constants';
-import {isVisible, isDisabled} from './utils';
+import {isDisabled} from './utils';
 import ExampleTopicsInputs from './ExampleTopicsInputs';
-import styles from '../model-customization-workspace.module.scss';
-import {ModelCardInfo} from '../../types';
+import PublishStatus from './PublishStatus';
+import moduleStyles from './publish-notes.module.scss';
+import modelCustomizationStyles from '../model-customization-workspace.module.scss';
```
fisher-alice, line 19: Nice rename of styles.

apps/src/aichat/views/modelCustomization/PublishNotes.tsx
```diff
@@ -24,24 +29,34 @@ const PublishNotes: React.FunctionComponent = () => {
...
 
   return (
-    <div className={styles.verticalFlexContainer}>
-      {isVisible(visibility) && (
-        <div className={styles.customizationContainer}>
+    <div className={modelCustomizationStyles.verticalFlexContainer}>
+      <div>
+        {hasFilledOutModelCard(modelCardInfo)
+          ? renderPublishOkNotification()
+          : renderCompleteToPublishNotification()}
```
fisher-alice: Not sure if this is what others would expect, but it seems like once the model card is published the first time, that unless the model card is modified so that one of the fields is no longer filled out, it remains published but can be updated.

This implementation seems to imply that if a user publishes, and then updates the model card, the model card is then unpublished since message states 'Ready to publish'. Or if the user publishes then refreshes the browser, the model card is no longer published. Is that the intention? This is probably more of a product question @samantha-code .

apps/src/aichat/views/modelCustomization/PublishNotes.tsx
```diff
@@ -24,24 +29,34 @@ const PublishNotes: React.FunctionComponent = () => {
...
+    }
+  }, [dispatch, modelCardInfo]);
 
-  const getInputTag = (property: keyof ModelCardInfo) => {
-    return property === 'botName' ? 'input' : 'textarea';
-  };
+  const onPublish = useCallback(() => {
+    dispatch(updateAiCustomization());
+    dispatch(setHasPublished(true));
+    dispatch(setViewMode(ViewMode.PRESENTATION));
```
fisher-alice: When the user clicks on 'Publish' and then the 'User View' is displayed, I imagine that the user may want to immediately update Model Card fields and click on 'Edit Mode'. What do you think about saving state so that the 'Publish' tab and not 'Setup' remains open to encourage an iterative process? Also related to comment below, that 'User View' remains displayed?

apps/src/aichat/views/modelCustomization/PublishNotes.tsx
fisher-alice: Seems like either approach is reasonable but defer to @sanchitmalhotra126 on a preference.

apps/src/aichat/views/AichatView.tsx
```diff
@@ -90,7 +90,7 @@ const AichatView: React.FunctionComponent = () => {
 
   return (
     <div id="aichat-lab" className={moduleStyles.aichatLab}>
-      {!levelAichatSettings?.hidePresentationPanel && (
+      {!levelAichatSettings?.hidePresentationPanel && hasPublished && (
```
fisher-alice: Should a `useEffect` be added so that when level changes, `hasPublished` is updated? Currently, if a user goes from a level that has a completed model card and publishes and then goes to a level with an incomplete model card, the presentation mode and view mode toggle are still displayed.

fisher-alice (approved):
Nice updates! Just left a couple questions and one being a more of a product question. Thanks!

apps/src/aichat/views/modelCustomization/PublishNotes.tsx
bencodeorg: Yeah good question, it's definitely a bit confusing. I was chatting with Sam about this yesterday and we were struggling a bit to come up with the right solution. Not sure I have the exact right answer, but I think we felt like this was "close enough" and we'd interate.

apps/src/aichat/redux/aichatRedux.ts
sanchitmalhotra126: I could go either way but my instinct is to group it with `ModelCardInfo` since it pertains to the model card and is a persisted value. Might be easiest to just add a `hasPublished` to `ModelCardInfo` and make the default `false`? That way any levelbuilder-created customizations will always start as unpublished.

Edit: you know what though that being said, if it makes it more difficult when it comes to iterating through keys, maybe easier to leave it separate for now actually (it's reminding me of how `hidePresentationPanel` being a key of `LevelAiCustomizations` made things harder to work with). It does mean we'd have to update how we save to a student's project though

apps/src/aichat/views/AichatView.tsx
```diff
@@ -74,7 +74,7 @@ const AichatView: React.FunctionComponent = () => {
     ],
     size: 'm',
     selectedButtonValue: viewMode,
-    onChange: setViewMode,
+    onChange: viewMode => dispatch(setViewMode(viewMode as ViewMode)),
```
sanchitmalhotra126, line 112: Why is this cast necessary?

apps/src/aichat/views/modelCustomization/PublishNotes.tsx
sanchitmalhotra126: Yeah, I mean nothing wrong with back to back dispatches as such, but if the goal here is that we only update the publish state after saving has completed, we should move this into the async thunk. In fact, if we need to save the publish state to the project, it might actually be better to do this in the opposite order - on clicking save, if the model card is not filled out, set hasPublished to false so that when you do save, that value is saved to the project.

apps/src/aichat/views/modelCustomization/PublishNotes.tsx
```diff
@@ -70,18 +85,80 @@ const PublishNotes: React.FunctionComponent = () => {
...
       </div>
     </div>
   );
 };
 
+const getInputTag = (property: keyof ModelCardInfo) => {
+  return property === 'botName' ? 'input' : 'textarea';
+};
+
+const hasFilledOutModelCard = (modelCardInfo: ModelCardInfo) => {
```
sanchitmalhotra126: Would be good to move this to a selector in aichatRedux since it's a state computation that directly drives some UI

apps/src/aichat/views/presentation/PresentationView.tsx
```diff
@@ -10,12 +10,12 @@ import moduleStyles from './presentation-view.module.scss';
...
-  const currentAiCustomizations = useAppSelector(
-    state => state.aichat.currentAiCustomizations
+  const previouslySavedAiCustomizations = useAppSelector(
+    state => state.aichat.previouslySavedAiCustomizations
   );
   const {systemPrompt, temperature, retrievalContexts} =
-    currentAiCustomizations;
-  const modelCardInfo = currentAiCustomizations.modelCardInfo;
+    previouslySavedAiCustomizations;
+  const modelCardInfo = previouslySavedAiCustomizations.modelCardInfo;
```
sanchitmalhotra126: nit: could collapse this into `const {systemPrompt, temperature, retrievalContexts, modelCardInfo} = useAppSelector(state => state.aichat.previouslySavedAiCustomizations);`?

sanchitmalhotra126 (commented):
Nice work, this UI looks great! A few comments but nothing major.

One thing to consider in a follow-up maybe - what happens if you click publish and save back to back? Do we kick off two simultaneous saves? I don't think that's a huge deal, but it might be good to look into disabling one or the other button while a save is in progress?

apps/src/aichat/redux/aichatRedux.ts
bencodeorg: Those both sound reasonable, thanks to you both! I'll look into this as a follow-up.

apps/src/aichat/views/modelCustomization/PublishNotes.tsx
bencodeorg: Do you think if we had different status messages based on whether you have a currently published model or not would be the right path? eg, "Ready to update published model" and "In order to update your published model, you must complete the model card"?

apps/src/aichat/views/modelCustomization/PublishNotes.tsx
bencodeorg: Done (I think), LMK if this looks ok!

apps/src/aichat/views/presentation/PresentationView.tsx
bencodeorg: Sure looks like it, done!

apps/src/aichat/views/modelCustomization/PublishNotes.tsx
bencodeorg: > What do you think about saving state so that the 'Publish' tab and not 'Setup' remains open to encourage an iterative process? 

I think that makes sense! Currently, we don't manage the selected tab in redux, and the Tabs component I think would need to refactored to manage the selected tab as a prop. We also should probably switch over to using Denys's new Tabs component rather than my home baked one. Would it be ok to do that as a follow-up?

> Also related to comment below, that 'User View' remains displayed?

Not sure I'm following this part, happy to chat offline if easier!

apps/src/aichat/views/AichatView.tsx
bencodeorg: I think because onChange has to be string => void, I get this without it. Is there a better way?

```
ERROR in ./src/aichat/views/AichatView.tsx:77:48
TS2345: Argument of type 'string' is not assignable to parameter of type 'ViewMode'.
    75 |     size: 'm',
    76 |     selectedButtonValue: viewMode,
  > 77 |     onChange: viewMode => dispatch(setViewMode(viewMode)),
       |                                                ^^^^^^^^
    78 |   };
    79 |
    80 |   const chatWorkspaceHeader = (

Found 1 error in 6688 ms.
```

apps/src/aichat/views/AichatView.tsx
bencodeorg: hmm good q -- if we add this as part of the stored `AiCustomizations`, it should reset on each level, yeah?

apps/src/aichat/views/modelCustomization/PublishNotes.tsx
bencodeorg: Check out the updates in the last two commits...there's at least one thing to figure out (I have an `any` in there 😬 ), but I tried to refactor to use async thunks. I don't see the approach I have here (a shared bit of async code that takes in `thunkAPI` as an arg), which usually isn't good. LMK what you think!

bencodeorg (commented):
Just pushed a couple commits, LMK what y'all think!

apps/src/aichat/redux/aichatRedux.ts
```diff
@@ -88,57 +100,84 @@ const initialState: AichatState = {
...
-      })
+    if (!hasFilledOutModelCard(modelCardInfo)) {
+      thunkAPI.dispatch(setHasPublished(false));
+    }
+    await updateAiCustomizationShared(thunkAPI);
+  }
+);
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const updateAiCustomizationShared = async (thunkAPI: any) => {
```
sanchitmalhotra126: nit: maybe rename to `updateAndSaveAiCustomizations` or just `saveAiCustomizations`? Also I think if you just pass in `state` and `dispatch` as separate args, you should be able to get rid of the `any`? Maybe even reduce down to just the specific state fields you need (current/previousAiCustomizations). We have a similar pattern in Lab2 where different async thunks make use of common shared functions, and just pass in the specific data they need from the `thunkApi`. Ex: https://github.com/code-dot-org/code-dot-org/blob/f0320b0cffd1f042aefcabb1d8497ca0b4a277ce/apps/src/lab2/lab2Redux.ts#L388-L396

apps/src/aichat/redux/aichatRedux.ts
```diff
@@ -330,6 +375,33 @@ const aichatSlice = createSlice({
...
+  }
+
+  return true;
+};
+
+// Selectors
+const selectModelCardInfo = (state: {aichat: AichatState}) =>
+  state.aichat.currentAiCustomizations.modelCardInfo;
+
+export const selectHasFilledOutModelCard = createSelector(
```
sanchitmalhotra126, line 416: TIL about `createSelector`!

apps/src/aichat/redux/aichatRedux.ts
```diff
@@ -330,6 +375,33 @@ const aichatSlice = createSlice({
...
+    } else if (!modelCardInfo[typedKey].length) {
+      return false;
+    }
+  }
+
+  return true;
+};
+
+// Selectors
+const selectModelCardInfo = (state: {aichat: AichatState}) =>
```
sanchitmalhotra126: Does this need to be a separate selector or could this just be declared inline in `selectHasFilledOutModelCard`?

apps/src/aichat/views/AichatView.tsx
sanchitmalhotra126: Ahh right, makes sense. This seems like the right way!

apps/src/aichat/views/AichatView.tsx
sanchitmalhotra126: ^ yep, this. We just need to make sure we're reading from the current project sources and/or setting it in aichatRedux whenever the level changes

apps/src/aichat/views/modelCustomization/PublishNotes.tsx
```diff
@@ -70,18 +81,60 @@ const PublishNotes: React.FunctionComponent = () => {
...
       </div>
     </div>
   );
 };
 
+const getInputTag = (property: keyof ModelCardInfo) => {
+  return property === 'botName' ? 'input' : 'textarea';
+};
+
+const renderPublishOkNotification = () => {
```
sanchitmalhotra126: I don't exactly know what the performance implications are, but thoughts on just making these components instead of functions that create components every time the parent component is rendered?

apps/src/aichat/redux/aichatRedux.ts
```diff
@@ -88,57 +100,84 @@ const initialState: AichatState = {
...
+export const publishModel = createAsyncThunk(
+  'aichat/publishModelCard',
+  async (_, thunkAPI) => {
+    thunkAPI.dispatch(setHasPublished(true));
+    await updateAiCustomizationShared(thunkAPI);
+    thunkAPI.dispatch(setViewMode(ViewMode.PRESENTATION));
+  }
+);
+
+export const saveModelCard = createAsyncThunk(
```
sanchitmalhotra126: nice refactoring! mind dropping some comments around how the code is organized between these various thunks and the shared function?

sanchitmalhotra126 (commented):
Looks good, nice refactoring! A few more small comments

apps/src/aichat/views/modelCustomization/PublishNotes.tsx
fisher-alice: I defer to @samantha-code but I think that different status messages would be helpful to the user. 

I see Sanchit suggested making this a separate component https://github.com/code-dot-org/code-dot-org/pull/57949#discussion_r1563422721
If this was a separate component, we could also add a warning for `RetrievalCustomization` when a user adds/deletes a retrieval but still has to update/save.

apps/src/aichat/views/modelCustomization/PublishNotes.tsx
```diff
@@ -70,18 +81,60 @@ const PublishNotes: React.FunctionComponent = () => {
...
+          onClick={onSave}
+          className={modelCustomizationStyles.updateButton}
+        />
         <Button
           text="Publish"
           iconLeft={{iconName: 'upload'}}
-          disabled={isDisabled(visibility)}
-          onClick={onUpdate}
-          className={styles.updateButton}
+          disabled={isDisabled(visibility) || !hasFilledOutModelCard}
```
fisher-alice, line 100: If the model card's visibility is `readyonly` but `hidePresentationPanel` is set to `false`, should the user still be able to see model card in presentation mode? I would assume yes, but currently user is not able to toggle to 'User View' anymore since button is disabled and `hasPublished ` would be set to `false`.

apps/src/aichat/redux/aichatRedux.ts
fisher-alice: >nit: maybe rename to updateAndSaveAiCustomizations or just saveAiCustomizations?

+1 on `saveAiCustomizations`

apps/src/aichat/views/modelCustomization/PublishNotes.tsx
fisher-alice: That would be great as a follow-up and +1 on using Denys's new Tabs component - thanks!

fisher-alice: >Also related to comment below, that 'User View' remains displayed?
>Not sure I'm following this part, happy to chat offline if easier!

Ah sorry. I think once we save `hasPublished` to sources, this will not be an issue when a user moves from level to level. I think I was referring to how when a model card has been published, the toggle remain visible.

apps/src/aichat/redux/aichatRedux.ts
fisher-alice: Cool refactoring - learned a lot here.

fisher-alice (approved):
Left a question about if the model card is`readonly` but I think this is a solid checkpoint since I know you're planning a fast follow-up.
Learned a lot and thanks for the iteration and discussion!

apps/src/aichat/views/modelCustomization/PublishNotes.tsx
bencodeorg: Oh good callout! Added a commit here for that case, LMK if it looks ok:

https://github.com/code-dot-org/code-dot-org/pull/57949/commits/7f8b29e603b60ff56176db0c7d0a2e8e4d0d8838

apps/src/aichat/views/modelCustomization/PublishNotes.tsx
bencodeorg: hmm good q, I think I just had these returning HTML originally before refactoring :) I think I moved them out of the render function just for readability (and I just added some more logic to cover an edge case [Alice identified here](https://github.com/code-dot-org/code-dot-org/pull/57949#discussion_r1564107940)). Is there a difference between this and an inline ternary that returns a component on each render? Isn't the component created each time in that case as well, or is React smarter in that case and keeps them around or something?

apps/src/aichat/redux/aichatRedux.ts
bencodeorg: I think inline is fine, I was inspired by Molly's work here -- updated!

https://github.com/code-dot-org/code-dot-org/blob/d27cc4266545f8ed3c437d3e9185ec7ad7dc8b78/apps/src/javalab/redux/editorRedux.ts#L272-L287

apps/src/aichat/redux/aichatRedux.ts
bencodeorg: Updated here -- went with `saveAiCustomization` for the thunk that is exported, and `saveAiCustomizationShared` for the shared code since it aligns exactly with what happens when you use `saveAiCustomization`. LMK if that works for y'all!

https://github.com/code-dot-org/code-dot-org/pull/57949/commits/a180faa64237d3628ae74d1752deedb097da0c9c

apps/src/aichat/views/AichatView.tsx
```diff
@@ -71,9 +78,17 @@ const AichatView: React.FunctionComponent = () => {
...
-  );
+  // Showing presentation view when:
+  // 1) levelbuilder hasn't explicitly configured the toggle to be hidden, and
+  // 2) we have a published model card (either by the student, or in readonly form from the levelbuilder)
+  const showPresentationToggle = () => {
+    return (
+      !levelAichatSettings?.hidePresentationPanel &&
+      (hasPublished ||
+        (levelAichatSettings?.visibilities &&
+          isDisabled(levelAichatSettings.visibilities.modelCardInfo)))
```
fisher-alice, line 89: Not sure if this is P0 - If levelbuilders mark visibility of `modelCardInfo` as `readonly`, then it is assumed that all of the fields in model card should be filled out. Maybe worth checking with Dan on this assumption or do levelbuilders need a reminder about this.

apps/src/aichat/views/modelCustomization/PublishNotes.tsx
sanchitmalhotra126: Yeah, I think the key difference is using functions to render vs values? I don't think it matters a whole lot since these are small, but because they're static components, I was thinking we could benefit from memoizing them. I think we can either just make them static values if they're outside the main component (`const publishOkNotification = <PublishStatus ... props />`), or use `useMemo` inside the component (`const publishOkNotification = useMemo(renderPublishOkNotification))`)

apps/src/aichat/redux/aichatRedux.ts
```diff
@@ -78,64 +89,122 @@ const initialState: AichatState = {
...
 };
 
 // THUNKS
 
 // This thunk saves a student's AI customizations using the Project Manager (ie, to S3 typically),
 // then does a comparison between the previous and current saved customizations in order to
 // output a message to the chat window with the list of customizations that were updated.
-export const updateAiCustomization = createAsyncThunk(
-  'aichat/updateAiCustomization',
+export const saveAiCustomization = createAsyncThunk(
```
sanchitmalhotra126: very minor / optional - name this `updateAiCustomizations` to match "Update" button and keep the shared function just `saveAiCustomizations`?

apps/src/aichat/views/modelCustomization/PublishNotes.tsx
fisher-alice: Thanks! Left a question at https://github.com/code-dot-org/code-dot-org/pull/57949#discussion_r1566243950 next to specific line.

apps/src/aichat/views/AichatView.tsx
sanchitmalhotra126: If this is a P0, we can add a levelbuilder-side check for this as well

sanchitmalhotra126 (approved):

apps/src/aichat/views/modelCustomization/PublishNotes.tsx
fisher-alice: Mentioned at https://github.com/code-dot-org/code-dot-org/pull/57949#discussion_r1564100432, but if a separate component, could use in other customization panels like 'Retrieval' (if retrieval context added/deleted but not updated yet).

apps/src/aichat/redux/aichatRedux.ts
fisher-alice: It seems like `saveAiCustomizationShared` is similar to an internal or private function. Is there a convention we could establish for these 'shared' functions that are used internally? 'Shared 'seems to have multiple meanings so may be a little unclear to a reader.

apps/src/aichat/redux/aichatRedux.ts
fisher-alice: Commented on this already but I think this would be more clear as well.

apps/src/aichat/views/AichatView.tsx
bencodeorg: I already did it, so I'll leave this for now (I think its a sensible default). We can clarify the finer points with levelbuilders / validate down the road

apps/src/aichat/redux/aichatRedux.ts
bencodeorg: Ok you guys win, updated

apps/src/aichat/views/modelCustomization/PublishNotes.tsx
bencodeorg: I think this is already ready as a reusable component, just would need to be renamed from `PublishStatus`.

d'oh re: static value / no args, updated

fisher-alice: Awesome - thanks!

apps/src/aichat/redux/aichatRedux.ts
fisher-alice: 😁 Thanks for iterating on this!

PR #57957, opened 2024-04-11
[Google Blockly] Support block limits

mikeharv:
## About Block Limits

In 2016, we built a custom CDO Blockly feature for Maze, Artist, and Play Lab called block limits. This feature was designed to encourage students to think about their programming solutions in more efficient ways. Originally, this feature prevented students from using more than a set limit of a particular type of block. Google Blockly supports this with a `maxInstances` option ([documentation](https://developers.google.com/blockly/guides/configure/web/configuration_struct), [example](https://blockly.games/maze?lang=en&level=3&&skin=0)). However, our feature was later relaxed so that students could still drag out as many blocks as they wanted, but their pass/fail feedback would still reflect the limits. This was intended to allow students to start with an inefficient solution and then refactor (e.g. using a loop).

image

How it works:
- To set a block limit of 1, a `limit="1"` attribute is added to the block element of the toolbox XML. 
- In the student view of the level, an SVG element is rendered at the top left of the block initially showing the limit.
- As students drag out blocks (and enable them by connecting to stack), the counter over the block counts down.
- If the student has used more than the limit, the SVG counter shows `!` and changes color.
- If the student otherwise passes the level but with block than the limit of any available block, they get fail the level with a generated feedback string. The feedback includes a restatement of the limit and a rendering of the block.

This branch makes all of the above work with Google Blockly:
image

## Technical Details

**Storing Limits**
The first challenge was to find a way to be able to read and store the block limits. Blockly doesn't expect extra state to be serialized as attributes in the block element, instead preferring the use of mutators. Since we were already storing the toolbox xml string on the wrapper, I created a util `createBlockLimitMap()` that parses the xml to find these values. They are added to a map (`<string, number>`. This map is placed directly on the wrapper so that it can accessed as `Blockly.blockLimitMap` in other files.

**Event Handler**
Because the SVG will need to be updated, I registered a workspace change listener for blocks. This handler creates or updates a `blockCountMap` that also put on the wrapper instance. This is populated with the current workspace count of blocks whose types are represented in the limit map. (If a block type doesn't have a limit, we don't bother to count the instances of that block.)
Once we have the counts, we then create (or update) an SVG for each flyout block with a corresponding type. The SVG is given the number of blocks "remaining" (subtracting the count from the limit).

**Limit SVG**
In CDO Blockly, the limit is managed directly by the `BlockSvg` class.  Since this mainline doesn't support extending this class for purposes like this, I created a new `BlockSvgLimit` class that does pretty much the same thing. When needed, we will create an instance of this class and append it to the block. The actual SVG drawing is the same as the original with the exception of two minor changes:

- The colors have been updated in accordance with guidance from the Design team (@markabarrett). The new colors are listed in `cdoCss`.
- Bubble height and text position have been tweaked for better centering and to account for changing themes.

Original code for reference: https://github.com/code-dot-org/blockly/blob/v4.1.0/core/ui/block_svg/block_svg.js#L879-L951

## Validation

Again, this feature doesn't prevent students from using too many blocks, but it does prevent that from passing if they have. The functions `blockLimitExceeded` and `getBlockLimit` are used to make this determination. These functions were already defined in `cdoUtils` but not fully implemented. They now work by reading from the needed map(s) to determine if a block type and limit needs to be added to the failure feedback string.
The CDO Blockly versions of these functions are here: https://github.com/code-dot-org/code-dot-org/blob/mike/block-limits/apps/src/blockly/cdoBlocklyWrapper.js#L253-L267
The `BlockLimits` functions referenced at the link above are part of the CDO Blockly wrapper and available here: https://github.com/code-dot-org/blockly/blob/v4.1.0/core/ui/block_space/block_limits.js

## Warning!!

## Links

Jira: https://codedotorg.atlassian.net/browse/CT-495
Slack #Design thread: https://codedotorg.slack.com/archives/C0SUN2SSF/p1712781058747139

## Testing story

I manually tested that the new SVG works as expected and is resized appropriately for a variety of numeric values. Students can pass or a fail a level as expected based whether they've exceeded a block limit. The feedback string renders the correct block and limit. I also made sure that the bubbles update in real time if the theme is changed.

We already have integration tests for block limits that use maze, so once Maze is migrated our coverage of this feature will pivot to cover the new implementation for Google Blockly. 

## Follow-up work

This work does not cover levelbuilder's ability to set block limits and save them when editing toolbox blocks.. If that feature is still desired we would need to:
* show the new svg on workspace blocks if we are in toolbox edit mode
* register a new block context menu option to update values in the block limit map
* save values from the map into the xml when saving toolbox blocks

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/blockly/addons/blockSvgLimit.ts
+++ b/apps/src/blockly/addons/blockSvgLimit.ts
@@ -1,18 +1,17 @@
 import {BlockSvg} from 'blockly';
 
-const BUBBLE_SIZE = 18;
-const HALF_BUBBLE_SIZE = BUBBLE_SIZE / 2;
 /**
  * Represents a bubble on a Blockly block, displaying the count of blocks remaining
  * based on a limit initially stated in the toolbox XML.
  */
 export default class BlockSvgLimit {
-  protected element_: BlockSvg;
-  protected count: number;
-  protected className: string;
-  protected limitGroup_: SVGElement | undefined;
-  protected limitRect_: SVGElement | undefined;
-  protected limitText_: SVGElement | undefined;
+  private readonly blockSvg: BlockSvg;
+  private count: number;
+  private readonly bubbleSize: number = 18;
+  private readonly halfBubbleSize: number;
+  private limitGroup: SVGElement | undefined;
+  private limitRect: SVGElement | undefined;
+  private limitText: SVGElement | undefined;
 
   /**
    * Constructs an SVG group to track a block limit.
@@ -20,15 +19,14 @@ export default class BlockSvgLimit {
    * @param {string} [count] - The initial count to display.
    */
   constructor(element: BlockSvg, count: number) {
-    this.element_ = element;
+    this.blockSvg = element;
     this.count = count;
-    this.className = 'blocklyLimit';
+    this.halfBubbleSize = this.bubbleSize / 2;
 
-    this.limitGroup_ = undefined;
-    this.limitRect_ = undefined;
-    this.limitText_ = undefined;
-    this.element_;
-    this.initChildren();
+    this.limitGroup = undefined;
+    this.limitRect = undefined;
+    this.limitText = undefined;
+    this.initializeSvgElements();
   }
 
   /**
@@ -36,8 +34,7 @@ export default class BlockSvgLimit {
    * on the count.
    * @param {number} newCount The new count to display.
    */
-  updateCount(newCount: number) {
-    // Update the count and text content
+  public updateCount(newCount: number) {
     this.count = newCount;
     this.updateTextAndClass();
   }
@@ -47,65 +44,55 @@ export default class BlockSvgLimit {
    * has blocks remaining. If over the limit, displays an exclamation mark and changes
    * the bubble and text color.
    */
-  updateTextAndClass() {
-    if (!this.limitText_ || !this.limitGroup_) {
+  private updateTextAndClass() {
+    if (!this.limitText || !this.limitGroup) {
       return;
     }
     if (this.count >= 0) {
-      this.limitText_.textContent = `${this.count}`;
-      Blockly.utils.dom.removeClass(this.limitGroup_, 'overLimit');
-      Blockly.utils.dom.removeClass(this.limitText_, 'overLimit');
+      this.limitText.textContent = `${this.count}`;
+      Blockly.utils.dom.removeClass(this.limitGroup, 'overLimit');
+      Blockly.utils.dom.removeClass(this.limitText, 'overLimit');
     } else {
-      this.limitText_.textContent = '!';
-      Blockly.utils.dom.addClass(this.limitGroup_, 'overLimit');
-      Blockly.utils.dom.addClass(this.limitText_, 'overLimit');
+      this.limitText.textContent = '!';
+      Blockly.utils.dom.addClass(this.limitGroup, 'overLimit');
+      Blockly.utils.dom.addClass(this.limitText, 'overLimit');
     }
     this.render();
   }
 
   /**
-   * Disposes of the limit group and its children, cleaning up all references.
-   */
-  dispose() {
-    this.limitGroup_?.remove();
-    this.limitGroup_ = undefined;
-    this.limitRect_ = undefined;
-    this.limitText_ = undefined;
-  }
-
-  /**
    * Initializes the SVG elements that make up the limit bubble.
    */
-  initChildren() {
-    this.limitGroup_ = Blockly.utils.dom.createSvgElement(
+  private initializeSvgElements() {
+    this.limitGroup = Blockly.utils.dom.createSvgElement(
       'g',
       {
-        class: this.className,
+        class: 'blocklyLimit',
       },
-      this.element_.getSvgRoot()
+      this.blockSvg.getSvgRoot()
     );
 
-    this.limitRect_ = Blockly.utils.dom.createSvgElement(
+    this.limitRect = Blockly.utils.dom.createSvgElement(
       'rect',
       {
-        height: BUBBLE_SIZE,
-        width: BUBBLE_SIZE,
-        x: -HALF_BUBBLE_SIZE,
-        y: -HALF_BUBBLE_SIZE,
-        rx: HALF_BUBBLE_SIZE,
-        ry: HALF_BUBBLE_SIZE,
+        height: this.bubbleSize,
+        width: this.bubbleSize,
+        x: -this.halfBubbleSize,
+        y: -this.halfBubbleSize,
+        rx: this.halfBubbleSize,
+        ry: this.halfBubbleSize,
       },
-      this.limitGroup_
+      this.limitGroup
     );
 
-    this.limitText_ = Blockly.utils.dom.createSvgElement(
+    this.limitText = Blockly.utils.dom.createSvgElement(
       'text',
       {
         class: 'blocklyText blocklyLimit',
         'dominant-baseline': 'central',
         'text-anchor': 'middle',
       },
-      this.limitGroup_
+      this.limitGroup
     );
     this.updateTextAndClass();
   }
@@ -115,30 +102,33 @@ export default class BlockSvgLimit {
    * current count text. Called automatically when the instance is created or
    * updated.
    */
-  render() {
-    if (!this.limitGroup_ || !this.limitRect_ || !this.limitText_) {
+  private render() {
+    if (!this.limitGroup || !this.limitRect || !this.limitText) {
       // If we haven't initialized the children yet, do nothing.
       return;
     }
 
-    const svgGroup = this.element_.getSvgRoot();
-    svgGroup.append(this.limitGroup_);
+    const svgGroup = this.blockSvg.getSvgRoot();
+    svgGroup.append(this.limitGroup);
 
-    const textBBox = (this.limitText_ as SVGGraphicsElement).getBBox();
-    const rectWidth = Math.max(textBBox.width + HALF_BUBBLE_SIZE, BUBBLE_SIZE);
+    const textBBox = (this.limitText as SVGGraphicsElement).getBBox();
+    const rectWidth = Math.max(
+      textBBox.width + this.halfBubbleSize,
+      this.bubbleSize
+    );
     const rectHeight = Math.max(
-      textBBox.height + HALF_BUBBLE_SIZE / 2,
-      BUBBLE_SIZE
+      textBBox.height + this.halfBubbleSize / 2,
+      this.bubbleSize
     );
 
     // Stretch the bubble to to fit longer numbers as text.
-    this.limitRect_.setAttribute('width', `${rectWidth}`);
-    this.limitRect_.setAttribute('height', `${rectHeight}`);
+    this.limitRect.setAttribute('width', `${rectWidth}`);
+    this.limitRect.setAttribute('height', `${rectHeight}`);
     // Center the text in the bubble.
-    this.limitText_.setAttribute('x', `${rectWidth * 0.5 - HALF_BUBBLE_SIZE}`);
-    this.limitText_.setAttribute(
+    this.limitText.setAttribute('x', `${rectWidth / 2 - this.halfBubbleSize}`);
+    this.limitText.setAttribute(
       'y',
-      `${Math.ceil(rectHeight * 0.5) - HALF_BUBBLE_SIZE}`
+      `${Math.ceil(rectHeight / 2) - this.halfBubbleSize}`
     );
   }
 }
--- a/apps/src/blockly/addons/blockSvgLimit.ts
+++ b/apps/src/blockly/addons/blockSvgLimit.ts
@@ -108,9 +108,6 @@ export default class BlockSvgLimit {
       return;
     }
 
-    const svgGroup = this.blockSvg.getSvgRoot();
-    svgGroup.append(this.limitGroup);
-
     const textBBox = (this.limitText as SVGGraphicsElement).getBBox();
     const rectWidth = Math.max(
       textBBox.width + this.halfBubbleSize,
--- a/apps/src/blockly/addons/blockSvgLimit.ts
+++ b/apps/src/blockly/addons/blockSvgLimit.ts
@@ -130,9 +130,6 @@ export default class BlockSvgLimit {
     // Stretch the bubble to to fit longer numbers as text.
     this.limitRect_.setAttribute('width', `${rectWidth}`);
     // Center the text in the bubble.
-    this.limitText_.setAttribute(
-      'x',
-      `${Math.round(rectWidth * 0.5) - HALF_BUBBLE_SIZE}`
-    );
+    this.limitText_.setAttribute('x', `${rectWidth * 0.5 - HALF_BUBBLE_SIZE}`);
   }
 }
--- a/apps/src/blockly/addons/blockSvgLimit.ts
+++ b/apps/src/blockly/addons/blockSvgLimit.ts
@@ -126,10 +126,19 @@ export default class BlockSvgLimit {
 
     const textBBox = (this.limitText_ as SVGGraphicsElement).getBBox();
     const rectWidth = Math.max(textBBox.width + HALF_BUBBLE_SIZE, BUBBLE_SIZE);
+    const rectHeight = Math.max(
+      textBBox.height + HALF_BUBBLE_SIZE / 2,
+      BUBBLE_SIZE
+    );
 
     // Stretch the bubble to to fit longer numbers as text.
     this.limitRect_.setAttribute('width', `${rectWidth}`);
+    this.limitRect_.setAttribute('height', `${rectHeight}`);
     // Center the text in the bubble.
     this.limitText_.setAttribute('x', `${rectWidth * 0.5 - HALF_BUBBLE_SIZE}`);
+    this.limitText_.setAttribute(
+      'y',
+      `${Math.ceil(rectHeight * 0.5) - HALF_BUBBLE_SIZE}`
+    );
   }
 }
--- a/apps/src/blockly/addons/blockSvgLimit.ts
+++ b/apps/src/blockly/addons/blockSvgLimit.ts
@@ -1,0 +1,148 @@
+import color from '@cdo/apps/util/color';
+import {BlockSvg} from 'blockly';
+
+const BUBBLE_SIZE = 18;
+const HALF_BUBBLE_SIZE = BUBBLE_SIZE / 2;
+/**
+ * Represents a bubble on a Blockly block, displaying the count of blocks remaining
+ * based on a limit initially stated in the toolbox XML.
+ */
+export default class BlockSvgLimit {
+  protected element_: BlockSvg;
+  protected count: number;
+  protected className: string;
+  protected limitGroup_: SVGElement | undefined;
+  protected limitRect_: SVGElement | undefined;
+  protected limitText_: SVGElement | undefined;
+
+  /**
+   * Constructs an SVG group to track a block limit.
+   * @param {Element} element - The block associated with the limit.
+   * @param {string} [count] - The initial count to display.
+   */
+  constructor(element: BlockSvg, count: number) {
+    this.element_ = element;
+    this.count = count;
+    this.className = 'blocklyLimit';
+
+    this.limitGroup_ = undefined;
+    this.limitRect_ = undefined;
+    this.limitText_ = undefined;
+    this.element_;
+    this.initChildren();
+  }
+
+  /**
+   * Updates the displayed count within the limit bubble and adjusts styling based
+   * on the count.
+   * @param {number} newCount The new count to display.
+   */
+  updateCount(newCount: number) {
+    // Update the count and text content
+    this.count = newCount;
+    this.updateTextAndClass();
+  }
+
+  /**
+   * Updates the text and class of the limit bubble depending on whether the student
+   * has blocks remaining. If over the limit, displays an exclamation mark and changes
+   * the bubble and text color.
+   */
+  updateTextAndClass() {
+    if (!this.limitText_ || !this.limitGroup_) {
+      return;
+    }
+    if (this.count >= 0) {
+      this.limitText_.textContent = `${this.count}`;
+      Blockly.utils.dom.removeClass(this.limitGroup_, 'overLimit');
+      Blockly.utils.dom.removeClass(this.limitText_, 'overLimit');
+    } else {
+      this.limitText_.textContent = '!';
+      Blockly.utils.dom.addClass(this.limitGroup_, 'overLimit');
+      Blockly.utils.dom.addClass(this.limitText_, 'overLimit');
+    }
+    this.render();
+  }
+
+  /**
+   * Disposes of the limit group and its children, cleaning up all references.
+   */
+  dispose() {
+    this.limitGroup_?.remove();
+    this.limitGroup_ = undefined;
+    this.limitRect_ = undefined;
+    this.limitText_ = undefined;
+  }
+
+  /**
+   * Initializes the SVG elements that make up the limit bubble.
+   */
+  initChildren() {
+    const position = this.element_.getRelativeToSurfaceXY();
+    // Google Blockly's block ids are randomly generated and can
+    // include invalid characters for element ids. Remove everything
+    // except alphanumeric characters and whitespace, then collapse
+    // multiple adjacent whitespace to single spaces.
+    const safeCharBlockId = this.element_.id
+      .replace(/[^\w\s\']|_/g, '')
+      .replace(/\s+/g, ' ');
+
+    this.limitGroup_ = Blockly.utils.dom.createSvgElement(
+      'g',
+      {
+        class: this.className,
+      },
+      this.element_.getSvgRoot()
+    );
+
+    this.limitRect_ = Blockly.utils.dom.createSvgElement(
+      'rect',
+      {
+        height: BUBBLE_SIZE,
+        width: BUBBLE_SIZE,
+        x: -HALF_BUBBLE_SIZE,
+        y: -HALF_BUBBLE_SIZE,
+        rx: HALF_BUBBLE_SIZE,
+        ry: HALF_BUBBLE_SIZE,
+      },
+      this.limitGroup_
+    );
+
+    this.limitText_ = Blockly.utils.dom.createSvgElement(
+      'text',
+      {
+        class: 'blocklyText blocklyLimit',
+        'dominant-baseline': 'central',
+        'text-anchor': 'middle',
+      },
+      this.limitGroup_
+    );
+    this.updateTextAndClass();
+  }
+
+  /**
+   * Renders the limit bubble, adjusting its size and position based on the
+   * current count text. Called automatically when the instance is created or
+   * updated.
+   */
+  render() {
+    if (!this.limitGroup_ || !this.limitRect_ || !this.limitText_) {
+      // If we haven't initialized the children yet, do nothing.
+      return;
+    }
+
+    const svgGroup = this.element_.getSvgRoot();
+    svgGroup.append(this.limitGroup_);
+
+    const textBBox = (this.limitText_ as SVGGraphicsElement).getBBox();
+    const rectWidth = Math.max(textBBox.width + HALF_BUBBLE_SIZE, BUBBLE_SIZE);
+
+    // Stretch the bubble to to fit longer numbers as text.
+    this.limitRect_.setAttribute('width', `${rectWidth}`);
+    // Center the text in the bubble.
+    this.limitText_.setAttribute(
+      'x',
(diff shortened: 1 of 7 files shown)


apps/src/blockly/addons/blockSvgLimit.ts
```diff
@@ -0,0 +1,144 @@
...
+const HALF_BUBBLE_SIZE = BUBBLE_SIZE / 2;
+/**
+ * Represents a bubble on a Blockly block, displaying the count of blocks remaining
+ * based on a limit initially stated in the toolbox XML.
+ */
+export default class BlockSvgLimit {
+  protected element_: BlockSvg;
+  protected count: number;
+  protected className: string;
+  protected limitGroup_: SVGElement | undefined;
```
ebeastlake: Why do some of these have suffix underscores, and some don't? (I should probably know this.) I was thinking internal vs. external use, but className doesn't appear to be used outside the `blockSvgLimit` file. Maybe we could remove `className` from the constructor. There's a spot where you assign the `class: 'blocklyText blocklyLimit'` and that one isn't stored anywhere -- I'm not sure this one needs to be either.

apps/src/blockly/addons/blockSvgLimit.ts
```diff
@@ -0,0 +1,144 @@
...
+   */
+  constructor(element: BlockSvg, count: number) {
+    this.element_ = element;
+    this.count = count;
+    this.className = 'blocklyLimit';
+
+    this.limitGroup_ = undefined;
+    this.limitRect_ = undefined;
+    this.limitText_ = undefined;
+    this.element_;
```
ebeastlake: Should this be removed?

apps/src/blockly/addons/blockSvgLimit.ts
```diff
@@ -0,0 +1,144 @@
...
+    this.initChildren();
+  }
+
+  /**
+   * Updates the displayed count within the limit bubble and adjusts styling based
+   * on the count.
+   * @param {number} newCount The new count to display.
+   */
+  updateCount(newCount: number) {
+    // Update the count and text content
```
ebeastlake: I think this comment is redundant with the JSDocs.

apps/src/blockly/addons/blockSvgLimit.ts
```diff
@@ -0,0 +1,144 @@
+import {BlockSvg} from 'blockly';
+
+const BUBBLE_SIZE = 18;
+const HALF_BUBBLE_SIZE = BUBBLE_SIZE / 2;
+/**
+ * Represents a bubble on a Blockly block, displaying the count of blocks remaining
+ * based on a limit initially stated in the toolbox XML.
+ */
+export default class BlockSvgLimit {
```
ebeastlake: nit: Would renaming this to `BlockLimitIndicator` be clearer? Since this class isn't the SVG block itself?

apps/src/blockly/addons/blockSvgLimit.ts
```diff
@@ -0,0 +1,144 @@
...
+   * Constructs an SVG group to track a block limit.
+   * @param {Element} element - The block associated with the limit.
+   * @param {string} [count] - The initial count to display.
+   */
+  constructor(element: BlockSvg, count: number) {
+    this.element_ = element;
+    this.count = count;
+    this.className = 'blocklyLimit';
+
+    this.limitGroup_ = undefined;
```
ebeastlake, line 27: I think if you do
```
protected limitGroup?: SVGElement;
protected limitRect?: SVGElement;
protected limitText?: SVGElement;
```
You might not need to explicitly set these as undefined?

apps/src/blockly/addons/blockSvgLimit.ts
ebeastlake: Also, why are these protected instead of private?

apps/src/blockly/addons/blockSvgLimit.ts
```diff
@@ -0,0 +1,144 @@
...
+  protected limitGroup_: SVGElement | undefined;
+  protected limitRect_: SVGElement | undefined;
+  protected limitText_: SVGElement | undefined;
+
+  /**
+   * Constructs an SVG group to track a block limit.
+   * @param {Element} element - The block associated with the limit.
+   * @param {string} [count] - The initial count to display.
+   */
+  constructor(element: BlockSvg, count: number) {
```
ebeastlake: super nit: Maybe use `blockElement` and `initialCount`? (When I see `element`, I think of HTML elements).
```
this.blockElement_ = element;
this.count = initialCount;
```

apps/src/blockly/addons/blockSvgLimit.ts
```diff
@@ -0,0 +1,144 @@
...
+    this.limitGroup_?.remove();
+    this.limitGroup_ = undefined;
+    this.limitRect_ = undefined;
+    this.limitText_ = undefined;
+  }
+
+  /**
+   * Initializes the SVG elements that make up the limit bubble.
+   */
+  initChildren() {
```
ebeastlake: super nit: `initializeSvgElements` might be a more descriptive name for this function? These don't strike me as "children" in the classic way I think of child elements.

apps/src/blockly/addons/blockSvgLimit.ts
ebeastlake: If we have `protected` modifiers on the attributes above, should we add modifiers to the functions in this class? It seems like `updateCount` is public, but `updateTextAndClass`, `initChildren`, and `render` are private?

apps/src/blockly/addons/blockSvgLimit.ts
```diff
@@ -0,0 +1,144 @@
...
+   * has blocks remaining. If over the limit, displays an exclamation mark and changes
+   * the bubble and text color.
+   */
+  updateTextAndClass() {
+    if (!this.limitText_ || !this.limitGroup_) {
+      return;
+    }
+    if (this.count >= 0) {
+      this.limitText_.textContent = `${this.count}`;
+      Blockly.utils.dom.removeClass(this.limitGroup_, 'overLimit');
```
ebeastlake: Curious: Do we have to use this through the Blockly utils instead of just doing `this.limitGroup_.classList.remove`?

apps/src/blockly/addons/blockSvgLimit.ts
```diff
@@ -0,0 +1,144 @@
...
+      Blockly.utils.dom.addClass(this.limitGroup_, 'overLimit');
+      Blockly.utils.dom.addClass(this.limitText_, 'overLimit');
+    }
+    this.render();
+  }
+
+  /**
+   * Disposes of the limit group and its children, cleaning up all references.
+   */
+  dispose() {
```
ebeastlake: Where is this used?

apps/src/blockly/addons/blockSvgLimit.ts
```diff
@@ -0,0 +1,144 @@
...
+    const rectHeight = Math.max(
+      textBBox.height + HALF_BUBBLE_SIZE / 2,
+      BUBBLE_SIZE
+    );
+
+    // Stretch the bubble to to fit longer numbers as text.
+    this.limitRect_.setAttribute('width', `${rectWidth}`);
+    this.limitRect_.setAttribute('height', `${rectHeight}`);
+    // Center the text in the bubble.
+    this.limitText_.setAttribute('x', `${rectWidth * 0.5 - HALF_BUBBLE_SIZE}`);
```
ebeastlake: Could this just be `rectWidth / 2`? And similarly below.

apps/src/blockly/addons/blockSvgLimit.ts
```diff
@@ -0,0 +1,144 @@
...
+   * updated.
+   */
+  render() {
+    if (!this.limitGroup_ || !this.limitRect_ || !this.limitText_) {
+      // If we haven't initialized the children yet, do nothing.
+      return;
+    }
+
+    const svgGroup = this.element_.getSvgRoot();
+    svgGroup.append(this.limitGroup_);
```
ebeastlake: What do lines 124 and 125 do?

apps/src/blockly/addons/blockSvgLimit.ts
```diff
@@ -0,0 +1,144 @@
+import {BlockSvg} from 'blockly';
+
+const BUBBLE_SIZE = 18;
```
ebeastlake: If you wanted to, I think you could add these to the constructor and make them
`private readonly bubbleSize: number = 18;`

apps/src/blockly/addons/cdoUtils.ts
```diff
@@ -318,11 +318,48 @@ export function isWorkspaceReadOnly() {
...
-  return false;
+/**
+ * Checks if any block type's usage count exceeds its defined limit and returns
+ * the type of the first block found to exceed.
+ * @returns {string | undefined} The type of the first block that exceeds its limit,
+ * or undefined if no block exceeds the limit.
+ */
+export function blockLimitExceeded(): string | undefined {
+  const blockLimitMap = Blockly.blockLimitMap;
+  const blockCountMap = Blockly.blockCountMap;
```
ebeastlake: nit: I think you can destructure these, i.e. 
```
const { blockLimitMap, blockCountMap } = Blockly;
```

apps/src/blockly/addons/cdoUtils.ts
```diff
@@ -318,11 +318,48 @@ export function isWorkspaceReadOnly() {
...
+  for (const [type, count] of blockCountMap) {
+    const limit = blockLimitMap.get(type);
+    // If a limit is defined and the count exceeds this limit, return the type
+    if (limit !== undefined && count > limit) {
+      return type;
+    }
+  }
+
+  // If no count exceeds the limit, return undefined
+  return undefined;
```
ebeastlake: I think returning `null` is a more intentional choice (i.e., nothing was found, and we explicitly returned null) vs. undefined.

apps/src/blockly/addons/cdoUtils.ts
```diff
@@ -318,11 +318,48 @@ export function isWorkspaceReadOnly() {
...
+ */
+export function getBlockLimit(type: string): number {
+  const blockLimitMap = Blockly.blockLimitMap;
+
+  // Check if the block limit map is defined and has the requested type
+  if (blockLimitMap && blockLimitMap.has(type)) {
+    return blockLimitMap.get(type) || 0;
+  }
+
+  // Return undefined if the map is not defined or the type is not found
```
ebeastlake: I think this comment is inaccurate since it references returning `undefined`. Also, what you have in this function might be the same as:

```
return Blockly.blockLimitMap?.get(type) ?? 0;
```

apps/src/blockly/addons/cdoUtils.ts
```diff
@@ -503,6 +540,44 @@ export function getLevelToolboxBlocks(customCategory: string) {
   }
 }
 
+/**
+ * Creates a map of block types and limits, based on limit attribtues found
```
ebeastlake: nit: Misspelled `attribtues`

apps/src/blockly/addons/cdoUtils.ts
```diff
@@ -503,6 +540,44 @@ export function getLevelToolboxBlocks(customCategory: string) {
...
+ * in the block XML for the current toolbox.
+ * @returns {Map} A map of block limits
```
ebeastlake: nit: `{Map<string, number>}` might be useful information to have here?

apps/src/blockly/addons/cdoUtils.ts
```diff
@@ -503,6 +540,44 @@ export function getLevelToolboxBlocks(customCategory: string) {
...
+ */
+export function createBlockLimitMap() {
+  const parser = new DOMParser();
+  // This method only works for string toolboxes.
```
ebeastlake, line 540: What's an example of a non-string toolbox?

mikeharv: All of our toolboxes should be strings, but Blockly also supports XML (not stringified) and JSON:
https://developers.google.com/blockly/guides/configure/web/toolbox#xml-string

If we ever want to support JSON toolboxes (e.g. to align with Music Lab), we need to build in support here.

apps/src/blockly/addons/blockSvgLimit.ts
mikeharv: A lot of these ended up being `protected` because I had initially copied some boiler plate from `svgFrame` which has two derived classes for blocks and workspaces.

apps/src/blockly/addons/blockSvgLimit.ts
mikeharv: Agreed!

apps/src/blockly/addons/blockSvgLimit.ts
mikeharv: It's a limit indicator _for_ an SVG block. Similar to `BlockSvgFrame` and `WorkspaceSvgFrame` in terms of naming convention. Since we wouldn't put one on an (unrendered) `Block`, could we do `BlockSvgLimitIndicator`?

apps/src/blockly/addons/blockSvgLimit.ts
mikeharv: Agreed on the `initialCount` parameter being used to set `this.count`. Before I got to this comment, I had already renamed `element` to `blockSvg`. That feels more descriptive to me than block element. Does that work?

apps/src/blockly/addons/blockSvgLimit.ts
mikeharv: No, I don't think we technically need to since they effectively do the same thing. 
https://github.com/google/blockly/blob/blockly-v11.0.0.-beta.6/core/utils/dom.ts#L105-L112

I think I'd lean towards keeping the Blockly util since Blockly is also managing our CSS rules to an extent.

apps/src/blockly/addons/blockSvgLimit.ts
mikeharv: Ha, good call. It's not. An an earlier uncommitted version, I was disposing and re-creating everything each time there was an update to the count. Will remove!

apps/src/blockly/addons/blockSvgLimit.ts
mikeharv: Yes, it could be. Happy to switch if there's a preference!

apps/src/blockly/addons/blockSvgLimit.ts
mikeharv: Sounds good!

apps/src/blockly/addons/cdoUtils.ts
mikeharv: Sounds good. I was aligning to the cdo version which uses `goog.object.findKey` (https://github.com/google/closure-library/blob/7818ff7dc0b53555a7fb3c3427e6761e88bde3a2/closure/goog/object/object.js#L261-L268), but for our purposes `null` works fine and is more intentional.

apps/src/blockly/addons/cdoUtils.ts
mikeharv: Nice!

apps/src/blockly/addons/blockSvgLimit.ts
mikeharv: Initially, I was using them to actually put the group onto the block, because I didn't realize that Blockly's `createSvgElement` util allowed you to specify the parent for the same purpose. These lines were redundant!

apps/src/blockly/addons/cdoUtils.ts
```diff
@@ -503,6 +529,44 @@ export function getLevelToolboxBlocks(customCategory: string) {
...
+  // Define blockLimitMap
+  const blockLimitMap = new Map<string, number>();
+
+  // Select all block elements and convert NodeList to array
+  const blocks = Array.from(xmlDoc.querySelectorAll('block'));
+
+  // Iterate over each block element using forEach
+  blocks.forEach(block => {
+    // Check if the block has a limit attribute
+    const limitAttribute = block.getAttribute('limit');
```
ebeastlake: I think you can simplify this check to:

```
const limitAttribute = block.getAttribute('limit');
const limit = parseInt(limitAttribute);  // Directly parse the attribute

if (!isNaN(limit)) {
```

`null` and `undefined` should not throw an error in `parseInt` I think and then you can check whether they're numbers.

apps/src/blockly/addons/cdoUtils.ts
```diff
@@ -503,6 +529,44 @@ export function getLevelToolboxBlocks(customCategory: string) {
...
+    if (limitAttribute !== null) {
+      // Check if the limit attribute is a valid number
+      const limit = parseInt(limitAttribute);
+      if (!isNaN(limit)) {
+        // Extract type and add to blockLimitMap
+        const type = block.getAttribute('type');
```
ebeastlake: If `getAttribute` returns an empty string, do you want to call `blockLimitMap.set`?

mikeharv: I don't think it can? All block element should have a type. The only way around that would be to manually add `<block type=""/>` to the level XML, but I don't think we need to be overly concerned about a graceful failure in a case like that.

apps/src/blockly/eventHandlers.ts
```diff
@@ -7,6 +7,9 @@ import {BlockChange} from 'blockly/core/events/events_block_change';
 import {BlockMove} from 'blockly/core/events/events_block_move';
 import {BlockCreate} from 'blockly/core/events/events_block_create';
```
ebeastlake: Can you do 
```
import { BlockChange, BlockMove, BlockCreate, ThemeChange } from 'blockly/core/events';
```
?

apps/src/blockly/eventHandlers.ts
```diff
@@ -117,3 +120,78 @@ export function reflowToolbox() {
     modalWorkspace?.getFlyout()?.reflow();
   }
 }
+export function updateBlockLimits(event: Abstract) {
+  // This check is to update bubbles that show block limits whenever
+  // blocks on the main workspace are updated.
+  if (
```
ebeastlake, line 127: nit: We could add an `isRelevantEvent` helper for this.

```
if (!isRelevantEvent(event)) {
  return;
}
```

apps/src/blockly/eventHandlers.ts
```diff
@@ -117,3 +120,78 @@ export function reflowToolbox() {
...
+  const blockLimitMap = Blockly.blockLimitMap;
+
+  if (!blockEvent.workspaceId || !blockLimitMap || !(blockLimitMap?.size > 0)) {
+    return;
+  }
+  const eventWorkspace = Blockly.Workspace.getById(
+    blockEvent.workspaceId
+  ) as ExtendedWorkspaceSvg;
+  const allWorkspaceBlocks = eventWorkspace?.getAllBlocks();
+  if (!allWorkspaceBlocks) {
```
ebeastlake: Will `.getAllBlocks()` always return an array? Do you need to check for this?

mikeharv: It will always return an array. In this case the check is really checking whether there's a valid eventWorkspace or not, rather than checking that we have a block array.

apps/src/blockly/addons/cdoUtils.ts
mikeharv: TypeScript actually yells at you if you try `parseInt(limitAttribute)` because `Argument of type 'string | null' is not assignable to parameter of type 'string'.` However, we could probably pass it in as a template string which would coerce `null` to `'null'`.

apps/src/blockly/eventHandlers.ts
mikeharv: I can potentially do this:

```
import type {
  BlockChange,
  BlockMove,
  BlockCreate,
  ThemeChange
} from 'blockly/core/events/events';
```

apps/src/blockly/eventHandlers.ts
mikeharv: Would that help with readability? Could also do something like 
`![...expectedEventTypes].includes(event.type)`

apps/src/blockly/addons/blockSvgLimitIndicator.ts
```diff
@@ -0,0 +1,131 @@
...
+
+/**
+ * Represents a bubble on a Blockly block, displaying the count of blocks remaining
+ * based on a limit initially stated in the toolbox XML.
+ */
+export default class BlockSvgLimitIndicator {
+  private readonly blockSvg: BlockSvg;
+  private count: number;
+  private readonly bubbleSize: number = 18;
+  private readonly halfBubbleSize: number;
```
molly-moen: why don't we set `halfBubbleSize` to 9 here since it's always `bubbleSize / 2`? Also should those `bubbleSize` and `halfBubbleSize` be constants rather than fields since they should never change?

apps/src/blockly/addons/blockSvgLimitIndicator.ts
```diff
@@ -0,0 +1,131 @@
...
+      Blockly.utils.dom.addClass(this.limitText, 'overLimit');
+    }
+    this.render();
+  }
+
+  /**
+   * Initializes the SVG elements that make up the limit bubble.
+   */
+  private initializeSvgElements() {
+    this.limitGroup = Blockly.utils.dom.createSvgElement(
```
molly-moen: We could avoid having to do a bunch of null checks on `limitGroup` and `limitRect` by assigning them in the constructor rather than a helper method. Then we would know they are never actually undefined.

apps/src/blockly/addons/cdoUtils.ts
```diff
@@ -317,13 +317,39 @@ export function bindBrowserEvent(
...
+  // If no count exceeds the limit, return null.
+  return null;
 }
 
-export function getBlockLimit() {
-  return 0;
+/**
+ * Retrieves the block limit for a given block type from the block limit map.
+ * @param {string} type The type of the block to check the limit for.
+ * @returns {number} The limit for the specified block type, or 0 if not found.
```
molly-moen: Do we never set the limit to 0? Is using 0 as the limit a pattern that already exists? It seems a little odd to me to mean 0 == no limit, but if we already have the logic in place we can leave it as-is. Otherwise I would do something like null or -1.

apps/src/blockly/eventHandlers.ts
```diff
@@ -117,3 +122,78 @@ export function reflowToolbox() {
...
   }
 }
+export function updateBlockLimits(event: Abstract) {
+  // This check is to update bubbles that show block limits whenever
+  // blocks on the main workspace are updated.
+  if (
+    event.type !== Blockly.Events.BLOCK_CHANGE &&
+    event.type !== Blockly.Events.BLOCK_MOVE &&
+    event.type !== Blockly.Events.BLOCK_CREATE &&
+    event.type !== Blockly.Events.THEME_CHANGE
```
molly-moen: why do we need to check theme change events?

molly-moen (commented):
looks good, a couple comments/questions!

apps/src/blockly/eventHandlers.ts
ebeastlake: We could add an explicit workspace check instead, right?
```
if (!eventWorkspace) {
return;
}
```

apps/src/blockly/eventHandlers.ts
```diff
@@ -117,3 +122,78 @@ export function reflowToolbox() {
...
+  ) as ExtendedWorkspaceSvg;
+  const allWorkspaceBlocks = eventWorkspace?.getAllBlocks();
+  if (!allWorkspaceBlocks) {
+    return;
+  }
+  // Define a Map to store block counts for each type
+  const blockCountMap = new Map<string, number>();
+  Blockly.blockCountMap = blockCountMap;
+  // Initialize block counts based on blockLimitMap
+  blockLimitMap.forEach((_, type) => {
```
ebeastlake, line 157: Why do we need to initialize all of these to zero?

apps/src/blockly/eventHandlers.ts
```diff
@@ -117,3 +122,78 @@ export function reflowToolbox() {
...
+
+  // Count the enabled blocks of each type
+  allWorkspaceBlocks
+    .filter(block => blockLimitMap.has(block.type) && block.isEnabled())
+    .forEach(block => {
+      const type = block.type;
+      if (blockCountMap.has(type)) {
+        const currentCount = blockCountMap.get(type) || 0;
+        blockCountMap.set(type, currentCount + 1);
+      }
```
ebeastlake: Are lines 156 through 170 the same as:

```
 allWorkspaceBlocks.forEach(block => {
    if (!block.isEnabled() || !Blockly.blockLimitMap.has(block.type)) {
      return;
    }
    const currentCount = blockCountMap.get(block.type) || 0;
    blockCountMap.set(block.type, currentCount + 1);
  });
  ```

File: apps/src/blockly/eventHandlers.ts [hunk 30]
```diff
@@ -117,3 +122,78 @@ export function reflowToolbox() {
...
+  const flyoutBlocks = flyout
+    .getWorkspace()
+    .getTopBlocks() as ExtendedBlockSvg[];
+
+  // Get the flyout blocks that have limits
+  const limitedFlyoutBlocks = flyoutBlocks.filter(block => {
+    return blockLimitMap.has(block.type);
+  });
+
+  limitedFlyoutBlocks.forEach(flyoutBlock => {
```
ebeastlake: nit: If you're going to do a filter and then a forEach, I have a slight preference for just doing a forEach with an early return,

```
flyoutBlocks.forEach(block => {
    const hasLimit = blockLimitMap.has(block.type);
    if (!hasLimit) {
      return;
    }
   ...
}
```
Since you retrieve the limit later anyway, you might be able to do...
```
flyoutBlocks.forEach(block => {
    const limit = blockLimitMap.get(block.type);
    if (!limit) { // depending on how to handle zero
      return;
    }
    const blockUsedCount = blockCountMap.get(flyoutBlock.type) || 0;
   ...
}
```

apps/src/blockly/addons/cdoUtils.ts
ebeastlake: Oh, right! Maybe TypeScript would be happy with
```
const limit = parseInt(block.getAttribute('limit') ?? '');
if (!isNaN(limit)) {
```

apps/src/blockly/addons/cdoUtils.ts
ebeastlake: Sounds good. I based this conversation on MDN, https://developer.mozilla.org/en-US/docs/Web/API/Element/getAttribute, but this is blocks not `Element`s!

apps/src/blockly/addons/blockSvgLimitIndicator.ts
ebeastlake: Ha, whoops, I was assuming making something private and readonly was perhaps better than a constant. Mike had it that way before. Sorry for giving bad feedback, Mike 🙃

molly-moen: No worries, I think a classic constant is more readable here. If I see `this.` I assume it's a changeable field. But I could not be following correct typescript protocol :)

apps/src/blockly/addons/cdoUtils.ts
mikeharv: The blocks are xml elements (renaming to make that clear). The documentation you linked says:

> All modern web browsers return `null` when the specified attribute does not exist on the specified element.

This is why I do a null check below. I think we may be overanalyzing this though because Blockly guarantees that all block elements will have a type.

apps/src/blockly/addons/blockSvgLimitIndicator.ts
mikeharv: Awesome idea!

apps/src/blockly/addons/cdoUtils.ts
mikeharv: To my surprise, we _do_ have a handful of levels where we set the limit to 0. (I would have thought in that case we'd just leave the block out of the toolbox...)

I'll try setting it up to work with `null`.

apps/src/blockly/eventHandlers.ts
mikeharv: If you switch in/out of high contrast mode, the size of all `blocklyText` changes, including the on the indicators. Will leave a comment!

apps/src/blockly/eventHandlers.ts
mikeharv: I think we want to make sure that both maps include the same keys (block types). Initializing to 0 ensures that blocks that the student hasn't used yet are still list in the count map.

apps/src/blockly/addons/blockSvgLimitIndicator.ts
mikeharv: Reverted!

molly-moen (approved):

ebeastlake (approved):

PR #57966, opened 2024-04-11
School Association UI Tweaks

hannahbergam:
This PR handles several UI updates as detailed below. I did switch back to the form saving the raw ncesId (which might be something like 'clickToAdd') so I added those strings to the cleanup work (thus assigning an empty ncesId) in the finish_sign_up.js file. 

- Divider line in dropdown (opt groups)
- US at the top
- Dropdown appears at same time as zip but is disabled
    - In addition, creates disabled states for label and button below
- Fixed width styling on input box (followup: Ben is working on fixing the width to the dropdowns)
- Update the string for add school
- Adds a new button ‘I don’t teach in a school setting' to below the dropdown
- Moves the static options to the bottom of the schools list
- Moves ‘select a school’ option to top of schools list

https://github.com/code-dot-org/code-dot-org/assets/37230822/56d4baa8-4544-4cbf-a41c-53f32c77b1af

Screenshot 2024-04-11 at 12 28 51 PM

Screenshot 2024-04-11 at 4 49 16 PM

## Warning!!

## Links

- spec: []()
- jira ticket: https://codedotorg.atlassian.net/browse/ACQ-1746

## PR Checklist:

- [ ] Code is well-commented


apps/src/templates/SchoolDataInputs.jsx
```diff
@@ -29,8 +29,13 @@ export default function SchoolDataInputs({
   const [country, setCountry] = useState('');
   const [zipSearchReady, setZipSearchReady] = useState(false);
 
-  let COUNTRY_ITEMS = [{value: 'selectCountry', text: i18n.selectCountry()}];
-  for (const item of Object.values(COUNTRIES)) {
+  let COUNTRY_ITEMS = [
+    {value: 'selectCountry', text: i18n.selectCountry()},
+    {value: 'US', text: 'United States'},
```
TurnerRiley: Should 'United States' be a translatable string?

apps/src/templates/SchoolDataInputs.jsx
```diff
@@ -29,8 +29,13 @@ export default function SchoolDataInputs({
...
+  ];
+  for (const item of Object.values(COUNTRIES).filter(function (item) {
+    return item.label !== 'US';
+  })) {
```
TurnerRiley: Nit: Could the filter logic either be moved to the line before or could the words be changed from "item"? It's a bit confusing with two different `item`s being used

TurnerRiley (approved):
Super stoked about how all this looks! Just had a couple tiny nits/questions

apps/src/templates/SchoolZipSearch.jsx
```diff
@@ -3,23 +3,24 @@ import PropTypes from 'prop-types';
...
 import Button from '@cdo/apps/templates/Button';
 import analyticsReporter from '@cdo/apps/lib/util/AnalyticsReporter';
 import {EVENTS, PLATFORMS} from '@cdo/apps/lib/util/AnalyticsConstants';
 
-const SELECT_A_SCHOOL = 'selectASchool';
-const CLICK_TO_ADD = 'clickToAdd';
-const NO_SCHOOL_SETTING = 'noSchoolSetting';
+export const SELECT_A_SCHOOL = 'selectASchool';
+export const CLICK_TO_ADD = 'clickToAdd';
+export const NO_SCHOOL_SETTING = 'noSchoolSetting';
```
bethanyaconnor, line 15: not necessarily something to address in this PR, but I wonder if exporting these consts is indicative of needing a constants file of some kind

bethanyaconnor (approved):
Looks good after addressing Turner's comments!

apps/src/templates/SchoolDataInputs.jsx
hannahbergam: I'm happy to add one- curious about the status of the current discussion with Jorge about this. For the rest of the countries I'm using the strings from the geographyConstants.js file (which are not translated). Do we have a plan to move toward a solution where all the countries are in their own locale? Otherwise, the US will be the only translated option.

apps/src/templates/SchoolDataInputs.jsx
hannahbergam: I can definitely rename one! Is it less confusing to name the second one 'country' even though both 'items' represent the same thing?

hannahbergam: Or I could rename them both to country so it is more clear what they represent?

apps/src/templates/SchoolDataInputs.jsx
bethanyaconnor: hm interesting! Yeah, that might be a question for Jorge and possibly a follow-up in a different PR

apps/src/templates/SchoolDataInputs.jsx
bethanyaconnor: Curious what Turner thinks but it might be clearer if the filter logic is on a separate line. IMO if you do that, using the same variable for both would be reasonable.

So something like 
```
nonUsCountries = Object.values(COUNTRIES).filter(item => item.value !== 'US')
for (const item of nonUsCountries) {
....
}
```

apps/src/templates/SchoolDataInputs.jsx
hannahbergam: Okay I found [where I was reading about it](https://codedotorg.atlassian.net/browse/ACQ-1720)! Not entirely the same but very relevant. I think for now I will add a string to translate 'United States' and add a comment to this ticket about other places on our site where all countries are displayed.

hannahbergam: Talked to Dani and [created this ticket](https://codedotorg.atlassian.net/browse/ACQ-1788) for ACQ Up Next

apps/src/templates/SchoolDataInputs.jsx
hannahbergam: Ahh clutch okay I updated it. I also added the translation and a couple comments to hopefully make it all clearer!

apps/src/templates/SchoolZipSearch.jsx
hannahbergam: I made a ticket for this! Good call https://codedotorg.atlassian.net/browse/ACQ-1789

bethanyaconnor (approved):
🎉

apps/src/templates/SchoolDataInputs.jsx
TurnerRiley: Yep I agree with Bethany!

apps/src/templates/SchoolDataInputs.jsx
TurnerRiley: Oh I didn't realize we didn't translate country names! Thank you for making the ticket!

PR #57969, opened 2024-04-11
P20-574 Allow printing translated lesson plans

mgc1194:
When trying to print lesson plans in languages other than English the following error occurs:
```
<Error>
<Code>NoSuchKey</Code>
<Message>The specified key does not exist.</Message>
<Key>csd1-2021/fallback/teacher-lesson-plans/Introduccion-a-la-resolucion-de-problemas.pdf</Key>
<RequestId>ADGWR3FMTQK60HJ6</RequestId>
<HostId>oBfid1myx4TvC+w0uJ7Anj1MELTvMRAj76dJq5g3Y8X8UtKvzjMbELwOXZkp0S9j3abT/GONaQQ=</HostId>
</Error>
```
This happens because, during the build process, we only generate PDFs for English resources, as not all resources are translated into other languages, and it would take a long time to generate PDFs for all languages. The current workaround for teachers is to use the `cmd+p` shortcut instead of the listed printing options.

This PR does:
- Removes the links to PDF resources when the locale is different to English and enables the default option to print the current window.
- Updates the front-end Firehose client to log the locale when `AppOptions` is not present.
- Logs a firehose event when a lesson plan is printed.

## Links

- Jira ticket: [P20-574](https://codedotorg.atlassian.net/browse/P20-574)

## Testing story
New behaviour when not in English:
https://github.com/code-dot-org/code-dot-org/assets/66776217/50a23419-6554-4958-9a41-2bd3003ffdc5

Same options in English:
Screenshot 2024-07-09 at 22 20 49

## PR Checklist:

- [ ] Code is well-commented


apps/src/templates/lessonOverview/LessonOverview.jsx
```diff
@@ -62,14 +63,14 @@ class LessonOverview extends Component {
     );
   }
 
-  recordAndNavigateToPdf = (e, firehoseKey, url) => {
-    // Prevent navigation to url until callback
-    e.preventDefault();
+  recordAndHandleResource = (e, firehoseKey, action, url = null) => {
+    e.preventDefault(); // Prevent navigation to url until callback
+    const event = action === 'navigate' ? 'open-pdf' : 'print-from-browser';
```
kobryan0619: Optional: Do you think it makes sense to put `'navigate'` , `'open-pdf`' etc in constants at the top of this file instead of using these strings (ex. `const OPEN_PDF = 'open-pdf'`)

kobryan0619 (approved):
Looks good - one optional change, but otherwise looks great!

apps/src/templates/lessonOverview/LessonOverview.jsx
mgc1194: Probably.
It's a large component and I thought about making a separate react component, but I was not sure what's the teacher-tools team's preference.
I will go ahead and move those variables to the top. 
I will also add some screenshots to the PR description.

apps/src/templates/lessonOverview/LessonOverview.jsx
```diff
@@ -110,6 +115,34 @@ class LessonOverview extends Component {
     return options;
   };
 
+  renderPrintOptions = () => {
+    const pdfDropdownOptions = this.compilePdfDropdownOptions();
+
+    if (pdfDropdownOptions.length > 0 && userLocale === 'en-US') {
+      return pdfDropdownOptions.map(option => (
+        <a
```
wilkie, line 132: these act like buttons but are `a` tags... is that a limitation of this component?

apps/src/templates/lessonOverview/LessonOverview.jsx
```diff
@@ -345,6 +364,9 @@ class LessonOverview extends Component {
   }
 }
 
+// Get the current locale.
```
wilkie: I'm curious if this component could work with the localesRedux? would be nice to standardize the provider of the locale code.

mgc1194: Agree on this. Do you have an example of the `localesRedux` being used?

apps/src/templates/lessonOverview/LessonOverview.jsx
mgc1194: I just kept the same tags so the style was consistent across users without having to dig too much into the styles.

apps/src/lib/util/firehose.js
```diff
@@ -118,6 +119,8 @@ class FirehoseClient {
   getLocale() {
     if (window.appOptions) {
       return window.appOptions.locale;
+    } else {
+      return cookies.get('language_') || 'en-US';
```
artem-vavilov: Please try to use this function if possible: https://github.com/code-dot-org/code-dot-org/blob/staging/apps/src/util/currentLocale.ts

mgc1194: Awesome! Thanks, I thought we had some function, but I could not find it.

apps/src/templates/lessonOverview/LessonOverview.jsx
wilkie: currentLocale function also works. I'm not sure the locale redux is initialized on these pages anyway.

apps/src/templates/lessonOverview/LessonOverview.jsx
mgc1194: Actually, it does not matter whether we pass links or buttons, the parent component re-tags everything as links:
https://github.com/code-dot-org/code-dot-org/blob/5d7b3eed7c8bbf024d7764919002f65afb93ccf1/apps/src/templates/DropdownButton.js#L105-L120

apps/src/templates/lessonOverview/LessonOverview.jsx
```diff
@@ -62,14 +70,15 @@ class LessonOverview extends Component {
     );
   }
 
-  recordAndNavigateToPdf = (e, firehoseKey, url) => {
-    // Prevent navigation to url until callback
-    e.preventDefault();
+  recordAndHandleResource = (e, firehoseKey, action, url = null) => {
+    e.preventDefault(); // Prevent navigation to url until callback
+    const event =
+      action === ResourceActions.NAVIGATE ? 'open-pdf' : 'print-from-browser';
```
mgc1194, line 77: I kept `'open-pdf'` and `'print-from-browser'` strings here, instead of declaring them at the beginning of the code, since they are exclusively used in `firehoseClient.putRecord`.

wilkie (approved):

apps/src/templates/lessonOverview/LessonOverview.jsx
```diff
@@ -110,6 +123,41 @@ class LessonOverview extends Component {
     return options;
   };
 
+  renderPrintOptions = () => {
+    const pdfDropdownOptions = this.compilePdfDropdownOptions();
+
+    if (pdfDropdownOptions.length > 0 && currentLocale() === 'en-US') {
```
artem-vavilov: Can `DefaultLocale` from `sharedConstants` be used in this context?
https://github.com/code-dot-org/code-dot-org/blob/staging/apps/src/util/currentLocale.ts#L3

artem-vavilov (approved):
Nice work, Mario! 🙌🏻

PR #57973, opened 2024-04-11
Updated product tour UI test

Nokondi:
Updated UI and unit tests for the AI rubric product tour. Update includes the following:
- UI test checks for correct data displayed in the rubric during the tour and after the tour ends
- UI test restarts tour using button in rubric header
- UI test backtracks through tour using Back button
- UI test checks that tour does not automatically load after teacher completes it and reloads the page
- Eyes test for the complete tour
- Unit tests to confirm that tour does not run on non-assessment levels


davidsbailey:
what do you think about covering this item from the jira ticket: 
* no product tour on non-eval levels

you could append it to the current product tour scenario, just navigate and then confirm that the product tour is not enabled?

davidsbailey (commented):
please re-request review when you are ready for another look. thanks!

davidsbailey (approved):
Looks good, thanks for the updates!

PR #57975, opened 2024-04-11
restore product tour test, fixing leak in dialogHelperTest

davidsbailey:
The product tour unit test (currently disabled) has been passing when run individually but failing when run as part of the apps test suite. a binary search of all unit tests revealed the dialogHelperTest was to blame. the root cause is that when this test stubs out and restores jQuery, it leaves a new `window.jQuery = undefined` property on the window object. Root cause for why this breaks the product tour test was not confirmed, but one guess is that the introjs library, which the product tour depends on, checks for the presence of jQuery in a way that is sensitive to this difference: https://github.com/usablica/intro.js/blob/3cab7911743415f25bd5d9ba250bf69f9e8273f9/src/util/cloneObject.ts#L14

The solution is to fix the leak by removing the jQuery property from the window object after dialogHelperTest runs.

alternatives considered -- run the leaky test last: https://github.com/code-dot-org/code-dot-org/pull/57939

thanks to @Nokondi and @wilkie for all the time spent narrowing down this elusive test issue!

## Testing story

passing drone run as well as local `yarn test:unit`


wilkie (approved):

Nokondi (approved):
Great work

davidsbailey:
Thanks Mark, thank you both for all the hard work too!

PR #57977, opened 2024-04-11
Gen AI: Move model descriptions to a JSON file and populate model selection UI from redux and json

cnbrenci:
Hooked up ModelDescription JSON to the UI, namely the `SetupCustomizations`, `CompareModelsDialog`, `ModelDescriptionPanel`, and `PresentationView`.

- Moved modelDescriptions into a json. 
- Added selectedModelId to the AiCustomizations for use in redux and setting in levelbuilder. 
- Added availableModelIds to the levelAichatSettings so they can be set from levelbuilder.
- Populate technical details in Presentation View from the model descriptions.

## Links

[Create and store JSON file with curated models info and fetch curated models JSON to populate model selection dialog](https://codedotorg.atlassian.net/browse/LABS-643)

## Testing story

https://github.com/code-dot-org/code-dot-org/assets/24951544/c65c2004-9ebf-4554-9be4-d4e6352f544d

https://github.com/code-dot-org/code-dot-org/assets/24951544/0872ce7b-6136-4821-938b-b19cdf0907ee

Manual test in UI shows that the correct values are being pulled from the json for the model selection drop down, the compare panel, and the technical details in presentation view. Also that the selected model shows on the left when opening the comparison dialog.

Second video shows that since the selectedModel is part of the AiCustomizations, it gets saved to S3 and persists between page loads.

## Follow-up work

The second part will be adding editing in levelbuilder and testing that it works together https://codedotorg.atlassian.net/browse/LABS-644


thomasoniii (approved):
lgtm 🚀

apps/src/aichat/views/modelCustomization/CompareModelsDialog.tsx
```diff
@@ -1,17 +1,29 @@
...
-const CompareModelsDialog: React.FunctionComponent<{onClose: () => void}> = ({
-  onClose,
-}) => {
-  const [chosenModelLeft, setChosenModelLeft] = useState<string>('llama2');
-  const [chosenModelRight, setChosenModelRight] = useState<string>('mistral');
+const CompareModelsDialog: React.FunctionComponent<{
+  onClose: () => void;
+  availableModels: ModelDescription[];
+}> = ({onClose, availableModels}) => {
+  const selectedModelId = useAppSelector(
```
fisher-alice, line 16: I am noticing that on project template back levels (/allthethings levels 3-5), the selected model is not being updated, even on current level. However, the other `aiCustomization` properties are. Unsure what the discrepancy is.
Tagging @sanchitmalhotra126 who implemented project template backed levels.

apps/src/aichat/views/modelCustomization/CompareModelsDialog.tsx
```diff
@@ -1,17 +1,29 @@
...
+    state =>
+      state.aichat.currentAiCustomizations.selectedModel ??
```
fisher-alice: Just curious about `??` operator which returns the first argument if it is not `null` nor `undefined`.
In `modelCustomization/constants`, the default `selectedModel` is assign `""`. So at some point, could `selectedModel` be assigned `""`?

apps/src/aichat/views/modelCustomization/CompareModelsDialog.tsx
fisher-alice: Looking at my Network tab, I see in `main.json` that the `selectedModel` is updated so that the value is being saved to S3. But the correct value is not being displayed for some reason. This is happening locally for level 1, so wondering if my sources got in a bad state. Curious if this is repro'd by others.

bencodeorg: I believe it's because we initialize based on the settings that are available in levelbuilder, and this property isn't settable yet in levelbuilder, so we ignore whatever is in S3. More details here:

https://github.com/code-dot-org/code-dot-org/blob/5f38bb1035034cd00f678ec54f7b8feb9e6b12d9/apps/src/aichat/redux/aichatRedux.ts#L247-L281

bencodeorg: Cassi says she sees persistence between page loads though in description 🤔

bencodeorg: (I should say, I repro what you're seeing Alice :))

apps/src/aichat/constants.ts
```diff
@@ -16,3 +18,7 @@ const initialChatMessages: ChatCompletionMessage[] = [
 ];
 
 export {initialChatMessages};
+
+const models: ModelDescription[] = modelsJson as ModelDescription[];
+
+export {models};
```
bencodeorg: Nit: just `export const` for both of these?

apps/src/aichat/views/modelCustomization/constants.ts
```diff
@@ -61,11 +63,13 @@ export const DEFAULT_LEVEL_AICHAT_SETTINGS: LevelAichatSettings = {
   initialCustomizations: EMPTY_AI_CUSTOMIZATIONS,
   visibilities: DEFAULT_VISIBILITIES,
   hidePresentationPanel: false,
+  availableModels: [],
 };
 
 export const AI_CUSTOMIZATIONS_LABELS: {
   [key in keyof AiCustomizations]: string;
 } = {
+  selectedModel: 'Selected Model',
```
bencodeorg: Nit: "Selected model" for consistency with other text

apps/src/aichat/views/modelCustomization/SetupCustomization.tsx
```diff
@@ -17,11 +17,12 @@ import {
 } from './constants';
 import {isVisible, isDisabled} from './utils';
 import CompareModelsDialog from './CompareModelsDialog';
+import {models as allModels} from '../../constants';
```
bencodeorg: Nit: could we use a consistent name for this? We name to different names on import in different files (`allModels`, `modelDescriptions`)?

apps/src/aichat/constants.ts
```diff
@@ -3,7 +3,9 @@ import {
   ChatCompletionMessage,
   Role,
   AITutorInteractionStatus as Status,
+  ModelDescription,
 } from './types';
+import modelsJson from '../../static/aichat/modelDescriptions.json';
```
sanchitmalhotra126: minor: you can replace this path with `@cdo/static/...` since we have a mapping for that in our tsconfig

apps/src/aichat/constants.ts
```diff
@@ -16,3 +18,7 @@ const initialChatMessages: ChatCompletionMessage[] = [
 ];
 
 export {initialChatMessages};
+
+const models: ModelDescription[] = modelsJson as ModelDescription[];
```
sanchitmalhotra126: Curious, do you need to cast this actually? TS should be able to pick up the structure from the JSON directly so long as the fields match up

apps/src/aichat/views/modelCustomization/CompareModelsDialog.tsx
sanchitmalhotra126: One idea - when we have the levelbuilder side working, we could also make it so that `selectedModel` is always initialized to first available model if not already set by levelbuilder/student, so we don't need this extra fallback here.

bencodeorg: curious about this too, we have very few usages of this pattern in our code base for some reason

apps/src/aichat/views/modelCustomization/ModelDescriptionPanel.tsx
```diff
@@ -55,18 +32,19 @@ const ModelDescriptionPanel: React.FunctionComponent<{
   const shouldScroll = userWantsScroll && contentNeedsScroll;
 
   const selectedModel =
```
sanchitmalhotra126: Could we just pass in the whole `model` object so we don't need this lookup/fallback logic here?

apps/src/aichat/views/modelCustomization/SetupCustomization.tsx
```diff
@@ -32,6 +33,17 @@ const SetupCustomization: React.FunctionComponent = () => {
...
 
+  /** defaults to all models if not set */
+  const availableModelIds = useAppSelector(
+    state =>
+      (state.lab.levelProperties as AichatLevelProperties | undefined)
+        ?.aichatSettings?.availableModels ?? allModels.map(model => model.id)
+  );
+  const availableModels = allModels.filter(model =>
+    availableModelIds.includes(model.id)
+  );
```
sanchitmalhotra126: nit: if `availableModels` is undefined, it looks like we map `allModels` to its IDs and then just map those back to `allModels`. Would it be better to let `availableModelIds` be undefined if not set on the level, and then default to `allModels` is `availableModelIds` is undefined?

apps/src/aichat/views/modelCustomization/CompareModelsDialog.tsx
fisher-alice: Yeah, it's interesting that on some levels, the selected model is persisting. I noticed that on the level that it was persisting, `levelAichatSettings` is `undefined` so that `reconciledAiCustomizations` is assigned `EMPTY_AI_CUSTOMIZATIONS` which includes `selectedModel` and default `visibility` is set to `editable`.

Once I ran `bundle exec rake seed:custom_levels`, then NONE of the levels persisted as expected based on code above. We've updated the `levelProperties` a lot the past few days!

So yeah, this should not be a problem after Cassie's follow-up with levelbuilder. Thanks!

apps/src/aichat/views/presentation/PresentationView.tsx
```diff
@@ -8,18 +8,20 @@ import styles from '@cdo/apps/aichat/views/model-customization-workspace.module.
...
   const modelCardInfo = currentAiCustomizations.modelCardInfo;
-
-  // These are temporary constants. They will be retrieved from s3.
-  const EXAMPLE_MODEL_NAME = 'Model A';
-  const EXAMPLE_MODEL_TRAINING_DATA = 'Model A Training Data';
+  const selectedModelDescriptions = modelDescriptions.find(
+    model => model.id === selectedModel
+  );
+  const modelName = selectedModelDescriptions?.name ?? '';
+  const trainingData = selectedModelDescriptions?.trainingData ?? '';
```
sanchitmalhotra126: nit: could do `{modelName, trainingData} = modelDescriptions.find(model => model.id === selectedModel)`?

sanchitmalhotra126 (commented):
Nice! Excited to see this come in! Some comments, nothing major

fisher-alice (approved):
Nice work! I left a small question and maybe woud update the 2nd video. Thanks!

apps/src/aichat/views/modelCustomization/ModelDescriptionPanel.tsx
bencodeorg: +1, was thinking the same thing :)

bencodeorg (commented):
Apologies, lots of cooks in this review kitchen :) Just some nits, looks great generally!

apps/src/aichat/constants.ts
cnbrenci: Oh, I do not! Thanks

apps/src/aichat/constants.ts
cnbrenci: so fresh so clean

apps/src/aichat/views/modelCustomization/CompareModelsDialog.tsx
cnbrenci: I was using it because I'm used to using that operator in c# >.> but I think what I actually want is || here. The differnce being that '' counts as falsy with || but not ??

cnbrenci: I'm used to || returning a boolean. Javascript is weird.

cnbrenci: I think changing to || essentially does what Sanchit is suggesting, since if not set by levelbuilder, selectedModel will be '', and availableModels will have been empty and therefore defaulted to all models.

apps/src/aichat/views/modelCustomization/CompareModelsDialog.tsx
cnbrenci: Hmm I didn't check the other levels. I will re-look at this for the part 2 level builder changes.

apps/src/aichat/views/modelCustomization/ModelDescriptionPanel.tsx
cnbrenci: Good idea!

apps/src/aichat/views/modelCustomization/SetupCustomization.tsx
cnbrenci: ohh, yeah. That's better.

apps/src/aichat/views/modelCustomization/SetupCustomization.tsx
cnbrenci: Yeah, really good point. Updating them to be exported as modelDescriptions. And updating the names to always have Id at the end when it's the id to differentiate from the full object.

apps/src/aichat/views/modelCustomization/constants.ts
cnbrenci: oops, good catch.

apps/src/aichat/views/modelCustomization/ModelDescriptionPanel.tsx
cnbrenci: So, the problem with that is the id is a piece of state that we're calling onChange on in the SimpleDropdown and it needs to be a string. I can't think of a good way to make it work without just moving that lookup somewhere else. But I'm also feeling kinda brain mush at the moment so I might be missing something obvious.

apps/src/aichat/views/modelCustomization/CompareModelsDialog.tsx
```diff
@@ -1,17 +1,29 @@
...
+import {ModelDescription} from '../../types';
 
-const CompareModelsDialog: React.FunctionComponent<{onClose: () => void}> = ({
-  onClose,
-}) => {
-  const [chosenModelLeft, setChosenModelLeft] = useState<string>('llama2');
-  const [chosenModelRight, setChosenModelRight] = useState<string>('mistral');
+const CompareModelsDialog: React.FunctionComponent<{
+  onClose: () => void;
+  availableModels: ModelDescription[];
```
cnbrenci, line 14: I think it makes sense to pass the selectedModel in here instead of getting it from redux?

apps/src/aichat/types.ts
```diff
@@ -64,6 +65,13 @@ export interface ModelCardInfo {
   exampleTopics: string[];
 }
 
+export interface ModelDescription {
```
sanchitmalhotra126, line 81: nit: mind adding a comment for what this interface describes?

apps/src/aichat/views/presentation/PresentationView.tsx
cnbrenci: ohhhh fancy!

apps/src/aichat/views/modelCustomization/ModelDescriptionPanel.tsx
sanchitmalhotra126: Hmm, yeah you're right that since the model ID is what is stored in the dropdown, we'll have to do a lookup somewhere. Not to turn this into a bigger refactor but I'm thinking, would it make sense to just pull the state of which model is currently selected into this component? I don't think `CompareModelsDialog` really needs to keep track of which model is currently chosen in each panel, right? It just needs to pass in an initial model ID for each, and then the rest of the state management can happen inside each `ModelDescriptionPanel`. That way all the lookup/fallback logic can also just happen here, instead of in both `CompareModelsDialog` and `ModelDescriptionPanel`. Thinking something like

```
// Helper function
const getModelFromId = (modelId: string): ModelDescription => { ... get model, or default to first if not found };

const ModelDescriptionPanel: React.FunctionComponent<...> = ({initialSelectedModelId: string, dropdownName: string}) => {

  // Store the selected model in state
  const [selectedModel, setSelectedModel] = useState<ModelDescription>(getModelFromId(initialSelectedModelId));

  // Dropdown change handler
  const onDropdownChange = (value: string) => {
    setSelectedModel(getModelFromId(value));
  }

  // Now you always have selectedModel available for other fields (name, training data, etc)
  ... rest of the render function
}
```

apps/src/aichat/types.ts
```diff
@@ -80,4 +88,6 @@ export interface LevelAichatSettings {
   visibilities: {[key in keyof AiCustomizations]: Visibility};
   /** If the presentation panel is hidden from the student. */
   hidePresentationPanel: boolean;
+  /** list of ModelDescription.ids to limit the models available to choose from in the level */
+  availableModels: string[];
```
sanchitmalhotra126: super nit: maybe `availableModelIds` to not confuse with other instances where `availableModels` refers to the full model objects?

apps/src/aichat/views/modelCustomization/SetupCustomization.tsx
fisher-alice: That would be super helpful - thanks!

apps/src/aichat/views/modelCustomization/CompareModelsDialog.tsx
```diff
@@ -1,17 +1,29 @@
...
+    state =>
+      state.aichat.currentAiCustomizations.selectedModel ??
+      availableModels[0].id
+  );
+  const [chosenModelLeft, setChosenModelLeft] =
+    useState<string>(selectedModelId);
+  const [chosenModelRight, setChosenModelRight] = useState<string>(
+    availableModels.find(model => model.id !== selectedModelId)?.id ??
+      selectedModelId
+  );
```
fisher-alice, line 25: Nit: unrelated to your changes, but I think the `<div>` on line 33 below is not necessary.

apps/src/aichat/views/modelCustomization/ModelDescriptionPanel.tsx
fisher-alice: Nice! Would we still need `availableModels` passed down as a prop?
And when a prop is needed at multiple levels of components, just curious when a`context` would be appropriate to use.

cnbrenci: Yeah, that's true. Removing the constraint of having the left and right selected states in `CompareModelsDialog` does make it simpler. Thanks for mapping out the impl! I'll do that refactor. 

Re: Alice
I think we'll still need `availableModels` passed in to populate the dropdown, since we need both id and name for that.
From what I just read about `context` (thanks for mentioning that so I can learn about it :)) Seems like it would be way over-kill for this situation & the use-cases described [here](https://react.dev/learn/passing-data-deeply-with-context#before-you-use-context) are quite broad. But I'm not sure where the line is. I see a bunch of uses of `useContext` in our code base in more specific cases. 🤷🏻

apps/src/aichat/views/modelCustomization/CompareModelsDialog.tsx
cnbrenci: ohhhh. I think the reason why mine was working was because I had deleted main.json for my level, and I didn't have any old version without the new property. I'll have to follow up about `bundle exec rake seed:custom_levels`. Wrote it down & resolving this in this PR to pick up in the next.

cnbrenci: It looks like it does actually work now, since I changed the ??s to ||s

PR #57984, opened 2024-04-12
[AI Tutor] CT-427: Allow students to provide feedback on AI Tutor responses

ebeastlake:
This PR allows students to provide thumbs up/thumbs down feedback on AI Tutor responses. See demo below.

https://github.com/code-dot-org/code-dot-org/assets/26844240/3c5989ee-c558-4812-b3d0-f46e8bbf98d2

After providing feedback, we can inspect the database to see that these flags have been recorded correctly. Note: Once an interaction has been given feedback (thumbs up/down), if a user "changes their mind" and reverts the feedback, a row will persist in the database for that response with both thumbs_up and thumbs_down equal to `null`.

Screenshot 2024-04-17 at 10 18 21 AM

To implement the feedback changes, I modified a couple of things under the hood:
1. Previously, every message had an id that represented its order in the stored messages array for the current session (an array guaranteed to be zero-indexed). Id was a required field on `ChatCompletionMessage`. Now, id is an optional parameter on `ChatCompletionMessage`, and we don’t assign it until after the message has been added to state, we've gotten the chat completion response, and then the response is saved. The id on the `ChatCompletionMessage` matches the id in the interactions database once saved. 
- I split the UserMessage and AssistantMessage into their components (as opposed to using one Message for both UserMessage and AssistantMessage) because I added specific CSS to manage the rendering the feedback buttons alongside the message.

TODO: 

- [x] Add annotations for models/ai_tutor_interaction_feedback. 
- [x] Add links to tickets for follow-up work. 

## Follow-up work

- Move buttons to a hamburger menu with a "copy response" option. 
- Add tests: https://codedotorg.atlassian.net/browse/CT-506
- Get design feedback from Mark: Scheduled for Thursday, no ticket required

## PR Checklist:

- [ ] Code is well-commented


apps/src/aiTutor/chatApi.ts
```diff
@@ -79,24 +78,21 @@ export async function getChatCompletionMessage(
 
   // For now, response will be null if there was an error.
   if (!response) {
-    return {status: Status.ERROR, id: userMessageId};
+    return {status: Status.ERROR};
```
ebeastlake, line 81: Removed ids here because we no longer assign them before we `getChatCompletionMessage`.

apps/src/aiTutor/interactionsApi.ts
```diff
@@ -13,14 +13,19 @@ export async function savePromptAndResponse(
...
         'Content-Type': 'application/json',
         'X-CSRF-Token': await getAuthenticityToken(),
       },
       body: JSON.stringify(interactionData),
     });
+    if (!response.ok) {
+      throw new Error('Network response was not ok');
+    }
+    const data = await response.json();
+    return data;
```
ebeastlake, line 28: We need to return the saved message so we can pull the id off it.

apps/src/aiTutor/redux/aiTutorRedux.ts
```diff
@@ -126,7 +118,8 @@ export const askAITutor = createAsyncThunk(
       aiResponse: chatApiResponse?.assistantResponse,
     };
 
-    await savePromptAndResponse(interactionData);
+    const savedInteraction = await savePromptAndResponse(interactionData);
+    thunkAPI.dispatch(updateLastChatMessage({id: savedInteraction.id}));
```
ebeastlake, line 123: Don't add the id until after the message is saved.

apps/src/aiTutor/redux/aiTutorRedux.ts
```diff
@@ -143,29 +136,32 @@ const aiTutorSlice = createSlice({
...
-        chatMessage.status = status;
+      if (state.chatMessages.length > 0) {
+        const lastMessage = state.chatMessages[state.chatMessages.length - 1];
+        const payloadKeys = Object.keys(action.payload) as Array<
+          keyof ChatCompletionMessage
+        >;
+
+        payloadKeys.forEach(key => {
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
+          (lastMessage as any)[key] = action.payload[key];
```
ebeastlake: I tried to make this a method that would update the most recent method with any keys that were contained in the payload, but TypeScript was not happy 😅 Opinions @molly-moen? I could split this out into two methods to update the status and/or id.

apps/src/aiTutor/views/UserMessage.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+const statusToStyleMap = {
+  [Status.OK]: style.userMessage,
+  [Status.UNKNOWN]: style.userMessage,
+  [Status.PROFANITY_VIOLATION]: style.inappropriateMessage,
+  [Status.PII_VIOLATION]: style.tooPersonalMessage,
+  [Status.ERROR]: style.errorMessage,
+};
+
+const getMessageText = (status: string, chatMessageText: string) => {
```
ebeastlake, line 28: I added `statusToStyleMap` and `getMessageText` helpers, but nothing else has changed. One of my next tasks is ensuring flagging the user messages appropriately works, so this will get tested again.

apps/src/templates/Button.jsx
```diff
@@ -206,7 +206,9 @@ class Button extends React.Component {
               <FontAwesome icon="spinner" className="fa-spin" />
             </span>
           )}
-          <span className={moduleStyles.textSpan}>{!isPending && text}</span>
+          {text && (
```
ebeastlake, line 209: @levadadenys Buttons without text are very slightly off-center without this change. Is this safe to make?

Before:
Screenshot 2024-04-17 at 10 40 59 AM

After:
Screenshot 2024-04-17 at 10 40 28 AM

dashboard/app/controllers/ai_tutor_interactions_controller.rb
```diff
@@ -9,7 +9,7 @@ def create
     return render(status: :not_acceptable, json: {error: 'Status is unacceptable.'}) unless valid_status
     @ai_tutor_interaction = AiTutorInteraction.new(ai_tutor_interaction_params)
     if @ai_tutor_interaction.save
-      render json: {message: "successfully created AiTutorInteraction with id: #{@ai_tutor_interaction.id}"}, status: :created
+      render json: {message: "successfully created AiTutorInteraction", id: @ai_tutor_interaction.id}, status: :created
```
ebeastlake, line 12: I moved the ID out of the message and into a dedicated field so it's easier to pull off the response.

apps/src/aiTutor/redux/aiTutorRedux.ts
molly-moen: what was the error you were getting?

ebeastlake:
```
Type 'string | number | undefined' is not assignable to type 'never'.
Type 'undefined' is not assignable to type 'never'.
```

My guess is that TypeScript can't be sure that the type of `action.payload[key]` matches the type of `lastMessage[key]` because it's not making type inference when doing dynamic assignments like this.

ebeastlake: The `(lastMessage as any)` fixed it, but I assume we want to avoid that.

molly-moen: gotcha yeah it can't track the key being part of the `ChatCompletionMessage` function. It's not ideal to do this because we want to avoid `any`, but it might be ok given we know the keys should be accurate. However, would something like this work?
```
state.chatMessages[state.chatMessages.length - 1] = {...lastMessage, ...action.payload}
```
(Alternatively you could pop/push `lastMessage` to avoid the wordiness with setting the value)

apps/src/aiTutor/interactionsApi.ts
```diff
@@ -13,14 +13,19 @@ export async function savePromptAndResponse(
...
+    const response = await fetch('/ai_tutor_interactions', {
       method: 'POST',
       headers: {
         'Content-Type': 'application/json',
         'X-CSRF-Token': await getAuthenticityToken(),
       },
       body: JSON.stringify(interactionData),
     });
+    if (!response.ok) {
+      throw new Error('Network response was not ok');
```
molly-moen, line 25: Is there a reason we throw an error here but if there's an error retrieving the feedbacks we log to the metrics reporter?

apps/src/aiTutor/views/AssistantMessage.tsx
```diff
@@ -0,0 +1,87 @@
...
+  const handleFeedbackSubmission = async (
+    thumbsUp: boolean,
+    messageId?: number
+  ) => {
+    if (!messageId) {
+      return;
+    }
+
+    const feedbackData = {
+      thumbsUp: thumbsUp ? (feedbackState.thumbsUp ? null : true) : null,
```
molly-moen, line 32: why is it if `feedbackState.thumbsUp` is true we set `thumbsUp` to `null`?

dashboard/app/models/ai_tutor_interaction_feedback.rb
```diff
@@ -0,0 +1,23 @@
...
+#
+# Indexes
+#
+#  index_ai_tutor_feedback_on_interaction_and_user  (ai_tutor_interaction_id,user_id) UNIQUE
+#
+class AiTutorInteractionFeedback < ApplicationRecord
+  belongs_to :ai_tutor_interaction
+  belongs_to :user
+
+  validates :thumbs_up, inclusion: {in: [true, false]}, allow_nil: true
```
molly-moen: for my understanding: why do we need this given that `thumbs_up` and `thumbs_down` are nullable booleans?

molly-moen (commented):
Looks good, a couple clarifying questions!

apps/src/templates/Button.jsx
levadadenys: Oh, this is not a Design system button.
To use DesignSystem Button you need to use this one https://github.com/code-dot-org/code-dot-org/blob/2b4ae36a680597fb148bdb4085f65b766b1e4884/apps/src/componentLibrary/button/Button.tsx

Also, for iconOnly button you'll need to pass `isIconOnly={true}` prop in it

levadadenys (commented):
Great improvement overall!
Though, old Button is used here, to use Deisgn System `Button` use this one: https://github.com/code-dot-org/code-dot-org/tree/staging/apps/src/componentLibrary/button

apps/src/aiTutor/interactionsApi.ts
ebeastlake: Ah, good catch. So here are the patterns we currently have...

```
export async function savePromptAndResponse() {
  try {
    const response = await fetch('/ai_tutor_interactions', {...});
    if (!response.ok) {
      throw new Error('...');
    }
    const data = await response.json();
    return data;
  } catch (error) {
    MetricsReporter.logError({...});
  }
}
```

```
export const fetchAITutorInteractions = async () => {
  ...
  try {
    const response = await fetch(url, {...});
    if (!response.ok) {
      throw new Error('...');
    }
    const data = await response.json();
    return data;
  } catch (error) {
    MetricsReporter.logError({...});
    return null;
  }
};
```

```
export async function saveFeedback(
  aiTutorInteractionId: number,
  feedbackData: FeedbackData
) {
  try {
    await fetch(`/ai_tutor_interactions/${aiTutorInteractionId}/feedbacks`, {...});
  } catch (error) {
    MetricsReporter.logError({...});
  }
}
```

So actually, even though `savePromptAndResponse` is reacting to HTTP status codes in addition to errors thrown by `fetch`, all of the functions here end execution within the catch block after logging the error. I'm trying to decide if that's okay. (It doesn't seem ideal, but I don't know what the UI should be if the user is successfully having a conversation and we're not providing feedback.)  I should probably file a follow-up task to ensure all this error handling is what we want.

apps/src/aiTutor/views/AssistantMessage.tsx
ebeastlake: If we're clicking thumbs up again (`thumbsUp` === `true`) and the existing state (`feedbackState.thumbsUp`) is true, then we want to nullify the provided feedback value to "toggle" it to its original state.

ebeastlake: The truth table should be:

| thumbsUp (clicked thumbs up) | feedbackState.thumbsUp (already clicked?) | Result (new selected value)                         |
|----------|------------------------|---------------------------------|
| F        | F                      | null                            |
| F        | T                      | null                            |
| T        | F                      | true                            |
| T        | T                      | null                            |

molly-moen: Ah I see thanks! Can you add a comment that this is a toggle value?

dashboard/app/models/ai_tutor_interaction_feedback.rb
ebeastlake: Since `thumbs_up` and `thumbs_down` are nullable booleans in the schema, you mean? I think the answer is we don't need these, and I copied conventions from examples that are a little stricter than our codebase. They're mainly redundant (though they do prevent someone from having to look at the schema to know what these should be). Would you like to remove? I'm good either way 🤷‍♀️

apps/src/aiTutor/redux/aiTutorRedux.ts
ebeastlake: That was a good call, thanks! I'll remember to merge/spread like this in the future to avoid weirdness with dynamic assignment.

apps/src/templates/Button.jsx
ebeastlake: Thank you, Denys! I updated it to use the Design System button and changed the colors to black. Mark plans to suggest tweaks and these will probably undergo more iteration. 

Screenshot 2024-04-18 at 4 53 22 PM

dashboard/app/models/ai_tutor_interaction_feedback.rb
molly-moen: yeah I think we don't need it? It makes sense to me to remove but I'm not a ruby expert.

apps/src/aiTutor/interactionsApi.ts
molly-moen: yeah totally fine to handle as a follow up

molly-moen (approved):
LGTM!

PR #57990, opened 2024-04-12
[ProgressV2] Create a new feature feedback endpoint for progress feedback

lfryemason:
## Warning!!

Based on this PR to add LTI feedback: https://github.com/code-dot-org/code-dot-org/pull/57517/files

This adds a simple create and show feedback endpoint for a thumbs up or thumbs down feedback form. This feedback is specifically intended for new features and will be used for the new progress view feedback form. 

This feedback is reusable because it uses a `form_key` identifier to determine which feedback form the user has answered. For the progress implementation, this will be `progress_v2` which is currently the only allowed value.

Adds a migration to create the table and API endpoints. These endpoints will be called in a future PR.

## Links
[TEACH-1012](https://codedotorg.atlassian.net/browse/TEACH-1012)

## Testing story
Unit tests for controller

## Follow-up work
Creating the FE component that will call the added endpoints.

## Security
Authenticating users calling the feedback form are teachers.

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/dashboard/app/controllers/new_feature_feedback_controller.rb
+++ b/dashboard/app/controllers/new_feature_feedback_controller.rb
@@ -1,5 +1,3 @@
-# frozen_string_literal: true
-
 class NewFeatureFeedbackController < ApplicationController
   before_action :authenticate_user!
 
--- a/dashboard/app/controllers/new_feature_feedback_controller.rb
+++ b/dashboard/app/controllers/new_feature_feedback_controller.rb
@@ -1,0 +1,32 @@
+# frozen_string_literal: true
+
+class NewFeatureFeedbackController < ApplicationController
+  before_action :authenticate_user!
+
+  def create
+    @feedback = NewFeatureFeedback.new(feedback_params)
+
+    @feedback.user = current_user
+    @feedback.locale = I18n.locale
+
+    respond_to do |format|
+      if @feedback.save
+        format.json {render json: @feedback, status: :created}
+      else
+        format.json {render json: @feedback.errors.full_messages, status: :unprocessable_entity}
+      end
+    end
+  end
+
+  def show
+    @feedback = NewFeatureFeedback.find_by(user: current_user, form_key: params.require(:form_key))
+
+    respond_to do |format|
+      format.json {render json: @feedback}
+    end
+  end
+
+  private def feedback_params
+    params.require(:feedback).permit(:satisfied, :form_key)
+  end
+end
--- a/dashboard/app/controllers/new_feature_feedback_controller.rb
+++ b/dashboard/app/controllers/new_feature_feedback_controller.rb
@@ -5,7 +5,6 @@ class NewFeatureFeedbackController < ApplicationController
     @feedback = NewFeatureFeedback.new(feedback_params)
 
     @feedback.user = current_user
-    @feedback.locale = I18n.locale
 
     respond_to do |format|
       if @feedback.save
--- a/dashboard/app/models/new_feature_feedback.rb
+++ b/dashboard/app/models/new_feature_feedback.rb
@@ -1,0 +1,26 @@
+# == Schema Information
+#
+# Table name: new_feature_feedbacks
+#
+#  id         :bigint           not null, primary key
+#  user_id    :integer          not null
+#  form_key   :string(255)      not null
+#  satisfied  :boolean          not null
+#  locale     :string(255)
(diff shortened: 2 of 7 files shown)


dashboard/app/controllers/new_feature_feedback_controller.rb
```diff
@@ -0,0 +1,30 @@
+class NewFeatureFeedbackController < ApplicationController
+  before_action :authenticate_user!
+
+  def create
+    @feedback = NewFeatureFeedback.new(feedback_params)
+
+    @feedback.user = current_user
+    @feedback.locale = I18n.locale
```
etaderhold: This feels like it might be duplicative. Don't we store the user's locale preference somewhere else already, and couldn't we just look it up there instead of storing it again?

dashboard/app/models/new_feature_feedback.rb
```diff
@@ -0,0 +1,26 @@
+# == Schema Information
+#
+# Table name: new_feature_feedbacks
+#
+#  id         :bigint           not null, primary key
+#  user_id    :integer          not null
+#  form_key   :string(255)      not null
```
etaderhold: You could use an integer enum here instead of a string to keep the storage size down. See `SectionInstructor.status` for an example.

dashboard/app/models/new_feature_feedback.rb
```diff
@@ -0,0 +1,26 @@
...
+#
+#  index_new_feature_feedbacks_on_satisfied  (satisfied)
+#  index_new_feature_feedbacks_on_user_id    (user_id) UNIQUE
+#
+# A simple thumbs up/down feedback form.
+# Intended for use by a front-end banner for new feature feedback.
+class NewFeatureFeedback < ApplicationRecord
+  belongs_to :user
+
+  validates :user_id, uniqueness: true
```
etaderhold: Is the intent to allow the same user to leave feedback on multiple features over time? This validation line will stop them after their first feedback. Validating on uniqueness of the (`user_id`, `form_key`) pair might be a better way to go. The unique index on the database table would also need to be modified in this case.

dashboard/app/controllers/new_feature_feedback_controller.rb
lfryemason: Looks like it, we already have a :locale variable on the user object.

dashboard/test/controllers/new_feature_feedback_controller_test.rb
```diff
@@ -0,0 +1,66 @@
...
+    assert_response :success
+    response = JSON.parse(@response.body)
+    assert_equal feedback.id, response['id']
+    assert_equal @user.id, response['user_id']
+  end
+
+  test 'show - returns nothing when no Feedback exists for the current user' do
+    get :show, params: {form_key: 'progress_v2'}, format: :json
+
+    assert_response :ok
```
etaderhold, line 83: Could you check for an empty response body here?

dashboard/db/migrate/20240412134643_create_new_feature_feedbacks.rb
```diff
@@ -0,0 +1,14 @@
...
+  def change
+    create_table :new_feature_feedbacks do |t|
+      t.belongs_to :user, type: :integer, null: false, foreign_key: true
+
+      t.integer :form_key, null: false
+      t.boolean :satisfied, null: false, index: true
+
+      t.index [:user_id, :form_key], unique: true
+
+      t.datetime :created_at, null: false
```
etaderhold: This is a more standard way of creating `created_at` and `updated_at` columns.

```suggestion
      t.timestamps
```

dashboard/test/controllers/new_feature_feedback_controller_test.rb
```diff
@@ -0,0 +1,66 @@
...
+
+  test 'create - returns validation errors when satisfied param is invalid' do
+    assert_no_difference('NewFeatureFeedback.count') do
+      post :create, params: {feedback: {satisfied: nil, form_key: 'progress_v2'}}, format: :json
+    end
+
+    assert_response :unprocessable_entity
+    assert_equal '["Satisfied is not included in the list"]', response.body
+  end
+
```
etaderhold, line 56: A test to ensure users can't leave feedback on the same thing twice would be nice.

etaderhold (approved):

PR #57992, opened 2024-04-12
fix(lms): allow list the lti iframe buster page from frame policies

stephenliang:
LMS providers iframe sandbox apps such as Code.org. Due to Code.org's anti-clickjacking CSP, Code.org cannot be executed within an iframe. To get around this, an iframe buster was created that informs the user to click a button to launch Code.org in a new window. 

However, the frame content security policies still applied to the iframe buster path. The initial implementation allowlisted schoology specifically, however schoology allows for districts to have their own custom domains resulting in an unbounded allowlist.

Instead of applying frame CSPs to the iframe buster path, allow that specific page to load in an iframe. The frame buster page does not render the normal dashboard/rails and only a minimal page to have a button to launch Code.org. 

Another security consideration is that you can clickjack the iframe buster page. However, this route, `/lti/v1/iframe` is only accessible from a LMS logged in user path which implies that the user is originating from one of our onboarded LMS platforms. The iframe buster page is also of minimal content (just the launch Code.org button) and the user would not be entering keystrokes on this page.

## Testing story

1. Validate that the Code.org iframe buster page can be rendered on Schoology
2. Validate that the CSP headers continue to be returned for other paths

## PR Checklist:

- [x] Code is well-commented

### DIFF
--- a/dashboard/app/controllers/lti_v1_controller.rb
+++ b/dashboard/app/controllers/lti_v1_controller.rb
@@ -128,9 +128,9 @@ class LtiV1Controller < ApplicationController
       auth_url_base = CDO.studio_url('/lti/v1/authenticate', CDO.default_scheme)
 
       query_params = {
-        id_token: ERB::Util.url_encode(params[:id_token]),
-        state: ERB::Util.url_encode(params[:state]),
-        new_tab: ERB::Util.url_encode("true"),
+        id_token: params[:id_token],
+        state: params[:state],
+        new_tab: "true",
       }
 
       @auth_url = "#{auth_url_base}?#{query_params.to_query}"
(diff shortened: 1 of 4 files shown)


dashboard/app/controllers/lti_v1_controller.rb
```diff
@@ -125,11 +125,16 @@ def authenticate
...
-      state = params[:state]
-      token_hash = {id_token: id_token, state: state}
-      redirect_to "#{lti_v1_iframe_url}?#{token_hash.to_query}"
-      return
+      auth_url_base = CDO.studio_url('/lti/v1/authenticate', CDO.default_scheme)
+
+      query_params = {
+        id_token: ERB::Util.url_encode(params[:id_token]),
+        state: ERB::Util.url_encode(params[:state]),
+        new_tab: ERB::Util.url_encode("true"),
```
unlox775-code-dot-org: manually URL encoding each parameter seems unnecessary. Is the intent here to do double URL encoding?

it seems HASH.to_query does escaping, and would protect against the XSS we originally had here

cat5inthecradle (approved):

dashboard/app/controllers/lti_v1_controller.rb
stephenliang: I originally went with the `ERB::Util.encode` and then later added `to_query`. Confirmed that to_query already encodes so the `ERB::Util.url_encode` encode isn't necessary and so removed it.

PR #57998, opened 2024-04-12
P20-870: Auto-verify LTI teachers on sign-up

artem-vavilov:
## After Deploy Action
We need to run this query on prod to verify already created LTI teachers:
```ruby
lti_user_ids = AuthenticationOption.where(credential_type: AuthenticationOption::LTI_V1).select(:user_id)

User.where(id: lti_user_ids, user_type: User::TYPE_TEACHER).find_each do |lti_teacher|
  lti_teacher.verify_teacher!
end
```

## Links
- JIRA ticket: [P20-870](https://codedotorg.atlassian.net/browse/P20-870)

### DIFF
--- a/dashboard/app/models/user.rb
+++ b/dashboard/app/models/user.rb
@@ -285,6 +285,8 @@ class User < ApplicationRecord
     Services::Lti.create_lti_user_identity(self)
   end
 
+  after_create :verify_teacher!, if: -> {teacher? && Policies::Lti.lti?(self)}
+
   before_destroy :soft_delete_channels
 
   before_validation on: :create, if: -> {gender.present?} do
@@ -1490,6 +1492,10 @@ class User < ApplicationRecord
 
   def teacher?
     user_type == TYPE_TEACHER
+  end
+
+  def verify_teacher!
+    self.permission = UserPermission::AUTHORIZED_TEACHER
   end
 
   # This method just checks if a user has the authorized teacher permission
--- a/dashboard/test/models/user_test.rb
+++ b/dashboard/test/models/user_test.rb
@@ -781,7 +781,7 @@ class UserTest < ActiveSupport::TestCase
     assert lti_teacher.verified_teacher?
   end
 
-  test 'LTI student should be verified after creation' do
+  test 'LTI student should not be verified after creation' do
     lti_integration = create(:lti_integration)
     auth_id = "#{lti_integration[:issuer]}|#{lti_integration[:client_id]}|#{SecureRandom.alphanumeric}"
 
--- a/dashboard/test/models/user_test.rb
+++ b/dashboard/test/models/user_test.rb
@@ -770,6 +770,28 @@ class UserTest < ActiveSupport::TestCase
     assert_empty user.lti_user_identities
   end
 
+  test 'LTI teacher should be verified after creation' do
+    lti_integration = create(:lti_integration)
+    auth_id = "#{lti_integration[:issuer]}|#{lti_integration[:client_id]}|#{SecureRandom.alphanumeric}"
+
+    lti_teacher = build(:teacher)
+    lti_teacher.authentication_options << build(:lti_authentication_option, user: lti_teacher, authentication_id: auth_id)
+    lti_teacher.save!
+
+    assert lti_teacher.verified_teacher?
+  end
+
+  test 'LTI student should be verified after creation' do
+    lti_integration = create(:lti_integration)
+    auth_id = "#{lti_integration[:issuer]}|#{lti_integration[:client_id]}|#{SecureRandom.alphanumeric}"
+
+    lti_student = build(:student)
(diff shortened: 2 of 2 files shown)


dashboard/test/empty_test.rb
```diff
@@ -0,0 +1,19 @@
+$start_time = Time.now
```
mgc1194: Why the test is called empty_test? What is this test testing?

dashboard/app/models/user.rb
```diff
@@ -285,6 +285,8 @@ class User < ApplicationRecord
     Services::Lti.create_lti_user_identity(self)
   end
 
+  after_create :verify_teacher!, if: -> {teacher? && Policies::Lti.lti?(self)}
+
```
mgc1194, line 289: LGTM

dashboard/test/empty_test.rb
artem-vavilov: It's from another task 😅 removed

mgc1194 (approved):

carl-codeorg (approved):
LGTM. Out of curiosity, how were teachers getting verified prior to this? If `verify_teacher!` had to be added here, presumably it didn't previously exist.

artem-vavilov:
> LGTM. Out of curiosity, how were teachers getting verified prior to this? If `verify_teacher!` had to be added here, presumably it didn't previously exist.

I believe we usually set permissions from the Admin panel: https://github.com/code-dot-org/code-dot-org/blob/05da6827e8825fadd090962fa4fb2cdefc50540c/dashboard/app/controllers/admin_users_controller.rb#L279-L289

dashboard/app/models/user.rb
```diff
@@ -285,6 +285,8 @@ class User < ApplicationRecord
     Services::Lti.create_lti_user_identity(self)
   end
 
+  after_create :verify_teacher!, if: -> {teacher? && Policies::Lti.lti?(self)}
```
daynew, line 288: should this be `before_save` for the times when a teacher's Code.org LTI account was accidentally created as a student account and later updated to a teacher account when we do another roster sync?

If it should be done only on create, should this be `before_create` so the changes are automatically saved? Right now I'm suspicious this doesn't actually save the changes to the DB because the modification is happening after the create has happened.

Another thought, does this also need to handle roster syncs where a teacher becomes a student? I can't remember if that is something we support. We should probably unverify the account if they become a student.

dashboard/test/models/user_test.rb
```diff
@@ -770,6 +770,28 @@ def assert_fails_email_uniqueness_validation(user)
...
   end
 
+  test 'LTI teacher should be verified after creation' do
+    lti_integration = create(:lti_integration)
+    auth_id = "#{lti_integration[:issuer]}|#{lti_integration[:client_id]}|#{SecureRandom.alphanumeric}"
+
+    lti_teacher = build(:teacher)
+    lti_teacher.authentication_options << build(:lti_authentication_option, user: lti_teacher, authentication_id: auth_id)
+    lti_teacher.save!
+
```
daynew, line 780: should `lti_teacher.reload` be called to verify the model changes were saved to the DB?

daynew:
```
lti_user_ids = AuthenticationOption.where(credential_type: AuthenticationOption::LTI_V1).select(:user_id)

User.where(id: lti_user_ids, user_type: User::TYPE_TEACHER).find_each do |lti_teacher|
  lti_teacher.verify_teacher!
end
```
does `lti_teacher.save!` need to be called?

dashboard/app/models/user.rb
artem-vavilov: The user must be persistent to be granted permission:
https://github.com/code-dot-org/code-dot-org/blob/9cee0fc395242a7c0f26363bc61eaa3e31cac80e/dashboard/app/models/concerns/user_permission_grantee.rb#L28-L32

artem-vavilov:
> ```
> lti_user_ids = AuthenticationOption.where(credential_type: AuthenticationOption::LTI_V1).select(:user_id)
> 
> User.where(id: lti_user_ids, user_type: User::TYPE_TEACHER).find_each do |lti_teacher|
>   lti_teacher.verify_teacher!
> end
> ```
> 
> does `lti_teacher.save!` need to be called?

When an `has_many` association is assigned like this `permissions <<`, the user record will be updated by default.
https://github.com/code-dot-org/code-dot-org/blob/9cee0fc395242a7c0f26363bc61eaa3e31cac80e/dashboard/app/models/concerns/user_permission_grantee.rb#L31

dashboard/test/models/user_test.rb
artem-vavilov: It's not strictly necessary, but it would be better to play it safe

daynew:
```
When an has_many association is assigned like this permissions <<, the user record will be updated by default.
```
I didn't know that, thanks!

daynew (approved):

PR #57999, opened 2024-04-12
fix(lms): do not create a duplicative authentication option for teachers

stephenliang:
Previously during LMS new user initialization for teachers via NRPS, the email address of the teacher was added using `User.update_primary_contact_info`. This function will look up the user and check for an existing auth option, for which none will be found for a new user. As a fallback, this function will create a new auth option using credential type `EMAIL` and return.

Subsequently, the `initialize_lti_user_from_nrps` function will add the LTI authentication option with the same email. This results in a two authentication options with the same email which fails the `AuthenticationOption` `email_must_be_unique` validation causing a 422 error to be thrown back to the user.

In the SSO user creation flow, `PartialRegistration` updates the primary contact info to the first authentication option, causing the subsequent call to `update_primary_contact_info` to select the LTI auth option.

This PR updates the logic to set the primary contact info to the LTI authentication option directly bypassing the second auth option that was created.

## Testing story

1. Added test to emulate duplicate user condition
2. Reproduced issue on Canvas and was able to sync successfully with teacher created

## PR Checklist:

- [x] Code is well-commented

### DIFF
--- a/dashboard/lib/services/lti.rb
+++ b/dashboard/lib/services/lti.rb
@@ -105,7 +105,6 @@ module Services
 
       if account_type == ::User::TYPE_TEACHER && email_address.present?
         user.user_type = ::User::TYPE_TEACHER
-        user.update_primary_contact_info(new_email: email_address, new_hashed_email: ::User.hash_email(email_address))
       else
         user.user_type = ::User::TYPE_STUDENT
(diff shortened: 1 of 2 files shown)


carl-codeorg (approved):
LGTM!

PR #53642, opened 2023-09-06
Set `EnforcedStyle: indented` for Two `Layout/Multiline*Indentation` Rules

Hamms:
Specifically:

`Layout/MultilineMethodCallIndentation`:
> Checks the indentation of the method name part in method calls that span more than one line.

`Layout/MultilineOperationIndentation`:
> Checks the indentation of the right hand side operand in binary operations that span more than one line.

Both rules largely work the same, but apply in different circumstances; handling them at the same time helps keep our codebase consistent.

In both cases, we have the option to either use the `indented` style that this PR proposes or the default style, but I think `indented` is a lot cleaner. See [this PR](https://github.com/code-dot-org/code-dot-org/pull/53644) for an example of what applying the default style rules to our codebase would look like, and more of my thoughts on why it's not an ideal option. 

We also have the option of intentionally disabling this rule and allowing developers to continue to apply whatever indentation style they'd prefer.

What does everyone think?

## Links

- https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Layout/MultilineMethodCallIndentation
- https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Layout/MultilineOperationIndentation
- https://docs.rubocop.org/rubocop/cops_layout.html#layoutmultilinemethodcallindentation
- https://docs.rubocop.org/rubocop/cops_layout.html#layoutmultilineoperationindentation

PR #53620, opened 2023-09-05
Fix Violations of and Reenable `Layout/SpaceAroundEqualsInParameterDefault`

Hamms:
> Checks that the equals signs in parameter default assignments have or don’t have surrounding space depending on configuration.

Fix applied automatically with `bundle exec rubocop --autocorrect-all --only Layout/SpaceAroundEqualsInParameterDefault`

I don't have a strong preference here, and don't personally find the more-space-consuming option more readable than the compact option. But in this case the official Ruby style guide does explicitly recommend using spaces around the `=` operator in method parameters, so I'm inclined to go with what they say.

## Links

- https://rubystyle.guide/#spaces-around-equals
- https://docs.rubocop.org/rubocop/cops_layout.html#layoutspacearoundequalsinparameterdefault
- https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Layout/SpaceAroundEqualsInParameterDefault

PR #58006, opened 2024-04-13
Fix Violation of and Reenable `Layout/SpaceAroundEqualsInParameterDefault`

Hamms:
> Checks that the equals signs in parameter default assignments have or don’t have surrounding space depending on configuration.

This is a pretty simple one; we only have a single violation in the whole codebase.

Also, regenerate `.rubocop_todo.yml`; it's been a while since it was last generated, and it was starting to get a bit out of date.

## Links

- https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Layout/SpaceAroundEqualsInParameterDefault
- https://docs.rubocop.org/rubocop/cops_layout.html#layoutspacearoundequalsinparameterdefault

## Testing story

Relying on existing tests to verify that this layout change doesn't result in any change to actual functionality.

## PR Checklist:

- [ ] Code is well-commented


.rubocop_todo.yml
```diff
@@ -1,68 +1,61 @@
...
-# Offense count: 111
-# This cop supports safe auto-correction (--auto-correct).
+# Offense count: 110
+# This cop supports safe autocorrection (--autocorrect).
 # Configuration parameters: EnforcedStyle, IndentationWidth.
 # SupportedStyles: aligned, indented
 Layout/LineEndStringConcatenationIndentation:
   Enabled: false
 
-# Offense count: 292
```
bethanyaconnor, line 37: Why does this claim there were nearly 300 offenses when you only found 1?

Hamms: This is why I decided to regenerate the todo file! I haven't verified this, but my assumption is that those violations got removed as a result of other changes made sometime since January 2023 when the file was last regenerated.

In short: because this file was outdated.

bethanyaconnor (approved):

.rubocop_todo.yml
bethanyaconnor: It looks like you tried to reenable this [last year](https://github.com/code-dot-org/code-dot-org/pull/53620) -- any ideas how it reverted? (Obviously not a blocker for this PR -- just curious)

Hamms: Lol, nice! I guess we found where the other changes got made.

Looks like the rule got accidentally redisabled as part of [an unrelated Rubocop fix](https://github.com/code-dot-org/code-dot-org/pull/53642/files#diff-29beaabe278fd17493296745cecb67919f0906b47b8246ab770f5517615d9ef7) I did about a week later; I'd guess I ran into a merge conflict, and resolved it incorrectly.

PR #49395, opened 2022-12-06
P20-12 i18n docummentation content

mgc1194:
**What:** This PR adds documentation of applab, gamelab and webal to the sync. 
**Why:** Part of the documentation was already made translatable in a [previous PR](https://github.com/code-dot-org/code-dot-org/pull/47511) but the actual content regarding programming expressions was not translatable yet.
**How:** Each programming environment is sync-in, from the database to individual files located in `i18n/source/docs`. This maintains modularity of the programming environments and allows different redaction/restoration approaches depending on the environment. 
During the sync out, redacted files are restored and merged into a single file `programming_environments.{locale}.json` that is distributed for each locale.
In the model `programming_expressions.rb` the recurrent scope needed to localize the content is served through `expression_scope` and a function `get_localized_property` optimizes the calls to the `I18n` function.

Both `sync-in` and `programming_expressions` model incorporate nested code according to the structure of the database.

**Note:** This PR substitutes [PR #47868](https://github.com/code-dot-org/code-dot-org/pull/47868)

## Links

- jira ticket: [P20-12](https://codedotorg.atlassian.net/browse/P20-12)/[FND-2060](https://codedotorg.atlassian.net/browse/FND-2060)
- tested links: `http://localhost-studio.code.org:3000/docs/ide/applab/expressions/button`

## Testing story

Sync ran successfully. Screenshots include examples of newly added content.

Screenshot 2022-12-06 at 23 34 56

For languages with no available translation, the webpage still renders english content.

Screenshot 2022-12-06 at 23 35 56

## Deployment strategy

Merge with staging after HoC restricted deployment.

## Follow-up work

- Javalab and Spritelab need to be added to the sync.

## PR Checklist:

- [ ] Tests provide adequate coverage
- [ ] Privacy and Security impacts have been assessed
- [ ] Code is well-commented
- [ ] New features are translatable or updates will not break translations
- [ ] Relevant documentation has been added or updated
- [ ] User impact is well-understood and desirable
- [ ] Pull Request is labeled appropriately
- [ ] Follow-up work items (including potential tech debt) are tracked and linked

PR #58008, opened 2024-04-13
Return Nil if there is no translation for palette_ params

mgc1194:
In PR #49395 I added a method to localize the parameters in programming expressions.

In that implementation, I assumed the default value returned when the I18n API did not find any localized content, was the same as the original content (an Array of Hashes).

However, after a [regression in the content](https://codedotorg.slack.com/archives/C0T0PNTM3/p1712964952362009?thread_ts=1712934698.712219&cid=C0T0PNTM3), we analyzed more in-depth what the `I18n.t` method was returning and found out only the first element of the array was return when defaulting.

```
palette_params: [{"name"=>"[id]", "type"=>"string", "required"=>true, "description"=>"The unique identifier for the canvas screen element. The id is used for referencing the canvas functions or other UI element modification functions. Must begin with a letter, contain no spaces, and may contain letters, digits, - and _."}, {"name"=>"[width]", "type"=>"number", "description"=>"The horizontal width in pixels of the canvas. If not specified the app window width is used."}, {"name"=>"[height]", "type"=>"number", "description"=>"The vertical height in pixels of the canvas. If not specified the app window height is used."}]

i18n_params: {"name"=>"[id]", "type"=>"string", "required"=>true, "description"=>"The unique identifier for the canvas screen element. The id is used for referencing the canvas functions or other UI element modification functions. Must begin with a letter, contain no spaces, and may contain letters, digits, - and _."}
```

Instead, I decided to return `Nil Value` by default (when no localized content is found) and then, when the localized content is not `Nil`, merge the localized parameters with the non-localized ones.

Localized params look like this: `{:"[height]"=>{:type=>"número"}, :"[width]"=>{:type=>"número"}}`

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/dashboard/app/models/programming_expression.rb
+++ b/dashboard/app/models/programming_expression.rb
@@ -305,10 +305,10 @@ class ProgrammingExpression < ApplicationRecord
     i18n_params = I18n.t(
       :palette_params,
       scope: expression_scope,
-      default: palette_params,
+      default: nil,
       smart: true
     )
-    if i18n_params != localized_params
+    unless i18n_params.nil?
       localized_params&.each do |param|
         param_key = param['name'].to_sym
         unless i18n_params[param_key].nil?


dashboard/app/models/programming_expression.rb
```diff
@@ -305,10 +305,10 @@ def get_localized_params
     i18n_params = I18n.t(
       :palette_params,
       scope: expression_scope,
-      default: palette_params,
+      default: nil,
       smart: true
     )
-    if i18n_params != localized_params
+    unless i18n_params.nil?
```
wilkie, line 311: Exactly right.

dashboard/app/models/programming_expression.rb
```diff
@@ -305,10 +305,10 @@ def get_localized_params
...
       localized_params&.each do |param|
         param_key = param['name'].to_sym
```
wilkie, line 313: You would still need a `next unless param.key?('name')` here to fix the honeybadger error in the thread... I guess... since we are here already.

wilkie (approved):

dashboard/app/models/programming_expression.rb
wilkie: An alternative would be to have `default: {}` and then remove the condition altogether.

dashboard/app/models/programming_expression.rb
mgc1194: My theory is that if there is no name param, the string won't be picked up in the sync, and there won't ever be a translation available, returning nil.

wilkie:
This is a good change for the sake of keeping the code logical. Technically, it does not matter too much. The default would always yield some kind of dict that differs from localized_params and then it would just kind of fail to do anything with it leading to it to return the right thing anyway.

But that's obviously very confusing... but nothing that's causing a user-facing issue.

dashboard/app/models/programming_expression.rb
mgc1194: I'd rather not process the content if translations aren't available. We would only iterate over the parameters when there's a translation, avoiding errors such as the one found in honeybadger.

dashboard/app/models/programming_expression.rb
wilkie: it's still going to render incorrectly... they need to fix the content, imo. we should not mitigate it by assuming the inappropriate metadata can happen.

PR #58012, opened 2024-04-15
Add method to destroy a MailJet contact

bethanyaconnor:
Adds a method to destroy a MailJet contact. In a follow-up, I'll call this as part of the account purger.

Note that, for some reason, the API to delete a contact is only [available through cURL](https://dev.mailjet.com/email/guides/contact-management/#gdpr-delete-contacts).

### DIFF
--- a/lib/test/cdo/test_mailjet.rb
+++ b/lib/test/cdo/test_mailjet.rb
@@ -7,7 +7,7 @@ class MailJetTest < Minitest::Test
   end
 
   def test_create_contact
-    email = 'fake.email@email.com'
+    email = 'fake.email@test.xx'
     name = 'Fake Name'
 
     Mailjet::Contact.expects(:create).with(is_excluded_from_campaigns: true, email: email, name: name)
@@ -22,7 +22,7 @@ class MailJetTest < Minitest::Test
   end
 
   def test_create_contact_with_existing_contact
-    email = 'fake.email@email.com'
+    email = 'fake.email@test.xx'
     name = 'Fake Name'
 
     mock_existing_contact = mock('Mailjet::Contactdata')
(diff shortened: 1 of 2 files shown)


lib/test/cdo/test_mailjet.rb
```diff
@@ -67,4 +67,31 @@ def test_send_template_email
...
+    mock_contact.stubs(:id).returns(123)
+    Mailjet::Contact.expects(:find).with(email).returns(mock_contact)
+
+    mock_http_request = mock('Net::HTTP::Delete')
+    Net::HTTP::Delete.expects(:new).with(URI.parse("https://api.mailjet.com/v4/contacts/#{mock_contact.id}")).returns(mock_http_request)
+    mock_http_request.expects(:basic_auth).once
+
+    Net::HTTP.any_instance.expects(:request).with(mock_http_request)
+
+    MailJet.delete_contact(email)
```
TurnerRiley, line 84: Nit: could there be another use of `Mailjet::Contact.expects(:find).with(email).returns(mock_contact)` after this last line but with `.returns(nil)` instead to confirm it was deleted?

TurnerRiley (approved):
Looks great!

lib/test/cdo/test_mailjet.rb
bethanyaconnor: I'm not sure that would confirm it's actually deleted though, because we're not actually calling the MailJet service. Or am I misunderstanding?

TurnerRiley: Oh gotcha, I thought maybe we could check it with that line to have the format of 1) check that it returns the mock_contact like you'd expect, 2) delete it, 3) have the same check but expect it to be nil sort of deal. Totally fine if that wouldn't work!

bethanyaconnor: When I walk through those steps, I'm not sure what you're imagining for step 3. Can you maybe write some pseudocode?

For clarity, what I'm understanding is:

1. to change line 76 to be 
```
Mailjet::Contact.expects(:find).with(email).returns(mock_contact).then.returns(nil)
```

2. already done in line 84!

3. not sure

TurnerRiley: Resolved offline: I misunderstood what the line was doing and this test makes sense!

PR #58016, opened 2024-04-15
fix(lms): url encode for iframe passthrough

stephenliang:
This change URL encodes the `state` and `nonce` parameters. Since these parameters are used for redirects or window opens, the encoding is appropriate to ensure the values are passed through to the next page correctly.

## Testing story

1. Ran adhoc schoology instance
2. Tested launching schoology to Code.org

## PR Checklist:

- [x] Code is well-commented


carl-codeorg (approved):
LGTM. Not sure if it is any better, but there's also the `URI::encode_www_form` method that can handle the whole params object without calling url_encode individually for each file.

nicklathe (approved):
LGTM, nice catch.

PR #58029, opened 2024-04-16
Revert "Fix Violations of and Reenable `Style/RescueModifier`"

Hamms:
Reverts code-dot-org/code-dot-org#57847

The original PR failed linting during the staging build, because of some stray inline rescues in HAML code which didn't get picked up by Drone or by my local testing:

```
RAILS_ENV=staging RACK_ENV=staging bundle exec haml-lint dashboard pegasus shared
rake aborted!
'RAILS_ENV=staging RACK_ENV=staging bundle exec haml-lint dashboard pegasus shared' returned 65
warning: parser/current is loading parser/ruby30, which recognizes 3.0.6-compliant syntax, but you are running 3.0.5.
Please see https://github.com/whitequark/parser#compatibility-with-ruby-mri.
dashboard/app/views/layouts/application.html.haml:12 [W] RuboCop: Style/RescueModifier: Avoid using `rescue` in its modifier form.
dashboard/app/views/levels/_admin.html.haml:118 [W] RuboCop: Style/RescueModifier: Avoid using `rescue` in its modifier form.
pegasus/sites.v3/code.org/views/theme_common_head_after.haml:1 [W] RuboCop: Style/RescueModifier: Avoid using `rescue` in its modifier form.
pegasus/sites.v3/hourofcode.com/public/events/2013/index.haml:21 [W] RuboCop: Style/RescueModifier: Avoid using `rescue` in its modifier form.
pegasus/sites.v3/hourofcode.com/public/events/schools/index.haml:21 [W] RuboCop: Style/RescueModifier: Avoid using `rescue` in its modifier form.
pegasus/sites.v3/hourofcode.com/public/events/special/index.haml:21 [W] RuboCop: Style/RescueModifier: Avoid using `rescue` in its modifier form.

1116 files inspected, 6 lints detected
Tasks: TOP => lint => lint:all => lint:haml
(See full trace by running task with --trace)

[CDO]/lib/cdo/rake_utils.rb:81:in `system_stream_output'
[CDO]/lib/cdo/rake_utils.rb:66:in `rake_stream_output'
[CDO]/aws/ci_build:89:in `block in build'
[CDO]/aws/ci_build:65:in `chdir'
[CDO]/aws/ci_build:65:in `build'
[CDO]/aws/ci_build:112:in `block in main'
[CDO]/lib/cdo/rake_utils.rb:329:in `capture'
[CDO]/aws/ci_build:111:in `main'
[CDO]/aws/ci_build:281:in `<main>'
[...]

PR #55312, opened 2023-12-05
Run Linting in Drone CI

Hamms:
Specifically, I extended our existing pre-commit linting logic to be able to check against the entire subset of files changes between two branches rather than just the subset of files that are staged for commit. This definitely feels a little bit hacky, but my attempts at getting this to work using 

This approach has the primary advantage of being quite fast; I tested changesets that affected around a dozen files at a time, and it took less than three seconds to complete; even faster if it does find a linting problem.

This does also have one notable disadvantage, which is that it is now possible for a linting problem to cause a CI build to fail before it has the chance to run unit tests, forcing a developer through another "push changes and wait for build" cycle before they are able to see the results of running unit tests. Our expectation is that pre-commit-hook linting will catch the vast majority of potential linting failures before ever hitting drone; for the fraction of issues that do make it through, the pain of waiting for another drone build should be offset by not having to deal with the pain of dealing with a build failure in staging.

If anyone has feedback about this, particularly any concerns about how this might negatively impact your development experience, please let the developer tools working group know!

## Testing story

I manually introduced linting errors to verify that they get caught in Drone, and used the bash `time` command to monitor how long it took.

## Follow-up work

Note that we currently have two independent linter stacks; [`lib/rake/lint.rake`](https://github.com/code-dot-org/code-dot-org/blob/staging/lib/rake/lint.rake) contains the "standard" linting logic which executes as part of the staging build and lints everything, [`tools/hooks/lint.rb`](https://github.com/code-dot-org/code-dot-org/blob/staging/tools/hooks/lint.rb) contains the linter we use in pre-commit hooks. These do represent two distinct use cases; the first is intended to be thorough, the second intended to be fast. There is however also a lot of repeated code between the two of them, and with this change they are drifting even farther apart in terms of core functionality.

It would probably be worthwhile to do a pass over these two sets of logic and consolidate as much of their functionality as possible into a single place. Out of scope for this change, unfortunately.

PR #58030, opened 2024-04-16
Revert "Revert "Fix Violations of and Reenable `Style/RescueModifier`""

Hamms:
Reverts code-dot-org/code-dot-org#58029, reapplying https://github.com/code-dot-org/code-dot-org/pull/57847

The original PR failed linting when it hit staging, because neither my manual local testing nor automated Drone testing linted haml, and we had a few violations of the new rule hidden in some of our haml templates.

I manually fixed those violations, and verified with `bundle exec haml-lint` that there are no more.


pegasus/sites.v3/hourofcode.com/public/events/2013/index.haml
```diff
@@ -13,12 +13,17 @@ nav: events_nav
...
+  rescue
+    0
+  end
 
 %h1= hoc_s(:events_all_old_title).gsub(/\@year/, hoc_year.to_s)
 
 %ul
   %li{class: "hoc-event-country"}
-    %span{style: "font-weight: bold;"}= "United States (#{us_events.count rescue '0'} #{hoc_s(:events)})"
+    %span{style: "font-weight: bold;"}= "United States (#{us_events_count} #{hoc_s(:events)})"
```
Hamms, line 26: Note that this is actually a *slight* refactoring of this very old (and likely entirely unused) page, to better align with how newer pages work. Because `us_events` is a group of collections, simply calling `.count` on it will return the number of collections in the group, not the number of events in all the collections.

davidsbailey (approved):
updates look good! yay for haml linting but boo for not running it until staging. 

do you have a sense of when the appropriate time to run haml-lint would be, of the places you mentioned that it's not currently being run?

Hamms:
Oh, to clarify; Drone *will* lint HAML files, but only [HAML files that have been modified in the PR itself](https://github.com/code-dot-org/code-dot-org/blob/12e028b4adf8c263038bde9009e6af3251ba74c5/tools/hooks/lint.rb#L103).

We could consider linting every file in the codebase for every drone run, but at a cost of slowing down all our drone runs. See https://github.com/code-dot-org/code-dot-org/pull/55312 for some previous discussion on the topic

PR #58037, opened 2024-04-16
[ProgressV2] create a new feedback banner for progress table v2

lfryemason:
## Warning!!

https://github.com/code-dot-org/code-dot-org/assets/25193259/26f4c5e8-2829-4fef-b58e-90b2960c02cd

Shows a feedback banner for the new progress view.

Banner is only showed after the second time a user has clicked on the new progress view and only on the new progress view. Banner should only be shown until it has been answered for each user.

Based off of [this](https://github.com/code-dot-org/code-dot-org/pull/57517/files) PR, however we are not using local storage due to race conditions with loading feedback. This could be added later as an optimization.

Testing:
Unit tests will be added as a fast follow. We are trying to get this banner in before a release later today. Extensive manual testing was taken instead of unit tests and unit tests will be added ASAP.

## Testing story
Manual testing steps:
* banner is never shown on v1 table
* banner is not shown when `switch to new progress view` is clicked
* Banner is shown if v2 table page is hard reloaded
* Banner is not shown if feedback has already been given
* Answered banner shows the `share more` link
* Closed banner is never re-shown

## PR Checklist:

- [ ] Code is well-commented


dju90:
For the record, the destination survey will have a title and description at the top - just waiting for a DTP to see those changes in prod
image

apps/src/utils.js
```diff
@@ -406,7 +406,7 @@ export function showUnusedBlockQtip(targetElement) {
 /**
  * @param {string} key
  * @param {string} defaultValue
- * @return {string}
+ * @return {string} returns the value of the key in localStorage, null if not set or the defaultValue if there is an error
```
kobryan0619, line 409: Thanks for adding this documentation!

apps/src/util/sharedVoices.js
```diff
@@ -0,0 +1,33 @@
+/* eslint-disable */
```
kobryan0619, line 1: I am a bit confused on how this did get generated here given I don't see any changes to `shared_constants.rb` in this PR.  For my own learning, can you help me understand this?

lfryemason: This was a problem with staging that occurred a few days ago. Slack [link](https://codedotorg.slack.com/archives/C0T0PNTM3/p1712940607887079), I'll go ahead and remove the changes

kobryan0619 (approved):
Looks good!

apps/src/templates/sectionProgressV2/progressV2FeedbackRedux.js
```diff
@@ -0,0 +1,106 @@
...
+  })
+    .then(response => {
+      if (!response.ok) throw new Error('Failed to fetch feedback');
+      return response.json();
+    })
+    .then(progressV2Feedback => {
+      dispatch({
+        type: PROGRESS_V2_FEEDBACK_FETCH_SUCCESS,
+        progressV2Feedback:
+          progressV2Feedback === null ? {empty: true} : progressV2Feedback,
```
kobryan0619: Note to future self: this may be a bigger part of the unit test.

apps/src/templates/sectionProgressV2/SectionProgressSelector.jsx
```diff
@@ -23,12 +24,17 @@ function SectionProgressSelector({
   progressTableV2ClosedBeta,
   sectionId,
 }) {
+  // Only show the feedback banner's default state if the user has not manually selected a view.
```
etaderhold, line 27: Just trying to wrap my head around this...when you load the page the `showFeedbackBannerLocked` will be false by default, so if you're coming straight into progress v2 the banner will show up if you haven't answered/closed it yet. But if your original page load brings you to progress v1 and you click the toggle the banner will not show up. Also if you toggle v2->v1->v2 the banner will be suppressed at the end. Do I have it right?

apps/src/templates/sectionProgressV2/progressV2FeedbackRedux.js
```diff
@@ -0,0 +1,106 @@
+import {getAuthenticityToken} from '@cdo/apps/util/AuthenticityTokenStore';
+
+const PROGRESS_V2_FEEDBACK_URL = '/new_feature_feedback';
+const progressV2FeedbackKey = 'progress_v2';
```
etaderhold:
```suggestion
const PROGRESS_V2_FEEDBACK_KEY = 'progress_v2';
```

apps/src/utils.js
etaderhold: I kind of wonder if this whole `defaultValue` parameter uses incorrect assumptions about the local storage interface. If I had to guess, the implementer of this method probably intended the `defaultValue` to be used for all cases where the object with `key` was not present in local storage.

The MDN docs about the Storage interface however mention that `null` will be returned in this case, and they don't mention any circumstance where `getItem` would throw an exception, so this might only come up in cases where `localStorage` is not actually a `Storage` object to begin with. When would this even occur?

Note I'm _not_ suggesting changing this behavior now as it's probably relied on in a bunch of different places. Thanks for adding the clarifying comment!

apps/src/templates/sectionProgressV2/progressV2FeedbackRedux.js
```diff
@@ -0,0 +1,106 @@
...
+      return {
+        ...state,
+        isLoading: false,
+        progressv2Feedback: action.progressV2Feedback,
+      };
+    case PROGRESS_V2_FEEDBACK_CREATE_FAILURE:
+      return {...state, isLoading: false, error: action.error};
+    default:
+      return state;
+  }
```
etaderhold: This makes it clearer that the state change operation is currently the same for the fetch and create requests, perhaps making a future change to this less error-prone if the change is inadvertently applied to one place and not the other.

```suggestion
  switch (action.type) {
    case PROGRESS_V2_FEEDBACK_FETCH_REQUEST:
    case PROGRESS_V2_FEEDBACK_CREATE_REQUEST:
      return {...state, isLoading: true};
    case PROGRESS_V2_FEEDBACK_FETCH_SUCCESS:
    case PROGRESS_V2_FEEDBACK_CREATE_SUCCESS:
      return {
        ...state,
        isLoading: false,
        progressV2Feedback: action.progressV2Feedback,
      };
    case PROGRESS_V2_FEEDBACK_FETCH_FAILURE:
    case PROGRESS_V2_FEEDBACK_CREATE_FAILURE:
      return {...state, isLoading: false, error: action.error};
    default:
      return state;
  }
```

apps/src/templates/sectionProgressV2/ProgressFeedbackBanner.jsx
```diff
@@ -0,0 +1,126 @@
...
+
+  const answer = satisfied => {
+    if (bannerStatus === BANNER_STATUS.UNANSWERED) {
+      setAnswered(true);
+      createProgressV2Feedback(satisfied);
+    }
+  };
+
+  const close = () => {
+    setBannerStatus(BANNER_STATUS.CLOSED);
```
etaderhold, line 84: I wonder if it would be smart to persist this closed state in local storage so that we don't keep showing the banner on every page reload to people who have hit the X in the past. Could be frustrating to keep saying "I don't want to answer this question" and yet having it shown again and again.

etaderhold (approved):

apps/src/templates/sectionProgressV2/SectionProgressSelector.jsx
lfryemason: Yep!

apps/src/utils.js
lfryemason: I completely agree that this seems weird and if I were to use the local storage, I might make a new item that does use the default value if the item is not found. I added the extra documentation because it took me about an hour to find what was wrong because I made assumptions about the function before I finally went deeper into the MDN function.

I'm hoping that this was not an MDN function that changed after our implementation expected the old behavior, but since everything seems to be working correctly, it does not seem to be.

apps/src/templates/sectionProgressV2/progressV2FeedbackRedux.js
lfryemason: Thank you for this great suggestion. This actually found an error where I had both `progressV2Feedback` and `progressv2Feedback`

apps/src/templates/sectionProgressV2/ProgressFeedbackBanner.jsx
lfryemason: The `share more` banner is only shown if the user just answered the first question. This is because we only set the state to `ANSWERED` if we have called the `answer` method. Otherwise, if feedback exists we set the state to `CLOSED`

etaderhold: Ah, I missed that the component doesn't show an "X" to close during the initial thumbs up/down state. Maybe it should? At a previous employer I worked on a similar survey and we'd stop showing it to you after you clicked the X a certain number of times. Maybe 3?

That's more of a product question about how badly we want full participation on this question, or if we'd rather let people opt out.

lfryemason: Ah, there is no close button on the feedback component before the user has answered the question. This was the designs from LTI, but maybe we should reconsider for our use.

apps/src/templates/sectionProgressV2/progressV2FeedbackRedux.js
etaderhold: Yay! I didn't even notice that initially, but a bias toward DRY definitely saves the day sometimes.

PR #58043, opened 2024-04-16
Add AiChatSession table

bencodeorg:

## Description

Creates the new `AiChatSession` table to store student interactions in our generative AI lab. A few notes:
- The idea is that a row would be unique per a student's AI customizations on a given level. I believe in our `AiChatController` we'll need to diff against all existing entries on the same level to see whether to replace an existing record (ie, this is a new chat log for an existing set of model configuration?), or insert a new row if there is no existing record.
- There was a `section_id` in the initially spec @sanchitmalhotra126 threw together, but I don't think we generally can associate student activity on a level with a specific section (it can be derived through user -> follower -> section, however), so I removed it.
- This is heavily inspired by the [existing `AiTutorInteraction` ](https://github.com/code-dot-org/code-dot-org/blob/staging/dashboard/app/models/ai_tutor_interaction.rb)model. ~~There are also a lot of indices in there -- I didn't include them for now (other than querying per student), but it is not a problem to add these if we want to support more complex query behavior~~. I switched to use a composite index (user ID - level ID - script ID) with user ID as the top-level index, which should give us indices to allow searching for individual students (for any teacher view), as well as level-specific queries we might need in order to update existing rows.
- I set the table to use utf8mb4 charset to support emojis.
- ~~I'm kind of curious about the `json` column type vs `text`, but I think text will suit us for now.~~ Used JSON columns for model customizations and messages.

## Testing story

Successfully ran this migration up and down locally.


dashboard/app/models/ai_chat_session.rb
```diff
@@ -0,0 +1,20 @@
...
+#  model_customizations :text(65535)
+#  messages             :text(65535)
+#  created_at           :datetime         not null
+#  updated_at           :datetime         not null
+#
+# Indexes
+#
+#  index_ai_chat_sessions_on_user_id  (user_id)
+#
+class AiChatSession < ApplicationRecord
```
sanchitmalhotra126: nit: so far in our code base we've been naming it `Aichat`/`aichat`, should we do the same here to keep it consistent?

sanchitmalhotra126 (approved):
Nice, LGTM!

> The idea is that a row would be unique per a student's AI customizations on a given level. I believe in our AiChatController we'll need to diff against all existing entries on the same level to see whether to replace an existing record (ie, this is a new chat log for an existing set of model configuration?), or insert a new row if there is no existing record.

Yeah, I was thinking one way to handle this would be to have the aichat_controller return a `chat_session_id` that corresponded to the model ID that was created for the current chat session. That way the client could just send back this ID so we know whether to overwrite an existing entry or create a new one. We could then have the front end just remove the ID whenever it wanted to create a new session (i.e. if some params changed), but we could also double check to make sure that if an ID was provided, the customizations should match. Thoughts?

> I'm kind of curious about the json column type vs text, but I think text will suit us for now.

Makes sense to me - maybe json would be useful for the model_customizations field? But yeah text works just fine since I don't think we'll be doing any complex look ups with that field specifically.

dashboard/app/models/ai_chat_session.rb
bencodeorg: Oh good call, updated!

bencodeorg:
> _Yeah, I was thinking one way to handle this would be to have the aichat_controller return a chat_session_id that corresponded to the model ID that was created for the current chat session. That way the client could just send back this ID so we know whether to overwrite an existing entry or create a new one. We could then have the front end just remove the ID whenever it wanted to create a new session (i.e. if some params changed), but we could also double check to make sure that if an ID was provided, the customizations should match. Thoughts?_

That sounds great, Alice and I were chatting yesterday about how we already do the diffing of the customizations on the front end and how we might be able to leverage that here. Do you think we could get away without the "double check"? I guess I'm wondering if we're double checking anyway, doing the sort of `chat_session_id` management (which I think sounds like a great idea generally) on the front end seems like an extra step.

[edit: maybe the "double check" is different / better than my original idea since we wouldn't be comparing each of a student's sessions for a given level, just the one that we're about to write to. If we wanted to be comprehensive, we could in go at that point and check all of the other student's sessions for the level before writing a new record]

[edit 2: we probably want to do this "double check" when we don't have a session ID, so that we don't create duplicate records if a student just refreshes the page and has the same ai customizations]

[edit 3: actually, maybe a new page load with the same customizations is in fact a new session, since we don't reload their chat history from the previous page view? A tangential complication is that we give users the ability to clear the chat window, which clears the chat history in Redux and could mess up our logging...]

dashboard/app/models/aichat_session.rb
```diff
@@ -0,0 +1,20 @@
+# == Schema Information
+#
+# Table name: aichat_sessions
+#
+#  id                   :bigint           not null, primary key
+#  user_id              :integer
+#  level_id             :integer
```
fisher-alice, line 7: `level_id` and `project_id` are sent as strings from the front-end but I see they're integers in `ai_tutor_interaction.rb` so unsure about this.

dashboard/app/models/aichat_session.rb
```diff
@@ -0,0 +1,20 @@
...
+#  script_id            :integer
+#  project_id           :integer
```
fisher-alice, line 9: Are we planning to convert `channel_id` to `project_id`?

dashboard/app/models/aichat_session.rb
bencodeorg: Hmm good callout, I'll make sure that level ID gets stored correctly. And yeah we'll be storing the project_id (integer) rather than the channel_id (string), to your other question

dashboard/app/models/aichat_session.rb
bencodeorg: Yes, was imagining using `storage_decrypt_channel_id`? I think by default when saving a project identifier in a Rails model, we should be saving `project_id` rather than `channel_id` -- makes for easier lookups of any project metadata we might need (should be as simple as `aichat_session.project...`), and the word channel confuses me!

fisher-alice: Sounds good - thanks for the explanation!

fisher-alice (approved):
Thanks for clarification!

dashboard/app/models/aichat_session.rb
cnbrenci: This is the right thing to do - we don't want to store channel_ids in any new tables going forward. (we had originally been doing that in datablock storage and infra strongly suggested against it for those exact reasons :))

dashboard/app/models/aichat_session.rb
```diff
@@ -0,0 +1,20 @@
...
+#  model_customizations :text(65535)
```
cnbrenci: If we're on the fence between text & json, would it be easier to start with json and change from json->text later vs. the other way around, since the json type enforces valid json?

bencodeorg: [tl;dr changed to json type columns...a few assorted thoughts below]

hmm yeah that makes sense...for our purposes, I'm not sure if makes much of a difference. Was reading thru docs here, seems like there are some storage benefits, and there are handy additions when writing raw SQL that are possible when the column type is JSON:

https://dev.mysql.com/doc/refman/5.7/en/json.html

Tangentially, I stumbled across this bit of guidance about foreign keys when I was thinking about whether to include foreign keys here or not. It kind of made me wonder if enforcing valid JSON at the database level is considered good or bad by Rails ppl:

https://guides.rubyonrails.org/active_record_migrations.html#active-record-and-referential-integrity

> The Active Record way claims that intelligence belongs in your models, not in the database. As such, features such as triggers or constraints, which push some of that intelligence back into the database, are not recommended.

I don't see any explicit guidance on JSON vs text columns elsewhere though. And databases obviously already enforce types to some degree (strings vs ints etc), so sort of a gray area.

bencodeorg: @sanchitmalhotra126 you mentioned just model customizations in JSON in the comment below, any objection to messages as well? Per what's coming into the controller right now, I think the messages will also be an array of objects and would be fine to serialize as JSON

> _Makes sense to me - maybe json would be useful for the model_customizations field? But yeah text works just fine since I don't think we'll be doing any complex look ups with that field specifically._

cnbrenci: FWIW, there is a precedent in the codebase 
image

(I think you can tell which camp I'm in based on the search results 😁 )

bencodeorg: haha you're a pioneer! The contact rollups examples do actually write raw SQL against the JSON field, so the benefits are clear there. I think the JSON type only came out in MySQL 5.7 (our current version), so I'm guessing we'll see more usages as time goes on

https://github.com/code-dot-org/code-dot-org/blob/90fc57339095d6f3588836e299e2f00b1347a149/dashboard/app/models/contact_rollups_pardot_memory.rb#L253-L258

cnbrenci (approved):
🥳 🚀

PR #58044, opened 2024-04-16
[JIMT][LABS-698] relocate generated scripts

thomasoniii:
This is just a tidying up PR.

Our `apps/script/generateSharedConstants.rb` file would generate javascript files and place them into directories local to where they were used. This isn't bad in and of itself, but as we generate more files, we risk having more generated scripts across the repo which will be tougher to keep track of.

Further, each time we'd add a new script, we'd need a new entry in `.gitignore` for it or else we'll get warnings about untracked files.

This PR just centralizes everything. Files are somewhat arbitrarily places into `apps/generated-scripts`, to make it clear that they are part of the `apps` content and explicitly _not_ considered source files in the `src` directory.

Then, every place one of those files is imported was changed to use a new `@cdo/generated-scripts` alias which points to that directory.

And finally, `.gitignore` was updated to ignore that new directory, as opposed to the individual files.

As long as the tests are passing, this should be fine. It's adding no new functionality, so the only questions may be around if this is organized the way we'd like it to be.

If I've tagged you as a reviewer on this PR, that means you either commented on the original thread in [slack/developers](https://codedotorg.slack.com/archives/C0T0PNTM3/p1712940607887079) about this or you were the last person on the blame for a generated file in `generateSharedConstants.rb` (or you're student-labs, which gets these PRs by default)

## Links

- jira ticket: [LABS-698](https://codedotorg.atlassian.net/browse/LABS-698)


thomasoniii:
One thing of note that I waffled on - I'd ended up just sticking all of the constants into the single `generated-scripts` directory.

We could do that, or we could maintain a hierarchy (maybe `applab`, `gamelab`, common files in the root dir, etc). I go back and forth. Does anyone have a strong opinion?

.gitignore
davidsbailey, line 1: one more line to remove, which does not yet appear to be merged into your branch https://github.com/code-dot-org/code-dot-org/blob/a4b116f86623891652eaf98368d84bdd2fafabdf/.gitignore#L106

thomasoniii: Ha. I knew my staging was a few days old, but I figured I'd chance just sending it in and that the gitignore hadn't been updated yet. Oops. 😊 

Fixed now!

jmkulwik (approved):
Thanks for doing this!

PR #58048, opened 2024-04-16
Music: fix small validation bug

sanchitmalhotra126:
## Warning!!

## Description

Fix a small validation bug spotted in the elementary pilot. When a student plays multiple sounds together, we should consider the "played_sounds_together" conditions satisfied if they played the required number of sounds or more. However, we were accidentally only checking the exact number of sounds, and so if they played more than the passing threshold, they wouldn't pass the level.

## Links

https://codedotorg.slack.com/archives/C043WM7TCH3/p1713295216389199

## Testing story

Tested with elementary and intro progression.


thomasoniii (approved):
lgtm 🚀

breville (approved):
Thanks for this.  I think the original design was intentional so that we didn't create an unbounded number of satisfied conditions, but this change seems reasonable.

sanchitmalhotra126:
> Thanks for this. I think the original design was intentional so that we didn't create an unbounded number of satisfied conditions, but this change seems reasonable.

Fair point, and I don't mind reinstating the limit if you think it's a good idea? The issue was just that we were just using the wrong variable - I just removed the limit as an added change while I was here.

breville:
> > Thanks for this. I think the original design was intentional so that we didn't create an unbounded number of satisfied conditions, but this change seems reasonable.
> 
> Fair point, and I don't mind reinstating the limit if you think it's a good idea? The issue was just that we were just using the wrong variable - I just removed the limit as an added change while I was here.

Oh this seems alright.  I think there's a bigger architectural discussion about how we can have conditions that are a bit more expressive, but that can be done separately!

PR #58049, opened 2024-04-16
chore(ejs): enable ejs javascript strict mode to remove `with` syntax

stephenliang:
Note: This change is required for the jest migration as jest enables strict mode during babel transpilation.

Currently, Code.org uses EJS to templatize files. EJS, by default, utilizes the `with` keyword which is deprecated and per MDN, "this feature may cease to work at any time.". To remove this keyword, the EJS strict mode option is enabled (`with` is prohibited in strict).

To replace the functionality previously provided by `with` (add variables within a given block scope), the variables are destructured from the `locals` variable (which is provided by EJS). In strict mode, variables are not scoped globally by default and therefore assigning the variables within the EJS strict mode provides the same functionality as `with` but with extra syntax needed to use an EJS template.

Previously, the code would be able to function:

```
<% hello %>
```

Now with strict mode enabled, the same code would need to become:

```
<% locals.hello %>
```

To avoid mistakes, a variable reassignment block was created to existing EJS files. This will enable the variables to continue to be used as if the with block was available.

To identify the locals, all EJS exported default functions were inspected, for example:

```
 return DropletFunctionTooltipMarkup({
    funcName: tooltipInfo.functionName,
    functionName: tooltipInfo.functionName,
    isProperty: tooltipInfo.isProperty,
    tipPrefix: tooltipInfo.tipPrefix,
    functionShortDescription: tooltipInfo.description,
    parameters: tooltipInfo.parameterInfos,
    signatureOverride: tooltipInfo.signatureOverride,
    showExamplesLink: this.showExamplesLink,
    currentParameterIndex: currentParameterIndex,
  });
```

The object was extracted and converted from key/value ES6 object pairs to

```
var funcName = locals.funcName;
var functionName = locals.functionName;
...
```

This enables the local variable to be used in the same way it was expected with the “with” keyword.

## Links

- [with deprecation and removal notice](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/with)
- [with in strict mode](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode#simplifying_scope_management)
- [ejs documentation](https://ejs.co/#docs)

## Testing story

These files are primarily used in the labs and exporter functionality. I ran through each lab and was able to progress in levels and export the labs as well.

## Deployment strategy

To minimize risk to AP CSP, this change will only be merged after April 30th.

## Follow-up work

1. Jest Migration

## Privacy

N/A

## Security

N/A

## PR Checklist:

- [x] Code is well-commented

### DIFF
--- a/apps/package.json
+++ b/apps/package.json
@@ -85,7 +85,7 @@
     "copy-webpack-plugin": "^9.0.1",
     "css-loader": "^6.2.0",
     "dedent": "^1.5.1",
-    "ejs": "^2.6.1",
+    "ejs": "^3.1.10",
     "ejs-webpack-loader": "^2.2.2",
     "enzyme": "^3.9.0",
     "enzyme-adapter-react-16": "^1.15.6",
--- a/apps/src/blockTooltips/DropletParameterTooltip.html.ejs
+++ b/apps/src/blockTooltips/DropletParameterTooltip.html.ejs
@@ -1,4 +1,18 @@
 <% var i18n = require('@cdo/locale'); %>
+<%
+/**
+ * This block destructures the local variables for use within this ejs template.
+ */
+var funcName = locals.funcName;
+var functionName = locals.functionName;
+var isProperty = locals.isProperty;
+var tipPrefix = locals.tipPrefix;
+var functionShortDescription = locals.functionShortDescription;
+var parameters = locals.parameters;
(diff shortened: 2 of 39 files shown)


wilkie (approved):
Happy to approve on the Platform side of things! So good to be moving steadily toward a solid modern structure.

PR #58051, opened 2024-04-16
[Python Lab] Enable running multi-file projects

molly-moen:
## Warning!!

## Summary
Now that we can create multi-file projects in python lab, we can enable running multi-file projects. I ended up using a different strategy from my initial proof of concept (#55783). That worked for files in the same folder but didn't work for nested folders. Instead I utilized Pyodide's [filesystem API](https://pyodide.org/en/stable/usage/file-system.html), which has the advantage of being able to run via JavaScript for easier readability.

While working on this, I discovered the files stick around between runs. Even if you update the files, if the user code imported one of them you would not see the update on subsequent runs. This is because pyodide has a single interpreter for the entire lifetime of the browser session. In order to work around this issue I needed to do two things:
- After each run, find any imported modules in `sys.modules` that came from user-written files and delete them.
- Delete the files between each run (this is so a deleted file can't be imported in a future run).

Now you can run projects with imports from nested files!

## Links

- jira ticket: [CT-494](https://codedotorg.atlassian.net/browse/CT-494)

## Testing story
Tested locally.

## Follow-up work
The single interpreter issue in pyodide may cause confusing behavior with variables too. I filed a [task](https://codedotorg.atlassian.net/browse/CT-505) to look into this.

## PR Checklist:

- [ ] Code is well-commented


apps/src/weblab2/CDOIDE/cdoIDEContext/utils.ts
```diff
@@ -27,11 +27,20 @@ import {
 
 const DEFAULT_NEW_FILE_CONTENTS = 'Add your changes to ${fileName}';
 
-export const getNextFileId = (files: ProjectFile[]) =>
-  String(Math.max(...files.map(f => Number(f.id))) + 1);
+export const getNextFileId = (files: ProjectFile[]) => {
+  if (files.length === 0) {
```
molly-moen: I discovered if you deleted all the files in the project and tried to create a new one it would have an id of `'-Infinity'` and only one file and one folder could be created. This allows for a default id.

apps/src/lab2/constants.ts
```diff
@@ -16,3 +16,5 @@ export const BLOCKLY_LABS: AppName[] = [
   'poetry',
   'spritelab',
 ];
+
+export const MAIN_PYTHON_FILE = 'main.py';
```
molly-moen, line 20: I was using `'main.py'` more in this pr, so I added a constant for it.

apps/src/lab2/responseValidators.ts
```diff
@@ -40,8 +40,11 @@ const PythonSourceResponseValidator: ResponseValidator<
       throw new ValidationError('Python sources must be a JSON object');
     }
     const source = responseToValidate.source as MultiFileSource;
-    if (!source?.files || getFileByName(source.files, 'main.py') === null) {
-      throwMissingFieldError('main.py');
+    if (
```
molly-moen, line 43: this change is just using the new constant

apps/src/weblab2/CDOIDE/cdoIDEContext/utils.ts
```diff
@@ -27,11 +27,20 @@ import {
...
 
-export const getNextFolderId = (folders: ProjectFolder[]) =>
-  String(Math.max(...folders.map(f => Number(f.id))) + 1);
+export const getNextFolderId = (folders: ProjectFolder[]) => {
+  if (folders.length === 0) {
+    return '1';
+  } else {
+    return String(Math.max(...folders.map(f => Number(f.id))) + 1);
+  }
+};
```
thomasoniii: Good catch!

I'd suggest fixing it like this:
```
export const getNextFolderId = (folders: ProjectFolder[]) =>
  String(Math.max(0, ...folders.map(f => Number(f.id))) + 1);
```

That way we don't need the conditional, but we'll still start at "1" if the list is empty.

apps/src/pythonlab/patches/pythonScriptUtils.ts
```diff
@@ -23,3 +14,98 @@ export function applyPatches(originalCode: string) {
...
+  for (const file of Object.values(source.files)) {
+    if (file.name !== excludedFileName) {
+      const filePath = getModuleName(file.id, source);
+      result += `
+if "${filePath}" in sys.modules:
+  del sys.modules['${filePath}']
+`;
+    }
+  }
+  return result;
```
thomasoniii: This is fine, but as a non-blocking suggestion I hate doing string concatenation like this, especially where the newlines are significant. I'd skew towards building an array and joining it.

```
const result = []
...
if (file.name !== excludedFileName) {
  const filePath = getModuleName(file.id, source);
  result.push(
    [`if "${filePath}" in sys.modules:`, `  del sys.modules['${filePath}']`].join("\n")
  );
}

return result.join("\n")
```

Again, this one's personal preference and not required. I also waffle on whether the embedded python should be joined or kept as a formatted template literal. But the outer list of results feels like it should be in a joined list.

apps/src/pythonlab/patches/pythonScriptUtils.ts
```diff
@@ -23,3 +14,98 @@ export function applyPatches(originalCode: string) {
...
+      pyodide.FS.writeFile(`${currentPath}${file.name}`, file.contents);
+    });
+  Object.values(source.folders)
+    .filter(f => f.parentId === currentFolderId)
+    .forEach(folder => {
+      // Create folder if it doesn't exist.
+      const newPath =
+        currentPath.length === 0
+          ? `${folder.name}`
+          : `${currentPath}${folder.name}`;
```
thomasoniii: Can't this just be 
```
`${currentPath}{folder.name}`
```
w/o the ternary?

It's already assuming that `currentPath` is defined, and if the length is 0, it'd just output an empty string anyway, right?

apps/src/pythonlab/patches/pythonScriptUtils.ts
```diff
@@ -23,3 +14,98 @@ export function applyPatches(originalCode: string) {
...
+
+// For the given fileId, return the full path to the file, including the file name.
+const getFilePath = (fileId: string, source: MultiFileSource) => {
+  let path = source.files[fileId].name;
+  let folderId = source.files[fileId].folderId;
+  while (source.folders[folderId]) {
+    path = source.folders[folderId].name + '/' + path;
+    folderId = source.folders[folderId].parentId;
+  }
+  return path;
```
thomasoniii: same as above, non-blocking personal preference would be to do this with an array:

```
const path = [source.files[fileId].name]
let folderId = source.files[fileId].folderId;
  while (source.folders[folderId]) {
    path.unshift(source.folders[folderId].name);
    folderId = source.folders[folderId].parentId;
  }
  return path.join('/');
```

apps/src/pythonlab/patches/pythonScriptUtils.ts
```diff
@@ -23,3 +14,98 @@ export function applyPatches(originalCode: string) {
...
+// For the given fileId, return the module version of the file. For example, a file at
+// path folder1/folder2/file.py would have a module name of "folder1.folder2.file".
+const getModuleName = (fileId: string, source: MultiFileSource) => {
+  let path = source.files[fileId].name.replace('.py', '');
+  let folderId = source.files[fileId].folderId;
+  while (source.folders[folderId]) {
+    path = source.folders[folderId].name + '.' + path;
+    folderId = source.folders[folderId].parentId;
+  }
+  return path;
```
thomasoniii: and same non-blocking array comment here, even with the same code as above and just a different delimiter.

I dunno if these two things are conceptually similar enough to merit pulling that code out into a function to prevent the duplication. They feel different enough that keeping it duplicated is probably fine, but I dunno.

thomasoniii (approved):
couple of non-blocking comments -
- I suggested a different way to fix the folderId error w/o using a conditional
- couple of suggestions about building paths with arrays that are joined on delimiters instead of formatted strings, but those ones are really just personal preference so I'll defer to your interest level.
- there's one note where I think we can get rid of a ternary

apps/src/pythonlab/patches/pythonScriptUtils.ts
molly-moen: I think you're right, at one point I had a `/` in here but it got moved to somewhere else.

apps/src/pythonlab/patches/pythonScriptUtils.ts
molly-moen: After updating to use the array method there is a clean way to extract out some of this code into a common method, so I'll do that!

apps/src/weblab2/CDOIDE/cdoIDEContext/utils.ts
molly-moen: Nice that's much cleaner

apps/src/pythonlab/patches/pythonScriptUtils.ts
molly-moen: I decided to keep the embedded python as a formatted string because it makes it a little easier to read and tell that there's an if statement going on (and because python is picky about tabs). But added the outer array!

PR #58054, opened 2024-04-16
Add poster attribute to video fallback player

kelbyhawn:
Adds a `poster` attribute to videos using the fallback player (`video_with_fallback.haml` view) on https://code.org. The thumbnail is pulled from the YouTube video similar to how the `display_video_thumbnail` view works as seen [here](https://github.com/code-dot-org/code-dot-org/blob/12e028b4adf8c263038bde9009e6af3251ba74c5/pegasus/sites.v3/code.org/views/display_video_thumbnail.haml#L13). 

The work in this PR is done in preparation for adding Swiper video carousels ([see Jira ticket](https://codedotorg.atlassian.net/browse/ACQ-1359)). There's a bug between our code and Swiper that prevents the video lightbox from working properly, so this a workaround for that, but this improves the UX/UI for the fallback video player in general, yay! Standalone videos (not in a carousel) will still use the `dsplay_video_thumbnail` view, so this isn't as relevant for them.

_Note:_ for cases where YouTube is blocked, this works best on videos that have downloadable backups hosted in S3 — all of our newer videos have these, but some legacy videos may not in which case the video won't show up. This is not a new issue, but may become more visible with this change if a video carousel is used that has older videos without backups.

**Jira ticket:** [ACQ-1787](https://codedotorg.atlassian.net/browse/ACQ-1787)

----

## Fallback player (when YouTube is blocked)

https://github.com/code-dot-org/code-dot-org/assets/9256643/df0a3ca4-b9bb-40e9-ac71-97ae13595c4f

## YouTube player (YT is not blocked)

https://github.com/code-dot-org/code-dot-org/assets/9256643/c354af16-3624-4291-9a27-de54dee4113d

### DIFF
--- a/pegasus/sites.v3/code.org/views/video_with_fallback.haml
+++ b/pegasus/sites.v3/code.org/views/video_with_fallback.haml
@@ -1,4 +1,4 @@
-- poster_path ||= nil
+- poster_path ||= "//i.ytimg.com/vi/#{video_code}/mqdefault.jpg"
 
 %figure{style: "width:100%; height: 100%; margin: 0 auto;"}
   .videodiv{style: "position:relative;"}
--- a/pegasus/sites.v3/code.org/views/video_with_fallback.haml
+++ b/pegasus/sites.v3/code.org/views/video_with_fallback.haml
@@ -1,3 +1,4 @@
+- download_path ||= nil
 - poster_path ||= "//i.ytimg.com/vi/#{video_code}/mqdefault.jpg"
 
 %figure{style: "width:100%; height: 100%; margin: 0 auto;"}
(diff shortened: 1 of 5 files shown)


TurnerRiley (approved):
Wooo this looks fantastic!

pegasus/sites.v3/code.org/views/video_with_fallback.haml
```diff
@@ -1,5 +1,8 @@
+- download_path ||= nil
+- poster_path ||= "//i.ytimg.com/vi/#{video_code}/mqdefault.jpg"
```
bethanyaconnor, line 2: I do wonder if schools that block YouTube also block ytimg 🤔 If ytimg is blocked, users will just see the black rectangles right? They would still be able to play the video?

kelbyhawn: Hypothetically yes, but the existing video lightboxes (`display_video_lightbox` view) would have the same issue since it's using that too. If they do get a black box thumbnail, at the very least the captions will still show up 😅

bethanyaconnor: fair fair. Sounds good!

PR #58060, opened 2024-04-17
Music: localize scripts

breville:
Localize the `music-intro-2024` script, and also a `music-onboard` script which contains new versions of some Panels levels that we'll swap into `music-intro-2024` for launch.

### DIFF
--- a/dashboard/lib/script_constants.rb
+++ b/dashboard/lib/script_constants.rb
@@ -256,6 +256,8 @@ module ScriptConstants
     HELLO_WORLD_EMOJI = 'hello-world-emoji'.freeze,
     HELLO_WORLD_RETRO = 'hello-world-retro'.freeze,
     K1HOC_2017 = 'k1hoc2017'.freeze,
+    MUSIC_INTRO_2024 = 'music-intro-2024'.freeze,
+    MUSIC_ONBOARD = 'music-onboard'.freeze,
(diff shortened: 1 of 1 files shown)


breville:
@code-dot-org/i18n @castro-jorge A couple questions:
1. Is this the appropriate way to ensure that the various levels inside `music-intro-2024` get their instructions/feedback/etc. localized?
2. There are a few levels (e.g. `musiclab_intro_panels_start_onboard`) that we'd like to get localized now, but we don't want them to be part of `music-intro-2024` yet, though we will modify that script to include them on launch day.  In this PR, the new levels are included in `music-onboard` which this PR also marks to be localized.  We don't intend to ever publish this script, but it acts as a vehicle to get them localized ahead of time.  Is this a good approach?  Is there a better way?

castro-jorge:
@breville marking `music-intro-2024` and `music-onboard` as FREEZE enables translatability for the strings under those scripts. So yes, this enables us to translate every string created under those two scripts.

How does _migrating_ the strings from `music-onboard` to `music-intro-2024` work? It is my understanding that if you change the tags of a string, this will create a new string in Crowdin that should inherit the translations from the original string (of course, this works if there are no updates to the string).

thomasoniii (approved):
lgtm 🚀

PR #58062, opened 2024-04-17
Music: default pack

breville:
This adds support for selecting the default pack of sounds, which allows the user to use the non-restricted sound packs without selecting a restricted one.

It's listed alongside the restricted packs in the pack selection UI.  It's not listed as a pack in the sound selection UI, since the non-restricted packs are already enumerated there.

In the project, the pack will be saved as `"default"`.

In the library, the pack can be set up like so:

```
    {
      "name": "Code.org sounds",
      "artist": "Code.org",
      "id": "default",
      "path": "packs/default",
      "imageSrc": "default.png",
      "restricted": true,
      "sounds": [
        { "name": "Preview", "src": "preview", "length": 2, "type": "preview" }
      ]
    }
```

### Pack selection UI

music-default-pack-0

### Header

Screenshot 2024-04-17 at 1 57 24 PM

### DIFF
--- a/apps/src/music/constants.ts
+++ b/apps/src/music/constants.ts
@@ -72,6 +72,8 @@ export const MIN_NUM_MEASURES = 30;
 export const LEGACY_DEFAULT_LIBRARY = 'default';
 export const DEFAULT_LIBRARY = 'intro2024';
 
+export const DEFAULT_PACK = 'default';
+
 export const DEFAULT_BPM = 120;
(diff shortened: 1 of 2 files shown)


breville:
@sanchitmalhotra126 When this is chosen, and we are using Tone.js, will we be using a key of C and a BPM of 120?

thomasoniii (approved):
lgtm 🚀

PR #58067, opened 2024-04-17
modified text size

kobryan0619:
## Warning!!

Modified font size of text in feedback bar.

- This should have no impact to the LTI Feedback Bar since the font size was already 16px.  
- The ProgressFeedbackBar went from font size of 13 to font size of 16.  
- I checked with Molly on this as well.
- The "Share More" bar will have the same font size (16) as well. 

Screenshots from the ProgressFeedbackBar
image

Screenshots from the LTI bar
image

Note: The icons are not showing up for me in either the "old" or "new" view - I think that might have to be with how I am forcing that bar to show up.  I am modifying the state of the component in the browser.  I think the icons aren't showing up because it isn't loading them initially... 

## Follow-up work
 - Look into icons not showing up in LTI bar...

### DIFF
--- a/apps/src/lib/ui/feedback/FeedbackBanner.scss
+++ b/apps/src/lib/ui/feedback/FeedbackBanner.scss
@@ -10,6 +10,11 @@
   color: $dark_charcoal;
   line-height: 2rem;
 
+  & .share-more,
+  & .feedback-title {
+    font-size: 16px;
+  }
+
   & .feedback-banner-greeting {
     display: inline-block;
     margin-right: $indent;
@@ -28,7 +33,7 @@
   }
 
   & i {
-    color: #0F0F0F;
+    color: #0f0f0f;
     font-size: 1.3rem;
   }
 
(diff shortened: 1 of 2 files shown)


dju90:
Looks much better than before! Thank you so much 🙏

MollyPeredo:
Looks good to me! Thanks for taking this across the finish line.

etaderhold (approved):

artem-vavilov (approved):
Nice work! 🙌🏻

apps/src/lib/ui/feedback/FeedbackBanner.scss
```diff
@@ -10,6 +10,11 @@
   color: $dark_charcoal;
   line-height: 2rem;
 
+  & .share-more,
+  & .feedback-title {
+    font-size: 16px;
```
artem-vavilov, line 15: Wouldn't it be better to set the font size for the entire block?

PR #58071, opened 2024-04-17
Rearrange the My PL Page section order

kelbyhawn:
Moves the Self-paced section on top of the My Workshops section on https://studio.code.org/my-professional-learning. I also wrapped the My Workshops section in a `<section>` element for margin purposes. There's still some extra space under the Survey banner, but I can clean that up as work continues on this page. 

**Jira ticket:** [ACQ-1793](https://codedotorg.atlassian.net/browse/ACQ-1793)

----

## Before
Before

## After
After


hannahbergam (approved):
LGTM! As an aside- the top of that page feels so empty🙁

kelbyhawn:
> LGTM! As an aside- the top of that page feels so empty🙁

I know 🥲 I think the banner is so sparse because it'll have tabs, and once I fix the spacing between the elements it'll get a little better.

PR #58072, opened 2024-04-17
Add speculative fix for font issues in resource PDFs behind DCDO flag 

kobryan0619:
This PR is intended to fix the missing font issues we have had with Curriculum PDFs getting generated, specifically with resource PDFs.

Since we have struggled to recreate the issue on machines outside of staging, this PR includes a DCDO flag to be used on Staging to test this speculative fix. 

## Context

After we updated `puppeteer` to generate PDFs of resources in the curriculum, we noticed that when these individual resources were combined into a single "Resources" PDF, we would get missing fonts in combined PDFs.  This showed up both in the "cover pages" that were between lessons as well as within lessons.  Below are some examples of what that looked like:

```
GPL Ghostscript 9.50: Missing glyph CID=90, glyph=005a in the font AAAAAA+LiberationSans-Bold . The output PDF may fail with some viewers.
GPL Ghostscript 9.50: Missing glyph CID=212, glyph=00d4 in the font DAAAAA+Consolas . The output PDF may fail with some viewers.
```

This was consisitent with these two fonts and would seem to happen most frequently with CSA units which notably used the `consolas` font more frequently than other curriculums.   The impact to students and teachers was that their resources would at times be unreadable because key characters would show up as boxes in the text.   See image below for what this would look like.

image

[This spreadsheet](https://docs.google.com/spreadsheets/d/1A1P15815F9HgP-nrT-bLvfHwMnLwOvdgGg0kqscNBuU/edit?gid=0#gid=0) took a snapshot of different units and what warnings popped up.

While investigating, we saw that on my local machine, these errors would not be generated, however on `staging` these errors were consistently popping up.  We tried creating an adhoc to generate PDFs, but again, we couldn't recreate the issue that `staging` had.  Elijah was able to recreate the issue on his local machine.

We initially thought it might be a font issue with different fonts available on different machines, but again, that research yielded no difference between the machines. 

Through some google research, we noticed that this was an issue others reported as well.  This [Stack Overflow ](https://stackoverflow.com/questions/52350486/rename-pdf-embedded-font/52352156#52352156)response yielded an idea that we are trying in this PR.  The short summary of this stack overflow answer is that when individual PDFs are created, there is a subset of a font that is embedded into that document and when you combine pdfs using ghostscript there are conflicts between those subsets which causes the issue of missing fonts later on. 

## Future work

- As of right now, this PR simply fixes curriculum resources by re-generating PDFs using `ghostscript`.  There is a world in which we update how the PDFs are initially generated in the first place so that we are not "re-generating" the exact same PDF twice.  However, before making this change, we want to test that indeed using `ghostscript` will fix our core issue.


dashboard/lib/services/curriculum_pdfs/resources.rb
```diff
@@ -155,6 +155,15 @@ def generate_lesson_resources_title_page(lesson, directory = "/tmp/")
           end
           raise "File #{path.inspect} does not exist after generation" unless File.exist?(path)
 
+          if DCDO.get('use-ghostscript-to-generate-pdfs', false)
```
kobryan0619, line 158: This is for the cover pages between the lessons which have had issues with missing fonts.

dashboard/lib/services/curriculum_pdfs/resources.rb
```diff
@@ -179,16 +188,28 @@ def fetch_url_to_path(url, path)
...
             service.get_file(file_id, download_dest: path)
-            return path
           elsif url.end_with?(".pdf")
             IO.copy_stream(URI.parse(url)&.open, path)
-            return path
+          end
+
+          if File.exist?(path)
+            # Regenerate the PDF using ghostscript
+            if DCDO.get('use-ghostscript-to-generate-pdfs', false)
```
kobryan0619, line 206: This is for the resources themselves.

lib/dynamic_config/dcdo.rb
```diff
@@ -56,6 +56,7 @@ def frontend_config
       'student-email-post-enabled': DCDO.get('student-email-post-enabled', false),
       'progress-v2-metadata-enabled': DCDO.get('progress-v2-metadata-enabled', false),
       'show-updated-lms-content': DCDO.get('show-updated-lms-content', false),
+      'use-ghostscript-to-generate-pdfs': DCDO.get('use-ghostscript-to-generate-pdfs', false),
```
Hamms: I don't believe we need to add this flag to the subset of DCDO flags that get read by clientside code, since we're only ever reading it serverside.

dashboard/lib/services/curriculum_pdfs/resources.rb
```diff
@@ -176,19 +185,37 @@ def fetch_url_to_path(url, path)
...
+            pdf_generated_initially = true
+          end
+
+          if pdf_generated_initially
+            # Regenerate the PDF using ghostscript
+            if DCDO.get('use-ghostscript-to-generate-pdfs', false)
+              new_path = File.join(File.dirname(path), "optimized_#{File.basename(path)}")
+              gs_command = "gs -sDEVICE=pdfwrite -dNOPAUSE -dBATCH -dQUIET -sOutputFile=#{new_path} #{path}"
+              system(gs_command)
+              puts "I regenerated #{path} to #{new_path} using Ghostscript."
```
Hamms: nit: looks like some stray `puts`es, here and below

dashboard/lib/services/curriculum_pdfs/resources.rb
```diff
@@ -155,6 +155,15 @@ def generate_lesson_resources_title_page(lesson, directory = "/tmp/")
           end
           raise "File #{path.inspect} does not exist after generation" unless File.exist?(path)
 
+          if DCDO.get('use-ghostscript-to-generate-pdfs', false)
+            # Regenerate the PDF using Ghostscript
+            optimized_path = File.join(directory, "optimized_#{filename}")
+            gs_command = "gs -q -sDEVICE=pdfwrite -dNOPAUSE -dBATCH -sOutputFile=#{optimized_path} #{path}"
+            system(gs_command)
+            return optimized_path
```
Hamms: I get that we want to keep this fix scoped to just the curriculum pdfs for now, but I'd still like to keep all of our manual system invocations of `gs` isolated to [`collate.rb`](https://github.com/code-dot-org/code-dot-org/blob/staging/lib/pdf/collate.rb). Can we add a new helper method to that class that can be called here? Something like `PDF.regenerate_with_ghostscript`, `PDF.regenerate_embedded_font_names`, or even just `PDF.normalize_for_merging`. We can then add a nice verbose comment to that method explaining all the context we uncovered about why this is necessary!

dashboard/lib/services/curriculum_pdfs/resources.rb
```diff
@@ -176,19 +185,37 @@ def fetch_url_to_path(url, path)
...
+            if DCDO.get('use-ghostscript-to-generate-pdfs', false)
+              new_path = File.join(File.dirname(path), "optimized_#{File.basename(path)}")
+              gs_command = "gs -sDEVICE=pdfwrite -dNOPAUSE -dBATCH -dQUIET -sOutputFile=#{new_path} #{path}"
+              system(gs_command)
+              puts "I regenerated #{path} to #{new_path} using Ghostscript."
+              return new_path
+            else
+              # return the path generated by gdrive service
+              puts "I did not regenerate #{path} using Ghostscript."
+              return path
```
Hamms, line 211: I do want to note that this logic is slightly different now; previously, if a given url did not pass any of the three checks, we did not return anything. Now, we do. I'm not sure if we expect that to ever happen or if that's a case we care about supporting, but I want to call it out anyway in case we do.

What if instead of adding the `pdf_generated_initially` flag, we just extract the regeneration logic into a method (as recommended above) and update the three `return path` calls to something like `return PDF.regenerate_embedded_font_names(path)`?

Hamms (commented):
Have a few nits and suggestions about minor implementation details, but the broader plan of regenerating PDFs with ghostscript and initially doing so behind a DCDO flag for testing purposes sounds good to me!

I would definitely like to see a short writeup either in the PR description or code comments (or both!) summarizing what we discovered about the font embeddings and probably linking to that stackoverflow post where we found the suggestion to try passing individual pdfs through gs before merging. Just to make sure that any future developers discovering this changeset can easily figure out why we decided on this approach.

Thanks!

lib/dynamic_config/dcdo.rb
kobryan0619: Hmm... I did a bit more digging on this to better understand what you're suggesting.  Paraphrasing what I believe to be true, give me a thumbs up if I have this right:

1. This code needed to configure DCDO flags in the front end / javascript
2. Since we will be only doing `DCDO.get` in back end code (ruby code), we don't need to add this DCDO flag here.
3. We will be able to still use DCDO flags when we merge this code into staging by doing `DCDO.set('use-ghostscript-to-generate-pdfs', true)` which will both create the flag, and set it to true.  The `ruby` file will be able to access this flag at this point.
4. Until that point, something like `DCDO.get('use-ghostscript-to-generate-pdfs', false)` will always return `false` because of the second parameter. 

That's the story I have woven together with some digging and context clues - let me know if that's all right.

dashboard/lib/services/curriculum_pdfs/resources.rb
kobryan0619: Those were somewhat intentionally left in as debugging helpers - they are probably part security blanket supports for me, but I can take them out :)

dashboard/lib/services/curriculum_pdfs/resources.rb
kobryan0619: I actually think the logic is the same because of the `pdf_generated_initially` flag.  The flag is set to `false` initially, and only ever gets switched back to `true` if a PDF is created. 

Then, this else statement only gets called if `pdf_generated_initially` is `true` AND if the `DCDO` flag is false. 

If `pdf_generated_initially` is never set to `true` the new PDFs wouldn't be re-generated.  

I looked into perhaps moving the "check for DCDO flag" logic into the `collate.rb` method, but that didn't make sense because if the DCDO flag was false, it wouldn't regenerate anything... so if you keep the DCDO flag in this file, you'd need to add it to each of the three paths here too...

For those reasons, and assuming my logic is right that this is the same outcome as before, I think it makes sense to keep this logic in this place with the `pdf_generated_initially` flag.

Hamms: Ohh, I misread the nested `if`s, and thought the `return path` else clause was attached to `if pdf_generated_initially`, not the DCDO check. Agreed that this is consistent with the prior logic, and that it's definitely cleaner to implement with a flag than by adding the DCDO check to each of the three cases. Thanks for the thorough analysis!

lib/pdf/collate.rb
```diff
@@ -75,6 +75,17 @@ def self.merge_local_pdfs(output_file, *filenames)
     end
   end
 
+  # PDFs generated by puppeteer and then combined using ghostscript tended to have
+  # missing fonts which would make them unreadable in some PDF viewers.  This issue
```
Hamms, line 79: I'd go as far as to specify here that puppeteer generates PDFs with embedded font names that can be the same between different PDFs, leading to problems when we try to merge the resulting PDFs. Whereas ghostscript adds a unique hash to each embedding, allowing for a seamless merge.

I also tracked down the SO answer that led us to this conclusion; probably worth linking https://stackoverflow.com/questions/52350486/rename-pdf-embedded-font/52352156#52352156

lib/dynamic_config/dcdo.rb
Hamms: Yep, that's completely correct! Nice digging :smile: 

For some additional context; DCDO was originally exclusively serverside, and is still intended to be mostly used that way; the `frontend_config` functionality is an optional add-on.

dashboard/lib/services/curriculum_pdfs/resources.rb
Hamms: Yeah, they're definitely nice to have for local testing! Slightly less useful when running on a remote server, though, and possibly quite noisy. We can always put em back if we have to debug something, but hopefully that won't be necessary :crossed_fingers:

Hamms (approved):
I'd like to see a tiny bit more information in one comment, to hopefully make things a little bit clearer for future developers. But everything else looks good to me; thanks for all the hard work and deep-dive investigation on this one!

PR #57947, opened 2024-04-10
Move Instructed PL Sections to My PL page

TurnerRiley:
Moves the Professional Learning sections the user is teaching from studio.code.org/home to [studio.code.org/my-professional-learning](https://studio.code.org/my-professional-learning).

Following the spec, this PR should:
- Show the table of the PL sections the user is teaching
   - Show it only for facilitators (under the "My Facilitator Center" tab), universal instructors (under the "Instructors" tab), and peer reviewers (under the "Instructors" tab)
   - Show the "Create Section" component
- Show the co-teacher invites banners for professional learning sections that were on the Teacher Dashboard
- Remove the PL sections table from the Teacher Dashboard and replace with a short message with a redirect to the My PL page

### Note about recreating the Owned Sections Table
Unfortunately, there are a lot of different table setups used across the Teacher Dashboard and the My PL page so it was not as simple as moving the `OwnedSectionsTable` component over because the styling was pretty off as the My PL page's styles, the imported styles from `tableConstants.js`, and the in-file styles within a `styles` object were all blending together. I figured taking the time to refactor `OwnedSectionsTable` to maintain its current styles on the Teacher Dashboard but have different functionality and styling on the My PL page would be scope creep. So, my solution was to make a similar `OwnedPlSectionsTable` that is the same content copied over but with anything unnecessary removed. I also consolidated the style updates for this in the `tableConstants.js` file so that it wasn't as confusing to track. If this seems like a concern I'm open to pushback and/or can make a ticket for building out this table following [the way Kelby did](https://github.com/code-dot-org/code-dot-org/pull/57811) with the `SelfPacedProgressTable` so that the styles can be better reused.

### View of page by permission
Facilitator:
(note: this Facilitator also has a non-PL section to demonstrate that this table only shows PL sections (both active and archived))
Facilitator

Universal instructor:
universalinstructor

Peer reviewer:
peer_reviewer

Teacher (should not be able to see it):
Teacherview

### Co-teacher invite
coteacherinvite

### Teacher Dashboard message where table used to be
teacherdashboard

## Links
Jira ticket: [here](https://codedotorg.atlassian.net/jira/software/c/projects/ACQ/boards/64?assignee=60d62161f65054006980bd71&selectedIssue=ACQ-1426)
Spec: [here](https://docs.google.com/document/d/1Ee02kw_ei2x6wOS06K768dC-kUhGDlydjoE6BZipbPE/edit)
[...]

PR #58075, opened 2024-04-17
Cleanup OwnedSections component

TurnerRiley:
As part of moving the Owned Professional Learning Sections table from the Teacher Homepage to the My PL page in [this PR](https://github.com/code-dot-org/code-dot-org/pull/57947), `OwnedSectionsTable` had to be split into `OwnedSectionsTable` and `OwnedPlSectionsTable` so that styling could be consistent for each on their respective pages. This then left some cleanup work:
- Remove the `isPlSections` parameter to the tables
- Add a test file for `OwnedPlSectionsTable`
- While I was in here, I converted both `OwnedSectionsTableTest.js` and `OwnedPlSectionsTableTest.js` into React testing library
   - Both files run nearly identical tests, I just kept the test section data in its original file and moved the PL section data over to the pl table test file

## Links
Jira ticket: [here](https://codedotorg.atlassian.net/jira/software/c/projects/ACQ/boards/64?assignee=60d62161f65054006980bd71&selectedIssue=ACQ-1794)
PR this follows up: [here](https://github.com/code-dot-org/code-dot-org/pull/57947)

## Testing story
Updating unit tests


apps/test/unit/templates/teacherDashboard/OwnedSectionsTableTest.js
```diff
@@ -260,231 +108,173 @@ describe('OwnedSectionsTable Sorting', () => {
...
-      const text = courseLinkCol.find(Button).prop('text');
-      assert.equal(button, '<Button />');
-      assert.equal(link, '/catalog');
-      assert.equal(text, 'Find a course');
-      expect(
-        courseLinkCol.find({
-          className: skeletonizeContent.skeletonizeContent,
-        })
-      ).to.have.lengthOf(0);
+  it('courseLinkFormatter provides links to course information and section information', () => {
```
hannahbergam, line 209: This test covers a lot! Non-blocking, but it feels like the format of looping through all sections and then checking that 'at least one of them' covers each of a few different states is slightly confusing and not quite future proof. I understand why you combined the concept of checking for assignmentPath counts together, but then it makes it hard to tell which section should have which count. Would it be possible to break this up?

apps/test/unit/templates/teacherDashboard/OwnedSectionsTableTest.js
```diff
@@ -260,231 +108,173 @@ describe('OwnedSectionsTable Sorting', () => {
...
-    const wrapper = shallow(
-      <OwnedSectionsTable
-        sectionIds={[1]}
-        sectionRows={sectionRowData.slice(0, 1)}
-        onEdit={() => {}}
-      />
-    );
-    const header = wrapper.find(Table.Header);
-    assert.equal(header.length, 1);
+  it('can be sorted correctly by section name', () => {
```
hannahbergam, line 137: Love these simplified sort tests!

apps/test/unit/templates/teacherDashboard/OwnedPlSectionsTableTest.js
```diff
@@ -0,0 +1,263 @@
...
+import {assert, expect} from '../../../util/reconfiguredChai';
+
+const plSectionRowData = [
+  {
+    id: 1,
+    name: 'first section',
+    studentCount: 3,
+    code: 'ABC',
+    courseOfferingsAreLoaded: true,
+    grades: ['K'],
```
hannahbergam: This is a very small thing, but do we have grades for PL sections?

hannahbergam (commented):
Wow this was a ton of work! A couple questions/nits but great job!

apps/test/unit/templates/teacherDashboard/OwnedPlSectionsTableTest.js
bethanyaconnor: no, actually, PL sections can [only have the grade](https://github.com/code-dot-org/code-dot-org/blob/0659952954efa4569b999231004d9117cb444455/dashboard/app/models/sections/section.rb#L121) `pl`

apps/test/unit/templates/teacherDashboard/OwnedPlSectionsTableTest.js
```diff
@@ -0,0 +1,263 @@
...
+    loginType: SectionLoginType.picture,
```
bethanyaconnor: Relatedly, they can only be email logins as well!

bethanyaconnor (approved):
Thanks for following up and cleaning this up!!

apps/test/unit/templates/teacherDashboard/OwnedSectionsTableTest.js
TurnerRiley: Yes for sure! I'll simplify the loop tests so they're more readable and updatable in the future

apps/test/unit/templates/teacherDashboard/OwnedPlSectionsTableTest.js
TurnerRiley: Oh good point! I had just copied these over from the test file and didn't notice, I'll clean that up!

apps/test/unit/templates/teacherDashboard/OwnedPlSectionsTableTest.js
TurnerRiley: Ah gotcha, I'll fix that! Thanks for spotting that!

PR #58086, opened 2024-04-18
GenAI: Adding the model selection to levelbuilder editor. 

cnbrenci:
## Warning!!

The [AP CSP Create Performance Task](https://apcentral.collegeboard.org/courses/ap-computer-science-principles/exam) is in progress. The most critical dates are from April 3 - April 30, 2024. Please consider any risk introduced by this PR that could affect our students taking AP CSP. Code.org students taking AP CSP primarily use App Lab for their Create Task, however a small percent use Game Lab. Carefully consider whether your change has any risk of altering, changing, or breaking anything in these two labs. Even small changes, such as a different button color, are considered significant during this time period. Reach out to the Student Learning team or Curriculum team for more details.

## Description

 - Added an additional section to the level editor for ai chat levels that includes selecting an initial value for the model to be used in the level as well as setting which models will be available to choose from and compare in the level.
 - Updated modelDescription for minstrel with example text provided by Sam.
 - Updated PresentationView to include overview in the technical details.
 - Updated UI to hide the Selected model dropdown and Compare Models button if visibility set to hidden
 - Updated UI to disable the Selected model dropdown and hide the Compare Models button if visibility is set to readonly.

Requirements fulfilled with this PR: 
 - Levelbuilders are able to set a default model for a level.
 - Levelbuilders are able to define which models students can pick from (as a list of checkboxes in levelbuilder)
 - Levelbuilders are able to hide the compare models dialog in a level.

## Links

- jira ticket: [LABS-644](https://codedotorg.atlassian.net/browse/LABS-644)

## Testing story

Tested locally - video shows the levelbuilder UI setting the correct state in the LevelAichatSettings. 

https://github.com/code-dot-org/code-dot-org/assets/24951544/f35b912b-0575-4fb3-a5ba-549459d55f8d

SetupCustomization with visibility set to READONLY:
image

SetupCustomization with visibility set to HIDDEN:
image

SetupCustomization with visibility set to EDITABLE:
image

SetupCustomization with visibility set to EDITABLE but with only one model selected (no additional models added):
image
image

I loaded up a few existing aichat levels with my changes and found that they defaulted correctly to disabling the model selection dropdown and hiding the compare models button:
image

## Follow-up work

In levelbuilder, if you select an additional model and then switch the visibility to read-only, the model stays selected and gets saved as an `additionalAvailableModel`.   Clearing the state of `additionalAvailableModels` in that case would be a bit of a refactor because I believe I'd need to lift that state up and pass through the setter as a callback.


apps/src/lab2/levelEditors/aichatSettings/EditAichatSettings.tsx
```diff
@@ -88,13 +89,30 @@ const EditAichatSettings: React.FunctionComponent<{
...
   );
 
+  const setModelSelectionValues = useCallback(
+    (availableModelIds: string[], selectedModelId: string = '') => {
+      setAichatSettings({
+        ...aichatSettings,
+        availableModelIds: availableModelIds,
+        initialCustomizations: {
+          ...aichatSettings.initialCustomizations,
+          ['selectedModelId']:
```
thomasoniii: the key should probably just be `selectedModelId`

apps/src/lab2/levelEditors/aichatSettings/ModelSelectionFields.tsx
```diff
@@ -0,0 +1,117 @@
...
+import {UpdateContext} from './UpdateContext';
+import FieldSection from './FieldSection';
+import {Visibility} from '@cdo/apps/aichat/types';
+
+const unionModelIds = (
+  additionalIds: Set<string>,
+  selectedId: string
+): string[] => {
+  return Array.from(new Set(additionalIds).add(selectedId));
+};
```
thomasoniii: Don't change this, but FWIW, there is a `Set.union` method as well. We probably shouldn't use it yet, since it's newly added to Chrome 122.

That could be:
```
Array.from(additionalIds.union(new Set(selectedId)))
```

apps/src/lab2/levelEditors/aichatSettings/ModelSelectionFields.tsx
```diff
@@ -0,0 +1,117 @@
...
+            labelText=""
+            onChange={e => {
+              setModelSelectionValues(
+                unionModelIds(additionalAvailableModelIds, e.target.value),
+                e.target.value
+              );
+            }}
+            items={modelDescriptions.map(model => {
+              return {value: model.id, text: model.name};
+            })}
```
thomasoniii: Just noting that both the `onChange` and `items` lists aren't memoized.

They probably should be on principle, but I dunno how urgent it is to actually do it atm.

apps/src/lab2/levelEditors/aichatSettings/ModelSelectionFields.tsx
```diff
@@ -0,0 +1,117 @@
...
+                      );
+                      setModelSelectionValues(
+                        unionModelIds(
+                          additionalAvailableModelIds,
+                          selectedModelId
+                        ),
+                        selectedModelId
+                      );
+                    }
+                  }}
```
thomasoniii: ditto as above - not memoized

thomasoniii (approved):
couple of minor notes (though the noted callbacks probably should be memoized), but otherwise lgtm 🚀

dashboard/config/levels/custom/aichat/AI Chat Test Level - 2.level
```diff
@@ -12,11 +12,46 @@
...
+
+          ]
+        },
+        "visibility": "editable"
+      },
+      "hidePresentationPanel": true
+    },
     "aichat_settings": {
       "initialCustomizations": {
+        "selectedModelId": "chatgpt",
```
bencodeorg: Do we handle levels that don't have this configuration in them, or do we need to update them all?

Separately, notice that we're writing the old style of initial customizations above:

```
      "botName": {
        "value": "",
        "visibility": "editable"
      },
      "temperature": {
        "value": 0.5,
        "visibility": "editable"
      },
```

But I get that too when I try to make an edit (guessing it's bc we have the old config in our database entries for these levels). So maybe fine for now.

apps/src/lab2/levelEditors/aichatSettings/UpdateContext.ts
```diff
@@ -20,4 +20,8 @@ export const UpdateContext = createContext({
     property: keyof ModelCardInfo,
     value: ModelCardInfo[keyof ModelCardInfo]
   ) => {},
+  setModelSelectionValues: (
+    availableModelIds: string[],
+    selectedModelId: string = ''
```
bencodeorg: context newbie...what does this default value do?

apps/src/lab2/levelEditors/aichatSettings/ModelSelectionFields.tsx
bencodeorg: +1, I think would help with readability too to move this elsewhere (and memoize)

apps/src/lab2/levelEditors/aichatSettings/ModelSelectionFields.tsx
```diff
@@ -0,0 +1,117 @@
...
+      labelText="Model Selection"
+      description="The initial model selected for the chatbot."
+      inputType="custom"
+      inputNode={
+        <>
+          <SimpleDropdown
+            labelText=""
+            onChange={e => {
+              setModelSelectionValues(
+                unionModelIds(additionalAvailableModelIds, e.target.value),
```
bencodeorg: Do we need to union these each time before calling `setModelSelectionValues`? Or could we just pass in the available and selected to `setModelSelectionValues`, union them inside that function, then set state appropriately?

bencodeorg (commented):
Looks great! Appreciate the quality of the levelbuilder experience here. A couple small questions, and I'm curious how this plays with existing levels if we don't update their level config?

apps/src/lab2/levelEditors/aichatSettings/ModelSelectionFields.tsx
cnbrenci: Oh good call, that would be more DRY.

apps/src/lab2/levelEditors/aichatSettings/UpdateContext.ts
cnbrenci: I'm not sure if I need it in both places, but this was a default for that parameter. I actually meant to remove that.

cnbrenci: Oh and this is just a default value so they're empty functions I guess.

dashboard/config/levels/custom/aichat/AI Chat Test Level - 2.level
fisher-alice: If I manually clean file, run `bundle exec rake seed:custom_levels`, then edit the file via levelbuilder UI, the old config didn't reappear for me.
Either way, we should probably clean up the old config.`

fisher-alice: > Do we handle levels that don't have this configuration in them, or do we need to update them all?

I believe that the following handles:
https://github.com/code-dot-org/code-dot-org/blob/53f6f88e0de265636288548606ca077a3969d41d/apps/src/aichat/views/modelCustomization/SetupCustomization.tsx#L42-L47

There is a weird case for an existing .level file that if these are not defined, then the compare models button doesn't appear. But for new `aichat` levels, I don't think this should be an issue.

dashboard/config/levels/custom/aichat/AI Chat Test Level.level
```diff
@@ -12,8 +12,42 @@
...
+
+          ]
+        },
+        "visibility": "editable"
+      },
+      "hidePresentationPanel": false
+    },
     "aichat_settings": {
       "initialCustomizations": {
+        "selectedModelId": "minstral",
```
fisher-alice: Spelling is 'mistral' [ref](https://huggingface.co/docs/transformers/main/en/model_doc/mistral).

apps/src/lab2/levelEditors/aichatSettings/EditAichatSettings.tsx
```diff
@@ -88,13 +89,30 @@ const EditAichatSettings: React.FunctionComponent<{
     [aichatSettings, setAichatSettings, initialCustomizations]
   );
 
+  const setModelSelectionValues = useCallback(
```
fisher-alice, line 92: Not a strong feeling, but what do you think about separating this function into two different functions - one to update the `selectedModelId` and the other to update the `availableModelIds`? I think it may make things a bit easier to read where these functions are called. Also, one property is included in `initialCustomizations` and the other property is not.

cnbrenci: I thought so too, and I was originally using setPropertyValue for the selectedModelId in initialCustomizations, but since these are still in the same piece of state, they weren't reliably updating correctly, so I had to combine them to set both values at the same time. My next commit will make the code a bit simpler tho at least.

apps/src/lab2/levelEditors/aichatSettings/ModelSelectionFields.tsx
cnbrenci: Just wrapping my head around these hooks. Super helpful to point these out. Thank you!!

apps/src/lab2/levelEditors/aichatSettings/ModelSelectionFields.tsx
cnbrenci: Pulled the onChange functions and items calculation up above the return, and I think I have the dependencies and stuff correct. LMK if it looks OK!

dashboard/config/levels/custom/aichat/AI Chat Test Level - 2.level
cnbrenci: I think they should work with the defaulting behavior, but I'm not 100% confident there aren't cases where the state of levels might be in weird states on different folks' machines? 

Thanks @fisher-alice for answering this and the extra info!! I want to understand better the process you're going through to manually clean the file and re-seed and stuff. Will we need to go and do something similar in each environment now, if we're done updating the level config? Could you walk me through what you did? Thanks!!

dashboard/config/levels/custom/aichat/AI Chat Test Level.level
cnbrenci: OOPS thanks for catching that!!

apps/src/lab2/levelEditors/aichatSettings/EditAichatSettings.tsx
cnbrenci: Is latest what you meant?
https://github.com/code-dot-org/code-dot-org/blob/3d5167a780b9b1c907541e65b4f19616080cf330/apps/src/lab2/levelEditors/aichatSettings/EditAichatSettings.tsx#L92-L107

apps/src/lab2/levelEditors/aichatSettings/ModelSelectionFields.tsx
```diff
@@ -0,0 +1,107 @@
...
+    useState<Set<string>>(
+      new Set(
+        aichatSettings.availableModelIds.filter(
+          id => id !== aichatSettings.initialCustomizations.selectedModelId
+        )
+      )
+    );
+  const selectedModelId = aichatSettings.initialCustomizations.selectedModelId;
+
+  const modelDropdownItems = useMemo(() => {
```
bencodeorg: hah I also got twisted up in React hooks and at second glance I think this could just be a const outside of the React component, it can't really change yeah?

apps/src/lab2/levelEditors/aichatSettings/EditAichatSettings.tsx
```diff
@@ -88,13 +89,31 @@ const EditAichatSettings: React.FunctionComponent<{
...
+    (additionalModelIds: Set<string>, selectedModelId: string) => {
+      const availableModelIds = Array.from(
```
bencodeorg, line 94: Nice :)

dashboard/config/levels/custom/aichat/AI Chat Test Level - 2.level
bencodeorg: > There is a weird case for an existing .level file that if these are not defined, then the compare models button doesn't appear. But for new aichat levels, I don't think this should be an issue.

> I think they should work with the defaulting behavior, but I'm not 100% confident there aren't cases where the state of levels might be in weird states on different folks' machines?

Yeah I'm not too worried about breaking people's local setups, but I know Dan made a level progression this week, so I think it'd be worth figuring out if this is going to break it (ie, hide compare models button)? It probably wouldn't be too hard to just manually update the level config on that handful of files.

bencodeorg (approved):
Code looks good to me! I do think it's worth checking out Dan's new level progression and see how this change affects it before merging, and/or if there are any manual level config changes we should make to support it.

apps/src/lab2/levelEditors/aichatSettings/EditAichatSettings.tsx
cnbrenci: Yeah, way better, thanks for pointing it out :-D

apps/src/lab2/levelEditors/aichatSettings/ModelSelectionFields.tsx
cnbrenci: Yeah, this shouldn't ever change since the only dependency is the modelDescriptions which is imported from consts

cnbrenci: OK, moved outside the component!

dashboard/config/levels/custom/aichat/AI Chat Test Level - 2.level
cnbrenci: To test Dan's level progression, lmk if these are the right steps to take or if there's an easier way :) - 
I should go find the .level files from the content scoops, grab the files and put them in my local, and then seed those scripts into my db, and try to run them?

cnbrenci: er, I can run seed:all and it will seed everything that exists in the repo to my local db which will include the new level progression, as long as I've merged in the latest staging that has the new levels??

fisher-alice: >er, I can run seed:all and it will seed everything that exists in the repo to my local db which will include the new level progression, as long as I've merged in the latest staging that has the new levels??

Yes - that would do it! Thanks for checking this!

apps/src/lab2/levelEditors/aichatSettings/EditAichatSettings.tsx
fisher-alice: Oh interesting. I actually had a similar issue with updating back when working on edubot project. And thanks for iterating on making code more simple!

apps/src/aichat/views/modelCustomization/SetupCustomization.tsx
```diff
@@ -73,16 +76,19 @@ const SetupCustomization: React.FunctionComponent = () => {
           name="model"
           size="s"
           className={styles.selectedModelDropdown}
+          disabled={isDisabled(selectedModelId)}
```
fisher-alice, line 97: Outside of scope of this PR - asked a question [in Slack](https://codedotorg.slack.com/archives/C06FELVTV0Q/p1713795342004929) to Design about contrast of read-only text.

apps/src/aichat/views/presentation/PresentationView.tsx
```diff
@@ -17,8 +17,11 @@ const PresentationView: React.FunctionComponent = () => {
   const {selectedModelId, systemPrompt, temperature, retrievalContexts} =
     savedAiCustomizations;
   const modelCardInfo = savedAiCustomizations.modelCardInfo;
-  const {name: modelName = '', trainingData = ''} =
-    modelDescriptions.find(model => model.id === selectedModelId) ?? {};
+  const {
+    name: modelName = '',
+    trainingData = '',
+    overview = '',
+  } = modelDescriptions.find(model => model.id === selectedModelId) ?? {};
```
fisher-alice, line 24: Interesting syntax!

apps/src/lab2/levelEditors/aichatSettings/ModelSelectionFields.tsx
```diff
@@ -0,0 +1,105 @@
...
+          </BodyFourText>
+          {modelDescriptions.map(model => {
+            return (
+              <div key={model.id} className={moduleStyles.fieldRow}>
+                <Checkbox
+                  key={model.id}
+                  name={model.id}
+                  label={model.name}
+                  checked={
+                    additionalAvailableModelIds.has(model.id) ||
```
fisher-alice: Nit: If `additionalAvailableModelIds` includes a model that is not the selected model, then 'read only' is selected, the non-selected model remains checked.
Could add to condition:
```
(additionalAvailableModelIds.has(model.id) &&
  !shouldDisableAdditionalModelSelection) ||
  model.id === selectedModelId
```

dashboard/config/levels/custom/aichat/Ai Chat Test Level - 3.level
```diff
@@ -0,0 +1,45 @@
+<Aichat>
```
fisher-alice: Planning to add this file to a script?

apps/src/lab2/levelEditors/aichatSettings/ModelSelectionFields.tsx
```diff
@@ -0,0 +1,105 @@
...
+  const onCheckboxChange = useCallback(
+    (e: React.ChangeEvent<HTMLInputElement>) => {
+      if (e.target.checked) {
+        setAdditionalAvailableModelIds(
+          additionalAvailableModelIds.add(e.target.name)
+        );
+        setModelSelectionValues(additionalAvailableModelIds, selectedModelId);
+      } else {
+        additionalAvailableModelIds.delete(e.target.name);
+        setAdditionalAvailableModelIds(additionalAvailableModelIds);
```
fisher-alice: Nit - curious why you don't use line 46 as a parameter for line 47 as you did for when checkbox is checked above.

fisher-alice (approved):
Love the refactor - much simpler to read/understand. Left small questions/comments. Looks great! 🎉

apps/src/lab2/levelEditors/aichatSettings/ModelSelectionFields.tsx
cnbrenci: `add()` modifies the `Set` and returns the `Set`, so I'm passing in the return val of the `add` operation, but `delete()` modifies the `Set` and returns a `bool` whether the value was in the `Set` or not. So I can't pass in the return val. Gotta modify first and then pass the modified `Set`

apps/src/lab2/levelEditors/aichatSettings/ModelSelectionFields.tsx
```diff
@@ -0,0 +1,105 @@
...
+    aichatSettings.visibilities.selectedModelId !== Visibility.EDITABLE;
+  const [additionalAvailableModelIds, setAdditionalAvailableModelIds] =
+    useState<Set<string>>(
+      new Set(
+        aichatSettings.availableModelIds.filter(
+          id => id !== aichatSettings.initialCustomizations.selectedModelId
+        )
+      )
+    );
+  const selectedModelId = aichatSettings.initialCustomizations.selectedModelId;
```
sanchitmalhotra126: nit: assign this const earlier to reuse in the few assignments above?

apps/src/lab2/levelEditors/aichatSettings/ModelSelectionFields.tsx
```diff
@@ -0,0 +1,105 @@
...
+  return {value: model.id, text: model.name};
+});
+
+const ModelSelectionFields: React.FunctionComponent = () => {
+  const {setModelSelectionValues, aichatSettings} = useContext(UpdateContext);
+  const shouldDisableAdditionalModelSelection =
+    aichatSettings.visibilities.selectedModelId !== Visibility.EDITABLE;
+  const [additionalAvailableModelIds, setAdditionalAvailableModelIds] =
+    useState<Set<string>>(
+      new Set(
```
sanchitmalhotra126: Would it be easier to just work with an array here? Realistically i don't think we'd have duplicated IDs right? And if we did I think it would cause a whole lot of problems elsewhere too

dashboard/config/levels/custom/aichat/AI Chat Test Level - 2.level
sanchitmalhotra126: Yeah just to +1, we should make sure to reseed our DBs before making level update changes as we're actively iterating on the level properties structure (shouldn't be as much of an issue once the structure is finalized)

apps/src/lab2/levelEditors/aichatSettings/EditAichatSettings.tsx
```diff
@@ -88,13 +89,31 @@ const EditAichatSettings: React.FunctionComponent<{
     [aichatSettings, setAichatSettings, initialCustomizations]
   );
 
+  const setModelSelectionValues = useCallback(
+    (additionalModelIds: Set<string>, selectedModelId: string) => {
```
sanchitmalhotra126: Similar to the previous comment - can `additionalModelIds` just be a string array? Looks like we create a new `Set` out of it anyway which should remove duplicates

apps/src/lab2/levelEditors/aichatSettings/ModelSelectionFields.tsx
cnbrenci: Hmm. Yeah the reason I went with a Set originally was because we could have duplicates when we merge together the selectedModelId and the additional ones, but I am actually copying the set when I do that anyway, so I could just use an array for this. good call.

apps/src/lab2/levelEditors/aichatSettings/ModelSelectionFields.tsx
cnbrenci: I fixed this in a different way so we don't save unnecessary stuff in the aiCustomizations in that case. See line 27-33

cnbrenci: I'm actually re-thinking this because it seems it's an antipattern to reset a component's state based on props

cnbrenci: Yeah I think I'm going to punt this one and log a follow-up bug to be triaged. I think I'd need a bigger refactor to move the `additionalAvailableModelIds` up to the parent component, because when I try to reset the array based on `shouldDisableAdditionalModelSelection` I get React errors because it has to re-render while it's already re-rendering.  I don't want to just hide them in the way that you suggested, because it's going to save that list of values even if it's not showing it, and I'd rather show an honest view of what will be saved.

fisher-alice: SGTM! Thanks for thinking this through.

apps/src/aichat/views/modelCustomization/SetupCustomization.tsx
```diff
@@ -11,41 +11,58 @@ import {
...
-  const availableModelIds =
-    useAppSelector(
-      state =>
-        (state.lab.levelProperties as AichatLevelProperties | undefined)
-          ?.aichatSettings?.availableModelIds
-    ) || [];
-  const availableModels = availableModelIds.length
-    ? modelDescriptions.filter(model => availableModelIds.includes(model.id))
-    : modelDescriptions;
+  const savedModelIds = useAppSelector(
```
bencodeorg: Nit: any reason not to keep this named as `availableModelIds`?

apps/src/aichat/views/modelCustomization/SetupCustomization.tsx
```diff
@@ -11,41 +11,58 @@ import {
...
 const SetupCustomization: React.FunctionComponent = () => {
   const dispatch = useAppDispatch();
 
   const [isShowingModelDialog, setIsShowingModelDialog] =
     useState<boolean>(false);
 
-  const {temperature, systemPrompt, selectedModelId} = useAppSelector(
-    state => state.aichat.fieldVisibilities
-  );
+  const {
```
bencodeorg, line 31: Maybe a comment here about why we have this default?

bencodeorg (approved):
lgtm!

apps/src/aichat/views/presentation/PresentationView.tsx
sanchitmalhotra126: Could you just provide the defaults instead of the empty object after the `??` and then not need to default to an empty string for each? Something like
```
const { modelName, trainingData, overview } = 
   modelDescriptions.find(model => model.id === selectedModelId) ?? 
   { modelName: '', trainingData: '', overview: ''}
```

sanchitmalhotra126 (approved):

PR #58073, opened 2024-04-17
Add DCDO flag for aichat chat_completion

fisher-alice:
This PR adds a DCDO flag for the `AichatController` `chat_completion` endpoint. I also added a corresponding unit test.

## Warning!!

The [AP CSP Create Performance Task](https://apcentral.collegeboard.org/courses/ap-computer-science-principles/exam) is in progress. The most critical dates are from April 3 - April 30, 2024. Please consider any risk introduced by this PR that could affect our students taking AP CSP. Code.org students taking AP CSP primarily use App Lab for their Create Task, however a small percent use Game Lab. Carefully consider whether your change has any risk of alterering, changing, or breaking anything in these two labs. Even small changes, such as a different button color, are considered significant during this time period. Reach out to the Student Learning team or Curriculum team for more details.

## Links
[jira](https://codedotorg.atlassian.net/browse/LABS-664)

## PR Checklist:

- [ ] Tests provide adequate coverage
- [ ] Privacy and Security impacts have been assessed
- [ ] Code is well-commented
- [ ] New features are translatable or updates will not break translations
- [ ] Relevant documentation has been added or updated
- [ ] User impact is well-understood and desirable
- [ ] Pull Request is labeled appropriately
- [ ] Follow-up work items (including potential tech debt) are tracked and linked

PR #58031, opened 2024-04-16
Restrict aichat controller to pilot

fisher-alice:
This PR restricts the access to `chat_completion` of `aichat_controller` to users in the `SingleUserExperiment` with pilot name`gen-ai-customizing-llm` and students of teachers in the pilot.

Note that I added `has_required_params` to `aichiat_controller` so that a `bad_request` response would be returned to the client instead of just erroring out. I also included dashboard unit tests.

## Warning!!

The [AP CSP Create Performance Task](https://apcentral.collegeboard.org/courses/ap-computer-science-principles/exam) is in progress. The most critical dates are from April 3 - April 30, 2024. Please consider any risk introduced by this PR that could affect our students taking AP CSP. Code.org students taking AP CSP primarily use App Lab for their Create Task, however a small percent use Game Lab. Carefully consider whether your change has any risk of alterering, changing, or breaking anything in these two labs. Even small changes, such as a different button color, are considered significant during this time period. Reach out to the Student Learning team or Curriculum team for more details.

## PR Checklist:

- [ ] Tests provide adequate coverage
- [ ] Privacy and Security impacts have been assessed
- [ ] Code is well-commented
- [ ] New features are translatable or updates will not break translations
- [ ] Relevant documentation has been added or updated
- [ ] User impact is well-understood and desirable
- [ ] Pull Request is labeled appropriately
- [ ] Follow-up work items (including potential tech debt) are tracked and linked

PR #58087, opened 2024-04-18
Aichat controller - make request to sagemaker endpoint and return assistant response

fisher-alice:
This PR follows up https://github.com/code-dot-org/code-dot-org/pull/57944 and makes an actual request to the Sagemaker endpoint within `AichatController` `chatCompletion`. Thanks to @sanchitmalhotra126 who created [this branch](https://github.com/code-dot-org/code-dot-org/compare/staging...sanchit/wip-aichat-sagemaker) which I used as a starting point which includes how to make the Sagemaker request and how to format the inputs.

This PR includes:
- additions to `Gemfile` and `Gemfile.lock`,
- removing the initial assistant chat message when a new chat sessions begins in `aichat`,
- refactoring the assignment of the `ChatCompletionMessage` `id`,
- Adding the `aichat_helper` file which defines helper functions used within `aichat_controller` including:
  -  `format_inputs_for_sagemaker_request` (formats frontend inputs)
  -  `request_sagemaker_chat_completion`
  - `get_sagemaker_assistant_response` (extracts latest assistant response from sagemaker respone)

 
I also added a test as a follow-up for adding the DCDO flag to controller https://github.com/code-dot-org/code-dot-org/pull/58073 and a test for when `storedMessages` param is not included in request.

Reminder: in order to have access to `AichatController` `chat_completion`, you must add the user to the pilot 'gen-ai-customizing-llms'. https://github.com/code-dot-org/code-dot-org/pull/58031

Screencast video with Sagemaker endpoint locally without then with retrieval contexts:

https://github.com/code-dot-org/code-dot-org/assets/107423305/6534234f-2553-4a3c-a90d-5e2d890e482f

## Warning!!

## Links
[jira - E2E retrieval](https://codedotorg.atlassian.net/browse/LABS-666)
[jira - E2E temperature tuning](https://codedotorg.atlassian.net/browse/LABS-665)

## PR Checklist:

- [ ] Code is well-commented


dashboard/app/helpers/aichat_helper.rb
```diff
@@ -0,0 +1,56 @@
...
+    SAGEMAKER_CLIENT.invoke_endpoint(
+      endpoint_name: SAGEMAKER_MODEL_ENDPOINT, # required
+      body: input_json.to_json.to_s, # required
+      # target_model: 'mistral-7b-inst.tar.gz',
+      content_type: "application/json"
+    )
+  end
+
+  def self.get_sagemaker_assistant_response(sagemaker_response)
+    parsed_response = JSON.parse(sagemaker_response.body.string)
```
fisher-alice, line 51: Not sure the best way to handle an error with these steps - `rescue`?

bencodeorg:
First off,  🎉 🎉 🎉 !!!

apps/src/aichat/redux/aichatRedux.ts
```diff
@@ -267,15 +259,14 @@ const aichatSlice = createSlice({
   initialState,
   reducers: {
     addChatMessage: (state, action: PayloadAction<ChatCompletionMessage>) => {
-      const newMessageId =
-        state.chatMessages[state.chatMessages.length - 1].id + 1;
+      const newMessageId = state.chatMessages.length + 1;
```
bencodeorg: Is this change needed for this PR? I think this might cause issues when you delete a single status message from the chat window?

dashboard/app/controllers/aichat_controller.rb
```diff
@@ -1,13 +1,15 @@
...
+
   # params are
   # newMessage: string
   # storedMessages: Array of {role: <'user', 'system', or 'assistant'>; content: string} - does not include user's new message
   # aichatParameters: {temperature: number; retrievalContexts: string[]; systemPrompt: string;}
   # chatContext: {userId: number; currentLevelId: string; scriptId: number; channelId: string;}
   # POST /aichat/chat_completion
   def chat_completion
-    return render status: :forbidden, json: {} unless can_request_aichat_chat_completion?
+    return render status: :forbidden, json: {} unless AichatHelper.can_request_aichat_chat_completion?
```
bencodeorg, line 12: Nit: if you include, I don't think you need to reference the module name?

dashboard/app/controllers/aichat_controller.rb
```diff
@@ -20,27 +22,20 @@ def chat_completion
...
-
-  private def can_request_aichat_chat_completion?
-    DCDO.get('aichat_chat_completion', true)
+    return render(status: :ok, json: payload.to_json)
   end
 
   private def has_required_params?
     begin
-      params.require([:newMessage, :storedMessages, :aichatParameters, :chatContext])
+      params.require([:newMessage, :aichatParameters, :chatContext])
```
bencodeorg: storedMessages not required any more? It looks like in `format_inputs_for_sagemaker_request` we assume that it is an array (ie, use `.push` on it)?

dashboard/app/helpers/aichat_helper.rb
```diff
@@ -0,0 +1,56 @@
...
+  SAGEMAKER_CLIENT = Aws::SageMakerRuntime::Client.new
+  SAGEMAKER_MODEL_ENDPOINT = "mistral-7b-inst-v01" # "BioMistral-7B"
+  BASE_PROMPT = "You are a helpful chatbot for children. Limit your responses to a small paragraph."
+  ASSISTANT = 'assistant'
+  USER = 'user'
+
+  # The instruction-tuned version of Mistral accepts formatted instructions where conversation roles
+  # must start with a user prompt and alternate between user and assistant.
+  def self.format_inputs_for_sagemaker_request(aichat_params, stored_messages, new_message)
+    stored_messages.push({role: USER, content: new_message})
```
bencodeorg: Feels a little weird to me to be directly modifying what we got from params...could we create a new array here?

dashboard/app/helpers/aichat_helper.rb
```diff
@@ -0,0 +1,56 @@
...
+        inputs += wrap_as_instructions(msg[:content])
+      elsif msg[:role] == ASSISTANT
+        inputs += msg[:content] + "</s>"
+      end
+    end
+    {
+      inputs: inputs,
+      parameters: {
+        temperature: aichat_params[:temperature].to_f,
+        max_new_tokens: 300,
```
bencodeorg: Pull these up into your list of constants?

dashboard/test/controllers/aichat_controller_test.rb
```diff
@@ -72,6 +74,13 @@ class AichatControllerTest < ActionController::TestCase
 
   test 'can_request_aichat_chat_completion returns false when DCDO flag is set to `false`' do
     DCDO.stubs(:get).with('aichat_chat_completion', true).returns(false)
-    assert_equal false, @controller.send(:can_request_aichat_chat_completion?)
+    assert_equal false, AichatHelper.can_request_aichat_chat_completion?
+  end
+
+  test 'returns forbidden when DCDO flag is set to `false`' do
```
bencodeorg, line 90: Very nice!

dashboard/app/helpers/aichat_helper.rb
```diff
@@ -0,0 +1,56 @@
...
+  ASSISTANT = 'assistant'
+  USER = 'user'
+
+  # The instruction-tuned version of Mistral accepts formatted instructions where conversation roles
+  # must start with a user prompt and alternate between user and assistant.
+  def self.format_inputs_for_sagemaker_request(aichat_params, stored_messages, new_message)
+    stored_messages.push({role: USER, content: new_message})
+    inputs = BASE_PROMPT + aichat_params[:systemPrompt]
+    inputs += aichat_params[:retrievalContexts].join(" ") if aichat_params[:retrievalContexts]
+    inputs = "<s> " + wrap_as_instructions(inputs)
```
bencodeorg: For my understanding, what is the `<s>` tag? And we'd end up with just one of these, but potentially many `</s>` from the loop below?

dashboard/app/helpers/aichat_helper.rb
```diff
@@ -0,0 +1,56 @@
...
+        max_new_tokens: 300,
+        top_p: 0.9
+      }
+    }
+  end
+
+  def self.request_sagemaker_chat_completion(input_json)
+    SAGEMAKER_CLIENT.invoke_endpoint(
+      endpoint_name: SAGEMAKER_MODEL_ENDPOINT, # required
+      body: input_json.to_json.to_s, # required
```
bencodeorg: Doesn't `to_json` already return a string?

dashboard/app/helpers/aichat_helper.rb
bencodeorg: hmm not sure...I think it's fair to say we could do error handling as a follow-up, not sure how specific we want to be about handling errors (ie, if bad json received, or if Sagemaker is down -> some generic failure message)?

dashboard/app/controllers/aichat_controller.rb
```diff
@@ -20,27 +22,20 @@ def chat_completion
     filter_result = ShareFiltering.find_failure(new_message_text, locale) if new_message_text
     # If the content is inappropriate, we skip sending to endpoint and instead hardcode a warning response on the front-end.
     return render(status: :ok, json: {status: filter_result.type, flagged_content: filter_result.content}) if filter_result
-    # TODO: Format input to send to Sagemaker.
+
+    input_json = AichatHelper.format_inputs_for_sagemaker_request(params[:aichatParameters], params[:storedMessages], params[:newMessage])
```
bencodeorg: Maybe something we can fix later, but `aichatParameters` is a bit confusing to me (aren't they all aichat params??)...`modelCustomizations` or something.

dashboard/app/helpers/aichat_helper.rb
```diff
@@ -0,0 +1,56 @@
...
+      # target_model: 'mistral-7b-inst.tar.gz',
```
bencodeorg: Intentionally commented out?

bencodeorg (commented):
A bunch of small comments/questions but this is exciting!!

dashboard/app/helpers/aichat_helper.rb
```diff
@@ -0,0 +1,56 @@
...
+  SAGEMAKER_MODEL_ENDPOINT = "mistral-7b-inst-v01" # "BioMistral-7B"
+  BASE_PROMPT = "You are a helpful chatbot for children. Limit your responses to a small paragraph."
+  ASSISTANT = 'assistant'
+  USER = 'user'
+
+  # The instruction-tuned version of Mistral accepts formatted instructions where conversation roles
+  # must start with a user prompt and alternate between user and assistant.
+  def self.format_inputs_for_sagemaker_request(aichat_params, stored_messages, new_message)
+    stored_messages.push({role: USER, content: new_message})
+    inputs = BASE_PROMPT + aichat_params[:systemPrompt]
```
cnbrenci, line 19: Does this need whitespace between?

dashboard/app/helpers/aichat_helper.rb
```diff
@@ -0,0 +1,56 @@
...
+
+  def self.get_sagemaker_assistant_response(sagemaker_response)
+    parsed_response = JSON.parse(sagemaker_response.body.string)
+    generated_text = parsed_response[0]["generated_text"]
+    parts = generated_text.split("[/INST]")
+    parts.last
+  end
+
+  def self.wrap_as_instructions(message)
+    "[INST] #{message} [/INST]"
```
cnbrenci: Might be good to put these tags into consts at the top `INSTRUCTIONS_START_TAG="[INST]"` `INSTRUCTIONS_END_TAG="[/INST]"` and same with the `<s>` tag which would also give an opportunity to give it a descriptive name.

dashboard/app/helpers/aichat_helper.rb
```diff
@@ -0,0 +1,56 @@
+module AichatHelper
+  SAGEMAKER_CLIENT = Aws::SageMakerRuntime::Client.new
+  SAGEMAKER_MODEL_ENDPOINT = "mistral-7b-inst-v01" # "BioMistral-7B"
+  BASE_PROMPT = "You are a helpful chatbot for children. Limit your responses to a small paragraph."
+  ASSISTANT = 'assistant'
```
cnbrenci: nit: inconsistent `"` and `'` throughout this file

dashboard/app/helpers/aichat_helper.rb
fisher-alice: Thanks! Logged as a follow-up https://codedotorg.atlassian.net/browse/LABS-720

dashboard/app/controllers/aichat_controller.rb
fisher-alice: Hmmm... I tried removing and got an `undefined method` error so will keep for now.

apps/src/aichat/redux/aichatRedux.ts
fisher-alice: >Is this change needed for this PR?

Since `initialChatMessage` is now empty, I had to update this line so that we didn't get an error the first time this is assigned.
I also realized that `id` was assigned to a user message in both `submtChatContents` and then reassigned here so made `id` in `ChatCompletionMessage` optional so can be assigned only once.

>I think this might cause issues when you delete a single status message from the chat window?

~~This doesn't seem to be causing an issue when I delete update status messages.~~

fisher-alice: Actually, I see it can. Thanks! will rethink.

dashboard/app/controllers/aichat_controller.rb
bencodeorg: Ah I think b/c they're class methods you prob need `self.`...doesn't matter though, this is fine

apps/src/aichat/redux/aichatRedux.ts
fisher-alice: Created a helper function to get `lastMessageId` and assign `id` to `ChatCompletionMessage` before dispatching `addChatMessage`. This may end up being moot if we assign `id` after getting response from backend but works for now. Thanks!

dashboard/app/controllers/aichat_controller.rb
fisher-alice: Yeah, the `require` interpreted an empty array as not meeting param requirement so removed. But now added the check to `storedMessages` array separately.

dashboard/app/helpers/aichat_helper.rb
fisher-alice: Here's a [ref](https://huggingface.co/mistralai/Mistral-7B-Instruct-v0.1#instruction-format)
`<s>` is a begin of sentence token and from reference:
"In order to leverage instruction fine-tuning, your prompt should be surrounded by [INST] and [/INST] tokens. The very first instruction should begin with a begin of sentence id. The next instructions should not. The assistant generation will be ended by the end-of-sentence token id."
And yes, it seems like there are multiple sentences, each sentence including an instruction surrounded by the instructions tokens and then the assistant response and the end of sentence token. Interestingly, a begin of sentence token is not required after an assistant response.

dashboard/app/controllers/aichat_controller.rb
fisher-alice: Updated and also updated `chatContext`. Lmk what you think.

dashboard/app/helpers/aichat_helper.rb
fisher-alice: I'll remove for now. I'll ask Sanchit about this when he's back in office.

dashboard/app/helpers/aichat_helper.rb
fisher-alice: I don't believe so.

dashboard/app/helpers/aichat_helper.rb
fisher-alice: Nice idea - updated!

dashboard/app/helpers/aichat_helper.rb
fisher-alice: Actually, I'll add since I see retrieval contexts are separated by whitespace as well. Thanks!

dashboard/app/helpers/aichat_helper.rb
fisher-alice: @sanchitmalhotra126 - do you mind sharing why `target_model` is commented out? Thanks!

dashboard/app/helpers/aichat_helper.rb
fisher-alice: Added a comment and resource link as a ref.

dashboard/app/helpers/aichat_helper.rb
```diff
@@ -0,0 +1,64 @@
...
+  SENTENCE_BEGIN_TOKEN = "<s>"
+  SENTENCE_END_TOKEN = "</s>"
+  MAX_NEW_TOKENS = 300
+  SAGEMAKER_CLIENT = Aws::SageMakerRuntime::Client.new
+  SAGEMAKER_MODEL_ENDPOINT = "mistral-7b-inst-v01" # "BioMistral-7B"
+  TOP_P = 0.9
+
+  # The instruction-tuned version of Mistral accepts formatted instructions where conversation roles
+  # must start with a user prompt and alternate between user and assistant.
+  # Mistral-7B-Instruction LLM instruction format doc at https://huggingface.co/mistralai/Mistral-7B-Instruct-v0.1.
```
fisher-alice, line 16: More on mistral tokenization at https://docs.mistral.ai/guides/tokenization/

apps/src/aichat/redux/aichatRedux.ts
```diff
@@ -191,17 +203,19 @@ const saveAiCustomization = async (
...
-        id: 0,
+        id: newMessageId,
         role: Role.MODEL_UPDATE,
         chatMessageText:
           AI_CUSTOMIZATIONS_LABELS[property as keyof AiCustomizations],
         status: Status.OK,
         timestamp: getCurrentTime(),
       })
     );
+    newMessageId++;
```
bencodeorg: This I think (?) this will work, but I think might be a little easier to manage if the helper was `getMessageId()` (rather than `getLastMessageId`), which looked at some const (`currentMessageId` ?), returned that value, and incremented it so the next call gets an unused ID. And would make the helper simpler than inspecting the existing messages / assuming the last message will have the highest ID.

I do think we'll continue to need message IDs specifically for front end use, since (unlike AI Tutor) we aren't logging at the individual message level, so we'll never get message IDs from the server. [A little more detail in this PR](https://github.com/code-dot-org/code-dot-org/pull/58103), curious for your thoughts!

apps/src/aichat/redux/aichatRedux.ts
```diff
@@ -267,15 +276,9 @@ const aichatSlice = createSlice({
...
-        state.chatMessages[state.chatMessages.length - 1].id + 1;
-      const newMessage = {
-        ...action.payload,
-        id: newMessageId,
-      };
-      state.chatMessages.push(newMessage);
+      state.chatMessages.push(action.payload);
     },
-    removeChatMessage: (state, action: PayloadAction<number>) => {
+    removeChatMessage: (state, action: PayloadAction<number | undefined>) => {
```
bencodeorg: Do we need this? Looks like we still require messages to have an ID?

bencodeorg (approved):
🎉

apps/src/aichat/redux/aichatRedux.ts
fisher-alice: Thanks - just updated as you suggested.

cnbrenci (approved):
🎉 🚀

Gemfile
```diff
@@ -240,6 +240,8 @@ gem 'aws-sdk-glue'
 gem 'aws-sdk-rds'
 gem 'aws-sdk-route53'
 gem 'aws-sdk-s3'
+gem 'aws-sdk-sagemaker'
```
sanchitmalhotra126, line 243: My bad - I had this in my branch, but don't know if we actually need the `aws-sagemaker` gem? I think we may only need `sagemakerruntime`

apps/src/aichat/types.ts
```diff
@@ -65,7 +65,7 @@ export interface AiCustomizations {
   modelCardInfo: ModelCardInfo;
 }
 
-export type AichatParameters = Omit<AiCustomizations, 'modelCardInfo'>;
+export type AichatModelCustomizations = Omit<AiCustomizations, 'modelCardInfo'>;
```
sanchitmalhotra126, line 68: Could we add some comments here to differentiate this from `AiCustomizations`?

dashboard/app/helpers/aichat_helper.rb
```diff
@@ -0,0 +1,64 @@
...
+  USER = "user"
+  BASE_PROMPT = "You are a helpful chatbot for children. Limit your responses to a small paragraph. "
+  INSTRUCTIONS_BEGIN_TOKEN = "[INST]"
+  INSTRUCTIONS_END_TOKEN = "[/INST]"
+  SENTENCE_BEGIN_TOKEN = "<s>"
+  SENTENCE_END_TOKEN = "</s>"
+  MAX_NEW_TOKENS = 300
+  SAGEMAKER_CLIENT = Aws::SageMakerRuntime::Client.new
+  SAGEMAKER_MODEL_ENDPOINT = "mistral-7b-inst-v01" # "BioMistral-7B"
+  TOP_P = 0.9
```
sanchitmalhotra126, line 12: @tshaffercodeorg just confirming that this value is okay?

dashboard/app/helpers/aichat_helper.rb
```diff
@@ -0,0 +1,64 @@
...
+
+  # The instruction-tuned version of Mistral accepts formatted instructions where conversation roles
+  # must start with a user prompt and alternate between user and assistant.
+  # Mistral-7B-Instruction LLM instruction format doc at https://huggingface.co/mistralai/Mistral-7B-Instruct-v0.1.
+  def self.format_inputs_for_sagemaker_request(aichat_params, stored_messages, new_message)
+    all_messages = [*stored_messages, {role: USER, content: new_message}]
+    inputs = BASE_PROMPT + aichat_params[:systemPrompt]
+    inputs += aichat_params[:retrievalContexts].join(" ") if aichat_params[:retrievalContexts]
+    inputs = SENTENCE_BEGIN_TOKEN + wrap_as_instructions(inputs)
+    # Filter only messages from user or assistant - ignore model update messages.
```
sanchitmalhotra126, line 22: We should ideally filter out model update messages on the front end before sending

sanchitmalhotra126 (commented):
I know this is already merged but great work! Just leaving a few comments that I noticed for potential follow-ups

dashboard/app/helpers/aichat_helper.rb
bencodeorg: This change should result in model update messages being filtered before sending:

https://github.com/code-dot-org/code-dot-org/pull/58103/files

fisher-alice:
> I know this is already merged but great work! Just leaving a few comments that I noticed for potential follow-ups

Thanks! Here's a follow-up PR to address your comments. https://github.com/code-dot-org/code-dot-org/pull/58157

dashboard/app/helpers/aichat_helper.rb
tshaffercodeorg: Regarding SAGEMAKER_MODEL_ENDPOINT? Our two endpoints currently are `mistral-7b-inst-base` (v0.1) and `mistral-7b-bio` (BioMistral) respectively. The other parameters seem fine to me although I'm unsure what Aws::SageMakerRuntime::Client.new does (probably something with the ruby script)

tshaffercodeorg: Correction, the above information is for our dev AWS account, not the prod one. Since this script is targeting our prod account, the above parameter value is correct.

fisher-alice: Thanks! Confirmed that this is correct with @tshaffercodeorg offline as well.

sanchitmalhotra126: Oh yep, I was specifically talking about the `TOP_P` value - I just set that to some default while testing, but is there another value we should use?

dashboard/app/helpers/aichat_helper.rb
sanchitmalhotra126: Yep you can remove this! This was being used for some testing with [multi-model endpoints](https://docs.aws.amazon.com/sagemaker/latest/dg/multi-model-endpoints.html), where you could specify a single endpoint, and a specific model within that endpoint. For now, we're just using single model endpoints, so only `endpoint_name` is needed

fisher-alice: Thanks for the explanation!

PR #58092, opened 2024-04-18
Music: pack dialog buttons

breville:
This adds a new optional interaction model to the pack dialog, so that we can see how it compares to the existing model.

### New model

In the new interaction model, exposed using the `show-pack-dialog-buttons=true` URL parameter, the user selects a restricted pack, and can then press `Continue` to use it.  The `skip` button will not use a restricted pack, leaving only the non-restricted sounds.  A selected pack can be deselected as well.

https://github.com/code-dot-org/code-dot-org/assets/2205926/3a9a8403-6abb-46ef-b294-acbb61555c02

### Existing model

The existing model continues to work, and features the default pack inline, as added in https://github.com/code-dot-org/code-dot-org/pull/58062.  There are no additional buttons in the dialog, and clicking a pack chooses it immediately.

https://github.com/code-dot-org/code-dot-org/assets/2205926/0e8db48c-783f-452d-ae27-616b1b0c2025


thomasoniii (approved):
lgtm 🚀

markabarrett:
New selection confirmation feels good. Non-blocking polish items-

- The label for the refresh button is 1px off vertical center (needs a nudge down). 
- The start over modal is light mode, other modals are dark mode. 
- Continue and Skip button could be replaced with DSCO component so they have click states.

PR #58093, opened 2024-04-18
P20-299: Delay "Thank You" email to be sent 24 hrs after approval

artem-vavilov:
## Summary
The delay is intended to provide notice to a parent when a student may no longer be monitoring the "parent's email."   

## Links
- JIRA ticket: [P20-299](https://codedotorg.atlassian.net/browse/P20-299)


dashboard/lib/services/child_account.rb
```diff
@@ -32,7 +32,9 @@ def self.grant_permission_request!(permission_request)
       update_compliance(user, Policies::ChildAccount::ComplianceState::PERMISSION_GRANTED)
       user.save!
       parent_email = permission_request.parent_email
-      ParentMailer.parent_permission_confirmation(parent_email).deliver_now
+      ParentMailer.
+        parent_permission_confirmation(parent_email).
+        deliver_later(wait: Policies::ChildAccount::PERMISSION_GRANTED_MAIL_DELAY)
```
artem-vavilov, line 37:
```ruby
#<MailDeliveryJob:0x00007fadf3b4ed28
 @arguments=["ParentMailer", "parent_permission_confirmation", "deliver_now", {:args=>["test@example.com"]}],
 @exception_executions={},
 @executions=0,
 @job_id="5e3d5045-4c90-44d5-8886-e65f4fa77352",
 @priority=nil,
 @provider_job_id="3a33e438-1e83-48e6-b015-3e686ce60c72",
 @queue_name="mailers",
 @scheduled_at=1713533474.4725819,
 @timezone="UTC">
```

dashboard/app/jobs/concerns/active_job_metrics.rb
```diff
@@ -0,0 +1,104 @@
...
+          timestamp: Time.now,
+          dimensions: [
+            {name: 'Environment', value: CDO.rack_env},
+          ],
+        },
+      ]
+    )
+  rescue => exception
+    Honeybadger.notify(exception, error_message: 'Error reporting ActiveJob metrics')
+  end
```
cat5inthecradle, line 50: Today, with EvaluateRubricJob as the only job hitting the queue, we can assume that a non-zero `JobCount` is a possible indicator of a stuck queue. Now that we're adding a job that will sit in the queue for 24 hours, we should revisit what metrics are actually useful.

This metric is currently only reporting the environment metric - should we update it to use `common_dimensions` like the others?

artem-vavilov: Depends on what you need this metric for. If you want to identify long-running jobs, we could count jobs using `Delayed::Job.where("locked_at IS NOT NULL AND locked_at < ?", 5.minutes.ago)` instead.

artem-vavilov: Depends on what you need this metric for. If you want to identify long-running jobs, we could count them like:
```ruby
Delayed::Job.where("locked_at IS NOT NULL AND locked_at < ?", 5.minutes.ago)
```

dashboard/app/jobs/concerns/active_job_metrics.rb
```diff
@@ -0,0 +1,104 @@
...
+
+  protected def common_dimensions
+    [
+      {name: 'Environment', value: CDO.rack_env},
+      {name: 'JobName', value: self.class.name},
+      {name: 'JobId', value: job_id},
+      {name: 'QueueName', value: queue_name},
+    ]
+  end
```
cat5inthecradle: I'm not sure if `job_id` is a good idea here. This is more expensive and it makes looking at metrics in cloudwatch a bit more complicated because you have to create a custom Multi-Source query. You won't be able to discover problematic job_id's via metrics, you need to know the job ID first. Can we rely on Honeybadger for tracing failures instead of these metrics?

In terms of cost, we are billed per unique custom metric. We have 4 main metric types here: JobCount, WaitTime, ExecutionTime, TotalTime. This gets multiplied by each unique dimension - so 4 per JobName (2) per environment (4) per queue(2, but it's 1:1 with JobName, so counts as 1). That gets us up to 32. If we add a Job_id dimension, we increase to 32N, where N is the number of jobs we process per month, which is going to be a lot.

cat5inthecradle (changes requested):
The refactor into ActiveJobMetrics looks good, but I don't want to include `job_id` as a dimension for these metrics.

dashboard/app/jobs/concerns/active_job_metrics.rb
artem-vavilov: Should I also remove `QueueName`?

dashboard/app/jobs/concerns/active_job_metrics.rb
wilkie: 5 minutes seems like the right amount of time considering what we would expect the AI job timing to be.

Noticing that there are often jobs that take longer than, say, 20 minutes is valuable. Since that only will be recorded if that job completes, and we would want to know if there is a job that is stuck, we would need one of these actions (on enqueue, probably) to also post up info about the current runtime of the current running task (aka queue/wait time) along with the queue length. If both numbers go up, the queue isn't stable.

dashboard/app/jobs/concerns/active_job_metrics.rb
cat5inthecradle: I think we'd be fine with just JobName

wilkie (approved):
I like the refactoring and I like that we are pursuing using the worker queues for emails! This paves the way to turn some of our cronjob-activated mailers into jobs like this, too.

cat5inthecradle (approved):

dashboard/app/jobs/mail_delivery_job.rb
```diff
@@ -8,4 +10,16 @@
 # @see https://apidock.com/rails/v6.1.7.7/ActionMailer/MessageDelivery/deliver_later
 class MailDeliveryJob < ActionMailer::MailDeliveryJob
   include ActiveJobMetrics
+
+  rescue_from StandardError, with: :report_exception
```
artem-vavilov, line 14: NOTE: The mail exceptions will be reported without retrying the job

cat5inthecradle: If we have a problem and a bunch of email can't get sent, how should we go about re-queuing up those messages?

artem-vavilov: The same as you re-queuing up regular failed jobs

artem-vavilov: The same as we re-queuing up regular failed jobs:
```ruby
Delayed::Job.where(queue: MailDeliveryJob.default_queue_name).where.not(failed_at: nil).find_each(&:invoke_job)
```
or
```ruby
find_each(&:reschedule_at)
```

artem-vavilov: But I believe we currently delete failed jobs: https://github.com/collectiveidea/delayed_job/blob/b66bb64437a8606a414a390531ac73108911434e/README.md?plain=1#L444-L445

cat5inthecradle: Yes, I believe we currently delete failed jobs. We might want to re-think that.

artem-vavilov: We could also add Web UI for managing jobs: https://github.com/ejschmitt/delayed_job_web

artem-vavilov: https://github.com/code-dot-org/code-dot-org/pull/58093/commits/8e230818025c4c183ea5319e99556c42b5fde500

dashboard/app/jobs/concerns/active_job_metrics.rb
cat5inthecradle: If I want to get an early warning that our queue workers aren't keeping up (either too slow, or stopped), then we could filter this down to just available jobs:

```sql
WHERE run_at < now AND locked_at IS NULL
```

We want that to be zero usually. We could also track a `WaitTimeOfOldestWaitingJob` metric (needs a better name).

cat5inthecradle: Lets just add the JobName dimension to this metric for now. We could consider adding an "AvailableJobCount" metric later if we think it would be useful.

```suggestion
  protected def report_job_count
    Cdo::Metrics.push(
      METRICS_NAMESPACE, [
        {
          # Same metric as "bin/cron/report_activejob_metrics"
          metric_name: 'JobCount',
          value: Delayed::Job.count,
          unit: 'Count',
          timestamp: Time.now,
          dimensions: common_dimensions,
        },
      ]
    )
  rescue => exception
    Honeybadger.notify(exception, error_message: 'Error reporting ActiveJob metrics')
  end
```

dashboard/config/initializers/delayed_job_config.rb
```diff
@@ -0,0 +1,3 @@
+# Prevents auto-deletion of failed jobs.
+# The failed jobs will be marked with non-null `failed_at`.
+Delayed::Worker.destroy_failed_jobs = false
```
cat5inthecradle, line 3: @davidsbailey This will cause failed EvaluateRubricJobs to pile up in the database - which is handy for troubleshooting, or if something downstream breaks and we want to retry a batch.

But we might want to follow this up with some code to scan the jobs table and clean up any failed jobs that have since had their projects evaluated. I'll start a Slack discussion about how the job failure & retry logic should work here.

cat5inthecradle (approved):

wilkie (approved):
If it is no problem for Infra, I like it, too. Not deleting the failed jobs gives one more piece of evidence for potentially diagnosing EvaluateRubricJob failures, so that might be nice. Though, I agree that we might want to clean them up still... but that's our problem.

Nice work, Artem!

davidsbailey:
Nice work Artem, it's good to see more people using Active Job! Thanks Darin for keeping an eye out for EvaluateRubricJob monitoring.

PR #58095, opened 2024-04-18
Add deleting MailJet contacts to deleted_accounts_helper

bethanyaconnor:
The delete method for MailJet was written in a [previous PR](https://github.com/code-dot-org/code-dot-org/pull/58012) so this is just hooking it up! Before deleting a MailJet contact, we do check if one actually exists, so this shouldn't affect existing users.


TurnerRiley (approved):
LGTM!

carl-codeorg (approved):
:shipit:

PR #58096, opened 2024-04-18
P20-881: Unlock existing CPA students who have requested parental permission

artem-vavilov:
## Summary
Unlocks CPA students created before the policy went into effect who have requested parental permission but have not yet received approval

## Links
- JIRA ticket: [P20-823](https://codedotorg.atlassian.net/browse/P20-823)

### DIFF
--- a/dashboard/lib/policies/child_account.rb
+++ b/dashboard/lib/policies/child_account.rb
@@ -13,6 +13,15 @@ class Policies::ChildAccount
 
     # The student's account has been approved by their parent.
     PERMISSION_GRANTED = SharedConstants::CHILD_ACCOUNT_COMPLIANCE_STATES.PERMISSION_GRANTED
+
+    # def self.locked_out?(student)
+    # def self.request_sent?(student)
+    # def self.permission_granted?(student)
+    SharedConstants::CHILD_ACCOUNT_COMPLIANCE_STATES.to_h.each do |key, value|
+      define_singleton_method("#{key.downcase}?") do |student|
+        student.child_account_compliance_state == value
+      end
+    end
   end
 
   # The individual US State child account policy configuration
@@ -30,7 +39,11 @@ class Policies::ChildAccount
   # parent permission before the student can start using their account.
   def self.compliant?(user)
     return true unless parent_permission_required?(user)
-    user.child_account_compliance_state == ComplianceState::PERMISSION_GRANTED
+    # CPA Part 2: unlock students created before the policy went into effect
+    # who have requested parental permission but have not yet received approval.
+    return true if ComplianceState.request_sent?(user) && user_predates_policy?(user)
+
+    ComplianceState.permission_granted?(user)
   end
 
   # Checks if a user is affected by a state policy but was created prior to the
--- a/dashboard/test/lib/policies/child_account_test.rb
+++ b/dashboard/test/lib/policies/child_account_test.rb
@@ -1,6 +1,30 @@
 require 'test_helper'
 
 class Policies::ChildAccountTest < ActiveSupport::TestCase
+  class ComplianceStateTest < ActiveSupport::TestCase
+    setup do
+      @student = build(:student)
+    end
+
+    test 'locked_out?' do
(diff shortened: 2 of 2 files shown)


dashboard/test/lib/policies/child_account_test.rb
```diff
@@ -12,6 +36,8 @@ class Policies::ChildAccountTest < ActiveSupport::TestCase
       [[:non_compliant_child, :not_U13], true],
       [[:non_compliant_child, :migrated_imported_from_clever], true],
       [[:non_compliant_child, :with_lti_auth], true],
+      [[:non_compliant_child, :with_pending_parent_permission, {created_at: '2023-06-30T23:59:59Z'}], true],
+      [[:non_compliant_child, :with_pending_parent_permission, {created_at: '2023-07-01T00:00:00Z'}], false],
```
mgc1194, line 40: Wow, you went for seconds of precision here! If these tests pass, the code must be very reliable.

dashboard/lib/policies/child_account.rb
```diff
@@ -13,6 +13,15 @@ module ComplianceState
 
     # The student's account has been approved by their parent.
     PERMISSION_GRANTED = SharedConstants::CHILD_ACCOUNT_COMPLIANCE_STATES.PERMISSION_GRANTED
+
+    # def self.locked_out?(student)
+    # def self.request_sent?(student)
+    # def self.permission_granted?(student)
+    SharedConstants::CHILD_ACCOUNT_COMPLIANCE_STATES.to_h.each do |key, value|
```
mgc1194, line 20: I'm no sure if you wanted these comments to be here.

artem-vavilov: Just wanted to provide more clarity in case someone searches these methods

dashboard/test/lib/policies/child_account_test.rb
artem-vavilov: It's a pretty straightforward comparison that isn't dependent on the current time, so there shouldn't be any issue

dashboard/lib/policies/child_account.rb
```diff
@@ -30,7 +39,11 @@ module ComplianceState
   # parent permission before the student can start using their account.
   def self.compliant?(user)
     return true unless parent_permission_required?(user)
-    user.child_account_compliance_state == ComplianceState::PERMISSION_GRANTED
+    # CPA Part 2: unlock students created before the policy went into effect
+    # who have requested parental permission but have not yet received approval.
+    return true if ComplianceState.request_sent?(user) && user_predates_policy?(user)
```
artem-vavilov, line 44: https://github.com/code-dot-org/code-dot-org/blob/e22716b4eb86d375914e3747618da89e358b358d/dashboard/lib/policies/child_account.rb#L49-L53

carl-codeorg: In the [application controller](https://github.com/code-dot-org/code-dot-org/blob/staging/dashboard/app/controllers/application_controller.rb#L354), we already don't redirect to the lockout page if `user_predates_policy?` is true. Is there a phase of CPA where we will remove that exemption from the redirect in the application_controller, but still want to allow pre-existing users to avoid the lockout if their request is pending?

artem-vavilov: cc @daynew

daynew: The next milestone will deal with updating the routing logic: https://codedotorg.atlassian.net/browse/P20-138

artem-vavilov: @daynew, do you know why we use the `UTC` time zone instead of `MST`, which is Colorado's standard time?
https://github.com/code-dot-org/code-dot-org/blob/e22716b4eb86d375914e3747618da89e358b358d/dashboard/lib/policies/child_account.rb#L33

daynew: Not for good reason. If you want to us MST instead, feel free.

daynew (approved):

carl-codeorg (approved):

PR #58099, opened 2024-04-18
Add sending welcome email through mailjet behind a DCDO flag

bethanyaconnor:
Hooks up using MailJet for our welcome email behind a DCDO flag! I put this behind a DCDO flag so that we can:
- test on an adhoc
- turn on the feature at a time I can easily monitor it
- turn off the feature if something breaks. This is a critical path and, while I believe I've done my best to not disrupt sign ups, it is always possible

## Warning!!

## PR Checklist:

- [ ] Code is well-commented


TurnerRiley (approved):
LGTM!

dashboard/test/controllers/registrations_controller_test.rb
```diff
@@ -370,6 +372,18 @@ class RegistrationsControllerTest < ActionController::TestCase
...
+    MailJet.stubs(:enabled?).returns(true)
+    MailJet.expects(:create_contact_and_send_welcome_email).once
+    assert_creates(User) do
+      post :create, params: {user: teacher_params}
+    end
+
+    assert_empty ActionMailer::Base.deliveries
+  end
+
   test 'create new teacher with es-MX locale sends localized welcome email' do
```
mgc1194, line 387: You may need to update the i18n tests for localized emails.
As fas as I know, Dani was experimenting with Mailjet-Crowdin integration to provide localized emails, without exposing a security risk on our end (a.k.a. processing markdown as HTML).

dmcavoy: We have the mailjet to crowdin sync enabled but I haven't requested translation of the email yet because we don't yet have a way to identify what language the email should be sent in. @mgc1194 do you know of a way we could get that information right now? How accurate is it?

mgc1194: Would the language the user has during the signup be the one we want to send the email to?
Once the user has selected a language other than English, it gets stored in the language cookie.

dmcavoy: We want to add a preferred language field to the sign up process in the future but it did not make the cut for this round. Could be something we consider for early next year

bethanyaconnor: Thanks Mario -- good catch!

It sounds like, for now, this is worth restricting to English users and Dani and I can investigate the non-EN behavior we want. @mgc1194 is it correct that we only have this email translated into `es-MX` right now?

bethanyaconnor: The sign up language seems like it would be a good proxy, though, at least for the welcome email. I can add that as a user property on the MailJet side if we want to use it

mgc1194: @bethanyaconnor you are right, it looks like the welcome email is only supported in English or Spanish-Mexico
https://github.com/code-dot-org/code-dot-org/blob/cf842e965eead6830efcff728848132024729377/dashboard/app/mailers/teacher_mailer.rb#L9-L12

bethanyaconnor: Does the update look good? (Assuming the strange Drone errors go away)

dmcavoy: How do we know if a teacher is Spanish-Mexico? What is the logic it is using right now?

dmcavoy: It looks like the email might have gotten translated: https://crowdin.com/editor/codeorg-markdown/182198/en-esmx?view=comfortable&filter=basic&value=0. I can try syncing it back to MailJet. If we do that then we would need to send over language to send in to MailJet

dmcavoy: @castro-jorge FYI on this thread

bethanyaconnor: > How do we know if a teacher is Spanish-Mexico? What is the logic it is using right now?

The language of the request. I would assume that's the language the site was in when the user signed up.

PR #58100, opened 2024-04-18
Unit tests for ProgressFeedbackBanner

kobryan0619:
## Warning!!

<hr>

Added tests for the `ProgressFeedbackBannerTest` component

## Links

[Ticket](https://codedotorg.atlassian.net/jira/software/projects/TEACH/boards/65?selectedIssue=TEACH-1028)

## Testing story

It is all tests!

### DIFF
--- a/apps/test/unit/templates/sectionProgressV2/ProgressFeedbackBannerTest.jsx
+++ b/apps/test/unit/templates/sectionProgressV2/ProgressFeedbackBannerTest.jsx
@@ -1,4 +1,4 @@
-import {fireEvent, render, screen} from '@testing-library/react';
+import {fireEvent, render, screen, waitFor} from '@testing-library/react';
 import React from 'react';
 import sinon from 'sinon';
 
@@ -61,25 +61,25 @@ describe('UnconnectedProgressFeedbackBanner', () => {
     expect(shareMoreText).to.not.exist;
   });
 
-  it('clicking thumbs up attempts to send feedback and asks for more detailed feedback', () => {
-    render(
-      <UnconnectedProgressFeedbackBanner {...defaultProps} canShow={true} />
-    );
-    const thumbsUpButton = screen.getByTitle(
-      i18n.progressV2_feedback_thumbsUp()
-    );
-    expect(thumbsUpButton).to.be.visible;
-    fireEvent.click(thumbsUpButton);
-    expect(fakeCreate).to.have.been.calledOnce;
-    const shareMoreText = screen.getByText(
-      i18n.progressV2_feedback_shareMore()
-    );
-    const shareMoreLink = screen.getByRole('link', {
-      name: i18n.progressV2_feedback_shareMoreLinkText(),
-    });
-    expect(shareMoreLink).to.be.visible;
-    expect(shareMoreText).to.be.visible;
-  });
+  //   it('clicking thumbs up attempts to send feedback and asks for more detailed feedback', () => {
+  //     render(
+  //       <UnconnectedProgressFeedbackBanner {...defaultProps} canShow={true} />
+  //     );
+  //     const thumbsUpButton = screen.getByTitle(
+  //       i18n.progressV2_feedback_thumbsUp()
+  //     );
+  //     expect(thumbsUpButton).to.be.visible;
+  //     fireEvent.click(thumbsUpButton);
+  //     expect(fakeCreate).to.have.been.calledOnce;
+  //     const shareMoreText = screen.getByText(
+  //       i18n.progressV2_feedback_shareMore()
+  //     );
+  //     const shareMoreLink = screen.getByRole('link', {
+  //       name: i18n.progressV2_feedback_shareMoreLinkText(),
+  //     });
+  //     expect(shareMoreLink).to.be.visible;
+  //     expect(shareMoreText).to.be.visible;
+  //   });
 
   //   it('clicking thumbs down attempts to send feedback and asks for more detailed feedback', () => {
   //     render(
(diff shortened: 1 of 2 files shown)


apps/test/unit/templates/sectionProgressV2/ProgressFeedbackBannerTest.jsx
```diff
@@ -0,0 +1,144 @@
...
+    fetchProgressV2Feedback: fakeFetch,
+    createProgressV2Feedback: fakeCreate,
+    errorWhenCreatingOrLoading: null,
+  };
+
+  afterEach(() => {
+    fakeCreate.resetHistory();
+  });
+
+  it('renders correctly with initial state', () => {
```
lfryemason: Could we also assert that it will call `fetchProgressV2Feedback`

apps/test/unit/templates/sectionProgressV2/ProgressFeedbackBannerTest.jsx
```diff
@@ -0,0 +1,144 @@
...
+import sinon from 'sinon';
+
+import {UnconnectedProgressFeedbackBanner} from '@cdo/apps/templates/sectionProgressV2/ProgressFeedbackBanner';
+import i18n from '@cdo/locale';
+
+import {expect} from '../../../util/reconfiguredChai';
+
+describe('UnconnectedProgressFeedbackBanner', () => {
+  const fakeFetch = sinon.spy();
+  const fakeCreate = sinon.spy();
```
lfryemason, line 12: optional nit: I prefer to have an empty function (`() => {}`) as the default and override the default whenever we need to add the spy.

apps/test/unit/templates/sectionProgressV2/ProgressFeedbackBannerTest.jsx
lfryemason:
```suggestion
  it('renders empty if can't show', () => {
```

apps/test/unit/templates/sectionProgressV2/ProgressFeedbackBannerTest.jsx
```diff
@@ -0,0 +1,144 @@
...
+      i18n.progressV2_feedback_question()
+    );
+    const shareMoreText = screen.queryByText(
+      i18n.progressV2_feedback_shareMore()
+    );
+    expect(questionText).to.not.exist;
+    expect(shareMoreText).to.not.exist;
+  });
+
+  it('renders correctly when user has not answered survey question', () => {
```
lfryemason: Can we also have `does not show if user has answered survey question already`

apps/test/unit/templates/sectionProgressV2/ProgressFeedbackBannerTest.jsx
```diff
@@ -0,0 +1,144 @@
...
+    const shareMoreText = screen.queryByText(
+      i18n.progressV2_feedback_shareMore()
+    );
+    expect(questionText).to.not.exist;
+    expect(shareMoreText).to.not.exist;
+  });
+
+  it('clicking thumbs up attempts to send feedback and asks for more detailed feedback', async () => {
+    render(
+      <UnconnectedProgressFeedbackBanner {...defaultProps} canShow={true} />
```
lfryemason: optional nit: seems like we are overriding canShow more often than not. should we make that the default and override false otherwise?

apps/test/unit/templates/sectionProgressV2/ProgressFeedbackBannerTest.jsx
```diff
@@ -0,0 +1,144 @@
...
+      i18n.progressV2_feedback_shareMore()
+    );
+    const shareMoreLink = screen.getByRole('link', {
+      name: i18n.progressV2_feedback_shareMoreLinkText(),
+    });
+    expect(shareMoreLink).to.be.visible;
+    expect(shareMoreText).to.be.visible;
+  });
+
+  it('user is able to close the banner', async () => {
```
lfryemason, line 141: Can we also have `attempts reset and reload if error from createFeedback`.

lfryemason (commented):
Thank you so much. These look great, just a couple other tests that I would want to add

apps/test/unit/templates/sectionProgressV2/ProgressFeedbackBannerTest.jsx
kobryan0619: I thought of that too... I also think of default as the "normal" status.  Most of the time this doesn't show, but in most of the tests we are testing what IS shown.  BUT, I get that we would have less repetition if we changed the default.   I'll make the change!

apps/test/unit/templates/sectionProgressV2/ProgressFeedbackBannerTest.jsx
kobryan0619: Got it... this makes sense - it helps it stay more self contained.  I was trying to avoid adding lines of code into individual tests by keeping this as the default.  That being said, I still declared the spies outside the individual test - I think this was the right move to be able to still use the `afterEach`.  Open to feedback on this choice though.

apps/test/unit/templates/sectionProgressV2/ProgressFeedbackBannerTest.jsx
```diff
@@ -0,0 +1,221 @@
...
+      );
+      const shareMoreText = screen.queryByText(
+        i18n.progressV2_feedback_shareMore()
+      );
+      expect(questionText).to.not.exist;
+      expect(shareMoreText).to.not.exist;
+    });
+  });
+
+  it('attempts reset and reload if error from createFeedback', async () => {
```
kobryan0619: @lfryemason I had a hard time with this test.  After some research, I am thinking that perhaps how this is implemented in the component might benefit from some refactoring.

lfryemason (approved):
Approving as long as we comment out (and slightly cleanup) the troublesome test.

PR #58103, opened 2024-04-18
Gen AI: chat history prefactor

bencodeorg:
## Warning!!

## Description

Preparing front end code to support logging chat history as well as showing chat history from two separate "sessions" in the UI. A "session" is identified by a:
- user
- level
- model customizations (temperature, etc)
- (new addition) existing chat history (ie, when a user hits the clear button, new session)

I iterated on this a bit, so I'm curious for other thoughts on where I landed. I consulted the approach Emily is taking for managing message IDs in AI Tutor [here](https://github.com/code-dot-org/code-dot-org/pull/57984) -- I'm taking a similar approach and using Rails IDs on the front end to identify a session.

The idea is to tag each message with the Rails ID for the `AichatSession` to which it belongs, and wipe the current session ID any time something happens where we want a new session (level change, clicking clear chat button, updating model customizaton). When we send a message that is not associated with a session (eg, first message a user sends on page load, or after changing a model customization), we go back after we get a response from the server and update the message with the provided session ID.

We keep track of the current session ID as well, such that we can only send messages to Rails from the current session.

A couple alternatives I considered:
- Using a boolean to keep track of messages in the current session. This also might work, but we'd have to go back and reset all old messages to be false (ie, not in current session) any time a customization changes, whereas this session ID approach should work without needing to do that.
- Doing a soft delete of messages when a user clears their chat history (see Slack discussion [here](https://codedotorg.slack.com/archives/C06FELVTV0Q/p1713394565462959?thread_ts=1713394111.370279&cid=C06FELVTV0Q) and [here](https://codedotorg.slack.com/archives/C05DMS18MEX/p1713456174984099?thread_ts=1713292032.551679&cid=C05DMS18MEX)). We thought this would be helpful, but the benefits here weren't clear to me upon further inspection. LMK if I'm missing something.

## Testing story

I hard coded a session ID to be returned from the Rails controller and tested various scenarios:
- clearing chat history or updating model customization set the session ID to undefined on the front end, resulting in a blank array of messages sent to Rails.
- any existing chat history with session IDs matching the current session ID were sent to the model
- a user message with no session ID was tagged appropriated with the current session ID returned from Rails

Also noted that things continue to work even without the hard coding of a session ID in the response for now (I haven't included a session ID in the response yet).

## Follow-up work

The follow-up here will be to add the Rails logic to write chat history to our `AichatSession` table, and return the session ID as appropriate.


dashboard/app/controllers/aichat_controller.rb
```diff
@@ -40,7 +40,7 @@ def request_chat_completion(payload)
 
   private def has_required_params?
     begin
-      params.require([:newMessage, :storedMessages, :aichatParameters, :chatContext])
+      params.require([:newMessage, :aichatParameters, :chatContext])
```
bencodeorg: I think Alice has this change in a PR as well -- I think Rails treats a blank array as a missing param, which is now possible without the initial message.

apps/src/aichat/views/UserChatMessageEditor.tsx
```diff
@@ -38,7 +38,7 @@ const UserChatMessageEditor: React.FunctionComponent = () => {
           isIconOnly
           icon={{iconName: 'paper-plane'}}
           onClick={handleSubmit}
-          disabled={isWaitingForChatResponse}
+          disabled={isWaitingForChatResponse || !userMessage}
```
bencodeorg, line 41: Unrelated change  to disable the send message button until you have something in the window.

apps/src/aichat/redux/aichatRedux.ts
```diff
@@ -267,24 +302,44 @@ const aichatSlice = createSlice({
...
       const newMessage = {
         ...action.payload,
         id: newMessageId,
       };
       state.chatMessages.push(newMessage);
     },
-    removeChatMessage: (state, action: PayloadAction<number>) => {
-      const updatedMessages = state.chatMessages.filter(
-        message => message.id !== action.payload
+    removeModelUpdateMessage: (state, action: PayloadAction<number>) => {
```
bencodeorg, line 316: Tried to make this reducer name specific (and only work if the message type is `MODEL_UPDATE` so that it is less likely to be accidentally used by a future developer to delete a single actual chat message.

apps/src/aichat/views/AichatView.tsx
```diff
@@ -60,6 +61,7 @@ const AichatView: React.FunctionComponent = () => {
         studentAiCustomizations,
       })
     );
+    dispatch(setNewChatSession());
```
bencodeorg: Is this the right place to put something that I want to happen when the level changes?

apps/src/aichat/constants.ts
```diff
@@ -1,19 +1,4 @@
-import {
-  ChatCompletionMessage,
-  Role,
-  AichatInteractionStatus as Status,
-  ModelDescription,
-} from './types';
+import {ModelDescription} from './types';
 import modelsJson from '@cdo/static/aichat/modelDescriptions.json';
 
-export const initialChatMessages: ChatCompletionMessage[] = [
```
bencodeorg: Alice and I can race to merge this :)

apps/src/aichat/views/AichatView.tsx
fisher-alice: I believe so since listed dependencies (`initialSources`, `levelAichatSettings` change when level changes.
But I wonder if it make make sense to move this to `ChatWorkspace` - we already have a `useEffect` to clear chat messages based on `currentLevelId`.

apps/src/aichat/redux/aichatRedux.ts
```diff
@@ -191,6 +191,20 @@ const saveAiCustomization = async (
...
+    changedProperties.some(property =>
+      [
+        'selectedModelId',
+        'temperature',
+        'systemPrompt',
+        'retrievalContexts',
+      ].includes(property)
+    )
+  ) {
+    dispatch(setNewChatSession());
```
fisher-alice, line 219: For any customization change, we set a new chat session. When we clear chat history, we call on `newNewChatSession` afterwards. Would it make sense to include setting the new chat session within `clearChatHistory` or is there a case when we wouldn't start a new chat session after clearing?

fisher-alice (approved):
Looks good! Left a couple questions but nothing blocking.

apps/src/aichat/views/AichatView.tsx
bencodeorg: Ok so in combination with your comment [here](https://github.com/code-dot-org/code-dot-org/pull/58103/files/1847f2eb2aedfb5a941d3ef601d9817206feff5d#r1575161767), I:

a) made it so any clearing of messages also clears the session (good call!), and 
b) moved the message clearing effect into `AichatView` from `ChatWorkspace`.

My thinking is that the chat history, since we're treating it as something that initiates a new session, it feels like something that should be managed at the top level, rather than in a more specific component like `ChatWorkspace`. LMK what you think!

fisher-alice: Nice!  I agree this it makes more sense to move to `AichatView`.  Looks great - thanks for the refactor.

apps/src/aichat/redux/aichatRedux.ts
```diff
@@ -205,6 +205,20 @@ const saveAiCustomization = async (
     savedAiCustomizations,
     trimmedCurrentAiCustomizations
   );
+
+  if (
+    changedProperties.some(property =>
+      [
+        'selectedModelId',
+        'temperature',
```
sanchitmalhotra126, line 213: _technically_ temperature doesn't have to trigger a session reset. But if it's better to just be consistent I think it's fine to leave it as part of this list. I'll defer to @dancodedotorg / @samantha-code from a product/curriculum perspective

apps/src/aichat/redux/aichatRedux.ts
```diff
@@ -241,16 +259,33 @@ export const submitChatContents = createAsyncThunk(
...
       aichatContext
     );
+    console.log('chatApiResponse', chatApiResponse);
+
+    // TODO: error handling
+    thunkAPI.dispatch(setChatSessionId(chatApiResponse.sessionId));
+    thunkAPI.dispatch(
+      updateChatMessageSession({
+        id: newMessage.id,
+        sessionId: chatApiResponse.sessionId,
```
sanchitmalhotra126, line 285: Discussed offline - an alternate approach here could be to treat the list of messages & updates in the chat window as more of a generic timeline of events, and just push an internal-only "new chat session" event whenever we reset the chat session. That way we don't need to go back and manually update the first message whenever the chat session refreshes. We could even store the associated session ID with that new chat session event, and just grab it whenever we need to send it to the server. I'd be okay either way but curious what you think after spending some time with this?

apps/src/aichat/types.ts
```diff
@@ -9,6 +9,7 @@ export type ChatCompletionMessage = {
   chatMessageText: string;
   status: AichatInteractionStatus;
   timestamp?: string;
+  sessionId?: number;
```
sanchitmalhotra126, line 15: nit: might adding a comment on what this refers to (and why it's optional)?

sanchitmalhotra126 (commented):
Nice! One architectural question/idea but otherwise looks good

apps/src/aichat/redux/aichatRedux.ts
dancodedotorg: From a lesson design / teacher cognitive load perspective: I'd rather it be consistent and always reset.

apps/src/aichat/types.ts
bencodeorg: Yep good call, will do!

apps/src/aichat/redux/aichatRedux.ts
bencodeorg: hmm yeah honestly I'm torn on this stuff...I think one downside to storing the session ID with the new chat session event would be that when we want to wipe it (eg, when a model customization changes), we'd have to go find that message and update it.

I guess having iterated on this a bit, it feels cleaner to me to rely on explicit session IDs per message, rather than ordering of messages. Deciding which messages to send to the server is pretty clean as well (if there's no session ID, send nothing. if there is one, send all matching messages). And having to go back and update the last message doesn't feel particularly gross either, because we already have the ID of the message to update on hand once the request comes back from the server.

sanchitmalhotra126 (approved):

PR #58106, opened 2024-04-18
Integrates deduplicate logic in school search

hannahbergam:
Updates since I first requested review:

I removed the query limit
I wrote a separate zip search. I utilized the same file as it allows me to reuse the serialization there without a refactor
I updated the regex requirements both in the route of the search and in the front end validation
I updated the zip search component to reset the selected school to 'select a school' when the zip code changes

--------------------------------------------------
This integrates into the school search/dropdown the logic we decided on to dedup schools.

A note: I ended up writing a new zip search so as not to impact the search on production (or its many associated tests). This was already in our roadmap, but I've pulled it forward.

This allowed me to write new tests more easily as well.

How I convinced myself this will work:

I started by re-seeding schools to make sure I had refreshed data:
Screenshot 2024-04-18 at 12 04 10 PM

First I determined that pulling the 'max' of last_known_year_open would in fact return the last import:
Screenshot 2024-04-18 at 12 34 12 PM

Then I determined that the only school_types we have in our database are represented in one of these two categories: either 'private', or filtered in with their recent status for public/charter:

Screenshot 2024-04-18 at 12 34 20 PM

Then I added two 'puts' statements to see the count of rows before and after the filter and played around with the dropdown for some popular and local districts (zip codes I used included 98112, 98059, 10024, 10026, 10027). A note- 40 is what is returned when no zip search is actually run, so there are pairs of 40's in between each zip return. Generally, the dedup logic weeded out anywhere between 0 and 3 schools:
Screenshot 2024-04-18 at 12 33 29 PM

I have even more confidence this logic is correct because initially, I didn't have the extra single quotes around the current_import_year variable, and the before/after row counts were nearly cut in half. When I realized the variable wasn't being injected as a string, the counts became much closer to the original returns, only slightly reduced. 

## Warning!!

## Links

- spec: []()
- jira ticket: https://codedotorg.atlassian.net/browse/ACQ-1734
- https://codedotorg.atlassian.net/browse/ACQ-1717

## PR Checklist:

- [ ] Code is well-commented


TurnerRiley (approved):
Woooo stoked about this!!! Thank you for the detailed "how I trust this works" walkthrough!

dashboard/app/controllers/api/v1/schools_controller.rb
```diff
@@ -50,4 +50,13 @@ def search
     end
     render json: search_results
   end
+
+  # GET /dashboardapi/v1/schoolzipsearch/:q/:limit
+  def zip_search
+    search_results = Api::V1::SchoolAutocomplete.get_zip_matches(
```
bethanyaconnor: Do we even need to use SchoolAutocomplete? Or can this just be `School.where(zip: params.require(:q)`? (Relatedly, can `q` change to `zip`?)

dashboard/config/routes.rb
```diff
@@ -1007,6 +1007,7 @@
     # @see http://guides.rubyonrails.org/routing.html#specifying-constraints
     get '/dashboardapi/v1/districtsearch/:q/:limit', to: 'api/v1/school_districts#search', defaults: {format: 'json'}, constraints: {q: /[^\/]+/}
     get '/dashboardapi/v1/schoolsearch/:q/:limit(/:use_new_search)', to: 'api/v1/schools#search', defaults: {format: 'json'}, constraints: {q: /[^\/]+/}
+    get '/dashboardapi/v1/schoolzipsearch/:q/:limit', to: 'api/v1/schools#zip_search', defaults: {format: 'json'}, constraints: {q: /[^\/]+/}
```
bethanyaconnor: Can we make this regex more restrictive??

dashboard/lib/api/v1/school_autocomplete.rb
```diff
@@ -43,6 +43,23 @@ def self.get_matches(query, limit, use_new_search)
...
+    limit = format_limit(limit)
+    rows = School.limit(limit)
+    query = "#{query[0, 5]}%"
+    rows = rows.where("zip LIKE ?", query)
+    # For private schools, we don't yet have a way to determine churn/inactive
+    # schools, so we surface them all to teachers.
+    # For public and charter schools, we filter on 'known open' schools as
+    # determined by nces 'status' (see school.rb for OPEN_SCHOOL_STATUSES logic)
+    # to prevent showing duplicate/inactive schools to the user.
+    current_import_year = School.maximum(:last_known_school_year_open)
```
bethanyaconnor: not sure what the best option here is but could we hard code this? I'm torn between not wanting to hard code a constant we change every year and this extra query. 🤔

dashboard/test/lib/api/v1/school_autocomplete_test.rb
bethanyaconnor, line 1: Thanks for these tests!

dashboard/lib/api/v1/school_autocomplete.rb
hannahbergam: I'm surprised to hear this! I think the NCES ingest process is already a bit cumbersome so was hoping to not add another step. That said, because I didn't see any performance lag when I updated to this version, I wasn't thinking of performance implications. Should we discuss offline?

dashboard/config/routes.rb
hannahbergam: Sorry I should have just reverted this back to a draft once I realized the bigger direction it was going. I definitely will update this regex!

dashboard/app/controllers/api/v1/schools_controller.rb
hannahbergam: Followup- do we want or need a limit anymore? If we're already limited by zip code I'm inclined to remove it

dashboard/lib/api/v1/school_autocomplete.rb
hannahbergam: We spoke offline and decided to cache the result

hannahbergam: Caching is added:) Resolving!

apps/src/templates/SchoolZipSearch.jsx
```diff
@@ -26,7 +26,8 @@ export default function SchoolZipSearch({fieldNames, zip, disabled}) {
 
   useEffect(() => {
     if (!disabled) {
-      const searchUrl = `/dashboardapi/v1/schoolsearch/${zip}/40`;
+      setSelectedSchoolNcesId(SELECT_A_SCHOOL);
```
bethanyaconnor, line 29: remind me why this is needed?

dashboard/test/controllers/api/v1/schools_controller_test.rb
```diff
@@ -53,7 +57,8 @@ class Api::V1::SchoolsControllerTest < ActionController::TestCase
     zip: "98936",
     school_type: "public",
     latitude: "46.552088",
-    longitude: "-120.381461"
+    longitude: "-120.381461",
+    last_known_school_year_open: "2021-2022"
```
bethanyaconnor, line 61: does anywhere test that this isn't included in results?

bethanyaconnor: oooo I see this is a test. My bad -- I thought this was setup 😅

bethanyaconnor (approved):
One question but this looks great! So excited!

apps/src/templates/SchoolZipSearch.jsx
hannahbergam: I realized that if I typed in a zip code and then erased it and typed a new zip, the selected school would still remain in the dropdown even though it was no longer in the dropdown options. This clears that out between searches to avoid that scenario!

bethanyaconnor: Maybe curious if it's better to set this to null or an empty string. but this is turning into a nit so not blocking :)

hannahbergam: That all happens on the back end- it gets cleared out to an empty string if the school isn't found/selected! This is to reset that dropdown to the 'select a school' value

PR #58111, opened 2024-04-18
Gen AI lab: persist model card publish state, fix model list bug

bencodeorg:
## Warning!!

## Description

Follow-up to https://github.com/code-dot-org/code-dot-org/pull/57949 -- persist whether a student has published their model in S3.

We discussed a couple of options in [this comment thread](https://github.com/code-dot-org/code-dot-org/pull/57949#discussion_r1561663697) - I think either approach (putting this flag separate from `ModelCardInfo`, or inside of it) would result in some ugliness / special casing. Since we're already special casing within `ModelCardInfo` for the list of examples and topics, I felt like adding another special case wasn't too bad.

This keeps our read/write to S3 logic simple (ie, write something with the `AiCustomization` type).

## Testing story

Tested manually that loading/reloading the page once a model is published/unpublished, that state is retained. Also tested that I was able to make levelbuilder changes to model card info successfully.

### DIFF
--- a/apps/src/aichat/redux/aichatRedux.ts
+++ b/apps/src/aichat/redux/aichatRedux.ts
@@ -78,7 +78,6 @@ export interface AichatState {
   savedAiCustomizations: AiCustomizations;
   fieldVisibilities: {[key in keyof AiCustomizations]: Visibility};
   viewMode: ViewMode;
-  hasPublished: boolean;
 }
 
 const initialState: AichatState = {
@@ -90,7 +89,6 @@ const initialState: AichatState = {
   savedAiCustomizations: EMPTY_AI_CUSTOMIZATIONS,
   fieldVisibilities: DEFAULT_VISIBILITIES,
   viewMode: ViewMode.EDIT,
-  hasPublished: false,
 };
 
 // THUNKS
@@ -119,11 +117,11 @@ export const updateAiCustomization = createAsyncThunk(
 export const publishModel = createAsyncThunk(
   'aichat/publishModelCard',
   async (_, thunkAPI) => {
+    const {dispatch} = thunkAPI;
+    dispatch(setModelCardProperty({property: 'isPublished', value: true}));
+
     const rootState = (await thunkAPI.getState()) as RootState;
     const {currentAiCustomizations, savedAiCustomizations} = rootState.aichat;
-    const {dispatch} = thunkAPI;
-
-    dispatch(setHasPublished(true));
     await saveAiCustomization(
       currentAiCustomizations,
       savedAiCustomizations,
@@ -138,14 +136,16 @@ export const publishModel = createAsyncThunk(
 export const saveModelCard = createAsyncThunk(
   'aichat/saveModelCard',
   async (_, thunkAPI) => {
-    const rootState = (await thunkAPI.getState()) as RootState;
-    const {currentAiCustomizations, savedAiCustomizations} = rootState.aichat;
     const {dispatch} = thunkAPI;
-
-    const {modelCardInfo} = currentAiCustomizations;
+    const modelCardInfo = await (thunkAPI.getState() as RootState).aichat
+      .currentAiCustomizations.modelCardInfo;
     if (!hasFilledOutModelCard(modelCardInfo)) {
-      dispatch(setHasPublished(false));
+      dispatch(setModelCardProperty({property: 'isPublished', value: false}));
     }
+
+    const {currentAiCustomizations, savedAiCustomizations} = await (
+      thunkAPI.getState() as RootState
+    ).aichat;
     await saveAiCustomization(
       currentAiCustomizations,
       savedAiCustomizations,
@@ -305,9 +305,6 @@ const aichatSlice = createSlice({
     setViewMode: (state, action: PayloadAction<ViewMode>) => {
       state.viewMode = action.payload;
     },
-    setHasPublished: (state, action: PayloadAction<boolean>) => {
-      state.hasPublished = action.payload;
-    },
     setStartingAiCustomizations: (
       state,
       action: PayloadAction<{
@@ -336,6 +333,10 @@ const aichatSlice = createSlice({
             [customization]: studentAiCustomizations[customization],
           };
         }
+      }
+
+      if (studentAiCustomizations.modelCardInfo.isPublished) {
+        reconciledAiCustomizations.modelCardInfo.isPublished = true;
       }
 
       state.savedAiCustomizations = reconciledAiCustomizations;
@@ -398,7 +399,9 @@ const hasFilledOutModelCard = (modelCardInfo: ModelCardInfo) => {
   for (const key of Object.keys(modelCardInfo)) {
     const typedKey = key as keyof ModelCardInfo;
 
-    if (typedKey === 'exampleTopics') {
+    if (typedKey === 'isPublished') {
+      continue;
+    } else if (typedKey === 'exampleTopics') {
       if (
         !modelCardInfo['exampleTopics'].filter(topic => topic.length).length
       ) {
@@ -428,7 +431,6 @@ export const {
   setShowWarningModal,
   updateChatMessageStatus,
   setViewMode,
-  setHasPublished,
   setStartingAiCustomizations,
   setSavedAiCustomizations,
   setAiCustomizationProperty,
--- a/apps/src/aichat/redux/aichatRedux.ts
+++ b/apps/src/aichat/redux/aichatRedux.ts
@@ -99,7 +99,7 @@ const initialState: AichatState = {
 export const updateAiCustomization = createAsyncThunk(
   'aichat/updateAiCustomization',
   async (_, thunkAPI) => {
-    const rootState = (await thunkAPI.getState()) as RootState;
+    const rootState = thunkAPI.getState() as RootState;
     const {currentAiCustomizations, savedAiCustomizations} = rootState.aichat;
     const {dispatch} = thunkAPI;
 
@@ -120,7 +120,7 @@ export const publishModel = createAsyncThunk(
     const {dispatch} = thunkAPI;
     dispatch(setModelCardProperty({property: 'isPublished', value: true}));
 
-    const rootState = (await thunkAPI.getState()) as RootState;
+    const rootState = thunkAPI.getState() as RootState;
     const {currentAiCustomizations, savedAiCustomizations} = rootState.aichat;
     await saveAiCustomization(
       currentAiCustomizations,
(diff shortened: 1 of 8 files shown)


apps/src/aichat/views/modelCustomization/PublishNotes.tsx
```diff
@@ -64,7 +64,7 @@ const PublishNotes: React.FunctionComponent = () => {
                     readOnly={isDisabled(visibility)}
                   />
                 )}
-                {property !== 'exampleTopics' && (
+                {property !== 'exampleTopics' && property !== 'isPublished' && (
```
bencodeorg, line 67: Typescript did not like `!['exampleTopics', 'isPublished'].includes(property)` for some reason

apps/src/aichat/views/modelCustomization/SetupCustomization.tsx
```diff
@@ -34,12 +34,13 @@ const SetupCustomization: React.FunctionComponent = () => {
...
-        ?.aichatSettings?.availableModelIds
-  );
-  const availableModels = availableModelIds
+  const availableModelIds =
+    useAppSelector(
+      state =>
+        (state.lab.levelProperties as AichatLevelProperties | undefined)
+          ?.aichatSettings?.availableModelIds
+    ) || [];
+  const availableModels = availableModelIds.length
```
bencodeorg, line 43: I was getting page load errors here after updating some levelbuilder config -- I think this was returning true b/c `availableModels` was a blank array (truthy), causing an error on line 48 below.

apps/src/aichat/redux/aichatRedux.ts
```diff
@@ -338,6 +335,10 @@ const aichatSlice = createSlice({
         }
       }
 
+      if (studentAiCustomizations.modelCardInfo.isPublished) {
```
bencodeorg: Feels a bit hacky :) LMK if there are better ideas here.

apps/src/aichat/views/modelCustomization/SetupCustomization.tsx
bencodeorg: cc @cnbrenci, LMK if this sounds right!

cnbrenci: Yees, this should also be fixed with 

https://github.com/code-dot-org/code-dot-org/blob/65860261b1e055aa65471fec69990ddd21731d8b/apps/src/aichat/views/modelCustomization/constants.ts#L68

so it won't default to empty.

Sorry about that!

apps/src/aichat/views/AichatView.tsx
```diff
@@ -68,7 +68,7 @@ const AichatView: React.FunctionComponent = () => {
   const showPresentationToggle = () => {
     return (
       !levelAichatSettings?.hidePresentationPanel &&
-      (hasPublished ||
+      (isPublished ||
```
fisher-alice, line 71: Just a note that we discussed in [this prior comment thread](https://github.com/code-dot-org/code-dot-org/pull/57949#discussion_r1566243950) the possibility of adding a levelbuilder-side check/reminder to make sure model card fields are filled out if read-only since model card will automatically be available in presentation mode.

apps/src/aichat/redux/aichatRedux.ts
```diff
@@ -138,14 +136,16 @@ export const publishModel = createAsyncThunk(
 export const saveModelCard = createAsyncThunk(
   'aichat/saveModelCard',
   async (_, thunkAPI) => {
-    const rootState = (await thunkAPI.getState()) as RootState;
-    const {currentAiCustomizations, savedAiCustomizations} = rootState.aichat;
     const {dispatch} = thunkAPI;
-
-    const {modelCardInfo} = currentAiCustomizations;
+    const modelCardInfo = await (thunkAPI.getState() as RootState).aichat
```
fisher-alice: `await` is outside of parentheses here while it's within parentheses above on line 123. I wonder if `await` is even needed.

apps/src/aichat/redux/aichatRedux.ts
```diff
@@ -138,14 +136,16 @@ export const publishModel = createAsyncThunk(
...
+      .currentAiCustomizations.modelCardInfo;
     if (!hasFilledOutModelCard(modelCardInfo)) {
-      dispatch(setHasPublished(false));
+      dispatch(setModelCardProperty({property: 'isPublished', value: false}));
     }
+
+    const {currentAiCustomizations, savedAiCustomizations} = await (
```
fisher-alice: Nit: I see you call on `getState` twice here, once for `modelCardInfo` in `currentAiCustomizations`, then to retrieve `currentAiCustomizations` and `saved...`. Could call once on `getState` once - but this may be cleaner?

apps/src/aichat/redux/aichatRedux.ts
```diff
@@ -305,9 +305,6 @@ const aichatSlice = createSlice({
     setViewMode: (state, action: PayloadAction<ViewMode>) => {
       state.viewMode = action.payload;
     },
-    setHasPublished: (state, action: PayloadAction<boolean>) => {
-      state.hasPublished = action.payload;
-    },
     setStartingAiCustomizations: (
       state,
       action: PayloadAction<{
```
fisher-alice, line 310: Just trying to remember - are we setting `levelAichatSettings` as optional because we've refactored `aichat` levels so some of the older levels may not have this defined?

apps/src/aichat/redux/aichatRedux.ts
fisher-alice: Is this needed? I thought since `isPublished` is a property of `modelCardInfo`, it would already be 'reconciled' above.

apps/src/lab2/levelEditors/aichatSettings/ModelCardFields.tsx
```diff
@@ -16,7 +16,7 @@ const ModelCardFields: React.FunctionComponent = () => {
   return (
     <div className={moduleStyles['model-card-fields']}>
       {MODEL_CARD_FIELDS_LABELS_ICONS.map(([property, label, _]) => {
-        if (property === 'exampleTopics') {
+        if (property === 'exampleTopics' || property === 'isPublished') {
```
fisher-alice, line 19: Just confirming my understanding. This is not a levelbuilder configurable property because:
- if `modelCardInfo` is editable, it is dependent on whether model card is filled out
- if `modelCardinfo` is read-only, it will always be 'published'.

apps/src/aichat/views/modelCustomization/PublishNotes.tsx
fisher-alice: Nit: checking for these two properties here and in `ModelCardFields`. Maybe a helper function could be defined in `modelCustomiztion/utils`?

fisher-alice: Any more thoughts on customizing `PublishStatus` if `isPublished` is already set to `true` but user is just updating a model card field?

fisher-alice (commented):
Just left a few minor questions - mostly for my own understandings. Thanks!

apps/src/aichat/views/AichatView.tsx
bencodeorg: Sorry, does that relate to this change?

apps/src/aichat/redux/aichatRedux.ts
bencodeorg: I don't know 😬

apps/src/aichat/views/modelCustomization/PublishNotes.tsx
bencodeorg: Sorry I haven't had a chance to think about the warning messages further, just trying to get this publish state saved to S3 here

apps/src/lab2/levelEditors/aichatSettings/ModelCardFields.tsx
bencodeorg: Yeah that's what I was thinking...to your earlier comment about levelbuilder / model card info validation, there's probably some work to figure out what the clearest way is to message the relationship between model card info visibilities and whether the presentation view appears. But I feel like we can do that separately?

apps/src/aichat/redux/aichatRedux.ts
bencodeorg: I thought I needed it, but I totally cannot remember why and I think you're right from what I can see...removing!

apps/src/aichat/views/modelCustomization/PublishNotes.tsx
bencodeorg: And re: helper function, TS being annoying! Even if I move the exact same logic into a helper, I get this error:

```
ERROR in ./src/aichat/views/presentation/PresentationView.tsx:62:15
TS2322: Type 'string | boolean | string[]' is not assignable to type 'string | string[]'.
  Type 'boolean' is not assignable to type 'string | string[]'.
    60 |               title={label}
    61 |               titleIcon={iconName}
  > 62 |               expandedContent={modelCardInfo[property]}
       |               ^^^^^^^^^^^^^^^
    63 |               key={property}
    64 |             />
    65 |           );
```

apps/src/aichat/redux/aichatRedux.ts
bencodeorg: Yeah I had to split it up because we want to grab the state to save to S3 after we've potentially updated the `isPublished` state.

apps/src/aichat/views/AichatView.tsx
bencodeorg: Ok seeing your later comment here, does my comment there cover your question?

https://github.com/code-dot-org/code-dot-org/pull/58111/files#r1572693287

apps/src/aichat/redux/aichatRedux.ts
bencodeorg: Ahh I think I got a bad warning from my editor, it incorrectly identified getState():

image

Removed those async statements!

apps/src/aichat/views/AichatView.tsx
fisher-alice: Yes - thanks so much!

apps/src/aichat/redux/aichatRedux.ts
fisher-alice: Once `aichat` .level settings are stable, we'll probably need to do an audit of the .level files. Seems like we'd want `levelAichatSettings` to always be present. Thanks!

apps/src/aichat/views/modelCustomization/PublishNotes.tsx
fisher-alice: I appreciate you trying and sharing! I think it's fine as is - thanks!

apps/src/lab2/levelEditors/aichatSettings/ModelCardFields.tsx
fisher-alice: Sounds good. I can create a jira ticket for this then - thanks!

fisher-alice (approved):
Looks good - thanks for answering my questions!

apps/src/lab2/levelEditors/aichatSettings/ModelCardFields.tsx
fisher-alice: https://codedotorg.atlassian.net/browse/LABS-718
@samantha-code - let us know if this is P0 or not. Thanks!

samantha-code: This is not a P0 - with the close relationship with Dan we can make sure the comminicate this for pilot but we would need to be clear about this by July when other T&L members will be picking up this lab to write lessons

apps/src/aichat/redux/aichatRedux.ts
sanchitmalhotra126: `levelAichatSettings` are optional only because `levelProperties` can technically be undefined in lab2 - that said though I agree with making them required in this function at least, and moving the check up into AichatView, since for Aichat levels, they should always be defined.

PR #58114, opened 2024-04-19
[Python Lab] Enable file writing

molly-moen:
## Warning!!

## Summary
After #58051 we could read from existing files in the user's project. File writing also would not throw any errors and code could read from a file after writing to it. However, we wouldn't show any new files in the user's project after they wrote to them. This change updates the user's project correctly after each run with any new csv or text files, and any new folders. This required a few changes:
- After each run, iterate over the working directory in pyodide and send back an updated `source` object with new folders and new csv and text files. If we saw a new file that was not a csv or text file, we sent an error message.
- Update the pyodide message handler to handle updating the source when it receives an updated source.
- Instead of deleting files based on the original source, delete all files and folders in the working directory after each run. To save time this is done after reading each file/folder in the above step.
- Add a new redux thunk that handles setting the python source and saving via the project manager in one go. This allowed us to reuse logic in `PythonLabView`. 

Since I was testing some file parsing code I also made the Python Console monospaced and updating it to respect spacing.

Now you can create files and folders!
https://github.com/code-dot-org/code-dot-org/assets/33666587/fd6379af-a210-49af-9ec7-a0215626ed19

## Links

- jira ticket: [https://codedotorg.atlassian.net/browse/CT-456](CT-456)

## Testing story
Tested locally. I mostly used basic python methods for writing to files (`open`, `os` methods). I also tested using pandas to read from files, as that is what curriculum plans to use.

## Follow-up work
We have some details to figure out on the edge cases here. I started a [slack thread](https://codedotorg.slack.com/archives/C06JELCAPT9/p1713475557377839) and made a [follow-up task](https://codedotorg.atlassian.net/browse/CT-509). The main concern is we don't currently have the ability to prevent the user from running code that creates or edits non csv or text files. We can prevent them from showing up in the file browser, but it may cause confusion that users don't see an error message about this until after their program completes.

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/pythonlab/PythonlabView.tsx
+++ b/apps/src/pythonlab/PythonlabView.tsx
@@ -72,17 +72,17 @@ const defaultConfig: ConfigType = {
 };
 
 const PythonlabView: React.FunctionComponent = () => {
-  const [currentProject, setCurrentProject] = useState<MultiFileSource>();
+  //const [currentProject, setCurrentProject] = useState<MultiFileSource>();
   const [config, setConfig] = useState<ConfigType>(defaultConfig);
   const initialSources = useAppSelector(state => state.lab.initialSources);
   const channelId = useAppSelector(state => state.lab.channel?.id);
   const dispatch = useAppDispatch();
+  const source = useAppSelector(state => state.pythonlab.source);
 
   // TODO: This is (mostly) repeated in Weblab2View. Can we extract this out somewhere?
   // https://codedotorg.atlassian.net/browse/CT-499
   const setProject = useMemo(
     () => (newProject: MultiFileSource) => {
-      setCurrentProject(newProject);
       dispatch(setSource(newProject));
       if (Lab2Registry.getInstance().getProjectManager()) {
         const projectSources = {
@@ -96,18 +96,17 @@ const PythonlabView: React.FunctionComponent = () => {
 
   useEffect(() => {
     // We reset the project when the channelId changes, as this means we are on a new level.
-    setCurrentProject(
-      (initialSources?.source as MultiFileSource) || defaultProject
+    dispatch(
+      setSource((initialSources?.source as MultiFileSource) || defaultProject)
     );
-    dispatch(setSource(initialSources?.source as MultiFileSource));
   }, [channelId, dispatch, initialSources]);
 
   return (
     <div className={moduleStyles.pythonlab}>
       <div className={moduleStyles.editor}>
-        {currentProject && (
+        {source && (
           <CDOIDE
-            project={currentProject}
+            project={source}
             config={config}
             setProject={setProject}
             setConfig={setConfig}
--- a/apps/src/pythonlab/PythonlabView.tsx
+++ b/apps/src/pythonlab/PythonlabView.tsx
@@ -7,9 +7,8 @@ import {LanguageSupport} from '@codemirror/language';
 import {python} from '@codemirror/lang-python';
 import {CDOIDE} from '@cdo/apps/weblab2/CDOIDE';
 import {MultiFileSource} from '@cdo/apps/lab2/types';
-import Lab2Registry from '../lab2/Lab2Registry';
 import {useAppDispatch, useAppSelector} from '@cdo/apps/util/reduxHooks';
-import {setSource} from './pythonlabRedux';
+import {setAndSaveSource, setSource} from './pythonlabRedux';
 import PythonConsole from './PythonConsole';
 import {MAIN_PYTHON_FILE} from '@cdo/apps/lab2/constants';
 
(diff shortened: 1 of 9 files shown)


apps/src/pythonlab/PythonlabView.tsx
```diff
@@ -72,42 +71,34 @@ const defaultConfig: ConfigType = {
 };
 
 const PythonlabView: React.FunctionComponent = () => {
-  const [currentProject, setCurrentProject] = useState<MultiFileSource>();
```
molly-moen, line 75: `currentProject` wasn't necessary, it was duplicate state with `source` in redux.

apps/src/pythonlab/patches/pythonScriptUtils.ts
```diff
@@ -54,59 +61,162 @@ export function writeSource(
...
-    } catch (e) {
-      // TODO: log error better. We catch this because it should not prevent
-      // future runs.
-      console.warn(`error unlinking Pyodide file ${filePath}, ${e}`);
+  const workingDir = pyodide.FS.cwd();
+  const directoryData = pyodide.FS.lookupPath(workingDir, {}).node;
+  const directoryContents = Object.values(
+    directoryData.contents
+  ) as PyodidePathContent[];
+  const newSource = {...source};
```
ebeastlake: I think spread syntax effectively only goes one level deep and copies references for any nested objects or arrays. Does that matter here? (Alternative: Use lodash to `cloneDeep`.)

apps/src/pythonlab/patches/pythonScriptUtils.ts
```diff
@@ -54,59 +61,162 @@ export function writeSource(
...
+          sendMessage({
+            type: 'error',
+            message: `You cannot write a ${fileExtension} file in Python Lab`,
+            id: id,
+          });
+        }
+      }
+      // Delete the file now that we have handled it.
+      pyodide.FS.unlink(fullPath);
+      // Do not create or iterate through folders that start with '.' as these are hidden folders.
```
ebeastlake, line 151: Do we need/want to skip hidden files? They'll have the same prefix, right?

apps/src/pythonlab/patches/pythonScriptUtils.ts
```diff
@@ -54,59 +61,162 @@ export function writeSource(
...
+        Object.values(content.contents),
+        source,
+        newPath,
+        newFolderId,
+        id,
+        pyodide,
+        sendMessage
+      );
+      // Now that we've handled the contents, delete the folder.
+      pyodide.FS.rmdir(newPath);
```
ebeastlake, line 183: Two questions:
1. Do we need to add a follow-up task to handle errors on operations like `unlink` and `rmdir`?
2. Can we think of any scenarios in which it would be less risky or more performant to process everything, track paths to delete, and then delete it all at the end?

ebeastlake (commented):
This is awesome progress! 🎉  Just a couple of minor questions.

apps/src/pythonlab/patches/pythonScriptUtils.ts
molly-moen: good catch. I'm not sure if it matters, this is passed down from redux so I think editing it probably wouldn't change the underlying redux value. I'll copy it to be safe though.

apps/src/pythonlab/patches/pythonScriptUtils.ts
molly-moen: Late breaking update that we will show hidden folders and files. I'll add that to the follow up ticket.

apps/src/pythonlab/patches/pythonScriptUtils.ts
molly-moen: 1. Good call I'll add to the follow-up task to handle errors here. 
2. I'm not sure. We delete everything in the working directory by the end of the function. I don't think there's a way for us to delete everything in bulk (`rmdir` only works on an empty directory, [api reference](https://emscripten.org/docs/api_reference/Filesystem-API.html#FS.rmdir)). On my tests this process is very fast for a handful of files, and student projects probably aren't too big?

ebeastlake: Yeah, I think you'd need to add to "paths to delete" and then cycle through and rmdir or unlink them. I just had a neuron fire from a "deleting as you iterate" perspective. But it sounds like it's working fine and probably not an issue because we pass 
```
const directoryContents = Object.values(
    directoryData.contents
  ) as PyodidePathContent[];
  ``` 
  as an argument so it's not changing that array as we delete.

ebeastlake (approved):

apps/src/pythonlab/patches/pythonScriptUtils.ts
molly-moen: yep exactly, I think there isn't a risk of the array changing from underneath us because we have a static object we are looking at.

PR #58118, opened 2024-04-19
Update teacher dashboard header styling

etaderhold:
Update the styling on teacher dashboard page header text per the new design system.

Before:
image

After:
image

## Warning!!

## Links

[JIRA](https://codedotorg.atlassian.net/browse/TEACH-966)

## Testing story

Tested in left-to-right and right-to-left languages.

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/templates/teacherDashboard/TeacherDashboardHeader.jsx
+++ b/apps/src/templates/teacherDashboard/TeacherDashboardHeader.jsx
@@ -3,6 +3,7 @@ import React from 'react';
 import {connect} from 'react-redux';
 
 import {disabledBubblesSupportArticle} from '@cdo/apps/code-studio/disabledBubbles';
+import Link from '@cdo/apps/componentLibrary/link';
 import {sectionShape} from '@cdo/apps/templates/teacherDashboard/shapes';
 import {SectionLoginType} from '@cdo/apps/util/sharedConstants';
 import i18n from '@cdo/locale';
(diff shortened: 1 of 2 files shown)


apps/src/templates/teacherDashboard/TeacherDashboardHeader.jsx
```diff
@@ -88,32 +91,43 @@ function TeacherDashboardHeader({
...
 
   return (
-    <div style={styles.headerContainer}>
-      <SmallChevronLink
-        href="/home#classroom-sections"
-        text={i18n.viewAllSections()}
-        iconBefore
-        style={styles.linkPadding}
-      />
+    <div className={dashboardStyles.headerContainer}>
```
kobryan0619, line 94: Looks good!  I am wondering if `SmallCheveronLink` will be updated by Denys at some point, OR if we should update it now?  This looks fantastic but I am wondering this could be it's own reusable component (ex. `SmallCheveronLinkBrandRefresh`) ???

kobryan0619 (approved):
Looks good - one comment about reusable components and how to handle this "in between brand refresh" situation.  Not blocking though

apps/src/templates/teacherDashboard/TeacherDashboardHeader.jsx
etaderhold: The `SmallChevronLink` was only used in one other place, and it's really a specialized version of a more general case where you want a link with an icon next to it. The problem with putting the icon inside of the `Link` was that our new link styling puts an underline under the icon when you do that. I think support for non-underlined icons next to links in our standard components would be a useful feature to add, but the best way to do that in a standardizable way would probably require some careful thought and I feel like it's beyond the scope of this task.

kobryan0619: Makes sense!  I didn't realize the new styling would conflict here - thanks for that context!

PR #58121, opened 2024-04-19
fix: require the `shared_resources` module to instantiate shared css

stephenliang:
The SASS css file compilation process gets corrupted with an error that shared css could not be located. This is due to the shared css path not being added to the SASS compilation path. If the test `test_hoc_routes` runs by itself without other tests, this test will cause the SASS module to run but since this test doesn't require the module `shared_resources`, the shared path is not added. Normally, the rake task :ci runs which includes the test `test_pegasus_documents` which requires `shared_resources`. This file injects the `shared/css` path to the `ruby-sass` gem which allows a successful compilation in the `test_hoc_routes` unit test.

I also was able to verify that order does not matter, while `:ci` does result in tests not having a predictable order, it doesn't matter if it's PegasusTest comes before HOCTest or vice versa.

However, if you run test_hoc_routes by itself, it will cause the CSS file to be corrupted because `shared_resources` never gets required (and thus is missing the path needed to compile SCSS files correctly).

## Warning!!

## Links

- [Slack thread](https://codedotorg.slack.com/archives/C0T0PNTM3/p1713408245984199)

## Testing story

1. Ran `test_hoc_routes` without any changes, broken CSS resulted.
2. Ran `test_hoc_routes` with PR changes, CSS does not break.

## Follow-up work

N/A

## Privacy

N/A

## Security

N/A
## PR Checklist:

- [x] Code is well-commented

### DIFF
--- a/pegasus/test/test_hoc_routes.rb
+++ b/pegasus/test/test_hoc_routes.rb
@@ -2,6 +2,7 @@ require_relative './test_helper'
 require 'minitest/autorun'
 require 'rack/test'
 require 'mocha/mini_test'
+require 'shared_resources'
 require_relative 'fixtures/mock_pegasus'
 
 # Set up the rack environment to be test and recreate the Gatekeeper and DCDO in that mode.


carl-codeorg (approved):
Definitely works for me locally. Can now run the test on its own without causing CSS issues.

PR #58126, opened 2024-04-19
Added eyes test for assessments in teacher progress dashboard.

kobryan0619:
## Warning!!

<hr>

Added eyes tests that will check that the teacher can see the difference between:

- A project submitted but no feedback given
- A project submitted and feedback was given
- A project was submitted and the teacher asked the student to keep working
Baselines below:
image

image

image

Since I was on a roll, I also added the following UI tests:

- Level data opens and closes when lesson cell is clicked 
- When open a lesson cell and reload the page, the open lesson is still open
- Teacher can click the level box and the level opens in a new tab 

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/templates/sectionProgressV2/LevelDataCell.jsx
+++ b/apps/src/templates/sectionProgressV2/LevelDataCell.jsx
@@ -134,15 +134,12 @@ function LevelDataCell({
       headers={getHeadersForCell(studentId, level.id, parentLevelId, lessonId)}
     >
       <Link
+        id={'ui-test' + level.path.replaceAll('/', '-') + '-cell-data'}
         href={navigateToLevelOverviewUrl(level.url, studentId, sectionId)}
         openInNewTab
         external
         onClick={levelClickedAmplitude(sectionId, level.kind === 'assessment')}
-        className={classNames(
-          styles.expandedLevelLink,
-          linkClassName,
-          'ui-test-' + level.levelNumber
-        )}
+        className={classNames(styles.expandedLevelLink, linkClassName)}
       >
         {itemType ? (
           <ProgressIcon itemType={itemType} />
--- a/apps/src/templates/sectionProgressV2/LevelDataCell.jsx
+++ b/apps/src/templates/sectionProgressV2/LevelDataCell.jsx
@@ -138,7 +138,11 @@ function LevelDataCell({
         openInNewTab
         external
         onClick={levelClickedAmplitude(sectionId, level.kind === 'assessment')}
-        className={classNames(styles.expandedLevelLink, linkClassName)}
+        className={classNames(
+          styles.expandedLevelLink,
+          linkClassName,
+          'ui-test-' + level.levelNumber
+        )}
       >
         {itemType ? (
           <ProgressIcon itemType={itemType} />
--- a/apps/src/templates/sectionProgressV2/LevelDataCell.jsx
+++ b/apps/src/templates/sectionProgressV2/LevelDataCell.jsx
@@ -134,7 +134,7 @@ function LevelDataCell({
       headers={getHeadersForCell(studentId, level.id, parentLevelId, lessonId)}
     >
       <Link
-        id={'ui-test' + level.path.replaceAll('/', '-') + '-cell-data'}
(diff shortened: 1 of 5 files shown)


dashboard/test/ui/features/teacher_tools/teacher_dashboard/teacher_dashboard_progress_v2.feature
```diff
@@ -33,6 +33,80 @@ Scenario: Teacher can open and close Icon Key and details
...
+  And I get levelbuilder access
+  When I click selector "a:contains(Untitled Section)" once I see it to load a new page
+  And I wait until element "#uitest-teacher-dashboard-nav" is visible
+  And check that the URL contains "/teacher_dashboard/sections/"
+  And I wait until element "#uitest-course-dropdown" is visible
+  Then I append "/?enableExperiments=section_progress_v2" to the URL
+
+  # toggle to V2 progress view
+  Then I click selector "#ui-test-toggle-progress-view"
+  And element "#ui-test-progress-table-v2" is visible
```
kobryan0619: I recognize that a lot of these steps are repeated... it is just kind of the process to getting to the V2 view under the current situation.  Once we move permanently over to V2 or "default to V2" many of these steps will be cut down... 

I could also put these steps into their own `steps.rb` function, but that feels a bit overkill for just this test file...

apps/src/templates/sectionProgressV2/LevelDataCell.jsx
```diff
@@ -138,7 +138,11 @@ function LevelDataCell({
         openInNewTab
         external
         onClick={levelClickedAmplitude(sectionId, level.kind === 'assessment')}
-        className={classNames(styles.expandedLevelLink, linkClassName)}
+        className={classNames(
+          styles.expandedLevelLink,
+          linkClassName,
+          'ui-test-' + level.levelNumber
```
lfryemason: Can we put this in the id field instead of the className field or because it's a custom component that won't work?

It looks like `Link` accepts an id.

Also, could we name this something more relevant? e.g. `ui-test-level-cell-#`

dashboard/test/ui/features/teacher_tools/teacher_dashboard/teacher_dashboard_progress_v2.feature
```diff
@@ -33,6 +33,80 @@ Scenario: Teacher can open and close Icon Key and details
...
```
lfryemason: Because lines 62-75 (or some subset) are used a bunch, could we make a new step for them? Maybe in a new file in `dashboard/test/ui/features/step_definitions/` or in `section_management_steps.rb`.

lfryemason (commented):
One other thing to consider: https://codedotorg.slack.com/archives/C045UAX4WKH/p1713826360879069

dashboard/test/ui/features/teacher_tools/teacher_dashboard/teacher_dashboard_progress_v2.feature
kobryan0619: Added!

apps/src/templates/sectionProgressV2/LevelDataCell.jsx
kobryan0619: I thought a lot about this before this PR - I was nervous to use an id because all the id for levels would be the same (ex. If you opened lesson 2, level 1, the id of that cell would be the same as the cell if you opened lesson 3, level 1).  I didn't want to have duplicate id names.  I went with class names because it would at least be not "guaranteed" to have different names.  Additionally, the lesson number is not passed through already in the data to the component, and the `level.id` is hard for the test to see... THUS I landed on just using the class identifier. 

However, I know that the id is probably makes more sense to a general reader - even if they don't know ALL the reasons why it wasn't given an ID to start.  For that reason, I decided to pass in the `level.path` to the id.  This makes it a lot longer of a name... which folks might question the need of... but it is descriptive, and it is stored in the ID (vs className).

lfryemason (approved):
This is awesome! Thank you!

PR #58130, opened 2024-04-19
Use design system Tags component on Curriculum Catalog

bethanyaconnor:
Switches to using the `Tags` component over the custom -- and extremely similar -- version in the curriculum catalog. The design system `Tags` component was released in December 2023, after the curriculum catalog was launched, which is presumably why this wasn't originally used.

I'm doing this now as I had some issues when trying to create a wide view of a curriculum catalog card for the congrats page, where the topic tokens had some broken styling. This seemed to fix it and, in general seemed like a good cleanup.

What it looks like now (same as before):
Screenshot 2024-04-19 at 12 17 45 PM

## Warning!!

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/componentLibrary/tags/_Tag.tsx
+++ b/apps/src/componentLibrary/tags/_Tag.tsx
@@ -72,6 +72,7 @@ const Tag: React.FunctionComponent<TagProps> = ({
       )}
     >
       <div
+        tabIndex={0} // eslint-disable-line jsx-a11y/no-noninteractive-tabindex
         role="tooltip"
         aria-describedby={tooltipId}
         className={moduleStyles.tag}
--- a/apps/src/componentLibrary/tags/CHANGELOG.md
+++ b/apps/src/componentLibrary/tags/CHANGELOG.md
@@ -2,5 +2,8 @@
 
 All notable changes to this project will be documented in this file.
 
+## [0.2.0](https://github.com/code-dot-org/code-dot-org/pull/58130)
+* Added `tabIndex={0}` to ensure tooltip shows up during tab navigation. See PR for more discussion.
+
 ## [0.1.0](https://github.com/code-dot-org/code-dot-org/pull/55420)
 * Initial release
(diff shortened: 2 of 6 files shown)


apps/src/componentLibrary/tags/_Tag.tsx
```diff
@@ -72,6 +72,7 @@ const Tag: React.FunctionComponent<TagProps> = ({
       )}
     >
       <div
+        tabIndex={0} // eslint-disable-line jsx-a11y/no-noninteractive-tabindex
```
bethanyaconnor, line 75: Definitely open to feedback here but I needed this to maintain parity for keyboard users being able to tab over this component and see the tool tip. If there's a better way of doing this or if this should be conditional based on a prop, let me know!

kelbyhawn: Using tooltips on non-interactive elements is discouraged, but this seems to be the best solution without redesigning this part of the card! 😅

kelbyhawn (approved):
This looks great! I answered your question re: the `tabindex`, but maybe @hannahbergam has thoughts?

apps/src/componentLibrary/tags/_Tag.tsx
levadadenys: Thanks for updating this!
Yeah, adding tabIndex might not be a best solution overall, but is probably the best solution in current situation.

Once `Tooltip` component will be ready I'll probably be coming back to this component and updating it, will also take a fresh look on `Tags` overall and maybe rework it slightly

levadadenys: nit: making changes into Design System components also requires manually update the corresponding [CHANGELOG.md](https://github.com/code-dot-org/code-dot-org/blob/staging/apps/src/componentLibrary/tags/CHANGELOG.md), please add some small description there (`0.2.0` should be a good version for this update)

levadadenys (commented):
Thanks a lot for this update!
LGTM overall, just two comments:

1. Is related to Changelog (see comment)
2. `Tags` , not `Tips`. 😄  Please update the Pr heading to avoid possible future misunderstandings

bethanyaconnor:
> Tags , not Tips. 😄 Please update the Pr heading to avoid possible future misunderstandings

🤦‍♀️ wow, that's a bad one. I updated the title and description!

apps/src/componentLibrary/tags/_Tag.tsx
bethanyaconnor: done!

levadadenys (approved):
LGTM! Thanks again for updating this!

PR #48647, opened 2022-10-14
Incubator: teacher homepage banner & page

breville:
This adds a new banner to the teacher homepage ("en" only) at https://studio.code.org/home, and a new page at https://studio.code.org/incubator.

If necessary, the banner on the teacher homepage can be hidden with the `"hide_incubator_link"` DCDO flag.

It also adds the ability for a `HeaderBanner` to have an image on the right side, on wider displays, used here for the lightbulb image.  It flips to the left for RTL.

### Teacher homepage

Screenshot 2022-10-25 at 4 22 48 PM

### Incubator page

Screenshot 2022-10-25 at 4 11 58 PM

PR #58139, opened 2024-04-20
Incubator: update test

breville:
We were seeing some instances of this error when running this unit test:
```
ActiveRecord::RecordNotUnique: Mysql2::Error: Duplicate entry '80' for key 'PRIMARY'
```

A specific key was never needed in the final version of https://github.com/code-dot-org/code-dot-org/pull/48647 because we didn't end up using the idea that only a subset of users would get the banner.  It's now removed.


carl-codeorg (approved):
LGTM! :shipit:

thomasoniii (approved):
lgtm 🚀

PR #58142, opened 2024-04-20
Music add play validations

breville:
This adds two new validations: `played_chords` (aka "play notes") and `played_patterns` (aka "play drums").  

They work the same as the existing `played_sounds` validation.  

The existing `played_sounds`, `played_sounds_together`, `played_sound_triggered`, `played_sound_id` validations now works specifically for "play" blocks (and not for "play notes" or "play drums" blocks).


thomasoniii (approved):
lgtm 🚀

apps/src/music/progress/MusicValidator.ts
```diff
@@ -47,62 +49,97 @@ export default class MusicValidator extends Validator {
...
 
-        this.conditionsChecker.addSatisfiedCondition({
-          name: MusicConditions.PLAYED_SOUND_ID.name,
-          value: eventData.id,
-        });
-      }
-
-      if (eventData.when <= currentPlayheadPosition) {
-        playedNumberSounds++;
+        if (eventData.when <= currentPlayheadPosition) {
```
sanchitmalhotra126: Could we just check `eventData.when <= currentPlayheadPosition` once outside of these nested ifs, and early return if false?

apps/src/music/progress/MusicValidator.ts
```diff
@@ -47,62 +49,97 @@ export default class MusicValidator extends Validator {
...
-        });
-      }
+      this.conditionsChecker.addSatisfiedCondition({
+        name: MusicConditions.PLAYED_SOUNDS.name,
+        value: numberSounds,
+      });
+    }
+
+    // Check for up to a certain number of patterns played.
+    for (
```
sanchitmalhotra126: worth creating a small helper for these? something like `addPlayedConditions(conditionName: string, playedNumber: number)`?

apps/src/music/progress/MusicValidator.ts
breville: Good idea; done.

apps/src/music/progress/MusicValidator.ts
breville: Good idea; done.

sanchitmalhotra126 (approved):

PR #58145, opened 2024-04-21
Use upstream ProxySQL 2.6.2

snickell:
Switch back to using upstream ProxySQL and upgrade from version 2.0.x to version 2.6.2

We previously used a code.org fork of ProxySQL in order to get this patch: https://github.com/sysown/proxysql/pull/3234

That patch was applied to upstream, but we never switched back to upstream. As a result, our version of ProxySQL has been frozen at 2.0, which potentially isn't great for maintaining compatibility with the most recent aurora upgrades.

We broke the ability to provision "full stack" adhocs when we moved our fork of ProxySQL to our GitHub Organization because we haven't published a binary build, so this is a good opportunity to fix our provisioning logic. Being on mainline ProxySQL also lowers the risks with our upcoming Aurora 3 / MySQL 8 upgrade.

This is still a high risk change and should be deployed on a Friday afternoon in a limited 2nd production release so we can quickly revert, if needed.

### DIFF
--- a/cookbooks/cdo-mysql/recipes/proxy.rb
+++ b/cookbooks/cdo-mysql/recipes/proxy.rb
@@ -13,9 +13,6 @@
 #   action :upgrade
 # end
 
-# Use patched ProxySQL with fix for Aurora 2.09 bug.
-# See: https://github.com/sysown/proxysql/issues/3082
-# Can we stop using the fork? Our patch was merged upstream https://github.com/sysown/proxysql/pull/3234
 proxysql_filename = 'proxysql_2.6.2-ubuntu20_amd64.deb'
 proxysql_file = "#{Chef::Config[:file_cache_path]}/#{proxysql_filename}"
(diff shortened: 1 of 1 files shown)


sureshc:
[Canary test](https://codedotorg.slack.com/archives/C0T0PNTM3/p1714165673426189) was successful today.

Let's merge this after the AP Computer Science Principles "create task" deadline has passed next week.

snickell:
@sureshc we're clear of AP CS P, do you still want this merged in?

sureshc (approved):

sureshc:
I plan to merge and deploy this tonight or at some point this weekend. This change is not blocking the Aurora 3 / MySQL 8 upgrade, but it reduces risk if there are any ProxySQL / Aurora compatibility issues with the upgrade because we'll be on the latest release and on the mainline.

snickell:
I like your thinking

PR #58156, opened 2024-04-22
Enable unit tests in Python Lab

molly-moen:
## Warning!!

## Summary
This PR enables writing unit tests using the `unittest` package in Python Lab. There were pretty minor changes required to get this up and running fully:
First, `unittest` writes output to `stderr`, which we weren't currently handling (it was going to the browser console by default). I updated `stderr` to do the same thing as `stdout`. We can discuss if we want different styling on this.

Second, there were two default values on `unittest.main` that warranted a default override. The main one was if `exit` is set to `True` (the default), we get a lengthy error message in the output because of the way pyodide runs. The fix for this is to set `exit` to `False`. The second was by default the output is not very verbose. For example, for a run with 3 tests you would see:

```
...
----------------------------------------------------------------------
Ran 3 tests in 0.000s

OK
```

This didn't seem ideal for showing results to students. I therefore set the default verbosity to `2`, which gives us results as follows:
```
test_isupper (__main__.TestStringMethods.test_isupper) ... ok
test_split (__main__.TestStringMethods.test_split) ... ok
test_upper (__main__.TestStringMethods.test_upper) ... ok

----------------------------------------------------------------------
Ran 3 tests in 0.003s

OK
```

Curriculum confirmed we would like more verbose as the default.

In order to override these defaults, I made a patch for `unittest`, similar to what I did for matplotlib, that changes the default values. At some point we may want to move these patches into a wheel and import them, but for now this works well and is easy to iterate on.

Here's what unit testing looks like as of now (only works if it's in the `main.py` file):
Screenshot 2024-04-22 at 11 24 36 AM

## Links
- jira ticket: [CT-464](https://codedotorg.atlassian.net/browse/CT-464)

## Testing story
Tested locally.

## Follow-up work
This is the first piece to getting validation working in Python Lab. I will follow this up with a mini-spec and tasks to lay out how we fully do validation in Python Lab.

## PR Checklist:

- [ ] Code is well-commented


molly-moen:
I found a better way to do this and will have an alternate pr up soon!

PR #58167, opened 2024-04-22
Revert "Revert "[AI Tutor] CT-427: Allow students to provide feedback on AI Tutor responses""

ebeastlake:
Reverts code-dot-org/code-dot-org#58162

When I addressed PR feedback on [this PR](https://github.com/code-dot-org/code-dot-org/pull/57984/files#diff-af9dcddf9984305b99f100e07fc85cedf2165e756e597436a893b3f63b0f1558), I updated the button style I was using from the old button to the new design system button. I confirmed that it was working, but I missed t[his revert](https://github.com/code-dot-org/code-dot-org/pull/58167/commits/521752169fee22f41594d87948eb16b759289451) of a `text && <span />` condition on the old button style (which causes icon-only buttons to render off center by default).

The unreverted change was caught when it caused eyes diffs in several spots: https://codedotorg.slack.com/archives/C0T0PNTM3/p1713818395211139. I reverted the entire PR until I could put up the revised PR. This is the revision!


ebeastlake:
I confirmed that the "X" in the corner here is rendered in the left of the div (as it is in the baseline) instead of the right.

Correct baseline:

Screenshot 2024-04-24 at 1 49 28 PM

View in this PR: 

Screenshot 2024-04-24 at 1 47 35 PM

molly-moen (approved):

PR #58172, opened 2024-04-22
increase section experiment limit to 2K

davidsbailey:
see [slack](https://codedotorg.slack.com/archives/C051P2V2RN0/p1713820338514529). we are running into this limit while trying to add users to the ai-rubrics experiment. we have 1000 of these objects in production today, so unless there are signs of performance problems, I'd like to propose increasing the limit.

### DIFF
--- a/dashboard/app/models/experiments/single_section_experiment.rb
+++ b/dashboard/app/models/experiments/single_section_experiment.rb
@@ -35,7 +35,7 @@ class SingleSectionExperiment < Experiment
 
   # set a limit on the total number of SingleSectionExperiment records, to mitigate performance problems
   # when calling Experiment.get_all_enabled on hot codepaths
-  MAX_COUNT = 1_000
+  MAX_COUNT = 2_000
 
   def max_count
     MAX_COUNT


sureshc:
During peak utilization (5AM PDT to 5PM PDT) our web application is issuing about 10 SELECT statements per second on the `experiments` table and is retrieving about 4500 rows per query with an average latency of ~5ms (10X slower than our overall SELECT latency). These queries are somewhat intensive ... they retrieve all active experiments and then inspect the result set within the web application server to determine which experiments apply to the current user session. Increasing the number of experiments will increase the size of these result sets, but likely will not degrade overall database performance.

These data is from Performance Insights on the writer database instance
image

NewRelic confirms that ActiveRecord Experiment find is being executed about 10 times per second during peak usage. The latency from the web application server is 10ms.

image

Several different Rails controller actions are invoking this ActiveRecord find method. I haven't looked at how they invoke the Experiment model, but we must be caching the result set in memory because our SELECT throughput is very low (10/second) for something that has to be checked several thousand times per minute across a wide variety of high throughput controller actions:

image

sureshc (approved):
Our application is inefficient when querying this data, but is likely caching the result set, so increasing the size of the result set should not significantly degrade database system performance.

PR #58173, opened 2024-04-22
Music: test main site embed

breville:
A test to see if embedding the Music mini-player on the main site, using an `iframe`, works.  

The temporary URL is https://code.org/music/embed?channels=[channels]&library=[library].

### DIFF
--- a/pegasus/sites.v3/code.org/public/music/embed.haml
+++ b/pegasus/sites.v3/code.org/public/music/embed.haml
@@ -1,0 +1,3 @@
+%h1 Project beats embed
+
+%iframe{width: 500, height: 300, src: CDO.studio_url("/musiclab/embed?channels=channels=GMXc39iEI-pz-bECeFZGkxeHDPEdJat1k6Wpp6QxLjQ&library=launch2024", CDO.default_scheme)}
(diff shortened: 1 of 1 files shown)


bethanyaconnor (approved):
Fun! I tested this out locally and it worked! The only issue is that that fixed height feels a bit weird but I'm sure that's on your radar

Screenshot 2024-04-23 at 7 56 50 AM

I'm not super familiar with iframes -- was there anything you needed to do to get it to work or did in 🪄 just work 🪄 ?

breville:
> Fun! I tested this out locally and it worked! The only issue is that that fixed height feels a bit weird but I'm sure that's on your radar

Oh yeah, we'll have to do more styling work, but this should suffice for testing the loading and playback capability.  Glad it worked locally, which didn't happen with my setup. :)

> I'm not super familiar with iframes -- was there anything you needed to do to get it to work or did in 🪄 just work 🪄 ?

There was a bit of work with the response headers on the server side, [here](https://github.com/code-dot-org/code-dot-org/pull/57673/files#r1547394586), though it was originally done so that we could do these embeds on third party domains as well.  Most of this testing is to ensure that cross-domain requests (for both restricted and non-restricted samples) work, in this case between dashboard and pegasus on production.

PR #58175, opened 2024-04-23
Music: use pack dialog buttons

breville:
Switch to always use the pack dialog buttons added as an option in https://github.com/code-dot-org/code-dot-org/pull/58092.


thomasoniii (approved):
lgtm 🚀

PR #58178, opened 2024-04-23
Music: fix empty timeline

breville:
This fixes a bug in which a timeline with no elements was almost infinitely scrollable horizontally.

### Before

Prior to this change, an empty timeline had a really big invisible element for the function bounds, allowing the user to scroll almost infinitely into empty space.

Screenshot 2024-04-23 at 1 05 21 PM

### After

With this change, the function bounds are no longer rendered if there's nothing inside them.

Screenshot 2024-04-23 at 1 05 44 PM

The function bounds still work appropriately.

Screenshot 2024-04-23 at 1 05 48 PM

### DIFF
--- a/apps/src/music/views/TimelineSimple2Events.tsx
+++ b/apps/src/music/views/TimelineSimple2Events.tsx
@@ -4,13 +4,21 @@ import {useMusicSelector} from './types';
 import {FunctionEvents} from '../player/interfaces/FunctionEvents';
 
 /**
- * Compute the bounds for the given function, given the list of unique sounds and all functions.
+ * Compute the extents for the given function, given the list of unique sounds and all functions.
+ * Returns null if the function doesn't generate any sound events and doesn't call any functions.
  */
-const getFunctionBounds = (
+interface FunctionExtents {
+  left: number;
+  right: number;
+  top: number;
+  bottom: number;
+}
+
+const getFunctionExtents = (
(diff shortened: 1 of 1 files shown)


apps/src/music/views/TimelineSimple2Events.tsx
```diff
@@ -110,24 +119,34 @@ const TimelineSimple2Events: React.FunctionComponent<
...
-              top: 32 + uniqueFunction.top * eventHeight,
-              height:
-                (uniqueFunction.bottom - uniqueFunction.top) * eventHeight - 3,
-            }}
-          >
-            &nbsp;
-          </div>
-        ))}
+        {uniqueFunctionExtentsArray
+          .filter(uniqueFunction => uniqueFunction)
```
sanchitmalhotra126: minor - perform the `filter` in the `useMemo` above where `uniqueFunctionExtentsArray` is assigned, to save on the formatting change here?

sanchitmalhotra126 (commented):
Nice catch!

apps/src/music/views/TimelineSimple2Events.tsx
breville: Good question.  This prompted a bigger refactor, which is now committed.

sanchitmalhotra126 (approved):
Nice!

PR #58179, opened 2024-04-23
Music: save pack to channel

breville:
This is a way to save a Music Lab project's pack ID to the `Project` database model's `value` JSON column.  We want to do this so that we can later display an image based on this pack ID in places where we enumerate Music Lab projects, such as playlist UI and project UI.  By having the pack ID available in the database, we won't need to hit S3 for sources when we do that enumeration.

It turns out that a simple way to do this is to simply include the `labConfig` information that we are already saving as part of the sources.  For Music Lab, this conveniently includes both the `library` and the `packId`, both of which could prove useful when enumerating projects.

We should keep an eye on `labConfig` for this and future Lab2 labs to ensure that it continues to include metadata that we want to store in the database.

Here's an example of a `value` JSON with this change:

```
{"name":"Untitled Project","level":"https://studio.code.org/projects/music","createdAt":"2024-04-23T04:44:07.000+00:00","updatedAt":"2024-04-23T04:46:22.595+00:00","id":"AOEdvCQ5T7Zyr4YaxX5xbA","isOwner":true,"publishedAt":null,"projectType":"music","labConfig":{"music":{"library":"launch2024","packId":"electro"}}} 
```

Thanks to @molly-moen and @sanchitmalhotra126 for helping think through this.

### DIFF
--- a/apps/src/lab2/types.ts
+++ b/apps/src/lab2/types.ts
@@ -18,7 +18,9 @@ export interface Channel {
   publishedAt: string | null;
   createdAt: string;
   updatedAt: string;
-  // Optional lab-specific configuration for this project
+  // Optional lab-specific configuration for this project.  If provided, this will be saved
+  // to the Project model in the database along with the other entries in this interface,
+  // inside the value field JSON.
   labConfig?: {[key: string]: object};
 }
 
--- a/apps/src/lab2/types.ts
+++ b/apps/src/lab2/types.ts
@@ -18,6 +18,8 @@ export interface Channel {
   publishedAt: string | null;
   createdAt: string;
   updatedAt: string;
+  // Optional lab-specific configuration for this project
+  labConfig?: {[key: string]: object};
 }
 
 export type DefaultChannel = Pick<Channel, 'name'>;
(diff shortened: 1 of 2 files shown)


apps/src/lab2/types.ts
```diff
@@ -18,6 +18,8 @@ export interface Channel {
   publishedAt: string | null;
   createdAt: string;
   updatedAt: string;
+  // Optional lab-specific configuration for this project
```
molly-moen: would it be worth including in the comment here that this will be saved to the projects table?

molly-moen (approved):
Looks good, one commenting suggestion!

apps/src/lab2/types.ts
breville: Thanks for catching!  Added some comments.

sanchitmalhotra126 (approved):
Nice solution!

I imagine we might want to consider even moving `labConfig` out of sources and only saving in the projects table at some point? Since labConfig is logically more like project metadata than it is like project source code, it does make sense that it would be saved in the projects table.

breville:
> I imagine we might want to consider even moving `labConfig` out of sources and only saving in the projects table at some point? Since labConfig is logically more like project metadata than it is like project source code, it does make sense that it would be saved in the projects table.

That does sound reasonable.  Hopefully another lab or two using Lab2 and it will become clear!

PR #58188, opened 2024-04-23
Switch TTS Bucket Path

wilkie:
This updates the `TextToSpeech` module to change the path where the files are stored to be more robust toward handling data deduplication.

The motivation is that when courses are cloned, this causes a major spike in our usage of our third-party TTS generation service which is needlessly expensive. During these spikes, something around 90% of the generated content is content we already have.

Also, we have no way of knowing the provenance of the generated files in the current scheme. We are both missing a lot of audio files (which indeed silently error out on the production site itself) and have a ton of disconnected files that are no longer used hanging out in our bucket. This update doesn't ensure that this doesn't happen, but rather also records a metadata file such that we can know which files are similarly abandoned in the future and perhaps write a script to periodically clean the bucket.

On that note, this update also includes a one-off script that will copy the current files to the new scheme. See "deployment" for details on the order of operations.

## Warning!!

## Data

This is a report generated by analyzing the bucket metadata showing the redundant work of the current implementation that would be 100% avoided by the new approach:

```
There are 29967 unique strings with TTS generated audio files.
There are 57070 audio files that have been generated.

Maximum number of redundant tts strings: 180
  Within: sharon22k/180/100/2ca6aec7e2b9280a05a9d98590324740 ex: csa-prework-m4-predict-level-2023.mp3
  For which, 55 were made the same day: 2024-01-12

The highest volume day was: 2021-09-15
  On that day, 2474 strings were generated. Only some were new.
  For instance, string 11cc9ad331b803cf33f2ca1e0aa34390/ml-intro ai lab-shapes v1.mp3 is new.
  For instance, string 1626a98ef8edf7263915099c3529fd77/ml-mini project-ai lab 2.mp3 is new.
  There were 2453 redundant strings generated.

On 2020-01-21 there was a dramatic generation.
  On that day, 2217 strings were generated. Only some were new.
  For instance, string 0028a3f8bd24a1099e546c74a9c958f2/Overworld Plant Crops.mp3 is new.
  For instance, string 007b52e163344d32ca3dffa4fc66fde2/Course 4 Artist Vars 4.mp3 is new.
  There were 1556 redundant strings generated.

On 2022-05-19 there was a dramatic generation.
  On that day, 2053 strings were generated. Only some were new.
  For instance, string 14089f06d9474ce94ea56e6020b58036/CSA U2L9-L5_2022.mp3 is new.
  For instance, string 1651affc172939c9afb5f3cef2a6bee4/CSA U1L12-L5_2022.mp3 is new.
  There were 2033 redundant strings generated.

On 2023-06-28 there was a dramatic generation.
  On that day, 1992 strings were generated. Only some were new.
  For instance, string 1797bf3a29a185365951772a29971f26/U6_L03_Lists Practice_Choice_1-1.mp3 is new.
  For instance, string 3c5bf2cb21491819b2493da7949812e3/U6_L03_Lists Practice_Choice_2-4.mp3 is new.
  There were 1983 redundant strings generated.

On 2020-02-17 there was a dramatic generation.
  On that day, 1879 strings were generated. Only some were new.
  For instance, string 01fdbbbc6a4daaf588ec78427ea92f5d/courseF_farmer_ramp11.mp3 is new.
  For instance, string 0239c0248cca1709124972373f999d11/iceage_repeat.mp3 is new.
  There were 1265 redundant strings generated.

In total, in the 5 days with the most volume, we generated 10615 strings.
Of those, roughly 9290 were strings we already had.

We performed 801% more work than necessary.
```

## Links

- jira ticket: [P20-280](https://codedotorg.atlassian.net/browse/P20-280) - High call rate to Acapela-TTS
- jira ticket: [P20-762](https://codedotorg.atlassian.net/browse/P20-762) - Transition generated Acapela files to using MD5+SHA1 hashes

## Testing story

Updated the existing tests to check the old and new behavior based on stubbing the experiment value.

## Deployment strategy

We will do the following:

1. Deploy this change
2. Run the one-off script to copy the bucket contents for known text
3. Set the `updated_tts_path` DCDO variable to `true`

If there is any problem, we turn off that flag and it will revert to using the old contents.

Eventually, we will be able to permanently set that flag in a subsequent PR and then delete the old files.

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/dashboard/app/models/concerns/text_to_speech.rb
+++ b/dashboard/app/models/concerns/text_to_speech.rb
@@ -37,13 +37,20 @@ TTSSafeScrubber.tags = ['xml']
 module TextToSpeech
   extend ActiveSupport::Concern
 
-  # Pull the VOICES out of the SharedConstants
-  VOICES = SharedConstants::VOICES
-
-  # TODO: this concern actually depends on the SerializedProperties
-  # concern ... I'm not sure how best to deal with that.
+  # Pull the VOICES out of the SharedConstants (updating the locale keys to
+  # match the expectation of the I18n gem since these locale keys are using
+  # the JavaScript i18n expectations)
+  VOICES = SharedConstants::VOICES.transform_keys do |locale|
+    lang, country = locale.to_s.split('_')
+    :"#{lang}-#{country.upcase}"
+  end.freeze
 
   included do
+    # We require the SerializedProperties to be included first
+    unless respond_to? :serialized_attrs
+      raise "Module must include SerializedProperties before TextToSpeech"
+    end
+
     before_save :tts_update
 
     serialized_attrs %w(
@@ -72,7 +79,7 @@ module TextToSpeech
     use_new_path = DCDO.get('updated_tts_path', false)
     if use_new_path
       # New path
-      "generated/#{content_md5}/#{content_sha}/#{loc_voice[:VOICE]}-#{loc_voice[:SPEED]}-#{loc_voice[:SHAPE]}.mp3"
+      "#{locale}/#{content_md5}/#{content_sha}/#{loc_voice[:VOICE]}-#{loc_voice[:SPEED]}-#{loc_voice[:SHAPE]}.mp3"
     else
       # Old path
       "#{loc_voice[:VOICE]}/#{loc_voice[:SPEED]}/#{loc_voice[:SHAPE]}/#{content_md5}/#{name}.mp3"
--- a/dashboard/app/models/concerns/text_to_speech.rb
+++ b/dashboard/app/models/concerns/text_to_speech.rb
@@ -93,12 +93,16 @@ module TextToSpeech
       AWS::S3.upload_to_bucket(TTS_BUCKET, filename, resp.body, no_random: true)
 
       # Also upload metadata so we know what the text is supposed to be
-      metadata_path = "#{filename}.mp3"
-      metadata = {
-        level: name,
+      metadata_path = "#{filename.rpartition('.').first}.json"
+      metadata = {}
+      if AWS::S3.exists_in_bucket(TTS_BUCKET, metadata_path)
+        # Pull down the existing metadata
+        metadata = JSON.parse(AWS::S3.download_from_bucket(TTS_BUCKET, metadata_path))
+      end
+      metadata[name] = {
         key: key,
         locale: locale,
-        text: text,
+        text: text
       }
       AWS::S3.upload_to_bucket(TTS_BUCKET, metadata_path, metadata.to_json, no_random: true)
     end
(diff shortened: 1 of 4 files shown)


dashboard/app/models/concerns/text_to_speech.rb
```diff
@@ -37,38 +37,20 @@ def image(link, title, alt_text)
...
-  }.freeze
-
-  # TODO: this concern actually depends on the SerializedProperties
-  # concern ... I'm not sure how best to deal with that.
+  # Pull the VOICES out of the SharedConstants (updating the locale keys to
+  # match the expectation of the I18n gem since these locale keys are using
+  # the JavaScript i18n expectations)
+  VOICES = SharedConstants::VOICES.transform_keys do |locale|
+    lang, country = locale.to_s.split('_')
+    :"#{lang}-#{country.upcase}"
```
mgc1194, line 48: What if the locale does not have an underscore? My understanding is that the system's locale can be a two-letter code as well. Maybe using a `gsub` statement could be safer since it would return the same string if no matches are found.

mgc1194: I see this is coming from shared constants:
https://github.com/code-dot-org/code-dot-org/blob/ab089c08ff56c57817d4523ee1ba7803e1be70f4/lib/cdo/shared_constants.rb#L655-L681

mgc1194: is there a reason to switch the underscore to a dash?

wilkie: Can you recommend a `gsub` that would convert the JavaScript locale code `en_us` to the style the Ruby localization system expects (`en-US`)?

mgc1194: We could do something like:
```
def js_to_rb_locale(string)
  captures = string.tr('_', '-').match(/([A-Za-z]{2})(-[A-Za-z]*)?/).captures
  if captures[1]
    (captures[0] + captures[1].upcase).to_sym
  else
    captures[0].to_sym
  end
end
```
But this would be over-engineering. Your solution looks great if we are confident we are expecting the two-letter locale.

mgc1194 (approved):

daynew (approved):
Read through the PR and it makes sense to me. Thank you for all the comments/documentation.

dashboard/app/models/concerns/text_to_speech.rb
wilkie: I can handle this by adding that `return` line to ignore the locale if it does not have an underscore:

```
VOICES = SharedConstants::VOICES.transform_keys do |locale|
  return locale unless locale.to_s.include?('_')
  lang, country = locale.to_s.split('_')
  :"#{lang}-#{country.upcase}"
end.freeze
```

They all should have them, though.

mgc1194 (approved):

PR #58195, opened 2024-04-23
Gen AI: log chat history

bencodeorg:
## Warning!!

## Description

Logs chat history between students and our models. Here's the general idea:
- Update an existing session if a session ID is provided from the front end, but confirms first that the contents to be overwritten are identical to other information in the request. If not identical, create a new session.
- If no session ID provided, we write a new session.

## Testing story

I tested manually the following scenarios:
- I added new messages to the chat window (without any changes to model customizations), the same session ID was returned from the server, and the record in the database was updated after each response (via manual inspection of the database entry). 
- As I made an update to a model customization, I saw a new session ID being returned from the server.

I also added new controller tests to confirm the appropriate logging behavior in these core scenarios.

## Follow-up work

I do think there's a bit of error handling follow-up work to do here. I'm thinking we have some issue with Sagemaker, and we're returning an error to the student, we probably do not (?) want to log the user's chat message, since there was no actual interaction with a model?

I'm also thinking that if we do have an error, we'll only send the user's last message from the front end (rather than every message they've written that received an error response).


apps/src/aichat/redux/aichatRedux.ts
```diff
@@ -247,8 +247,7 @@ export const submitChatContents = createAsyncThunk(
     } = state.aichat;
 
     const aichatContext: AichatContext = {
-      userId: state.currentUser.userId,
-      currentLevelId: state.progress.currentLevelId,
+      currentLevelId: parseInt(state.progress.currentLevelId || ''),
```
bencodeorg, line 250: Converted this to an int so we don't have to deal with it on the server side.

apps/src/aichat/types.ts
```diff
@@ -16,8 +16,7 @@ export type ChatCompletionMessage = {
 };
 
 export type AichatContext = {
-  userId: number;
-  currentLevelId: string | null;
+  currentLevelId: number | null;
```
bencodeorg, line 19: Removed sending user ID from the front end because we should be able to derive the user id from the signed in user, and don't want to allow one user to update another user's history.

dashboard/test/controllers/aichat_controller_test.rb
```diff
@@ -42,44 +50,44 @@ class AichatControllerTest < ActionController::TestCase
 
   test 'pilot teacher has access test' do
     sign_in(@genai_pilot_teacher)
-    post :chat_completion, params: @valid_params
+    post :chat_completion, params: @valid_params, as: :json
```
bencodeorg, line 53: This took me a bit to debug after changing the default `storedMessages` param to an empty array (an empty array is what gets sent by the front end for the first message when a user opens the page, so seems like a sensible default).

Related to an earlier change Alice had to make to handle an empty array passed as a param in the aichat controller, I think this comment describes the issue:
https://github.com/rails/rails/issues/27215#issuecomment-263972010

dashboard/test/controllers/aichat_controller_test.rb
```diff
@@ -42,44 +50,44 @@ class AichatControllerTest < ActionController::TestCase
...
   # Post request with a profane messages param returns a failure
   test 'returns failure when chat message contains profanity' do
     sign_in(@genai_pilot_student)
     ShareFiltering.stubs(:find_failure).returns(ShareFailure.new(ShareFiltering::FailureType::PROFANITY, 'damn'))
-    post :chat_completion, params: @profanity_violation_params
-    assert_equal json_response["status"], ShareFiltering::FailureType::PROFANITY
-    assert_equal json_response["flagged_content"], "damn"
+    post :chat_completion, params: @profanity_violation_params, as: :json
+    assert_equal ShareFiltering::FailureType::PROFANITY, json_response["status"]
+    assert_equal "damn", json_response["flagged_content"]
```
bencodeorg, line 81: Argument order for `assert_equal` is expected, observed, so I switched to that. Makes failed tests easier to read :)

bencodeorg:
Working through an existing issue related to projects in tests, but should be good to review!

dashboard/app/controllers/aichat_controller.rb
```diff
@@ -44,4 +46,66 @@ def has_required_params?
...
+        params[:storedMessages] != JSON.parse(session.messages)
+      return false
+    end
+
+    true
+  end
+
+  private def update_session(session, assistant_message)
+    session.messages = updated_message_list(assistant_message).to_json
+    session.save
```
sanchitmalhotra126, line 85: Question - I've sometimes seen models saved with `.save!` which seems like it throws an error if saving fails. I'm guessing we don't want to do that here since we don't want to disrupt the user if the message sent successfully? But in that case should we do something with the return value (true/false) of `session.save`?

sanchitmalhotra126 (approved):
Nice, LGTM!

dashboard/app/controllers/aichat_controller.rb
bencodeorg: Yeah good q, I was thinking we probably need an error handling pass here (I had a few notes in the description, but didn't explicitly cover the case you're mentioning). I was wondering how we want to handle if logging isn't working for whatever reason, do we want to treat that as the whole service is down (ie, don't respond to the student)? I could see that, if we want to be sure that we have a record of every chat interaction that happens. I could also see considering logging as non-blocking, so curious for your opinions.

dashboard/test/controllers/aichat_controller_test.rb
fisher-alice: Oh nice! Thanks for sharing the comment.

dashboard/app/controllers/aichat_controller.rb
bencodeorg: Relatedly, thinking about profanity filtering...I do think we'd want to log any profane input from the student, even if we filter it before it gets sent to Sagemaker. So will have to think on how we manage keeping that in the chat history, but excluded from the conversation history that we send to Sagemaker.

dashboard/app/controllers/aichat_controller.rb
```diff
@@ -26,14 +26,16 @@ def chat_completion
...
+      content: latest_assistant_response,
     }
-    return render(status: :ok, json: payload.to_json)
+
+    session_id = log_chat_session(assistant_message)
+    return render(status: :ok, json: assistant_message.merge({sessionId: session_id}).to_json)
   end
 
-  def has_required_params?
+  private def has_required_params?
```
fisher-alice, line 38: Thanks for adding `private` back!

apps/src/aichat/aichatCompletionApi.ts
```diff
@@ -33,6 +34,7 @@ export async function postAichatCompletionMessage(
     storedMessages,
     aichatModelCustomizations,
     aichatContext,
+    ...(sessionId ? {sessionId} : {}),
```
fisher-alice, line 37: Is it possible to replace this line with just `sessionId`?

dashboard/app/controllers/aichat_controller.rb
fisher-alice, line 38: Thanks for adding `private` back!

dashboard/app/helpers/aichat_helper.rb
```diff
@@ -19,7 +19,6 @@ def self.format_inputs_for_sagemaker_request(aichat_params, stored_messages, new
     inputs = BASE_PROMPT + aichat_params[:systemPrompt]
     inputs += aichat_params[:retrievalContexts].join(" ") if aichat_params[:retrievalContexts]
     inputs = SENTENCE_BEGIN_TOKEN + wrap_as_instructions(inputs)
-    # Filter only messages from user or assistant - ignore model update messages.
```
fisher-alice, line 22: Outside of scope of your PR, but I wonder if we should rename this file with something like `aichat_sagemaker_helper`.

fisher-alice: Related [discussion](https://github.com/code-dot-org/code-dot-org/pull/57944#discussion_r1561682568) from past PR.

dashboard/app/controllers/aichat_controller.rb
fisher-alice: Interesting thought! I wonder when you begin working on profanity filtering if it would make sense to store messages as a hash with content and status. I see that you've already typed as `:json`.

fisher-alice: > if logging isn't working for whatever reason, do we want to treat that as the whole service is down (ie, don't respond to the student)? I could see that, if we want to be sure that we have a record of every chat interaction that happens.

I agree that it seems safer to treat as the whole service not working if there was a saving error.

dashboard/test/controllers/aichat_controller_test.rb
```diff
@@ -24,8 +30,14 @@ class AichatControllerTest < ActionController::TestCase
...
   setup do
+    @assistant_response = "This is an assistant response from Sagemaker"
     AichatHelper.stubs(:request_sagemaker_chat_completion).returns({status: 200, json: {body: {}}})
-    AichatHelper.stubs(:get_sagemaker_assistant_response).returns("This is an assistant response from Sagemaker")
+    AichatHelper.stubs(:get_sagemaker_assistant_response).returns(@assistant_response)
+    @controller.stubs(:storage_decrypt_channel_id).returns([123, 456])
+  end
+
+  teardown do
+    @controller.unstub(:storage_decrypt_channel_id)
```
fisher-alice: Just for my learning, this is necessary?

dashboard/test/controllers/aichat_controller_test.rb
```diff
@@ -9,10 +9,16 @@ class AichatControllerTest < ActionController::TestCase
     @genai_pilot_teacher = create :teacher, pilot_experiment: @genai_pilot.name
     pilot_section = create(:section, user: @genai_pilot_teacher)
     @genai_pilot_student = create(:follower, section: pilot_section).student_user
+
+    @default_model_customizations = {temperature: 0.5, retrievalContexts: ["test"], systemPrompt: "test"}
     @common_params = {
-      storedMessages: [{role: "user", content: "this is a test!"}],
-      aichatModelCustomizations: {temperature: 0.5, retrievalContexts: ["test"], systemPrompt: "test"},
-      aichatContext: {userId: 1, currentLevelId: "test", scriptId: 1, channelId: "test"}
+      storedMessages: [],
```
fisher-alice, line 15: Thanks for testing the empty array!

dashboard/test/controllers/aichat_controller_test.rb
```diff
@@ -90,7 +102,72 @@ class AichatControllerTest < ActionController::TestCase
   test 'returns forbidden when DCDO flag is set to `false`' do
     AichatHelper.stubs(:can_request_aichat_chat_completion?).returns(false)
     sign_in(@genai_pilot_teacher)
-    post :chat_completion, params: @valid_params
+    post :chat_completion, params: @valid_params, as: :json
     assert_response :forbidden
   end
+
```
fisher-alice, line 104: Nice thorough tests!

fisher-alice (approved):
🎉  - Just left a couple small questions.

dashboard/test/controllers/aichat_controller_test.rb
bencodeorg: tl;dr I don't think so, removed!

I saw that tricky project-related test issue we've seen in [this build](https://drone.cdn-code.org/code-dot-org/code-dot-org/5548/1/6) -- I added this as a speculative fix, and the next build passed. I just reran the build at the old commit, and it passed, so I don't think it was doing anything :). Reverted!

```
=====ERROR["test_find_channel_owner_returns_nil_for_channel_with_no_owner", "UserTest", 444.14566143699994]
 test_find_channel_owner_returns_nil_for_channel_with_no_owner#UserTest (444.15s)
Minitest::UnexpectedError:         Sequel::DatabaseError: Mysql2::Error::TimeoutError: Timeout waiting for a response from the last query. (waited 30 seconds)
            test/testing/projects_test_utils.rb:31:in `with_storage_id_for'
            test/testing/projects_test_utils.rb:9:in `with_channel_for'
            test/testing/projects_test_utils.rb:35:in `with_anonymous_channel'
            test/models/user_test.rb:4861:in `block in <class:UserTest>'
```

dashboard/app/helpers/aichat_helper.rb
bencodeorg: Sounds reasonable to me!

apps/src/aichat/aichatCompletionApi.ts
bencodeorg: tl;dr maybe?

I'm not sure how Rails would handle a param with an undefined value (maybe it would ignore it?), but a lot of the logic in Rails depends on the presence / absence of a sessionId param, so it felt safer to just fully not include it if it is undefined. Here's me playing with the difference between the two approaches:

image

dashboard/app/controllers/aichat_controller.rb
bencodeorg: _> Interesting thought! I wonder when you begin working on profanity filtering if it would make sense to store messages as a hash with content and status. I see that you've already typed as :json._

Yeah we're already storing each message as a hash with keys of `role` and `content`, might need to include a `status` key or something to handle profanity filtering...will noodle on it (and may have some questions for you based on your initial work here 🙏 )

apps/src/aichat/aichatCompletionApi.ts
fisher-alice: Gotcha - thanks for the explanation! Then leaving as is makes sense.

dashboard/app/controllers/aichat_controller.rb
sanchitmalhotra126: Re: error handling, yeah I'm a bit torn because a failure of logging is not necessary a fatal error if the student can continue chatting with the chatbot, but it would result in missed messages. I'd probably defer to @dancodedotorg / @samantha-code and ask: is it more important to try for a seamless student experience, or that we ensure that every chat interaction is logged, even if it means interruptions to students?

dancodedotorg: ^ took that question to Slack for Sam's input

PR #58198, opened 2024-04-23
Add Test button to Python Lab

molly-moen:
## Warning!!

## Summary
This adds a "test" button to Python Lab that, when clicked, will find all tests in the project and run them. Tests need to be in python files that start with `test`, and either need to be in the top-level directory or in a directory with an `__init__.py` file (this signifies that the folder is a module). If that directory is nested, all the directories above it also need `__init__.py` files. This is consistent with python's usual behavior for test discovery. We could be more permissive or use a different pattern for the file name, but any file that matches the pattern will be run when we run tests, so allowing all files could be confusing (their `main.py` could add confusing output to the console, for example). 

This change replaces the work in the closed PR #58156. In that PR I was requiring the user to run `unittest.main()` themselves, and I patched it to provide better default values. This way was better, because it allows users to write all their test files and then just click "test". I kept the update from that PR to output to `syserr`, as that is where `unittest` puts its output (and it makes sense to stream `syserr` to the console anyway).

Now a test run looks like this (note the new "Test" button)
Screenshot 2024-04-23 at 3 38 08 PM

## Links

- jira ticket: [CT-523](https://codedotorg.atlassian.net/browse/CT-523)
- [Validation requirements doc](https://docs.google.com/document/d/1RSb2XCSu-F78SvAy7enmHbAmHJU_p11AIDDe6MCpUMY/edit#heading=h.9ca950bkiejz)

## Testing story
Tested with some random unit tests that are in the main folder/nested folders.

## Follow-up work
See the requirements doc in links for more details on the validation work.


apps/src/pythonlab/PythonConsole.tsx
```diff
@@ -19,23 +19,27 @@ const PythonConsole: React.FunctionComponent = () => {
...
     // For now, restrict running python code to levelbuilders.
     if (parsedData.permissions.includes('levelbuilder')) {
-      dispatch(appendSystemMessage('Running program...'));
-      let mainFound = false;
       if (source) {
-        const code = getFileByName(source.files, MAIN_PYTHON_FILE)?.contents;
-        if (code) {
-          mainFound = true;
-          runPythonCode(code, source);
+        if (runTests) {
```
ebeastlake: Long-term, I think this would be slightly easier to read if we kept these conditions "flatter", but no need to change it for this PR!

apps/src/pythonlab/pythonHelpers/scripts.ts
```diff
@@ -0,0 +1,17 @@
+// Find all tests in the project and run them. The test files need to start with 'test'
+// and end with '.py'. The tests be in the top-level folder or in a folder with an
```
ebeastlake: nit: The tests can be

apps/src/pythonlab/pythonHelpers/scripts.ts
```diff
@@ -0,0 +1,17 @@
...
+// __init__.py file (the file can be empty). An init file signifies that the folder is a module.
+// In order to restrict to find tests in a specific folder or file we can parameterize this
+// and update the start_dir and/or pattern.
```
ebeastlake: Do we want to go ahead and expose the ability to customize the start_dir and pattern via the JavaScript API? (I don't know how likely we are to need it.)

apps/src/pythonlab/pythonHelpers/scripts.ts
```diff
@@ -0,0 +1,17 @@
...
+export const FIND_AND_RUN_ALL_TESTS = `
+import unittest
+
+def discover_and_run(start_dir = '.', pattern = 'test*.py'):
+    loader = unittest.TestLoader()
+    test_suite = loader.discover(start_dir, pattern)
+    runner = unittest.TextTestRunner(verbosity=2)
+    result = runner.run(test_suite)
+    return result
```
ebeastlake: It seems that the way you have `asyncRun` set up propagates all meaningful errors from Python (e.g., not being able to find the `start_dir` or something) even though we don't have any explicit Python handling set up. 

If we wanted more user-friendly/customized output, presumably, we could eventually return something like:
```
return {
  'total': result.testsRun,
  'failures': len(result.failures),
  'errors': len(result.errors),
  'passed': result.testsRun - len(result.failures) - len(result.errors)
}
   ```

And then parse it on the other side, but I think you had a follow-up ticket for that already!

ebeastlake (approved):
Nice progress here! Some minor questions/comments, but nothing blocking.

apps/src/pythonlab/pythonHelpers/scripts.ts
molly-moen: The return here wasn't actually doing anything, and `TextTestRunner` is already outputting reasonable results to stderr, which we put on the console, so I'm just going to remove the return here.

apps/src/pythonlab/pythonHelpers/scripts.ts
molly-moen: We are pretty likely to need to customize `pattern` to restrict to the levelbuilder-defined validation file, so I'll parameterize the file pattern now.

apps/src/pythonlab/PythonConsole.tsx
molly-moen: Since I was already cleaning things up I flattened this up a bit, lmk if that's what you had in mind!

ebeastlake: Looks great, thanks!

PR #58205, opened 2024-04-24
Music: show pack image in mini-player

breville:
With the pack ID now available in project enumeration, courtesy of https://github.com/code-dot-org/code-dot-org/pull/58179, the Music Lab mini-player is able to display the pack image.

This change also tidies up styling, and improves the RTL layout.

### LTR

Screenshot 2024-04-25 at 8 37 23 AM

### RTL

Screenshot 2024-04-25 at 8 37 12 AM

### Kudos

Thanks to @markabarrett for making a compelling case to include this.

### DIFF
--- a/apps/src/music/views/MiniMusicPlayer.tsx
+++ b/apps/src/music/views/MiniMusicPlayer.tsx
@@ -13,7 +13,6 @@ import {setUpBlocklyForMusicLab} from '../blockly/setup';
 import Lab2Registry from '../../lab2/Lab2Registry';
 import moduleStyles from './MiniMusicPlayer.module.scss';
 import FontAwesomeV6Icon from '@cdo/apps/componentLibrary/fontAwesomeV6Icon/FontAwesomeV6Icon';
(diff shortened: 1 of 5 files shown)


apps/src/music/views/MiniMusicPlayer.tsx
```diff
@@ -97,6 +98,27 @@ const MiniPlayerView: React.FunctionComponent<MiniPlayerViewProps> = ({
     setCurrentProjectId(undefined);
   }, []);
 
+  const getPackImageUrl = (packId: string) => {
```
sanchitmalhotra126: Would it be helpful to make this a function in `MusicLibrary` in case it's useful elsewhere?

sanchitmalhotra126 (approved):
Nice!

apps/src/music/views/MiniMusicPlayer.tsx
breville: Great question.  I had talked myself out of needing it yet, but have now done it.

PR #58212, opened 2024-04-24
starting ui for invitation

kobryan0619:
This PR creates a modal inviting users to use the new section progress view.  This PR:

- Creates the pop-up with the right elements. 
- Allows the user to dismiss the pop-up or accept the invitation
- Logs the events to amplitude.
- Save data that the invitation has been seen or delayed. 
- Accurately process a "delayed" invitation

image

It does not: 
- "Remember" the status of the invitation if a user switches between the tabs in the dashboard.  It DOES however remember the status of the invitation if a teacher reloads or logs out and back in.

## Links

[Ticket](https://codedotorg.atlassian.net/jira/software/projects/TEACH/boards/65?selectedIssue=TEACH-870)

## Testing story

- Tests coming in separate PR.

## Follow-up work

- Fix issue with tabs not recognizing the data.
- fix issue with image being slow to load


apps/src/templates/sectionProgressV2/InviteToV2ProgressModal.jsx
```diff
@@ -0,0 +1,164 @@
...
+        } else {
+          return true;
+        }
+      }
+    };
+
+    setInvitationOpen(showInvitation());
+  }, [dateProgressTableInvtationDelayed, hasSeenProgressTableInvite]);
+
+  const handleDismiss = () => {
```
kobryan0619: @lfryemason it is these event handlers that contain the events getting duplicated.

PR #58213, opened 2024-04-24
Add more Music Lab landing page strings

kelbyhawn:
Adds additional strings for the upcoming https://code.org/music landing page.

**Jira ticket:** [ACQ-1805](https://codedotorg.atlassian.net/browse/ACQ-1805)

### DIFF
--- a/pegasus/sites.v3/hourofcode.com/i18n/en.yml
+++ b/pegasus/sites.v3/hourofcode.com/i18n/en.yml
@@ -2011,6 +2011,7 @@
     introducing: "Introducing"
     music_lab: "Music Lab"
     introducing_music_lab: "Introducing Music Lab"
+    introducing_music_lab_strong: "Introducing **Music Lab**"
     button:
       try_music_lab: "Try Music Lab"
       learn_more: "Learn more"
@@ -2023,7 +2024,7 @@
       heading: "Create music with code"
       desc: "Get creative with your favorite artist's music and code new songs and mixes!"
     featuring_songs_by: "Featuring songs by:"
-      artist_list: "%{artist_list}, and more!"
+    artist_list: "%{artist_list}, and more!"
     artist:
       singer_songwriter: "%{artist_name}, Singer-songwriter"
       band: "%{artist_name} from %{band_name}"
--- a/pegasus/sites.v3/hourofcode.com/i18n/en.yml
+++ b/pegasus/sites.v3/hourofcode.com/i18n/en.yml
@@ -2024,7 +2024,7 @@
       heading: "Create music with code"
       desc: "Get creative with your favorite artist's music and code new songs and mixes!"
     featuring_songs_by: "Featuring songs by:"
-    artist_list: "%{artist_list}, and more!"
+    artist_list: "%{artist_list} and more!"
     artist:
       singer_songwriter: "%{artist_name}, Singer-songwriter"
       band: "%{artist_name} from %{band_name}"
--- a/pegasus/sites.v3/hourofcode.com/i18n/en.yml
+++ b/pegasus/sites.v3/hourofcode.com/i18n/en.yml
@@ -2023,6 +2023,10 @@
       heading: "Create music with code"
       desc: "Get creative with your favorite artist's music and code new songs and mixes!"
     featuring_songs_by: "Featuring songs by:"
+      artist_list: "%{artist_list}, and more!"
+    artist:
+      singer_songwriter: "%{artist_name}, Singer-songwriter"
+      band: "%{artist_name} from %{band_name}"
     student_projects:
       heading: "Listen to what students are creating"
       subheading: "Getting started is easy!"
--- a/pegasus/sites.v3/hourofcode.com/i18n/en.yml
+++ b/pegasus/sites.v3/hourofcode.com/i18n/en.yml
@@ -2011,7 +2011,7 @@
     introducing: "Introducing"
     music_lab: "Music Lab"
     introducing_music_lab: "Introducing Music Lab"
-    introducing_music_lab_strong: "Introducing **Music Lab**"
+    introducing_music_lab_strong: "Introducing **%{music_lab}**"
     button:
       try_music_lab: "Try Music Lab"
       learn_more: "Learn more"


pegasus/sites.v3/hourofcode.com/i18n/en.yml
```diff
@@ -2023,6 +2024,10 @@
       heading: "Create music with code"
       desc: "Get creative with your favorite artist's music and code new songs and mixes!"
     featuring_songs_by: "Featuring songs by:"
+    artist_list: "%{artist_list}, and more!"
+    artist:
+      singer_songwriter: "%{artist_name}, Singer-songwriter"
+      band: "%{artist_name} from %{band_name}"
```
kelbyhawn, line 2030: Here's how these will look in the markup:

Screenshot 2024-04-24 at 9 42 40 AM

pegasus/sites.v3/hourofcode.com/i18n/en.yml
```diff
@@ -2011,6 +2011,7 @@
     introducing: "Introducing"
     music_lab: "Music Lab"
     introducing_music_lab: "Introducing Music Lab"
+    introducing_music_lab_strong: "Introducing **Music Lab**"
```
kelbyhawn, line 2014: Found a good css workaround to add the icon on the `<strong>` element that wraps `**Music Lab**`. This screenshot is just showing an example locally, I'll add the css when I build out the page.

Screenshot 2024-04-24 at 10 22 03 AM

pegasus/sites.v3/hourofcode.com/i18n/en.yml
```diff
@@ -2023,6 +2024,10 @@
       heading: "Create music with code"
       desc: "Get creative with your favorite artist's music and code new songs and mixes!"
     featuring_songs_by: "Featuring songs by:"
+    artist_list: "%{artist_list}, and more!"
```
breville, line 2027: Will all languages work if we programmatically generate a comma-separated list and insert it here?

bethanyaconnor: To Brendan's question: Are you using a programmatically inserted list here because we don't want the names translated? Or because we want to be able to easily change the list? Or some other reason?

kelbyhawn: I initially did it this way because we don't want the names translated, but now that you mention it there is probably a high chance that the list will change.

bethanyaconnor: I'm okay moving forward with this. I think Brendan brings up a valid concern, but I'm not sure of the solution, other than spelling out all the names.

I think we do use names in Dance Party strings, though, right?

pegasus/sites.v3/hourofcode.com/i18n/en.yml
mgc1194: I'm curious if we want **Music Lab** to be translated or if there is any intent to keep it as a brand name.
I am asking this because we used to translate CS Discoveries and then turn it into a trademark name.
I could see **Music Lab** being something we end up registering.

pegasus/sites.v3/hourofcode.com/i18n/en.yml
mgc1194: I have seen a list of artist names in some Dance Party strings in Crowdin before. 
There are no specific rules on how to handle proper names. While it is true that [some languages](https://linguistics.stackexchange.com/questions/6737/what-languages-use-the-serial-comma) don't use comma-separated values, Kelby's approach allows for simultaneously updating the list for all languages.

It would be a question of preference: Do we prefer to acustome more to other languages? or Do we prefer to have an accurate list for all users?

pegasus/sites.v3/hourofcode.com/i18n/en.yml
kelbyhawn: Good question @mgc1194! I asked Karim and Michael to weigh in on [Slack](https://codedotorg.slack.com/archives/C04TH8400AU/p1714067901608119).

pegasus/sites.v3/hourofcode.com/i18n/en.yml
kelbyhawn: Who is the decider in this case? And I'll remove the Oxford comma, we don't have that on code.org/dance (which **does not** have artists name translations in a similar case where this will be used): https://github.com/code-dot-org/code-dot-org/blob/0c302bcd22c299f2136e32d0e8a23ff8ce5112ca/pegasus/sites.v3/code.org/views/dance/featured_artists.haml#L13

pegasus/sites.v3/hourofcode.com/i18n/en.yml
kelbyhawn: Closing the loop: we are going to keep this as-is for now. See the Slack convo above for more context.

pegasus/sites.v3/hourofcode.com/i18n/en.yml
bethanyaconnor: I think you're the decider Kelby ;) 

What I'm hearing Mario say is -- if we think that list might change, it's probably a good idea to go with your change so that the list is always accurate. If it's definitely not going to change, it could be good to make sure the string is properly formatted for all languages.

kelbyhawn: Ooo how exciting haha! I'm going to go with what I have, Dani shared more artist info with me yesterday and I expect it to change. We can always revisit if needed.

pegasus/sites.v3/hourofcode.com/i18n/en.yml
kelbyhawn: Re-opening the loop 😅 The XT has decided they **do not** want "Music Lab" to be translated, so I updated this string to make that so. Will update existing strings if needed in a separate PR.

Screenshot of this update in the markup:
Screenshot 2024-04-26 at 9 04 24 AM

bethanyaconnor (approved):

pegasus/sites.v3/hourofcode.com/i18n/en.yml
breville: TIL [Intl.ListFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/ListFormat).

PR #58214, opened 2024-04-24
Lab2: enable share mode

sanchitmalhotra126:
## Warning!!

## Description

This enables the "share" / "play" mode for Lab2 labs (i.e. the view presented when navigating to a /projects/\<app\>/\<channelId\> link, without /view or /edit). Previously, we were just redirecting the user back to /view or /edit, but we now allow access to this page.

One callout - because the share page is presented without a header, the code in header.js responsible for setting the current level ID never runs, so we instead grab the level ID from app options and store it in Redux in Lab2Wrapper. Longer term there's a discussion to be had about decoupling the code that initializes progress from header.js, as I'm sure there would be instances where we'd want some progress-related information available even if the header is not displayed.

Music Lab in "share" view (currently just displays the existing UI, but will eventually display the Music Lab share UI as part of LABS-654)

Screenshot 2024-04-24 at 3 34 16 PM (2)

## Links

Related to https://codedotorg.atlassian.net/browse/LABS-654

## Testing story

Tested this with standalone projects.

### DIFF
--- a/apps/src/lab2/lab2Redux.ts
+++ b/apps/src/lab2/lab2Redux.ts
@@ -57,6 +57,8 @@ export interface LabState {
   validationState: ValidationState;
   // Level properties for the current level.
   levelProperties: LevelProperties | undefined;
+  // If this lab should presented in a "share" or "play-only" view, which may hide certain UI elements.
+  isShareView: boolean | undefined;
 }
 
 const initialState: LabState = {
@@ -67,6 +69,7 @@ const initialState: LabState = {
   initialSources: undefined,
   validationState: getInitialValidationState(),
   levelProperties: undefined,
+  isShareView: undefined,
 };
 
 // Thunks
@@ -275,6 +278,9 @@ const labSlice = createSlice({
       state.levelProperties = action.payload.levelProperties;
       state.initialSources = action.payload.initialSources;
     },
+    setIsShareView(state, action: PayloadAction<boolean>) {
+      state.isShareView = action.payload;
+    },
   },
   extraReducers: builder => {
     builder.addCase(setUpWithLevel.fulfilled, state => {
@@ -423,8 +429,13 @@ async function cleanUpProjectManager() {
   Lab2Registry.getInstance().clearProjectManager();
 }
 
-export const {setIsLoading, setPageError, clearPageError, setValidationState} =
-  labSlice.actions;
+export const {
+  setIsLoading,
+  setPageError,
+  clearPageError,
+  setValidationState,
+  setIsShareView,
+} = labSlice.actions;
 
 // These should not be set outside of the lab slice.
 const {setChannel, onLevelChange} = labSlice.actions;
--- a/apps/src/lab2/projects/ProjectContainer.tsx
+++ b/apps/src/lab2/projects/ProjectContainer.tsx
@@ -80,7 +80,7 @@ const ProjectContainer: React.FunctionComponent<ProjectContainerProps> = ({
     return () => {
       // If we have an early return, we will abort the promise in progress.
       // An early return could happen if the level is changed mid-load.
-      promise.abort();
+      promise?.abort();
     };
   }, [
     channelId,
--- a/apps/src/lab2/projects/utils.ts
+++ b/apps/src/lab2/projects/utils.ts
@@ -36,7 +36,7 @@ export function getServerLevelId(): number | undefined {
  * Returns if the lab should presented in a share/play-only view,
  * if present in App Options. Only used in standalone project levels.
  */
-export function isShareView(): boolean | undefined {
+export function getIsShareView(): boolean | undefined {
   const appOptions = getScriptData('appoptions') as PartialAppOptions;
   return appOptions.share;
 }
--- a/apps/src/lab2/projects/utils.ts
+++ b/apps/src/lab2/projects/utils.ts
@@ -27,7 +27,7 @@ export function getStandaloneProjectId(): string | undefined {
  * This is specifically used in scenarios where the level ID is not provided
  * by other means (for example via header.js)
  */
-export function getServerLevelId(): number | undefined {
+export function getAppOptionsLevelId(): number | undefined {
   const appOptions = getScriptData('appoptions') as PartialAppOptions;
   return appOptions.levelId;
 }
--- a/apps/src/lab2/projects/utils.ts
+++ b/apps/src/lab2/projects/utils.ts
@@ -7,8 +7,6 @@ interface PartialAppOptions {
   channel: string;
   levelId: number;
   share: boolean;
-  iframeEmbed: boolean;
-  iframeEmbedAppAndCode: boolean;
 }
 
 /**
@@ -24,24 +22,23 @@ export function getStandaloneProjectId(): string | undefined {
   return appOptions.channel;
 }
 
+/**
+ * Returns the level ID provided by App Options, if available.
+ * This is specifically used in scenarios where the level ID is not provided
+ * by other means (for example via header.js)
+ */
 export function getServerLevelId(): number | undefined {
   const appOptions = getScriptData('appoptions') as PartialAppOptions;
   return appOptions.levelId;
 }
 
+/**
+ * Returns if the lab should presented in a share/play-only view,
+ * if present in App Options. Only used in standalone project levels.
+ */
 export function isShareView(): boolean | undefined {
   const appOptions = getScriptData('appoptions') as PartialAppOptions;
   return appOptions.share;
-}
-
-export function isIframeEmbed(): boolean | undefined {
-  const appOptions = getScriptData('appoptions') as PartialAppOptions;
-  return appOptions.iframeEmbed;
-}
-
-export function isIframeEmbedAppAndCode(): boolean | undefined {
-  const appOptions = getScriptData('appoptions') as PartialAppOptions;
-  return appOptions.iframeEmbedAppAndCode;
 }
 
 /**
--- a/apps/src/lab2/projects/utils.ts
+++ b/apps/src/lab2/projects/utils.ts
@@ -1,4 +1,4 @@
-import getScriptData from '@cdo/apps/util/getScriptData';
+import getScriptData, {hasScriptData} from '@cdo/apps/util/getScriptData';
 import {ProjectFile} from '../types';
 
 // Partial definition of the App Options structure, only defining the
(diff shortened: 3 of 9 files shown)


apps/src/lab2/views/Lab2Wrapper.tsx
```diff
@@ -28,6 +31,16 @@ const Lab2Wrapper: React.FunctionComponent<Lab2WrapperProps> = ({children}) => {
...
       state.lab.pageError?.errorMessage || state.lab.pageError?.error?.message
   );
+  const dispatch = useAppDispatch();
+
+  // Store the level ID provided by App Options in redux. This is needed
+  // on pages without a header, such as the share view.
+  const serverLevelId = getServerLevelId();
+  useEffect(() => {
+    if (serverLevelId) {
+      dispatch(setCurrentLevelId(serverLevelId.toString()));
```
sanchitmalhotra126: This is effectively a replacement for https://github.com/code-dot-org/code-dot-org/blob/8cf1b3b09edb95051da918d2db669bd81fa8aa15/apps/src/code-studio/header.js#L140 which is not called here because there's no header in the share view.

apps/src/lab2/projects/utils.ts
```diff
@@ -21,9 +24,24 @@ export function getStandaloneProjectId(): string | undefined {
...
+
+export function isIframeEmbed(): boolean | undefined {
+  const appOptions = getScriptData('appoptions') as PartialAppOptions;
+  return appOptions.iframeEmbed;
+}
+
+export function isIframeEmbedAppAndCode(): boolean | undefined {
+  const appOptions = getScriptData('appoptions') as PartialAppOptions;
+  return appOptions.iframeEmbedAppAndCode;
 }
```
breville: Not very familiar with all of these.  Do we want to support all of them in Lab2?

sanchitmalhotra126: I included them since we eventually may want to? iframeEmbed is the same as the share view, just embedded in an iframe. iframeEmbedAppAndCode is supposed to present a readonly view of the full workspace in an iframe.

Edit: but also fine if we'd rather only expose what we're actually using (just 'share' for now)

breville: Over the Lab2 process I think we've found value in reducing our surface area to key features we want to support going forward, so I'd encourage just supporting must-have scenarios to begin with.  Though maybe add a comment alluding to additional scenarios we could re-introduce if there is demand.

apps/src/lab2/lab2Redux.ts
```diff
@@ -57,6 +57,8 @@ export interface LabState {
   validationState: ValidationState;
   // Level properties for the current level.
   levelProperties: LevelProperties | undefined;
+  // If this lab should presented in a "share" or "play-only" view, which may hide certain UI elements.
```
fisher-alice, line 60: Nit: '...lab should [be] presented...'

apps/src/lab2/lab2Redux.ts
```diff
@@ -67,6 +69,7 @@ const initialState: LabState = {
   initialSources: undefined,
   validationState: getInitialValidationState(),
   levelProperties: undefined,
+  isShareView: undefined,
```
fisher-alice, line 72: Nit: why not initialize to 'false'?

apps/src/lab2/projects/ProjectContainer.tsx
```diff
@@ -80,7 +80,7 @@ const ProjectContainer: React.FunctionComponent<ProjectContainerProps> = ({
     return () => {
       // If we have an early return, we will abort the promise in progress.
       // An early return could happen if the level is changed mid-load.
-      promise.abort();
+      promise?.abort();
```
fisher-alice, line 83: Why is `?` now needed?

apps/src/lab2/projects/utils.ts
```diff
@@ -1,11 +1,12 @@
...
-import {ProjectFile, ProjectType} from '../types';
+import {ProjectFile} from '../types';
 
 // Partial definition of the App Options structure, only defining the
 // pieces we need in this component.
 interface PartialAppOptions {
   channel: string;
-  projectType: ProjectType;
+  levelId: number;
+  share: boolean;
```
fisher-alice, line 9: Maybe name as `isShareView` for clarity.

apps/src/lab2/projects/utils.ts
```diff
@@ -21,9 +22,23 @@ export function getStandaloneProjectId(): string | undefined {
   return appOptions.channel;
 }
 
-export function getProjectType(): string {
+/**
+ * Returns the level ID provided by App Options, if available.
+ * This is specifically used in scenarios where the level ID is not provided
+ * by other means (for example via header.js)
+ */
+export function getServerLevelId(): number | undefined {
```
fisher-alice: Nit: If `serverLevelId` and `levelId` have the same value, is there a reason this is not called `getLevelId`?

apps/src/lab2/views/Lab2Wrapper.tsx
```diff
@@ -29,6 +37,28 @@ const Lab2Wrapper: React.FunctionComponent<Lab2WrapperProps> = ({children}) => {
...
   );
 
+  // Store some server-provided data in redux.
+
+  const dispatch = useAppDispatch();
+  const currentLevelId = useAppSelector(state => state.progress.currentLevelId);
+
+  // Store the level ID provided by App Options in redux if necessary.
+  // This is needed on pages without a header, such as the share view.
+  const serverLevelId = getServerLevelId();
```
fisher-alice: Same question as above about name - a little unclear whether `serverLevelId` is actually the same as `levelId`.

apps/src/lab2/projects/utils.ts
```diff
@@ -1,11 +1,12 @@
 import getScriptData from '@cdo/apps/util/getScriptData';
-import {ProjectFile, ProjectType} from '../types';
+import {ProjectFile} from '../types';
 
 // Partial definition of the App Options structure, only defining the
 // pieces we need in this component.
 interface PartialAppOptions {
```
fisher-alice, line 6: Should `AppOptions` also be updated in `lab2/types` with `share`?

fisher-alice (approved):
Thanks for taking care of this! Just left a handful of minor questions/comments.

dashboard/app/helpers/levels_helper.rb
```diff
@@ -732,6 +732,12 @@ def blockly_options
     app_options
   end
 
+  def lab2_options
+    app_options = {channel: view_options[:channel], level_id: @level.id}
+    app_options.merge! level_view_options(@level.id)
```
molly-moen: do we need all of the `level_view_options`?

apps/src/lab2/lab2Redux.ts
sanchitmalhotra126: I think I wanted to distinguish between this value being unset (not passed in from the server) vs explicitly set to true or false

apps/src/lab2/projects/ProjectContainer.tsx
sanchitmalhotra126: If none of the previous conditions are met, `promise` would be undefined, so this is to prevent an undefined exception.

apps/src/lab2/projects/utils.ts
sanchitmalhotra126: So this is reflecting a slice of the app_options hash created on the server, which uses the key `share` (see here: https://github.com/code-dot-org/code-dot-org/blob/ff1c0a4ffd5bf2e619297166d0f1ac45ae394ba6/dashboard/app/controllers/projects_controller.rb#L380

apps/src/lab2/views/Lab2Wrapper.tsx
sanchitmalhotra126: Yeah I don't love this name 🙂 we have other instances of "serverLevelId" in the codebase but it's not immediately clear what it means. I guess I'm trying to find a way to describe the level ID sent via app options, vs the level ID available in progress redux. Maybe just `appOptionsLevelId`?

apps/src/lab2/projects/utils.ts
sanchitmalhotra126: Hmm good question, we don't even use the `AppOptions` type in `lab2/types` as far as I'm aware, so maybe we just need to move this interface into `lab2/types`

dashboard/app/helpers/levels_helper.rb
sanchitmalhotra126: Technically no at this point; from what I can tell though, this just adds `share`, `iframe_embed`, and `iframe_embed_app_and_code`. Do you think we should just manually pluck out the values we care about? I'd be fine either way

apps/src/lab2/views/Lab2Wrapper.tsx
fisher-alice: I think 'appOptionsLevelId' is more clear - thanks!.

~~But you made a point earlier about names being consistent with those from the front end so not sure.~~ Nvm - I see this is different.

fisher-alice: > we have other instances of "serverLevelId" in the codebase but it's not immediately clear what it means. 

Yeah - there's also a `serverScriptLevelId` in `appOptions`!

apps/src/lab2/projects/utils.ts
fisher-alice: > we don't even use the AppOptions type in lab2/types as far as I'm aware

Interesting! So it's more for documentation? I was wondering if we could use `Omit` like we did [here](https://github.com/code-dot-org/code-dot-org/blob/731456039cdbb9b61e11b5c83d761f379d483aa1/apps/src/aichat/types.ts#L74), but I see that in `AppOptions` we use `level: Level` instead of `levelId: number`.

dashboard/app/helpers/levels_helper.rb
fisher-alice: I was curious about which properties these were so printed the values. So in terms of readability, I think manually assigning values may be more clear.

molly-moen: My motivation for only sending the values we actually use is it makes it clear what in app options lab2 is still dependent on, and forces us to think about new usage of app options.

apps/src/lab2/projects/utils.ts
sanchitmalhotra126: Yeah, that `AppOptions` type was added in the early days of Lab2 - our philosophy since then has been to try to use AppOptions as little as possible, which is why we have this `PartialAppOptions` where we only define the properties we use.

dashboard/app/helpers/levels_helper.rb
sanchitmalhotra126: Yep agreed, that makes sense. Will update!

molly-moen (approved):

PR #58215, opened 2024-04-24
Revert "Revert "Integrates deduplicate logic in school search""

hannahbergam:
Reverts code-dot-org/code-dot-org#58196
This is a clean revert: no additional changes.

After some investigation with @bethanyaconnor this morning, we believe this could have been an issue with the way our unit tests are parallelized on the 16 test servers. Those databases didn't align with our test database. I'd like to merge the revert later this afternoon to do some testing (and potentially see if a test build restart can help solve the issue) and will revert again if necessary.


bethanyaconnor (approved):
Looks good as long as the DOTD is okay with ceding the test server to you for a while! (I know you've already asked, so good to go there)

PR #54972, opened 2023-11-15
LTI teacher email input support

daynew:
LTI doesn't guarantee to provide an Email when a teacher is signing up for a Code.org account. However, at Code.org, we want every Teacher account to have an email associated with it so we can reach out to them or find their User data if there are issues.

This PR adds an email input field to the account signup page if the LMS doesn't provide an email address and the user is a Teacher.

## Links
* [JIRA](https://codedotorg.atlassian.net/browse/P20-556)

## Testing story
* Unit tests
* Manual test

PR #58227, opened 2024-04-24
Persist school_info from finish registration page

carl-codeorg:
When we [modified the partial registration code](https://github.com/code-dot-org/code-dot-org/pull/54972) to allow account creation via LTI launch, we inadvertently started omitting the nested attributes for `school_info` when creating new teacher accounts.

This adds handling for `school_info` to make it behave more like `authentication_option`s, where we explicitly take the nested attributes from the params and attach them to the rehydrated user object.

Screenshot 2024-04-24 at 1 17 50 PM

```
[development] dashboard > User.find_by_email('teacher@schoolinfo.com').school_info
=>
#<SchoolInfo:0x00007f7ede271730
 id: 5,
 country: "US",
 school_type: "private",
 zip: 11111,
 state: "Alabama",
 school_district_id: nil,
 school_district_other: nil,
 school_district_name: nil,
 school_id: nil,
 school_other: nil,
 school_name: "school info test",
 full_address: nil,
 created_at: Thu, 25 Apr 2024 03:17:52.000000000 UTC +00:00,
 updated_at: Thu, 25 Apr 2024 03:17:52.000000000 UTC +00:00,
 validation_type: "none">
```

```
[development] dashboard > UserSchoolInfo.where(user_id: 23)
  UserSchoolInfo Load (24.7ms)  SELECT `user_school_infos`.* FROM `user_school_infos` WHERE `user_school_infos`.`user_id` = 23
=>
[#<UserSchoolInfo:0x00007f7edc451818
  id: 3,
  user_id: 23,
  start_date: Thu, 25 Apr 2024 03:17:52.000000000 UTC +00:00,
  end_date: nil,
  school_info_id: 5,
  last_confirmation_date: Thu, 25 Apr 2024 03:17:52.000000000 UTC +00:00,
  created_at: Thu, 25 Apr 2024 03:17:52.000000000 UTC +00:00,
  updated_at: Thu, 25 Apr 2024 03:17:52.000000000 UTC +00:00>]
```

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/dashboard/lib/services/user.rb
+++ b/dashboard/lib/services/user.rb
@@ -46,7 +46,7 @@ class Services::User
     if params_school_info
       # Disable validation because the school info fields are optional on the registration page
       params_school_info[:validation_type] = SchoolInfo::VALIDATION_NONE
-      user.school_info = SchoolInfo.create(params_school_info)
+      user.assign_attributes(school_info_attributes: params_school_info)
     end
   end
 end
--- a/dashboard/lib/services/user.rb
+++ b/dashboard/lib/services/user.rb
@@ -3,11 +3,9 @@ class Services::User
   # new values to the given User instance.
   def self.assign_form_params(user, params)
     assign_auth_option_params(user, params)
-    assign_school_info_params(user, params)
-    # Remove nested attributes, otherwise we might accidentally create new
-    # nested objects rather than updating existing ones.
-    # '_attributes' is how active_record identifies nested attrs.
-    flat_params = params.reject {|key, _| key.ends_with? '_attributes'}
+    # Remove nested attributes for authentication_options to prevent
+    # duplicates when email is ommitted from the LTI launch response
+    flat_params = params.except('authentication_options_attributes')
     user.assign_attributes(flat_params.compact)
   end
 
@@ -32,20 +30,6 @@ class Services::User
         # Apply the form params to the AuthenticationOption
         user_ao.assign_attributes(params_ao.compact)
       end
-    end
-  end
-
-  # The user model accepts nested attributes for school_info. On the finish
-  # signup form, users can associate themselves with a school. This method
-  # looks for the school_info attributes in the form params and updates the
-  # User object. The user_school_infos record is created separately, in an
-  # after_save hook in the User model, so it doesn't need to be created here.
-  def self.assign_school_info_params(user, params)
-    # "school_info_attributes"=>{"country"=>"US", "school_type"=>"public", "school_state"=>"Washington", "school_name"=>"Foo School", "school_zip"=>"98021"}
-    params_school_info = params['school_info_attributes']
-    if params_school_info
-      # Disable validation because the school info fields are optional on the registration page
-      user.assign_attributes(school_info_attributes: params_school_info)
     end
   end
 end
--- a/dashboard/lib/services/user.rb
+++ b/dashboard/lib/services/user.rb
@@ -3,6 +3,7 @@ class Services::User
   # new values to the given User instance.
   def self.assign_form_params(user, params)
     assign_auth_option_params(user, params)
+    assign_school_info_params(user, params)
     # Remove nested attributes, otherwise we might accidentally create new
     # nested objects rather than updating existing ones.
     # '_attributes' is how active_record identifies nested attrs.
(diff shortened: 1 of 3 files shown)


dashboard/lib/services/user.rb
```diff
@@ -33,4 +34,19 @@ def self.assign_auth_option_params(user, params)
...
+  # looks for the school_info attributes in the form params and updates the
+  # User object. The user_school_infos record is created separately, in an
+  # after_save hook in the User model, so it doesn't need to be created here.
+  def self.assign_school_info_params(user, params)
+    # "school_info_attributes"=>{"country"=>"US", "school_type"=>"public", "school_state"=>"Washington", "school_name"=>"Foo School", "school_zip"=>"98021"}
+    params_school_info = params['school_info_attributes']
+    if params_school_info
+      # Disable validation because the school info fields are optional on the registration page
+      params_school_info[:validation_type] = SchoolInfo::VALIDATION_NONE
+      user.school_info = SchoolInfo.create(params_school_info)
```
carl-codeorg: @bethanyaconnor Do we need to do some kind of lookup for existing school_infos here? Or does the deduplication logic in `SchoolInfoDeduplicator` handle this?

artem-vavilov: I believe the purpose of this is to create `school_info` when the user saves:
```suggestion
      user.assign_attributes(school_info_attributes: params_school_info)
```

dashboard/lib/services/user.rb
```diff
@@ -3,6 +3,7 @@ class Services::User
   # new values to the given User instance.
   def self.assign_form_params(user, params)
     assign_auth_option_params(user, params)
+    assign_school_info_params(user, params)
     # Remove nested attributes, otherwise we might accidentally create new
     # nested objects rather than updating existing ones.
     # '_attributes' is how active_record identifies nested attrs.
```
bethanyaconnor: Maybe a question for the future but why is removing all nested attributes necessary? If there was a specific attribute that was causing problems, could we handle that one separately and allow Rails to handle the rest?

dashboard/lib/services/user.rb
```diff
@@ -33,4 +34,19 @@ def self.assign_auth_option_params(user, params)
...
+  # signup form, users can associate themselves with a school. This method
+  # looks for the school_info attributes in the form params and updates the
+  # User object. The user_school_infos record is created separately, in an
+  # after_save hook in the User model, so it doesn't need to be created here.
+  def self.assign_school_info_params(user, params)
+    # "school_info_attributes"=>{"country"=>"US", "school_type"=>"public", "school_state"=>"Washington", "school_name"=>"Foo School", "school_zip"=>"98021"}
+    params_school_info = params['school_info_attributes']
+    if params_school_info
+      # Disable validation because the school info fields are optional on the registration page
+      params_school_info[:validation_type] = SchoolInfo::VALIDATION_NONE
```
bethanyaconnor: is the `preprocess_school_info` step skipped now? [That function](https://github.com/code-dot-org/code-dot-org/blob/356a72ad8184170f48984a78a82c17b25efe3419/dashboard/app/models/user.rb#L364) also sets this validation type to none

dashboard/test/models/user_test.rb
```diff
@@ -5194,4 +5194,29 @@ def hide_scripts_in_sections(section1, section2)
     create :follower, section: section, user: student
     assert_empty teacher.reload.followers
   end
+
+  test 'persists nested attributes when creating a teacher from a partial registration' do
```
bethanyaconnor, line 5198: Thanks for adding this test!

dashboard/lib/services/user.rb
carl-codeorg: I honestly don't remember - I think it might have also worked to handle auth options in a special way and leave the rest. The change was specifically around handling auth options, because we wanted to be able to set the email address if it was missing in the LTI response. The idea was to ensure that we are updating the email in the existing auth_option instead of creating another one with the user-input email.

dashboard/lib/services/user.rb
carl-codeorg: I was getting validation errors testing locally, but I'll do another pass and see if that function is being run or not

dashboard/lib/services/user.rb
carl-codeorg: @daynew do you remember why we didn't just reject `authentication_options_attributes` and let other nested attributes behave normally?

hannahbergam: I'd like to +1 Bethany's comment above. It feels more future-proof and true to Rails to handle auth as a special case but maintain the original nested attribute processing. That said, if it's a tougher change then I understand it just being a potential follow up.

dashboard/lib/services/user.rb
carl-codeorg: Ok just removed that line and it seems to still work fine. The preprocess_school_info function is getting called. Good catch!

dashboard/lib/services/user.rb
carl-codeorg: Ok I refactored based on this feedback. Still seems to work - lmk what you think. Still want to wait and see if @daynew knows of a reason for rejecting nested attributes other than auth options before merging anything though.

daynew: At the time I was leaning towards being more explicit about how nested attributes were handled. In retrospect, I think the code should look like how it is in this PR.

daynew (approved):

bethanyaconnor (approved):
Thank you so much for your quick work here! Looks great!

PR #58232, opened 2024-04-24
Update Legacy `user_id` Implementation to Support New Session Store

Hamms:
In response to us moving session data out of the cookie and into a persistent store, also update the legacy implementation of `user_id` we have in Rack to read the user id from that store rather than the cookie.

This is definitely a little bit gross; it involves not just pulling in some Dashboard-specific configuration details and manually initializing a Dashboard-specific class, but also actually referencing the underlying Dashboard application itself, all in some code that intentionally lives in the `lib/` directory.

Note that this PR is opened not against staging, but against https://github.com/code-dot-org/code-dot-org/pull/58019

## Testing story

Added a unit test for the new functionality. Also noticed while I was doing so that we didn't have any existing unit tests for any of the existing functionality in this class, so I added a couple.

## Deployment strategy

Merging this PR will not actually deploy any change; it will just update my main feature branch with this new functionality. That branch can then be independently reviewed and deployed at an appropriate time.

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/lib/cdo/rack/request.rb
+++ b/lib/cdo/rack/request.rb
@@ -88,27 +88,18 @@ module Cdo
     end
 
     def user_id
-      @user_id ||= user_id_from_session_cookie
+      @user_id ||= user_id_from_session_store
     end
 
-    def user_id_from_session_cookie
+    def user_id_from_session_store
       session_cookie_key = "_learn_session"
       session_cookie_key += "_#{rack_env}" unless rack_env?(:production)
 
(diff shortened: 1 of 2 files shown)


sureshc:
This isn’t too gross! 😂
I’m still digesting the tests so we can be extra careful here. Will try to complete my feedback by tomorrow morning.

cat5inthecradle (approved):

unlox775-code-dot-org (approved):

PR #58236, opened 2024-04-24
Fix Violations of and Reenable `Style/FloatDivision`

Hamms:
> Checks for division with integers coerced to floats. It is recommended to either always use `fdiv` or coerce one side only. This cop also provides other options for code consistency.

Fix applied automatically with `bundle exec rubocop --autocorrect-all --only Style/FloatDivision`

This seems like a pretty safe one, with few violations.

## Links

- https://docs.rubocop.org/rubocop/cops_style.html#stylefloatdivision
- https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/FloatDivision

## Testing story


bethanyaconnor (approved):
> Relying on existing tests to verify that this does not result in any change to functionality.

Looking at where these changes are, I'm a bit doubtful of test coverage. That said, they're very low risk places to make changes (knock on wood...) and a low risk change at that, so I think this is fine!

Hamms:
Oh, good point. I saw that we had an `admin_reports_controller_test.rb`, but didn't go so far as actually looking at what's in it; that is indeed pretty sparse. Fortunately, I think you're right that this is pretty low risk, particularly because [we cast the value to a float on an earlier line anyway](https://github.com/code-dot-org/code-dot-org/blob/9381cfecfe93551d6bd77bebb0589e39b1ba27cd/dashboard/app/controllers/admin_reports_controller.rb#L89)

Thanks for checking!

PR #58238, opened 2024-04-24
AI Chat: hide customization panel when all settings are hidden

sanchitmalhotra126:
## Warning!!

## Description

Hides the entire model customization panel if the visibility of all settings is set to "hidden". Also included a few other related quality of life fixes that I ended up adding as I was doing this (will add comments in the PR).
- `getTypedKeys` TypeScript util to get typed keys when iterating through an interface
- Add a check for selected model ID when deciding whether to hide the Setup panel
- Perform some sanitization when saving AI Chat settings in levelbuilder, to filter out unrecognized keys, and set missing values to defaults. Helpful when the format changes after data has already been written to a level.

UI when all fields are hidden:
Screenshot 2024-04-24 at 4 34 03 PM (2)

### Sanitizing Level AI Chat Settings

Before saving - manually added some extraneous data and removed some required fields (look for "extraKey-" prefixed fields)
Screenshot 2024-04-24 at 4 48 06 PM

After saving, without actually updating anything in levelbuilder - notice that "extraKey-" fields are gone, and other missing fields are present with default values.
Screenshot 2024-04-24 at 4 49 25 PM

## Links

https://codedotorg.atlassian.net/browse/LABS-701

## Testing story

Tested with allthethings levels.

### DIFF
--- a/apps/src/aichat/redux/aichatRedux.ts
+++ b/apps/src/aichat/redux/aichatRedux.ts
@@ -478,16 +478,10 @@ const hasFilledOutModelCard = (modelCardInfo: ModelCardInfo) => {
   return true;
 };
 
-const allFieldsHidden = (
-  fieldVisibilities: AichatState['fieldVisibilities']
-) => {
-  for (const key of getTypedKeys(fieldVisibilities)) {
-    if (fieldVisibilities[key] !== Visibility.HIDDEN) {
-      return false;
-    }
-  }
-  return true;
-};
+const allFieldsHidden = (fieldVisibilities: AichatState['fieldVisibilities']) =>
+  getTypedKeys(fieldVisibilities).every(
+    key => fieldVisibilities[key] === Visibility.HIDDEN
+  );
 
 // Selectors
 export const selectHasFilledOutModelCard = createSelector(
--- a/apps/src/aichat/redux/aichatRedux.ts
+++ b/apps/src/aichat/redux/aichatRedux.ts
@@ -29,6 +29,7 @@ import {
   ViewMode,
   Visibility,
 } from '../types';
+import {getTypedKeys} from '@cdo/apps/types/utils';
 
 const haveDifferentValues = (
   value1: AiCustomizations[keyof AiCustomizations],
@@ -460,18 +461,16 @@ const aichatSlice = createSlice({
 });
 
 const hasFilledOutModelCard = (modelCardInfo: ModelCardInfo) => {
-  for (const key of Object.keys(modelCardInfo)) {
-    const typedKey = key as keyof ModelCardInfo;
-
-    if (typedKey === 'isPublished') {
+  for (const key of getTypedKeys(modelCardInfo)) {
+    if (key === 'isPublished') {
       continue;
-    } else if (typedKey === 'exampleTopics') {
+    } else if (key === 'exampleTopics') {
       if (
         !modelCardInfo['exampleTopics'].filter(topic => topic.length).length
       ) {
         return false;
       }
-    } else if (!modelCardInfo[typedKey].length) {
+    } else if (!modelCardInfo[key].length) {
       return false;
     }
   }
@@ -479,11 +478,26 @@ const hasFilledOutModelCard = (modelCardInfo: ModelCardInfo) => {
   return true;
 };
 
+const allFieldsHidden = (
+  fieldVisibilities: AichatState['fieldVisibilities']
+) => {
+  for (const key of getTypedKeys(fieldVisibilities)) {
+    if (fieldVisibilities[key] !== Visibility.HIDDEN) {
+      return false;
+    }
+  }
+  return true;
+};
+
 // Selectors
 export const selectHasFilledOutModelCard = createSelector(
-  (state: {aichat: AichatState}) =>
-    state.aichat.currentAiCustomizations.modelCardInfo,
+  (state: RootState) => state.aichat.currentAiCustomizations.modelCardInfo,
   hasFilledOutModelCard
+);
+
+export const selectAllFieldsHidden = createSelector(
+  (state: RootState) => state.aichat.fieldVisibilities,
+  allFieldsHidden
 );
 
 registerReducers({aichat: aichatSlice.reducer});
--- a/apps/src/aichat/views/ModelCustomizationWorkspace.tsx
+++ b/apps/src/aichat/views/ModelCustomizationWorkspace.tsx
@@ -10,8 +10,13 @@ import {useAppSelector} from '@cdo/apps/util/reduxHooks';
 import {AichatLevelProperties} from '@cdo/apps/aichat/types';
 
 const ModelCustomizationWorkspace: React.FunctionComponent = () => {
-  const {temperature, systemPrompt, retrievalContexts, modelCardInfo} =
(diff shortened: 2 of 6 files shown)


apps/src/types/utils.ts
```diff
@@ -14,3 +14,12 @@ export function convertOptionalStringToBoolean(
     return value === 'true';
   }
 }
+
+/**
+ * Return the keys of an object as a typed array.
+ */
+export function getTypedKeys<K extends string | number | symbol>(object: {
```
sanchitmalhotra126, line 21: Added this TypeScript util so we don't have to manually cast keys whenever iterating through the keys of an interface (@molly-moen lmk if you think this should go somewhere else, I chose this file a bit arbitrarily).

apps/src/lab2/levelEditors/aichatSettings/EditAichatSettings.tsx
```diff
@@ -27,6 +30,50 @@ import ModelSelectionFields from './ModelSelectionFields';
 import VisibilityDropdown from './VisibilityDropdown';
 import {UpdateContext} from './UpdateContext';
 import moduleStyles from './edit-aichat-settings.module.scss';
+import {getTypedKeys} from '@cdo/apps/types/utils';
+
+function sanitizeSettings(settings: LevelAichatSettings) {
```
sanchitmalhotra126, line 35: Added a function to sanitize the settings whenever they're saved in levelbuilder. This will make sure we remove any unrecognized keys, and set any missing values to defaults. Useful when the properties structure changes after levels have already been created.

apps/src/aichat/views/ModelCustomizationWorkspace.tsx
```diff
@@ -20,7 +25,9 @@ const ModelCustomizationWorkspace: React.FunctionComponent = () => {
   );
 
   const showSetupCustomization =
-    isVisible(temperature) || isVisible(systemPrompt);
+    isVisible(temperature) ||
+    isVisible(systemPrompt) ||
+    isVisible(selectedModelId);
```
sanchitmalhotra126, line 30: Noticed we were hiding the Setup panel if temperature and system prompt were hidden, but selected model ID was still visible. This makes sure we'll still show the Setup panel if the model selection UI is visible (unless this was intentional?)

cnbrenci: Not intentional. Just an oversight. Thanks!

apps/src/aichat/redux/aichatRedux.ts
```diff
@@ -460,32 +461,45 @@ const aichatSlice = createSlice({
...
 
+const allFieldsHidden = (
+  fieldVisibilities: AichatState['fieldVisibilities']
+) => {
+  for (const key of getTypedKeys(fieldVisibilities)) {
+    if (fieldVisibilities[key] !== Visibility.HIDDEN) {
+      return false;
+    }
+  }
+  return true;
```
thomasoniii: Not a required change, but just noting that this could be condensed down into a single `every` call:

```
return getTypedKeys(fieldVisibilities).every(key => fieldVisibilities[keys] === Visibility.HIDDEN)
```

If every field is hidden, then it returns `true`, and if any of them are _not_ hidden, it'll return false.

Note that `every` is also smart enough to bail out early if it hits a failure - so if the first element is not hidden, it'll stop immediately and not test the full array.

apps/src/lab2/levelEditors/aichatSettings/EditAichatSettings.tsx
```diff
@@ -27,6 +30,50 @@ import ModelSelectionFields from './ModelSelectionFields';
...
+
+  // Set default values for missing keys
+  for (const validKey of validKeys) {
+    if (field[validKey] === undefined) {
+      field[validKey] = defaults[validKey];
+    }
+  }
+
+  return field;
+}
```
thomasoniii: **Actually, ignore everything I wrote in this comment and read the next one instead. I'll leave this one around as a historical artifact.**

Unless I'm misreading it, this function is impure and will change the original object.

The wrapper `sanitizeSettings` function is passing in a sub-object to the `sanitizeFIeld` function, but _that_ function is destructively iterating over the subobject and deleting keys from it.

At the end of `sanitizeSettings`, it will still return a new wrapper object, but the _original_ `settings` object that was handed in will have its `visibilities` and `initialCustomizations` keys changed in place. Right?

I think just adding:
```
const sanitizedField = {...field}
```

at the start of the function, and changing the later refs from `field` -> `sanitizedField` should fix it.

Also, on line 63, we already have the keys for the default, so can swap this:
```
!getTypedKeys<keyof M>(defaults).includes(key)
```
to this:
```
!validKeys.includes(key)
```

Finally, really just as a note, this is using `Array.includes`, which is _probably_ fine here but is also doing a linear scan of the array every time, and actually hiding that the first loop is O(n^2) and not just O(n).

Since the list of keys is (presumably) very short, it probably doesn't matter. But if the list of keys is "big" and/or this function is called frequently, it might be beneficial to just make a `Set` of `validKeys`:

```
const validKeysSet = new Set(validKeys)
// and then later swap this:
validKeys.includes(key)
// for this:
validKeysSet.has(key)
```

We could grind out a little more performance by yanking out that set creation to an earlier spot, maybe by handing it into the `sanitizeField` function as another arg? Or keeping a memoized copy somewhere else? But I think our returns are starting to diminish at that point.

thomasoniii (approved):
couple of minor non-blocking function swapping suggestions, but otherwise lgtm 🚀

apps/src/lab2/levelEditors/aichatSettings/EditAichatSettings.tsx
thomasoniii: Wait. Ignore everything in that previous comment and do this instead.

That double looping really bugged me, and we can actually get rid of it, solve our overwrite issue, _and_ simplify the function simultaneously. And improve our runtime to O(n) from O(n^2 * n)!

```
function sanitizeField<M extends object>(field: M, defaults: M) {
  // iterate over our default keys. This way, we know that we're _only_ going to keep keys that are valid
  // and we don't need to worry about deleting ones that shouldn't be there
  // we're going to reduce that list into a new object to return
  return getTypedKeys<keyof M>(defaults).reduce(( newField, key) => (
    // and in our reduce, it's easy - since we know that we're only dealing with valid keys, then keep
    // whatever value is in our existing object if present, and otherwise stick in the default.
    { ...newField, [key] : field[key] ?? defaults[key] }
  ), {})
}
```

And I even forewent my usual insistence on naming the accumulator `bucket` for some added clarity. 😄 

I think that'll do it, though of course...
```
"Beware of bugs in the above code; I have only proved it correct, not tried it." - Donald Knuth
```

apps/src/types/utils.ts
molly-moen: this seems as good a place as any!

apps/src/lab2/levelEditors/aichatSettings/EditAichatSettings.tsx
sanchitmalhotra126: Slick, I love it! Thanks, will update!

PR #58242, opened 2024-04-25
Music: show pack image in mini-player (fix)

breville:
Follow-up to https://github.com/code-dot-org/code-dot-org/pull/58205 to include a missed file.


cnbrenci (approved):
🚀

PR #58246, opened 2024-04-25
Music: Add strings for effects dropdown labels

cnbrenci:
## Warning!!

## Description

Adding strings for effects labels

## Links

[jira ticket](https://codedotorg.atlassian.net/browse/LABS-682)

## Follow-up work

Putting these strings into use. I believe here is where the dropdown is where these values will be used:
https://github.com/code-dot-org/code-dot-org/blob/22d55bad670b4a861638d61baa2f8fd82b419c8a/apps/src/music/blockly/blocks/simple2.js#L205-L228

### DIFF
--- a/apps/i18n/music/en_us.json
+++ b/apps/i18n/music/en_us.json
@@ -78,5 +78,10 @@
   "soundsFilterEffects": "Effects",
   "soundsFilterVocals": "Vocals",
   "packDialogTitle": "Choose a featured track",
-  "packDialogBody": "Level up your mix by adding sounds from a featured track."
+  "packDialogBody": "Level up your mix by adding sounds from a featured track.",
+  "effectsLabelsFull": "Full",
+  "effectsLabelsHigh": "High",
+  "effectsLabelsMedium": "Medium",
+  "effectsLabelsLow": "Low",
+  "effectsLabelsOff": "Off"
 }


apps/i18n/music/en_us.json
```diff
@@ -78,5 +78,10 @@
   "soundsFilterEffects": "Effects",
   "soundsFilterVocals": "Vocals",
   "packDialogTitle": "Choose a featured track",
-  "packDialogBody": "Level up your mix by adding sounds from a featured track."
+  "packDialogBody": "Level up your mix by adding sounds from a featured track.",
+  "effectsLabelsFull": "Full",
```
cnbrenci, line 82: Given these proposed labels:
```
re: effects, thoughts on the following dropdown labels to improve clarity?
Volume: Full (High?) / Medium / Low, or 100% / 50% / 25%
Delay: Off / High / Low
Filter: Off / Medium / Low
```

I combined the same labels into 1 string (i.e. 'Off' is effectsLabelsOff rather than a string for Delay and a string for Filter with the same value). The values are the same in English, but I'm wondering if the context could make them different in other languages, and if translators use the key name to understand that context? If that's the case, it may be better to split them up into different strings with the same value. Thoughts?

cnbrenci: Chatted with Mario and have an open question here https://codedotorg.slack.com/archives/C0T0UQH0R/p1714067725809699

If I hear back that the context makes a difference I can check in new strings, but seems lower pri than getting these in first.

cnbrenci: The jist is that right now duplicate strings will just get marked as such and take the same translations anyway. But International might have opinions on how they want to separate duplicate strings with different context.

fisher-alice (approved):

PR #58261, opened 2024-04-25
Align Statsig user config to match JS client

nicklathe:
In an effort to align the backend Statsig Metrics wrapper with the frontend wrapper, the following changes are necessary:

* Move the `user_type` to the Statsig user object (previously this was a value in the metadata hash)
* Add a `get_enabled_experiments` as an optional parameter, which will be a list of experiments a user might be enrolled in. The reason for this being optional is to save on db queries where experiments may not be relevant

Both of these will be values in the custom_ids hash in the StatsigUser class, per the [Statsig docs](https://docs.statsig.com/server/rubySDK#type-statsiguser).

Example usage:

```
metadata = {
    'some_data' => 'foo',
}
Metrics::Events.log_event(
    user: user,
    event_name: 'lti_user_signin',
    metadata: metadata,
    get_enabled_experiments: true,
)
```

**Old formatted event:**
Screenshot 2024-04-25 at 5 29 22 PM

**New formatted event:**
Screenshot 2024-04-26 at 12 36 12 PM

## Warning!!

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/dashboard/lib/metrics/events.rb
+++ b/dashboard/lib/metrics/events.rb
@@ -49,7 +49,7 @@ module Metrics
           event_name: event_name,
           event_value: event_value,
           metadata: metadata,
-      }.compact
+        }.compact
         puts "Logging Event: #{event_details.to_json}"
       end
     end
--- a/dashboard/lib/metrics/events.rb
+++ b/dashboard/lib/metrics/events.rb
@@ -45,11 +45,14 @@ module Metrics
       private def log_event_to_stdout(user:, event_name:, event_value:, metadata:, enabled_experiments:)
         event_details = {
           user_id: user.id,
-          enabled_experiments: enabled_experiments,
+          custom_ids: {
+            user_type: user.user_type,
+            enabled_experiments: enabled_experiments,
+          }.compact,
           event_name: event_name,
           event_value: event_value,
           metadata: metadata,
-        }.compact
+        }
         puts "Logging Event: #{event_details.to_json}"
       end
     end
--- a/dashboard/lib/metrics/events.rb
+++ b/dashboard/lib/metrics/events.rb
@@ -4,14 +4,14 @@ module Metrics
   module Events
     class << self
       # Logs an event, delegating to the appropriate handler based on environment
-      def log_event(user:, event_name:, event_value: nil, metadata: {})
+      def log_event(user:, event_name:, event_value: nil, metadata: {}, enabled_experiments: nil)
         event_value = event_name if event_value.nil?
         managed_test_environment = CDO.running_web_application? && CDO.test_system?
 
         if CDO.rack_env?(:development)
-          log_event_to_stdout(user: user, event_name: event_name, event_value: event_value, metadata: metadata)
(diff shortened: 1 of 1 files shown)


hannahbergam (approved):
Yay- this looks great! @dmcavoy I'm not exactly sure how their metrics will work across client and server side integrations, but I imagine this is the closest we can get. Thanks so much for jumping on this Nick!

hannahbergam (commented):
Actually one more thought- don't we have a way to query a user's current experiments from the backend?

nicklathe:
> Actually one more thought- don't we have a way to query a user's current experiments from the backend?

Good point, let me look into that.

nicklathe:
It doesn't look like Statsig supports listing experiments, at least for server-side SDKs. It does support [Feature Gates](https://docs.statsig.com/feature-flags/implement/server#step-4-check-the-feature-gate).

hannahbergam:
> It doesn't look like Statsig supports listing experiments, at least for server-side SDKs. It does support [Feature Gates](https://docs.statsig.com/feature-flags/implement/server#step-4-check-the-feature-gate).

I am referring to Code.org experiments here. Perhaps we can discuss offline?

dashboard/lib/metrics/events.rb
```diff
@@ -4,8 +4,9 @@ module Metrics
   module Events
     class << self
       # Logs an event, delegating to the appropriate handler based on environment
-      def log_event(user:, event_name:, event_value: nil, metadata: {}, enabled_experiments: nil)
+      def log_event(user:, event_name:, event_value: nil, metadata: {}, get_enabled_experiments: false)
```
hannahbergam, line 7: Thank you for adding this! And makes sense that it can be optional to save the query. I wonder if we could cache this somehow (but that can be a future consideration)!

mgc1194 (approved):
Nice! Thanks!
Could you link in this PR where the front-end wrapper is defined?
I'm curious to compare both.

dashboard/lib/metrics/events.rb
nicklathe: Sounds good, yeah caching could be an interesting thing to explore 👍. We can evaluate later to see how many event metrics are part of an experiment vs. how many aren't, and dial this in to favor the majority as it makes sense.

nicklathe:
@mgc1194 [Statsig wrapper](https://github.com/code-dot-org/code-dot-org/blob/staging/apps/src/lib/util/StatsigReporter.js), implemented in the [AnalyticsReporter](https://github.com/code-dot-org/code-dot-org/blob/staging/apps/src/lib/util/AnalyticsReporter.js)

dashboard/lib/metrics/events.rb
hannahbergam: I think the idea with this configuration isn't necessarily about *only* seeing users inside experiments: it's about viewing *whether or not the user is in experiments. Perhaps @dmcavoy could provide more context here. My instinct is that we *always want experiment data instead of having to opt in.

hannahbergam: I'm so sorry for all the churn on this!! I just spoke with @bethanyaconnor offline about this and I think our implementation of experiments is messy enough between the backend and front end to keep this as an opt-in to save the additional query. Resolving!

PR #58269, opened 2024-04-26
Lab2: teacher panel

breville:
With this change, the teacher panel can be used to switch between users and levels without a page reload when used with Lab2 levels.  

It revisits and builds upon work done last year in https://github.com/code-dot-org/code-dot-org/pull/58165.

The goal is that the existing teacher panel continues working as-is for non-Lab2 levels.

Lab2 levels already get level changes without a page reload, and the teacher panel stays intact while that happens.  The bigger change was to make sure the teacher panel can change user ID of the user whose work is being viewed.  

On the backend, the `ProjectsController` action to retrieve project data takes an optional parameter for the user ID, so that a teacher can retrieve the work of a student in one of their sections.  Permission is checked at this time.

The teacher panel often modifies URL parameters before triggering a reload.  Even though we trigger fewer reloads now, the code has been left intact; upon manual reload, the panel will be restored to the appropriate state.  

Section selection still triggers a reload, so that the new section information can be loaded at initialization time.  It could be nice to avoid this reload too, but this change already does a lot.

WIP: This is being shared now to get a first pass of feedback, but changes are to be expected, and tests should be updated.

### Video

This video demonstrates changing "view as", "user to view", and "current level", all without a page reload.

https://github.com/code-dot-org/code-dot-org/assets/2205926/acc8632f-d066-4193-ad73-0f14a12dd16f


apps/src/lab2/projects/projectsApi.ts
```diff
@@ -7,13 +7,17 @@ const rootUrl = '/projects/';
...
+  userId?: string
 ): Promise<Response> {
   let requestString = rootUrl;
   if (scriptId !== undefined) {
     requestString += `script/${scriptId}/`;
   }
   requestString += `level/${levelId}`;
+  if (userId !== '') {
+    requestString += `/user/${userId}`;
+  }
```
thomasoniii, line 20: This can be called where `userId` is undefined, right? (the param is optional)

Should it use the same `userId !== undefined` check as the `scriptId` one?

thomasoniii (approved):
one note about changing a guard condition, but otherwise lgtm 🚀

apps/src/code-studio/components/progress/ViewAsToggle.jsx
```diff
@@ -94,8 +96,11 @@ export default connect(
     viewAs: state.viewAs,
   }),
   dispatch => ({
-    changeViewType(viewAs) {
-      dispatch(changeViewType(viewAs));
+    changeViewType(viewAs, isAsync) {
+      if (viewAs === ViewType.Participant) {
+        dispatch(setUserId(null));
```
sanchitmalhotra126: Should clearing the user ID happen in the changeViewType thunk action in viewAsRedux rather than here, because we also clear the user_id from query params in that function? That being said, not sure if there's an antipattern of updating state for a different slice inside a thunk action

apps/src/code-studio/progressRedux.ts
```diff
@@ -45,6 +45,7 @@ export interface ProgressState {
   lessons: Lesson[] | null;
   lessonGroups: LessonGroup[] | null;
   scriptId: number | null;
+  userId: string | null;
```
sanchitmalhotra126: Could this be something like `viewAsUserId` or similar? Don't want this to be confused with the current user's user ID (which I believe is represented by `state.currentUser.userId`).

Also I was wondering if this should be moved to `viewAsRedux`, but I see that viewAsRedux's state is simply the ViewType value, so would probably be more involved to refactor that into an object. viewAsRedux and progressRedux seem pretty closely related anyway.

apps/src/code-studio/components/progress/teacherPanel/TeacherPanel.jsx
```diff
@@ -73,18 +83,26 @@ class TeacherPanel extends React.Component {
...
   componentDidMount() {
     const initialViewAs = queryParams('viewAs') || ViewType.Instructor;
-
     if (this.props.viewAs !== initialViewAs) {
       this.props.setViewType(initialViewAs);
     }
 
+    const initialUserId = this.getSelectedUserId();
+    if (initialUserId) {
+      console.log('Setting user ID from URL:', initialUserId);
```
sanchitmalhotra126: Log statements still required?

apps/src/code-studio/components/progress/teacherPanel/TeacherPanel.jsx
```diff
@@ -128,9 +146,21 @@ class TeacherPanel extends React.Component {
     });
   };
 
+  isCurrentLevelLab2 = () => {
+    const currentLevel = this.props.levels?.find(level => level.isCurrentLevel);
+    return currentLevel?.usesLab2;
+  };
```
sanchitmalhotra126: Minor: could do this calculation in the `connect` function below and just pass `isCurrentLevelLab2` in here as a prop?

apps/src/lab2/lab2Redux.ts
```diff
@@ -84,6 +84,7 @@ export const setUpWithLevel = createAsyncThunk(
       scriptId?: number;
       levelPropertiesPath: string;
       channelId?: string;
+      userId: string;
```
sanchitmalhotra126: Is this still needed?

apps/src/lab2/projects/ProjectContainer.tsx
```diff
@@ -28,6 +28,9 @@ const ProjectContainer: React.FunctionComponent<ProjectContainerProps> = ({
   const currentLevelId = useSelector(
     (state: {progress: ProgressState}) => state.progress.currentLevelId
   );
+  const userId = useSelector(
+    (state: {progress: ProgressState}) => state.progress.userId || ''
+  );
```
sanchitmalhotra126: nit: can use `useAppSelector` here to skip the type cast (but also fine if we want to keep it consistent with the rest and update them all later)

apps/src/lab2/projects/ProjectManager.ts
```diff
@@ -209,7 +212,9 @@ export default class ProjectManager {
       '/projects/' +
       this.lastChannel.projectType +
       '/' +
-      this.channelId
+      this.channelId +
+      '/' +
+      this.userId
```
sanchitmalhotra126: Why do we need the userId for the shareUrl here?

apps/src/lab2/projects/ProjectManagerFactory.ts
```diff
@@ -39,13 +41,18 @@ export default class ProjectManagerFactory {
    */
   static async getProjectManagerForLevel(
     projectManagerStorageType: ProjectManagerStorageType,
+    userId: string,
```
sanchitmalhotra126: In the same vein as @thomasoniii's comment, should `userId` be optional here like `scriptId`?

dashboard/app/controllers/projects_controller.rb
```diff
@@ -310,19 +310,37 @@ def create_new
...
+    end
+
     error_message = under_13_without_tos_teacher?(level)
     return render(status: :forbidden, json: {error: error_message}) if error_message
+
     # get_storage_id works for signed out users as well, it uses the cookie to determine
     # the storage id.
-    user_storage_id = get_storage_id
+    user_storage_id = if user_id
+                        # todo: handle this not finding a valid storage id
```
sanchitmalhotra126: Should we fall back to the current user's channel ID in this case? Or throw an error? (Curious what we do for non-lab2 labs in this situation)

sanchitmalhotra126 (commented):
Nice! A handful of minor comments but overall the logic looks good and makes sense to me.

apps/src/code-studio/components/progress/teacherPanel/TeacherPanel.jsx
```diff
@@ -128,9 +146,21 @@ class TeacherPanel extends React.Component {
...
+
   onSelectUser = (id, selectType) => {
     this.logToFirehose('select_student', {select_type: selectType});
-    const isAsync = this.props.pageType === pageTypes.scriptOverview;
+
+    const isLab2Level = this.isCurrentLevelLab2();
+    if (isLab2Level) {
+      console.log('Teacher panel: select user is async because level is Lab2.');
```
molly-moen: remove this log?

apps/src/code-studio/progressRedux.ts
molly-moen: +1 to this question, also in other places where you've added `userId` is it a similar field, as in it's the user id of the user we are viewing, and not the current user? I think if it's not the actual user's id we should have a more specific name than `userId`.

dashboard/app/controllers/projects_controller.rb
```diff
@@ -310,19 +310,37 @@ def create_new
     )
   end
 
-  # GET /projects(/script/:script_id)/level/:level_id
+  # GET /projects(/script/:script_id)/level/:level_id(/user/:user_id)
   # Given a level_id and the current user (or signed out user), get the existing project
   # or create a new project for that level and user. If a script_id is provided, get or
   # create the project for that level, script and user
```
molly-moen: mind updating the comment to explain how `user_id` works here, and that it is likely different from the current user?

apps/src/code-studio/components/progress/ViewAsToggle.jsx
breville: I remember wondering about this too, and having a similar architectural question.  

The first problem encountered was that importing `progressRedux` in `viewAsRedux` triggered a new circular dependency: 
```
progressRedux.ts -> progressHelpers.js -> viewAsRedux.js -> progressRedux.ts
```

apps/src/code-studio/progressRedux.ts
breville: Good call; I've made the switch to `viewAsUserId` in a variety of places on the client. 

I'm not quite sure why `viewAsRedux` is so small, but so far thought it easier to avoid additional refactoring to keep this change as small as possible.

apps/src/code-studio/components/progress/teacherPanel/TeacherPanel.jsx
breville: No; removed.

apps/src/code-studio/components/progress/teacherPanel/TeacherPanel.jsx
breville: Good call; done.

apps/src/lab2/projects/ProjectContainer.tsx
breville: Good call, though in the spirit of reducing changes, might be good for a potential follow-up.

apps/src/code-studio/components/progress/teacherPanel/TeacherPanel.jsx
breville: Done.

dashboard/app/controllers/projects_controller.rb
breville: Yes; done.

apps/src/lab2/projects/projectsApi.ts
breville: Good point.  It wasn't really utilizing the optional part before, but it does now.

apps/src/lab2/projects/ProjectManager.ts
breville: Apparently it's not.  Removed!

apps/src/lab2/projects/ProjectManagerFactory.ts
breville: Good call; now done.

apps/src/lab2/lab2Redux.ts
breville: I believe so; we pass this into `getProjectManagerForLevel` just below.

dashboard/app/controllers/projects_controller.rb
```diff
@@ -310,19 +310,37 @@ def create_new
...
   def get_or_create_for_level
     script_id = params[:script_id]
+    user_id = params[:user_id]
     level = Level.find(params[:level_id])
+
+    # If viewing another user's work, ensure that we have permission.
+    if user_id
+      script_level = level.script_levels.find_by_script_id(script_id)
+      user = User.find(user_id)
+      unless can?(:view_as_user, script_level, user)
```
breville, line 333: This is based on: https://github.com/code-dot-org/code-dot-org/blob/f280b95523fb6a3aabbe624250146256a9480cef/dashboard/app/controllers/api_controller.rb#L479

apps/src/lab2/projects/ProjectManager.ts
```diff
@@ -50,12 +50,14 @@ export default class ProjectManager {
   private reduceChannelUpdates: boolean;
   private initialSaveComplete: boolean;
   private forceReloading: boolean;
+  private userId: string | undefined;
```
sanchitmalhotra126: We can remove `userId` from this class entirely now right?

dashboard/app/controllers/projects_controller.rb
breville: Pulling on this thread led to some new work that has arrived in commit https://github.com/code-dot-org/code-dot-org/pull/58269/commits/76b5dbc623485032bc51cf44b6e85b960f0b79e8.  With that work, we:
- no longer create a channel if one doesn't exist when a teacher is looking at a student's work;
- update the blockly workspace to be read-only when a teacher is viewing a student's work;
- handle the situation in which a student has not done any work yet, in which case the starter code is shown.

apps/src/lab2/lab2Redux.ts
```diff
@@ -135,14 +136,28 @@ export const setUpWithLevel = createAsyncThunk(
...
             );
+
       // Only set the project manager and initiate load
       // if this request hasn't been cancelled.
       if (thunkAPI.signal.aborted) {
         return;
       }
 
+      // We might be a teacher attempting to view a student level that hasn't been
+      // started, and there is no project manager available.
```
sanchitmalhotra126, line 150: Just pointing out that this will be useful for other cases where we don't want to load project as well, such as editing start sources (which @mikeharv will be working on)

dashboard/app/controllers/projects_controller.rb
```diff
@@ -310,26 +310,55 @@ def create_new
...
-    # Find the channel for the user and level if it exists, or create a new one.
-    channel_token = ChannelToken.find_or_create_channel_token(level, request.ip, user_storage_id, script_id, {hidden: true})
-    script_name = !script_id.nil? && Unit.find(script_id)&.name
+
+    if user_id
+      # If viewing another user's work, ensure that we have permission.
+      script_level = level.script_levels.find_by_script_id(script_id)
+      user = User.find(user_id)
+      unless can?(:view_as_user, script_level, user)
+        return render(status: :forbidden, json: {error: "Access denied."})
```
sanchitmalhotra126, line 334: Just flagging this for later, but do we need to handle this in any special way on the client? I'm curious what we do for other labs

sanchitmalhotra126 (approved):
LGTM! Thanks for iterating on this, cool to see this finally coming through!

apps/src/lab2/projects/ProjectManagerFactory.ts
```diff
@@ -34,23 +34,31 @@ export default class ProjectManagerFactory {
...
+      scriptId,
+      userId
+    );
     if (response.ok) {
       const responseBody = await response.json();
       if (responseBody && responseBody.channel) {
         channelId = responseBody.channel;
         reduceChannelUpdates = responseBody.reduceChannelUpdates;
+      } else if (responseBody && responseBody.started === false) {
+        return null;
```
molly-moen, line 62: Mind adding a comment that this condition will only be hit if we are viewing another user's project?

molly-moen (approved):
Looks good!

apps/src/lab2/projects/ProjectManager.ts
breville: Good catch; done.

apps/src/lab2/projects/ProjectManagerFactory.ts
breville: Good call; done.

PR #58275, opened 2024-04-26
Music: sounds panel revision

breville:
This change introduces an alternate sound panel which emphasizes exploration of the sound library.

The idea behind this change came from thinking about a few inputs: students weren't always finding sounds they wanted; there was a desire from the beginning of the Music Lab project to center the sound exploration experience; an observation that students were drawn to iconic representations of sounds; and the insight we gained when making the pack dialog a [two-step sequence](https://github.com/code-dot-org/code-dot-org/pull/58092).  Most importantly, the sound exploration just wasn't always fun, but it should be!

The existing sounds panel had some room for change: it was functional but businesslike; the preview button was tiny; and clicking on one sound would dismiss the panel.

The vision for this change was to make some substantial changes: sound selection sticks around while you are exploring sounds; everything is big and iconic and uses our bright timeline colors; and clicking anywhere on a sound tile plays a preview.

That has led to this new implementation, which we will test out as an alternative.  For now, the new panel can be viewed by adding the `&sounds-panel-2=true URL` parameter.

We also wanted to make it more clear that the folder and sounds lists are vertically scrollable, and so when opening the panel with the default sound selected, the two lists do a small vertical scroll into place using a new `ScrollIntoView` component.  If the user interrupts the scroll manually, then it ends immediately.

### New sounds panel

https://github.com/code-dot-org/code-dot-org/assets/2205926/c340ddc1-ea3f-4557-9fe4-761e0cc951f5

### Opens at selected sound

https://github.com/code-dot-org/code-dot-org/assets/2205926/be2f108e-0cd2-4c9d-b76e-522dad3b18a1

### Interrupting the scroll

https://github.com/code-dot-org/code-dot-org/assets/2205926/e3f3c6a6-c337-46e2-95de-ff9337fd5f22


markabarrett:
This is SO much better! Way to take the initiative on this and make it more fun and easier to use.

apps/src/music/views/MusicView.jsx
```diff
@@ -304,8 +304,9 @@ class UnconnectedMusicView extends React.Component {
     this.executeCompiledSong();
 
     Globals.setShowSoundFilters(
-      AppConfig.getValue('show-sound-filters') === 'true' ||
-        levelData?.showSoundFilters
+      AppConfig.getValue('show-sound-filters') !== 'false' &&
+        (AppConfig.getValue('show-sound-filters') === 'true' ||
+          levelData?.showSoundFilters)
```
thomasoniii, line 309: Why does it have the extra check for `!== 'false'`?

This looks like a boolean value, can it have something other than `true` or `false` in it?

I'm guessing this is protecting against a nullish case, where it's undefined. If so, doesn't this work:

```
AppConfig.getValue('show-sound-filters') !== 'false' || levelData?.showSoundFilters
```

It's checking to see if the value is _not_ false (so empty values default to `true`), or if the `showSoundFilters` is enabled.

apps/src/music/views/SoundsPanel2.tsx
```diff
@@ -0,0 +1,487 @@
...
+      className={classNames(
+        'sounds-panel-folder-row',
+        classNames(styles.folderRow, isSelected && styles.folderRowSelected)
+      )}
+      onClick={() => onSelect(folder)}
+      onKeyDown={event => {
+        if (event.key === 'Enter') {
+          onSelect(folder);
+        }
+      }}
```
thomasoniii: probably should be memoized.

React 19 can't get here fast enough. 😝

apps/src/music/views/SoundsPanel2.tsx
```diff
@@ -0,0 +1,487 @@
...
+    return Math.sin((x * Math.PI) / 2);
+  }
+
+  // Initial render.
+  useEffect(() => {
+    if (!doScroll) {
+      return;
+    }
+
+    setInterval(() => {
```
thomasoniii: This is fine and _probably_ doesn't need to be changed, but just noting that animating things in an interval is a little problematic - intervals aren't guaranteed to fire at the requested time, they're only guaranteed not to fire _before_ the requested time. So if the main thread is locked up doing other things, this could result in a very slow animation.

If this were rewritten to use [requestAnimationFrame](https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame) we'd be able to guarantee the exact duration of the animation, where it'd drop frames if necessary to fit the timeframe.

Personally, I think [react-interpolation-animation](https://www.npmjs.com/package/react-interpolation-animation) works very nicely for this, but of course I'm biased. 😅

apps/src/music/views/SoundsPanel2.tsx
```diff
@@ -0,0 +1,487 @@
...
+  if (filter === 'all') {
+    rightColumnSoundEntries = possibleSoundEntries.filter(
+      soundEntry => soundEntry.sound.type !== 'preview'
+    );
+  } else {
+    rightColumnSoundEntries = possibleSoundEntries.filter(
+      soundEntry =>
+        soundEntry.sound.type === filter && soundEntry.sound.type !== 'preview'
+    );
+  }
```
thomasoniii: Non-blocking, but I'd be inclined to crush this down into a single case and put the conditional inline:

```
rightColumnSoundEntries = possibleSoundEntries.filter(
  soundEntry =>
    soundEntry.sound.type !== 'preview' && (soundEntry.sound.type === filter || filter === 'all')
);
```

apps/src/music/views/SoundsPanel2.tsx
```diff
@@ -0,0 +1,487 @@
...
+    currentSoundRef.current = ref;
+  };
+
+  let possibleSoundEntries: SoundEntry[] = [];
+  let rightColumnSoundEntries: SoundEntry[] = [];
+
+  if (mode === 'packs') {
+    folders.sort((a, b) =>
+      a.restricted === b.restricted ? 0 : a.restricted ? -1 : 1
+    );
```
thomasoniii, line 294: just noting that this is sorting the array in place and confirming that we're okay doing that.

`Array.prototype.toSorted` is still too new for us to use. 😞

thomasoniii (approved):
couple of minor logic Qs/suggestions, but nothing blocking. lgtm 🚀

apps/src/music/views/SoundsPanel2.tsx
```diff
@@ -0,0 +1,487 @@
...
+  SoundFolder,
+  SoundType,
+} from '../player/MusicLibrary';
+import SoundStyle from '../utils/SoundStyle';
+import FocusLock from 'react-focus-lock';
+import SegmentedButtons from '@cdo/apps/componentLibrary/segmentedButtons';
+import musicI18n from '../locale';
+
+/*
+ * Renders a UI for previewing and choosing samples. This is currently used within a
```
sanchitmalhotra126: Mind adding a comment around what the "2" refers to (/how this differs from v1)?

apps/src/music/views/SoundsPanel2.tsx
```diff
@@ -0,0 +1,487 @@
...
+  libraryGroupPath: string;
+  playingPreview: string;
+  folder: SoundFolder;
+  currentValue: SoundFolder;
+  onSelect: (path: SoundFolder) => void;
+  onPreview: (path: string) => void;
+  currentFolderRefCallback: (ref: HTMLDivElement) => void;
+}
+
+const FolderPanelRow: React.FunctionComponent<FolderPanelRowProps> = ({
```
sanchitmalhotra126, line 51: Is `FolderPanelRow` identical between SoundsPanel and SoundsPanel2? If so, should we reuse?

apps/src/music/views/SoundsPanel2.tsx
```diff
@@ -0,0 +1,487 @@
...
+      onClick={onSoundClick}
+      onKeyDown={event => {
+        if (event.key === 'Enter') {
+          onSoundSelect();
+        }
+      }}
+      ref={isSelected ? currentSoundRefCallback : null}
+      aria-label={sound.name}
+      tabIndex={0}
+      role="button"
```
sanchitmalhotra126, line 189: Can this be a `<button>` component for better tab navigation?

apps/src/music/views/SoundsPanel2.tsx
```diff
@@ -0,0 +1,487 @@
...
+          )}
+        >
+          {sound.name}
+        </div>
+      </div>
+    </div>
+  );
+};
+
+// ScrollIntoView: This component does an eased scroll of the container's content,
```
sanchitmalhotra126: Very cool! This feels like something that should be in its own file for potential reuse (even if it stays within /music for now)

apps/src/music/views/SoundsPanel2.tsx
```diff
@@ -0,0 +1,487 @@
...
+
+          scrollStep.current++;
+
+          if (scrollStep.current > delay + frames) {
+            // The automatic scroll has reached its destination.
+            scrollStep.current = undefined;
+          }
+        }
+      }
+    }, 1000 / 60);
```
sanchitmalhotra126: Should this interval rate also be configurable via props?

apps/src/music/views/SoundsPanel2.tsx
```diff
@@ -0,0 +1,487 @@
...
+        )}
+        <div id="sounds-panel-body" className={styles.soundsPanelBody}>
+          {mode === 'packs' && (
+            <ScrollIntoView
+              id="sounds-panel-left"
+              className={styles.leftColumn}
+              doScroll={isDefaultSoundSelected}
+              delay={0}
+              frames={50}
+              distanceY={70}
```
sanchitmalhotra126: nit: constants for these (since we presumably want the same animation for both <ScrollIntoView> containers)?

apps/src/music/views/soundsPanel2.module.scss
```diff
@@ -0,0 +1,193 @@
...
+      width: calc(40% - 00px);
+      overflow-y: scroll;
+    }
+
+    .rightColumn {
+      background-color: $neutral_dark;
+      width: calc(67% + 00px);
+      overflow-y: scroll;
+      display: grid;
+      grid-template-columns: calc(33.3333% - 0px) calc(33.3333% - 0px) calc(33.3333% - 0px);
```
sanchitmalhotra126: nit: could widths just be `calc(100% / 3)`?

apps/src/music/views/soundsPanel2.module.scss
```diff
@@ -0,0 +1,193 @@
...
+
+    &Left {
+      width: 55px;
+      height: 55px;
+      margin: 0 6px 0 4px;
+      float: left;
+    }
+
+    &Middle {
+      width: calc(100% - 75px - 58px);
```
sanchitmalhotra126: nit: some constants and/or comments to indicate what these px values refer to?

apps/src/music/views/soundsPanel2.module.scss
```diff
@@ -0,0 +1,193 @@
...
+    display: flex;
+    justify-content: space-between;
+    align-items: center;
+    cursor: pointer;
+    box-sizing: border-box;
+    flex-direction: column;
+    border-radius: 5px;
+
+    &Selected {
+      background-color: $neutral_dark70 !important;
```
sanchitmalhotra126: Why is the `!important` needed?

dashboard/app/assets/stylesheets/application.scss
```diff
@@ -898,6 +898,10 @@ $small-footer-standard-width: 400px;
       color: $neutral_light;
     }
   }
+
+  .blocklyDropDownContent {
+    max-height: initial !important;
+  }
```
sanchitmalhotra126: Why is this needed? Also, would this affect any other blockly dropdowns?

sanchitmalhotra126 (commented):
Awesome! This looks and feels great - thanks for taking the time to get this UI revision out! A few questions/comments

apps/src/music/views/MusicView.jsx
breville: If the URL parameter is not provided, it can be `undefined`.  So this is a bit awkward, but I think captures both the ability to force the UI to be hidden, and the ability to force it to be shown.

With the suggested code, providing `?show-sound-filters=false` when viewing a level that has `levelData?.showSoundFilters` set to `true` doesn't actually hide the filters.

apps/src/music/views/SoundsPanel2.tsx
breville: Presumably this means `useCallback`?  Done.

apps/src/music/views/SoundsPanel2.tsx
breville: Interesting. I've used `requestAnimationFrame` a bunch for big canvas animations but good point about using it for this kind of animation as well.  Let's keep an eye on animation quality and maybe do this as a follow-up.

apps/src/music/views/SoundsPanel2.tsx
breville: Nice; done.

apps/src/music/views/SoundsPanel2.tsx
breville: Yes, this is intentional.

apps/src/music/views/SoundsPanel2.tsx
breville: Good call; done.

apps/src/music/views/SoundsPanel2.tsx
breville: Yes, this is intentional.

apps/src/music/views/SoundsPanel2.tsx
breville: The keyboard navigation has been working well for a while in the previous version of this panel, so we might consider this for a follow-up when we next focus on overall navigation.

apps/src/music/views/SoundsPanel2.tsx
breville: Good call; done.

apps/src/music/views/SoundsPanel2.tsx
breville: I think 60fps is reasonable for the component to own, but will make a constant.

apps/src/music/views/SoundsPanel2.tsx
breville: Good call; done.  

The animation is almost the same, though with a slight delay for the sounds scroll, to illustrate that the two sections can scroll independently.

apps/src/music/views/soundsPanel2.module.scss
breville: Much better; done!

apps/src/music/views/soundsPanel2.module.scss
breville: Good point.  Tidied the styling up so this is no longer here.

apps/src/music/views/soundsPanel2.module.scss
breville: It was to avoid the hover overriding the selected color, but I've now found another way that seems to work.

apps/src/music/views/SoundsPanel2.tsx
breville: It has diverged a tiny bit now, and I have been anticipating a bit more to come.  I think it might be safest to keep it separate to minimize the risk of regressing the earlier version.

dashboard/app/assets/stylesheets/application.scss
breville: Oh, I meant to mention this.  Blockly seems to [enforce](https://github.com/google/blockly/blob/76e179d65513998ac0e405f3ae0b9772a401c083/core/css.ts#L127) a maximum height of its dropdowns of `300px`, and for this panel, especially with the filters, a little more height seemed good.  

It gets us from this:
Screenshot 2024-05-08 at 10 39 34 AM

### 

To this:
Screenshot 2024-05-08 at 10 39 51 AM

It looks like the `!important` isn't required, so I've taken that out.  I scoped the CSS to only affect Music Lab levels, and I don't think we have any other dropdowns at the moment that are affected by this change in `max-height`.

FYI @mikeharv

sanchitmalhotra126 (approved):
LGTM!

PR #58068, opened 2024-04-17
[LABS-654] Music: Play page pt.1

levadadenys:
### [LABS-654] Music: Play page pt.1

* created a screen for Music Play page

**Please note (!):**  this is not a full implementation of the page (not all of the needed functionality is connected and adjusted yet). For more context, please take a looks into [jira ticket](https://codedotorg.atlassian.net/browse/LABS-654), I've left a comment describing what we have  todo/done so far.

Chrome (high res):
Знімок екрана 2024-04-22 о 13 02 09

Chrome low res (nevermind missign icons, that's just an issue with loading the icons when accessing locally from other machine):

Safari:
Знімок екрана 2024-04-22 о 13 03 15

Chrome (IOS):

Safari (IOS):

## Links
- jira ticket: [LABS-654](https://codedotorg.atlassian.net/browse/LABS-654)

## PR Checklist:

- [ ] Tests provide adequate coverage
- [ ] Privacy and Security impacts have been assessed
- [ ] Code is well-commented
- [ ] New features are translatable or updates will not break translations
- [ ] Relevant documentation has been added or updated
- [ ] User impact is well-understood and desirable
- [ ] Pull Request is labeled appropriately
- [ ] Follow-up work items (including potential tech debt) are tracked and linked

PR #58293, opened 2024-04-26
Music Play View - part 2

fisher-alice:
This PR updates the Music Lab play view and follows up https://github.com/code-dot-org/code-dot-org/pull/58068.

Note that the adhoc based on this branch is available at https://adhoc-alice-play-view-updates-studio.cdn-code.org/home.
[Example project in play view](https://adhoc-alice-play-view-updates-studio.cdn-code.org/projects/music/lpef7b3VcqvntTmBLG2c1w).

Updates in this PR include:
- Adds `MusicPlayView` to `MusicLabView` so that when `isShareView` is set to `true`, the play view is rendered.
- In `MusicView`, when `isPlayView` is set to `true`, a headless `musicBlocklyWorkspace` is initialized, i.e., the UI of the blockly workspace is not rendered. Thanks to @sanchitmalhotra126 for pairing with me on this! I based this part of the PR on Sanchit's [WIP branch](https://github.com/code-dot-org/code-dot-org/compare/staging...sanchit/wip-music-play-view).
- Adds functionality to the play/stop slider and shows progress as song plays.
- Adds functionality to the 'View code' button (view/edit mode) and Make my Own' button (remix).
- Adds `redirectToView` to `projectsApi`.
- Adds the music play image from product.
- Updated styling of play view based on product/design preferences.

The 'Share' button in play view is displayed on most mobile devices. But you can view this button if you add the query param '?canShare=true' to the end of the project play view url, e.g., https://adhoc-alice-play-view-updates-studio.cdn-code.org/projects/music/lpef7b3VcqvntTmBLG2c1w?canShare=true 

## After Update

Play view card on desk top:
Screenshot 2024-04-30 at 6 00 23 PM

Play view full screen on desk top:
Screenshot 2024-04-30 at 6 00 18 PM

Play view on mobile:
Screenshot 2024-04-30 at 6 02 24 PM

Screencast video of play view on desktop:

https://github.com/code-dot-org/code-dot-org/assets/107423305/9b93f33a-6027-4bb2-859d-a6667dc00927

## Follow-up work
Follow-up will be to refactor the slider and include adding color to progress.

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/lab2/views/Lab2Wrapper.module.scss
+++ b/apps/src/lab2/views/Lab2Wrapper.module.scss
@@ -1,16 +1,6 @@
 @import 'color.scss';
 @import './common.scss';
 
-.shareViewContainer {
-  position: fixed;
-  top: 0;
-  bottom: 0;
-  left: 0;
-  height: 100%;
-  width: 100%;
-  z-index: 1,
-}
-
 .labContainer {
   position: fixed;
   top: 52px;
--- a/apps/src/lab2/views/Lab2Wrapper.module.scss
+++ b/apps/src/lab2/views/Lab2Wrapper.module.scss
@@ -23,6 +23,11 @@
     pointer-events: none;
   }
 
+  &ShareView {
+    top: 0;
+    bottom: 0;
+  }
+
   .pageErrorContainer {
     @extend %full-container-overlay;
     animation: fadein 0.5s;
--- a/apps/src/lab2/views/Lab2Wrapper.module.scss
+++ b/apps/src/lab2/views/Lab2Wrapper.module.scss
@@ -1,6 +1,17 @@
 @import 'color.scss';
 @import './common.scss';
 
+.overlay {
+  position: fixed;
+  top: 0;
+  bottom: 0;
+  left: 0;
+  height: 100%;
+  width: 100%;
+  opacity: 0.8;
+  background-color: $light_black;
+  z-index: 1,
+}
 
 .labContainer {
   position: fixed;
--- a/apps/src/lab2/views/Lab2Wrapper.module.scss
+++ b/apps/src/lab2/views/Lab2Wrapper.module.scss
@@ -1,7 +1,7 @@
 @import 'color.scss';
 @import './common.scss';
 
-.overlay {
+.shareViewContainer {
   position: fixed;
   top: 0;
   bottom: 0;
--- a/apps/src/lab2/views/Lab2Wrapper.module.scss
+++ b/apps/src/lab2/views/Lab2Wrapper.module.scss
@@ -8,8 +8,6 @@
   left: 0;
   height: 100%;
   width: 100%;
-  opacity: 0.8;
-  background-color: $light_black;
   z-index: 1,
 }
 
--- a/apps/src/lab2/views/Lab2Wrapper.tsx
+++ b/apps/src/lab2/views/Lab2Wrapper.tsx
@@ -73,8 +73,8 @@ const Lab2Wrapper: React.FunctionComponent<Lab2WrapperProps> = ({children}) => {
       <div
         id="lab-container"
         className={classNames(
-          moduleStyles.labContainer,
-          isLoading && moduleStyles.labContainerLoading
+          !isShareView && moduleStyles.labContainer,
+          isShareView && moduleStyles.shareViewContainer
         )}
       >
         {children}
--- a/apps/src/lab2/views/Lab2Wrapper.tsx
+++ b/apps/src/lab2/views/Lab2Wrapper.tsx
@@ -73,8 +73,8 @@ const Lab2Wrapper: React.FunctionComponent<Lab2WrapperProps> = ({children}) => {
       <div
         id="lab-container"
         className={classNames(
-          isLoading && moduleStyles.labContainerLoading,
           moduleStyles.labContainer,
+          isLoading && moduleStyles.labContainerLoading,
           isShareView && moduleStyles.labContainerShareView
         )}
       >
--- a/apps/src/lab2/views/Lab2Wrapper.tsx
+++ b/apps/src/lab2/views/Lab2Wrapper.tsx
@@ -73,8 +73,9 @@ const Lab2Wrapper: React.FunctionComponent<Lab2WrapperProps> = ({children}) => {
       <div
         id="lab-container"
         className={classNames(
-          !isShareView && moduleStyles.labContainer,
-          isShareView && moduleStyles.shareViewContainer
+          isLoading && moduleStyles.labContainerLoading,
+          moduleStyles.labContainer,
+          isShareView && moduleStyles.labContainerShareView
         )}
       >
         {children}
--- a/apps/src/mixins.scss
+++ b/apps/src/mixins.scss
@@ -32,18 +32,3 @@
   }
 }
 
-@mixin slider-thumb {
-  height: 20px;
-  width: 4px;
-  border-radius: 1px;
-  background: white;
-  box-shadow: 1px 1px $neutral_dark80;
-}
-
-@mixin slider-track {
-  width: 100%;
-  height: 6px;
-  background: $light_gray_700;   
-  border-radius: 10px;
-  border: 0.5px solid $neutral_dark;
-}
--- a/apps/src/mixins.scss
+++ b/apps/src/mixins.scss
@@ -31,4 +31,3 @@
     border: none !important;
   }
 }
-
--- a/apps/src/mixins.scss
+++ b/apps/src/mixins.scss
@@ -31,3 +31,19 @@
     border: none !important;
   }
 }
+
+@mixin slider-thumb {
+  border: 1px solid $neutral_dark;
+  height: 32px;
+  width: 6px;
+  border-radius: 1px;
+  background: yellow;
+}
+
+@mixin slider-track {
+  width: 100%;
+  height: 6px;
+  background: $light_gray_700;   
+  border-radius: 1px;
+  border: 0.5px solid $neutral_dark;
+}
--- a/apps/src/mixins.scss
+++ b/apps/src/mixins.scss
@@ -33,11 +33,11 @@
 }
 
 @mixin slider-thumb {
-  border: 1px solid $neutral_dark;
-  height: 32px;
-  width: 6px;
+  height: 20px;
+  width: 4px;
   border-radius: 1px;
   background: white;
+  box-shadow: 1px 1px $neutral_dark80;
 }
 
 @mixin slider-track {
--- a/apps/src/mixins.scss
+++ b/apps/src/mixins.scss
@@ -37,13 +37,13 @@
   height: 32px;
   width: 6px;
   border-radius: 1px;
-  background: yellow;
+  background: white;
 }
 
 @mixin slider-track {
   width: 100%;
   height: 6px;
   background: $light_gray_700;   
-  border-radius: 1px;
+  border-radius: 10px;
   border: 0.5px solid $neutral_dark;
 }
--- a/apps/src/music/views/MusicPlayView.tsx
+++ b/apps/src/music/views/MusicPlayView.tsx
@@ -19,13 +19,13 @@ interface MusicPlayViewProps {
 const MusicPlayView: React.FunctionComponent<MusicPlayViewProps> = ({
   setPlaying,
 }) => {
-  const {isPlaying, lastMeasure, currentPlayheadPosition} = useAppSelector(
-    state => state.music
-  );
-  const progressValue =
-    lastMeasure === 1
+  const isPlaying = useAppSelector(state => state.music.isPlaying);
+  const progressValue = useAppSelector(state => {
+    const {currentPlayheadPosition, lastMeasure} = state.music;
+    return lastMeasure === 1
       ? 0
       : ((currentPlayheadPosition - 1) / (lastMeasure - 1)) * 100;
+  });
   const progressSliderValue = progressValue.toString();
 
   const projectName = useAppSelector(state => state.lab.channel?.name);
@@ -47,9 +47,12 @@ const MusicPlayView: React.FunctionComponent<MusicPlayViewProps> = ({
   // Requires HTTPS connection.
   // For testing purposes, we can pass a query parameter canShare=true to force the button to appear.
   const canShareInternal = queryParams('canShare') === 'true';
-  const canShare =
-    (navigator && navigator.canShare && navigator.canShare(shareData)) ||
-    canShareInternal;
+  const canShare = useMemo(() => {
+    return (
+      (navigator && navigator.canShare && navigator.canShare(shareData)) ||
+      canShareInternal
+    );
+  }, [canShareInternal, shareData]);
 
   const projectManager = Lab2Registry.getInstance().getProjectManager();
   const onShareProject = useCallback(() => {
@@ -57,9 +60,7 @@ const MusicPlayView: React.FunctionComponent<MusicPlayViewProps> = ({
   }, [shareData]);
   const onViewCode = useCallback(() => {
     if (projectManager) {
-      projectManager.flushSave().then(() => {
-        projectManager.redirectToView();
-      });
+      projectManager.redirectToView();
     }
   }, [projectManager]);
   const onRemix = useCallback(() => {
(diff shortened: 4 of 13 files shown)


apps/src/music/views/music-play-view.module.scss
```diff
@@ -37,13 +50,12 @@
...
       align-items: center;
-      gap: 0.5rem;
+      gap: 1rem;
 
       input[type='range'] {
         overflow: hidden;
         width: 100%;
-        height: 4px;
-        accent-color: $light_gray_200;
+        accent-color: yellow;
```
levadadenys: please note: worth checking how it looks like in different browsers, since this won't affect all of the browsers. (when using $light_gray_200 it's just not that visible so to say, but still there's a difference between different browsers)

https://css-tricks.com/styling-cross-browser-compatible-range-inputs-css/

https://www.cssportal.com/style-input-range/

apps/src/music/views/MusicPlayView.tsx
```diff
@@ -30,76 +42,107 @@ const MusicPlayView: React.FunctionComponent<MusicPlayViewProps> = ({
...
-        </div>
+      <div className={moduleStyles.musicPlayViewCardContainer}>
+        <img
+          className={moduleStyles.musicPlayViewImage}
+          src={musicPlayViewLogo}
+          alt="Code.org play view logo"
+        />
+        <div className={moduleStyles.musicPlayViewCard}>
+          <div className={moduleStyles.musicPlayViewInfoSection}>
+            <Heading2 className={moduleStyles.musicPlayViewInfoText}>
```
levadadenys: nit: (just a comment, no need to update this) Coming back to the conversation re setting a className to each specific text element, now that we have a live example, in my opinion it's making code harder to read and adding extra complexity. Personal preference, but I like more when all the general styles are stored in some-level-container and classNames applied to text only when text is styled in 'exceptional' way (like emphasised text, error message, etc)

cc @sanchitmalhotra126

levadadenys (commented):
LGTM overall! 😍 
Couple non blocking comments.
Not approving only because PR is in Draft state

dashboard/app/controllers/projects_controller.rb
```diff
@@ -589,9 +589,12 @@ def limited_gallery?
   # Redirect to the correct view/edit page for Lab2 projects. If a project owner is on a /view
   # page, redirect to /edit. If a non-owner is on an /edit page, redirect to /view.
   # For legacy (non-Lab2) labs, this is handled on the front-end.
-  private def redirect_edit_view_for_lab2
+  private def redirect_view_for_lab2
     return nil unless @level.uses_lab2?
-
+    # If the user is on the play view '/projects/channel_id', set `response_content`.`
+    if params[:share] == true
```
fisher-alice: Not sure if this is okay to do this for `lab2` labs or should I only do for music lab projects for now? @sanchitmalhotra126

apps/src/music/views/MusicPlayView.tsx
fisher-alice: @sanchitmalhotra126 - lmk what you think. Thanks y'all!

apps/src/music/views/music-play-view.module.scss
fisher-alice: Thanks so much! I've been noticing the differences so I appreciate the resources.

apps/src/music/views/music-play-view.module.scss
```diff
@@ -37,23 +51,48 @@
     .musicPlayViewPlaySection {
       display: flex;
       align-items: center;
-      gap: 0.5rem;
+      gap: 1rem;
+
+      input[type=range] {
```
fisher-alice: Reference: https://css-tricks.com/styling-cross-browser-compatible-range-inputs-css/
Style range inputs so that styling is consistent across Firefox, Edge, Safari, and Chrome.

apps/src/music/views/music-play-view.module.scss
levadadenys: sorry, I've duplicated the links in the initial comment. Updated it. added https://www.cssportal.com/style-input-range/  which allows to generate cross-browser styles for input type range more easily

dashboard/app/controllers/projects_controller.rb
```diff
@@ -599,6 +600,14 @@ def limited_gallery?
     return redirect_to "/projects/#{params[:key]}/#{params[:channel_id]}/view" if !is_owner && request.path.ends_with?('/edit')
   end
 
+  private def set_lab2_responsive_view_options
+    return nil unless @level.uses_lab2?
+  
+    # If the user is on the play view '/projects/channel_id', set `response_content`.`
+    if params[:share] == true
```
fisher-alice: This is checked in `application.html.haml` [here](https://github.com/code-dot-org/code-dot-org/blob/4923cc336bdec49e275bc235dcb13087d9aa6139/dashboard/app/views/layouts/application.html.haml#L16).

apps/src/lab2/views/Lab2Wrapper.module.scss
```diff
@@ -1,6 +1,15 @@
 @import 'color.scss';
 @import './common.scss';
 
+.shareViewContainer {
```
sanchitmalhotra126: What do we need this for (or what does the view look like without this)?

apps/src/mixins.scss
```diff
@@ -31,3 +31,19 @@
     border: none !important;
   }
 }
+
+@mixin slider-thumb {
```
sanchitmalhotra126: Wondering if it makes sense to have these here, or to just keep in the music-play-view CSS file...typically I've seen shared mixins treated more like functions that take arguments to encapsulate shared bits of styling, but these seem pretty specific to the MusicPlayView? Alternatively I could see us creating a common slider component that uses these styles, but then we'd include these in the CSS module for that component rather than here. Another idea would be to create a mixin that handles all the variations between browsers internally and takes a few arguments such as color, height, etc.

apps/src/music/views/MusicPlayView.tsx
```diff
@@ -3,25 +3,37 @@ import React, {memo, useCallback, useMemo} from 'react';
...
+  setPlaying,
 }) => {
+  const {isPlaying, lastMeasure, currentPlayheadPosition} = useAppSelector(
+    state => state.music
+  );
+  const progressValue =
+    lastMeasure === 1
+      ? 0
+      : ((currentPlayheadPosition - 1) / (lastMeasure - 1)) * 100;
+  const progressSliderValue = progressValue.toString();
```
sanchitmalhotra126: Out of curiosity, why is this conversion necessary?

apps/src/music/views/MusicPlayView.tsx
```diff
@@ -3,25 +3,37 @@ import React, {memo, useCallback, useMemo} from 'react';
...
-  projectName,
+  setPlaying,
 }) => {
+  const {isPlaying, lastMeasure, currentPlayheadPosition} = useAppSelector(
+    state => state.music
+  );
+  const progressValue =
+    lastMeasure === 1
+      ? 0
+      : ((currentPlayheadPosition - 1) / (lastMeasure - 1)) * 100;
```
sanchitmalhotra126: Might be good to put this whole calculation inside the selector so that way the `progressValue` is automatically memoized and this component only updates when that value changes, rather than any time the position updates? Actually speaking it may end up being the same amount of updates, but for example if we were to round this number, then it would be helpful since we'd only update when the actual value changed. Ex:
```
const progressValue = useAppSelector(state => {
  const {currentPlayheadPosition, lastMeasure} = state.music;
  // ... calculate value
  return progressValue;
}

File: apps/src/music/views/MusicPlayView.tsx [hunk 10]
```diff
@@ -30,76 +42,106 @@ const MusicPlayView: React.FunctionComponent<MusicPlayViewProps> = ({
...
 
+  const projectManager = Lab2Registry.getInstance().getProjectManager();
   const onShareProject = useCallback(() => {
     navigator?.share(shareData);
   }, [shareData]);
   const onViewCode = useCallback(() => {
-    console.log('view code');
-  }, []);
+    if (projectManager) {
+      projectManager.flushSave().then(() => {
```
sanchitmalhotra126: I don't think we need to `flushSave` here since the project isn't actually changing in this view, right?

apps/src/music/views/MusicPlayView.tsx
```diff
@@ -30,76 +42,106 @@ const MusicPlayView: React.FunctionComponent<MusicPlayViewProps> = ({
...
-  // Share button will only appear when users browser supports the Web Share API
+  // Share button will only appear when user's browser supports the Web Share API.
   // (Can be mobile browsers and some desktop browser like macOS Safari)
-  // Requires HTTPS connection (adhoc or production page).
-  // Was unable to access it on localhost with our routing (http://localhost-studio.code.org:3000/
-  // is not considered a safe URL by browser, only http://localhost:3000/ is).
+  // Requires HTTPS connection.
+  // For testing purposes, we can pass a query parameter canShare=true to force the button to appear.
+  const canShareInternal = queryParams('canShare') === 'true';
   const canShare =
```
sanchitmalhotra126: nit: `useMemo` for this?

apps/src/music/views/MusicPlayView.tsx
sanchitmalhotra126: Hmm, I think I personally prefer the classnames still since it's easier to look up the styles in the stylesheet? That being said, might be good to make these names less verbose if it makes it easier to read by removing the `musicPlayView` prefix (e.g. just `card`, `infoSection`, etc). The class names are already scoped to the specific module and get converted to IDs in production, so we don't need to worry about naming collisions.

apps/src/music/views/music-play-view.module.scss
sanchitmalhotra126: Cool stuff!

dashboard/app/controllers/projects_controller.rb
sanchitmalhotra126: Wondering if we should technically use the `sharing` value computed on line 374 so that we're consistent with what we pass into `level_view_options`. It shouldn't matter for now since we don't do anything special for embedded projects in lab2 currently, but just to be safe? Curious if @breville has thoughts

sanchitmalhotra126 (commented):
Looks pretty good overall! A few comments/questions

apps/src/lab2/views/Lab2Wrapper.module.scss
fisher-alice: The view looks like this without this styling.
Screenshot 2024-05-01 at 11 38 45 AM

So I added this container to cover up the header area and extended the container to the edges.

apps/src/mixins.scss
fisher-alice: I can keep in the music-play-view CSS file for now then. I am going to probably refactor the slider as a follow-up PR. Thanks!

apps/src/music/views/MusicPlayView.tsx
fisher-alice: From https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/range - The [value](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#value) attribute contains a string which contains a string representation of the selected number. The value is never an empty string ("").

But I believe it works as a number as well. I was also thinking since`min` and `max` are defined as strings, just being consistent. But looks like `min` and `max` could be assigned numbers. I don't have a preference on this. Can change if you prefer as numbers.

apps/src/lab2/views/Lab2Wrapper.module.scss
sanchitmalhotra126: Oh interesting - could we also just update `.labContainer` to have a `top` of 0 in share view? Would that accomplish the same thing?

(Edit: also noticed some other potential cleanup we could do with the container styles, but that's outside of the scope of this PR and probably not something we want to do this close to launch. Logged https://codedotorg.atlassian.net/browse/LABS-759 for now)

apps/src/music/views/MusicPlayView.tsx
fisher-alice: Thanks! Will shorten style names.

dashboard/app/controllers/projects_controller.rb
fisher-alice: Yeah - I was wondering about this earlier. Thanks! Will wait on Brendan to weigh in.

apps/src/music/views/MusicPlayView.tsx
sanchitmalhotra126: Gotcha, no strong preference either way, what you have is fine!

apps/src/lab2/views/Lab2Wrapper.module.scss
fisher-alice: Sure - wil do! And thanks for logging the ticket.

apps/src/lab2/views/Lab2Wrapper.tsx
```diff
@@ -73,8 +73,9 @@ const Lab2Wrapper: React.FunctionComponent<Lab2WrapperProps> = ({children}) => {
       <div
         id="lab-container"
         className={classNames(
+          isLoading && moduleStyles.labContainerLoading,
           moduleStyles.labContainer,
-          isLoading && moduleStyles.labContainerLoading
+          isShareView && moduleStyles.labContainerShareView
```
sanchitmalhotra126: super minor - mind reordering this with loading second so the diff is clear (that we only added a new style, not updated an existing style)

apps/src/mixins.scss
```diff
@@ -31,3 +31,4 @@
     border: none !important;
   }
 }
+
```
sanchitmalhotra126: nit: revert?

apps/src/music/views/MusicPlayView.tsx
sanchitmalhotra126: Oh but might as well move the `toString` into the selector too?

apps/src/music/views/MusicPlayView.tsx
```diff
@@ -30,76 +42,105 @@ const MusicPlayView: React.FunctionComponent<MusicPlayViewProps> = ({
...
+  const projectManager = Lab2Registry.getInstance().getProjectManager();
   const onShareProject = useCallback(() => {
     navigator?.share(shareData);
   }, [shareData]);
   const onViewCode = useCallback(() => {
-    console.log('view code');
-  }, []);
+    if (projectManager) {
+      projectManager.redirectToView();
+    }
```
sanchitmalhotra126: nit/optional - could use optional chaining to make this `projectManager?.redirectToView()`

apps/src/music/views/music-play-view.module.scss
```diff
@@ -1,59 +1,112 @@
 @import 'color';
 
-.musicPlayViewContainer {
+@mixin slider-thumb {
```
sanchitmalhotra126, line 3: Minor - technically it would make more sense to make these [placeholder selectors](https://sass-lang.com/documentation/style-rules/placeholder-selectors/) (`%slider-thumb`) and use them with `@extend`, i.e.
```
input[type="range"]::-moz-range-thumb {
  @extend %slider-thumb;
}
```
since mixins are intended to be used more like functions. It accomplishes the same thing but using placeholders has some minor benefits in reducing the size of compiled css files.

sanchitmalhotra126 (approved):
A few minor things but LGTM! Nice work!

dashboard/app/controllers/projects_controller.rb
fisher-alice: Discussed offline - I'll go ahead and update to use `sharing`.

apps/src/music/views/music-play-view.module.scss
fisher-alice: Cool! Thanks so much. Learning a lot of CSS from this PR.

fisher-alice: Interestingly, the placeholder selector does not work with webkit, but it does work with firefox.
So we can either use mixins, or have duplicate declarations. I don't think it makes sense to have a placeholder if used only once. Lmk what you think. Thanks!

fisher-alice: @sanchitmalhotra126 and I messaged offline and agree to keep as mixins.

apps/src/music/views/music-play-view.module.scss
bencodeorg: Slick, need this for AI Chat too!

PR #58298, opened 2024-04-27
Music: small rendering fix

breville:
Fixes a React warning by ensuring we have unique keys for lists of elements in the Timeline.


thomasoniii (approved):
lgtm 🚀

apps/src/music/views/TimelineSimple2Events.tsx
```diff
@@ -173,6 +173,7 @@ const TimelineSimple2Events: React.FunctionComponent<
       <div id="timeline-events-function-extents">
         {uniqueFunctionExtentsArray.map((functionExtents, index) => (
           <FunctionExtentsSimple2
+            key={index}
```
levadadenys, line 175: Don't have much context on this array, but wanted to ask if the order of this array may change?
Using element's index should be ok if array order stays the same, though if the order might be changed - might potentially cause some bugs on reordering - more info here
https://react.dev/learn/rendering-lists#keeping-list-items-in-order-with-key

breville: Very interesting point.  As the Timeline content changes, it is certainly possible that keeping these keys consistent between renders would be counterproductive.  

I've updated the PR with a new approach that ensures we have unique values for each new execution of the code, which is what drives the major changes to the layout of the Timeline.  Am curious if it's sufficient for now, though I'm sure there are cleverer ways we could render (and optimize rendering) of this Timeline in the future.

breville: We've actually gone back to the simpler approach for now.  @thomasoniii and I talked things through, and we think that it's probably alright to use the index, and it keeps things simpler as we approach launch.  We'll keep an eye out for rendering issues, though.

sanchitmalhotra126 (approved):

PR #58299, opened 2024-04-28
Music: pack dialog grid

breville:
The pack dialog's grid of packs now has fixed-width entries.

### Screenshots

Here are a couple different window widths, showing how the grid has fixed-width entries, with the number of columns being dynamic:

Screenshot 2024-04-28 at 10 12 59 PM

###

music-pack-dialog-grid-1

### DIFF
--- a/apps/src/music/views/PackDialog.module.scss
+++ b/apps/src/music/views/PackDialog.module.scss
@@ -49,63 +49,70 @@
       margin-bottom: 14px;
     }
 
-    .packs {
+    .packsContainer {
       width: 100%;
       height: 100%;
       overflow-y: auto;
       flex: 1;
-      display: grid;
-      grid-template-columns: 25% 25% 25% 25%;
-
       background-color: $neutral_dark;
       border: solid $neutral_dark 2px;
       box-sizing: border-box;
       border-radius: 5px;
+      padding: 16px;
 
-      .pack {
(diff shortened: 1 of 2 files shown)


thomasoniii (approved):
lgtm 🚀

apps/src/music/views/PackDialog.module.scss
```diff
@@ -49,63 +49,70 @@
...
+            }
 
-          &Preview {
-            margin-right: 10px;
+            &Preview {
+              margin-right: 10px;
 
-            .preview {
-              color: $neutral_dark40;
+              .preview {
```
sanchitmalhotra126, line 109: Minor - I don't know how strict we've been about this in our codebase but I do think there's a preference to avoid a lot of nesting in CSS files (typically no more than 3 levels deep). So for example I would personally probably pull out the `.preview` class into a top level selector (unless there's a strict need that this rule only apply to `preview` elements inside `.pack` elements)

sanchitmalhotra126 (approved):

apps/src/music/views/PackDialog.module.scss
breville: Oh, I didn't know about this.  For these files I've quite enjoyed having the structure of the CSS match the structure of the code.

PR #58306, opened 2024-04-29
Fix and test PL unit completion for contained levels

bethanyaconnor:
Fixes a case where, when a unit contained a bubble choice level, greater than 100% completion was being shown. To fix that, I changed the logic to look for the levels, instead of the units, associated with a UserLevel.

This did come with a side effect of not counting predict levels correctly. Predict levels are set up such that the Blockly level is in the ScriptLevel and contains the FreeResponse/Multi level. However, the UserLevel is created for the FreeResponse/Multi level. I added some special handling for that case.

I added a test for both BubbleChoice and predict levels.

I largely used the investigation in https://github.com/code-dot-org/code-dot-org/pull/57121 as a starting point here!

## Warning!!

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/dashboard/app/models/user.rb
+++ b/dashboard/app/models/user.rb
@@ -1936,7 +1936,15 @@ class User < ApplicationRecord
     pl_user_scripts = user_scripts.select {|us| us.script.pl_course?}
     pl_scripts = pl_user_scripts.map(&:script)
 
-    user_levels = UserLevel.where(user: self, script: pl_scripts)
+    levels = pl_scripts.map(&:levels).flatten
+    level_ids = levels.pluck(:id)
+
+    levels.filter {|l| l.contained_levels.count > 0 && !l.is_a?(BubbleChoice)}.each do |level|
+      contained_levels = level.contained_levels
+      level_ids += contained_levels.pluck(:id)
+    end
+
+    user_levels = UserLevel.where(user: self, script: pl_scripts, level_id: level_ids)
     return [] if user_levels.empty?
     user_levels_by_script = user_levels.group_by(&:script_id)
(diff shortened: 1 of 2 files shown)


bethanyaconnor:
Tagging Teacher Tools in case they have any insight!

etaderhold (approved):
This change LGTM.

I do wonder about the difference between this behavior of looking at all contained levels individually, vs. what we do on the progress table where we just look at the first contained level for non-bubble-choice levels. Are there good reasons to do things differently in these two places, or would it be better to standardize on one way?

https://github.com/code-dot-org/code-dot-org/blob/7c97b3f391c1798a705c655023eccbd05dcc41a6/dashboard/app/models/levels/level.rb#L809

bethanyaconnor:
> I do wonder about the difference between this behavior of looking at all contained levels individually, vs. what we do on the progress table where we just look at the first contained level for non-bubble-choice levels. 

I think, in practice, levels can only have one (or zero) contained level, so they're effectively the same. But I would love to standardize on that and actually enforce it. Do you want me to change this logic to only look at the first contained level?

etaderhold:
> I think, in practice, levels can only have one (or zero) contained level, so they're effectively the same.

I ran `Level.all.filter{|level| !level.is_a?(BubbleChoice)}.filter {|level| (level.contained_levels.length rescue 0) > 1 }` in a Rails console in my local dev environment and it came back with just two levels with more than one contained level. So yes, agreed that it's nearly always the same in practice.

> But I would love to standardize on that and actually enforce it. Do you want me to change this logic to only look at the first contained level?

Sounds fine!

PR #58309, opened 2024-04-29
Gen AI: detect user profanity violation

bencodeorg:
## Warning!!

## Description

Puts user content in gen AI lab through our profanity filter. If we detect profanity, we hide the user's message and show a notification that the content has been marked as profane. I've refactored our existing model status message UI to re-use when a user message is marked as profane.

We continue to send profane messages to/from the server for logging purposes, but exclude them from what gets sent to Sagemaker. We now log a message status (ok, profanity, etc) in addition to type and content.

https://github.com/code-dot-org/code-dot-org/assets/25372625/a27b1aab-beaa-4079-8b5f-7f90796aa7e8

## Testing story

I set up profanity filtering to work locally, and observed messages being flagged appropriately (see video above). I also added some new unit tests to cover this behavior.

## Follow-up work

- Handle profanity coming from Sagemaker: this PR doesn't handle profanity coming from Sagemaker -- we should probably handle this as part of a more general "server error" PR, since I assume we don't actually want to tell the user that the model was trying to send them profane content (probably something more generic, like "an error has occurred"?)
- Handle PII from user: this should be a quick follow-up, but wanted to get feedback on the overall approach here first.


apps/src/aichat/redux/aichatRedux.ts
```diff
@@ -276,17 +277,27 @@ export const submitChatContents = createAsyncThunk(
...
       updateChatMessageSession({
         id: newMessage.id,
-        sessionId: chatApiResponse.sessionId,
+        sessionId: chatApiResponse.session_id,
       })
     );
 
+    if (chatApiResponse?.status === 'profanity') {
+      // Logging to allow visibility into flagged content.
+      console.log(chatApiResponse);
```
levadadenys, line 291: nit: I see that you've added this in a separate commit, but still wanted to ask if this should be removed before merging?

apps/src/aichat/redux/aichatRedux.ts
```diff
@@ -296,14 +307,12 @@ export const submitChatContents = createAsyncThunk(
...
       };
-      thunkAPI.dispatch(addChatMessage(assistantChatMessage));
-    } else {
-      // TODO: Update most recent user message's status if PII or profanity violation.
-      // latest message's id is stored at `newMessageId`.
-      console.log('Did not receive assistant response.');
+      return thunkAPI.dispatch(addChatMessage(assistantChatMessage));
     }
+
+    console.log('Could not handle response.');
```
levadadenys, line 315: nit: should it be removed before merging?

levadadenys (commented):
LGTM from Design System  perspective. (Don't have enough context to approve from back end and gen ai front end (logic) perspective )

lib/cdo/shared_constants.rb
```diff
@@ -708,15 +708,19 @@ module SharedConstants
...
+  AI_INTERACTION_STATUS = {
     ERROR: 'error',
     PII_VIOLATION: 'pii_violation',
     PROFANITY_VIOLATION: 'profanity_violation',
     OK: 'ok',
     UNKNOWN: 'unknown',
   }.freeze
 
+  AI_TUTOR_INTERACTION_STATUS = AI_INTERACTION_STATUS
+  AICHAT_INTERACTION_STATUS = AI_INTERACTION_STATUS
```
fisher-alice: Nit: Since you're importing `AichatInteractionStatus` as `Status` in `aichatRedux` and `ChatMessage`, maybe just using `AiInteractionStatus`?

apps/src/aichat/redux/aichatRedux.ts
```diff
@@ -351,13 +360,13 @@ const aichatSlice = createSlice({
     setShowWarningModal: (state, action: PayloadAction<boolean>) => {
       state.showWarningModal = action.payload;
     },
-    updateChatMessageStatus: (
+    updateUserChatMessageStatus: (
```
fisher-alice, line 363: I believe I updated `newMessage` with status `Status.OK` as the default in https://github.com/code-dot-org/code-dot-org/pull/57944 on [line 260](https://github.com/code-dot-org/code-dot-org/blob/99a91995d5b86aa754ae5c5e4bc946d401683ed0/apps/src/aichat/redux/aichatRedux.ts#L260) above.
What do you think about the setting the default back to `Status.UNKNOWN` and when we get the response, always updating `newMessageId` (to OK or whichever error)?

I'm wondering if it's safer to mark as `UNKNOWN` until we get confirmation that it doesn't have PII or profanity included. If so, then `updateChatMessageSession` and `updateChatMessageStatus` could be combined since `status` would always be updated once response is received.

fisher-alice: But then that leaves the question of what do we do with a message with status `UNKNOWN`. Currently we [display](https://github.com/code-dot-org/code-dot-org/blob/99a91995d5b86aa754ae5c5e4bc946d401683ed0/apps/src/aichat/views/ChatMessage.tsx#L27).

apps/src/aichat/views/ChatNotificationMessage.tsx
```diff
@@ -0,0 +1,38 @@
+import React from 'react';
+import classNames from 'classnames';
+
+import Button from '@cdo/apps/componentLibrary/button';
+import FontAwesomeV6Icon from '@cdo/apps/componentLibrary/fontAwesomeV6Icon';
+
+import styles from './chat-notification-message.module.scss';
+
+const ChatNotificationMessage: React.FunctionComponent<{
```
fisher-alice, line 9: Nice!

apps/src/aichat/views/ChatNotificationMessage.tsx
```diff
@@ -0,0 +1,38 @@
...
+  return (
+    <div className={classNames(styles.container, containerClass)}>
+      <div>
+        <FontAwesomeV6Icon
+          iconName={iconName}
+          className={classNames(iconClass, styles.icon)}
+        />
+        {content}
+      </div>
+      {onRemove && (
```
fisher-alice, line 25: So at this point, we can only remove model update messages?

apps/src/aichat/views/chat-notification-message.module.scss
```diff
@@ -0,0 +1,25 @@
...
+  margin: 3px 0;
+  min-height: 25px;
+
+  .removeButton {
+    background-color: transparent;
+    color: $light_gray_700;
+
+    &:hover {
+      background-color: transparent;
+      color: $light_gray_700;
```
fisher-alice, line 22: Would it be useful to have a slight color change on hover for the user?

apps/src/aichat/redux/aichatRedux.ts
fisher-alice: I also wonder since we already have messages marked as 'profane' stored on the backend, do we need/want them to be stored in redux?

apps/src/aichat/redux/aichatRedux.ts
bencodeorg: Yeah sorry I meant to add a github comment here, this was an explicit recommendation from Brendan and Dan from their previous experiences using our profanity filter:

https://codedotorg.slack.com/archives/C05N0D3GUER/p1714009688665089?thread_ts=1713960007.526659&cid=C05N0D3GUER

apps/src/aichat/redux/aichatRedux.ts
bencodeorg: This is just a rewording of the existing console log ("'Did not receive assistant response.") if we can't handle the response. We need to do an error handling pass as a follow-up, so hoping we'd be able to remove this and do fail more gracefully at that point.

dashboard/app/controllers/aichat_controller.rb
```diff
@@ -15,24 +15,46 @@ def chat_completion
...
+    sagemaker_response = AichatSagemakerHelper.request_sagemaker_chat_completion(input)
     latest_assistant_response = AichatSagemakerHelper.get_sagemaker_assistant_response(sagemaker_response)
-    assistant_message = {
-      role: "assistant",
-      content: latest_assistant_response,
-    }
 
-    session_id = log_chat_session(assistant_message)
-    return render(status: :ok, json: assistant_message.merge({sessionId: session_id}).to_json)
+    assistant_message = {role: "assistant", content: latest_assistant_response, status: SharedConstants::AICHAT_INTERACTION_STATUS[:OK]}
```
fisher-alice: Just curious - you mentioned that follow-up would include checking response from Sagemaker. Are we planning to just use WebPurify as we do for user messages?

dashboard/test/controllers/aichat_controller_test.rb
```diff
@@ -72,16 +72,45 @@ class AichatControllerTest < ActionController::TestCase
...
+    params = @valid_params.merge(
+      storedMessages: [
+        {status: 'profanity_violation', role: 'user', content: 'damn'},
+        ok_message
+      ]
+    )
+
+    # Note that second expected argument filters out the previous profane message
+    # in what we send to Sagemaker.
+    AichatSagemakerHelper.expects(:format_inputs_for_sagemaker_request).
```
fisher-alice, line 100: Do you mind explaining again why you needed `to_unsafe_h` to test this function? I didn't quite understand from [your comment](https://github.com/code-dot-org/code-dot-org/blob/99a91995d5b86aa754ae5c5e4bc946d401683ed0/dashboard/app/controllers/aichat_controller.rb#L39) in `aichat_controller.rb`.

fisher-alice (approved):
Looks great! Just left a couple minor comments and a question about storing profane messages on the front-end.

dashboard/app/controllers/aichat_controller.rb
bencodeorg: Yes that's my understanding, does that sound right to you?

apps/src/aichat/views/ChatNotificationMessage.tsx
bencodeorg: Yep!

apps/src/aichat/views/chat-notification-message.module.scss
bencodeorg: Maybe? It at least changes the cursor to clickable, not sure how design would want that styled (red x? circle surrounding x turns red?)

dashboard/test/controllers/aichat_controller_test.rb
bencodeorg: Yeah...so if you're passing in params[:key] directly, you're getting an instance of `ActionController::Params`:

https://api.rubyonrails.org/classes/ActionController/Parameters.html#method-i-5B-5D

And if you try to set an `expects` on that argument, you get different instances, and the test fails.

You can compare hashes for equality, so `to_unsafe_h` allows you do that.

There certainly might be a better way to do this, but I couldn't figure it out. I thought you might be able to "record" the arguments passed to a stubbed function (then call `to_h` or something on the recorded argument and do a comparison), but couldn't figure out how to do that?

Here's some info on why to use "strong params" generally, which made me think in this case using `to_unsafe_h` isn't actually unsafe (we do it in a place or two already in our code base):

https://www.writesoftwarewell.com/why-use-strong-parameters-in-rails/

apps/src/aichat/redux/aichatRedux.ts
bencodeorg: [tl;dr these are tricky, happy to chat further offline if any of this is unclear!]

mmm good questions...a couple thoughts:

re: keeping profane messages in redux or not, we do need some sort of entry in the list of messages so that we can display the warning message, yeah? In general, we do rely on the messages in the UI matching the messages stored in Rails for a given session, which was part of my reason for wanting to exclude the delete button on these notifications, and I think it'd be more complicated to manage different messages in Redux vs. Rails. I did have the thought working on this though that keeping the message list in sync between Rails and Redux feels fragile, and I was hoping we might be able to refactor to simplify that. Open to ideas on how to do that!

re: unknown status, could you clarify what you mean by "safer" to use the unknown status (we don't do anything with the unknown status, yeah)? I was inclined to keep things simple and just use the ok status for now if we don't use unknown. But happy to update if we change that going forward.

lib/cdo/shared_constants.rb
bencodeorg: Yep I like it, updated! [commit here](https://github.com/code-dot-org/code-dot-org/pull/58309/commits/87e92be4f8793696608186bdcaa0e8979077eee6)

apps/src/aichat/redux/aichatRedux.ts
fisher-alice: > we do need some sort of entry in the list of messages so that we can display the warning message, yeah? In general, we do rely on the messages in the UI matching the messages stored in Rails for a given session

I was wondering if we could reassign the content of the profane message in redux to something like 'this message was marked as profane' so that we're not sending messages with profanity/pii included in them back and forth in long conversations. But I you brought up the issue of keeping things in sync since we're replacing the chat message history in the same chat session.

I wonder if we could create a new chat session when a profane/pii user message is sent, like a model update? Either way - I think what's in this PR works and makes sense!

fisher-alice: > could you clarify what you mean by "safer" to use the unknown status (we don't do anything with the unknown status, yeah)? I was inclined to keep things simple and just use the ok status for now if we don't use unknown. But happy to update if we change that going forward.

I was thinking of the case of a user sending a message with profanity/pii and for whatever reason, we don't get a server response. I believe that message remains as `OK`. But I think error handling can be handled later.

dashboard/test/controllers/aichat_controller_test.rb
fisher-alice: Gotcha - thanks so much for the explanation!

apps/src/aichat/redux/aichatRedux.ts
bencodeorg: Ooh that makes sense re: profanity and error handling, good callout! Like you said, I do think it makes sense to cross that bridge when we get into error handling.

bencodeorg: And re: profanity and a new session, I don't think we want a new session if a profane message is sent (we'd still want to retain the previous chat history in what's sent to the model, yeah?)

apps/src/aichat/redux/aichatRedux.ts
sanchitmalhotra126: Do we want to also log to Cloudwatch (via Lab2MetricsReporter) in case we can't get access to a student's console?

apps/src/aichat/redux/aichatRedux.ts
sanchitmalhotra126: Adding my two cents:
- I like the idea of using the UNKNOWN status for undetermined student messages, but agree that if we're not doing any different styling with it currently, then OK is fine too.
- I think we generally want to be diligent about having a single source of truth of all chat messages, and agree that it should be the front end rather than the back end. I think it's best to think about the backend controller as "stateless" where we just overwrite the previous chat session with the new messages sent from the front end, rather than relying on the data in the DB to know where we are (other than things like comparing customizations to check if we need to create a new session for example). That being said, it may help for us to start thinking about a more generic data model that contains all chat messages, profanity status, model updates, session ID, etc etc as separate from what gets presented in the actual chat window. In other words, the content of the chat window is some subset/function of the larger data model, which is our source of truth. Don't think this actually impacts the code right now, but just my thoughts on where and how we would want to store this data.

dashboard/app/controllers/aichat_controller.rb
```diff
@@ -15,24 +15,46 @@ def chat_completion
...
+    # Use to_unsafe_h here to allow testing this function.
+    # Safe params are primarily targeted at preventing "mass assignment vulnerability"
+    # which isn't relevant here.
+    input = AichatSagemakerHelper.format_inputs_for_sagemaker_request(
+      params.to_unsafe_h[:aichatModelCustomizations],
+      params.to_unsafe_h[:storedMessages].filter {|message| message[:status] == SharedConstants::AI_INTERACTION_STATUS[:OK]},
+      params.to_unsafe_h[:newMessage]
+    )
+    sagemaker_response = AichatSagemakerHelper.request_sagemaker_chat_completion(input)
     latest_assistant_response = AichatSagemakerHelper.get_sagemaker_assistant_response(sagemaker_response)
```
sanchitmalhotra126, line 48: I think this was commented elsewhere, but we'd also want to verify this response for profanity as well, right? (would be a separate task)

dashboard/app/controllers/aichat_controller.rb
```diff
@@ -15,24 +15,46 @@ def chat_completion
...
+          role: 'user',
+          content: params[:newMessage],
+          status: SharedConstants::AI_INTERACTION_STATUS[:PROFANITY_VIOLATION]
+        }
+      ]
+      session_id = log_chat_session(new_messages)
+      return render(
+        status: :ok,
+        json: {
+          status: filter_result.type,
```
sanchitmalhotra126, line 32: Should we convert this into an `AI_INTERACTION_STATUS`?

dashboard/app/controllers/aichat_controller.rb
```diff
@@ -15,24 +15,46 @@ def chat_completion
...
-    session_id = log_chat_session(assistant_message)
-    return render(status: :ok, json: assistant_message.merge({sessionId: session_id}).to_json)
+    assistant_message = {role: "assistant", content: latest_assistant_response, status: SharedConstants::AI_INTERACTION_STATUS[:OK]}
+    new_messages = [
+      {role: 'user', content: params[:newMessage], status: SharedConstants::AI_INTERACTION_STATUS[:OK]},
+      assistant_message,
+    ]
+    session_id = log_chat_session(new_messages)
+
+    render(status: :ok, json: assistant_message.merge({session_id: session_id}).to_json)
```
sanchitmalhotra126, line 57: super minor: should we keep `sessionId` camel case for the front end?

sanchitmalhotra126 (approved):
LGTM overall! A few comments/questions

dashboard/app/controllers/aichat_controller.rb
bencodeorg: Yes, doing this as a follow-up now!

dashboard/app/controllers/aichat_controller.rb
bencodeorg: I dunno...it's a different string ("profanity" vs "profanity_violation"), and feels slightly different (ie, not an interaction status, but a violation type that will also potentially include things like "address", etc that won't match 1:1 with an interaction status)

apps/src/aichat/redux/aichatRedux.ts
bencodeorg: Thanks for the thoughts! Agree that there is some sort of state management simplification that should be made here ("stateless" on the back end sounds promising)...will noodle on it a bit.

apps/src/aichat/redux/aichatRedux.ts
bencodeorg: Yeah that's a good call...I was thinking of logging server-side for any profanity detected from the model, so i could do the same for this? Do you know if this is the preferred Ruby logging client?

https://github.com/code-dot-org/code-dot-org/blob/staging/lib/cdo/aws/metrics.rb

dashboard/app/controllers/aichat_controller.rb
bencodeorg: haha I switched to the opposite :) It felt consistent with the fact we don't "tee up" the back end by converting all params we send to Rails into snake case? Relatedly, I was thinking it'd be nice to do something like what AI Tutor is doing in their controller:

https://github.com/code-dot-org/code-dot-org/blob/f280b95523fb6a3aabbe624250146256a9480cef/dashboard/app/controllers/ai_tutor_interactions_controller.rb#L18-L26

dashboard/app/controllers/aichat_controller.rb
sanchitmalhotra126: Gotcha, yeah maybe a follow-up then but it would be good to have some sort of constant mapping for this on the front end too (rather than just checking the string `'profanity'`?)

dashboard/app/controllers/aichat_controller.rb
sanchitmalhotra126: Ahh gotcha, yep not a huge deal either way and I do like what AI Tutor is doing in the snippet you linked, so maybe we can take a pass at cleaning this up later

PR #58315, opened 2024-04-29
Update ruby-prof gem to support ARM chipset

sureshc:
The current version of ruby-prof that we install (0.15.9) is failing to install on development environments that use the ARM chipset. Upgrade to the latest (1.7.0).

We only use this gem in `development` and `test` environments so this change is likely low risk.

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -91,7 +91,7 @@ group :development, :test do
   gem 'brakeman'
   gem 'database_cleaner-active_record', '~> 2.1.0'
   gem 'haml-rails' # haml (instead of erb) generators
-  gem 'ruby-prof'
+  gem 'ruby-prof', '>= 1.7.0'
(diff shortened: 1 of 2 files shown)


molly-moen (approved):

simonguest (approved):
Matches the version I've been successfully using in development.

Hamms (commented):
Could we formalize this requirement in the Gemfile itself, rather than relying on Gemfile.lock to preserve this level of specificity all on its own?

sureshc:
> Could we formalize this requirement in the Gemfile itself, rather than relying on Gemfile.lock to preserve this level of specificity all on its own?

Agreed. Done!

Hamms (approved):
LGTM, thanks!

PR #58319, opened 2024-04-29
feat: add DCDO flag for posting email param to finish signup page

stephenliang:
This PR adds scaffolding to enable email addresses to bypass the partial registration cache and go directly into the finish signup page. This is accomplished by adding a new DCDO flag `student-email-post-enabled` which is off by default.

## Links

- [jira ticket](https://codedotorg.atlassian.net/browse/P20-834)
- [design](https://docs.google.com/document/d/1G9nXdae8Gt2Ea93enIJEdjSnF3xB9Rb-W4pCDnt46vo/edit#heading=h.25gc3zi9dtz9)

## Testing story

1. With DCDO disabled, signed up for account
2. With DCDO disabled, signed up via Google OAuth
3. With DCDO enabled, signed up for account
4. With DCDO enabled, signed up via Google OAuth

## Deployment strategy

This will be deployed via a DCDO feature flag

## PR Checklist:

- [x] Code is well-commented

### DIFF
--- a/dashboard/app/controllers/registrations_controller.rb
+++ b/dashboard/app/controllers/registrations_controller.rb
@@ -24,7 +24,13 @@ class RegistrationsController < Devise::RegistrationsController
     if PartialRegistration.in_progress?(session)
       user_params = params[:user] || ActionController::Parameters.new
       user_params[:user_type] ||= session[:default_sign_up_user_type]
-      @user = User.new_with_session(user_params.permit(:user_type), session)
+      if DCDO.get('student-email-post-enabled', false)
+        user_params[:email] ||= params[:email]
+
+        @user = User.new_with_session(user_params.permit(:user_type, :email), session)
+      else
+        @user = User.new_with_session(user_params.permit(:user_type), session)
+      end
     else
       save_default_sign_up_user_type
       SignUpTracking.begin_sign_up_tracking(session, split_test: true)
@@ -53,13 +59,16 @@ class RegistrationsController < Devise::RegistrationsController
     @user = User.new(begin_sign_up_params)
     @user.validate_for_finish_sign_up
     SignUpTracking.log_begin_sign_up(@user, session)
+    is_signup_post_enabled = DCDO.get('student-email-post-enabled', false)
 
     if @user.errors.blank?
       PartialRegistration.persist_attributes(session, @user)
-      redirect_to new_user_registration_path
+      redirect_to new_user_registration_path unless is_signup_post_enabled
     else
-      render 'new' # Re-render form to display validation errors
+      render 'new' unless is_signup_post_enabled # Re-render form to display validation errors
     end
+
+    render 'new' if is_signup_post_enabled
   end
 
   #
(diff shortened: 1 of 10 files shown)


dashboard/app/controllers/registrations_controller.rb
```diff
@@ -53,13 +59,16 @@ def begin_sign_up
...
     SignUpTracking.log_begin_sign_up(@user, session)
+    is_signup_post_enabled = DCDO.get('student-email-post-enabled', false)
 
     if @user.errors.blank?
       PartialRegistration.persist_attributes(session, @user)
-      redirect_to new_user_registration_path
+      redirect_to new_user_registration_path unless is_signup_post_enabled
     else
-      render 'new' # Re-render form to display validation errors
+      render 'new' unless is_signup_post_enabled # Re-render form to display validation errors
```
carl-codeorg: In this else case, we don't render new if `is_signup_post_enabled`, but then won't we still render new on line 71 because it has the opposite condition?

dashboard/app/models/user.rb
```diff
@@ -913,7 +913,10 @@ def oauth_only?
 
   def self.new_with_session(params, session)
     return super unless PartialRegistration.in_progress? session
-    new_from_partial_registration session do |user|
+    new_from_partial_registration(session) do |user|
+      if DCDO.get('student-email-post-enabled', false)
```
carl-codeorg: Have you tested this without the if block? I would have thought that `Services::User.assign_form_params` should automatically set email if it's in the params. But you might have to specifically do a `.permit` or something to get it to work.

dashboard/app/controllers/registrations_controller.rb
stephenliang: If `is_signup_post_enabled` is true, then it will re-render `new` if there are errors via line 71. If there are no errors, it will still render `new` via line 71. 

If `is_signup_post_enabled` is false, then it will re-render new via line 68 if there are errors. If no errors, it will redirect via line 66. Line 71 is skipped

dashboard/app/models/user.rb
stephenliang: You're right! Not sure why I needed to do this, but I'll remove this.

dashboard/app/controllers/registrations_controller.rb
carl-codeorg: In that case, why do we need the else? If the conditional is removed from line 71, it should handle the case where there are errors and `is_signup_post_enabled` is false.

stephenliang: The else is needed to prevent the redirect and render, rails will error if you do both. However I cleaned this up a bit to be a little cleaner, let me know what you think.

carl-codeorg: Oh I didn't catch that, good to know. That looks a lot cleaner with the `and return` instead of the else! It's probably a PEBKAC situation, but I find it hard to read Ruby control flow that includes if/elses with `unless` and `if` keywords inside.

carl-codeorg (approved):

PR #58322, opened 2024-04-29
Refactor the BorderedCallToAction component

kelbyhawn:
Refactors the `BorderedCallToAction` to use new `Typography` and `Button` design system components which makes the component responsive and better for accessibility.
- Had to update a couple unit tests due to the new `Button` component (thank you for the help @TurnerRiley!)
  - I decided to keep these in Enzyme as opposed to refactoring to React Testing Library to limit scope creep
- This component is used in `SetUpSections` and `SetUpCourses` in Storybook if you want to see it in action

**Jira ticket:** [ACQ-1829](https://codedotorg.atlassian.net/browse/ACQ-1829)

----

### Dashed border
Before_Dashed

### Solid
Before_Solid

### Dashed border
After_Bordered

### Solid
After_Solid

## Responsive

https://github.com/code-dot-org/code-dot-org/assets/9256643/fd15c8ea-e6e7-45f9-8928-51fffaad9cbf


apps/src/templates/studioHomepages/BorderedCallToAction.jsx
```diff
@@ -25,82 +25,36 @@ const BorderedCallToAction = ({
...
-        onClick={onClick || (() => navigateToHref(buttonUrl))}
+        onClick={onClick}
         className={buttonClass}
         color={buttonColor}
+        size={'s'}
+        type={buttonType}
         text={buttonText}
-        style={styles.button}
+        href={buttonUrl}
+        useAsLink={!!buttonUrl}
```
hannahbergam, line 41: Oooo interesting... this is the first time I'm realizing the button was designed this way. (This is not on you but I'm surprised we would have a state where we provide an href value but *don't useAsLink)

apps/src/templates/studioHomepages/borderedCallToAction.module.scss
```diff
@@ -0,0 +1,37 @@
+@import 'color';
```
hannahbergam, line 1: I imagine a lot of these are translated from the styles in the old component.. I think the proof is in the screenshots!

hannahbergam (approved):
I trust we're all good with the extra space between the header and text? Of course it seems large to me after staring at the difference between old and new, but now I can't decide😂

apps/src/templates/studioHomepages/BorderedCallToAction.jsx
```diff
@@ -25,82 +25,36 @@ const BorderedCallToAction = ({
...
       />
     </div>
   );
 };
 
-BorderedCallToAction.defaultProps = {
```
bethanyaconnor, line 44: Do all callers set this?

bethanyaconnor (approved):

apps/src/templates/studioHomepages/borderedCallToAction.module.scss
kelbyhawn: Correct! I refactored it a bit too to clean things up and make it responsive.

apps/src/templates/studioHomepages/BorderedCallToAction.jsx
kelbyhawn: What's a caller in this context?

apps/src/templates/studioHomepages/BorderedCallToAction.jsx
kelbyhawn: Oh that's a great point! I just tried to see if it would automatically set `useAsLink` if there's a `buttonUrl` and it does not and gives an error.

kelbyhawn:
> I trust we're all good with the extra space between the header and text? Of course it seems large to me after staring at the difference between old and new, but now I can't decide😂

I went through this experience too! 😅 Decided to leave it as-is since it adheres to the Design System styles and negates the need for override styles.

apps/src/templates/studioHomepages/BorderedCallToAction.jsx
bethanyaconnor: oh sorry, I mean anywhere that uses this component. I'm asking why this was safe to delete

kelbyhawn: Oh! Yep, it's safe to delete — the prop this sets is only used on the old `Button` component, the new `Button` component handles colors more efficiently.

kelbyhawn: I believe the new Button component sets purple as the default color automatically so we don't need to set it again here.

PR #58331, opened 2024-04-30
Add more templates/ folders to eslint import order

lfryemason:
## Warning!!

## PR Checklist:

- [ ] Code is well-commented


apps/src/templates/rubrics/introjs.scss
```diff
@@ -445,4 +445,3 @@ tr.introjs-showElement > th {
   z-index: 1;
   opacity: 0;
 }
-/*# sourceMappingURL=introjs.css.map */
```
Hamms: I've noticed these comments getting automatically removed by the linter in some of my own PRs; I'm not entirely sure why they're being removed because they don't actually appear to violate any linting rules, but I do think we want to preserve them.

Hamms (commented):
One `sourceMappingURL` comment got unintentionally removed, but otherwise this LGTM!

apps/src/templates/rubrics/introjs.scss
lfryemason: Asking in [this](https://codedotorg.slack.com/archives/C051P2V2RN0/p1714507708508309) thread to see if we can resolve this permanently

lfryemason: @Nokondi has suggested we let it be removed. I'm going to keep this PR as is.

Hamms: Are we confident we don't want this comment? Note this isn't just an arbitrary comment; it's intended to enable sourcemapping for this compiled asset. This is the kind of thing where if we remove it, it won't break it an obvious way but may make it difficult down the line to diagnose something else that's broken.

Of course, it could already be broken for all I know :upside_down_face:

Hamms (approved):
LGTM; thanks for taking care of the `sourceMappingURL` line!

PR #58333, opened 2024-04-30
Music: enable Amplitude analytics on standalone projects

sanchitmalhotra126:
## Warning!!

## Description

Enable Amplitude analytics on standalone music projects (rather than /projectbeats)

Screenshot 2024-04-30 at 12 06 14 PM

## Links

https://codedotorg.atlassian.net/browse/LABS-727

## Testing story

Tested with standalone projects and script levels - confirmed that events are only emitted for standalone projects.

### DIFF
--- a/apps/src/music/analytics/AnalyticsReporter.ts
+++ b/apps/src/music/analytics/AnalyticsReporter.ts
@@ -5,13 +5,17 @@ import {
   identify,
   setSessionId,
   flush,
+  setUserId,
 } from '@amplitude/analytics-browser';
 import Lab2Registry from '@cdo/apps/lab2/Lab2Registry';
-import {isDevelopmentEnvironment} from '@cdo/apps/utils';
+import {
+  getEnvironment,
+  isDevelopmentEnvironment,
+  isProductionEnvironment,
+} from '@cdo/apps/utils';
 import {Block} from 'blockly';
-
-const BlockTypes = require('../blockly/blockTypes').BlockTypes;
-const FIELD_SOUNDS_NAME = require('../blockly/constants').FIELD_SOUNDS_NAME;
+import {BlockTypes} from '../blockly/blockTypes';
+import {FIELD_SOUNDS_NAME} from '../blockly/constants';
 
 const API_KEY_ENDPOINT = '/musiclab/analytics_key';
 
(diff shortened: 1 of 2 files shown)


apps/src/music/analytics/AnalyticsReporter.ts
```diff
@@ -171,29 +165,6 @@ export default class AnalyticsReporter {
...
-
-  onInstructionsVisited(page: number) {
-    if (!this.sessionInProgress) {
-      this.log('No session in progress');
-      return;
-    }
-
-    this.currentInstructionsPage = page;
-    this.maxInstructionsSeen = Math.max(this.maxInstructionsSeen, page);
-  }
```
sanchitmalhotra126, line 195: Removing some data we don't log any more

apps/src/music/analytics/AnalyticsReporter.ts
```diff
@@ -267,6 +236,19 @@ export default class AnalyticsReporter {
   }
 
   log(message: string) {
-    console.log(`[AMPLITUDE ANALYTICS EVENT]: ${message}`);
+    console.log(`[MUSIC AMPLITUDE ANALYTICS EVENT]: ${message}`);
+  }
+
+  private formatUserId(userId: number) {
```
sanchitmalhotra126, line 242: Copied from https://github.com/code-dot-org/code-dot-org/blob/f280b95523fb6a3aabbe624250146256a9480cef/apps/src/lib/util/AnalyticsReporter.js#L126 I'd imagine later we might want to consolidate/replace with this metrics reporter, so just copying over for now

apps/src/music/analytics/AnalyticsReporter.ts
```diff
@@ -267,6 +236,19 @@ export default class AnalyticsReporter {
...
+    const userIdString = userId.toString() || 'none';
+    if (!userId) {
+      return userIdString;
+    }
```
thomasoniii: non-blocking, but this feels backwards to me. Would it make more sense as:

```
if (!userId) {
  return 'none'
}
const userIdString = userId.toString()
```

My thinking is that we're explicitly assuming the `userId` exists (and is truthy) on line 243, but then on 244 we're checking to see if it doesn't exist (or, is at least non-truthy) and then bombing out later.

thomasoniii (approved):
another minor suggestion/opinion question, but otherwise lgtm 🚀

apps/src/music/analytics/AnalyticsReporter.ts
sanchitmalhotra126: Oh yep good call - copied this from another class but didn't catch this, thanks

breville (approved):

breville (approved):

PR #58335, opened 2024-04-30
UI changes from School Association BugBash

hannahbergam:
This PR addresses a few styling updates:

1. The zip error is now red and has top padding
2. The dropdowns are full width
3. The inputs are white background and full width
4. The buttons below the school dropdown now use the new button component, and override the styling on hover so they become underlined in that state instead of having a purple background

Screenshot 2024-04-30 at 2 18 55 PM
Screenshot 2024-04-30 at 2 19 08 PM
Screenshot 2024-04-30 at 2 19 15 PM

before:
Screenshot 2024-04-30 at 2 24 10 PM
Screenshot 2024-04-30 at 2 24 01 PM
Screenshot 2024-04-30 at 2 25 35 PM

## Links

- spec: []()
- jira ticket: https://codedotorg.atlassian.net/browse/ACQ-1855
- https://codedotorg.atlassian.net/browse/ACQ-1857
- https://codedotorg.atlassian.net/browse/ACQ-1856

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/templates/SchoolZipSearch.jsx
+++ b/apps/src/templates/SchoolZipSearch.jsx
@@ -97,7 +97,6 @@ export default function SchoolZipSearch({fieldNames, zip, disabled}) {
           <SimpleDropdown
             id="uitest-school-dropdown"
             disabled={disabled}
-            className={style.dropdown}
             name={fieldNames.ncesSchoolId}
             itemGroups={[
               {
--- a/apps/src/templates/SchoolZipSearch.jsx
+++ b/apps/src/templates/SchoolZipSearch.jsx
@@ -6,7 +6,7 @@ import style from './school-association.module.scss';
 import classNames from 'classnames';
 import {SimpleDropdown} from '@cdo/apps/componentLibrary/dropdown';
 import SchoolNameInput from '@cdo/apps/templates/SchoolNameInput';
-import Button from '@cdo/apps/templates/Button';
+import {Button} from '@cdo/apps/componentLibrary/button';
 import analyticsReporter from '@cdo/apps/lib/util/AnalyticsReporter';
 import {EVENTS, PLATFORMS} from '@cdo/apps/lib/util/AnalyticsConstants';
 
@@ -115,7 +115,9 @@ export default function SchoolZipSearch({fieldNames, zip, disabled}) {
           <Button
             text={i18n.noSchoolSetting()}
             disabled={disabled}
-            styleAsText={true}
+            color={'purple'}
+            type={'tertiary'}
+            size={'xs'}
             onClick={e => {
               e.preventDefault();
               setSelectedSchoolNcesId(NO_SCHOOL_SETTING);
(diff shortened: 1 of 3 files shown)


kelbyhawn (approved):
Looks great! 🎉 

I helped with these if someone else wants to take a look.

apps/src/templates/SchoolZipSearch.jsx
```diff
@@ -6,7 +6,7 @@ import style from './school-association.module.scss';
 import classNames from 'classnames';
 import {SimpleDropdown} from '@cdo/apps/componentLibrary/dropdown';
 import SchoolNameInput from '@cdo/apps/templates/SchoolNameInput';
-import Button from '@cdo/apps/templates/Button';
+import {Button} from '@cdo/apps/componentLibrary/button';
```
bethanyaconnor, line 9: yay for using the new component library button!

apps/src/templates/school-association.module.scss
```diff
@@ -9,45 +9,56 @@
...
     margin-top: 0.37rem !important;
     border-radius: 0.25rem;
     border: 1px solid;
-    background-color: inherit;
     text-indent: 0.75rem;
   }
 
+  select {
+    font-weight: 400 !important;
+    width: 100% !important;
```
bethanyaconnor, line 30: Is `!important` needed here in order to override the component library styling? I wonder if it would be better to use classnames to do so?

hannahbergam: Yes unfortunately. And do we have a norm around using classnames? We moved this one out of a classname to standardize across the components included in this stylesheet.

apps/src/templates/school-association.module.scss
```diff
@@ -9,45 +9,56 @@
...
   }
 
   .disabledLabel {
-      color: $light_gray_300 !important;
-    }
-  }
-
-.dropdown {
-  & > div {
-    select {
```
hannahbergam, line 48: @bethanyaconnor it used to be down here in case that helps

apps/src/templates/school-association.module.scss
bethanyaconnor: won't this also apply to other selects on the page if they exist? (they may not on the sign up page but they might on other pages we plan to use this component on)

There is some [somewhat hidden documentation](https://github.com/code-dot-org/code-dot-org/tree/staging/apps/src/componentLibrary/typography#additional-styling--rewriting-existing-styles-of-the-components) on using classnames, though all selects that need this styling would need to have that classname

hannahbergam: Will other components matter given these are nested in the outer classname .outerContainer? And wouldn't the other components have to import this stylesheet? Maybe @kelbyhawn could be a helpful person to bring in here as well.

bethanyaconnor: OH totally missed `.outerContainer` (it was above github's preview 😭)

bethanyaconnor: > And wouldn't the other components have to import this stylesheet?

I'm thinking of the case of adding this to https://code.org/yourschool which has other selects. When we update that page, this stylesheet will be brought in.

If this only applies to selects within `.outerContainer`, I think this is fine

kelbyhawn: I suggest renaming the `.outerContainer` class to something more specific like `.schoolAssociationWrapper` just in case!

hannahbergam: Renamed accordingly! I tested locally to make sure it was all still picked up:)

bethanyaconnor (approved):

PR #58336, opened 2024-04-30
LTI Dynamic Registration

nicklathe:
This PR adds functionality to support the [LTI Dynamic Registration](https://www.imsglobal.org/spec/lti-dr/v1p0) flow. It adds the following:

- `/lti/v1/dynamic_registration` to the `iframe_path_allowlist` (see comment below)
- New routes (GET and POST)
- New LTI::V1::DynamicRestrationController
- New UI page to initiate registration flow and collect admin email
- New method and constants in `Policies::Lti`
- New CacheClient
- Unit tests

Because Canvas requires this flow to take place in an iframe, it is necessary to add the new path to the [iframe_path_allowlist](https://github.com/code-dot-org/code-dot-org/pull/58336/files#diff-19cc5be92c36ff06b63767f0ff922d2b9b7b9b8bebe4eaf38e0f331a14b0b528R70). The only thing this page does is collect an email, and POST to the `lti/v1/dynamic_registration` endpoint. Unlike other cases where we want to "jailbreak" out of the iframe, Canvas requires the entire registration process to happen in the iframe in order to preserve the expected user experience. If desirable, you can read about Canvas' implementation of this dynamic registration spec [in their docs](https://canvas.instructure.com/doc/api/file.registration.html).

**New UI**
Screenshot 2024-05-08 at 2 16 14 PM

## Links

- [LTI Tech Spec, Dynamic Registration Proposal](https://docs.google.com/document/d/1tTwAF40cnn2oUvK2CmRO-a5qYXkCR5rR5gBJS2-zV_U/edit#heading=h.keeofzme42ac)
- [Jira task](https://codedotorg.atlassian.net/browse/P20-883)

## Testing story
```
bundle exec spring testunit ./test/controllers/lti_v1_controller_test.rb

Since there is no EDITOR or BETTER_ERRORS_EDITOR environment variable, using Textmate by default.==================================                                           ] 74% Time: 00:00:07,  ETA: 00:00:03
  55/55: [===================================================================================================================================================================] 100% Time: 00:00:10, Time: 00:00:10

Finished in 10.89758s
55 tests, 136 assertions, 0 failures, 0 errors, 0 skips
```
```
bundle exec spring testunit ./test/lib/policies/lti_test.rb

] 44% Time: 00:00:00,  ETA: 00:00:01
DEPRECATED: Use assert_nil if expecting nil from test/lib/policies/lti_test.rb:136. This will fail in Minitest 6.===============                                              ] 72% Time: 00:00:01,  ETA: 00:00:00
  29/29: [===================================================================================================================================================================] 100% Time: 00:00:01, Time: 00:00:01

Finished in 1.62851s
29 tests, 35 assertions, 0 failures, 0 errors, 0 skips
```

## PR Checklist:

- [ ] Code is well-commented


dashboard/app/controllers/lti_v1_controller.rb
```diff
@@ -407,6 +408,90 @@ def new_integration
...
+    registration_token = params[:registration_token]
+    begin
+      openid_config = JSON.parse(HTTParty.get(openid_configuration_url).body).symbolize_keys
+    rescue => exception
+      Honeybadger.notify(exception, context: {message: 'Error getting LMS openid_configuration'})
+      return render status: :internal_server_error
+    end
+
+    registration_id = SecureRandom.uuid
+    puts registration_id
```
nicklathe: TODO: Remove `puts` statement

dashboard/app/controllers/lti_v1_controller.rb
```diff
@@ -407,6 +408,90 @@ def new_integration
...
+    registration_data = {
+      registration_token: registration_token,
+      registration_endpoint: openid_config[:registration_endpoint],
+      issuer: openid_config[:issuer],
+      registration_id: registration_id,
+    }
+
+    write_cache(registration_id, registration_data, 1.hour)
```
nicklathe: TODO: Add comment on why 1 hour, due to the TTL on the registration token we get from Canvas

dashboard/lib/policies/lti.rb
```diff
@@ -78,6 +78,41 @@ module MessageType
...
+        email: "$Person.email.primary",
+        full_name: "$Person.name.full",
+        given_name: "$Person.name.given",
+        family_name: "$Person.name.family",
+        display_name: "$Person.name.display",
+        section_ids: "$Canvas.course.sectionIds",
+        section_names: "$com.instructure.User.sectionNames"
+      },
+      claims: %w[sub iss name given_name family_name nickname picture email locale],
+      messages: [
```
nicklathe, line 108: TODO: Add more placements, match existing JSON blob placements

dashboard/lib/policies/lti.rb
```diff
@@ -113,6 +148,19 @@ def self.issuer_name(issuer)
...
 
+  def self.find_platform_by_issuer(issuer)
+    LMS_PLATFORMS.each do |_, platform|
+      return platform if platform[:issuer] == issuer
+    end
+    # Return nil if no platform matches the issuer
+    nil
+  end
+
+  # def self.generate_dyanmic_config_payload(admin_email)
```
nicklathe: TODO: Remove this method, not needed anymore.

dashboard/app/controllers/lti_v1_controller.rb
```diff
@@ -407,6 +408,90 @@ def new_integration
...
+        <h1>Hello, World!</h1>
+      </body>
+      </html>
+    HTML
+  end
+
+  # POST /lti/v1/dynamic_registration
+  # Initiate LTI Dynamic Registration Request and create LTI Integration
+  def create_registration
+    admin_email = params[:email]
```
nicklathe: TODO:

1. Consider making these params required and...
2. Wrapping more of this code in an exception handler

dashboard/app/controllers/lti_v1_controller.rb
```diff
@@ -407,6 +408,90 @@ def new_integration
...
+    # Get cached values
+    cache_key = params[:registration_id]
+    registration_data = read_cache(cache_key)
+
+    platform = Policies::Lti.find_platform_by_issuer(registration_data[:issuer])
+    # Create registration request
+    body = Policies::Lti::DYNAMIC_REGISTRATION_CONFIG
```
nicklathe: TODO: Currently I'm serializing this to JSON in the http post method argument, should I do it here instead?

dashboard/app/controllers/lti_v1_controller.rb
```diff
@@ -407,6 +408,90 @@ def new_integration
...
+      registration = JSON.parse(response.body).symbolize_keys
+    rescue => exception
+      Honeybadger.notify(exception, context: {message: 'Error creating dynamic registration'})
+      return render status: :internal_server_error
+    end
+
+    existing_integration = Queries::Lti.get_lti_integration(platform[:issuer], registration[:client_id])
+    if existing_integration.nil?
+      Services::Lti.create_lti_integration(
+        name: 'Test Dynamic Registration',
```
nicklathe: TODO: Change name. Should we prompt for a school/district name on the form, or should we use the `https://canvas.instructure.com/lti/account_name` value from the openid_config url response? This value is `Code.org` in our case. We don't really use this name for anything other than helping with visual identification when we are looking at the DB record.

nicklathe: UPDATE: Ali wants to use the value from `https://canvas.instructure.com/lti/account_name`. Add this to the object we are caching.

dashboard/app/controllers/lti_v1_controller.rb
```diff
@@ -407,6 +408,90 @@ def new_integration
...
+    begin
+      openid_config = JSON.parse(HTTParty.get(openid_configuration_url).body).symbolize_keys
+    rescue => exception
+      Honeybadger.notify(exception, context: {message: 'Error getting LMS openid_configuration'})
+      return render status: :internal_server_error
+    end
+
+    registration_id = SecureRandom.uuid
+    puts registration_id
+    registration_data = {
```
nicklathe: TODO: Add `https://canvas.instructure.com/lti/account_name` value to this, we can call it `lms_account_name` or something. It will be used for the `name` value in the LTI Integration we create later.

dashboard/app/controllers/lti_v1_controller.rb
```diff
@@ -407,6 +408,93 @@ def new_integration
...
+  def create_registration
+    admin_email = params[:email]
+    # Get cached values
+    cache_key = params[:registration_id]
+    registration_data = read_cache(cache_key)
+
+    # Don't proceed if expected registration data doesn't exist
+    return unauthorized_status if registration_data.nil?
+
+    platform = Policies::Lti.find_platform_by_issuer(registration_data[:issuer])
```
nicklathe: TODO: Check for `nil` here, handle case of missing non-existent platform type

dashboard/app/controllers/lti_v1_controller.rb
```diff
@@ -407,6 +408,96 @@ def new_integration
     render template: Policies::Lti.early_access? ? 'lti/v1/integrations/early_access' : 'lti/v1/integrations'
   end
 
+  # GET /lti/v1/dynamic_registration
+  # Initiate LTI Dynamic Registration flow
+  def new_registration
```
artem-vavilov: I recommend creating a separate `Lti::V1::DynamicRegistrationController` to avoid overloading the already bloated `LtiV1Controller`

apps/src/lib/ui/lti/registration/LtiDynamicRegistrationPage.tsx
```diff
@@ -0,0 +1,90 @@
...
+import {Button} from '@cdo/apps/componentLibrary/button';
+import i18n from '@cdo/locale';
+import styles from './styles.module.scss';
+
+interface LtiDynamicRegistrationProps {
+  logoUrl: string;
+  registrationID: string;
+}
+
+export const LtiDynamicRegistrationPage = ({
```
stephenliang, line 13: You should be able to create a RTL unit test for this page and stub out the ajax calls and the `postMessage` window call.

dashboard/app/controllers/lti_v1_controller.rb
nicklathe: Yeah, that's probably a good idea. I remember now that there are some examples where you've done that already. Will take a look 👍

artem-vavilov: You can also check this: https://github.com/code-dot-org/code-dot-org/pull/58464/files#diff-a3ae4dbdd699d7712c2672e40f767aa0a7ff3c5a36ad8adb624d676be890f444

dashboard/lib/policies/lti.rb
```diff
@@ -113,6 +180,14 @@ def self.issuer_name(issuer)
     I18n.t(:lti_v1, scope: [:section, :type])
   end
 
+  def self.find_platform_by_issuer(issuer)
+    LMS_PLATFORMS.each do |_, platform|
```
carl-codeorg: Could potentially shorten this using `find` like:
```
LMS_PLATFORMS.find {|_, platform| platform[:issuer] == issuer}
```

dashboard/app/controllers/lti_v1_controller.rb
```diff
@@ -407,6 +408,96 @@ def new_integration
...
+      openid_configuration_url = params[:openid_configuration]
+      registration_token = params[:registration_token]
+      openid_config = JSON.parse(HTTParty.get(openid_configuration_url).body)
+    rescue => exception
+      message = 'Error getting LMS openid_configuration'
+      Honeybadger.notify(exception, context: {message: message})
+      return render status: :internal_server_error, json: {error: message}
+    end
+    # Create registration ID to pass to be used by the view as a hidden form
+    # field. This ID will be useed to read the registration_data back out from
```
carl-codeorg: Typo: useed -> used

dashboard/app/controllers/lti_v1_controller.rb
```diff
@@ -407,6 +408,96 @@ def new_integration
...
+    }
+
+    begin
+      response = HTTParty.post(registration_data[:registration_endpoint], body: body, headers: headers)
+      registration_response = JSON.parse(response.body).symbolize_keys
+    rescue => exception
+      message = 'Error creating registration'
+      Honeybadger.notify(exception, context: {message: message})
+      return render status: :internal_server_error, json: {error: message}
+    end
```
carl-codeorg: Would it be worth pulling this section into its own function or class (maybe a service or client)? Would keep the size of this route down, and in tests you could just stub the new function instead of having to stub a specific http request.

dashboard/app/controllers/lti_v1_controller.rb
nicklathe: ✅

dashboard/lib/clients/lti_dynamic_registration_client.rb
```diff
@@ -0,0 +1,18 @@
+class LtiDynamicRegistration
```
carl-codeorg: Nice, minor nitpick is to namespace it like:

```
module Lti
  class DynamicRegistration
```

I should have done this with the LTI advantage client but we hadn't started modularizing our LTI classes yet.

nicklathe: Ah got it. Will update this.

dashboard/lib/clients/cache_client.rb
```diff
@@ -0,0 +1,16 @@
+class CacheClient
```
nicklathe, line 1: The `lti_v1_controller` has some private methods to help with reading and writing to the cache. However, those aren't available outside that controller, so I made a client that can be used by the smaller modular controllers. I chose not to update any existing cache code in the `lti_v1_controller` in an effort to keep this PR smaller, however that could be done via a followup PR.

dashboard/app/controllers/lti_v1_controller.rb
nicklathe: ✅ This work has been done.

carl-codeorg (approved):
LGTM

apps/src/lib/ui/lti/registration/LtiDynamicRegistrationPage.tsx
```diff
@@ -0,0 +1,94 @@
...
+      type: 'POST',
+      url: '/lti/v1/dynamic_registration',
+      data: {
+        email: email,
+        registration_id: registrationID,
+      },
+      dataType: 'json',
+      success: () => {
+        if (typeof onSubmit === 'function') {
+          onSubmit();
```
stephenliang: Curious why you moved this up one level, it's still good but just wondering

dashboard/config/routes.rb
```diff
@@ -607,10 +607,16 @@
     match '/lti/v1/sync_course', to: 'lti_v1#sync_course', via: [:get, :post]
     post '/lti/v1/integrations', to: 'lti_v1#create_integration'
     get '/lti/v1/integrations', to: 'lti_v1#new_integration'
+    # post '/lti/v1/dynamic_registration', to: 'lti_v1#create_registration'
+    # get '/lti/v1/dynamic_registration', to: 'lti_v1#new_registration'
```
stephenliang: Could we remove these?

apps/src/lib/ui/lti/registration/LtiDynamicRegistrationPage.tsx
nicklathe: I was following another example in our code, thinking it may be best practice to have a method like this passed in via the component props, as it seems to make it easier to test/mock. Definitely not a React pro, and heavily relied on other examples in our code.

dashboard/config/routes.rb
nicklathe: Ah yes, good catch.

apps/src/lib/ui/lti/registration/LtiDynamicRegistrationPage.tsx
stephenliang: Usually I wouldn't recommend moving this one layer up as the parent component doesn't really use anything from this `onSubmit`. Feel free to move it back or merge it as is though.

stephenliang (approved):

apps/src/lib/ui/lti/registration/LtiDynamicRegistrationPage.tsx
nicklathe: Roger that (and agreed), I moved it back.

PR #54038, opened 2023-10-02
Allows parent permission request from account email if parent created account

wilkie:
This updates the `LockoutPanel` component to take, instead of a `studentEmail`, a `disallowEmail` property. This is set to the account email (as a hash since students' emails are not stored plaintext) _unless_ the parent created the account during registration via the `I'm creating this account on behalf of a child` checkbox. When this happens the `parent_email` for the `User` will be plaintext and match the account email which is encoded only in its hashed form as `user.hashed_email`. Therefore, I added a `parent_created?` method in the `User` to return `true` when the hash of the `parent_email`, if exists, matches the `hashed_email` of the account.

When this is `true`, the `/lockout` controller action does not set the `disallowEmail` data such that the `LockoutPanel` component does not have it. Thus, the component will not reject the account email address and will not disable the submit button.

I have also added a UI test that will create such a parent-created student account. Using this account, the test navigates the lockout panel and checks that the button to submit is enabled when the account email is entered. This is in contrast to another UI test that is simply a student created account, where that test successfully tests that the button is disabled in the similar situation of entering the account email.

To that end, a few little UI test steps were added or refined to make that possible. I could not simply use the "type in email" step since that bypasses the component's check for input events to enable/disable the button. We have an existing pattern of using "press keys" steps for actually simulating input, so I added one for the email in the "account steps" definitions. I augmented the prior tests to use that when necessary, which actually makes them more accurate tests.

Took the time to remove the `TODO` sections of the UI tests since they are now resolvable.

## Links

- jira ticket: [https://codedotorg.atlassian.net/browse/P20-435](P20-435)
- jira (support) ticket: [https://codedotorg.atlassian.net/browse/P20-419](P20-419)

## Testing story

Manual testing.

Added UI test.

## PR Checklist:

- [ ] Tests provide adequate coverage
- [ ] Privacy and Security impacts have been assessed
- [ ] Code is well-commented
- [ ] New features are translatable or updates will not break translations
- [ ] Relevant documentation has been added or updated
- [ ] User impact is well-understood and desirable
- [ ] Pull Request is labeled appropriately
- [ ] Follow-up work items (including potential tech debt) are tracked and linked

PR #58337, opened 2024-05-01
P20-57: Add pre-lockdown parent permission modal

artem-vavilov:
## Summary
The modal will appear daily for students who are not compliant with the Child Account Policy until parental permission is requested at least once, and while the CPA experience is `cpa_all_user_lockout_warning`.

**Note:** Use the parameter `?show_parental_permission_modal` to force the display of the modal.

### Update parent permission request
update

## Links
- JIRA ticket: [P20-57](https://codedotorg.atlassian.net/browse/P20-57)
- Feature flag task: [P20-880](https://codedotorg.atlassian.net/browse/P20-880)
- Related PR: https://github.com/code-dot-org/code-dot-org/pull/54038


apps/src/redux/parentalPermissionRequestReducer.ts
```diff
@@ -0,0 +1,182 @@
...
+      return state;
+  }
+}
+
+export const fetchPendingPermissionRequest = (
+  dispatch: Dispatch<ParentalPermissionRequestAction>
+) => {
+  dispatch({type: FETCH_PENDING_PERMISSION_REQUEST_PERFORM});
+
+  fetch(FETCH_PENDING_PERMISSION_REQUEST_URL, {
```
carl-codeorg: nit: could probably clean up this function and the one below it by using async/await instead of promise chaining.

artem-vavilov: Are you suggesting we use async/await functions instead of more common for React the Redux/Reducer approach?

carl-codeorg: Is the Redux/Reducer approach incompatible with async functions?

carl-codeorg: My comment was just regarding convention - when possible, favor async/await over promise chaining or old school callbacks. If it isn't possible because of some nuance of Redux, feel free to ignore

artem-vavilov: Oh, I see... https://github.com/code-dot-org/code-dot-org/pull/58337/commits/8451021fb833b9732259a9e80c89ac1b20f2e721

apps/src/templates/policy_compliance/ParentalPermissionModal/index.story.tsx
mgc1194, line 1: I liked placing the new components under a new directory.
I am also working on some components, however, I placed them under `apps/src/templates/ChildAccountPolicy`. 
I'd like to hear Dayne's opinion on what directory name we should use, and stay consistent.
I don't mind changing my PR and I'd love to hear more about the pattern used here, with `index.tsx` and `_modal.js`

apps/i18n/common/en_us.json
```diff
@@ -1986,6 +1986,13 @@
...
+  "policyCompliance_parentalPermissionModal_newRequestForm_title": "Heads up! Your account may soon be locked.",
+  "policyCompliance_parentalPermissionModal_newRequestForm_text1": "Starting on {lockoutDate}, **you will need your parent or guardian’s permission to keep your personal login** with Code.org.",
+  "policyCompliance_parentalPermissionModal_newRequestForm_text2": "To avoid having your account locked, please share your parent or guardian’s email address so they can grant you permission.",
+  "policyCompliance_parentalPermissionModal_updateRequestForm_title": "Thanks! We’ve contacted your parent or guardian.",
+  "policyCompliance_parentalPermissionModal_updateRequestForm_text1": "Didn’t receive anything? Update your parent or guardian’s email below or send another request. Note, your account will be locked if we do not receive your parent/guardian’s permission by {lockoutDate}.",
+  "policyCompliance_parentalPermissionModal_updateRequestForm_text2": "You can also provide this email address at any time from [your profile]({profileUrl}).",
+  "policyCompliance_parentalPermissionModal_lastEmailSentAt": "Last email sent at {sendingTime}",
```
mgc1194, line 2005: I've already created some strings under the namespace `childAccountPolicy_`.
Would it be reasonable to stick to that namespace? If you feel strongly about using `policyCompliance_` I'm okey switching the strings I've added, but eitherway we should stay consistent.
https://github.com/code-dot-org/code-dot-org/blob/9cb8ea19ea80d65a38a253c8c0b7bcebc4f48493/apps/i18n/common/en_us.json#L412-L413

apps/src/templates/sessions/LockoutPanel.jsx
```diff
@@ -6,8 +6,8 @@ import {getStore} from '../../redux';
 import {isEmail} from '@cdo/apps/util/formatValidation';
 import cookies from 'js-cookie';
 import * as color from '../../util/color';
-import headerImage from './images/lockout_penguin.png';
-import headerThanksImage from './images/dancing_penguin.png';
+import headerImage from '@cdo/static/common_images/penguin/yelling.png';
+import headerThanksImage from '@cdo/static/common_images/penguin/dancing.png';
```
mgc1194, line 10: I'm very pro this change.

apps/src/util/currentLocale.ts
```diff
@@ -0,0 +1,4 @@
+import {get} from 'js-cookie';
+import {DefaultLocale} from '@cdo/generated-scripts/sharedConstants';
+
+export default () => get('language_') || DefaultLocale;
```
mgc1194, line 4: Does this mean we are not surfacing the locale anywhere else in the front end?
This will also allow tacking client-side events with language data.

dashboard/config/routes.rb
```diff
@@ -1139,10 +1139,11 @@
...
     # Policy Compliance
-    get '/policy_compliance/child_account_consent/', to:
-      'policy_compliance#child_account_consent'
-    post '/policy_compliance/child_account_consent/', to:
-      'policy_compliance#child_account_consent_request'
+    namespace :policy_compliance do
+      get :pending_permission_request
+      get :child_account_consent
+      post :child_account_consent, action: :child_account_consent_request
+    end
```
mgc1194, line 1150: I see why you have been using the `policy_compliance` namespace. I think the Child Account Policy is definitely a type of policy compliance, so it is FERPA or GDPR.
I'm not sure what would be the best term to use across all components @daynew.

dashboard/test/lib/test_cpa.rb
```diff
@@ -42,20 +43,34 @@ def test_cpa_experience_before_new_user_lockout
...
     result = Cpa.cpa_experience(@request, current_time: current_time)
     assert_equal Cpa::NEW_USER_LOCKOUT, result
   end
 
+  def test_cpa_experience_after_all_user_lockout_warning
+    current_time = DateTime.parse('2023-01-02T00:00:01Z')
+    schedule = {
+      Cpa::NEW_USER_LOCKOUT => '2023-01-01T00:00:00Z',
+      Cpa::ALL_USER_LOCKOUT_WARNING => '2023-01-02T00:00:00Z',
+      Cpa::ALL_USER_LOCKOUT => '2023-01-03T00:00:00Z'
```
mgc1194, line 62: Just a note for all reviewers, I'm not sure what the scheduled dates are, but we all need to make sure the dates are accurate. I'm seeing 2023 and I want to make sure it is not supposed to date in 2024.

dashboard/lib/policies/child_account.rb
```diff
@@ -38,6 +39,12 @@ module ComplianceState
   # when a student may no longer be monitoring the "parent's email."
   PERMISSION_GRANTED_MAIL_DELAY = 24.hours
 
+  # The maximum number of daily requests a student can send to their parent.
+  MAX_STUDENT_DAILY_PARENT_PERMISSION_REQUESTS = 3
+
+  # The maximum number of times a student can resend a request to a parent.
+  MAX_PARENT_PERMISSION_RESENDS = 3
+
```
mgc1194, line 47: Cool! I was wondering if there was something preventing students from spamming with permission requests.

dashboard/lib/cpa.rb
```diff
@@ -3,12 +3,14 @@
 # Support for the Colorado Privacy Act (CPA) compliance.
 module Cpa
   NEW_USER_LOCKOUT = 'cpa_new_user_lockout'
+  ALL_USER_LOCKOUT_WARNING = 'cpa_all_user_lockout_warning'
   ALL_USER_LOCKOUT = 'cpa_all_user_lockout'
 
   # There are three phases for the Colorado Privacy Act:
```
mgc1194: I would update this sentence to `four phases`. I got confused with the phases at first.

dashboard/config/locales/en.yml
```diff
@@ -1481,3 +1481,11 @@ en:
...
     alt_text_no_name: "Certificate for completion of %{course_name}"
     one_hour_of_code: "One Hour of Code"
+  child_account:
+    parental_permission_request:
+      errors:
+        invalid_parent_email: The parent/guardian's email address is invalid
+        already_granted: Parental permission has already been granted
+        parent_email_is_own: A request cannot be sent to your own email address
+        resend_limit_reached: The limit for resend requests has been reached
+        daily_limit_reached: The daily request limit has been reached
```
mgc1194: Could we have all these new strings ordered alphabetically, including the parent keys? It would make them more discoverable. I have the feeling that we already have some strings in this file for CPA, but it is such a massive file is hard to know what already exists.

mgc1194 (commented):
I left a few comments I'd like to address before giving approval.
I addressed everything I'm familiar with but this is a big PR with a lot of important changes and optimizations, so I would still want some more senior engineer to review the changes.

wilkie (approved):
Took me a good while to review as it is touching a lot of different things, so apologies.

Looks really really good. I like the adoption of the Forms namespace and I like the cleanup around the code that is dealing with those `STATE_POLICY` constants.

The modal itself looks very well-engineered. Good work!

dashboard/test/lib/test_cpa.rb
wilkie: These are in the tests and they just need to be more or less in an order. This is the proper order. Seems good to have them be a set of dates that is easy to think about and understand quickly.

apps/src/templates/sessions/LockoutPanel.jsx
wilkie: It makes sense to have the images be near the components that they are used uniquely within and then update them when the images are used in a second container by moving them to the common path.

I am considering namespacing this path so is at least slightly clearer what the context of these images are, but maybe that's not too necessary. A common path just becomes a swamp of random images... which might be nice to avoid. Not a very important problem, though.

apps/src/templates/policy_compliance/ParentalPermissionModal/index.story.tsx
wilkie: It seems pretty reasonable to place CPA components within that existing namespace. The `policy_compliance` would be named after the controller that supports it. Much like how there is a `sessions` path and a `rubrics` path that similarly group components that are backed by those controllers. Hmm... I think all three of these directories are entirely my fault... I just like congruence between controller namespaces and view/component namespaces when possible.

I applied my own opinion there because ultimately we have no actual convention. (nor is there a perfect answer anyway) Seems like another strategy that teacher tools is employing is to namespace them by project feature. (`sectionProgressV2`) That also makes sense as a logical way to handle the cognitive load of understanding the interactions between different modules and components... especially in cases that might require a coördinated effort across different systems.

And then there are just honestly way too many that just exist in the templates root. Then again, I'm guilty of putting something in the `src` root rather recently. heh.

apps/i18n/common/en_us.json
wilkie: I think if your `ChildAccountPolicy` component were in the `policy_compliance` path, a good name for the string key would then be `policyCompliance_childAccountPolicy_learnMore`, etc. I generally didn't use an `_` when I would do this... that seems nicer though.

That said, I don't think I care too much that the actual name of the component is there... just the major grouping. These strings end up moving around as things are updated (or reused elsewhere) that they shouldn't be too tied to the component name, in my opinion, and be more focused on their relationship to their function. So, for the `LockoutPanel`, which I put within the `sessions` grouping because at the time it was supported by the sessions controller, I think I made strings like `sessionLockoutNewAccountHeader`, etc. So, I expect this to be used by stuff within the `sessions` path, so `session`... they have to do with the "`Lockout`" function... and then the general key.

If I was redesigning everything from scratch, I'd probably want to define the strings in the components themselves and then pull them out instead of trying to define every string in one file. Haha what can you do though.

apps/src/util/currentLocale.ts
wilkie: We do in the homepage/script data and `appOptions.localeCode` often via the `localesRedux` which this kind of replaces by looking at the `language_` cookie instead of the locale set within the rendered page. It might be nice to update the redux instead to use the default locale on its initial state (I think it is hard-coded there) and then update the redux to `setLocaleCode` to the cookie value.

The one complication that I'm not fully thinking through is that the locale on the ruby side is in the form `en-US` and on the JS side it is `en_us`. Not sure how much that matters in different situations.

artem-vavilov: `window.appOptions` are set only on lab pages:

apps/src/templates/policy_compliance/ParentalPermissionModal/index.story.tsx
mgc1194: I will be taking these considerations in my PRs :) Thank you Willkie for al the context.

apps/src/templates/policy_compliance/ParentalPermissionModal/index.tsx
```diff
@@ -0,0 +1,354 @@
...
+  );
+
+  const [
+    {action, isLoading, error, parentalPermissionRequest},
+    parentalPermissionRequestDispatch,
+  ] = useReducer(parentalPermissionRequestReducer, {isLoading: false});
+  const prevParentalPermissionRequest = usePrevious(parentalPermissionRequest);
+
+  const reportEvent = (eventName: string, payload: object = {}) => {
+    analyticsReporter.sendEvent(eventName, payload, PLATFORMS.STATSIG);
```
daynew: Did @alivira request we send these events to Statsig?

dashboard/config/routes.rb
carl-codeorg: One option could be doing a sub-namespace like:

```
namespace :policy_compliance do
  namespace :child_accounts
    get :pending_permission_request
    get :consent
    post :consent, action: :consent_request
  end
  namespace :gdpr
    get :information_request
    post :deletion_request
  end
end
```

But since we don't actually seem to have any GDPR or other complaince-related routes, it might be YAGNI.

carl-codeorg (approved):
:shipit:

dashboard/lib/cpa.rb
```diff
@@ -35,13 +37,17 @@ def self.cpa_experience(current_request, current_time: DateTime.now)
     # Verify the schedule is well defined.
     return nil unless schedule
     return nil unless schedule[Cpa::NEW_USER_LOCKOUT]
+    return nil unless schedule[Cpa::ALL_USER_LOCKOUT_WARNING]
```
daynew, line 41: We need to make sure this configuration value is in Prod DCDO before it is DTP'd, otherwise CPA will be disabled for everyone, right?

Should I add it now?
```
{"cpa_new_user_lockout":"2023-07-05T23:15:00+00:00","cpa_all_user_lockout_warning":"2024-07-01T00:00:00MST","cpa_all_user_lockout":"2024-07-01T00:00:00MST"}
```

dashboard/lib/cpa.rb
```diff
@@ -3,12 +3,14 @@
...
   NEW_USER_LOCKOUT = 'cpa_new_user_lockout'
+  ALL_USER_LOCKOUT_WARNING = 'cpa_all_user_lockout_warning'
   ALL_USER_LOCKOUT = 'cpa_all_user_lockout'
 
-  # There are three phases for the Colorado Privacy Act:
+  # There are four phases for the Colorado Privacy Act:
   # 1. Nothing - nil
   # 2. New User Accounts must be compliant - 'cpa_new_user_lockout'
-  # 3. All User Accounts must be compliant - 'cpa_all_user_lockout'
+  # 3. Existing accounts warned of lockout - 'cpa_all_user_lockout_warning'
```
daynew, line 12: There is an "example" cpa_schedule on line 24, could you add this there too?

daynew (approved):
This was a lot of work, thank you! Having all the tests written really helped me be confident it works.

dashboard/lib/cpa.rb
artem-vavilov: @daynew correct, and thank you!

apps/src/templates/policy_compliance/ParentalPermissionModal/index.tsx
artem-vavilov: You are right, it should be Amplitude only

daynew:
Could you also update [CPA README](https://docs.google.com/document/d/1cxE_FeAR7I5mXepCaBr4EFVhTW9KBWWygjE0cg_pM08/edit#heading=h.czv2xbkhqc4v) with any query string params and DCDO configs you added?

dashboard/lib/cpa.rb
artem-vavilov: https://github.com/code-dot-org/code-dot-org/pull/58337/commits/6b83cf15a8babdfa2d6996ac99b49fbbcb570518

artem-vavilov: https://github.com/code-dot-org/code-dot-org/blob/6b83cf15a8babdfa2d6996ac99b49fbbcb570518/dashboard/lib/cpa.rb#L24-L29

apps/src/templates/policy_compliance/ParentalPermissionModal/index.tsx
artem-vavilov: https://github.com/code-dot-org/code-dot-org/blob/6b83cf15a8babdfa2d6996ac99b49fbbcb570518/apps/src/templates/policy_compliance/ParentalPermissionModal/index.tsx#L58-L60

dashboard/lib/cpa.rb
artem-vavilov: Correct, and thank you for reminding

PR #58338, opened 2024-05-01
AI Chat: pass selected model ID to back end

sanchitmalhotra126:
## Warning!!

## Description

This hooks up the selected model ID in the front end to the endpoint we actually invoke on the back end, to target the desired model. I updated our JSON file with our three existing endpoints and some placeholder information, but we'll want to do a formal pass on the overview and training data for each.

Demo (just to showcase the ability to target different models - we're aware that the Mental Health Mistral model produces some less than ideal responses, but will address updating the available model options separately)

https://github.com/code-dot-org/code-dot-org/assets/85528507/91cff21f-04c6-460a-b50a-7bbcf246546c

## Links

https://codedotorg.atlassian.net/browse/LABS-669
Also @bencodeorg to your question in the JIRA, looks like we are logging the selected model ID to the DB already (screenshot of my local DB below):

Screenshot 2024-04-30 at 5 42 07 PM

## Testing story

Tested with our existing endpoints.


bencodeorg (approved):
Nice!

apps/src/lab2/levelEditors/aichatSettings/EditAichatSettings.tsx
```diff
@@ -46,8 +47,14 @@ function sanitizeSettings(settings: LevelAichatSettings) {
     DEFAULT_VISIBILITIES
   );
 
+  // Ensure that only valid model IDs are included.
+  const filteredModelIds = (settings.availableModelIds || []).filter(id =>
+    modelDescriptions.map(model => model.id).includes(id)
+  );
```
thomasoniii: non-blocking, but I think this'd be a little better with a `some` call.

```
const filteredModelIds = (settings.availableModelIds || []).filter(id =>
  modelDescriptions.some(model => model.id === id)
);
```

It's O(n^2) instead of O(n^3) (not that it really matters with such small loops), and I lean towards it being a little easier to read.

OTOH, I also acknowledge it's not _that_ much easier to read, and maybe it loses something in legibility going to a `some` call instead of a more clear sounding `includes` call. 🤷 

I dunno. Dealer's choice.

It could also be an external set:
```
const validModelIds = new Set(modelDescriptions.map(model => model.id));
const filteredModelIds = (settings.availableModelIds || []).filter(id => validModelIds.includes(id) );
```

That's clearer and faster (O(2n)), but it's also using an ugly external variable for just one thing.

Anyway, I'm just rambling.

thomasoniii (approved):
minor non-blocking suggestion, but otherwise lgtm 🚀

apps/src/lab2/levelEditors/aichatSettings/EditAichatSettings.tsx
sanchitmalhotra126: Ah yep, good call, I'll update this - might just go with `some` since it's a relatively small set

dashboard/config/levels/custom/aichat/AI Chat Test Level - 2.level
```diff
@@ -2,7 +2,7 @@
   <config><![CDATA[{
```
fisher-alice, line 2: Thanks for updating these!

fisher-alice (approved):
🎉

PR #58340, opened 2024-05-01
[AI Tutor] CT-534: Add copy code button

ebeastlake:
This PR:
- Adds a `CodeBlockWithCopy` component, which renders its children in a code block but adds a header with the ability to copy the text,
- Modifies the `SafeMarkdown` component to add an optional parameter, `includeCodeCopyButton`. When `includeCodeCopyButton` is true (only for AI Tutor, so far), `<pre>` blocks will render using the (new) `CodeBlockWithCopy` component.

Here's a quick demo:

https://github.com/code-dot-org/code-dot-org/assets/26844240/09918f59-a7bd-44f5-9ebc-7429357f4ba4

## Links

Jira ticket: https://codedotorg.atlassian.net/browse/CT-534

## Testing story

I also manually tested other scenarios to make sure non-code blocks are doing something reasonable too.

Screenshot 2024-04-30 at 10 13 14 PM
Screenshot 2024-04-30 at 10 13 40 PM
Screenshot 2024-04-30 at 10 14 58 PM

## PR Checklist:

- [ ] Code is well-commented


apps/src/templates/SafeMarkdown.jsx
```diff
@@ -29,23 +30,58 @@ class SafeMarkdown extends React.Component {
...
-    const rendered = parser.processSync(this.props.markdown).contents;
+  getParser() {
+    const components = {
+      // Use React component wrappers for Blockly XML elements to prevent
+      // React from warning us about invalid components.
+      ...blocklyComponentWrappers,
+      ...(this.props.includeCodeCopyButton ? {pre: CodeBlockWithCopy} : {}),
+    };
+
+    const parser = Parser.create()
```
ebeastlake, line 48: This logic matches what was stored in `markdownToReact` and `markdownToReactExternalLinks` before, but is just extracted into a `getParser` function that includes the `CodeBlockWithCopy` binding in the `components` if `includeCodeCopyButton` is true.

ebeastlake:
@breville @Hamms What do you all think of this high-level approach to modifying `SafeMarkdown` to be able to add a copy button to "blocks"? I can [add unit tests](https://github.com/code-dot-org/code-dot-org/blob/b9dfd79a7c3c7c80642674ac6e7ec15c3fe41a51/apps/test/unit/templates/SafeMarkdownTest.js#L4) for the new functionality, but wanted to check the high-level approach first.

Hamms:
In general, it would be best for our components to focus on doing just one thing; in this case, rendering markdown in a way that's safe to do for untrusted sources. Even the existing `openExternalLinksInNewTab` functionality is outside the scope of what this component should be doing, so I'm pretty hesitant to add yet more domain-specific exceptions and extra cases to support to this component.

Instead, what do you think about implementing this as a Remark plugin? Consider for example our existing `expandableImages` functionality, which allows you to create images that can clicked on to view in a modal, or the new `clickableText` plugin that Brendan added recently.

ebeastlake:
Thanks for the feedback @Hamms! `outside the scope of what this component should be doing` was how I was feeling about this addition to SafeMarkdown and the reason I asked. Thanks for the pointer to Remark plugins! Closing this PR for now.

ebeastlake:
See back and forth with Elijah, above.

PR #58341, opened 2024-05-01
Music: miniplayer image & text updates

breville:
The miniplayer will no longer show pack images for now, and will instead show a musical note image with a background radial blur, with its color provided by an optional per-pack `color` field in the library JSON, which looks like this:

```
"color": "#EF5A00",
```

We also show the pack's name and artist.

Screenshot 2024-05-01 at 2 41 41 PM


apps/src/music/player/MusicLibrary.ts
```diff
@@ -299,13 +305,20 @@ export interface SoundData {
...
+}
+
 export interface SoundFolder {
   name: string;
   artist?: string;
   id: string;
   type?: SoundFolderType;
   path: string;
   imageSrc: string;
+  imageRgb?: RGB;
```
sanchitmalhotra126: Minor: authoring wise, would be easier to just store a hex code?

apps/src/music/views/MiniMusicPlayer.tsx
```diff
@@ -14,6 +14,8 @@ import Lab2Registry from '../../lab2/Lab2Registry';
 import moduleStyles from './MiniMusicPlayer.module.scss';
 import FontAwesomeV6Icon from '@cdo/apps/componentLibrary/fontAwesomeV6Icon/FontAwesomeV6Icon';
 
+const noteImage = require(`@cdo/static/music/music-note.png`);
```
sanchitmalhotra126: Should be able to use `import` here

apps/src/music/views/MiniMusicPlayer.tsx
```diff
@@ -135,7 +160,15 @@ const MiniPlayerView: React.FunctionComponent<MiniPlayerViewProps> = ({
               />
             </div>
 
-            <div className={moduleStyles.name}>{project.name}</div>
+            <div className={moduleStyles.body}>
+              <div className={moduleStyles.name}>{project.name}</div>
+              {project.labConfig?.music?.packId && (
+                <div className={moduleStyles.details}>
+                  {getPackDetails(project.labConfig.music.packId)?.name} &bull;{' '}
+                  {getPackDetails(project.labConfig.music.packId)?.artist}
```
sanchitmalhotra126: nit: pull out `getPackDetails(project...)` into a local var? (also maybe `project.labConfig.music.packId` since it's used in many other places)

sanchitmalhotra126 (approved):
Nice solution!

apps/src/music/views/MiniMusicPlayer.tsx
```diff
@@ -102,6 +104,26 @@ const MiniPlayerView: React.FunctionComponent<MiniPlayerViewProps> = ({
...
   }
 
+  const getPackImageBackgroundStyle = (packId: string) => {
+    const rgb = MusicLibrary.getInstance()?.getPackImageRgb(packId);
+
+    if (!rgb) {
+      return undefined;
+    }
+
+    return `radial-gradient(rgb(${rgb.r}, ${rgb.g}, ${rgb.b}), rgb(0,0,0))`;
```
thomasoniii: Non-blocking, and this is fine as is, but...

part of me feels like this should be broken apart into at least two methods - the intent feels a little buried. So we've got `getPackImageBackgroundStyle`, but that's always _only_ going to return a radial gradient, and it's also always _only_ going to return a radial gradient that goes to black.

I feel like it'd be better if we had a separate `getRadialGradient` function somewhere:

```
function getRadialGradient(from : RGB, to : RGB) {
  return `radial-gradient(rgb(${from.r}, ${from.g}, ${from.b}), rgb(${to.r}, ${to.g}, ${to.b}))`;
}
```

and then the return in this function is:
```
return getRadialGradient(rgb, {r : 0, g : 0, b : 0})
```

It's a little overengineered in this case, but it's still separating out the functionality - so we have the parent function that gets whatever the background style is from a `packId`, and then a helper function that gets gradients given `RGB` input objects.

Anyway, again, nothing urgent.

thomasoniii (approved):
one minor suggestion I waffled about, and otherwise +1 to everything @sanchitmalhotra126 said in his review.

beyond those, lgtm as well 🚀

apps/src/music/player/MusicLibrary.ts
fisher-alice: +1

fisher-alice (approved):
Cool solution!

apps/src/music/views/MiniMusicPlayer.tsx
breville: I like it.  Done.

apps/src/music/views/MiniMusicPlayer.tsx
breville: Done.

apps/src/music/views/MiniMusicPlayer.tsx
breville: Oh, great point.  Had somehow forgotten we get to create new variables inside these `.map` handlers.

apps/src/music/player/MusicLibrary.ts
breville: You're right.  Changed to that.

PR #53351, opened 2023-08-17
Disable linking of personal login accounts for some students under CPA

carl-codeorg:
U13 Colorado student accounts created after the start of CPA (July 2023) are locked out of Code.org until they get parent permission. U13 CO student accounts that existed prior to July 2023 are not affected by the global CPA lockout, but will be starting in 2024. We also don't have state location information for these accounts, because they were created prior to the addition of the US State field.

These pre-7/23 accounts are the subject of this PR. Even though they're grandfathered in and can access their Code.org accounts without parent permission, we can't allow them to link "personal accounts" (e.g. Google, Microsoft, and Facebook logins) because this counts as collection of new personal data under the CPA.

To determine whether a student is in Colorado, this PR adds a query that looks at the state of the school that their teacher is associated with. When a student visits the Account Settings page, their `us_state` is backfilled if we can get it from their teacher. Eventually we will need to populate all students' `us_state` to comply with the second phase of CPA in 2024, but for now the backfill is only in the account settings page. This reduces the impact on page load time that would come with checking every user on login, or forcing users to enter state info in a pop-up/modal.

This adds:
- A controller for the account settings /users/edit page, which handles the logic for enabling/disabling personal account linking, as well as backfilling us_state.
- A new component LockoutLinkedAccounts, based heavily off of LockoutPanel. It is used to send parent permission requests just like the LockoutPanel, but is different because it lives on the Account Settings page and exists to get permission for unlocking the personal account linking buttons.

## Links

[JIRA ticket](https://codedotorg.atlassian.net/browse/P20-62)

[Figma Mockups](https://www.figma.com/file/wR0Myx8T2dHMaeeb1GAabb/CPA-Flows?type=design&node-id=801-3178&mode=design) - See step 3: connect a personal login flow

## PR Checklist:

- [ ] Tests provide adequate coverage
- [ ] Privacy and Security impacts have been assessed
- [ ] Code is well-commented
- [ ] New features are translatable or updates will not break translations
- [ ] Relevant documentation has been added or updated
- [ ] User impact is well-understood and desirable
- [ ] Pull Request is labeled appropriately
- [ ] Follow-up work items (including potential tech debt) are tracked and linked

PR #58343, opened 2024-05-01
CPA Account Settings Locks and Polish

wilkie:
Carl's [prior PR](https://github.com/code-dot-org/code-dot-org/pull/53351) did already add the section to perform the parental permission on the Account Settings page. This PR adds the remaining locks over sections of the Account Settings page for partially locked out legacy accounts. This specifically adds:

* Better visualization of the locked out SSO/third-party auth connections.
* Some accessible information about why those connections are disabled.
* Locking out the personal login section.
* Some end-to-end UI tests that click through two scenarios for such accounts.

Some of the challenges with the approach have to do with the connection management being done as a React component but the form to add a personal login still exists as a legacy form within the haml of the account settings page... they do not reasonably share the same asset pipeline so the lock image is redundantly copied to two places until the personal login portion is made a component. Doing that was out of the scope.

## Links

- jira ticket: [P20-59](https://codedotorg.atlassian.net/browse/P20-59) (includes a link to the Figma design mockups)

## Testing story

Updates the unit tests for the ManageLinkedAccounts component and the unit tests for the registrations controller.

Adds a few UI tests and some Cucumber steps and test controller actions that aid in testing, end-to-end, the CPA process from the user's perspective.

## Deployment strategy

This will also be hidden behind the `cpa-partial-lockout` and the general CPA experiment values.

## PR Checklist:

- [ ] Code is well-commented


apps/src/lib/ui/accounts/ManageLinkedAccounts.jsx
```diff
@@ -71,6 +75,14 @@ class ManageLinkedAccounts extends React.Component {
     return true;
   };
 
+  shouldDisableConnectButton = authOption => {
+    return (
+      PERSONAL_LOGIN_TYPES.includes(authOption.credentialType) &&
+      !this.formatEmail(authOption) &&
```
carl-codeorg: Why is `!this.formatEmail(authOption)` included in the conditions for disabling the buttons?

apps/src/lib/ui/accounts/ManageLinkedAccounts.jsx
```diff
@@ -91,6 +103,13 @@ class ManageLinkedAccounts extends React.Component {
     }
   };
 
+  connectDisabledStatus = authOption => {
```
carl-codeorg, line 111: Is it possible to combine `connectDisabledStatus` and `disconnectDisabledStatus` into one `getDisabledStatus` function? In my mind, it seems like it'd be cleaner to have one function, and more consistent with the `OauthConnection` having one `disabledStatus` prop and one function that gets the tooltip regardless of whether it's a connect or disconnect button. But I haven't thought through all the possible implications.

apps/src/lib/ui/accounts/ManageLinkedAccounts.jsx
```diff
@@ -275,21 +333,18 @@ class OauthConnection extends React.Component {
                 type="submit"
                 style={styles.button}
                 text={buttonText}
-                disabled={
-                  !!disconnectDisabledMessage ||
-                  (!isConnected && this.shouldDisableConnectButton())
-                }
+                disabled={!!disabledMessage}
```
carl-codeorg, line 343: Much cleaner!

apps/src/lib/ui/accounts/ManageLinkedAccounts.jsx
```diff
@@ -154,26 +195,51 @@ class ManageLinkedAccounts extends React.Component {
...
-                  }
-                  error={option.error}
-                  personalAccountLinkingEnabled={
-                    this.props.personalAccountLinkingEnabled
-                  }
-                />
-              );
-            })}
+            {/* The options that are always available. */}
+            {this.formatAuthOptions()
```
carl-codeorg: Since we already called `this.formatAuthOptions()` and performed a similar filter when we declared `lockedOptions`, is it possible to instead only call `this.formatAuthOptions()` once, and traverse the array only once to determine `lockedOptions` and `unlockedOptions`? Since N is super small here, it'd only be a small optimization, but would clean up this area a bit.

carl-codeorg (commented):
Looking great, just had a couple small comments/questions. Thanks for taking this lockout functionality across the finish line! It's been in the back of my mind since last July.

apps/src/lib/ui/accounts/ManageLinkedAccounts.jsx
wilkie: It's a tricky refactor. I thought of that, too, and I guess I can build a list when I map the other.

wilkie: Yep. I'll update it to use `_.partition(...)`.

apps/src/lib/ui/accounts/ManageLinkedAccounts.jsx
wilkie: Good question. When there's an id/email, then it is already connected and it shouldn't disable the button because it's then a disconnect button. The easier way to express this is just `!authOption.id` so I've updated it to that. I've also added a comment breaking down the conditional and addressing the logic.

wilkie:
Dang it. Flaky UI tests masked the real issue. Somebody snuck some changes to how the CPA experiment works and my new UI tests broke. Investigating.

apps/src/lib/ui/accounts/ManageLinkedAccounts.jsx
```diff
@@ -161,21 +171,18 @@ class ManageLinkedAccounts extends React.Component {
...
         personalAccountLinkingEnabled={this.props.personalAccountLinkingEnabled}
       />
     );
   };
 
   render() {
-    const lockedOptions = this.formatAuthOptions().filter(option => {
-      return this.shouldDisableConnectButton(option);
-    });
+    const [lockedOptions, unlockedOptions] = _.partition(
```
carl-codeorg, line 183: Thanks for indulging my nitpick, and nice use of underscore! I always forget that we use it as our utility library of choice.

carl-codeorg (approved):
LGTM! (once the pesky UI tests are passing)

apps/src/lib/ui/accounts/ManageLinkedAccounts.jsx
carl-codeorg: Ahh ok, thanks for the explanation. That makes sense now, and I think doing `!authOption.id` is much easier to grok.

apps/src/lib/ui/accounts/ManageLinkedAccounts.jsx
wilkie: Yep nice to learn about the partition method! I was about to write it myself before coming to my senses and checking.

dashboard/app/controllers/test_controller.rb
```diff
@@ -397,4 +397,17 @@ def create_user
...
+
+  # Accepts a parental request that was submitted by the current user
+  def accept_parental_request
+    permission_request = ParentalPermissionRequest.find_by(user: current_user)
+    Services::ChildAccount.grant_permission_request!(permission_request)
+    head :ok
+  end
+
+  # Ages an account via its created_at property
+  def make_account_older
```
daynew: I don't see this being used. Maybe this is an orphan from a merge with one of my changes above which updated `create_user` to allow arbitrary `:created_at` values?

dashboard/test/ui/features/platform/policy_compliance.feature
```diff
@@ -145,3 +145,39 @@ Feature: Policy Compliance and Parental Permission
...
+  Scenario: Student should not be able to connect a third-party account until their account is unlocked
+    Given I create an old account for a young student in Colorado who has never signed in named "Coco Student" and go home
+
+    # Find the locked buttons to connect an account
+    Given I am on "http://studio.code.org/users/edit?cpa-partial-lockout=1&cpa_schedule={%22cpa_new_user_lockout%22:%222023-07-05T23:15:00%2B00:00%22,%22cpa_all_user_lockout_warning%22:%222024-07-01T00:00:00%2B00:00%22,%22cpa_all_user_lockout%22:%222024-07-01T00:00:00%2B00:00%22}"
+    Then I wait to see "#manage-linked-accounts"
+    Then I wait until "form[action=\'/users/auth/google_oauth2?action=connect\'] button" is disabled
+
+    # Navigate the lockout process
+    Given I am on "http://studio.code.org/lockout"
```
daynew: I don't think a user can navigate to this page unless they are locked out? This test is meant to simulate a user who is in the warning/pre-lockout phase, right?

Why navigate to the lockout page? Doesn't `users/edit` have the widget for requesting permission?

dashboard/test/ui/features/platform/policy_compliance.feature
```diff
@@ -145,3 +145,39 @@ Feature: Policy Compliance and Parental Permission
   Scenario: Existing under 13 account in Colorado should not be locked out.
     Given I create a young student in Colorado who has never signed in named "Sally Student" before CPA exception and go home
     Given I am on "http://studio.code.org/home"
+
+  Scenario: Student should not be able to connect a third-party account until their account is unlocked
+    Given I create an old account for a young student in Colorado who has never signed in named "Coco Student" and go home
+
+    # Find the locked buttons to connect an account
+    Given I am on "http://studio.code.org/users/edit?cpa-partial-lockout=1&cpa_schedule={%22cpa_new_user_lockout%22:%222023-07-05T23:15:00%2B00:00%22,%22cpa_all_user_lockout_warning%22:%222024-07-01T00:00:00%2B00:00%22,%22cpa_all_user_lockout%22:%222024-07-01T00:00:00%2B00:00%22}"
```
daynew: Is this UI test going to continue working as intended after July 1, 2024?
I wonder if we need to add a query string param for overriding the date used in our calculations as well.

UI tests are tricky when testing date related behavior 🤔

daynew (changes requested):
Could you double-check some thing related to the UI test?

Also, I tried checking out these changes and manually doing the steps in the UI test but it didn't show a locked "Manage Linked Accounts" section. Maybe we could pair to figure out what's wrong with my setup?

dashboard/app/controllers/test_controller.rb
wilkie: Yep! since you added a way to allow `:created_at`, it isn't needed. I originally did what you did, actually, and then thought I shouldn't permit extra things in that weird `users.json` route... then you did the BEST thing and got rid of it haha. I'll remove it.

wilkie:
> Could you double-check some thing related to the UI test?
> 
> Also, I tried checking out these changes and manually doing the steps in the UI test but it didn't show a locked "Manage Linked Accounts" section. Maybe we could pair to figure out what's wrong with my setup?

Yep. The test for compliance got more permissive between my first implementation and now. I am very much going to want to talk over this because that permissiveness is making me question things.

dashboard/test/ui/features/platform/policy_compliance.feature
wilkie: Absolutely right. Updating it to work with the account settings page instead.

dashboard/test/ui/features/platform/policy_compliance.feature
wilkie: I can easily find out!! Will run the tests with a machine using a future date.

wilkie:
I'm trying to record a demo and describe the 'bug bashing' steps to see the locking as it is for some specific circumstances.

dashboard/app/views/devise/registrations/edit.html.haml
```diff
@@ -130,7 +130,9 @@
   = form_for(resource, as: resource_name, url: registration_path(resource_name), html: {method: :put, id: 'lti-sync-settings-form'}) do |f|
     = f.hidden_field :lti_roster_sync_enabled
 
-- if current_user.student? && experiment_value('cpa-partial-lockout', request) && Cpa.cpa_experience(request) && Policies::ChildAccount.user_predates_policy?(current_user)
+-# The CPA checks for whether or not the account is a personal login... so we check everything except that
+- locked = experiment_value('cpa-partial-lockout', request) && Policies::ChildAccount.user_predates_policy?(current_user, assume_permission_required: true)
```
wilkie: could change this to:

```
- locked = experiment_value('cpa-partial-lockout', request) && Policies::ChildAccount.underage?(current_user) && !Policies::ChildAccount::ComplianceState.permission_granted?(current_user)
```

which I think is the same idea: We have a partial lockout (can see the account settings) where you cannot link accounts or create a personal login.

This `locked` when true will show the parental permission field and lock the personal account creation until you are granted permission (or become old enough).

wilkie:
Quickly adding the us state to the account settings page:

image

Soon to be there... but I need to pass the US State into the ManageLinkedAccounts component so it can render the needs a state permission vs the parental permission (or do what Artem did and pass the string in... but I'd rather localize it in JS-land)

dashboard/app/views/devise/registrations/edit.html.haml
```diff
@@ -75,6 +75,12 @@
       .field
         = f.label :gender_student_input, t('signup_form.gender'), class: "label-bold"
         = f.text_field :gender_student_input, maxlength: 50
+    .field
+      - location = Geocoder.search(request.ip).try(:first)
+      - country_code = location&.country_code.to_s.upcase
+      = f.label :us_state, t('activerecord.attributes.user.us_state'), class: 'label-bold'
+      = f.select :us_state, us_state_options, {include_blank: t('school_info.select_state')}, required: true
+      = f.hidden_field :country_code, value: country_code
```
daynew, line 83:
```suggestion
      = f.hidden_field :country_code, value: country_code
      = f.hidden_field :user_provided_us_state, value: true
```
During the Live Site issue, we added this attribute to indicate the US State was provided by the user. This was added because we decided that we didn't want to trust the values we backfilled using the the teacher's US State.

dashboard/app/views/devise/registrations/edit.html.haml
```diff
@@ -130,7 +136,9 @@
   = form_for(resource, as: resource_name, url: registration_path(resource_name), html: {method: :put, id: 'lti-sync-settings-form'}) do |f|
     = f.hidden_field :lti_roster_sync_enabled
 
-- if current_user.student? && experiment_value('cpa-partial-lockout', request) && Cpa.cpa_experience(request) && Policies::ChildAccount.user_predates_policy?(current_user)
+-# The CPA checks for whether or not the account is a personal login... so we check everything except that
+- locked = experiment_value('cpa-partial-lockout', request) && Policies::ChildAccount.underage?(current_user) && !Policies::ChildAccount::ComplianceState.permission_granted?(current_user)
```
daynew: Should this also be limited to USA?

wilkie:
https://github.com/code-dot-org/code-dot-org/assets/31674/c2406230-c7cb-45a2-89d9-9aca4a85c00e

This shows two students that are 'sponsored' where they log in via the section and a picture. These students are locked out of the account linking / personal account creation flow until they provide a state.

If that state is a policy state, such as Colorado, they are then locked out of those flows until they get parental permission. The parental permission request is now visible above which can be used to unlock when the request is permitted via the parental request email (simulated above with a database update in the console).

The `cpa-partial-lockout` experiment does gate the personal account creation lockout. Without the experiment, this is still allowed.

Note: It wrongfully says 'pending' in the parental request because I incorrectly reset the state in-between runs of recording. It indeed said 'Not requested' or whatever initially and it would never get into this weird state normally.

dashboard/app/views/devise/registrations/edit.html.haml
wilkie: If they do not have a policy, `underage?` returns false. That's true for those outside the USA, right?

dashboard/app/views/devise/registrations/edit.html.haml
wilkie: Oh! I did not see this in the state modal where I copied it from: https://github.com/code-dot-org/code-dot-org/blob/staging/dashboard/app/views/layouts/_student_information_interstitial.html.haml

daynew: Oh you know what, it is probably added on the controller side, my bad

dashboard/app/views/devise/registrations/edit.html.haml
daynew: oh of course.

dashboard/app/views/devise/registrations/edit.html.haml
wilkie: It won't be! Good catch! The interstitial modal uses a different route to update the state and that sets the flag. The general user update path won't do that. I'll patch it.

wilkie:
Updated text:

image

I apparently need to rebase. Pending!

wilkie:
Added clamping the parental permission section to the account settings so that you see the "Granted" status:

image

wilkie:
This has been rebased... but I'm noticing I should update the prompt before the "Permission: Granted" to not say "You need parental permission ..." anymore. Doing that and then running the tests for the granted state.

wilkie:
Rebased again but this should be in its more-or-less final state conditional on the tests reflecting the right things and my rebases didn't delete anything.

daynew (approved):

PR #58351, opened 2024-05-01
Error on profane response from model

bencodeorg:
This PR applies the profanity filter to responses received from Sagemaker. If a message is found to be profane, we display a generic "There was an error getting a response. Please try again" message, and do not include the user message that preceded the profane response in the next attempt to make a request to the model.

image

From a logging perspective, I'm looking to log the profane content somewhere (last step in this PR, location TBD!), but **not** as part of our normal logging session, since I didn't want to forward the content back and forth between Rails and the front end (or expose it at all).

## Testing story

Tested manually that a stubbed version of the model producing profanity was flagged and resulted in the error message in the screenshot below. Also tested that normal messages and logging continued to work as expected. Added a unit test to cover this case.


apps/src/aichat/views/aichatView.module.scss
```diff
@@ -11,7 +11,6 @@
   padding: 0;
   box-sizing: border-box;
   overflow: hidden;
-  user-select: none;
```
bencodeorg, line 14: Unrelated change to allow copying text (particularly messages). More discussion in [this Slack thread](https://codedotorg.slack.com/archives/C06UT1E2796/p1714149250477099).

dashboard/app/controllers/aichat_controller.rb
```diff
@@ -47,6 +47,27 @@ def chat_completion
     sagemaker_response = AichatSagemakerHelper.request_sagemaker_chat_completion(input, params[:aichatModelCustomizations][:selectedModelId])
     latest_assistant_response = AichatSagemakerHelper.get_sagemaker_assistant_response(sagemaker_response)
 
+    filter_result = ShareFiltering.find_failure(latest_assistant_response, locale)
+    if filter_result&.type == ShareFiltering::FailureType::PROFANITY
+      # log here that profanity happened
```
bencodeorg: I want to log somewhere that this happened, but I'm not totally sure what is the right way to do that. [Discussing here](https://codedotorg.slack.com/archives/C046G4TRLEN/p1714672239480129), but didn't want to block review on this otherwise.

apps/src/aichat/redux/aichatRedux.ts
```diff
@@ -278,41 +280,67 @@ export const submitChatContents = createAsyncThunk(
...
         status: Status.OK,
         chatMessageText: chatApiResponse.content,
-        // The accuracy of this timestamp is debatable since it's not when our backend
-        // issued the message, but it's good enough for user testing.
         timestamp: getCurrentTimestamp(),
         sessionId: chatApiResponse.session_id,
       };
-      return thunkAPI.dispatch(addChatMessage(assistantChatMessage));
-    }
+      thunkAPI.dispatch(addChatMessage(assistantChatMessage));
```
bencodeorg, line 322: This I think is out of scope for this PR, but wanted to get any thoughts other folks have...I chatted with Sanchit a little bit about this, but I'm getting anxious about the degree to which logging currently relies on keeping the state of chat messages in sync between what's in Redux and what's being stored in our logging model in Rails.

For example, we rely on this `assistantChatMessage` being added to the chat history in a way that matches the chat message we've added to the session in Rails here:

https://github.com/code-dot-org/code-dot-org/blob/1f8d78272f3897668876df0bc5a309a7abde61b9/dashboard/app/controllers/aichat_controller.rb#L51-L55

I don't have this fully thought out, I'm wondering if we could pass back and forth the entire list of messages instead of doing this appending on both sides (and relying on that appending to be the same).

apps/src/aichat/views/ChatMessage.tsx
```diff
@@ -24,7 +24,11 @@ const isUser = (role: string) => role === Role.USER;
 const isModelUpdate = (role: string) => role === Role.MODEL_UPDATE;
 
 const displayUserMessage = (status: string, chatMessageText: string) => {
-  if (status === Status.OK || status === Status.UNKNOWN) {
+  if (
+    status === Status.OK ||
+    status === Status.UNKNOWN ||
+    status === Status.ERROR
```
bencodeorg, line 30: If we mark the response from the model in an error state, we also mark the preceding student message as in an "error" state, such that it doesn't get sent to the model on the next request cycle (we only send messages in the "ok" state, in addition to the current message from the user). From a UI perspective, though, treating it like other messages. Curious for other thoughts on a) whether that UI seems ok, and b) whether marking it as an "error" message seems ok.

apps/src/aichat/redux/aichatRedux.ts
thomasoniii: What exactly is the concern? It sounds like the worry is that we have a list of messages on the client in redux, and we're trying to propagate that same list of messages to the server, _and_ that the list could potentially get out of sync. So the speculation is maybe to send the entire list of chat messages. Am I following it?

If so, maybe just sending a unique incrementing ID could do it. Maybe we namespace it with some unique value?

So when we start up, we assign that we're dealing with chat `X` (or whatever) and then we send along an incrementing id `13` for this message - so we tack into the packet `messageID: 'X,13'` (or however it's formatted. Maybe they're fixed width).

Then the server keeps track of if we've see the previous message. If so, cool - we're gtg. If not, then it fails back to the client with a message of the _last_ ID it's successfully received. The client then transmits the next ID up through the present. If the server receives a duplicate message ID, it just discards and ignores it (maybe the other request shows up late).

It's a little more work to implement, but it keeps us still sending only the updated messages, and re-transmitting only the missed ones if anything's dropped.

Or maybe I'm misunderstanding the issue. 😄 

Regardless, I do agree with the sentiment of it being out of scope for this PR, so we should prob bump to a slack thread to discuss further.

thomasoniii (approved):
a note for follow up discussion about the message logging, but still lgtm 🚀

apps/src/aichat/redux/aichatRedux.ts
```diff
@@ -278,41 +280,67 @@ export const submitChatContents = createAsyncThunk(
...
 
-    console.log('Could not handle response.');
+      thunkAPI.dispatch(
+        updateUserChatMessageStatus({
```
fisher-alice, line 325: ❤️

apps/src/aichat/redux/aichatRedux.ts
fisher-alice: +1 on being out of scope of this PR, but if we want to add some kind of sync check, we have handy the `ChatCompletionMessage` `id` - maybe we could implement check as a P1?

apps/src/aichat/views/ChatMessage.tsx
fisher-alice: I'm curious what @samantha-code or @dancodedotorg what they think. This is how it looks now:

Screenshot 2024-05-03 at 3 55 42 PM

fisher-alice: Adding an option to just not displaying the user message that resulted in profane assistant response.

dashboard/app/controllers/aichat_controller.rb
fisher-alice: Just confirming that logging wb part of this PR?

dashboard/app/controllers/aichat_controller.rb
```diff
@@ -47,6 +47,27 @@ def chat_completion
...
+          content: params[:newMessage],
+          status: SharedConstants::AI_INTERACTION_STATUS[:ERROR]
+        },
+      ]
+
+      return render(
+        status: :ok,
+        json: {
+          status: SharedConstants::AICHAT_ERROR_TYPE[:PROFANITY_MODEL],
+          error: 'Profane content generated by model',
```
fisher-alice: Are we using this `error` value on the front end?

dashboard/test/controllers/aichat_controller_test.rb
```diff
@@ -109,6 +109,24 @@ class AichatControllerTest < ActionController::TestCase
     assert_equal 1, (JSON.parse(session.messages).count {|message| message["status"] == 'profanity_violation'})
   end
 
+  test 'returns failure when model response contains profanity' do
```
fisher-alice: Maybe rename as returns 'error'

fisher-alice (approved):
I know you'll add logging as part of ths PR but looks good so far!
I added a suggestion to ask product/curric about the UI question you had. But love how this is taking shape!

apps/src/aichat/views/ChatMessage.tsx
bencodeorg: Raised here, no objections: https://codedotorg.slack.com/archives/C06FELVTV0Q/p1714773355042179

dashboard/app/controllers/aichat_controller.rb
bencodeorg: Yep! Just added.

dashboard/app/controllers/aichat_controller.rb
bencodeorg: Nope! Removed.

dashboard/test/controllers/aichat_controller_test.rb
bencodeorg: Thanks for flagging this -- I updated this to "returns model profanity status when model response contains profanity" and the user profanity test similarly to "returns user profanity status when chat message contains profanity".

I decided not to say "error" because we're actually returning a success status, and also added a couple other assertions (ie, that the response status is success, and that the right keys are in the JSON response).

dashboard/app/helpers/aichat_sagemaker_helper.rb
```diff
@@ -18,7 +18,7 @@ def self.create_sagemaker_client
   # Mistral-7B-Instruction LLM instruction format doc at https://huggingface.co/mistralai/Mistral-7B-Instruct-v0.1.
   def self.format_inputs_for_sagemaker_request(aichat_params, stored_messages, new_message)
     all_messages = [*stored_messages, {role: USER, content: new_message}]
-    inputs = aichat_params[:systemPrompt]
+    inputs = aichat_params[:systemPrompt] + " "
```
bencodeorg, line 20: @fisher-alice unrelated change, but I noticed these were getting appended together without a space, does this seem ok?

dashboard/test/controllers/aichat_controller_test.rb
fisher-alice: 🎉

dashboard/app/helpers/aichat_sagemaker_helper.rb
fisher-alice: Sure. Thanks!

apps/src/aichat/redux/aichatRedux.ts
bencodeorg: Thanks for the thoughts y'all -- just set a reminder to start a Slack thread on this on Monday :)

apps/src/aichat/views/ChatMessage.tsx
sanchitmalhotra126: Discussed briefly last week - I do think it feels a little weird to call this an "Error" state when it wasn't truly a user error, but also understand the need for not marking it "OK" so it gets filtered out. Not sure if I have a better idea of what to use and I don't want to add another constant just for the sake of it, so fine with this for now :) happy to revisit model architecture more broadly on Slack/our sync

dashboard/app/controllers/aichat_controller.rb
```diff
@@ -47,6 +47,33 @@ def chat_completion
...
+      new_messages = [
+        {
+          role: 'user',
+          content: params[:newMessage],
+          status: SharedConstants::AI_INTERACTION_STATUS[:ERROR]
+        }
+      ]
+      session_id = log_chat_session(new_messages)
+
+      Honeybadger.notify(
```
sanchitmalhotra126, line 61: Is HB the right tool here? My understanding was that we want to use HoneyBadger primarily for server errors that we can triage and should try to limit/eliminate. But in this case, is this an error we would actively try to drive down?

sanchitmalhotra126 (approved):
LGTM! As you mentioned, we can continue discussion model architecture offline, but good with this going out as is

dashboard/app/controllers/aichat_controller.rb
bencodeorg: Yeah good q -- tl;dr is I'm not sure. Some more discussion in the Slack thread linked in [this comment](https://github.com/code-dot-org/code-dot-org/pull/58351#discussion_r1588431675) -- I'm happy to revisit the approach here down the line.

PR #54951, opened 2023-11-15
Adds recording of metrics for tokens used in ai rubrics and enables retries on rubric jobs.

wilkie:
This adds CloudWatch metrics to track the tokens used in total and as part of either prompt or response.

This also adds the ability to retry on a rate limiting action bubbled up from the aiproxy service. Additionally, timeouts from the aiproxy service (say, if it is itself restarting, etc) will trigger a retry. The retries are exponentially backed off.

## Links

- jira ticket: [AITT-244](https://codedotorg.atlassian.net/browse/AITT-244)
- jira ticket: [AITT-245](https://codedotorg.atlassian.net/browse/AITT-245)

## Testing story

Added a new matcher to the existing Metrics matchers to detect the correct use of dimensionality of the metric.

Otherwise, it is amending to our set of unit tests for the EvaluateRubricJob class.

## Follow-up work

I need to ensure that the aiproxy can bubble up a 429. I tested this by adding a dedicated 429 triggering route on the proxy service. I could commit and push that, at least.

## PR Checklist:

- [ ] Tests provide adequate coverage
- [ ] Privacy and Security impacts have been assessed
- [ ] Code is well-commented
- [ ] New features are translatable or updates will not break translations
- [ ] Relevant documentation has been added or updated
- [ ] User impact is well-understood and desirable
- [ ] Pull Request is labeled appropriately
- [ ] Follow-up work items (including potential tech debt) are tracked and linked

PR #58354, opened 2024-05-01
log to cloudwatch instead of HB on aiproxy request retry

davidsbailey:
Finishes https://codedotorg.atlassian.net/browse/AITT-597. Some relevant background here: https://github.com/code-dot-org/code-dot-org/pull/54951/files#r1394758841

## Testing story

new unit test coverage

## Follow-up steps

resolve https://app.honeybadger.io/projects/3240/faults/106895906/01HVM50BD81SEVA9545BV6XN4Q after this PR ships (currently it is paused)

### DIFF
--- a/dashboard/app/jobs/evaluate_rubric_job.rb
+++ b/dashboard/app/jobs/evaluate_rubric_job.rb
@@ -231,10 +231,7 @@ class EvaluateRubricJob < ApplicationJob
   RETRIES_ON_TIMEOUT = 2
 
   # Retry just once on a timeout. It is likely to timeout again.
-  retry_on Net::ReadTimeout, Timeout::Error, wait: 10.seconds, attempts: RETRIES_ON_TIMEOUT do |job, _error|
-    options = job.arguments.first
-    script_level = ScriptLevel.find(options[:script_level_id])
-
+  retry_on Net::ReadTimeout, Timeout::Error, wait: 10.seconds, attempts: RETRIES_ON_TIMEOUT do |_job, _error|
     Cdo::Metrics.push(
       AI_RUBRIC_METRICS_NAMESPACE,
       [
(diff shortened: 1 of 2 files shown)


cat5inthecradle:
I think this is a great candidate for structured cloudwatch logs, from which we could infer some metrics. We don't have this implemented yet, but it has popped up in a few conversations, including this week's issues with the Test server.

This PR looks good to me, just noting is the loss of context that you were previously able to send to Honeybadger.

Are there any additional dimensions worth adding? I'm thinking maybe the name/url of the particular resource that timed out.

davidsbailey:
@cat5inthecradle please tell me more about structured cloudwatch logs! I would love to log user id, script level id, etc so that we can see if a certain project is failing repeatedly, though I guess we'll find that out another way if the project fails more times than the number of retries. another wish is to send along error details from aiproxy and have those show up in these logs, so that we don't have to do forensics to match up this occurrence with the raw logs from aiproxy.

I'm open to adding more dimensions, but at the moment all of our traffic from this job is going to the assessment endpoint on aiproxy so I'm not sure if that particular piece of data adds anything?

davidsbailey:
as an aside, new relic has "custom metrics" and "custom events". metrics are like any other metrics, and events could be searched with a SQL-like interface. I thought this worked really well, and wish we had something like this within AWS

davidsbailey:
lastly... I'm actually tempted to log the error details to firehose, since that actually implements exactly what I need

cat5inthecradle:
Firehose gets the logs into S3 which makes them queryable via AWS Athena. You can do the same thing with Cloudwatch Logs. Cloudwatch Logs also integrate more closely with metrics.

You use MetricFilters for that, something like this in Cloudformation:

```yaml
  TimeoutMetricFilter:
    Type: "AWS::Logs::MetricFilter"
    Properties:
      LogGroupName: 
        Ref: TimeoutLogGroup
      FilterPattern: '{ $.event = "timeout" }'
      MetricTransformations:
        - MetricName: "TimeoutEventCount"
          MetricNamespace: "YourApplication"
          MetricValue: "1"
          DefaultValue: 0
```

Firehose might be a better intermediate step than Metrics because it allows for wider events. I guess I wasn't aware we were doing that on the backend though.

davidsbailey:
> This PR looks good to me, just noting is the loss of context that you were previously able to send to Honeybadger.

> Firehose might be a better intermediate step than Metrics because it allows for wider events. 

Adding firehose logging in https://github.com/code-dot-org/code-dot-org/pull/58411, then will switch to writing to cloudwatch logs once that is working.

> I guess I wasn't aware we were doing that on the backend though.

Can you clarify what you mean by this last part?

PR #58367, opened 2024-05-01
Create flag to control showing the diversity survey to even or odd us…

etaderhold:
This change introduces a DCDO flag `diversity_audience` to control who will see the diversity survey. This mirrors an existing `nps_audience` flag to control the NPS survey audience.

Possible values:
* `all` (default) Show survey to all users who otherwise qualify
* `none` Suppress survey for all users
* `odd` Suppress survey for even-numbered user IDs
* `even` Suppress survey for odd-numbered user IDs

This is needed to avoid cluttering the homepage with multiple surveys at once when the NPS survey goes live later this month.

## Links

[JIRA](https://codedotorg.atlassian.net/browse/TEACH-1056)

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/dashboard/app/helpers/survey_results_helper.rb
+++ b/dashboard/app/helpers/survey_results_helper.rb
@@ -13,6 +13,11 @@ module SurveyResultsHelper
     return false unless has_any_students?
     return false unless has_any_student_under_13?
 
+    diversity_audience = DCDO.get('diversity_audience', 'all')
+    return false if diversity_audience == 'none'
+    return false if diversity_audience == 'odd' && current_user.id.even?
+    return false if diversity_audience == 'even' && current_user.id.odd?
+
     # There is no reason not to show the survey, so show the survey.
     return true
   end
--- a/dashboard/test/helpers/survey_results_helper_test.rb
+++ b/dashboard/test/helpers/survey_results_helper_test.rb
@@ -44,11 +44,25 @@ class SurveyResultsHelperTest < ActionView::TestCase
   end
 
   test 'show diversity survey' do
+    current_survey = SurveyResult::DIVERSITY_2024
+
     stubs(:language).returns "en"
     stubs(:request).returns(stub(location: stub(try: "RD")))
     follower = create :follower, user: @teacher
     follower.student_user.update(age: 10)
-    assert show_diversity_survey? SurveyResult::DIVERSITY_2024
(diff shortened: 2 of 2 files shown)


dashboard/app/helpers/survey_results_helper.rb
```diff
@@ -13,6 +13,11 @@ def show_diversity_survey?(kind)
     return false unless has_any_students?
     return false unless has_any_student_under_13?
 
+    diversity_audience = DCDO.get('diversity_audience', 'all')
```
dju90, line 16: Not the expert here, but would it be safer to set the default value of the `DCDO.get` to 'none', then have `return true if diversity_audience == 'all'`? That way, if we forget to set the DCDO flag at all, nothing will happen.

etaderhold: The `..., 'all')` argument at the end specifies what should be returned if the flag is unset. Same effect as what you propose.

dju90: Hmm, isn't it currently the opposite? If the 'diversity_audience' flag is unset, `diversity_audience` will be set to 'all' - and then won't we hit `return true` on line 22?

etaderhold: Yes if the flag is unset `diversity_audience` will be `all`, which will not cause any of these three lines looking at `diversity_audience` to return false, so we get to the end and return true, which means the survey will be shown.

dju90: Ah ok, no difference in how we're thinking through the logic. Just making a note for posterity that we will still need to "turn on" each year's diversity survey via PR (a la #57997) in addition to setting this DCDO `diversity_audience` flag to 'all' (or nil). Thanks Eric!

dju90 (approved):

dju90:
Not a blocker, just for my own learning: when do we need to add a new DCDO flag to `lib/dynamic_config/dcdo.rb` and when do we not (like in this PR)?

etaderhold:
> Not a blocker, just for my own learning: when do we need to add a new DCDO flag to `lib/dynamic_config/dcdo.rb` and when do we not (like in this PR)?

That file is used to send the current values of certain DCDO flags for use in the React frontend. We don't need to read this flag there, so no need to put it in that file.

PR #58368, opened 2024-05-01
Adding a table to track users as they go through the Child Account Po…

daynew:
As a PM responsible for Code.org's implementation of the Colorado Privacy Act, I want data about how users move through our pipeline. For example, how many parent permission emails are sent? What is the average amount of time it takes for a user to get parent permission once an email is sent?

This PR creates a new DB table where we will log Child Account Policy events which our PM's can query and answer the questions they have. See the JIRA task for the DB schema to implement.

## Links
* [JIRA](https://codedotorg.atlassian.net/browse/P20-893)

## Testing story
* `bundle exec rake db:migrate`
* `bundle exec rake db:rollback`

## Follow-up work
* A follow-up PR will actually log the events.

### DIFF
--- a/dashboard/db/migrate/20240501215821_create_cap_user_events.rb
+++ b/dashboard/db/migrate/20240501215821_create_cap_user_events.rb
@@ -2,7 +2,6 @@ class CreateCapUserEvents < ActiveRecord::Migration[6.1]
   def change
     create_table :cap_user_events do |t|
       t.string :name, null: false, limit: 64
-      t.datetime :occurred_at, null: false
       t.references :user, null: false, foreign_key: true, type: :integer, index: true
       t.string :policy, null: false, limit: 16
 
--- a/dashboard/db/migrate/20240501215821_create_cap_user_events.rb
+++ b/dashboard/db/migrate/20240501215821_create_cap_user_events.rb
@@ -1,0 +1,15 @@
+class CreateCapUserEvents < ActiveRecord::Migration[6.1]
+  def change
+    create_table :cap_user_events do |t|
+      t.string :name, null: false, limit: 64
+      t.datetime :occurred_at, null: false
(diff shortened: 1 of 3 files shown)


dashboard/db/schema.rb
```diff
@@ -168,6 +168,18 @@
     t.string "callout_text"
   end
 
+  create_table "cap_user_events", charset: "utf8mb3", collation: "utf8mb3_unicode_ci", force: :cascade do |t|
+    t.string "name", limit: 64, null: false
+    t.datetime "occurred_at", null: false
```
wilkie: Now that I see the schema as a whole, is the `occurred_at` property redundant when we have the `timestamps` and the `created_at` property already?

daynew: It definitely is redundant. I will update this.

dashboard/db/migrate/20240501215821_create_cap_user_events.rb
```diff
@@ -0,0 +1,15 @@
+class CreateCapUserEvents < ActiveRecord::Migration[6.1]
+  def change
+    create_table :cap_user_events do |t|
```
carl-codeorg, line 3: Nit: spelling out `child_account_policy` would make it easier to know what the table is at a glance.

carl-codeorg (approved):
LGTM

dashboard/db/migrate/20240501215821_create_cap_user_events.rb
daynew: I'm on the fence but slightly leaning towards the shorter name because ChildAccountPolicyUserEvents is approaching a length which makes me uncomfortable (I acknowledge this is totally just a gut feeling). I think it should be pretty easy to discover what "Cap" stands for given it's usage in the code base.

PR #58371, opened 2024-05-02
Music: Rename effects dropdown labels & make them dynamic based on selected effect

cnbrenci:
This PR replaces effect block labels with strings from #58246 and #58263 and changes the effect values available depending on the effect selected in the block to match the following:

Volume: Full / Medium / Low
Delay: Off / High / Low
Filter: Off / Medium / Low

Effects blocks in toolbox after change:
image

Interaction of the drop downs

https://github.com/code-dot-org/code-dot-org/assets/24951544/23c3d152-663a-43a2-a576-cc278b528687

## Links

[LABS-682: Rename effects dropdown labels](https://codedotorg.atlassian.net/browse/LABS-682)

## Testing story

- [x] Verified the main.json saved only contains the values which haven't changed from before
- [x] Test against a project created without the change that uses the block
- [x] Test to make sure the effects still work correctly

## Deployment strategy

No special rollout needed. Only labels in the dropdown changed, not the values, so that pre-existing projects saved with old values are still compatible.

## Follow-up work

- [ ] Automated testing


apps/src/music/blockly/toolbox.js
```diff
@@ -65,20 +67,23 @@ const toolboxBlocks = {
...
       [FIELD_EFFECTS_NAME]: 'volume',
+      [FIELD_EFFECTS_VALUE]: DEFAULT_EFFECT_VALUE,
     },
   },
   [BlockTypes.SET_FILTER_EFFECT_AT_CURRENT_LOCATION_SIMPLE2]: {
     kind: 'block',
     type: BlockTypes.SET_EFFECT_AT_CURRENT_LOCATION_SIMPLE2,
     fields: {
       [FIELD_EFFECTS_NAME]: 'filter',
+      [FIELD_EFFECTS_VALUE]: DEFAULT_EFFECT_VALUE,
```
bencodeorg, line 80: For my understanding, is this a behavior change? Or just being more explicit?

apps/src/music/blockly/blocks/simple2.js
```diff
@@ -210,20 +212,12 @@ export const setEffectAtCurrentLocationSimple2 = {
...
-          ['volume', 'volume'],
-          ['filter', 'filter'],
-          ['delay', 'delay'],
-        ],
+        options: FIELD_EFFECT_NAME_OPTIONS,
       },
       {
-        type: 'field_dropdown',
+        // This input is replaced with a field_dropdown by the extension
+        type: 'input_dummy',
```
bencodeorg, line 219: Also for my learning, is this a standard practice (my blockly is rusty 😬 )?

bencodeorg (approved):
Blockly gets my brain in knots but looks good as far as I can tell!

apps/src/music/blockly/blocks/simple2.js
```diff
@@ -10,6 +10,8 @@ import {
   DOCS_BASE_URL,
   FIELD_TRIGGER_START_NAME,
   TriggerStart,
+  EFFECTS_FIELD_EXTENSION,
```
fisher-alice: Nit: Is there a reason this is named `EFFECTS_FIELD_EXTENSION` instead of `FIELD_EFFECTS_EXTENSION`?

apps/src/music/blockly/constants.js
```diff
@@ -52,3 +55,33 @@ export const TriggerStart = {
...
+  [musicI18n.effectsLabelsDelay(), 'delay'],
+];
+
+// Even though these strings don't necessarily match the labels, they must
+// stay the same to avoid breaking existing projects saved with these values.
+export const DEFAULT_EFFECT_VALUE = 'normal';
+export const MEDIUM_EFFECT_VALUE = 'medium';
+export const LOW_EFFECT_VALUE = 'low';
+
+export const FIELD_EFFECTS_OPTIONS = {
```
fisher-alice: Maybe rename as `FIELD_EFFECTS_VALUE_OPTIONS` to compare with `FIELD_EFFECTS_NAME_OPTIONS` above.

apps/src/music/blockly/extensions.js
```diff
@@ -111,3 +115,34 @@ export const playMultiMutator = {
     this.updateShape_(this.extraSoundInputCount_ - 1);
   },
 };
+
+/*
+ * Extension for the effects field, which replaces the input_dummy in the block
```
fisher-alice, line 120: In general, nice comments!

apps/src/music/blockly/extensions.js
```diff
@@ -111,3 +115,34 @@ export const playMultiMutator = {
...
+ * with a dynamic field_dropdown and updates the effect value dropdown
+ * based on the selected effect name.
+ */
+export const effectsFieldExtension = function () {
+  // Set the initial state when the block gets created
+  const thisBlock = this;
+  const valuesDropdown = new Blockly.FieldDropdown(function () {
+    return FIELD_EFFECTS_OPTIONS[thisBlock.getFieldValue(FIELD_EFFECTS_NAME)];
```
fisher-alice: Nit: instead of function, maybe simpler to provide the array `FIELD_EFFECTS_OPTIONS[thisBlock.getFieldValue(FIELD_EFFECTS_NAME)]` as you did in line 143 below.

apps/src/music/blockly/extensions.js
```diff
@@ -111,3 +115,34 @@ export const playMultiMutator = {
...
+  valuesDropdown.setValue(DEFAULT_EFFECT_VALUE);
+  this.getInput(FIELD_EFFECTS_VALUE).appendField(
+    valuesDropdown,
+    FIELD_EFFECTS_VALUE
+  );
+
+  // Set up handler to update the effect value when the effect name changes
+  const effectNameField = this.getField(FIELD_EFFECTS_NAME);
+  const baseHandler = effectNameField.onItemSelected_;
+  effectNameField.onItemSelected_ = (menu, menuItem) => {
```
fisher-alice: Are we overriding this function? If so, could add a comment.

apps/src/music/blockly/extensions.js
```diff
@@ -111,3 +115,34 @@ export const playMultiMutator = {
...
+    // Update the effect value dropdown's options to match the newly selected effect
+    const newNameValue = menuItem.opt_value;
```
fisher-alice: 'Value' seems to be overloaded here as it's also be used for the effect values and in this case, I think we want the effect name. Maybe use `newNameEffect`?

apps/src/music/blockly/blocks/simple2.js
fisher-alice: FWIW, I see it at https://github.com/code-dot-org/code-dot-org/blob/7c1b5bee73ca9ec2992642b3fef2b825f8b5ce33/apps/src/blockly/customBlocks/googleBlockly/behaviorBlocks.ts#L89.

fisher-alice (approved):
Nice job! Just left a handful of minor comments.

apps/src/music/blockly/blocks/simple2.js
sanchitmalhotra126: Yeah, I do think this is a blockly standard as per these docs as well: https://developers.google.com/blockly/guides/create-custom-blocks/define-blocks#block_inputs

sanchitmalhotra126 (approved):
Nice! +1 to the other comments but otherwise LGTM!

apps/src/music/blockly/blocks/simple2.js
cnbrenci: Not thrilled about the name, but I got it from this doc: https://developers.google.com/blockly/guides/create-custom-blocks/fields/built-in-fields/dropdown#dynamic_dropdowns

bencodeorg: Nice, thanks for the info y'all!

apps/src/music/blockly/toolbox.js
cnbrenci: Without this, the blocks in the toolbox did not have the correct effect values until they were dragged into the workspace. They all started with "Full" as the effect value label

apps/src/music/blockly/blocks/simple2.js
cnbrenci: Hmm. In my head it makes more sense in that order, but now that you point it out, I'm noticing that the `FIELD_EFFECTS_EXTENSION` matches the convention better. Thanks for pointing that out! I'll update.

apps/src/music/blockly/extensions.js
cnbrenci: Yeah, field dropdown options have a label and a value, and then the block has an effect name and effect value dropdown >.> 

maybe 'selectedEffectName' would be clearer? cause it's the effect name selected in the effect name dropdown. (it is technically the value of the option, but yeah it's confusing since value is overloaded)

Also just realized I'm calling it effectNameField here instead of fieldEffectName like it is everywhere else. Lemme update the naming to both and hopefully it'll be more clear! Thanks for the feedback!

apps/src/music/blockly/extensions.js
cnbrenci: If I make this change, all the Effects block in the toolbox have the default starting value "Full". It needs the function so that it can evaluate the value based on the field value of `FIELD_EFFECTS_NAME` at the time the block gets instantiated.

fisher-alice: Gotcha - thanks for the explanation!

cnbrenci: You're welcome!

PR #58372, opened 2024-05-02
Adds invitation for participants to view the new progress view

kobryan0619:
This PR creates a modal inviting users to use the new section progress view.  This PR:

- Creates the pop-up with the right elements. 
- Allows the user to dismiss the pop-up or accept the invitation
- Logs the events to amplitude.
- Save data that the invitation has been seen or delayed. 
- Accurately process a "delayed" invitation

image

(NEW): It DOES
- "Remember" the status of the invitation if a user switches between the tabs in the dashboard. 

## Links

[Ticket](https://codedotorg.atlassian.net/jira/software/projects/TEACH/boards/65?selectedIssue=TEACH-870)

## Testing story

- Tests coming in separate PR.

## Follow-up work

- Fix issue with tabs not recognizing the data.
- fix issue with image being slow to load


MollyPeredo:
Looks great! Just one tweak - can you remove the teal border-radius on the "close" button?
image

kobryan0619:
> Looks great! Just one tweak - can you remove the teal border-radius on the "close" button?
> image

https://github.com/code-dot-org/code-dot-org/assets/24883357/c1a8ef63-13f9-4426-bf0e-7419adab3459


That's actually a focus indicator that moves through out the component as a user tab navigates.  This outline is also what shows up in other modals - is there a different focus indicator you'd prefer for this?

apps/src/templates/currentUserRedux.js
```diff
@@ -213,6 +215,9 @@ export default function currentUser(state = initialState, action) {
       progressTableV2ClosedBeta: progress_table_v2_closed_beta,
       isLti: is_lti,
       isTeacher: user_type === UserTypes.TEACHER,
+      dateProgressTableInvtationDelayed:
```
lfryemason: nit: Invtation -> Invitation
```suggestion
      dateProgressTableInvitationDelayed:
```

apps/src/templates/sectionProgress/SectionProgress.jsx
```diff
@@ -174,6 +179,12 @@ class SectionProgress extends Component {
         className={dashboardStyles.dashboardPage}
         data-testid="section-progress-v1"
       >
+        {allowUserToSelectV2View && (
```
lfryemason: Nit: could we move this logic to the `SectionProgressSelector`? That class contains all the logic for the banners and determines whether to show v1 or v2. I feel like this popup fits better there.

dju90:
> It does not: "Remember" the status of the invitation if a user switches between the tabs in the dashboard. It DOES however remember the status of the invitation if a teacher reloads or logs out and back in.

Can you clarify what this means? Does it mean that if I click "close" on the invitation, then switch to a different dashboard tab and then switch back, the invitation will re-appear?

apps/src/templates/sectionProgressV2/InviteToV2ProgressModal.jsx
```diff
@@ -0,0 +1,140 @@
...
+    });
+  };
+
+  const setHasSeenProgressTableInvite = () => {
+    return $.post(`/api/v1/users/has_seen_progress_table_v2_invitation`, {
+      has_seen_progress_table_v2_invitation: true,
+    });
+  };
+
+  if (invitationOpen) {
```
lfryemason, line 108: Optional nit: I find `  if (!invitationOpen) { return null} ...` to be slightly clearer because it shows the early return before the big blob of content to show. Completely optional nit pick

lfryemason (commented):
Looks great, a couple nit-picks and that's it. Code structure and logic are all sound!

MollyPeredo:
> > Looks great! Just one tweak - can you remove the teal border-radius on the "close" button?
> > image
> 
>  Screen.Recording.2024-05-02.at.9.29.28.AM.mov 
> That's actually a focus indicator that moves through out the component as a user tab navigates. This outline is also what shows up in other modals - is there a different focus indicator you'd prefer for this?

Ahhhhh, got it. This is perfect then, thanks!

apps/src/templates/sectionProgressV2/InviteToV2ProgressModal.jsx
```diff
@@ -0,0 +1,185 @@
...
+} from '@cdo/apps/templates/currentUserRedux';
+import i18n from '@cdo/locale';
+
+import styles from './progress-v2-invitation.module.scss';
+
+const newProgressViewGraphic = require('@cdo/static/teacherDashboard/progressOpenBetaAnnouncementGraphic.png');
+
+function InviteToV2ProgressModal({
+  sectionId,
+
```
lfryemason, line 26:
```suggestion
```
nit

apps/src/templates/sectionProgressV2/InviteToV2ProgressModal.jsx
```diff
@@ -0,0 +1,185 @@
...
+  // from redux
```
lfryemason, line 27: Hmm, I have complicated thoughts about this kind of comment. On one hand I can see how it would be useful to the reader, on the other, these kinds of comments tend to get stale and unreliable very quickly. 

Do you see value in keeping this in the codebase long term or was this better for your initial iteration? 

I think my personal preference is to remove it, especially since there is already a list of props that are from redux in the `connect` function at the bottom of the file. But, I'll leave it up to you as an optional nit.

apps/src/templates/sectionProgressV2/InviteToV2ProgressModal.jsx
```diff
@@ -0,0 +1,185 @@
...
+  setDateProgressTableInvitationDelayed,
+}) {
+  const [invitationOpen, setInvitationOpen] = React.useState(false);
+
+  React.useEffect(() => {
+    const timeSinceInvitationLastDelayed = () => {
+      const startingDate = new Date(dateProgressTableInvitationDelayed);
+      const today = new Date();
+      const differenceInMilliseconds = today.getTime() - startingDate.getTime();
+      const differenceInDays = differenceInMilliseconds / (1000 * 3600 * 24);
```
lfryemason, line 41:
```suggestion
      const differenceInDays = differenceInMilliseconds / MILLISECONDS_IN_ONE_DAY;
```
Optional nit: move this to a constant with a meaningful name.

apps/src/templates/sectionProgressV2/InviteToV2ProgressModal.jsx
```diff
@@ -0,0 +1,185 @@
...
+  dateProgressTableInvitationDelayed,
+  hasSeenProgressTableInvite,
+  setHasSeenProgressTableInvite,
+  setShowProgressTableV2,
+  setDateProgressTableInvitationDelayed,
+}) {
+  const [invitationOpen, setInvitationOpen] = React.useState(false);
+
+  React.useEffect(() => {
+    const timeSinceInvitationLastDelayed = () => {
```
lfryemason, line 37:
```suggestion
    const numDaysSinceInvitationLastDelayed = () => {
```
Optional nit.

lfryemason (approved):
Fantastic! A few optional nits, but I have no problem merging as is!

apps/src/templates/sectionProgressV2/InviteToV2ProgressModal.jsx
kobryan0619: Yeah, this is probably another place where it would be nice, as an org/team to get on the same page.  I see this `from redux` comment kinda a lot in our codebase.  I don't know if it is just because it is the "codebase I grew up with" or if there is a bigger meaning here, but I actually like the separation between "here's the props passsed in from another component" distinction.

PR #58373, opened 2024-05-02
Add Music lab progressSlider component

fisher-alice:
This PR refactors the progress slider of the music lab play view that was added by https://github.com/code-dot-org/code-dot-org/pull/58293 and adds color to the progress bar.

Because `<input>` and `<progress>` elements need to be styled differently for different browsers and the styling was getting complicated, I opted to create a new small component `ProgressSlider` which displays the progress of the song being played with a slider and playhead. A different color is rendered on the bar to represent the percent progress.

https://github.com/code-dot-org/code-dot-org/assets/107423305/4a6dc2f8-8232-4b3d-ab12-ad8bf87de9a2

## PR Checklist:

- [ ] Code is well-commented


apps/src/music/views/ProgressSlider.tsx
```diff
@@ -0,0 +1,42 @@
...
+*/
+const ProgressSlider: React.FunctionComponent<ProgressSliderProps> = ({
+  percentProgress,
+}) => {
+  if (percentProgress < 0) {
+    percentProgress = 0;
+  } else if (percentProgress > 100) {
+    percentProgress = 100;
+  }
+  const sliderWidth = parseInt(moduleStyles.sliderWidth.replace('px', ''));
```
fisher-alice, line 18: Importing `sliderWidth` variable from `progress-slider.module.scss`.

apps/src/music/views/ProgressSlider.tsx
```diff
@@ -0,0 +1,42 @@
...
+  
+*/
+const ProgressSlider: React.FunctionComponent<ProgressSliderProps> = ({
+  percentProgress,
+}) => {
+  if (percentProgress < 0) {
+    percentProgress = 0;
+  } else if (percentProgress > 100) {
+    percentProgress = 100;
+  }
```
thomasoniii: This is fine as is, but for another suggestion:

```
const boundedPercentProgress = Math.min(Math.max(0, percentProgress), 100)
```

I kinda like that it _sorta_ reads in a row that `percentProgress` should be between `0` and `100`.

I'm not sure if this version adds enough clarity or removes enough redundancy to be worth it, though.

thomasoniii (approved):
trivial non-blocking range suggestion, but otherwise lgtm 🚀

apps/src/music/views/ProgressSlider.tsx
fisher-alice: I like it and I also like your variable name. Will update - thanks!

apps/src/music/views/ProgressSlider.tsx
```diff
@@ -0,0 +1,38 @@
...
+interface ProgressSliderProps {
+  percentProgress: number;
+}
+
+/*
+  * A progress slider that displays the current progress of a music lab song.
+  * The slider is a gray bar that fills up with a white color as the song progresses.
+  * A playhead moves along the slider to show the current progress of the song.
+  * @param percentProgress - The percentage of the song that has been played.
+  * @param sliderWidth - The width of the slider in pixels.
```
sanchitmalhotra126: nit: is `sliderWidth` meant to be an optional prop as well? (or if not, can we remove it from this doc?)

sanchitmalhotra126 (approved):
Nice refactor! The new component looks slick, and I can see it being useful outside Music Lab as well!

apps/src/music/views/ProgressSlider.tsx
fisher-alice: Thanks for catching! I originally had it as a prop, but removed and forgot to update comment. Will update!

PR #58382, opened 2024-05-02
feat(email): do not render the 'finish signup' if email was not passed in

stephenliang:
This change ensures that the finish signup page will not be rendered unless the email parameter is passed in for the DCDO post email flow

## Links

- [jira ticket](https://codedotorg.atlassian.net/browse/P20-836)
- [design](https://docs.google.com/document/d/1G9nXdae8Gt2Ea93enIJEdjSnF3xB9Rb-W4pCDnt46vo/edit#heading=h.25gc3zi9dtz9)

## Testing story

1. With DCDO disabled, go to finish sign up page via email, then go to home page, and back to trying to sign up. Should go straight to finish sign up.
2. With DCDO disabled, go to finish sign up page via Google, then go to home page, and back to trying to sign up. Should go straight to finish sign up.
3. With DCDO enabled, go to finish sign up page via email, then go to home page, and back to trying to sign up. Should restart from the first page
4. With DCDO enabled, go to finish sign up page via Google, then go to home page, and back to trying to sign up. Should restart from the first page

## Deployment strategy

This will be deployed via a DCDO feature flag

## PR Checklist:

- [x] Code is well-commented

### DIFF
--- a/dashboard/app/models/concerns/partial_registration.rb
+++ b/dashboard/app/models/concerns/partial_registration.rb
@@ -24,6 +24,14 @@ module PartialRegistration
     end
   end
 
+  def self.can_finish_signup?(params, session)
+    if DCDO.get('student-email-post-enabled', false)
+      params&.dig(:user)&.dig(:email)&.present? && in_progress?(session)
+    else
+      in_progress?(session)
+    end
+  end
+
   def self.in_progress?(session)
     session[SESSION_KEY] && CDO.shared_cache.exist?(session[SESSION_KEY])
   end
--- a/dashboard/app/models/concerns/partial_registration.rb
+++ b/dashboard/app/models/concerns/partial_registration.rb
@@ -26,7 +26,7 @@ module PartialRegistration
 
   def self.can_finish_signup?(params, session)
     if DCDO.get('student-email-post-enabled', false)
-      params&.dig(:user)&.dig(:email)&.present? && in_progress?(session)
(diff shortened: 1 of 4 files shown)


dashboard/app/models/concerns/partial_registration.rb
```diff
@@ -24,6 +24,14 @@ def new_from_partial_registration(session, &block)
     end
   end
 
+  def self.can_finish_signup?(params, session)
+    if DCDO.get('student-email-post-enabled', false)
+      params&.dig(:user)&.dig(:email)&.present? && in_progress?(session)
```
carl-codeorg: Is it safe to do `params&.dig(:user, :email)`, or is there a possibility that params[:user] doesn't have a `dig` method?

stephenliang: I do see that `dig` is available on this class: https://api.rubyonrails.org/classes/ActionController/Parameters.html#method-i-dig

carl-codeorg: Then is it possible to clean this line up by just calling `.dig(:user, :email)` and save the additional `.dig`?

stephenliang: Yep looks like `params` is never nil on the controller (even if no params are passed as it will always have the controller and path), so just got rid of the first dig.

carl-codeorg (approved):
This LGTM

PR #58386, opened 2024-05-02
Fixes Zip Search Glitches

hannahbergam:
I was able to fix the zip search lag!

The issue was that the child component was running a search in the time it took for the parent component to pass in the new 'disabled' state.

Moving the all of the zip logic (including the input box itself) down into the child component directly ties it to the search and allows the child component to manage its own disabled state. I didn't alter any of what is in the input or its label- it was a direct copy/paste.

I added a comment, moved a constant further up in the component and renamed a couple things to hopefully be more clear!

How I know this works:

I no longer see any console logs for failed zip search calls. This means we aren't calling it any extra times! I can also search a bunch of zip codes, delete digits, add them back in, and the dropdown has been populated every time. But I'd love for another developer to checkout this branch and try the same gut checks. 

https://github.com/code-dot-org/code-dot-org/assets/37230822/9640c144-6c5b-41d7-817e-51b561bcb48a

(the experiment url to add is ?enableExperiments=school_association_v2)

## Links

- spec: []()
- jira ticket: https://codedotorg.atlassian.net/browse/ACQ-1858

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/templates/SchoolZipSearch.jsx
+++ b/apps/src/templates/SchoolZipSearch.jsx
@@ -19,13 +19,21 @@ const SEARCH_DEFAULTS = [
 ];
 const ZIP_REGEX = new RegExp(/(^\d{5}$)/);
 
+// Controls the logic and components surrounding a zip input box and its error
+// messaging, the api school search filtered on zip, and the school dropdown
+// that search populates.
 export default function SchoolZipSearch({fieldNames}) {
   const [selectedSchoolNcesId, setSelectedSchoolNcesId] =
     useState(SELECT_A_SCHOOL);
   const [inputManually, setInputManually] = useState(false);
   const [dropdownSchools, setDropdownSchools] = useState([]);
   const [zip, setZip] = useState('');
-  const [isDisabled, setIsDisabled] = useState(true);
+  const [isSchoolDropdownDisabled, setIsSchoolDropdownDisabled] =
+    useState(true);
+
+  const labelClassName = isSchoolDropdownDisabled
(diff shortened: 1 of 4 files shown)


kelbyhawn:
I can check this locally! Can you remind me what flag I need to set to see it?

kelbyhawn (approved):
Woo hoo, exciting! 🎉 

Turns out we can't test this locally unless all schools are seeded locally, so approving based on the video, but if anyone else already has all the schools seeded, feel free to look too!

apps/src/templates/SchoolZipSearch.jsx
```diff
@@ -40,8 +55,16 @@ export default function SchoolZipSearch({fieldNames, zip, disabled}) {
             error.message
           );
         });
+      setIsSchoolDropdownDisabled(false);
```
bethanyaconnor, line 58: I wonder if it makes sense to set this to false when schools are loaded 🤔

bethanyaconnor: Otherwise, I'm wondering if this should be in state at all. Checking a regex could be done on every render

bethanyaconnor (approved):
One thought/question/wonder but looks good!

apps/src/templates/SchoolZipSearch.jsx
hannahbergam: We had a conversation about this offline- to keep from having a duplicate regex check, we opted to leave the structure as-is.

PR #58392, opened 2024-05-02
Scroll to bottom on new message, allow new lines

bencodeorg:
Two unrelated improvements to the chat UI:
- scroll to the bottom of the chat window when a new message is appended (currently, it doesn't scroll, and you have to manually scroll to the bottom)
- don't remove newlines that are included in the model response

## Links

- Jira: https://codedotorg.atlassian.net/browse/LABS-735
- Jira: https://codedotorg.atlassian.net/browse/LABS-741

| Before (newlines stripped) | After (newlines included) |
| - | - |
| image | image |

https://github.com/code-dot-org/code-dot-org/assets/25372625/0185c0a9-106a-4488-8e75-38808e8c25ca

## Testing story

Tested manually (see video/screenshots above). I also noticed that this change preserves newlines in a user message as well (ie, if you hit shift+enter while entering a message, that newline is reflected in the chat workspace)

### DIFF
--- a/apps/src/aichat/views/chatMessage.module.scss
+++ b/apps/src/aichat/views/chatMessage.module.scss
@@ -4,6 +4,7 @@
   padding: 10px 10px;
   border-radius: 10px;
   overflow: hidden;
+  white-space: pre-line;
 }
 
 .userMessage {
--- a/apps/src/aichat/views/ChatWorkspace.tsx
+++ b/apps/src/aichat/views/ChatWorkspace.tsx
@@ -30,9 +30,10 @@ const ChatWorkspace: React.FunctionComponent = () => {
 
   useEffect(() => {
     if (conversationContainerRef.current) {
-      console.log(conversationContainerRef.current.scrollTop);
-      conversationContainerRef.current.scrollTop =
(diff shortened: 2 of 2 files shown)


apps/src/aichat/views/chatMessage.module.scss
```diff
@@ -4,6 +4,7 @@
   padding: 10px 10px;
   border-radius: 10px;
   overflow: hidden;
+  white-space: pre-line;
```
fisher-alice, line 7: Just curious how you decided to select `pre-line` vs the other possible options.(e.g. `pre-wrap`).

fisher-alice (approved):
Cool! Just a left a question for my own understanding. Thanks!

apps/src/aichat/views/chatMessage.module.scss
bencodeorg: Honestly I wasn't sure which was better :) One concrete thing is that I noticed the bot responses seem like they often (always?) have a space at the start, which `pre-line` strips (and `pre-wrap` does not).

image

PR #58393, opened 2024-05-03
Enable choice levels for lab2

sanchitmalhotra126:
Make sublevels of bubble choice levels work with Lab2. The main work here is making sure the sublevel level ID/position is passed around to the right places, so lab2 can request the correct level properties, post milestones for the correct level, and navigate to the right place on clicking continue.

~~There are definitely a few things I'm feeling iffy about which I've tagged below, so I'd love feedback on the approach if folks have any thoughts.~~ Edit: after feedback and finding some edge cases, I've revised my original approach. More details included below as well as in comments on the PR.

### Details
The primary change here is actually pretty straightforward: set `state.progress.currentLevelId` to the current level OR current sublevel's ID (instead of only the top level ID, when viewing a sublevel). The issue is that this value is used in a bunch of places throughout the code base, so I tried to do my due diligence in making sure those places weren't affected, and made changes if so. Most areas required no or minimal changes, and any significant changes are called out in the PR. Here's a roundup of every use I checked:
- Lab2 code (Aichat/Dance Lab2/Music/Lab2 core): directly benefit from this change because they depend on the level ID referring to the actual level being presented.
- [StudioApp.js](https://github.com/code-dot-org/code-dot-org/blob/315a6c4c9afd97e1c19bf69bcd3d81beb3a2b823/apps/src/StudioApp.js#L575): only used on challenge levels, and after speaking with curriculum, it does not seem like there is a way for a sublevel to also be a challenge level. Even then, this value is used to check level progress, which will return the correct progress for the level or sublevel.
- [HeaderPopup.jsx](https://github.com/code-dot-org/code-dot-org/blob/315a6c4c9afd97e1c19bf69bcd3d81beb3a2b823/apps/src/code-studio/components/header/HeaderPopup.jsx#L37): used to fetch progress data from the server, but the value of `currentLevelId` is only used to check if we're on an overview page, so it works in either case.
- [LessonProgress.jsx](https://github.com/code-dot-org/code-dot-org/blob/315a6c4c9afd97e1c19bf69bcd3d81beb3a2b823/apps/src/code-studio/components/progress/LessonProgress.jsx#L59): used to determine if the level has changed, which works the same with this change.
- SummaryTopLinks.jsx/SummaryResponses.jsx: summaries are not available for sublevels as far as I can tell, so this is unaffected.
- progressRedux/progressReduxSelectors/progressHelpers: the majority of the changes have been made here to ensure that sublevels are being parsed and processed correctly, and that the `isCurrentLevel` flag is assigned correctly if the current level is a sublevel.
- [teacherPanelRedux.js](https://github.com/code-dot-org/code-dot-org/blob/315a6c4c9afd97e1c19bf69bcd3d81beb3a2b823/apps/src/code-studio/teacherPanelRedux.js#L79): **minor change made here** because the `/api/teacher_panel_progress/` API does not seem to work properly with a sublevel ID. I retained the existing behavior here, so this API will always use the top level ID, not the sublevel ID.
- ContainedLevelResetButton.jsx: no change made here but this PR actually seems to **fix a bug** where contained levels that are also sublevels had an issue with resetting their "Delete" button.

### Alternate Approaches
Given that this value is used in so many places, why not introduce a separate value to track the sublevel index/ID and make this more of a scoped change? I started down this path initially but quickly realized that I'd have to add special cases/branching in quite a few places outside of Lab2 to make sure the sublevel ID was being stored and updated correctly. Rather than doubling down on a temporary solution with a lot of special case code, I felt it was better to explore solving this problem in a more consistent way. I'm hopeful that this solution makes the progress system a little more straightforward, rather than introducing more complication.

## Links

https://codedotorg.atlassian.net/browse/LABS-739

## Testing story

Tested choice levels in a Gen AI draft progression (see http://localhost-studio.code.org:3000/s/gen-ai-foundations-draft/lessons/2/levels/11/) as well as a few non-lab2 choice levels to make sure nothing broke. Also tested all the affected areas mentioned above and confirmed existing functionality did not change as far as I could verify.

TODO: Will create a Lab2 bubble choice level in allthethings and add to the existing bubble choice eyes test.


apps/src/code-studio/progressRedux.ts
```diff
@@ -453,21 +455,6 @@ const userProgressFromServer = (
   });
 };
 
-/**
```
sanchitmalhotra126, line 456: Some minor cleanup here: previously, we were fetching the current level directly out of `state.progress.lessons` without performing `processLevel` like we do in some other places (in `progressReduxSelectors` for example). This confused me a bit at first because there are some slight differences in the level structure before and after processing; before, the level's `id` is the script level ID, while the level ID is stored in its `ids` array and/or `activeId` field. After processing, we assign the `id` to the `activeId` if present. To be consistent with the rest of progressRedux and just to make things more clear, I opted to use a function from `progressReduxSelectors` instead which will perform `processLevel`, and add an explicit `scriptLevelId` field to the processed level, so we know which ID we're dealing with.

apps/src/code-studio/progressReduxSelectors.js
```diff
@@ -117,6 +123,25 @@ export const getLevelPropertiesPath = state => {
   }
 };
 
+/**
+ * Get the ID of the current level, or sublevel if we're on a sublevel.
+ */
+export const getCurrentLevelOrSublevelId = state => {
```
sanchitmalhotra126: This is the main thing I'm stumped on; currently, `state.progress.currentLevelId` refers to the current non-sublevel ID. So if we're on a sublevel of a bubble choice level, that ID will refer to the parent bubble choice level and not the sublevel. Initially I thought it would make sense to update `state.progress.currentLevelId` to the current level or sublevel ID, but we use this field in a bunch of other places so I'm wary of breaking existing behavior. For Lab2, we need the level ID of the specific sublevel we're on in order to fetch the right level properties, etc, so we need some way to get that level's ID (which is what this selector does). It doesn't feel great, because it's ambiguous what `currentLevelId` even means then, but I thought this might be safer for now. Curious what folks think (cc @molly-moen @breville)

dashboard/app/controllers/script_levels_controller.rb
```diff
@@ -219,7 +219,7 @@ def level_properties
     @script_level = ScriptLevelsController.get_script_level(@script, params)
     raise ActiveRecord::RecordNotFound unless @script_level
 
-    @level = @script_level.level
+    @level = select_level
```
sanchitmalhotra126, line 222: This will automatically select the current level or sublevel based on the supplied params

dashboard/app/models/script_level.rb
```diff
@@ -319,7 +319,7 @@ def summarize(include_prev_next = true, for_edit: false, user_id: nil)
       end
 
       summary = {
-        id: id,
+        id: id.to_s,
```
sanchitmalhotra126, line 322: Updated this to a string since I noticed that there was a dedicated effort some time back to make sure all level IDs were passed to the client as strings. We added this ID for lab2 purposes more recently, but because it wasn't a string, there was some inconsistency in how it was handled on the client. Changed to a string just to be consistent and safer.

apps/src/code-studio/progress.js
```diff
@@ -359,6 +365,7 @@ function initializeStoreWithProgress(
       isFullProgress: isFullProgress,
       isLessonExtras: isLessonExtras,
       currentPageNumber: currentPageNumber,
+      currentSublevelPosition: currentSublevelPosition,
```
thomasoniii: non-blocking, but could we crunch these down to the shorthand properties where possible?

e.g., `currentSublevelPosition` instead of `currentSublevelPosition : currentSublevelPosition`.

I think there's a linter rule we could enable to do that automatically as well...

thomasoniii (approved):
lgtm 🚀

apps/src/code-studio/progressReduxSelectors.js
molly-moen: How much do the functionality of a level and a sublevel overlap? Mostly I'm wondering if we provide a sublevel id to the apis that we are currently using for "regular" levels, will any break? Do sublevel queries always happen with the parent level and sublevel ids or do they just use the sublevel?

davidsbailey:
There is some end-to-end test coverage of progress on these types of levels in `bubble_choice.feature`. my recommendation would be to extend those tests to cover a lab2 level to make sure this is working and stays working.

apps/src/code-studio/progressReduxSelectors.js
sanchitmalhotra126: Yeah I believe since sublevel IDs are valid IDs, any APIs using them should work as expected (as in, they would be target the specific sublevel)? I think the only places where the parent level and sublevel are required together is for script level APIs like `/s/<script>/lessons/<lesson>/levels/<level>/sublevel/<sublevel>/level_properties`. 

I did come across another issue though which is that if you open the unit overview dropdown from the header, it makes a server to call to re-fetch progress which resets the current level ID (without any sublevel) and causes Lab2 to break again. So I'm now more inclined try making `currentLevelId` refer to the actual sublevel ID and see how that goes.

molly-moen: yeah I was wondering if we should add a `parentLevelId` field in redux that is null if this isn't a sublevel, and has the parent id otherwise.

sanchitmalhotra126: Oh yeah, that's a good idea. Maybe the current script level ID (which would be the parent level ID) in addition to the current level ID (which would be the sublevel's ID if on a sublevel, or parent level's level ID)

apps/src/code-studio/progress.js
sanchitmalhotra126: Oh yeah, I kept it the same for consistency but would be nice to use shorthands for these. Will take a look

sanchitmalhotra126:
Moving this back to draft while I revise the approach (setting progress.currentLevelId to actual level/sublevel ID instead of creating a separate selector for sublevel ID)

apps/src/code-studio/teacherPanelRedux.js
```diff
@@ -75,9 +76,11 @@ const getLevelProgressQueryParams = state => {
       script_id: state.progress.scriptId,
     };
   } else {
+    const currentLevel = getCurrentLevel(state);
+    // If we are on a sublevel, use the level ID of the main (parent) level
     return {
       script_id: state.progress.scriptId,
-      level_id: state.progress.currentLevelId,
+      level_id: currentLevel.parentLevelId || state.progress.currentLevelId,
```
sanchitmalhotra126: The `level_id` value here is used by the `api/teacher_panel_progress` API, which doesn't work with a sublevel ID (because the API tries to look up the associated script level, which doesn't exist for sublevels). The change here retains the original behavior to use the top-level/parent level ID if we're on a sublevel.

sanchitmalhotra126:
Hi folks, thanks for your patience - this is ready for review again. I've revised the approach since first posting this and captured my thinking in the description.

sanchitmalhotra126:
> There is some end-to-end test coverage of progress on these types of levels in `bubble_choice.feature`. my recommendation would be to extend those tests to cover a lab2 level to make sure this is working and stays working.

I'm planning on doing this in a separate PR since it will require creating a new level progression with a bubble choice level that has lab2 sublevels. If it's alright, I can plan to merge that as a fast follow and/or branch off of this PR and merge together.

molly-moen (approved):
Looks good to me!

apps/src/types/progressTypes.ts
```diff
@@ -50,16 +50,19 @@ export interface Level {
   display_as_unplugged: boolean;
   freePlay: boolean;
   icon: string | null;
-  id: number;
+  id: string;
   ids: string[];
   inactiveIds: string[];
   is_concept_level: boolean;
   kind: string;
+  levelNumber: number;
```
fisher-alice, line 58: Could you explain what `levelNumber` is and compare it with position? thanks!

apps/src/types/progressTypes.ts
```diff
@@ -50,16 +50,19 @@ export interface Level {
...
   position: number;
   title: number;
   url: string;
   path: string;
+  scriptLevelId: string;
   status?: string;
+  sublevels?: Level[];
```
fisher-alice, line 65: Is the significant difference between `Level` and `LevelWithProgress` that `LevelWithProgress` always tracks `status` for progress?
And just checking that it's possible to have a level with sublevels for which neither the level nor sublevel track progress?

apps/src/code-studio/progressReduxSelectors.js
```diff
@@ -103,12 +103,25 @@ export const getLevelPropertiesPath = state => {
     const lessonPosition = state.progress.lessons?.find(
       lesson => lesson.id === state.progress.currentLessonId
     ).relative_position;
-    const levelNumber =
-      levelsForLessonId(
+
```
fisher-alice, line 106: Not related to your change, but do you happen to know why we use `relative_position` above instead of `position`?

apps/src/code-studio/progressRedux.ts
fisher-alice: Nice cleanup!

apps/src/code-studio/progressRedux.ts
```diff
@@ -333,15 +363,22 @@ export function sendSuccessReport(appType: string): ProgressThunkAction {
     if (!state.currentLessonId || !levelId) {
       return;
     }
-    const currentLevel = getLevelById(
-      state.lessons,
-      state.currentLessonId,
-      levelId
-    );
+    const currentLevel = getCurrentLevel(getState());
```
fisher-alice, line 366: I think this function `sendSuccessReport` is only used by lab2. But other functions in this file are used across labs. Would it be okay to tag in comments functionality that is lab2 specific in this file?

apps/src/code-studio/progressRedux.ts
```diff
@@ -333,15 +363,22 @@ export function sendSuccessReport(appType: string): ProgressThunkAction {
...
+    if (currentLevel.parentLevelId) {
+      const parentLevel = levelById(
+        state,
+        state.currentLessonId,
+        currentLevel.parentLevelId
+      );
+      scriptLevelId = parentLevel.scriptLevelId;
+    } else {
+      scriptLevelId = currentLevel.scriptLevelId;
+    }
```
fisher-alice, line 381: Out of curiosity and not important, do you happen to know where the sublevel posts a milestone? I see the progress bubble does turn green on the parent level page.

apps/src/templates/progress/progressHelpers.js
```diff
@@ -267,9 +267,10 @@ export function lessonProgressForSection(sectionLevelProgress, lessons) {
  * script.lessons.levels) contains more data than we need. This parses the parts
  * we care about to conform to our `levelType` object.
  */
-export const processedLevel = level => {
+export const processedLevel = (level, parentLevelId) => {
+  const id = level.activeId || level.id;
```
fisher-alice, line 271: Not related to your PR, but just wrapping my head around `Level`. https://github.com/code-dot-org/code-dot-org/blob/staging/apps/src/types/progressTypes.ts#L46-L64
Is there an `activeId` in the case that a user has had multiple attempts of a level? Thus we store `ids` and `inactiveIds`.

apps/src/templates/progress/progressHelpers.js
```diff
@@ -292,9 +296,13 @@ export const processedLevel = level => {
       typeof level.page_number !== 'undefined'
         ? level.page_number
         : PUZZLE_PAGE_NONE,
+    // Script level ID doesn't apply for sublevels. Set to undefined if we have a parent level.
```
fisher-alice, line 299: Thanks for the comment!

dashboard/app/controllers/script_levels_controller.rb
fisher-alice: Is this what fixed the exisiting bug with contained levels that are also sublevels?

dashboard/app/models/levels/level.rb
```diff
@@ -525,7 +525,8 @@ def summarize
       name: name,
       display_name: display_name,
       is_validated: validated?,
-      can_have_feedback: can_have_feedback?
+      can_have_feedback: can_have_feedback?,
+      uses_lab2: uses_lab2?,
```
fisher-alice, line 529: Could you explain where this is used in frontend?

fisher-alice (approved):
Thanks for the detailed PR description and comments. Learned a lot and left a handful of questions for my understanding.

apps/src/code-studio/browserNavigation.js
```diff
@@ -37,11 +39,18 @@ export function setupNavigationHandler(lessonData) {
       return;
     }
 
-    const levelIndex = levelNumber - 1;
-    const levelId = lessonData.levels[levelIndex].activeId;
+    let sublevelNumber;
+    if (values.length > sublevelNumberStringIndex) {
+      sublevelNumber = Number(values[sublevelNumberStringIndex]);
+      if (!Number.isInteger(sublevelNumber) || sublevelNumber <= 0) {
```
etaderhold, line 45: In the above three cases where the URL bits don't pass the validation we bail out and return, but here if there's a provided sublevel number that isn't a positive integer we proceed as if a sublevel number was not provided.

I just want to call out that difference in behavior between the cases and make sure you've thought through anything that could go wrong here.

davidsbailey:
> > There is some end-to-end test coverage of progress on these types of levels in `bubble_choice.feature`. my recommendation would be to extend those tests to cover a lab2 level to make sure this is working and stays working.
> 
> I'm planning on doing this in a separate PR since it will require creating a new level progression with a bubble choice level that has lab2 sublevels. If it's alright, I can plan to merge that as a fast follow and/or branch off of this PR and merge together.

OK - the end-to-end tests are what I would look to in order to see that this is working in all the right scenarios. Do we already have users using lab2 levels within bubble choice levels? If not then it seems ok to defer until another time before people start using this feature.

@etaderhold are you available to review the parts that pertain to progress and teacher dashboard? I'm afraid my knowledge may be out of date here.

sanchitmalhotra126:
> > > There is some end-to-end test coverage of progress on these types of levels in `bubble_choice.feature`. my recommendation would be to extend those tests to cover a lab2 level to make sure this is working and stays working.
> > 
> > 
> > I'm planning on doing this in a separate PR since it will require creating a new level progression with a bubble choice level that has lab2 sublevels. If it's alright, I can plan to merge that as a fast follow and/or branch off of this PR and merge together.
> 
> OK - the end-to-end tests are what I would look to in order to see that this is working in all the right scenarios. Do we already have users using lab2 levels within bubble choice levels? If not then it seems ok to defer until another time before people start using this feature.

We don't yet, but they will be as part of the upcoming Gen AI pilot (https://levelbuilder-studio.code.org/s/gen-ai-customizing-pilot-v1/lessons/2/levels/9). It'll be available for teachers next week but with the caveat that some features (like bubble choice levels) may not work until 5/20, so we have a little bit of leeway before actual usage starts. The existing end-to-end tests (UI/eyes/integration) are passing with these changes.

apps/src/code-studio/progressRedux.ts
```diff
@@ -313,6 +314,35 @@ export function navigateToLevelId(levelId: string): ProgressThunkAction {
...
+    if (
+      sublevelIndex !== undefined &&
+      level.sublevels &&
+      sublevelIndex < level.sublevels.length
+    ) {
+      const sublevel = level.sublevels[sublevelIndex];
+      newLevelId = sublevel.id;
+    }
+
+    dispatch(setCurrentLevelId(newLevelId));
```
etaderhold: I generally try to avoid using the `let` keyword. It adds some mental overhead for reviewers and future readers of the code trying to understand that the variable actually had something assigned to it in every case where it should, and trying to understand where the variable is reassigned.

In this case avoiding all that seems easy enough.

```suggestion
    if (
      sublevelIndex !== undefined &&
      level.sublevels &&
      sublevelIndex < level.sublevels.length
    ) {
      dispatch(setCurrentLevelId(level.sublevels[sublevelIndex].id));
    } else {
      dispatch(setCurrentLevelId(level.id));
    }
```

apps/src/code-studio/progressReduxSelectors.js
```diff
@@ -211,21 +224,42 @@ export const levelsForLessonId = (state, lessonId) => {
...
+      return level;
+    }
+
+    if (level.sublevels) {
+      for (const sublevel of level.sublevels) {
+        if (sublevel.isCurrentLevel) {
+          return sublevel;
+        }
+      }
+    }
```
etaderhold:
```suggestion
  return getCurrentLevels(state)
    .flatMap(level => [level, ...level.sublevels])
    .find(level => level.isCurrentLevel)
```

apps/src/code-studio/progressReduxSelectors.js
```diff
@@ -211,21 +224,42 @@ export const levelsForLessonId = (state, lessonId) => {
...
+      return level;
+    }
+
+    if (level.sublevels) {
+      for (const sublevel of level.sublevels) {
+        if (sublevel.id === levelId) {
+          return sublevel;
+        }
+      }
+    }
```
etaderhold:
```suggestion
  return levelsForLessonId(state, lessonId)
    .flatMap(level => [level, ...level.sublevels])
    .find(level => level.id === levelId);
```

etaderhold (approved):
As I mentioned in our conversation earlier this week there's potentially a lot of surface area for this change and I'm not familiar with all of it, nor is it likely that anyone is. We therefore need to rely pretty heavily on our test coverage for existing functionality, and if some bug makes it through that's an excellent opportunity to add a regression test.

apps/src/code-studio/progressReduxSelectors.js
fisher-alice: Cool! Or as my daughter would say, fire.

apps/src/code-studio/progressRedux.ts
fisher-alice: Personally, I think that the current implementation may be easier to read. For me, it compartmentalizes the assignment of `newLevelId` and then what is done with the variable.

etaderhold: For me I see a `let` and alarm bells just start going off in my head. If a mutable variable is needed, something complicated/delicate/error-prone is likely to be coming up next, and I need to be extra careful when reading or touching that code. 

I do see your point about giving `newLevelId` a name and that improves readability. How about this? 
```suggestion
    if (
      sublevelIndex !== undefined &&
      level.sublevels &&
      sublevelIndex < level.sublevels.length
    ) {
      const newLevelId = level.sublevels[sublevelIndex].id;
      dispatch(setCurrentLevelId(newLevelId));
    } else {
      const newLevelId = level.id;
      dispatch(setCurrentLevelId(newLevelId));
    }

```

fisher-alice: Nice! And thanks for taking time looking at a student-labs PR. Appreciate getting different perspectives.

apps/src/code-studio/browserNavigation.js
sanchitmalhotra126: Yep good call, I think this is fine because we'll just default to the parent level is the sublevel number is somehow invalid.

apps/src/code-studio/progressRedux.ts
sanchitmalhotra126: Looks fine to me, will update!

apps/src/code-studio/progressReduxSelectors.js
sanchitmalhotra126: Good call!

apps/src/types/progressTypes.ts
sanchitmalhotra126: I think it's very similar, and I don't think `position` is actually assigned - I believe this type is meant to match the output of `processedLevel` here: https://github.com/code-dot-org/code-dot-org/blob/5f3acce40a0b26755c9bd762443ea391e2aaa280/apps/src/templates/progress/progressHelpers.js#L271

apps/src/types/progressTypes.ts
sanchitmalhotra126: The `LevelWithProgress` type i believe comes from the result of this function: https://github.com/code-dot-org/code-dot-org/blob/5f3acce40a0b26755c9bd762443ea391e2aaa280/apps/src/code-studio/progressReduxSelectors.js#L126
I think generally some cleanup that can happen around here with these types and associated code to make things a bit clearer. A level and associated sublevels will only have progress info if they've gone through the above function.

apps/src/code-studio/progressReduxSelectors.js
sanchitmalhotra126: Not sure :) it seems like `relative_position` refers to the `relative_position` in the model while `position` refers to `absolute_position`. I found this comment, but don't have much context beyond that: https://github.com/code-dot-org/code-dot-org/blob/5f3acce40a0b26755c9bd762443ea391e2aaa280/dashboard/app/models/lesson.rb#L76

apps/src/code-studio/progressRedux.ts
sanchitmalhotra126: 👍 done

apps/src/code-studio/progressRedux.ts
sanchitmalhotra126: My understanding is that progress is stored separately for the level and the sublevel, but there's logic that displays the level's bubble as green if one of its sublevels has been completed. The success report is sent to the activities_controller#milestone with the script level ID of the parent level and the level ID of the sublevel.

apps/src/templates/progress/progressHelpers.js
sanchitmalhotra126: Yeah this was confusing for me too :) not 100% sure what the multiple IDs refer to, but from I've seen, these levels typically just have one id in the `ids` field. Worth calling out that the "level" model passed into this function typically represents a script level, and this function converts it into a type that matches the `Level` interface in progressTypes. It's also a bit confusing because types are currently split between [progressTypes.ts](https://github.com/code-dot-org/code-dot-org/blob/5f3acce40a0b26755c9bd762443ea391e2aaa280/apps/src/types/progressTypes.ts#L4) and [progressTypes.js](https://github.com/code-dot-org/code-dot-org/blob/5f3acce40a0b26755c9bd762443ea391e2aaa280/apps/src/templates/progress/progressTypes.js#L4) so I think we need to do some cleanup to reconcile these with the actual shape used in these files.

dashboard/app/controllers/script_levels_controller.rb
sanchitmalhotra126: Nope that was actually just automatically fixed by making `state.progress.currentLevelId` refer to the sublevel ID - the issue was that the ContainedLevelResetButton (the delete button under a contained level submission) uses `state.progress.currentLevelId` to determine if the student has attempted that level: https://github.com/code-dot-org/code-dot-org/blob/5f3acce40a0b26755c9bd762443ea391e2aaa280/apps/src/templates/instructions/ContainedLevelResetButton.jsx#L81-L82
Previously though, when on a sublevel that was a contained level, this ID was referring to the parent level, it was pulling progress from the parent bubble choice level. And the parent level will register an attempt if any of its sublevels have been attempted. So for example if there was a bubble choice levels with two contained sublevels, if a student had attempted the first sublevel and then navigated to the second sublevel, the delete button would be enabled even though they hadn't attempted that level. With this change, when they go to the second sublevel, we will correctly determine that there has not been any attempt made for this sublevel, and the delete button will be disabled.

fisher-alice: Cool - thanks for the explanation!

apps/src/types/progressTypes.ts
fisher-alice: Thanks! Interesting that `level.title` is used to assign to `levelNumber` and `bubbleText`.

apps/src/code-studio/progressReduxSelectors.js
fisher-alice: Thank you!

apps/src/code-studio/progressRedux.ts
fisher-alice: Thanks - that makes sense!

apps/src/templates/progress/progressHelpers.js
fisher-alice: Did a little digging, and see that `activeId` seems to be referring to the active `script_level` variant.  https://codedotorg.atlassian.net/browse/PLAT-371  includes plan to add new active variant to a script level. Interesting [Slack thread](https://codedotorg.slack.com/archives/C02EEGWLHR8/p1641330249083000?thread_ts=1641321218.080800&cid=C02EEGWLHR8) in which @davidsbailey discusses how we want to get rid of level variants, but 'need other ideas on how to support fixing a badly broken level in a unit that already has progress in it.'

PR #58401, opened 2024-05-03
[Labs-715][GenAI] Continue and Update buttons pushed off screen

levadadenys:
### [Labs-715][GenAI] Continue and Update buttons pushed off screen

* Update  min width in order to prevent ui to be jumpy on when screen width is small

Fixed the issue by setting the `min-width` to every column and adding `overflow: auto`, so that in case the screen width is too small - user will be able to just scroll horizontally. 200px is a value got from visual comparison (screen collapses when columns width is smaller than 196.7px, we use x8 factor for design system, so used the closes x8 factor here, which is 200px )

Before:
scrolling-bug

After:

https://github.com/code-dot-org/code-dot-org/assets/22244040/3db6b0b1-4e2d-4856-8031-b63dd6de559d

## Links

- jira ticket: [LABS-715](https://codedotorg.atlassian.net/browse/LABS-715)

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/aichat/views/aichatView.module.scss
+++ b/apps/src/aichat/views/aichatView.module.scss
@@ -19,15 +19,18 @@
   .instructionsArea {
     width: 30%;
     height: 100%;
+    min-width: 200px;
   }
 
   .chatWorkspaceArea {
     flex: 1;
+    min-width: 200px;
   }
 
   .customizationArea {
     height: 100%;
     width: 30%;
+    min-width: 200px;
   }
 
   .presentationArea {
@@ -41,6 +44,7 @@
   display: flex;
   flex: 1;
   gap: 2px;
+  overflow: auto;
 }
 
 .clearButton {
--- a/apps/src/aichat/views/aichatView.module.scss
+++ b/apps/src/aichat/views/aichatView.module.scss
@@ -1,5 +1,11 @@
 @import 'color.scss';
 
+//200px is a value got from visual comparison (screen collapses when columns width is smaller than 196.7px,
+// we use x8 factor for design system, so used the closes x8 factor here, which is 200px).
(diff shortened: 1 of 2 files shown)


thomasoniii (approved):
lgtm 🚀

apps/src/aichat/views/aichatView.module.scss
```diff
@@ -19,15 +19,18 @@
   .instructionsArea {
     width: 30%;
     height: 100%;
+    min-width: 200px;
```
sanchitmalhotra126: Mind adding a comment similar to your PR description around why this value was chosen? Optional - also, create a sass constant for this value?

sanchitmalhotra126 (approved):
Nice solution!

apps/src/aichat/views/aichatView.module.scss
```diff
@@ -19,15 +19,18 @@
   .instructionsArea {
     width: 30%;
     height: 100%;
```
fisher-alice, line 27: I wonder if we should also prevent buttons from disappearing if user resizes window vertically.

levadadenys: It's also included in this PR. If user resizes window vertically and height is too small - scroll for y axis will appear

levadadenys: https://github.com/code-dot-org/code-dot-org/assets/22244040/76bb45e2-94bd-482e-b0c5-bfba56522765

fisher-alice: Nice! Thanks for the demo.

fisher-alice (approved):
🎉

PR #58409, opened 2024-05-03
Create CloudWatch Alarms for the new Session Store Redis Cluster

Hamms:
Specifically, create alarms for four important metrics: `EngineCPUUtilization`, `Evictions`, `NetworkBytesIn`, and `NetworkBytesOut`. I largely based my choice of metrics on [this blog post](https://aws.amazon.com/blogs/database/monitoring-best-practices-with-amazon-elasticache-for-redis-using-amazon-cloudwatch/), which is referenced in [this AWS documentation](https://docs.aws.amazon.com/AmazonElastiCache/latest/red-ug/CacheMetrics.WhichShouldIMonitor.html). Three sets of metrics referenced in that blog post which I did not include in this PR are those related to Current Connections, Replication and Latency; these metrics are more important for readwrite-heavy workloads, whereas our use case is significantly more storage-heavy. Please let me know if you think any of these are important enough to include.

My thinking on the selected limits for each metric is as follows:

- `EngineCPUUtilization`: the docs recommend keeping this below 90%, so I picked that to alarm on. We consistently stay under 6% utilization, so we could also consider adding a low-urgency alarm at a much lower threshold, something like 20%, to warn us of unexpected but still manageable behavior. I've left that out of this PR for now, but let me know if you think it's worth adding.
- `Evictions`: We unfortunately don't have any data for this one yet, so the current limit is highly speculative, and should be adjusted once the feature has been live long enough to see our regular eviction rate (should take about two months to begin evicting old sessions). I've set this alarm to LowPriority rather than Urgent for now, and selected a threshold we should only hit once we start evicting as many sessions as we create.
- `NetworkBytesIn` and `NetworkBytesOut`: The `cache.r7g.xlarge` nodes we're using have a 12.5Gbps limit, and we generally have approximately two bytes out for every one byte in, so I set the threshold for in at 25% of the total and threshold for out at 50% of the total, so even if both max out at once we still have a 25% buffer. Note that like with `EngineCPUUtilization` we usually stay significantly under this maximum (~80Mbps in, ~160Mbps out), and so could consider adding a LowPriority alarm at a lower threshold as an early warning.

## Links

- https://docs.aws.amazon.com/AmazonElastiCache/latest/red-ug/CacheMetrics.WhichShouldIMonitor.html
- https://aws.amazon.com/blogs/database/monitoring-best-practices-with-amazon-elasticache-for-redis-using-amazon-cloudwatch/

## Testing story

Tested locally with both `stack:lint` and `stack:validate`, and verified that the template generates as expected:

`bundle exec rake stack:lint RAILS_ENV=production VERBOSE=1 | view -`
`bundle exec rake stack:validate RAILS_ENV=production VERBOSE=1 | view -`

Also tested on an adhoc, by temporarily modifying the base cloudformation template to create these resources even in environments without frontends. Verified that the resources are created, configured, and referenced as expected.

## Deployment strategy

Change should be deployable via a standard DTP, although as with all cloudformation template changes that include production-specific logic we should plan to be available for a revert if necessary.

## Follow-up work

I plan to update some of our documentation to include suggestions for testing cloudformation template changes.

## PR Checklist:

- [ ] Code is well-commented


aws/cloudformation/components/session_store.yml.erb
```diff
@@ -15,11 +18,163 @@
...
+  # https://aws.amazon.com/blogs/database/monitoring-best-practices-with-amazon-elasticache-for-redis-using-amazon-cloudwatch/
+  SessionStoreCPUUsageAlarm:
+    Type: AWS::CloudWatch::Alarm
+    DependsOn: SessionStoreGroup
+    Properties:
+      AlarmName: <%="Redis Session Store CPU Usage Too High (#{stack_name})" %>
+      AlarmDescription: AWS Documentation recommends maintaining
+        EngineCPUUtilization below 90%. Note that we use EngineCPUUtilization
+        rather than CPUUtilization because the cache.r7g.xlarge nodes we're
+        using have 4vCPUs
```
cat5inthecradle, line 44: is this the correct multiline syntax? The lint or validate rake tasks should be sufficient to confirm.

aws/cloudformation/components/session_store.yml.erb
```diff
@@ -15,11 +18,163 @@
...
+      ActionsEnabled: true
+      AlarmActions:
+        - !Sub "arn:aws:sns:${AWS::Region}:${AWS::AccountId}:CDO-Urgent"
```
cat5inthecradle: This pages the DOTD. Alternatively we could just page Infra and use `!Ref InfraUrgent` (see alarms.yml.erb)

```suggestion
        - !Ref InfraUrgent
```

cat5inthecradle (commented):
My confidence in this deploying correctly is low - simply because I never expect fresh cloudformation configuration to work on the first try. If we can test this in an adhoc, that would be awesome, but we should definitely run `lint` and `validate` first. We should also just expect the cloudformation update step to fail during the DTP because life isn't fair.

If we're willing to stand by to quickly revert or fix forward, I'm fine giving it a go.

aws/cloudformation/components/session_store.yml.erb
Hamms: I verified with `lint`, `validate`, and also on an adhoc; see "Testing story" in the PR description

aws/cloudformation/components/session_store.yml.erb
Hamms: Updated! Also changed the `CDO-LowPriority` one to `InfraLowPriority`

unlox775-code-dot-org (approved):

PR #58411, opened 2024-05-03
retry aiproxy requests on 503 and 504

davidsbailey:
## Description

Starts https://codedotorg.atlassian.net/browse/AITT-464. The sequence or PRs looks like this:
* https://github.com/code-dot-org/code-dot-org/pull/58354
* https://github.com/code-dot-org/code-dot-org/pull/58411 (this PR)
* https://github.com/code-dot-org/aiproxy/pull/74 

this PR handles 503 and 504 requests from OpenAI, and logs metrics and firehose rather than honeybadger in those cases. 

### response codes

Here is some discussion of why 503 and 504 are reasonable to return in this case:
* https://stackoverflow.com/questions/18709834/http-status-424-or-500-for-error-on-external-dependency
* https://stackoverflow.com/questions/25398364/http-status-code-for-external-dependency-error/25398605#25398605

One concern with using these response codes is that the same codes might also be returned due to other infrastructure problems. To address this, I've added an "agent" dimension to each metric and firehose event, which will be "openai" when the problem is due to an openai error and "none" when the problem is between dashboard and aiproxy.

### timeouts

This PR and https://github.com/code-dot-org/aiproxy/pull/74 also change some timeouts, in order to: (1) to avoid the problem where the active job gives up on the request a moment before aiproxy returns a 504, and (2) make the timeouts a bit longer in hopes of reducing how often they happen in the first place.

Together with https://github.com/code-dot-org/aiproxy/pull/74 , we have the following timeouts, which should ensure that openai timeouts are returned to active job as 504 before any other timeouts are reached:
* 150s: aiproxy request to openai
* 165s: active job request to aiproxy
* 180s: aiproxy load balancer. see https://github.com/code-dot-org/aiproxy/blob/d53e627366fb4fdf209d86128ecdaec8c3b2763e/cicd/3-app/aiproxy/template.yml#L65

## command line output

Performing the job from rails console to confirm that firehose logging is working, with an artificially low timeout:
```
[development] dashboard > EvaluateRubricJob.perform_now(user_id: 219, requester_id: 1, script_level_id: 160880)
...
Stopped retrying EvaluateRubricJob due to a Net::ReadTimeout, which reoccurred on 2 attempts.
Skipped sending records to analysis-events:
["{\"study\":\"ai-rubrics\",\"study_group\":\"v0\",\"event\":\"retry-on-timeout\",\"data_string\":\"Net::ReadTimeout: Net::ReadTimeout with #\\u003cTCPSocket:(closed)\\u003e\",\"data_json\":\"{\\\"user_id\\\":219,\\\"requester_id\\\":1,\\\"script_level_id\\\":160880,\\\"script_name\\\":\\\"csd3-2023\\\",\\\"lesson_number\\\":11,\\\"level_name\\\":\\\"CSD U3 Sprites scene challenge_2023\\\"}\",\"created_at\":\"2024-05-06T16:28:25.476+00:00\",\"environment\":\"development\",\"device\":\"\\\"server-side\\\"\"}"]
```

## Testing story

- new unit tests
- end-to-end testing with local aiproxy:
  - since it is hard to simulate openai returning 500, I modified my local aiproxy with https://github.com/code-dot-org/aiproxy/pull/74 to unconditionally `return 'openai error', 503`
  - to test the 504 pathway, I temporarily reduced the `OPENAI_API_TIMEOUT` in aiproxy to 1 second.
  - in each case, I was able to verify the metrics showing up in cloudwatch and firehose

## Deployment strategy

Ideally we deploy this PR before the aiproxy PR, so that we know how to handle the new response codes before aiproxy starts returning them. If we do it in the other order, we might see the same error cases show up differently in Honey Badger, drawing unnecessary attention from the DOTD.

## Follow-up work

Move from firehose logging to cloudwatch logging once that is supported.

### DIFF
--- a/dashboard/app/jobs/evaluate_rubric_job.rb
+++ b/dashboard/app/jobs/evaluate_rubric_job.rb
@@ -252,7 +252,7 @@ class EvaluateRubricJob < ApplicationJob
   # Retry just once on a timeout. It is likely to timeout again.
   retry_on Net::ReadTimeout, Timeout::Error, wait: 10.seconds, attempts: ATTEMPTS_ON_TIMEOUT do |job, error|
     log_metric(metric_name: :TimeoutError)
-    log_to_firehose(job: job, error: error, event_name: 'retry-on-timeout')
+    log_to_firehose(job: job, error: error, event_name: 'timeout-error')
   end
 
   ATTEMPTS_ON_SERVICE_UNAVAILABLE = 3
@@ -262,7 +262,7 @@ class EvaluateRubricJob < ApplicationJob
   retry_on ServiceUnavailableError, wait: :exponentially_longer, attempts: ATTEMPTS_ON_SERVICE_UNAVAILABLE do |job, error|
     agent = error.message.downcase.include?('openai') ? 'openai' : 'none'
     log_metric(metric_name: :ServiceUnavailable, agent: agent)
-    log_to_firehose(job: job, error: error, event_name: 'retry-on-503', agent: agent)
+    log_to_firehose(job: job, error: error, event_name: 'service-unavailable', agent: agent)
   end
 
   ATTEMPTS_ON_GATEWAY_TIMEOUT = 3
@@ -272,7 +272,7 @@ class EvaluateRubricJob < ApplicationJob
   retry_on GatewayTimeoutError, wait: :exponentially_longer, attempts: ATTEMPTS_ON_GATEWAY_TIMEOUT do |job, error|
     agent = error.message.downcase.include?('openai') ? 'openai' : 'none'
     log_metric(metric_name: :GatewayTimeout, agent: agent)
-    log_to_firehose(job: job, error: error, event_name: 'retry-on-504', agent: agent)
+    log_to_firehose(job: job, error: error, event_name: 'gateway-timeout', agent: agent)
   end
 
   def self.log_metric(metric_name:, agent: nil, value: 1)
--- a/dashboard/app/jobs/evaluate_rubric_job.rb
+++ b/dashboard/app/jobs/evaluate_rubric_job.rb
@@ -231,10 +231,10 @@ class EvaluateRubricJob < ApplicationJob
     rubric_ai_evaluation.save!
   end
 
-  RETRIES_ON_RATE_LIMIT = 3
+  ATTEMPTS_ON_RATE_LIMIT = 3
 
   # Retry on any reported rate limit (429 status) 'exponentially_longer' waits 3s, 18s, and then 83s.
-  retry_on TooManyRequestsError, wait: :exponentially_longer, attempts: RETRIES_ON_RATE_LIMIT do |job, error|
+  retry_on TooManyRequestsError, wait: :exponentially_longer, attempts: ATTEMPTS_ON_RATE_LIMIT do |job, error|
     # Job arguments are always serializable, so we just pull out the hash
     # and send it as context.
     options = job.arguments.first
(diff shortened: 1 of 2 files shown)


dashboard/app/jobs/evaluate_rubric_job.rb
```diff
@@ -228,25 +247,75 @@ def self.log_metrics(response)
...
 
-  RETRIES_ON_TIMEOUT = 2
+  ATTEMPTS_ON_TIMEOUT = 2
 
   # Retry just once on a timeout. It is likely to timeout again.
-  retry_on Net::ReadTimeout, Timeout::Error, wait: 10.seconds, attempts: RETRIES_ON_TIMEOUT do
+  retry_on Net::ReadTimeout, Timeout::Error, wait: 10.seconds, attempts: ATTEMPTS_ON_TIMEOUT do |job, error|
+    log_metric(metric_name: :TimeoutError)
+    log_to_firehose(job: job, error: error, event_name: 'timeout-error')
+  end
```
davidsbailey: the `retry_on` block is only executed if/after all N attempts fail. the metric and event names are updated to reflect this.

davidsbailey:
apologies for the commit spam folks -- @cat5inthecradle @Nokondi this is ready for review.

cat5inthecradle:
> One concern with using these response codes is that the same codes might also be returned due to other infrastructure problems. To address this, I've added an "agent" dimension to each metric and firehose event, which will be "openai" when the problem is due to an openai error and "none" when the problem is between dashboard and aiproxy.

This makes sense and it's a tricky problem to solve. I've also seen folks use `server` or `host` when tracking which request in the stack failed, as a more general solution that focuses on the HTTP request instead of the specific functionality.

dashboard/app/jobs/evaluate_rubric_job.rb
```diff
@@ -228,25 +247,75 @@ def self.log_metrics(response)
...
+          script_name: script_level.script.name,
+          lesson_number: script_level.lesson.relative_position,
+          level_name: script_level.level.name,
+          agent: agent
+        }.to_json
+      }
+    )
+  end
+
   def perform(user_id:, requester_id:, script_level_id:, rubric_ai_evaluation_id: nil)
```
cat5inthecradle, line 311: personal preference nitpick - The `perform` method is the entrypoint and it's clearer if it's the first method in the class.

dashboard/app/jobs/evaluate_rubric_job.rb
```diff
@@ -212,10 +231,10 @@ def self.log_metrics(response)
     rubric_ai_evaluation.save!
   end
 
-  RETRIES_ON_RATE_LIMIT = 3
+  ATTEMPTS_ON_RATE_LIMIT = 3
 
-  # Retry on any reported rate limit (429 status) 'exponentially_longer' waits 3s, 18s, and then 83s.
-  retry_on TooManyRequestsError, wait: :exponentially_longer, attempts: RETRIES_ON_RATE_LIMIT do |job, error|
+  # Retry on any reported rate limit (429 status). With 3 attempts, 'exponentially_longer' waits 3s, then 18s.
+  retry_on TooManyRequestsError, wait: :exponentially_longer, attempts: ATTEMPTS_ON_RATE_LIMIT do |job, error|
```
cat5inthecradle, line 237: I'm not fully understanding why we're notifying Honeybadger for rate limit but not notifying Honeybadger for the other errors.

cat5inthecradle (approved):
The retry logic all makes sense to me. I am curious to understand more about the logging/alerting approach. Seems to me that any request failure to AIProxy should notify Honeybadger. Apologies if I'm missing some prior conversation on that.

dashboard/app/jobs/evaluate_rubric_job.rb
davidsbailey: this was an oversight -- if there's agreement on the overall approach, I'll update the rate limit handling to be more like the other ones I updated in this PR.

Nokondi (approved):
looks good!

davidsbailey:
> The retry logic all makes sense to me. I am curious to understand more about the logging/alerting approach. Seems to me that any request failure to AIProxy should notify Honeybadger. Apologies if I'm missing some prior conversation on that.

why do you think we should notify HB in these cases? It seems to me like honeybadger errors should only happen if there is a programming error. otherwise, what are we supposed to do in response to receiving the notice?

davidsbailey:
@cat5inthecradle more generally I'd love to develop an eng-wide best practice on when to log to honeybadger, especially as cloudwatch logging becomes available

dashboard/app/jobs/evaluate_rubric_job.rb
davidsbailey: this could get messy, as methods are also interspersed with directives like `retry_on`. tackling in a follow-up PR

davidsbailey: * https://github.com/code-dot-org/code-dot-org/pull/58510

I tried to follow https://rubystyle.guide/#consistent-classes, at least the parts of it I agreed with 😅 this is why the directives like `retry_on` still come before the `perform` method.

PR #58412, opened 2024-05-04
Styling to prevent text from being pushed down from top of banner

dju90:
## Testing story
Testing locally
_Before_
banner_before

_After_
banner_after

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/templates/sectionProgressV2/coming-soon-banner.module.scss
+++ b/apps/src/templates/sectionProgressV2/coming-soon-banner.module.scss
@@ -4,9 +4,11 @@
 
 
 .comingSoonBanner {
+  align-items: flex-start;
   background-color: $light_secondary_100;
   border: 0;
   color: $light_gray_700;
+  display: flex;
   line-height: 2rem;
 
   & .icon {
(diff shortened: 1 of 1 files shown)


apps/src/templates/sectionProgressV2/coming-soon-banner.module.scss
```diff
@@ -26,6 +28,7 @@
   }
 
   & .bannerText {
+    flex: 1 1 80%;
```
lfryemason:
```suggestion
    flex-grow: 1;
```
Do we need to set flex-shrink or flex-basis or can we just use flex-grow?

dju90:
Also tested on:
Chrome
banner_after_chrome

Firefox
banner_after_firefox

apps/src/templates/sectionProgressV2/coming-soon-banner.module.scss
dju90: Thanks! Tested this change locally 👍

lfryemason (approved):

PR #58426, opened 2024-05-06
feat(email): add script to scan for legacy partial registration keys

stephenliang:
This oneoff script scans the shared memcached instance for keys that do not have a TTL and matches one of the partial registration keys. In `DRY_RUN=true`, the keys will be printed out. In `DRY_RUN=false` the keys will be removed.

This script only works for memcached caches as filesystem caches aren't utilized in production and thus not stored.

## Links

- [jira ticket](https://codedotorg.atlassian.net/browse/P20-836)

## Testing story

1. With DCDO flag enabled, start registration. Script should not remove the key
2. With DCDO flag disabled, start registration. Script should remove the key
3. With `DRY_RUN=false`, script should remove the key.
4. With `DRY_RUN=true`, script should NOT remove the key
5. Ensured other keys are NOT removed

## Deployment strategy

This script will be executed with one other person after the TTLs have been published for a week.

## PR Checklist:

- [x] Code is well-commented

### DIFF
--- a/bin/oneoff/remove_no_ttl_partial_registrations.rb
+++ b/bin/oneoff/remove_no_ttl_partial_registrations.rb
@@ -7,7 +7,7 @@ require_relative '../../dashboard/config/environment'
 
 # This script inspects all cache entries in the partial registration namespace for the presence of a TTL.
 # If no TTL is present, the cache entry is removed.
-DRY_RUN = false
+DRY_RUN = true
 
 puts "Processing... DRYRUN=#{DRY_RUN}"
 
--- a/bin/oneoff/remove_no_ttl_partial_registrations.rb
+++ b/bin/oneoff/remove_no_ttl_partial_registrations.rb
@@ -1,0 +1,71 @@
+#!/usr/bin/env ruby
+
+require 'socket'
+require 'timeout'
+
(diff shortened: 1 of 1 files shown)


bin/oneoff/remove_no_ttl_partial_registrations.rb
```diff
@@ -0,0 +1,71 @@
...
+    Timeout.timeout(30) do
+      sock.write("lru_crawler metadump all\r\n")
+
+      # read keys until encountered an end (for memcached, this will be the Timeout)
+      loop do
+        data = sock.readline
+        break if !data || data.empty? || data == "END\r\n" || data == "ERROR\r\n"
+        matches = data.scan(/^key=(?<key>.*) exp=-1/)
+        break if matches.empty?
+        all_keys_without_expiration += matches.flatten!
```
carl-codeorg: Based on [the Ruby docs](https://apidock.com/ruby/v2_5_5/Array/%2B), it might improve performance to use `concat` if we end up manipulating a large array.

stephenliang: Thanks, updated this

carl-codeorg (approved):
LGTM :shipit:

PR #58434, opened 2024-05-06
P20-119: Add pre-lockdown parental permission banner

artem-vavilov:
## Summary
The banner will appear for students who are not compliant with the Child Account Policy while the CPA experience is `cpa_all_user_lockout_warning`.

**Note:** Use the parameter `?show_parental_permission_banner` to force the display of the banner.

## Links
- JIRA ticket: [P20-119](https://codedotorg.atlassian.net/browse/P20-119)
- Feature flag task: [P20-880](https://codedotorg.atlassian.net/browse/P20-880)
- Related PR: https://github.com/code-dot-org/code-dot-org/pull/58337
- [Figma](https://www.figma.com/file/wR0Myx8T2dHMaeeb1GAabb/CPA-Flows?type=design&node-id=1811-108&mode=design&t=dgLypeReXsde2qGp-0)


apps/src/templates/Notification.jsx
```diff
@@ -124,7 +125,9 @@ const Notification = ({
     return null;
   }
 
-  const colorStyles = styles.colors[type];
+  const colorStyles = color
+    ? {backgroundColor: color, borderColor: color}
+    : styles.colors[type];
```
wilkie: Since type is now not required, but either `color` or `type` seem to need to exist... and since `Notification` isn't TypeScript and can't support a discriminated union for its properties... maybe we need to at least raise some error if neither key is given. Or have a default type?

apps/src/templates/policy_compliance/ParentalPermissionModal/index.story.tsx
```diff
@@ -33,18 +33,22 @@ const useReducerStub = sinon.stub(React, 'useReducer');
...
+
+  return Template(state);
 };
 
 export const UpdateRequestForm = () => {
   const state = {
     parentalPermissionRequest: {
       parent_email: 'parent@email.com',
-      requested_at: new Date().toISOString(),
+      requested_at: Date(),
```
wilkie: Seems so haphazard to trust the localized string of the date instead of something more consistent. Is this really the way one is supposed to handle dates? I'm sure `moment` can magically handle anything but it feels so wrong.

apps/src/templates/Notification.jsx
wilkie: I'm also wondering if it would be more consistent to add your own type instead of adding the ability to directly manipulate the color so that updating the styling the notifications at-large can be made consistent.

artem-vavilov: I agree with your point about focusing on consistency. However, I bilieve that creating a custom type for each minor change could lead to a bloated component over time. Directly managing styles might be a more efficient approach, ensuring the component stays streamlined and its purpose remains clear

apps/src/templates/policy_compliance/ParentalPermissionModal/index.story.tsx
artem-vavilov: I'll revert this change

apps/src/templates/Notification.jsx
wilkie: Yep, that makes sense. Leave it as you had it.

artem-vavilov: https://github.com/code-dot-org/code-dot-org/pull/58434/commits/8567fc1e26051fc54589a8a9b36f6511870de802

apps/src/sites/studio/pages/policy_compliance/parental_permission/_modal.js
```diff
@@ -1,20 +1,57 @@
...
+import analyticsReporter from '@cdo/apps/lib/util/AnalyticsReporter';
+import {EVENTS, PLATFORMS} from '@cdo/apps/lib/util/AnalyticsConstants';
 
 const SHOW_DELAY = 86400; // 1 day
 
 document.addEventListener('DOMContentLoaded', () => {
   const renderModal = () => {
-    const lockoutDate = new Date(getScriptData('lockoutDate'));
+    const reportEvent = (eventName, payload = {}) => {
+      analyticsReporter.sendEvent(eventName, payload, PLATFORMS.STATSIG);
```
daynew: This should be Amplitude, right?

apps/src/templates/policy_compliance/ParentalPermissionBanner/index.tsx
```diff
@@ -0,0 +1,130 @@
...
+  const [consentStatus, setConsentStatus] = useState(
+    currentUser.childAccountComplianceState
+  );
+  const [showModal, setShowModal] = useState(false);
+  const formattedLockoutDate = moment(lockoutDate)
+    .lang(getCurrentLocale())
+    .format('ll');
+
+  const reportEvent = (eventName: string, payload: object = {}) => {
+    analyticsReporter.sendEvent(eventName, payload, PLATFORMS.STATSIG);
```
daynew: Amplitude

apps/src/templates/policy_compliance/ParentalPermissionBanner/index.tsx
```diff
@@ -0,0 +1,130 @@
...
+    setShowModal(false);
+
+    const newConsentStatus = parentalPermissionRequest
+      ? parentalPermissionRequest.consent_status
+      : consentStatus;
+
+    setConsentStatus(newConsentStatus);
+
+    reportEvent(EVENTS.CPA_PARENT_EMAIL_BANNER_CLOSED, {
+      consentStatus: newConsentStatus,
```
daynew, line 75: Does the analytics library already include `in_section`?

apps/src/lib/util/AnalyticsConstants.js
```diff
@@ -269,6 +269,11 @@ const EVENTS = {
   PUBLISH_MODEL_CARD_INFO: 'Student publishes their model card info',
 
   // CPA
+  CPA_PARENT_EMAIL_BANNER_CLICKED: 'cpa_parent_email_banner_clicked',
+  CPA_PARENT_EMAIL_BANNER_SUBMITTED: 'cpa_parent_email_banner_submitted',
+  CPA_PARENT_EMAIL_BANNER_UPDATED: 'cpa_parent_email_banner_updated',
+  CPA_PARENT_EMAIL_BANNER_RESEND: 'cpa_parent_email_banner_resend',
+  CPA_PARENT_EMAIL_BANNER_CLOSED: 'cpa_parent_email_banner_closed',
```
daynew: Although it wasn't originally requested, I think we will also want 
```
CPA_PARENT_EMAIL_BANNER_SHOWN: 'cpa_parent_email_banner_shown',
```
This is the first step in the experience. Please add this metric. I updated the JIRA task.

daynew (changes requested):
We should add a "shown" metric for this banner.

apps/src/lib/util/AnalyticsConstants.js
artem-vavilov: https://github.com/code-dot-org/code-dot-org/pull/58434/commits/0e74324359b554b5b92108e76e8065710bd0d3e0

apps/src/templates/policy_compliance/ParentalPermissionBanner/index.tsx
artem-vavilov: https://github.com/code-dot-org/code-dot-org/pull/58434/commits/eb58bb2047c42fafc9a9c50478276fb10785fc46

daynew (approved):

apps/src/templates/policy_compliance/ParentalPermissionModal/index.story.tsx
wilkie: I appreciate your effort on this. I'm never quite confident about handling date serialization. I normally default to the iso format whenever there is doubt, but I'm still ready to be convinced otherwise.

apps/src/sites/studio/pages/home/_homepage.js
```diff
@@ -80,6 +81,12 @@ function showHomepage() {
   }
 
   const announcement = getTeacherAnnouncement(announcementOverride);
+  const parentalPermissionBanner = homepageData.parentalPermissionBanner && (
+    <ParentalPermissionBanner
+      key="parentla-permission-bnner"
```
wilkie: two subtle typos in the key: `parental-permission-banner`

apps/src/sites/studio/pages/policy_compliance/parental_permission/_modal.js
```diff
@@ -1,20 +1,60 @@
...
 
 const SHOW_DELAY = 86400; // 1 day
 
 document.addEventListener('DOMContentLoaded', () => {
   const renderModal = () => {
-    const lockoutDate = new Date(getScriptData('lockoutDate'));
+    const inSection = getScriptData('inSection');
+
+    const reportEvent = (eventName, payload = {}) => {
+      payload.isSection = inSection;
```
wilkie: is this supposed to be `isSection` or `inSection` in the event?

dashboard/app/views/scripts/show.html.haml
```diff
@@ -6,7 +6,8 @@
 - scriptOverviewData = {scriptData: @script_data, plcBreadcrumb: @plc_breadcrumb}
 
 - content_for(:head) do
-  %script{ src: webpack_asset_path('js/scripts/show.js'), data: {scriptoverview: scriptOverviewData.to_json }}
+  %script{src: webpack_asset_path('js/scripts/show.js'),
+          data: {scriptoverview: scriptOverviewData.to_json, parental_permission_banner: current_user && parental_permission_banner_data(current_user, request).to_json}}
```
wilkie: Technically, this sets the field to `"null"` when there is no banner data... and that does work since it is parsed in `show.js`. The logic in `show.js` seems to support this dataset field being actual `null` or undefined. Could we have the data field not exist in that case?

wilkie (commented):
Looks great! Some comments about some possible typos and a small nit but other than that looks like some great work as usual. I'm happy to approve when those typos are addressed.

apps/src/templates/policy_compliance/ParentalPermissionModal/index.story.tsx
artem-vavilov: When working with an API, the date must be in ISO format. However, I think this is not as crucial for a storybook example

PR #58436, opened 2024-05-07
Datablock Storage: use CSRF token from AuthenticityTokenStore.ts

### LINKED ISSUE #57430: Datablock Storage: Get CSRF token using AuthenticityTokenStore.ts
While there's not a clear CSRF attack we could do on DatablockStorageController.rb (because all GET requests in the method are non-data-mutating), there's enough methods that it would be better if we didn't disable CSRF authing for the controller.

We originally disabled this because our cached-for-scale pages (e.g. curriculum content) do not have a valid CSRF token.

@bethanyaconnor pointed us to:
https://github.com/code-dot-org/code-dot-org/pull/51439/files#diff-520783f85ff905ddf1b768264fa818b570d16da8d67de93bb9e1ebeb26176e17

Which has a `getAuthenticityToken()` that does a fetch from `AuthenticityTokenController`

We should use it and re-enable CSRF-auth on DatablockStorageController.

snickell:
Fixes [#57430](https://github.com/code-dot-org/code-dot-org/issues/57430)

### DIFF
--- a/apps/src/storage/datablockStorage.js
+++ b/apps/src/storage/datablockStorage.js
@@ -1,4 +1,5 @@
 import _ from 'lodash';
+import {getAuthenticityToken} from '@cdo/apps/util/AuthenticityTokenStore';
 
 // DatablockStorage powers the "Data" tab in App Lab and Game Lab, including the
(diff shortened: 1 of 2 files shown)


cnbrenci (approved):
LGTM 🚀

PR #58441, opened 2024-05-07
Gen AI: handle tab list changing on level change

bencodeorg:
Dan noticed a bug where the tab content in the model customization panel disappears if you have a tab selected, then switch to at level that has been configured to hide that tab (see Jira for video / more details). This is a result of not properly resetting the selected tab index when the tab list changes, which this PR fixes.

## Links

- Jira: https://codedotorg.atlassian.net/browse/LABS-766

## Testing story

Tested on the levels where Dan observed this, and was able to navigate between panels successfully.

## Follow-up work

We should a) probably migrate to Denys's new `Tabs` component, which at a skim looks like it might have the same issue (FYI @levadadenys)


sanchitmalhotra126 (approved):
LGTM!

Relatedly, should we also reset the active index back to 0 whenever current level ID changes, even if it's within bounds? Do we want to always switch back to the "Setup" tab when changing levels?

bencodeorg:
> LGTM!
> 
> Relatedly, should we also reset the active index back to 0 whenever current level ID changes, even if it's within bounds? Do we want to always switch back to the "Setup" tab when changing levels?

Playing around with this, I kind of liked that you kept your tab selected (if possible) when you switched between levels. Might be a nice to have to let levelbuilders configure this one day though. I think we should also move to Denys's tabs component, which might be a good time to figure out the finer points.

levadadenys (commented):
For DSCO `Tabs` we assume we use a value/key/id instead of array index, so this issue might not appear, or appear in a different way haha (In case user will somehow select the tab which is not rendered or remove currently active tab), though in those cases tabs logic should probably be handled by parent component which is also supported.

Overall, +1 it's worth trying migrate to DSCO `Tabs` component when there'll be time for it and see how it goes

PR #58442, opened 2024-05-07
[Blockly] always remove 'block_id_' prefix before highlighting blocks

mikeharv:
Block highlighting during program execution was not working on Google Blockly. This feature is especially important for our sequencing-based labs like Artist/Maze, and even more especially when there is a "Step" button available. In the CDO Blockly days, we could assume that generated block ids were just integers, however Google Blockly uses a random string of characters. This was causing a regular expression to not match for any Google Blockly block ids. Without stripping the prefix, Blockly is unable to find the block by id and highlight it. We can use a simple `.replace()` with a new simpler regular expression. The expression's `^` ensures that the `'block_id_'` pattern must occur right at the start of the string.

**CDO Blockly - Baseline, no regressions**

https://github.com/code-dot-org/code-dot-org/assets/43474485/89a892e2-ede9-473a-9349-680c63bb4793

**Google Blockly before**

https://github.com/code-dot-org/code-dot-org/assets/43474485/2414f976-151b-4e02-9a56-d8659ba36565

**Google Blockly after**

https://github.com/code-dot-org/code-dot-org/assets/43474485/302f8689-5660-4290-a57c-cfde0d6d74d7

Separately, I also noticed a difference in how each Blockly version handles (or doesn't handle) the currently selected block when highlighting. CDO Blockly's `highlightBlock` function unselects the selected block:
https://github.com/code-dot-org/blockly/blob/v4.1.0/core/ui/block_space/block_space.js#L782-L804

However, Google Blockly does not: https://github.com/google/blockly/blob/171befa7462b4a5288ce0fa775814c267f424e2c/core/workspace_svg.ts#L1284-L1314

We can match the intended behavior by manually unselecting the selected block before calling `highlightBlock`. I created a `cdoUtils` function to handle this extra step for Google Blockly only.

## Links

Jira - https://codedotorg.atlassian.net/browse/CT-555

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/StudioApp.js
+++ b/apps/src/StudioApp.js
@@ -1588,10 +1588,7 @@ StudioApp.prototype.resizeToolboxHeader = function () {
 StudioApp.prototype.highlight = function (id, spotlight) {
   if (this.isUsingBlockly() && !isEditWhileRun(getStore().getState())) {
     if (id) {
-      var m = id.match(/^block_id_(\d+)$/);
-      if (m) {
-        id = m[1];
-      }
+      id = id.replace(/^block_id_/, '');
     }
 
     Blockly.mainBlockSpace.highlightBlock(id, spotlight);
--- a/apps/src/StudioApp.js
+++ b/apps/src/StudioApp.js
@@ -1591,7 +1591,7 @@ StudioApp.prototype.highlight = function (id, spotlight) {
       id = id.replace(/^block_id_/, '');
     }
 
-    Blockly.mainBlockSpace.highlightBlock(id, spotlight);
+    Blockly.cdoUtils.highlightBlock(id, spotlight);
   }
 };
 
--- a/apps/src/blockly/addons/cdoUtils.ts
+++ b/apps/src/blockly/addons/cdoUtils.ts
@@ -701,3 +701,12 @@ export function processToolboxXml(toolboxString: string) {
   const modifiedXmlString = new XMLSerializer().serializeToString(xmlDoc);
   return modifiedXmlString;
 }
+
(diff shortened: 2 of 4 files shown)


apps/src/StudioApp.js
```diff
@@ -1588,13 +1588,10 @@ StudioApp.prototype.resizeToolboxHeader = function () {
 StudioApp.prototype.highlight = function (id, spotlight) {
   if (this.isUsingBlockly() && !isEditWhileRun(getStore().getState())) {
     if (id) {
-      var m = id.match(/^block_id_(\d+)$/);
-      if (m) {
-        id = m[1];
-      }
+      id = id.replace(/^block_id_/, '');
```
molly-moen, line 1591: for my information: why is `block_id` appended to the front of the id?

molly-moen (approved):
Looks good, just one question for my understanding

apps/src/StudioApp.js
mikeharv: In many of our old block definitions, we actually add the block id to the generated code. Then as the interpreter does its thing, it's able to pass back those ids to Blockly to handle the highlighting. Here's one example (of close to 200): https://github.com/code-dot-org/code-dot-org/blob/d00dff787a1c604d198bdd1dc4319ad071f9447e/apps/src/maze/blocks.js#L110-L114

We strip these values out of the generated code when we show it to the student. The second screenshot in that work shows an example of `block_id_` in the generated code: https://github.com/code-dot-org/code-dot-org/pull/57903

PR #58445, opened 2024-05-07
Lab2: add source validator for labs with JSON sources

sanchitmalhotra126:
Quick convenience fix; in local development, we're running into a lot of AI Chat levels that share channel IDs with old labs that have incompatible sources, and have to manually go into S3 to delete the old sources. For other labs (Blockly/Python lab), the response validators will let us fall back to empty sources, so this adds similar behavior for AI Chat and any other labs that use JSON sources; if the JSON cannot be parsed, we'll fall back to empty sources.

## Testing story

Tested with AI chat levels in the gen-ai-foundations-draft script.

### DIFF
--- a/apps/src/lab2/responseValidators.ts
+++ b/apps/src/lab2/responseValidators.ts
@@ -6,7 +6,7 @@ import {
   ProjectSources,
 } from './types';
 import Lab2Registry from './Lab2Registry';
-import {BLOCKLY_LABS} from './constants';
+import {BLOCKLY_LABS, LABS_WITH_JSON_SOURCES} from './constants';
 
 // Validator for Blockly sources.
 const BlocklySourceResponseValidator: ResponseValidator<
@@ -46,6 +46,20 @@ const PythonSourceResponseValidator: ResponseValidator<
   return sourceValidatorHelper(response, pythonValidator);
 };
 
+const JsonSourceResponseValidator: ResponseValidator<
+  ProjectSources
+> = response => {
+  const jsonValidator = (responseToValidate: Record<string, unknown>) => {
+    try {
+      JSON.parse(responseToValidate.source as string);
+    } catch (e) {
+      throw new ValidationError('Error parsing JSON: ' + e);
+    }
+  };
+
+  return sourceValidatorHelper(response, jsonValidator);
+};
+
 // Default source validator. This just checks if there is a source field.
 const DefaultSourceResponseValidator: ResponseValidator<
   ProjectSources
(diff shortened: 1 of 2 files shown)


apps/src/lab2/responseValidators.ts
```diff
@@ -46,6 +46,21 @@ const PythonSourceResponseValidator: ResponseValidator<
...
 
+// Validator for labs that use JSON sources
+const JsonSourceResponseValidator: ResponseValidator<
+  ProjectSources
+> = response => {
+  const jsonValidator = (responseToValidate: Record<string, unknown>) => {
+    try {
+      JSON.parse(responseToValidate.source as string);
+    } catch (e) {
+      throw new ValidationError('Error parsing JSON: ' + e);
```
bencodeorg, line 57: Looks like you're following the existing pattern, but for my understanding, where does this get caught? I can't seem to trace it properly 🤦

bencodeorg (approved):
Thanks for doing this!

apps/src/lab2/responseValidators.ts
sanchitmalhotra126: No worries this code is a bit tricky to follow 🙂 this `ValidationError` is eventually caught by ProjectManager when it attempts to load sources for the project: https://github.com/code-dot-org/code-dot-org/blob/7b86d9cbbe8bfff2eac03f2ce9adacbd9ceef61f/apps/src/lab2/projects/ProjectManager.ts#L88-L91

It takes care of defaulting to empty sources and thereby not passing the invalid sources to labs.

bencodeorg: There it is! Thanks

molly-moen (approved):

apps/src/lab2/responseValidators.ts
```diff
@@ -46,6 +46,21 @@ const PythonSourceResponseValidator: ResponseValidator<
   return sourceValidatorHelper(response, pythonValidator);
 };
 
+// Validator for labs that use JSON sources
+const JsonSourceResponseValidator: ResponseValidator<
```
fisher-alice, line 50: nit: technically, Blocky sources are also json sources I think.

apps/src/lab2/responseValidators.ts
fisher-alice: Thanks for the explanation. Where are sources assigned `{}`?

fisher-alice (approved):
🙏  Thank you for doing this - I was so happy when I saw this PR title pop up in notifications!

apps/src/lab2/responseValidators.ts
sanchitmalhotra126: Good call, will update the comment to specify

levadadenys (approved):
Thanks a lot for doing this!

apps/src/lab2/responseValidators.ts
sanchitmalhotra126: Oh, they're just never assigned (they stay undefined from https://github.com/code-dot-org/code-dot-org/blob/806ab49f82ad35393b2873d20f8c6cac956a0264/apps/src/lab2/projects/ProjectManager.ts#L79

PR #58449, opened 2024-05-07
Gen AI: format markdown, bot icon to top of message

bencodeorg:
Two unrelated changes to the chat experience in gen AI:
- support markdown responses
- move the bot icon to the top of the response (this is what was spec'd in design, and is particularly noticeable with longer messages)

| Before | After |
| - | - |
| image | image |

## Links

- Jira: https://codedotorg.atlassian.net/browse/LABS-741

## Testing story

Tested manually in responses containing markdown, responses containing normal text, and errored responses.

### DIFF
--- a/apps/src/aichat/views/chatMessage.module.scss
+++ b/apps/src/aichat/views/chatMessage.module.scss
@@ -8,13 +8,13 @@
 }
 
 .userMessage {
-  margin: 0 0 15px 50px;
+  margin: 0 0 5px 50px;
   background-color: $brand_teal_5;
   border-radius: 16px 0 16px 16px;
 }
 
 .assistantMessageContainer {
-  margin: 0 50px 15px 0;
+  margin: 0 50px 5px 0;
   display: flex;
   align-items: start;
 }
@@ -45,6 +45,7 @@
   background-color: $light_gray_50;
   border-radius: 0 16px 16px 16px;
   margin-top: 14px; // a quarter of height of bot icon
+  overflow-x: scroll;
 }
 
 .inappropriateMessage {
--- a/apps/src/aichat/views/chatMessage.module.scss
+++ b/apps/src/aichat/views/chatMessage.module.scss
@@ -16,6 +16,7 @@
 .assistantMessageContainer {
   margin: 0 50px 15px 0;
   display: flex;
+  align-items: start;
 }
 
 .modelUpdateMessageTextContainer {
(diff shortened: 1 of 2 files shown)


sanchitmalhotra126 (approved):
🎉 🎉

apps/src/aichat/views/chatMessage.module.scss
```diff
@@ -43,6 +44,7 @@
 .assistantMessage {
   background-color: $light_gray_50;
   border-radius: 0 16px 16px 16px;
+  margin-top: 14px; // a quarter of height of bot icon
```
fisher-alice, line 47: Nit: icon seems a bit high when bot responses are single line.

fisher-alice (approved):
Nice!

apps/src/aichat/views/chatMessage.module.scss
fisher-alice: Still a nit but here are screenshots.

14px: 
Screenshot 2024-05-07 at 1 02 29 PM

4px: 
Screenshot 2024-05-07 at 1 02 17 PM

bencodeorg: Thanks, I'll play with it a bit! FWIW here's what the design mocks look like, need to figure out how to push down the bot message without adding extra space between the user and bot messages

image

bencodeorg: @fisher-alice want to give this another spin and see what you think? i think the main culprit of things being spaced out is a wide margin-bottom we had on bot and user messages, which I tightened up.

I also noticed that long single lines of code in an ordered list were overflowing the container (and were unscrollable), so I allowed scroll in the x direction for assistant messages. I cannot for the life of me get the bot to return it any more, but can simulate by modifying the dom directly:

image

Something like this in a response should repro the issue -- I don't see any other issues with content scrolling rather than just flowing onto the next line as I introduced `overflow-x: scroll`, but not totally sure):

```
<ol><li><code class="language-python">def add_numbers(num1, num2):
    sum = num1 + num2
    return sum
</code></li></ol>
```

fisher-alice: Nice work finding the edge case and fixing! I was able to repro by making window narrower and saw your fix resolved it. 🎉 

Screenshot 2024-05-08 at 10 45 48 AM

fisher-alice: > i think the main culprit of things being spaced out is a wide margin-bottom we had on bot and user messages, which I tightened up.

LGTM!

fisher-alice (approved):

PR #58450, opened 2024-05-07
Add amplitude event to code.org/music and update activity links

kelbyhawn:
Adds Amplitude (and Statsig) events to https://code.org/music, and updates the existing urls to the activity to include a `/reset` at the end in a few places.

## Links
Jira ticket: [ACQ-1869](https://codedotorg.atlassian.net/browse/ACQ-1869)

## Testing story
Local testing

----

Screenshot 2024-05-07 at 2 19 38 PM

### DIFF
--- a/pegasus/sites.v3/code.org/public/music/index.haml
+++ b/pegasus/sites.v3/code.org/public/music/index.haml
@@ -15,7 +15,7 @@ theme: responsive_full_width
         =hoc_s("music_lab.banner.heading")
       %p.body-one.white
         =hoc_s("music_lab.banner.desc")
-      %a.link-button.white{href: CDO.studio_url("/s/music-intro-2024")}
+      %a.link-button.white{href: CDO.studio_url("/s/music-intro-2024/reset")}
         =hoc_s("music_lab.button.try_music_lab")
       .afe.flex-container.align-items-baseline.wrap
         %p.add-margin-top.no-margin-bottom.white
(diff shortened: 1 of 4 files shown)


pegasus/sites.v3/code.org/public/music/index.haml
```diff
@@ -86,3 +86,5 @@ theme: responsive_full_width
       =hoc_s(:dance_afe_disclaimer)
 
 = view :swiper_page_music_lab
+
+= view :analytics_event_log_helper, event_name: AnalyticsConstants::MUSIC_PAGE_VISITED_EVENT
```
hannahbergam, line 90: Is the analytics event log helper automatically sending all events to both?

hannahbergam: I just double checked- that .js file would require a refactor to send Statsig events by default

hannahbergam: We can pair on it this PM if you want! (Also... if Statsig isn't a requirement here then please ignore me)

kelbyhawn: I don't know how that Statsig console message originally showed up, and I can't get it to show up again, weird! 🤔  

@dmcavoy do you want Statsig set up for Pegasus pages too?

dmcavoy: Not right now. Lets keep this just amplitude

hannahbergam (commented):
The screenshot is of two user properties events, not 'page visited' events. You should be able to see the event show up in the console under the two you showed (which happen on nearly every page load). Can you update the screenshot with the specific events?

hannahbergam (approved):
I see the new screenshot yay!

PR #55531, opened 2024-01-02
Add level URL to Dance analytics

fisher-alice:
This is a follow-up to https://github.com/code-dot-org/code-dot-org/pull/55318 which added Amplitude logging for Dance Party validation. @breville [suggested](https://github.com/code-dot-org/code-dot-org/pull/55318#issuecomment-1858635396) that we record the page path on which these events occur, especially for pass/fail events that look identical across levels. 

Although `state` values `nextLevelUrl` and `serverLevelId` already exist, this PR adds `currentScriptLevelUrl`, and then includes this path in the Amplitude event. For pages with '/s/', the URL is assigned. 

Screenshot 2024-01-03 at 11 21 26 AM

For pages with '/level/' `undefined` is assigned. 

Screenshot 2024-01-03 at 11 23 57 AM

Note that `serverLevelId` is already assigned the level id in `pageConstants`.

Screenshot 2024-01-03 at 11 24 36 AM

Now the Amplitude event includes:
- `levelPath`: page URL
- `result`: 'PASSED' or 'FAILED'
- `message`: feedback message key
- `userId`

Consulted with @samantha-code and she confirmed the level URL would be very helpful. I asked if she would want to log only failed events, and she said that 'Passed events can be logged tor (it would be interesting to do a query to see how many students passed without first having failed the level)' so logging both passed and failed validation level events.

## PR Checklist:

- [ ] Tests provide adequate coverage
- [ ] Privacy and Security impacts have been assessed
- [ ] Code is well-commented
- [ ] New features are translatable or updates will not break translations
- [ ] Relevant documentation has been added or updated
- [ ] User impact is well-understood and desirable
- [ ] Pull Request is labeled appropriately
- [ ] Follow-up work items (including potential tech debt) are tracked and linked

PR #58459, opened 2024-05-07
[Gen ai] Log level path instead of level id on Amplitude/Statsig

fisher-alice:
This PR logs the level path instead of level id when sending an event to Amplitude/Statsig. This was requested by @samantha-code .

Screenshot 2024-05-07 at 4 14 51 PM

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/aichat/redux/aichatRedux.ts
+++ b/apps/src/aichat/redux/aichatRedux.ts
@@ -17,7 +17,6 @@ import {
 } from '@cdo/generated-scripts/sharedConstants';
 import analyticsReporter from '@cdo/apps/lib/util/AnalyticsReporter';
 import {EVENTS, PLATFORMS} from '@cdo/apps/lib/util/AnalyticsConstants';
-import {ProgressState} from '@cdo/apps/code-studio/progressRedux';
 
 import {
   AI_CUSTOMIZATIONS_LABELS,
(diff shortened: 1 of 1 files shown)


apps/src/aichat/redux/aichatRedux.ts
```diff
@@ -165,12 +166,25 @@ export const saveModelCard = createAsyncThunk(
...
       savedAiCustomizations,
       EVENTS.SAVE_MODEL_CARD_INFO,
-      levelId,
+      levelPath,
       dispatch
     );
   }
 );
 
+const getLevelPath = (progress: ProgressState) => {
```
bencodeorg: Is there a reason not to just grab the URL (something like `window.location.pathname`)?

fisher-alice: Oh cool! I don't think so. I'll go ahead and use `window.location.pathname` instead. Thanks!

fisher-alice: FYI - when I added level url to Dance AI analytics, I added `currentScriptLevelUrl`.  https://github.com/code-dot-org/code-dot-org/pull/55531
But now I see I could've used `window.location.pathname` instead. Wonder if I should remove this seemingly unnecessary property.

bencodeorg: Yeah that seems redundant to me, but I could be missing something :)

bencodeorg (approved):

PR #58308, opened 2024-04-29
Add Music Lab dashboard banner content to announcements.json

kelbyhawn:
Adds Music Lab banner content to `announcements.json` so it can be translated ahead of the Music Lab launch. These are **not live** yet and will go live with a DCDO flag that I'll add in these tickets next sprint: [ACQ-1676](https://codedotorg.atlassian.net/browse/ACQ-1676) • [ACQ-1677](https://codedotorg.atlassian.net/browse/ACQ-1677)

**Jira ticket:** [ACQ-1847](https://codedotorg.atlassian.net/browse/ACQ-1847)

----

Screenshot 2024-04-29 at 9 28 45 AM

PR #58460, opened 2024-05-07
Add Music Lab announcement banners to Teacher and Student homepages

kelbyhawn:
Adds the DCDO logic to show the Music Lab announcement banners on the Teacher and Student homepages on https://studio.code.org for the launch in a couple weeks.
- This will not show until the `music-lab-launch-2024` DCDO flag is set to true
- The object for this was added in this PR: https://github.com/code-dot-org/code-dot-org/pull/58308
- Removes outdated AI TA and Dance Party AI banner objects

## Links
Jira tickets:
- Teacher - [ACQ-1676](https://codedotorg.atlassian.net/browse/ACQ-1676)
- Student - [ACQ-1677](https://codedotorg.atlassian.net/browse/ACQ-1677)

## Testing story
Tested locally that the DCDO flag works

----

## Teacher homepage
Teacher

## Student homepage
Student

### DIFF
--- a/dashboard/config/marketing/announcements.json
+++ b/dashboard/config/marketing/announcements.json
@@ -1,7 +1,7 @@
 {
   "pages": {
-    "/home": "ai-teaching-assistant",
-    "/student-home": "hoc-2023-dance-ai"
+    "/home": "music-lab-launch",
+    "/student-home": "music-lab-launch"
   },
   "banners": {
     "superhero": {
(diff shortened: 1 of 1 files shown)


dashboard/config/marketing/announcements.json
```diff
@@ -20,24 +20,8 @@
...
-      "dcdo": "ai-teaching-assistant-launch",
-      "image": "/images/action-blocks/action-block-ai-ta.png",
-      "title": "AI Teaching Assistant",
-      "body": "Elevate teaching confidence and streamline assessment of student progress. This tool heralds a new chapter in personalized learning and instructional excellence.",
-      "buttonText": "Learn more about AI Teaching Assistant",
-      "buttonUrl": "https://code.org/ai-ta",
-      "buttonId": "ai-teaching-assistant"
-    },
     "music-lab-launch": {
+      "dcdo": "music-lab-launch-2024",
```
bethanyaconnor, line 24: I do wonder what happens if multiple banners are supposed to be shown 🤔

bethanyaconnor (approved):

dashboard/config/marketing/announcements.json
kelbyhawn: Since this banner is used for marketing purposes, it makes sense to only show one at a time, so this may be more of a timing problem. I don't think we can put multiples banners up, even if they can be displayed with DCDO flags.

bethanyaconnor: oh sorry, I was curious what actually happens if multiple of these DCDO flags are set. Totally agree that only one should show at one time.

Really, I'm asking if there's some potential work here to make sure we don't accidentally launch multiple banners at the same time (wayyyyy out of scope of this PR)

kelbyhawn: Oh! To get a banner to show you have to update the `pages` object (see code below), which only takes one option at a time per page. Hypothetically it should be impossible to show more than one. 

I just tried setting both the `ai-teaching-assistant-launch` and `music-lab-launch-2024` DCDO flags to true locally, and only the Music Lab banner is showing up since that's what's set on the `pages` object. 

The only time I could see us _wanting_ to set multiples and controlling them with DCDO flags is if there is a tight launch turnaround and we want to turn one off (`ai-teaching-assistant-launch` for example) and one on (`music-lab-launch-2024`) without having to manually update this file. Like if we wanted to keep the AI TA banner up right up until Music Lab launched.

https://github.com/code-dot-org/code-dot-org/blob/7c1b5bee73ca9ec2992642b3fef2b825f8b5ce33/dashboard/config/marketing/announcements.json#L3-L4

bethanyaconnor: Got it thank you! Seems like it's working pretty well :)

PR #58461, opened 2024-05-07
Gen AI: report model response time

bencodeorg:
Reports the time it takes to get a response from the model after user submits a message. Also include the ~~path (ie, /s/allthethings/...)~~ model ID as a dimension to see if things vary by ~~level~~ model (the path as a dimension isn't a good idea, see discussion below).

image

## Links

- Jira: https://codedotorg.atlassian.net/browse/LABS-760

## Testing story

Tested manually (see screenshot above).

### DIFF
--- a/apps/src/aichat/redux/aichatRedux.ts
+++ b/apps/src/aichat/redux/aichatRedux.ts
@@ -285,6 +285,7 @@ export const submitChatContents = createAsyncThunk(
     thunkAPI.dispatch(addChatMessage(newMessage));
 
     // Post user content and messages to backend and retrieve assistant response.
+    const startTime = Date.now();
     const chatApiResponse = await postAichatCompletionMessage(
       newUserMessageText,
       currentSessionId
@@ -296,6 +297,14 @@ export const submitChatContents = createAsyncThunk(
       aichatContext,
       currentSessionId
     );
+    Lab2Registry.getInstance()
+      .getMetricsReporter()
+      .reportLoadTime('AichatModelResponseTime', Date.now() - startTime, [
+        {
+          name: 'Pathname',
+          value: window.location.pathname,
+        },
+      ]);
 
     // Regardless of response type,
     // assign last user message to session.
--- a/apps/src/aichat/redux/aichatRedux.ts
+++ b/apps/src/aichat/redux/aichatRedux.ts
@@ -301,8 +301,8 @@ export const submitChatContents = createAsyncThunk(
       .getMetricsReporter()
       .reportLoadTime('AichatModelResponseTime', Date.now() - startTime, [
         {
-          name: 'Pathname',
-          value: window.location.pathname,
+          name: 'ModelId',
+          value: aiCustomizations.selectedModelId,
         },
       ]);
 


bencodeorg:
Any opinions on any other information we should log other than the level path? We could log the message content?

levadadenys (approved):
LGTM!
I also think it might be worth to also log the input message content so that it will be possible to also check what's causing response time to be short/long.

fisher-alice:
> LGTM! I also think it might be worth to also log the input message content so that it will be possible to also check what's causing response time to be short/long.

+1 on user's message.
We could add system prompt when editable and maybe length of assistant response? (Is the response time long bc the response is long-winded or was it a "difficult" question to respond to?) Maybe consult with Sam as well?

fisher-alice (approved):
LGTM! Just added a couple suggestions to the question you left.

apps/src/aichat/redux/aichatRedux.ts
```diff
@@ -296,6 +297,14 @@ export const submitChatContents = createAsyncThunk(
...
       currentSessionId
     );
+    Lab2Registry.getInstance()
+      .getMetricsReporter()
+      .reportLoadTime('AichatModelResponseTime', Date.now() - startTime, [
+        {
+          name: 'Pathname',
+          value: window.location.pathname,
+        },
+      ]);
```
fisher-alice, line 307: This is question is out of scope for the PR, but related to the dimensions being reported and directed at @sanchitmalhotra126 . I see that the common dimension  `AppName` is included in the reporting. But curious why the other `commonProperties` (e.g., `channelId`, `currentLevelId`,...) are not included when calling [on `getCommonDimensions`](https://github.com/code-dot-org/code-dot-org/blob/be7f35b66d20a6cf09dbc66b00daff61b3868520/apps/src/lab2/Lab2MetricsReporter.ts#L89).

bencodeorg:
Thinking about this a bit more, since this runs on every request, I am a little hesitant to throw the content of the user message in there (ie, we'd be logging all user generated content to another system).

This also logs to "cloudwatch metrics", not CloudWatch Logs, which (as I understand it) is aimed at quantitative metrics rather than comprehensive logs. The additional information are "dimensions" that you might want to split your metric (response time) by. There is a separate helper (Lab2MetricsReporter.logInfo) that would log to CloudWatch Logs that we could use if we want more comprehensive logs, but I think this task is focused on the performance piece (ie, developer-focused, rather than product-focused).

I guess I think anything we add here is likely to be logged in perpetuity, so I'd be inclined to start small and add as needed. Taking the suggestions above, response length seems like the only thing that would fit given the considerations I've mentioned here, but not sure how much we need that at this point? Thoughts @fisher-alice @levadadenys @sanchitmalhotra126 ?

levadadenys:
> Thinking about this a bit more, since this runs on every request, I am a little hesitant to throw the content of the user message in there (ie, we'd be logging all user generated content to another system).
> 
> This also logs to "cloudwatch metrics", not CloudWatch Logs, which (as I understand it) is aimed at quantitative metrics rather than comprehensive logs. The additional information are "dimensions" that you might want to split your metric (response time) by. There is a separate helper (Lab2MetricsReporter.logInfo) that would log to CloudWatch Logs that we could use if we want more comprehensive logs, but I think this task is focused on the performance piece (ie, developer-focused, rather than product-focused).
> 
> I guess I think anything we add here is likely to be logged in perpetuity, so I'd be inclined to start small and add as needed. Taking the suggestions above, response length seems like the only thing that would fit given the considerations I've mentioned here, but not sure how much we need that at this point? Thoughts @fisher-alice @levadadenys @sanchitmalhotra126 ?

Agree, makes sense to start small and add more later if needed, once we know what we need to log for sure. As for logging response length - I'm not sure it that'll be helpful, but agree it should fit what you've metioned above. if it'll be useful - let's log it, if not - not 😄  (Cause from my understand right now we'll be having a general picture of which level has which response time)

In the future, if needed - we can add some conditions like if response time is bigger then N -> also log something (input msg, input msg length) , etc. Was also thinking about sharing some 'tags' instead of sharing the full input message, though a good question is how to implement that (unless we use ai to parse the input mesage haha).

bencodeorg:
@levadadenys love the idea of logging (to CloudWatch Logs) details like the message content, system prompt, etc when a request exceeds a certain roundtrip time! Not sure what the right threshold would be, maybe we can get some initial data from this and add logging once we have a better idea.

fisher-alice:
> Thinking about this a bit more, since this runs on every request, I am a little hesitant to throw the content of the user message in there (ie, we'd be logging all user generated content to another system).
> 
> This also logs to "cloudwatch metrics", not CloudWatch Logs, which (as I understand it) is aimed at quantitative metrics rather than comprehensive logs. The additional information are "dimensions" that you might want to split your metric (response time) by. There is a separate helper (Lab2MetricsReporter.logInfo) that would log to CloudWatch Logs that we could use if we want more comprehensive logs, but I think this task is focused on the performance piece (ie, developer-focused, rather than product-focused).
> 
> I guess I think anything we add here is likely to be logged in perpetuity, so I'd be inclined to start small and add as needed. Taking the suggestions above, response length seems like the only thing that would fit given the considerations I've mentioned here, but not sure how much we need that at this point? Thoughts @fisher-alice @levadadenys @sanchitmalhotra126 ?

Thanks for explaining the purpose of the cloudwatch metrics and the difference between metrics and logs.
Starting small and adding more once we gather data SGTM!
+1 on liking @levadadenys 's idea of setting a threshold after which we log.

sanchitmalhotra126:
Chiming in a bit late here but I think y'all have a good understanding of things :) a couple clarifications from my end
- Yep to clarify, this will publish a metric (via the Cloudwatch Metrics API) rather than publish a log message to Cloudwatch logs.
- Accordingly, we need to be a bit careful with adding metric [dimensions](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/cloudwatch_concepts.html#Dimension). Each additional dimension creates an additional metric, and aggregating/summing across dimensions can be a little tricky (there's a way to do it, but currently it limits metric to the last 3 hours only). We can think of dimensions as each individual line we might want to see on a graph; so for example, we probably want to add the selected model ID as a dimension as we'd probably want the different model latencies split out on a graph. However, we probably don't want to use path name since it's less likely that we want to have distinct metrics for each level, rather than aggregating latencies across all levels.
- As y'all pointed out, cloudwatch logs is the right place for logging any additional information like user message, system prompt, etc. We do something similar in Music Lab where we publish a metric with report a load time metric with just the latency and a dimension for whether the load was on the initial level load or after changing some code (since we expect those two numbers to be significantly different): https://github.com/code-dot-org/code-dot-org/blob/6de9ca5d26ea06a21a65d778297d9065dbd78362/apps/src/music/views/MusicView.jsx#L550-L559 but also publish a log to cloudwatch with more information, like the number of sounds loaded: https://github.com/code-dot-org/code-dot-org/blob/6de9ca5d26ea06a21a65d778297d9065dbd78362/apps/src/music/views/MusicView.jsx#L563-L567

Edit: also note that if using the Lab2MetricsReporter, log statements will already include metadata like level ID, channel ID, user ID, path name, etc.

apps/src/aichat/redux/aichatRedux.ts
sanchitmalhotra126: Yep, added some clarifications around metric dimensions in a separate comment. I think we probably don't want to include Pathname as a dimension since we don't want to emit distinct metrics for each level, but it would be good to include the selected model ID as a dimensions so we can view those metrics separately? Trying to think if there's any other fields we want to similarly bucketize by, but that's the first that comes to mind.

bencodeorg:
@sanchitmalhotra126 thanks for the clarification around dimensions! Switched from path to model ID.

apps/src/aichat/redux/aichatRedux.ts
fisher-alice: This makes sense - thanks for the explanations!

PR #58462, opened 2024-05-07
Add all remaining templates folders to import/order

lfryemason:

## PR Checklist:

- [ ] Code is well-commented


apps/src/templates/projects/frozenProjectInfoDialog/FrozenProjectInfoDialog.tsx
```diff
@@ -1,18 +1,21 @@
...
+import i18n from '@cdo/locale';
+
 import {
   FrozenProjectInfoDialogState,
   hideFrozenProjectInfoDialog,
 } from './frozenProjectInfoDialogRedux';
-import Typography from '@cdo/apps/componentLibrary/typography/Typography';
+
 // eslint-disable-next-line @typescript-eslint/ban-ts-comment
 // @ts-ignore
```
Hamms: I suspect these lines need to be moved to stay with the `@cdo/locale` import

apps/src/templates/rubrics/RubricContainer.jsx
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable import/order */
```
Hamms: Why is this file not getting fixed up? Worth adding an explanatory comment?

apps/src/templates/rubrics/introjs.scss
```diff
@@ -445,4 +445,3 @@ tr.introjs-showElement > th {
   z-index: 1;
   opacity: 0;
 }
-/*# sourceMappingURL=introjs.css.map */
```
Hamms: It's back! :smile:

Hamms (commented):
Found a couple of nitpicks; not bad, given the scope of this changeset!

apps/src/templates/rubrics/introjs.scss
lfryemason: Uuuugh, haven't merged my fix yet

apps/src/templates/projects/frozenProjectInfoDialog/FrozenProjectInfoDialog.tsx
lfryemason: Good catch!

apps/src/templates/rubrics/RubricContainer.jsx
lfryemason: This is because some of the imports had a comment and were purposefully grouped.

I added a comment and also ran the order linter before re-adding the groupings.

Hamms: Gotcha. How do you feel about disabling the rule just for that one block of import statements, rather than for the whole file? Something like:

```js
// product Tour
/* eslint-disable import/order */
import './introjs.scss';
import {Steps} from 'intro.js-react';
import {INITIAL_STEP, STEPS, DUMMY_PROPS} from './productTourHelpers';
/* eslint-enable import/order */
```

apps/src/templates/VersionHistory.jsx
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable import/order */
+// Disabling import order due to circular dependencies.
```
Hamms: Could you expand on this comment a bit more? I'd like to capture information here about what the specific circular dependency is and exactly what about the code would need to change for us to no longer need this.

If possible, I'd also ideally like to only ignore this rule for the specific imports that need to ignore it, and to still apply it to all other imports in the file.

apps/src/templates/projects/projectsRedux.js
```diff
@@ -1,4 +1,7 @@
 /** @file Redux actions and reducer for the Projects Gallery */
+
+/* eslint-disable import/order */
+// Disabling import order due to circular dependencies.
```
Hamms: same as above; I'd love to scope this disable directive down to just the imports that need it, and document more details about what exactly goes wrong when we don't disable.

apps/src/templates/rubrics/RubricContainer.jsx
```diff
@@ -1,27 +1,34 @@
...
-import {TAB_NAMES} from './rubricHelpers';
-import aiBotOutlineIcon from '@cdo/static/ai-bot-outline.png';
-import HttpClient from '@cdo/apps/util/HttpClient';
-import analyticsReporter from '@cdo/apps/lib/util/AnalyticsReporter';
-import {EVENTS} from '@cdo/apps/lib/util/AnalyticsConstants';
+import RubricTabButtons from './RubricTabButtons';
+
+import style from './rubrics.module.scss';
 
+/* eslint-disable import/order */
```
Hamms, line 29: please also reenable the rule after this group of import! I'd rather not risk new imports getting added below this not being linted

apps/src/templates/rubrics/RubricContainer.jsx
```diff
@@ -1,27 +1,34 @@
-import React, {useCallback, useEffect, useState, useRef} from 'react';
-import PropTypes from 'prop-types';
-import style from './rubrics.module.scss';
-import i18n from '@cdo/locale';
+/* eslint-disable import/order */
```
Hamms: Is this disable directive intentional, or do we only need the one below?

apps/src/templates/rubrics/introjs.scss
```diff
@@ -445,4 +445,5 @@ tr.introjs-showElement > th {
...
+
+/* eslint-disable import/order */
```
Hamms: I don't think this belongs here :smile: 

Also looks like the unintended removal above is still present in the changeset

apps/src/templates/rubrics/RubricContainer.jsx
```diff
@@ -1,31 +1,38 @@
-import React, {useCallback, useEffect, useState, useRef} from 'react';
-import PropTypes from 'prop-types';
-import style from './rubrics.module.scss';
-import i18n from '@cdo/locale';
+// Disabling import/order rule for this file because product tour imports are grouped.
```
Hamms: nit: stray comment

apps/src/templates/projects/projectsRedux.js
```diff
@@ -1,12 +1,15 @@
...
 import $ from 'jquery';
 import _ from 'lodash';
+import {combineReducers} from 'redux';
+
 import {Galleries} from './projectConstants';
 import {PUBLISH_SUCCESS} from './publishDialog/publishDialogRedux';
-import {DELETE_SUCCESS} from './deleteDialog/deleteProjectDialogRedux';
-import {channels as channelsApi} from '../../clientApi';
-import LibraryClientApi from '@cdo/apps/code-studio/components/libraries/LibraryClientApi';
+// Disabling import order due to circular dependencies errors from build.
```
Hamms: I would still like to see some more specificity here about what exactly the circular dependency is. One purpose for comments like this is to help future developers identify when the exception to the norm we've introduced could be removed.

apps/src/templates/projects/projectsRedux.js
```diff
@@ -1,12 +1,15 @@
...
+import {DELETE_SUCCESS} from './deleteDialog/deleteProjectDialogRedux'; /* eslint-disable-line import/order */
+import {channels as channelsApi} from '../../clientApi'; /* eslint-disable-line import/order */
+import LibraryClientApi from '@cdo/apps/code-studio/components/libraries/LibraryClientApi'; /* eslint-disable-line import/order */
```
Hamms: Is there a specific reason you're disabling these line-by-line rather than as a block?

apps/src/templates/VersionHistory.jsx
```diff
@@ -1,12 +1,18 @@
 import PropTypes from 'prop-types';
 import React from 'react';
-import VersionRow from './VersionRow';
+
+// Disabling import order due to circular dependencies errors from build.
```
Hamms: As above, I'd like to see some more details in this comment, and I'm curious why we're disabling the rule line-by-line rather than as a group.

Hamms (approved):
A couple comment nits, but for the most part this LGTM!

PR #58463, opened 2024-05-07
P20-831: Optimize user search by username

artem-vavilov:
## Summary

Testing the `users` table locally with over 3.25 million records

### Before `(9483.1ms)`
```SQL
EXPLAIN for: SELECT `users`.* FROM `users` WHERE `users`.`username` LIKE '%art%'
+----+-------------+-------+------------+------+---------------+------+---------+------+---------+----------+-------------+
| id | select_type | table | partitions | type | possible_keys | key  | key_len | ref  | rows    | filtered | Extra       |
+----+-------------+-------+------------+------+---------------+------+---------+------+---------+----------+-------------+
|  1 | SIMPLE      | users | NULL       | ALL  | NULL          | NULL | NULL    | NULL | 3135067 |    11.11 | Using where |
+----+-------------+-------+------------+------+---------------+------+---------+------+---------+----------+-------------+
```

### After `(0.6ms)`
```SQL
EXPLAIN for: SELECT `users`.* FROM `users` WHERE `users`.`username` = 'artem'
+----+-------------+-------+------------+------+----------------------------------------+----------------------------------------+---------+-------+------+----------+-------+
| id | select_type | table | partitions | type | possible_keys                          | key                                    | key_len | ref   | rows | filtered | Extra |
+----+-------------+-------+------------+------+----------------------------------------+----------------------------------------+---------+-------+------+----------+-------+
|  1 | SIMPLE      | users | NULL       | ref  | index_users_on_username_and_deleted_at | index_users_on_username_and_deleted_at | 768     | const |    1 |    100.0 | NULL  |
+----+-------------+-------+------------+------+----------------------------------------+----------------------------------------+---------+-------+------+----------+-------+
```

## Links
- JIRA ticket: [P20-831](https://codedotorg.atlassian.net/browse/P20-831)

### DIFF
--- a/dashboard/app/controllers/admin_search_controller.rb
+++ b/dashboard/app/controllers/admin_search_controller.rb
@@ -16,7 +16,7 @@ class AdminSearchController < ApplicationController
 
     # If requested, filter...
     if params[:usernameFilter].present?
-      users = users.where(User.arel_table[:username].matches("%#{params[:usernameFilter]}%"))
+      users = users.where(User.arel_table[:username].matches("#{params[:usernameFilter]}%"))
     end
     if params[:studentNameFilter].present?
       users = users.where(User.arel_table[:name].matches("%#{params[:studentNameFilter]}%"))
--- a/dashboard/app/controllers/admin_search_controller.rb
+++ b/dashboard/app/controllers/admin_search_controller.rb
@@ -16,7 +16,7 @@ class AdminSearchController < ApplicationController
 
     # If requested, filter...
     if params[:usernameFilter].present?
-      users = users.where(User.arel_table[:username].matches("#{params[:usernameFilter]}%"))
+      users = users.where(username: params[:usernameFilter])
     end
     if params[:studentNameFilter].present?
       users = users.where(User.arel_table[:name].matches("%#{params[:studentNameFilter]}%"))


dashboard/app/controllers/admin_search_controller.rb
```diff
@@ -16,7 +16,7 @@ def find_students
 
     # If requested, filter...
     if params[:usernameFilter].present?
-      users = users.where(User.arel_table[:username].matches("%#{params[:usernameFilter]}%"))
+      users = users.where(User.arel_table[:username].matches("#{params[:usernameFilter]}%"))
```
carl-codeorg: What would the performance look like if we got rid of the wildcards entirely and did `User.arel_table[:username].eq(params[:usernameFilter])`? In your test, the improvement was pretty huge by removing the leading %. If we get another large improvement by going to `.eq` it might be worth switching over, considering the large size of the users table and the fact that the CX team says they normally have the exact username they want to search.

carl-codeorg (approved):
LGTM! Thanks for taking this on!

carl-codeorg (commented):
Btw after this is deployed, it's probably worth posting in #customer-support to inform them of the change.

PR #58469, opened 2024-05-08
[Design2-108] DSCO - Add Aria/AriaDescribedBy attributes to components where needed

levadadenys:
### [Design2-108] DSCO - Add Aria/AriaDescribedBy attributes to components where needed

* added helpers, add getAriaPropsFromProps helper function
* added `aria-` attributes support for `Button` components
* added `aria-` attributes support for `...Dropdown` components
* add `aria-` attributes support for `FontAwesomeV6Icon` component

## Links
- jira ticket: [Design2-108](https://codedotorg.atlassian.net/browse/DESIGN2-108)

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/componentLibrary/common/helpers.ts
+++ b/apps/src/componentLibrary/common/helpers.ts
@@ -1,0 +1,23 @@
+import {AriaAttributes} from 'react';
+
+type PrimitiveValue =
+  | string
+  | number
+  | bigint
+  | boolean
+  | symbol
+  | null
+  | undefined;
+
+export const getAriaPropsFromProps = (props: {
+  [key: string]: PrimitiveValue;
+}) => {
+  const ariaProps: {[key: string]: PrimitiveValue} = {};
+  Object.keys(props).forEach(key => {
+    if (key.startsWith('aria-')) {
+      ariaProps[key] = props[key];
+    }
+  });
+
+  return ariaProps as AriaAttributes;
+};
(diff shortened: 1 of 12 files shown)


apps/src/componentLibrary/common/helpers.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export const getAriaPropsFromProps = (props: {
+  [key: string]: PrimitiveValue;
+}) => {
+  const ariaProps: {[key: string]: PrimitiveValue} = {};
+  Object.keys(props).forEach(key => {
+    if (key.startsWith('aria-')) {
+      ariaProps[key] = props[key];
+    }
+  });
```
thomasoniii, line 20: this is fine and doesn't need to be changed, but I think it reads a little nicer with a `fromEntries` and a `filter`.

```
const ariaProps: {[key: string]: PrimitiveValue} = Object.fromEntries(
  Object.entries(props).filter(([key]) => key.startsWith('aria-'))
)
```

thomasoniii (approved):
minor non-blocking syntactic sugar suggestion, but otherwise lgtm 🚀

sanchitmalhotra126 (approved):

apps/src/componentLibrary/common/helpers.ts
levadadenys: Agree it reads better in your suggestion. Though, I got interested and compared the perfomance of both implementation and the current one turned out to be faster. 
The more props there'll be in an array - the bigger gap would be it was 12% average faster when there were ~15 props and it turned out to be 27% average faster when there were. ~25+ props pased. Giving that this function will be used a lot and the number of props passed here is unknown (it's possible there'll be 10 and more props for one component) I suggest we stick with current version. Reason of execution time difference smaller is a number of operations executed in the current version.

PR #58470, opened 2024-05-08
Sets up AB test scaffolding for school assocation

hannahbergam:
This PR sets up experiment scaffolding in the StatsigReporter to support the new AB test for school association in the finish_sign_up file. 

I opted to abstract the calls in the statsigReporter so we would only have to import Statsig in one place. Once this is merged, the access to each flow will be determined on the Statsig experiment page for school association. I added a new parameter called 'showNewFlow' that will be 'false' for the control group and 'true' for the test group (see pic below).

Though this is a bit weird to test locally (I have to use a production key on a local environment and Statsig complains that there is an environment mismatch), I was able to work through all the designated testing steps from the Statsig platform and verify that once I added my Stable Id to the manual override page, I was able to put myself in the new flow.

I removed the experiment flag entry for now, and had to disable the ui test accordingly. I created [this ticket](https://codedotorg.atlassian.net/browse/ACQ-1880) to clean those up post-launch.

This would not make sense for an adhoc, because we would still have to manually add each device id to the 'manual override' list. If folks are looking to test this locally, I would be happy to walk them through it in a meeting!

If we moved both of the groups to 'false', we would be able to merge this PR without any users entering the experiment.

Screenshot 2024-05-08 at 3 41 53 PM
Screenshot 2024-05-08 at 3 38 40 PM
Screenshot 2024-05-08 at 3 38 46 PM

## Links

- spec: []()
- jira ticket: https://codedotorg.atlassian.net/browse/ACQ-1546

## PR Checklist:

- [ ] Code is well-commented


TurnerRiley (approved):
LGTM! Super excited to see this go live! Is the reason for disabling the UI test that there's a chance the test run of it will end up with a `true` value for `isInSchoolAssociationExperiment`?

hannahbergam:
> LGTM! Super excited to see this go live! Is the reason for disabling the UI test that there's a chance the test run of it will end up with a `true` value for `isInSchoolAssociationExperiment`?

Yay thanks!! It's just that it relies explicitly on the experiment flag, which I removed from the code so that we wouldn't have a confusing state of the experiment flag and the statsig code interacting

apps/src/sites/studio/pages/devise/registrations/_finish_sign_up.js
```diff
@@ -238,7 +245,7 @@ $(document).ready(() => {
 
   function renderSchoolInfo() {
     if (schoolInfoMountPoint) {
-      if (experiments.isEnabled(experiments.SCHOOL_ASSOCIATION_V2)) {
```
hannahbergam, line 241: @TurnerRiley Here is where I remove the experiment flag:)

PR #58471, opened 2024-05-08
Added tests for Progress View V2 Invitation

kobryan0619:
Test following the[ PR for the invitation modal to V2 Progress view.](https://github.com/code-dot-org/code-dot-org/pull/58372)

Note: eyes tests will be in future PR

### DIFF
--- a/apps/src/templates/sectionProgressV2/InviteToV2ProgressModal.jsx
+++ b/apps/src/templates/sectionProgressV2/InviteToV2ProgressModal.jsx
@@ -164,6 +164,8 @@ InviteToV2ProgressModal.propTypes = {
   setShowProgressTableV2: PropTypes.func.isRequired,
 };
 
+export const UnconnectedInviteToV2ProgressModal = InviteToV2ProgressModal;
+
 export default connect(
   state => ({
     dateProgressTableInvitationDelayed:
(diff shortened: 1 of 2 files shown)


apps/src/templates/sectionProgressV2/InviteToV2ProgressModal.jsx
```diff
@@ -34,11 +34,13 @@ function InviteToV2ProgressModal({
   const [invitationOpen, setInvitationOpen] = React.useState(false);
 
   React.useEffect(() => {
-    const timeSinceInvitationLastDelayed = () => {
+    const numDaysSinceInvitationLastDelayed = () => {
+      const MILLISECONDS_IN_ONE_DAY = 1000 * 3600 * 24;
```
lfryemason: Nit: move this constant outside the `InviteToV2ProgressModal` function (to the top of the file).

apps/test/unit/templates/sectionProgressV2/InviteToV2ProgressModalTest.jsx
```diff
@@ -0,0 +1,157 @@
...
+        {...DEFAULT_PROPS}
+        {...propOverrides}
+      />
+    );
+  }
+
+  it('renders the dialog with required elements', () => {
+    renderDefault();
+
+    expect(screen.getByText(i18n.progressTrackingAnnouncement())).to.exist;
```
lfryemason:
```suggestion
    screen.getByText(i18n.progressTrackingAnnouncement());
```
getByText will throw an error if it is not found. We don't need the `expect` statement and in fact get a much much better error message if we omit it.
Same for a bunch below.

lfryemason (approved):

PR #58475, opened 2024-05-08
Refactor Project-related code in CodeBridge Labs

molly-moen:
We had some duplicated code between Web Lab 2 and Python Lab for saving and loading projects. This PR de-duplicated that code by moving the project sources previously stored in `pythonlabRedux` to a new `lab2ProjectRedux` store. Right now only Web Lab 2 and Python Lab use this slice, but other lab2 labs could utilize it as well. I put this in a new slice because lab2Redux was already getting a bit big and this piece of state is pretty big on its own, and we could end up with more selectors/thunks going forward.

Now CodeBridge labs just need to use the `useSource` hook to get the current source and a `setSource` function. I kept `useInitialSources` as its own hook even though it's only referenced in `useSource`, happy to combine those as a follow-up if desired.

## Links

- jira ticket: [CT-499](https://codedotorg.atlassian.net/browse/CT-499)

## Testing story
I tested in normal and start mode in Python Lab, and normal mode in Web Lab 2 (as we don't fully support start mode for web lab 2 yet). Reading/writing from files and running works, and you can save start code.

I got a bit tripped up by `useMemo` functions running more than expected as I was working on this. I eventually figured out that my main issue was creating new objects like `{source: mySourceObject}` and passing them as parameters or returning them. Those were then always treated as new objects by the memoization logic. The solution was to not do that :)

## Follow-up work
I created a new `redux` folder under `lab2` to put the new slice in. I did not move `lab2Redux` because that would have ballooned the size of the PR, and since Music Lab uses `lab2Redux` I didn't want to make any changes that could cause issues there.

## PR Checklist:

- [ ] Code is well-commented


apps/src/codebridge/hooks/useInitialSources.ts
```diff
@@ -5,22 +5,33 @@ import {getAppOptionsEditBlocks} from '@cdo/apps/lab2/projects/utils';
...
 import {useAppSelector} from '@cdo/apps/util/reduxHooks';
 
+// Hook for getting the initial sources for the current level.
+// If we are in start mode, we show the start sources or the default sources
+// if the level has no start sources.
+// If the user has a project, we prioritize that over the start sources. If there is
+// no project or start sources, we use the default sources.
 export const useInitialSources = (defaultSources: ProjectSources) => {
-  const labState = useAppSelector(state => state.lab);
+  const labInitialSources = useAppSelector(state => state.lab.initialSources);
```
molly-moen, line 14: I split out the values we needed here because `labState` changes on every project save (channel metadata changes), which was causing `useMemo` to fire too often.

thomasoniii (approved):
lgtm 🚀

apps/src/codebridge/hooks/useInitialSources.ts
```diff
@@ -5,22 +5,33 @@ import {getAppOptionsEditBlocks} from '@cdo/apps/lab2/projects/utils';
...
+  const levelStartSource = useAppSelector(
+    state => state.lab.levelProperties?.source
+  );
+  // We memoize this object so that is doesn't cause an unexpected re-render.
```
ebeastlake: nit: so that it

ebeastlake:
🧹

apps/src/codebridge/hooks/useInitialSources.ts
```diff
@@ -5,22 +5,33 @@ import {getAppOptionsEditBlocks} from '@cdo/apps/lab2/projects/utils';
...
+// Hook for getting the initial sources for the current level.
+// If we are in start mode, we show the start sources or the default sources
+// if the level has no start sources.
+// If the user has a project, we prioritize that over the start sources. If there is
+// no project or start sources, we use the default sources.
 export const useInitialSources = (defaultSources: ProjectSources) => {
-  const labState = useAppSelector(state => state.lab);
+  const labInitialSources = useAppSelector(state => state.lab.initialSources);
+  const levelStartSource = useAppSelector(
+    state => state.lab.levelProperties?.source
```
sanchitmalhotra126, line 16: Just confirming that `levelProperties.source` is the right place for starter code to go? I know that in legacy labs we use either `start_blocks` or `start_sources` (maybe only for Javalab?). Music lab is a weird special case because the starter code is in `levelData` but I think it's called `startSources` there too? (either way, I'd like to move it out of `levelData` at some point anyway)

apps/src/codebridge/hooks/useSource.ts
```diff
@@ -0,0 +1,48 @@
...
+import {setAndSaveProjectSource} from '@cdo/apps/lab2/redux/lab2ProjectRedux';
+import {MultiFileSource, ProjectSources} from '@cdo/apps/lab2/types';
+import {useAppDispatch, useAppSelector} from '@cdo/apps/util/reduxHooks';
+
+import {useInitialSources} from './useInitialSources';
+
+// Hook for handling the project source for the current level.
+// Returns the current project source and a function to save the source.
+// This also handles displaying the levelbuilder save button in start mode.
+export const useSource = (defaultSources: ProjectSources) => {
```
sanchitmalhotra126, line 15: Slick!

sanchitmalhotra126 (approved):
LGTM, nice refactoring! Looking forward to moving the other lab2 labs onto this system

apps/src/codebridge/hooks/useInitialSources.ts
molly-moen: yeah `source` is the start source

PR #58478, opened 2024-05-08
[Python Lab] Support hidden support files

mikeharv:
This branch makes it possible for levelbuilders to use start mode to mark files hidden. The purpose of this is to be able to have support files included in project while not making the visible to students. Here is what the update start mode looks like:

image

1. File tabs will now show an `eye-slash` icon if the file won't be visible to students
2. An open or closed eye is also shown beside the file names in the browser
3. Another icon is added to the button bar that toggles a file's `visible` property. (It intentionally shows the opposite icon to the one beside the filename label.)

<hr/>

The only change to the student view is that only visible files will show up. This includes file tabs, and in the file browser. Code that is part of hidden files can still be imported and run as shown in the console below:

image

## Technical Details
* I tried to stick to existing conventions as much as possible. The new case in the reducer is modeled after `RENAME_FILE`, and I used the existing button bar rather than defining new components for now.
* I had to modify how we determine the single "active" file to exclude hidden files. If there are no active/visible files, we default to the first file. For example, if the curriculum writer left `hidden.py` active when saving, we would make `main.py` active for the student. This mostly relates to the file tab style and which file is loaded into the editor.
* I also added filtering for files listed in the file browser. If a file is hidden, we do not show it unless we're in start mode.
* The eye icons are a temporary UI as we're later likely to move to a single dropdown menu for files in the browser.
* As far as naming function goes, I went with `toggleFileVisibility` initially as we don't yet know the current visible state. Once we have the boolean we need, it gets passed as a second argument into `setFileVisibility`.
* There may a couple cases where there is some repeated logic to check for file visibility and/or start mode. I'm open to suggestions for optimization (or anything else)!

## Links

Jira - https://codedotorg.atlassian.net/browse/CT-549

## Follow-up work

If this flow looks good, it should be easy to add a similar path for setting a file as validation.

## PR Checklist:

- [ ] Code is well-commented


apps/src/lab2/projects/utils.ts
```diff
@@ -75,3 +77,19 @@ export function getFileByName(
...
+ */
+export function getActiveFileForProject(project: MultiFileSource) {
+  const files = Object.values(project.files);
+  const isStartMode = getAppOptionsEditBlocks() === START_SOURCES;
+  const activeFiles = files.filter(
+    f => f.active && (f.visible !== false || isStartMode)
+  );
+
+  // If the only active file is hidden, make the first file active.
+  return activeFiles?.[0] || files[0];
```
molly-moen: will this throw if `files` is empty? It's possible to have no files in a project if the user deletes all files.

apps/src/codebridge/FileBrowser/index.tsx
```diff
@@ -110,6 +127,20 @@ const InnerFileBrowser = React.memo(
                   <span onClick={() => deleteFile(f.id)}>
                     <i className="fa-solid fa-trash" />
                   </span>
+                  {isStartMode && (
+                    <span
+                      onClick={e => {
+                        e.stopPropagation();
```
molly-moen: why do we need `e.stopPropagation` here but not for other on clicks in this file?

apps/src/codebridge/FileBrowser/index.tsx
```diff
@@ -259,6 +291,18 @@ export const FileBrowser = React.memo(() => {
     [renameFolder, project.folders]
   );
 
+  const toggleFileVisibility: FilesComponentProps['toggleFileVisibility'] =
+    useMemo(
+      () => fileId => {
+        const file = project.files[fileId];
+        // file.visible could be undefined.
+        // We only want to show the file if it's currently explicitly hidden.
+        const visibility = file.visible === false;
```
molly-moen: could we avoid some of these explicit checks by changing the field from `visible` to `hidden`, then we could just do checks on `file.hidden`, and if it's undefined or false we will show the file?

apps/src/codebridge/FileBrowser/index.tsx
```diff
@@ -87,19 +92,31 @@ const InnerFileBrowser = React.memo(
...
+          )
           .sort((a, b) => a.name.localeCompare(b.name))
           .map(f => (
             <li key={f.id}>
               <span className="label">
-                <span onClick={() => openFile(f.id)}>{f.name}</span>
+                <span onClick={() => openFile(f.id)}>
+                  {isStartMode && (
+                    <i
+                      className={`fa-solid ${
```
molly-moen, line 111: would it be worth making a helper method for getting the icon based on the visibility? I see that repeated a few places.

molly-moen (commented):
Looks good, couple comments!

apps/src/codebridge/FileBrowser/index.tsx
```diff
@@ -87,19 +92,31 @@ const InnerFileBrowser = React.memo(
...
                 )}
               </li>
             );
           })}
         {Object.values(files)
-          .filter(f => f.folderId === parentId)
+          .filter(
+            f => f.folderId === parentId && (f.visible !== false || isStartMode)
+          )
           .sort((a, b) => a.name.localeCompare(b.name))
```
ebeastlake, line 104: Non-blocking: It seems like
```
Object.values(files)
          .filter(
            f => f.folderId === parentId && (f.visible !== false || isStartMode)
          )
          .sort((a, b) => a.name.localeCompare(b.name))
```

Could be a good candidate for extracting into a helper or separate variable, similar to how you did `getActiveFileForProject`?

apps/src/codebridge/codebridgeContext/codebridgeContextProjectReducer.ts
```diff
@@ -76,13 +78,25 @@ export const projectReducer = (project: ProjectType, action: ReducerAction) => {
...
+        action.payload
+      );
+
+      return {
+        ...project,
+        files: {
+          ...project.files,
+          [fileId]: {...project.files[fileId], visible: visibility},
+        },
+      };
```
ebeastlake, line 92: I think @molly-moen and I have chatted about the fact that I've used lodash a lot more on previous teams to do deep merges like this. The syntax would be:
```
  const updatedProject = _.merge({}, project, {
    files: {
      [fileId]: {
        visible: visibility
      }
    }
  });
```
No strong feelings either way, but something to think about for deep merges!

apps/src/lab2/projects/utils.ts
ebeastlake: If we render a useful "placeholder" screen if no (active?) files exist, it might be okay for `getActiveFileForProject` to return null (?).

ebeastlake:
Nice work! No blocking comments from me pending Molly's approval.

apps/src/codebridge/FileBrowser/index.tsx
mikeharv: I'm not sure. The logic is repeated in two places with the opposite result for each. E.g. if the file is hidden, label shows a closed eye but the button bar has an open eye. We could have a boolean parameter to specify which we want, but I wonder if it's worth it. I think it's also likely that we're going to cover the button bar to a single ellipsis icon with a dropdown menu, which should remove the similar logic from one of two places.

apps/src/codebridge/codebridgeContext/codebridgeContextProjectReducer.ts
mikeharv: Right now there are around 10 functions that work in a similar way, without lodash. Since you don't have strong feelings, I think I'd prefer to handle that feedback with a follow-up task to refactor the whole file.

apps/src/codebridge/FileBrowser/index.tsx
mikeharv: Hmm, we don't...

mikeharv: Hmm, we don't...

apps/src/lab2/projects/utils.ts
mikeharv: It doesn't throw an error. In the case of opening/activating files (in `codebridgeContextProjectReducer`) and rendering the editor (`codebridge/Editor/index`),  there are null checks already. That being said, Molly's point about deleting a file helped me find a related bug. If you delete your last visible file, we'd suddenly show your first hidden file. I'm reworking this so that we always return a visible file (or undefined).

apps/src/codebridge/FileBrowser/index.tsx
mikeharv: I had this nagging thought about that while I was working, but I think my conscience felt I was in too deep. You're totally correct though and I'm going to rework this. "Hidden" also better aligns to how we've been describing this feature as well.

apps/src/codebridge/FileBrowser/index.tsx
mikeharv: This is a little simpler now that I've renamed/reworked `.visible` to `.hidden`. PTAL

apps/src/codebridge/codebridgeContext/codebridgeContextProjectReducer.ts
ebeastlake: Yep, no worries! And not even necessarily something to bother refactoring -- just something to keep in mind when you get into deep merging in the future.

apps/src/codebridge/FileBrowser/index.tsx
molly-moen: that's fair, makes sense to leave as is

molly-moen (approved):
Nice!

apps/src/codebridge/FileBrowser/index.tsx
mikeharv: FWIW, switching to `.hidden` makes it slightly less complicated too.

PR #58483, opened 2024-05-08
Fix odd spacing on My PL page

TurnerRiley:
In the "Joined PL Sections" section of the My PL page, if the user isn't in any sections then the spacing looks odd and leaves little room before the "Recommended for you" section below. So, this PR adds conditional formatting if the user isn't enrolled in any PL sections.

### Previous "Joined PL Sections" with no sections (odd spacing)
old my pl no section

### Previous "Joined PL Sections" with sections (no change)
old my pl with section

### New "Joined PL Sections" with no sections (fixed spacing)
New joined pl no section

### New "Joined PL Sections" with sections (no change)
New joined pl with section

## Links
Jira ticket: [here](https://codedotorg.atlassian.net/jira/software/c/projects/ACQ/boards/64?assignee=60d62161f65054006980bd71&selectedIssue=ACQ-1865)

## Testing story
Local testing.

### DIFF
--- a/apps/src/code-studio/pd/professional_learning_landing/LandingPage.jsx
+++ b/apps/src/code-studio/pd/professional_learning_landing/LandingPage.jsx
@@ -104,6 +104,9 @@ function LandingPage({
     workshopsAsParticipant?.length === 0 &&
     plCoursesStarted?.length === 0;
 
+  const joinedPlSectionsStyling =
+    joinedPlSections?.length > 0 ? '' : style.joinedPlSectionsWithNoSections;
+
   // Load PL section info into redux
   const dispatch = useDispatch();
   useEffect(() => {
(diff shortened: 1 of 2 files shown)


hannahbergam (approved):
Makes sense to me! I could go either way on adding this kind of logic in vs. outside of the render, so happy to approve as-is:)

bethanyaconnor (approved):
Thank you for doing this!!

kelbyhawn (approved):
Woo hoo, thanks for doing this! 🎉

TurnerRiley:
> Makes sense to me! I could go either way on adding this kind of logic in vs. outside of the render, so happy to approve as-is:)

Oh true, actually for consistency's-sake within the file, I'll move it to outside of the render!

PR #58484, opened 2024-05-08
Gen AI: dedupe error logging

bencodeorg:
Checking out logs from today's bug bash, and noticed duplicate logs being generated. A result of a [last minute change](https://github.com/code-dot-org/code-dot-org/pull/58351/commits/fd722fc84c044be03dc3fac29998810b4f30cab7) to this PR, where I accidentally started logging twice when we observe profanity from the model 🤦 . This change resolves that duplication.

## Testing story

Tested manually that I could repro before this change, and no longer repro after.


thomasoniii (approved):
lgtm 🚀

PR #58486, opened 2024-05-09
[AI Tutor] CT-562: Add s3 system prompt

ebeastlake:
We want to keep our system prompt hidden from the user. I moved it to the `cdo-ai/tutor` bucket in S3. It now gets loaded server-side and appended to the conversation history before it gets sent to OpenAI.

So currently the flow is:
```
conversation history from client
 --> prepend system prompt on server
 --> system prompt & conversation history get sent to Open AI
 --> server returns most recent message
 --> client adds most recent message to history
```

☝️ We might want to modify this at some point, but since this is a last-minute change, I wanted to keep it as close to the original implementation as possible.

Other changes in this PR:
- Level instructions need to go to the server so they can be added to the system prompt
- There was an issue where 500 and other unexpected errors weren't failing gracefully in the UI; this PR changes that

## Links

Jira ticket: https://codedotorg.atlassian.net/browse/CT-562

## Testing story

Added unit tests for:
- read_file_from_s3
- prepend_system_prompt
- add_content_to_system_prompt

I tested manually that the code compilation and validation cases were still working as expected.

Screenshot 2024-05-09 at 9 02 24 PM

Screenshot 2024-05-09 at 9 08 49 PM

And also that error states are being handled as expected (by blocking network requests from the console).

Screenshot 2024-05-09 at 9 09 48 PM

I also tested that the questionable answers in the AI Tutor Bug Bash doc are all resolved by the new system prompt, see comments here: https://docs.google.com/document/d/1PAT43dP4lhuhhecQIaeb2ZnsH8dzU-1UaDwA-N4QAC8/edit#heading=h.kgu6m7hz2xke

## Follow-up work

I followed a file-up ticket to capture adding eyes tests for AI Tutor UI: https://codedotorg.atlassian.net/browse/CT-571

## PR Checklist:

- [ ] Code is well-commented


dashboard/app/controllers/openai_chat_controller.rb
```diff
@@ -39,7 +49,40 @@ def validated_level?
...
+      content: system_prompt,
+      role: "system"
+    }
+
+    messages.unshift(system_prompt_message)
+    messages
+  end
+
+  private def read_file_from_s3(key_path)
+    if [:development, :test].include?(rack_env)
```
molly-moen: is this to make local testing easier? Why do we need this for `test` too?

dashboard/app/controllers/openai_chat_controller.rb
```diff
@@ -39,7 +49,40 @@ def validated_level?
...
+      local_path = File.join("local-aws", S3_AI_BUCKET, key_path)
+      if File.exist?(local_path)
+        puts "Note: Reading AI prompt from local file: #{key_path}"
+        return File.read(local_path)
+      end
+    end
+    s3_client.get_object(bucket: S3_AI_BUCKET, key: key_path).body.read
```
molly-moen: will this cache the response? I'm guessing the system prompt won't change that much?

molly-moen (commented):
Couple questions, otherwise looks good

dashboard/app/controllers/openai_chat_controller.rb
ebeastlake: There's a thread relevant to this comment and the one above here: https://codedotorg.slack.com/archives/C051P2V2RN0/p1715274926003069?thread_ts=1715274404.871189&cid=C051P2V2RN0

I copied the dev/test logic from Teacher Tools. I originally had it cached and then disabled it because I was troubleshooting unrelated test failures that mentioned a cache, and Dave said it shouldn't be pilot-blocking. If it's okay with you, I'll reenable the caching, assuming it passes Drone. No strong feelings on the dev/test thing -- though it probably makes sense to disable it for test if we don't know the application :D

dashboard/app/controllers/openai_chat_controller.rb
ebeastlake: Locally, this is so I can try out an updated system prompt without replacing the contents in production. (AI TA might have needed this in test for their test suite, but we don't.) I'm removing the :test env here for now.

davidsbailey: nice catch and good call! the `:test` part looks like a mistake on our end because it would have no effect in automated testing and would potentially break running our tests locally.

dashboard/test/controllers/openai_chat_controller_test.rb
```diff
@@ -48,6 +57,14 @@ class OpenaiChatControllerTest < ActionController::TestCase
...
   response: :success
 
+  # Post request without a messages param returns a bad request
+  test_user_gets_response_for :chat_completion,
+  name: "no_messages_test",
+  user: :ai_tutor_access,
+  method: :post,
+  params: {},
+  response: :bad_request
+
```
davidsbailey, line 67: 🤩

davidsbailey (approved):
great job on thorough test coverage!

ebeastlake:
@molly-moen @davidsbailey Thoughts on this approach to caching? https://github.com/code-dot-org/code-dot-org/pull/58486/files/ff7859efb863474271ec287dfd9a66127a5b2a1a..6daa4f98764fed7b707937ceb5fe834ec799e2f2

I disabled it in dev because I feel like caching causes more problems than it solves locally, but the risk is this logic is only unit-tested before it hits test and prod. Should I be caching in the test environment? 🤔 Does it matter?

molly-moen (approved):
🎉

dashboard/app/controllers/openai_chat_controller.rb
```diff
@@ -72,14 +72,27 @@ def add_content_to_system_prompt(system_prompt, level_instructions, test_file_co
   end
 
   private def read_file_from_s3(key_path)
-    if [:development, :test].include?(rack_env)
+    cache_key = "s3_file_#{key_path}"
```
davidsbailey: since we don't know how else the shared cache is being used, I would suggest including the bucket name in the cache key.

molly-moen: @ebeastlake I don't know much about caching besides "we should probably do it here". It makes sense to me to run it on prod and maybe test too? I don't see a downside to running caching on test.

davidsbailey (approved):
LGTM after addressing comments

dashboard/app/controllers/openai_chat_controller.rb
```diff
@@ -72,14 +72,27 @@ def add_content_to_system_prompt(system_prompt, level_instructions, test_file_co
...
         return File.read(local_path)
       end
     end
-    s3_client.get_object(bucket: S3_AI_BUCKET, key: key_path).body.read
+
+    content = s3_client.get_object(bucket: S3_AI_BUCKET, key: key_path).body.read
+    # In production, cache the content after fetching it from S3
+    # Note: We will hit this codepath in dev if the file is not found locally
+    if Rails.env.production?
+      CDO_SHARED_CACHE.write(cache_key, content, expires_in: 1.hour)
```
davidsbailey: please use `CDO.shared_cache` for consistency

ebeastlake: Good catch, thanks Dave!

PR #58491, opened 2024-05-09
Music: fix preview plays

breville:
With this change, plays of sounds marked with a `type` of `"preview"` no longer have their pitch shifted or rate changed.  This allows previews of packs (both in the pack dialog and the sound panel) to play in their original form, even if the current project is set to a different BPM or key.

### DIFF
--- a/apps/src/music/player/MusicPlayer.ts
+++ b/apps/src/music/player/MusicPlayer.ts
@@ -353,6 +353,7 @@ export default class MusicPlayer {
           effects: soundEvent.effects,
           originalBpm: soundData.bpm || DEFAULT_BPM,
           pitchShift: this.calculatePitchShift(soundData),
+          preview: soundData.type === 'preview',
         },
       ];
     } else if (event.type === 'pattern') {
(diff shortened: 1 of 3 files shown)


thomasoniii (approved):
lgtm 🚀

apps/src/music/player/MusicPlayer.ts
```diff
@@ -353,6 +353,7 @@ export default class MusicPlayer {
           effects: soundEvent.effects,
           originalBpm: soundData.bpm || DEFAULT_BPM,
           pitchShift: this.calculatePitchShift(soundData),
+          preview: soundData.type === 'preview',
```
sanchitmalhotra126: Can we instead check this in `calculatePitchShift` and just return a pitch shift of 0 if `soundData.type === preview` (similar to how we do for `'beat'`)?

Looks like we'd still need some field to also indicate that we don't want tempo adjustment but then maybe we can rename this to `disableTempoAdjustment` or similar?

breville: Sure thing.  Done.

sanchitmalhotra126 (approved):

PR #58493, opened 2024-05-09
Add finish button to last level of a lab2 progression

fisher-alice:
This PR adds a 'Finish' button to the last level (lab2) of a lesson. After a user clicks on the 'Finish' button, we post to `milestone` so that the level bubble turns green (or purple for an assessment level). Then a message is displayed below the disabled button.

Before clicking:

Screenshot 2024-05-09 at 10 41 37 PM

After clicking:

Screenshot 2024-05-09 at 10 41 25 PM

Note that currently, validation has not been added to `aichat` levels. Also, for music lab levels, the 'Continue' or 'Finish' buttons appear after validation conditions are satisfied.
When the 'Continue'/'Finish' button is clicked, we post to `milestone` [with `idealPassResult`](https://github.com/code-dot-org/code-dot-org/blob/0caf8a106432725986d8037ca27a43b4e7b56898/apps/src/code-studio/progressRedux.ts#L351-L352) so that we will always pass or attain a green (or purple for assessment levels) bubble after clicking the button.

Future work includes sending accurate data to the backend and consuming the server response, i.e., the `redirect` url.

Scrreencast videos AFTER update:

https://github.com/code-dot-org/code-dot-org/assets/107423305/3a74fca0-0e0c-4716-85bc-b0866a39f911

https://github.com/code-dot-org/code-dot-org/assets/107423305/ab3700c9-39a7-4a34-a800-4814ef0bfde6

## Links
[jira](https://codedotorg.atlassian.net/browse/LABS-768)

## Testing story
Tested locally in both music lab and aichat progressions.

## Follow-up work
Currently, an assessment level does not turn 'bubble' after 'Continue' or 'Finish' button is clicked until the page reloads. Investigate ~~and resolve if possible~~. 
Update: this is existing behavior but not noticeable in legacy labs since page always reloads on level change. Here's an example in a CSP progression. Note that bubble turns green immediately for an app lab level even before page reload but not for an assessment level - at 0:25 in video.

https://github.com/code-dot-org/code-dot-org/assets/107423305/811fd750-e8e0-4fa7-b1fc-a64d7a0f6bf5

I see in `ProgressBubble.jsx` that assessment levels render a progress bubble differently.
https://github.com/code-dot-org/code-dot-org/blob/48feb7d3930f4cb6560a01abb9faaf056f7697d8/apps/src/templates/progress/ProgressBubble.jsx#L70-L85

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/lab2/views/components/Instructions.tsx
+++ b/apps/src/lab2/views/components/Instructions.tsx
@@ -68,6 +68,9 @@ const Instructions: React.FunctionComponent<InstructionsProps> = ({
   const showNextButton =
     (!hasConditions || satisfied) && levelIndex + 1 < currentLevelCount;
 
+  const showFinishButton =
+    (!hasConditions || satisfied) && levelIndex + 1 === currentLevelCount;
+
   const dispatch = useAppDispatch();
 
   const {theme} = useContext(ThemeContext);
@@ -90,6 +93,8 @@ const Instructions: React.FunctionComponent<InstructionsProps> = ({
       message={message || undefined}
       messageIndex={index}
       showNextButton={showNextButton}
+      showFinishButton={showFinishButton}
+      beforeNextLevel={beforeNextLevel}
       onNextPanel={onNextPanel}
       theme={theme}
       {...{baseUrl, layout, imagePopOutDirection, handleInstructionsTextClick}}
@@ -108,6 +113,10 @@ interface InstructionsPanelProps {
   imageUrl?: string;
   /** If the next button should be shown. */
   showNextButton?: boolean;
+  /** If the finish button should be shown. */
+  showFinishButton?: boolean;
+  /** Additional callback to fire before navigating to the next level. */
+  beforeNextLevel?: () => void;
   /** Callback to call when clicking the next button. */
   onNextPanel?: () => void;
   /** If the instructions panel should be rendered vertically or horizontally. Defaults to vertical. */
@@ -135,6 +144,8 @@ const InstructionsPanel: React.FunctionComponent<InstructionsPanelProps> = ({
   messageIndex,
   imageUrl,
   showNextButton,
+  showFinishButton,
+  beforeNextLevel,
   onNextPanel,
   layout = 'vertical',
   imagePopOutDirection = 'right',
@@ -150,6 +161,15 @@ const InstructionsPanel: React.FunctionComponent<InstructionsPanelProps> = ({
   const vertical = layout === 'vertical';
 
   const canShowNextButton = showNextButton && onNextPanel;
+
+  const canShowFinishButton = showFinishButton;
+
+  const onFinish = () => {
+    if (beforeNextLevel) {
+      beforeNextLevel();
+    }
+    console.log('Congrats! You finished!');
+  };
 
   return (
     <div
@@ -216,7 +236,7 @@ const InstructionsPanel: React.FunctionComponent<InstructionsPanelProps> = ({
             />
           </div>
         )}
-        {(message || canShowNextButton) && (
+        {(message || canShowNextButton || canShowFinishButton) && (
           <div
             key={messageIndex + ' - ' + message}
             id="instructions-feedback"
@@ -238,9 +258,19 @@ const InstructionsPanel: React.FunctionComponent<InstructionsPanelProps> = ({
                   id="instructions-feedback-button"
                   type="button"
                   onClick={onNextPanel}
-                  className={moduleStyles.buttonNext}
+                  className={moduleStyles.buttonContinue}
                 >
                   {commonI18n.continue()}
+                </button>
+              )}
+              {canShowFinishButton && (
+                <button
+                  id="instructions-feedback-button"
+                  type="button"
+                  onClick={onFinish}
+                  className={moduleStyles.buttonContinue}
+                >
+                  {commonI18n.finish()}
                 </button>
               )}
             </div>
--- a/apps/src/lab2/views/components/Instructions.tsx
+++ b/apps/src/lab2/views/components/Instructions.tsx
@@ -9,7 +9,6 @@ import {
   currentLevelIndex,
 } from '@cdo/apps/code-studio/progressReduxSelectors';
 import {useAppDispatch, useAppSelector} from '@cdo/apps/util/reduxHooks';
-import Button from '@cdo/apps/componentLibrary/button/Button';
 import {Heading6} from '@cdo/apps/componentLibrary/typography';
 import {LabState} from '../../lab2Redux';
 import {ProjectLevelData} from '../../types';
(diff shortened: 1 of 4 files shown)


apps/src/lab2/views/components/Instructions.tsx
```diff
@@ -233,15 +261,27 @@ const InstructionsPanel: React.FunctionComponent<InstructionsPanelProps> = ({
...
+              {canShowFinishButton && (
+                <Button
+                  text={commonI18n.finish()}
+                  disabled={isFinished}
+                  onClick={onFinish}
+                  className={moduleStyles.button}
+                />
+              )}
+              {finalMessage && isFinished && (
+                <EnhancedSafeMarkdown
```
sanchitmalhotra126: Out of curiosity, why do we need to use markdown here (instead of just CSS-styled text)?

apps/src/lab2/views/components/Instructions.tsx
```diff
@@ -134,22 +147,37 @@ const InstructionsPanel: React.FunctionComponent<InstructionsPanelProps> = ({
...
   };
 
   const vertical = layout === 'vertical';
 
-  const canShowNextButton = showNextButton && onNextPanel;
+  const canShowContinueButton = showContinueButton && onNextPanel;
+
+  const canShowFinishButton = showFinishButton;
+
+  const onFinish = () => {
```
sanchitmalhotra126: nit: `useCallback`

apps/src/lab2/views/components/Instructions.tsx
```diff
@@ -106,8 +115,12 @@ interface InstructionsPanelProps {
...
   /** Optional image URL to display. */
   imageUrl?: string;
-  /** If the next button should be shown. */
-  showNextButton?: boolean;
+  /** If the continue button should be shown. */
+  showContinueButton?: boolean;
+  /** If the finish button should be shown. */
+  showFinishButton?: boolean;
+  /** Additional callback to fire before navigating to the next level. */
+  beforeNextLevel?: () => void;
```
sanchitmalhotra126: nit: maybe better to call this `beforeFinish` or something to more clearly indicate what this callback is used for? Technically this inner component doesn't know that `beforeNextLevel` happens to already be wrapped by the `onNextPanel` callback

apps/src/lab2/views/components/instructions.module.scss
```diff
@@ -145,16 +145,15 @@
...
-  color: $neutral_light;
-  width: 100%;
-  padding: 5px 10px;
-  border-radius: 4px;
-  height: 38px;
-  margin: 0 0 10px 0;
+.button {
+  width: 100%;  
+  margin-bottom: 10px !important;
+  &:disabled, &[aria-disabled="true"] {
```
sanchitmalhotra126, line 158: Curious, do the default DSCO button styles not work here?

sanchitmalhotra126 (commented):
Nice! This seems like a solid temp solution for now, and we can revisit in more depth once we have a better understanding of what we want to do across all Lab2 labs. Just a few minor comments.

apps/src/lab2/views/components/Instructions.tsx
```diff
@@ -106,8 +115,12 @@ interface InstructionsPanelProps {
   messageIndex?: number;
   /** Optional image URL to display. */
   imageUrl?: string;
-  /** If the next button should be shown. */
-  showNextButton?: boolean;
```
cnbrenci, line 110: Nit: optional suggestion for this interface & in general - IMO it's preferrable to give a variable/param a more descriptive name rather than a short name with extra description in a comment. That way you know the full meaning wherever that var is used as you're reading the code, and it's less likely to get out of date as the code changes (easier to forget to update the comment than the var name).

some examples: 
`text` -> `primaryDisplayText`
`showContinueButton` -> `shouldShowContinueButton`
`showFinishButton` -> `shouldShowFinishButton`
`beforeNextLevel` -> `onBeforeNextLevelNavigation` 
`onNextPanel` -> `onNextButtonClick`

These name changes would cause the extra comments to be redundant and they could be removed.

apps/src/lab2/views/components/Instructions.tsx
```diff
@@ -134,22 +147,37 @@ const InstructionsPanel: React.FunctionComponent<InstructionsPanelProps> = ({
...
+
+  const onFinish = () => {
+    if (beforeNextLevel) {
+      beforeNextLevel();
+    }
+    setIsFinished(true);
+  };
+
+  const finalMessage =
+    '<strong>You finished this lesson! Check in with your teacher for the next activity.</strong>';
```
cnbrenci: Does this need to be translated?

apps/src/lab2/views/components/Instructions.tsx
fisher-alice: Thanks for the question! I was copying what was done with  `message` above but on second thought, I'll use a `Typography` element.

apps/src/lab2/views/components/instructions.module.scss
fisher-alice: Thanks for catching this! I added this back when I was using `<button>` and added this before deciding to use the DSCO button. Will remove.

apps/src/lab2/views/components/Instructions.tsx
fisher-alice: Thanks! I probably lean towards longer descriptive names, but I've gotten feedback in the past to shorten variable names. So I think I'll keep as is for now, but definitely worth a team discussion!

apps/src/lab2/views/components/Instructions.tsx
fisher-alice: This is a temporary UI solution and will most likely be iterated on in the next couple weeks. We wanted something in place for the pilot so I think maybe for now, keeping as a constant makes sense until we land on a more permanent solution. Thanks for bring up!

apps/src/lab2/views/components/Instructions.tsx
fisher-alice: That makes sense - thanks!

apps/src/lab2/views/components/instructions.module.scss
```diff
@@ -145,16 +145,9 @@
...
-  border-color: $brand_secondary_default;
-  color: $neutral_light;
-  width: 100%;
-  padding: 5px 10px;
-  border-radius: 4px;
-  height: 38px;
-  margin: 0 0 10px 0;
+.button {
+  width: 100%;  
+  margin-bottom: 10px !important;
```
fisher-alice, line 150: @levadadenys - lmk if there's a way to avoid using `!important` here. Thanks!

apps/src/lab2/views/components/Instructions.tsx
```diff
@@ -134,22 +148,44 @@ const InstructionsPanel: React.FunctionComponent<InstructionsPanelProps> = ({
...
+
+  const onFinish = useCallback(() => {
+    if (beforeFinish) {
+      beforeFinish();
+    }
+    setIsFinished(true);
+  }, [beforeFinish]);
+
+  // When the level changes, the Finish button is active again if user returns to last level.
+  useEffect(() => {
```
fisher-alice, line 182: @sanchitmalhotra126 @cnbrenci - since you reviewed, I added this so that 'Finish' button is no longer disabled when user leaves the last level.

apps/src/lab2/views/components/instructions.module.scss
fisher-alice: FYI - After discussing with @breville , we're going to keep the style as is to not making any changes, even if subtle, before the music lab launch. After the two launches, we can make a lab2-wide decision on what buttons to use for common components. Thanks!

apps/src/lab2/views/components/Instructions.tsx
sanchitmalhotra126: Not blocking but does it work if you use `canShowFinishButton` as the dependency for this useEffect? i.e. reset the finish button state whenever we go from showing the finish button to not showing the finish button? Asking only because I'd like to avoid using redux in this inner component if possible (the idea was this inner component could be used standalone outside of the lab2/progress redux framework, but currently it's only used by the parent Instructions component, so not a big deal).

sanchitmalhotra126 (approved):

apps/src/lab2/views/components/Instructions.tsx
fisher-alice: Oh no! I didn't see this until after I merged. Sorry about that! I can do a fast folllow-up for this. Thanks!

fisher-alice: https://github.com/code-dot-org/code-dot-org/pull/58528

PR #51602, opened 2023-04-28
Adds CPA messaging to create section modal.

wilkie:
Currently, stuck behind the "cpa_experience" flag.

Fairly straightforward in the end. Tricky to remove styling specific to maintaining the height of the modal. I'm not really sure, and can not get guidance, about why it was done this way. Bisecting the code reveals folks maintaining this height. Fixed heights like this are not reasonable for localization... we never know how long the strings will be.

The "Learn More" is a link via the `<Button>` component which I learned can be rendered as an accessible link. It reveals a pop-over containing the messaging text. For screen-reader accessibility, it would be nice if the appropriate "`aria-expands`" tag was being used, however. But, the pop-over text is readable by screen-readers as part of the modal itself.

For _right-to-left_ languages, the modal renders reasonably. We already have a "Learn more" translation for most languages. The icon appears to the right correctly. The Learn More modal does not place the close button in the appropriate corner, however. This is a style bug with `StylizedBaseDialog` / `BaseDialog` and beyond the scope of this feature.

## Links

- jira ticket: [P20-63](https://codedotorg.atlassian.net/browse/P20-63)

## Testing story

Manual testing via the flag and storybook testing for seeing rendering.

Flag testing via query parameter. After deploy, can see changed modal via adding `?cpa_experience=1` to the address URL.

Results below.

**BEFORE**

image

**AFTER**

image

**CLICKING LEARN MORE**

image

**WITH SSO OPTIONS**

image

**CLICKING LEARN MORE**

image

**RIGHT TO LEFT**

My notes about this in the summary at the top.

image

image

## PR Checklist:

- [ ] Tests provide adequate coverage
- [ ] Privacy and Security impacts have been assessed
- [ ] Code is well-commented
- [ ] New features are translatable or updates will not break translations
- [ ] Relevant documentation has been added or updated
- [ ] User impact is well-understood and desirable
- [ ] Pull Request is labeled appropriately
- [ ] Follow-up work items (including potential tech debt) are tracked and linked

PR #57919, opened 2024-04-10
added CAP Notification to SectionSetUpContainer

mgc1194:
As part of the Colorado Privacy Act requirements, we want to show all US teachers, that create a new section using student personal logins, a warning message. This warning message intends to persuade teachers to use other student authentication methods (secret words, picture passwords, or school-managed accounts SSO) if any of their students are under 13.
The Notification also links to the parental permission support article.

This PR adds a warning banner on the Create New Section page. The banner shows when teachers try to create new sections that are set up using students' personal emails and are in the USA.

Unit tests have been added to cover the use of the newly added Notice.

## Links

- spec: [Colorado Privacy Act Part Two](https://docs.google.com/document/d/1eDd8ow0kp7Ganv6aJ6GjiynPQSxB2XO21OxWuhm1Z1c/edit#heading=h.qi8xxyq72bct)
- jira ticket: [P20-816](https://codedotorg.atlassian.net/browse/P20-816)

## Testing story
Teacher creates a new section using **personal email**:
Screenshot 2024-04-10 at 09 19 38

Teacher creates a new section using **secret words**:
Screenshot 2024-04-10 at 09 23 16

Teacher creates a new section using **picture password**:
Screenshot 2024-04-10 at 09 23 44

The banner renders in RTL languages:
Screenshot 2024-04-10 at 09 33 34

## PR Checklist:

- [ ] Tests provide adequate coverage
- [ ] Privacy and Security impacts have been assessed
- [ ] Code is well-commented
- [ ] New features are translatable or updates will not break translations
- [ ] Relevant documentation has been added or updated
- [ ] User impact is well-understood and desirable
- [ ] Pull Request is labeled appropriately
- [ ] Follow-up work items (including potential tech debt) are tracked and linked

PR #58497, opened 2024-05-09
P20-903: Display U13 Warning on Section Creation to USA teachers

artem-vavilov:
## Links
- JIRA ticket: [P20-903](https://codedotorg.atlassian.net/browse/P20-903)
- Related PR: https://github.com/code-dot-org/code-dot-org/pull/51602


dashboard/app/controllers/api/v1/users_controller.rb
```diff
@@ -41,6 +41,7 @@ def current
         ai_tutor_access_denied: !!current_user.ai_tutor_access_denied,
         has_seen_progress_table_v2_invitation: current_user.has_seen_progress_table_v2_invitation?,
         date_progress_table_invitation_last_delayed: current_user.date_progress_table_invitation_last_delayed,
+        us_state: current_user.us_state,
```
juanmanzojr: Do we currently have `us_state` for teachers? I was under the impression we dont retrieve `us_state` for teachers but instead will use the school's state.

artem-vavilov: @wilkie @daynew do you know something about this?

daynew: You will need to look up the school_info associated with the teacher. There is `state` attribute on the SchoolInfo model. When @mgc1194 and I were looking at it, we saw values like `CO` and `Colorado`, so you will need to do some pattern matching.

artem-vavilov: https://github.com/code-dot-org/code-dot-org/pull/58497/commits/fb4fd19d373cee964fb82d5d4e3df2fc4c46841f

dashboard/app/models/user.rb
```diff
@@ -2737,6 +2737,20 @@ def uuid
     id && Digest::UUID.uuid_v5(Dashboard::Application.config.secret_key_base, id.to_s)
   end
 
+  # @return [String, nil] the user's US state code in the ISO 3166-2:US standard
+  def us_state_code
+    state = student? ? us_state : school_info&.state
+    return if state.blank?
+
+    # {CO: 'Colorado', DC: 'District Of Columbia', ...}
+    state_code_to_name_map = CS.states(:US)
```
daynew, line 2746: I like this way normalizing the "state" values.

apps/src/templates/teacherDashboard/LoginTypePicker.jsx
```diff
@@ -88,6 +89,11 @@ class LoginTypePicker extends Component {
     const withClever =
       providers && providers.includes(OAuthSectionTypes.clever);
     const hasThirdParty = withGoogle | withMicrosoft | withClever;
+    const currentUser = getStore().getState().currentUser;
+    const showStudentsToSectionPermissionWarning =
+      currentUser.isTeacher &&
+      (['US'].includes(currentUser.countryCode) || !!currentUser.usStateCode) &&
```
daynew: `country_code` is an attribute on the User model which I think is only set for students when we are asking for their us_state. So this won't be helpful for Teacher accounts. To be more consistent, we should look at the detected country in the `request` object: `request.country.to_s.upcase` (or something similar). I see the Mario recently did something similar to pass the value from the backend to the front end: https://github.com/code-dot-org/code-dot-org/pull/57919/files

daynew (changes requested):
Looks good except I think we should make the country detection consistent with all the CAP features.

apps/src/templates/teacherDashboard/LoginTypePicker.jsx
artem-vavilov: https://github.com/code-dot-org/code-dot-org/pull/58497/commits/98b792955f4ca60e8d3f679e895ebb9b514c96a4

dashboard/app/helpers/users_helper.rb
```diff
@@ -527,4 +527,10 @@ def lti_user_info_required?(user)
 
     Policies::Lti.lti?(current_user)
   end
+
+  def country_code(user, request)
+    return user.country_code if user.student?
+
+    user.country_code.presence || request.country.to_s.upcase
```
daynew, line 534: if `request.country` == 'RD', we usually treat it as 'US'. Should we convert to 'RD' -> 'US' here or update `(['US'].includes(currentUser.countryCode) || !!currentUser.usStateCode)` to check for RD too? I will let you decide.

daynew (approved):

apps/src/templates/teacherDashboard/LoginTypePicker.jsx
```diff
@@ -88,6 +89,14 @@ class LoginTypePicker extends Component {
...
       providers && providers.includes(OAuthSectionTypes.clever);
     const hasThirdParty = withGoogle | withMicrosoft | withClever;
+    const currentUser = getStore().getState().currentUser;
+    const inUSA =
+      ['US', 'RD'].includes(currentUser.countryCode) ||
+      !!currentUser.usStateCode;
+    const showStudentsToSectionPermissionWarning =
+      inUSA &&
+      currentUser.isTeacher &&
+      experiments.isEnabledAllowingQueryString(experiments.CPA_EXPERIENCE);
```
mgc1194, line 98:
```
(inUSA &&
currentUser.isTeacher) ||
experiments.isEnabledAllowingQueryString(experiments.CPA_EXPERIENCE);
```
Should this be an OR statement? My understanding is that the experiment is a way for us to override the conditions for the banner to appear.

mgc1194 (commented):
Looks great! As always, leaving the code better than when you found it Artem!!
I left one small comment, I´m not sure if you are setting the query parameters somewhere else.

apps/src/templates/teacherDashboard/LoginTypePicker.jsx
artem-vavilov: cc @wilkie

wilkie: Yep. Should be an `||` as Mario suggests, I think.

With the `&&` it would only show the banner if the `cpa_experience` DCDO flag is set (or you set it in the URL or via `enableExperiments=`). That flag is not set in production, so it wouldn't show at all. I think the intent is to actually roll this out.

That flag should just force you into a situation where things are acting like they would if you were a targeted user and the entire feature is live.

I think I like updating it to an `||` so it's live on deploy that it can still be forced on for testing via that flag.

wilkie (approved):
Looks great! I do think we should do as Mario suggests with the experiment logic and roll out this warning now that it has its dedicated audience.

Great work!

apps/src/templates/teacherDashboard/LoginTypePicker.jsx
artem-vavilov: https://github.com/code-dot-org/code-dot-org/pull/58497/commits/26a8e90f87a4f12377cdfe2d65b9ec8442c77107

mgc1194 (approved):

PR #58498, opened 2024-05-09
Gen AI: hide scrollbar when not needed

bencodeorg:
Seeing a scrollbar on bot responses on Windows/Chrome when scroll isn't needed. Setting to `auto` allows for scroll (and shows scrollbar) when needed, but hides otherwise.

**Before (fixed now)**

image

## Links

- [css overflow docs](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow)

## Testing story

Tested on Dan's machine that the scrollbar was visible in a case where it was needed, and is otherwise not shown.

### DIFF
--- a/apps/src/aichat/views/chatMessage.module.scss
+++ b/apps/src/aichat/views/chatMessage.module.scss
@@ -45,7 +45,7 @@
   background-color: $light_gray_50;
   border-radius: 0 16px 16px 16px;
   margin-top: 14px; // a quarter of height of bot icon
-  overflow-x: scroll;
+  overflow-x: auto;
 }
 
 .inappropriateMessage {


apps/src/aichat/views/chatMessage.module.scss
```diff
@@ -45,7 +45,7 @@
   background-color: $light_gray_50;
   border-radius: 0 16px 16px 16px;
   margin-top: 14px; // a quarter of height of bot icon
-  overflow-x: scroll;
+  overflow-x: auto;
```
sanchitmalhotra126, line 48: Do we want to allow horizontal scrolling at all actually?

bencodeorg: Yes, there's a case where content was overflowing but wasn't scrollable (tbh I don't totally understand when it overflows vs. just wraps to the next line). More detail [here](https://github.com/code-dot-org/code-dot-org/pull/58449#discussion_r1593089650)

bencodeorg: If there's a way to force the example in that comment to wrap that would also work

bencodeorg: A `white-space: nowrap` style we have applied to `code` HTML elements seems to prevent it from wrapping in the case I added this change for initially...I could potentially override that styling here (see screenshot where I've disabled it) rather than allowing the content to scroll? But a) I'm not sure if this example is the only situation where we'd have content coming back from the model that won't wrap, and b) I kinda like the scroll experience in this case better than the wrap?

image

sanchitmalhotra126: Gotcha, yeah this is fine for now then

sanchitmalhotra126 (approved):

PR #58507, opened 2024-05-09
disabled last scenario in ai_evaluate_student_code

Nokondi:
skipping final scenario in ai_evaluate_student_code ui test pending investigation into failures


snickell (approved):
TY @Nokondi 🙏

PR #58510, opened 2024-05-09
Rearrange eval rubric job file and extract helper classes

davidsbailey:
Follow-up from https://github.com/code-dot-org/code-dot-org/pull/58411#discussion_r1594321347 . also extracts some code which tutor may want to share soon.

no expected changes in behavior.

## Testing story

relying on existing test coverage, including
* unit tests, updated in this PR
* UI tests which run the job's success path using the test_ai_proxy_controller


Nokondi (approved):
I like this refactor - looks much cleaner.

PR #58512, opened 2024-05-09
Disable test that was flaky from the new V2 dashboard

kobryan0619:
Skipping a test after we tried debugging and couldn't find a solution - it appears to be flaky.  [Thread here](https://codedotorg.slack.com/archives/C0T0PNTM3/p1715290571199679)

[Adding a ticket to jira to debug](https://codedotorg.atlassian.net/jira/software/projects/TEACH/boards/65?selectedIssue=TEACH-1092)


snickell (approved):
🙏

PR #58533, opened 2024-05-10
Cleaning up School Infos validation and dedup processing

hannahbergam:
We used to send a honeybadger error if a user was trying to submit data that didn't align with our school database data. This usually meant they were selecting a school type (i.e. private, charter, public) that mismatched their accurate school type. Unfortunately, the old search didn't filter the search results on that input, so these mismatches were entirely possible. 

In the new search, the zip code will just return the schools without asking for a school type at all, so it's no longer relevant. And for the old search, we want to override the info with school data anyway (which means the Honeybadger event is no longer necessary). 

Bethany and I looked at this together and started working on it a few weeks ago, with her draft work in this pr: https://github.com/code-dot-org/code-dot-org/pull/58259/files.

## Links

- spec: []()
- jira ticket: https://codedotorg.atlassian.net/browse/ACQ-1843

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/dashboard/lib/school_info_deduplicator.rb
+++ b/dashboard/lib/school_info_deduplicator.rb
@@ -19,7 +19,7 @@ module SchoolInfoDeduplicator
 
     final_attr = process_school_info_attributes(school_info_attr)
 
-    return false unless SchoolInfo.new(final_attr).valid?
+    return nil unless SchoolInfo.new(final_attr).valid?
 
     SchoolInfo.where(final_attr).first
   end
--- a/dashboard/lib/school_info_deduplicator.rb
+++ b/dashboard/lib/school_info_deduplicator.rb
@@ -15,7 +15,7 @@ module SchoolInfoDeduplicator
   # Returns the SchoolInfo if it already exists.
   # Returns nil if the SchoolInfo is new or if it is invalid.
   def get_duplicate_school_info(school_info_attr)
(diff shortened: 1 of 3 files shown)


dashboard/lib/school_info_deduplicator.rb
```diff
@@ -15,22 +15,13 @@ def deduplicate_school_info(school_info_attr, self_object)
...
-    school_info =
-      if attr[:school_id]
-        SchoolInfo.where(school_id: attr[:school_id], validation_type: SchoolInfo::VALIDATION_FULL).first
-      else
-        SchoolInfo.where(attr).first
-      end
-    if school_info
-      return school_info
-    end
+    return false unless SchoolInfo.new(final_attr).valid?
```
bethanyaconnor: Thinking through this -- the only change I see here is that this return is now later. If I create a school info with an invalid school_id, it throws an error, but then we wouldn't find it above anyway. So it would now return nil instead of false.

Is there anywhere that would care about false vs nil? If not, could we maybe change this to nil?

bethanyaconnor (approved):
One rambling thinking-out-loud comment that ended in a nit but otherwise this looks great! Thanks for this clean up!

dashboard/lib/school_info_deduplicator.rb
hannahbergam: Ahhh yes good point- I updated to nil!

PR #58534, opened 2024-05-10
Add "Start Over" to CodeBridge

molly-moen:
This adds a start over button to the CodeBridge workspace. To get this working I had to do a bit of refactoring:
- Instead of configuring the file tabs and editor as separate components in the layout, they are combined as one`workspace` component. In order to enable this I removed the direct passing of the editor component and instead pass the language mapping and editable file types to the config. This also had the benefit of moving the header for the editor above the file tabs.
- I added the same start over icon button to the new "Workspace" header that we use in Music Lab.
- I updated the `useSources` hook to return a start over function as well.  I was able to reuse logic from Music Lab/lab2 for this.
- I re-used the same start over dialog from Music Lab/lab2. This dialog specifically mentioned "blocks" in the text. I updated it to check the lab type, and show "code" instead if the lab is text-based.

The new flow now looks like this (it is very similar in Web Lab 2):
https://github.com/code-dot-org/code-dot-org/assets/33666587/1ea7ec39-0321-43ae-a6a4-ff7e7b92c51f

## Links
- jira ticket: [CT-552](https://codedotorg.atlassian.net/browse/CT-552)

## Testing story
Tested locally in Python Lab and Web Lab 2

## Follow-up work
We have a separate task to add full version history.

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/i18n/common/en_us.json
+++ b/apps/i18n/common/en_us.json
@@ -2633,6 +2633,7 @@
   "startOverTitle": "Are you sure you want to start over?",
   "startOverBody": "This will reset the puzzle to its start state and reset all the data you've added or changed.",
   "startOverWorkspace": "This will reset the workspace to its start state and remove all the blocks you've added or changed.",
+  "startOverWorkspaceText": "This will reset the workspace to its start state and remove all the code you've added or changed.",
   "startWithUnit": "Start with unit:",
   "statsTableFailure": "Sorry, something went wrong. Please reload the page to try again.",
   "stayHere": "Stay here",
--- a/apps/src/codebridge/Editor/index.tsx
+++ b/apps/src/codebridge/Editor/index.tsx
@@ -39,7 +39,6 @@ export const Editor = ({langMapping, editableFileTypes}: EditorProps) => {
     return <div>Cannot currently edit files of type {file.language}</div>;
   }
 
-  console.log(`rendering editor for file ${file.id}/${1}`);
   return (
     <div className="editor-container">
       {file && (
--- a/apps/src/codebridge/Editor/index.tsx
+++ b/apps/src/codebridge/Editor/index.tsx
@@ -1,5 +1,4 @@
 import {useCodebridgeContext} from '@codebridge/codebridgeContext';
-import {editableFileType} from '@codebridge/utils';
 import {LanguageSupport} from '@codemirror/language';
 import React, {useCallback, useMemo} from 'react';
 
@@ -7,19 +6,17 @@ import {getActiveFileForProject} from '@cdo/apps/lab2/projects/utils';
 import CodeEditor from '@cdo/apps/lab2/views/components/editor/CodeEditor';
 
 import './styles/editor.css';
+import {editableFileType} from '../utils';
 
-export const Editor = (
-  langMapping: {[key: string]: LanguageSupport},
-  editableFileTypes: string[]
-) => {
+interface EditorProps {
+  langMapping: {[key: string]: LanguageSupport};
+  editableFileTypes: string[];
+}
+
+export const Editor = ({langMapping, editableFileTypes}: EditorProps) => {
   const {project, saveFile} = useCodebridgeContext();
 
   const file = getActiveFileForProject(project);
-  // this is a stupid hack. the low level code-mirror editor won't update itself
-  // automatically if the doc is changed externally. So in lieu of doing it better,
-  // for now we just key the component off of the file ID + an incrementing value that
-  // hits every time the format button is pressed. That'll force a re-render and make it work.
-  // let's swap this out with something better.
 
   const onChange = useCallback(
     (value: string) => {
@@ -42,6 +39,7 @@ export const Editor = (
     return <div>Cannot currently edit files of type {file.language}</div>;
   }
 
+  console.log(`rendering editor for file ${file.id}/${1}`);
   return (
     <div className="editor-container">
       {file && (
--- a/apps/src/codebridge/Workspace/HeaderButtons.tsx
+++ b/apps/src/codebridge/Workspace/HeaderButtons.tsx
@@ -15,8 +15,6 @@ const WorkspaceHeaderButtons: React.FunctionComponent = () => {
   const onClickStartOver = useCallback(() => {
     if (dialogControl) {
       dialogControl.showDialog(DialogType.StartOver, resetProject);
-    } else {
-      console.log('no dialog control');
     }
   }, [dialogControl, resetProject]);
 
--- a/apps/src/codebridge/Workspace/HeaderButtons.tsx
+++ b/apps/src/codebridge/Workspace/HeaderButtons.tsx
@@ -1,0 +1,37 @@
+import React, {useCallback, useContext} from 'react';
+
+import Button from '@cdo/apps/componentLibrary/button';
+import {
+  DialogContext,
+  DialogType,
+} from '@cdo/apps/lab2/views/dialogs/DialogManager';
+
+import {useCodebridgeContext} from '../codebridgeContext';
+
+const WorkspaceHeaderButtons: React.FunctionComponent = () => {
+  const dialogControl = useContext(DialogContext);
+  const {resetProject} = useCodebridgeContext();
+
+  const onClickStartOver = useCallback(() => {
+    if (dialogControl) {
+      dialogControl.showDialog(DialogType.StartOver, resetProject);
+    } else {
+      console.log('no dialog control');
+    }
+  }, [dialogControl, resetProject]);
+
+  return (
+    <div>
+      <Button
+        icon={{iconStyle: 'solid', iconName: 'refresh'}}
+        isIconOnly
+        color={'black'}
+        onClick={onClickStartOver}
+        ariaLabel={'Start Over'}
+        size={'xs'}
+      />
+    </div>
+  );
+};
+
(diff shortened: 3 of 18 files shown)


apps/src/lab2/views/dialogs/StartOverDialog.tsx
```diff
@@ -11,13 +13,21 @@ const StartOverDialog: React.FunctionComponent<BaseDialogProps> = ({
   handleConfirm,
   handleCancel,
 }) => {
+  const currentAppName = useAppSelector(
```
molly-moen, line 16: @sanchitmalhotra126 I'm pretty confident this won't impact Music Lab since we are only using the new translated string if we are in a text-based lab. LMK if this is too risky during the launch period.

apps/src/weblab2/Weblab2View.tsx
```diff
@@ -208,8 +204,6 @@ const Weblab2View = () => {
               if (configName === 'project') {
                 setSource(newConfig as ProjectType);
               } else if (configName === 'config' || configName === 'layout') {
-                (newConfig as ConfigType).EditorComponent =
```
molly-moen, line 211: @thomasoniii is it ok that I removed this since we don't support providing an editor component anymore?

apps/i18n/common/en_us.json
```diff
@@ -2633,6 +2633,7 @@
   "startOverTitle": "Are you sure you want to start over?",
   "startOverBody": "This will reset the puzzle to its start state and reset all the data you've added or changed.",
   "startOverWorkspace": "This will reset the workspace to its start state and remove all the blocks you've added or changed.",
+  "startOverWorkspaceText": "This will reset the workspace to its start state and remove all the code you've added or changed.",
```
mikeharv, line 2636: This is maybe a product question, but I wonder if we could just use this string (referencing "code") for all labs, since the blocks are the code.

apps/src/codebridge/Editor/index.tsx
```diff
@@ -1,25 +1,22 @@
...
+
+export const Editor = ({langMapping, editableFileTypes}: EditorProps) => {
   const {project, saveFile} = useCodebridgeContext();
 
   const file = getActiveFileForProject(project);
-  // this is a stupid hack. the low level code-mirror editor won't update itself
-  // automatically if the doc is changed externally. So in lieu of doing it better,
-  // for now we just key the component off of the file ID + an incrementing value that
-  // hits every time the format button is pressed. That'll force a re-render and make it work.
-  // let's swap this out with something better.
```
mikeharv, line 22: I saw this comment recently but couldn't tell what it was referring to exactly. Thanks for cleaning it up!

apps/src/lab2/types.ts
```diff
@@ -260,3 +260,11 @@ export interface ExtraLinksData {
   can_delete: boolean;
   level_name: string;
 }
+
+export const TEXT_BASED_LABS: AppName[] = [
+  'applab',
+  'javalab',
+  'weblab',
+  'pythonlab',
+  'weblab2',
```
mikeharv: Is this an exhaustive list? Are applab and weblab compatible with Lab2? If "not yet", maybe we should also include gamelab? Not sure if there are other worth considering too, like ai lab.

Alternatively, I wonder if it might be more robust to actually check whether the level uses Blockly or not.

apps/src/pythonlab/PythonlabView.tsx
```diff
@@ -70,15 +70,15 @@ const defaultConfig: ConfigType = {
   gridLayoutRows: '32px 232px auto',
   gridLayoutColumns: '300px auto',
   gridLayout: `
-    "instructions file-tabs"
-    "instructions editor"
-    "file-browser editor"
+    "instructions workspace"
+    "instructions workspace"
```
mikeharv, line 74: I think we can clean this up now, by removing one of these duplicate lines. Above on line 70 we can just specify two rows. Instead of giving the rows a fixed height, we could try make it [flexible](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_grid_layout/Basic_concepts_of_grid_layout#the_fr_unit) like
```
gridLayoutRows: '1fr 1fr',
```
which would give each 50% of the available space. 

We could do similar improvements to the weblab2 layouts since they also have a couple of identical rows now.

mikeharv (approved):
Looks good! I think it's worth cleaning up the grid rows before merging, now that we don't need to worry about file tabs separately.

apps/src/lab2/types.ts
molly-moen: No this was my loose list :) I put labs that I knew were text-based, and included those not on lab2 so it wasn't something we had to worry about in the future. But you are probably right that having a half-correct list isn't great. I can update this for now to just be lab2 labs that are text based, and when a lab is migrated when can add it to the list. I'm not sure if we can check consistently if a level uses Blockly, is that a setting on every lab type?

apps/src/pythonlab/PythonlabView.tsx
molly-moen: I might leave this for now since I'm going to work next on moving the console into Code Bridge, and this will get updated again.

apps/src/codebridge/Editor/index.tsx
molly-moen: yeah I was pretty sure this was out of date so I removed it, can you confirm @thomasoniii ?

apps/i18n/common/en_us.json
molly-moen: I kept them separate because Music Lab uses this string and they are in a string lock now.

apps/src/lab2/views/dialogs/StartOverDialog.tsx
sanchitmalhotra126: Seems fine to me 👍 just pulled this locally and confirmed on my end as well.

apps/src/lab2/types.ts
mikeharv: All Blockly levels, including Music, should extend the Blockly level class. I've seen us do `level.is_a? Blockly` in some places. I'm not sure if that's helpful or not. 
I'd vote for either being robust like this (if it works) or having a short list that just has supported lab2 labs if that doesn't work.

apps/i18n/common/en_us.json
mikeharv: Good point!

apps/src/lab2/types.ts
molly-moen: We can't do that on the frontend without sending that as a level property (which I don't think we do now). I'll pare down the list and comment that it's only lab2 labs.

apps/src/codebridge/Editor/index.tsx
thomasoniii: yep, totally just out of date and safe to nuke. good catch!

apps/src/codebridge/Workspace/HeaderButtons.tsx
```diff
@@ -0,0 +1,35 @@
...
+    <div>
+      <Button
+        icon={{iconStyle: 'solid', iconName: 'refresh'}}
+        isIconOnly
+        color={'black'}
+        onClick={onClickStartOver}
+        ariaLabel={'Start Over'}
+        size={'xs'}
+      />
+    </div>
```
thomasoniii, line 31: Does this need to be wrapped in the extra div? Is it just anticipating more buttons being added here?

apps/src/pythonlab/PythonlabView.tsx
thomasoniii: The duplicate line can safely be deleted, and the layoutRows can be simplified as well.

But also, not urgent especially if more layout changes are pending.

apps/src/weblab2/Weblab2View.tsx
```diff
@@ -19,31 +18,27 @@ const weblabLangMapping: {[key: string]: LanguageSupport} = {
...
 
 const verticalLayout = {
   gridLayoutRows: '32px 300px auto auto',
   gridLayoutColumns: '300px auto',
-  gridLayout: `    "instructions file-tabs file-tabs"
-      "instructions editor editor"
-      "file-browser editor editor"
+  gridLayout: `    "instructions workspace workspace"
+      "instructions workspace workspace"
+      "file-browser workspace workspace"
```
thomasoniii, line 34: same as the PythonLab comments above, we can simplify the `gridLayoutRows` and redundant `gridLayout` line.

apps/src/weblab2/Weblab2View.tsx
thomasoniii: Oh. That's a shame. I thought the flexibility would've been nice.

Anyway, sure, if not supported can be safely tossed.

thomasoniii (approved):
minor parroting on the grid layout notes, but otherwise lgtm 🚀

apps/src/pythonlab/PythonlabView.tsx
molly-moen: I'm going to fast follow this with an update to both layouts!

apps/src/codebridge/Workspace/HeaderButtons.tsx
molly-moen: Yeah just anticipating more buttons.

PR #58537, opened 2024-05-10
Adding bubble choice options to multi and external level editors

kakiha11:
This adds the ability for levelbuilders to add bubble choice descriptions and thumbnail URLs to Multi and External/Markdown level types.

This shows the fields added to a multi level editor

Screenshot 2024-05-10 at 1 50 01 PM

This shows a bubble choice level with one multi and one external/markdown level that have descriptions and thumbnail URLs added

Screenshot 2024-05-10 at 1 49 10 PM

## Links

[Slack Thread](https://codedotorg.slack.com/archives/C036XVC3CBF/p1715118727411519)

## PR Checklist:

- [ ] Code is well-commented


davidsbailey:
I am already seeing this option on levelbuilder for multi levels: https://levelbuilder-studio.code.org/levels/60298/edit what am I missing?
Screenshot 2024-05-10 at 4 30 50 PM

Also, when I pull down the PR, and make an edit in this field and press save, I am not seeing anything serialized to the filesystem. if this is working for you, maybe you can provide some details or a video of how you validated so I can try to follow the same steps?

kakiha11:
@davidsbailey Here are screenshots from the dashboard console on a multi level where I updated both the display name and the bubble choice description

Before

Screenshot 2024-05-14 at 1 18 37 PM

After

Screenshot 2024-05-14 at 1 19 25 PM

Is there another way I could verify this update is happening?

davidsbailey:
> Is there another way I could verify this update is happening?

Yes, when you make an edit to a level on levelbuilder, it needs to be updated in the database and in the filesystem. if it is not updated in the filesystem, then the change will not make it off of levelbuilder and into other environments. For example, when I go and make an edit to a different part of a multi level:
Screenshot 2024-05-14 at 2 15 10 PM

I can see my updates to one of the old fields, but not to the new field:
Screenshot 2024-05-14 at 2 15 44 PM

if you are seeing both, then it's possible I have something misconfigured on my end. if not, then it is time to look at the `dsl_text` you are sending from the client to make sure it includes the new data:

https://github.com/code-dot-org/code-dot-org/assets/8001765/d008befc-c5ce-4f75-89ed-b5d7203bdb62

kakiha11:
It turns out DSL-defined levels save to scripts/ and they only save the DSL text, not the .level XML. This means while the old bubble choice form was allowing the db to modified, it wasn't done in a way that could be dumped back out to a level file because there is no level file. There was already support for `display_name` in the DSL model so we added `bubble_choice_description` and `thumbnail_url` to the model to unblock the K-5 curriculum launch. The right long-term fix might be to move away from DSL-defined levels.

Tested that you can now do include multi, match, and external(markdown) in a bubble choice and the changes display correctly, and persist to disk. 

Screenshot 2024-06-25 at 9 51 37 PM


Here is an example of how you can use the new DSL properties in the levelbuilder DSL Editor
```
name 'test_match'
title 'title'
description 'description here'
question 'Question'
answer 'Answer 1'
allow_multiple_attempts false
display_name 'Click this match level!'
bubble_choice_description 'Description set in DSL'
thumbnail_url 'https://images.code.org/ad8ae8b5d46683d8906a2a72ff856d73-FareFinder Logo Design.webp'
```

Here is what a bubble choice level can look like with a multi, match, and external level added

image

snickell (approved):
Looks good, great to see it persisting to disk and not just to DB. Would be nice to have a UI, but I don't think layering more hack UIs over the DSL makes sense (like the teacher only markdown form, which is custom and just edits the DSL in the end).

PR #58539, opened 2024-05-10
Fix bug with flickering pop-up

kobryan0619:
Liam noticed that when the progress page re-loaded, the invitation to see the new progress view would flicker open.

  

https://github.com/code-dot-org/code-dot-org/assets/24883357/116207ca-4184-401c-ab58-6397002b3159

This PR addresses that issue.

## PR Checklist:

- [ ] Code is well-commented


apps/src/templates/sectionProgressV2/InviteToV2ProgressModal.jsx
```diff
@@ -36,6 +36,14 @@ function InviteToV2ProgressModal({
   const [invitationOpen, setInvitationOpen] = React.useState(false);
 
   React.useEffect(() => {
+    if (
+      dateProgressTableInvitationDelayed === undefined ||
+      hasSeenProgressTableInvite === undefined
+    ) {
+      // Do not proceed if data has not been fully loaded.
```
kobryan0619, line 43: Intentionally left this comment in here.

kobryan0619: Note that once this data is fully loaded (for even a brand new user), the values are set to `null` and `false`
image

lfryemason (approved):

apps/test/unit/templates/sectionProgressV2/InviteToV2ProgressModalTest.jsx
```diff
@@ -35,7 +35,10 @@ describe('UnconnectedInviteToV2ProgressModal', () => {
   }
 
   it('renders the dialog with required elements', () => {
-    renderDefault();
+    renderDefault({
+      hasSeenProgressTableInvite: false,
```
lfryemason, line 39: Should we just add these to `DEFAULT_PROPS`?

apps/test/unit/templates/sectionProgressV2/SectionProgressSelectorTest.jsx
```diff
@@ -141,8 +141,6 @@ describe('SectionProgressSelector', () => {
   it('sets user preference when link clicked', () => {
     renderDefault();
 
-    const remindLaterLink = screen.getByText('Remind me later');
```
lfryemason, line 144: Why are we removing this?

lfryemason (approved):
Two non-blocking comments, otherwise looks good to merge

apps/test/unit/templates/sectionProgressV2/SectionProgressSelectorTest.jsx
kobryan0619: In the original tests we added this in because the pop-up got in the way of viewing the screen with the SectionProgressSelector in it.  With the changes to how the pop-up is rendered, it will not render the pop-up until it has data back from Redux about if the pop-up `hasBeeenSeen` or if there is a `dateLastDelayed`.  We could make add these reducers to this test, but given that this test is more about testing the SectionProgressSelector and we already have eyes tests that ensure that this works, I was thinking that wasn't necessary.  Thoughts?

PR #58559, opened 2024-05-13
Lab2: fix progress merge

breville:
Follow-up to https://github.com/code-dot-org/code-dot-org/pull/58269.  

This fixes an issue in the teacher panel when a teacher switches between viewing users with different amounts of progress in a Lab2 level.  The server response for level progress was being merged into the current state, but because we don't reload the whole page, switching to a user with less progress than the last user didn't remove any progress, leading to more green bubbles filled than should be.  

Now, this particular code path uses an optional parameter to skip the progress merge.

### Before

The green bubbles filled for a student's progress remain filled when switching back to the teacher.

https://github.com/code-dot-org/code-dot-org/assets/2205926/bd2d26c0-81a0-417e-bb7b-edfe01eae313

### After

Now, the green bubbles filled for a student's progress are cleared when switching back to the teacher.

https://github.com/code-dot-org/code-dot-org/assets/2205926/d95c82b4-dfbe-4cb4-9cbc-73d5be223dbc

### DIFF
--- a/apps/src/code-studio/components/progress/teacherPanel/TeacherPanel.jsx
+++ b/apps/src/code-studio/components/progress/teacherPanel/TeacherPanel.jsx
@@ -390,7 +390,7 @@ export default connect(
       updateQueryParam('user_id', userId);
       updateQueryParam('version');
       if (isAsync) {
-        dispatch(queryUserProgress(userId));
+        dispatch(queryUserProgress(userId, false));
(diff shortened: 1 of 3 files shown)


thomasoniii (approved):
lgtm 🚀

sanchitmalhotra126 (approved):

etaderhold (approved):

PR #58568, opened 2024-05-13
P20-893: Log CAP user events

artem-vavilov:
## Summary
CAP user events:
- `parent_email_submit`: Logged when a student requests parental permission for the first time.
- `parent_email_update`: Logged when a student makes subsequent requests for parental permission.
- `permission_granting`: Logged when a student's compliance state is changed to "granted".
- `account_locking`: Logged when a student's compliance state is changed to "locked".
- `account_purging`: Logged when a student's account is purged due to non-compliance with the CAP for more than 7 days.

## Links
- JIRA ticket: [P20-893](https://codedotorg.atlassian.net/browse/P20-893)


dashboard/app/models/concerns/serialized_properties.rb
```diff
@@ -29,6 +29,10 @@ def init_internals
     super
   end
 
+  def property_before_save(key)
+    properties_previous_change&.any? ? properties_previous_change.dig(0, key) : properties[key]
+  end
```
wilkie, line 34: This is to support knowing what the prior value was to set `state_before` in the user event and to encapsulate that wonky logic to handle `nil`.

dashboard/test/models/concerns/serialized_properties_test.rb
```diff
@@ -0,0 +1,120 @@
+require 'test_helper'
+
+class SerializedPropertiesTest < ActiveSupport::TestCase
```
wilkie, line 3: We didn't have any tests for `SerializedProperties` so I added some.

wilkie (approved):
I did add some code to this and I really like the modularization it adds. Great work! Since I added some code to this, and own part of it, I would like to see a second approval.

dashboard/lib/policies/child_account.rb
```diff
@@ -79,17 +80,17 @@ def self.user_predates_policy?(user)
     parent_permission_required?(user) && user.created_at < STATE_POLICY[user.us_state][:start_date]
   end
 
+  def self.state_policy(user)
+    return unless user.us_state
```
daynew: You should probably merge with the latest `staging` changes and make sure the country_code check is still there: https://github.com/code-dot-org/code-dot-org/blob/1a66d6262ab0cd5bc91f11045724a8e4757ebfa0/dashboard/lib/policies/child_account.rb#L120

daynew (approved):
Are we planning on backfilling the events by updating the database? Or should we create SQL query to simulate the backfilled events?

artem-vavilov:
> Are we planning on backfilling the events by updating the database? Or should we create SQL query to simulate the backfilled events?

cc @wilkie

wilkie:
We will backfill the database so it is picked up by redshift/trevor/etc for Ali. Drafted PR with the script: https://github.com/code-dot-org/code-dot-org/pull/58790

PR #58573, opened 2024-05-13
Music: Update doc link

cnbrenci:
Change the ? icon in the top right corner of music lab levels to say "Music Lab Documentation" and point to the /music/ doc link instead of Project Beats.

## Links

[[After May 14][Anyone] Music: Change doc link to studio.code.org/docs/ide/music
](https://codedotorg.atlassian.net/browse/LABS-686)

## Testing story

Tested manually 

image

Clicking `Music Lab Documentation` opens up [/docs/ide/music/](https://studio.code.org/docs/ide/music/) in a new tab, which loads the Music Lab Documentation on local deployment but is giving 403 forbidden currently. 

## Followup

I'll do a separate PR to clean up the other doc links so they match the changes to the music lab one.


lib/cdo/help_header.rb
```diff
@@ -59,10 +59,9 @@ def self.get_help_contents(options)
     end
 
     if options[:level] && options[:level].game == Game.music
-      options[:request].path.starts_with?('/projectbeats')
```
cnbrenci, line 62: @breville  Just want to confirm that this is right? Or should it be changed to /music?

dashboard/config/locales/en.yml
```diff
@@ -390,7 +390,7 @@ en:
       sprite_lab_documentation: "Sprite Lab Documentation"
       sprite_lab_tutorials: "Sprite Lab Tutorials"
       web_lab_documentation: "Web Lab Documentation"
-      project_beats_documentation: "Project Beats Documentation"
+      music_lab_documentation: "Music Lab Documentation"
```
breville: nit: can we make the `_documentation` entries alphabetical?

lib/cdo/help_header.rb
breville: I suspect there was a missing `||` because I don't see how this line does anything!  I think it's okay to remove it completely since we'll be redirecting users from `/projectbeats` to a new music project.

lib/cdo/help_header.rb
```diff
@@ -59,10 +59,9 @@ def self.get_help_contents(options)
...
       entries << {
-        title: I18n.t("#{loc_prefix}project_beats_documentation"),
-        url: "https://studio.code.org/docs/ide/projectbeats/",
+        title: I18n.t("#{loc_prefix}music_lab_documentation"),
+        url: "https://studio.code.org/docs/ide/music/",
```
breville: So long as opening in a new tab still works, it might be good to use a relative URL here so that it stays in the same environment.

dashboard/config/locales/en.yml
cnbrenci: yep!

lib/cdo/help_header.rb
cnbrenci: It does open in a new tab.

so that'd be `CDO.code_org_url('/docs/ide/music/')` I believe. I wonder if I should change them all... and why specifically the doc links are all pointing directly to prod.

breville: Since we're already on dashboard, I'm not sure it even needs to be wrapped in `CDO.code_org_url()`?

cnbrenci: You are correct.

lib/cdo/help_header.rb
```diff
@@ -59,10 +59,9 @@ def self.get_help_contents(options)
     end
 
     if options[:level] && options[:level].game == Game.music
-      options[:request].path.starts_with?('/projectbeats')
       entries << {
-        title: I18n.t("#{loc_prefix}project_beats_documentation"),
-        url: "https://studio.code.org/docs/ide/projectbeats/",
+        title: I18n.t("#{loc_prefix}music_lab_documentation"),
+        url: "/docs/ide/music/",
```
breville: nit: it might look nicer in the browser without the final `/`?

breville (approved):

breville:
Should we update [here](https://github.com/code-dot-org/code-dot-org/blob/4d9875ee510beaf6e8e6802d6d79d1211e6d2694/apps/src/music/blockly/constants.js#L51), too?

PR #58583, opened 2024-05-13
Ensures that a user who has already seen the V2 version will not get the pop-up

kobryan0619:
Some of our testers have already been using the V2 dashboard.  This PR will ensure that any user who has already used our new view of the dashboard, will not get the pop-up invitation to use it. 

1. [Jira Ticket](https://codedotorg.atlassian.net/jira/software/projects/TEACH/boards/65?selectedIssue=TEACH-1089)
2. Tests are included :) 

Note: this will have merge conflicts with Liam's change.  Once Liam's change to add the amplitude events and modify when the banner shows up is in staging, I will re-merge staging here.

### DIFF
--- a/apps/src/templates/sectionProgressV2/InviteToV2ProgressModal.jsx
+++ b/apps/src/templates/sectionProgressV2/InviteToV2ProgressModal.jsx
@@ -32,13 +32,15 @@ function InviteToV2ProgressModal({
   setHasSeenProgressTableInvite,
   setShowProgressTableV2,
   setDateProgressTableInvitationDelayed,
+  showProgressTableV2,
 }) {
   const [invitationOpen, setInvitationOpen] = React.useState(false);
 
   React.useEffect(() => {
     if (
       dateProgressTableInvitationDelayed === undefined ||
-      hasSeenProgressTableInvite === undefined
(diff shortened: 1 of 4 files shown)


apps/src/templates/sectionProgressV2/InviteToV2ProgressModal.jsx
```diff
@@ -32,13 +32,15 @@ function InviteToV2ProgressModal({
...
+  showProgressTableV2,
 }) {
   const [invitationOpen, setInvitationOpen] = React.useState(false);
 
   React.useEffect(() => {
     if (
       dateProgressTableInvitationDelayed === undefined ||
-      hasSeenProgressTableInvite === undefined
+      hasSeenProgressTableInvite === undefined ||
+      showProgressTableV2 === undefined
```
lfryemason: Due to how the user properties block serializes properties, `false` will be stripped to `nil`. Therefore this logic won't work.

kobryan0619: Modified to instead use the timestamp.  For this section though, we are checking for `undefined` so that if the state is not done loading it won't "glitch" and show the pop-up unintentionally.

PR #58585, opened 2024-05-13
feat(lms): create account linking landing page for lti providers

stephenliang:
This PR introduces a design library component `Card`. Cards are used in multiple areas at Code.org but there doesn't appear to be a unified component leading to different implementations. This is a first attempt at creating a common component that can be used across the application.

Using the `Card` component, the LMS onboard flow includes a disambiguation screen when users sign on to determine whether the user wishes to create a new account or link to an existing account.

## Links

- [figma](https://www.figma.com/design/88sbry7xnl1wcyrOgbgFtn/Sign-Up-Flow?node-id=2767-1089&t=FtjUUxHtmNV1PCBK-0)
- [jira](https://codedotorg.atlassian.net/browse/P20-916)

## Testing story

image

## PR Checklist:

- [x] Code is well-commented


apps/src/lib/ui/lti/link/LtiLinkAccountPage/types.ts
```diff
@@ -0,0 +1 @@
+export type LtiProvider = 'Canvas' | 'Schoology';
```
carl-codeorg: I think we actually have to map the values from the keys of [LMS_PLATFORMS in the LTI policy module](https://github.com/code-dot-org/code-dot-org/blob/staging/dashboard/lib/policies/lti.rb#L50-L79) into these human readable LMS names. Could do that here or just do it in the controller before we redirect to this page. Might be cleaner to do it in the controller so that we only ever end up with user-friendly strings in the frontend.

Example of the component locally with the value `canvas_cloud` which should map to "Canvas"
Screenshot 2024-05-14 at 6 37 33 PM

dashboard/app/controllers/lti_v1_controller.rb
```diff
@@ -197,7 +197,7 @@ def authenticate
         PartialRegistration.persist_attributes(session, user)
         session[:user_return_to] = destination_url
         if DCDO.get('lti_account_linking_enabled', false)
-          redirect_to lti_v1_account_linking_landing_path and return
+          redirect_to lti_v1_account_linking_landing_path, lti_provider: integration[:platform_name] and return
```
carl-codeorg:
```suggestion
          redirect_to lti_v1_account_linking_landing_path lti_provider: integration[:platform_name] and return
```
Locally this is only working for me without the comma

apps/src/componentLibrary/card/Card/index.tsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+
+import styles from './card.module.scss';
+
+interface CardProps {
+  children: React.ReactNode;
+  'data-testid': string;
+}
+
+export const Card = ({children, ...props}: CardProps) => {
```
carl-codeorg, line 10: Way to go above and beyond and create this Card component! Good thing you took this story; I was going to hack it together with existing components and guess-and-check CSS 😂

apps/src/lib/ui/lti/link/LtiLinkAccountPage/WelcomeBanner/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+  }, [ltiProvider]);
+
+  const ltiProviderIcon = getLtiProviderIcon();
+
+  return (
+    <div className={styles.welcomeContainer}>
+      {ltiProviderIcon && (
+        <div className={styles.welcomeBannerContainer}>
+          <img src={ltiProviderIcon} alt={ltiProvider} />
+          <FontAwesomeV6Icon iconName={'exchange'} />
```
carl-codeorg: Tiny nitpick: this exchange icon is grey in the Figma

carl-codeorg (changes requested):
Couple comments around the `lti_provider` param. Overall looks great; thanks for adding the new component!

apps/src/lib/ui/lti/link/LtiLinkAccountPage/types.ts
stephenliang: Hmm unfortunately I won't be able to pass this up to the TS union type as a dynamic value - not even as a shared constant via the shared constants file. We would probably need to write something that creates a Typescript definition file (similar to the shared constants generator) for this to work correctly. I'm thinking we can potentially roll this into a future API contract discussion where we refactor our APIs to have proper TS definition generation, what do you think?

stephenliang: I dug into this a little more I figured out what you mean, I've updated the PR to publish both the raw type and the friendly name from the controller.

carl-codeorg (approved):
LGTM!

PR #58589, opened 2024-05-13
Music: fix drum preview highlight speeding up

sanchitmalhotra126:
Fixes the behavior of the preview highlight in the play drums block, finally addressing a long standing piece of tech debt in the PatternPanel component. Instead of PatternPanel creating its own timer to tick through the preview animation, we now provide an "onTick" callback from the ToneJSPlayer layer, so that the player can serve as the source of truth for timing. ToneJS's sampler doesn't have any sort of callbacks for when a sample is played unfortunately, but they do have a Clock class which is meant to be highly accurate and understands ToneJS time values (ex. 16th notes). So i opted to use this Clock class to start a timer within the ToneJS player whenever a sequence is previewed to provide accurate tick timing callbacks.

Also makes a slight behavior change where the preview button will change from play to stop when the preview is started, and needs to be explicitly stopped again if the user wants to restart the preview mid-playback. This is just to prevent the play button from being spammed to reduce the likelihood of of any race conditions occurring.

## Links

https://codedotorg.atlassian.net/browse/LABS-789

## Testing story

Tested on standalone projects with the drums and notes blocks.

### DIFF
--- a/apps/src/music/blockly/FieldChord.ts
+++ b/apps/src/music/blockly/FieldChord.ts
@@ -6,7 +6,7 @@ import {ChordEventValue} from '../player/interfaces/ChordEvent';
 import MusicLibrary from '../player/MusicLibrary';
 import {getNoteName} from '../utils/Notes';
 import {generateGraphDataFromChord, ChordGraphNote} from '../utils/Chords';
-import {LoadFinishedCallback} from '../types';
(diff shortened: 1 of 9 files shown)


breville (approved):
Great stuff.  Seems to work nicely.

PR #58592, opened 2024-05-14
Datablock Storage: raise StudentFacingError if called without a table_name

### LINKED ISSUE #58591: ActiveRecord::NotNullViolation in datablock_storage # create_record: Mysql2::Error: Field 'table_name' doesn't have a default value
From Honeybadger: https://app.honeybadger.io/projects/3240/faults/107785410

snickell:
I believe this will fixes [#58591](https://github.com/code-dot-org/code-dot-org/issues/58591), which we got from honeybadger.

### DIFF
--- a/dashboard/app/controllers/datablock_storage_controller.rb
+++ b/dashboard/app/controllers/datablock_storage_controller.rb
@@ -326,7 +326,11 @@ class DatablockStorageController < ApplicationController
   end
 
   private def where_table
-    DatablockStorageTable.where(project_id: @project_id, table_name: params[:table_name])
+    if params[:table_name]
+      DatablockStorageTable.where(project_id: @project_id, table_name: params[:table_name])
+    else
+      raise StudentFacingError, "You must specify a table"
+    end
   end
 
   private def table_or_create


dashboard/app/controllers/datablock_storage_controller.rb
```diff
@@ -326,7 +326,11 @@ def use_firebase_storage
   end
 
   private def where_table
-    DatablockStorageTable.where(project_id: @project_id, table_name: params[:table_name])
+    if params[:table_name]
+      DatablockStorageTable.where(project_id: @project_id, table_name: params[:table_name])
+    else
+      raise StudentFacingError, "You must specify a table"
```
cnbrenci, line 332: We should be able to test this in the same test harness thing that we used to test getColumn to make sure that error is being routed all the way to the student console! I will go find after I finish reviewing the others.

snickell: I mean, we could add a controller test, but I'm pretty 🤷 on that ("tests that don't break aren't good tests"), and we can't do testing that involves crossing the JS <-> backend (we never figured out how to test that).

cnbrenci: Because when called through the block itself, there's always a table_name sent, yeah? and we just aren't sure how this is getting called without one. is that right?

cnbrenci (approved):

dashboard/app/controllers/datablock_storage_controller.rb
snickell: yup, that's IT!

PR #58598, opened 2024-05-14
Enforce MAX_TABLE_ROW_COUNT, MAX_TABLE_COUNT, MAX_RECORD_LENGTH, MAX_NUM_KVPS and MAX_VALUE_LENGTH

snickell:
- Enforces MAX_TABLE_ROW_COUNT, Fixes: [#57002](https://github.com/code-dot-org/code-dot-org/issues/57002)
- Enforces MAX_TABLE_COUNT, Fixes: [#57003](https://github.com/code-dot-org/code-dot-org/issues/57003)
- Enforces MAX_RECORD_LENGTH, Fixes: [#57001](https://github.com/code-dot-org/code-dot-org/issues/57001)
- Enforces MAX_NUM_KVPS, Fixes [#57000](https://github.com/code-dot-org/code-dot-org/issues/57000)
- Enforces MAX_VALUE_LENGTH, Fixes [#56999](https://github.com/code-dot-org/code-dot-org/issues/56999)
- Stops skipping tests for all of the above now that they're working

PR #58601, opened 2024-05-14
DatablockStorageTable.create_records uses `insert_all` rather than one-by-one

### LINKED ISSUE #58596: ActiveRecord::AdapterTimeout in datablock_storage # import_csv
Honeybadger issue: https://app.honeybadger.io/projects/3240/faults/107843187

snickell:
I believe this will speed-up import_csv in particular, and may also help with DB load issues. We currently do all rows in an import_csv in a single transaction, but the transaction involves N_rows number of calls to DatablockStorageRecord.create.

This PR:
1) pre-computes all the IDs and does a single `insert_all()` call (using the [association](https://guides.rubyonrails.org/association_basics.html), so it doesn't need updating!)
2) enforces a max size on CSV data that import_csv will consider trying to import (=max_rows * max_row_size, about 80MB)
3) Locks the DatablockStorageTable row for a table _instead_  of locking the min-id record. When we designed the locking logic, we didn't have a table row that we could lock, this simplifies locking (may make debugging Suresh's graphs easier) and may fix locking edge cases like https://github.com/code-dot-org/code-dot-org/issues/58593 too)

Fixes: [#58596](https://github.com/code-dot-org/code-dot-org/issues/58596)

THIS IS A FOLLOWUP PR TO: https://github.com/code-dot-org/code-dot-org/pull/58598 which should be merged FIRST, sorry if this one still contains the diff of #58598 tomorrow, I changed the base branch, but github isn't updating the files changed diff)


dashboard/app/models/datablock_storage_table.rb
```diff
@@ -159,39 +156,39 @@ def if_shared_table_copy_on_write
...
-      max_record_id = DatablockStorageRecord.where(project_id: project_id, table_name: table_name).maximum(:record_id)
-      next_record_id = (max_record_id || 0) + 1
+      max_record_id = DatablockStorageRecord.where(project_id: project_id, table_name: table_name).maximum(:record_id) || 0
 
-      cols_in_records = Set.new
-      record_jsons.each do |record_json|
-        record_json.each do |_key, json_value|
-          raise StudentFacingError.new(:INVALID_RECORD), 'Invalid record: nested objects and arrays are not permitted' if json_value.is_a?(Hash) || json_value.is_a?(Array)
+      records.insert_all(
+        record_jsons.map.with_index(max_record_id + 1) do |record_json, record_id|
```
cnbrenci, line 180: this is neat

cnbrenci (approved):
Smart to limit the file size! 🚀

PR #58615, opened 2024-05-14
[Sprite Lab] add minimumScale option

mikeharv:
During the piloting of the new Game Design module, some great student projects were collected. One of which features a custom sprite costume that isn't tiling as expected with the new grid blocks:
image (198)

The reason for this is that the image created by the student was 64x64, which is smaller than our expected minimum.
image (199)

When sprites are typically created in Sprite Lab, we automatically scale them down to 100x100, but we don't scale up small images. In the case of the grid block, it's important for the sprites to tile as cleanly as possible. A quick solution here is to add an option for a minimum scale. This makes it so the 64x64 costume image tiles cleanly:
image (200)

When sprites are created in other ways, there is no change to how they are scaled. See the animal sprite and smaller "dirt" sprite below:

image

In considering edges, I created the smallest image that the Piskel editor would allow (2x2). This image can be tiled as well, although I was surprised to learn P5 seems to automatically blur small images when they're scaled up. This also matches the existing behavior in the costume picking block field.

image
image

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/p5lab/spritelab/CoreLibrary.js
+++ b/apps/src/p5lab/spritelab/CoreLibrary.js
@@ -576,6 +576,7 @@ export default class CoreLibrary {
       sprite.initialAngle = opts.initialAngle;
     }
 
+    const minimumScale = opts.minimumScale || 100;
     sprite.baseScale = 1;
     sprite.setScale = function (scale) {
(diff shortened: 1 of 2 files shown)


molly-moen (approved):
nice!

PR #58616, opened 2024-05-14
Delete the i18n poetry.*.md.partial files

bethanyaconnor:
Fixes an issue where code.org/poetry is broken in non-EN languages. I did a little bit of digging and I believe when `poetry.md.partial` was moved to `poetry.haml`, the change didn't take effect in non-EN languages.

Screenshot 2024-05-14 at 9 46 30 AM

I'm definitely interested in figuring out how to avoid this in the future. Some thoughts:
- testing in at least one non-EN language when making major page changes
- maybe a linter rule or something to detect this specific issue (removed the `.md.partial` file but not the translations)

Open to other ideas!

## PR Checklist:

- [ ] Code is well-commented


kelbyhawn (approved):
Thanks for catching this and fixing! 

The next step we can take is to check all of the existing `.md.partial` translated pages and see if there are others that can be deleted, and make a note/ticket for any that have yet to be updated.

bethanyaconnor:
I found a bunch of redirects that still have translations hanging around -- made a ticket: https://codedotorg.atlassian.net/browse/ACQ-1891

bethanyaconnor:
ooo jk found a list of the files to look at: https://github.com/code-dot-org/code-dot-org/blob/27a3004564a84bd11ce0735f21d13454cd78810e/bin/i18n/resources/pegasus/markdown/sync_in.rb#L13

@code-dot-org/platform -- for this specific PR, do I need to remove this file from that list above?

mgc1194 (approved):
Thanks! 
We don't have a system that automatically deletes translations when the source is removed, so we'll have to make sure translations are removed at the same time as partials are migrated to haml. 
We prefer not loosing translations in case source files are accidentally deleted.

mgc1194:
> ooo jk found a list of the files to look at:
> 
> https://github.com/code-dot-org/code-dot-org/blob/27a3004564a84bd11ce0735f21d13454cd78810e/bin/i18n/resources/pegasus/markdown/sync_in.rb#L13
> 
> @code-dot-org/platform -- for this specific PR, do I need to remove this file from that list above?

Yes, you should edit this list with the markdowns we actually support.
I think if a markdown is not found, the sync throws a warning but it does not fail.

PR #58622, opened 2024-05-14
Add more folder to import/order lint rule

lfryemason:

## PR Checklist:

- [ ] Code is well-commented


apps/src/weblab2/Weblab2View.tsx
```diff
@@ -1,18 +1,18 @@
-import React, {useState} from 'react';
-
-import './styles/Weblab2View.css';
-
-import {Config} from './Config';
+import './styles/Weblab2View.css'; // eslint-disable-line import/order
```
Hamms, line 3: can you also add a comment explaining why this is necessary, and which specifically of the other imports in the file we need this to before?

Hamms (approved):
I'd like to see one explanatory comment added, otherwise this all LGTM. Thank you!

PR #58626, opened 2024-05-14
fixed race condition

kobryan0619:
Fixed race condition causing the modal to still pop-up on reload of the "accept" action on the pop-up modal

## Links
[
Link to Slack thread](https://codedotorg.slack.com/archives/C06ERUABG01/p1715721046521149)

## Testing story
Modified tests


apps/src/templates/sectionProgressV2/InviteToV2ProgressModal.jsx
```diff
@@ -89,10 +89,10 @@ function InviteToV2ProgressModal({
     analyticsReporter.sendEvent(EVENTS.PROGRESS_V2_ACCEPT_INVITATION, {
       sectionId,
     });
-    setHasSeenProgressTableInviteData();
+    setHasSeenProgressTableInviteData(true);
     setHasSeenProgressTableInvite(true);
-    onShowProgressTableV2Change();
     setShowProgressTableV2(true);
+    onShowProgressTableV2Change();
```
lfryemason: We need to remove this, otherwise we will still have race condition. This function is calling `setShowProgressTableV2`.

dashboard/app/controllers/api/v1/users_controller.rb
```diff
@@ -215,7 +215,15 @@ def post_has_seen_progress_table_v2_invitation
     return head :unauthorized unless current_user
 
     current_user.has_seen_progress_table_v2_invitation = !!params[:has_seen_progress_table_v2_invitation].try(:to_bool)
-    current_user.save
+
```
lfryemason, line 218: Could we add some BE unit tests?

lfryemason: This can be a fast follow if we manual test and want to push through for release.

apps/src/templates/sectionProgressV2/InviteToV2ProgressModal.jsx
```diff
@@ -89,16 +90,11 @@ function InviteToV2ProgressModal({
...
     setHasSeenProgressTableInvite(true);
-    onShowProgressTableV2Change();
     setShowProgressTableV2(true);
-  }, [
-    sectionId,
-    setHasSeenProgressTableInvite,
-    setShowProgressTableV2,
-    onShowProgressTableV2Change,
-  ]);
+    new UserPreferences().setShowProgressTableV2(true);
```
lfryemason: Why did we add this line? Won't this make a redundant post call?

lfryemason (approved):

apps/src/templates/sectionProgressV2/InviteToV2ProgressModal.jsx
kobryan0619: Sorry - I just updated the PR.  I added it because I thought we needed it after removing the `onShowProgressTableV2Change` but we don't.  It still feels like something is missing, but I as far as I can tell we copied over the needed behavior from `onShowProgressTableV2Change`.

lfryemason (approved):

lfryemason (approved):

PR #58627, opened 2024-05-14
Updating client side finish sign up event, moving experiment call

hannahbergam:
This PR covers two updates and may be easier to review one commit at a time!

First: 

The current finish sign up event fires each time the button is pressed, which means we may get multiple events for the same user if they haven't selected at minimum: user type, display name (and for teachers): email opt in preference. 

These additions to the event add has_school (by checking for the existence of an nces id in both the old flow and the new flow), and display name and email opt in preference (by checking if either radio button is filled in). 

I was able to test this on the old flow in all sorts of permutations (see the list below and one expanded example) and on the new flow by setting up the new experiment name and checking that the school designation still works (photo after). I was also able to confirm that the student flow still works. 

Second:

This PR moves the experiment allocation point. Now, we only call the experiment once we have a user type = teacher. This required me to change the order of the render, so I'd love if someone could check out this branch and move through the flow locally to double check my own local testing. Happy to pair on this- it will require editing one line in the StatsigAnalytics file and adding a manual override for your session ID to the experiment.

Screenshot 2024-05-14 at 2 17 12 PM
Screenshot 2024-05-14 at 2 17 27 PM
Screenshot 2024-05-14 at 3 54 23 PM

## Links

- spec: []()
- jira ticket: https://codedotorg.atlassian.net/browse/ACQ-1889
- https://codedotorg.atlassian.net/browse/ACQ-1890

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/sites/studio/pages/devise/registrations/_finish_sign_up.js
+++ b/apps/src/sites/studio/pages/devise/registrations/_finish_sign_up.js
@@ -53,15 +53,10 @@ let userInRegionalPartnerVariant = experiments.isEnabled(
   experiments.OPT_IN_EMAIL_REG_PARTNER
 );
 
-const isInSchoolAssociationExperiment = statsigReporter.getIsInExperiment(
-  'school_association_update_2024',
-  'showNewFlow',
-  false
-);
-
 $(document).ready(() => {
   const schoolInfoMountPoint = document.getElementById('school-info-inputs');
   let user_type = $('#user_user_type').val();
+  let isInSchoolAssociationExperiment = false;
   init();
 
   function init() {
@@ -75,7 +70,6 @@ $(document).ready(() => {
         'width:135px;';
     }
     setUserType(user_type);
-    renderSchoolInfo();
     renderParentSignUpSection();
   }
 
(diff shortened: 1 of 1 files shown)


apps/src/sites/studio/pages/devise/registrations/_finish_sign_up.js
```diff
@@ -221,6 +215,12 @@ $(document).ready(() => {
     fadeInFields(TEACHER_ONLY_FIELDS);
     hideFields(STUDENT_ONLY_FIELDS);
     toggleVisShareEmailRegPartner(isInUnitedStates);
+    isInSchoolAssociationExperiment = statsigReporter.getIsInExperiment(
```
bethanyaconnor, line 218: could this be done in `renderSchoolInfo` to avoid having to declare it above?

bethanyaconnor: ooo I see ... `renderSchoolInfo` is called multiple times in the old flow. 

I'm so excited for this code to be cleaned up!

bethanyaconnor (approved):

PR #58628, opened 2024-05-14
Control When Project Tables are Truncated For Controller Unit Tests

sureshc:
The logic to `TRUNCATE` dashboard projects tables before running controller unit tests was executing any time `common_test_helper` was `require`d. Explicitly `TRUNCATE` before each controller unit test so that these tables are not modified until we explicitly do so. We no longer need to reset the auto-increment counter because `TRUNCATE` does that. 

We still need Rails to rollback transactions after executing each test in case they write to non-project tables.

This change potentially unblocks #57507 which in turn is needed to support upgrading to MySQL 8.

## Follow-up work

Move the VCR configuration and the project table truncate to a `setup` method within the module and change all controller unit tests to explicitly invoke that setup method.

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/shared/test/common_test_helper.rb
+++ b/shared/test/common_test_helper.rb
@@ -101,12 +101,5 @@ module SetupTest
     # so reset them to ensure that they are not reused across tests.
     BucketHelper.s3 = nil if defined?(BucketHelper)
     AWS::S3.s3 = nil
-
-    # Reset AUTO_INCREMENT, since it is unaffected by transaction rollback.
-    DASHBOARD_TEST_TABLES.each do |table|
-      # rubocop:disable CustomCops/DashboardDbUsage
-      DASHBOARD_DB.execute("ALTER TABLE `#{table}` AUTO_INCREMENT = 1")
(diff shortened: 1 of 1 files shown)


shared/test/common_test_helper.rb
```diff
@@ -17,40 +17,41 @@
...
+  DASHBOARD_TEST_TABLES = %w(channel_tokens user_project_storage_ids projects project_commits code_review_comments code_reviews).freeze
 
-# Truncate database tables to ensure repeatable tests.
-DASHBOARD_TEST_TABLES = %w(channel_tokens user_project_storage_ids projects project_commits code_review_comments code_reviews).freeze
-DASHBOARD_TEST_TABLES.each do |table|
-  # rubocop:disable CustomCops/DashboardDbUsage
-  DASHBOARD_DB[table.to_sym].truncate
-  # rubocop:enable CustomCops/DashboardDbUsage
-end.freeze
+  VCR.configure do |c|
```
sureshc: @Hamms suggests moving this back outside the module declaration.

Hamms (approved):

PR #58630, opened 2024-05-15
Fix Lab2 bubble choice UI test

sanchitmalhotra126:
Seems like there are some issues with loading Music Lab in UI tests on Safari/iPad, so we'll just use a different Lab2 level instead (level type isn't important, just needs to be a lab2 level)

https://codedotorg.slack.com/archives/C0T0PNTM3/p1715724643985969

## Testing story

Tested locally with Sauce Labs


sanchitmalhotra126:
Merging ahead of Drone to try and fix test build

PR #58633, opened 2024-05-15
Only validate the username if it is changing

daynew:
We received a Zendesk ticket where a User reported an issue where they couldn’t update the state they lived in. Upon investigation, we found that the User couldn’t update their state because their username was invalid (it had a white space in it “big husky”). The User model runs most of its validations before every save we don’t want it to get into a bad state. 

We queried Redshift and found that there are about 6,000 users with a whitespace in their name.

As a workaround, we are only going to validate the username whenever it changes.

## Links
* [JIRA](https://codedotorg.atlassian.net/browse/P20-933)

## Testing story
* Unit tests

## Follow-up work
* [Long term fix](https://codedotorg.atlassian.net/browse/P20-932)


carl-codeorg (approved):
:shipit:

PR #58649, opened 2024-05-15
Add Music Lab Amplitude analytics on session end - count of samples played

fisher-alice:
This PR adds Amplitude analytics that are reported at the end of a session when the user has played samples in a music lab [standalone project](https://github.com/code-dot-org/code-dot-org/blob/4ff077e255571bd39479f7671486a98e655b8692/apps/src/music/views/MusicView.jsx#L153) (not on music lab levels as requested by product). A DCDO flag is also added since this PR is close to launch.

The `SessionEndPayload` now includes `soundsPlayed` which records the number of times a sample is actually played during the user session. Note that there was [discussion](https://codedotorg.slack.com/archives/C04TH8400AU/p1715639719809219) about whether to count the number of scheduled plays vs the number of actual plays, and we opted for the actual number. The count also includes samples played from a trigger and a sample played [while in preview mode](https://github.com/code-dot-org/code-dot-org/blob/c3f04ecf635231538e81c85dc6100d69baac6d69/apps/src/music/player/MusicPlayer.ts#L187).

Update on May 16: Videos are slightly out of date. A sound sample's `optionalSoundPath` is now included in the sound id.
For example, 'electro/bass/bass_synth_candy'. 

Here's a screencast video with a log statement added when `onSoundsPlayed` is executed [here](https://github.com/code-dot-org/code-dot-org/blob/815cfe92fc38905b6cc497fdd4bb971e3ff8337f/apps/src/music/analytics/AnalyticsReporter.ts#L182) (logging is since removed). The program includes two 'play sound' blocks that will be executed consecutively.

The following scenarios are demo'd:
1. When 'Run' button is clicked,  sample count is updated and displayed when each of the two samples begin to play  because `onPlaySound` is scheduled to be executed at the same time that the sample is actually played.
2. When 'Run' is clicked, and then 'Stop' is clicked before second sample is scheduled to play, `soundsPlayed` is NOT updated with the second scheduled sample.
3. When 'Run' is clicked and then the trigger key is clicked, the sample count is updated at the beginning of the sample scheduled by the trigger. This happens twice in the video.
4. When the user ends the session by clicking on 'My Dashboard', there is [a breakpoint set](https://github.com/code-dot-org/code-dot-org/blob/815cfe92fc38905b6cc497fdd4bb971e3ff8337f/apps/src/music/analytics/AnalyticsReporter.ts#L253) in `endSession` where the payload is logged and `soundsPlayed` is included.

https://github.com/code-dot-org/code-dot-org/assets/107423305/339d1172-107b-4f68-8b32-97dab8385c44

Here's a screenshot of the data logged in Amplitude:
Screenshot 2024-05-15 at 9 38 27 PM

Here's a screencast video of samples being played in preview and the logging of the sample counts in the dev console.

https://github.com/code-dot-org/code-dot-org/assets/107423305/4811393f-2997-4373-b71a-673d9fd4766a

## Links

[jira](https://codedotorg.atlassian.net/browse/LABS-733)

## Testing story
Local testing is described in PR descrption above.
For the DCDO flag, I set the flag to `false` via the dashboard-console using `DCDO.set('music-lab-samples-report', false)` and saw the the `onSoundsPlayed` was not executed. Then I set to flag to `{}` and since the default is `true`, `onSoundsPlayed` was executed. And checked when flag is set to `true` and function was executed as expected.

## Follow-up work
Currently statsig analytics is not set up in `AnalyticsReporter.ts` as it is in `AnalyticsReporter.js`. Should we add?
Update: Checked with @moneppo and he said not at this time. [Slack discussion](https://codedotorg.slack.com/archives/C04TH8400AU/p1715873875325729?thread_ts=1715639719.809219&cid=C04TH8400AU)

## PR Checklist:

- [ ] Code is well-commented


apps/src/music/analytics/AnalyticsReporter.ts
```diff
@@ -165,6 +168,19 @@ export default class AnalyticsReporter {
...
+    if (!this.sessionInProgress) {
+      this.log('No session in progress');
+      return;
+    }
+
+    if (!this.soundsPlayed[id]) {
+      this.soundsPlayed[id] = 1;
+    } else {
+      this.soundsPlayed[id]++;
+    }
```
thomasoniii: not blocking, and I'm not convinced this is any better, but as an alternative:

```
this.soundsPlayed[id] = 1 + (this.soundsPlayed[id] ?? 0)
```

clearly I've just got nullish coalescing operators on my mind this morning. 😄

apps/src/music/player/MusicLibrary.ts
```diff
@@ -188,6 +188,13 @@ export default class MusicLibrary {
     return `${baseUrl}${this.libraryJson.path}/${folder.path}/${optionalSoundPath}${soundData.src}.mp3`;
   }
 
+  getSoundIdFromUrl(soundUrl: string): string {
+    const soundUrlParts = soundUrl.split('/');
+    const source = soundUrlParts[soundUrlParts.length - 1].slice(0, -4);
+    const folder = soundUrlParts[soundUrlParts.length - 2];
```
thomasoniii: Dangit. I wanted to suggest using `Array.prototype.at` here, since it can take negative offsets, but that was only added to Safari in 15.4, and we're currently supporting down to 13 (14 this summer). Chrome has had it since 92 and Firefox since 90.

It _would have been_ simplified to:

```
const source = soundUrlParts.at(-1).slice(0, -4);
const folder = soundUrlParts.at(-2);
```

So just noting it as something that could be tweaked when our min reqs catch up.

thomasoniii (approved):
lgtm 🚀

apps/src/music/player/MusicLibrary.ts
fisher-alice: Good to know - thanks!

apps/src/music/analytics/AnalyticsReporter.ts
fisher-alice: Sure - I can update! I was first introduced to nullish coalescing operators by Cassi. https://github.com/code-dot-org/code-dot-org/pull/58086/files#r1574847284
Thanks!

apps/src/music/player/ToneJSPlayer.ts
```diff
@@ -259,6 +260,14 @@ class ToneJSPlayer implements AudioPlayer {
       .start(this.playbackTimeToTransportTime(sample.playbackPosition));
 
     this.activePlayers.push(player);
+    const library = MusicLibrary.getInstance();
+    if (!library) {
+      return null;
```
sanchitmalhotra126: nit: just `return` here since this method returns void

apps/src/music/player/types.ts
```diff
@@ -52,7 +52,7 @@ export interface AudioPlayer {
   cancelPreviews(): void;
 
   /** Schedule a sample to played */
-  scheduleSample(sample: SampleEvent): void;
+  scheduleSample(sample: SampleEvent, callback: (id: string) => void): void;
```
sanchitmalhotra126: Could we call this something like `onSamplePlayed` or `onSampleStart`?

apps/src/music/player/MusicLibrary.ts
```diff
@@ -188,6 +188,13 @@ export default class MusicLibrary {
     return `${baseUrl}${this.libraryJson.path}/${folder.path}/${optionalSoundPath}${soundData.src}.mp3`;
   }
 
+  getSoundIdFromUrl(soundUrl: string): string {
```
sanchitmalhotra126: Want to clarify this logic with @breville - I was thinking we could compute the ID from the URL by removing the baseUrl and the first part before the first slash (which should be the folder path), because some sound IDs may have an `"optionalSoundPath"` as well. Does that sound right?

apps/src/music/views/MusicView.jsx
```diff
@@ -639,6 +642,7 @@ class UnconnectedMusicView extends React.Component {
     this.player.stopSong();
     this.playingTriggers = [];
 
+    // Clear the timeline of triggered events when song is stopped.
```
sanchitmalhotra126, line 644: Thanks for adding!

apps/src/music/player/MusicPlayer.ts
```diff
@@ -180,7 +184,7 @@ export default class MusicPlayer {
       blockId: 'preview',
       soundType: 'beat',
     };
-
+    this.analyticsReporter?.onSoundsPlayed(id);
```
sanchitmalhotra126: The result of `this.convertEventSamples(preview)[0]` that's computed below should return a sample that has a URL, in the same format as the sample that's passed to `scheduleSample` in ToneJSPlayer. So you should be able to use the same function to convert the URL back to an ID. It's a little roundabout, but will ensure that we're using the same ID format consistently for reporting.

fisher-alice: Thanks! will update.

apps/src/music/player/MusicLibrary.ts
fisher-alice: I updated this function so that the `optionalSoundPath` is included. Lmk what you think. Thanks!

breville: This is a little fragile because the presence of `packs/` comes only from the library JSON, and in future a different library could opt to not use that directory at all.  

It's also not guaranteed to always use the same unique identifier that we use for sounds in code, which is always just `[folder.id]/[sound.src]` - the optional sound path is something that is inserted when we generate the sample URL - though this should be less of an issue.

But since this code needs to take a peek inside the system for gathering analytics, I think the approach is alright for now.  I would just add a comment that we are making an assumption that the current llibrary uses `packs/` and would probably add a guard for cases when that string isn't found.

sanchitmalhotra126: Agreed. @breville is it a safe assumption that the first directory after the baseUrl is always the folder path (`packs` or otherwise)? If so, I was thinking something like:

```
getSoundIdFromUrl(soundUrl: string): string {
  const baseUrl = soundUrl.includes(baseAssetUrlRestricted) ? baseAssetUrlRestricted : getBaseAssetUrl();
  
  // Remove the URL, library path, and first slash
  const path = soundUrl.substring(baseUrl.length + this.libraryJson.path.length + 1);
  
  // Remove folder by finding first slash, and remove '.mp3' extension
  const firstSlashIndex = path.indexOf('/');
  return path.substring(firstSlashIndex + 1, path.length - 4);
}
```

apps/src/dcdo.js
```diff
@@ -16,8 +16,8 @@ export class DCDO {
    * due to the caching behavior of the page. Analyze the HTTP headers of the pages you are
    * interested in to understand what kind of caching they use and if that will be a concern.
    * @param {string} key The key for the DCDO config to lookup.
-   * @param {object} defaultValue The value to return if the given key is not defined in DCDO.
-   * @return {object} The value for the given key. Uses the defaultValue if no key found.
+   * @param {object | boolean} defaultValue The value to return if the given key is not defined in DCDO.
+   * @return {object | boolean} The value for the given key. Uses the defaultValue if no key found.
```
sanchitmalhotra126, line 20: Thanks for updating this!

apps/src/music/player/MusicLibrary.ts
sanchitmalhotra126: All this being said though, looking at the code I'm realizing that it's probably just simpler and easier to add an `id` field to `SampleEvent`, since we have access to the original ID before we convert to a SampleEvent...that way, both the ToneJSPlayer and MusicPlayer just need to pass the ID in and don't need to rely on the library (and all the `null` checks) to convert the ID back to a URL. 😅 sorry for all the runaround here!

breville: That sounds great!  Otherwise it would be a case of trying to reverse the behaviour of [`generateSoundUrl`](https://github.com/code-dot-org/code-dot-org/blob/348a49b4628006be03dc83cfcf2b0162ee11f583/apps/src/music/player/MusicLibrary.ts#L182-L189), which has two possible base URLs (for restricted versus non-restricted sounds) and the optional sound path to contend with.

sanchitmalhotra126:
Made a few updates to try get this merged today
- Add ID to SampleEvent interface and use that for reporting, rather than reverse-engineering ID from URL
- Remove unused function from MusicLibrary

breville (approved):

apps/src/music/player/MusicLibrary.ts
fisher-alice: Thanks to you both! And thanks @sanchitmalhotra126 so much for your help and making latest updates when I was dealing with power outage issues.

PR #58653, opened 2024-05-15
P20-558: Fix joins to sections

artem-vavilov:
## Issue
I couldn’t reproduce the bug locally, but based on the available information from production,
it seems that the `current_user` sections are already preloaded on prod.
Therefore, we need to refresh the preloaded collection of sections whenever a new section is added.

https://github.com/code-dot-org/code-dot-org/blob/4ff077e255571bd39479f7671486a98e655b8692/apps/src/templates/studioHomepages/JoinSection.jsx#L60-L68

## Links
- JIRA ticket: [P20-558](https://codedotorg.atlassian.net/browse/P20-558)

### DIFF
--- a/dashboard/app/controllers/api/v1/sections_controller.rb
+++ b/dashboard/app/controllers/api/v1/sections_controller.rb
@@ -175,7 +175,7 @@ class Api::V1::SectionsController < Api::V1::JSONApiController
       return
     end
     render json: {
-      sections: current_user.sections_as_student.map(&:summarize_without_students),
+      sections: current_user.sections_as_student.reload.map(&:summarize_without_students),
       studentSections: current_user.sections_as_student_participant.map(&:summarize_without_students),
       plSections: current_user.sections_as_pl_participant.map(&:summarize_without_students),
       result: result


dashboard/app/controllers/api/v1/sections_controller.rb
```diff
@@ -175,7 +175,7 @@ def join
       return
     end
     render json: {
-      sections: current_user.sections_as_student.map(&:summarize_without_students),
+      sections: current_user.sections_as_student.reload.map(&:summarize_without_students),
       studentSections: current_user.sections_as_student_participant.map(&:summarize_without_students),
       plSections: current_user.sections_as_pl_participant.map(&:summarize_without_students),
```
daynew, line 180: should we also reload `studentSections` and `plSections`?

dashboard/app/controllers/api/v1/sections_controller.rb
```diff
@@ -175,7 +175,7 @@ def join
       return
     end
     render json: {
-      sections: current_user.sections_as_student.map(&:summarize_without_students),
+      sections: current_user.sections_as_student.reload.map(&:summarize_without_students),
```
artem-vavilov, line 178: Reload `current_user.sections_as_student` after a new section has been added
https://github.com/code-dot-org/code-dot-org/blob/3d20e67d0eac412842fb553f0af79cb6a9f1aeab/dashboard/app/controllers/api/v1/sections_controller.rb#L147

daynew (approved):
This sounds like a reasonable thing to try. I have just one question I'd like you to address before merging.

dashboard/app/controllers/api/v1/sections_controller.rb
artem-vavilov: Both these methods simply filter the `sections_as_student` collection data that will be already reloaded:
https://github.com/code-dot-org/code-dot-org/blob/4ff077e255571bd39479f7671486a98e655b8692/dashboard/app/models/user.rb#L2014-L2020

PR #58654, opened 2024-05-15
Panels: remove deprecated level data when saving level

sanchitmalhotra126:
A quick fix to automatically clean up unused data in Panels levels whenever they're saved.

Example diff after saving level:
Screenshot 2024-05-15 at 1 42 26 PM

## Testing story

Tested locally with Music Lab panels levels.


breville (approved):
Thanks!

PR #58655, opened 2024-05-15
Music: use sounds panel revision

breville:
With this change, the revised sounds panel introduced in https://github.com/code-dot-org/code-dot-org/pull/58275 becomes the default.

Two UI issues have been fixed:
- When the current sound was the default, and the sounds panel was opened, and another sound was clicked, we instantly scrolled to that new sound.
- When the current sound was not the default, and the sounds panel was opened, and the default sound was clicked, we did an eased scroll to the default sound.

The `ScrollIntoView` component has been renamed `EaseIntoView` to alleviate confusion with the browser's existing `scrollIntoView` function.

A DCDO flag of `"music-lab-use-sounds-panel-1"` can be set to `true` or the `sounds-panel-2=false` URL parameter can continue to be used to switch back to the legacy sounds panel.

### DIFF
--- a/apps/src/music/views/SoundsPanel2.tsx
+++ b/apps/src/music/views/SoundsPanel2.tsx
@@ -253,8 +253,9 @@ const SoundsPanel: React.FunctionComponent<SoundsPanelProps> = ({
     useRef(null);
 
   // Capture whether the initially-selected sound is the default sound, in which
-  // case we will do a smooth scroll into view, rather than scroll directly
-  // to the current sound.
+  // case we will do a smooth ease into view, rather than scroll directly
+  // to the current sound.  Uses useRef so that changes to this value don't trigger
+  // additional executions of the "Initial render" useEffect, below.
   const isDefaultSoundSelected = useRef(
     currentValue === library.getDefaultSound()
   );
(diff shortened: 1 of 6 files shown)


thomasoniii (approved):
lgtm 🚀

apps/src/music/views/SoundsPanel2.tsx
```diff
@@ -252,7 +252,12 @@ const SoundsPanel: React.FunctionComponent<SoundsPanelProps> = ({
   const currentSoundRef: React.MutableRefObject<HTMLDivElement | null> =
     useRef(null);
 
-  const isDefaultSoundSelected = currentValue === library.getDefaultSound();
+  // Capture whether the initially-selected sound is the default sound, in which
+  // case we will do a smooth scroll into view, rather than scroll directly
+  // to the current sound.
+  const isDefaultSoundSelected = useRef(
```
sanchitmalhotra126, line 259: To check my understanding: we're using `useRef` so that the below `useEffect` only runs on initial render, and not whenever `currentValue` changes, right? If so, mind updating the comment to capture that?

sanchitmalhotra126 (approved):

apps/src/music/views/SoundsPanel2.tsx
breville: Yes, that's right.  Comment updated.

PR #58666, opened 2024-05-15
Lab2: fix viewing student work on a sublevel

sanchitmalhotra126:
Fix for viewing student work as a teacher on a lab2 sublevel. This was broken because the get_or_create_for_level API needed a script level ID to check if the teacher could view the student's work, which is not present on a sublevel. The fix here is just to explicitly pass through the script level ID of the current or parent level so that we can correctly check if the teacher can view the student's work.

Demo (disregard the duplicate levels 8 and 9, that's a local bug with my DB seeding).

https://github.com/code-dot-org/code-dot-org/assets/85528507/80faa9d5-c639-434e-a673-c818aa39fecb

## Links

https://codedotorg.atlassian.net/browse/LABS-803

## Testing story

Tested on Lab2 allthethings progression, music intro script, and Gen AI pilot script. Also tested with individual levels and standalone projects to make sure no behavior regressed.


apps/src/lab2/projects/ProjectContainer.tsx
```diff
@@ -25,26 +26,24 @@ const ProjectContainer: React.FunctionComponent<ProjectContainerProps> = ({
   channelId,
   appName,
 }) => {
-  const currentLevelId = useSelector(
-    (state: {progress: ProgressState}) => state.progress.currentLevelId
-  );
-  const userId = useSelector(
-    (state: {progress: ProgressState}) =>
-      state.progress.viewAsUserId || undefined
+  const currentLevelId = useAppSelector(state => state.progress.currentLevelId);
```
breville, line 29: Thanks for updating all of these to use `useAppSelector`.

dashboard/test/controllers/projects_controller_test.rb
```diff
@@ -474,6 +474,28 @@ def sign_in_with_request(user)
     refute_nil @response.body['channel']
   end
 
+  test 'get_or_create_for_level with uses script level ID if provided' do
```
breville: Typo?

dashboard/app/controllers/projects_controller.rb
```diff
@@ -328,7 +328,9 @@ def get_or_create_for_level
 
     if user_id
       # If viewing another user's work, ensure that we have permission.
-      script_level = level.script_levels.find_by_script_id(script_id)
+      script_level = params[:script_level_id] ?
+        ScriptLevel.cache_find(params[:script_level_id].to_i) :
+        level.script_levels.find_by_script_id(script_id)
```
breville: Is there a potential malicious case in which the request provides a `script_level_id` which  is used here to check for permission, but which is unrelated to the `level_id` which is used below to actually retrieve the channel?

sanchitmalhotra126: Oh true, that's a good point...yeah it would make sense to have a stricter check here. Ideally we should be able to use the script ID in combination with sublevel ID to find the script level ID. I'll look into that.

sanchitmalhotra126: Updated this with a check in `projects_controller` which ensures that if a script level ID is provided, it either matches the provided level ID or is associated with a valid parent level of provided sublevel ID. Added a unit test for this too. Let me know if you think this feels strict enough!

breville: I don't know these subsystems very well, but that seems like a good improvement, with good test coverage too, thanks!

breville (approved):
I'm not too familiar with these systems, but glad the extra validation is in place.  The approach and the testing looks solid.  Thanks for tackling this complex system!

molly-moen (approved):

kobryan0619 (approved):
As we say in the midwest, "uffda".   This all looks good though.  I focused on the items related to scriptId and leveId.  I am wondering if we should have UI/Eyes test coverage on this at all?

sanchitmalhotra126:
> As we say in the midwest, "uffda". This all looks good though. I focused on the items related to scriptId and leveId. I am wondering if we should have UI/Eyes test coverage on this at all?

Thanks! Good call on UI tests - I'll add a check to the existing lab2 bubble choice UI tests.

PR #58667, opened 2024-05-16
Applab: fix screen import ID conflict between screen and element

bencodeorg:
Applab has a feature that allows you to import "screens" from other projects:

| image | image|
| - | - |

When using this feature, we have code that:
- handles a conflict between two projects with screens with the same IDs, and
- handles conflicts between two projects with individual elements (eg, buttons) with the same IDs.

A recent curriculum update ([more detail here](https://codedotorg.slack.com/archives/C05DMS18MEX/p1716247158785079?thread_ts=1716230450.073589&cid=C05DMS18MEX)) exposed the fact that we don't handle two other sets of potential conflicts:

Conflict between... | Current project | Imported screen | Result |
| - | - | - | - |
| | screen ID | element ID | error, dialog hangs |
| | element ID | screen ID | Import "succeeds", silently wipes element with name conflict from existing project |

The change here is to treat these conflicts the same way that we treat conflicts between individual elements (eg, a button in the existing project matches a button in the screen to import). I include a slight rewording of the error message to support these cases.

**Updated experience**
Conflicts between screen IDs and element IDs appear in list of IDs preventing import. In this case, the project trying to do the import has a button with the ID `screen1`, and the screen to import has the ID `screen1`

image

## Links

- jira: https://codedotorg.atlassian.net/browse/LABS-691
- [slack discussion](https://codedotorg.slack.com/archives/C05DMS18MEX/p1716247158785079?thread_ts=1716230450.073589&cid=C05DMS18MEX)

## Testing story

I tested manually:
- Each of the "cross-type" collision types (ie, screen ID-element ID collisions) blocks import gracefully per the screenshot above.
- Screen ID-screen ID collision still offers the user the opportunity to overwrite their existing screen.
- Element ID-element ID collision continues to blocks import of a screen.
- Screen imports without any of these collisions succeed.

### DIFF
--- a/apps/src/applab/ImportScreensDialog.jsx
+++ b/apps/src/applab/ImportScreensDialog.jsx
@@ -25,6 +25,13 @@ const SCALE = 0.1;
 const MARGIN = 10;
 const ICON_HEIGHT = applabConstants.APP_HEIGHT * SCALE;
 
+export const IMPORT_FAILURE_MESSAGE = `
+  Cannot import the following screens because their IDs or
+  contained design element IDs are already used in your existing
+  project. Fix the IDs in either project so they aren't
+  duplicated between the two projects before trying to import the following.
+`;
+
 // TODO: possibly refactor AssetRow to make it work here instead of
 // or with this component
 class AssetListItemUnwrapped extends React.Component {
@@ -206,13 +213,7 @@ export class ImportScreensDialog extends React.Component {
             {nonImportableScreens.length > 0 && (
               <div style={styles.section}>
                 <h2 style={multiCheckboxStyles.header}>Cannot Import</h2>
-                <p style={styles.subtext}>
-                  Cannot import the following screens because their IDs or
-                  contained design element IDs are already used in your existing
-                  project. Fix the IDs in either project so they aren't
-                  duplicated across different screens before trying to import
-                  the following.
-                </p>
+                <p style={styles.subtext}>{IMPORT_FAILURE_MESSAGE}</p>
                 <ul style={multiCheckboxStyles.list}>
                   {nonImportableScreens.map(screen => (
                     <li key={screen.id} style={multiCheckboxStyles.listItem}>
--- a/apps/src/applab/ImportScreensDialog.jsx
+++ b/apps/src/applab/ImportScreensDialog.jsx
@@ -101,7 +101,7 @@ class ScreenListItemUnwrapped extends React.Component {
             )}
           {screen.conflictingIds.length > 0 && (
             <p style={styles.warning}>
-              Uses existing element IDs:{' '}
+              Uses existing element or screen IDs:{' '}
               {quotedCommaJoin(screen.conflictingIds)}.
             </p>
           )}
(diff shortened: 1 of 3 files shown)


breville (approved):
Excellent write-up and clarity around the entire test matrix.

apps/src/applab/import.js
```diff
@@ -38,11 +38,24 @@ function getImportableScreen(dom) {
   const id = dom.id;
   const willReplace = designMode.getAllScreenIds().includes(id);
   const conflictingIds = [];
+
+  // Catch edge case where the to be imported screen ID
```
fisher-alice: Nit - thought sentence had a typo at first:
'where the ID of the imported screen conflicts with an existing element ID'.

apps/src/applab/ImportScreensDialog.jsx
```diff
@@ -207,8 +207,8 @@ export class ImportScreensDialog extends React.Component {
               <div style={styles.section}>
                 <h2 style={multiCheckboxStyles.header}>Cannot Import</h2>
                 <p style={styles.subtext}>
-                  Cannot import the following screens because they contain
-                  design elements with IDs already used in your existing
+                  Cannot import the following screens because their IDs or
+                  contained design element IDs are already used in your existing
                   project. Fix the IDs in either project so they aren't
                   duplicated across different screens before trying to import
```
fisher-alice: This is not your change, but wondering if the wording could be updated.
'Fix the IDs in either project so that they aren't duplicated between the two projects before trying to import.'
We already ensure that element/screen IDs are not duplicated across screens within the same project.
@dancodedotorg - lmk what you think. Thanks!

apps/src/applab/import.js
```diff
@@ -38,11 +38,24 @@ function getImportableScreen(dom) {
...
+  // conflicts with an existing element ID.
+  if (!willReplace && !elementUtils.isIdAvailable(id)) {
+    conflictingIds.push(id);
+  }
+
   Array.from(dom.children).forEach(child => {
     if (!elementUtils.isIdAvailable(child.id)) {
       var existingElement = elementUtils.getPrefixedElementById(child.id);
```
fisher-alice: Nit: use `const`.

fisher-alice (approved):
Nice and +1 on well-written description! Just left a couple nits. Thanks!

apps/test/unit/applab/ImportScreensDialogTest.js
```diff
@@ -351,11 +351,11 @@ describe('ImportScreensDialog', () => {
                 <div>
                   <h2>Cannot Import</h2>
                   <p>
-                    Cannot import the following screens because they contain
-                    design elements with IDs already used in your existing
-                    project. Fix the IDs in either project so they aren't
-                    duplicated across different screens before trying to import
-                    the following.
+                    Cannot import the following screens because their IDs or
```
sanchitmalhotra126: nit: constant for this string so it can be reused in this test?

sanchitmalhotra126 (approved):
Nice sleuthing!

apps/src/applab/ImportScreensDialog.jsx
bencodeorg: Took me a sec to follow so just to clarify, I think the specific proposed diff is:

"Fix the IDs in either project so they aren't duplicated ~~across different screens~~ **between the two projects** before trying to import"

This change makes sense to me though, updated! I think we can make an executive decision here unless Dan objects.

dancodedotorg: I trust y'all - don't wait up on me

PR #58668, opened 2024-05-16
Gen AI: more robust chat logging

bencodeorg:
In service of simplifying our chat logging, a couple of related changes.

Break up chat history in UI into three pieces:
- `chatMessagesPast`: past messages that are from a previous "session" and are only now needed from a UI perspective (ie, to show students what they did before). Added to when `chatMessagesCurrent` is wiped (see below).
- `chatMessagesCurrent`: current set of messages that we're sending to the model representing existing conversation history. Append to this when we get a response from the model, and wipe and move existing messages to `chatMessagesPast` when a new session is created (eg, by a change in system prompt).
- `chatMessagePending`: the message a user has input and is awaiting response from the server. Wiped once we receive a response from the server.

This change allows us to standardize the response from Rails in all cases to be a list of messages that should be appended to the chat history in the UI.

image

Supporting changes:
- Update message removal logic to support removing messages from either of the lists of chat messages.
- New selector that combines all chat messages for UI and copying messages.
- Update to controller such that we only call `log_chat_session` and return a list of messages in a single place.
- New user message sent to model is an object rather than a string (and has a key `chatMessageText` rather than `content` for consistency).

## Links

- slides with a bit of graphical description: [link](https://docs.google.com/presentation/d/1qyL-ySGhlAK9ADi83U3cOx0CSj9us_W0WQw1vPrtREg/edit#slide=id.g27154527de7_0_33)

## Testing story

I updated Rails-side unit tests to match the new behavior. We have no end-to-end testing at this point, so I manually tested:
- Normal request/response with the model. Maintains the same session ID across multiple messages.
- Profanity from student is hidden.
- Profanity generated by model generates an error message.
- Model update and error notifications are dismissable. Model updates result in a new session.
- Copying and clearing chat history succeeds.


sanchitmalhotra126 (commented):
Great refactor! I know this is still WIP so some high-level thoughts on the design
- Love splitting up the chat history into clear categories. Really helps with the mental model of what's going and cleans up the code quite a bit
- Likewise, I like that the chat API has a strict contract of returning a list of messages. Makes it easier and more predictable to parse on the front end
- This will also allow us to have tighter controls on what fields can be updated via redux, vs what fields should only be updated internally in thunks. Clearer division of responsibilities between internal redux code and other UI code.
- Something I've been noodling on relatedly is potentially reorganizing the types of the content in the chat message array(s). Currently, since we started with only messages, we're kind of shoehorning other types of UI (like model updates, and later on, error messages) into the chat message format, when they are distinct UI types. This can (probably should) be it's own separate refactor, but generally I was thinking about having a top level "ChatUpdate"/"ChatItem" interface, from which different sub-types extend (messages, model updates, errors, etc). Ex:
```
interface ChatItem {
  type: 'message' | 'modelUpdate' | 'error' // (would it be helpful to even separate user and assistant message types here?)
  ...
}

interface ChatMessage extends ChatItem {
  type: 'message' // force all ChatMessages to have type 'message'
  text: string;
  role: Role // this can now just be 'User' & 'Assistant'
  ... // do we even still need an ID?
}

interface ModelUpdate extends ChatItem {
  type: 'modelUpdate';
  updatedField: keyof AiCustomizations;
  updatedValue: AiCustomizations[keyof AiCustomizations];
  timestamp: string;
  ...
}

interface Error extends ChatItem {
  type: 'error';
  errorMessage: string;
  ...
}
```

With your proposed change, we could use these types to make restrictions on what content can go where. For example, I'd imagine that `chatMessagesPast` could be an array of `ChatItem[]` but `chatMessagesCurrent` can only be a list of actual chat messages, so `ChatMessage[]`. And `chatMessagePending` can only be a single `ChatMessage`.

apps/src/aichat/aichatCompletionApi.ts
```diff
@@ -47,19 +47,18 @@ export async function postAichatCompletionMessage(
...
+    chatMessageText: message.chatMessageText,
+    status: message.status,
+  };
 };
 
 type AichatCompletionMessage = {
   role: Role;
-  content: string;
+  chatMessageText: string;
+  status: string;
```
fisher-alice: Would it be helpful to type as `AichatInteractionStatusValue`. 
I know there's issues with inferring types from shared constants.

apps/src/aichat/redux/aichatRedux.ts
```diff
@@ -47,10 +43,20 @@ import {
...
+  messageListKey: MessageListKey;
+};
+
 export interface AichatState {
-  // All user and assistant chat messages - includes too personal and inappropriate user messages.
-  // Messages will be logged and stored.
-  chatMessages: ChatCompletionMessage[];
+  // Messages from previous chat sessions that we track purely for visibility to the user
+  // and do not send to the model as history.
+  chatMessagesPast: ChatCompletionMessage[];
```
fisher-alice, line 57: Really appreciate storing past and current separately!

apps/src/aichat/redux/aichatRedux.ts
```diff
@@ -251,6 +259,22 @@ export const onSaveFail = () => (dispatch: AppDispatch) => {
   dispatch(endSave());
 };
 
+type AichatCompletionMessage = {
```
fisher-alice: I see this type defined here and in `aichatCompletionApi.ts`. Maybe extract and move to `types.ts`?

apps/src/aichat/redux/aichatRedux.ts
```diff
@@ -251,6 +259,22 @@ export const onSaveFail = () => (dispatch: AppDispatch) => {
...
+  role: Role;
+  chatMessageText: string;
+  status: string;
+};
+
+const decorateMessage = (
```
fisher-alice: I see this does the inverse of `formatMessageForAichatCompletion`. Would it be accurate to say that `CompletionMessage` is a subtype of `AichatCompletionMessagwe`?

apps/src/aichat/redux/aichatRedux.ts
```diff
@@ -562,6 +517,32 @@ const aichatSlice = createSlice({
   },
 });
 
+const findModelUpdateMessage = (
```
fisher-alice: This name threw me off a little - maybe `getModeUpdateMessageLocation`?

dashboard/app/controllers/aichat_controller.rb
```diff
@@ -14,49 +16,52 @@ def chat_completion
...
-      )
+
+      return {
+        messages: messages,
+        flagged_content: filter_result.content,
+      }
+    end
+
+    messages_for_model = params.to_unsafe_h[:storedMessages].filter do |message|
+      message[:status] == SharedConstants::AI_INTERACTION_STATUS[:OK] &&
```
fisher-alice, line 40: I think we already filter for messages in `conversationMessages` on the front-end for user and assistant responses. Is this for redundancy?

dashboard/app/controllers/aichat_controller.rb
```diff
@@ -14,49 +16,52 @@ def chat_completion
...
-      new_messages = [
+      messages = [
+        get_user_message(SharedConstants::AI_INTERACTION_STATUS[:ERROR]),
         {
-          role: 'user',
-          content: params[:newMessage],
-          status: SharedConstants::AI_INTERACTION_STATUS[:ERROR]
+          role: "assistant",
+          status: SharedConstants::AI_INTERACTION_STATUS[:ERROR],
+          chatMessageText: '[redacted - model generated profanity]',
```
fisher-alice, line 62: Nit: Wonder if we should defined 'redacted' for users.

fisher-alice (approved):
Really nice refactor! Left some minor questions but overall, I think this made the code more simple to understand.

apps/src/aichat/redux/aichatRedux.ts
fisher-alice: We could use 'Omit' as we did 'AichatModelCustomizations' type.

apps/src/aichat/redux/aichatRedux.ts
bencodeorg: Good call, it's also potentially finding error notification messages since I first wrote this :). Will update!

dashboard/app/controllers/aichat_controller.rb
bencodeorg: Sorry, not sure I'm following...what do you mean "define"?

dashboard/app/controllers/aichat_controller.rb
bencodeorg: haha yeah, good catch -- it felt safer to me to be redundant in the absence of any testing of our front end API currently, ok with you?

dashboard/app/controllers/aichat_controller.rb
bencodeorg: Also maybe worth noting that this message doesn't actually get displayed to users -- we just display a generic "error getting response" message.

apps/src/aichat/redux/aichatRedux.ts
bencodeorg: Ended up using `Pick`, which I think is just the opposite of Omit -- lmk what you think!

bencodeorg: Also renamed the two "inversion" functions and moved 'em to utils to make clearer they're related! Naming a little tricky but did my best :)

dashboard/app/controllers/aichat_controller.rb
fisher-alice: Sure - just confirming.

dashboard/app/controllers/aichat_controller.rb
fisher-alice: Oh - gotcha. Sounds good!

apps/src/aichat/types.ts
```diff
@@ -12,6 +12,11 @@ export type ChatCompletionMessage = {
   timestamp?: string;
 };
 
+export type AichatCompletionMessage = Pick<
```
fisher-alice, line 15: Cool! TIL about `Pick`.

apps/src/aichat/redux/utils.ts
```diff
@@ -78,3 +80,28 @@ export const allFieldsHidden = (fieldVisibilities: FieldVisibilities) =>
   getTypedKeys(fieldVisibilities).every(
     key => fieldVisibilities[key] === Visibility.HIDDEN
   );
+
+// Opposite of decorateMessageFromModelResponse.
+// Preps messages to send to model for response.
+export const trimMessageForModelInput = (
```
fisher-alice: I can see what you're saying about the tricky naming of this function. 

Nit: 'trimMessage' seems to imply that the message text/content is trimmed. What about `prepMessageForModelInput`? I don't feel strongly either way - just another option.

bencodeorg:
One thing I forgot to mention in the description is that I updated the key representing the content of messages from `content` to `chatMessageText` so they match on the front end and back end -- we may want to do update to match Ruby conventions and do something like what AI Tutor does here to snake case their params:

https://github.com/code-dot-org/code-dot-org/blob/099da69970f686c8896bc58da754b73ce142caec/dashboard/app/controllers/ai_tutor_interactions_controller.rb#L19-L26

PR #58672, opened 2024-05-16
Music: fix zero key

breville:
Fixes an issue in which switching to a pack with a `key` of `0` was not applying that key.

### DIFF
--- a/apps/src/music/player/MusicLibrary.ts
+++ b/apps/src/music/player/MusicLibrary.ts
@@ -244,9 +244,9 @@ export default class MusicLibrary {
     }
     const folder = this.getFolderForFolderId(this.currentPackId);
     // Read key from the folder, or the first sound that has a key if not present on the folder.
-    return folder?.key !== undefined
-      ? folder?.key
-      : folder?.sounds.find(sound => sound.key !== undefined)?.key;
+    return (
+      folder?.key ?? folder?.sounds.find(sound => sound.key !== undefined)?.key
+    );
   }
 }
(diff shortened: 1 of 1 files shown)


apps/src/music/player/MusicLibrary.ts
```diff
@@ -244,9 +244,9 @@ export default class MusicLibrary {
     }
     const folder = this.getFolderForFolderId(this.currentPackId);
     // Read key from the folder, or the first sound that has a key if not present on the folder.
-    return (
-      folder?.key || folder?.sounds.find(sound => sound.key !== undefined)?.key
-    );
+    return folder?.key !== undefined
+      ? folder?.key
+      : folder?.sounds.find(sound => sound.key !== undefined)?.key;
```
thomasoniii: nullish coalescing operator for the win!

```
folder?.key ?? folder?.sounds.find(sound => sound.key !== undefined)?.key
```

Lets us keep the simpler logic w/o adding the ternary.

thomasoniii (changes requested):
Made a note on the PR - use the nullish coalescing operator instead of a ternary.

apps/src/music/player/MusicLibrary.ts
breville: Nice, thanks!  Done.

sanchitmalhotra126 (approved):
Thanks for catching this!

thomasoniii (approved):
lgtm 🚀

PR #58684, opened 2024-05-16
Add help and tips to Code Bridge

molly-moen:
This PR enables editing and showing help and tips in Python Lab. The ability to see help and tips is enabled in Web Lab 2 as well, but for now I'm keeping level editing changes scoped to Python Lab.

I made a rough approximation of the [figma](https://www.figma.com/design/vcNGOFsIks0HhRk6GEOfZf/Upper-Lab?node-id=2974-20725&t=U77y0HmBE4Az1f6I-0) version of help and tips. It is a dropdown on what was previously only the instructions panel which will only show up if the level has help and tips. I've renamed the instructions component "Info Panel" since it now encompasses more than just instructions.

We did not have a component like the one in the figma (an icon-only dropdown button), so I patched one together using the design system `Button` components and some copied css from the design system dropdown components. Since this is still an approximate UI I didn't spend time perfecting it.

### Figma
Screenshot 2024-05-16 at 9 28 37 AM

### Current UX
https://github.com/code-dot-org/code-dot-org/assets/33666587/a11df9a8-428f-4b73-975a-2daae94cb025

The hover color is light gray because that is the default in the design system.

## Links

- jira ticket: [CT-461](https://codedotorg.atlassian.net/browse/CT-461)

## Testing story
Tested locally. 

## Follow-up work
If this is the final design, there is some polish work to do on the dropdown.

## PR Checklist:

- [ ] Code is well-commented


apps/src/lab2/views/components/editor/CodeEditor.tsx
```diff
@@ -84,7 +84,7 @@ const CodeEditor: React.FunctionComponent<CodeEditorProps> = ({
 
   return (
     <div id="code-editor">
-      <div ref={editorRef} className={classNames('codemirror-container')} />;
+      <div ref={editorRef} className={classNames('codemirror-container')} />
```
molly-moen, line 87: this just removes a rogue `;` that was showing up under the editor

dashboard/config/levels/custom/pythonlab/Allthethings Python Lab 1.level
```diff
@@ -27,8 +28,12 @@
         }
       }
     },
-    "long_instructions": "## Welcome to Python Lab\r\n\r\nPress the Run button and see what happens."
+    "long_instructions": "## Welcome to Python Lab\r\n\r\nPress the Run button and see what happens.",
+    "reference_links": [
+      "/courses/csa-2023/guides/instantiating-objects"
+    ],
+    "map_reference": "/docs/concepts/html"
```
mikeharv: Nit: After reseeding this level, the map link wasn't working for me. Adding a slash fixes that.

```suggestion
    "map_reference": "/docs/concepts/html/"
```

apps/src/codebridge/InfoPanel/index.tsx
```diff
@@ -0,0 +1,99 @@
...
+      id="codebridge-info-panel"
+      headerContent={currentPanel}
+      rightHeaderContent={renderHeaderButton()}
+      className={moduleStyles.infoPanel}
+    >
+      {isDropdownOpen && (
+        <form className={moduleStyles.dropdownContainer}>
+          <ul>
+            {panelOptions().map(panel => {
+              return (
```
mikeharv: Nit: If this map function just returns the list item, I think we can actually remove the function's outer curly brackets and explicit `return` keyword.

apps/src/codebridge/InfoPanel/index.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+  const panelOptions = () => {
+    // For now, always include Instructions panel.
+    // TODO: support hiding this panel completely if there are no instructions.
+    const panelOptions = [Panels.Instructions];
+    if (mapReference || referenceLinks) {
+      panelOptions.push(Panels.HelpAndTips);
+    }
+    return panelOptions;
+  };
```
mikeharv: Does this need to be a function or could `panelOptions` just be defined as an array? Is there a need to recalculate the options each time `panelOptions()` is called or would it be sufficient to do it once?

Going a step further, would it be worth memoizing `panelOptions` so that it only re-calculates when there are changes to `mapReference` or `referenceLinks`?

apps/src/codebridge/InfoPanel/index.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+  const showCurrentPanel = () => {
+    switch (currentPanel) {
+      case Panels.Instructions:
+        return <InstructionsView />;
+      case Panels.HelpAndTips:
+        return <HelpAndTips />;
+    }
+  };
```
mikeharv: In your Start Over PR, you used `useCallback` for functions like `resetToStartSource` and `onClickStartOver`. Could we do the same thing here to avoid re-creating this function unless the dependencies change? 

I have the same question for `changePanel` and `renderHeaderButton` in this file.

apps/src/codebridge/InfoPanel/styles/info-panel.module.scss
```diff
@@ -0,0 +1,74 @@
...
+    padding: 0;
+
+    li {
+      list-style-type: none;
+      display: flex;
+      align-items: flex-end;
+      align-self: stretch;
+
+      .dropdownItem {
+        /* Reset's every elements apperance */
```
mikeharv: nit: "element's"

mikeharv (commented):
Looks good, a couple questions. Pulled, reseeded, and tested and seems to be working great!

apps/src/codebridge/InfoPanel/styles/info-panel.module.scss
thomasoniii: also nit: "Resets". 🤣

apps/src/codebridge/InfoPanel/index.tsx
thomasoniii: I'm pretty sure this can be a statically wired dispatch lookup instead.

Move it outside the component and name as:
```
const getCurrentPanel = {
  [Panels.Instructions] : InstructionsView,
  [Panels.HelpAndTips] : HelpAndTips
}
```

Then call it as:
```
{ getCurrentPanel[currentPanel]() }
```
(annoyingly, react will not render `<CurrentPanel[currentPanel] />` (AFAIK), so we need the janky syntax)

Or, to make it prettier, but less compact:
```
const CurrentPanel = getCurrentPanel[currentPanel]
...
<CurrentPanel/>
```

Or, if you still want it inline:
```
const CurrentPanel = useMemo(() => {
  const panelLookup = {
    [Panels.Instructions] : InstructionsView,
    [Panels.HelpAndTips] : HelpAndTips
  }
  return panelLookup[currentPanel]
}, [currentPanel]);
...
<CurrentPanel/>
```

apps/src/codebridge/InfoPanel/index.tsx
molly-moen: I get errors when I do that. I believe the reason is this is inline code rather than a function, and any code inside the html needs to be surrounded by curly braces.

apps/src/codebridge/InfoPanel/index.tsx
molly-moen: The reason to keep it as a function is we could switch levels and go from having to not having help and tips, and want to update accordingly.
It's a pretty cheap operation, so I don't think we need to memoize. The [react docs](https://react.dev/reference/react/useMemo#skipping-expensive-recalculations) say you don't need to memoize for cheap operations.

apps/src/codebridge/InfoPanel/index.tsx
molly-moen: I'll update this to be a static lookup.
To the question on `useCallback`, the reason I added it around the start over logic was because I was seeing some issues with unexpected resets of the project code due to us firing a callback too often. However, I probably was overly cautious and didn't clean up ones that didn't need to be wrapped in useCallback.

apps/src/codebridge/InfoPanel/index.tsx
mikeharv: What errors are you getting? I was curious and tried modifying locally and it seems to work. You should be able to remove the curly braces and `return` - and the return statement's `;`, which I missed initially!
I believe you only need the curly braces and return if the arrow function includes other steps.

apps/src/codebridge/InfoPanel/index.tsx
mikeharv: Sounds good to me! These hooks are still pretty new to me, so I've been trying to test my understanding of how (and why) to use them. Thanks for explaining!

mikeharv (approved):
Curious about my question of removing the `return`, but definitely not a blocker!

apps/src/codebridge/InfoPanel/index.tsx
molly-moen: oh I misunderstood what you were saying. I thought you meant remove the curly braces around `panelOptions().map...`. Updated to remove them from the inner function.

apps/src/codebridge/InfoPanel/index.tsx
molly-moen: I ended up changing `panelOptions` to be a state value instead, because I found a bug where if you switch levels and were previously on a panel that does not exist on the new level it will still show as your selected panel, but be blank. So I needed to add a `useEffect` that depended on the panel options changing.

breville:
I'm curious about some broader architectural questions.  I see that we have [added](https://github.com/code-dot-org/code-dot-org/pull/58684/files#diff-2ef686a0ce8571958a051ecc8bc4331a70e0fe625d4cadd4e11a94ccef5df243R148-R150) some support to Lab2 for this, but the component itself lives in CodeBridge.  Do we consider this work a Lab2 feature?

PR #58688, opened 2024-05-16
P20-903: Display U13 Warning on Section Creation to USA teachers

artem-vavilov:
## Error
```
==============[0m[1000D[KERROR["test_returns_nil_if_student_us_state_is_\"??\"", "UserTest::UsStateCodeTest", 56.4121803099988]
 test_returns_nil_if_student_us_state_is_"??"#UserTest::UsStateCodeTest (56.41s)
Minitest::UnexpectedError:         Errno::EACCES: Permission denied @ rb_sysopen - /usr/local/lib/ruby/gems/3.0.0/gems/city-state-1.1.0/lib/db/states.us
            app/models/user.rb:2746:in `us_state_code'
            test/models/user_test.rb:18:in `block in <class:UsStateCodeTest>'
```
## Links
- JIRA ticket: [P20-903](https://codedotorg.atlassian.net/browse/P20-903)
- Related PR: https://github.com/code-dot-org/code-dot-org/pull/58497
- Revert PR: https://github.com/code-dot-org/code-dot-org/pull/58670

### DIFF
--- a/dashboard/test/models/user_test.rb
+++ b/dashboard/test/models/user_test.rb
@@ -24,7 +24,7 @@ class UserTest < ActiveSupport::TestCase
     end
 
     test 'returns teacher school US state code when state is name' do
-      teacher = create(:teacher, school_info: create(:school_info, state: 'district of Columbia'))
+      teacher = create(:teacher, school_info: create(:school_info, state: 'washington dc'))
       assert_equal 'DC', teacher.us_state_code
     end
   end
--- a/dashboard/test/models/user_test.rb
+++ b/dashboard/test/models/user_test.rb
@@ -13,19 +13,29 @@ class UserTest < ActiveSupport::TestCase
       assert_equal 'CO', student.us_state_code
     end
 
-    test 'returns nil if student us_state is "??"' do
+    test 'returns nil if student us_state is unknown' do
       student = create(:student, :unknown_us_region)
       assert_nil student.us_state_code
     end
 
     test 'returns teacher school US state code' do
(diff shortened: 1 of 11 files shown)


dashboard/test/models/user_test.rb
```diff
@@ -7,6 +7,28 @@ class UserTest < ActiveSupport::TestCase
...
+
+    test 'returns teacher school US state code' do
+      teacher = create(:teacher, school_info: create(:school_info, state: 'ny'))
+      assert_equal 'NY', teacher.us_state_code
+    end
+
+    test 'returns teacher school US state code when state is name' do
+      teacher = create(:teacher, school_info: create(:school_info, state: 'washington dc'))
+      assert_equal 'DC', teacher.us_state_code
+    end
```
wilkie, line 29: Maybe we should cover the case where the `school_info` does not have a state recorded or has an unknown state ... maybe it is a school in Canada, Europe, etc. I think you handle that case in your implementation.

wilkie: Basically, cover the case where the teacher's state would be `nil` I think.

artem-vavilov: It stores only U.S. states

artem-vavilov: https://github.com/code-dot-org/code-dot-org/pull/58688/commits/e187410d8cb2af0c28f764763d420bca925c9955

wilkie (approved):
Great work! 👍

PR #58707, opened 2024-05-17
Check for new value in us_state validation

juanmanzojr:
This PR also includes all the changes made within this [PR](https://github.com/code-dot-org/code-dot-org/pull/58697)
(Validate us_state on create and update, make sure country_code is also recorded in state modal)
,which was reverted as I could not get a fix forward in time for DTP. 

Checks to make sure that a new value for `us_state` is coming in before finishing the `us_state` validation.
We currently have some users who have invalid states. We do not want to prevent any other updates from raising an error because of the changes we made to when the us_state validation is run.

## Links

- jira ticket: [here](https://codedotorg.atlassian.net/browse/P20-939)

## Testing story
New Test added to check us_state validation does not disrupt the updates of users with invalid states.


dashboard/app/models/user.rb
```diff
@@ -2877,6 +2877,8 @@ def self.us_state_dropdown_options
       errors.add(:us_state, :blank)
```
daynew, line 2873: I think this is going to result in all the users who managed to create a new account without setting a US state being considered invalid. Whenever they go to update something on their user model, like their display name, validation will fail because US State is blank. You can verify this by creating a unit test where `country_code: 'US'` and `us_state: nil` then call `user.validate!`. We should try to have unit tests for all "states" our users are in. Let's pair to brainstorm a list of all of them and then write the tests together.

dashboard/app/models/user.rb
```diff
@@ -2877,6 +2877,8 @@ def self.us_state_dropdown_options
...
       return
     end
+    # Check if us_state value will change, check after making sure us_state is not blank
+    return unless will_save_change_to_properties? && properties_change&.first&.[]("us_state") != us_state
```
daynew: I'm guessing `us_state_changed?` doesn't work because it is a serialized attribute?
The `.first` is concerning to me because I'm wondering what is `.second` or `.last`. I'm not familiar with `properties_change`.

daynew (changes requested):

dashboard/app/models/user.rb
juanmanzojr: Correct, it was a bit tricky because of `us_state` being part of serialized attributes.
I used the documentation provided[ here](https://api.rubyonrails.org/classes/ActiveModel/Dirty.html#method-i-2A_change) for` _change` which Returns the old and the new value of the attribute.

daynew (approved):
Thank you for creating all these unit tests!

PR #58708, opened 2024-05-17
Add new string, update unit5 self paced PL link, update AFE career tour

dmcavoy:
* Added string "Middle and High School" which can be translated and then used for the homepage (Did I put this in the right spot @kelbyhawn?)
* Update link on self paced pl page for CSD Unit 5 to not be year based. Tested that going to the non-year based worked. 
* Removed Keeping Data Safe Career Tour and replaced with Music Career Tour

Screenshot 2024-05-16 at 8 47 01 PM


pegasus/cache/i18n/en-US.yml
```diff
@@ -375,6 +375,7 @@
   homepage_slot_text_blurb_educators: "Teach your students"
   homepage_slot_text_link_elementary: "Elementary school"
   homepage_slot_text_link_middle: "Middle school"
+  homepage_slot_text_link_middle_and_high: "Middle and High School"
```
kelbyhawn: This needs to go in the [i18n gsheet](https://docs.google.com/spreadsheets/d/1Tq7VqZALgRA0wYk0HDfEOTyRI0TM2Dir2rloXIPGCgU/edit#gid=0&range=342:343).

kelbyhawn (commented):
Looks great, thanks for doing this! I commented in the code, but the new string needs to go in the [i18n ghseet](https://docs.google.com/spreadsheets/d/1Tq7VqZALgRA0wYk0HDfEOTyRI0TM2Dir2rloXIPGCgU/edit#gid=0&range=342:343).

kelbyhawn (approved):
Thanks for moving that string over to the gsheet! (I remember this PR now 😅)

PR #58712, opened 2024-05-17
CPA regression mitigations

daynew:
This PR implements two bug fixes:
* Microsoft, Facebook, and Email accounts created before this fix is deployed should not be age gated until July 1st, 2024.
* Don't Age Gate users based on the "interpolated" `us_state` attribute.

This PR also adds a unit test to verify that we don't age gate the user if they didn't set a `us_state` when creating their account.

## Links

* [Recategorize Email, Microsoft, and Facebook as P2](https://codedotorg.atlassian.net/browse/P20-937)
* [CPA should not trust the interpolated state value](https://codedotorg.atlassian.net/browse/P20-938)
* [Creating new accounts should require US State to be defined.](https://codedotorg.atlassian.net/browse/P20-939)

## Testing story
* Unit tests

## PR Checklist:

- [ ] Code is well-commented


wilkie (approved):
LGTM

dashboard/lib/policies/child_account.rb
```diff
@@ -76,10 +73,24 @@ def self.show_cap_state_modal?(user)
...
+
+    user.created_at < policy_start_date ||
+      user.authentication_options.any?(&:google?) ||
+      # [P20-937] CPA students created after the policy came into effect but
+      # before '2024/05/18' might have had their us_state field not managed
+      # properly, so we will allow them to be treated the same as users who
+      # predate CPA starting in 2023/07/1.
+      (
+        user.created_at < Date.parse('2024-05-18T00:00:00Z') &&
+        user.authentication_options.any? do |ao|
```
carl-codeorg: Are we getting much value from checking these remaining auth option types? We're already handling Google on line 81, and LTI, Clever, and word/picture are all exempted elsewhere, right? So by including email, Microsoft, and Facebook, we're pretty much covering everything. Could we just do `user.created_at < Date.parse('2024-05-18T00:00:00Z')` to keep it simple?

dashboard/lib/policies/child_account.rb
```diff
@@ -89,6 +100,9 @@ def self.lockout_date(user)
   end
 
   private_class_method def self.state_policy(user)
+    # If the country_code is not set, then us_state value was inherited
+    # from the teacher and we don't trust it.
+    return unless user.country_code
```
carl-codeorg, line 99: What's the rationale for putting this here instead of in `self.user_predates_policy?` It seems like if we contain all these changes to a single function, it's easier to remove/disable them later.

daynew: I put it here because this is the only place that `user.us_state` is read in this file. And there is already logic in other parts of the code to handle the case where state_policy is `nil`.

dashboard/lib/policies/child_account.rb
daynew: The rational was that I was trying to closely match the fix with the bug report, but you are right, this could be simplified to just checking the created date.

wilkie: oh interesting. yeah I suppose there is no test for a provider that would trigger this to be something different than Facebook, etc... because there isn't any situation?

dashboard/lib/policies/child_account.rb
carl-codeorg: Makes sense - I'm good with it staying here as long as we have a plan to remove it once we hit phase 2

dashboard/lib/policies/child_account.rb
```diff
@@ -78,19 +83,8 @@ def self.user_predates_policy?(user)
...
-      # predate CPA starting in 2023/07/1.
-      (
-        user.created_at < Date.parse('2024-05-18T00:00:00Z') &&
-        user.authentication_options.any? do |ao|
-          [AuthenticationOption::EMAIL,
-           AuthenticationOption::MICROSOFT,
-           AuthenticationOption::FACEBOOK].include?(ao.credential_type)
-        end
-      )
+      user.created_at < CPA_CREATED_AT_EXCEPTION_DATE ||
```
carl-codeorg, line 86: Nice, definitely a lot cleaner!

carl-codeorg (approved):

PR #58716, opened 2024-05-17
[Sprite Lab] use new bitmap field config options

mikeharv:
A new version of the bitmap field plugin has been published which includes new customization options:
* https://github.com/google/blockly-samples/pull/2351

These options allow for customization that was previously only possible by extending some fairly major methods in the class. Now we can simply indicate that we want a fixed height and to hide the "Randomize" button when we create in the field instance.

**Before and after** _No differences_

image

## PR Checklist:

- [ ] Code is well-commented


molly-moen (approved):
yay for contributing back to the plugin!

PR #58718, opened 2024-05-17
[Lab2] update styles for instructions code

mikeharv:
Now that an upper lab sample progression has been shared, I decided to quickly check how well the instruction panel component would render the type of text found. 

This helped me realize that we're not correctly styling `code` text in the instructions: 
image

I compared the styles with a published lab (Java Lab): 
image

Assuming we'd like consistency here, this PR provides the needed color and background color. For the background color, I followed a precedent I found for a recent similar fix for AI Tutor: https://github.com/code-dot-org/code-dot-org/pull/58364

This results in our familiar purple being used for code text in this CodeBridge level:
image

## Links

Slack thread: https://codedotorg.slack.com/archives/C06JELCAPT9/p1715895412065549
Data Science Sample progression: https://docs.google.com/document/d/1gE8tuXM2g5ZqsbomSWC0AvU5EJSvDuc7xICHwsdTHUc/edit#heading=h.w7zi4jv346jl

## Testing story

It looks like the original color was added intentionally based on this commit: https://github.com/code-dot-org/code-dot-org/commit/6e10321bb2d288f2d976f37f9e2f1431ddb092fc

However, I tested that this bug existed in Python Lab as well as Music Lab (where I'm guessing we must not be using this formatting anywhere). It looks like the new Music Lab tutorial uses separate styles for its callout links It would be good to confirm that changing this wouldn't cause any problems.
DOM view of clickable text element in Music Lab tutorial level:
image

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/lab2/views/components/instructions.module.scss
+++ b/apps/src/lab2/views/components/instructions.module.scss
@@ -186,9 +186,7 @@
   }
 
   code {
-    // Color copied for consistency with bootstrap styles auto-applied to SafeMarkdown component
-    background-color: #f7f7f9;
-    color: $purple;
+    color: $light_primary_700;
   }
 
   p {
--- a/apps/src/lab2/views/components/instructions.module.scss
+++ b/apps/src/lab2/views/components/instructions.module.scss
@@ -155,7 +155,8 @@
   border-radius: 4px;
   height: 38px;
   margin: 0 0 10px 0;
-  &:disabled, &[aria-disabled="true"] {
+  &:disabled,
+  &[aria-disabled='true'] {
     cursor: not-allowed;
     border-color: $light_gray_200;
     background-color: $light_gray_200;
@@ -166,7 +167,10 @@
 .markdownText {
   line-height: 0;
 
-  h1, h2, h3, h4 {
+  h1,
+  h2,
+  h3,
+  h4 {
     margin: 0 0 10px 0;
     font-family: $barlowSemiCondensed-semibold;
     font-size: 1.75em;
@@ -182,7 +186,9 @@
   }
 
   code {
-    color: $neutral_light;
+    // Color copied for consistency with bootstrap styles auto-applied to SafeMarkdown component
+    background-color: #f7f7f9;
+    color: $purple;
   }
 
   p {


apps/src/lab2/views/components/instructions.module.scss
```diff
@@ -182,7 +186,9 @@
   }
 
   code {
-    color: $neutral_light;
+    // Color copied for consistency with bootstrap styles auto-applied to SafeMarkdown component
+    background-color: #f7f7f9;
```
breville: is there a reason we don't just use the existing background color for the instructions panel?

breville:
Quickly tested `/s/music-intro-2024` and didn't see any regressions to existing instructions styling.

apps/src/lab2/views/components/instructions.module.scss
mikeharv: Just that usually we use a slightly darker background color for code.  I grabbed this value from a Java Lab level, but I believe it's the same slightly-gray color in all other labs too.

breville: Interesting.  It must be subtle, because I don't quite see it in the screenshots.  At any rate, maybe worth hoisting into a constant, at least in this file, if not in `color.scss`?

mikeharv: I agree that it's subtle. I think it's because in other labs the primary background is pure white, but in the lab2 component it's `$neutral_light`.

breville: It was my earlier understanding that `$neutral_light` is the desired background color where we'd formerly use `white`, so Music Lab and associated Lab2 components use it quite a bit.

Given that, it might be worth picking a new background color, or not using one, for code?  When there is only a tiny bit of a difference, it can feel more like a subtle mismatch than a deliberate contrast.

markabarrett: Brendan, Mike, it would be good to demo what we have been doing design-side around variable names and light/dark mode.

mikeharv: Per Mark's suggestion in a Slack DM, I've updated the styles to use teal on white: 
image (201)

Let me know if either of you have thoughts/concerns about this. Brendan this also means we're using the existing background now, to your original question.

breville (approved):

PR #58727, opened 2024-05-17
Update csf and csd language list on curriculum marketing pages

bethanyaconnor:
How I built this string:
- I copy pasted the local language name from the cdo-languages gsheet for each language in the translateable string
- I checked with Jorge and made a couple of small adjustments
  - Jorge told me that English is not needed in these strings

CSF:
Screenshot 2024-05-17 at 3 48 01 PM

CSD:
Screenshot 2024-05-17 at 3 27 20 PM


kelbyhawn (approved):
🎉

PR #58737, opened 2024-05-17
Delete duplicate files to fix the case-insensitive filesystem glitch on macOS

artem-vavilov:


davidsbailey (approved):

Hamms:
The source of truth for this level file data is levelbuilder; will this PR put us in a state in which the content in the codebase is inconsistent with the content in levelbuilder's database? What will happen if any of these levels get saved on levelbuilder?

PR #58746, opened 2024-05-19
Music: sounds panel preview variant

breville:
This is a new variant of the original `SoundsPanel` which borrows one idea from `SoundsPanel2`: when a sound is clicked, it plays a preview and changes the selection, but doesn't dismiss the panel.  It can be tried using the `sounds-panel-1-preview=true` URL parameter.

### DIFF
--- a/apps/src/music/views/SoundsPanel.tsx
+++ b/apps/src/music/views/SoundsPanel.tsx
@@ -1,6 +1,6 @@
 import React, {useCallback, useEffect, useRef, useState} from 'react';
 import classNames from 'classnames';
-import {getBaseAssetUrl} from '../appConfig';
+import AppConfig, {getBaseAssetUrl} from '../appConfig';
 import styles from './soundsPanel.module.scss';
 import FontAwesome from '@cdo/apps/templates/FontAwesome';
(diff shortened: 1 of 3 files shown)


apps/src/music/views/SoundsPanel.tsx
```diff
@@ -162,10 +184,10 @@ const SoundsPanelRow: React.FunctionComponent<SoundsPanelRowProps> = ({
         styles.soundRow,
         isSelected && styles.soundRowSelected
       )}
-      onClick={() => onSelect(folder.id + '/' + sound.src)}
+      onClick={onSoundClick}
       onKeyDown={event => {
         if (event.key === 'Enter') {
-          onSelect(folder.id + '/' + sound.src);
+          onSoundSelect();
```
sanchitmalhotra126, line 190: Just confirming, you don't need to pass `folder.id + '/' + sound.src` to these functions because `soundPath` already stores this value right?

sanchitmalhotra126 (approved):

apps/src/music/views/SoundsPanel.tsx
breville: Yes.

PR #58761, opened 2024-05-20
Don't throw a honeybadger error on importing a blank csv

### LINKED ISSUE #58754: import_csv: ArgumentError: Empty list of attributes passed
From honeybadger: https://app.honeybadger.io/projects/3240/faults/107882182

This one's actually been going on for a while, its not related to our recent read_records changes, not sure how we missed it earlier.

snickell:
Fixes [#58754](https://github.com/code-dot-org/code-dot-org/issues/58754), which came from a honeybadger error

Was able to repro the honeybadger error by importing a blank CSV. After this PR, its succeeds silently.

`records.insert_all()`, invoked by create_records which is invoked by import_csv throws `ArgumentError: Empty list of attributes passed` if called with an empty list of record_json. This means importing a blank .csv file silently raises an error, when its really a reasonable thing to import (blank table). We detect this condition and successfully insert nothing.


cnbrenci:
Do we want this to be a supported path, actually? I can't imagine someone would do this on purpose, and since it's a destructive operation on the existing table content, it feels like we should throw up a modal and block the action instead?

snickell:
I went back and forth on this, but landed here. I take it this means you disagree? I'm working on implementing the other approach, should have something in a sec.

snickell:
Now we raise this dialog on the PR, whatcha think?
image

cnbrenci (approved):
Yaay! I think this is better, yeah. Cause then if they accidentally upload an empty file their table doesn't get wiped. Thank you for appeasing me 😁

PR #58764, opened 2024-05-20
Raise StudentFacingError.new(:TABLE_NAME_INVALID) if table name is too long

snickell:
Fixes [#58763](https://github.com/code-dot-org/code-dot-org/issues/58763)

Fixes Honeybadger Error: https://app.honeybadger.io/projects/3240/faults/107904108

Tested by:
- Manually trying to use the data browser add table button to add a table_name > 800 chars. Confirm it brings up a UX error when it fails: 
image
- Adds an integration test that the correct error is thrown to trigger the UX error.

PR #58765, opened 2024-05-20
Datablock Storage: use UX expected enum values for StudentFacingError(:VALUE)

### LINKED ISSUE #58762: Raise expected VALUE for StudentFacingError(VALUE)
To get appropriate UX errors from Data Browser forms, we need to raise hardcoded VALUE for the StudentError.

Here's some from DataOverView.jsx for clicking the create table button: https://github.com/code-dot-org/code-dot-org/blob/84d65fa95b143fc2852b9bc53955d1d505fea094/apps/src/storage/dataBrowser/DataOverview.jsx/#L39-L42
```
          (error.type === WarningType.MAX_TABLES_EXCEEDED ||
            error.type === WarningType.TABLE_NAME_INVALID ||
            error.type === WarningType.TABLE_RENAMED ||
            error.type === WarningType.DUPLICATE_TABLE_NAME)
```

We can be sure the corresponding `raise StudentError(VALUE)` use one of these, like `:TABLE_NAME_INVALID`.

Another source of these errors would be the enum itselfl (https://github.com/code-dot-org/code-dot-org/blob/19368167d875b535d69e6dddcf0b7f3f11c2a79b/apps/src/storage/constants.js/#L5-L15):
```
export const WarningType = makeEnum(
  'CANNOT_CONVERT_COLUMN_TYPE',
  'DUPLICATE_TABLE_NAME',
  'IMPORT_FAILED',
  'KEY_INVALID',
  'KEY_RENAMED',
  'MAX_TABLES_EXCEEDED',
  'TABLE_NAME_INVALID',
  'TABLE_RENAMED'
);
```

- [x] CANNOT_CONVERT_COLUMN_TYPE, already called by coerce_column
- [x] 'DUPLICATE_TABLE_NAME', already called in add_shared_table
- [x] 'DUPLICATE_TABLE_NAME', call in create_table?, our current behavior seems reasonable, we just take you to the pre-existing table
- [x]  'IMPORT_FAILED', fixed in https://github.com/code-dot-org/code-dot-org/pull/58765, called on import_csv
- [x] 'KEY_INVALID', raise KEY_INVALID instead of MAX_KEY_LENGTH_EXCEEDED [here](https://github.com/code-dot-org/code-dot-org/blob/abab88c87d826995c52d2709eff5a0325b2a87c6/dashboard/app/models/datablock_storage_kvp.rb/#L49-L50), fixed in https://github.com/code-dot-org/code-dot-org/pull/58765
- [x] 'KEY_RENAMED', I believe this was [only for compat during the dynamodb => firebase conversion](https://github.com/code-dot-org/code-dot-org/blob/abab88c87d826995c52d2709eff5a0325b2a87c6/apps/src/storage/firebaseStorage.js/#L135-L140)
- [x]  'MAX_TABLES_EXCEEDED', already raised by datablock_storage_table#validate_max_table_count
- [x]  'TABLE_NAME_INVALID', fixed by https://github.com/code-dot-org/code-dot-org/pull/58764
- [x]  'TABLE_RENAMED', I believe this was [only for compat during the dynamodb => firebase conversion](https://github.com/code-dot-org/code-dot-org/blob/abab88c87d826995c52d2709eff5a0325b2a87c6/apps/src/storage/firebaseStorage.js/#L146-L158)

snickell:
Fixes #58762 

1. Raise UX errors on import_csv:
image

2. Raise UX errors when set_key_value key is too long:
image

### DIFF
--- a/dashboard/app/models/datablock_storage_kvp.rb
+++ b/dashboard/app/models/datablock_storage_kvp.rb
@@ -46,7 +46,7 @@ class DatablockStorageKvp < ApplicationRecord
       DatablockStorageKvp.insert_all(kvps)
     end
   rescue ActiveRecord::ValueTooLong
-    raise StudentFacingError.new(:MAX_KEY_LENGTH_EXCEEDED), "The key is too large, it must be shorter than #{columns_hash['key'].limit} bytes ('characters')"
+    raise StudentFacingError.new(:KEY_INVALID), "The key is too large, it must be shorter than #{columns_hash['key'].limit} bytes ('characters')"
   end
 
   def self.set_kvp(project_id, key, value)
--- a/dashboard/app/models/datablock_storage_table.rb
+++ b/dashboard/app/models/datablock_storage_table.rb
@@ -245,7 +245,7 @@ class DatablockStorageTable < ApplicationRecord
 
     create_records(new_records)
   rescue CSV::MalformedCSVError => exception
-    raise StudentFacingError.new(:INVALID_CSV), "Could not import CSV as it was not in the format we expected: #{exception.message}"
+    raise StudentFacingError.new(:IMPORT_FAILED), "Could not import CSV as it was not in the format we expected: #{exception.message}"
   end
 
   def add_column(column_name)


snickell:
We examined all possible enum StudentFacingError conditions, and only found two more (besides https://github.com/code-dot-org/code-dot-org/pull/58764) to fix in this PR. I believe we now support all the useful error conditions supported by the firebase UX.

cnbrenci (approved):
Awesome! 

Logged [#58791](https://github.com/code-dot-org/code-dot-org/issues/58791) as a follow-up from Brendan's comment in slack. We can prioritize it against other stuff on the backlog, but no use delaying getting a fix out for it.

PR #58768, opened 2024-05-20
Datablock Storage: test commands.js

snickell:
Gets this to pass w/o Firebase:
```
yarn test:unit --entry=./test/unit/storage/DatablockStorageTest.js
```

Also need to port this to Datablock Storage: apps/test/integration/levelSolutions/applab1/ec_data_blocks.js

This PR:
1. Makes DatablockStorage _fetch() sinon stub-able / mockable
2. Test at least 4 different basic blocks are working


snickell:
we went a different way with this one, this is mostly redundant with the UI tests in `dataBlocks.feature`

PR #58771, opened 2024-05-20
[JIMT][LABS-588] - added auth list of mime-types

thomasoniii:
Added list of supported mime-types to the filebucket `allowed_file_types` method that weblab uses.

Previously, this could be exploited - if a user were to alter the HTML in the upload modal to remove the `accept` attribute on the file input and upload a file within our size constraints, we would allow it (so long as we could figure out its mime type).

This is a little too wide open for security purposes - for example, a bad actor could upload a `.dmg` or a `.exe` or a `.bin` file to propagate malware.

Now, the allowed_file_type method has a built-in authorized list of mime types it'll accept, somewhat arbitrarily set to:

```
    allowed_mime_regexes = [
      "image/",
      "text/",
      "audio/",
      "video/",
      "application/json",
      "application/ld+json",
      "application/pdf",
      "application/rtf",
    ]
```

Note that this is still more permissive that the weblab uploader, which only has `accept=".jpg, .jpeg, .jfif, .gif, .png"`, but I broadened the scope in case this method is used elsewhere to other benign seeming file types.  We may want to expand/collapse this list as needed.

## Links

- jira ticket: [LABS-588](https://codedotorg.atlassian.net/browse/LABS-588)

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/dashboard/legacy/middleware/helpers/file_bucket.rb
+++ b/dashboard/legacy/middleware/helpers/file_bucket.rb
@@ -10,8 +10,29 @@ class FileBucket < BucketHelper
   end
 
   def allowed_file_type?(extension)
-    # Allow all file types recognized by Sinatra
-    Sinatra::Base.mime_type(extension)
+    # use Sinatra to get the mime_type
+    mime_type = Sinatra::Base.mime_type(extension)
+
+    # and compare it against of our allowed regex-ish strings.
+    allowed_mime_regexes = [
+      "image/",
+      "text/",
+      "audio/",
+      "video/",
+      "application/json",
+      "application/ld+json",
+      "application/gzip",
+      "application/pdf",
+      "application/rtf",
+    ]
+
+    allowed_mime_regexes.each do |type|
+      if Regexp.new(type).match(mime_type)
+        return true
+      end
+    end
+
+    return false
   end
 
   def get_manifest(channel_id)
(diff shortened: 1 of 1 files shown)


unlox775-code-dot-org (approved):
looks great!

I wonder if there is a simple way to have it actually read the file's content to see if it really is the mime type instead of relying on the extension alone. That would likely be an expensive check, especially to be doing every time we serve a file.

Maybe a note to self for future that we do a periodic scan through the few dozen files our users have uploaded and check if their contents match their file type. :-D

unlox775-code-dot-org (commented):
Are there any tests that already test for valid/invalid types?

thomasoniii:
> looks great!
> 
> I wonder if there is a simple way to have it actually read the file's content to see if it really is the mime type instead of relying on the extension alone. That would likely be an expensive check, especially to be doing every time we serve a file.
> 
> Maybe a note to self for future that we do a periodic scan through the few dozen files our users have uploaded and check if their contents match their file type. :-D

We could shell out to `file`? Maybe ruby has a gem that does the equivalent? I'm not gonna roll my own code to start checking magic numbers. 😉 

I think we could reasonably only do it on upload, since it won't let you rename the file to an invalid extension after it's uploaded, so all you could do is download an `dmg` with a `png` extension, or whatever.

...and I just discovered that that's not _quite_ true. While you can't _save_ a file with a rename to an invalid extension, you can temporarily rename it in the file browser. i.e., if you have the project open, you can rename it to dmg and then download it _as_ a dmg file, even though it won't save as that. sigh. Lemme see if I can prevent that from even showing up - I assume the code isn't checking the return of the rename call.

sanchitmalhotra126 (approved):

thomasoniii:
Follow up work candidate filed in https://codedotorg.atlassian.net/browse/LABS-851

PR #58772, opened 2024-05-20
Add MacOS case-sensitive volume instructions to SETUP.md

lfryemason:
Based on this thread, new engineers on Mac should create a new volume that is case-sensitive:
https://codedotorg.slack.com/archives/C046G4TRLEN/p1715982918657509

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/SETUP.md
+++ b/SETUP.md
@@ -9,6 +9,28 @@ You can do Code.org development using macOS, Ubuntu, or Windows (running Ubuntu 
 
 1. Request and Configure AWS access (code.org staff) or configure local secrets (open source contributors). See [Configure AWS Access or Secrets](#configure-aws-access-or-secrets) below. This step is not required until rake is first run below, but staff may wish to submit the request first so its ready when rake is.
 
+1. MacOS only
+    <details>
+      <summary>Create a case-sensitive volume</summary>
+
+
+    * Open Disk Utility
+    
+    * Select any hard drive
+    
+    * Hit `+` Volume in top right
+    
+      * Name - name you want for volume (e.g. git or projects or code)
+    
+      * Format: APFS (Case-sensitive)
+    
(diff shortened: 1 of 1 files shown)


cat5inthecradle:
This is a "best practice for macOs development" type of instruction, and not something unique to our codebase. I think we should have a high bar for adding instructions like this. Steps that aren't truly necessary clutter our SETUP.md and make our codebase less accessible to new engineers. There's probably a better place to advise this, like onboarding docs, or perhaps a new dev environment setup tips doc.

lfryemason:
@cat5inthecradle, I disagree for a few main reasons:
1) I think the setup document here should be the source of truth. Any instructions that would be useful for setup but are not in this document are going to get ignored most of the time. For example, if my computer dies tomorrow and I need to set this up again, I am not going to remember to go to any onboarding document.
2) While this may be "best practice" for mac development, not all of the people who are going to follow these instructions will know about it. Most developers only set up their environment once every 2-5 years, which is a long time to forget best practices and newer developers might never have set something up correctly. Having best practices included in this document will only help our codebase be more approachable and easier to navigate.

I could see moving this to a new section for tips/tricks in this document.

cat5inthecradle:
I still lean against merging this as is. I think we should strive to minimize the requirements we place on engineers, whereas this adds a pretty hefty one (there are others in here I'd like to remove too). I could get behind moving this to a tips section, or as part of our "Getting Started as a Developer at Code.org" document, but as it's currently proposed, this is written like a necessary first step with no explanation of why.

I think I'd like to see some explanatory text like "Because much of our curriculum and website content is stored in the repository, and because both our content-editing and production environments use case-sensitive file systems, engineers working on case-insensitive file systems may occasionally run into issues when files are renamed upstream. To avoid this issue on macOS, you can create a case-sensitive volume to contain the cloned repository"

PR #58776, opened 2024-05-20
Added DCDO flag for modal

kobryan0619:
Previously, the modal would pop-up for anyone who had access to `V2` - for the Open Beta, we don't want this pop-up to show.  Thus, a DCDO flag was added along with some logic to only show the modal if the DCDO flag is in the right position. 

## Links

[Ticket](https://codedotorg.atlassian.net/jira/software/projects/TEACH/boards/65?selectedIssue=TEACH-1096)

## Testing story

Added tests.  Tested locally too.

## Deployment strategy

This should be merged in right before we go to Open Beta.


lfryemason (approved):

dju90:
When this is merged, the modal will continue to show until we set the flag to `true`, is that correct?

PR #58783, opened 2024-05-20
Music: use default pack if project has already been started

sanchitmalhotra126:
As we prepare for launch, we found a scenario where if a student has started a project with a library that doesn't have restricted sounds, and then opens their project later on a library that does have restricted sounds, they'll be prompted with the pack selection dialog. This may be confusing for users because once they select a pack, they are locked into the standard flow which means that if they want to change their pack, they need to start over, which is much more impactful if a student has already worked a lot on a project. This change sets the pack to the "default" Code.org pack if any changes have been made on an existing project, so when the library does update, they will be able to continue working on their project seamlessly.

## Testing story

Scenarios I tried:
- Open a project that was edited with the current (non-restricted library) with ?library=launch2024-prep. Pack dialog **does not** appear (automatically set to Code.org sounds).
- Starting a project and making changes on current (non-restricted) library. Refresh with ?library=launch2024-prep. Pack dialog **does not** appear (automatically set to Code.org sounds)
- Starting a project and saving, but making no changes. Refresh with ?library=launch2024-prep. Pack dialog **does** appear (because current code matches starter code).
- Open an existing project with edits, but then reverting back to blank (only when run block). Refresh with ?library=launch2024-prep. Pack dialog **does not** appear (pack ID has already been set to "default").
- Opening a blank project with ?library=launch2024-prep. Refreshing with the pack dialog still open. Pack dialog **does** appear since a pack has still not yet been selected.

### DIFF
--- a/apps/src/music/views/MusicView.jsx
+++ b/apps/src/music/views/MusicView.jsx
@@ -55,6 +55,7 @@ import {isEqual} from 'lodash';
 import MusicLibrary from '../player/MusicLibrary';
 import {TRIGGER_FIELD} from '../blockly/constants';
 import MusicLabView from './MusicLabView';
+import DCDO from '@cdo/apps/dcdo';
 
 const BLOCKLY_DIV_ID = 'blockly-div';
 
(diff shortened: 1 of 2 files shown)


breville (approved):
Tested locally and appears to work well.  Thanks for doing this!

PR #57938, opened 2024-04-10
Use more reliable restart strategy for ActiveJob

cat5inthecradle:
Slack Discussion: https://codedotorg.slack.com/archives/C051P2V2RN0/p1712167298464339

We have been experiencing issues with the `restart` command leaving old workers that don't have the latest code. This PR rolls back to an older version of the `daemons` gem and switches us to use an explicit `stop` and `start`.

## PR Checklist:

- [ ] Tests provide adequate coverage
- [ ] Privacy and Security impacts have been assessed
- [ ] Code is well-commented
- [ ] New features are translatable or updates will not break translations
- [ ] Relevant documentation has been added or updated
- [ ] User impact is well-understood and desirable
- [ ] Pull Request is labeled appropriately
- [ ] Follow-up work items (including potential tech debt) are tracked and linked

PR #58786, opened 2024-05-20
Revert "Use more reliable restart strategy for ActiveJob"

cat5inthecradle:
Reverts code-dot-org/code-dot-org#57938


cat5inthecradle:
preparing this revert in case we notice issues with the new logic.

PR #58790, opened 2024-05-20
Adds script to backfill some of the CAP/CPA user events.

wilkie:
This script will backfill the Cap::UserEvent table with records related to any folks that need it.

This is motivated by users being in the flow prior to the merging of specific recording of user events in our own database.

I could have built a View for this, but it felt easier to just backfill the few records in this window than maintain a database view that might have some issues when cloned to RedShift/Trevor.

## Links

- jira ticket: [P20-931](https://codedotorg.atlassian.net/browse/P20-931)

## Deployment strategy

We would merge this and run the script on the production console. It does a 'dry run' where it writes out the records it would write. Then, you can commit that by running the script again. It should be idempotent.. if there are any failures, they will be found by running the script again.

There is a finite amount of old records since this is purely backfilling up to the date of the first actual user event record.

To handle the `ACCOUNT_LOCKED` event (when a user gets to the `l` state), we have no good means of deciding this outside of querying all the `User` records with a `child_account_compliance_lock_out_date`. So, I have to run this first:

```
echo 'select id,JSON_EXTRACT(properties, "$.child_account_compliance_lock_out_date") from users where JSON_EXTRACT(properties, "$.child_account_compliance_lock_out_date") is not null;' | ./bin/mysql-client-dashboard-reporting > lockout_dates.tsv
```

Which will create the `lockout_dates.tsv` file for consumption by this oneoff script.

I have already produced this file on production. There are only 7,000 or so users in the locked out state... not too bad.

## PR Checklist:

- [ ] Code is well-commented


bin/oneoff/cpa/backfill_user_events.rb
```diff
@@ -0,0 +1,112 @@
...
+parental_permissions.each do |ppr|
+  # New parental email
+  records << {
+    created_at: ppr[:created_at].iso8601,
+    updated_at: ppr[:created_at].iso8601,
+    policy: POLICY,
+    name: PARENT_EMAIL_SUBMIT,
+    user_id: ppr[:user_id],
+    state_before: nil,
+    state_after: 's',
```
daynew:
```suggestion
    state_before: 'l',
    state_after: 's',
```
I think we can assume the state was LOCKED_OUT ("l") because new users would immedietly be locked out.
The flow was always `nil` -> `LOCKED_OUT` -> `REQUEST_SENT` -> `PERMISSION_GRANTED`.

Does that sound right?

bin/oneoff/cpa/backfill_user_events.rb
```diff
@@ -0,0 +1,112 @@
...
+  }
+  records = records[...-1] if DASHBOARD_REPORTING_DB_READER[:cap_user_events].where(policy: records[-1][:policy], name: records[-1][:name], user_id: records[-1][:user_id]).count > 0
```
daynew: Preventing duplication, right?

bin/oneoff/cpa/backfill_user_events.rb
```diff
@@ -0,0 +1,112 @@
...
+      state_before: 's',
+      state_after: 's',
+    }
+    records = records[...-1] if DASHBOARD_REPORTING_DB_READER[:cap_user_events].where(policy: records[-1][:policy], name: records[-1][:name], user_id: records[-1][:user_id]).count > 0
+  end
+
+  # Now, look to see that they have the granted compliance state
+  user = DASHBOARD_REPORTING_DB_READER[:users].where(id: ppr[:user_id]).first
+  user_properties = JSON.parse(user[:properties])
+  next unless user_properties['child_account_compliance_state'] == SharedConstants::CHILD_ACCOUNT_COMPLIANCE_STATES.PERMISSION_GRANTED
```
daynew: I checked `clear_user_and_mark_purged` and `:properties` gets set to an empty Hash, so this should be safe to do even if the users account was deleted and purged. 👍

bin/oneoff/cpa/backfill_user_events.rb
```diff
@@ -0,0 +1,112 @@
...
+    KEYS.each do |key|
+      data[key] = DateTime.parse(data[key])
+    end
+    DASHBOARD_DB[:cap_user_events] << data
+  end
+else
+  puts
+  puts "Writing 'cpa-backfill-records.json' with the records we intend to write."
+  puts "  Refer and verify this file and then re-run the script and append '1' as an argument to commit these records."
+  puts
```
daynew: Should we write the log for both cases? That way it can be added to the Jira ticket as a record? Just in case we accidentally delete the JSON and then run script with `1`?

daynew (approved):
I looked through the script and compared it to how we actively log these events.

daynew (changes requested):
Is this missing the backfilling for the "LOCKED_OUT" event? Could we use the `child_account_compliance_lock_out_date` field to generate these?

bin/oneoff/cpa/backfill_user_events.rb
```diff
@@ -0,0 +1,112 @@
+#!/usr/bin/env ruby
+
+require_relative '../../../deployment'
+require_relative '../../../lib/cdo/shared_constants'
+require 'cdo/db'
+require 'json'
+
+# We cannot require the model easily, so we just repeat the event names
```
artem-vavilov: What if we place the script in the `dashboard/scripts/oneoff`?

bin/oneoff/cpa/backfill_user_events.rb
```diff
@@ -0,0 +1,112 @@
...
+  puts
+  puts "Writing"
+  puts
+
+  records.each do |data|
+    KEYS.each do |key|
+      data[key] = DateTime.parse(data[key])
+    end
+    DASHBOARD_DB[:cap_user_events] << data
+  end
```
artem-vavilov: We can use the [ActiveRecord Import](https://github.com/zdennis/activerecord-import?tab=readme-ov-file#hashes) to insert records in batches

wilkie: The only benefit I can think of in using the import gem is that it would run rails validations. I'm not that worried about the validations or to have inserts wrapped in transactions. I think it will be ok and I can avoid loading the whole app while running on the production server.

bin/oneoff/cpa/backfill_user_events.rb
wilkie: We have similar backfill scripts already in the `bin/oneoff` path. It was getting "unwieldy" so I, as a compromise, put it in its own directory (then, now, I made that directory `platform` to show team ownership) which is maybe a little silly. But I do not want to hide oneoff scripts in the dashboard path when there is a more-used alternative.

wilkie:
See my expanded description above under "Deployment Strategy" for new information about how I will be handling the "account_locked" event.

wilkie:
I continue to assume that the "account_purged" event has never happened.

daynew (approved):
Created a production clone and ran the script. It worked as expected. Thank you!

PR #58797, opened 2024-05-21
Music: update UI test

breville:
Update UI test to skip the pack dialog when viewing a standalone Music Lab project.


apps/src/music/views/PackDialog.tsx
```diff
@@ -220,7 +220,7 @@ const PackDialog: React.FunctionComponent<PackDialogProps> = ({player}) => {
           <div className={styles.buttonContainer}>
             <button
               onClick={setPackToDefault}
-              className={styles.skip}
+              className={classNames('skip-button', styles.skip)}
```
sanchitmalhotra126, line 223: nit: might be good to rename to something like `.ui-test-music-skip-button` so it's not accidentally affected by any of the application-wide styles?

PR #58800, opened 2024-05-21
Music: miniplayer uses pack tempo and key

breville:
The miniplayer will now use the the tempo and key of a pack if the song uses one.

### DIFF
--- a/apps/src/music/views/MiniMusicPlayer.tsx
+++ b/apps/src/music/views/MiniMusicPlayer.tsx
@@ -73,8 +73,8 @@ const MiniPlayerView: React.FunctionComponent<MiniPlayerViewProps> = ({
 
     // If there is a pack ID, give the player its BPM and key.
     const currentLibrary = MusicLibrary.getInstance();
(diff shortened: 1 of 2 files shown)


apps/src/music/views/MiniMusicPlayer.tsx
```diff
@@ -71,6 +71,17 @@ const MiniPlayerView: React.FunctionComponent<MiniPlayerViewProps> = ({
     sequencerRef.current.clear();
     workspaceRef.current.executeCompiledSong();
 
+    // If there is a pack ID, give the player its BPM and key.
+    const currentLibrary = MusicLibrary.getInstance();
+    const packId = project.labConfig?.music.packId;
+    if (currentLibrary && packId) {
```
sanchitmalhotra126: We should also set the player back to the default BPM/key if there's no pack ID, right?

breville: Great catch.  Done.

sanchitmalhotra126 (approved):

PR #58807, opened 2024-05-21
Fix sad bee on My Professional Learning page when a workshop has no sessions

bethanyaconnor:
We got a report that the My Professional Learning page was showing a sad bee to a long time regional partner. I found the corresponding [Honeybadger error](https://app.honeybadger.io/projects/3240/faults/105696509), which looks like we're not handling a start and end time of null, which this PR handles.

I dug into our database a bit and it looks like in 2015 fifteen workshops were created without sessions. There were over 100 enrollments across these workshops, but I assume the vast majority of those users are no longer active. 

## Testing story

I did not explicitly test this case on the My Professional Learning page, instead relying on the added unit test. I confirmed that the added test threw and error before this change and passed after.

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/dashboard/app/models/pd/workshop.rb
+++ b/dashboard/app/models/pd/workshop.rb
@@ -683,6 +683,7 @@ class Pd::Workshop < ApplicationRecord
   end
 
   def workshop_date_range_string
+    return "N/A" if sessions.empty?
     if workshop_starting_date == workshop_ending_date
       workshop_starting_date.strftime('%B %e, %Y')
     else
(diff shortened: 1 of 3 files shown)


kelbyhawn (approved):
I think this makes sense, but Hannah if you want to take a look please do!

hannahbergam (approved):
Thanks for unit testing! I'd be interested to see what this looks like live but because it's such a small audience it seems fine to verify post-merge. (Happy to assume identity after to look!)

bethanyaconnor:
Good call on verifying once it's live!

PR #58811, opened 2024-05-21
Add dotd flaky test tracking reminders

lfryemason:
Adds two reminders:

When a DTT fails, it posts a new message with a link to the flaky test tracker
Every day at 8:00 AM PT, the DOTD will get a DM from the DOTD app

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/aws/ci_build
+++ b/aws/ci_build
@@ -181,7 +181,7 @@ def main
       if CDO.test_system?
         DevelopersTopic.set_dtt 'no (robo-DTT failed)'
         ChatClient.log "*** Before you manually re-run a test, please ensure that the team who owns that test is actively tracking de-flaking. ***", color: 'yellow'
-        ChatClient.log "*** Please update the <https://docs.google.com/spreadsheets/d/1DD6ebHM-2fsiMxxYCzhJYd1DzA69Z20qrCItaER9vto|Flaky Test Tracker> ***", color: 'red'
+        ChatClient.log "*** Please update the <https://docs.google.com/spreadsheets/d/1DD6ebHM-2fsiMxxYCzhJYd1DzA69Z20qrCItaER9vto|Flaky Test Tracker> with any failing tests. ***", color: 'red'
       end
     end
 
--- a/aws/ci_build
+++ b/aws/ci_build
@@ -160,6 +160,8 @@ def main
 
         # Schedule a reminder for the dotd to check Zendesk in 2 hours
         Slack.remind(Slack.user_id(DevelopersTopic.dotd), Time.now.to_i + 7200, 'Reminder: check <https://codeorg.zendesk.com/agent/filters/44863373|Zendesk>')
+        # Remind dotd to update the flaky test tracker on DTP as a once-ish a day reminder.
+        Slack.remind(Slack.user_id(DevelopersTopic.dotd), Time.now.to_i, 'Please update the <https://docs.google.com/spreadsheets/d/1DD6ebHM-2fsiMxxYCzhJYd1DzA69Z20qrCItaER9vto|Flaky Test Tracker>')
 
         # Check hoc_mode and hoc_launch only on successful DTPs, so that we get about 1 reminder per
         # day to bring staging, test and production to the same hoc_mode and hoc_launch after a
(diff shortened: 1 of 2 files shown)


aws/ci_build
```diff
@@ -160,6 +160,8 @@ def main
 
         # Schedule a reminder for the dotd to check Zendesk in 2 hours
         Slack.remind(Slack.user_id(DevelopersTopic.dotd), Time.now.to_i + 7200, 'Reminder: check <https://codeorg.zendesk.com/agent/filters/44863373|Zendesk>')
+        # Remind dotd to update the flaky test tracker on DTP as a once-ish a day reminder.
+        Slack.remind(Slack.user_id(DevelopersTopic.dotd), Time.now.to_i, 'Please update the <https://docs.google.com/spreadsheets/d/1DD6ebHM-2fsiMxxYCzhJYd1DzA69Z20qrCItaER9vto|Flaky Test Tracker>')
```
cat5inthecradle: Is this a one time reminder, or a recurring reminder?

lfryemason: It will remind each DTP

lfryemason: That was the best mechanism I could find. Ideally it would be once a day around 9:00 am PT, but I couldn't find a way to do that

davidsbailey: to do this once every morning, you could add this to bin/cron/update_dotd

davidsbailey: or create a new cronjob, if that's not too heavy-handed

lfryemason: Perfect! Updated to put into update_dotd instead.

bin/cron/update_dotd
```diff
@@ -25,6 +25,9 @@ def main
   # Set the new developers room topic
   new_topic = current_topic.sub(/^.+?;/, "DOTD: @#{primary_dotd_user};")
   Slack.update_topic('developers', new_topic) unless new_topic == current_topic
+
+  # Remind dotd to update the flaky test tracker at 8am PST.
+  Slack.remind(Slack.user_id(DevelopersTopic.dotd), Time.now.to_i + 3600, 'Please update the <https://docs.google.com/spreadsheets/d/1DD6ebHM-2fsiMxxYCzhJYd1DzA69Z20qrCItaER9vto|Flaky Test Tracker>')
```
davidsbailey: maybe update_dotd was not an ideal suggestion, because it gets run twice, to avoid needing to be updated twice a year for timezone changes: https://github.com/code-dot-org/code-dot-org/blob/b6bde7529a20316a396cfe871c1634d94e38e268/cookbooks/cdo-apps/templates/default/crontab.erb#L64

as a result, I would expect this reminder to get posted twice.

my recommendation would be to expand the `unless new_topic == current_topic` condition above and include this reminder within that block, as a hacky way of making this only get run once

davidsbailey: my other comment here, perhaps more important, is that this message seems a bit un-actionable. What are you expecting the dotd to do in response to this message? I am wondering if the message on each DTT failure is perhaps the most timely and effective way to prompt the dotd.

davidsbailey (approved):
LGTM after considering comments

aws/ci_build
cat5inthecradle: Oh, I guess the ZenDesk reminder works the same way. I was worried this might create a daily reminder, such that on the third day, the current and two prior DOTD's would get reminders.

bin/cron/update_dotd
lfryemason: @davidsbailey I changed the mechanics to tag the DOTD and print out the message when a test build fails. I also added a bit more context to the message.

I know this will tag the DOTD twice on a build failure, but I think that will be helpful temporarily to make sure that developers see the new message.

bin/cron/update_dotd
```diff
@@ -24,7 +24,11 @@ def main
 
   # Set the new developers room topic
   new_topic = current_topic.sub(/^.+?;/, "DOTD: @#{primary_dotd_user};")
-  Slack.update_topic('developers', new_topic) unless new_topic == current_topic
+  unless new_topic == current_topic
+    Slack.update_topic('developers', new_topic)
+
+    Slack.remind(Slack.user_id(DevelopersTopic.dotd), Time.now.to_i + 3600, msg)
```
davidsbailey: won't you need the `msg` to be defined in this file also? for verification, I would suggest running whatever is inside this block once in rails console on whichever machine runs this cronjob

davidsbailey (approved):

cat5inthecradle (approved):

PR #58814, opened 2024-05-21
P20-548: Fix translation of `Level#available_callouts`

artem-vavilov:
## Links
- JIRA ticket: [P20-548](https://codedotorg.atlassian.net/browse/P20-548)


mgc1194 (commented):
Is there unit test coverage for this scenario?

mgc1194 (approved):
The changes look great!.
It looks like [there is](https://github.com/code-dot-org/code-dot-org/blob/7966f13b0e9e564419d6959d19eb86f6e7546a6e/dashboard/test/models/level_test.rb#L672) some test coverage, although it would be ideal to have a test case for this scenario.

dashboard/test/test_helper.rb
```diff
@@ -1,5 +1,4 @@
 require 'test_reporter'
-require 'rspec'
```
artem-vavilov, line 2: It was added by mistake: https://github.com/code-dot-org/code-dot-org/pull/53573/files#diff-f9b0c262bdff2e77cde6d07ffbfda6bb970043fbd0d63bfb617dbafd57ffac7dR2

dashboard/test/models/levels/level_test.rb
```diff
@@ -0,0 +1,112 @@
...
+      end
+    end
+
+    context 'when level is custom with callout_json' do
+      let(:level_is_custom) {true}
+
+      let(:callout_element_id) {'expected_callout_element_id'}
+      let(:callout_localization_key) {'expected_callout_localization_json'}
+      let(:callout_text) {'expecte_callout_text'}
+      let(:callout_qtip_config) {'expectd_callout_qtip_config'}
```
daynew: nit-pick: "expected" is misspelled on both lines.

dashboard/app/models/levels/level.rb
```diff
@@ -203,16 +203,13 @@ def should_localize?
   def available_callouts(script_level)
     if custom?
       if callout_json.present?
-        return JSON.parse(callout_json).map do |callout_definition|
-          i18n_key = "data.callouts.#{name}.#{callout_definition['localization_key']}"
-          callout_text = (should_localize? &&
-            I18n.t(i18n_key, default: nil)) ||
-              callout_definition['callout_text']
+        callouts_i18n = should_localize? ? I18n.t(name, scope: %i[data callouts], default: {}).with_indifferent_access : {}
```
daynew, line 206: Is it the `with_indifferent_access` which fixes the issue reported in the JIRA?

Could you explain the root cause of this bug?

dashboard/test/models/levels/level_test.rb
```diff
@@ -0,0 +1,112 @@
...
+          callout_text: callout_text,
+          qtip_config: callout_qtip_config.to_json,
+          on: callout_on,
+        }
+      end
+
+      context 'if it is localizable and i18n string exists' do
+        let(:level_is_localizable) {true}
+
+        let(:callout_localization_key) {''}
```
daynew, line 1422: Nice, this matches what is in the bug report, right? `{"": "some string"}`

daynew (approved):
Just some questions. Nice tests!

dashboard/app/models/levels/level.rb
artem-vavilov: [`with_indifferent_access`](https://api.rubyonrails.org/v6.1.7.7/classes/ActiveSupport/HashWithIndifferentAccess.html) allows hash values to be accessed using either symbol (`:key`) or string (`'key'`) key types interchangeably, eliminating the need to worry about which type of key is used.

artem-vavilov: The issue is that `callout_definition['localization_key']` can be an empty string, which shouldn't be allowed. Consequently, [`I18n.t("data.callouts.#{name}.#{callout_definition['localization_key']}")`](https://github.com/code-dot-org/code-dot-org/blob/a84f0281231cb28c9b4616e010c99e66a90294ec/dashboard/app/models/levels/level.rb#L207-L210) will return a hash of all values associated with the one-level-higher key:
```bash
[development] dashboard > I18n.t("data.callouts.courseD_maze_ramp5.", locale: :'fr-FR')
=> {:""=>"Ces blocs n'ont pas de nombres sur eux. Tu peux en utiliser autant que tu veux !"}
```

artem-vavilov: Because [the I18n library does not allow a key to be an empty string](https://github.com/ruby-i18n/i18n/blob/55c7750a79aff26caa1a0b053c3a5b4432d23160/lib/i18n/backend/base.rb#L29), I adjusted the code to retrieve the value from the i18n hash instead as, in Ruby, a hash key can be any object.

artem-vavilov: Because [the I18n library does not allow a key to be an empty string](https://github.com/ruby-i18n/i18n/blob/55c7750a79aff26caa1a0b053c3a5b4432d23160/lib/i18n/backend/base.rb#L29), I modified the code to retrieve the value from the i18n hash instead (as, in Ruby, a hash key can be any object).

artem-vavilov: Because [the I18n library does not allow a key to be an empty string](https://github.com/ruby-i18n/i18n/blob/55c7750a79aff26caa1a0b053c3a5b4432d23160/lib/i18n/backend/base.rb#L29), I modified the code to retrieve the i18n string from the i18n hash instead (as, in Ruby, a hash key can be any object).

stephenliang (approved):

PR #58817, opened 2024-05-21
Music: update gallery

breville:
This gets the internal admin-only gallery at https://studio.code.org/musiclab/gallery working with packs.


dashboard/app/controllers/musiclab_controller.rb
```diff
@@ -14,9 +14,10 @@ def gallery
 
     @channel_ids = Project.
       where(project_type: "music").
-      last(15).
+      last(30).
```
sanchitmalhotra126, line 17: Would be great to make this DCDO (or otherwise) configurable at some point!

sanchitmalhotra126 (approved):

PR #58826, opened 2024-05-21
[Sprite Lab] fix bitmap field for smaller images

mikeharv:
See thread: https://codedotorg.slack.com/archives/C03DBDN67B7/p1716324039544319

### DIFF
--- a/apps/src/blockly/addons/cdoFieldBitmap.ts
+++ b/apps/src/blockly/addons/cdoFieldBitmap.ts
@@ -13,9 +13,10 @@ export class CdoFieldBitmap extends FieldBitmap {
   constructor(
     value?: number[][] | null,
     options?: object | null,
-    config?: object | null
+    config?: {fieldHeight?: number} | null
   ) {
     super(value, options, config);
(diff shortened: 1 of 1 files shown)


apps/src/blockly/addons/cdoFieldBitmap.ts
```diff
@@ -37,4 +38,11 @@ export class CdoFieldBitmap extends FieldBitmap {
     const bitmap = JSON.parse(fieldElement.textContent || '[]');
     this.setValue(bitmap);
   }
+
+  setValue(newValue: number[][], fireChangeEvent = true) {
+    super.setValue(newValue, fireChangeEvent);
+    if (newValue?.length) {
```
molly-moen, line 44: (can be a follow-up): Can you add a comment explaining why we need to do this and what the calculation is doing?

molly-moen (approved):

apps/src/blockly/addons/cdoFieldBitmap.ts
mikeharv: Yes! Will do a fast follow tomorrow

PR #58833, opened 2024-05-22
Updating Sauce Labs tunnel instructions

daynew:
I wanted to setup Sauce Labs tunnel for the first time on my laptop but I ran into problems because the commands didn't match what we have documented. It looks like the `sc` Sauce Connect Proxy CLI has been updated to require more arguments:
* `region` e.g. "us-west"
* `tunnel-name` e.g. "dayne_localhost"
  * The important part of the tunnel name is that it has to matches what you use when running `sc run --tunnel-name`

This PR:
* Creates a new config value `CDO.saucelabs_tunnel_name` so the UI tests can find the tunnel `sc` has started.
* Update the README file related to running the Sauce Labs tunnel.

An alternative to creating the config property `CDO.saucelabs_tunnel_name` is to set `CDO.circle_run_identifier` in your locals.yml. However, I think creating a new config value is more readable.

## Testing story
* Started the tunnel and successfully ran UI tests.

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/dashboard/test/ui/README.md
+++ b/dashboard/test/ui/README.md
@@ -55,7 +55,7 @@ If you want to run tests on Sauce Labs against localhost you need to set up your
 4. Start the sauce labs tunnel
     - `sc run -u <saucelabs_username> -k <saucelabs_authkey> -r us-west --tunnel-name <saucelabs_tunnel_name>`
 5. Run your UI test
-    - `./runner.rb -l -f features/platform/policy_compliance.feature -c Chrome --html`
+    - `./runner.rb -l -c Chrome --html -f features/platform/policy_compliance.feature`
         - The log output can be found in `log/*.html`
 #### Older versions of Sauce Connect Proxy CLI
 1. Login to Sauce Labs and download the [tunnel](https://app.saucelabs.com/tunnels).
--- a/dashboard/test/ui/README.md
+++ b/dashboard/test/ui/README.md
@@ -47,6 +47,17 @@ You can find the values for these settings in your saucelabs account settings (`
 
 If you want to run tests on Sauce Labs against localhost you need to set up your tunnel:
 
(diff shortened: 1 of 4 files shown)


dashboard/test/ui/README.md
```diff
@@ -47,6 +47,17 @@ You can find the values for these settings in your saucelabs account settings (`
 
 If you want to run tests on Sauce Labs against localhost you need to set up your tunnel:
 
+#### Latest version of Sauce Labs tunnel
```
breville: It might be worth mentioning which version these instructions were written for, so that if a newer version is released, we can double-check whether the instructions still work?

dashboard/test/ui/features/support/connect.rb
```diff
@@ -32,7 +32,8 @@ def saucelabs_browser(test_run_name)
     idleTimeout: 60,
     seleniumVersion: Selenium::WebDriver::VERSION
   }
-  sauce_options[:tunnelIdentifier] = CDO.circle_run_identifier if CDO.circle_run_identifier
+  tunnel_name = CDO.circle_run_identifier || CDO.saucelabs_tunnel_name
```
Hamms, line 39: We only ever expect `saucelabs_tunnel_name` to be used in the context of using `sc` for local testing, right? If so, could we add a comment to that effect? Specifically, something that explains to future code archaeologists why we're considering it as an alternative to `circle_run_identifier` here, but not on line 31 above.

Hamms (approved):
One comment requested, but otherwise this looks good to me! I'm looking forward to not getting confused about the discrepancy regarding tunnel names between official docs and our docs the next time I need to use local sauceconnect; thank you! :smile:

breville (approved):
Thanks for taking the time to do this!

PR #58846, opened 2024-05-22
Add index on user_id to pd_enrollments table

bethanyaconnor:
We've received reports of the My Professional Learning page loading slowly. I dug into New Relic data and confirmed that we have several instances of the page taking >10 seconds to load. I don't have a root cause of this slowness yet, but I did notice that we call a function `Pd::Enrollments.for_user` a couple of times, which isn't fully indexed. In particular, the email column does have an index, while the user_id column does not, but this method [queries on both](https://github.com/code-dot-org/code-dot-org/blob/5747cb43b611806d365d40ea77d16b80537936f6/dashboard/app/models/pd/enrollment.rb#L99). Given we are querying on this column frequently, adding an index seemed like low hanging fruit that may offer a modest improvement.

I did look into whether we could simply stop querying by `user_id` and found that about 12000 enrollments user a different email on the enrollment than on the user. This may be due to the fact that we allow teachers to supply an alternative email for PD in case they don't have access to their school email during the summer, but I didn't look too deeply and instead would like to stick with the conservative approach of not changing the logic of this method.

A git hook spit out this message:

```
Looks like you are adding a column, changing a column or adding an index in this migration:
dashboard/db/migrate/20240522163806_add_index_to_pd_enrollments_user_id.rb
Making these types of changes on a large table (>10M rows) needs to be reviewed and
tested with the Infrastructure team to avoid negatively impacting production database performance.
The may cause MySQL to rebuild the entire table.
For more information see https://dev.mysql.com/doc/refman/5.7/en/innodb-online-ddl-operations.html#online-ddl-column-operations.
```

The pd_enrollments table has just over 220,000 rows as of 5/22, so I believe this migration should be safe. That said, I'm tagging the infrastructure team just in case.

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/dashboard/app/models/pd/enrollment.rb
+++ b/dashboard/app/models/pd/enrollment.rb
@@ -25,6 +25,7 @@
 #  index_pd_enrollments_on_code            (code) UNIQUE
 #  index_pd_enrollments_on_email           (email)
 #  index_pd_enrollments_on_pd_workshop_id  (pd_workshop_id)
+#  index_pd_enrollments_on_user_id         (user_id)
 #
 
 require 'cdo/code_generation'
--- a/dashboard/db/migrate/20240522163806_add_index_to_pd_enrollments_user_id.rb
+++ b/dashboard/db/migrate/20240522163806_add_index_to_pd_enrollments_user_id.rb
@@ -1,0 +1,5 @@
+class AddIndexToPdEnrollmentsUserId < ActiveRecord::Migration[6.1]
+  def change
+    add_index :pd_enrollments, :user_id
+  end
+end
--- a/dashboard/db/schema.rb
+++ b/dashboard/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2024_05_14_181513) do
+ActiveRecord::Schema.define(version: 2024_05_22_163806) do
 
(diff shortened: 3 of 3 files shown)


jordan-springer (approved):

jordan-springer:
Re: "I did look into whether we could simply stop querying by user_id and found that about 12000 enrollments user a different email on the enrollment than on the user. This may be due to the fact that we allow teachers to supply an alternative email for PD in case they don't have access to their school email during the summer, but I didn't look too deeply and instead would like to stick with the conservative approach of not changing the logic of this method."


@nataliazm99 is this a known data issue for us with how we are using PD data?

dashboard/app/models/pd/workshop.rb
```diff
@@ -683,6 +683,7 @@ def workshop_ending_date
   end
 
   def workshop_date_range_string
+    return I18n.t('not_applicable_abbreviation') if sessions.empty?
```
hannahbergam: Was this work from a different context?

bethanyaconnor: ....weird. Yes, give me one second to fix

hannahbergam (approved):
This looks good to me!

nataliazm99:
@jordan-springer To your [question above](https://github.com/code-dot-org/code-dot-org/pull/58846#issuecomment-2132287286): we know about this situation, that is why for PD we rely on the `studio_person_id` field and then go through a data cleanup process with the programs team.

PR #58850, opened 2024-05-22
Gen AI: chat logs to CSV script

bencodeorg:
Generates a CSV of chat logs for a given section on a given date. Plan for now would be to run this on `production-console`, then copy down the generated CSV to a dev machine to send to a teacher. Note that this is a placeholder for the pilot until we can build a more self-serve solution for teachers.

Here's an example of what the generated CSV would look like: [link](https://docs.google.com/spreadsheets/d/1u2oqHCMVRQWyLv-LuUVrIi5WfV23z256PCYMX2W60pU/edit#gid=1545249476)

To run: `./bin/oneoff/aichat_logs_to_csv --section_id 123 --date 2024-05-21`

## Links

- jira: https://codedotorg.atlassian.net/browse/LABS-843

## Testing story

Tested manually with two students in a section that I created that the CSV generated looked reasonable.

## Follow-up work

We will probably need to do a bit of follow-up work to support requests for a date range (rather than single date) or a single student (rather than a whole section), but this should work for the one request we've gotten for logging.

### DIFF
--- a/bin/oneoff/aichat_logs_to_csv
+++ b/bin/oneoff/aichat_logs_to_csv
@@ -16,8 +16,8 @@ def main
 
   CSV.open("aichat_logs_#{options[:section_id]}_#{options[:date]}.csv", "w") do |csv|
     headers = %w[
-      session_id user_id username name level_id script_id project_id created_at
-      updated_at selectedModelId temperature systemPrompt retrievalContexts
+      session_id user_id username name path created_at updated_at
+      selectedModelId temperature systemPrompt retrievalContexts
       status role content
     ]
     csv << headers
@@ -33,14 +33,16 @@ def main
       messages = JSON.parse(session.messages)
       user = session.user
 
+      level = Level.find(session.level_id)
+      script_level = level.script_levels.find_by_script_id(session.script_id)
+      path = script_level.path
+
       base_row = [
         session.id,
         session.user_id,
         user.username,
         user.name,
-        session.level_id,
-        session.script_id,
-        session.project_id,
+        path,
         session.created_at,
         session.updated_at,
         model_customizations["selectedModelId"],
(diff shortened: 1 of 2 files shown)


bin/oneoff/aichat_logs_to_csv
```diff
@@ -0,0 +1,65 @@
...
+    sessions = AichatSession.where(
+      user: student_users_in_section,
+      created_at: options[:date].beginning_of_day..options[:date].end_of_day
+    )
+
+    sessions.each do |session|
+      model_customizations = JSON.parse(session.model_customizations)
+      messages = JSON.parse(session.messages)
+      user = session.user
+
```
fisher-alice, line 35: I think maybe the script level path may be more useful to the teacher than the level_id and script_id. You could provide the level path:
```
level = Level.find(session.level_id)
script_level = level.script_levels.find_by_script_id(session.script_id)
path = script_level.path
```

fisher-alice (approved):
Looks good and ran locally with success! Just left one suggestion.

bin/oneoff/aichat_logs_to_csv
bencodeorg: Good call, updated!

bencodeorg:
Merging without Drone, since this is a oneoff script I'd like to get to production today (linting appears to be passing)

bin/oneoff/aichat_logs_to_csv
sanchitmalhotra126: Just calling out that this may not work for sublevels (since I've been deep in similar code for bubble choice levels in lab2 😅). Sublevels I believe will not have a `script_levels` field, but if we have access to the sublevel position, we can compute this path with `build_script_level_path(script_level, {sublevel_position: index + 1})`

bencodeorg: haha I was wondering this as well :) Kind of expecting something might go wrong running this and might require a bit of manual editing, good to know!

PR #58853, opened 2024-05-22
Music: increment counter metric on level load

sanchitmalhotra126:
Add some more metric reporting so we can better track the number of loads. This metric is reported with a dimension that categorizes it as either a project or level load, and a dimension that identifies the "mode" in which the project/level is (edit, view/readonly, or share).

## Testing story

Tested locally with music intro script and standalone projects. Tried edit, view, and share modes, and teacher viewing student work on the intro script.

### DIFF
--- a/apps/src/music/views/MusicView.jsx
+++ b/apps/src/music/views/MusicView.jsx
@@ -327,6 +327,23 @@ class UnconnectedMusicView extends React.Component {
         (AppConfig.getValue('show-sound-filters') === 'true' ||
           levelData?.showSoundFilters)
     );
+
+    Lab2Registry.getInstance()
+      .getMetricsReporter()
+      .incrementCounter('LevelLoad', [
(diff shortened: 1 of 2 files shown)


apps/src/music/views/MusicView.jsx
```diff
@@ -327,6 +327,23 @@ class UnconnectedMusicView extends React.Component {
         (AppConfig.getValue('show-sound-filters') === 'true' ||
           levelData?.showSoundFilters)
     );
+
+    Lab2Registry.getInstance()
+      .getMetricsReporter()
+      .incrementCounter('LevelLoad', [
```
sanchitmalhotra126, line 333: This is something I could even see being helpful cross-Lab2 eventually, perhaps? (@molly-moen @breville)

breville (approved):

PR #58855, opened 2024-05-22
Using us ip and persisting school data

hannahbergam:
This PR combines two upgrades: detecting US IP addresses to prepopulate country, and saving school selections in session storage so if a user tries to save with other form errors, their fields will keep the information they entered in their prior attempt. 

I tried to cover as many edge cases in one round as I could think of (while utilizing attempts to save incomplete registration fields to show the most common use case of this work). 

https://github.com/code-dot-org/code-dot-org/assets/37230822/5779dcd7-feea-4797-8fd7-4c6628e98caa

## Links

- spec: []()
- jira ticket: https://codedotorg.atlassian.net/browse/ACQ-1905
- https://codedotorg.atlassian.net/browse/ACQ-1906

## Testing story

See screen recording above!

Additionally, I added tests to the schoolDataInputs test file so that I could check the functionality of School Association on the whole.

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/templates/SchoolZipSearch.jsx
+++ b/apps/src/templates/SchoolZipSearch.jsx
@@ -23,18 +23,24 @@ const SEARCH_DEFAULTS = [
   {value: NO_SCHOOL_SETTING, text: i18n.noSchoolSetting()},
 ];
 const ZIP_REGEX = new RegExp(/(^\d{5}$)/);
+const SCHOOL_ZIP = 'schoolZip';
+const SCHOOL_ID = 'schoolId';
 
 // Controls the logic and components surrounding a zip input box and its error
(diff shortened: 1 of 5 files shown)


apps/src/templates/SchoolZipSearch.jsx
bethanyaconnor, line 1: Is it worth adding tests for any of the logic in this file? (And maybe `SchoolDataInputs`?)

hannahbergam: I ended up adding a bunch to SchoolDataInputs because it's much easier to test the behavior from a user perspective if they update the country (which is in that parent component). Let me know how you feel about the coverage I added!

bethanyaconnor (approved):
Looks good!

PR #58867, opened 2024-05-23
[AI Tutor] CT-602: Update find_project_and_version_id and add tests

ebeastlake:
Currently, project_id and project_version_id (expected in the AI Tutor Interactions schema, see: https://docs.google.com/document/d/18nEmRjYT0v18jBsovFWyaL-8eayHHFCSUnjXOeMxjq0/edit#heading=h.hm8mqg2r3w5m) are never populated in our database because the parameter they're gated in, `isProjectBacked`, is false for all of our levels.

From @molly-moen: isProjectBacked may be a very confusing field. "Project backed" can mean "multiple levels share the same project", rather than "this level has a project". I'm not 100% sure if that's the case here but that's usually how we define it

It turns out Molly was exactly right! This PR updates `find_project_and_version_id` to be called on every level. I refactored the code, adding early returns since we expect this could be called on levels _without_ this data now, and added tests.

## Links

Jira ticket: https://codedotorg.atlassian.net/browse/CT-602

## Testing story

I tested this locally and confirmed that now, the project ID and project version ID are saved to the database.

Screenshot 2024-05-22 at 6 41 03 PM

## Follow-up work

We noticed this bug when a student's code was flagged as profane, and we could not reference it. For now, we'll have the project_id and project_version_id handy for manual lookups in these cases. However, if we need easier access to the student's code, we can revisit that as a separate issue. 

## PR Checklist:

- [ ] Code is well-commented


molly-moen (approved):

davidsbailey:
> From @molly-moen: isProjectBacked may be a very confusing field. "Project backed" can mean "multiple levels share the same project", rather than "this level has a project". I'm not 100% sure if that's the case here but that's usually how we define it

Great catch Molly and Emily! If it's not too much trouble, it would be great to rename this to something more accurate like `isTemplateBacked`, to help us avoid making the same mistake again.

ebeastlake:
Good call @davidsbailey! Ticket here: https://codedotorg.atlassian.net/browse/CT-607

dashboard/app/controllers/ai_tutor_interactions_controller.rb
```diff
@@ -26,11 +26,12 @@ def ai_tutor_interaction_params
...
     ai_tutor_interaction_params[:user_id] = current_user.id
     ai_tutor_interaction_params[:ai_model_version] = SharedConstants::AI_TUTOR_CHAT_MODEL_VERISON
-    if params[:isProjectBacked]
-      project_data = find_project_and_version_id(ai_tutor_interaction_params[:level_id], ai_tutor_interaction_params[:script_id])
-      ai_tutor_interaction_params[:project_id] = project_data[:project_id]
-      ai_tutor_interaction_params[:project_version_id] = project_data[:version_id]
-    end
+
+    # TODO: There might be a way to restrict this to only do this lookup for levels that use the projects system, but isProjectBacked
+    # isn't it. For now, since all the levels this is enabled for use projects/are JavaLab levels, we can just do this for all of them.
```
davidsbailey, line 31: I believe what you want is `level.channel_backed?`. Seems fine to defer until you have non-project-backed levels to enable this for.

davidsbailey (approved):

PR #58879, opened 2024-05-23
Music: triggered immediately sounds sometimes don't play

sanchitmalhotra126:
The issue here was just that we were rounding playback time to the nearest sixteenth when converting to ToneJS's transport time. Now, we'll convert to the nearest 1/1000 of a sixteenth note, which is the precision that the ToneJS Transport gives us when we ask for its current time.

Also includes a small optimization to make sure we clean up any existing effect busses when creating new ones.

https://github.com/code-dot-org/code-dot-org/assets/85528507/90b53908-eb25-4942-9537-c9d31d89f1d9

## Links

https://codedotorg.atlassian.net/browse/LABS-846

## Testing story

Tested locally on standalone projects with a few different songs of varying tempos.

### DIFF
--- a/apps/src/music/player/ToneJSPlayer.ts
+++ b/apps/src/music/player/ToneJSPlayer.ts
@@ -281,20 +281,6 @@ class ToneJSPlayer implements AudioPlayer {
       player.toDestination();
     }
 
-    const startPosition = sample.playbackPosition;
-    const currentTime = this.getCurrentPlaybackPosition();
-    if (startPosition < currentTime && startPosition - currentTime > -0.01) {
-      console.log(`Offset: ${startPosition - currentTime}`);
-      console.log(
-        `Sample time ${startPosition}; Current time ${currentTime}. Adjusting start time.`
-      );
-      console.log(
-        `start transport time: ${this.playbackTimeToTransportTime(
-          startPosition
-        )}`
-      );
-      console.log(`current transport time: ${Transport.position}`);
-    }
     player
       .sync()
       .start(this.playbackTimeToTransportTime(sample.playbackPosition));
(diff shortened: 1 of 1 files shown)


cnbrenci (approved):
Heck yeah, slap that bell 🔔

apps/src/music/player/ToneJSPlayer.ts
```diff
@@ -405,6 +407,8 @@ class ToneJSPlayer implements AudioPlayer {
   }
 
   private generateEffectBusses() {
+    // Dispose of all existing effect busses
+    Object.values(this.effectBusses).forEach(node => node.dispose());
```
breville, line 411: Related?

apps/src/music/player/ToneJSPlayer.ts
```diff
@@ -375,8 +375,10 @@ class ToneJSPlayer implements AudioPlayer {
   ): BarsBeatsSixteenths {
     const bar = Math.floor(playbackPosition);
     const beat = Math.floor((playbackPosition - bar) * 4);
-    const sixteenths = Math.floor((playbackPosition - bar - beat / 4) * 16);
```
breville, line 378: Was this failing sometimes because the `floor` was scheduling the sound too far in the past?

apps/src/music/player/ToneJSPlayer.ts
sanchitmalhotra126: Not strictly, but a small cleanup to make sure we're disposing of existing effect busses before creating new ones. Can break this out into a separate change if it would be safer?

apps/src/music/player/ToneJSPlayer.ts
sanchitmalhotra126: Yep I believe so, and it wasn't immediately clear because the sound would still appear on the timeline at the precise position

breville (approved):

apps/src/music/player/ToneJSPlayer.ts
breville: I think it's fine to include here.

PR #58880, opened 2024-05-23
Fix Violations of and Reenable `Style/IfWithBooleanLiteralBranches`

Hamms:
> Checks for redundant `if` with boolean literal branches. It checks only conditions to return boolean value (`true` or `false`) for safe detection. The conditions to be checked are comparison methods, predicate methods, and double negation (!!). `nonzero?` method is allowed by default. These are customizable with `AllowedMethods` option.

This is a pretty straightforward one! We only have a single violation, and I'd say the fix makes it more readable. It's also in a test file, so it's about as safe as changes get.

Fix applied automatically with `bundle exec rubocop --autocorrect-all --only Style/IfWithBooleanLiteralBranches`

## Links

- https://docs.rubocop.org/rubocop/cops_style.html#styleifwithbooleanliteralbranches
- https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/IfWithBooleanLiteralBranches

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -225,13 +225,6 @@ Style/GlobalVars:
 Style/HashLikeCase:
   Enabled: false
 
-# Offense count: 1
-# This cop supports unsafe autocorrection (--autocorrect-all).
-# Configuration parameters: AllowedMethods.
-# AllowedMethods: nonzero?
-Style/IfWithBooleanLiteralBranches:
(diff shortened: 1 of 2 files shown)


davidsbailey (approved):
easy peasy!

nicklathe (approved):
LGTM!

PR #58892, opened 2024-05-23
Add exemplars to Code Bridge

molly-moen:
Add Java Lab-style exemplars to Code Bridge labs (Python lab and Web lab 2). This style of exemplar is a single example, and the code is saved (encrypted) to the level itself so we don't need to have live production project links working for each exemplar. This is the preferred style of exemplars by curriculum.

This PR spans a few areas in order to get exemplars working end to end:
- A couple updates to the `level_properties` api. We now send the `exampleSolutions` links (links to the exemplar page, for old labs these were project links but now they are the level link with a url parameter) and the `exemplarSources` if they are requested via a url parameter and the user can access them (levelbuilder for editing, authorized teacher for viewing). 
- Previously the backend would serve a "no exemplar for this level" page if the level did not have an exemplar and you are on the exemplar url (adding `?exemplar=true` to the script url). To enable switching between pages without a page reload, lab2 levels do not serve that page anymore and still load the level. In `LabViewsRenderer` we now handle showing a "no exemplar" page if there is no exemplar and the user is trying to view it for the level.
- The teacher panel now pulls in the `exampleSolutions` links from lab2 if they exist and shows the "example solution" button if there is an exemplar for the level.
- We skip creating a channel id in `lab2Redux` if we are viewing or editing exemplars.
- Codebridge handles loading in the exemplar code if we are viewing or editing exemplars.
- Codebridge also handles updating the header with the exemplar edit button if we are editing exemplars.

The backend still handles not showing the exemplar page if the user is not authorized to view it. I also previously added the exemplar edit page to the extra links box, which takes you to the correct edit page.

### Teacher Panel with an exemplar
This UI is unchanged from the normal UI
Screenshot 2024-05-23 at 2 09 41 PM

### Editing an exemplar
Screenshot 2024-05-23 at 2 15 04 PM

### Viewing an exemplar
Screenshot 2024-05-23 at 2 15 15 PM

### Page if there is no exemplar
Screenshot 2024-05-23 at 2 15 38 PM

## Links
- jira ticket: [https://codedotorg.atlassian.net/browse/CT-462](CT-462)

## Testing story
Tested locally. I verified that only authorized teachers can view exemplars on levels that have them, and levelbuilders can edit exemplars.

## Follow-up work
Currently you can type in the editor while viewing an exemplar. It won’t save but this is confusing behavior. I have a [follow-up task](https://codedotorg.atlassian.net/browse/CT-608) to fix this, which depends on us adding a view only mode ([task for that](https://codedotorg.atlassian.net/browse/CT-455))

## PR Checklist:

- [ ] Code is well-commented


sanchitmalhotra126 (commented):
A few questions for my understanding - I'm realizing I haven't dealt with exemplars much so I might just be misunderstanding stuff here :)

- To clarify, `exemplarSources` refers to one example solution (i.e. one set of sources) here and `exampleSolutions` is a list of project links. However since we only have one exemplar per level, `exampleSolutions` is a list of just one link, which is the existing script level URL with `?exemplar=true`, right?
- If this is the case, do we need the `exampleSolutions` array at all? Could we have the teacher panel just dispatch a redux action that tells Lab2 to show exemplar code (or fallback if it has none)?
- Relatedly, would it make sense to always include the exemplar code if possible (i.e. if verified instructor or levelbuilder) so we don't have to make a separate call to get level properties with the `include_exemplar` param?

Also, this is definitely out of scope for this PR, but I'm wondering if there's an opportunity here for us to rethink some of these paradigms as we have been doing for Lab2 so far...for example, I'm curious how levelbuilders would feel about editing start and exemplar sources directly in levelbuilder? I could imagine us setting up a react view that just houses lab2 with a single lab, and render it within the level edit page. I'm mainly thinking in terms of how we can best separate "content" for the level (instructions, validations, starter code, exemplar code, etc) from the student's project. Again, this is a broader discussion that should probably happen elsewhere, but just capturing some ideas as I was looking at this.

apps/src/codebridge/hooks/useInitialSources.ts
```diff
@@ -15,23 +19,41 @@ export const useInitialSources = (defaultSources: ProjectSources) => {
...
+    state => state.lab.levelProperties?.exemplarSources
+  );
   // We memoize this object so that it doesn't cause an unexpected re-render.
   const projectStartSource: ProjectSources | undefined = useMemo(
     () => (levelStartSource ? {source: levelStartSource} : undefined),
     [levelStartSource]
   );
   const isStartMode = getAppOptionsEditBlocks() === START_SOURCES;
+  const isEditingExemplar = getAppOptionsEditingExemplar();
+  const isViewingExemplar = getAppOptionsViewingExemplar();
```
mikeharv, line 32: How expensive is it to `getScriptData('appoptions')`? We have six utils now that get these options and then return one particular value (`channel`, `editBlocks`,  `isEditingExemplar`, `isViewingExemplar`, `levelId`, `share`). 

I'm wondering if it would be more efficient in any way to have a single util that returns an object with all of the values we potentially need than to be repeatedly parsing JSON data from a script element in the DOM.

apps/src/lab2/views/components/NoExemplarPage.tsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import moduleStyles from './no-exemplar-page.module.scss';
+
+const NoExemplarPage: React.FunctionComponent = () => {
+  return (
+    <div className={moduleStyles.noExemplarContainer}>
+      No exemplar exists for this level.
+    </div>
```
mikeharv, line 8: How would a user potentially navigate to this page?

dashboard/app/helpers/levels_helper.rb
```diff
@@ -740,6 +740,11 @@ def lab2_options
     if level_options[:edit_blocks]
       app_options[:edit_blocks] = level_options[:edit_blocks]
     end
+    # Also pass through if we are exemplar edit or view mode.
```
mikeharv: Missing "in"?

mikeharv (commented):
To @sanchitmalhotra126's first question, my interpretation was that we need `exampleSolutions`. For older labs, it's actually a list of project links on production (and not the existing script level URL with `?exemplar=true`). I think we need to use the same structure for old and new labs because we're using the existing teacher panel to house the button(s).

apps/src/codebridge/hooks/useInitialSources.ts
molly-moen: I think it's very cheap, it's a lookup followed by a json parse of a pretty small string.

apps/src/lab2/views/components/NoExemplarPage.tsx
molly-moen: They could either add `?exemplar=true` to the current level url or if they are already on an exemplar level and click a bubble in their current progression for a level that doesn't have an exemplar.

molly-moen:
@sanchitmalhotra126 yep, @mikeharv is correct that we need `exampleSolutions` in order to reuse the existing teacher panel logic. 
For the question on always including, I went back and forth on this a lot, the reason I added the parameter was to avoid sending data (which could theoretically be somewhat long, as its the source for an example project) we wouldn't need most of the time. But I'm not committed to this pattern and could change it. Do you think it's fine to always send the exemplar sources even when we don't need them?

apps/src/lab2/views/components/NoExemplarPage.tsx
mikeharv: That makes sense. Does this style of exemplar link also open in a new tab? I'm more familiar with the project link style, but I'm trying to understand how a user would get back to the standard view.

molly-moen: Yeah it opens in a new tab.

mikeharv: Thanks. I had been wondering if we should consider adding a link her to take teachers back to the level view (or to take levelbuilders to the exemplar edit view), but given that our current flow is to close the tab, that doesn't feel as necessary.

mikeharv (approved):

molly-moen:
> For the question on always including, I went back and forth on this a lot, the reason I added the parameter was to avoid sending data (which could theoretically be somewhat long, as its the source for an example project) we wouldn't need most of the time. But I'm not committed to this pattern and could change it. Do you think it's fine to always send the exemplar sources even when we don't need them?

Replying to myself, I think it is cleaner to just always send them for verified teachers/levelbuilders, so I'll update this PR to do that.

PR #58893, opened 2024-05-23
Music: add additional logging for sound load errors

sanchitmalhotra126:
Our largest error type by far is "Error loading sounds" but we currently don't have great insight into why these errors are occurring. This adds some more logging in the hopes of helping debug and ideally root cause any issues.

- The "Error loading sounds" log payload now includes an error message for each failed sound, as well as the overall number of sounds attempted, so we can get a better of idea of what fraction of sounds failed in a particular session.
- I also added counter metrics for load attempts and load failures, so we can get a better sense of overall error rate.

Example log payload in Cloudwatch
Screenshot 2024-05-23 at 2 54 32 PM

## Links

https://codedotorg.atlassian.net/browse/LABS-398

## Testing story

Tested locally using devtools to block URLs.

### DIFF
--- a/apps/src/music/player/SoundCache.ts
+++ b/apps/src/music/player/SoundCache.ts
@@ -38,7 +38,7 @@ class SoundCache {
       updateLoadProgress?: (progress: number) => void;
     } = {}
   ): Promise<void> {
-    const failedSounds: string[] = [];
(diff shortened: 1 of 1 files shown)


thomasoniii (approved):
lgtm 🚀

breville (approved):
Thanks for this!

PR #58895, opened 2024-05-23
Use /blockly url to serve pyodide files

molly-moen:
We couldn't serve the pyodide files from the `/assets/js` path on prod because only the hashed version of the files was hosted there. However, the unhashed files are available via `/blockly/js`. After discussing with infra we decided this was a reasonable url to use instead, as we are already referencing this path in other labs, and using it unblocks us from hosting pyodide from our server.

## Links
- [slack thread](https://codedotorg.slack.com/archives/C03CK49G9/p1713201165878559)
- jira ticket: [CT-512](https://codedotorg.atlassian.net/browse/CT-512)

## Testing story
Tested that you can access pyodide locally with this configuration. This url already works on production, so we are reasonably confident this will work end to end on production.

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/pythonlab/pyodideWebWorker.js
+++ b/apps/src/pythonlab/pyodideWebWorker.js
@@ -8,9 +8,10 @@ import {loadPyodide, version} from 'pyodide';
 
 async function loadPyodideAndPackages() {
   self.pyodide = await loadPyodide({
-    indexURL: `https://cdn.jsdelivr.net/pyodide/v${version}/full`,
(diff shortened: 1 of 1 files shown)


snickell (approved):
✅, we discussed this with infra, and as of now, this is the most effective solution we found. We confirmed that this path /is/ cached by cloudfront.

Hamms (approved):
LGTM! Thanks for the explanatory comment, too!

PR #58898, opened 2024-05-24
Datablock Storage: rename_column creates a new column if the old column doesn't exist

### LINKED ISSUE #56734: Data Browser: When you initially create a new table, you can't seem to name the first column until you create a row
Is this an issue in the Firebase version too? This may be a react/UX issue not a datablock storage issue, more investigation required.

snickell:
Fixes: [#56734](https://github.com/code-dot-org/code-dot-org/issues/56734)

Firebase Storage had a weird behavior where rename_column was used to create a new column (basically if the old_column_name doesn't exist, it creates a new column). Because the databrowser UX relied on this behavior, this PR adds the same behavior to datablock storage.

### DIFF
--- a/dashboard/app/models/datablock_storage_table.rb
+++ b/dashboard/app/models/datablock_storage_table.rb
@@ -281,7 +281,11 @@ class DatablockStorageTable < ApplicationRecord
     end
 
     # Second rename the column in the table definition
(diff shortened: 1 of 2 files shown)


cnbrenci (approved):
heck yea

PR #58912, opened 2024-05-24
P20 941/instrument registration events

nicklathe:
In order to add event metrics to registration events (portal and dynamic), it was necessary to change the Statsig wrapper to allow for a `nil` user, as these events happen outside of a logged in user session. This PR does the following:

* Changes Stagsig wrapper to allow for nil user
* Adds comments on Statsig wrapper with usage examples
* Instruments metrics for both portal and dynamic registration flows
* Standardizes Dynamic Registration client to a single Class, to align with other clients (technically out of scope with this PR, but I was already here and figured I'd make the improvement. The PR isn't that big so I think it's fine, but can move this to another PR if people want)

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/dashboard/lib/clients/lti_dynamic_registration_client.rb
+++ b/dashboard/lib/clients/lti_dynamic_registration_client.rb
@@ -1,20 +1,18 @@
-module Lti
-  class DynamicRegistration
-    def initialize(registration_token, registration_endpoint)
-      raise ArgumentError unless registration_token && registration_endpoint
-      @registration_token = registration_token
-      @registration_endpoint = registration_endpoint
-    end
+class LtiDynamicRegistrationClient
+  def initialize(registration_token, registration_endpoint)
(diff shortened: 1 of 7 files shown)


dashboard/lib/clients/lti_dynamic_registration_client.rb
```diff
@@ -1,20 +1,18 @@
-module Lti
-  class DynamicRegistration
-    def initialize(registration_token, registration_endpoint)
-      raise ArgumentError unless registration_token && registration_endpoint
-      @registration_token = registration_token
-      @registration_endpoint = registration_endpoint
-    end
+class LtiDynamicRegistrationClient
```
nicklathe, line 1: This change is out of scope for this PR, however while I was in here debugging, I figured I'd align this to how the other clients are configured.

stephenliang (approved):

PR #58915, opened 2024-05-24
Gen AI: export chat history to CSV on sublevels

bencodeorg:
Fixes the chat history CSV export script to handle progress that occurred on a sublevel (eg, `/s/gen-ai-customizing-pilot-v1/lessons/6/levels/9/sublevel/2`). Follow-up to https://github.com/code-dot-org/code-dot-org/pull/58850.

## Testing story

Tested manually/locally that chat history generated on a sublevel appeared appropriately in the generated CSV, with an appropriate path. Also tested that chat history that occurred on a normal level continued to export appropriately.


sanchitmalhotra126 (approved):
Thanks for adding this!

PR #58917, opened 2024-05-24
Show save indication while saving and error if save fails

sanchitmalhotra126:
Two closely related updates to save/update behavior and UI
- while a save is in progress, disable Update/Save/Publish and chat submit buttons, and show spinners where appropriate. This is to prevent triggering another save while one is in progress, and gives the student some indication that a save is currently happening
- Show an error message if saving fails. Currently, we don't handle save failures and just show the success message ("XYZ updated at time"), even though we didn't successfully update the project. Instead now we'll just show a dismissible error message. 
  - Note that this error UI is temporary; guidance from design is to have an auto-dismissing toast notification (see [thread](https://codedotorg.slack.com/archives/C06FELVTV0Q/p1716507981002499) for more details). However, since the experience is currently a little misleading if a save fails, I opted to go ahead with this temporary UI to at least have some indication in place, and follow-up later with a different UI.

Save in-progress (successful) UI:

https://github.com/code-dot-org/code-dot-org/assets/85528507/6244d399-e366-4d4b-89f3-87892a5e6f0b

Save error UI (manually blocked requests in devtools):

https://github.com/code-dot-org/code-dot-org/assets/85528507/786a20cc-0757-44ec-a9cb-ac2f7575e0fd

## Links

https://codedotorg.atlassian.net/browse/LABS-692

## Testing story

Tested locally on allthethings; manually blocked requests in devtools to simulate save errors.

## Follow-up work

Better error UI (auto-dismissing toast): https://codedotorg.atlassian.net/browse/LABS-853

### DIFF
--- a/apps/src/aichat/redux/aichatRedux.ts
+++ b/apps/src/aichat/redux/aichatRedux.ts
@@ -77,7 +77,7 @@ const saveTypeToAnalyticsEvent: {[key in SaveType]: string} = {
   saveModelCard: EVENTS.SAVE_MODEL_CARD_INFO,
 };
 
-export type SaveType = 'updateChatbot' | 'publishModelCard' | 'saveModelCard';
+type SaveType = 'updateChatbot' | 'publishModelCard' | 'saveModelCard';
 
 export interface AichatState {
   // All user and assistant chat messages - includes too personal and inappropriate user messages.
@@ -269,7 +269,7 @@ export const onSaveFail = () => (dispatch: AppDispatch) => {
   dispatch(
     addChatMessage({
       id: getNewMessageId(),
-      role: Role.ERROR,
+      role: Role.ERROR_NOTIFICATION,
       chatMessageText: 'Error updating project. Please try again.',
       status: Status.ERROR,
       timestamp: getCurrentTime(),
@@ -401,13 +401,13 @@ const aichatSlice = createSlice({
     addChatMessage: (state, action: PayloadAction<ChatCompletionMessage>) => {
       state.chatMessages.push(action.payload);
     },
-    removeModelUpdateMessage: (state, action: PayloadAction<number>) => {
+    removeUpdateMessage: (state, action: PayloadAction<number>) => {
       const updatedMessages = [...state.chatMessages];
       const messageToRemovePosition = updatedMessages.findIndex(
         message => message.id === action.payload
       );
 
-      // Only allow removing individual messages that are model updates,
+      // Only allow removing individual messages that are model updates and error notifications,
       // as we want to retain user and bot message history
       // when requesting model responses within a chat session.
       // If we want to clear all history
@@ -415,7 +415,8 @@ const aichatSlice = createSlice({
       if (
         messageToRemovePosition < 0 ||
         (updatedMessages[messageToRemovePosition].role !== Role.MODEL_UPDATE &&
-          updatedMessages[messageToRemovePosition].role !== Role.ERROR)
+          updatedMessages[messageToRemovePosition].role !==
+            Role.ERROR_NOTIFICATION)
       ) {
         return;
       }
@@ -596,7 +597,7 @@ const {startSave} = aichatSlice.actions;
 registerReducers({aichat: aichatSlice.reducer});
 export const {
   addChatMessage,
-  removeModelUpdateMessage,
+  removeUpdateMessage,
   setNewChatSession,
   setChatSessionId,
   clearChatMessages,
--- a/apps/src/aichat/redux/aichatRedux.ts
+++ b/apps/src/aichat/redux/aichatRedux.ts
@@ -71,6 +71,14 @@ const findChangedProperties = (
 const getCurrentTimestamp = () => moment(Date.now()).format('YYYY-MM-DD HH:mm');
 const getCurrentTime = () => moment(Date.now()).format('LT');
 
+const saveTypeToAnalyticsEvent: {[key in SaveType]: string} = {
+  updateChatbot: EVENTS.UPDATE_CHATBOT,
+  publishModelCard: EVENTS.PUBLISH_MODEL_CARD_INFO,
+  saveModelCard: EVENTS.SAVE_MODEL_CARD_INFO,
+};
+
+export type SaveType = 'updateChatbot' | 'publishModelCard' | 'saveModelCard';
+
 export interface AichatState {
   // All user and assistant chat messages - includes too personal and inappropriate user messages.
   // Messages will be logged and stored.
@@ -89,7 +97,7 @@ export interface AichatState {
   // If a save is currently in progress
   saveInProgress: boolean;
   // The type of save action being performed (customization update, publish, model card save, etc).
-  currentSaveType: string | undefined;
+  currentSaveType: SaveType | undefined;
 }
 
 const initialState: AichatState = {
@@ -115,7 +123,7 @@ export const updateAiCustomization = createAsyncThunk(
   async (_, {dispatch, getState}) => {
     await saveAiCustomization(
       (getState() as RootState).aichat.currentAiCustomizations,
-      EVENTS.UPDATE_CHATBOT,
+      'updateChatbot',
       dispatch
     );
   }
@@ -130,10 +138,9 @@ export const publishModel = createAsyncThunk(
     dispatch(setModelCardProperty({property: 'isPublished', value: true}));
     await saveAiCustomization(
       (getState() as RootState).aichat.currentAiCustomizations,
-      EVENTS.PUBLISH_MODEL_CARD_INFO,
+      'publishModelCard',
       dispatch
     );
-    dispatch(setViewMode(ViewMode.PRESENTATION));
   }
 );
 
@@ -150,7 +157,7 @@ export const saveModelCard = createAsyncThunk(
 
     await saveAiCustomization(
       currentAiCustomizations,
-      EVENTS.SAVE_MODEL_CARD_INFO,
+      'saveModelCard',
       dispatch
     );
   }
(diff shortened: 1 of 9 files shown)


apps/src/aichat/redux/aichatRedux.ts
```diff
@@ -70,6 +71,14 @@ const findChangedProperties = (
 const getCurrentTimestamp = () => moment(Date.now()).format('YYYY-MM-DD HH:mm');
 const getCurrentTime = () => moment(Date.now()).format('LT');
 
+const saveTypeToAnalyticsEvent: {[key in SaveType]: string} = {
+  updateChatbot: EVENTS.UPDATE_CHATBOT,
+  publishModelCard: EVENTS.PUBLISH_MODEL_CARD_INFO,
+  saveModelCard: EVENTS.SAVE_MODEL_CARD_INFO,
+};
+
+export type SaveType = 'updateChatbot' | 'publishModelCard' | 'saveModelCard';
```
bencodeorg: Does this get used elsewhere?

apps/src/aichat/types.ts
```diff
@@ -26,6 +26,7 @@ export enum Role {
   USER = 'user',
   SYSTEM = 'system',
   MODEL_UPDATE = 'update',
+  ERROR = 'error',
```
bencodeorg: "error" as a role feels weird, but I can't think of a better option. [Ties into your comment here](https://github.com/code-dot-org/code-dot-org/pull/58668#pullrequestreview-2071960253) about types of these messages, which I agree we could improve going forward.

apps/src/aichat/redux/aichatRedux.ts
```diff
@@ -201,51 +198,85 @@ const saveAiCustomization = async (
...
+            propertyUpdated: property,
+            levelPath: window.location.pathname,
+          },
+          PLATFORMS.BOTH
+        );
+      }
+    });
+
+    // Update our last saved customizations to match the current customizations
+    dispatch(setSavedAiCustomizations(currentAiCustomizations));
```
bencodeorg, line 258: Not sure if it matters, but is there a reason not to do this at the top of this function? It seems like maybe? the most important save callback, but is done after a bunch of other stuff.

apps/src/aichat/redux/aichatRedux.ts
```diff
@@ -383,7 +414,8 @@ const aichatSlice = createSlice({
       // and start a new session, see clearChatMessages.
       if (
         messageToRemovePosition < 0 ||
-        updatedMessages[messageToRemovePosition].role !== Role.MODEL_UPDATE
+        (updatedMessages[messageToRemovePosition].role !== Role.MODEL_UPDATE &&
```
bencodeorg, line 417: I think we should probably update the comment above and rename this reducer now that we have these error messages

apps/src/aichat/redux/aichatRedux.ts
```diff
@@ -177,8 +175,7 @@ const getNewMessageId = () => {
 // model customizations (setup, retrieval, and "publish" tab)
 const saveAiCustomization = async (
   currentAiCustomizations: AiCustomizations,
-  savedAiCustomizations: AiCustomizations,
-  eventDescription: string,
+  saveType: SaveType,
```
bencodeorg, line 178: Nice simplification of these method signatures :)

apps/src/aichat/redux/aichatRedux.ts
```diff
@@ -201,51 +198,85 @@ const saveAiCustomization = async (
...
+    if (currentSaveType === 'publishModelCard') {
+      dispatch(setViewMode(ViewMode.PRESENTATION));
     }
-  });
+  };
+
+// Thunk called when a save has failed.
+export const onSaveFail = () => (dispatch: AppDispatch) => {
+  dispatch(
+    addChatMessage({
```
bencodeorg, line 270: Did you confirm that these don't show up in conversation history logs? I think they won't b/c no sessionId, but worth confirming.

apps/src/aichat/views/AichatView.tsx
```diff
@@ -61,6 +65,24 @@ const AichatView: React.FunctionComponent = () => {
 
   const allFieldsHidden = useAppSelector(selectAllFieldsHidden);
 
+  const projectManager = Lab2Registry.getInstance().getProjectManager();
+  // Attach save listeners whenever the project manager updates
+  useEffect(() => {
```
bencodeorg, line 70: This is slick!

apps/src/aichat/views/modelCustomization/RetrievalCustomization.tsx
```diff
@@ -110,13 +103,7 @@ const RetrievalCustomization: React.FunctionComponent = () => {
...
       </div>
       <div className={modelCustomizationStyles.footerButtonContainer}>
-        <Button
-          text="Update"
-          onClick={onUpdate}
-          iconLeft={{iconName: 'edit'}}
-          className={modelCustomizationStyles.updateButton}
-          disabled={isReadOnly}
-        />
+        <UpdateButton allFieldsDisabled={isReadOnly} />
```
bencodeorg: I don't think this prop name makes sense in this context...`isDisabled` isn't quite right either since the button can also be disabled based on `saveInProgress` within this button, but seems closer at least (`disabledByDefault`?)

bencodeorg (commented):
A few small comments but this looks great!

apps/src/aichat/types.ts
fisher-alice: We could temporarily call it `ERROR_NOTIFICATION`.

apps/src/aichat/views/AichatView.tsx
```diff
@@ -61,6 +65,24 @@ const AichatView: React.FunctionComponent = () => {
...
+    if (!projectManager) {
+      return;
+    }
+    projectManager.addSaveNoopListener(() => {
+      // No save occurred
```
fisher-alice: Wrt comment, isn't `endSave` called on whether a save is successful or not?

apps/src/aichat/views/AichatView.tsx
fisher-alice: +1 - love how this hooks into the `ProjectManager`!

apps/src/aichat/redux/aichatRedux.ts
fisher-alice: I confirmed they're not sent to backend.

apps/src/aichat/views/modelCustomization/RetrievalCustomization.tsx
fisher-alice: I think this is okay in that for the button,  `disabled={allFieldsDisabled || saveInProgress}`.

bencodeorg: Sorry maybe I wasn't clear, I'm saying the prop name `allFieldsDisabled` doesn't make sense to me here, since we're passing in the value `isReadOnly`?

I'm guessing Sanchit started refactoring from `SetupCustomization` (where this prop name makes sense), then re-used it here without picking a more general prop name?

fisher-alice: Gotcha.  `isDisabled` does seem a bit better.

apps/src/aichat/redux/aichatRedux.ts
bencodeorg: Looks like no as far as I can tell, removing export.

apps/src/aichat/redux/aichatRedux.ts
bencodeorg: This seems fine, leaving as-is.

apps/src/aichat/types.ts
bencodeorg: I think that's better in the short term, updated!

apps/src/aichat/redux/aichatRedux.ts
bencodeorg: Updated to `removeUpdateMessage`

apps/src/aichat/views/AichatView.tsx
bencodeorg: Yes I think you're right -- moved comment up a line to hopefully make this a bit clearer?

apps/src/aichat/views/modelCustomization/RetrievalCustomization.tsx
bencodeorg: Updated to `isDisabledDefault`

bencodeorg (commented):
Alrighty @fisher-alice, took a pass thru the feedback. LMK what you think!

apps/src/aichat/views/modelCustomization/RetrievalCustomization.tsx
fisher-alice: Nice rename.

fisher-alice (approved):
Love the refactor @sanchitmalhotra126 - enjoyed learning more about lab2's `ProjectManager`.
And thanks for carrying this through to the end @bencodeorg!

sanchitmalhotra126:
Thanks @fisher-alice and @bencodeorg for getting this through!

apps/src/aichat/redux/aichatRedux.ts
sanchitmalhotra126: Yeah, not a strong reason either way, I think my thinking at the time was to update redux as the last step after all other actions were complete, but any time after the save completes works.

apps/src/aichat/types.ts
sanchitmalhotra126: Yep I didn't feel great tacking on another role here but this name sounds good to me. Would definitely like to revisit these types and refactor a bit as we've discussed

PR #58921, opened 2024-05-24
Delaying the CPA exception date because the DTP is delayed

daynew:
The DTP was cancelled for today which means the next DTP will be after this exception date. The exception date needs to be  after the next DTP. If the DTP gets delayed again, I will use a DCDO value instead.

## Testing story
* Unit tests


stephenliang (approved):

wilkie (approved):

mgc1194 (commented):
Just a change in dates.
Thank you!

mgc1194 (approved):

PR #58922, opened 2024-05-24
remove unused AssignmentSelector component

davidsbailey:
This PR removes dead code which I discovered during investigating a prototype for curriculum modularity. EDIT: also removes most references to the now-mostly-unused `course_offering.category` field, which has been replaced by `marketing_initiative`.

## Testing story

inspected the codebase to confirm there are no code references to this component from outside the source file and unit test which are being removed.

I am relying on existing test coverage to cover the following usages which eventually hit `valid_course_offerings` and `summarize_for_assignment_dropdown`:
```
Dave-MBP:~/src/cdo/apps/src (remove-assignment-selector)$ grep -R --color 'asyncLoadSectionData(' * 
code-studio/pd/professional_learning_landing/LandingPage.jsx:    dispatch(asyncLoadSectionData());
code-studio/components/libraries/ShareTeacherLibraries.jsx:    asyncLoadSectionData();
templates/teacherDashboard/teacherSectionsRedux.js:      .then(() => dispatch(asyncLoadSectionData()))
templates/manageStudents/ManageStudentsActionsCell.jsx:      dispatch(asyncLoadSectionData(sectionId));
templates/manageStudents/manageStudentsRedux.js:        dispatch(asyncLoadSectionData(sectionId));
templates/manageStudents/manageStudentsRedux.js:        dispatch(asyncLoadSectionData(sectionId));
templates/manageStudents/manageStudentsRedux.js:            asyncLoadSectionData(state.teacherSections.selectedSectionId)
templates/studioHomepages/CoteacherInviteNotification.jsx:        asyncLoadSectionData();
templates/studioHomepages/TeacherSections.jsx:    asyncLoadSectionData();
```

I also did a bit of code searching and did not see any uses of `category` which looked like they might break.

## Follow-up work

next step is to remove this reference to category: https://github.com/code-dot-org/code-dot-org/blob/447f28022912057f0a8c4122087b3a8b70f5eb1d/dashboard/app/models/course_offering.rb#L415-L416

this will depend on updating test fixtures (https://github.com/code-dot-org/code-dot-org/pull/58951) and then fixing a few more unit tests.


apps/src/templates/teacherDashboard/AssignmentSelector.jsx
```diff
@@ -1,328 +0,0 @@
...
-import React, {Component} from 'react';
-
-import fontConstants from '@cdo/apps/fontConstants';
-import {
-  CourseOfferingCategories,
-  ParticipantAudiencesByType,
-} from '@cdo/apps/generated/curriculum/sharedCourseConstants';
-import i18n from '@cdo/locale';
-
-import {translatedCourseOfferingCategories} from './AssignmentSelectorHelpers';
```
lfryemason, line 12: Should we rename this file or move it into `CourseOfferingEditor` (the only other place it is used)?

lfryemason (commented):
WOO! I love deleting code!

apps/src/templates/teacherDashboard/AssignmentSelector.jsx
davidsbailey: ooh, good catch! It seemed suspicious to me that we'd only need the translations for the `category` in levelbuilder, so I poked around and I believe we can remove the `category` field entirely! PTAL

lfryemason (approved):

PR #58926, opened 2024-05-25
Batch entire foorm data processing sequence

cat5inthecradle:
This PR has been replaced with two options:
1. Instrument the old script #59097
2. Fully batch the process #59098

---

The existing process, "process_foorm_data" was failing for an initially unknown reason. It was presumed that the size of the data was the issue. This PR changes the script to batch the entire ETL into configurable batches (starting at 10,000), rather than simply batching the initial DB retrieval.

There are three possible solutions here, so I'm primarily looking for a suggestion on which to use. I prefer 1 or 3.

1. Move forward with "process_foorm_data_batch" with it's batching and improved error handling.
3. Just resize the redshift column and continue using the old script
4. Resize the column and use the old script, but improve the error handling and logging to make issues like this easier to detect in the future.

### Old Process: "process_foorm_data"

1. Load Submissions from the database in batches of 1000
5. Transform these submissions in memory, flattening them into multiple rows
6. Concatenate the rows to a csv string (one string to hold all 300,000 submissions, which flattens to over 3 million rows)
7. Upload the csv to S3
8. Replace the foorm submissions table in redshift with the data in the CSV.

The old process was failing during steps 1-3, somewhere in the last 25% of data. Presumably due to memory limitations with the CSV string, but I don't have a solid answer for why (EDIT: I do now, see below)

### New Process: "process_foorm_data_batch"

1. In batches of 10,000 do the following
2. Load BATCH_SIZE submissions from the database
3. Process them one at a time and add them to a csv string (just for this batch)
4. Upload the csv to a file in S3
5. Load the CSV into a temp table in redshift
9. Merge the temp table into the foorm submissions table in redshift on `submission_id`
10. Repeat with the next batch.

### Why was the old script failing?

After debugging the new script, I discovered why the old script was failing. A new very large survey answer was exceeding the column size on the redshift table. We can get around this by resizing the column in redshift.

I have manually resized the column, which may allow the regular script to succeed.

```
ALTER TABLE analysis_pii.foorm_submissions_reshaped
ALTER COLUMN response_value TYPE varchar(2000);
```

## Testing story

I manually uploaded this file to production-daemon, and manually inserted the new function in to "dashboard/lib/pd/foorm/submission_analytics_parser.rb".

I ran tests of the script, iterating on issues. This test run appears like it will be successful: https://codedotorg.slack.com/archives/C3S8AJY9K/p1717176681278969

## Deployment strategy

Before merge, update the "crontab.yml.erb" to use this script instead of the old script. Or rename this script to overwrite the old one.

## Follow-up work

Configure lifecycle rules on the S3 bucket prefix where these temp csv files are being saved. They are NOT being deleted, to support troubleshooting of this process. They should live no longer than 1 week, though we could choose to delete them as we go, a line of code that we could comment out if we wish to troubleshoot.

## Privacy

Submissions may contain PII, and are destined for the dashboard_pii schema in Redshift. The S3 temp files and the Redshift temp table will contain PII and are only useful for troubleshooting this process. If these are a concern, the script should be modified to clean them up as it goes - but I think there is value in keeping them for a short time to troubleshoot.

## Security

S3 bucket and redshift pii shema are protected.

## PR Checklist:

- [ ] Code is well-commented


dashboard/test/lib/pd/foorm/submission_analytics_parser_test.rb
```diff
@@ -5,6 +5,16 @@ class SubmissionAnalyticsParserTest < ActiveSupport::TestCase
...
 
+    test 'reshape_all_submissions_into_csv returns a CSV string' do
+      csv = SubmissionAnalyticsParser.reshape_all_submissions_into_csv
+      assert csv.is_a? String
+    end
+
+    test 'reshape submissions_batch_into_csv returns a CSV string' do
+      csv = SubmissionAnalyticsParser.reshape_submissions_batch_into_csv(0, 100)
+      assert csv.is_a? String
+    end
```
cat5inthecradle: I believe these functions can and should be tested, but as of now they are not.

bin/cron/process_foorm_data
```diff
@@ -75,7 +75,7 @@ def write_submissions_to_redshift(client, arn)
       submission_id    int,
       item_name        varchar,
       matrix_item_name varchar,
-      response_value   varchar,
+      response_value   varchar(2000),
```
cat5inthecradle, line 101: This is the simple fix. If we do just this, we could keep using the `process_foorm_data` script and skip the rest of this PR.

jordan-springer: is there any more benefit in making it `varchar(max)`? do we know any constraints on the response_value field?

jordan-springer (commented):
looks good!

bin/cron/process_foorm_data
cat5inthecradle: I think `max` is overkill at ~65536 and while `max` vs `n` should be the same from a storage perspective, for performance `max` is slower. [AWS advises against defaulting to the max column size](https://docs.aws.amazon.com/redshift/latest/dg/c_best-practices-smallest-column-size.html).

2000 is a little over 3x bigger than the size of the data that caused the problem. As far as I can tell, the maximum _possible_ size is ~65536, similar to `max`, but that's because we defaulted to the largest possible size in our mysql migration too.

cat5inthecradle:
Closing in favor of one of these:

- Instrument the old script https://github.com/code-dot-org/code-dot-org/pull/59097
- Fully batch the process https://github.com/code-dot-org/code-dot-org/pull/59098

PR #52069, opened 2023-05-24
CPA LockoutPanel UI

wilkie:
This adds `LockoutPanel` and the associated route, images, haml page, and JS initialization in our pipeline.

There is no behavior, just the UI, as per our ticket. Gluing this UI and the API/data model together is a next step.

But I did go ahead and add a route for the page for convenience.

Added each respective thingy to the `sessions` namespace... not sure if that's a good place. Felt right and better than creating something new or have it float around with everything else.

### Behaviors of the UI:

* Not currently masked behind the CPA experiment flag... not necessary: not in any common path or updating any already used UI.
* `/lockout` route is a placeholder... any failed authentication on parental permission should just redirect there.
* Information is passed by adding data attributes on the containing div currently. Easy to change.
* Submit button is disabled until a valid email is typed in.
* Resend button (styled as a link) replaces the email with the pending email and submits.
* Delete by date is rendered as UTC so there are no strange timezone shenanigans when rendering it. Let's be sure to generate a UTC delete by date.
* Last request date is rendered in the current locale / timezone. We should still store it as UTC.

### New Strings for i18n:

`sessionLockoutNewAccountHeader`: "Just one more thing!"
`sessionLockoutPendingHeader`: "Thanks! We've contacted your parent."
`sessionLockoutPrompt`: "We need your parent to approve your account before you can get started. Please supply us with your parent’s email address so they can grant you permission."
`sessionLockoutPendingPrompt`: "We sent an email to {pendingEmail}. Didn’t receive anything? Update your parent’s email below or send another request."
`sessionLockoutNote`: "Note: Your account will be deleted if we do not receive your parent’s permission by {deleteDate}."
`sessionLockoutParentEmailField`: "Parent Email:"
`sessionLockoutParentStatusField`: "Parent Permission:"
`sessionLockoutSubmit`: "Send permission request"
`sessionLockoutStatusPending`: "Pending"
`sessionLockoutStatusNotSubmitted`: "Not Submitted"
`sessionLockoutResendEmail`: "Resend permission email"
`sessionLockoutUpdateSubmit`: "Update and send"
`signOutButton`: "Sign out",

## Links

- jira ticket: [P20-61](https://codedotorg.atlassian.net/browse/P20-61)

## Testing story

Manual testing of the basic boilorplate behaviors. Then, contributing one Storybook with two templates:

### New Accounts

image

### When there is a pending email request:

image

### New Accounts (right-to-left)

image

### Pending email (right-to-left)

image

## Deployment strategy

[...]

PR #58929, opened 2024-05-27
P20-138: Add CAP Lock-out flow for existing accounts

artem-vavilov:
## Links
- Jira ticket: [P20-138](https://codedotorg.atlassian.net/browse/P20-138)
- Figma: [2. Lock-out Flow (Post-July 1st 2023) Existing Accounts](https://www.figma.com/design/wR0Myx8T2dHMaeeb1GAabb/CPA-Flows?node-id=815-1518&t=i2PIn6xnnN6NaGC2-0)
- Related PR: https://github.com/code-dot-org/code-dot-org/pull/52069


dashboard/lib/policies/child_account.rb
```diff
@@ -94,6 +94,14 @@ def self.lockout_date(user)
     state_policy(user).try(:[], :lockout_date)
   end
 
+  # Checks if the user is locked out due to non-compliance with CAP.
+  def self.locked_out?(user)
+    user_lockout_date = lockout_date(user)
```
artem-vavilov, line 108: https://github.com/code-dot-org/code-dot-org/blob/55b1a5afdc90eb4606535921df91a8eb2291cfe5/dashboard/lib/policies/child_account.rb#L91-L95

dashboard/lib/policies/child_account.rb
```diff
@@ -94,6 +94,14 @@ def self.lockout_date(user)
...
+    return false unless user_lockout_date
+
+    DateTime.now >= user_lockout_date || !user_predates_policy?(user)
```
wilkie, line 111: This is the same as:

```ruby
# Not locked out if the user account predates the policy and is in their grace period
return false if user_predates_policy?(user) && DateTime.now < user_lockout_date

# Otherwise, the account is subject to the policy, so we assume the account is locked out
return true
```

Where it might be easier to follow the logic if it is a list of exceptions since you have a `return false` earlier and the exceptions are the important parts to understand on a future reading.

The logic does hold though. It looks good overall.

dashboard/test/controllers/sessions_controller_test.rb
```diff
@@ -352,4 +321,92 @@ class LtiAccountLinkingSignInPageTest < ActionDispatch::IntegrationTest
...
+    end
+
+    it 'assigns @pending_email with empty string' do
+      get :lockout
+      _(assigns[:pending_email]).must_equal ''
+    end
+
+    it 'assigns @disallowed_email with empty string' do
+      get :lockout
+      _(assigns[:disallowed_email]).must_equal ''
```
wilkie, line 349: I think you'll be the first to actually use minitest expectations and minitest/spec in the codebase in spite of it being an option for a long while. 😅

I'm happy with this.

wilkie: I know the `_` can be opinionated... I wonder if anybody has an opinion on `_(...)` vs the `expect(...)` alias, but I like the `_` better. Easier to scan for and I already easily know that it is representing an expectation and an assert. All good to me!

wilkie (approved):
Looks good. I have one comment that you can elect to address to make the logic in the `locked_out?` method a little easier to understand by being a little more verbose there. That logic is correct, though, so it is technically optional to address.

I really like seeing better use of Minitest and it would be nice to have many of our other test files conform to the `describe` spec syntax just like you do with a block per method/route. Nice use of the `around` block for handling timestamps.

Great job!

dashboard/lib/policies/child_account.rb
artem-vavilov: @dayne @wilkie I believe it makes sense to check `user_predates_policy?` (users created before `July 1, 2023`) only before `July 1, 2024`, since all users created after that date should be locked out anyway based on legal requirements:
https://github.com/code-dot-org/code-dot-org/blob/2ee8ac6ac31dfb9f19948d792fe0b0f1baa117e4/dashboard/app/controllers/application_controller.rb#L354
https://github.com/code-dot-org/code-dot-org/blob/55b1a5afdc90eb4606535921df91a8eb2291cfe5/dashboard/lib/policies/child_account.rb#L79-L89

artem-vavilov: @daynew Shouldn't we also update the method `user_predates_policy?` to check users created before `July 1, 2024` instead of `July 1, 2023` now?

daynew: I think we should leave the `user_predates_policy?` logic for now. I think we will end up deprecating this function once we implement the "Grace Period" feature.

daynew (approved):
Test seem to match the behavior we talked about! Thank you for doing all this refactoring.

PR #58930, opened 2024-05-28
Music: increase timeline rows

breville:
Previously, the timeline showed a maximum of 12½ rows.  Now, it can show up to 27½.

### before

Screenshot 2024-05-27 at 10 45 50 PM
Screenshot 2024-05-27 at 10 45 55 PM
Screenshot 2024-05-27 at 10 45 58 PM

### after

Screenshot 2024-05-27 at 10 44 35 PM
Screenshot 2024-05-27 at 10 44 39 PM
Screenshot 2024-05-27 at 10 44 43 PM
Screenshot 2024-05-27 at 10 44 49 PM
Screenshot 2024-05-27 at 10 44 53 PM
Screenshot 2024-05-27 at 10 44 56 PM
Screenshot 2024-05-27 at 10 44 59 PM

### DIFF
--- a/apps/src/music/views/Timeline.tsx
+++ b/apps/src/music/views/Timeline.tsx
@@ -14,6 +14,10 @@ import {
 import {useMusicSelector} from './types';
 import usePlaybackUpdate from './hooks/usePlaybackUpdate';
 
+// The height of the primary timeline area for drawing events.  This is the height of each measure's
+// vertical bar.
+const timelineHeight = 130;
+// The width of one measure.
 const barWidth = 60;
 // Leave some vertical space between each event block.
 const eventVerticalSpace = 2;
@@ -22,7 +26,10 @@ const paddingOffset = 10;
 // Start scrolling the playhead when it's more than this percentage of the way across the timeline area.
 const playheadScrollThreshold = 0.75;
 
-const getEventHeight = (numUniqueRows: number, availableHeight = 130) => {
+const getEventHeight = (
+  numUniqueRows: number,
+  availableHeight = timelineHeight
+) => {
   // While we might not actually have this many rows to show,
   // we will limit each row's height to the size that would allow
   // this many to be shown at once.
--- a/apps/src/music/views/Timeline.tsx
+++ b/apps/src/music/views/Timeline.tsx
@@ -22,13 +22,13 @@ const paddingOffset = 10;
 // Start scrolling the playhead when it's more than this percentage of the way across the timeline area.
 const playheadScrollThreshold = 0.75;
 
-const getEventHeight = (numUniqueRows: number, availableHeight = 110) => {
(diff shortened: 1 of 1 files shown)


apps/src/music/views/Timeline.tsx
```diff
@@ -22,13 +22,13 @@ const paddingOffset = 10;
...
 
-const getEventHeight = (numUniqueRows: number, availableHeight = 110) => {
+const getEventHeight = (numUniqueRows: number, availableHeight = 130) => {
   // While we might not actually have this many rows to show,
   // we will limit each row's height to the size that would allow
   // this many to be shown at once.
   const minVisible = 5;
 
-  const maxVisible = 10;
+  const maxVisible = 26;
```
fisher-alice, line 38: Has there been discussion about making timeline panel scrollable for projects with > `maxVisible` unique sounds?

apps/src/music/views/Timeline.tsx
```diff
@@ -22,13 +22,13 @@ const paddingOffset = 10;
 // Start scrolling the playhead when it's more than this percentage of the way across the timeline area.
 const playheadScrollThreshold = 0.75;
 
-const getEventHeight = (numUniqueRows: number, availableHeight = 110) => {
+const getEventHeight = (numUniqueRows: number, availableHeight = 130) => {
```
fisher-alice: Nit: use a constant.

fisher-alice (approved):
LGTM!

apps/src/music/views/Timeline.tsx
breville: I've tried it from time to time, but it would require a deeper dive.  

Some issues to consider include:

- calculating the timeline height in this class rather than in `TimelineSimple2Events`.
- testing across all browsers to see how it feels; some browsers lock the primaxy axis once you start scrolling in one, which hasn't always matched user intent.
- deciding whether we want to be consistent with Blockly's two-axis scrolling which allows the user to drag the workspace as well as scroll it natively.
- deciding on the right experience when playback begins and we auto-scroll the timeline horizontally.  It might be better to intelligently adjust vertical scroll in that case rather than, or in addition to, allowing the user to vertically scroll.
- whether users are likely to miss events that appear in corners; an advantage of scrolling in one axis only is that a full sweep back and forth can reveal everything.  

That said, this is best explored practically, so I might separately prepare a small URL parameter that lets us start playing with vertical scrolling.

apps/src/music/views/Timeline.tsx
breville: Good call; done.

apps/src/music/views/Timeline.tsx
fisher-alice: Interesting - thanks for sharing the many complex considerations!

PR #58932, opened 2024-05-28
Bump ios versions [test all browsers]

bethanyaconnor:
Bumps iPhone ~and iPad~ version from 15.4 and 14.5 respectively to 16.2.

We are tracking whether this will help prevent flaky tests that we are seeing more frequently recently on IOS versions.

Note: this also bumps our Appium versions used by saucelabs and requires `appium:` prepended to some config settings.

UPDATE: Removing iPad for now, while we work around a safari quirk where the sidebar opens by default.

## Links
Example flaky test jira ticket: https://codedotorg.atlassian.net/browse/TEACH-1102
https://codedotorg.atlassian.net/browse/TEACH-1128

## PR Checklist:

- [ ] Code is well-commented


cat5inthecradle:
The P95 cutoff for iOS users is version 15, which is still receiving security updates from apple as recently as earlier this month.

The exact p95 threshold is 15.6, but there's a cluster of users on 15.5 before numbers drop off; that's an argument for staying on 15.4/5. **I would recommend updating no farther than 15.8.2 until either A) Apple stops releasing updates for iOS 15 or B) we start running into iOS 15 specific issues.**

And a completely separate question: **should we ALSO test against the latest iOS, as that's the bulk of our users?**

cat5inthecradle:
Additionally, [it's expected that we'll see iOS 18 released in September 2024](https://www.macrumors.com/roundup/ios-18/), which is shortly after the school year starts. We strive to keep our requirements consistent through the school year, and we're already warning people that v15 will become our new minimum in July - but since v18 will be right on the heels of that change, perhaps we should preemptively set our minimum version to v16 in July in anticipation that they'll drop support for v15

(ipads are also such a small % of active users and much easier to apply mid-school-year updates on, that perhaps we don't need to worry about bumping it in September)

cat5inthecradle (approved):
Approving this because I think I've convinced myself that we're okay to start testing on 16 in preparation for a September release of iOS 18, which will likely drop Apple support for 15.

cat5inthecradle (approved):
Edited the description to reflect the removal of the iPad update.

PR #58940, opened 2024-05-28
Add more folders to eslint import order rule

lfryemason:

## PR Checklist:

- [ ] Code is well-commented


apps/src/applab/DesignModeBox.jsx
```diff
@@ -1,8 +1,13 @@
 import PropTypes from 'prop-types';
 import React from 'react';
+
 import color from '../util/color';
+
+// DesignToolbox must be imported before DesignProperties to avoid circular dependency
```
Hamms: Ideal comment, thank you!

apps/src/applab/DesignToolbox.jsx
```diff
@@ -1,8 +1,13 @@
 import PropTypes from 'prop-types';
 import React from 'react';
-import DesignToolboxElement from './DesignToolboxElement';
+
 import applabMsg from '@cdo/applab/locale';
+
+// DesignToolboxElement must be imported before ThemeDropdown to avoid circular dependency
```
Hamms: :+1:

apps/src/applab/applab.js
```diff
@@ -8,40 +8,52 @@ import $ from 'jquery';
...
-import {Provider} from 'react-redux';
-import {getStore} from '../redux';
 import {actions, reducers} from './redux/applab';
-import {add as addWatcher} from '../redux/watchedExpressions';
 import {changeScreen} from './redux/screens';
-import * as applabConstants from './constants';
+
 const {ApplabInterfaceMode} = applabConstants;
+
+// Disabling import order in order to import ApplabInterfaceMode
```
Hamms: Could you add a clue here about what specifically goes wrong, or how one could reproduce locally? I'm presuming we get an error of some kind during the build, but it would be nice to spell it out explicitly.

apps/src/applab/designMode.js
```diff
@@ -3,30 +3,35 @@ import 'jquery-ui/ui/effects/effect-drop';
...
 import ReactDOM from 'react-dom';
 import {Provider} from 'react-redux';
 import RGBColor from 'rgbcolor';
-import objectFitImages from 'object-fit-images';
+
+import firehoseClient from '../lib/util/firehose';
+import logToCloud from '../logToCloud';
+
+import sanitizeHtml from './sanitizeHtml';
+// Disabling import order because build fails with circular dependencies
```
Hamms: :+1:

apps/src/bounce/bounce.js
```diff
@@ -127,7 +135,7 @@ var initWallMap = function () {
 /**
  * PIDs of async tasks currently executing.
  */
-import * as timeoutList from '../lib/util/timeoutList';
+import * as timeoutList from '../lib/util/timeoutList'; // eslint-disable-line import/order
```
Hamms, line 139: Why are we disabling this one?

apps/src/bounce/BounceVisualizationColumn.jsx
```diff
@@ -1,12 +1,19 @@
 import React from 'react';
+
 var msg = require('@cdo/locale');
 
+// Disabling import order in order to add require statements first
```
Hamms, line 5: For this and the other "in order to add require statements" in this file, can you clarify why it's required for the `require` statements to come first, and how it breaks if they don't? I'd also appreciate some clarification on why it might be safe, and what the referenced investigation would look like.

Hamms (commented):
Mostly looks good, but there are a couple of comments I'd like to see expanded on a bit

apps/src/bounce/bounce.js
lfryemason: Because it seems to be very specifically placed. I think the added context of having it placed here is probably better than having it up with the other imports.

apps/src/calc/calc.js
```diff
@@ -22,28 +22,40 @@ var Calc = module.exports;
...
+
+var CalcVisualizationColumn = require('./CalcVisualizationColumn');
+var Equation = require('./equation');
+var EquationSet = require('./equationSet');
+var ExpressionNode = require('./expressionNode');
+var InputIterator = require('./inputIterator');
+var calcMsg = require('./locale');
+
+// Disabling import order in order to add require statements first.
+// This might be safe to remove but needs investigation.
```
Hamms: Could you add the same explanation to this block as you did to the others above? Thanks!

Hamms (approved):
One more comment I'd like to see get updated, then LGTM!

PR #58941, opened 2024-05-28
[Play Lab] Start Breaking CS in Algebra

mikeharv:
Update: This branch now includes the removal of all unit/integration tests related to Calc, Eval, and Algebra Game (Play Lab with functional blocks). See additional PR for changes not detailed below.
- https://github.com/code-dot-org/code-dot-org/pull/58954

Now that CS in Algebra has been fully deprecated, we can start removing code that made it work. 🎉 

This first branch removes a couple of calls related `installStringPicker` and `hasUnfilledFunctionalInput` which cause a crash when using Google Blockly. 
image

The functions are defined in the CDO Blockly repo. As a general rule of thumb, any Play Lab (ie. Studio) code that references functional block is going to relate exclusively to calc and eval and can likely be cleaned up. In this case, removing these lines is sufficient to get all levels of the Play Lab Hour of Code progression to load without error. (That being said, there are still code generation errors if you try to actually _run_ the code.)

Still, it's exciting to see that we can some stuff working by just removing old logic that we don't actually need to migrate.

image

No regressions were found in Play Lab using CDO Blockly.

Note that this branch is based upon this PR:
- https://github.com/code-dot-org/code-dot-org/pull/58941

Without the changes in the base branch, we'd still be crashing due to `interpolateMsg` being undefined.

## PR Checklist:

- [ ] Code is well-commented


molly-moen (approved):
🧹

PR #58946, opened 2024-05-28
Clean up files manually added to apps build by Storybook deploy

bencodeorg:
[Stephen noted](https://codedotorg.slack.com/archives/C0T0PNTM3/p1716411724975939) that the difference between our staging and test builds that produce the "different package for same contents" that often blocks staging/test builds is a result of a couple of files being manually added to our staging build as part of our Storybook deploy process. This change cleans up that diff, which should reduce the frequency of these errors.

I pursued a "cleaner" approach of separating out and running the Storybook build after we do the S3 upload, but I felt it ended up a bit more complicated than it was worth. Here's the description for posterity:

The "upload to S3" step is actually a part of (really a prerequisite for) the "dashboard" build step ([the "package" dependency listed here](https://github.com/code-dot-org/code-dot-org/blob/0d63b81a4963a5fab567576dc6a1d87d86337937/lib/rake/build.rake#L48)), rather than the "apps" build step. I didn't feel confident messing with that setup (and it might make sense if you think of the apps build itself as "standalone" and the need for it to be on S3 as a requirement for our Rails app). As a result, I wanted to put the Storybook step after the dashboard build step. To do this, I needed to separate out the logic that determines when we build apps (ie, if there have been changes in the apps directory since the last build), and apply this to both the Storybook step and the main apps build step. [A rough draft of this is here for future reference](https://github.com/code-dot-org/code-dot-org/pull/58925).

## Testing story

I tested this script manually locally and saw that the two files no longer appeared in my build directory.

### DIFF
--- a/apps/script/deploy-storybook.sh
+++ b/apps/script/deploy-storybook.sh
@@ -67,5 +67,9 @@ echo "cleaning up!"
 # clean up after ourselves
 popd
 rm -rf $DIR_TO_DEPLOY
+# Remove these files manually copied over above
+# so they are not retained in our staging apps build package.
+rm ./build/package/css/application.css
+rm ./build/package/css/font-awesome.css
 
 echo "if everything worked, your changes should be up on https://code-dot-org.github.io/cdo-styleguide/"
--- a/lib/rake/build.rake
+++ b/lib/rake/build.rake
@@ -11,10 +11,12 @@ namespace :build do
   desc 'Builds apps.'
   timed_task_with_logging :apps do
     Dir.chdir(apps_dir) do
+      # Only rebuild if any of the apps_build_trigger_paths have changed since last build.
       commit_hash = apps_dir('build/commit_hash')
+      if !RakeUtils.git_staged_changes?(*apps_build_trigger_paths) &&
+          File.exist?(commit_hash) &&
+          File.read(commit_hash) == calculate_apps_commit_hash
 
-      # This will never be hit in current configuration
-      if apps_unchanged?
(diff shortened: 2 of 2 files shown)


apps/script/deploy-storybook.sh
```diff
@@ -67,5 +67,9 @@ echo "cleaning up!"
 # clean up after ourselves
 popd
 rm -rf $DIR_TO_DEPLOY
+# Remove these files manually copied over above
+# so they are not retained in our staging apps build package.
+rm ./build/package/css/application.css
+rm ./build/package/css/font-awesome.css
```
stephenliang, line 73: The only concern is that technically production will no longer have `application.css` and `font-awesome.css` (as they used to be added by the storybook script) and I wonder if we use them today. Otherwise it looks good

stephenliang (approved):

apps/script/deploy-storybook.sh
bencodeorg: Script tags for these files are "hard coded" into HAML that (I think?) is loaded on every studio.code.org page, so I don't expect any issues with these files being removed...or am I misunderstanding your concern?

https://github.com/code-dot-org/code-dot-org/blob/ad1fa5624d940f6f3ca0d18aff73748e3f3ec0b8/dashboard/app/views/layouts/application.html.haml#L23-L24

stephenliang: I think we can do an eyes test when it hits DTT to confirm it doesn't change anything. I believe these files are technically out there in the bundle but may not actually be referenced. I think just making sure that storybook and combined with eyes will be sufficient.

PR #58951, opened 2024-05-29
Regen curriculum test fixtures

davidsbailey:
Regenerate test fixtures via `./dashboard/scripts/generate_fixtures.rb`, to unblock some other work.

unit test fixes fall into two categories:
* some hard-coded script ids were changed, so I updated the unit test code not to depend on these ids
* addition of `course_offering.marketing_initiative` broke some assumptions in QuickAssignHelperTest, which I'm addressing by removing all test fixtures prior to those tests (these are restored after each test run)

## Testing story

passing drone run


Hamms (approved):
This makes sense to me, but is it worth looking into how much work it would take to refactor the relevant tests to use factories instead of fixtures? It would be nice to use this as an opportunity to reduce the amount of fixture-dependent tests we rely on, in service of our long-term goal of deprecating fixtures entirely.

davidsbailey:
Since I'm already in two levels deep of cleanup here, I'm gonna go ahead and merge this PR as "making things better" (now less fragile to regenerate fixtures) and address the topic of follow-up work separately.

> This makes sense to me, but is it worth looking into how much work it would take to refactor the relevant tests to use factories instead of fixtures? It would be nice to use this as an opportunity to reduce the amount of fixture-dependent tests we rely on, in service of our long-term goal of deprecating fixtures entirely.

I think so, but I also think we want to rip out the feature that these tests are covering. We shouldn't need to redirect from urls like /s/1 any more. I noticed we do have some traffic on these urls, and believe that google search results are the culprit:
Screenshot 2024-06-03 at 1 02 18 PM

perhaps the very next step should be to get these urls out of google search results?

davidsbailey:
> perhaps the very next step should be to get these urls out of google search results?

the top result for how to remove a url from google search results is to make it return 404... so we could also just move straight to ripping out the feature of handling requests to /s/ by script id

davidsbailey:
https://github.com/code-dot-org/code-dot-org/pull/59038

PR #58958, opened 2024-05-29
[Python Lab] Improve error messages

molly-moen:
Pyodide's error messages are confusing to an end user because they refer to an internal stack trace in pyodide. For example, for an indentation error in `main.py` you see:
```
Traceback (most recent call last):
  File "/lib/python311.zip/_pyodide/_base.py", line 573, in eval_code_async
    await CodeRunner(
          ^^^^^^^^^^^
  File "/lib/python311.zip/_pyodide/_base.py", line 267, in __init__
    self.ast = next(self._gen)
               ^^^^^^^^^^^^^^^
  File "/lib/python311.zip/_pyodide/_base.py", line 145, in _parse_and_compile_gen
    mod = compile(source, filename, mode, flags | ast.PyCF_ONLY_AST)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<exec>", line 30
    print("hi")
    ^^^^^
IndentationError: expected an indented block after 'if' statement on line 29
```
`<exec>` refers to `main.py`. In addition, the line numbers look wrong to students because we prepend patches to their code internally.

To fix this, we cut out the first part of the error message, adjust the line numbers, and replace the reference to `<exec>` with a reference to `main.py`. The resulting error is
```
main.py, line 4
    print("hi")
    ^^^^^
IndentationError: expected an indented block after 'if' statement on line 3
```

If there is an error in a non-main file, we handle that too. The original message prefixes all our files with `/home/pyodide` (the folder pyodide runs from in the virtual filesystem), so we remove that prefix. For example, an error in a nested file looks like
```
main.py, line 1
test1.py, line 1
    import not_a_module
ModuleNotFoundError: No module named 'not_a_module'
```

Finally, if a user tries to import a package that is supported by pyodide but we aren't hosting, pyodide gives them a message about trying to import it. We don't want users to do that, so we replace that message with a simpler message:
```
ModuleNotFoundError: The module 'bleach' is not supported in Python Lab.
```

As part of this I also improved the styling of console errors (previously they were the same as system messages but prefixed with `Error:`. For now I made them `lightest_red`, since that seemed to be better contrast then our other reds:
Screenshot 2024-05-29 at 9 27 28 AM

## Links

- jira ticket: [CT-502](https://codedotorg.atlassian.net/browse/CT-502)

## Testing story
Tested locally with a variety of python errors that I sourced from chat gpt. There are probably error messages I missed, so we should file follow-up tasks if we find wonky error messages going forward. This should still work after #58924 is in, because I adjust the main.py error line based on the `ALL_PATCHES` constants.

## PR Checklist:

- [ ] Code is well-commented


snickell (approved):
👍

I don't _love_ munging the error string - for starters it could be potentially brittle as python versions change (it probably won't be, but..... it still makes me a little nervous!). But I think its worth it for dev experience, and after thinking for a bit, I'm not thinking of a better way to do this.

There's perhaps some sort of potentially principled approach on the python side that involves munging `__file__` variables inside imported modules.... but it would require a custom MetaPathFinder (https://docs.python.org/3/library/importlib.html#importlib.abc.MetaPathFinder), and for all that complexity..... it might actually increase the risk since its starting to do weird things!

All things considered, I think this is as good an approach as we're gonna get.

apps/src/pythonlab/pythonHelpers/messageHelpers.ts
```diff
@@ -0,0 +1,92 @@
...
+  ) {
+    mainErrorLine++;
+  }
+  if (mainErrorLine >= errorLines.length) {
+    // If we never find the main.py error, return the entire message.
+    return errorMessage;
+  }
+  const mainLineNumber = parseInt(
+    errorLines[mainErrorLine].match(/line (\d+)/)![1]
+  );
```
mikeharv, line 45: Is it possible that a line number won't be found?

apps/src/pythonlab/pythonHelpers/messageHelpers.ts
```diff
@@ -0,0 +1,92 @@
...
+    !mainErrorRegex.test(errorLines[mainErrorLine])
+  ) {
+    mainErrorLine++;
+  }
+  if (mainErrorLine >= errorLines.length) {
+    // If we never find the main.py error, return the entire message.
+    return errorMessage;
+  }
+  const mainLineNumber = parseInt(
+    errorLines[mainErrorLine].match(/line (\d+)/)![1]
```
mikeharv: Could we re-use `mainErrorLineRegex` for the regex here?

mikeharv (commented):
Nice! I had been curious how/when we were going to tackle this, and specifically how hard it was going to be to figure out the correct/adjusted line numbers to show the student.

mikeharv (approved):

apps/src/pythonlab/pythonHelpers/messageHelpers.ts
molly-moen: It's not because we've found a main error line, which matches `/File "<exec>", line \d+.*/`

apps/src/pythonlab/pythonHelpers/messageHelpers.ts
molly-moen: Yeah I'll update it to include a capture group and re-use it. (edit: thought you were referring to `mainErrorRegex`, I'll just reuse `mainErrorLineRegex`)

PR #58960, opened 2024-05-29
[Python Lab] Add support for scipy

molly-moen:
Add support for scipy in our self-hosted version of pyodide. Scipy required us to add a `.zip` file for its `openblas` dependency, so I updated our `.gitattributes` to include any `zip` files in the `pyodide` folder in LFS. We have some other zip files in the repo that are not in LFS so I didn't include zip more broadly.

## Links

- jira ticket: [https://codedotorg.atlassian.net/browse/CT-447](https://codedotorg.atlassian.net/browse/CT-447) This ticket was for scipy, math and pandas. I already added pandas, and math is included by default.

## Testing story
Tested locally that you can now import and use scipy in Python Lab.

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -22,6 +22,9 @@ pegasus/sites.v3/** filter=lfs diff=lfs merge=lfs -text
 pegasus/sites/** filter=lfs diff=lfs merge=lfs -text
 shared/images/** filter=lfs diff=lfs merge=lfs -text
 
+# INCLUDE any .zip files that appear in the pyodide folder.
+apps/lib/pyodide/*.zip filter=lfs diff=lfs merge=lfs -text
+
 # HOW TO EXCLUDE A FILE/FILES FROM GIT LFS:
 #
 # 1. Add a new exclusion pattern in the immediately following section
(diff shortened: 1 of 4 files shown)


snickell (changes requested):
I'd much rather see a non-directory-specific `*.zip` pattern added in .gitignore right after `*.whl`. A goal here is to keep the .gitattributes as simple as possible, not have folder by folder rules. There's no cases where a zip file (almost random binary by definition) shouldn't be treated as a binary file and put in LFS.

Ideally, .zip would have already been in the patterns at the time I converted (so they weren't in the history), but that ship has sailed.

If you agree, directions for adding a pattern to LFS are in the `.gitignore` file.

molly-moen:
@snickell updated! I wasn't sure if it was ok to add an inclusion when we already had some `.zip` files, good to know it's fine to do so.

snickell:
Thanks for being careful! yeah, moving things into LFS is very take-backable and generally safe, its the other direction that has the pitfalls.

I don't see any new zip files being added by the pattern change.... did you do the renormalize command and nothing extra showed up?

molly-moen:
> Thanks for being careful! yeah, moving things into LFS is very take-backable and generally safe, its the other direction that has the pitfalls.
> 
> I don't see any new zip files being added by the pattern change.... did you do the renormalize command and nothing extra showed up?

@snickell I was moving too fast and didn't renormalize, but then I ran the command and nothing got added, I think because all the `*.zip` files are already covered by `pegasus` inclusions.

snickell (approved):
Oh yeah, that totally makes sense. TY pegasus pattern for already having us covered 😅

LGTM

PR #58964, opened 2024-05-29
Revert "Revert "Write UI tests for My Professional Learning page""

kelbyhawn:
Updates the Self-paced test to hopefully fix test flakiness. I think the flakiness was caused by there not being enough time to register that a course was started before going back to the My PL page. This attempts to fix this by adding an extra step to wait for an additional level to load on the self-paced course.

Reverts code-dot-org/code-dot-org#58962

## Links
Slack convo: [here](https://codedotorg.slack.com/archives/C0T0PNTM3/p1717009073054799)
Sauce labs passing test: [here](https://app.saucelabs.com/tests/60e61c34b98c45f89172855022ca3b22#57)

## Testing story
Tested locally on SauceLabs

### DIFF
--- a/dashboard/test/ui/features/teacher_tools/pl_landing_page.feature
+++ b/dashboard/test/ui/features/teacher_tools/pl_landing_page.feature
@@ -117,7 +117,7 @@ Feature: Professional Learning landing page
     Then I am on "http://studio.code.org/s/alltheselfpacedplthings/lessons/1/levels/3"
     And I wait until element "a:contains(Submit)" is visible
     Then I click selector "a:contains(Submit)"
-    Then I am on "http://studio.code.org/s/alltheselfpacedplthings/lessons/1/levels/4"
+    Then I wait until I am on "http://studio.code.org/s/alltheselfpacedplthings/lessons/1/levels/4"
     And I wait until element "a:contains(Submit)" is visible
     Then I am on "http://studio.code.org/my-professional-learning"
 
--- a/dashboard/test/ui/features/teacher_tools/pl_landing_page.feature
+++ b/dashboard/test/ui/features/teacher_tools/pl_landing_page.feature
@@ -117,6 +117,8 @@ Feature: Professional Learning landing page
     Then I am on "http://studio.code.org/s/alltheselfpacedplthings/lessons/1/levels/3"
     And I wait until element "a:contains(Submit)" is visible
     Then I click selector "a:contains(Submit)"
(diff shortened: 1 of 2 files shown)


dashboard/test/ui/features/teacher_tools/pl_landing_page.feature
```diff
@@ -117,6 +117,8 @@ Feature: Professional Learning landing page
     Then I am on "http://studio.code.org/s/alltheselfpacedplthings/lessons/1/levels/3"
     And I wait until element "a:contains(Submit)" is visible
     Then I click selector "a:contains(Submit)"
+    Then I am on "http://studio.code.org/s/alltheselfpacedplthings/lessons/1/levels/4"
```
breville: I think this does a navigation to that URL?  

If we intend to test that the above line's submission takes us to this page, I think this skips that aspect of testing, because it forces the navigation to this URL, rather than making sure the above submission took us there.

I think maybe we want `I wait until I am on`?

kelbyhawn: Just updated! Thanks for the `I wait until I am on` tip ✨

bethanyaconnor (approved):
Were you able to reproduce the flakiness locally or is this fix speculative? I'm approving either way -- just curious!

breville:
> Were you able to reproduce the flakiness locally or is this fix speculative?

It was purely an observation and I'd encourage folks to double-check my assertion.

kelbyhawn:
> Were you able to reproduce the flakiness locally or is this fix speculative? I'm approving either way -- just curious!

Bethany and I just talked about this in check-in, but closing the loop here. 

I didn't try to reproduce locally because I didn't know how until just now (by adding `[test all browsers]` to the commit message 😄). I'm going to merge since it passed drone already, and if it continues to be flaky I'll do more investigation.

breville:
> I didn't try to reproduce locally because I didn't know how until just now (by adding [test all browsers] to the commit message 😄). 

That gives a directive to drone.  You should be able to test (somewhat) locally by setting up a SauceLabs tunnel and running the correct `runner.rb` command...

kelbyhawn:
> > I didn't try to reproduce locally because I didn't know how until just now (by adding [test all browsers] to the commit message 😄).
> 
> That gives a directive to drone. You should be able to test (somewhat) locally by setting up a SauceLabs tunnel and running the correct `runner.rb` command...

Yep I know. Bethany was asking about testing the flakiness of the test (not the test itself in SauceLabs which I already did). We discussed using the drone directive in our check-in to possibly test for drone flakiness before merging.

PR #58967, opened 2024-05-29
Fix for local codebridge project loading

molly-moen:
If you've made edits to a project on codebridge and the page hot reloads, the project will be overwritten by the initial sources from your initial page load. This seems to be because we aren't reloading the level data but are reloading the component, which ends up resetting the initial sources. The fix for this is to explicitly compare the channel id to the last seen channel id, and only update the project if the channel id has changed (or we haven't seen a channel id yet).

## Testing story
I noticed this issue while doing local development, my project changes kept disappearing. With this fix, I can change the project, then make a change that triggers hot reloading, and the change is still there.

## PR Checklist:

- [ ] Code is well-commented


thomasoniii (approved):
lgtm 🚀

PR #58968, opened 2024-05-29
[ProgressV2M] Add buttons to show time spent and last updated to student names

lfryemason:
Adds a new feature flag to gate all metadata (time spent and last updated) work behind:
`progress-v2-metadata-enabled`

Adds a caret icon to the left of each student name (when feature flag is enabled). Clicking on the student name should switch from a right caret to a down caret. This also adds the expanded student ID to the redux table to be used by other components.

New flag set to false:
image

New flag set to true:
https://github.com/code-dot-org/code-dot-org/assets/25193259/334b4ddd-ae9c-493e-9fd5-6ab3290393ed

## Links
[TEACH-1070](https://codedotorg.atlassian.net/browse/TEACH-1070)

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/templates/sectionProgress/sectionProgressRedux.js
+++ b/apps/src/templates/sectionProgress/sectionProgressRedux.js
@@ -1,3 +1,5 @@
+import _ from 'lodash';
+
 import {SET_SCRIPT} from '@cdo/apps/redux/unitSelectionRedux';
 
 import firehoseClient from '../../lib/util/firehose';
@@ -127,10 +129,10 @@ export default function sectionProgress(state = initialState, action) {
   if (action.type === EXPAND_METADATA_FOR_STUDENTS) {
     return {
       ...state,
-      expandedMetadataStudentIds: [
+      expandedMetadataStudentIds: _.uniq([
         ...state.expandedMetadataStudentIds,
         ...action.studentIds,
-      ],
+      ]),
     };
   }
   if (action.type === COLLAPSE_METADATA_FOR_STUDENTS) {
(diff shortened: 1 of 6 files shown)


apps/src/templates/sectionProgress/sectionProgressRedux.js
```diff
@@ -109,6 +124,23 @@ export default function sectionProgress(state = initialState, action) {
       },
     };
   }
+  if (action.type === EXPAND_METADATA_FOR_STUDENTS) {
+    return {
+      ...state,
+      expandedMetadataStudentIds: [
+        ...state.expandedMetadataStudentIds,
```
etaderhold, line 133: Do you intend a use case to add/remove multiple students at once? Perhaps an "expand all" button somewhere? It seems to be only one at a time in the current implementation though. 

Regardless it might be useful to ensure that duplicates can't be added here.

etaderhold (approved):

apps/src/templates/sectionProgress/sectionProgressRedux.js
lfryemason: 1) Yep, we want an expand all and collapse all button.

2) I think duplicates should be fine because the collapse method removes all duplicates. Everything should just check whether an id is in the table. I think that the performance gain will not outweigh the cost of de-duplicating.

etaderhold: The collapse action seems to only remove duplicates of whatever you're removing.

Consider a small section with student IDs `[1,2,3,4,5]`.

Slightly malicious scenario:
1. Expand all. `expandedMetadataStudentIds = [1, 2, 3, 4, 5]`
2. Collapse student 1. `expandedMetadataStudentIds = [2, 3, 4, 5]`
3. Expand all. `expandedMetadataStudentIds = [2, 3, 4, 5, 1, 2, 3, 4, 5]`
4. Collapse student 1. `expandedMetadataStudentIds = [2, 3, 4, 5, 2, 3, 4, 5]`
5. Expand all. `expandedMetadataStudentIds = [2, 3, 4, 5, 2, 3, 4, 5, 1, 2, 3, 4, 5]`
6. ...

Your point about premature optimization is well taken though. 👍

lfryemason: Very good example. I added the deduplication

PR #58972, opened 2024-05-29
Python Lab: reset globals on each run

molly-moen:
Pyodide has a single interpreter for its lifetime, so we could get some inconsistent results in some programs. For example, before this change, the below code would print `x is not in locals` on first run, but `x is in locals` on subsequent runs.
```
if 'x' in locals():
  print("x is in locals")
else:
  print("x is not in locals")

x = 5
```

The fix for this is to save the python globals we see on pyodide load, and delete any globals that were not in that original list after each run.

## Links

- jira ticket: [CT-505](https://codedotorg.atlassian.net/browse/CT-505)

## Testing story
Tested locally. The above test code now always prints `x is not in locals`, and other code still works as expected.

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/pythonlab/pythonHelpers/pythonScriptUtils.ts
+++ b/apps/src/pythonlab/pythonHelpers/pythonScriptUtils.ts
@@ -210,6 +210,21 @@ export async function importPackagesFromFiles(
   }
 }
 
+// Delete any python globals that were not present in
+// the original globals on pyodide start (originalGlobals param).
+// This allows us to ensure each pyodide run has a clean state.
+export function resetGlobals(
+  pyodide: PyodideInterface,
+  originalGlobals: Map<string, string>
+) {
+  const newGlobals = pyodide.globals.toJs();
+  newGlobals.forEach((_value: string, key: string) => {
+    if (!originalGlobals.has(key)) {
+      pyodide.globals.delete(key);
+    }
+  });
+}
+
 // For the given fileId, return the module version of the file. For example, a file at
 // path folder1/folder2/file.py would have a module name of "folder1.folder2.file".
 function getModuleName(fileId: string, source: MultiFileSource) {
--- a/apps/src/pythonlab/pyodideWebWorker.js
+++ b/apps/src/pythonlab/pyodideWebWorker.js
@@ -1,6 +1,7 @@
 import {
   getUpdatedSourceAndDeleteFiles,
   importPackagesFromFiles,
+  resetGlobals,
   writeSource,
 } from './pythonHelpers/pythonScriptUtils';
 import {DEFAULT_FOLDER_ID} from '@codebridge/constants';
@@ -20,11 +21,13 @@ async function loadPyodideAndPackages() {
 }
 
 let pyodideReadyPromise = null;
+let pyodideGlobals = null;
 async function initializePyodide() {
   if (pyodideReadyPromise === null) {
     pyodideReadyPromise = loadPyodideAndPackages();
   }
   await pyodideReadyPromise;
+  pyodideGlobals = self.pyodide.globals.toJs();
 }
 
 // Get pyodide ready as soon as possible.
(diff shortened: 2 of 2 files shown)


snickell (commented):
Is there a reason we don't just restart the interpreter entirely? There's state beyond globals that can be modified by execution, and I'm skeptical of our ability to clean up sufficiently so its not a potential long tail of subtle bugs. For example, I suspect there could be problems with python refusing to "re-import" modules, even if the source has changed.

If the downside is it takes a little time to re-initialize the interpreter (say 1s), I think this would be a worthwhile price to pay for not having subtle weird bugs confusing new programmers. Unless its very clear there's no approach to this, I think it'd be worth spending a little time figuring out if its possible/not-possible.

snickell:
Just in terms of time to initialize the interpreter, I feel the reload speed of this page is totally reasonable for the upside of having a truly clean slate: https://pyodide.org/en/stable/console.html

On my system, that feels like just over a second. It will probably be slower on a chromebook, but I feel like if this was a local python interpreter, we wouldn't be risking state pollution between runs to optimize startup time.

snickell:
I made a simple jsfiddle based on [console.html](https://pyodide.org/en/stable/console.html) to confirm that you can load separate pyodides, and to get a sense of how long they take to load: https://jsfiddle.net/bgL6wq9n/

Every time you click the "new terminal" button it does a loadPyodide, and pops up an alert how long the loadPyodide took:
image

1. The interpreters do seem to be separate state-wise
2. On my (fast) system it takes about 1s to load a new interpreter, chromebooks are gonna be significantly slower.

snickell:
Made another jsfiddle, this time to experiment with resetGlobals() behavior.

Using this:
https://jsfiddle.net/8q9rp1uL/2/

Here's a very simple example, pressed the "reset globals" button after assigning 5 to boo:
image

That works as expected!

But here's another one that illustrates a potentially serious problem (possibly the first of many) with resetGlobals (vs restarting the interpreter). Lets say the student has their own file boo.py, which they import using `import boo`. Changing boo.py does not mark the module for re-import, and reset global isn't enough to get the module to reload. The problem is there's state outside of globals.:
image

apps/src/pythonlab/pythonHelpers/pythonScriptUtils.ts
```diff
@@ -210,6 +210,21 @@ export async function importPackagesFromFiles(
...
 
+// Delete any python globals that were not present in
+// the original globals on pyodide start (originalGlobals param).
+// This allows us to ensure each pyodide run has a clean state.
+export function resetGlobals(
+  pyodide: PyodideInterface,
+  originalGlobals: Map<string, string>
+) {
+  const newGlobals = pyodide.globals.toJs();
+  newGlobals.forEach((_value: string, key: string) => {
```
snickell, line 221: I've always wished JS hadn't tried to make Map.forEach "compatible" with Array.forEach.... the reversed value/key always makes me gulp 🤣

snickell (approved):
👍 from me.

Molly tested on a chromebook (=slower, but also with wifi throttling enabled), and got almost 25s for a load+first-run. This is so slow that it would render pyodide non-viable. She also pointed me to code where we wipe user-deleted modules on re-runs: https://github.com/code-dot-org/code-dot-org/blob/d5c165f2309ff87cd0690f68ae190803805a39df/apps/src/pythonlab/pythonHelpers/pythonScriptUtils.ts#L30

I'd love to see a measure of chromebook loadPyodide + first-run time NOT including wifi/network time, but assuming its a similar order of magnitude, I agree this optimization is worth it.

molly-moen:
Noting here the 25 seconds on chromebook is without wifi throttling enabled. [Slack thread](https://codedotorg.slack.com/archives/C05DK21DAHK/p1717087322424489) with more details.

PR #58973, opened 2024-05-29
Add music lab to featured projects

fisher-alice:
This PR allows music lab projects to be added to the featured project gallery. Because thumbnail images are not generated for music lab projects, a customized thumbnail image is used for the featured projects public gallery and the admin featured projects page. However, the music lab image for the public gallery (via the project card) is slightly different in that more space is allowed for the music note.

An aside - at first, I also removed `studio` type labs from featured project types because it is no longer possible to convert a `studio` free-play level to a standalone project. But it is still possible to create a `studio` standalone project 'from scratch' so reverted the removal.

Note that a project still needs to be in a 'published' state in order for it to be displayed in the public featured gallery. Follow-up work includes providing a way for an admin user to publish a project since currently an engineer is needed to publish a project.

This is part 1 for the [task: Music: Ability to select featured projects to show on the home page (code.org/music)](https://codedotorg.atlassian.net/browse/LABS-842). Part 2 is to retrieve the channel ids of the active music lab featured projects to use for the mini-player at code.org/music.

## After update

Screenshot of public featured project gallery:
Screenshot 2024-05-31 at 12 27 51 PM

Screenshot of music lab projects after being added as featured projects at /projects/featured.
music-thumbnails-featured-admin

## Links
[Slack thread](https://codedotorg.slack.com/archives/C05DMS18MEX/p1716325573252699)

## Testing story
I tested locally by creating two different music lab project, published them, then made them active featured projects using dashboard-console commands.

## PR Checklist:

- [ ] Code is well-commented


apps/src/templates/projects/FeaturedProjectsTable.jsx
```diff
@@ -82,7 +84,10 @@ export const styles = {
 // Cell formatters.
 const thumbnailFormatter = function (thumbnailUrl, {rowData}) {
   const projectUrl = `/projects/${rowData.type}/${rowData.channel}/`;
-  thumbnailUrl = thumbnailUrl || PROJECT_DEFAULT_IMAGE;
+  thumbnailUrl =
+    thumbnailUrl ||
+    PROJECT_DEFAULT_THUMBNAIL_IMAGE_OVERRIDE[rowData.type] ||
+    PROJECT_DEFAULT_IMAGE;
```
thomasoniii: I see that this code is basically re-used down in `ProjectCard.jsx`. Could it be broken out into a utility function?

maybe as

```
const thumbnailUrl = getThumbnailUrl(path, type)

const getThumbnailUrl = (path, type) =>
  path
  || PROJECT_DEFAULT_THUMBNAIL_IMAGE_OVERRIDE[rowData.type] 
  || PROJECT_DEFAULT_IMAGE
```

where `getThumbnailUrl` is defined in some external file with access to those `PROJECT_DEFAULT_*` constants.

Then this function becomes `thumbnailUrl = getThumbnailUrl(thumbnailUrl, rowData.type)` (could also re-name the parameter so it's not re-assigned...`const thumbnailUrl = getThumbnailUrl(givenThumbnailUrl, rowData.type)`, maybe?)

And then `ProjectCard` lines 100-104 becomes `getThumbnailUrl(projectData.thumbnailUrl, type)`

There may be a better name for it.

I'm just worried about needing to keep these two bits of code in sync and forgetting to update one or the other down th eline.

thomasoniii (approved):
suggestion about adding an abstraction to getting the thumbnail url, but otherwise lgtm 🚀

apps/src/templates/projects/FeaturedProjectsTable.jsx
fisher-alice: Sounds good. Added function to a new `projectUtils.ts` file. Thanks!

apps/src/templates/projects/projectUtils.ts
```diff
@@ -0,0 +1,18 @@
+import {
+  PROJECT_DEFAULT_IMAGE,
+  PROJECT_DEFAULT_THUMBNAIL_IMAGE_OVERRIDE,
+  PROJECT_DEFAULT_CARD_IMAGE_OVERRIDE,
+} from './projectConstants';
+
+export const getThumbnailUrl = (
+  path: string,
+  type: keyof typeof PROJECT_DEFAULT_THUMBNAIL_IMAGE_OVERRIDE
```
bencodeorg: Trying to understand TS...isn't `type` often going to be something other than music (which I think is the only valid value given the contents of the constant)? I guess it compiles but TS confuses me sometimes...

apps/src/templates/projects/ProjectCardGrid.jsx
```diff
@@ -86,14 +86,13 @@ class ProjectCardGrid extends Component {
             <ProjectAppTypeArea
               labKey="featured"
               labName={i18n.featuredProjects()}
-              labViewMoreString={i18n.projectTypeDanceViewMore()}
```
bencodeorg, line 89: What does this change do?

apps/src/templates/projects/ProjectCardGrid.jsx
```diff
@@ -86,14 +86,13 @@ class ProjectCardGrid extends Component {
...
               projectList={projectLists.featured}
               numProjectsToShow={numProjects}
               galleryType={this.props.galleryType}
               navigateFunction={this.onSelectApp}
               isDetailView={false}
               hideViewMoreLink={true}
-              hideWithoutThumbnails={true}
```
bencodeorg, line 96: Does this change have any potential impact on other lab types?

apps/src/templates/projects/ProjectCardGrid.jsx
fisher-alice: The public gallery update included not showing the 'View more of ...' button. So this value is not being used since `hideViewMoreLink` has the value of `true`. I left a comment [here](https://github.com/code-dot-org/code-dot-org/blob/8b8bd34cd3a63df769b309e189eaa7b969ede883/apps/src/templates/projects/ProjectAppTypeArea.jsx#L34-L37).

apps/src/templates/projects/ProjectCardGrid.jsx
fisher-alice: Since the public gallery only displays curated featured projects, admin will select projects for which thumbnail images are available. At '/projects/featured' (the admin featured projects page), the thumbnail images are displayed alongside a link to the project.

I believe in the past, since any project could be published, this was a way to filter projects without thumbnail images from being included in the public gallery (projects without any content).
As new labs are added, either they'll have a way to generate a corresponding thumbnail OR we'll need to provide a thumbnail image as we are doing for music lab. To my knowledge, legacy labs that are publishable generate thumbnail images.

bencodeorg: Makes sense, thanks!

bencodeorg (approved):

apps/src/templates/projects/projectUtils.ts
fisher-alice: Good question - I tried `string` at first and TS didn't like it. I'll ping Sanchit when he's back in office next week to get his thoughts. Thanks!

fisher-alice: Tagging @sanchitmalhotra126 now that he's back in office.

sanchitmalhotra126: Hmm yeah, I think because `PROJECT_DEFAULT_THUMBNAIL_IMAGE_OVERRIDE` is defined in a JS file, TS is reading its type to be strictly `{[key: 'music']: string}` rather than a generic `{[key: string]: string}` and therefore complains when you try to use a `string` to index into it. 

One solution would be to define this constant in TS where you can explicitly type it as a `{[key: string]: string}` and then you can have the `type` param be a `string`. Otherwise it might just be best to cast the param inside the function - have the `type` param be `string` and cast it to `keyof typeof PROJECT_DEFAULT_THUMBNAIL_IMAGE_OVERRIDE` internally. 

Using `keyof typeof` for the `type` param does technically compile but only because all the references to this function are in JS. If/when a TS file tries to use this function, it will fail to compile unless the provided `type` is always `'music'`.

fisher-alice: Thanks Sanchit! I converted `projectConstants` to a .ts file. Lmk what you think.

apps/src/templates/projects/projectConstants.ts
```diff
@@ -63,3 +65,32 @@ export const convertChannelsToProjectData = function (projects) {
...
+};
+
+export const PROJECT_DEFAULT_THUMBNAIL_IMAGE_OVERRIDE: {
+  [projectType: string]: string;
+} = {
+  music: '/shared/images/fill-70x70/courses/logo_music.png',
+};
+
+// Project data from the channels API.
+type project = {
```
sanchitmalhotra126: Hmm, we have a version of this interface in lab2 `types` (where it's called `Channel`) and I'm wondering if we want to reuse/consolidate...ideally it would be great to have a single source of truth for these types (so there's a consistent understanding of what a "project" is across the codebase) but probably out of scope of this PR to do that kind of refactor. Maybe ok to keep here for now but add a comment that this type is duplicated in lab2/types? cc @molly-moen 

also nit: capitalize `Projects`

apps/src/templates/projects/projectUtils.ts
```diff
@@ -0,0 +1,15 @@
...
+} from './projectConstants';
+
+export const getThumbnailUrl = (path: string, type: string) =>
+  path ||
+  PROJECT_DEFAULT_THUMBNAIL_IMAGE_OVERRIDE[type] ||
+  PROJECT_DEFAULT_IMAGE;
+
+export const getProjectCardImageUrl = (
+  path: string,
+  type: keyof typeof PROJECT_DEFAULT_CARD_IMAGE_OVERRIDE
```
sanchitmalhotra126: should this also be `string`?

apps/src/templates/projects/projectConstants.ts
fisher-alice: Oh good to know about `Channel` interface - thanks Sanchit! I [updated the interface ](https://github.com/code-dot-org/code-dot-org/pull/58973/commits/71f6bd58a37c1ec1615eb90f334ccc35d943c9b2) adding two optional properties needed for `convertChanneslToProjectData`. @molly-moen lmk if this looks okay to you. If not, I can revert this commit. Thanks!

fisher-alice: The other option that @sanchitmalhotra126 and I discussed was converting the new `projectUtils.ts` file to a .js file. 🤷‍♀️

molly-moen: Seems fine to add those fields there, they already exist on the channel itself, right?

fisher-alice: Yes, they do. Thanks!

PR #58983, opened 2024-05-30
AI Tutor: add inappropriate boolean column to ai tutor interaction feedbacks

Erin007:
[CT-609](https://codedotorg.atlassian.net/browse/CT-609) partial

In addition to 👍  or 👎 feedback, users need to be able to flag an AI Tutor response as "inappropriate". This sets up the column on the AITutorInteractionsFeedback table to store whether the response has been flagged or not.

### DIFF
--- a/dashboard/db/migrate/20240530170255_add_flagged_inappropriate_to_ai_tutor_interaction_feedbacks.rb
+++ b/dashboard/db/migrate/20240530170255_add_flagged_inappropriate_to_ai_tutor_interaction_feedbacks.rb
@@ -1,0 +1,6 @@
+class AddFlaggedInappropriateToAiTutorInteractionFeedbacks < ActiveRecord::Migration[6.1]
+  def change
+    add_column :ai_tutor_interaction_feedbacks, :inappropriate, :boolean
+    add_index :ai_tutor_interaction_feedbacks, :inappropriate
+  end
+end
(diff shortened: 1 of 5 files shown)


dashboard/db/migrate/20240530170255_add_flagged_inappropriate_to_ai_tutor_interaction_feedbacks.rb
```diff
@@ -0,0 +1,6 @@
+class AddFlaggedInappropriateToAiTutorInteractionFeedbacks < ActiveRecord::Migration[6.1]
+  def change
+    add_column :ai_tutor_interaction_feedbacks, :inappropriate, :boolean
+    add_index :ai_tutor_interaction_feedbacks, :inappropriate
```
ebeastlake, line 4: What's the motivation for adding an index on this column?

Erin007: My thinking was that we'd want to query for all of the inappropriate ai tutor interaction feedbacks, but sounds like we're headed in a [different direction](https://codedotorg.slack.com/archives/C04VD4ZB6BT/p1717091636478109?thread_ts=1717090761.195869&cid=C04VD4ZB6BT). revisions incoming!

Erin007:
After discussion with Product/Design we're sticking with just 👍 👎 as the options.

PR #58989, opened 2024-05-30
Handle no project identifier in aichat request

bencodeorg:
We saw Honeybadger errors come in for a missing channel ID yesterday. I repro this situation if a teacher views a student's work via the teacher panel but the student hasn't actually viewed the level (ie, has no channel ID).

The change here is to handle if no channel ID is provided (ie, don't include it in the stored session).

## Links

- honeybadger error: https://app.honeybadger.io/projects/3240/faults/108087043

## Testing story

Tested manually that the observed repro case does not fail any more (and logs successfully) with this change in place.

### DIFF
--- a/dashboard/app/controllers/aichat_controller.rb
+++ b/dashboard/app/controllers/aichat_controller.rb
@@ -117,9 +117,11 @@ class AichatController < ApplicationController
       return false
     end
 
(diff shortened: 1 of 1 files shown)


sanchitmalhotra126 (approved):

fisher-alice (approved):
Nice!

PR #58997, opened 2024-05-30
AI Tutor: AssistantMessageFeedback component

Erin007:
[CT-541](https://codedotorg.atlassian.net/browse/CT-541)
Common Chatbot components [doc](https://docs.google.com/document/d/1kUWgyT_FbjCvXN2Lc28cMF6NWNrfv62O5SlVdjEIwFQ/edit)
[Figma](https://www.figma.com/design/lOLzGeTfH44a5YDS7GrzdA/AI-Tutor?node-id=2386-8167&t=I89XwOK2y2MlJ50n-0) 

I extracted the 👍 👎 feedback icons into their own component, `AssistantMessageFeedback`, which will make it slightly easier to expand on the feedback if/when we want to add UI to collect details about the reasoning for the rating ([example](https://codedotorg.slack.com/archives/C04VD4ZB6BT/p1717098777862579?thread_ts=1717090761.195869&cid=C04VD4ZB6BT). In the new component I modified the styling to be more consistent with the proposed design for a common feedback component. There is now a "Is this helpful?" prompt and the icons live below the assistant's message. 

BEFORE 

Screenshot 2024-05-31 at 11 57 32 AM

AFTER 

Screenshot 2024-05-31 at 12 14 02 PM


apps/src/aiTutor/views/AssistantMessageFeedback.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+const AssistantMessageFeedback: React.FC<AssistantMessageProps> = ({
+  messageId,
+}) => {
+  const [feedbackState, setFeedbackState] = useState<FeedbackData>({
+    thumbsUp: false,
+    thumbsDown: false,
+  });
+
+  const handleFeedbackSubmission = async (
```
ebeastlake, line 19: If we want this to be more generalizable in the future, we could pass `handleFeedbackSubmission` as a callback and the checked state as a prop, but that seems like work for a future PR.

ebeastlake (approved):
Nice!

PR #59000, opened 2024-05-30
Code Bridge: small console updates

molly-moen:
A couple updates to the code bridge console:
- On level change, reset the console. Previously the contents would stick around until you manually clicked the clear button, which is a bit confusing when all your code changes
- Change the clear console button from a broom to an eraser, to match current mocks.
- Remove the hard-coded `[PYTHON LAB]` prefix from the start of every system message, and instead make the prefix dependent on the lab app name. For now only python lab adds a system message prefix, as it's the only one that uses this console.

For the clear console logic, I did something similar to #58967 so that the console wouldn't be cleared on hot reload. This has an interesting side effect that is only present if you are in a level progression that switches from Python Lab to another lab2 lab (and therefore never reload the page). In this case the console contents persist, because the level never changed from the console's point of view. This seems fine to me as the console contents were for that level anyway, and this is probably a very rare situation to come across. 

## Links

- jira ticket: [CT-614](https://codedotorg.atlassian.net/browse/CT-614)

## Testing story
Tested locally.

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/codebridge/Console/index.tsx
+++ b/apps/src/codebridge/Console/index.tsx
@@ -15,6 +15,7 @@ const Console: React.FunctionComponent = () => {
   const previousLevelId = useRef(levelId);
   const appName = useAppSelector(state => state.lab.levelProperties?.appName);
 
+  // TODO: Update this with other apps that use the console as needed.
   const systemMessagePrefix = appName === 'pythonlab' ? '[PYTHON LAB] ' : '';
 
   useEffect(() => {
--- a/apps/src/codebridge/Console/index.tsx
+++ b/apps/src/codebridge/Console/index.tsx
@@ -13,6 +13,9 @@ const Console: React.FunctionComponent = () => {
   const dispatch = useDispatch();
   const levelId = useAppSelector(state => state.lab.levelProperties?.id);
   const previousLevelId = useRef(levelId);
+  const appName = useAppSelector(state => state.lab.levelProperties?.appName);
+
+  const systemMessagePrefix = appName === 'pythonlab' ? '[PYTHON LAB] ' : '';
 
   useEffect(() => {
     // If the level changes, clear the console.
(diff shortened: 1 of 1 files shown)


apps/src/codebridge/Console/index.tsx
```diff
@@ -11,6 +11,19 @@ import moduleStyles from './console.module.scss';
 const Console: React.FunctionComponent = () => {
   const codeOutput = useAppSelector(state => state.codebridgeConsole.output);
   const dispatch = useDispatch();
+  const levelId = useAppSelector(state => state.lab.levelProperties?.id);
+  const previousLevelId = useRef(levelId);
+  const appName = useAppSelector(state => state.lab.levelProperties?.appName);
+
+  const systemMessagePrefix = appName === 'pythonlab' ? '[PYTHON LAB] ' : '';
```
ebeastlake, line 19: Should we add a TODO to reference updating this with other appNames eventually?

ebeastlake (approved):

apps/src/codebridge/Console/index.tsx
molly-moen: good call, will update

apps/src/codebridge/Console/index.tsx
```diff
@@ -11,6 +11,20 @@ import moduleStyles from './console.module.scss';
 const Console: React.FunctionComponent = () => {
   const codeOutput = useAppSelector(state => state.codebridgeConsole.output);
   const dispatch = useDispatch();
+  const levelId = useAppSelector(state => state.lab.levelProperties?.id);
+  const previousLevelId = useRef(levelId);
+  const appName = useAppSelector(state => state.lab.levelProperties?.appName);
+
+  // TODO: Update this with other apps that use the console as needed.
+  const systemMessagePrefix = appName === 'pythonlab' ? '[PYTHON LAB] ' : '';
```
thomasoniii, line 19: this is fine as is, but over the slightly longer term this should probably be broken out into a separate function `getSystemMessagePrefix` or something, so we have a more clear separation of concerns

thomasoniii (approved):
added a future refactoring note, but otherwise lgtm 🚀

apps/src/codebridge/Console/index.tsx
molly-moen: Yep makes sense, this was a quick measure to remove the hard coded python lab string.

PR #59003, opened 2024-05-30
AI Tutor: user message tweaks

Erin007:
Addresses 2 pieces of 🐞 Bug Bash feedback for AI Tutor: 

1.) Users can no longer submit empty messages, which prevents us from making empty calls to the chatAPI and saves resources. 

2.) Users can now submit their message by pressing 'Enter' rather than having to always click the 'Submit' button.

### DIFF
--- a/apps/src/aiTutor/views/AITutorFooter.tsx
+++ b/apps/src/aiTutor/views/AITutorFooter.tsx
@@ -110,6 +110,9 @@ const AITutorFooter: React.FC<AITutorFooterProps> = ({renderAITutor}) => {
           onChange={e => setUserMessage(e.target.value)}
           value={userMessage}
           disabled={disabled}
+          onKeyDown={(e) => (
+            e.key === 'Enter' ? handleSubmit(ActionType.GENERAL_CHAT) : null
+          )}
         />
         <Button
           className={style.submitButton}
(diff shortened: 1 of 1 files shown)


apps/src/aiTutor/views/AITutorFooter.tsx
```diff
@@ -110,11 +112,16 @@ const AITutorFooter: React.FC<AITutorFooterProps> = ({renderAITutor}) => {
           onChange={e => setUserMessage(e.target.value)}
           value={userMessage}
           disabled={disabled}
+          onKeyDown={e =>
+            e.key === 'Enter' && !userMessageIsEmpty
+              ? handleSubmit(ActionType.GENERAL_CHAT)
+              : null
```
ebeastlake: nit: I think swapping this to a dedicated handler instead of using a ternary to return null when we're not using the return value could be slightly cleaner, but logic looks good! 
```
const handleKeyPress = (e: React.KeyboardEvent) => {
    if (e.key === 'Enter' && userMessage.trim() !== '') {
      handleSubmit(ActionType.GENERAL_CHAT);
    }
  };
```

ebeastlake (approved):
🎉

PR #59004, opened 2024-05-30
Gen AI: basic error handling

bencodeorg:
Makes an initial attempt at broadening error handling on the front end -- ie, if we receive any error from Rails/submitting a request/etc, we display a generic error response:

image

This case is the same as what we do when we receive profanity generated by the model, so I've reused that logic here. I haven't done any special casing to handle different errors, as I don't see much differentiation currently needed from a user perpsective whether we receive a failure from WebPurify, Sagemaker, or Rails.

## Links

- jira ticket: [LABS-820](https://codedotorg.atlassian.net/browse/LABS-820)

## Testing story

Tested manually that:
- I saw the error response pictured above when I intentionally threw an error or hard-coded profanity to be returned by the controller. 
- I was able to make the usual successful requests when no intentional error states were put in place. 
- I saw the appropriate messages being logged to our conversation history after this change.

### DIFF
--- a/apps/src/aichat/redux/aichatRedux.ts
+++ b/apps/src/aichat/redux/aichatRedux.ts
@@ -320,17 +320,44 @@ export const submitChatContents = createAsyncThunk(
 
     // Post user content and messages to backend and retrieve assistant response.
     const startTime = Date.now();
-    const chatApiResponse = await postAichatCompletionMessage(
-      newUserMessageText,
-      currentSessionId
-        ? storedMessages.filter(
-            message => message.sessionId === currentSessionId
-          )
-        : [],
-      aiCustomizations,
-      aichatContext,
-      currentSessionId
-    );
+
+    let chatApiResponse;
+    try {
+      chatApiResponse = await postAichatCompletionMessage(
+        newUserMessageText,
+        currentSessionId
+          ? storedMessages.filter(
+              message => message.sessionId === currentSessionId
+            )
+          : [],
+        aiCustomizations,
+        aichatContext,
+        currentSessionId
+      );
+    } catch (error) {
+      Lab2Registry.getInstance()
+        .getMetricsReporter()
+        .logError('Error in aichat completion request', error as Error);
+
+      const assistantChatMessage: ChatCompletionMessage = {
+        id: getNewMessageId(),
+        role: Role.ASSISTANT,
+        status: Status.ERROR,
+        chatMessageText: 'error',
+        timestamp: getCurrentTimestamp(),
+      };
+      thunkAPI.dispatch(addChatMessage(assistantChatMessage));
+
+      thunkAPI.dispatch(
+        updateUserChatMessageStatus({
+          id: newMessage.id,
+          status: Status.ERROR,
+        })
+      );
+
+      return;
+    }
+
     Lab2Registry.getInstance()
       .getMetricsReporter()
       .reportLoadTime('AichatModelResponseTime', Date.now() - startTime, [
--- a/apps/src/aichat/redux/aichatRedux.ts
+++ b/apps/src/aichat/redux/aichatRedux.ts
@@ -339,21 +339,7 @@ export const submitChatContents = createAsyncThunk(
         .getMetricsReporter()
         .logError('Error in aichat completion request', error as Error);
 
-      const assistantChatMessage: ChatCompletionMessage = {
-        id: getNewMessageId(),
-        role: Role.ASSISTANT,
-        status: Status.ERROR,
-        chatMessageText: 'error',
-        timestamp: getCurrentTimestamp(),
-      };
-      thunkAPI.dispatch(addChatMessage(assistantChatMessage));
-
-      thunkAPI.dispatch(
-        updateUserChatMessageStatus({
-          id: newMessage.id,
-          status: Status.ERROR,
-        })
-      );
+      updateMessagesOnError(newMessage, thunkAPI.dispatch);
 
       return;
     }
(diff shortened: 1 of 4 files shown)


levadadenys (approved):

apps/src/aichat/redux/aichatRedux.ts
```diff
@@ -604,6 +603,27 @@ const allFieldsHidden = (fieldVisibilities: AichatState['fieldVisibilities']) =>
...
   );
 
+const updateMessagesOnError = (
+  newMessage: ChatCompletionMessage,
+  dispatch: ThunkDispatch<unknown, unknown, AnyAction>
+) => {
+  const assistantChatMessage: ChatCompletionMessage = {
+    id: getNewMessageId(),
+    role: Role.ASSISTANT,
+    status: Status.ERROR,
```
fisher-alice, line 613: Now that we have `Status.ERROR_NOTIFICATION` would that status make more sense here?

apps/src/aichat/redux/aichatRedux.ts
```diff
@@ -604,6 +603,27 @@ const allFieldsHidden = (fieldVisibilities: AichatState['fieldVisibilities']) =>
...
+    chatMessageText: 'error',
+    timestamp: getCurrentTimestamp(),
+  };
+  dispatch(addChatMessage(assistantChatMessage));
+
+  dispatch(
+    updateUserChatMessageStatus({
+      id: newMessage.id,
+      status: Status.ERROR,
```
fisher-alice, line 622: In `ChatMessage.tsx`, the `chatMessageText` [is displayed](https://github.com/code-dot-org/code-dot-org/blob/ea984eaabd11973b70e13530e7e30d4c2246673b/apps/src/aichat/views/ChatMessage.tsx#L29-L32) with `ERROR` status. Just confirming that this is the case when a user's message results in an error (profane model response or other) and we do want to display the user's message.

apps/src/aichat/redux/aichatRedux.ts
```diff
@@ -604,6 +603,27 @@ const allFieldsHidden = (fieldVisibilities: AichatState['fieldVisibilities']) =>
...
 
+const updateMessagesOnError = (
+  newMessage: ChatCompletionMessage,
+  dispatch: ThunkDispatch<unknown, unknown, AnyAction>
+) => {
+  const assistantChatMessage: ChatCompletionMessage = {
+    id: getNewMessageId(),
+    role: Role.ASSISTANT,
+    status: Status.ERROR,
+    chatMessageText: 'error',
```
fisher-alice: Nit: I wonder if we should assign the string constant here.
'There was an error getting a response. Please try again.'

fisher-alice (approved):
Nice! Just left a couple small questions.

apps/src/aichat/redux/aichatRedux.ts
bencodeorg: It makes sense to me to keep the user message that resulted in an error response in view, so the user gets some feedback that they did indeed attempt to submit a message and we had an issue processing it. I could imagine it being confusing if your message just disappeared...what do you think?

Worth noting that we do exclude these "unfulfilled" user messages from future conversation history that is submitted to the model here:

https://github.com/code-dot-org/code-dot-org/blob/ea984eaabd11973b70e13530e7e30d4c2246673b/dashboard/app/controllers/aichat_controller.rb#L44

fisher-alice: Thanks! and yes, I agree it would be confusing if your message just disappeared.

I wonder if it would make sense to indicate via UI that this user message is in an error state to reflect that it's not part of the chat conversation with the chatbot. This is outside the scope of this PR but if the chat workspace is acting as a kind of history for the user, it could be useful to distinguish an errored/unfulfilled user message and an okay/fulfilled one.

apps/src/aichat/redux/aichatRedux.ts
bencodeorg: I think of the "error response to your submission to model" as different than the "error occurred elsewhere and we're using the chat window to notify you of the error occurring". Looking at the [rough draft shared UI components for Code.org chatbots](https://docs.google.com/document/d/1kUWgyT_FbjCvXN2Lc28cMF6NWNrfv62O5SlVdjEIwFQ/edit), it does look like the preferred color for this situation would be red instead of yellow, so I updated to that. Thoughts on this?

image

apps/src/aichat/redux/aichatRedux.ts
bencodeorg: Good call, updated!

apps/src/aichat/redux/aichatRedux.ts
fisher-alice: >I think of the "error response to your submission to model" as different than the "error occurred elsewhere and we're using the chat window to notify you of the error occurring". 

Yeah - that makes sense. Thanks!

And agree that it would be good to use the preferred color outlined in doc - good call!

fisher-alice: I now see that Emily [left a comment](https://docs.google.com/document/d/1kUWgyT_FbjCvXN2Lc28cMF6NWNrfv62O5SlVdjEIwFQ/edit?disco=AAABNnOTo9I) about this questioning the color. So maybe leave for now and ask Sam when she gets back or when conversation is resolved?

bencodeorg: It's already non-standard colored (ie, yellow), which I think is out of sync with Emily's observation anyway, so I'd be inclined to roll with what's currently laid out in the doc (ie, red)...happy to update once Mark weighs in though!

PR #59005, opened 2024-05-30
Gen AI: move redux utils into separate file

bencodeorg:
We have a decent number of utils stashed in our main `aichatRedux` file, which makes finding your way to the core functionality a bit tricky at times. This PR moves those utility functions into their own file.

## Testing story

Relying a bit on linting/TS/webpack to make sure I haven't missed anything here -- there are no (intended) changes here other than moving things around. I did some basic playing with the lab (chatting, saving customizations, etc) and didn't observe any issues.


apps/src/aichat/redux/aichatRedux.ts
```diff
@@ -645,7 +572,6 @@ export const {
   setNewChatSession,
   setChatSessionId,
   clearChatMessages,
-  setIsWaitingForChatResponse,
```
bencodeorg, line 659: This is the only intended diff that's not just moving things from one place to another -- my editor is telling me this is unused, and I didn't see any usages.

fisher-alice: Oh yeah - I think I added this and then added the extra reducers afterwards so this is not needed. Thanks for cleaning up!

apps/src/aichat/redux/aichatRedux.ts
```diff
@@ -403,6 +354,29 @@ export const submitChatContents = createAsyncThunk(
   }
 );
 
+// Helper that is used when we receive an error response
```
fisher-alice, line 368: Thanks for adding the comment!

apps/src/aichat/redux/utils.ts
```diff
@@ -0,0 +1,80 @@
...
+        !modelCardInfo['exampleTopics'].filter(topic => topic.length).length
+      ) {
+        return false;
+      }
+    } else if (!modelCardInfo[key].length) {
+      return false;
+    }
+  }
+
+  return true;
```
thomasoniii, line 74: I think this could be smooshed down a lot, though I'm not convinced it's more legible. 

Assuming I got all my logic right...

```
// note that we return the negation - so if we find a key === 'exampleTopics' AND the modelCardInfo.exampleTopics
// has a length, OR we have no length on the modelCardInfo[key], then we return a true value (result of the `some`)
// but we want to return a false in that case, so we negate it. If we find nothing, then the `some` is false so 
// we negate it and return `true`
return !getTypedKeys(modelCardInfo).some( key => 
  (key === 'exampleTopics' && !modelCardInfo.exampleTopics.filter(topic => topic.length).length)
  || !modelCardInfo[key].length
```

thomasoniii (approved):
couple of minor non-blocking suggestions/nits, but otherwise lgtm 🚀

apps/src/aichat/types.ts
```diff
@@ -66,6 +66,8 @@ export interface AiCustomizations {
 // The customizations will be included in request to LLM endpoint.
 export type AichatModelCustomizations = Omit<AiCustomizations, 'modelCardInfo'>;
 
+export type FieldVisibilities = {[key in keyof AiCustomizations]: Visibility};
```
fisher-alice, line 74: So nice.

fisher-alice (approved):
I appreciate so much this refactor. 🎉  🧹 🎉

apps/src/aichat/redux/utils.ts
bencodeorg: ugh yeah you're right this isn't pretty / is quite wordy...I like your suggestion but think it doesn't cover the `isPublished` case though, which is a boolean. It'd have to be:

```
return !getTypedKeys(modelCardInfo).some( key => 
  (key === 'exampleTopics' && !modelCardInfo.exampleTopics.filter(topic => topic.length).length)
  || (key !== 'isPublished' && !modelCardInfo[key].length)
```

The various negations/logic combinations you have to do in your head for this to work feels harder to understand than the quite verbose version we have currently IMO...what do you think?

fisher-alice: FWIW, I think that the current implementation may be easier to understand for a future reader.

PR #59010, opened 2024-05-31
Nicklathe/fix dynamic registration

nicklathe:
This PR address a few Dynamic Registration improvements and bug fixes.

Fixes:
* Specifies the same scopes in registration as we request with the authentication token. This was preventing course syncing from working, as we were requesting different scopes in our authentication token request than we were specifying during the registration step.

Enhancements:
* Adds padding between the email label in the input text box
* Adds icon URL
* Removes undesirable placements
* Uses the `CDO.studio_url` and `CDO.dashboard_site_host` helper to set dynamic domain based on environment

Cleanup:
* Removes unused imports left over from a previous dynamic registration PR

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/dashboard/lib/policies/lti.rb
+++ b/dashboard/lib/policies/lti.rb
@@ -86,15 +86,15 @@ class Policies::Lti
     grant_types: ["client_credentials", "implicit"],
     initiate_login_uri: CDO.studio_url('/lti/v1/login', CDO.default_scheme),
     redirect_uris: [CDO.studio_url('/lti/v1/authenticate', CDO.default_scheme)],
-    client_name: "Code.org",
+    client_name: "Code.org - Nick Dynamic",
     jwks_uri: CDO.studio_url('/oauth/jwks', CDO.default_scheme),
     token_endpoint_auth_method: "private_key_jwt",
     contacts: ["platform@code.org"],
     scope: "https://purl.imsglobal.org/spec/lti-ags/scope/score https://purl.imsglobal.org/spec/lti-nrps/scope/contextmembership.readonly",
     "https://purl.imsglobal.org/spec/lti-tool-configuration" => {
-      domain: "studio.code.org",
-      description: "Code.org",
-      target_link_uri: "https://studio.code.org/lti/v1/sync_course",
+      domain: CDO.studio_url('', CDO.default_scheme),
+      description: "Code.org - Nick Dynamic",
+      target_link_uri: CDO.studio_url('/lti/v1/sync_course', CDO.default_scheme),
       custom_parameters: {
         email: "$Person.email.primary",
         full_name: "$Person.name.full",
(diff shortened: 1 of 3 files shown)


dashboard/lib/policies/lti.rb
```diff
@@ -90,11 +90,11 @@ module MessageType
     jwks_uri: CDO.studio_url('/oauth/jwks', CDO.default_scheme),
     token_endpoint_auth_method: "private_key_jwt",
     contacts: ["platform@code.org"],
-    scope: "https://purl.imsglobal.org/spec/lti-ags/scope/score https://purl.imsglobal.org/spec/lti-nrps/scope/contextmembership.readonly",
+    scope: ALL_SCOPES.join(' '),
     "https://purl.imsglobal.org/spec/lti-tool-configuration" => {
-      domain: "studio.code.org",
-      description: "Code.org",
-      target_link_uri: "https://studio.code.org/lti/v1/sync_course",
+      domain: CDO.studio_url('', CDO.default_scheme),
```
stephenliang: Per the [LTI spec](https://www.imsglobal.org/spec/lti-dr/v1p0#lti-configuration-0) it looks like domain shouldn't include the protocol which I believe this change will add. 

> The primary domain covered by this tool; protocol must not be included. For example mytool.example.org

nicklathe: Hmmm yeah you're right. I'll look into if we have a helper method that leaves the protocol off.

nicklathe: Looks like I can just leave the second argument blank, and it defaults to an empty string.

nicklathe: Looks like I should use [CDO.dashboard_site_host](https://github.com/code-dot-org/code-dot-org/blob/staging/lib/cdo.rb#L114)

nicklathe: Tested on multiple environments:

Prod
```
[production] dashboard > CDO.dashboard_site_host
=> "studio.code.org"
```

Test
```
[test] dashboard > CDO.dashboard_site_host
=> "test-studio.code.org"
```

Local development
```
[development] dashboard > CDO.dashboard_site_host
=> "localhost-studio.code.org:3000"
```

dashboard/lib/policies/lti.rb
```diff
@@ -90,11 +90,11 @@ module MessageType
...
     contacts: ["platform@code.org"],
-    scope: "https://purl.imsglobal.org/spec/lti-ags/scope/score https://purl.imsglobal.org/spec/lti-nrps/scope/contextmembership.readonly",
+    scope: ALL_SCOPES.join(' '),
     "https://purl.imsglobal.org/spec/lti-tool-configuration" => {
-      domain: "studio.code.org",
-      description: "Code.org",
-      target_link_uri: "https://studio.code.org/lti/v1/sync_course",
+      domain: CDO.dashboard_site_host,
+      description: "Code.org LTI Integration",
+      target_link_uri: CDO.studio_url('/lti/v1/sync_course', CDO.default_scheme),
```
carl-codeorg: Does `lti_v1_sync_course_path` work for the path segment here? Could also add a new const DEFAULT_TARGET_LINK_URI and set it to `lti_v1_sync_course_path`

carl-codeorg (approved):
One tiny non-blocking comment, otherwise LGTM

dashboard/lib/policies/lti.rb
nicklathe: `lti_v1_sync_course_path` doesn't seem to work here. I like the idea of adding a new const in this file and referencing it.

PR #59012, opened 2024-05-31
Fix Progress v2 UI test

lfryemason:
Fix a flaky progress test `Teacher can view student work, ask student to keep working, on rubric level`
This was caused because the test was leaving a page after submitting feedback too quickly. This caused a popup for `Are you sure you want to leave`. By waiting an extra three seconds, we should remove this popup.

This also removes the test `Teacher can view validated level`
This functionality is covered by unit tests [here](https://github.com/code-dot-org/code-dot-org/blob/staging/apps/test/unit/templates/sectionProgressV2/LevelDataCellTest.jsx#L95). All ui except for the validated checkmark is also covered by other UI tests
This test can't be tested on Drone because it uses java lab. This makes it very hard to debug now and in the future. This test was not working when merged, even after a couple rounds of fixes and is currently disabled. I don't think that the minimal coverage is worth the effort to fix the test to work or continue to maintain it in the future.

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/dashboard/test/ui/features/teacher_tools/teacher_dashboard/teacher_dashboard_progress_v2.feature
+++ b/dashboard/test/ui/features/teacher_tools/teacher_dashboard/teacher_dashboard_progress_v2.feature
@@ -151,6 +151,7 @@ Scenario: Teacher can view student work, ask student to keep working, on rubric 
   And I wait to see "#ui-test-feedback-input"
   And I click selector "#keep-working" once I see it
   And I press "#ui-test-submit-feedback" using jQuery
+  And I wait for 3 seconds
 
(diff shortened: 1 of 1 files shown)


etaderhold (commented):
Thanks for figuring out this timing issue!

Regarding the deleted test is there no validated non-Javalab level in `allthethings` that could be used instead?

etaderhold (approved):
If it's not a simple swap to fix that test I agree with your assessment about the low value and am fine with deleting it.

lfryemason:
It is not a simple swap. I'm going to remove the test and keep it in mind to add later

PR #59017, opened 2024-05-31
Use Google Blockly for embedded blocks in lesson plans

mikeharv:
While [investigating](https://codedotorg.slack.com/archives/C0T0PNTM3/p1717160381952269) a build failure related to a deleted programming expression, I noticed that the affected lesson plan was failing to load completely:

image
image

The error indicates a problem with a field class that should be present in Google Blockly. However, setting a breakpoint revealed that we're running CDO Blockly:

image

Updating which script gets used in `show.html.haml` was enough to remove the error. However, I then noticed that the blocks in the code section were being created multiple times, evidently due to some change in the value of `blockRef` in the EmbeddedBlock component.

image

We can prevent multiple workspaces from being shown to the teacher user by finding the previous workspace SVG (if it exists) and removing it. With this change, the lesson plans displays correctly to the user.

image

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/templates/codeDocs/EmbeddedBlock.jsx
+++ b/apps/src/templates/codeDocs/EmbeddedBlock.jsx
@@ -11,6 +11,10 @@ export default function EmbeddedBlock({blockName, link, ariaLabel}) {
   useEffect(() => {
     if (blockName && blockRef.current) {
       const blocksDom = parseElement(`<block type='${blockName}' />`);
+      const previousBlockSpace = blockRef.current.querySelector(
+        'svg.readOnlyBlockSpace'
+      );
+      previousBlockSpace?.remove();
       const blockSpace = Blockly.createEmbeddedWorkspace(
         blockRef.current,
         blocksDom,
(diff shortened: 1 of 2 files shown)


etaderhold (approved):

etaderhold (approved):

apps/src/templates/codeDocs/EmbeddedBlock.jsx
```diff
@@ -11,6 +11,10 @@ export default function EmbeddedBlock({blockName, link, ariaLabel}) {
   useEffect(() => {
     if (blockName && blockRef.current) {
       const blocksDom = parseElement(`<block type='${blockName}' />`);
+      const previousBlockSpace = blockRef.current.querySelector(
```
molly-moen, line 14: In the screenshot it looked like we were creating duplicate blocks. Is there a way we can just skip creating a new block rather than deleting the existing one? If not, can you add a comment here explaining why we need to do this?

ebeastlake:
I filed a ticket to capture @molly-moen's comment so we can merge this PR and fix the lesson plans without the follow-up getting lost: https://codedotorg.atlassian.net/browse/CT-622

molly-moen (approved):

PR #59018, opened 2024-05-31
jimt/CT-623 - popup buttons

thomasoniii:
It's still ugly as sin, but this at least implements a disappearing kebab menu upon mouseover, instead of the inlined row of buttons.

Also adds a prettier header.

Old:
Screenshot 2024-05-31 at 4 59 08 PM

vs new:
Screenshot 2024-05-31 at 4 58 54 PM
Screenshot 2024-05-31 at 4 58 59 PM

Couple known things:
- ~this simple button isn't smart enough to navigate away when you click elsewhere, so we need a better handler for it.~
- Added a better handler so click aways clears out the pop up menu
- the kebab vanishes when you pop up the menu and mouse away, we probably should keep it.
- there's some sort of layering issue with the top + button for the new folder/new file popup - so it's cheating and right aligning it under the button. If it's left aligned, it disappears under the editor. Weird.
## Links

- jira ticket: [CT-623](https://codedotorg.atlassian.net/browse/CT-623)

## PR Checklist:

- [ ] Code is well-commented


apps/src/codebridge/PopUpButton/PopUpButton.scss
```diff
@@ -0,0 +1,14 @@
+@import 'color.scss';
```
molly-moen: can you make this a css module?

apps/src/codebridge/PopUpButton/PopUpButton.tsx
```diff
@@ -0,0 +1,56 @@
...
+      />
+      {isOpen && buttonRect && offsetParent && (
+        <div
+          className="popup-button-menu"
+          onClick={() => setIsOpen(false)}
+          style={{
+            top:
+              buttonRect.top + buttonRect.height + 5 - (offsetParent.top || 0),
+            [alignment]: buttonRect[alignment] - (offsetParent[alignment] || 0),
+            //left: 100,
```
molly-moen: remove commented out lines?

apps/src/codebridge/FileBrowser/index.tsx
```diff
@@ -296,29 +303,41 @@ export const FileBrowser = React.memo(() => {
...
-          renameFolderPrompt={renameFolderPrompt}
-          toggleFileVisibility={toggleFileVisibility}
-        />
-      </ul>
+      <PanelContainer
+        id="file-browser"
+        headerContent={'Files'}
+        className={moduleStyles['file-browser']}
+        rightHeaderContent={
+          <PopUpButton iconName="plus" alignment="right">
```
molly-moen: maybe worth extracting this into a helper method for readability?

apps/src/codebridge/FileBrowser/styles/fileBrowser.css
```diff
@@ -53,3 +68,12 @@ span.caret-container {
 .file-browser ul li span.button-bar i {
   margin-right: 4px;
 }
+
+li .button-kebab {
+  display: none;
+  height: 13px !important;
```
molly-moen: why do we need !important here?

apps/src/codebridge/FileBrowser/styles/fileBrowser.css
```diff
@@ -28,16 +28,31 @@
...
 .file-browser ul li span.button-bar {
   display: flex;
-  justify-content: flex-end;
+  justify-content: flex-start;
+  flex-direction: column;
   cursor: pointer;
 }
 
+.file-browser ul li span.button-bar span:hover {
+  background-color: #54595e;
```
molly-moen: can we use color constants here instead (from color.scss, would require changing the file to scss)? Ideally this file should be a css module but that can be a follow up.

molly-moen (commented):
Looks good, couple clean up suggestions!

apps/src/codebridge/FileBrowser/index.tsx
thomasoniii: I broke it out into a separate `FileBrowserHeaderPopUpButton` component.

apps/src/codebridge/FileBrowser/styles/fileBrowser.css
thomasoniii: I dunno. 🤷 

At the time, it was necessary and if I didn't have it in there it was oversizing the buttons so they were a little too tall, but presumably that was due to some other css glitch because after the extra bit of refactoring I did to address the comments, we seem to work okay w/o it. So I dropped it Whee! CSS!

apps/src/codebridge/FileBrowser/styles/fileBrowser.css
thomasoniii: Done!
Also moved it into a scss module.

thomasoniii:
Just noting, in addition to addressing the PR comments I also added a file icon to the browser and updated the click handler so a click elsewhere in the doc when the menu is open will close it.

apps/src/codebridge/PopUpButton/PopUpButton.tsx
```diff
@@ -0,0 +1,72 @@
...
+  children,
+  iconName,
+  className,
+  alignment = 'left',
+}: PopUpButtonProps) => {
+  const [isOpen, _setIsOpen] = useState(false);
+  const [buttonRect, setButtonRect] = useState<DOMRect>();
+  const [offsetParent, setOffsetParent] = useState<DOMRect>();
+
+  const setIsOpenFalse = useCallback(() => {
```
molly-moen, line 24: (for my understanding) why do we need `_setIsOpen`, `setIsOpen`, and `setIsOpenFalse`?

apps/src/codebridge/PopUpButton/PopUpButton.tsx
```diff
@@ -0,0 +1,72 @@
...
+  return (
+    <>
+      <Button
+        className={className}
+        size="xs"
+        icon={{iconStyle: 'solid', iconName}}
+        color="black"
+        isIconOnly
+        onClick={e => {
+          setButtonRect((e.target as HTMLElement).getBoundingClientRect());
```
molly-moen: nit: extract this out into a helper method?

molly-moen (approved):
Looks good!

apps/src/codebridge/PopUpButton/PopUpButton.tsx
thomasoniii: done, and also re-tooled and fiddled with the logic to help address the previous comment about the three methods, getting rid of the `_setIsOpen`/`setIsOpen` wrapper method.

There may be different ways to implement all of this, so take it with a grain of salt as a personal pref. The nuisance was the event listener which was tacked onto the document. _That_ event listener is the thing that responds to clicks elsewhere in the doc to close our pop up menu, and we want to clean it up when we no longer need it. In this case, that's when the pop up menu closes. Annoyingly, the only way we can remove an event listener is by handing in the same function we'd originally added.

So we need to keep a stable copy of that function, which is the `setIsOpenFalse` that does nothing more than set `isOpen` to false, as well as removing itself as the click handler. I'd originally added the additional `setIsOpen` wrapper as a way of keeping the logic for adding/removing the click handler closer to changing the `isOpen` value itself, as opposed to associating it more closely with the click handler. But that's personal preference and it could've gone in the other place.

...which is where I ended up moving it anyway. When I pulled out the embedded callback to stick in a separate helper with a `useCallback`, I lost access to the current `isOpen` value (w/o making it a callback dependency and making `useCallback` moot), and at that point I needed to move the event listener adding/removing into the `setState` callback anyway (to get at the `newIsOpen` value). Anyway, `setState` function is now effectively the old wrapper `setIsOpen` that set the value + added/removed the listeners, so it's in place there now; just embedded into the overall click handler.

`setIsOpenFalse` is admittedly possibly too short as a name, but conversely I felt that `setIsOpenFalseAndRemoveEventListener` was too long. Definitely open to better suggestions on it, though.

As I've often repeated, there are only 2 difficult tasks in computer science: Cache invalidation, naming things, and off by one errors.

molly-moen: thanks for the explanation!

PR #59023, opened 2024-06-01
Music: sounds panel preview variant update

breville:
Follow-up to https://github.com/code-dot-org/code-dot-org/pull/58746.

This makes some updates to the new sounds panel variant, still shown with the `sounds-panel-1-preview=true` URL parameter.  The variant now doesn't show the preview icon, rows have extra height, and the panel is taller.  Both the original and the variant now cancel preview sounds when dismissed.

### variant

Screenshot 2024-05-31 at 10 51 26 PM

### original

Screenshot 2024-05-31 at 10 51 33 PM

### DIFF
--- a/apps/src/music/blockly/FieldSounds.js
+++ b/apps/src/music/blockly/FieldSounds.js
@@ -158,6 +158,8 @@ class FieldSounds extends GoogleBlockly.Field {
   }
 
   dropdownDispose_() {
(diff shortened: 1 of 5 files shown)


thomasoniii (approved):
lgtm 🚀

sanchitmalhotra126 (approved):
Nice!

PR #59024, opened 2024-06-01
Music: Pack dialog updates

breville:
This makes some small updates to the pack dialog:

- the preview button becomes a stop button when the preview is playing.
- the preview button is slightly bigger.
- the preview button has the same vertical position for each entry.
- the preview stops playing when the dialog is closed.

It also introduces a variant behind the `?pack-dialog-preview=true` URL parameter that tries an alternative:

- the preview plays each time an entry is selected.
- again, the preview stops playing when the dialog is closed.

### DIFF
--- a/apps/src/music/views/PackDialog.tsx
+++ b/apps/src/music/views/PackDialog.tsx
@@ -9,7 +9,12 @@ import classNames from 'classnames';
 import FontAwesome from '@cdo/apps/templates/FontAwesome';
 import MusicPlayer from '../player/MusicPlayer';
 import {DEFAULT_PACK} from '../constants';
+import AppConfig from '../appConfig';
 import musicI18n from '../locale';
+
+// A variant for PackDialog that plays previews as sounds are selected.
+const usePackDialogPreview =
+  AppConfig.getValue('pack-dialog-preview') === 'true';
 
 interface PackEntryProps {
   playingPreview: string | null;
@@ -17,6 +22,7 @@ interface PackEntryProps {
   isSelected: boolean;
   onSelect: (path: SoundFolder) => void;
   onPreview: (path: string) => void;
+  onStopPreview: () => void;
 }
 
 const PackEntry: React.FunctionComponent<PackEntryProps> = ({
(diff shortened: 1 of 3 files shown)


apps/src/music/views/PackDialog.tsx
```diff
@@ -33,23 +40,35 @@ const PackEntry: React.FunctionComponent<PackEntryProps> = ({
...
+      }
+    }
+  }, [folder, isPlayingPreview, onPreview, onSelect, soundPath]);
+
   const onPreviewClick = useCallback(
     (e: Event) => {
-      if (soundPath && !isPlayingPreview) {
+      if (isPlayingPreview) {
+        onStopPreview();
+      } else if (soundPath && !isPlayingPreview) {
```
thomasoniii, line 57: just noting that `!isPlayingPreview` is unneeded and redundant - since if it were true, it would've been caught by the prior conditional.

That said, it might be useful to keep in if it adds clarity.

apps/src/music/views/PackDialog.tsx
```diff
@@ -33,23 +40,35 @@ const PackEntry: React.FunctionComponent<PackEntryProps> = ({
...
   const imageSrc = library?.getPackImageUrl(folder.id);
 
+  const onEntryClick = useCallback(() => {
+    onSelect(folder);
+
+    if (usePackDialogPreview) {
+      if (soundPath && !isPlayingPreview) {
+        onPreview(soundPath);
+      }
+    }
```
thomasoniii, line 50: Should this be smooshed into a single conditional with another and?

```
if (usePackDialogPreview && soundPath && !isPlayingPreview) {
...
```
?

thomasoniii (approved):
some minor logical sugar suggestions, but otherwise lgtm 🚀

apps/src/music/views/PackDialog.tsx
breville: I like keeping the `usePackDialogPreview` one separate because if the experiment is a success then we'll just take that line out when we switch to the new variant.

sanchitmalhotra126 (approved):

PR #59026, opened 2024-06-02
Music: callouts updates

breville:
This makes some updates to the Music Labs callouts.  

When the `clickable-text-with-glow=true` URL parameter is provided, a glow now plays behind each item of clickable text in instructions and feedback when it first appears, designed as a hint that it's clickable.  The same glow shows on hover, too.  

The goal is that it doesn't interfere with regular instruction reading, but still suggests there is something to do.  The yellow color is chosen to match the arrow that will show upon click.

Thanks to @snickell for revitalizing this effort with a prototype. 

Also, a callout can now open a toolbox category.

### Glow sequence

https://github.com/code-dot-org/code-dot-org/assets/2205926/830e0a98-2450-4c04-a8da-0fb34a1dd785

### Glow on hover

https://github.com/code-dot-org/code-dot-org/assets/2205926/d2536e33-6692-40ab-af8f-7fa8349a4697

### Opening a toolbox category

https://github.com/code-dot-org/code-dot-org/assets/2205926/4699b28c-8668-4db4-8935-302fdc792bc6

### DIFF
--- a/apps/src/templates/EnhancedSafeMarkdown.jsx
+++ b/apps/src/templates/EnhancedSafeMarkdown.jsx
@@ -87,10 +87,14 @@ export class ClickableTextWrapper extends React.Component {
 
   renderClickableText(node) {
     const clickableTextAll = node.querySelectorAll('b.clickable-text');
-    for (const clickableText of clickableTextAll) {
+    clickableTextAll.forEach((clickableText, index) => {
       const id = clickableText.dataset.id;
+      const extraClass = ` clickable-text-${index}`;
+      if (!clickableText.className.includes(extraClass)) {
+        clickableText.className += extraClass;
+      }
       clickableText.onclick = () => this.props.handleInstructionsTextClick(id);
-    }
+    });
   }
 
   render() {
--- a/apps/src/templates/EnhancedSafeMarkdown.jsx
+++ b/apps/src/templates/EnhancedSafeMarkdown.jsx
@@ -3,10 +3,15 @@ import React from 'react';
 import ReactDOM from 'react-dom';
 import {connect} from 'react-redux';
 
+import {queryParams} from '@cdo/apps/code-studio/utils';
 import {openDialog} from '@cdo/apps/redux/instructionsDialog';
 
 import SafeMarkdown from './SafeMarkdown';
 import {renderExpandableImages} from './utils/expandableImages';
+
+// A variant for clickable text that shows a glow on hover and appearance.
+const showClickableTextWithGlow =
+  queryParams('clickable-text-with-glow') === 'true';
 
 export class UnconnectedExpandableImagesWrapper extends React.Component {
   static propTypes = {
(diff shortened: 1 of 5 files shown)


apps/src/templates/EnhancedSafeMarkdown.jsx
```diff
@@ -87,10 +87,14 @@ export class ClickableTextWrapper extends React.Component {
 
   renderClickableText(node) {
     const clickableTextAll = node.querySelectorAll('b.clickable-text');
-    for (const clickableText of clickableTextAll) {
+    clickableTextAll.forEach((clickableText, index) => {
       const id = clickableText.dataset.id;
+      const extraClass = ` clickable-text-${index}`;
+      if (!clickableText.className.includes(extraClass)) {
+        clickableText.className += extraClass;
+      }
```
thomasoniii: significant whitespace always feels icky to me.

First off, is this always valid? I'm not sure what guarantees we have about the list of classnames, but since this is expecting the classname to _always_ have a space in front of it, is there a case where we could only have a classname we're looking for at the start of the string? i.e., we have `className = 'foo'` and we look to see if ` foo` appears in the string and fail as a result?

As a non-blocking suggestion, what about using a set? Something like this:

```
const clickableTextClassNames = new Set(clickableText.className.split(/\s+/));
clickableTextAll.forEach( (clickableText, index) => {
  const id = clickableText.dataset.id;
  const className = `clickable-text-${index}`; // leading space not required
  clickableTextClassNames.add(className)
  clickableText.onclick = () => this.props.handleInstructionsTextClick(id)
})

clickableText.className = Array.from(clickableTextClassNames).sort().join(' ');
```

thomasoniii (approved):
non-blocking question/suggestion about how additional classnames are added, but overall lgtm 🚀

apps/src/templates/EnhancedSafeMarkdown.jsx
breville: Agreed that this is not great, and more generally it's unfortunate that we have to protect against adding the classname repeatedly at all.  The new code is clever but I find significantly more complex than the proposed code, and the proposed code does seem to work reliably, since we are detecting first that it has the class `.clickable-text`, and the addition of ` clickable-text-${index}` does seem to always be applied just once.

dashboard/app/assets/stylesheets/application.scss
```diff
@@ -4099,9 +4099,56 @@ a {
   }
 }
 
+$clickable-text-glow: 0 0 1px $realyellow,
```
levadadenys, line 4102: nit: this color (value is `#ff0`) is not in scope of design system, just double checking if maybe we want to use color from design system? `$light_caution_500` for example(value is `#F9CB28`) cc @markabarrett

dashboard/app/assets/stylesheets/application.scss
```diff
@@ -4099,9 +4099,56 @@ a {
...
+    &-with-glow {
+      animation: clickable-text-glow-keyframes 2s ease-in-out;
+
+      &:hover {
+        text-shadow: $clickable-text-glow;
+      }
+    }
+  }
+
+  .clickable-text-0 {
```
levadadenys: Don't have full context here, but are we sure that there'll always be 5 clickable text, or that there won't be more than 5 clickable-texts?


nit: can be refactored like this, plus if we'll need more then 5 clickable texts - it'll be easier to  change that number
```
@for $i from 0 through 4 {
  .clickable-text-#{$i} {
    animation-delay: #{$i + 1}s;
  }
}
```

levadadenys (commented):
LGTM overall!
Have couple design system/styles related questions/nits

dashboard/app/assets/stylesheets/application.scss
breville: Good call.  It's a known issue, which is part of the reason this is behind a URL parameter.  It currently matches the arrow image, so both should be updated in tandem if necessary.

dashboard/app/assets/stylesheets/application.scss
breville: Five was just a best guess based on existing content.  We can bump later if there is demand!  

Nice technique; now used.

PR #59029, opened 2024-06-03
Pass down less info about enrolled workshops

bethanyaconnor:
The `EnrolledWorkshops` component fetches info about enrolled workshops, so we were passing down a bunch of unnecessary data. In fact, we were only using `workshopsAsParticipant` to check if the user enrolled in any workshops, so this change passes down that info only. I'm hoping this cuts down on unnecessary queries and data transfer.

### DIFF
--- a/apps/src/code-studio/pd/professional_learning_landing/LandingPage.jsx
+++ b/apps/src/code-studio/pd/professional_learning_landing/LandingPage.jsx
@@ -80,7 +80,7 @@ function LandingPage({
   lastWorkshopSurveyCourse,
   deeperLearningCourseData,
   currentYearApplicationId,
-  workshopsAsParticipant,
+  hasEnrolledInWorkshop,
   workshopsAsFacilitator,
   workshopsAsOrganizer,
   workshopsAsRegionalPartner,
@@ -101,7 +101,7 @@ function LandingPage({
 
   const showGettingStartedBanner =
     !currentYearApplicationId &&
-    workshopsAsParticipant?.length === 0 &&
+    !hasEnrolledInWorkshop &&
     plCoursesStarted?.length === 0;
 
   const joinedPlSectionsStyling =
@@ -438,7 +438,7 @@ LandingPage.propTypes = {
   lastWorkshopSurveyCourse: PropTypes.string,
   deeperLearningCourseData: PropTypes.array,
   currentYearApplicationId: PropTypes.number,
-  workshopsAsParticipant: PropTypes.array,
+  hasEnrolledInWorkshop: PropTypes.bool,
   workshopsAsFacilitator: PropTypes.array,
   workshopsAsOrganizer: PropTypes.array,
   workshopsAsRegionalPartner: PropTypes.array,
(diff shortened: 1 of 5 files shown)


apps/src/code-studio/pd/professional_learning_landing/LandingPage.jsx
```diff
@@ -101,7 +101,7 @@ function LandingPage({
 
   const showGettingStartedBanner =
     !currentYearApplicationId &&
-    workshopsAsParticipant?.length === 0 &&
+    !hasEnrolledInWorkshop &&
```
TurnerRiley, line 104: Thank you for simplifying this into a variable that explains what it's actually being used for!

dashboard/app/controllers/pd/professional_learning_landing_controller.rb
```diff
@@ -16,11 +16,6 @@ def index
       PLC_COURSE_ORDERING.index(enrollment[:courseName]) || PLC_COURSE_ORDERING.size
     end
 
-    workshops_as_participant_data = Pd::Enrollment.for_user(current_user).map do |enrollment|
-      workshop = enrollment.workshop
-      workshop.summarize_for_my_pl_page.merge({feedback_given: enrollments_with_pending_surveys.include?(enrollment)})
-    end
```
TurnerRiley, line 22: Love how much this is cutting down!

dashboard/app/controllers/pd/professional_learning_landing_controller.rb
```diff
@@ -36,7 +31,7 @@ def index
       last_workshop_survey_course: last_enrollment_with_pending_survey.try(:workshop).try(:course),
       summarized_plc_enrollments: summarized_plc_enrollments,
       current_year_application_id: Pd::Application::TeacherApplication.find_by(user: current_user, application_year: Pd::SharedApplicationConstants::APPLICATION_CURRENT_YEAR)&.id,
-      workshops_as_participant: workshops_as_participant_data,
+      has_enrolled_in_workshop: Pd::Enrollment.for_user(current_user).any?,
```
TurnerRiley, line 34: Now that I look at how it's being used, I'm wondering do we want to specify enrollments for the user for the current year? Or should it just be if the user has ever been enrolled before?

TurnerRiley (approved):
Looks great! Just one question

dashboard/app/controllers/pd/professional_learning_landing_controller.rb
bethanyaconnor: @dmcavoy thoughts?

dmcavoy: Is this a question of if we want to only show workshops for the current year?

bethanyaconnor: oh sorry. It's a question of whether the "Getting Started" banner should show if a user has enrolled in a workshop in a previous year but not this year. I think the spec said that we shouldn't show the banner in this case

dmcavoy: I think if they have done any workshop we don't need to show the getting started

PR #51059, opened 2023-03-31
Upgrade eslint and eslint-related packages

molly-moen:
Upgrade of various linting packages:
- upgrade from `babel-eslint` to `@babel/eslint-parser` ([package name changed](https://www.npmjs.com/package/babel-eslint))
- `eslint`
- `eslint-config-prettier`
- `eslint-plugin-babel`
- `eslint-plugin-mocha`
- `eslint-plugin-prettier`
- `eslint-plugin-react-hooks`

I also added linting packages for typescript, and set up configurations for typescript files in `.eslintrc.js`, but have not adding typescript linting to `yarn lint` yet. That will be done as a follow-up so this PR stays scoped to the various upgrades.

I kept our non-typescript files linted with the babel parser, and we will use `@typescript-eslint/parser` for our typescript files. There were a few discrepancies between the two parsers that made it hard to do an easy swap out, and it made sense to me to have typescript-specific parsing for typescript files only. However, we could revisit this if we find it confusing to have two linters (or at least unify rules if we discover major differences).

I fixed a few new linter errors that were obviously safe to fix (such as 3 global variables that were not used, and one case of a react hook dependency array that was safe). For the rest, I either added ignores for rules that would need many ignore comments, or adding ignore comments to relevant lines. We will follow up with fixes/investigations for these violations. The categories of violations are:
- [no-prototype-builtins](https://eslint.org/docs/latest/rules/no-prototype-builtins). This seems to be a new rule, that I am currently ignoring because we had many violations. The fix here seems straightforward, so I created a [task](https://codedotorg.atlassian.net/browse/SL-774) on student learning to update our usage here.
- [no-redeclare](https://eslint.org/docs/latest/rules/no-redeclare). I needed to add an exception for `builtInGlobals` to this rule. I made a [task](https://codedotorg.atlassian.net/browse/SL-775) on Student Learning to start investigating this and determine whether to follow the new rule or not.
- react-hooks/exhaustive-deps seems to have become more strict (there is no official documentation for this rule, the best I rule itself links to [this github issue](https://github.com/facebook/react/issues/14920), and now alarms on nested objects and functions that are not in the array. This one needs to be resolved on a case-by-case basis. Here is a [ticket](https://codedotorg.atlassian.net/browse/SL-777) for the student learning violations and a [thread](https://codedotorg.slack.com/archives/C045UAX4WKH/p1680817629449789) with teacher tools about fixing their violations.
[...]

PR #59031, opened 2024-06-03
use useRef in for blockRef in EmbeddedBlock

mikeharv:
Alternative/Follow-up to:
- https://github.com/code-dot-org/code-dot-org/pull/59017

On Friday we observed that a lesson plan was crashing due to a Sprite Lab block failing to render with CDO Blockly. After updating lessons to use Google Blockly, the block rendered correctly - but it rendered four times. As a quick fix, I updated the steps of the `useEffect` to check for an existing workspace SVG and remove it.

~~As an alternative, we could consider updating the dependency array to only include `blockName`. This change makes it so that `useEffect` hook runs once instead of four times.~~

In the component, `blockRef` was created using `createRef()`, which returns a new/different object each time the component is remounted. By switching to the more appropirate `useRef()`, we can keep the reference stable which avoid running the hook multiple times. We can also remove the manual SVG removal and the eslint disable rule. :tada:

## Links

https://codedotorg.atlassian.net/browse/CT-622

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/templates/codeDocs/EmbeddedBlock.jsx
+++ b/apps/src/templates/codeDocs/EmbeddedBlock.jsx
@@ -1,12 +1,12 @@
 import {Link} from '@dsco_/link';
 import PropTypes from 'prop-types';
-import React, {createRef, useEffect} from 'react';
+import React, {useRef, useEffect} from 'react';
 
 import {shrinkBlockSpaceContainer} from '@cdo/apps/templates/instructions/utils';
 import {parseElement} from '@cdo/apps/xml';
 
 export default function EmbeddedBlock({blockName, link, ariaLabel}) {
-  const blockRef = createRef();
+  const blockRef = useRef();
 
   useEffect(() => {
     if (blockName && blockRef.current) {
@@ -21,8 +21,7 @@ export default function EmbeddedBlock({blockName, link, ariaLabel}) {
       );
       shrinkBlockSpaceContainer(blockSpace, true);
     }
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [blockName]);
+  }, [blockName, blockRef]);
 
   return (
     <div>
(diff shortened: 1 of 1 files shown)


apps/src/templates/codeDocs/EmbeddedBlock.jsx
```diff
@@ -26,7 +22,7 @@ export default function EmbeddedBlock({blockName, link, ariaLabel}) {
       shrinkBlockSpaceContainer(blockSpace, true);
     }
     // eslint-disable-next-line react-hooks/exhaustive-deps
```
molly-moen: do you know if we can remove this eslint disable here? I wonder if it was the initial cause of our issues.

molly-moen: (it being ignoring the exhaustive dependencies rule)

mikeharv: If we remove the ignore rule it complains about `blockRef` being missing. (It expects both `blockName` and `blockRef`.) The rule was added when we updated the linter: https://github.com/code-dot-org/code-dot-org/pull/51059

I checked with @bethanyaconnor who originally built the feature. She said she ran into the duplicate block issue with CDO Blockly too but that adjusting the `useEffect` dependencies at the time fixed it. FWIW, we don't see duplicate blocks with CDO Blockly today, but it does look like we're infinitely creating read-only blockspaces inside the same embedded code div.

mikeharv: Found a fix with @bethanyaconnor! https://github.com/code-dot-org/code-dot-org/pull/59031/commits/5e9c54c36d12133d9b87d47fc30e36707fce3bf0

Since this component gets remounted multiple times as part of `LessonOverview`, we want to make sure we're keeping the same reference. Using `useRef` instead of `createRef` is typically the way to do that in functional components. This way, `blockRef` doesn't change so it can included as a dependency. The latest commit removes the eslint disable.

molly-moen (approved):
🎉

bethanyaconnor (approved):

PR #1295, opened 2015-02-13
Make /s/20-hour/stage/:x/puzzle/:y canonical for the K-8 script

joshlory:
https://github.com/code-dot-org/code-dot-org/pull/1227 added stages to the K-8 script so http://studio.code.org/s/20-hour/stage/2/puzzle/1 resolves.  However it currently redirects to the canonical URL: http://studio.code.org/s/1/level/2.

This change makes /s/20-hour/stage/:x/puzzle/:y canonical.  Once search engines update, the /s/1/level/:n links will be deprecated.

PR #59038, opened 2024-06-03
only find unit by name on script and script level paths

davidsbailey:
Follow-up from https://github.com/code-dot-org/code-dot-org/pull/58951#pullrequestreview-2089216860 .

Remove deprecated routes, in part so that we can make our tests a bit less dependent on test fixtures.

We stopped using script and script level paths like /s/1/... in our product a long time ago: 
* https://github.com/code-dot-org/code-dot-org/pull/1295

There is a little bit of traffic to /s/1 and other numeric script id paths, likely because this path is showing up in google search results. Making these paths return 404 should be enough to make Google stop showing these pages in search results.

## Testing story

new + updated unit tests

### DIFF
--- a/dashboard/app/controllers/script_levels_controller.rb
+++ b/dashboard/app/controllers/script_levels_controller.rb
@@ -11,6 +11,7 @@ class ScriptLevelsController < ApplicationController
   before_action :disable_session_for_cached_pages
   before_action :redirect_admin_from_labs, only: [:reset, :next, :show, :lesson_extras]
   before_action :set_redirect_override, only: [:show]
+  before_action :check_script_id_is_name, only: [:show, :lesson_extras]
 
   # Return true if request is one that can be publicly cached.
   def cachable_request?(request)
@@ -594,6 +595,13 @@ class ScriptLevelsController < ApplicationController
     if params[:script_id] && params[:no_redirect]
       VersionRedirectOverrider.set_unit_redirect_override(session, params[:script_id])
     end
+  end
+
+  # showing script levels by script id is no longer supported.
+  private def check_script_id_is_name
+    script_id = request.params[:script_id]
+    is_id = script_id.to_i.to_s == script_id.to_s
+    raise ActiveRecord::RecordNotFound if is_id
   end
 
   private def redirect_script(script, locale)
--- a/dashboard/app/controllers/script_levels_controller.rb
+++ b/dashboard/app/controllers/script_levels_controller.rb
@@ -597,7 +597,10 @@ class ScriptLevelsController < ApplicationController
     end
   end
 
-  # showing script levels by script id is no longer supported.
+  # showing script levels by script id is no longer supported. Other codepaths
+  # still need underlying helper methods to support lookup by id, so we filter
+  # out numerical ids on a per-action basis rather than removing support for
+  # ids from those methods.
   private def check_script_id_is_name
     script_id = request.params[:script_id]
     is_id = script_id.to_i.to_s == script_id.to_s
--- a/dashboard/app/controllers/script_levels_controller.rb
+++ b/dashboard/app/controllers/script_levels_controller.rb
@@ -602,6 +602,8 @@ class ScriptLevelsController < ApplicationController
   # out numerical ids on a per-action basis rather than removing support for
   # ids from those methods.
   private def check_script_id_is_name
+    # Unfortunately, scripts routes sometimes pass the name and sometimes pass
+    # the id, making params[:script_id] a misnomer when passing the name.
     script_id = request.params[:script_id]
     is_id = script_id.to_i.to_s == script_id.to_s
     raise ActiveRecord::RecordNotFound if is_id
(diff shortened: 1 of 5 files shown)


dashboard/app/controllers/script_levels_controller.rb
```diff
@@ -596,6 +597,13 @@ def self.get_script(request)
     end
   end
 
+  # showing script levels by script id is no longer supported.
```
bethanyaconnor: Do you mind adding a comment about why we have to explicitly check rather than just not searching for a unit by id? It took me a couple minutes to realize that we don't want to change `Unit.get_from_cache` to not look up by ID.

bethanyaconnor (approved):

dashboard/app/controllers/scripts_controller.rb
```diff
@@ -260,13 +261,17 @@ def get_rollup_resources
     end
   end
 
-  private def get_unit
+  private def get_unit_by_name
     unit_id = params[:id]
```
Hamms: How much work would it be to update the parameter used here to be called `name`? It was already a little weird to use the name `id` when it actually represented "either name or id", but it's even more confusing now.

Hamms (commented):
I don't want to get too deep into bikeshedding, but I'd love it if we could also update some of our variable/parameter names to make it more clear what the values are supposed to represent!

dashboard/app/controllers/scripts_controller.rb
davidsbailey: I agree that this is confusing! This would be a relatively large change, affecting the following routes: https://github.com/code-dot-org/code-dot-org/blob/710ce4a10c6ec4d741b9573dbdf75ec571440cb5/dashboard/config/routes.rb#L453-L513

given that I am already 3 layers deep on cleanup I'd like to pass on adding anything this big to my queue at this time.

davidsbailey: I've done the best I can with adding comments and renaming local variables. If anyone wants to do the bigger fix, AFAIK that would mean doing something like:
```
 resources :scripts, path: '/s/', param: :script_name_or_id do 
```
and then updating the controller actions for all the routes listed above to look for that instead of `params[:id]` or `params[:script_id]`.

Hamms (approved):
The improved comments make me feel much more comfortable with merging this change, thank you! I'm definitely still interested in pursuing the rename as a follow-up, but this LGTM as a step in the right direction.

bethanyaconnor (approved):
Thanks for additional comments!

dashboard/app/controllers/scripts_controller.rb
```diff
@@ -5,10 +5,11 @@ class ScriptsController < ApplicationController
   before_action :require_levelbuilder_mode_or_test_env, only: [:edit, :update, :new, :create]
   before_action :authenticate_user!, except: [:show, :vocab, :resources, :code, :standards]
   check_authorization
-  before_action :set_unit, only: [:show, :vocab, :resources, :code, :standards, :edit, :update, :destroy]
+  before_action :set_unit_by_name, only: [:show, :vocab, :resources, :code, :standards, :edit]
   before_action :render_no_access, only: [:show]
   before_action :set_redirect_override, only: [:show]
-  authorize_resource class: 'Unit'
+  authorize_resource class: 'Unit', except: [:update, :destroy]
+  load_and_authorize_resource class: 'Unit', only: [:update, :destroy]
```
vijayamanohararaj, line 12: @davidsbailey , it looks like this change introduced a regression as delete scripts API call uses the script name, and this is resulting in a look up by ID. I have a PR going to fix the destroy and update actions (https://github.com/code-dot-org/code-dot-org/pull/59426)

Can you help me understand why you made this change to split the flow for update and destroy actions.

davidsbailey: @vijayamanohararaj apologies, I must not have tested the destroy route. I made the overall change in order to be able to deprecate the feature where we show scripts by id, and in the process i had to categorize all actions based on whether they took a name or id.

davidsbailey: can you say more specifically what is broken in levelbuidler? a screenshot or video would be helpful.

davidsbailey: I'm not sure I am following the PR you linked to. is there a file missing from it? 

in any case, ideally, update and destroy would take an id. I'm not sure I see any reason they should take a name.

vijayamanohararaj: Here is the repro,

On levelbuilder scripts page, delete any script, click yes on the confirmation dialog
Actual:
We see below exception
image

>in any case, ideally, update and destroy would take an id. I'm not sure I see any reason they should take a name.

This is making me think that we expect the deletion to use ID and not the name. Sounds like that should be the fix in that case.

The PR right now is only a set of unit tests to assert the expected behavior, so doesn't have any changes yet. From what you are saying, it looks like the set of tests I was adding to change id to name for destroy and update should be reverted, correct?

davidsbailey: that makes sense. nice TDD!

yes, it seems like delete and update should both use the id, not the name. this makes it easier to use cancan to load_and_authorize_resource.

vijayamanohararaj: Perfect, thanks for the confirmation, I'll work on that fix.

PR #59039, opened 2024-06-03
Add Project Template Level support to Python Lab

molly-moen:
Add project template-backed level support to Python Lab (adding support for Web Lab 2 will be just be a matter of updating the level edit page). 

This was a pretty straightforward change! Just marking the level as template backed did the main portion of the work, because our channel id fetching per-level still worked, if we are moving between two project-backed levels we will fetch the same channel as the previous level and re-load the same contents. The only other piece of work was adding the icon to the header to indicate the level is project backed. In order to get this to show up properly I had to add a `z-index` to the panel container header div. I verified this did not cause any change for any lab2 lab.

As I was working on this I noticed I broke start mode in #58967. I fixed this by updating the check to be against levelId rather than channelId (all the changes in `useSource` are to fix this issue).

I think it would be similarly straightforward to bring project-backed levels to other lab2 labs. Tagging student labs as an fyi.

### Project template backed level
Screenshot taken while hovering over the workspace header icon.
Screenshot 2024-06-03 at 2 06 37 PM

## Links

- jira ticket: [CT-618](https://codedotorg.atlassian.net/browse/CT-618)

## Testing story
Tested locally.

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/codebridge/Workspace/index.tsx
+++ b/apps/src/codebridge/Workspace/index.tsx
@@ -4,16 +4,26 @@ import {FileTabs} from '@codebridge/FileTabs';
 import React from 'react';
 
 import PanelContainer from '@cdo/apps/lab2/views/components/PanelContainer';
+import ProjectTemplateWorkspaceIcon from '@cdo/apps/templates/ProjectTemplateWorkspaceIcon';
+import {useAppSelector} from '@cdo/apps/util/reduxHooks';
 
 import HeaderButtons from './HeaderButtons';
 
 import moduleStyles from './workspace.module.scss';
 const Workspace = () => {
   const {config} = useCodebridgeContext();
+  const isProjectTemplateLevel = useAppSelector(
+    state => state.lab.levelProperties?.isProjectTemplateLevel
+  );
+
+  const headerContent = (
+    <>Workspace {isProjectTemplateLevel && <ProjectTemplateWorkspaceIcon />}</>
+  );
+
   return (
     <PanelContainer
       id="editor-workspace"
-      headerContent={'Workspace'}
+      headerContent={headerContent}
       rightHeaderContent={<HeaderButtons />}
       className={moduleStyles.workspace}
     >
--- a/apps/src/codebridge/Workspace/index.tsx
+++ b/apps/src/codebridge/Workspace/index.tsx
@@ -3,6 +3,7 @@ import {Editor} from '@codebridge/Editor';
 import {FileTabs} from '@codebridge/FileTabs';
 import React from 'react';
 
+import {isProjectTemplateLevel} from '@cdo/apps/lab2/lab2Redux';
 import PanelContainer from '@cdo/apps/lab2/views/components/PanelContainer';
 import ProjectTemplateWorkspaceIcon from '@cdo/apps/templates/ProjectTemplateWorkspaceIcon';
 import {useAppSelector} from '@cdo/apps/util/reduxHooks';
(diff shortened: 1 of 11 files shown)


apps/src/codebridge/Workspace/index.tsx
```diff
@@ -4,16 +4,26 @@ import {FileTabs} from '@codebridge/FileTabs';
...
 import PanelContainer from '@cdo/apps/lab2/views/components/PanelContainer';
+import ProjectTemplateWorkspaceIcon from '@cdo/apps/templates/ProjectTemplateWorkspaceIcon';
+import {useAppSelector} from '@cdo/apps/util/reduxHooks';
 
 import HeaderButtons from './HeaderButtons';
 
 import moduleStyles from './workspace.module.scss';
 const Workspace = () => {
   const {config} = useCodebridgeContext();
+  const isProjectTemplateLevel = useAppSelector(
```
sanchitmalhotra126: We actually already have a [isProjectTemplateLevel](https://github.com/code-dot-org/code-dot-org/blob/478de6f89be9d4c3a3c55a0b8fd1e128145d3a3e/apps/src/lab2/lab2Redux.ts#L275) selector in lab2redux that you should be able to use here 🙂 (edit: this was added when we enabled project template levels for AI Chat)

molly-moen: thanks! Will update

apps/src/codebridge/hooks/useSource.ts
```diff
@@ -30,6 +30,7 @@ export const useSource = (defaultSources: ProjectSources) => {
   );
   const previousChannelIdRef = useRef<string | null>(null);
   const levelId = useAppSelector(state => state.lab.levelProperties?.id);
+  const [hasSetInitialSource, setHasSetInitialSource] = useState(false);
```
mikeharv: No suggestion, but `setHasSetInitialSource` is hard for my brain to parse.

mikeharv (commented):
Code changes look good to me.
Your description says:
> if we are moving between two project-backed levels we will fetch the same channel as the previous level and re-load the same contents.

What would happen if levels 1-2 share template A while level 3-4 share template B? Presumably, we would need to fetch the new channel when moving to level 3. I wanted to ask because the quote doesn't specify whether it needs to be the same template.

molly-moen:
> Your description says:
    if we are moving between two project-backed levels we will fetch the same channel as the previous level and re-load the same contents.
What would happen if levels 1-2 share template A while level 3-4 share template B? Presumably, we would need to fetch the new channel when moving to level 3. I wanted to ask because the quote doesn't specify whether it needs to be the same template.

@mikeharv unclear wording on my part. For each level, whether template-backed or not, we follow the same pattern. We fetch the channel id from the backend and load the code into the level. If we switch between two levels with the same template, that ends up being the same channel and code, so it's technically a little bit of wasted work. But it means that if we then go to a different template-backed level there's no problem, we just load a different channel and code.

apps/src/codebridge/hooks/useSource.ts
molly-moen: yeah I couldn't think of a better name, if you have one I'm all ears

mikeharv:
@molly-moen Thanks for confirming. That's what I was assuming would happen. Just thought I'd double-check my understanding!

mikeharv (approved):

sanchitmalhotra126 (approved):

apps/src/codebridge/hooks/useSource.ts
molly-moen: I ended up reworking this so I didn't need this check anymore. Instead of gating setting initial sources on a channel id change, we now check based on level id changing. Both start and exemplar mode don't have channel ids, but they have level ids, and it's not possible for the channel to change but not the level.

PR #59040, opened 2024-06-03
fixed student selector styling

Nokondi:
Fixed issue causing options in the student selector dropdown to be misaligned when a student has a long name.
- Set minimum width for status bubble to prevent text dropping to a second line
- Truncate last names of students with long names

Before:
image

After:
Screenshot 2024-06-03 160009

## Links
[AITT-632](https://codedotorg.atlassian.net/browse/AITT-632)

## Testing story
Manually tested

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/templates/rubrics/StudentSelector.jsx
+++ b/apps/src/templates/rubrics/StudentSelector.jsx
@@ -84,9 +84,16 @@ function StudentSelector({
           label: (
             <div className={style.studentDropdownOptionContainer}>
               <div className={style.studentDropdownOption}>
-                <BodyThreeText className={style.submitStatusText}>{`${
-                  student.name
-                } ${student.familyName || ''}`}</BodyThreeText>
+                <BodyThreeText className={style.submitStatusText}>
+                  {student.familyName
+                    ? student.familyName.length < 10
+                      ? `${student.name} ${student.familyName}`
+                      : `${student.name} ${student.familyName.substring(
+                          0,
+                          7
+                        )}...`
+                    : `${student.name}`}
+                </BodyThreeText>
                 {!!levelsWithProgress && (
                   <StudentProgressStatus
                     level={levelsWithProgress.find(
--- a/apps/src/templates/rubrics/StudentSelector.jsx
+++ b/apps/src/templates/rubrics/StudentSelector.jsx
@@ -22,6 +22,7 @@ import {reportingDataShape} from './rubricShapes';
 import style from './rubrics.module.scss';
 
 const NO_SELECTED_SECTION_VALUE = '';
+const MAX_NAME_LENGTH = 20;
 
 function StudentSelector({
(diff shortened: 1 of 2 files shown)


apps/src/templates/rubrics/StudentSelector.jsx
```diff
@@ -84,9 +84,16 @@ function StudentSelector({
           label: (
             <div className={style.studentDropdownOptionContainer}>
               <div className={style.studentDropdownOption}>
-                <BodyThreeText className={style.submitStatusText}>{`${
-                  student.name
-                } ${student.familyName || ''}`}</BodyThreeText>
+                <BodyThreeText className={style.submitStatusText}>
+                  {student.familyName
+                    ? student.familyName.length < 10
```
kobryan0619: A few questions about this solution...

1. Does this value (`10`) change depending on if a user is zoomed in or not?  Does the styling still break when zoomed in a bit?  Is that OK?
2. I am wondering about using the value of `10` here.  I feel like when Liam reviews my code he typically puts something like this in a constant at the top of the file.  How do you feel about that style choice here to avoid magic numbers?

apps/src/templates/rubrics/StudentSelector.jsx
```diff
@@ -84,9 +84,16 @@ function StudentSelector({
...
+                      ? `${student.name} ${student.familyName}`
+                      : `${student.name} ${student.familyName.substring(
```
kobryan0619: Are we missing a case here where the first name is really long and it might get cut off before it gets to the last name?  Maybe an extreme edge case?

apps/src/templates/rubrics/StudentSelector.jsx
```diff
@@ -84,9 +84,16 @@ function StudentSelector({
...
+                          0,
+                          7
```
kobryan0619: Same comment with the `7` and `10` - perhaps stored as `MAX_NAME_CHARACTERS`?

kobryan0619 (commented):
Some brief comments and questions.

apps/src/templates/rubrics/StudentSelector.jsx
Nokondi: Added a new constant to store the name length

apps/src/templates/rubrics/StudentSelector.jsx
Nokondi: I did a little research to see what the longest currently used names are around the world. Most of what I found topped out at 11 characters for first names. I resized the selector and bumped up the name length limit to 20 to account for this. There may still be some extreme cases of first names with more than 20 characters, but I would hope that the teacher could recognize that student's name even if it were cut off!

kobryan0619 (approved):
Looks good!

PR #59041, opened 2024-06-03
P20-813: Lock account linking until parent permission is granted

artem-vavilov:
## Screenshot
locked-account-linking

## Links
- JIRA ticket: [P20-813](https://codedotorg.atlassian.net/browse/P20-813)
- JIRA ticket: [P20-944](https://codedotorg.atlassian.net/browse/P20-944)
- Figma: [Account Details Locked](https://www.figma.com/file/wR0Myx8T2dHMaeeb1GAabb/CPA-Flows?type=design&node-id=950-863&mode=design&t=zdTDIzqs2lBe9lsf-4)

### DIFF
--- a/dashboard/app/controllers/omniauth_callbacks_controller.rb
+++ b/dashboard/app/controllers/omniauth_callbacks_controller.rb
@@ -11,6 +11,8 @@ class OmniauthCallbacksController < Devise::OmniauthCallbacksController
   # TODO: figure out how to avoid skipping CSRF verification for Powerschool
   skip_before_action :verify_authenticity_token, only: :powerschool
 
+  before_action :check_account_linking_lock, on: %i[connect_provider link_accounts]
+
   # Note: We can probably remove these once we've broken out all providers
   BROKEN_OUT_TYPES = [
     AuthenticationOption::CLEVER,
@@ -525,6 +527,16 @@ class OmniauthCallbacksController < Devise::OmniauthCallbacksController
 
     return errors.first unless errors.empty?
     I18n.t('auth.unable_to_connect_provider', provider: I18n.t("auth.#{auth_option.credential_type}"))
+  end
+
+  private def check_account_linking_lock
+    user = current_user || find_user_by_credential
+    return unless user
+
+    lock_reason = account_linking_lock_reason(user)
+    return unless lock_reason
+
+    redirect_to edit_user_registration_path, alert: lock_reason
   end
 
   # Determine whether to link a new LTI auth option to an existing account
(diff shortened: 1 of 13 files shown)


dashboard/test/controllers/omniauth_callbacks_controller_test.rb
```diff
@@ -1671,91 +1671,109 @@ class OmniauthCallbacksControllerTest < ActionController::TestCase
...
+          end
+        end
+      end
+    end
   end
 
-  test 'Microsoft SSO: links an LTI auth option to an existing account' do
-    DCDO.stubs(:get).with('lti_account_linking_enabled', false).returns(true)
-    lti_integration = create :lti_integration
+  describe '#link_accounts' do
```
carl-codeorg, line 1707: Nice, this really cleans up these tests.

dashboard/app/controllers/omniauth_callbacks_controller.rb
```diff
@@ -527,6 +529,16 @@ def login
...
   end
 
+  private def check_account_linking_lock
+    user = current_user || find_user_by_credential
+    return unless user
+
+    lock_reason = account_linking_lock_reason(user)
+    return unless lock_reason
+
+    redirect_to edit_user_registration_path, alert: lock_reason
```
carl-codeorg, line 539: If the user isn't logged in will this redirect them to the sign-in page? Just need to make sure that it works for users who haven't finished registration yet.

artem-vavilov: I implemented behavior similar to `OmniauthCallbacksController#connect_provider`. What should we do for `#link_account` then?
https://github.com/code-dot-org/code-dot-org/blob/abcfd4d83a47adbbaac93c787454243255ca0b07/dashboard/app/controllers/omniauth_callbacks_controller.rb#L88-L91

carl-codeorg: I think it's fine to either redirect them to the page they came from, or to the sign-in page, as long as the alert is present.

artem-vavilov:
This has already been implemented in https://github.com/code-dot-org/code-dot-org/pull/58343

dashboard/app/controllers/omniauth_callbacks_controller.rb
```diff
@@ -11,6 +11,8 @@ class OmniauthCallbacksController < Devise::OmniauthCallbacksController
   # TODO: figure out how to avoid skipping CSRF verification for Powerschool
   skip_before_action :verify_authenticity_token, only: :powerschool
 
+  before_action :check_account_linking_lock, on: %i[connect_provider link_accounts]
```
wilkie, line 14: I missed this again when I cherry-picked your changes, here. The `on` needs to be `only`. Why does it let you even do that!! 😢

PR #55650, opened 2024-01-09
[Google Blockly] Stop serializing Edit button fields. Start serializing name fields.

mikeharv:
I've noticed that whenever we save procedure callers as JSON, we are serializing the edit button field unnecessarily.

This hasn't seemed problematic, but I did find a case where it leads to some unexpected behavior. Because we were saving the translated value (e.g. 'Edit'), we were forcing the project to always use that value for that block even if it was later translated. The last value of the field is irrelevant since we can translate it freshly on load. See video:

https://github.com/code-dot-org/code-dot-org/assets/43474485/377c9072-59f6-424b-bd30-efac3d938782

Similarly, but the opposite case, we have not been serializing the name fields. By doing so, we can better align to the XML serialization that CDO Blockly supports. Doing so is also necessary if we want a levelbuilder to be able to get comprehensive block XML for a procedure block once we are on Google Blockly. While we aren't saving workspaces with XML, we still should be able to serialize individual blocks correctly, for tasks like embedding blocks in instructions.

This table demonstrates the differences between serializations of the same `growing` block, shown selected below:
image

<table>
<tbody>
<tr >
<td >
</td>
<td>CDO XML (Baseline)</td>
<td>Staging XML</td>
<td>Branch XML</td>
<td>Staging JSON</td>
<td>Branch JSON</td>
</tr>
<tr>
<td>

```



Serialization



```

</td>
<td>

```
<block type="gamelab_behavior_get">
   <mutation />
   <field name="VAR" id="growing">growing</field>
</block>
```

</td>
<td>

```
<block type="gamelab_behavior_get" id="k5dMADe{*c*nBC_m]0Zs">
   <mutation behaviorId="growing" />
   <field name="EDIT">Edit</field>
</block>
```

</td>
<td>

```
<block type="gamelab_behavior_get" id="=D?`@A[$Z#Qab%!JYEBL">
   <mutation behaviorId="growing" />
   <field name="NAME">growing</field>
</block>
```

</td>
<td>

```
{
    "type": "gamelab_behavior_get",
    "id": "k5dMADe{*c*nBC_m]0Zs",
    "extraState": {
        "behaviorId": "growing",
        "name": "growing"
    },
    "fields": {
        "EDIT": "Edit"
    }
}
```

</td>
<td>

```
{
    "type": "gamelab_behavior_get",
    "id": "=D?`@A[$Z#Qab%!JYEBL",
    "extraState": {
        "behaviorId": "growing",
        "name": "growing"
    },
    "fields": {
[...]

PR #59056, opened 2024-06-04
[Google Blockly] Support for embedded feedback blocks with functions

mikeharv:
While working on getting the modal editor to work with maze levels https://github.com/code-dot-org/code-dot-org/pull/59055 I noticed that a level's "Recommend Blocks" feature wasn't working correctly. 

For context, in older supported levels, we offer levelbuilders a place to provided XML for recommended (or required) blocks. 
image

Upon passing a level, we check whether the student's solution includes those blocks. If we can't find the blocks, we say so in the feedback (or we prevent passing in the case of required blocks). After passing a Maze level with functions using Google Blockly, I was surprised to see this pop up:

image

This is the expected modal for not using recommended blocks, but the block was not rendering. This is because when we create the embedded workspace, we don't provide an `options` object argument:
https://github.com/code-dot-org/code-dot-org/blob/mike/maze-modal-editor/apps/src/feedbackBlocks.js#L68

To fix this, I updated the function to provide an empty object as a default value. Now the block renders correctly without errors:

image

However, this is _still_ unexpected because this block is part of the start code, and it was part of my solution. Things break down when we are checking the student's given code blocks for a match with the expected recommended block, which happens here:
https://github.com/code-dot-org/code-dot-org/blob/mike/maze-modal-editor/apps/src/required_block_utils.js#L218

The function linked above checks two XML elements for equivalency, including attributes and children. If you look at the first screenshot above from levelbuilder, you'll see that the expected XML element has a single `<mutation/>` child. However, implementation of the new procedure blocks results in XML with both a mutation and a field: 
image

Initially, I implemented a fix for this where we would look for that field element and remove it before comparing elements. This felt safe enough, and it worked. We don't need both children any way as evidenced by our ability to load old projects and levels with Google Blockly. This didn't feel great though because now I wanted to understand why the serialization included the `<field/>` in the first place. 

CDO Blockly serializes all editable fields (aka titles), which does not include label fields.
https://github.com/code-dot-org/blockly/blob/v4.1.0/core/utils/xml.js#L76

Google Blockly serializes all fields that are serializable: 
https://github.com/google/blockly/blob/af907bf98ee8523ca6f8da6134e43d2af6b65b41/core/xml.ts#L149

Field labels are not normally serializable (https://developers.google.com/blockly/guides/create-custom-blocks/fields/built-in-fields/label#serialization), but for some reason our implementation of the field label on function call blocks _are_ serializable.  The reason for this comes down to a small oversight when working on behaviors for the Sprite Lab migration.

In January, I added serialization for behavior callers in Sprite Lab to get us closer to parity with CDO Blockly.
Unlike functions, the labels on behavior blocks need to be editable (or serializable): 
https://github.com/code-dot-org/code-dot-org/blob/mike/maze-modal-editor/apps/src/blockly/customBlocks/cdoBlockly/spritelabBlocks.js#L116

While doing that work https://github.com/code-dot-org/code-dot-org/pull/55650 I created a new extension that would ensure that our new procedure getters would have serializable label fields, however I unnecessarily did this for both functions and behaviors. This wasn't a problem (ie. we could still load things just fine), but it was redundant information. More importantly, the extra field tag was breaking the element equivalency check. To fix this, I simply removed the extension from our function caller block definition. (For clarity of purpose, I then renamed and moved the extension as it's only needed for behaviors.)

This results in the recommended block check succeeding which means you can now pass the level:

Click for authenticity: https://studio.code.org/blockly/media/skins/bee/win.mp3

image

This change is safe to make because we can still load projects just fine even if we have the redundant field tag in the XML. From the perspective of the recommended blocks feature, it also doesn't matter if the level sources have this field tag or not - we call `blockToDom` at the time of the check which creates new XML elements for the comparisons.

## PR Checklist:

- [ ] Code is well-commented


apps/src/blockly/customBlocks/googleBlockly/behaviorBlocks.ts
```diff
@@ -134,6 +134,17 @@ GoogleBlockly.Extensions.registerMutator(
   behaviorDefMutator
 );
 
+// This extension make the NAME fields of behavior getter blocks serializable.
+GoogleBlockly.Extensions.register(
```
molly-moen: for my understanding: why do we need this for behaviors but not functions? You said in your description it's redundant information for functions, is it not redundant for behaviors?

mikeharv: Great question. I had to look into it. The answer is we don't, but we used to. When we did the work to serialize these fields, it was to get us closer to parity with the XML structure CDO Blockly would create. As things work on staging now, you can't embed a behavior call block without a field element because we don't know where to get the name value from. This results in a behavior block without any text. Let's compare XML across Blockly versions:

CDO:
```
<block type="gamelab_behavior_get">
   <mutation />
   <field name="VAR" id="growing">growing</field>
</block>
```
The id attribute above is the behavior id and the text content is the behavior name.

Staging:
```
<block type="gamelab_behavior_get">
    <mutation behaviorId="growing" name="growing"/>
    <field name="NAME">growing</field>
</block>
```
The behavior name shows up in two places above.

At an earlier point in time, we weren't storing the behavior name in the mutation, so it wasn't redundant. Also, with either serialization, we could look at the `textContent` of the field element to find it. In the interest of simplifying and modernizing, I'm going to update this again so we stop serializing these fields. I will have to modify the deserialization hook `domToMutation` to find the name in one of these places:
1. name attribute of the mutation (new)
2. text content of the field (old)
3. use `''` only if neither can be found

I've tested locally that this works well, including with translations.

molly-moen (approved):
🎉

PR #59061, opened 2024-06-04
Add music mini-player analytics

fisher-alice:
This PR adds Amplitude analytics to the music home page's mini music player. A report is sent when a user clicks on the play button of a music project entry. The report includes user info, project name and channel.

## After update

Screencast video of signed-out user with dev console:

https://github.com/code-dot-org/code-dot-org/assets/107423305/18fa5b90-0be9-4420-a005-d1234580b72e

Screencast video of signed-in user with dev console:

https://github.com/code-dot-org/code-dot-org/assets/107423305/bebbff3d-9392-4a25-8fb9-9389fd7f9163

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/music/analytics/AnalyticsReporter.ts
+++ b/apps/src/music/analytics/AnalyticsReporter.ts
@@ -149,6 +149,13 @@ export default class AnalyticsReporter {
     });
   }
 
+  onMiniPlayerButtonClicked(buttonName: string, properties?: object) {
+    this.trackUIEvent('Mini-player button clicked', {
+      buttonName,
+      ...properties,
+    });
+  }
+
   onKeyPressed(keyName: string, properties?: object) {
     this.trackUIEvent('Key pressed', {
       keyName,
--- a/apps/src/music/analytics/AnalyticsReporter.ts
+++ b/apps/src/music/analytics/AnalyticsReporter.ts
@@ -149,15 +149,8 @@ export default class AnalyticsReporter {
     });
   }
 
-  onMiniplayerPlayButtonClicked(properties?: object) {
-    this.trackUIEvent('Miniplayer play button clicked', {
-      ...properties,
-    });
-  }
-
-  onMiniPlayerButtonClicked(buttonName: string, properties?: object) {
-    this.trackUIEvent('Mini-player button clicked', {
-      buttonName,
+  onMiniMusicPlayerButtonClicked(properties?: object) {
+    this.trackUIEvent('Mini music player play button clicked', {
       ...properties,
     });
   }
--- a/apps/src/music/analytics/AnalyticsReporter.ts
+++ b/apps/src/music/analytics/AnalyticsReporter.ts
@@ -149,6 +149,12 @@ export default class AnalyticsReporter {
     });
   }
 
+  onMiniplayerPlayButtonClicked(properties?: object) {
+    this.trackUIEvent('Miniplayer play button clicked', {
+      ...properties,
+    });
+  }
+
   onMiniPlayerButtonClicked(buttonName: string, properties?: object) {
     this.trackUIEvent('Mini-player button clicked', {
       buttonName,
--- a/apps/src/music/analytics/AnalyticsReporter.ts
+++ b/apps/src/music/analytics/AnalyticsReporter.ts
@@ -149,8 +149,9 @@ export default class AnalyticsReporter {
     });
   }
 
-  onMiniMusicPlayerButtonClicked(properties?: object) {
-    this.trackUIEvent('Mini music player play button clicked', {
+  onMiniMusicPlayerButtonClicked(buttonName: string, properties: object) {
+    this.trackUIEvent('Mini music player button clicked', {
+      buttonName,
       ...properties,
     });
   }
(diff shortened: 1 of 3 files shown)


apps/src/music/views/MiniMusicPlayer.tsx
```diff
@@ -53,57 +62,71 @@ const MiniPlayerView: React.FunctionComponent<MiniPlayerViewProps> = ({
...
-
-    // If there is a pack ID, give the player its BPM and key.
-    const currentLibrary = MusicLibrary.getInstance();
-    const packId = project.labConfig?.music.packId || null;
-    if (currentLibrary) {
-      currentLibrary.setCurrentPackId(packId);
-      playerRef.current.updateConfiguration(
-        currentLibrary.getBPM(),
-        currentLibrary.getKey()
+  const onPlaySong = useCallback(
```
fisher-alice, line 65: I was prompted to make these prettier changes causing these unintended diffs.

apps/src/music/views/MiniMusicPlayer.tsx
```diff
@@ -53,57 +62,71 @@ const MiniPlayerView: React.FunctionComponent<MiniPlayerViewProps> = ({
...
+            loadTimeMs,
+            channelId: project.id,
+          });
+        }
+      );
+
+      // Play sounds
+      playerRef.current.playSong(sequencerRef.current.getPlaybackEvents());
+      setCurrentProjectId(project.id);
+
```
fisher-alice, line 118: Unintended prettier diffs end here.

breville:
Good stuff.  Could we also log an event if the project is opened using the icon link on the right?  

Also, I'm not sure whether logging space is at a premium, but we could probably get away with just logging the channel ID rather than the project's friendly name as well.

apps/src/music/views/MiniMusicPlayer.tsx
```diff
@@ -35,15 +37,22 @@ const MiniPlayerView: React.FunctionComponent<MiniPlayerViewProps> = ({
...
+  // Setup library, workspace, and analyticsReporter on mount
   const onMount = useCallback(async () => {
     setUpBlocklyForMusicLab();
     workspaceRef.current.initHeadless();
     const library = await loadLibrary(libraryName);
     MusicLibrary.setCurrent(library);
+    analyticsReporter.startSession().then(() => {
+      analyticsReporter.setUserProperties(userId, userType, signInState);
+    });
     setIsLoading(false);
```
thomasoniii, line 54: this is probably fine, but just noting that these'll be done out of sequence - it'll fire `setIsLoading(false)`, and _then_ do the results of the promise succeeding and fire `analyticsReporter.setUserProperties(userId, userType, signInState);`

I don't know if it matters for anything going on here.

If it _does_ matter, you can just move the `setIsLoading(false)` call to the end of the `then` of the `startSession()` promise so it'll fire after `analyticsReporter.setUserProperties`.

apps/src/music/analytics/AnalyticsReporter.ts
```diff
@@ -149,6 +149,12 @@ export default class AnalyticsReporter {
     });
   }
 
+  onMiniMusicPlayerButtonClicked(properties?: object) {
+    this.trackUIEvent('Mini music player play button clicked', {
+      ...properties,
+    });
+  }
+
```
thomasoniii: non-blocking notes -

does the `properties` object really need to be destructured? I see that the rest of the file is always sending in a copy. Is amplitude's `track` function know to be destructive to the input and alter the object?

Alternatively, if we do need the destructuring (say the `track` function does damage the object), it looks like it could be moved out of the various event handlers and into `trackUIEvent`.

Changing line 175 in the `trackUIEvent` method from

```
track(eventType, payload).promise)
```
to
```
track(eventType, {...payload}).promise)
```

would be functionally identical, but we only need to duplicate the object once where it actually goes into `track` instead of all of the handlers doing it.

I'm still guessing that we _probably_ don't need to copy it at all and this is just an errant copy and paste that propagated, but keeping it centralized in the `trackUIEvent` method at least tidies it up.

Anyway, it's also veering off the PR purposes, so can definitely be tagged and bagged for future work instead.

thomasoniii (approved):
couple of tangential notes, but otherwise lgtm 🚀

apps/src/music/views/MiniMusicPlayer.tsx
fisher-alice: Looks like `isLoading` is used to set loading UI while fetching the library so I'll go ahead and just move it after `MusicLibrary.setCurrent(library);`. Thanks!

apps/src/music/analytics/AnalyticsReporter.ts
fisher-alice: Thanks Jim! Brendan asked if we could add a report on a user clicking on the 'Open project' button so that I updated code to:
```  onMiniMusicPlayerButtonClicked(buttonName: string, properties: object) {
    this.trackUIEvent('Mini music player button clicked', {
      buttonName,
      ...properties,
    });
  }
```
So I think now it makes sense to use the spread operator. Thanks!

fisher-alice:
Closing this PR and opening a duplicate based off of `staging`. Can test by using `channels` param in URL.

PR #59066, opened 2024-06-04
lab2 projects are view-only when frozen

fisher-alice:
This PR updates lab2 projects so that they are always view-only if the project is frozen (even for the owner). Projects are frozen when they are featured projects and they are 'active', i.e., they are currently being displayed in the public featured gallery at '/projects/public'.
When a project is selected as an active featured project, there is a notification icon which is displayed next to the project title. 
Screenshot 2024-06-04 at 5 01 15 PM

When clicked following dialog is displayed:
Screenshot 2024-06-04 at 5 01 23 PM

## After update

Screencast video:
This demos a project that is currently featured, e.g., 'active' featured project. In the personal projects gallery, a notification icon is displayed. And when the owner opens the project, it is view only.
Then the featured project is 'archived' and is no longer featured. At that point, the notification icon is no longer displayed and the project is editable for the project owner.

https://github.com/code-dot-org/code-dot-org/assets/107423305/8d284f93-cf57-4dcb-8a18-45d4fa562d73

## Links
[jira](https://codedotorg.atlassian.net/browse/LABS-871)

## Testing story
I tested locally a music lab (lab2) project to see that it was read-only for the project owner when actively featured and editable for the project owner when the project was no longer actively featured.
A unit test was also added.

## PR Checklist:

- [ ] Code is well-commented


apps/src/lab2/types.ts
```diff
@@ -23,6 +23,7 @@ export interface Channel {
   // Optional lab-specific configuration for this project.  If provided, this will be saved
   // to the Project model in the database along with the other entries in this interface,
   // inside the value field JSON.
+  frozen?: boolean;
```
bencodeorg: Does the comment above apply to this new property?

dashboard/app/controllers/projects_controller.rb
```diff
@@ -650,8 +650,10 @@ def limited_gallery?
 
     project = Projects.new(get_storage_id).get(params[:channel_id])
     is_owner = project[:isOwner]
+    is_frozen = project[:frozen]
 
-    return redirect_to "/projects/#{params[:key]}/#{params[:channel_id]}/edit" if is_owner && request.path.ends_with?('/view')
+    return redirect_to "/projects/#{params[:key]}/#{params[:channel_id]}/edit" if is_owner && request.path.ends_with?('/view') && !is_frozen
```
bencodeorg, line 655: Not sure how much it matters if the request ends in view or edit, but might be simpler if we could just have an initial step that always returns the view for a frozen project:

```
return redirect_to "/projects/#{params[:key]}/#{params[:channel_id]}/view" if is_frozen
```

dashboard/test/controllers/projects_controller_test.rb
```diff
@@ -534,4 +534,13 @@ def sign_in_with_request(user)
     assert_response :redirect
     assert_redirected_to "/projects/music/#{channel_id}/view"
   end
+
+  test 'on lab2 levels navigating to /edit redirects to /view if user is project owner and project is frozen' do
```
bencodeorg: Nice!

bencodeorg (approved):
lgtm!

apps/src/lab2/types.ts
bencodeorg: It does get stored inside the project json (right?), but it's not lab-specific?

dashboard/app/controllers/projects_controller.rb
fisher-alice: I tried this but page failed to load. Seems like `request.path.ends_with?` is needed.

Screenshot 2024-06-13 at 5 51 31 PM

Screenshot 2024-06-13 at 5 51 47 PM

bencodeorg: ah ok looks like it starts just redirecting in a loop...I guess we need to confirm that the URL you're redirecting to and from are different, which makes sense. Thanks for checking!

PR #59071, opened 2024-06-04
disable iPad_teacher_tools_level_types_multi

davidsbailey:
I could not get this test to pass today after 5 manual reruns. Disabling so that we do not get interrupted on every subsequent DTT this week.

the work to re-enable this is tracked in https://codedotorg.atlassian.net/browse/TEACH-1150


kobryan0619 (approved):
Looks good!

etaderhold (approved):

PR #59076, opened 2024-06-05
Add validation to Python Lab

molly-moen:
This PR covers the work to add and run validation code to python lab levels. It includes the following:
- Add the option to mark files as validation in start mode. This is done via the file dropdown in the file browser. Only one file can be a validation file, so if a validation file already exists, you need to change it to "visible" or "hidden" instead in order to mark a different file as validation.
- Update the file icons in start mode to show a flask if a file is validation.
- When a file is marked as validation in start mode, it is saved with `validation: true` in the level sources. We decided we did not need to encrypt the validation code because we have to send it as plain text over the network anyway (in Java Lab validation code is a separate encrypted field).
- When the test button is clicked, it will default to run the tests from the validation file, if one exists. Otherwise it will do what it did before, and run every test in a file that starts with `test`.

I also updated the error message we show when we have a file name conflict. I made the message similar to Java Lab; if the file name conflicts with a visible file, the message is `Filename <fileName> is already in use in this folder. Please choose a different name.`. If it conflicts with a hidden or validation file, the message is `Filename <fileName> is already in use in this folder in the level's support code. Please choose a different name.`.

For now, Web Lab 2 levels also have the same option in start mode to mark a single file as validation. We should revisit if we want to keep this functionality the same, remove the ability to mark files as validation in web lab 2, or allow multiple files be validation in Web Lab 2. Marking a file as validation in web lab 2 will hide it from users but have no extra functionality.

### Screen recording (start mode)
https://github.com/code-dot-org/code-dot-org/assets/33666587/7271626a-c760-46c8-a8d4-983f576f2b17

## Links
- [spec](https://docs.google.com/document/d/1RSb2XCSu-F78SvAy7enmHbAmHJU_p11AIDDe6MCpUMY/edit)
- jira ticket: [CT-524](https://codedotorg.atlassian.net/browse/CT-524), [CT-525](https://codedotorg.atlassian.net/browse/CT-525). I ended up doing both the start mode and run tasks together, because the run task was a very small change.

## Testing story
Tested locally

## Follow-up work
There's a separate task to gate the continue/finish button being enabled on validation passing ([CT-526](https://codedotorg.atlassian.net/browse/CT-526))

## PR Checklist:

- [ ] Code is well-commented


apps/src/codebridge/FileBrowser/index.tsx
```diff
@@ -48,11 +55,59 @@ const InnerFileBrowser = React.memo(
...
+      setFileVisibility(fileId, hidden);
+      // When we set the visibility, we also want to set the file to validation: false.
+      setFileIsValidation(fileId, false);
+    };
+
+    const startModeFileDropdownOptions = (file: ProjectFile) => {
+      // We only support one validation file per project, so if we already have one,
+      // do not show the option to mark another file as validation.
+      const options = [];
+      if (!file.validation && !hasValidationFile) {
```
mikeharv: Is it possible for the current file to be validation if there is no validation file? In other words, why is the `!file.validation` check needed?

apps/src/codebridge/FileBrowser/index.tsx
```diff
@@ -212,11 +254,13 @@ export const FileBrowser = React.memo(() => {
...
+          fileName,
+          folderId,
+          project.files
         );
-        if (existingFile) {
-          alert('File already exists');
+        if (duplicateNameError) {
+          alert(duplicateNameError);
           return;
         }
```
mikeharv, line 256: It looks like we duplicate this logic 3 times (for new/move/rename). Would it be worth making a `checkForDuplicateFilename` to use in each place?

apps/src/codebridge/FileBrowser/index.tsx
```diff
@@ -287,6 +335,25 @@ export const FileBrowser = React.memo(() => {
     [renameFile, project.files]
   );
 
+  const getDuplicateFilenameError = (
```
mikeharv: Could we document that this is intended to return `null` or a message string? Initially I wasn't sure if it was an oversight that we didn't always re-assign `message` before returning.

apps/src/codebridge/FileTabs/index.tsx
```diff
@@ -1,29 +1,22 @@
...
   const files = sortFilesByName(project.files, {mustBeOpen: true});
 
   const activeFile = getActiveFileForProject(project);
-  const isStartMode = getAppOptionsEditBlocks() === START_SOURCES;
 
   return (
     <div className="file-tabs">
       {files
-        .filter(f => !f.hidden || isStartMode)
+        .filter(f => shouldShowFile(f))
```
mikeharv, line 18: Nice refactor!

apps/src/codebridge/utils/fileUtils.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export function getFileIcon(file: ProjectFile) {
+  const isStartMode = getAppOptionsEditBlocks() === START_SOURCES;
+  let icon = 'fa-solid ';
+  if (isStartMode) {
+    icon += file.validation
+      ? 'fa-flask'
+      : file.hidden
+      ? 'fa-eye-slash'
+      : 'fa-eye';
```
mikeharv: I sometimes wonder if nested ternaries like this are harder to read than `if...else if...else`. No need to change, but thought I'd flag in case you or anyone else has an opinion.

mikeharv (commented):
Looks good, a couple questions. I didn't closely review the changes to level files.

Institutional knowledge question... How well understood do you feel like the 3-way toggle between visible/hidden/validation is by our internal users? 
Is it correct that validation files are also "hidden" from student users? Is it obvious that validation files will be hidden from students (in terms of UI/UX) or is that something you have to know? The reason I ask is that I could imagine someone thinking that selecting "Make file hidden" on a validation file makes sense (but in actuality they're just making it a non-validation file that is still hidden). I wonder if it could make sense to use a different option label for validation->hidden, like "Make standard hidden file" or "Unset validation file".

apps/src/codebridge/FileBrowser/index.tsx
```diff
@@ -287,6 +335,25 @@ export const FileBrowser = React.memo(() => {
...
+    fileName: string,
+    folderId: string,
+    projectFiles: Record<string, ProjectFile>
+  ) => {
+    let message = null;
+    const existingFiles = Object.values(projectFiles).filter(
+      f => f.name === fileName && f.folderId === folderId
+    );
+    if (existingFiles.length > 0) {
+      const existingFile = existingFiles[0];
```
thomasoniii: There's only a single file we're checking for, right? Assuming so, this should be a find:

```
const existingFile = Object.values(projectFiles).find(
  f => f.name === fileName && f.folderId === folderId
);
if (existingFile) {
```

apps/src/codebridge/utils/fileUtils.ts
thomasoniii: I definitely find nested ternaries harder to read than if/else ladders, but considering how many of them we have in the codebase I'm also not gonna demand it be replaced.

apps/src/pythonlab/pyodideRunner.ts
```diff
@@ -54,7 +53,19 @@ export async function runPythonCode(mainFile: string, source: MultiFileSource) {
...
-export async function runAllTests(source: MultiFileSource) {
-  // To run all tests in the project, we look for files that follow the regex 'test*.py'
-  await runPythonCode(getTestRunnerScript('test*.py'), source);
+export async function runAllTests(
+  source: MultiFileSource,
+  dispatch: Dispatch<AnyAction>
+) {
+  // If the project has a validation file, we just run those tests.
+  const validationFile = Object.values(source.files).filter(f => f.validation);
+  if (validationFile.length > 0) {
```
thomasoniii: same as above, this could be a `find`

```
const validationFile = Object.values(source.files).filter(f => f.validation);
```

It'll automatically stop on the first one, so the behavior is the same. It slightly masks the concept that there _might_ be more than one validation file, so it still might be useful to call that out in a comment.

apps/src/codebridge/FileBrowser/index.tsx
```diff
@@ -48,11 +55,59 @@ const InnerFileBrowser = React.memo(
...
     const {openFile, deleteFile, toggleOpenFolder, deleteFolder} =
       useCodebridgeContext();
     const isStartMode = getAppOptionsEditBlocks() === START_SOURCES;
+    const hasValidationFile = Object.values(files).some(f => f.validation);
+
+    const handleSetFileVisibility = (fileId: string, hidden: boolean) => {
+      setFileVisibility(fileId, hidden);
+      // When we set the visibility, we also want to set the file to validation: false.
+      setFileIsValidation(fileId, false);
+    };
```
thomasoniii: This is saying that we always set validation to false if we're changing the visibility. i.e., if we set visibility to true or false, we always set validation to false.

Is this correct behavior, or should it only be setting validation to false for one of the visibility cases?

thomasoniii (approved):
couple minor notes and questions, but otherwise lgtm 🚀

apps/src/codebridge/FileBrowser/index.tsx
molly-moen: This is the correct behavior, although this comment and the one below by @mikeharv are making me think of doing a bit of a refactor here.

> How well understood do you feel like the 3-way toggle between visible/hidden/validation is by our internal users?

In the model I had, if the validation boolean was `true`, the hidden boolean is ignored (the file is always hidden). 

To Mike's question, this functionally copies the behavior in Java Lab, which is the same internal user. In Java Lab things are implemented a bit differently, because validation code goes into a separate, encrypted json blob,  while the start/support files are in the same start code blob. I made all files in the same json here because we decided we didn't need to encrypt the validation. In Java Lab we also use different terminology; validation is the same, but visible files are called "starter" and hidden are called "support". I am thinking it would be good to use the same terminology here as well. Looking at these comments I will also change the two booleans (`hidden` and `validation`) into one "file type" field that can be one of `"starter" | "support" | "validation"`. This will make it clearer a file is only one of the three options.

apps/src/pythonlab/pyodideRunner.ts
molly-moen: will update!

apps/src/codebridge/utils/fileUtils.ts
molly-moen: I'm happy to change it!

apps/src/codebridge/FileBrowser/index.tsx
molly-moen: yep I'll add a comment here

apps/src/codebridge/FileBrowser/index.tsx
molly-moen: nope you're right, it's a redundant check

apps/src/codebridge/FileBrowser/index.tsx
molly-moen: I'll refactor this!

apps/src/codebridge/FileBrowser/index.tsx
molly-moen: good call, updating

mikeharv (approved):
This feels much cleaner and clearer to me. Thanks for making the changes.

PR #59078, opened 2024-06-05
P20-983: Lock-out students based on their CAP compliance state

artem-vavilov:
## Summary
Resolved the issue where we lock the user but don't reflect this in their CAP compliance state
## Links
- JIRA ticket: [P20-983](https://codedotorg.atlassian.net/browse/P20-983)


dashboard/app/controllers/application_controller.rb
```diff
@@ -351,7 +351,7 @@ def prevent_caching
       users_set_student_information_path,
     ].include?(request.path)
 
-    redirect_to lockout_path if Policies::ChildAccount.locked_out?(current_user)
+    redirect_to lockout_path if Policies::ChildAccount::ComplianceState.locked_out?(current_user)
```
daynew, line 354: What do you think about creating a method like `Services::ChildAccount.refresh` where all our time_based transition logic is contained? Otherwise all the conditional transition logic will end up living here, in the ApplicationController, for example if the user should be transitions from `nil` -> `grace_period`. We could even add a transition to from [grace_period, locked_out] -> permission_granted. Then all our unit tests can be in one place and the rest of the code base can simply call `refresh` and assume the user in in the correct CPA state.

By the way, I'm not asking for you to add all the other transition to `refresh`, yet. I'm only asking you to consider starting the pattern so we can follow it with the other transitions.

dashboard/app/controllers/application_controller.rb
```diff
@@ -334,8 +334,8 @@ def prevent_caching
     # Check that the child account policy is currently enabled.
     return unless ::Cpa.cpa_experience(request)
 
-    # Anonymous users are NOT affected by our Child Account Policy
-    return unless current_user
+    # Lock the user out if they do not comply with CAP during the lockout phase.
+    Services::ChildAccount.lock_out(current_user) if Policies::ChildAccount.lockable?(current_user)
```
daynew, line 340: What do you think about creating a method like `Services::ChildAccount.refresh` where all our transition logic is contained? Otherwise most of the conditional transition logic will end up living here, in the ApplicationController, for example if the user should be transitions from `nil` -> `grace_period`. We could even add a transition from `[grace_period, locked_out]` -> `permission_granted`. Then all our unit tests can be in one place and the rest of the code base can simply call `refresh` and assume the user in in the correct CPA state.

By the way, I'm not asking for you to add all the other transition to `refresh`, yet. I'm only asking you to consider starting the pattern so we can follow it with the other transitions.

daynew: By the way, feel free to choose a better name, I was having trouble thinking of a good one.

artem-vavilov: Correct, I plan to implement this in the upcoming tasks related to the "grace period"

artem-vavilov: Correct, I'm going to implement this in the upcoming tasks related to the "grace period"

artem-vavilov: In this task, I resolved the issue where we lock the user but don't reflect this in their compliance state

daynew (approved):

daynew (approved):
I approve of temporarily disabling the UI tests so we can unblock feature work.

PR #59086, opened 2024-06-05
[Google Blockly] Prevent error when looking for non-existent field values

mikeharv:
Play Lab (aka Studio) and Star Wars do some weird block definition sharing where the same block could look very different. For example, Play Lab lets you select which sprite to move but Star Wars only has a single player sprite. For the latter, we remove the label "actor" and the dropdown.
Play Lab `studio_move`: image
Star Wars `studio_move`: image

In the latter case we assume the value for the non-existent sprite dropdown to be "0". We check this by attempting to get the field value. If the field name is "SPRITE" (ie. for Play Lab), we look at the field. If we fail to get the field value (because the field name is `undefined`), then we just use '0'.

This breaks for Google Blockly because when `getFieldValue` calls `getField` with `undefined` an error is thrown because we expect a string. However, if we cast to a string first, it works as expected. (We try to find the value of the field called `'undefined'` which results in `null` and therefore use '0'.

I'm open to feedback about whether this feels too hacky, but, for what it's worth, it makes it so the block are fully usable. 

**Before:**
image

**After:**

image

## PR Checklist:

- [ ] Code is well-commented


apps/src/studio/studio.js
```diff
@@ -3161,17 +3161,17 @@ var registerHandlers = function (
...
-    // default title values to '0' for case when there is only one sprite
-    // and no title value is set through a dropdown
-    var titleVal1 = block.getFieldValue(nameParam1) || '0';
-    var titleVal2 = block.getFieldValue(nameParam2) || '0';
+  const blocks = Blockly.mainBlockSpace.getTopBlocks();
+  for (let x = 0; blocks[x]; x++) {
+    const block = blocks[x];
+    // default field values to '0' for case when there is only one sprite
+    // and no field value is set through a dropdown
+    const fieldVal1 = block.getFieldValue(String(nameParam1)) || '0';
```
molly-moen: could we instead check if `nameParam1` is `undefined`, and return `'0'` if it is and call `getFieldValue` if it's not?

mikeharv: Yes, that should be fine. I think we’d still need to check for a null return from getFieldValue - in case for some reason the name parameter is defined but we can’t find the field.

mikeharv: Updated to check if `nameParam1` is not a string.

molly-moen (approved):

PR #59088, opened 2024-06-05
Add time spent and last updated boxes to levels in new progress view

kobryan0619:
Adds the row for time spent and last updated.

From tech spec:
image

What it looks like in this PR:

image

Note that this has parity with the old view:
image

## Links

[Jira Ticket](https://codedotorg.atlassian.net/jira/software/projects/TEACH/boards/65?selectedIssue=TEACH-1075)
[Tech spec](https://docs.google.com/document/d/1VgHhApGUMcCFLafL6KelrkonzO3MpYgIeRn6wD-Vrlo/edit#heading=h.qi8xxyq72bct)

## Testing story

Coming in future PR

### DIFF
--- a/apps/src/templates/sectionProgressV2/LevelDataCell.jsx
+++ b/apps/src/templates/sectionProgressV2/LevelDataCell.jsx
@@ -13,6 +13,7 @@ import {commentLeft, studentNeedsFeedback} from '../progress/progressHelpers';
 import {studentLevelProgressType} from '../progress/progressTypes';
 
 import {ITEM_TYPE} from './ItemType';
+import {formatTimeSpent, formatLastUpdated} from './MetadataHelpers';
 import ProgressIcon from './ProgressIcon';
 
 import legendStyles from './progress-table-legend.module.scss';
@@ -73,6 +74,7 @@ function LevelDataCell({
   linkClassName,
   parentLevelId,
   lessonId,
+  metadataExpanded,
 }) {
   const itemType = React.useMemo(() => {
     if (expandedChoiceLevel) {
@@ -123,7 +125,7 @@ function LevelDataCell({
     }
   }, [studentLevelProgress, level, expandedChoiceLevel]);
 
-  return (
+  const levelCellUnexpanded = (
     <td
       className={classNames(
         styles.gridBox,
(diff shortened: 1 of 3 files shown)


apps/src/templates/sectionProgressV2/progress-table-v2.module.scss
kobryan0619, line 1: Note: It looks like there were some linting changes to this file... I assume that is fine.

apps/src/templates/sectionProgressV2/LevelDataCell.jsx
```diff
@@ -149,6 +151,30 @@ function LevelDataCell({
...
+        <div
+          className={classNames(styles.gridBox, styles.gridBoxMetadata, {
+            [styles.gridBoxChoiceSubLevel]: level.parentLevelId !== undefined,
+          })}
+        >
+          {!level.parentLevelId && formatTimeSpent(studentLevelProgress)}
+        </div>
+        <div
+          className={classNames(styles.gridBox, styles.gridBoxMetadata, {
+            [styles.gridBoxChoiceSubLevel]: level.parentLevelId !== undefined,
```
kobryan0619, line 168: I went back and forth on the right way to modify this for choice levels.  I decided to keep it all in this return statement rather than create a separate function that would slightly modify the return depending on if the cell is for a choice level or not.  Open to feedback on that.

MollyPeredo:
Looks great to me. Thanks, Kaitie!

dju90:
This looks great! How do the choice levels look when collapsed?

apps/src/templates/sectionProgressV2/LevelDataCell.jsx
etaderhold: Yeah it's more of an art than a science to decide when there's enough shared logic to keep two code paths together vs. splitting them up. I think you did well here.

etaderhold (approved):

kobryan0619:
@dju90 here's what it looks like opening and closing.

https://github.com/code-dot-org/code-dot-org/assets/24883357/9f0ca7b4-230c-46d7-9006-83aafd185c53

PR #59090, opened 2024-06-05
LTI account linking for roster sync users

carl-codeorg:
Currently, users are only put into the account linking flow if the LTI login route doesn't find a user based on the info from their ID token. This means that users who were provisioned via a roster sync, who actually have full users created for them, aren't sent into the flow.

This PR fixes this issue by also including LTI users who exist, but have a sign-in count of 0. These should be users who were provisioned via roster sync, but are logging in for the first time.

This also adds functionality to swap the newly-linked user into the sections that the synced user used to be in. Otherwise, the teacher would have to re-sync their classes to avoid having out of sync rosters after their students go through the account linking flow.

## Links

[JIRA](https://codedotorg.atlassian.net/browse/P20-945)

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/dashboard/app/controllers/lti_v1_controller.rb
+++ b/dashboard/app/controllers/lti_v1_controller.rb
@@ -170,6 +170,12 @@ class LtiV1Controller < ApplicationController
       destination_url = "#{target_link_uri}?#{redirect_params.to_query}"
 
       if user
+        # If this is the user's first login, send them into the account linking flow
+        if user.sign_in_count < 1 && DCDO.get('lti_account_linking_enabled', false)
+          PartialRegistration.persist_attributes(session, user)
+          render 'lti/v1/account_linking/landing', locals: {lti_provider: integration[:platform_name], email: Services::Lti.get_claim(decoded_jwt, :email)} and return
+        end
+
         sign_in user
 
         metadata = {
@@ -181,6 +187,7 @@ class LtiV1Controller < ApplicationController
           event_name: 'lti_user_signin',
           metadata: metadata,
         )
+
         # If on code.org, the user is a student and the LTI has the same user as a teacher, upgrade the student to a teacher.
         if lti_account_type == User::TYPE_TEACHER && user.user_type == User::TYPE_STUDENT
           @form_data = {
--- a/dashboard/lib/services/lti/account_linker.rb
+++ b/dashboard/lib/services/lti/account_linker.rb
@@ -23,9 +23,9 @@ module Services
         @rehydrated_user ||= find_cached_user || ::User.new_with_session(ActionController::Parameters.new, session)
       end
 
-      # The user with the auth option being linked might be in sections, if they were
-      # created via a roster sync. In this case, we need to swap the existing user into these
-      # sections to avoid the roster being in a bad state.
+      # The new user might already be in sections, if the account was created
+      # via a roster sync. In this case, we need to swap the pre-existing user
+      # into these sections to avoid the roster being in a bad state.
       private def handle_sections(user_to_remove, user_to_add)
         return if user_to_remove.sections_as_student.empty?
         user_to_remove.sections_as_student.each do |section|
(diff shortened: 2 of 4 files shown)


dashboard/app/controllers/lti_v1_controller.rb
```diff
@@ -170,6 +170,12 @@ def authenticate
       destination_url = "#{target_link_uri}?#{redirect_params.to_query}"
 
       if user
+        # If this is the user's first login, send them into the account linking flow
+        if user.sign_in_count < 1 && DCDO.get('lti_account_linking_enabled', false)
+          PartialRegistration.persist_attributes(session, user)
```
carl-codeorg, line 175: I also tried caching the user directly using `CDO.shared_cache` instead of using `PartialRegistration`. Since it was essentially doing the exact same thing but under a different cache key, it felt cleaner to just use a `PartialRegistration`. Plus this works with the rest of the code that checks for in-progress partial registrations. It also doesn't feel that far off semantically, since roster sync users who haven't logged in technically haven't finished the registration process; they still need to go through the account linking flow.

If anyone has strong preferences here, or better ideas, let me know.

stephenliang: You'll want to test this with the student email remediation DCDO as well since email will be removed in that mode

dashboard/lib/services/lti/account_linker.rb
```diff
@@ -12,13 +12,40 @@ def call
...
+      end
+
+      # The new user might already be in sections, if the account was created
+      # via a roster sync. In this case, we need to swap the pre-existing user
+      # into these sections to avoid the roster being in a bad state.
+      private def handle_sections(user_to_remove, user_to_add)
+        return if user_to_remove.sections_as_student.empty?
+        user_to_remove.sections_as_student.each do |section|
+          section.students << user_to_add
+          section.students.destroy(user_to_remove)
```
stephenliang, line 33: Looks like there's a test failure here:

```
=ERROR["test_Swaps_the_existing_user_into_the_defunct_user's_sections", "Services::Lti::AccountLinkerTest", 1075.3661866319999]
 test_Swaps_the_existing_user_into_the_defunct_user's_sections#Services::Lti::AccountLinkerTest (1075.37s)
Minitest::UnexpectedError:         ArgumentError: You are passing an instance of ActiveRecord::Base to `find`. Please pass the id of the object by calling `.id`.
            lib/services/lti/account_linker.rb:33:in `block in handle_sections'
            lib/services/lti/account_linker.rb:31:in `handle_sections'
            lib/services/lti/account_linker.rb:15:in `block in call'
            lib/services/lti/account_linker.rb:12:in `call'
            lib/services/base.rb:4:in `call'
            test/lib/services/lti/account_linker_test.rb:42:in `block in <class:AccountLinkerTest>'
```

dashboard/app/controllers/lti_v1_controller.rb
carl-codeorg: Good call, will test that out

dashboard/lib/services/lti/account_linker.rb
carl-codeorg: Will check it out - didn't encounter this locally

nicklathe (approved):
I don't have anything additional to Stephen's comments, once those are addressed :shipit:

dashboard/lib/services/lti/account_linker.rb
carl-codeorg: Not sure exactly what was causing this, but I couldn't repro this locally. Re-running the tests in drone caused other failures, but after rebasing post-mysql-8 upgrade, the tests all pass.

dashboard/app/controllers/lti_v1_controller.rb
carl-codeorg: Verified it works as intended with and without the email flag.

stephenliang (approved):

PR #59095, opened 2024-06-06
change seed order to fix intermittent contained level seeding problems

davidsbailey:
## background

see https://codedotorg.slack.com/archives/C046G4TRLEN/p1716912073888199 for context. the current seeding problems were coming from this contained level not having been seeded before the containing level was seeded: https://github.com/code-dot-org/code-dot-org/blob/e914f6fe7ec3d5268420154aac3fb2d79bca5ad1/dashboard/config/levels/custom/spritelab/k5_ai_pilot_sorting_fruit_vegetables_2024.level#L57-L58

`k5_ai_pilot_sorting_fruit_vegetables_2024` is a custom level, and `k5_ai_sorting_food_question` is a DSLDefined level of type `Multi`.

The general problem is that seeding of multi levels happens after seeding the custom levels that may contain them: https://github.com/code-dot-org/code-dot-org/blob/11de383801f91f9e86c5ce67262517c1e4820304/dashboard/lib/tasks/seed.rake#L177-L178

there has not been a root cause analysis on why seeding NOT consistently fail during setup of new development or adhoc environments (some speculation in slack). however, this PR fixes the root cause for the case when seeding DOES fail, which seems like the right thing to focus on.

## description

the solution is to seed all levels that can be contained (multi, match, etc) before seeding custom levels, while still seeding bubble choice levels after custom levels: https://github.com/code-dot-org/code-dot-org/blob/7988d5b0759e4faa788f2baf0a09c0b43d855aad/dashboard/lib/tasks/seed.rake#L177-L179

this code change can be seen in 1aa43a5493802b3a3c801001cfc39042022f072e, and the rest of the commits are mostly refactoring or other cleanup.

### collateral damage

seeding a single DSLDefined level used to be:
```
rake seed:dsls DSL_FILENAME=...
```
and is now:
```
rake seed:single_dsl DSL_FILENAME=...
```

After merge, the Tips and Tricks doc will be updated to reflect this.

## Testing story

It's hard to verify this fix because I am not consistently able to reproduce the problem. after deleting containing and contained levels, `rake seed:custom_levels` succeeds in recreating the containing level even without the contained level in the database. So, all I can really show as verification steps are that nothing new is broken:
* `rake seed:scripts` passes locally after deleting contained and containing levels
* adhoc successfully starts (including seeding all levels from scratch)

That leaves this as a speculative fix, though I'd argue it is still worth doing since both my and Molly's issues were solved by creating the multi level first.

### DIFF
--- a/dashboard/.gitignore
+++ b/dashboard/.gitignore
@@ -7,9 +7,6 @@
 # build dependency accounting stuff
 /config/scripts/.dropbox
 /config/scripts/.hints_imported
-/config/scripts/.matches_seeded
-/config/scripts/.multis_seeded
-/config/scripts/.dsls_seeded
 /config/scripts/.seeded
 
 # unit test and ui test artifacts
--- a/dashboard/lib/tasks/seed.rake
+++ b/dashboard/lib/tasks/seed.rake
@@ -232,7 +232,10 @@ namespace :seed do
 
   timed_task_with_logging child_dsls: :environment do
     DSLDefined.transaction do
+      # Allow developers to seed just one dsl-defined level, e.g.
+      # rake seed:child_dsls DSL_FILENAME=k-1_Artistloops_multi1.multi
       dsl_files = override_dsl_filenames(ENV['DSL_FILENAME'], CHILD_DSL_FILES)
+
       parse_dsl_files(dsl_files, CHILD_DSL_TYPES)
     end
   end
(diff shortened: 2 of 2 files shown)


etaderhold (approved):
LGTM

As I read through this I'm idly wondering if/when generating a legit dependency tree of items to be seeded might seem like a worthwhile investment into the robustness of the system.

davidsbailey:
> LGTM
> 
> As I read through this I'm idly wondering if/when generating a legit dependency tree of items to be seeded might seem like a worthwhile investment into the robustness of the system.

Great idea! There are other known triangular dependencies. For example, courses have to be seeded after scripts. however, lesson with resources or vocabularies (seeded as part of scripts) need the course to already have been seeded. so when you seed scripts for the first time, you get an error message like 
```
WARNING: unable to import resources into script xyz because course version is missing. This is only to be expected if the script is being seeded for the first time.
```

davidsbailey:
Finishes https://codedotorg.atlassian.net/browse/TEACH-1151

davidsbailey:
@etaderhold here is a start (updated): https://docs.google.com/drawings/d/1F0AnvWB7B9Xke2xWsxEiZ_B4hMx0CRelUIidk2pUjJ0/edit
Screenshot 2024-06-06 at 3 31 37 PM

PR #59106, opened 2024-06-06
[Design2-119] DSCO - rework Tooltip to use React Portal functionality

levadadenys:
### [Design2-119] DSCO - rework Tooltip to use React Portal functionality

* Reworked `WithTooltip` to use React Portal for rendering tooltip content outside the DOM hierarchy of the parent
  component. This allows the tooltip to be rendered outside the parent component's overflow boundaries.
* `WithTooltip` is now a recomended way to use `Tooltip` and `TooltipOverlay` components. 

Here're couple screenshots that verify the positioning issue being fixed now, keep in mind code updating these tabs hasn't been added in scope of this PR and won't be since that's a separate task.
Before fix (tooltips are there, but not visible since they 'overflow' parent container):
Знімок екрана 2024-06-06 о 16 58 44

After fix:
Знімок екрана 2024-06-06 о 16 49 48

## Links

- jira ticket: [Design2-119](https://codedotorg.atlassian.net/browse/DESIGN2-119)

## PR Checklist:

- [ ] Code is well-commented


bencodeorg:
Hey @levadadenys -- can you provide some commentary on why the accessibility issues around using React Portals noted in this thread aren't relevant / are avoided here?

At a glance, I _think_ what you're doing here may be ok from an accessibility perspective, but would love to hear your general thoughts on the issue before going deeper. The thread mentions a library called "Popper" -- curious if you considered that as an alternative? Would also be great to hear if you were able to do some testing with keyboard navigation / a screen reader to see what the experience was like. Thanks!

https://github.com/facebook/react/issues/24987

sanchitmalhotra126 (commented):
TIL about React Portals! Seems really handy, but also +1 to Ben's questions around accessibility and keyboard navigation. Changes LGTM otherwise.

levadadenys:
Hey @bencodeorg, @sanchitmalhotra126 ,

* `Tooltips` only have text and icons as a content, meaning there'll be no tab-navigable elements inside `tooltip`
* Screen reader understands the `tooltip` value because we're passing `aria-describedBy` to the element we're attaching `tooltip` to

That being said + investigating react-bootstrap and seeing that they're using similar approach as we do now for tooltips, I would say we should be good with this approach for now. (for `Tooltips` **only (!)**).

Now, more interesting and harder question are `Popover`, `Dialog` and `Modal` DSCO components that are we going to implement in the near future (especially `Popover` since it may be used in a 'tooltip like' way and have tab-navigable elemnts).

The plan here is now that we have React Portal implementation, once I'll be working on Popovers - I check the no react-portal + Popper library option, compare these two and choose what's best for us and for the web in general. Then, if needed update `Tooltip` component.

p.s. Though, I also wonder if popper library will help with the bug that is fixed in this PR, cause from the first glance - this library is using similar approach to what we were using before this PR. (will be able to tell more once be working on `Popover` component)

apps/src/componentLibrary/tooltip/Tooltip.tsx
```diff
@@ -60,31 +62,39 @@ export const TooltipOverlay: React.FunctionComponent<TooltipOverlayProps> = ({
...
-      )}
-      role="tooltip"
-    >
-      {iconLeft && <FontAwesomeV6Icon {...iconLeft} />}
-      <span className={moduleStyles.tooltipText}>{text}</span>
-      {iconRight && <FontAwesomeV6Icon {...iconRight} />}
-    </div>
-  );
-};
+const Tooltip = React.forwardRef<HTMLDivElement, TooltipProps>(
```
bencodeorg, line 65: Should we follow the pattern you've established previously to mark these as for internal use only and name it `_Tooltip`? I see you said in the docs that this could technically be used directly, but maybe we should just not advise people to use it directly?

apps/src/componentLibrary/tooltip/Tooltip.story.tsx
```diff
@@ -28,23 +28,17 @@ const SingleTemplate: StoryFn<TooltipProps> = args => (
...
     <div style={{display: 'flex', gap: '20px', flexWrap: 'wrap'}}>
       {args.components?.map(componentArg => (
-        <TooltipOverlay key={componentArg.tooltipId}>
-          <button
-            aria-describedby={componentArg.tooltipId}
-            type="button"
-            style={{margin: 0}}
-            onClick={() => {}}
-          >
+        <WithTooltip key={componentArg.tooltipId} tooltipProps={componentArg}>
```
bencodeorg, line 37: Love having these stories as documentation of how to use these components 🙌

apps/src/componentLibrary/tooltip/Tooltip.tsx
```diff
@@ -60,31 +62,39 @@ export const TooltipOverlay: React.FunctionComponent<TooltipOverlayProps> = ({
...
+      direction = 'onTop',
+      className,
+      size = 'm',
+      style = {},
+    },
+    ref
+  ) => {
+    return (
+      <div
+        ref={ref}
```
bencodeorg, line 81: Unfamiliar with `forwardRef`...what does it mean to assign the ref to an HTML element attribute `ref`?

apps/src/componentLibrary/tooltip/WithTooltip.tsx
```diff
@@ -13,16 +30,180 @@ const WithTooltip: React.FunctionComponent<WithTooltipProps> = ({
...
+      const direction = document.documentElement.dir || 'ltr'; // Default to 'ltr' if not specified
+
+      let styles: React.CSSProperties = {};
+
+      // Calculate the tooltip position based on the direction and its tail length
+      switch (tooltipProps.direction) {
+        case 'onRight':
+          styles = {
+            top: `${
+              rect.top +
```
bencodeorg: This math looks repeated / complicated with `onLeft` (maybe other repetition too?), could we extract it?

bencodeorg (approved):
Thanks for the notes about accessibility! Agreed that I think we're covered in this case where there aren't potential tab navigation issues, and looks like you've done a nice job with the aria attributes here!

apps/src/componentLibrary/tooltip/WithTooltip.tsx
levadadenys: it's `top` for `onRight` and `onLeft`,
`left` for `onTop` and `onBottom`,
`left` `ltr` for `onRight` and `left` `rtl` for `onLeft`
`left` `rtl` for `onRight` and `left` `ltr` for `onLeft`.

I agree this switch case is now bulky, though not sure about the best way of refactoring it yet. Will think more and try updating

apps/src/componentLibrary/tooltip/Tooltip.tsx
levadadenys: Yeah, was think about this also. Thought it might be better to keep it 'public' just because other internal components are internal cause they are not designed for any usage outside of Design System, while `Tooltip` can theoretically be used if user wants to create some custom positioned tooltip. At the same time, recommended way to use it is, of course, `WithTooltip`.


Though, now that you're asking about it, I'm now hesitant about leaving `Tooltip`  public

apps/src/componentLibrary/tooltip/Tooltip.tsx
levadadenys: Oh, `ref`s are awesome, though specific and not always needed feature of React! `ref` is used to access DOM element inside React code/component. Here it's used to get `Tooltip` DOM element height and width since we can't predict them in any other way.

`forwardRef` is a method of React that allows you to pass `ref` to the component as a prop. (Without forwardRef you'll be only able to set a ref to an html element inside your component, meaning you won't be able to access this dom element at current component's parent component and you won't be able to pass ref as a prop to this component)

You can read more about them here:
https://react.dev/learn/manipulating-the-dom-with-refs
https://react.dev/reference/react/forwardRef

apps/src/componentLibrary/tooltip/WithTooltip.tsx
bencodeorg: Lines 74-77 and 105-109 are exactly the same, yeah? I think just at least pulling out exactly copied math would be good, not sure if you need to refactor to fancily do all math with single helper or anything.

apps/src/componentLibrary/tooltip/Tooltip.tsx
bencodeorg: Oh cool I didn't realize you could assign a ref directly to a native HTML element, although that seems now fairly common now that I look at docs/other examples in our code. Thanks!

apps/src/componentLibrary/tooltip/Tooltip.tsx
bencodeorg: I'd vote to make it not public (ie, drive usage towards `WithTooltip` / away from custom implementations if possible), but up to you :)

levadadenys:
Updated the PR, cleaned up the tooltip positioning calculation logic to be more readable and cleaner.
PR is ready to be merged now.

Tooltips will be positioned correctly now, no matter of the size of the container of element triggering tooltip show. To do that it's required that `componentToWrap` (component we're going to attach tooltip to) will support `onFocus, onBlur, onMouseEnter, onMouseLeave` props. It's a known issue that know `WIthTooltip` won't work with components that doesn't pass these props to actual html element (e.g. Design System components right now). Planning to work on it in scope of a follow up PR. The idea here is either add support of native html element props to existing Design System components since we might actually need them anyways (?) at some point OR use `Popper` lib if that'll be easier, though the question about adding `onFocus, onBlur, onMouseEnter, onMouseLeave` and all other general html attributes/props to all the Design System components still remains active

cc @bencodeorg @sanchitmalhotra126

sanchitmalhotra126 (approved):

bencodeorg:
Much easier to understand to read the positioning math now, thanks for extracting!

Not blocking, but re: checking out what gets passed as a child to `WithTooltip`, could we do something like this (although I think throwing an error is a little harsh, maybe just a warning):

https://github.com/code-dot-org/code-dot-org/blob/38f42be85693508bdaa6c3eabf2421fe44e2e37b/apps/src/javalab/components/JavalabDropdown.jsx#L14-L26

levadadenys:
> Much easier to understand to read the positioning math now, thanks for extracting!
> 
> Not blocking, but re: checking out what gets passed as a child to `WithTooltip`, could we do something like this (although I think throwing an error is a little harsh, maybe just a warning):

@bencodeorg 
Not sure I understand. Do you mean add a check that children is `button` specifically? But the idea here is that almost everything can be wrapped up with `withTooltip`, meaning you should be able to attach `tooltip` to almost everything. Are storybook stories now are now making people assume it can be used only with buttons? I might need to do some additional work on stories and documentation then.

Merging PR now as is, will make additional changes in a follow up PR if needed.

PR #59107, opened 2024-06-06
Don't show ended workshops for Facilitators, Program Managers, and Workshop Organizers

bethanyaconnor:
This PR limits the workshops shown for facilitators, program managers, and workshop organizers to workshops that are in progress or upcoming. We've gotten multiple pieces of feedback around both the slowness of the page and the unwieldiness of having hundreds of workshops show up. There's a link to the workshop dashboard already on this page if users need to see ended workshops. As part of this, I adjusted the header for the table to be "In Progress and Upcoming Workshops" in these cases. 

This does not affect teachers looking for their enrolled workshops. Past workshops will still appear for them.

I also renamed `EnrolledWorkshopsTable` to simply `WorkshopsTable` to better reflect its expanded use. This component is not translateable but, instead of ballooning this PR even more, I made a [ticket](https://codedotorg.atlassian.net/browse/ACQ-1955) to address this in the future.

Screenshot:

Screenshot 2024-06-05 at 2 53 27 PM


apps/src/code-studio/pd/professional_learning_landing/EnrolledWorkshops.jsx
```diff
@@ -227,7 +230,7 @@ class EnrolledWorkshopsTable extends React.Component {
         </Modal>
         {this.props.workshops && (
           <section>
-            <h2>My Workshops</h2>
+            {this.props.tableHeader && <h2>{this.props.tableHeader}</h2>}
```
kelbyhawn: Non-blocking suggestion: you could use the DISCO `Heading2` component here.

kelbyhawn (approved):
This looks good to me! Thanks for making the followup translation ticket 😄

PR #59109, opened 2024-06-06
AI Component Library: consolidate AI Tutor and GenAI warning modals + styles

Erin007:
As we build out a suite of AI tools we want to ensure that they all feel like a consistent, cohesive part of the Code.org product. Similarly, we want to share UI components across projects for the sake of ease of maintenance and speed of development. 

This PR consolidates the use cases of the GenAI and AI Tutor `ChatWarningModal`.  In addition to removing redundant code,  AI Tutor now has the benefit of using a warning modal with translatable strings. 

Both use cases now see the shared modal. 

## Links

[CT-627](https://codedotorg.atlassian.net/browse/CT-627) <- I just made this up. I think there's a conversation to be had about who "owns" this work stream, how it gets labeled in Jira etc but wanted to get a foothold in so we collectively have a tangible example to start with. 

Shared AI Components are documented here: [Common Chatbot Components ](https://docs.google.com/document/d/1kUWgyT_FbjCvXN2Lc28cMF6NWNrfv62O5SlVdjEIwFQ/edit).

Designs for AI Components are in this [Figma](https://www.figma.com/design/lOLzGeTfH44a5YDS7GrzdA/AI-Tutor?node-id=2383-5497&t=xhOwOQKmd2n9BlH7-0) file. 

## Testing story

- [x] manually test AI Tutor 

Screenshot 2024-06-07 at 12 38 56 PM

- [x] manually test GenA 

Screenshot 2024-06-07 at 12 43 28 PM

## Follow-up work

- AI Tutor currently uses session storage to manage the frequency with which the chat warning modal shows. The `AITutorWarningModal` is a wrapper for the generic shared `ChatWarningModal` that handles this hide/show logic. If the other projects want something similar we may be able to also consolidate that behavior. Open question noted in [here](https://docs.google.com/document/d/1kUWgyT_FbjCvXN2Lc28cMF6NWNrfv62O5SlVdjEIwFQ/edit?disco=AAABPQnjfss). 

- There will likely be a teacher-facing version of the warning modal with slightly different wording. When that use case is more fleshed out and we know what the alternate strings will be, we can parameterize `ChatWarningModal` to have a teacher and student version depending on the view. 

## PR Checklist:

- [ ] Code is well-commented


apps/src/aiTutor/views/WarningModal.tsx
```diff
@@ -35,37 +28,7 @@ const WarningModal = () => {
...
-  return (
-    <AccessibleDialog onClose={onClose} className={style.chatWarningModal}>
-      <Heading2>Remember to chat responsibly!</Heading2>
-
-      <button type="button" onClick={onClose} className={style.xCloseButton}>
-        <i id="x-close" className="fa-solid fa-xmark" />
-      </button>
-      <hr />
-      <BodyTwoText>
-        <StrongText>All messages sent to AI Tutor are recorded</StrongText>
```
bencodeorg, line 47: Looks like there are small copy differences between the two (eg, using AI Tutor instead of AI Chat, not sure if there are others?)

Erin007: Yeah, also "Ask AI Tutor" -> "Start chatting" on the button. I think it's ultimately a Product question if we want to streamline or customize strings @samantha-code Do you have a preference?

bencodeorg (approved):
I'd vote to change these not to include any specific product (maybe Sam can help clarify here), but otherwise looks great!

https://github.com/code-dot-org/code-dot-org/blob/52659b45a40522a9b286193f17c6db8dfd007eff/apps/i18n/aichat/en_us.json#L3-L4

Erin007:
> I'd vote to change these not to include any specific product (maybe Sam can help clarify here), but otherwise looks great!
> 
> https://github.com/code-dot-org/code-dot-org/blob/52659b45a40522a9b286193f17c6db8dfd007eff/apps/i18n/aichat/en_us.json#L3-L4

+1 

@sanchitmalhotra126 also helped me uncover that we'll need to move the shared strings to a common file because Javalab doesn't currently have access to the JS bundle with the aichat locale strings. Working on updating.

apps/src/aiTutor/views/WarningModal.tsx
samantha-code: I think we should streamline unless there is a really obvious need to be different which I dont see here.

Erin007: Streamlined! To make them more generic I removed references to the specific tooling. 
Screenshot 2024-06-07 at 12 40 40 PM

apps/src/aiComponentLibrary/warningModal/ChatWarningModal.tsx
davidsbailey, line 1: I love the filesystem organization here 😀

davidsbailey (approved):
Thanks for getting the shared component started Erin! 💯 agree with not doing any of the follow-up work until we need it.

apps/src/aiComponentLibrary/warningModal/ChatWarningModal.tsx
```diff
@@ -1,14 +1,19 @@
...
-import aichatI18n from '@cdo/apps/aichat/locale';
-import moduleStyles from './chatWarningModal.module.scss';
 import {
   BodyTwoText,
   Heading3,
   StrongText,
 } from '@cdo/apps/componentLibrary/typography';
-import Button from '@cdo/apps/templates/Button';
 import AccessibleDialog from '@cdo/apps/templates/AccessibleDialog';
+import Button from '@cdo/apps/templates/Button';
```
levadadenys: nit: (optional since this PR is more about restructuring, not changing) Would be awesome to use Design System Button.

levadadenys (approved):
LGTM!
Structure looks nice! Thanks a lot for consolidating this! 
Have only 1 non-blocking nit/suggestion.

apps/src/aiComponentLibrary/warningModal/ChatWarningModal.tsx
Erin007: Updated! Thanks for the suggestion @levadadenys

PR #59111, opened 2024-06-06
[Code Bridge] Add confirmation on delete file/folder

molly-moen:
This PR adds a confirmation modal on delete file and folder in Code Bridge.
I added a generic confirmation modal to lab2 to get this to work, as we want to customize the message in the modal to include the file/folder name. I also made the confirm button text customizable, as for this modal we want it to be "delete", but other modals may want "ok" or "confirm". It was easiest to use the existing lab2 modal setup because it's already styled and set up to display over the entire app.

I confirmed this has no impact on the existing start over and skip to project modals.

### Screenshots
Delete file
Screenshot 2024-06-06 at 11 26 05 AM

Delete Folder
Screenshot 2024-06-06 at 11 25 59 AM

## Links

- jira ticket: [CT-593](https://codedotorg.atlassian.net/browse/CT-593)

## Testing story
Tested locally.

## PR Checklist:

- [ ] Code is well-commented


breville (approved):

apps/src/lab2/views/dialogs/DialogManager.tsx
```diff
@@ -83,6 +116,13 @@ const DialogManager: React.FunctionComponent<DialogManagerProps> = ({
 
 interface DialogControl {
   showDialog: (dialogType: DialogType, callback: () => void) => void;
+  showGenericDialog: (
+    dialogType: DialogType,
+    callback: () => void,
+    title: string,
+    message: string,
+    confirmText?: string
```
sanchitmalhotra126: Thoughts on just adding these as optional params to `showDialog` instead of having a separate function? I can imagine these being useful in other cases as well (for example in the Start Over dialog where we currently read the app type to determine what specific message to show, maybe instead labs can just pass in a string to use)

molly-moen: That seems fine to me, will update

apps/src/lab2/views/dialogs/GenericConfirmationDialog.tsx
```diff
@@ -0,0 +1,47 @@
...
+      </Typography>
+      <div className={moduleStyles.buttonContainer}>
+        <button
+          className={moduleStyles.cancel}
+          type="button"
+          onClick={handleCancel}
+        >
+          {commonI18n.cancel()}
+        </button>
+        <button
```
sanchitmalhotra126, line 35: Question/thoughts for later (non-blocking)
- might be good to use design system buttons here? (as well as on the other lab2 dialogs like Start Over)
- @levadadenys do we have any plans for a generic dialog component in our design system? I'm wondering if a simple dialog with title, content, confirm and cancel actions is generic enough to be part of the design sytem.

sanchitmalhotra126 (approved):

apps/src/lab2/views/dialogs/GenericConfirmationDialog.tsx
molly-moen: yeah we should probably update these to use the design system buttons, I filed a [task](https://codedotorg.atlassian.net/browse/CT-629) that is currently in the code tools backlog

thomasoniii (approved):
lgtm 🚀

mikeharv (approved):
Code changes look good! A couple product/design questions. Approving now since the answers to any of these questions could lead to follow-up tasks or nothing.

- Is there a design prototype for this confirmation? I was surprised that the delete button isn't read like it is in other non-CodeBridge places (modal function editor, animation tab, design tab)
- (Out of scope, but curious) Will there be a concept of undeletable files/folders? I'm imagining an option a start mode setting and/or always protected files/folders (e.g. main.py, /src/)
- For folder deletion, would it make sense to tell the user how many files (and subfolders) they're about to delete?
- Can these actions be undone?

molly-moen:
@mikeharv good questions! The main goal of this PR was to prevent us/curriculum from accidentally deleting files/folders, and we can iteratively improve on it.

>  Is there a design prototype for this confirmation? I was surprised that the delete button isn't red like it is in other non-CodeBridge places (modal function editor, animation tab, design tab)

There is no design for this. I considered making the button red but that was another configuration to add for something we had no design for in the first place. I already filed a task to change the buttons to design system buttons, so if we want color configuration we can add it as part of that task.

> (Out of scope, but curious) Will there be a concept of undeletable files/folders? I'm imagining an option a start mode setting and/or always protected files/folders (e.g. main.py, /src/)

So far I haven't heard of such a requirement! We don't have this in Java Lab.
    
> For folder deletion, would it make sense to tell the user how many files (and subfolders) they're about to delete?

It's something we could do as a follow up!
    
> Can these actions be undone?

Only via version history (once we have it) or start over.

molly-moen:
@mikeharv I made a polish task to cover your questions [CT-630](https://codedotorg.atlassian.net/browse/CT-630)

PR #59117, opened 2024-06-06
[Google Blockly] cleanup level methods

mikeharv:
As of the Sprite Lab migration, all Blockly labs in the Game Lab "branch" of our level models structure are now on Google Blockly. Rather than continue to define the `uses_google_blockly?` method in each Game Lab child, we can do this just one in the parent. See diagram:
image

Poetry, Dance, and Sprite Lab will now inherit the same method from the Game Lab class. It may be worth pointing out that Game Lab proper level use Droplet and not Blockly (all Droplet levels are technically considered "Blockly"). This method doesn't cause any problems since it's only used to inform which version of Blockly we would use if it was needed. The GamelabJr class also explicitly states that Droplet isn't used (including for its children Poetry and Dancelab).

This is just clean up; there are no functional changes.

## Testing story

I confirmed that Sprite Lab, Dance, and Poetry project still load with Google Blockly. I also confirmed that Game Lab projects still load without error and that Blockly never gets set to the global scope.

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/dashboard/app/models/levels/dancelab.rb
+++ b/dashboard/app/models/levels/dancelab.rb
@@ -57,10 +57,6 @@ class Dancelab < GamelabJr
     )
   end
 
-  def uses_google_blockly?
-    true
-  end
-
   def common_blocks(type)
   end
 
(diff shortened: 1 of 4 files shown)


molly-moen (approved):
Nice!

PR #59118, opened 2024-06-06
fix(lms): allow duplicate user accounts to be created for LMS users

stephenliang:
Currently, Code.org allows duplicate accounts with the same email address to be created due to the partial registration cache allowing the rehydration of invalid users. When the student email remediation DCDO was enabled, the rehydration of the email address allowed for the email uniqueness validation to occur (which was previously bypassed) and prevented LMS accounts from being created. 

Per the [slack discussion](https://codedotorg.slack.com/archives/C032LKQL53Q/p1717613645457869), we will be allowing duplicate accounts to be created for LMS users that are unlinked.

## Links

- [slack thread](https://codedotorg.slack.com/archives/C032LKQL53Q/p1717613645457869)
- [jira ticket](https://codedotorg.atlassian.net/browse/P20-984)

## Testing story

1. Added unit tests for this scenario
2. Created an email account user, then went to Canvas and created a fresh account with the same user. Previously, this would error out but now a duplicate account is created.

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/dashboard/app/models/user.rb
+++ b/dashboard/app/models/user.rb
@@ -816,6 +816,8 @@ class User < ApplicationRecord
   def email_and_hashed_email_must_be_unique
     # skip the db lookup if we are already invalid
     return if errors.present?
+    # allow duplicate accounts to be created for LMS users that are unlinked
+    return if authentication_options.length == 1 && authentication_options.first&.credential_type == AuthenticationOption::LTI_V1
 
     if ((email.present? && (other_user = User.find_by_email_or_hashed_email(email))) ||
         (hashed_email.present? && (other_user = User.find_by_hashed_email(hashed_email)))) &&
--- a/dashboard/app/models/user.rb
+++ b/dashboard/app/models/user.rb
@@ -817,7 +817,7 @@ class User < ApplicationRecord
     # skip the db lookup if we are already invalid
     return if errors.present?
     # allow duplicate accounts to be created for LMS users that are unlinked
-    return if authentication_options.length == 1 && authentication_options.first&.credential_type == AuthenticationOption::LTI_V1
+    return if authentication_options.length == 1 && authentication_options.first&.lti?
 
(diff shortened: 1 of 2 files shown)


dashboard/app/models/user.rb
```diff
@@ -816,6 +816,8 @@ def presence_of_email_or_hashed_email
   def email_and_hashed_email_must_be_unique
     # skip the db lookup if we are already invalid
     return if errors.present?
+    # allow duplicate accounts to be created for LMS users that are unlinked
+    return if authentication_options.length == 1 && authentication_options.first&.credential_type == AuthenticationOption::LTI_V1
```
carl-codeorg: AuthenticationOption has an `lti?` method you can use like:

```
return if authentication_options.length == 1 && authentication_options.first&.lti?
```

carl-codeorg (approved):
LGTM, just one small cleanup suggestion

dashboard/app/models/user.rb
wilkie: Is it possible for an account to having just an LTI auth option and then get another auth option (of any kind) which would suddenly forever invalidate the User?

wilkie: Is it possible for an account to having just an LTI auth option and then get another auth option (of any kind) which would suddenly forever invalidate the User?

stephenliang: This would be happening today if it was the case (this PR relaxes the rule instead of making it stricter). If the user gets _another_ account linked it might not lead to the situation where the user is erroring (because this check only occurs if changing an email). However, I do think this is a case where we need to think about it further longer term and we'll need to discuss the account linking design further as a team. Fundamentally, this PR won't affect lock out behavior due to an invalid state. However it will allow duplicates to be created in the system (which appears to be what we are looking to do)

nicklathe (approved):

PR #59121, opened 2024-06-06
[Levelbuilder] Use key for levels list

mikeharv:
At `/levels`, we show a filterable table of all levels by name. Many old levels of type "Blockly" share the same name "blockly". There are many pages of these levels which makes the list impossible to navigate.
image

In other places, such as the Extra links box, we use the level key instead of name to show a more helpful label. 
image

I propose that we do the same thing to the levels list to make the list possible to navigate when perusing Blockly levels.

image

For most (all?) other level types, there's no difference between `name` and `key` so these changes shouldn't be noticeable except for Blockly levels. Note that child types related to Blockly (e.g. GamelabJr) have descriptive names. These also won't be affected.
image

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/dashboard/app/views/levels/_list.html.haml
+++ b/dashboard/app/views/levels/_list.html.haml
@@ -88,9 +88,9 @@
         -# Level name column, linked to view action if permitted
         %td
           - if can? :show, level
-            = link_to level.name, level, title: t('crud.show')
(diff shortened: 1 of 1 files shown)


davidsbailey (approved):
💯

molly-moen (approved):

PR #59122, opened 2024-06-06
flip star wars to google blockly

mikeharv:
Flips Star Wars Event (droid game) and Star Wars Grid (BB-8 maze) Levels to Google Blockly 🎉 

image

The original (engineer-owned) Star Wars levels work a bit differently because they are actually just `Blockly` levels (rather than `Studio` or `StarWarsGrid` levels). To make sure these flip, I've added a skin check to the Blockly model. Due to NDA agreements at the time, `'hoc2015'` and `'hoc2015x` were used in code to represent the Star Wars skins. These skin are not available in other non-migrated labs (e.g. Maze, Artist).
Another type of Star Wars level `StudioEC` ("Studio Edit Code") also exists, but this does not need to be migrate because it uses Droplet instead of Blockly.

## Testing story

The current scripts that feature Star Wars have all been manually tested:
* `/s/starwarsblocks/lessons/1/levels/1`
* `/s/coursea-2023/lessons/6/levels/1`
* `/s/coursec-2023/lessons/8/levels/1`
* `/s/coursed-2023/lessons/7/levels/1`

The project created in final Free Play level in Event lessons can be shared with a `/c/` link. If remixed, we'll create a new project at `/projects/starwarsblocks_hour`. Projects can also be created from the gallery at `/projects/starwarsblocks`.

## Deployment strategy

As Blockly flips go, this one is considered low risk. Both level types have a relatively small number of blocks, all of which have been confirmed to be usable and generate correct code. Furthermore, as progress in all HoC/CSF levels is saved as XML we can potentially revert the flip if needed with minimal disruption.

It is likely this change will result in Eyes Tests failures. These should only occur in the Blockly area (under the "Blocks" and "Workspace" headings. Failing tests that show changes to the blocks or toolbox area should have their baselines updated. I'll work with the assigned DOTD to communicate/assist with this.

## PR Checklist:

- [ ] Code is well-commented


dashboard/app/models/levels/blockly.rb
```diff
@@ -89,6 +89,15 @@ class Blockly < Level
   # DCDO key for turning this feature on or off.
   BLOCKLY_I18N_IN_TEXT_DCDO_KEY = 'blockly_i18n_in_text'.freeze
 
+  def uses_google_blockly?
+    migrated_skins = ["hoc2015", "hoc2015x"]
+    skin = properties['skin']
+    if migrated_skins.include?(skin)
```
molly-moen, line 95: nit: could we do `return migrated_skins.include?(skin)` instead?

molly-moen (approved):
Looks good!

PR #59123, opened 2024-06-06
fix(statsig): add statsig stable id to the rails session

stephenliang:
Normally, statsig requires a user id to track a user's journey. However, anonymous users will inherently not have a user id because the user has not yet been created. Instead, use the 'stable id' feature which is a random uuid generated alongside the rails session. The reason why a new uuid is generated is to prevent leaking the rails session id to external parties.

A new method `log_statsig_event_with_session` has been created to ensure most calls are made with the user and only in unique circumstances (such as tracking an anonymous user) should this method be used.

## Links

- [jira ticket](https://codedotorg.atlassian.net/browse/P20-986)

## Testing story

Screenshot from 2024-06-06 14-42-05

## Follow-up work

- [change frontend statsig sdk to initialize with the same stable id as the backend](https://codedotorg.atlassian.net/browse/P20-985)

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/dashboard/lib/metrics/events.rb
+++ b/dashboard/lib/metrics/events.rb
@@ -48,6 +48,7 @@ module Metrics
 
       def log_event_with_session(session:, event_name:, event_value: nil, metadata: {})
         event_value = event_name if event_value.nil?
+        managed_test_environment = CDO.running_web_application? && CDO.test_system?
         statsig_user = StatsigUser.new({'userID' => session[:statsig_stable_id]})
 
         if CDO.rack_env?(:development)
--- a/dashboard/lib/metrics/events.rb
+++ b/dashboard/lib/metrics/events.rb
@@ -34,7 +34,7 @@ module Metrics
         if CDO.rack_env?(:development)
           log_event_to_stdout(user: user, event_name: event_name, event_value: event_value, metadata: metadata, enabled_experiments: enabled_experiments)
         elsif CDO.rack_env?(:production) || managed_test_environment
-          log_event_with_statsig(user: user, event_name: event_name, event_value: event_value, metadata: metadata, enabled_experiments: enabled_experiments)
+          log_statsig_event_with_cdo_user(user: user, event_name: event_name, event_value: event_value, metadata: metadata, enabled_experiments: enabled_experiments)
         else
           # We don't want to log in other environments, just return silently
           return
@@ -46,11 +46,21 @@ module Metrics
       )
       end
 
+      def log_event_with_session(session:, event_name:, event_value: nil, metadata: {})
+        statsig_user = StatsigUser.new({'userID' => session[:statsig_stable_id]})
+
+        log_statsig_event(statsig_user: statsig_user, event_name: event_name, event_value: event_value, metadata: metadata)
+      end
+
       # Logs an event to Statsig
-      private def log_event_with_statsig(user:, event_name:, event_value:, metadata:, enabled_experiments:)
-        # Re-initialize Statsig to make sure it's avaiable in current thread.
-        Cdo::StatsigInitializer.init
+      private def log_statsig_event_with_cdo_user(user:, event_name:, event_value:, metadata:, enabled_experiments:)
         statsig_user = build_statsig_user(user: user, enabled_experiments: enabled_experiments)
+        log_statsig_event(statsig_user: statsig_user, event_name: event_name, event_value: event_value, metadata: metadata)
+      end
+
+      private def log_statsig_event(statsig_user:, event_name:, event_value:, metadata:)
+        # Re-initialize Statsig to make sure it's available in current thread.
+        Cdo::StatsigInitializer.init
         Statsig.log_event(statsig_user, event_name, event_value, metadata)
       end
 
--- a/dashboard/lib/metrics/events.rb
+++ b/dashboard/lib/metrics/events.rb
@@ -47,9 +47,22 @@ module Metrics
       end
 
       def log_event_with_session(session:, event_name:, event_value: nil, metadata: {})
+        event_value = event_name if event_value.nil?
         statsig_user = StatsigUser.new({'userID' => session[:statsig_stable_id]})
 
-        log_statsig_event(statsig_user: statsig_user, event_name: event_name, event_value: event_value, metadata: metadata)
+        if CDO.rack_env?(:development)
+          log_event_to_stdout_with_session(session: session, event_name: event_name, event_value: event_value, metadata: metadata)
+        elsif CDO.rack_env?(:production) || managed_test_environment
+          log_statsig_event(statsig_user: statsig_user, event_name: event_name, event_value: event_value, metadata: metadata)
+        else
+          # We don't want to log in other environments, just return silently
+          return
+        end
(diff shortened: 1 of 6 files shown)


lib/cdo/statsig.rb
```diff
@@ -10,8 +10,8 @@ def self.init
       # environments. Anything else that runs on the Test server, as well as
       # code that executes in the continous integration builds, will run in local_mode.
       # This limits the number of metrics we emit, thereby lowering our credit usage.
-      local_mode = CDO.rack_env?(:production) || managed_test_environment ? false : true
-      options = StatsigOptions.new({'tier' => CDO.rack_env}, network_timeout: 10, logging_interval_seconds: 30, local_mode: local_mode)
+      CDO.rack_env?(:production) || managed_test_environment ? false : true
+      options = StatsigOptions.new({'tier' => CDO.rack_env}, network_timeout: 10, logging_interval_seconds: 30, local_mode: false)
```
wilkie: Is this intended?

stephenliang: Great catch!

dashboard/lib/metrics/events.rb
```diff
@@ -33,8 +33,9 @@ def log_event(user: nil, event_name:, event_value: nil, metadata: {}, get_enable
 
         if CDO.rack_env?(:development)
           log_event_to_stdout(user: user, event_name: event_name, event_value: event_value, metadata: metadata, enabled_experiments: enabled_experiments)
+          log_statsig_event_with_cdo_user(user: user, event_name: event_name, event_value: event_value, metadata: metadata, enabled_experiments: enabled_experiments)
```
nicklathe: This line should be removed right? Left over from local dev testing?

stephenliang: Thanks, this might be a good data point for the upcoming work to add some sort of flag to enable development testing of statsig to avoid committing these!

dashboard/test/integration/session_cookie_test.rb
```diff
@@ -12,12 +12,6 @@ class SessionCookieTest < ActionDispatch::IntegrationTest
     assert cookies['_learn_session_test']
   end
 
-  test 'no cookies if you do not do anything' do
-    get '/'
-
-    assert_nil cookies['_learn_session_test']
-  end
-
```
stephenliang, line 20: This was removed because the rails session is now instantiated even on the home page to track stable id. This was true even before this change because if you hit `/` it would redirect you to `/sign_in` and that will vend you a cookie.

dashboard/lib/metrics/events.rb
```diff
@@ -46,11 +46,21 @@ def log_event(user: nil, event_name:, event_value: nil, metadata: {}, get_enable
       )
       end
 
+      def log_event_with_session(session:, event_name:, event_value: nil, metadata: {})
```
nicklathe, line 49: Thinking through this, I'm wondering if this would be cleaner if we make this a private method, and conditionally call this or the `log_statsig_event_with_cdo_user` in the main `log_event` method depending on if `user` is nil?

dashboard/lib/metrics/events.rb
```diff
@@ -46,11 +46,21 @@ def log_event(user: nil, event_name:, event_value: nil, metadata: {}, get_enable
...
+        statsig_user = StatsigUser.new({'userID' => session[:statsig_stable_id]})
```
nicklathe, line 52: Or, what if we just move this line down to 76, where we initialize an empty string for a user. Then if the user is nil, we just init a statsig user with the :statsig_stable_id

dashboard/lib/metrics/events.rb
stephenliang: I wrote up some of the pros/cons I can think of, I'm leaning toward Option 1 to avoid possible bugs if someone forgets to input `user` and the analytics get chained incorrectly (some to stable id, some to user id).

### Option 1: Create a `log_event_with_session` method, the caller must use this if logging with user.

Pros: 
* Explicitly known that the session is used to store
* Oddball compared with `log_event`, so little more clear that this is special

Cons:
* Extra method, might be a bit confusing

### Option 2: Add `session` field to the parameter

Pros:
* Less methods to consider, always use `log_event` to log

Cons:
* If the caller accidentally forgets `user`, it will log with the stable id resulting in possible bugs where the metrics aren't chained correctly

nicklathe: If we go with option 1, I think this new method needs to handle the same cases as the `log_event` method, namely:

* set event_value to event_name if nil
* log_event_to_stdout if env == :development
* It should probably have exception handling too?

nicklathe: I think I'm comfortable with option 2, where we default to the stable id if user is nil. This should be easy to catch cases where `user` is accidentally left out, and can be corrected. I think the simplicity of this option outweighs the potential downsides.

As a followup, we can add some documentation (and update the commented usage at the top of the events.rb file) to clarify that `user` should be passed in whenever it is available.

stephenliang: Updated to use the same logic as above (might be some opportunity to clean up after we get Statsig working for anonymous users)

```
Logging Event: {"user_id":"1a02b095-9308-46ac-b1c1-7e254c74d333","event_name":"clever-begin-auth","event_value":"clever-begin-auth","metadata":{}}
Logging Event: {"user_id":"1a02b095-9308-46ac-b1c1-7e254c74d333","event_name":"clever-callback","event_value":"clever-callback","metadata":{"study":"account-sign-up-v5","study_group":"not-in-study"}}
```

dashboard/lib/metrics/events.rb
```diff
@@ -46,11 +46,34 @@ def log_event(user: nil, event_name:, event_value: nil, metadata: {}, get_enable
       )
       end
 
+      def log_event_with_session(session:, event_name:, event_value: nil, metadata: {})
+        event_value = event_name if event_value.nil?
+        statsig_user = StatsigUser.new({'userID' => session[:statsig_stable_id]})
+
+        if CDO.rack_env?(:development)
+          log_event_to_stdout_with_session(session: session, event_name: event_name, event_value: event_value, metadata: metadata)
+        elsif CDO.rack_env?(:production) || managed_test_environment
```
nicklathe, line 56: You'll need to add `managed_test_environment = CDO.running_web_application? && CDO.test_system?` in this method similar to `log_event`.

Think it would be worth abstracting this and the `event_value = event_name if event_value.nil?` into two private methods that could be called in each of the log functions?

stephenliang: I was thinking this would be a good candidate to either use a composition pattern or inheritance pattern. However I wanted to make sure the whole setup works well (with data flowing) before coming back and refactoring it. Thanks for catching this, I updated it :)

nicklathe: Sounds good, yeah there's some opportunity to make the code a bit more dry, but ultimately I appreciate the explicit nature of the two different log methods. As you said, we can address a refactor/cleanup later once we verify it's working.

nicklathe (approved):

PR #59124, opened 2024-06-06
Make a read-only version of Code Bridge

molly-moen:
This PR enables Code Bridge to be in read only mode. It decides if it is in read-only mode based on the lab2 [isReadOnlyWorkspace](https://github.com/code-dot-org/code-dot-org/blob/047e757d09ffa022f28bf95cc79b338ca6f2de94/apps/src/lab2/lab2Redux.ts#L259) selector. The logic for this currently is if the level has a channel and the user is the owner of the channel, we are not in read-only mode. I updated this logic to also check that we are not in start mode (which has no channel) This turns exemplar and share modes in Code Bridge into view-only, which is what we want. As we add special level types (contained and submittable) we will need to update this selector.

The following happens in read only mode:
- The user cannot edit code
- The user cannot start over (button is gone)
- The user cannot do any file or folder operations (create, move, rename, delete) (the file menu is gone)
- The user can still open and close files in the editor, and expand and collapse folders in the file browser.
- We don't attempt to save to the project manager

Even after updating `useSource` to not save to the project manager while in read only mode, I saw a case where the project manager still attempted to save on reload (due to [this logic](https://github.com/code-dot-org/code-dot-org/blob/047e757d09ffa022f28bf95cc79b338ca6f2de94/apps/src/lab2/projects/ProjectContainer.tsx#L108-L116)). This causes a "you have unsaved changes" pop-up to appear every time you try to reload the page, because for some reason we think we have unsaved changes, but then when we try to save them we can't because the user doesn't own the channel. In order to avoid this I added a check in the project container to not execute that code if we are in read-only mode. This should not cause any issues besides avoiding an annoying pop-up.

## Links
- jira ticket: [CT-455](https://codedotorg.atlassian.net/browse/CT-455) covers making read-only mode. This work automatically completed [CT-608](https://codedotorg.atlassian.net/browse/CT-608), make exemplar view-only.

## Testing story
Tested that exemplars and share mode are read-only for python lab and exemplars are read only in web lab 2 (we haven't set up standalone projects in web lab 2 yet). I also verified other lab2 labs are unaffected.

## Follow-up work
This work was a prerequisite to getting contained and submittable levels working.

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/codebridge/Workspace/HeaderButtons.tsx
+++ b/apps/src/codebridge/Workspace/HeaderButtons.tsx
@@ -21,7 +21,11 @@ const WorkspaceHeaderButtons: React.FunctionComponent = () => {
     }
   }, [dialogControl, resetProject]);
 
-  return !isReadOnly ? (
+  if (isReadOnly) {
+    return null;
+  }
+
+  return (
     <div>
       <Button
         icon={{iconStyle: 'solid', iconName: 'refresh'}}
@@ -32,7 +36,7 @@ const WorkspaceHeaderButtons: React.FunctionComponent = () => {
         size={'xs'}
       />
     </div>
-  ) : null;
+  );
 };
 
 export default WorkspaceHeaderButtons;
--- a/apps/src/codebridge/Workspace/HeaderButtons.tsx
+++ b/apps/src/codebridge/Workspace/HeaderButtons.tsx
@@ -1,16 +1,19 @@
 import React, {useCallback, useContext} from 'react';
 
 import Button from '@cdo/apps/componentLibrary/button';
+import {isReadOnlyWorkspace} from '@cdo/apps/lab2/lab2Redux';
 import {
   DialogContext,
   DialogType,
 } from '@cdo/apps/lab2/views/dialogs/DialogManager';
+import {useAppSelector} from '@cdo/apps/util/reduxHooks';
 
 import {useCodebridgeContext} from '../codebridgeContext';
 
 const WorkspaceHeaderButtons: React.FunctionComponent = () => {
   const dialogControl = useContext(DialogContext);
   const {resetProject} = useCodebridgeContext();
+  const isReadOnly = useAppSelector(isReadOnlyWorkspace);
 
   const onClickStartOver = useCallback(() => {
     if (dialogControl) {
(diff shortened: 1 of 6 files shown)


sanchitmalhotra126 (approved):
Lab2 changes LGTM!

apps/src/codebridge/Workspace/HeaderButtons.tsx
```diff
@@ -29,7 +32,7 @@ const WorkspaceHeaderButtons: React.FunctionComponent = () => {
         size={'xs'}
       />
     </div>
-  );
+  ) : null;
```
thomasoniii: Doesn't need to be changed, but I think this'd read more clearly with a negated early return instead:

```
if (isReadOnly) {
  return null
}
...
return <div><Button>...etc
```

apps/src/codebridge/hooks/useSource.ts
```diff
@@ -29,12 +33,25 @@ export const useSource = (defaultSources: ProjectSources) => {
...
+
+  const setSourceHelper = useMemo(
+    () => (newProjectSource: ProjectSources) => {
+      if (isReadOnly) {
+        // Don't attempt to save in read-only mode.
+        dispatch(setProjectSource(newProjectSource));
+      } else {
+        dispatch(setAndSaveProjectSource(newProjectSource));
+      }
+    },
```
thomasoniii: again, non-blocking syntactic sugar, but I think it reads a little more clearly like this:

```
const saveFunction = isReadOnly ? setProjectSource : setAndSaveProjectSource;
dispatch(saveFunction(newProjectSource);
```

Of course, if there's a change that the arguments may drift apart, then it should definitely stay the if/else

thomasoniii (approved):
minor non-blocking formatting type suggestions, but otherwise lgtm 🚀

mikeharv (commented):
It looks like this break start mode, and I think this is why:
> The logic for this currently is if the level has a channel and the user is the owner of the channel, we are not in read-only mode.

If we're editing start sources, there isn't a channel, correct?

molly-moen:
> It looks like this break start mode, and I think this is why:
> 
> > The logic for this currently is if the level has a channel and the user is the owner of the channel, we are not in read-only mode.
> 
> If we're editing start sources, there isn't a channel, correct?

@mikeharv good catch 🤦‍♀️ will update

apps/src/codebridge/Workspace/HeaderButtons.tsx
molly-moen: updated!

apps/src/codebridge/hooks/useSource.ts
molly-moen: updated!

mikeharv (approved):

PR #59131, opened 2024-06-07
CAP State Policies now dynamic

daynew:
The Child Account Policy configuration for each US State was being frozen as a constant. This is a problem because some of the values are sourced from DCDO, so we expect these configurations to change if we change the DCDO value.

This fixes a few minor issues we noticed related to our Child Account Policy (CAP) code. 
* The CPA State Policies configuration now checks DCDO with every request.
* Use DateTime instead of Date because it can handle timezone names (We use MST for Colorado timezone).
* If an engineer uses the Query String or Cookie override for the `cpa_schedule` config, then it is a JSON serialized string and needs to be deserialized into an object.

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/dashboard/lib/policies/child_account.rb
+++ b/dashboard/lib/policies/child_account.rb
@@ -27,19 +27,7 @@ class Policies::ChildAccount
   # P20-937 - We had a regression which we have chosen to mitigate by allowing
   # accounts created before the below date to have their lock-out delayed until
   # the CAP policy is set to lockout all users.
-  CPA_CREATED_AT_EXCEPTION_DATE = Date.parse('2024-05-26T00:00:00MST')
-
-  # The individual US State child account policy configuration
-  # max_age: the oldest age of the child at which this policy applies.
-  # start_date: the date on which this policy first went into effect.
-  STATE_POLICY = {
-    'CO' => {
-      name: 'CPA', # Colorado Privacy Act
-      max_age: 12,
-      lockout_date: DateTime.parse(DCDO.get('cpa_schedule', {Cpa::ALL_USER_LOCKOUT => Cpa::ALL_USER_LOCKOUT_DATE.iso8601})[Cpa::ALL_USER_LOCKOUT]),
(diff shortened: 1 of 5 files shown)


dashboard/lib/policies/child_account.rb
```diff
@@ -65,7 +53,7 @@ def self.compliant?(user)
   # state missing
   # We use Colorado as it is the only start date we have for now
   def self.user_predates_state_collection?(user)
-    user.created_at < STATE_POLICY['CO'][:start_date]
+    user.created_at < state_policies['CO'][:start_date]
```
artem-vavilov, line 57: Shouldn't we use `state_policy(user)` instead?

dashboard/lib/policies/child_account.rb
```diff
@@ -132,12 +120,29 @@ def self.personal_account?(user)
     end
   end
 
+  def self.state_policies
```
mgc1194, line 124: If this is not a private method, would we want to add unit tests?

daynew: good idea! https://github.com/code-dot-org/code-dot-org/pull/59131/commits/900608644fc424f3b1e1c683333b7b570cd29946

dashboard/lib/policies/child_account.rb
```diff
@@ -65,7 +53,7 @@ def self.compliant?(user)
   # state missing
   # We use Colorado as it is the only start date we have for now
   def self.user_predates_state_collection?(user)
-    user.created_at < STATE_POLICY['CO'][:start_date]
+    user.created_at < state_policy(user)[:start_date]
```
artem-vavilov: I believe we need to check if `state_policy(user)` is present

daynew: Actually we don't want to do this because the user might not have their state yet. We use this colorado date because it happens to also be the date we us-state collection. We should actually create a separate const....

artem-vavilov (approved):

PR #59140, opened 2024-06-07
Music: use callouts updates

breville:
This uses the callouts updates added in https://github.com/code-dot-org/code-dot-org/pull/59026, though instead of a yellow glow, it uses a brand-colour for the text itself, and updates the arrow to match.

https://github.com/code-dot-org/code-dot-org/assets/2205926/3ef3f3a2-9f31-487a-b802-989fe5800e29


sanchitmalhotra126 (approved):
Nice! I'd be curious to explore a bigger/more dramatic glow effect to make sure users see it, but defer to @markabarrett for design guidance

PR #59142, opened 2024-06-07
P20-987: Add ability to set DCDO via cookie for testing

artem-vavilov:
## Summary
Allows us to set DCDO via cookies for testing purposes in the `test` and `development` environments:
https://github.com/code-dot-org/code-dot-org/blob/79996e15f334d4ce49bda676ee8482a615696621/dashboard/test/ui/features/cookie_dcdo.feature#L7

## Links

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -112,6 +112,7 @@ group :development, :test do
   gem 'minitest-spec-context', '~> 0.0.3'
   gem 'minitest-stub-const', '~> 0.6'
   gem 'net-http-persistent'
+  gem 'request_store', '~> 1.6.0', require: false
   gem 'rinku'
   gem 'rspec', require: false
   gem 'selenium-webdriver', '~> 4.0'
--- a/Gemfile
+++ b/Gemfile
@@ -112,7 +112,6 @@ group :development, :test do
   gem 'minitest-spec-context', '~> 0.0.3'
   gem 'minitest-stub-const', '~> 0.6'
   gem 'net-http-persistent'
-  gem 'request_store', '~> 1.6.0', require: false
   gem 'rinku'
   gem 'rspec', require: false
   gem 'selenium-webdriver', '~> 4.0'
@@ -259,6 +258,7 @@ end
 # Reduce volume of production logs
 # Ref: https://github.com/roidrage/lograge/pull/252
 gem 'lograge', github: 'code-dot-org/lograge', ref: 'debug_exceptions'
+gem 'request_store', '~> 1.6.0', require: false
 
 # Enforce SSL
 gem 'rack-ssl-enforcer'
--- a/dashboard/test/ui/features/step_definitions/steps.rb
+++ b/dashboard/test/ui/features/step_definitions/steps.rb
@@ -1066,7 +1066,7 @@ def set_cookie_dcdo(key, value)
   end
 end
 
-Given(/^I set the DCDO flag "([^"]*)" to "(.*)"$/) do |key, json|
+Given(/^I set the DCDO key "([^"]*)" to "(.*)"$/) do |key, json|
   begin
     value = JSON.parse(json)
   rescue JSON::ParserError
--- a/dashboard/test/ui/features/step_definitions/steps.rb
+++ b/dashboard/test/ui/features/step_definitions/steps.rb
@@ -1058,12 +1058,7 @@ def set_cookie_dcdo(key, value)
 
   cookie_dcdo[key] = value
 
-  begin
-    set_cookie('DCDO', cookie_dcdo.to_json)
-  rescue Selenium::WebDriver::Error::InvalidCookieDomainError
-    navigate_to replace_hostname('http://studio.code.org')
-    retry
-  end
+  set_cookie('DCDO', cookie_dcdo.to_json)
 end
 
 Given(/^I set the DCDO key "([^"]*)" to "(.*)"$/) do |key, json|
--- a/dashboard/test/ui/features/step_definitions/steps.rb
+++ b/dashboard/test/ui/features/step_definitions/steps.rb
@@ -844,6 +844,22 @@ Then /^element "([^"]*)" is (not )?displayed$/ do |selector, negation|
   expect(element_displayed?(selector)).to eq(negation.nil?)
 end
 
+Then /^element "([^"]*)" attr "([^"]*)" includes (no )?"(.*)"$/ do |selector, attr, negation, attr_json|
+  actual_attr_value = @browser.find_element(:css, selector).attribute(attr)
+  expected_attr_value =
+    begin
+      JSON.parse(attr_json)
+    rescue JSON::ParserError
+      attr_json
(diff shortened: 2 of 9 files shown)


lib/dynamic_config/dcdo.rb
```diff
@@ -55,6 +55,8 @@ def frontend_config
       'student-email-post-enabled': DCDO.get('student-email-post-enabled', false),
       'progress-v2-metadata-enabled': DCDO.get('progress-v2-metadata-enabled', false),
       'show-updated-lms-content': DCDO.get('show-updated-lms-content', false),
+      # Used to test DCDO setting via cookie. See: 'dashboard/test/ui/features/cookie_dcdo.rb'
+      cookie_dcdo_test: DCDO.get('cookie_dcdo_test', nil),
```
wilkie, line 59: Do you even need to put this in here? I'm not seeing exactly now the UI test referenced there needs to put the value in the JS DCDO space. I admit about 50% of this mechanism is still magic to me, haha, so who knows.

artem-vavilov: It's just a straightforward method to verify that the functionality is working and that the DCDO returns the expected value.
https://github.com/code-dot-org/code-dot-org/blob/661fff3a9f6f822fecf7e1407941017da656a379/dashboard/app/views/layouts/application.html.haml#L26-L28

lib/cdo/rack/cookie_dcdo.rb
```diff
@@ -0,0 +1,32 @@
+require 'json'
+require 'request_store'
```
wilkie, line 2: So bizarre to see this gem for personal reasons!! I think this gem came out of the rstat.us days for me... Steve wrote this when being plagued by switching between puma and thin. I can't remember which of us had to diagnose a thread-local-storage bug and running into this problem. A whole dang lifetime ago!!!!! Wow!!!!!

If we are using `RequestStore`, though, we should put it in the `Gemfile` proper. In the `development`/`test` group with `require: false`. I think it is only coincidentally there at the moment.

lib/dynamic_config/dcdo.rb
wilkie: Oh! I see! That makes sense. 👍

wilkie (approved):
This looks very nice! Such quick work! I love some rack middleware especially when it opens the doors to better acceptance testing.

I would like to ensure the gem library you are using gets recorded in the Gemfile but other than that, this looks very nice. I'll pre-emptively approve so we can get it out the door! I'm excited to use it!

lib/cdo/rack/cookie_dcdo.rb
artem-vavilov: https://github.com/code-dot-org/code-dot-org/pull/59142/commits/1b1b95380d98307f8abd8e2c37deea7736d55a60

dashboard/test/ui/features/step_definitions/steps.rb
```diff
@@ -1032,6 +1048,34 @@ def set_cookie(key, value)
...
+
+  begin
+    set_cookie('DCDO', cookie_dcdo.to_json)
+  rescue Selenium::WebDriver::Error::InvalidCookieDomainError
+    navigate_to replace_hostname('http://studio.code.org')
+    retry
+  end
+end
+
+Given(/^I set the DCDO flag "([^"]*)" to "(.*)"$/) do |key, json|
```
daynew: nit-pick to match language in `dynamic_config_base.rb`
```suggestion
Given(/^I set the DCDO key "([^"]*)" to "(.*)"$/) do |key, json|
```

dashboard/test/ui/features/step_definitions/steps.rb
```diff
@@ -1032,6 +1048,34 @@ def set_cookie(key, value)
...
+      {}
+    end
+
+  cookie_dcdo[key] = value
+
+  begin
+    set_cookie('DCDO', cookie_dcdo.to_json)
+  rescue Selenium::WebDriver::Error::InvalidCookieDomainError
+    navigate_to replace_hostname('http://studio.code.org')
+    retry
```
daynew: Is there a risk of an infinite loop?

dashboard/test/ui/features/step_definitions/steps.rb
```diff
@@ -1032,6 +1048,34 @@ def set_cookie(key, value)
...
+    rescue Selenium::WebDriver::Error::NoSuchCookieError
+      {}
+    end
+
+  cookie_dcdo[key] = value
+
+  begin
+    set_cookie('DCDO', cookie_dcdo.to_json)
+  rescue Selenium::WebDriver::Error::InvalidCookieDomainError
+    navigate_to replace_hostname('http://studio.code.org')
```
daynew: What if the test is against pegasus? (code.org or hourofcode.com)? DCDO is also used by pegasus. Could you explain how this error occurs and how this navigate fixes the error?

daynew (approved):

dashboard/test/ui/features/step_definitions/steps.rb
artem-vavilov: The error occurs when we try to set a cookie for a domain while being on a different domain, which is not allowed. To avoid this, we should be within the expected domain scope `.code.org`:
https://github.com/code-dot-org/code-dot-org/blob/f8b90d55464aa91daa5dfaba200888d7ad634de3/dashboard/test/ui/features/step_definitions/steps.rb#L1043-L1046

dashboard/test/ui/features/step_definitions/steps.rb
artem-vavilov: If it didn't happen locally, then the chance of it occurring during CI testing is close to zero.

dashboard/test/ui/features/step_definitions/steps.rb
artem-vavilov: I believe Pegasus shares the same `.code.org` domain with Studio, and I wasn't aware of `hourofcode.com`. However, I can provide support for any domain if needed.

artem-vavilov: I believe Pegasus shares the same `.code.org` domain with Studio, and I wasn't aware of `hourofcode.com`.
However, after considering this, I realized it would be better to remove the rescue for `InvalidCookieDomainError`.

wilkie (approved):

Gemfile
```diff
@@ -258,6 +258,7 @@ end
 # Reduce volume of production logs
 # Ref: https://github.com/roidrage/lograge/pull/252
 gem 'lograge', github: 'code-dot-org/lograge', ref: 'debug_exceptions'
+gem 'request_store', '~> 1.6.0', require: false
```
artem-vavilov, line 261: @wilkie Since `lograge` (`debug_exceptions`) depends on `request_store`, we should allow `request_store` in the same envs.

PR #59145, opened 2024-06-07
Add experiment for allowing bigger playspace in Sprite Lab

bencodeorg:
Adds an experiment flag (`?enableExperiments=bigPlayspace`) that allows the playspace to be expanded up to a width of 75% of the viewport height (assumes that the limiting/smaller screen dimension in a visualization where height and width are the same is the height).

This is definitely still rough around the edges (eg, if you enable this experiment and view other lab types like App Lab, you get funky behavior), but the goal here is get something in front of curriculum and design folks to critique and see if it fits their needs.

https://github.com/code-dot-org/code-dot-org/assets/25372625/71e8c113-7041-4a9d-9ef9-06d34bff9033

## Links

- jira: https://codedotorg.atlassian.net/browse/LABS-858

## Testing story

I tested manually on a Sprite Lab allthethings level as well as two levels that were called out by curriculum as potentially being positively impacted by this change:
- /s/allthethings/lessons/36/levels/4
- /s/k5-ai-data-2024/lessons/1/levels/2
- /s/coursee-2023/lessons/5/levels/1

## Follow-up work

We'll need to do a lot more experimentation across device types, screen sizes, labs, etc. before we would ship a change this this without an experiment flag. I'm also not super happy with the scaling/resolution of the playspace at larger widths, so that's something else we'd want to look at. We'd maybe want a better experience for what happens to the workspace/instructions at these larger sizes (eg, full collapse button?) as well. This is probably not a comprehensive list of things we'd want to follow up on before making this change broadly :).

### DIFF
--- a/apps/src/p5lab/P5LabView.jsx
+++ b/apps/src/p5lab/P5LabView.jsx
@@ -4,8 +4,18 @@ import classNames from 'classnames';
 import {connect} from 'react-redux';
 import PropTypes from 'prop-types';
 import React from 'react';
-import AnimationTab from './AnimationTab/AnimationTab';
+
+import experiments from '@cdo/apps/util/experiments';
 import StudioAppWrapper from '@cdo/apps/templates/StudioAppWrapper';
+import InstructionsWithWorkspace from '@cdo/apps/templates/instructions/InstructionsWithWorkspace';
+import {isResponsiveFromState} from '@cdo/apps/templates/ProtectedVisualizationDiv';
+import CodeWorkspace from '@cdo/apps/templates/CodeWorkspace';
+import {getManifest} from '@cdo/apps/assetManagement/animationLibraryApi';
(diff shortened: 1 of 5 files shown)


apps/src/p5lab/P5LabView.jsx
```diff
@@ -4,8 +4,18 @@ import classNames from 'classnames';
 import {connect} from 'react-redux';
 import PropTypes from 'prop-types';
 import React from 'react';
-import AnimationTab from './AnimationTab/AnimationTab';
+
+import experiments from '@cdo/apps/util/experiments';
```
bencodeorg, line 8: This is all just moving around imports to match our current preferred import order/categories, with the exception of adding the `experiments` import

sanchitmalhotra126 (approved):
LGTM!

PR #59147, opened 2024-06-07
Add translation support to Codebridge

molly-moen:
Add support for translation in codebridge. This involves making a codebridge locale and importing it in the lab2 haml entrypoint. I translated the File Browser as a proof of concept, other translations are left as a follow-up. I also translated the workspace header, which uses a common translation string rather than a codebridge-specific one.

## Links
- jira ticket: [CT-628](https://codedotorg.atlassian.net/browse/CT-628)

## Testing story
Tested locally that the new strings work.

## Follow-up work
From now on we should use translatable strings when adding text to Codebridge/Python Lab. I also filed [CT-631](https://codedotorg.atlassian.net/browse/CT-631) to translate all the existing hard-coded strings.

## PR Checklist:

- [ ] Code is well-commented


thomasoniii (approved):
just confirming - a lot of these files have extra spaces added `{' '}`, and we do actually want them and it's not a linter artifact, right?

otherwise, lgtm 🚀

molly-moen:
> just confirming - a lot of these files have extra spaces added `{' '}`, and we do actually want them and it's not a linter artifact, right?
> 
> otherwise, lgtm 🚀

@thomasoniii yep they are intentional! The linter was adding a new line so we needed an explicit `' '` to have a space between the icon and the label.

PR #59148, opened 2024-06-07
Lab2: tidy up types

sanchitmalhotra126:
Quick cleanup of Lab2 types. Changes include:
- Grouping types in lab2/types.ts by category
- Removing redundant/unused types
- Removing deprecated fields from level data, like text, validations, and panels (these should be coming from the level properties directly).

Of these, only the third makes changes to existing behavior; now instead of falling back to level data validations, text, or panels, we will just ignore it. This should **only** affect levels that were created **before** we added custom GUI editors for these fields in the level edit page (i.e. validations and panels) and **have not** been updated since. I manually tested our production scripts that have these fields to make sure nothing was broken (see Testing Story below).

I was originally thinking of splitting the types.ts file up according to the various sections but held off on this for now to avoid having to change a bunch of imports.

## Testing story

Locally verified that music-intro-2024, music-tutorial-2024, and gen-ai-customizing-pilot-v1 were unchanged. I also checked music-accessibility, which does have one level which will need to be updated. I will manually update that level in levelbuilder before merging this (just need to open re-save the level without making changes).

## Follow-up work

- Update https://levelbuilder-studio.code.org/s/music-accessibility/lessons/1/levels/7 on levelbuilder before merging.

### DIFF
--- a/apps/src/lab2/constants.ts
+++ b/apps/src/lab2/constants.ts
@@ -27,3 +27,6 @@ export enum PERMISSIONS {
 export const START_SOURCES = 'start_sources';
 
 export const LABS_USING_NEW_SHARE_DIALOG = ['music', 'pythonlab'];
+
+// Text-based labs that are currently supported by lab2.
+export const TEXT_BASED_LABS: AppName[] = ['aichat', 'pythonlab', 'weblab2'];
--- a/apps/src/lab2/progress/ProgressContainer.tsx
+++ b/apps/src/lab2/progress/ProgressContainer.tsx
@@ -8,7 +8,6 @@ import {useAppDispatch, useAppSelector} from '@cdo/apps/util/reduxHooks';
 import React, {useCallback, useEffect, useRef} from 'react';
 import {useSelector} from 'react-redux';
 import {setValidationState} from '../lab2Redux';
-import {ProjectLevelData} from '../types';
 
 interface ProgressContainerProps {
   children: React.ReactNode;
@@ -43,21 +42,13 @@ const ProgressContainer: React.FunctionComponent<ProgressContainerProps> = ({
     new ProgressManager(onProgressChange)
   );
 
-  const levelDataValidations = useAppSelector(
-    state =>
-      (state.lab.levelProperties?.levelData as ProjectLevelData | undefined)
-        ?.validations
-  );
   const levelValidations = useAppSelector(
     state => state.lab.levelProperties?.validations
   );
 
   useEffect(() => {
-    // Use validations on level properties if present, otherwise fallback to validations in level data.
-    progressManager.current.onLevelChange(
-      levelValidations || levelDataValidations
-    );
-  }, [levelValidations, levelDataValidations]);
+    progressManager.current.onLevelChange(levelValidations);
+  }, [levelValidations]);
 
   return (
     <ProgressManagerContext.Provider value={progressManager.current}>
--- a/apps/src/lab2/types.ts
+++ b/apps/src/lab2/types.ts
@@ -143,8 +143,6 @@ export interface LevelProperties {
   toolboxBlocks?: string;
   source?: MultiFileSource;
   sharedBlocks?: BlockDefinition[];
-  // We are moving level validations out of level data and into level properties.
-  // Temporarily keeping them in both places to avoid breaking existing code.
   validations?: Validation[];
   // An optional URL that allows the user to skip the progression.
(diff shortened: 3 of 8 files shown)


apps/src/lab2/constants.ts
```diff
@@ -27,3 +27,6 @@ export enum PERMISSIONS {
 export const START_SOURCES = 'start_sources';
 
 export const LABS_USING_NEW_SHARE_DIALOG = ['music', 'pythonlab'];
+
+// Text-based labs that are currently supported by lab2.
```
breville, line 31: Does "text-based" here mean the student source is created in a text editor?

breville (approved):
Nice!

molly-moen (approved):

apps/src/lab2/types.ts
```diff
@@ -38,7 +49,7 @@ export interface ProjectSources {
 }
 
 // We will eventually make this a union type to include other source types.
-export type Source = BlocklySource;
+export type Source = BlocklySource | MultiFileSource;
```
fisher-alice, line 52: Do you mind explaining the difference between `Source` type here which includes `MultiFileSource` and is included as a property in `ProjectLevelData` and `source` in `LevelProperties` which has type `MultiFileSource`?
I'm wrapping my head around sources within these types as well as the `source` in `ProjectSources` interface included in `ProjectAndSources`.

apps/src/lab2/progress/ProgressContainer.tsx
```diff
@@ -8,7 +8,6 @@ import {useAppDispatch, useAppSelector} from '@cdo/apps/util/reduxHooks';
 import React, {useCallback, useEffect, useRef} from 'react';
 import {useSelector} from 'react-redux';
 import {setValidationState} from '../lab2Redux';
-import {ProjectLevelData} from '../types';
```
fisher-alice, line 11: Just confirming that this cleanup is finishing up the absorbing of fields in `level_data` back to `level_properties` as part of [this convo](https://codedotorg.slack.com/archives/C044AGTKW3D/p1690844666853189?thread_ts=1690842845.194209&cid=C044AGTKW3D).

apps/src/lab2/constants.ts
sanchitmalhotra126: Kind of, yes - this refers to any lab where student source is stored as text. Sometimes this is in a text-based IDE (python, weblab) but not always, for example in AI Chat.

I believe this is primarily used to determine what string to display in the start over message (referring to "remove all blocks" vs "remove all code")

apps/src/lab2/types.ts
sanchitmalhotra126: Yeah this is a bit confusing and worth cleaning up later :)
- `ProjectLevelData -> startSources` are start sources for labs that use the level data field. Currently this is only Music Lab. I added the union type to keep things consistent but in practice, this is always going to be **un-stringified** Blockly JSON since it's only in Music Lab.
- `LevelProperties -> source` I believe is start sources for CodeBridge projects specifically, which is why it's currently only a `MultiFileSource`.
- `ProjectAndSources -> sources` refers to the **student's** initial sources loaded for the lab, if present. These can be undefined if the student has not started the project yet. These sources are currently saved as either **stringified** blockly JSON, or **un-stringified** multi-file sources. The blockly sources are currently saved as strings just because that's how it was initially set up (and is consistent with legacy labs) but doesn't necessarily need to be so.

There are a couple cleanups I think we can do here (might even have JIRAs for these already):
- Unified location for all labs start sources (consolidate `LevelProperties -> source` and `ProjectLevelData -> startSources`)
- Store Blockly JSON as un-stringified JSON rather than a string. Makes the format consistent across start sources and user sources and prevents unnecessary stringifying/parsing.

apps/src/lab2/progress/ProgressContainer.tsx
sanchitmalhotra126: Yep! We had already started using the level_properties fields for localization purposes, so this only affects one level in the music-accessibility script as far as I can tell (which I will fix before merging).

apps/src/lab2/types.ts
fisher-alice: This is super helpful. And both your cleanup items SGTM, especially the first bullet. Unsure how involved the process of converting stored legacy Blockly from stringified to un-stringified JSON.
Thanks so much!

PR #59150, opened 2024-06-08
Add LTI account linking metrics

carl-codeorg:
Adds some new metrics for LTI account linking:

- `lti_account_linking_page_visit` - when the user is sent to the new account landing/disambiguation page
- `lti_existing_account_click` - when the user clicks the button to link an existing account instead of creating a new account
- `lti_new_account_click` - when the user clicks the button to create a fresh account instead of linking
- `lti_account_linked` - after a successful account linking operation

## Links

[JIRA](https://codedotorg.atlassian.net/browse/P20-921)

## Testing story

Tested locally and observed statsig output to stdout and to the browser console.

## Follow-up work

Two metrics from the ticket still need to be implemented:

- Clicks for the individual oauth linking buttons. Currently these are wired up the same as the normal oauth login buttons, so they'll publish the usual oauth events. Will have to circle back with Ali on whether or not we need to do special instrumentation of the buttons.
- Successfully creating an LTI account. Not sure how this differs from normal account creation metrics and doesn't seem directly related to account linking.

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/dashboard/app/controllers/lti_v1_controller.rb
+++ b/dashboard/app/controllers/lti_v1_controller.rb
@@ -199,6 +199,15 @@ class LtiV1Controller < ApplicationController
         PartialRegistration.persist_attributes(session, user)
         session[:user_return_to] = destination_url
         if DCDO.get('lti_account_linking_enabled', false)
+          metadata = {
+            'user_type' => user.user_type,
+            'lms_name' => integration[:platform_name],
+          }
+          Metrics::Events.log_event(
+            user: user,
+            event_name: 'lti_account_linking_page_visit',
+            metadata: metadata,
+          )
           render 'lti/v1/account_linking/landing', locals: {lti_provider: integration[:platform_name], email: email_address} and return
         end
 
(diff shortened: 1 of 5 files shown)


dashboard/app/controllers/lti_v1_controller.rb
```diff
@@ -199,6 +199,15 @@ def authenticate
         PartialRegistration.persist_attributes(session, user)
         session[:user_return_to] = destination_url
         if DCDO.get('lti_account_linking_enabled', false)
+          metadata = {
+            'user_type' => user.user_type,
```
nicklathe: `user_type` is actually automatically added to the custom_ids object https://github.com/code-dot-org/code-dot-org/blob/staging/dashboard/lib/metrics/events.rb#L61.

This functionality was added relatively recently, I still need to go remove all the places where I added it to the metadata.

dashboard/lib/services/lti/account_linker.rb
```diff
@@ -15,6 +17,19 @@ def call
           user.lti_roster_sync_enabled = true if user.teacher?
           PartialRegistration.delete(session)
         end
+
+        lti_integration = user.lti_user_identities.first.lti_integration
+        metadata = {
+          'user_type' => user.user_type,
```
nicklathe: Same comment as above 👍

nicklathe (approved):

stephenliang (approved):

PR #59169, opened 2024-06-10
Fix /catalog for when device compatibility is (None)

TurnerRiley:
Last week, [Dan found a bug](https://codedotorg.slack.com/archives/C04540KNGDQ/p1716219131534839) where if a new student course offering is created without setting the device compatibilities (at least , when its associated Curriculum Catalog expanded card is opened it will break the page.

This PR fixes this both on the side of the unit + course editors and on the Curriculum Catalog. In the unit + course editors now, a unit/course that will go on the catalog page (assignable and for students) can't be published (set to "preview" or "stable") without setting all device compatibilities. On the Curriculum Catalog, expanded card device compatibilities are only listed if a valid answer is provided to avoid the page breaking.

#### Unit editor
Does not allow editors to publish (i.e. put into "preview" or "stable" state) if at least one of the device compatibility dropdowns is still set to "(None)":

https://github.com/code-dot-org/code-dot-org/assets/56283563/686dcc5f-6ec8-4823-b58e-1ba4b16835a9

Allows editors to publish once all device compatibility dropdowns are set:

https://github.com/code-dot-org/code-dot-org/assets/56283563/14f3b8e6-ba44-48ce-8399-be8e0b8e3361

#### Course editor has the same behavior

https://github.com/code-dot-org/code-dot-org/assets/56283563/048d2a06-11c0-4a7c-b530-94f9cbeabf7b

#### Regular view with all device compatibilities set remains unchanged
For a unit course offering:
regular is unchanged

For a unit group course offering:
looks normal

#### If one device compatibility is not set, will not show that device
For a unit course offering:
one dc is still None

For a unit group course offering:
Doesnt show Null device for edited course

#### If no device compatibilities are set, will not show any of them
For a unit course offering:
null dc

For a unit group course offering:
csd no devices

## Links
Jira ticket: [here](https://codedotorg.atlassian.net/jira/software/c/projects/ACQ/boards/64?assignee=60d62161f65054006980bd71&selectedIssue=ACQ-1907)
Slack discussion where Dan found the bug: [here](https://codedotorg.slack.com/archives/C04540KNGDQ/p1716219131534839)

## Testing story
Local testing and updating unit tests.

### DIFF
--- a/apps/src/templates/curriculumCatalog/ExpandedCurriculumCatalogCard.jsx
+++ b/apps/src/templates/curriculumCatalog/ExpandedCurriculumCatalogCard.jsx
@@ -62,7 +62,7 @@ const ExpandedCurriculumCatalogCard = ({
     },
   };
 
-  const devices = JSON.parse(deviceCompatibility);
+  const devices = deviceCompatibility ? JSON.parse(deviceCompatibility) : {};
 
   const resoucesOrder = [
     'Lesson Plan',
@@ -233,22 +233,29 @@ const ExpandedCurriculumCatalogCard = ({
               </div>
               <hr className={style.horizontalDivider} />
               <div className={style.compatibilityContainer}>
-                {Object.keys(devices).map(device => (
-                  <div key={device} className={style.iconWithDescription}>
-                    <FontAwesome
-                      icon={iconData[devices[device]].icon}
-                      className={`fa-solid ${iconData[devices[device]].color}`}
-                    />
-                    <BodyTwoText>
-                      {device !== 'no_device'
-                        ? translatedCourseOfferingDeviceTypes[device]
-                            .charAt(0)
-                            .toUpperCase() +
-                          translatedCourseOfferingDeviceTypes[device].slice(1)
-                        : i18n.offline()}
-                    </BodyTwoText>
-                  </div>
-                ))}
+                {Object.keys(devices).map(
+                  device =>
+                    devices[device] !== '' && (
+                      <div key={device} className={style.iconWithDescription}>
+                        <FontAwesome
+                          icon={iconData[devices[device]].icon}
+                          className={`fa-solid ${
+                            iconData[devices[device]].color
+                          }`}
+                        />
+                        <BodyTwoText>
+                          {device !== 'no_device'
+                            ? translatedCourseOfferingDeviceTypes[device]
+                                .charAt(0)
+                                .toUpperCase() +
+                              translatedCourseOfferingDeviceTypes[device].slice(
+                                1
+                              )
+                            : i18n.offline()}
+                        </BodyTwoText>
+                      </div>
+                    )
+                )}
               </div>
               <hr className={style.horizontalDivider} />
               <div className={style.buttonsContainer}>
--- a/dashboard/app/models/course_offering.rb
+++ b/dashboard/app/models/course_offering.rb
@@ -240,19 +240,14 @@ class CourseOffering < ApplicationRecord
     false
   end
 
-  # Return if the course offering's device compatibilities JSON string either:
-  # - Is null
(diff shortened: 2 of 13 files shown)


apps/src/templates/curriculumCatalog/ExpandedCurriculumCatalogCard.jsx
```diff
@@ -233,22 +233,29 @@ const ExpandedCurriculumCatalogCard = ({
...
-                            .charAt(0)
-                            .toUpperCase() +
-                          translatedCourseOfferingDeviceTypes[device].slice(1)
-                        : i18n.offline()}
-                    </BodyTwoText>
-                  </div>
-                ))}
+                {Object.keys(devices).map(
+                  device =>
+                    devices[device] !== '' && (
```
TurnerRiley, line 238: The only code change for this block is adding `devices[device] !== '' &&` but because of the lint styling indents it looks like a much larger change

dashboard/app/models/unit.rb
```diff
@@ -1614,10 +1625,13 @@ def summarize_for_rollup(user = nil)
...
 
   def summarize_for_unit_edit
+    co = course_version&.course_offering
     include_lessons = false
     summary = summarize(include_lessons)
     summary[:lesson_groups] = lesson_groups.map(&:summarize_for_unit_edit)
-    summary[:courseOfferingEditPath] = edit_course_offering_path(course_version.course_offering.key) if course_version
+    summary[:courseOfferingEditPath] = edit_course_offering_path(co&.key) if course_version
+    # Course offering device compatibilites are required to publish student courses that are assignable.
+    summary[:missingRequiredDeviceCompatibilities] = co&.assignable? && co&.get_participant_audience == 'student' && missing_device_compatibility?
```
bethanyaconnor: why do we need to check `assignable?` here?

bethanyaconnor (commented):
This looks great for units! I think you also need to update the course edit page and the corresponding component `CourseEditor`. We probably want similar protections for curriculum whose root is a UnitGroup.

TurnerRiley:
> This looks great for units! I think you also need to update the course edit page and the corresponding component `CourseEditor`. We probably want similar protections for curriculum whose root is a UnitGroup.

Oh good call, thank you!

dashboard/app/models/unit.rb
TurnerRiley: This is to match which courses are shown on the Curriculum Catalog page (minus the condition of being published). The ones shown on the catalog page are student assignable published courses.

dashboard/app/models/course_offering.rb
```diff
@@ -240,6 +240,21 @@ def can_be_assigned?(user)
...
+  def missing_device_compatibility?
+    return true unless device_compatibility
+
+    device_compatibility_values = JSON.parse(device_compatibility).values
+    device_compatibility_values.any?(&:blank?)
+  end
+
+  def missing_required_device_compatibility?
+    # Course offering device compatibilites are required to publish student courses that are assignable.
+    assignable? && get_participant_audience == 'student' && missing_device_compatibility?
```
bethanyaconnor: I wonder if we shouldn't even check `assignable`? Sometimes, the course offering is made assignable after the course/unit is published and I don't think we have protections there.

bethanyaconnor (approved):
This looks great! Had one question -- curious for your thoughts.

dashboard/app/models/course_offering.rb
TurnerRiley: Sounds good! On it!

TurnerRiley: I'm also simplifying to one function since I don't think it needs to be two separate ones

PR #59170, opened 2024-06-10
define parameters_get block and generator

mikeharv:
Core Blockly uses the same `variables_get` block for global variables as well as parameters. However, our levels expect a similar `parameters_get` block, which is found in the modal function editor.  These blocks generate the same code but have a different appearance. Compare the `direction` parameter with the `left` variable:
image

Due to concerns about stretching an already thin palette of accessible block colors, I'm hesitant to define a new block style for this block. Instead, for now, I've opted to re-use our "sprite_blocks" style, because a) sprites and parameters are so far never both exposed within the same lab, and b) the color seems similar enough. 

image

Note that this branch just defines the block appearance and code generation. The work to actually add them to the mini-toolbox (as shown above) is in another branch: https://github.com/code-dot-org/code-dot-org/pull/59180 

## Links

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/blockly/addons/cdoVariables.ts
+++ b/apps/src/blockly/addons/cdoVariables.ts
@@ -6,6 +6,8 @@ import {BlocklyWrapperType} from '../types';
 export default function initializeVariables(
   blocklyWrapper: BlocklyWrapperType
 ) {
+  blocklyWrapper.JavaScript.forBlock.parameters_get =
+    blocklyWrapper.JavaScript.forBlock.variables_get;
   blocklyWrapper.Variables.DEFAULT_CATEGORY = 'Default';
 
   // TODO: Removing support for sprite variables as a separate variable type is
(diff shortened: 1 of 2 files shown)


apps/src/blockly/addons/cdoVariables.ts
```diff
@@ -6,6 +6,8 @@ import {BlocklyWrapperType} from '../types';
 export default function initializeVariables(
   blocklyWrapper: BlocklyWrapperType
 ) {
+  blocklyWrapper.JavaScript.forBlock.parameters_get =
```
molly-moen, line 9: what does this line do?

mikeharv: Tells Blockly to use the same generator function that it uses for standard variable getters.

molly-moen: Mind adding a comment explaining that? I could see it being confusing that we are doing things for parameters inside the variables function.

mikeharv: Yup, can do

molly-moen (approved):

PR #59174, opened 2024-06-10
P20-987: Add ability to set DCDO via cookie for testing

artem-vavilov:
## Summary

Allows us to set DCDO via cookies for testing purposes in the `test` and `development` environments:
https://github.com/code-dot-org/code-dot-org/blob/4117b99459d18cab0bb28dcffe844eecf0b10f6b/dashboard/test/ui/features/support/dashboard_helpers.rb#L15-L26

## Links
- Related PR: https://github.com/code-dot-org/code-dot-org/pull/59142
- Revert PR: https://github.com/code-dot-org/code-dot-org/pull/59157

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -258,6 +258,7 @@ end
 # Reduce volume of production logs
 # Ref: https://github.com/roidrage/lograge/pull/252
 gem 'lograge', github: 'code-dot-org/lograge', ref: 'debug_exceptions'
+gem 'request_store', '~> 1.6.0', require: false
 
 # Enforce SSL
 gem 'rack-ssl-enforcer'
--- a/dashboard/test/ui/features/cookie_dcdo.feature
+++ b/dashboard/test/ui/features/cookie_dcdo.feature
@@ -1,0 +1,19 @@
+Feature: Cookie DCDO
+  Scenario: Cookie DCDO flag is set
+    Given I am on "http://studio.code.org/users/sign_in"
+    Then element "script[data-dcdo]" attr "data-dcdo" includes ""cookie_dcdo_test":null"
+
+    # Tests assigning of a cookie DCDO flag
+    When I set the DCDO key "cookie_dcdo_test" to "Cookie DCDO assigning works"
+    And I reload the page
+    Then element "script[data-dcdo]" attr "data-dcdo" includes ""cookie_dcdo_test":"Cookie DCDO assigning works""
+
+    # Tests re-assigning of a cookie DCDO flag
+    When I set the DCDO key "cookie_dcdo_test" to "{"Cookie DCDO re-assigning":"works"}"
+    And I reload the page
+    Then element "script[data-dcdo]" attr "data-dcdo" includes ""Cookie DCDO re-assigning":"works""
+
+    # Tests cleaning of a cookie DCDO flag
+    When I delete the cookie named "DCDO"
+    And I reload the page
+    Then element "script[data-dcdo]" attr "data-dcdo" includes ""cookie_dcdo_test":null"
--- a/dashboard/test/ui/features/cookie_dcdo.feature
+++ b/dashboard/test/ui/features/cookie_dcdo.feature
@@ -4,14 +4,14 @@ Feature: Cookie DCDO
     Then element "script[data-dcdo]" attr "data-dcdo" includes ""cookie_dcdo_test":null"
 
     # Tests assigning of a cookie DCDO flag
-    When I set the DCDO key "cookie_dcdo_test" to "Cookie DCDO assigning works"
+    When I set the DCDO key "cookie_dcdo_test" to "assigning_works"
     And I reload the page
-    Then element "script[data-dcdo]" attr "data-dcdo" includes ""cookie_dcdo_test":"Cookie DCDO assigning works""
+    Then element "script[data-dcdo]" attr "data-dcdo" includes ""cookie_dcdo_test":"assigning_works""
 
     # Tests re-assigning of a cookie DCDO flag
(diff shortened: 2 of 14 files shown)


Gemfile
```diff
@@ -258,6 +258,7 @@ end
 # Reduce volume of production logs
 # Ref: https://github.com/roidrage/lograge/pull/252
 gem 'lograge', github: 'code-dot-org/lograge', ref: 'debug_exceptions'
+gem 'request_store', '~> 1.6.0', require: false
```
artem-vavilov, line 261: We need to allow `request_store` in any group where `lograge` (`debug_exceptions`) is enabled, since `lograge` depends on it.

dashboard/test/ui/features/support/dashboard_helpers.rb
```diff
@@ -9,6 +9,21 @@ def require_rails_env
     puts "Requiring rails env took #{finish - start} seconds"
     @rails_loaded = true
   end
+
+  # Sets DCDO for a test scenario.
+  # @see Rack::CookieDCDO
+  def set_dcdo(key, value)
```
Hamms, line 15: nit; can this be given a more explicit name? I'd like it to be made clear to someone reading the code that invokes this method that it is specifically setting a cookie which will override the actual value set in DCDO just for testing purposes, rather than actually setting a value on the DCDO store itself.

dashboard/test/ui/features/cookie_dcdo.feature
```diff
@@ -0,0 +1,19 @@
+Feature: Cookie DCDO
+  Scenario: Cookie DCDO flag is set
+    Given I am on "http://studio.code.org/users/sign_in"
+    Then element "script[data-dcdo]" attr "data-dcdo" includes ""cookie_dcdo_test":null"
+
+    # Tests assigning of a cookie DCDO flag
+    When I set the DCDO key "cookie_dcdo_test" to "assigning_works"
```
Hamms, line 7: I'd like this step to be more descriptive of what it's actually doing. What about something like,

```suggestion
    When I use a cookie to mock the DCDO key "cookie_dcdo_test" as "assigning_works"
```

Or even just,

```suggestion
    When I mock the DCDO key "cookie_dcdo_test" as "assigning_works"
```

lib/dynamic_config/dcdo.rb
```diff
@@ -55,6 +55,8 @@ def frontend_config
       'student-email-post-enabled': DCDO.get('student-email-post-enabled', false),
       'progress-v2-metadata-enabled': DCDO.get('progress-v2-metadata-enabled', false),
       'show-updated-lms-content': DCDO.get('show-updated-lms-content', false),
+      # Used to test DCDO setting via cookie. See: 'dashboard/test/ui/features/cookie_dcdo.rb'
+      cookie_dcdo_test: DCDO.get('cookie_dcdo_test', nil),
```
Hamms, line 59: Why is this necessary to include in the set of values we forward to frontend code?

lib/cdo/rack/cookie_dcdo.rb
```diff
@@ -0,0 +1,32 @@
...
+      RequestStore.store[:DCDO] = JSON.parse(Rack::Request.new(env).cookies['DCDO'] || '{}')
+
+      unless DCDO.instance_variable_get(:@_redefined)
+        # Redefines `DCDO#get` to return the cookie DCDO value if it exists.
+        DCDO.define_singleton_method(:get) do |key, *args|
+          RequestStore.store[:DCDO]&.key?(key) ? RequestStore.store[:DCDO][key] : super(key, *args)
+        end
+
+        DCDO.instance_variable_set(:@_redefined, true)
+      end
```
Hamms: A Rack middleware definitely doesn't feel like the best place to implement this modification to DCDO functionality. Any reason not to just put this logic in the class itself, gated by an `if CDO.cookie_dcdo` check?

Hamms (commented):
Can you clarify what purpose `RequestStore` is serving in this implementation? It looks to me like we're storing the value we want in the cookie, then moving it from the cookie to the RequestStore, then reading it out of the RequestStore. Why aren't we reading the value from the cookie directly? What am I missing?

lib/dynamic_config/dcdo.rb
artem-vavilov: It's just the easiest way to identify that the DCDO cookie stubbing is working,
but we can add a separate tag with data attr to test this.

Hamms: Ahh, of course. Yeah, if possible I'd love to do this in a way that keeps test-specific functionality isolated from things we do for the sake of the user experience in production. Even if that means creating a template we can render just for the sake of the test, like we do for https://github.com/code-dot-org/code-dot-org/blob/staging/pegasus/sites.v3/code.org/public/test_dcdo.haml

Hamms (approved):
Nits and concerns aside, this LGTM to me to merge in the short term so we can do some manual debugging of it on test.

I do want to see some changes here eventually, but those can be done in a follow-up PR

lib/cdo/rack/cookie_dcdo.rb
artem-vavilov: I usually prefer to avoid metaprogramming, but in this case, I have three reasons for doing so:
1. Since this is intended for testing purposes only, I didn't want to modify the original method, as that would burden it with redundant checks in production and other environments.
2. It's easier to understand the business logic when everything is kept together in one place.
3. This dynamic method redefinition will ensure that we override `DCDO#get` regardless of its origin, even if it isn't inherited from `DynamicConfigBase` where the original `#get` method is defined.

PR #59178, opened 2024-06-11
include hidden blocks in block counts for feedback

mikeharv:
This updates how we count get or count all student blocks in our feedback utilities so that blocks on the hidden workspace are counted.  We surface this count when we report the number of lines a student has written and when monitoring their lines of code compared with the "ideal block count":
image
image

The first changed function also ensures that checks for recommend and required blocks will now also include hidden blocks. I'm not aware of any levels where we set recommended/required blocks that a student would be expected to add with the modal function editor, but there's also no reason we shouldn't be able to.

## Links

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/feedback.js
+++ b/apps/src/feedback.js
@@ -1525,10 +1525,6 @@ FeedbackUtils.prototype.getCountableBlocks_ = function () {
       : []),
   ];
   const blocks = allBlocks.filter(FeedbackUtils.blockShouldBeCounted_);
-  console.log(
-    blocks.length,
-    blocks.map(b => b.type)
-  );
   return blocks;
 };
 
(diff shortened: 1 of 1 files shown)


apps/src/feedback.js
```diff
@@ -1466,9 +1466,14 @@ FeedbackUtils.prototype.hasAllBlocks_ = function (blocks) {
  * @return {Array<Object>} The blocks.
  */
 FeedbackUtils.prototype.getUserBlocks_ = function () {
-  var allBlocks = Blockly.mainBlockSpace.getAllUsedBlocks();
-  var blocks = allBlocks.filter(function (block) {
-    var blockValid = !block.disabled && block.type !== 'when_run';
+  const allBlocks = [
```
molly-moen, line 1469: nit: since we reuse this logic should we make a helper function that gets all blocks for the workspaces?

molly-moen (approved):

apps/src/feedback.js
mikeharv: Will do as a follow-up!

PR #59180, opened 2024-06-11
add parameters to mini-toolbox

mikeharv:
Requires:
- https://github.com/code-dot-org/code-dot-org/pull/59170

This adds parameter blocks to a mini-toolbox for procedure definition blocks, where allowed:
image

Levelbuilders can toggle a setting that enables/disables this feature.

This field is available for all Blockly labs: https://github.com/code-dot-org/code-dot-org/blob/d2f76326f4b659d188d548735f59061f63148b7b/dashboard/app/views/levels/editors/fields/_blockly.html.haml#L28

The value is added to the Blockly inject options here: https://github.com/code-dot-org/code-dot-org/blob/d2f76326f4b659d188d548735f59061f63148b7b/apps/src/StudioApp.js#L2882

However, `"disable_param_editing": "true"` doesn't appear to be set on any published Dance/Sprite Lab levels, so this shouldn't immediately affect live courses.

**Technical details**
* We will store `Blockly.enableParamEditing` similar to `Blockly.useModalFunctionEditor`. (Disabled by default, ie. unless the level explicitly doesn't disable them.)
* If we are not disabling parameter editing, we will add the `parameters_get` block type to the list of mini-toolbox blocks.
* When creating the flyout items for the toolbox, a special case will be added for parameter blocks, so that we creating one block instance for defined parameter:

image

Note that this only makes it so that blocks for existing parameters are available. It doesn't make it possible for users to add new parameters to a function. I plan to do this by adding a "New parameter..." button to the mini-toolbox flyout.
There is additional required follow-up to customize the dropdown options for the parameter blocks.

## Links

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/blockly/constants.ts
+++ b/apps/src/blockly/constants.ts
@@ -116,12 +116,13 @@ export function stripUserCreated(xmlString: string) {
 
 export enum BLOCK_TYPES {
   behaviorDefinition = 'behavior_definition',
-  danceWhenSetup = 'Dancelab_whenSetup',
-  procedureDefinition = 'procedures_defnoreturn',
-  whenRun = 'when_run',
   behaviorGet = 'gamelab_behavior_get',
-  spriteParameterGet = 'sprite_parameter_get',
+  danceWhenSetup = 'Dancelab_whenSetup',
+  parametersGet = 'parameters_get',
+  procedureDefinition = 'procedures_defnoreturn',
   procedureCall = 'procedures_callnoreturn',
+  spriteParameterGet = 'sprite_parameter_get',
+  whenRun = 'when_run',
   variableGet = 'variables_get',
   variableSet = 'variables_set',
 }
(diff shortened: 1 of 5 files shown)


apps/src/blockly/constants.ts
```diff
@@ -116,12 +116,13 @@ export function stripUserCreated(xmlString: string) {
...
   behaviorDefinition = 'behavior_definition',
+  behaviorGet = 'gamelab_behavior_get',
   danceWhenSetup = 'Dancelab_whenSetup',
+  parametersGet = 'parameters_get',
   procedureDefinition = 'procedures_defnoreturn',
-  whenRun = 'when_run',
-  behaviorGet = 'gamelab_behavior_get',
-  spriteParameterGet = 'sprite_parameter_get',
   procedureCall = 'procedures_callnoreturn',
+  spriteParameterGet = 'sprite_parameter_get',
```
mikeharv, line 124: In addition to adding this new type, I also alphabetized the full list.

apps/src/blockly/customBlocks/googleBlockly/proceduresBlocks.ts
```diff
@@ -143,13 +143,18 @@ GoogleBlockly.Extensions.register(
 GoogleBlockly.Extensions.register(
   'procedure_def_mini_toolbox',
   function (this: ProcedureBlock) {
-    // TODO: Add comment block here after https://codedotorg.atlassian.net/browse/CT-121
```
mikeharv, line 146: We ended up adding a description field to the block rather than adding a comment block to the minitoolbox.

molly-moen (approved):

PR #59192, opened 2024-06-12
P20-964: Add new CAP compliance state `grace_period` and remove `request_sent`

artem-vavilov:
## Links
- JIRA task: [P20-964](https://codedotorg.atlassian.net/browse/P20-964)


daynew (approved):
The implementation and tests changes make sense. I will assume all the transition logic will be handled in https://github.com/code-dot-org/code-dot-org/pull/59241

PR #55648, opened 2024-01-09
Music: intro script

breville:
This adds a new intro script for **Music Lab**, at https://studio.code.org/s/music-intro-2024.  

It will be the first public use of **Lab2**.

It's not final, and more specific validations will be added, but the general structure is in place.  

The levels are:

```
musiclab_intro_video_start
musiclab_intro_change_sound
musiclab_intro_play_sound
musiclab_intro_play_together
musiclab_intro_repeat
musiclab_intro_functions
musiclab_intro_trigger
musiclab_intro_video_end
musiclab_intro_freeplay
```

See https://github.com/code-dot-org/code-dot-org/pull/55755 for an enhancement to `musiclab_intro_play_sound`, adding support for clickable keywords in instructions to show callout arrows.

See https://github.com/code-dot-org/code-dot-org/pull/55758 which replaces `musiclab_intro_video_end` with `musiclab_intro_panels_end`, a new `Panels` level type which shows the user a series of panels that each contain an image and some text.

### Details

Some of the specific tasks done in this PR:

- Made standalone video Lab2 compatible
- Named all levels `musiclab_intro_` variants
- Added missing levels, removed extra ones
- Ordered levels correctly
- Drafted proposed text with lower word count
- Removed use of numbered lists
- Set up toolboxes
- Set up starter code
- Fixed code to support functions in starter code
- Created lesson plan (so levels have friendly names)
- Created `/s/music-intro-2024` (and shallow-cloned existing lesson into it)
- Added rudimentary validations to each level

PR #59193, opened 2024-06-12
[Google Blockly] only toggle definition block state on main workspace

mikeharv:
While working on adding support for functions with parameters, I discovered that function definition blocks are unexpectedly movable from within the modal editor.  This was happening due to some custom logic we had implemented during the block's `loadExtraState` hook which was intended to correctly set the deletable/editable/movable state based on its initial configuration. However, this hook executes _after_ the function editor's `showForFunction`, effectively undoing the work to make it always immovable. Similarly, I was observing that function blocks didn't have the right editable state either. We also shouldn't need to toggle the block's deletability since this is handled by the modal editor's Delete button (when the function is user-created).

A solution is to only toggle these block state values if we're not using the modal function editor.
In addition to fixing this issue, I also tested that it still works as needed in non-editor contacts. Music Lab functions can still be made immovable on the main workspace. https://github.com/code-dot-org/code-dot-org/pull/55648#discussion_r1458134249 Minecraft functions can still be uneditable on the main workspace. https://github.com/code-dot-org/code-dot-org/pull/55993#uneditable-functions

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/dashboard/test/ui/features/code_tools/google_blockly/modal_function_editor.feature
+++ b/dashboard/test/ui/features/code_tools/google_blockly/modal_function_editor.feature
@@ -31,7 +31,7 @@ Scenario: Can edit a function
   # Open Sprites flyout
   And I press "blockly-d"
   # Drag new sprite block to top of function
-  And I drag block number 3 to offset "40, 100"
+  And I drag block number 2 to offset "40, 100"
   # Close function
   And I press "closeModalFunctionEditor"
   And element "#modalFunctionEditor" is not visible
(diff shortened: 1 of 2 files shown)


dashboard/test/ui/features/code_tools/google_blockly/modal_function_editor.feature
```diff
@@ -31,7 +31,7 @@ Scenario: Can edit a function
   # Open Sprites flyout
   And I press "blockly-d"
   # Drag new sprite block to top of function
-  And I drag block number 3 to offset "40, 100"
+  And I drag block number 2 to offset "40, 100"
```
molly-moen, line 34: for my knowledge: why did this change?

molly-moen (approved):

dashboard/test/ui/features/code_tools/google_blockly/modal_function_editor.feature
mikeharv: Oh! I captured this in the commit message, but didn't leave a comment sorry. When we find blocks with this step, we are only counting draggable blocks. Now that the definition block is immovable like it's supposed to be, it no longer counts as a draggable block.

PR #59196, opened 2024-06-12
P20-991: Fix CAP non-compliant students locking

artem-vavilov:
## CAP rules
- New accounts created after `July 1, 2023` are locked out immediately
- Accounts created before `July 1, 2023` are locked out during the all-users lockout phase.

## Links
- JIRA ticket: [P20-991](https://codedotorg.atlassian.net/browse/P20-991)


daynew (approved):
I read through the code changes and the unit tests. This looks like it is mostly a refactoring in order to support the grace_period functionality in a future PR.

dashboard/lib/policies/child_account.rb
```diff
@@ -44,7 +44,6 @@ module ComplianceState
   # parent permission before the student can start using their account.
   def self.compliant?(user)
     return true unless parent_permission_required?(user)
-    return true if user_predates_policy?(user)
```
artem-vavilov, line 47: Because of this, "existing" users (created before 7/1/2023) have always been CAP compliant, ensuring they are never locked out.

juanmanzojr (approved):
LGTM!

PR #59198, opened 2024-06-13
Parameter Field

mikeharv:
Our new parameter blocks are currently using our custom variable field, which serves well enough for looks and code generation. However, parameter blocks need their own menu generator (which provides the dropdown options) and special handling for some of those options. This warrants creating a new class for a parameter field. Currently, this new field extends the base Blockly variable field class. As a follow-up, I'd like to look at having it our custom (CDO) variable field class class.

These screenshots come from CDO Blockly and help us establish our target:
image
image
You'll notice that the (top) variable getter lists all variable options (because Blockly variables are always global). Additionally, there are options to rename all or a single instance of the given variable. By contrast the (bottom) parameter getter block has no variable options, because switching to another random variable isn't guaranteed to make sense since they won't all be valid parameters for the given function. Additionally, the two custom options are different, serving to rename or delete (all instances of) the given parameter.

These screenshot show the work on this branch to give us parity with Google Blockly:
image
image

The following video demonstrates renaming a parameter and then deleting it. When perform these actions, we also need to update (or remove) all the other parameter blocks on the workspace. (Note that the work to add a "new parameters" button is not present on this branch, but is drafted here: https://github.com/code-dot-org/code-dot-org/pull/59201)

https://github.com/code-dot-org/code-dot-org/assets/43474485/5870afbb-8b9d-4237-a600-065ee2ae6ddd

## Renaming
When a parameter is renamed, we first rename the variable model. This updates all blocks that reference that variable. We then change the name of the parameter model within the procedure model context.

## Deleting
When a parameter is deleted, we first find the procedure model and then remove the parameter model. Following this, we remove all blocks that reference the variable.

## Finding the models
Because of the nature of mini-toolboxes, finding the right procedure/parameter/variable is a little tricky. If the parameter block is in the mini-toolbox, its workspace is the one in the flyout (and not the function editor workspace). The relationship hierarchy looks like this:

* Workspace (main or function editor)
  * Procedure map, variable map _<- We need these_
  * Definition block
    * Statement input
      * Other blocks with various block inputs
      * Parameter block(s)
    * Flyout field
      * Block flyout (mini-toolbox)
        * workspace
          * Parameter block(s)
  * Parameter block(s) (disconnected)

In order to delete/rename parameters, we need to be able to access the main active workspace, the definition block, and procedures/parameters/variables. For a block that is part of the funciton definition, we can easily access `block.workspace` or `block.getRootBlock()` to find what we need. If we try this on a flyout block, the workspace we'll get is the _flyout workspace_ which has its own (out of sync) maps for procedures/variables. Fortunately, we can check `block.isInFlyout` to determine what we need to do. I created some helper properties so that the flyout workspace has a reference to the main workspace to make this easier. https://github.com/code-dot-org/code-dot-org/pull/59198/commits/d5f7272f9725dd03bf2117ac73e3fad6a7f47d19
The last case is to handle a parameter block that is in neither of these places (e.g. disconnected, or connected to a root block that isn't a definition). We can't crash if a student tries to rename/delete in this case. If we determine that the root block isn't a valid type, we will search for all definition blocks and use the first one that has a matching parameter. (We can't assume there is only one definition block because the modal function editor isn't always enabled.)

This last edge case is depicted here:

https://github.com/code-dot-org/code-dot-org/assets/43474485/353df203-01be-4bd3-83d1-5a5c0f2a7c3d

## Links

## PR Checklist:

- [ ] Code is well-commented


apps/src/blockly/addons/cdoFieldParameter.ts
```diff
@@ -0,0 +1,262 @@
...
+      }
+    }
+    return {definitionBlock, workspace};
+  }
+  /**
+   * Override of createTextArrow_ to fix the arrow position on Safari.
+   * We need to add dominant-baseline="central" to the arrow element in order to
+   * center it on Safari.
+   *  @override */
+  createTextArrow_() {
```
mikeharv, line 197: This is copied directly from `CdoFieldVariable` and is here so that we don't have issue with dropdown arrows on any browsers. This repeated code is a good argument for "rebasing" this class as a follow-up.

apps/src/blockly/addons/cdoFieldParameter.ts
```diff
@@ -0,0 +1,265 @@
...
+    callback: (newName: string) => void,
+    prompt: boolean,
+    isDangerCancel: boolean,
+    defaultText?: string
+  ) {
+    Blockly.customSimpleDialog({
+      bodyText: promptText,
+      prompt,
+      promptPrefill: defaultText,
+      cancelText: confirmButtonLabel,
```
molly-moen, line 258: why is the cancel text the confirm text?

apps/src/blockly/addons/cdoFieldParameter.ts
```diff
@@ -0,0 +1,265 @@
...
+      // In case the block is disconnected from the definition, manually
+      // reset its field value.
+      this.getSourceBlock()?.setFieldValue(variable.getId(), 'VAR');
+    }
+  }
+  /**
+   * For a given parameter block, find the definition block and its workspace.
+   * If the block is part of the function definition, we want the root (top-most) block.
+   * If this block is in a mini-toolbox, we want the flyout's parent block.
+   * If neither of these are true (e.g. the block is disconnected on the workspace,
```
molly-moen, line 140: super nit: close parentheses after workspace?

molly-moen (approved):
Looks good!

apps/src/blockly/addons/cdoFieldParameter.ts
mikeharv: This struck me as odd too. We do it this way for the variable field as well as for the buttons in the modal function editor:
- https://github.com/code-dot-org/code-dot-org/blob/mike/new-param-button/apps/src/blockly/addons/cdoFieldVariable.ts#L133
- https://github.com/code-dot-org/code-dot-org/blob/mike/new-param-button/apps/src/blockly/addons/functionEditor.ts#L425

It didn't seem like I should break from this pattern without some discussion or potential re-working. The rationale seems to be that the dialog has a `isDangerCancel` property that allows us to make the delete buttons red:

https://github.com/code-dot-org/code-dot-org/blob/mike/new-param-button/apps/src/feedback.js#L1290

apps/src/blockly/addons/cdoFieldParameter.ts
mikeharv: Fix pushed to the other branch so that I could merge this one without waiting for drone again. https://github.com/code-dot-org/code-dot-org/pull/59201/commits/8bcfea38b7dbfc80340b18625c604256eeab5524

PR #55445, opened 2023-12-13
Add empty alt text to images to fix linter errors and reenable `jsx-a11y/alt-text`

kelbyhawn:
Adds empty `alt=""` attributes to images to fix `alt-text` errors in the `jsx-a11y` linter that was added [here](https://github.com/code-dot-org/code-dot-org/pull/55358).

**Followup:** I added TODO comments above each `` element that will be reviewed and updated w/ descriptive text if needed. See related Jira Ticket: [A11Y-279](https://codedotorg.atlassian.net/browse/A11Y-279)

**Jira ticket:** [A11Y-278](https://codedotorg.atlassian.net/browse/A11Y-278)

PR #59209, opened 2024-06-13
Added page for new images

kobryan0619:
This creates a page that allows levelbuilders to upload images to S3 similar to how they can do that for videos.  

Context:  In the past, when a levelbuilder wanted to add an image they had to go through an elaborate workaround for getting a URL to a new image.  This creates a separate place for levelbuilder users to do this.  This also may help some internationalization efforts Jorge has been working on (as discussed in this [thread](https://codedotorg.slack.com/archives/C03DBDN67B7/p1712173242574789)).

The video below shows the page working with two different images to see what the workflow would look like for someone uploading multiple images. 

https://github.com/code-dot-org/code-dot-org/assets/24883357/d25106c6-e4f7-430e-be80-e7e19fbbc4ae

## Future work

This has been added to Liam's document for levelbuilder quick wins as well, but it is important to know that this is just a starting point. 

Extensions include:

- (Extension for “Image Upload page) Allow the user to upload multiple images at one time (or one at a time and show all the URLs).  
- - (Extension for “Image Upload page) Have a button that can copy the URL created automatically 
- (Extension for “Image Upload page) Add the images link to the helpful links page  (note - this one is a fast follow - and is done as of 6/21)
- (Extension for “Image Upload page) Saved notification after saving is done (Note: this is probably least helpful since there is a visual reminder that the image has been created).
- Add library of images?

### DIFF
--- a/apps/src/lib/levelbuilder/lesson-editor/UploadImageForm.jsx
+++ b/apps/src/lib/levelbuilder/lesson-editor/UploadImageForm.jsx
@@ -67,7 +67,7 @@ export default function UploadImageForm() {
       {tempImageUrl && (
         <img
           src={tempImageUrl}
-          alt="uploaded image preview"
+          alt="preview of uploaded document"
           className={styles.imagePreview}
         />
       )}
--- a/apps/src/lib/levelbuilder/lesson-editor/UploadImageForm.jsx
+++ b/apps/src/lib/levelbuilder/lesson-editor/UploadImageForm.jsx
@@ -30,22 +30,6 @@ export default function UploadImageForm() {
     setFormDataForImage(formData);
     setIsUploading(false);
     setTempImageUrl(URL.createObjectURL(e.target.files[0]));
-
-    // // POST
-    // const csrfContainer = document.querySelector('meta[name="csrf-token"]');
-    // fetch('/level_assets/upload', {
-    //   method: 'post',
-    //   body: formData,
-    //   headers: {
-    //     'X-CSRF-Token': csrfContainer && csrfContainer.content,
-    //   },
-    // })
-    //   .then(response => response.json())
-    //   .then(handleResult)
-    //   .catch(err => {
-    //     setError(err);
-    //     setIsUploading(false);
-    //   });
   };
 
   const saveImageToS3 = () => {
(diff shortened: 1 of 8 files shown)


dancodedotorg:
Ahhhhh love this!!! Looks great to me! And love the "step by step" approach we're taking here - yes there's more to do, but this is a net improvement so let's take this win for now!!

dancodedotorg (approved):

apps/src/lib/levelbuilder/lesson-editor/UploadImageForm.jsx
```diff
@@ -0,0 +1,114 @@
...
+    setImgUrl(undefined);
+    setError(undefined);
+  };
+
+  const handleChange = e => {
+    if (!e.target.files[0]) {
+      resetState();
+      return;
+    }
+    resetState();
```
lfryemason:
```suggestion
    resetState();
    if (!e.target.files[0]) {
      return;
    }
```

apps/src/lib/levelbuilder/lesson-editor/UploadImageForm.jsx
```diff
@@ -0,0 +1,114 @@
...
+      setError(result);
+    }
+    setIsUploading(false);
+  };
+
+  return (
+    <div className={styles.topContainer}>
+      <h2>{i18n.uploadImage()}</h2>
+      {
+        // TODO: A11y279 (https://codedotorg.atlassian.net/browse/A11Y-279)
```
lfryemason: Could we just add some alt-text here that makes sense? I assume that this was added because someone didn't want to go through lots of different images to add alt-text, but since we are just adding one, I think we can add it.

lfryemason (commented):
Looks great and I'm so excited to see this!

One question, should we add a link to this page in an appropriate place? It seems like the curriculum writer would need to know the exact URL?

apps/src/lib/levelbuilder/lesson-editor/UploadImageForm.jsx
lfryemason: Maybe something like `uploaded image preview`

kobryan0619: Re: link to this page.  Yes, that is one of the follow ups from this ticket: `(Extension for “Image Upload page) Add the images link to the helpful links page`.  I documented this in your google doc as well and will be an easy fast follow.

kobryan0619: As I understand this comment (which was borrowed from a component that is used to upload images on a specific level) I think this might be a much bigger to do... right now this button just uploads it to S3 - there is no alt text stored on images currently in S3.  

So, I am not sure it is really easy to add alt-text when the image is uploaded to S3.  Instead it seems like alt-text should be added when a levelbuilder decides to use a level OR we need to add a table that includes URLs for images and alt text for those associated levels... again, feels like a much bigger project, unless I misunderstood the comment.

lfryemason: Agreed that would be a much bigger ask. These comments were added as part of [this](https://github.com/code-dot-org/code-dot-org/pull/55445/files) PR and are not all related to not storing alt text. Some seem to be just because Kelby didn't know what alt-text to put and would have needed further exploration.

What do you think about adding a generic alt-text. Something like `uploaded image preview`?

kobryan0619: I see, in that case - yes!  I added this in.

lfryemason (approved):

PR #59219, opened 2024-06-14
P20-987: Add ability to mock DCDO via cookie for testing

artem-vavilov:
## Summary

## Links
- Related PR: https://github.com/code-dot-org/code-dot-org/pull/59174
- Revert PR: https://github.com/code-dot-org/code-dot-org/pull/59216


Gemfile
```diff
@@ -258,6 +258,7 @@ end
 # Reduce volume of production logs
 # Ref: https://github.com/roidrage/lograge/pull/252
 gem 'lograge', github: 'code-dot-org/lograge', ref: 'debug_exceptions'
+gem 'request_store', '~> 1.6.0', require: false
```
artem-vavilov, line 261: We need to allow `request_store` in any group where `lograge` (`debug_exceptions`) is enabled, since `lograge` depends on it.

lib/cdo/rack/cookie_dcdo.rb
```diff
@@ -0,0 +1,34 @@
...
+      RequestStore.store[KEY] = JSON.parse(Rack::Request.new(env).cookies[KEY] || '{}')
+
+      unless DCDO.instance_variable_get(:@_redefined)
+        # Redefines `DCDO#get` to return the cookie DCDO value if it exists.
+        DCDO.define_singleton_method(:get) do |key, *args|
+          RequestStore.store[KEY]&.key?(key) ? RequestStore.store[KEY][key] : super(key, *args)
+        end
+
+        DCDO.instance_variable_set(:@_redefined, true)
+      end
```
artem-vavilov: I usually prefer to avoid metaprogramming, but in this case, I have three reasons for doing so:

1. Since this is intended for testing purposes only, I didn't want to modify the original method, as that would burden it with redundant checks in production and other environments.
2. This dynamic method redefinition will ensure that we override DCDO#get regardless of its origin, even if it isn't inherited from DynamicConfigBase where the original #get method is defined.
3. It's easier to understand the business logic when everything is kept together in one place.

dashboard/test/ui/features/support/cookie_helpers.rb
```diff
@@ -30,6 +30,12 @@ def decrypt_cookie(cookies)
 
     encryptor.decrypt_and_verify(message)
   end
+
+  # It's safer to retrieve a cookie from the list of all cookies
+  # because `#cookie_named` may raise `Selenium::WebDriver::Error::UnknownMethodError` during mobile testing.
+  def get_cookie(key)
+    @browser.manage.all_cookies.find {|cookie| cookie[:name] == key}
+  end
```
artem-vavilov, line 38: The error: https://cucumber-logs.s3.amazonaws.com/test/test/iPad_cookie_dcdo_output.html?versionId=Byg7lZiczljR7d7cQ5gpJNVzxOfKT99Y

pegasus/sites.v3/code.org/public/test_dcdo.haml
```diff
@@ -1,3 +1,3 @@
-= DCDO.get("test_dcdo_on_pegasus", nil).inspect
+#dcdo= DCDO.get("test_dcdo_on_pegasus", nil).inspect
```
Hamms: Was this change intentional?

lib/cdo/rack/cookie_dcdo.rb
Hamms: I suppose it does make sense to keep this test-only functionality isolated from the actual production implementation, but I'm still not a fan of the specific way we're going about it here. In particular, I really don't like that we've implemented the method redefinition somewhere that will be called repeatedly rather than as a onetime initialization.

What if we instead declare a module here in this file which can be prepended to the DCDO class in the initialize method? Similar to how we make modifications to existing classes in our initializers: [1](https://github.com/code-dot-org/code-dot-org/blob/c7f1bc053759520a7228b44727895088bb624148/dashboard/config/initializers/fast_localization.rb#L44), [2](https://github.com/code-dot-org/code-dot-org/blob/c7f1bc053759520a7228b44727895088bb624148/dashboard/config/initializers/mysql_avoid_utf8_alias_in_schema_dump.rb#L58), [3](https://github.com/code-dot-org/code-dot-org/blob/c7f1bc053759520a7228b44727895088bb624148/dashboard/config/initializers/trackable.rb#L21)

Hamms (commented):
Repeating one question I'm still not clear on from the last time I looked at this:

> Can you clarify what purpose RequestStore is serving in this implementation? It looks to me like we're storing the value we want in the cookie, then moving it from the cookie to the RequestStore, then reading it out of the RequestStore. Why aren't we reading the value from the cookie directly? What am I missing?

lib/cdo/rack/cookie_dcdo.rb
artem-vavilov: That's how I implemented this at first:
https://github.com/code-dot-org/code-dot-org/blob/6f3f113491d7a65cd148abdabcfda73d9e77c83e/lib/cdo/rack/cookie_dcdo.rb#L9-L30

pegasus/sites.v3/code.org/public/test_dcdo.haml
artem-vavilov: It's just easier to test this way: https://github.com/code-dot-org/code-dot-org/blob/90488fd4b41b2c15fef26ebcbce607016237124a/dashboard/test/ui/features/cookie_dcdo.feature#L6-L9

lib/cdo/rack/cookie_dcdo.rb
Hamms: I definitely prefer that!

pegasus/sites.v3/code.org/public/test_dcdo.haml
Hamms: D'oh, of course! I missed that connection; adding an identifier here makes sense to me. Could we go for something a little more descriptive, though? Maybe use `#fetched_dcdo_value` for this element and `#raw_dcdo_value` for the other element below the `<br>`, for consistency?

lib/cdo/rack/cookie_dcdo.rb
artem-vavilov: https://github.com/code-dot-org/code-dot-org/pull/59219/commits/b99ecbc8d24870a6f193fb9d1a1d3ed63fd244c7

pegasus/sites.v3/code.org/public/test_dcdo.haml
artem-vavilov: https://github.com/code-dot-org/code-dot-org/pull/59219/commits/62fb3747266683e6aa11be64d67d0a5a186cc845

Hamms (approved):
LGTM! I like how extracting the method redefinition out from the `call` method makes it more clear why the new `DCDO.get` method can't just grab the value it wants from the cookie directly.

Thank you!

PR #59223, opened 2024-06-14
Use `template` EnforcedStyle for `Style/FormatStringToken`

Hamms:
The default style for this cop is `annotated`, which looks like:

```
format('%<greeting>s', greeting: 'Hello')
```

Currently in our codebase, we consistently use the `template` style, which looks like:

```
format('%{greeting}', greeting: 'Hello')
```

Similar to the style used by classic string interpolation. I'm not sure whether this was intentional or if we simply defaulted to this style since most of us are likely more familiar with string interpolation than with the format method, but given how consistent we are (and how many violations there are with the default style), I think it makes sense to codify this behavior.

## Links

See https://github.com/code-dot-org/code-dot-org/pull/59224 for an example of what adopting the default style would look like.

- https://www.rubydoc.info/github/bbatsov/RuboCop/RuboCop/Cop/Style/FormatStringToken
- https://docs.rubocop.org/rubocop/cops_style.html#styleformatstringtoken

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -91,6 +91,9 @@ Style/EmptyElse:
 Style/EmptyMethod:
   EnforcedStyle: expanded
 
+Style/FormatStringToken:
+  EnforcedStyle: template
+
 Style/NumericLiterals:
   MinDigits: 7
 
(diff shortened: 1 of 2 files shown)


bethanyaconnor (approved):
Makes sense to me!

PR #59227, opened 2024-06-14
Gen AI: handle new message key on CSV export

bencodeorg:
Quick follow-up to https://github.com/code-dot-org/code-dot-org/pull/58668, see relevant comment [here](https://github.com/code-dot-org/code-dot-org/pull/58668#issuecomment-2168705151)) -- handles CSV export of chat messages now that we have a different key storing the content (`chatMessageText` rather than `content`) of chat messages.

### DIFF
--- a/bin/oneoff/aichat_logs_to_csv
+++ b/bin/oneoff/aichat_logs_to_csv
@@ -76,7 +76,7 @@ def main
           *base_row,
           message["status"],
           message["role"],
-          message["content"]
+          message["content"] || message["chatMessageText"]
         ]
         csv << row
       end


thomasoniii (approved):
one small question, but otherwise lgtm 🚀

bin/oneoff/aichat_logs_to_csv
```diff
@@ -76,7 +76,7 @@ def main
           *base_row,
           message["status"],
           message["role"],
-          message["content"]
+          message["content"] || message["chatMessageText"]
```
thomasoniii, line 79: This is fine, but just confirming - if the `message["content"]` is falsy, this'll still fall through to `message["chatMessageText"]`.

So we can't have `message["content"]` that is things like `0` or `""` or `"0"` (though, amusingly, we still _would_ allow `" "`).

Anyway, so long as that's the case, then it's good. Otherwise if we'd want to display it, then it should hop to a `??`.

(And if we want to prevent the blanking string cases, then maybe a regex like `/\S/` in a ternary)

bencodeorg: Thanks for raising this, I hadn't really thought it through (sort of a last minute Friday afternoon change here). Did a little Ruby digging, and I think we're actually ok here? Does this look right to you?

image

fisher-alice: Also - I don't think `??` operator is available in Ruby.

fisher-alice (approved):

PR #59230, opened 2024-06-14
Revert "Gen AI: more robust chat logging"

bencodeorg:
Reverts code-dot-org/code-dot-org#58668 -- there's a bug in our logging to Honeybadger that apparently was exercised on the test machine but not in Drone/locally. More detail [here](https://codedotorg.slack.com/archives/C03CM903Y/p1718402178669979?thread_ts=1718402147.403789&cid=C03CM903Y).


bencodeorg:
Merging without drone to unblock test.

PR #59238, opened 2024-06-17
AI Tutor: suggested prompt generic component + functionality 

Erin007:
We want to be able to suggest questions a student might ask AI Tutor, specifically "Why doesn't my code compile?" and "Why aren't my tests passing?". Suggested prompts is also a feature we'll likely want to use across our suite of AI chat tools.  

To this end, I built two simple, reusable components `SuggestedPrompt` and `SuggestedPrompts`.  A `SuggestedPrompt` is basically a clickable tag based on these [Figma designs](https://www.figma.com/design/lOLzGeTfH44a5YDS7GrzdA/AI-Tutor?node-id=2383-5497&t=annyKNmLs0LYQuDX-0) and `SuggestedPrompts` renders a list of individual suggested prompts. 

I then wrote a wrapper for `SuggestedPrompts` specific to AI Tutor, `AITutorSuggestedPrompts` that handles the functionality for prompting the student to ask about compilation and validation.

BEFORE -  Suggested prompts  were baked into a button in the AI Tutor footer below the general chat input box 

Screenshot 2024-06-17 at 11 29 13 AM

AFTER - Suggested prompts show within the chat workspace

https://github.com/code-dot-org/code-dot-org/assets/12300669/eedc5011-8bdc-4b5a-968a-e153b0b9154e

The hide/show logic for the compilation and validation prompts is the same as before. 

## Follow-up work

- I suspect the designs to evolve with continued iteration on AI Tutor 
- I think we're shifting AI Tutor out of Java Lab into the forthcoming Python Lab so the state that determines whether the prompts should show or not will need to change. Consider this a Proof of Concept for now until the dust settles 😁 
- As far as I know there aren't currently any other "Suggested Prompts" in other AI tools that need to be swapped out; can Teacher Tools & GenAI confirm? 
- I didn't touch the Design System `Tag` component because 1.) I'm not sure we'll want this as the final design 2.) I'm not confident we want tags to generally be clickable given their general purpose; nonetheless it's possible there's work to align more closely with the Design System 

## PR Checklist:

- [ ] Code is well-commented


apps/src/aiComponentLibrary/suggestedPrompt/SuggestedPrompt.tsx
```diff
@@ -0,0 +1,28 @@
...
+  show,
+}) => (
+  <>
+    {show ? (
+      <div onClick={onClick} className={moduleStyles.prompt}>
+        <span>{label}</span>
+      </div>
+    ) : null}
+  </>
+);
```
thomasoniii: Not blocking, but I'd suggest moving the conditional outside of the empty block so we can get rid of it entirely.

Further, I'd also recommend inverting the logic so we can cuddle the `null` case more closely to the top and minimize it getting lost with distance from the check.

```
=> (
  !show
    ? null
    : <div onClick={onClick} className={moduleStyles.prompt}>
        <span>{label}</span>
      </div>
```

(or something like that)

apps/src/aiComponentLibrary/suggestedPrompt/SuggestedPrompts.tsx
```diff
@@ -0,0 +1,30 @@
...
+    <div className={moduleStyles.prompts}>
+      {suggestedPrompts.map(prompt => (
+        <SuggestedPrompt
+          onClick={prompt.onClick}
+          label={prompt.label}
+          show={prompt.show}
+        />
+      ))}
+    </div>
+  );
```
thomasoniii: again, non-blocking, but just noting that since it's the single element returned we can drop the explicit block and return statement.

apps/src/aiTutor/views/AITutorFooter.tsx
```diff
@@ -38,73 +29,35 @@ const AITutorFooter: React.FC<AITutorFooterProps> = ({renderAITutor}) => {
...
-  );
+    analyticsReporter.sendEvent(EVENTS.AI_TUTOR_ASK_GENERAL_CHAT, {
+      levelId: level?.id,
+      levelType: level?.type,
+    });
+  }, [userMessage, studentCode, isWaitingForChatResponse, level, dispatch]);
 
   const disabled = !renderAITutor || isWaitingForChatResponse;
 
   const userMessageIsEmpty = userMessage.trim() === '';
```
thomasoniii, line 56: Again, non-blocking, but would it be better to trim when the value is set?

That'd be on line 71:
```
onChange={e => setUserMessage(e.target.value.trim())}
```

Then we wouldn't need to worry about calling trim again later and can just directly check against an empty string or `.length`.

apps/src/aiTutor/views/AITutorSuggestedPrompts.tsx
```diff
@@ -0,0 +1,115 @@
...
+      label: QuickActions[ActionType.VALIDATION],
+      onClick: () => handleClick(ActionType.VALIDATION),
+      show: showValidationOption,
+    },
+  ];
+
+  return (
+    <>
+      <SuggestedPrompts suggestedPrompts={suggestedPrompts} />
+    </>
```
thomasoniii: Can we drop the wrapping `<>` component, or is it needed for something else?

apps/src/aiTutor/views/AITutorSuggestedPrompts.tsx
```diff
@@ -0,0 +1,115 @@
...
+      switch (actionType) {
+        case ActionType.COMPILATION:
+          studentInput = QuickActions[ActionType.COMPILATION];
+          event = EVENTS.AI_TUTOR_ASK_ABOUT_COMPILATION;
+          break;
+        case ActionType.VALIDATION:
+          studentInput = QuickActions[ActionType.VALIDATION];
+          event = EVENTS.AI_TUTOR_ASK_ABOUT_VALIDATION;
+          break;
+      }
```
thomasoniii, line 77: This is fine and I'd probably discourage changing it. So this is purely a note.

But.

I always find it frustrating writing code when I need to use a mutable variable because an approach to using an immutable one isn't apparent. Fortunately, I know a pattern that works here, so I figured I'd just note it for future reference. I'm not convinced it adds enough clarity to actually be worth implementing. Unless the fact that mutable variables exist bugs you too. 😄 

```
const actionDispatch = {
  [ActionType.COMPILATION] : {
    studentInput : QuickActions[ActionType.COMPILATION],
    event: EVENTS.AI_TUTOR_ASK_ABOUT_COMPILATION
  },
  [ActionType.VALIDATION] : {
    studentInput : QuickActions[ActionType.VALIDATION],
    event: EVENTS.AI_TUTOR_ASK_ABOUT_VALIDATION
  }
}

const {studentInput, event} = actionDispatch[actionType]
```

One potential advantage to this approach is that it's just data and the `QuickActions` and `EVENTS` objects are already defined externally, so the `actionDispatch` (or whatever it's called) could be defined external to the component and not inlined at all here, which might improve readability.

The counter argument being that it's also moving the lookup farther away, which could be detrimental. So YMMV.

thomasoniii (approved):
added some minor non-blocking formatting type suggestions, but otherwise lgtm 🚀

apps/src/aiTutor/views/AITutorSuggestedPrompts.tsx
Erin007: Thanks for the info! I'm going to leave as is for now, but appreciate the attention to readability and explanation.

apps/src/aiTutor/views/AITutorFooter.tsx
Erin007: I see where you're coming from, but I don't want to prematurely trim the whitespace within the `onChange` of the text box because then the user can't put any spaces in their text. I endupwithsomethinglikethis 😬

davidsbailey:
> * As far as I know there aren't currently any other "Suggested Prompts" in other AI tools that need to be swapped out; can Teacher Tools & GenAI confirm?

Teacher Tools does not have suggested prompts today, but we are planning to use them in differentiation chat (see [requirements](https://docs.google.com/document/d/1vqNvCDKTykDL_qn_yHBW6CgJcD9zEp0QQFnj0CeHnNE/edit#heading=h.58rkn11oqmhn)), and so will appreciate having a shared component which facilitates this

davidsbailey (approved):

PR #59240, opened 2024-06-17
Revert "Revert "Gen AI: more robust chat logging""

bencodeorg:
Revives a refactor of our Gen AI logging code (https://github.com/code-dot-org/code-dot-org/pull/58668), which had to be reverted due to a failing unit test in a failed DTT -- [more detail in this thread](https://codedotorg.slack.com/archives/C03CM903Y/p1718402178669979?thread_ts=1718402147.403789&cid=C03CM903Y).

New code is in this commit: https://github.com/code-dot-org/code-dot-org/pull/59240/commits/a8d455d26eb4334bd3db6a2a903410d677d7dd33.

An undeclared variable used in some Honeybadger logging didn't show itself until being run during the DTT (I guess we don't exercise Honeybadger logging in other envs)? Updated to support generating a logging session at the point of logging to Honeybadger, then reusing that session when we send a response to the user.

Learned a couple things here:
- Ruby text editors/linters can't necessarily identify undeclared variables. [More discussion here](https://www.reddit.com/r/ruby/comments/10k4nug/can_text_editors_detect_undefined_variables_in/).
- Rails controller tests reuse the same controller instance across multiple requests 🤔  -- [other developers also confused by this choice here.](https://github.com/rails/rails/issues/24566)


fisher-alice (approved):
Thanks for the explanations/links and sharing your learnings.

PR #59241, opened 2024-06-17
P20-965: Add transition to grace period

artem-vavilov:
## Summary
- Implemented the transition of CAP non-compliant "pre-policy" students to the grace period phase.
- Implemented a scheduling lockout for students after their grace period ends.
- Implemented logging for the grace period start event.
- Updated the Parental Permission banner/modal to new conditions.

## Links
- JIRA ticket: [P20-965](https://codedotorg.atlassian.net/browse/P20-965)
- Related PR: https://github.com/code-dot-org/code-dot-org/pull/59192
- DOC: https://docs.google.com/document/d/1cxE_FeAR7I5mXepCaBr4EFVhTW9KBWWygjE0cg_pM08/edit


dashboard/app/controllers/application_controller.rb
```diff
@@ -353,7 +356,16 @@ def prevent_caching
       users_set_student_information_path,
     ].include?(request.path)
 
-    redirect_to lockout_path if Policies::ChildAccount::ComplianceState.locked_out?(current_user)
+    redirect_to lockout_path
+  rescue StandardError => exception
+    Honeybadger.notify(
+      exception,
+      error_message: 'Failed to lock out user',
```
daynew: Couldn't this also be a failure to transition to grace_period?

Maybe something like "Failed to apply the Child Account Policy to the user"

dashboard/app/jobs/cap/lockout_job.rb
```diff
@@ -0,0 +1,23 @@
+module CAP
+  class LockoutJob < ApplicationJob
+    rescue_from StandardError, with: :report_exception
+
+    # @param user_id [Integer] the student account ID
+    def perform(user_id:)
+      user = User.find(user_id)
+      Services::ChildAccount::LockoutHandler.call(user: user)
```
daynew: Should the job requeue the job if the calculated locked_out date is in the future? For example, maybe the job was queued up and then we changed the DCDO `cpa_schedule` configuration.

Maybe that would complicate this implementation to support an edge-case that probably won't happen...

dashboard/app/jobs/cap/lockout_job.rb
```diff
@@ -0,0 +1,23 @@
+module CAP
+  class LockoutJob < ApplicationJob
```
daynew, line 5:
```suggestion
  # Attempts to transition the User to the locked-out state if they
  # should no longer be able to access Code.org due to their State's 
  # Child Account Policy.
  class LockoutJob < ApplicationJob
```
Consider adding some documentation so other devs can quickly understand the purpose of this job.

dashboard/app/jobs/cap/lockout_job.rb
```diff
@@ -0,0 +1,23 @@
...
+    rescue_from StandardError, with: :report_exception
+
+    # @param user_id [Integer] the student account ID
+    def perform(user_id:)
+      user = User.find(user_id)
+      Services::ChildAccount::LockoutHandler.call(user: user)
+    end
+
+    private def report_exception(exception)
```
daynew, line 36: I wonder if this should be in the parent class ApplicationJob...

dashboard/lib/policies/child_account.rb
```diff
@@ -85,27 +85,34 @@ def self.pre_lockout_user?(user)
...
 
-    # CAP non-compliant students who were created:
-    # - before the policy took effect - should be locked out during the all users lockout phase.
-    # - after the policy took effect - should be locked out immediately.
-    user_predates_policy?(user) ? user_state_policy[:lockout_date] : user_state_policy[:start_date]
+    grace_period_duration = user_state_policy[:grace_period_duration]
+    return unless grace_period_duration
+
+    start_date = DateTime.parse(user.child_account_compliance_state_last_updated) if ComplianceState.grace_period?(user)
+    start_date = user_state_policy[:lockout_date] if approximate && start_date.nil?
```
daynew, line 96:
```suggestion
    start_date ||= user_state_policy[:lockout_date] if approximate
```
This would be the same, right?

dashboard/lib/policies/child_account.rb
```diff
@@ -85,27 +85,34 @@ def self.pre_lockout_user?(user)
...
 
-  # Checks if the user can be locked out due to non-compliance with CAP.
-  def self.lockable?(user)
-    return false if ComplianceState.locked_out?(user)
+  # The date on which the student's account will be locked if the account is not compliant.
+  # @param user [User] the student account
+  # @param approximate [Boolean] if true, return an approximate date if the exact date is not known
+  def self.lockout_date(user, approximate: false)
+    return if compliant?(user)
+    return if ComplianceState.locked_out?(user)
```
daynew, line 106: Why don't we return a date if the user is already locked out? If there is a good reason, we should probably add that to the method documentation above because that is unexpected behavior to me.

dashboard/app/helpers/child_account_helper.rb
```diff
@@ -4,14 +4,11 @@ module ChildAccountHelper
...
   # @return [Hash, nil]
   def parental_permission_banner_data(user, request)
-    student_lockout_date = Policies::ChildAccount.lockout_date(user)
+    student_lockout_date = Policies::ChildAccount.lockout_date(user, approximate: true)
 
-    if request.params.key?(:show_parental_permission_banner)
-      student_lockout_date ||= Cpa::ALL_USER_LOCKOUT_DATE
-    else
-      return unless student_lockout_date
-      return unless Cpa.cpa_experience(request) == Cpa::ALL_USER_LOCKOUT_WARNING
```
daynew, line 13: Since we removed this check, does that mean these banners will start to be shown?

dashboard/lib/services/child_account/grace_period_handler.rb
```diff
@@ -0,0 +1,45 @@
...
+
+        true
+      end
+
+      # Checks if the user can be transited to the grace period state.
+      private def eligible?
+        return false if Policies::ChildAccount.compliant?(user)
+        return false if Policies::ChildAccount::ComplianceState.locked_out?(user)
+        # If the user is already in the grace period, it cannot be started again.
+        return false if Policies::ChildAccount::ComplianceState.grace_period?(user)
```
daynew: Rather then checking all the states they shouldn't be, would it be simpler to instead be explicit about the state they can be?
```suggestion
        return false unless [nil, 's'].include?(user.child_account_compliance_state)
```
A user can only transition from `nil` -> `grace_period` or `request_sent` -> `grace_period`

daynew: I noticed the LockoutHandler returns `true` if the user is locked_out. Should the GracePeriodHandler return `true` if the user is already in the grace_period?

dashboard/lib/services/child_account/lockout_handler.rb
```diff
@@ -0,0 +1,25 @@
...
+
+      # @param user [User] the student account
+      def initialize(user:)
+        @user = user
+      end
+
+      # @return [true,false] true if the user is locked out, false otherwise
+      def call
+        return true if Policies::ChildAccount::ComplianceState.locked_out?(user)
+        return false if Policies::ChildAccount.compliant?(user)
```
daynew, line 14: Similar comment as above, should we be explicit about what start states are acceptable? (`nil`, `grace_period`, `request_sent`)? Then there is no chance of going from `permission_granted` to `locked_out` if a bug is introduced in `compliant?`.

dashboard/test/controllers/application_controller_test.rb
```diff
@@ -63,21 +63,50 @@ class ApplicationControllerTest < ActionDispatch::IntegrationTest
...
-    context 'when user is not lockable' do
-      let(:user_lockable?) {false}
+    context 'when user is locked out' do
+      let(:user_is_locked_out?) {false}
 
-      it 'does not lock out user' do
-        Services::ChildAccount.expects(:lock_out).with(user).never
+      it 'does not redirect to lockout page' do
         get root_path
+        refute_redirect_to lockout_path
```
daynew: Is the context description on this correct? The context and the tests don't seem to match. The context says the user is locked out but the text asserts the the user is *not* locked out.

dashboard/test/controllers/application_controller_test.rb
```diff
@@ -63,21 +63,50 @@ class ApplicationControllerTest < ActionDispatch::IntegrationTest
...
+      end
+    end
+
+    context 'when error is raised during locking out' do
+      let(:error) {StandardError.new('expected_error')}
+
+      before do
+        expect_lockout_handler_call.never
+        Services::ChildAccount::LockoutHandler.expects(:call).with(user: user).raises(error)
```
daynew:
```suggestion
        expect_lockout_handler_call.never
        expect_lockout_handler_call.raises(error)
```
Does this work? Is the `.never` necessary?

dashboard/test/lib/policies/child_account_test.rb
```diff
@@ -222,12 +228,94 @@ class ComplianceStateTest < ActiveSupport::TestCase
...
+        child_account_compliance_state_last_updated: user_cap_compliance_state_updated_at
+      )
+    end
+
+    before do
+      Policies::ChildAccount.stubs(:state_policy).with(user).returns(user_state_policy)
+    end
+
+    it 'returns end date of user grace period based on its start time' do
+      _(grace_period_end_date).must_be_close_to user_cap_compliance_state_updated_at.since(user_state_policy_grace_period_duration), 1.second
```
daynew, line 204: Out of curiosity, why `must_be_close_to` instead of an exact match? This is my first time seeing it.

dashboard/test/lib/policies/child_account_test.rb
```diff
@@ -222,12 +228,94 @@ class ComplianceStateTest < ActiveSupport::TestCase
...
+    context 'when user is not affected by state policy' do
+      let(:user_state_policy) {nil}
+
+      it 'returns nil' do
+        _(grace_period_end_date).must_be_nil
+      end
+    end
+
+    describe 'approximate' do
+      let(:grace_period_end_date) {Policies::ChildAccount.grace_period_end_date(user, approximate: true)}
```
daynew, line 232: should there be a test for `approximate: false`?

daynew (commented):
The Handler pattern is interesting. Could you link me to some reading material about it?

I also have a lot of questions to better help me understand these changes. Thank you!

dashboard/app/controllers/application_controller.rb
artem-vavilov: Correct, this could also be a failure to transition to the "grace period".

I added the message "Failed to lock out user" because this reporting applies to the entire "handle_cap_lockout" callback and the grace period transition attempt is just a part of this process.

We can replace it with a more general message or add a separate reporter specifically about the grace period transition failing.

dashboard/app/jobs/cap/lockout_job.rb
artem-vavilov: It's better for each job to handle its errors individually, as exceptions could be part of the job retry mechanism.

artem-vavilov: Here is a nice example:
https://github.com/code-dot-org/code-dot-org/blob/82de84516accaa49d7df59bcaa2885d2ab65c685/dashboard/app/jobs/evaluate_rubric_job.rb#L282-L284
https://github.com/code-dot-org/code-dot-org/blob/82de84516accaa49d7df59bcaa2885d2ab65c685/dashboard/app/jobs/evaluate_rubric_job.rb#L147-L151

dashboard/lib/policies/child_account.rb
artem-vavilov: Correct. I wasn't sure which variant to use, so I chose the one I found easier to understand from the first reading.

And here I used another approach:
https://github.com/code-dot-org/code-dot-org/blob/359e66f2be0c0bf68cf4da7383b2048516f96b26/dashboard/app/helpers/child_account_helper.rb#L9

dashboard/lib/policies/child_account.rb
artem-vavilov: As the method description states, the method returns the date when the account will (or should) be locked out, not when it was locked out.
https://github.com/code-dot-org/code-dot-org/blob/38ff97f57d25cdd6ea6ee749d4e3c644ae019209/dashboard/lib/policies/child_account.rb#L104-L107

This is useful in various scenarios, such as scheduling the account lockout job or determining if/when the Parental Permission banner/modal needs to be displayed to the user:
1. https://github.com/code-dot-org/code-dot-org/blob/359e66f2be0c0bf68cf4da7383b2048516f96b26/dashboard/lib/services/child_account/grace_period_handler.rb#L19-L22
2. https://github.com/code-dot-org/code-dot-org/blob/359e66f2be0c0bf68cf4da7383b2048516f96b26/dashboard/app/helpers/child_account_helper.rb#L6-L15

dashboard/app/helpers/child_account_helper.rb
artem-vavilov: Correct, we'll start displaying the banner/modal immediately after deployment.
If necessary, we could add another DCDO flag to manage this.

dashboard/lib/services/child_account/grace_period_handler.rb
artem-vavilov: "It depends on how we intend to use these services.

`LockoutHandler` returns true for already locked-out users because we use it not only to lock out the user but also to identify their locked-out state: https://github.com/code-dot-org/code-dot-org/blob/38ff97f57d25cdd6ea6ee749d4e3c644ae019209/dashboard/app/controllers/application_controller.rb#L342-L343

However, in `GracePeriodHandler`, we only need to determine whether the user should be transitioned to the grace period. We could make it return true for consistency with `LockoutHandler`, but this isn't currently used.

artem-vavilov: It depends on how we intend to use a service.

`LockoutHandler` returns true for already locked-out users because we use it not only to lock out the user but also to identify their locked-out state: https://github.com/code-dot-org/code-dot-org/blob/38ff97f57d25cdd6ea6ee749d4e3c644ae019209/dashboard/app/controllers/application_controller.rb#L342-L343

However, in `GracePeriodHandler`, we only need to determine whether the user should be transitioned to the grace period. We could make it return true for consistency with `LockoutHandler`, but this isn't currently used: https://github.com/code-dot-org/code-dot-org/blob/38ff97f57d25cdd6ea6ee749d4e3c644ae019209/dashboard/app/controllers/application_controller.rb#L339-L340

dashboard/test/controllers/application_controller_test.rb
artem-vavilov: https://github.com/code-dot-org/code-dot-org/pull/59241/commits/c55806da36e738f0ce17077ca322ac0d3d42b43b

dashboard/test/controllers/application_controller_test.rb
artem-vavilov: Unfortunately, `expect_lockout_handler_call` cannot be stubbed to raise any errors because it's already stubbed to return a value: https://github.com/code-dot-org/code-dot-org/blob/38ff97f57d25cdd6ea6ee749d4e3c644ae019209/dashboard/test/controllers/application_controller_test.rb#L17

dashboard/test/lib/policies/child_account_test.rb
artem-vavilov: There is already one; it is the main test case, as approximate is false by default: https://github.com/code-dot-org/code-dot-org/blob/38ff97f57d25cdd6ea6ee749d4e3c644ae019209/dashboard/test/lib/policies/child_account_test.rb#L258-L260

dashboard/test/lib/policies/child_account_test.rb
artem-vavilov: The issue is that `User#child_account_compliance_state_last_updated` loses milliseconds during serialization into a string when saved as a JSON property.

One approach to compare dates without considering milliseconds is to use `must_be_close_to` with a 1-second difference.

dashboard/test/controllers/application_controller_test.rb
artem-vavilov: The refactoring of the tests: https://github.com/code-dot-org/code-dot-org/pull/59241/commits/fca955a5a0f8474abbf51a750fd77ae3e916f25c

dashboard/lib/services/child_account/grace_period_handler.rb
artem-vavilov: I believe a more important question is: should we clear the "grace period" state in `GracePeriodHandler` if the user becomes CAP compliant during the grace period?

artem-vavilov: It's a bit more complicated: https://github.com/code-dot-org/code-dot-org/pull/59241/commits/cfd1806231b86bd4cb302f143818ea3eed806a46

artem-vavilov:
> The Handler pattern is interesting. Could you link me to some reading material about it?

It's not a commonly used pattern; it's simply a service object that encapsulates business logic related to some specific subject.

dashboard/lib/policies/child_account.rb
artem-vavilov: I can rename the method to #estimated_lockout_date if it makes it clearer.

artem-vavilov: I can rename the method to `#estimated_lockout_date` if it makes it clearer.

artem-vavilov: I can rename the method to #estimated_lockout_date if this makes it clearer.

artem-vavilov: I can rename the method to `#estimated_lockout_date` if this makes it clearer.

dashboard/app/controllers/application_controller.rb
daynew: The first person to see these messages is likely going to be the DOTD, so we want a high level description of the effect on users so they can quickly understand which team to notify.

dashboard/app/jobs/cap/lockout_job.rb
daynew: Thank you for the examples

dashboard/test/lib/policies/child_account_test.rb
daynew: Oops I missed that!

dashboard/lib/policies/child_account.rb
daynew: How about `upcoming_lockout_date`. To me, that makes it clear it only returns dates in the future.

dashboard/app/helpers/child_account_helper.rb
daynew: Yes, please add a DCDO flag for these features then. Please add them the [CPA README](https://docs.google.com/document/d/1cxE_FeAR7I5mXepCaBr4EFVhTW9KBWWygjE0cg_pM08/edit#heading=h.1wy1wvf8zc7y) as well (I added a TODO for you).

dashboard/app/controllers/application_controller.rb
artem-vavilov: What would you suggest naming it then?

artem-vavilov: How would you suggest describing this so the DOTD understands which team to notify?

dashboard/lib/policies/child_account.rb
artem-vavilov: This date may not always be in the future, so I'm not sure

dashboard/lib/services/child_account/grace_period_handler.rb
artem-vavilov: https://github.com/code-dot-org/code-dot-org/pull/59241/commits/cfd1806231b86bd4cb302f143818ea3eed806a46

dashboard/test/lib/services/child_account/grace_period_handler_test.rb
```diff
@@ -13,138 +13,148 @@ class Services::ChildAccount::GracePeriodHandlerTest < ActiveSupport::TestCase
...
+        let(:user_is_in_grace_period?) {true}
+
+        it 'ends grace period for user' do
+          expect_grace_period_starting.never
+          expect_cap_compliance_clearing.once
 
-      it 'returns false' do
-        _(eligible?).must_equal false
+          handle_user_grace_period_transition
+        end
```
daynew: Nice catch, this is a transition I hadn't considered. `grace_period` -> `nil` if we haven't started the all user lockout phase yet.

dashboard/lib/policies/child_account.rb
daynew: Hmm I think I understand a little but better. Is this the question the method is trying to answer: "What date should the user be locked out on?" That's pretty close to what you wrote in the documentation for the method.

I was confused because the original name, `lockout_date` makes it sound like a simple getter method like `birthday` or `name`. 

I have spent way too long thing about this. I can't really think of something better than `lockout_date`. I don't think you need to change anything.

dashboard/lib/services/child_account/grace_period_handler.rb
```diff
@@ -0,0 +1,46 @@
...
+        return if Policies::ChildAccount::ComplianceState.locked_out?(user)
+        # Only "pre-policy" created students can be transited to the grace period.
+        return unless Policies::ChildAccount.user_predates_policy?(user)
+        # Users can be transitioned to the grace period only after the "all users' lockout" phase has started.
+        return if all_users_lockout_start_date.nil? || all_users_lockout_start_date > DateTime.now
+
+        if Policies::ChildAccount.compliant?(user)
+          # Removes the "grace period" state if the user is now CAP compliant.
+          Services::ChildAccount.remove_compliance(user) if Policies::ChildAccount::ComplianceState.grace_period?(user)
+        else
```
daynew: This defines a transition **leaving** the `grace_period` state. It seems inconsistent because I was expecting GracePeriodHandler and LockedOutHandler to only deal with transitions going **into** their respective states. For example, there is a transition from `grace_period` -> `locked_out` and that transition is handled in the LockedOutHandler (not the GracePeriodHandler).
However, this is an edge case and maybe writing a whole new Handler would be excessive...

dashboard/app/controllers/application_controller.rb
daynew: "Failed to apply the Child Account Policy to the user. A user's account might not comply with our data privacy policies and should be investigated."

The note that it deals with accounts and privacy should be enough info to figure out which team owns resolving this issue.

dashboard/lib/policies/child_account.rb
daynew: Cool, I just saw us using `||=` a lot, so I felt like I should point out we could do it here too. Just a nit-pic.

dashboard/lib/services/child_account/grace_period_handler.rb
daynew: I read through the edge case you pointed out. I still wonder if the code would be easier to read if it was explicit about the start and end states. For example (but you don't need to copy this exactly):
```
if (start_state_1)
  transition(new_state) if (should_transition_from_state_1)
elsif(start_state_2)
  transition(new_state) if (should_transition_from_state_2)
...
```
Ultimately, this is feedback about readability, not functionality. If you think the code would be harder to read, then go ahead and leave it how you have it. I just wanted to offer a different perspective on how it could have been written.

daynew: > I believe a more important question is: should we clear the "grace period" state in GracePeriodHandler if the user becomes CAP compliant during the grace period?

That's a good catch. It's probably good to have it because it's a possible transition if we mess up the launch.

daynew (commented):
Overall looks good, just minor comments about code readability.
The only blocking change is having the DCDO flags and documentation, which I know you are working on now. After those are in, I will give approval.

dashboard/lib/services/child_account/grace_period_handler.rb
artem-vavilov: The `GracePeriodHandler` and `LockedOutHandler` are not limited to handling transitions into their respective states. They encompass all business logic related to the "grace period" and "locked out" states in general.

dashboard/app/jobs/cap/lockout_job.rb
artem-vavilov: https://github.com/code-dot-org/code-dot-org/pull/59241

artem-vavilov: https://github.com/code-dot-org/code-dot-org/pull/59241/commits/e3daef723810f073023a16729064f15c350ee46a

dashboard/app/controllers/application_controller.rb
artem-vavilov: Isn't this message a bit too long? Can we shorten it to just `"Failed to apply the Child Account Policy to the user"`?

dashboard/app/helpers/child_account_helper.rb
artem-vavilov: Added the DCDO `cap_student_warnings_enabled` to manage CAP student warnings: https://github.com/code-dot-org/code-dot-org/pull/59241/commits/024d98dc127e1f2fca343fc160271678b60ece87

artem-vavilov: Removed unused CPA phase "all users lockout warning": https://github.com/code-dot-org/code-dot-org/pull/59241/commits/dc22ae96bd169f38b4ae65c6a3397490a6960b55

dashboard/app/controllers/application_controller.rb
artem-vavilov: https://github.com/code-dot-org/code-dot-org/pull/59241/commits/1469cc5976a972ac733d2991f2b792ed61cb3ae4

dashboard/lib/services/child_account/grace_period_handler.rb
artem-vavilov: I'm not sure I understand this correctly, but doesn't this add even more conditions that need to be updated each time a new state is added?
I think the only thing that could really help is moving everything to be managed by a state machine.

dashboard/app/jobs/cap/lockout_job.rb
```diff
@@ -1,14 +1,36 @@
...
+    # @param reschedules [Integer] the number of times the job can be rescheduled due to a lockout failure
+    def perform(user_id:, reschedules: 0)
       user = User.find(user_id)
-      Services::ChildAccount::LockoutHandler.call(user: user)
+
+      user_is_locked_out = Services::ChildAccount::LockoutHandler.call(user: user)
+
+      # If the user was not locked out, attempt to reschedule the lockout job
+      # in case the reason is a re-estimation of the user's lockout date.
+      self.class.schedule_for(user, reschedules: reschedules.pred) unless user_is_locked_out || reschedules <= 0
```
daynew, line 33: Will this result in the job getting rescheduled if the user already has permission?
https://github.com/code-dot-org/code-dot-org/blob/38ff97f57d25cdd6ea6ee749d4e3c644ae019209/dashboard/lib/services/child_account/lockout_handler.rb#L14

dashboard/app/helpers/child_account_helper.rb
```diff
@@ -6,9 +6,12 @@ module ChildAccountHelper
...
     student_lockout_date = Policies::ChildAccount.lockout_date(user, approximate: true)
 
-    student_lockout_date ||= Cpa::ALL_USER_LOCKOUT_DATE if request.params.key?(:show_parental_permission_banner)
-
-    return unless student_lockout_date
+    if request.params.key?(:show_parental_permission_banner)
+      student_lockout_date ||= Cpa::ALL_USER_LOCKOUT_DATE
+    else
+      return unless student_lockout_date
+      return unless DCDO.get('cap_student_warnings_enabled', false)
```
daynew, line 13:
```suggestion
      return unless DCDO.get('cap_student_warnings_enabled', !Rails.env.production?)
```
I wonder if it would be a good idea to set defaults this way so dev, test, and staging have the features running by default.

dashboard/lib/cpa.rb
```diff
@@ -30,7 +29,6 @@ def self.cpa_experience(current_request, current_time: DateTime.now)
     # schedule [Map] A map of the CPA phases to dates. Example:
     # {
     #   “cpa_new_user_lockout”:         “2023-07-05T23:15:00+00:00”,
-    #   “cpa_all_user_lockout_warning”: “2024-05-01T00:00:00MDT”,
```
daynew, line 37:
```
# 3. Existing accounts warned of lockout - 'cpa_all_user_lockout_warning'
```
On line 17 above also needs to be deleted.

daynew (approved):
Looks good! Thanks for explaining a lot of this.

daynew (approved):
Looks good. Be careful with the merge conflict, it might be the same one I messed up when preparing our bug bash environment.

dashboard/app/jobs/cap/lockout_job.rb
artem-vavilov: No, because `Policies::ChildAccount.lockout_date` will return `nil` if the student is CAP-compliant or already locked out.

artem-vavilov: `self.class.schedule_for` will be called, but the job will not be scheduled because `Policies::ChildAccount.lockout_date` returns `nil` if the student is either CAP-compliant or already locked out.

https://github.com/code-dot-org/code-dot-org/blob/1469cc5976a972ac733d2991f2b792ed61cb3ae4/dashboard/app/jobs/cap/lockout_job.rb#L15-L16

daynew: Oh yeah and we already had a long discussion about that in this PR, sorry for forgetting!

daynew (approved):
Nice job iterating all the edge cases we have been talking about for the past month: `dashboard/test/integration/cap/cpa/lockout_test.rb`

dashboard/app/helpers/child_account_helper.rb
artem-vavilov: I also thought about this, but I wasn't sure if we want this

PR #59246, opened 2024-06-17
Add Localization Strings for Login Lockout

Hamms:
In preparation for eventually adding functionality which will lock teacher accounts after an excessive number of failed login attempts (see https://github.com/code-dot-org/code-dot-org/pull/57515), we first add some necessary user-facing strings for that functionality, so they can be translated in advance of the feature being enabled for users.

### High-level overview of new functionality

When a user repeatedly enters incorrect authentication credentials for the same account within a configurable window of time (default 1 hour), with this new feature in place they will after a configurable number of failures (default 10) see a "last attempt" message warning them that their account will be locked on the next failed attempt.

If they fail another attempt after seeing the message, their account will be temporarily locked. We can configure a variety of unlock strategies, but the one I'm recommending we use is a combination of the timeout and email strategies; the timeout strategy will automatically unlock their account after a configurable amount of time (default 1 hour), and the email strategy will send an email to the user with a link they can click to unlock the account immediately.

Because we are only able to reliable send emails to teachers (and because they are the users whose accounts have access to more sensitive information), right now we are scoping this functionality to only apply to teacher accounts. Student accounts will remain vulnerable to brute-force password attacks, but at least the teachers who manage those accounts will have their own accounts protected.

### Screenshots

string(s) | screenshot
--- | ---
last attempt | image
account locked | image
email |image
email link clicked | image
email link clicked after account already unlocked | image


dashboard/config/locales/devise.en.yml
```diff
@@ -11,7 +11,8 @@ en:
       inactive: "Your account was not activated yet."
       invalid: "Invalid email or password."
       invalid_token: "Invalid authentication token."
-      locked: "Your account is locked."
+      locked: "Your account is locked. You will receive an email with instructions about how to unlock your account in a few minutes."
```
Hamms: Other things we could possibly do here:

- inform them that their account will be unlocked over time even if they don't click on the link (it's a configurable amount of time; default is one hour)
- link them to the "resend unlock instructions" form, although it's not particularly user-friendly

cat5inthecradle:
For comparison, here is our password reset email
Screenshot from 2024-06-17 14-49-53


Given that, I might suggest similar formatting and fully expanding the link. I don't think there are any time limits? If so, then we should mention that.

Hamms:
There isn't a time limit per se, but we do have the option to unlock their account after a configurable amount of time has passed. Probably a good idea to mention that in the email

dashboard/config/locales/devise.en.yml
cat5inthecradle: I think less detail is good here, even though people can read the source code or this PR to discover the timeout.

carl-codeorg (approved):

PR #59247, opened 2024-06-17
Kaitie/add open and close all button

kobryan0619:
Opens and closes rows for all students.

https://github.com/code-dot-org/code-dot-org/assets/24883357/e4fd509d-d02c-47bc-85e8-155a5feeffcc

Issue: The dropdown is not tab navigable.  Molly said Denys might be able to support in this.   [This ticket documents this work. ](https://codedotorg.atlassian.net/jira/software/projects/TEACH/boards/65?selectedIssue=TEACH-1166)

## Links

[Figma](https://www.figma.com/design/gCktlnafGYQSNJafD3FUgA/Progress-Tracking?node-id=5155-7439&m=dev)
[Ticket](https://codedotorg.atlassian.net/jira/software/projects/TEACH/boards/65?selectedIssue=TEACH-1081)

## Testing story

Tests included


apps/src/templates/sectionProgressV2/OpenAllStudentProgressButton.jsx
```diff
@@ -0,0 +1,149 @@
...
+  };
+
+  const collapseMetaDataForAllStudents = () => {
+    collapseMetadataForStudents(getAllStudentIds);
+    setOpened(false);
+  };
+
+  const getMenuLocation = () => {
+    const rect = elementRef.current.firstChild.getBoundingClientRect();
+    console.log(
```
lfryemason: Probably don't need the console log anymore?

apps/src/templates/sectionProgressV2/OpenAllStudentProgressButton.jsx
```diff
@@ -0,0 +1,149 @@
...
+  students,
+  expandMetadataForStudents,
+  collapseMetadataForStudents,
+}) {
+  const [opened, setOpened] = useState(false);
+  const [menuLocation, setMenuLocation] = useState({menuTop: 0, menuLeft: 0});
+  const elementRef = useRef(null);
+  const resizeListener = useRef(null);
+
+  const getAllStudentIds = students.map(student => student.id);
```
lfryemason:
```suggestion
  const allStudentIds = React.useMemo(() => students.map(student => student.id));
```

If we useMemo than we only have to compute this list once and it will be cached.

apps/src/templates/sectionProgressV2/OpenAllStudentProgressButton.jsx
```diff
@@ -0,0 +1,149 @@
...
+      }
+    };
+  }, [updateMenuLocation, opened]);
+
+  const handleDropdownClick = () => {
+    setOpened(!opened);
+  };
+
+  const closeMenu = () => {
+    if (opened) setOpened(false);
```
lfryemason:
```suggestion
    setOpened(false);
```
It will be the same if we just call setOpened to false

apps/src/templates/sectionProgressV2/OpenAllStudentProgressButton.jsx
```diff
@@ -0,0 +1,149 @@
...
+  };
+
+  const onClose = () => {
+    closeMenu();
```
lfryemason: Could we just combine this method with `closeMenu`?

apps/src/templates/sectionProgressV2/OpenAllStudentProgressButton.jsx
```diff
@@ -0,0 +1,149 @@
+import throttle from 'lodash/debounce';
+import PropTypes from 'prop-types';
+import React, {useState, useEffect, useRef} from 'react';
+import {connect} from 'react-redux';
+
+import PopUpMenu from '@cdo/apps/lib/ui/PopUpMenu';
```
lfryemason: Could you explain why we are using PopUpMenu instead of SimpleDropdown or a custom component like SimpleDropdown that uses `<select>`?

apps/src/templates/sectionProgressV2/SectionProgressV2.jsx
```diff
@@ -101,6 +102,7 @@ function SectionProgressV2({
           {i18n.lessonsIn()}
 
           <UnitSelectorV2 className={styles.titleUnitSelectorDropdown} />
+          <OpenAllStudentProgressButton />
```
lfryemason: Should this be `<MoreOptionsDropdown>` or similar instead?

apps/src/templates/sectionProgressV2/OpenAllStudentProgressButton.jsx
```diff
@@ -0,0 +1,149 @@
...
+          isOpen={opened}
+          targetPoint={targetPoint}
+          offset={{x: -elementRef.current.firstChild.offsetWidth, y: 5}}
+          onClose={onClose}
+          className={style.menu}
+        >
+          <PopUpMenu.Item
+            className={style.menuItem}
+            onClick={expandMetaDataForAllStudents}
+            style={{marginTop: '4px', padding: '5px 14px'}}
```
lfryemason: Can we use SCSS instead of style?

lfryemason (commented):
Thanks for working on this!

apps/src/templates/sectionProgressV2/OpenAllStudentProgressButton.jsx
kobryan0619: I started with `PopUpMenu` because it was what I was most familiar with... I wasn't aware of `SimpleDropdown` until later - now I know to start with storybook to look for components first.  That being said, according to this [slack thread](https://codedotorg.slack.com/archives/C06ERUABG01/p1718729235946379?thread_ts=1718664570.407079&cid=C06ERUABG01), it sounds like that won't meet our needs either.  Let me check in with Denys on this a bit more directly though too.

apps/src/templates/sectionProgressV2/OpenAllStudentProgressButton.jsx
kobryan0619: Just to clarify, do you mean add this style to the className `menuItem`?  I tried that initially but it wasn't working.  When I looked at the `PopUpMenu` component I saw that `style` was its own prop (in addition to `className`).  When I put the styles I wanted into a class for `scss` it wouldn't work, however, if I added it in `style` directly as a prop, it worked (this made sense with what I saw in `PopUpMenu` as well).  I tried adding the styles to the `scss` file and then using it as a class in `style`... 

... this is getting long, BUT, through this I learned that `style` needs an object with CSS properties, and when passing through a class, it gets a string.  I got into a way to extract the CSS from the `scss` file, but that made the code way less readable.

lfryemason (approved):
Awesome!

PR #59249, opened 2024-06-17
Fix bug with dynamic registration platform name

nicklathe:
- Add new method to return top level platform name by issuer
  - Add unit test
  - Update registration data to ref new platform name

LTI Integration now has correct platform_name via dynamic registration
Screenshot 2024-06-17 at 1 05 41 PM

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/dashboard/lib/policies/lti.rb
+++ b/dashboard/lib/policies/lti.rb
@@ -162,6 +162,11 @@ class Policies::Lti
     LMS_PLATFORMS.values.find {|platform| platform[:issuer] == issuer}
   end
 
+  def self.find_platform_name_by_issuer(issuer)
+    platform_name, _ = LMS_PLATFORMS.find {|_, platform| platform[:issuer] == issuer}
+    platform_name.to_s
+  end
+
   # Returns the email provided by the LMS when creating the User through LTI
   # provisioning.
   def self.lti_provided_email(user)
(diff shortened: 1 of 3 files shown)


dashboard/lib/policies/lti.rb
```diff
@@ -162,6 +162,11 @@ def self.find_platform_by_issuer(issuer)
     LMS_PLATFORMS.values.find {|platform| platform[:issuer] == issuer}
   end
 
+  def self.find_platform_name_by_issuer(issuer)
+    platform_name, _ = LMS_PLATFORMS.find {|_, platform| platform[:issuer] == issuer}
+    platform_name.to_s
```
stephenliang, line 167: nit: This will throw if the platform can't be found, right? Don't think we guard against this

stephenliang (approved):

dashboard/lib/policies/lti.rb
carl-codeorg: Could do like &.to_s?

nicklathe: I think this is ok as is? Calling `.to_s` on nil returns an empty string.

```
irb(main):001> nil.to_s
=> ""
```

This is also handled in the unit test.

PR #59258, opened 2024-06-17
Use session[:user_return_to] in account linking flow

nicklathe:
After account linking completes, parse session key `session[:user_return_to]` and redirect to that url, or home_path if empty. This is handled by devise automatically for devise handled controller actions and routes, however we need to do this manually since the account linking flow doesn't call devise.

## Links

[Jira](https://codedotorg.atlassian.net/browse/P20-996)

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/dashboard/test/controllers/lti/v1/account_linking_controller_test.rb
+++ b/dashboard/test/controllers/lti/v1/account_linking_controller_test.rb
@@ -17,12 +17,14 @@ class Lti::V1::AccountLinkingControllerTest < ActionController::TestCase
       credential_type: AuthenticationOption::LTI_V1,
       email: @user.email,
     )
+    target_url = "some/test/path"
+    session[:user_return_to] = target_url
     partial_lti_teacher.authentication_options = [ao]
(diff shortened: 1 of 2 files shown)


dashboard/test/controllers/lti/v1/account_linking_controller_test.rb
```diff
@@ -17,12 +17,14 @@ class Lti::V1::AccountLinkingControllerTest < ActionController::TestCase
...
     )
+    target_url = "some/test/path"
+    session[:user_return_to] = target_url
     partial_lti_teacher.authentication_options = [ao]
     PartialRegistration.persist_attributes session, partial_lti_teacher
     User.any_instance.stubs(:valid_password?).returns(true)
 
     post :link_email, params: {email: @user.email, password: 'password'}
-    assert_redirected_to home_path
+    assert_redirected_to target_url
```
nicklathe, line 27: I guess there's room to test the default `home_path` case here, if `session[:user_return_to]` is empty, not sure if it's worth adding an additional test for this though.

carl-codeorg (approved):

PR #59260, opened 2024-06-17
better LTI account linking error handling

carl-codeorg:
In the (hopefully rare) event that an error occurs in `Services::Lti::AccountLinker.call` during an oauth login, this flashes a message, clears their cached partial registration, and redirects to the login page instead of throwing an unhandled 5XX error. If the user is in the middle of an oauth login, and enters the wrong password, MFA code, etc. that's all handled within the oauth provider's login flow. This PR only handles the case of a backend error when we try to do the actual rewiring of auth options.

This PR also adds some handling for bad email/pass logins. Previously, this would just throw a 401. Now, it redirects back to the linking/login page with a invalid credentials message. I had to add some hidden fields to the form to make sure we pass the proper LMS names back to the login page. Without this, we end up missing some strings that we need for the text on the page.

## Links

[JIRA](https://codedotorg.atlassian.net/browse/P20-947)

## PR Checklist:

- [ ] Code is well-commented


carl-codeorg:
Have a unit test that needs to be updated

dashboard/app/views/devise/sessions/_login_lti_account_linking.haml
```diff
@@ -33,6 +33,9 @@
           = f.label :password
           = f.password_field :password, autofocus: email != '', required: true
 
+        = f.hidden_field :lti_provider, value: params[:lti_provider]
```
nicklathe, line 36: Haha I added this same change to the disallow admin from linking accounts feature. Guess I can just rebase off this in now :)

nicklathe (approved):

PR #59269, opened 2024-06-18
AI Tutor: skip PII filter

Erin007:
[CT-642](https://codedotorg.atlassian.net/browse/CT-642)

Our PII filters block usage of AI Tutor when they detect an email, address or phone number. However, there are tasks in the curriculum that specifically reference fake emails in the instructions and starter code which lead to a false positive and the students' chats to be unnecessarily blocked. For example, s/csa6-2023/lessons/6/levels/7/sublevel/4 asks students to extract email addresses from a list of job applicants. 

Screenshot 2024-06-18 at 8 27 50 AM

Additionally,  Geocoder is erroneously flagging `for loops` as addresses.  

This is significantly impacting the user experience with AI Tutor. Therefore, we are going to allow PII through to OpenAI while still logging the violation.  I'm continuing to log the PII violations so that we'll have examples with which to test future solutions that hopefully better meet our needs without impinging on student usage of AI Tutor. 

## Testing story

Student chat with PII is now sent and AI responds appropriately. 

Screenshot 2024-06-25 at 11 54 32 AM

The student's message is still blocked for profanity. 

https://github.com/code-dot-org/code-dot-org/assets/12300669/eeb22ee9-6dc0-4633-9940-0333a98ccb00

## Follow-up work

There are investigations underway for testing a shared safety layer (i.e. profanity, toxic topics) for our AI chat tools. We're evaluating various solutions and hope to re-enable PII filtering if we can avoid false positives.

## Privacy

My understanding is that our privacy agreement with OpenAI is robust enough that this is ok. We only intend to display student chats to the student's teacher. 

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/aiTutor/chatApi.ts
+++ b/apps/src/aiTutor/chatApi.ts
@@ -24,7 +24,7 @@ const CHAT_COMPLETION_URL = '/openai/chat_completion';
 // We want to expose enough information to help troubleshoot false positives
 const logViolationDetails = (response: OpenaiChatCompletionMessage) => {
   console.info('Violation detected in chat completion response', {
-    type: response.status,
+    type: response.safety_status,
     content: response.flagged_content,
   });
   MetricsReporter.logWarning({
@@ -42,8 +42,7 @@ export async function postOpenaiChatCompletion(
   messagesToSend: OpenaiChatCompletionMessage[],
   levelId?: number,
   tutorType?: AITutorTypesValue,
-  levelInstructions?: string,
-  skipPIIFilter?: boolean
+  levelInstructions?: string
 ): Promise<OpenaiChatCompletionMessage | null> {
   const payload = levelId
     ? {
@@ -51,13 +50,11 @@ export async function postOpenaiChatCompletion(
         messages: messagesToSend,
         type: tutorType,
         levelInstructions,
-        skipPIIFilter,
       }
     : {
         messages: messagesToSend,
         type: tutorType,
         levelInstructions,
-        skipPIIFilter,
       };
 
   const response = await HttpClient.post(
@@ -100,17 +97,12 @@ export async function getChatCompletionMessage(
   ];
   let response;
 
-  // Some level instructions include data that looks like PII, optionally skip the PII filter for those levels.
-  // For example s/csa6-2023/lessons/6/levels/7/sublevel/4
-  const skipPIIFilter = levelInstructions?.includes('email');
-
   try {
     response = await postOpenaiChatCompletion(
       messagesToSend,
       levelId,
       tutorType,
-      levelInstructions,
-      skipPIIFilter
+      levelInstructions
     );
   } catch (error) {
     MetricsReporter.logError({
@@ -127,7 +119,7 @@ export async function getChatCompletionMessage(
         'There was an error processing your request. Please try again.',
     };
 
-  switch (response.status) {
+  switch (response.safety_status) {
     case ShareFilterStatus.Profanity:
       logViolationDetails(response);
       return {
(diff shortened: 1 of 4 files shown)


apps/src/aiTutor/chatApi.ts
```diff
@@ -92,12 +94,18 @@ export async function getChatCompletionMessage(
     {role: Role.USER, content: formattedQuestion},
   ];
   let response;
+
+  // Some level instructions include data that looks like PII, optionally skip the PII filter for those levels.
+  // For example s/csa6-2023/lessons/6/levels/7/sublevel/4
+  const skipPIIFilter = levelInstructions?.includes("email");
```
Erin007: This feels hack-y to me. Here's what I don't like about this solution: 
- we'll need to update this rule any time we find another exception (it's essentially hardcoded)
- it's very specific to the curriculum content in CSA
- it won't work with translated level instructions 
- I'm a little nervous about students sending real PII in scenarios where we've skipped the PII filter 

On the other hand, I like that it's not "all or nothing" and we still mostly have PII filtering coverage. 

Very open to other suggestions for how we might handle false positives in a less brittle way!

davidsbailey: Hey Erin, my first thought is, why not blank out the email from the message before sending it to the PII filter?

my second thought is, wait a minute, why are we doing PII filtering on the instructions in the first place? shouldn't we just be running it on the student's entry? In your pilot writeup you mentioned students sometimes pasting the instructions into the chat prompt, but I'm not sure if that's what's going on here.

davidsbailey: to be more specific, I would suggest using a simple regex to identify email addresses rather than looking for the word "email", perhaps copying the exact one we use in our PII filter: https://github.com/code-dot-org/code-dot-org/blob/504257222dd02b9f9e11e09de50c18e7e6b62996/lib/cdo/regexp.rb#L6

I would also think that this could be done on the backend, so that the PII logic can be consolidated to ruby code.

Erin007: Turns out the false positives with the PII filter are more widespread than flagging emails for these specific levels. The more common and negatively impactful issue is that Geocoder is flagging "for loops" as addresses. I've updated the PR to reflect a new, broader approach and reached out to Travis to triple check that this solution is viable.

davidsbailey:
Top level-comments:
* the code itself looks good
* where are we with unit testing for Tutor? I'm not seeing any unit test updates in this PR, and I'm not sure I see a unit test for  the .ts file in this PR
* assuming this plan is approved by Travis, what's the plan for addressing this issue (and in what timeframe)? while we might not need PII filtering for a pilot, if we want it for a broader launch then it'll be great to get more coverage on it during the pilot so we can hone in on what works.
* given the privacy consideration I'd like @ebeastlake to sign off as well

Erin007:
@davidsbailey 
- Good call out on testing. I updated the openai controller test to reflect the changes. I'd say in general we have opportunity for increased test coverage for AI Tutor; but I'm waiting for the dust to settle a bit on what we ultimately think we'll keep v scrap/drastically change. It's on my radar, though. 
- I heard back from Travis and he gave a 👍 The punchline is that we're covered by our agreement with OpenAI from a legal and policy perspective, we are okay sending this data using the OpenAI API service, because they are not allowed to review or use the information for training, etc. We'll need to re-assess if we use a different service. 
- I added a [Jira ticket](https://codedotorg.atlassian.net/browse/CT-652) to reflect the work to re-able PII filtering, but I don't have a definite "due date" for that work, yet

davidsbailey (approved):
LGTM, thanks for the thoughtful responses Erin! It sounds like Travis has provided a clear idea of the privacy implications here, so I'm good with merging without Emily's explicit PR approval.

PR #59271, opened 2024-06-18
Curriculum Launch 2024 - Homepage hero banner

kelbyhawn:
Adds the code.org homepage hero banner for the 2024 Curriculum Launch:
- Works with the `curriculum-launch-2024` DCDO flag
- Fully responsive and works w/ LTR and RTL languages

**Note:** I added the other banner strings here for the skinny banners ([ACQ-1969](https://codedotorg.atlassian.net/browse/ACQ-1969)) and teacher homepage banner ([ACQ-1968](https://codedotorg.atlassian.net/browse/ACQ-1968)) so they can get in the pipeline sooner.

## Links
Jira ticket: [ACQ-1967](https://codedotorg.atlassian.net/browse/ACQ-1967)
Figma mock: [here](https://www.figma.com/design/HwXIyBGQ0b2ZI213sWb3cK/2024-Curriculum-Launch?node-id=2436-4603&t=dFI9Qgia70mMmEsS-1)

## Testing story
Tested locally

## Followup
Add the blog post URL if we don't get it before this gets merged.

----

https://github.com/code-dot-org/code-dot-org/assets/9256643/8f7cd75c-c630-44ba-a6e4-f975c23bdb3a

PR #59278, opened 2024-06-18
Add banner for new curriculum on My PL page 

bethanyaconnor:
Also includes a small refactor to ensure no more than one banner is displayed for any user. The logic for that goes:
- if a teacher has never interacted with our PL before, show the Getting Started banner
- if the teacher has a workshop survey to fill out, show the banner to do so (can never happen at the same time as above)
- if neither of the two cases above, and the DCDO flag for curriculum launch is set, show the curriculum launch banner

Screenshot of banner:

Screenshot 2024-06-20 at 8 40 23 AM

I confirmed that this banner does not show up without the DCDO flag.


apps/src/code-studio/pd/professional_learning_landing/LandingPage.jsx
```diff
@@ -114,23 +110,6 @@ function LandingPage({
     dispatch(asyncLoadCoteacherInvite());
   }, [dispatch]);
 
-  const RenderGettingStartedBanner = () => (
```
bethanyaconnor, line 117: This was folded into `RenderBanner` below.

apps/src/code-studio/pd/professional_learning_landing/LandingPage.jsx
```diff
@@ -148,6 +127,54 @@ function LandingPage({
...
+          imageUrl={pegasus(
+            '/images/fill-540x300/professional-learning/banner-books-with-background.png'
+          )}
+          subHeading={i18n.plLandingCurriculumLaunchBannerSubHeading()}
+          description={i18n.plLandingCurriculumLaunchBannerDescription()}
+          buttons={[
+            {
+              url: pegasus('/educate/professional-learning'),
+              text: i18n.plLandingCurriculumLaunchBannerButtonText(),
+            },
```
bethanyaconnor, line 173: question for reviewers: should this button have an aria label. It does only say "Start here" so not sure if we need to add more context

kelbyhawn: I think technically it should, but since it's the only call to action in this section it's probs ok to not add one since the heading and description text clarifies where it goes.

But this button component should have an aria label prop, so it would be easy to add if you decide to!

apps/src/code-studio/pd/professional_learning_landing/LandingPage.jsx
```diff
@@ -148,6 +127,54 @@ function LandingPage({
...
+              url: pegasus('/educate/professional-learning'),
+              text: i18n.plLandingGettingStartedButton(),
+            },
+          ]}
+        />
+      );
+    } else if (lastWorkshopSurveyUrl) {
+      return RenderLastWorkshopSurveyBanner();
+    } else if (!!DCDO.get('curriculum-launch-2024', false)) {
+      return (
```
kelbyhawn, line 161: Not blocking suggestion: can you add a TODO clean up comment here so we know to remove this eventually after the launch?

apps/src/code-studio/pd/professional_learning_landing/LandingPage.jsx
```diff
@@ -312,8 +339,7 @@ function LandingPage({
   const RenderMyPlTab = () => {
     return (
       <>
-        {showGettingStartedBanner && RenderGettingStartedBanner()}
-        {lastWorkshopSurveyUrl && RenderLastWorkshopSurveyBanner()}
+        {RenderBanner()}
```
kelbyhawn, line 344: Love how clean this is now! 💖

lib/dynamic_config/dcdo.rb
```diff
@@ -55,6 +55,7 @@ def frontend_config
       'student-email-post-enabled': DCDO.get('student-email-post-enabled', false),
       'progress-v2-metadata-enabled': DCDO.get('progress-v2-metadata-enabled', false),
       'show-updated-lms-content': DCDO.get('show-updated-lms-content', false),
+      'curriculum-launch-2024': DCDO.get('curriculum-launch-2024', false),
```
kelbyhawn: If you merge in staging you can remove this since I added it in this PR: https://github.com/code-dot-org/code-dot-org/pull/59271

kelbyhawn (approved):
Left a couple comments, but looks great!

apps/src/code-studio/pd/professional_learning_landing/LandingPage.jsx
bethanyaconnor: Done!

apps/src/code-studio/pd/professional_learning_landing/LandingPage.jsx
bethanyaconnor: Added!

kelbyhawn (approved):
Thanks for adding the comment and aria label!

PR #59283, opened 2024-06-18
Lab2 project validator links

fisher-alice:
This PR adds the majority of the project validator info, links/button to the React version of the Extra Links box for lab2 labs. It is a follow-up to https://github.com/code-dot-org/code-dot-org/pull/58410 which added levelbuilder info/links.

As explained in the PR reference above, the reason that the current gray extra links box is not working correctly for lab2 levels is because it is populated via haml. Therefore, if the level changes, the box is not updated because the page is NOT reloaded.

This PR adds a new `projects` api `/extra_links` which returns the project and owner info including links to the project sources, and to projects included in the remix ancestry.
Note that for standalone projects, we do not display 'clone' or 'delete' buttons since curriculum writers rarely need to edit standalone project levels and accidental edits can be quite disruptive.
Also, a subset of project info including featured project status and remix history is only displayed for standalone projects.
The `projects_controller` `extra_links` method is only available for users with `project_validator` permission.

I added a 'Bookmark' button for standalone projects that are not yet included in the `featured_projects` table. Once a project is featured regardless of status, the 'Bookmark' button is not displayed.

## After update

Screenshot of only levelbuilder permission on aichat level:
Screenshot 2024-06-18 at 5 38 02 PM

Screenshot of only levelbuilder permission on standalone music project: Note that clone/delete. buttons are not displayed since this is a standalone project level.
Screenshot 2024-06-18 at 5 30 52 PM

Screenshot of only project validator permission on aichat level:
Screenshot 2024-06-24 at 1 33 51 PM

Screenshot of only project validator permission on standalone music project (no remix ancestry):
Screenshot 2024-06-18 at 5 34 03 PM

Screenshot of only project validator permission on standalone music project (with remix ancestry):
Screenshot 2024-06-18 at 5 34 25 PM

Screencast video of standalone project being bookmarked:

https://github.com/code-dot-org/code-dot-org/assets/107423305/03fb2ac1-b270-4edc-8505-a9b73e39873a

## Links
[jira](https://codedotorg.atlassian.net/browse/LABS-870)

## Testing story
I tested locally using an account with only project validator permission and an account with only levelbuilder permission - I tested on different lab2 levels (music, aichat, and python) including project template levels as well as music and python standalone projects to confirm that the expected info/links/buttons were displayed. I confirmed that the ancestry project links worked as well.

I added a couple dashboard unit tests to confirm access to  the new `projects_controller` method is restricted to those with `project_validator` access.

## Follow-up work
Follow-up work includes adding Blockly-related buttons, project abuse info/buttons, and exemplar unfreeze/freeze buttons (I think an api endpoint needs to be added for this) as tracked in https://codedotorg.atlassian.net/browse/CT-550.
https://codedotorg.atlassian.net/browse/LABS-891

## PR Checklist:

- [ ] Code is well-commented


apps/src/lab2/views/ExtraLinks.tsx
```diff
@@ -14,61 +19,74 @@ interface PermissionResponse {
   permissions: string[];
 }
 
-interface ExtraLinksResponse {
```
fisher-alice, line 17: Removed this since I think it dups `ExtraLinksLevelbuilderData` (renamed from `ExtraLinksData`.

apps/src/lab2/views/ExtraLinks.tsx
```diff
@@ -14,61 +19,74 @@ interface PermissionResponse {
...
 
   useEffect(() => {
     const permissionData = data
       ? (data as PermissionResponse)
       : {permissions: []};
-    if (permissionData.permissions.includes(PERMISSIONS.LEVELBUILDER)) {
+    if (
+      permissionData.permissions.includes(PERMISSIONS.LEVELBUILDER) ||
+      permissionData.permissions.includes(PERMISSIONS.PROJECT_VALIDATOR)
+    ) {
```
fisher-alice: Note that in `levels_controller` we check if user has permission to edit level (i.e., levelbuilder permission). But project validators are able to view level info even thought they are unable to edit.
https://github.com/code-dot-org/code-dot-org/blob/affebbbc320e0c1c4e5fb1f513220f96d6fae1f1/dashboard/app/controllers/levels_controller.rb#L508-L516

apps/src/lab2/views/ExtraLinks.tsx
```diff
@@ -14,61 +19,74 @@ interface PermissionResponse {
...
-    : {permissions: []};
+    if (permissionData.permissions.includes(PERMISSIONS.PROJECT_VALIDATOR)) {
+      try {
+        HttpClient.fetchJson<ExtraLinksProjectValidatorData>(
+          `/projects/${channelId}/extra_links`
+        ).then(response => {
+          setProjectValidatorLinkData(response.value);
+        });
+      } catch (e) {
+        console.error('Error fetching levelbuilder extra links', e);
```
molly-moen: nit: project validator extra links?

apps/src/lab2/views/ExtraLinksModal.tsx
```diff
@@ -27,10 +34,26 @@ const ExtraLinksModal: React.FunctionComponent<ExtraLinksModalProps> = ({
...
   const [deleteError, setDeleteError] = useState('');
+  const [featuredProjectStatus, setFeaturedProjectStatus] = useState<
+    string | undefined
+  >('');
+
+  const channelId: string | undefined = useAppSelector(
+    state => state.lab.channel && state.lab.channel.id
+  );
+  const levelName = levelbuilderLinkData.level_name;
+  const isStandaloneProject =
```
molly-moen: I wonder if we can get `isStandaloneProject` from the backend instead of deriving it here (maybe it's part of the extra links response?). I think it would be cleaner to only have 1 source of truth for the standalone project names rather than duplicating them.

apps/src/lab2/types.ts
```diff
@@ -256,7 +256,7 @@ export enum ProjectManagerStorageType {
   REMOTE = 'REMOTE',
 }
 
-export interface ExtraLinksData {
+export interface ExtraLinksLevelbuilderData {
```
molly-moen: I got confused thinking this data was only available to levelbuilders but it's actually available to both user types. Would it make more sense to name these interfaces `ExtraLinksLevelData` and `ExtraLinksProjectData` since they are the data about the level and the data about the project, respectively?

apps/src/lab2/views/ExtraLinks.tsx
```diff
@@ -14,61 +19,74 @@ interface PermissionResponse {
...
+  return loading ||
+    (!levelbuilderLinkData && !projectValidatorLinkData) ? null : (
     <>
       <Button
         onClick={() => setIsModalOpen(true)}
         text={'Extra Links'}
         className={moduleStyles.extraLinksButton}
       />
-      {linkData && (
+      {levelbuilderLinkData && (
```
molly-moen: will the user possibly see an incomplete modal if they click this and the level data has loaded but the project data has not? This might not be enough of a concern to try to work around.

apps/src/lab2/views/ExtraLinksModal.tsx
```diff
@@ -83,6 +106,145 @@ const ExtraLinksModal: React.FunctionComponent<ExtraLinksModalProps> = ({
...
 
+  const onBookmark = async () => {
+    try {
+      await HttpClient.put(
+        `/featured_projects/${channelId}/bookmark`,
+        undefined,
+        true,
+        {contentType: 'application/json;charset=UTF-8'}
+      );
+      setFeaturedProjectStatus('bookmarked');
```
molly-moen: nit: should we make a constant for `'bookmarked'`?

dashboard/app/controllers/projects_controller.rb
```diff
@@ -571,6 +571,39 @@ def limited_gallery?
     !project_validator && limited_project_gallery
   end
 
+  # GET /projects/:channel_id/extra_links
+  # Get the extra links for the project for use by project validators.
+  # This is used by lab2 levels that cannot use the haml 'extra links' box since
+  # this box will not refresh when changing levels.
+  def extra_links
```
molly-moen, line 578: how are permissions handled for this method?

apps/src/lab2/views/ExtraLinks.tsx
molly-moen: Can you update the comment on that method in levels_controller to indicate project validators can call it too, but they see a subset of that information?

molly-moen (commented):
Looks good, a couple comments/questions!

apps/src/lab2/types.ts
fisher-alice: Sure! I like that idea.

apps/src/lab2/views/ExtraLinksModal.tsx
fisher-alice: Sure - that makes sense. Will update.

apps/src/lab2/views/ExtraLinks.tsx
fisher-alice: On line 79 above, this will not render unless both level data and link data are defined. I can reset the link data to `null` at the beginning of the `useEffect`. Lmk what you think - thanks!

apps/src/lab2/views/ExtraLinksModal.tsx
fisher-alice: Oh yeah! Can use `FeaturedProjectStatus` from shared constants.

dashboard/app/controllers/projects_controller.rb
fisher-alice: This has been a question that's been floating around in my head so thanks for asking it!
I see that currently, a signed out user doesn't have access to this method but signed-in users do even though the information is not actually displayed via Extra Links.
I'll update and add a condition at the beginning of the definition to check for `project_validator` permission. Thanks!

apps/src/lab2/views/ExtraLinksModal.tsx
```diff
@@ -83,6 +102,144 @@ const ExtraLinksModal: React.FunctionComponent<ExtraLinksModalProps> = ({
...
+
+  const renderRemixAncestry = (remixList: string[]) => {
+    if (remixList.length === 0) {
+      return <li>Not a remix.</li>;
+    }
+    return remixList.map((link: string, index: number) => (
+      <li key={index}>
+        <a href={link}>{link}</a>
+      </li>
+    ));
```
thomasoniii: This is probably okay, but it might be better to key off of `link` itself. That way we don't even need the `index` arg, and we're safe against render errors if the list is re-ordered.

apps/src/lab2/views/ExtraLinksModal.tsx
```diff
@@ -83,6 +102,144 @@ const ExtraLinksModal: React.FunctionComponent<ExtraLinksModalProps> = ({
...
+            size={Button.ButtonSize.small}
+            color={Button.ButtonColor.purple}
+            onClick={onBookmark}
+            text={'Bookmark as featured'}
+          />
+        </>
+      );
+    }
+    return <div>Featured project status: {featuredProjectStatus}</div>;
+  };
```
thomasoniii: Totally nonblocking, just my preferred syntactic sugar...

For all of these `render*` functions, they're fine as is, but stylistically I kinda prefer making them actual components and referencing them that way. The change is simple:

```
renderFeaturedProjectInfo -> FeaturedProjectInfo

and then
 { renderFeaturedProjectInfo()} => <FeaturedProjectInfo />
```

(or whatever component name you want.

I'd consider the function calls a slightly older style, and it feels more react-ish as components.

Then, just other notes for the other components -
- The ones that return static markup can also be wrapped into components in this manner - just have them be a function that returns something (probably preferably a memoized function to ensure no re-renders)
- The ones taking args should be tweaked so the args come in as props
- I see that `renderProjectValidatorData` is wrapped up in a conditional where it checks `projectLinkData` and `isStandaloneProject`. I might add those in as additional props and have the component return `null` if either of them are false before falling into the rendering.

But, again, totally just personal preference so only bother to update if it looks better to you too. 😄

thomasoniii (approved):
one question about a `key`, and some non-blocking stylistic suggestions.

Otherwise lgtm 🚀

apps/src/lab2/views/ExtraLinksModal.tsx
fisher-alice: Thanks Jim - this makes sense and very appreciated! ~~Follow-up work is to add Report abuse info/links as well as blockly-related buttons so I will log this suggested refactor as part of the follow-up.
https://codedotorg.atlassian.net/browse/LABS-891~~

dashboard/app/controllers/projects_controller.rb
```diff
@@ -571,6 +571,43 @@ def limited_gallery?
...
+    return render(status: :forbidden, json: {error: 'This user does not have project validator access.'}) unless current_user.project_validator?
```
molly-moen: I think we can move this check into `ability.rb` instead

dashboard/app/controllers/projects_controller.rb
```diff
@@ -571,6 +571,43 @@ def limited_gallery?
...
+    src_channel_id = params[:channel_id]
+    if src_channel_id == "undefined"
+      return render json: {error: 'No channel id provided.'}, status: :ok
```
molly-moen: Should the status be `ok` in this situation?

apps/src/lab2/views/ExtraLinks.tsx
molly-moen: this seems fine then!

molly-moen (commented):
couple remaining questions, thanks for the updates!

apps/src/lab2/views/ExtraLinks.tsx
```diff
@@ -14,61 +16,75 @@ interface PermissionResponse {
...
+    useState<ExtraLinksProjectData | null>(null);
+  const permissionData = data
+    ? (data as PermissionResponse)
+    : {permissions: []};
+  const channelId = useAppSelector(
+    state => state.lab.channel && state.lab.channel.id
+  );
 
   useEffect(() => {
     const permissionData = data
```
sanchitmalhotra126: Should we also early return here if `loading` is true (still fetching permissions data from server)?

apps/src/lab2/views/ExtraLinks.tsx
```diff
@@ -14,61 +16,75 @@ interface PermissionResponse {
...
       ? (data as PermissionResponse)
       : {permissions: []};
-    if (permissionData.permissions.includes(PERMISSIONS.LEVELBUILDER)) {
+    setLevelLinkData(null);
+    setProjectLinkData(null);
+    if (
+      permissionData.permissions.includes(PERMISSIONS.LEVELBUILDER) ||
```
sanchitmalhotra126: nit: can pull these out into locals (`isLevelbuilder` or `hasLevelbuilderPermissions`) since this is checked in multiple if statements?

apps/src/lab2/views/ExtraLinks.tsx
```diff
@@ -14,61 +16,75 @@ interface PermissionResponse {
...
-  if (!permissionData.permissions.includes(PERMISSIONS.LEVELBUILDER)) {
+  if (
+    !permissionData.permissions.includes(PERMISSIONS.LEVELBUILDER) &&
+    !permissionData.permissions.includes(PERMISSIONS.PROJECT_VALIDATOR)
+  ) {
     return <></>;
   }
 
-  return loading || !linkData ? null : (
+  return loading || (!levelLinkData && !projectLinkData) ? null : (
```
sanchitmalhotra126: nit: early return here instead of ternary?

apps/src/lab2/views/ExtraLinks.tsx
```diff
@@ -14,61 +16,75 @@ interface PermissionResponse {
...
+  const [projectLinkData, setProjectLinkData] =
+    useState<ExtraLinksProjectData | null>(null);
+  const permissionData = data
+    ? (data as PermissionResponse)
+    : {permissions: []};
+  const channelId = useAppSelector(
+    state => state.lab.channel && state.lab.channel.id
+  );
 
   useEffect(() => {
```
sanchitmalhotra126, line 78: Looking at this again, might it be cleaner to pull out all this fetching logic into a separate function outside of the component, since all three of these server calls have to happen back to back? For example:

```
async function fetchExtraLinksData(levelId: number): Promise<ExtraLinksWithPermissions> {
  // Fetch permissions
  const permissionsResponse = await HttpClient.fetchJson<PermissionResponse>(
    '/api/v1/users/current/permissions'
  );
  const {permissions} = permissionsResponse.value;

  // Fetch level links data
  let levelLinkData: ExtraLinksLevelData;
  if (permissions.includes(PERMISSIONS.LEVELBUILDER)) {
    const levelLinkDataResponse =
      await HttpClient.fetchJson<ExtraLinksLevelData>(
        `/levels/${levelId}/extra_links`
      );
    levelLinkData = levelLinkDataResponse.value;
  }

  // Fetch project links data
  let levelProjectData: ExtraLinksProjectData;
  if (permissions.includes(PERMISSIONS.PROJECT_VALIDATOR)) {
    const levelProjectDataResponse =
      await HttpClient.fetchJson<ExtraLinksProjectData>(
        `/projects/${levelId}/extra_links`
      );
    levelProjectData = levelProjectDataResponse.value;
  }

  // Return fetched data and permissions
  return {
    levelLinkData,
    levelProjectData,
    permissions,
  };
}
```

And then, in the component we just need something like

```
  const [isLoading, setIsLoading] = useState(false);
  const [extraLinksWithPermissions, setExtraLinksWithPermissions] = useState(null);

  useEffect(() => {
    setIsLoading(true);
    fetchExtraLinksData(levelId).then(data => {
      setExtraLinksWithPermissions(data);
      setIsLoading(false);
    });
  }, [levelId]);

  // Do stuff with extraLinksWithPermissions once it exists
```

(couldn't think of a good name for the "extraLinksWithPermissions" bundle, but I'm sure there's something better)

apps/src/lab2/views/ExtraLinks.tsx
fisher-alice: Oh interesting. Yes that makes sense.

dashboard/app/controllers/projects_controller.rb
fisher-alice: I added this because I saw an error response for `/extra_links` when a project validator was on a non-standalone project level and I don't necessarily think of this as an error. I think I meant for this to be an `ok` response with a different response message...maybe something like:
`render json: {message: 'Levels that are not standalone projects do not include project info.'}, status: :ok`

What do you think?

molly-moen: By "non-standalone project level" do you mean any level that's not the "new <lab name> project" level or a level that doesn't have a project at all? We should still be able to get project validator information for a project-backed level, which isn't necessarily the standalone project level.

apps/src/lab2/views/ExtraLinks.tsx
fisher-alice: Oh! I was going to do this, but since we fetch permissions in the `useEffect`, I didn't end up assigning a local variable. Thoughts?

apps/src/lab2/views/ExtraLinksModal.tsx
fisher-alice: Hey Jim - doing some refactoring based on other comments so will also incorporate your feedback as well. Thanks!

apps/src/lab2/views/ExtraLinks.tsx
fisher-alice: Nice! Will update.

dashboard/app/controllers/projects_controller.rb
fisher-alice: I've noticed that for channel-backed legacy labs like Sprite lab, I don't think we display project/owner info on a non-standalone project level - this is 's/allthethings/lessons/36/levels/2' and level name is 'AllTheThings sprite lab with video'.
Screenshot 2024-06-21 at 5 33 09 PM

But once we get the share URL via `dashboard.project.getShareUrl()`, we can then get the project info.
Screenshot 2024-06-21 at 5 33 30 PM
The level name for the level via share URL is 'New Sprite Lab Project'.

I remember this debugging ZenDesk issues. Makes sense according to: https://github.com/code-dot-org/code-dot-org/blob/6a67ac98118222f97c0208ae60829d207e2cf06d/dashboard/app/views/levels/_admin.html.haml#L115

But for project validators, we will display report abuse and project-sharing info (will be included in follow-up).

apps/src/lab2/views/ExtraLinks.tsx
fisher-alice: I think with the refactor you suggested below, this may now be okay as is?

dashboard/app/models/user.rb
```diff
@@ -1541,6 +1541,11 @@ def can_view_student_ai_chat_messages?
       SingleUserExperiment.enabled?(user: self, experiment_name: AI_TUTOR_EXPERIMENT_NAME)
   end
 
+  # This method just checks if a user has project validator permission
+  def project_validator?
```
molly-moen: is this used anywhere?

dashboard/app/controllers/projects_controller.rb
molly-moen: Ok makes sense. As a follow-up we might be able to improve this to show the project info even for non-standalone project levels, since we know the channel id.

molly-moen (approved):
Looks good, one potential follow-up and a question on a method that might be unused. Good job!

dashboard/app/controllers/projects_controller.rb
fisher-alice: But this is definitely not ideal considering that `aichat` doesn't have a standalone project level yet. ~~Can look into how `channel_id` is assigned in redux in a follow-up and find a workaround for this. Currently for non-standalone project levels, `channel_id` is assigned`'undefined'`.~~
Will change this!

apps/src/lab2/views/ExtraLinks.tsx
```diff
@@ -14,61 +15,91 @@ interface PermissionResponse {
...
 }
 
 // If the user has permission to see extra links, fetch extra links for the level,
 // then display a modal with the link data.
 const ExtraLinks: React.FunctionComponent<ExtraLinksProps> = ({
   levelId,
 }: ExtraLinksProps) => {
-  const {loading, data} = useFetch('/api/v1/users/current/permissions');
+  const [isLoading, setIsLoading] = useState(false);
+  const [PermissionsAndLinkData, setPermissionsAndLinkData] =
```
sanchitmalhotra126: nit: lowercase naming `permissionsAndLinkData` so it doesn't get confused with the type name?

dashboard/app/controllers/projects_controller.rb
fisher-alice: Thanks! Think our lines got crossed. Will definitely log a ticket to look into this.

apps/src/lab2/views/ExtraLinks.tsx
```diff
@@ -14,61 +15,91 @@ interface PermissionResponse {
...
+    setIsLoading(true);
+    fetchPermissionsAndLinkData(levelId, channelId).then(data => {
+      setPermissionsAndLinkData(data);
+      setIsLoading(false);
+    });
+  }, [levelId, channelId]);
+  const {levelLinkData, projectLinkData, permissions} =
+    PermissionsAndLinkData || {};
+  if (
+    !permissions?.includes(PERMISSIONS.LEVELBUILDER) &&
```
sanchitmalhotra126: Actually, do we need to check the permissions at all? Could this just be
```
if (loading || permissionsAndLinkData === undefined) {
  return null;
}

const {levelLinkData, projectLinkData} = permissionsAndLinkData;
if (levelLinkData === undefined && projectLinkData === undefined) {
  return null;
}

// if we've gotten here, we know we have at least one type of data and some permissions to view it
```
In which case we don't even need to return the permissions and could just call it `extraLinksData` or something

apps/src/lab2/views/ExtraLinksModal.tsx
```diff
@@ -1,23 +1,27 @@
...
+import {ExtraLinksLevelData, ExtraLinksProjectData} from '../types';
+import moduleStyles from './extra-links.module.scss';
 
 // Extra Links modal. This is used to display helpful links for levelbuilders, and should
 // be extended to also include links for project validators as well. It replaces the haml
 // version of extra links, which doesn't work on lab2 after a level change.
 interface ExtraLinksModalProps {
-  linkData: ExtraLinksData;
+  levelLinkData: ExtraLinksLevelData;
+  projectLinkData: ExtraLinksProjectData | undefined;
```
sanchitmalhotra126: nit: why `| undefined` instead of optional?

sanchitmalhotra126 (commented):
Thanks for incorporating the refactor! a few minor follow-up comments

apps/src/lab2/views/ExtraLinks.tsx
fisher-alice: Oh nice! And can simplify even more with one condition:
```
return isLoading || (!levelLinkData && !projectLinkData) ? null : (
  <>
...
```

Thanks for all the refactor suggestions!

apps/src/lab2/views/ExtraLinks.tsx
```diff
@@ -14,61 +15,86 @@ interface PermissionResponse {
...
+  let projectLinkData: ExtraLinksProjectData | undefined;
+  if (permissions.includes(PERMISSIONS.PROJECT_VALIDATOR)) {
+    const levelProjectDataResponse =
+      await HttpClient.fetchJson<ExtraLinksProjectData>(
+        `/projects/${channelId}/extra_links`
+      );
+    projectLinkData = levelProjectDataResponse.value;
+  }
+
+  // Return fetched permissions and link data.
```
sanchitmalhotra126: nit: update comment now that permissions have been removed?

apps/src/lab2/views/ExtraLinks.tsx
```diff
@@ -14,61 +15,86 @@ interface PermissionResponse {
...
+    setIsLoading(true);
+    fetchExtraLinksData(levelId, channelId).then(data => {
+      setExtraLinksData(data);
+      setIsLoading(false);
+    });
+  }, [levelId, channelId]);
+  const {levelLinkData, projectLinkData} = extraLinksData || {};
 
-  return loading || !linkData ? null : (
+  return isLoading || (!levelLinkData && !projectLinkData) ? null : (
```
sanchitmalhotra126: Nice, much cleaner! One last thing, totally up to you - I personally prefer to early return rather than putting a whole react component in a ternary since it reads a little easier to me. Either way works, especially since it's not a lot of TSX. But for example I might do:
```
if (isLoading || (!levelLinkData && !projectLinkData)) {
  return null;
}

return <ExtraLinksModal ... />

File: dashboard/app/controllers/projects_controller.rb [hunk 23]
```diff
@@ -571,6 +571,42 @@ def limited_gallery?
...
+  # this box will not refresh when changing levels.
+  def extra_links
+    src_channel_id = params[:channel_id]
+    if src_channel_id == "undefined"
+      return render json: {message: 'No channel id provided.'}, status: :ok
+    end
+    project_info = {}
+    owner_info = {}
+    owner_info['storage_id'], project_info['id'] = storage_decrypt_channel_id(src_channel_id)
+    project_info['sources_link'] = "https://s3.console.aws.amazon.com/s3/buckets/#{CDO.sources_s3_bucket}/#{CDO.sources_s3_directory}/#{owner_info['storage_id']}/#{project_info['id']}/"
```
sanchitmalhotra126, line 586: Out of curiosity - is this how we construct S3 links for the legacy extra links box as well?

apps/src/lab2/views/ExtraLinksModal.tsx
```diff
@@ -27,10 +31,25 @@ const ExtraLinksModal: React.FunctionComponent<ExtraLinksModalProps> = ({
...
   const [deleteError, setDeleteError] = useState('');
+  const [featuredProjectStatus, setFeaturedProjectStatus] = useState<
+    string | undefined
+  >('');
+
+  const channelId: string | undefined = useAppSelector(
+    state => state.lab.channel && state.lab.channel.id
+  );
+
+  const isStandaloneProject = levelLinkData.is_standalone_project;
```
sanchitmalhotra126: minor: I think we already have access to this via lab2Redux in `state.lab.levelProperties.isProjectLevel`?

(though @molly-moen I think you included this on the backend in `levels_controller#extra_links` - would `is_project_level`/`isProjectLevel` also work here, or is there a reason we need to do a separate lookup?)

molly-moen: Oh good call, I think we could use `isProjectLevel` instead.

apps/src/lab2/views/ExtraLinks.tsx
fisher-alice: I don't really have a personal preference so can update.

dashboard/app/controllers/projects_controller.rb
fisher-alice: yes - https://github.com/code-dot-org/code-dot-org/blob/a0f1ed29a6b217c930590a020159b4b77bfcace5/dashboard/app/views/levels/_admin.html.haml#L123

dashboard/app/controllers/projects_controller.rb
fisher-alice: Hey Molly - I totally forgot I intentionally did not display project/owner info since this is not shown in legacy labs. I won't show all irrelevant project info (like featured project info or remix ancestry), but will show project id, S3 sources link, owner name and storage_id. Thanks!

apps/src/lab2/views/ExtraLinksModal.tsx
fisher-alice: @sanchitmalhotra126 - just tagging since you wrote a comment related to this below - related to https://github.com/code-dot-org/code-dot-org/pull/59283#discussion_r1651330227.

dashboard/app/controllers/projects_controller.rb
fisher-alice: Here's an updated screenshot of channel-backed level:
Screenshot 2024-06-24 at 1 33 51 PM

sanchitmalhotra126 (approved):
LGTM! Thanks for the iteration on this!

PR #59284, opened 2024-06-18
Add new fine-tuned models to AI Chat

sanchitmalhotra126:
Adds 3 new fine-tuned models to AI Chat. There are 4 listed in the JIRA but I was having trouble deploying one of them (Customer Support Mistral) and wanted to get these three available for testing while I continue to troubleshoot the fourth. Also added a small util function to validate the selected model ID whenever we load or reset a level, to make sure we're always using a valid model ID in case the levelbuilder or student-saved model ID is stale.

Screenshot 2024-06-18 at 2 21 45 PM

Screenshot 2024-06-18 at 2 19 13 PM

Screenshot 2024-06-18 at 2 18 24 PM

Screenshot 2024-06-18 at 2 20 47 PM

## Links

https://codedotorg.atlassian.net/browse/LABS-662

## Testing story

Tested each model locally, and verified that invalid model IDs get reset to the first available model.


apps/src/aichat/redux/aichatRedux.ts
```diff
@@ -46,6 +46,7 @@ import {
   getCurrentTimestamp,
   hasFilledOutModelCard,
 } from './utils';
+import {validateModelId} from '../views/modelCustomization/utils';
```
fisher-alice, line 49: Nice! Just confirming we need this for AI customizations that have been saved in the past and we don't need this check in `saveAiCustomization` because we're assuming that users can only selected from updated `modelDescriptions.json`. Is that correct?

fisher-alice (approved):
Nice!

apps/src/aichat/redux/aichatRedux.ts
sanchitmalhotra126: Yep, I think so - this is specifically for when we change the model descriptions JSON and levels or student code have old invalid IDs saved.

PR #59292, opened 2024-06-19
[Design2-118] DSCO - Tabs - add tooltip as a prop to every tab, update icon sizing to be the same as text size

levadadenys:
### [Design2-118] DSCO - Tabs - add tooltip as a prop to every tab, update icon sizing to be the same as text size

* added `tooltip` optional prop to `_Tab` component
* updated `tab` `icons` sizes

image

Before:
image

After:
Знімок екрана 2024-06-20 о 21 31 57

## Links
- jira ticket: [](https://codedotorg.atlassian.net/browse/DESIGN2-118)

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/componentLibrary/tabs/_Tab.tsx
+++ b/apps/src/componentLibrary/tabs/_Tab.tsx
@@ -54,7 +54,7 @@ const checkTabForErrors = (
   }
 };
 
-const renderButtonContent = (
+const renderTabButtonContent = (
   isIconOnly: boolean,
   icon?: FontAwesomeV6IconProps,
   text?: string,
@@ -90,7 +90,7 @@ const _Tab: React.FunctionComponent<TabsProps> = ({
   const handleClick = useCallback(() => onClick(value), [onClick, value]);
   checkTabForErrors(isIconOnly, icon, text);
 
-  const buttonContent = renderButtonContent(
+  const buttonContent = renderTabButtonContent(
     isIconOnly,
     icon,
     text,
--- a/apps/src/componentLibrary/tabs/_Tab.tsx
+++ b/apps/src/componentLibrary/tabs/_Tab.tsx
@@ -4,6 +4,7 @@ import React, {useCallback} from 'react';
 import FontAwesomeV6Icon, {
   FontAwesomeV6IconProps,
 } from '@cdo/apps/componentLibrary/fontAwesomeV6Icon';
+import {TooltipProps, WithTooltip} from '@cdo/apps/componentLibrary/tooltip';
 
 import moduleStyles from './tabs.module.scss';
 
@@ -18,6 +19,8 @@ export interface TabModel {
   iconRight?: FontAwesomeV6IconProps;
   /** Whether button should be icon only */
   isIconOnly?: boolean;
+  /** Tab tooltip props */
+  tooltip?: TooltipProps;
   /** Tab icon */
   icon?: FontAwesomeV6IconProps;
   /** Tab content */
@@ -51,6 +54,25 @@ const checkTabForErrors = (
   }
 };
 
+const renderButtonContent = (
+  isIconOnly: boolean,
+  icon?: FontAwesomeV6IconProps,
+  text?: string,
+  iconLeft?: FontAwesomeV6IconProps,
+  iconRight?: FontAwesomeV6IconProps
+) => {
+  if (isIconOnly && icon) {
+    return <FontAwesomeV6Icon {...icon} />;
+  }
+  return (
+    <>
+      {iconLeft && <FontAwesomeV6Icon {...iconLeft} />}
+      {text && <span>{text}</span>}
+      {iconRight && <FontAwesomeV6Icon {...iconRight} />}
+    </>
+  );
+};
+
 const _Tab: React.FunctionComponent<TabsProps> = ({
   isSelected,
   onClick,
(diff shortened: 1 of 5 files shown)


apps/src/componentLibrary/tabs/_Tab.tsx
```diff
@@ -67,30 +90,39 @@ const _Tab: React.FunctionComponent<TabsProps> = ({
...
-        {!isIconOnly && (
-          <>
-            {iconLeft && <FontAwesomeV6Icon {...iconLeft} />}
-            {text && <span>{text}</span>}
-            {iconRight && <FontAwesomeV6Icon {...iconRight} />}
-          </>
-        )}
-      </button>
+      {tooltip ? (
+        <WithTooltip tooltipProps={tooltip}>{buttonElement}</WithTooltip>
```
fisher-alice, line 122: Would it be possible to update `Tooltip` so that if the Tab has an icon, the tooltip caret could be set to the right (if iconRight) or left (if iconLeft)?

Here's what the tab current looks like with the tooltip:
tooltip-center

And am wondering if we could move the caret if there's an icon:
Screenshot 2024-06-20 at 2 47 44 PM

@markabarrett - curious what you think about this. Thanks!

fisher-alice: In this case, the icon and the label do not have the same meaning but in other cases I imagine they will. So could we add a way to set the position of tooltip caret regardless of whether there is can icon or not.

levadadenys: Yeah, right now it's not possible. And I don't think it's a good idea to add tooltip directly to an icon.
However, there's a native support of 'native tooltips' for icons which might be the best and the most accessible option: Just pass `title` prop to the needed `icon`, and it'll work in the following way. (I actually think this `title` prop should be passed in most of icons usage)

fisher-alice: After discussion with you and Mark, we've decided to take a different approach with Tabs and the lock icon which indicates 'view only'. Thanks so much for the information about the native support of native tooltips!

apps/src/componentLibrary/tabs/_Tab.tsx
```diff
@@ -51,6 +54,25 @@ const checkTabForErrors = (
   }
 };
 
+const renderButtonContent = (
```
fisher-alice: Nit: maybe rename as 'renderTabContent' and below 'tabContent' and 'tabElement'?

apps/src/componentLibrary/tabs/_Tab.tsx
fisher-alice, line 122: I tested locally implementing DSCO 
Tabs with a Tooltip prop in `aichat`. I observed that every once in a while, the tooltip location would shift.
I'm unsure why but just wanted to see if you had any ideas or if you can repro. Thanks!

https://github.com/code-dot-org/code-dot-org/assets/107423305/8f3974a2-24a9-4055-8b20-d467ff17ecb7

Screenshot 2024-06-21 at 3 26 41 PM

Screenshot 2024-06-21 at 3 26 50 PM

fisher-alice (approved):
Looks good! Left a minor suggestion and observation but excited that we now have ability to add a tooltip to a tab. 🎉

apps/src/componentLibrary/tabs/_Tab.tsx
levadadenys: Thanks for suggesting!
This is intended here to avoid confusion with the `tab` `content` which is rendered inside `tabPanel`. The best alternative here is `renderTabButtonContent`, `tabButtonElement` in my opinion, but they feel like to bulky and extra to me

apps/src/componentLibrary/tabs/_Tab.tsx
levadadenys: Interesting, thanks for sharing!
Looks like it's first render issue, also might be related to hovering the icon. I was fixing similar issue with first render in this PR. Can you please confirm if it only appears on fist render or hover on icon?
Definetely worth fixing, though not sure if it should be done in this PR or in an follow up PR (there's another possible rework of tooltip component ahead which might eliminate this bug)

fisher-alice: > Looks like it's first render issue, also might be related to hovering the icon. I was fixing similar issue with first render in this PR. Can you please confirm if it only appears on fist render or hover on icon? 

Yes, it was on first render and didn't always occur. Since I believe @markabarrett  now prefers NOT using a Tooltip in `aichat`, adding a tooltip as a prop is not as time-sensitive.

apps/src/componentLibrary/tabs/_Tab.tsx
fisher-alice: I personally prefer `renderTabButtonContent` but not a strong opinion so good either way.

apps/src/componentLibrary/tabs/_Tab.tsx
levadadenys: Created a separate ticket for this: https://codedotorg.atlassian.net/browse/DESIGN2-130
Suggesting merge as is for now since aside from adding tooltip, there're also style updates to the tab in this PR.

fisher-alice: SGTM! Thanks for logging the bug.

PR #59303, opened 2024-06-20
Unit deletion script updates to be more flexible

vijayamanohararaj:
Unit deletion script does not allow for deletion when the unit is not in `in_development` or when there is any references to the unit from the section or progress table.

As we are looking at deleting units in CSD-2024 for task https://codedotorg.atlassian.net/browse/TEACH-1161, these limitations are too strict and can be made more flexible without significant user impact.

The changes in this PR include

1. Updating the condition to check for unit state to only block deletion of units in `stable` state
2. Instead of blocking the deletion when user progress is present, change the flow to confirm with the user. 

## Links

https://codedotorg.atlassian.net/browse/TEACH-1161

## Testing story

Manually validated the following through dry run mode

1. A script in `preview` state to ensure it will allow deletion
2. A script that has progress to ensure it will allow deletion (validated both 'yes' and 'no' flows on confirmation)
3. A script in stable state is blocked
4. A script assigned to sections (even if not in stable state) is blocked.

## Deployment strategy

DTP

## Follow-up work

Tracked by https://codedotorg.atlassian.net/browse/TEACH-1161

## Privacy

N/A

## Security

N/A

## Caching

N/A

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/bin/curriculum/delete_unit.rb
+++ b/bin/curriculum/delete_unit.rb
@@ -42,7 +42,7 @@ require_relative '../../dashboard/config/environment'
 options.unit_names.each do |unit_nm|
   unit_to_del = Unit.find_by(name: unit_nm)
   raise "Unit with name #{unit_nm} not found" if unit_to_del.nil?
-  raise "Only units in_development can be deleted as they wouldn't have usage and hence any DB references." unless unit_to_del.published_state == "in_development"
+  raise "Published units (state set as 'Stable') cannot be deleted as they could have active sections and progress." unless unit_to_del.published_state != "stable"
 
   assigned_section_count = Section.where(script_id: unit_to_del.id).count
   raise "Unit is currently assigned to some sections, deleting will break loading dashboard for those teachers." unless assigned_section_count == 0
--- a/bin/curriculum/delete_unit.rb
+++ b/bin/curriculum/delete_unit.rb
@@ -46,6 +46,7 @@ options.unit_names.each do |unit_nm|
   raise "Unit with name #{unit_nm} not found" if unit_to_del.nil?
   raise "Published units (state set as 'Stable') cannot be deleted as they could have active sections and progress." unless unit_to_del.get_published_state != "stable"
 
+  # Include additional warning if the script is not in one of the safe states where its not available for use externally.
   if unit_to_del.get_published_state != "in_development"
     print "Published state for Unit [#{unit_nm}] is not in_development, script might have usage. " \
       "Please confirm with curriculum team and ensure comms has been sent before deletion. Continue? (Y/N)"
--- a/bin/curriculum/delete_unit.rb
+++ b/bin/curriculum/delete_unit.rb
@@ -40,6 +40,8 @@ raise "unit name is required. Use -h for options." if options.unit_names.nil?
 
 require_relative '../../dashboard/config/environment'
 options.unit_names.each do |unit_nm|
+  puts "Processing unit [#{unit_nm}]..."
+
   unit_to_del = Unit.find_by(name: unit_nm)
   raise "Unit with name #{unit_nm} not found" if unit_to_del.nil?
   raise "Published units (state set as 'Stable') cannot be deleted as they could have active sections and progress." unless unit_to_del.published_state != "stable"
@@ -48,9 +50,12 @@ options.unit_names.each do |unit_nm|
   raise "Unit is currently assigned to some sections, deleting will break loading dashboard for those teachers." unless assigned_section_count == 0
 
   user_progress_count = UserScript.where(script_id: unit_to_del.id).count
-  raise "Some users have existing progress in this unit" unless user_progress_count == 0
+  if user_progress_count > 0
+    print "Some users have existing progress in unit [#{unit_nm}]. Continue with deletion? (Y/N)"
+    next unless gets.chomp.casecmp('y').zero?
+  end
 
-  puts "No references to Unit found from sections or user progress."
+  puts "No references to Unit [#{unit_nm}] found from sections."
   if options.dry_run
     puts "Skipping deletion due to dry run mode."
   else
--- a/bin/curriculum/delete_unit.rb
+++ b/bin/curriculum/delete_unit.rb
@@ -44,7 +44,13 @@ options.unit_names.each do |unit_nm|
 
   unit_to_del = Unit.find_by(name: unit_nm)
   raise "Unit with name #{unit_nm} not found" if unit_to_del.nil?
-  raise "Published units (state set as 'Stable') cannot be deleted as they could have active sections and progress." unless unit_to_del.published_state != "stable"
+  raise "Published units (state set as 'Stable') cannot be deleted as they could have active sections and progress." unless unit_to_del.get_published_state != "stable"
+
+  if unit_to_del.get_published_state != "in_development"
+    print "Published state for Unit [#{unit_nm}] is not in_development, script might have usage. " \
+      "Please confirm with curriculum team and ensure comms has been sent before deletion. Continue? (Y/N)"
+    next unless gets.chomp.casecmp('y').zero?
+  end
 
   assigned_section_count = Section.where(script_id: unit_to_del.id).count
   raise "Unit is currently assigned to some sections, deleting will break loading dashboard for those teachers." unless assigned_section_count == 0
--- a/bin/curriculum/delete_unit.rb
+++ b/bin/curriculum/delete_unit.rb
@@ -44,6 +44,9 @@ options.unit_names.each do |unit_nm|
 
   unit_to_del = Unit.find_by(name: unit_nm)
   raise "Unit with name #{unit_nm} not found" if unit_to_del.nil?
+
+  # Ensure the unit getting deleted is not marked stable. Using the get_published_state method to address units that are part of unit groups
+  # This would return the state for the unit if not null else fall back to return state of unit group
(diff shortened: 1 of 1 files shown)


bin/curriculum/delete_unit.rb
```diff
@@ -40,17 +40,22 @@ def parse_options
...
-  raise "Only units in_development can be deleted as they wouldn't have usage and hence any DB references." unless unit_to_del.published_state == "in_development"
+  raise "Published units (state set as 'Stable') cannot be deleted as they could have active sections and progress." unless unit_to_del.published_state != "stable"
 
   assigned_section_count = Section.where(script_id: unit_to_del.id).count
   raise "Unit is currently assigned to some sections, deleting will break loading dashboard for those teachers." unless assigned_section_count == 0
 
   user_progress_count = UserScript.where(script_id: unit_to_del.id).count
-  raise "Some users have existing progress in this unit" unless user_progress_count == 0
+  if user_progress_count > 0
+    print "Some users have existing progress in unit [#{unit_nm}]. Continue with deletion? (Y/N)"
```
vijayamanohararaj: Just say continue to avoid confusion during dry run mode

bin/curriculum/delete_unit.rb
```diff
@@ -40,17 +40,22 @@ def parse_options
...
+    next unless gets.chomp.casecmp('y').zero?
+  end
 
-  puts "No references to Unit found from sections or user progress."
+  puts "No references to Unit [#{unit_nm}] found from sections."
```
vijayamanohararaj: May be include, "safe to move forward with deletion."

bin/curriculum/delete_unit.rb
```diff
@@ -40,17 +40,22 @@ def parse_options
 
 require_relative '../../dashboard/config/environment'
 options.unit_names.each do |unit_nm|
+  puts "Processing unit [#{unit_nm}]..."
+
   unit_to_del = Unit.find_by(name: unit_nm)
   raise "Unit with name #{unit_nm} not found" if unit_to_del.nil?
-  raise "Only units in_development can be deleted as they wouldn't have usage and hence any DB references." unless unit_to_del.published_state == "in_development"
+  raise "Published units (state set as 'Stable') cannot be deleted as they could have active sections and progress." unless unit_to_del.published_state != "stable"
```
davidsbailey: the rules around published state are a bit non-intuitive and I'm concerned that this is no longer enforcing the rules that it looks like its enforcing. here's how published state works (a few examples):

1. courese csd-2023 has published state stable, course csd3-2023 has published state `nil` (for units in a course it is inherited from the course)
2. when you mark a unit as "hidden", we actually overload the published_state field to represent this. for example, right now csd-2024 has published_state preview, and csd6b-2024 has published_state "in_development" (because it was hidden).
3. standalone units like coursea-2023 use their own published state field to determine their published state

so my concern is that this makes it easier to delete a unit in a published course, while giving the impression that it is protecting against this case by checking the published state. 

so, the original code (perhaps accidentally?) requires the unit to either be hidden, or to be truly in_development.

if you did want to get the published state of a unit, agnostic of whether it is in a course, you could use `unit.get_published_state`.

I don't feel strongly about having especially strong protections here... but I would like whatever we include here to not be misleading.

any preferred options given this analysis?

davidsbailey: to clarify i don't feel strongly about strong protections because it seems more important whether a section is assigned to the unit than anything else.

vijayamanohararaj: Thanks Dave for that input. I wasn't aware that published state for units can be inherited potentially from the overall course. Yes, the original was accidentally preventing a unit from getting deleted. 

I agree with your concern around misleading sense of protection because of the inheritance of published state between unit group. I see two potential options here, one to use the get_published_state to address this or not do any checks at all, but rather include a warning message asking the user of the script to confirm the script can be deleted.

I think doing a combination of both would be good, i.e., update the check to use get_published_state and also include a warning message, something like, "You are deleting a script that is not in_development (and this can extend in the future to include a set of safe states)  - please confirm with the curriculum team and involve CX to provide appropriate comms."  Thoughts?

davidsbailey: yes, sounds good to me, although there may be an extra step to NOT warn in the case where the unit is within a published unit group but has been marked as hidden.

vijayamanohararaj: Updated to include the warning and use `get_published_state` - using the method also ensured the case where unit is within a published unit group, but marked hidden.

Including output from local run. microbit was in preview state, while the other two had state set in_development
```vijaya@VijiMiniDskTp:~/source/code-dot-org$ bin/curriculum/delete_unit.rb -u microbit-2024,focus-on-coding4b-2024,focus-on-creativity4b-2024 --dry-run
GetSecretValue: development/cdo/slack_bot_token
WARNING: MYSQL_OPT_RECONNECT is deprecated and will be removed in a future version.
Processing unit [microbit-2024]...
WARNING: MYSQL_OPT_RECONNECT is deprecated and will be removed in a future version.
Published state for Unit [microbit-2024] is not in_development, script might have usage. Please confirm with curriculum team and ensure comms has been sent before deletion. Continue? (Y/N)n
Processing unit [focus-on-coding4b-2024]...
No references to Unit [focus-on-coding4b-2024] found from sections. All validations for deletion passed.
Skipping deletion due to dry run mode.
Processing unit [focus-on-creativity4b-2024]...
No references to Unit [focus-on-creativity4b-2024] found from sections. All validations for deletion passed.
Skipping deletion due to dry run mode.

File: bin/curriculum/delete_unit.rb [hunk 4]
```diff
@@ -40,17 +40,29 @@ def parse_options
...
   raise "Unit with name #{unit_nm} not found" if unit_to_del.nil?
-  raise "Only units in_development can be deleted as they wouldn't have usage and hence any DB references." unless unit_to_del.published_state == "in_development"
+  raise "Published units (state set as 'Stable') cannot be deleted as they could have active sections and progress." unless unit_to_del.get_published_state != "stable"
+
+  # Include additional warning if the script is not in one of the safe states where its not available for use externally.
+  if unit_to_del.get_published_state != "in_development"
+    print "Published state for Unit [#{unit_nm}] is not in_development, script might have usage. " \
+      "Please confirm with curriculum team and ensure comms has been sent before deletion. Continue? (Y/N)"
+    next unless gets.chomp.casecmp('y').zero?
+  end
```
davidsbailey, line 57: I think I see a couple of problems here. if the course is stable, but the unit is in development, we should not raise. also, if the course is in development, and the unit is NOT in development, then we shouldn't warn.

I think you have two main choices, one is to write a single, short condition that catches all bad cases and the other is to have different messages for the two cases.

if just one case, the original code is pretty good, and could just use a better comment. or you could do:
```
raise "..." unless unit.get_published_state != 'stable' || unit.published_state == 'in_development'
```

If you want to give just the right error message in all cases, it might be something like:
```
# this means "if the unit is a standalone unit"
if unit.is_course?
  raise "cannot delete standalone unit in stable state" if unit.pubished_state == 'stable'
else
  raise "cannot delete unit in a stable course unless unit is hidden. to fix this..." if unit.get_published_state == 'stable' && unit.published_state != 'in_development'
``

vijayamanohararaj: Good catch here, thank you! I like the more specific error, but shouldn't the condition to also allow for cases where course is not stable. For cases example, in case where course is in preview and unit within is in any state, we are probably ok to go ahead with deletion (unless there are section references)??.

davidsbailey: yes I may have missed a case. I think it depends on what you want the new behavior to be, if it is not the same as the original code in the repo.

vijayamanohararaj: Actually, looking through this more, the get_published_state method returns the unit's published_state as long as it is not null. So

>  if the course is stable, but the unit is in development

This would not raise the exception in current implementation as `get_published_state` would return in_development

if the course is in development, and the unit is NOT in development, In this case as well, unless the unit is marked stable, there shouldn't be an exception.

I'm a bit confused here - happy to hop on a call if that is easier.

davidsbailey: > Actually, looking through this more, the get_published_state method returns the unit's published_state as long as it is not null

I see, I had missed this. good catch! if there is subtle behavior like this you are relying on, I think it will be helpful to explain in a comment (or if possible to do accurately, in the error message)

vijayamanohararaj: Just including a comment for now as I'm also planning on following up with a change to not allow deletion of units that are still part of unit groups. That should simplify the script quiet a bit and can do more clean as part of that change.

davidsbailey: @vijayamanohararaj I am just remembering that the page that issues the delete command may not already have the unit id available on it. while it could be possible to provide it, I think it would also be acceptable to just go back to deleting by name. I would like to avoid the situation where the delete command accepts either name or id, though.

thank you for taking this on! sorry you ended up having to fix my mistake.

davidsbailey: sorry, it looks like i put this comment in the wrong PR. removing it here since we also talked about it in slack

davidsbailey (approved):
Thank you for the thorough consideration of all the issues here. Looks good!

PR #59304, opened 2024-06-20
remove old courses/index page

davidsbailey:
remove dead code. studio.code.org/courses now redirects to https://code.org/students, so the courses/index route is no longer needed.

## Testing story

* updated existing unit tests
* verified redirect still works locally


bethanyaconnor (approved):
🚮

bethanyaconnor:
Thank you so much for picking this up!

PR #59305, opened 2024-06-20
Creates experiment flag for New Sign-Up Flow

TurnerRiley:
Creates a new experiment flag for the new sign-up flow. All of the new changes will remain behind this flag so that we can easily set up the A/B experiment later like in [this PR](https://github.com/code-dot-org/code-dot-org/pull/58470/files).

## Links
Spec: [Step 1 here](https://docs.google.com/document/d/1-Bxt7orcLUsc6sHR8dKYj3XmfMNs3pG7AK8j0xeVIG8/edit)


kelbyhawn (approved):
🎉

bethanyaconnor (approved):
Change looks good! Just to be clear, the plan is to use this to hide the old flow using hacky JS if it's true, right?

TurnerRiley:
> Change looks good! Just to be clear, the plan is to use this to hide the old flow using hacky JS if it's true, right?

Yep!

PR #59314, opened 2024-06-20
Prevent new account creation during account linking flow

carl-codeorg:
If a user attempts to link an oauth account and succeeds in the oauth login even though they don't have an existing Code.org account, the current code will create a new account for them. However, the linking won't happen, which is a confusing UX.

This PR changes the linking flow so that if you successfully log in but we can't find an existing account, it kicks you back to the login screen with an error instead of creating a new account.

## Links

[JIRA](https://codedotorg.atlassian.net/browse/P20-994)

## PR Checklist:

- [ ] Code is well-commented


nicklathe (approved):
LGTM

PR #59341, opened 2024-06-21
Refactor AI Chat message types

sanchitmalhotra126:
A follow-up refactor, on the heels of Ben's awesome [chat logging refactor](https://github.com/code-dot-org/code-dot-org/pull/58668). The main motivation for this change is that, as our lab has grown, the `ChatCompletionMessage` type has gotten a bit overloaded and bloated since it now represents user chat messages, assistant chat messages, model update notifications, and error notifications, as well as potentially other content types as we support more features. This changes splits these different use cases out into discrete types, that all extend from a common `ChatItem` base type. Because Ben's change splits the list of chat items into separate sections (past, current, and pending), one other advantage of this is we can restrict parts of the chat to certain types, like enforce that the `chatMessagePending` must be an actual chat message (rather than a model update or notification).

Also made a couple of minor fixes/cleanups along the way
- added a `ChatApiResponse` type to represent the data returned from the server
- a few small TS optimizations
- privated some reducer actions in the aichatRedux slice since they're not meant to be used outside of that file
- removed some unused types
- fixed an issue where the "reset model" notification would always show the same timestamp
- show the model display name rather than model ID in the update notification
- consistent storage of chat item timestamp (store the UTC epoch millis on the model, and convert to appropriate format when displaying/copying).

## Testing story

Tested for regressions on allthethings and a few pilot levels.


apps/src/aichat/redux/aichatRedux.ts
```diff
@@ -554,32 +535,35 @@ export const selectAllFieldsHidden = createSelector(
...
+} = aichatSlice.actions;
 
 registerReducers({aichat: aichatSlice.reducer});
 export const {
-  addChatMessage,
+  addNotification,
   removeUpdateMessage,
   setNewChatSession,
   setChatSessionId,
-  setChatMessagePending,
```
bencodeorg, line 576: Nice, thanks for cleaning this up!

apps/src/aichat/types.ts
```diff
@@ -3,24 +3,48 @@ import {LevelProperties} from '@cdo/apps/lab2/types';
...
-  timestamp?: string;
-};
+}
 
-export type AichatCompletionMessage = Pick<
-  ChatCompletionMessage,
-  'role' | 'chatMessageText' | 'status'
->;
+export interface ModelUpdate extends ChatItem {
+  id: number;
```
bencodeorg, line 18: Maybe overkill, but any value in `Notification` and `ModelUpdate` both being derived from a type that has an `id` (`ChatItemRemovable`, since that's why we add an id field)?

apps/src/aichat/redux/aichatRedux.ts
```diff
@@ -295,12 +281,9 @@ export const submitChatContents = createAsyncThunk(
 
     let chatApiResponse;
     try {
-      const conversationMessages = chatMessagesCurrent.filter(
-        message => message.role === Role.USER || message.role === Role.ASSISTANT
-      );
       chatApiResponse = await postAichatCompletionMessage(
         newMessage,
-        conversationMessages,
+        chatItemsCurrent.filter(isChatMessage) as ChatMessage[],
```
bencodeorg, line 287: Slick :)

apps/src/aichat/views/ChatMessage.tsx
```diff
@@ -119,24 +141,30 @@ const displayModelUpdateMessage = (
 };
 
 const displayErrorMessage = (
```
bencodeorg: Rename this since it's being used to display non-error notifications?

apps/src/aichat/views/AichatView.tsx
```diff
@@ -32,30 +30,24 @@ import {
...
 import ModelCustomizationWorkspace from './ModelCustomizationWorkspace';
 import PresentationView from './presentation/PresentationView';
 import CopyButton from './CopyButton';
 import moduleStyles from './aichatView.module.scss';
 import aichatI18n from '../locale';
-import {getCurrentTime, getNewMessageId} from '../redux/utils';
+import {getNewMessageId} from '../redux/utils';
 
-const RESET_MODEL_NOTIFICATION: ChatCompletionMessage = {
+const getResetModelNotification = (): Notification => ({
```
bencodeorg, line 47: Good catch re: not having the right timestamp here! Looks like we're dumping the bold/non-bold here (making it all bold I think?), yeah? Seems ok to me, just confirming.

apps/src/aichat/types.ts
```diff
@@ -3,24 +3,48 @@ import {LevelProperties} from '@cdo/apps/lab2/types';
...
+  updatedField: keyof AiCustomizations;
+  updatedValue: AiCustomizations[keyof AiCustomizations];
+}
 
-type ChatMessageSuffix = {
+export interface Notification extends ChatItem {
```
bencodeorg, line 23: Can we remove certain `Role` types now (ERROR_NOTIFICATION, MODEL_UPDATE)?

bencodeorg (approved):
Looks great, thank you for doing this! 🎉

apps/src/aichat/types.ts
sanchitmalhotra126: Yeah good idea, I think as a next step I wanted to look into component-izing the actual chat UI components, in which case I can see `ModelUpdate` and `Notification` being rendered by some single `ChatNotification` component that would have a common interface. So might make sense to do it then?

apps/src/aichat/views/AichatView.tsx
sanchitmalhotra126: Yeah good catch, it does show up as all bold. When we move the chat components to common components, we could look into possibly supporting markdown so we can preserve the original formatting?

apps/src/aichat/types.ts
sanchitmalhotra126: Yep, done!

apps/src/aichat/views/ChatMessage.tsx
sanchitmalhotra126: Done!

apps/src/aichat/redux/aichatRedux.ts
```diff
@@ -29,38 +27,41 @@ import {
...
-  decorateMessageFromModelResponse,
   findChangedProperties,
   getNewMessageId,
-  getCurrentTime,
-  getCurrentTimestamp,
   hasFilledOutModelCard,
 } from './utils';
 
-type MessageListKey = 'chatMessagesCurrent' | 'chatMessagesPast';
+const messageListKeys = ['chatItemsPast', 'chatItemsCurrent'] as const;
```
fisher-alice, line 51: Just curious - Is this the preferred way to create a type as opposed to what was initially here?

apps/src/aichat/redux/aichatRedux.ts
```diff
@@ -29,38 +27,41 @@ import {
...
 type MessageLocation = {
   index: number;
-  messageListKey: MessageListKey;
+  messageListKey: (typeof messageListKeys)[number];
```
fisher-alice, line 54: Sorry - still trying to understand TS typing. Do you mind explaining this syntax?

```
(typeof messageListKeys)[number]
```

apps/src/aichat/redux/aichatRedux.ts
fisher-alice: Oh! I think it helps simplify these lines:
https://github.com/code-dot-org/code-dot-org/blob/33f65b5316207fbe4558a38b76d93c4a2bd1e243/apps/src/aichat/redux/aichatRedux.ts#L507-L508

apps/src/aichat/types.ts
```diff
@@ -3,24 +3,48 @@ import {LevelProperties} from '@cdo/apps/lab2/types';
...
+  id: number;
   text: string;
-  boldtypeText?: string;
-};
+  notificationType: 'error' | 'success';
+}
+
+// Type Predicates: checks if a ChatItem is a given type, and more helpfully,
+// automatically narrows to the specific type.
```
fisher-alice, line 30: TIL about type predicates!
It seems like the function returns a boolean, but you write that the function automatically narrows to the specific type as well.
Do you mean that it is used to check the type in a condition and within the body confirmed type is used without error?

apps/src/aichat/views/AichatView.tsx
fisher-alice: I wonder if this is more of a `ModelUpdate` type?

fisher-alice:
> Looks great, thank you for doing this! 🎉

Awesome refactor. Always learn a lot from your PRs! 🧹 🎉

PR #59344, opened 2024-06-21
Disallow admins from linking LTI accounts

nicklathe:
Adds logic to disallow linking for admins, and returns an error message to them. This alert is displayed after they attempt to link the account, as that's the point where the existing user account is looked up and determined to be an admin.

Since this error message will only be for internal Code.org users, a simple flash alert was used.

**Alert**
Screenshot 2024-06-26 at 9 00 24 AM

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/dashboard/app/controllers/lti/v1/account_linking_controller.rb
+++ b/dashboard/app/controllers/lti/v1/account_linking_controller.rb
@@ -31,6 +31,10 @@ module Lti
         head :bad_request unless PartialRegistration.in_progress?(session)
         params.require([:email, :password])
         existing_user = User.find_by_email_or_hashed_email(params[:email])
+        if existing_user.admin?
+          flash[:alert] = "Admins cannot link LTI accounts"
+          redirect_to user_session_path(lti_provider: params[:lti_provider], lms_name: params[:lms_name]) and return
+        end
         if existing_user&.valid_password?(params[:password])
           Services::Lti::AccountLinker.call(user: existing_user, session: session)
           sign_in existing_user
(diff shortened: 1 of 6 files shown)


dashboard/app/controllers/lti/v1/account_linking_controller.rb
```diff
@@ -31,6 +31,10 @@ def link_email
         head :bad_request unless PartialRegistration.in_progress?(session)
         params.require([:email, :password])
         existing_user = User.find_by_email_or_hashed_email(params[:email])
+        if existing_user.admin?
+          flash[:alert] = "Admins cannot link LTI accounts"
```
nicklathe: TODO: Add this string to i18n JSON

dashboard/test/controllers/lti/v1/account_linking_controller_test.rb
```diff
@@ -3,6 +3,7 @@
 class Lti::V1::AccountLinkingControllerTest < ActionController::TestCase
   setup do
     @user = create(:teacher, email: 'test@lti.com')
+    @admin = create :admin, email: 'admin@lti.com'
```
nicklathe: Interestingly, `@admin.admin` here is `true` but this test fails, and when this user is looked up in the link_email endpoint, it's nil.

dashboard/app/controllers/omniauth_callbacks_controller.rb
```diff
@@ -549,6 +549,10 @@ def login
   # For linking new LTI auth options to existing accounts
   private def link_accounts(user)
     if user
+      if user.admin
```
carl-codeorg: Any reason this check is `admin` sans `?`?

wilkie (approved):
LGTM

dashboard/app/controllers/omniauth_callbacks_controller.rb
nicklathe: Yeah, good catch. I meant to add the `?` in for this.

carl-codeorg (approved):

PR #59346, opened 2024-06-21
Add predict options to Python Lab level editing

molly-moen:
This is setup for enabling predict levels in Python Lab. In this PR I updated the python lab level edit page to show predict level edit options. See the linked spec below for details on how predict levels will be different in Python Lab vs older labs (TL;DR we are doing away with having a separate "contained" level that has the free response/multiple choice configuration).

I modeled the editing interface off of similar work for AI Chat, which has an `aichat_settings` json property which is edited via a React component. Here we have a `predict_settings` property, for now only on Python Lab, which controls whether a level is a predict level, what type of predict level it is, and any specific configuration for that type. Predict levels can either be free response or multiple choice questions.

Once we have proved out that this works, we may want to move the `predict_settings` property up to the `level` model so other lab2 lab types can use it without duplicated work. As I work on this I am aiming to do as much work as possible in shared code so we can reuse this for other lab2 labs.

### Screenshots
Level edit page if predict is turned off
Screenshot 2024-06-21 at 3 50 56 PM

Editing a free response predict level
Screenshot 2024-06-21 at 3 51 04 PM

Editing a multiple choice predict level
Screenshot 2024-06-21 at 3 51 28 PM

Users can select multiple correct answers, but when we save we verify that there is at least 1 correct answer specified.

## Links
- jira ticket: [CT-634](https://codedotorg.atlassian.net/browse/CT-634)

## Testing story
I verified that we can save the new fields to the level. Using them is a follow-up, so I probably missed some configuration in this initial pass. This PR was pretty big on its own so it made sense to do this work separately. I made a new all the things python lab level that will be the predict level example level.

## Follow-up work
There are many predict level tasks linked in the spec above and on the [Codebridge kanban board](https://codedotorg.atlassian.net/jira/software/c/projects/CT/boards/92?quickFilter=180).

## PR Checklist:

- [ ] Code is well-commented


apps/src/lab2/levelEditors/types.ts
```diff
@@ -0,0 +1,15 @@
+export interface LevelPredictSettings {
+  is_predict_level: boolean;
```
sanchitmalhotra126: minor: not sure if we have a consistent style for this, but should these be camel case since they're defined in apps code?

apps/src/lab2/levelEditors/predictSettings/EditPredictSettings.tsx
```diff
@@ -0,0 +1,119 @@
...
+}
+
+const EditPredictSettings: React.FunctionComponent<
+  EditPredictSettingsProps
+> = ({initialSettings}) => {
+  const [predictSettings, setPredictSettings] = useState<LevelPredictSettings>(
+    initialSettings || {is_predict_level: false}
+  );
+
+  const handleIsPredictToggle = () => {
```
sanchitmalhotra126, line 33: nit: `useCallback` for these functions?

dashboard/app/models/levels/pythonlab.rb
```diff
@@ -48,4 +50,16 @@ def self.create_from_level_builder(params, level_params)
   def uses_lab2?
     true
   end
+
+  # Ensure that if this is a multiple choice predict level, there is at least one correct answer
```
sanchitmalhotra126, line 55: Should we check this on the front end too, before saving?

apps/src/lab2/levelEditors/predictSettings/EditPredictSettings.tsx
```diff
@@ -0,0 +1,119 @@
...
+    setPredictSettings({
+      ...predictSettings,
+      is_predict_level: !predictSettings.is_predict_level,
+      question_type:
+        predictSettings.question_type || PredictQuestionType.FreeResponse,
+    });
+  };
+
+  const handleQuestionTypeChange = (
```
sanchitmalhotra126, line 42: Looks like we don't clear out settings when changing the question type, right? I can see that being useful so work isn't lost when toggling between types, but when saving to level, should we clear out unused data? Any potential issues if a level has both free response data and multiple choice data?

apps/src/lab2/levelEditors/predictSettings/MultipleChoiceFields.tsx
```diff
@@ -0,0 +1,108 @@
...
+    ) {
+      newCorrectAnswers.splice(newCorrectAnswers.indexOf(e.target.value), 1);
+    }
+    setPredictSettings({
+      ...predictSettings,
+      multiple_choice_answers: newCorrectAnswers,
+    });
+  };
+
+  return predictSettings.multiple_choice_options ? (
```
sanchitmalhotra126: nit: early return instead of ternary?

apps/src/lab2/levelEditors/predictSettings/EditPredictSettings.tsx
```diff
@@ -0,0 +1,119 @@
...
+      ...predictSettings,
+      question_type: e.target.value as PredictQuestionType,
+    });
+  };
+
+  const renderMultipleChoiceOptions = () => {
+    if (!predictSettings.multiple_choice_options) {
+      setPredictSettings({
+        ...predictSettings,
+        multiple_choice_options: [''],
```
sanchitmalhotra126: Could this be set in the initial useState assignment so it doesn't have to be checked every time?

sanchitmalhotra126 (approved):
Awesome! Wayyy simpler than creating contained levels - thanks for doing this!

apps/src/lab2/levelEditors/types.ts
molly-moen: yeah good point, I used snake case because that's what everything on the backend uses, but I see we are using camel case for the ai chat settings, and it was additional work to camelize them when we send them back to the frontend later. I'll update this.

dashboard/app/models/levels/pythonlab.rb
molly-moen: This will cause an error message to show up for levelbuilders when they click save. I didn't want to have annoying red errors while they were editing, and it's not a pattern we currently follow.

apps/src/lab2/levelEditors/predictSettings/EditPredictSettings.tsx
molly-moen: Yeah I debated this, it's probably a good idea to clear out the data on save. I was planning to ignore unused fields on the frontend but it's definitely cleaner to just delete them.

apps/src/lab2/levelEditors/predictSettings/EditPredictSettings.tsx
molly-moen: I'm a little fuzzy on when we should use `useCallback`. The [docs](https://react.dev/reference/react/useCallback) say to use it sparingly, and this case doesn't seem to fall into the cases they talk about. Is there a reason you think it would be useful here?

apps/src/lab2/levelEditors/predictSettings/EditPredictSettings.tsx
molly-moen: good call

apps/src/lab2/levelEditors/predictSettings/EditPredictSettings.tsx
sanchitmalhotra126: Hmm yeah I thought generally it was good practice to wrap most functions with it, but you're right that the docs say otherwise. Generally I think if the function is going to be passed as a prop to another component or used as a dependency for another hook then it's a good idea to wrap it - in this case though, the `setPredictSettings` function would be passed as a dependency if this was wrapped in a useCallback, so maybe there's some slight optimization there, but i don't think `setPredictSettings` is meant to change that often, so it's probably fine?

PR #59347, opened 2024-06-21
remove implementation of alternate units from the codebase

davidsbailey:
* Follows https://github.com/code-dot-org/code-dot-org/pull/63387. 
* Continues [TEACH-1545: remove "alternate units" implementation, but keep the unit_group_units join table](https://codedotorg.atlassian.net/browse/TEACH-1545)

In 2017, we added the ability to customize the appearance of each course based on whether the student viewing the course was in a certain experiment. The only instance of this ever launched was the "subgoals" experiment, which would show any student in the experiment https://studio.code.org/s/csp3-research-mxghyt instead of https://studio.code.org/s/csp3-2017 after they completed csp2-2017. 

It's time to remove this feature now that we are starting on work for [modular curriculum](https://docs.google.com/document/d/1e8Xs_azDTJRyJoGLSk7G2v23rpyA-ZQzgSVnle6exvg/edit?tab=t.0) because:
* removing the unused implementation will simplify the modularity work
* the feature has been unused for 7 years
* should the need arise again, modular curriculum will make it easy to solve the same problem, by creating two courses with some shared units

### terminology

a UnitGroupUnit is a join model which represents the presence of a unit in a unit group. these fall into 2 categories: 
* https://github.com/code-dot-org/code-dot-org/blob/2f6d9c3fc029b7ee13b5ce4dbab8a39048895bc2/dashboard/app/models/unit_group.rb#L34
* https://github.com/code-dot-org/code-dot-org/blob/2f6d9c3fc029b7ee13b5ce4dbab8a39048895bc2/dashboard/app/models/unit_group.rb#L36

default unit group units are the normal case. alternate unit group units are the special units that only show up if the user is in a certain experiment.

## Testing story

existing test coverage from:
* unit test files modified in this PR
* any UI tests covering units in courses (CSP, CSD, etc) provide some additional coverage

## Deployment strategy 

- [x] merge https://github.com/code-dot-org/code-dot-org/pull/63387
- [x] wait for a production deploy
- [x] confirm all alternate unit group units have been removed in production
- [x] merge this PR
- [ ] merge https://github.com/code-dot-org/code-dot-org/pull/63402


dashboard/config/scripts_json/csp3-a.script_json
davidsbailey, line 1: this unit is safe to remove because it was only used in ui tests that were previously removed. however, [csp3-research-mxghyt](https://studio.code.org/s/csp3-research-mxghyt) is not being removed because some students could have progress in that unit.

dashboard/test/ui/features/step_definitions/experiment_steps.rb
```diff
@@ -1,13 +1,3 @@
-When /^I enable the "([^"]*)" course experiment$/ do |experiment_name|
-  steps <<-STEPS
-    Given I am on "http://studio.code.org/experiments/set_course_experiment/#{experiment_name}"
-    And I get redirected to "/" via "dashboard"
-    And I get redirected to "/home" via "dashboard"
-    And I wait to see ".alert-success"
-    Then element ".alert-success" contains text "You have successfully joined the experiment"
-  STEPS
-end
```
davidsbailey, line 9: this ui test step has been unused since 2022: https://github.com/code-dot-org/code-dot-org/pull/45489

pegasus/test/fixtures/fake_dashboard.rb
```diff
@@ -54,121 +54,10 @@ module FakeDashboard
...
-      course_id: COURSE_CSP[:id],
-      script_id: UNIT_CSP2[:id],
-      position: 2
-    },
-    {
-      course_id: COURSE_CSP[:id],
-      script_id: UNIT_CSP2_ALT[:id],
-      position: 2,
-      experiment_name: 'csp2-alt-experiment',
-      default_script_id: UNIT_CSP2[:id]
```
davidsbailey, line 148: the pegasus test changes are a bit of bonus cleanup. I noticed them because we were using alternate units here, then discovered there was a bigger chunk of test code here that is unused.

daynew (commented):
`dashboard/config/scripts_json/csp3-a.script_json` is deleted in this PR. Do we want to also delete `dashboard/config/scripts_json/csp3-research-mxghyt.script_json`?

davidsbailey:
> `dashboard/config/scripts_json/csp3-a.script_json` is deleted in this PR. Do we want to also delete `dashboard/config/scripts_json/csp3-research-mxghyt.script_json`?

Yes, good idea! The unit won't go away, but it'll make it more obvious that we are not maintaining it. will do as soon as git push starts working again.

daynew (approved):
These changes make sense to me. It would be nice if we had a deprecation process for the Units no longer being supported, but that can be a problem for another day.

davidsbailey:
confirmed that the previous PR had the intended effect:
```
[production] dashboard > UnitGroupUnit.where.not(experiment_name: nil).count
=> 0
```

PR #59352, opened 2024-06-24
Add import order rule to all unit test files

lfryemason:
This is a big one, but because it is all unit test files, I feel very confident that there will be no problems as long as drone passes

## PR Checklist:

- [ ] Code is well-commented


apps/test/unit/p5lab/reducersTest.js
```diff
@@ -1,12 +1,22 @@
+// Require statements can change behavior based on order.
+// We are keeping the original order of the require statements.
```
Hamms, line 2: Is there a specific reason for this file why we want to preserve the original order?

Hamms (approved):
One non-blocking question, otherwise LGTM! Agreed that this seems safe, despite representing a large set of changes; always nice to be able to rely on tests. Getting close to full coverage here; exciting to see!

Thank you!

apps/test/unit/p5lab/reducersTest.js
lfryemason: I'll work on removing it in a quick follow up PR

PR #59365, opened 2024-06-24
feat(lms): add a state for LMS account linking to enforce the requirement of making a decision for linking

stephenliang:
This PR adds a new state to the user property bag which determines whether a LTI user has made a decision regarding account linking. Unless a user specifically opts out of account linking, the user will be sent to the account linking page to make a decision.

## Links

- [jira ticket](https://codedotorg.atlassian.net/browse/P20-1009)

## Testing story

Tested on [adhoc](https://adhoc-stephen-lms-lockout-studio.cdn-code.org/users/sign_in)

#### Scenario 1a: Click on “Continue to Code.org”

Expected: Opt-out of account linking

#### Scenario 1b: Click on “sign in with my account”

Expected: Remains opted-in to account linking (via opt-in to account linking by default)

#### Scenario 1c: Click on any of the links in the header or footer

Expected: Remains opted-in to account linking (via opt-in to account linking by default)

#### Scenario 1d: Close the browser window or tab

Expected: Remains opted-in to account linking (via opt-in to account linking by default)

#### Scenario 2a: Successfully login to account

Expected: Opt-out of account linking

#### Scenario 2b: Unsuccessfully login to account

Expected: Remains opted-in to account linking (via opt-in to account linking by default)

#### Scenario 2c: Click “go back”

Expected: Remains opted-in to account linking (via opt-in to account linking by default)

#### Scenario 2d: Click on any of the links in the header or footer

Expected: Remains opted-in to account linking (via opt-in to account linking by default)

#### Scenario 2e: Close the browser window or tab
Expected: Remains opted-in to account linking (via opt-in to account linking by default)

## PR Checklist:

- [x] Code is well-commented

### DIFF
--- a/apps/src/lib/ui/lti/link/LtiLinkAccountPage/cards/NewAccountCard/index.tsx
+++ b/apps/src/lib/ui/lti/link/LtiLinkAccountPage/cards/NewAccountCard/index.tsx
@@ -8,7 +8,7 @@ import FontAwesomeV6Icon from '@cdo/apps/componentLibrary/fontAwesomeV6Icon';
 import classNames from 'classnames';
 import styles from '@cdo/apps/lib/ui/lti/link/LtiLinkAccountPage/link-account.module.scss';
 import {Button, buttonColors} from '@cdo/apps/componentLibrary/button';
-import React, {useContext, useRef} from 'react';
+import React, {useContext, useRef, useState} from 'react';
 import i18n from '@cdo/locale';
 import {LtiProviderContext} from '../../context';
 import DCDO from '@cdo/apps/dcdo';
@@ -16,6 +16,7 @@ import RailsAuthenticityToken from '@cdo/apps/lib/util/RailsAuthenticityToken';
 import {navigateToHref} from '@cdo/apps/utils';
 import analyticsReporter from '@cdo/apps/lib/util/AnalyticsReporter';
 import {PLATFORMS} from '@cdo/apps/lib/util/AnalyticsConstants';
+import {getAuthenticityToken} from '@cdo/apps/util/AuthenticityTokenStore';
 
 const NewAccountCard = () => {
   const {ltiProviderName, newAccountUrl, emailAddress} =
@@ -25,8 +26,9 @@ const NewAccountCard = () => {
     'student-email-post-enabled',
     false
   );
+  const [isSaving, setIsSaving] = useState(false);
 
-  const handleNewAccountSubmit = () => {
+  const handleNewAccountSaved = () => {
     const eventPayload = {
       lms_name: ltiProviderName,
     };
@@ -40,6 +42,22 @@ const NewAccountCard = () => {
     } else {
       navigateToHref(newAccountUrl);
     }
+  };
+
+  const handleNewAccountSubmit = async () => {
+    setIsSaving(true);
+
+    fetch('/lti/v1/account_linking/new_account', {
+      method: 'POST',
+      headers: {
+        'Content-Type': 'application/json',
+        'X-CSRF-Token': await getAuthenticityToken(),
+      },
+    }).then(response => {
+      if (response.ok) {
+        handleNewAccountSaved();
+      }
+    });
   };
 
   return (
@@ -75,6 +93,8 @@ const NewAccountCard = () => {
           color={buttonColors.white}
           size="l"
           text={i18n.ltiLinkAccountNewAccountCardActionLabel()}
+          isPending={isSaving}
+          disabled={isSaving}
           onClick={handleNewAccountSubmit}
         />
       </CardActions>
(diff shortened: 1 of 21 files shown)


apps/src/lib/ui/lti/link/LtiLinkAccountPage/cards/NewAccountCard/index.tsx
```diff
@@ -42,6 +44,22 @@ const NewAccountCard = () => {
     }
   };
 
+  const handleNewAccountSubmit = async () => {
+    setIsSaving(true);
+
+    fetch('/lti/v1/account_linking/new_account', {
```
carl-codeorg, line 50: Since we're in an async function, is it cleaner to await the fetch? Also, do we need to add handling for a non-200 response? Uncaught front-end errors won't notify via Honeybadger IIRC

dashboard/app/controllers/application_controller.rb
```diff
@@ -356,6 +356,29 @@ def prevent_caching
     redirect_to lockout_path if Policies::ChildAccount::ComplianceState.locked_out?(current_user)
   end
 
+  # Check that the user has completed the LMS onboarding flow
+  protected def assert_lms_landing_policy
+    return unless current_user.authentication_options.first&.lti? && !current_user.lms_landing_opted_out
```
carl-codeorg: The first condition could be replaced with the `lti?` helper from Policies::Lti

dashboard/app/controllers/application_controller.rb
```diff
@@ -356,6 +356,29 @@ def prevent_caching
...
+
+    # URLs we should not redirect.
+    return if Set[
```
carl-codeorg, line 378: Do we need to add the `user_session_path` to allow them to get to the link existing account page?

dashboard/lib/services/lti.rb
```diff
@@ -299,5 +299,12 @@ def self.sync_course_roster(lti_integration:, lti_course:, nrps_sections:, curre
       end
       course_sync_result
     end
+
+    def self.initialize_lms_landing_session(session, lti_provider_name, new_cta_type)
+      session[:lms_landing] = {
+        lti_provider_name: lti_provider_name,
```
carl-codeorg, line 305: minor nit: could change this to `lti_provider` to match the name of the param in the view

dashboard/lib/services/lti/account_linker.rb
```diff
@@ -16,6 +16,7 @@ def call
           Services::Lti.create_lti_user_identity(user)
           handle_sections(rehydrated_user, user)
           user.lti_roster_sync_enabled = true if user.teacher?
+          user.lms_landing_opted_out = true
```
carl-codeorg, line 19: good catch adding this

dashboard/app/controllers/application_controller.rb
stephenliang: No, because when they click to link an account, the user is signed out and therefore there is no current user.

carl-codeorg: You're totally right, as long as we allow `finish_link` we're good!

apps/src/lib/ui/lti/link/LtiLinkAccountPage/cards/NewAccountCard/index.tsx
stephenliang: The await/async boils down the same stuff in the end and I think in this case it doesn't matter which we use. I went with Promise because I was going to chain, but ended up not needing to. If you feel strongly about await I can change it up, though it shouldn't matter much in this case.

carl-codeorg: Nah purely a conventional preference; I'm fine rolling with this since we're on the clock

dashboard/app/controllers/lti/v1/account_linking_controller.rb
```diff
@@ -42,6 +42,21 @@ def link_email
         end
       end
 
+      # POST /lti/v1/account_linking/new_account
+      def new_account
+        if current_user
+
```
carl-codeorg, line 48: accidental whitespace?

dashboard/app/controllers/application_controller.rb
```diff
@@ -370,6 +370,29 @@ def prevent_caching
     )
   end
 
+  # Check that the user has completed the LMS onboarding flow
+  protected def assert_lms_landing_policy
+    return unless Policies::Lti.lti?(current_user) && !current_user.lms_landing_opted_out
```
carl-codeorg, line 375: Users that already exist will have nil `lms_landing_opted_out` right? Do we need to check explicitly for false here, to avoid putting pre-existing users through the linking flow?

carl-codeorg: @alivira

carl-codeorg (approved):
:shipit:

PR #58949, opened 2024-05-28
remove CourseOffering category code references

davidsbailey:
This is a cleanup PR to remove the last references to the now-officially-dead category field on the CourseOffering object. A long time ago we used "category" to organize items in the AssignmentSelector. That component was just removed in https://github.com/code-dot-org/code-dot-org/pull/58922 . The new assignment experience uses marketing_initiative to categorize our course offerings instead.

## Testing story

* Updates to unit tests
* manually verified that marketing initiative is set everywhere it needs to be, as noted inline

## Follow-up work

* confirm that the category column is not needed by the data team (see [slack](https://codedotorg.slack.com/archives/C0SUN2W3D/p1718726696904969))
* remove the category column from the database table

PR #59369, opened 2024-06-24
updated calls to unit_in_category

mgc1194:
For some reason, after PR #58949 was merged, the i18n sync-in started failing with the error:
```
method_missing': undefined method unit_in_category?' for Unit:Class (NoMethodError)
````

The i18n scrip references the method `unit_in_category?` [from script_constants](https://github.com/code-dot-org/code-dot-org/blob/92ff38cfecd1638b70491f3c8c76b9bcf50a58da/dashboard/lib/script_constants.rb#L304), not the [method from the unit model](https://github.com/code-dot-org/code-dot-org/blob/92ff38cfecd1638b70491f3c8c76b9bcf50a58da/dashboard/app/models/unit.rb#L832).
Thus this error was unexpected.

This PR updates the call to `Unit.unit_in_category?` to  `unit.in_initiative?` unblock the i18n sync. 
(I have not figured out the root cause for this method overlap)

Update:
The class method unit_in_category? became an instance method `in_initiative?`.
We use that method to determine the directory in which i18n content is organized. The switch in methods made some Units change the directory we store their i18n content. 
I had to manually update crowdin and delete the previously synced-in files since we prevent duplicated files to be synced.

## PR Checklist:

- [ ] Code is well-commented

### CODE REVIEW

**Comment by wilkie** [comment]
Do we need to update: https://github.com/code-dot-org/code-dot-org/blob/7d740c866a26034cf736821ce3fcadc10cfa49c4/bin/test/i18n/resources/dashboard/curriculum_content/test_sync_in.rb#L55-L59

File: bin/i18n/resources/dashboard/course_content/sync_in.rb [hunk 1]
```diff
@@ -342,7 +342,7 @@ def process
               # We want to make sure to categorize HoC scripts as HoC scripts even if
               # they have a version year, so this ordering is important
               script_i18n_directory =
-                if Unit.unit_in_category?('hoc', script.name)
+                if ScriptConstants.unit_in_category?('hoc', script.name)
```
wilkie: I would write `:hoc` instead of `'hoc'`, which conforms to other places this function is called in our codebase: https://github.com/code-dot-org/code-dot-org/blob/7d740c866a26034cf736821ce3fcadc10cfa49c4/dashboard/lib/certificate_image.rb#L329-L333

And then you wouldn't need the change to `script_constants.rb`

artem-vavilov: Shouldn't we use `Unit.in_initiative?('HOC')` instead?
https://github.com/code-dot-org/code-dot-org/pull/58949/files#diff-debae58161437b537cd58e1c7ab8fc3ddda361cbbe871088a127b2592e4429d7L832-R834

davidsbailey (approved):
Thanks Artem! I agree, a root cause investigation could be quite time consuming, so it seems better to try this fix first.

Just curious if you have already verified that this fixes the problem?

bin/i18n/resources/dashboard/course_content/sync_in.rb
davidsbailey: I would love to get all uses of `unit_in_category?` moved over to `in_initiative?`. However I am not sure I understand the environment in which the i18n sync runs. Does it have all units seeded into its database? if so, the next thing to check is whether all units in the hoc category are also properly marked as in the HOC initiative. I can try that and get back to you with an update.

davidsbailey: mi = marketing initiative
sc = script constants
here are the discrepancies I see between the two data sources:
```
[development] dashboard > mi_offerings = CourseOffering.where(marketing_initiative: 'HOC').map(&:key).compact.sort

[development] dashboard > sc_names = ScriptConstants::CATEGORIES[:hoc].compact
[development] dashboard > sc_offerings = Unit.where(name: sc_names).map(&:get_course_version).map(&:course_offering).map(&:key).compact.sort

[development] dashboard > mi_offerings - sc_offerings
=> ["pixelation"]
[development] dashboard > sc_offerings - mi_offerings
=> ["ai-ethics", "hour-of-code", "how-ai-works"]
```

davidsbailey: If you want to depend on marketing initiative, you will be tied to the field that is used by Acquisition Products to control what shows up on the studio.code.org/catalog page. this could be good, if you don't want to miss new HOC tutorials as they are added. but it could be bad if you want more control. @artem-vavilov @mgc1194 let me know if you think you'd like to start using marketing initiative and I'll be happy to help. I would suggest a separate PR for this.

davidsbailey: sorry to go on but here is the difference with unit names instead of course offering names, since it looks like those might be what you use in the i18n code:
```
[development] dashboard > mi_names = CourseOffering.where(key: mi_offerings).all.map(&:course_versions).flatten.map(&:content_root).map(&:name).compact.sort

[development] dashboard > sc_names - mi_names
=> ["Hour of Code", "ai-ethics-2023", "generation-ai", "how-ai-works-2023"]
[development] dashboard > mi_names - sc_names
=> ["pixelation"]
```

mgc1194: We only need the category/initiative to determine the files' directory before uploading them to Crowdin. The possible folders are `Hour of Code`, `other`, or `year`, giving additional context to translators and helping prioritize some files over others. All translatable scripts are still being translated.
However, we prevent uploading duplicated files when they appear in multiple directories, therefore changing the initiative of a Unit will prevent from that unit to sync i18n content unless files are manually moved in the i18n directory.

mgc1194: It sounds like the teacher tools preference is using the `in_initiative?` method, so I will go ahead and use that method instead.

davidsbailey: > It sounds like the teacher tools preference is using the in_initiative? method, so I will go ahead and use that method instead.

yes, sounds great! the more calls to `unit_in_category?` we can eliminate from i18n code the better. This will get us one step closer to removing the big chunks of script_constants which are duplicate state from whats in course_offerings.marketing_initiative.

mgc1194: The only call left to `unit_in_category?` I could find is in [`certificate_image.rb`](https://github.com/code-dot-org/code-dot-org/blob/243954975c96d28e16e2e6704aa419ac6208c7ef/dashboard/lib/certificate_image.rb#L331)

There are no more calls to `unit_in_category?` in the i18n side of things.

davidsbailey: Sorry I misspoke -- there are other dependencies from i18n code on the following parts of ScriptConstants:
* https://github.com/code-dot-org/code-dot-org/blob/5259a94ee7b7c67a537cc8efa64b347cf2105689/dashboard/lib/script_constants.rb#L375-L380
* https://github.com/code-dot-org/code-dot-org/blob/5259a94ee7b7c67a537cc8efa64b347cf2105689/dashboard/lib/script_constants.rb#L310-L332

The fix for these would be different, possibly one of:
1. write special logic for translatability, e.g. "course is CSF, CSD or HOC and is in preview or published state", so that we no longer explicitly have to set when a course is ready to be translated
2. add a new flag to levelbuilder, figure out who (non-engineering) should mark each course / unit as translatable when it is time
3. do nothing, leave the code how it is

since teacher tools appears to make the updates to the list of which units are translatable, I will consult with the rest of the teacher tools team on which option would be best.

davidsbailey:
Just making sure this didn't get buried. For the following units...
```
[development] dashboard > sc_names - mi_names
=> ["Hour of Code", "ai-ethics-2023", "generation-ai", "how-ai-works-2023"]
[development] dashboard > mi_names - sc_names
=> ["pixelation"]
```
I would expect that you are now looking for these files in a different location, based on this check: https://github.com/code-dot-org/code-dot-org/blob/3909ebcc7f7a6cab55da96af825afae28f8740c9/bin/i18n/resources/dashboard/course_content/sync_in.rb#L342-L346

I don't have visibility into whether you've accounted for this, so just raising it again in case this was overlooked.

PR #59370, opened 2024-06-24
Script to remove section association to prepare for unit deletion.

vijayamanohararaj:
The script is to automate parts of the workflow to delete a unit. If a section has deleted units assigned to them, it breaks loading the teacher dashboard page. Hence, as part of the unit deletion process, we additionally remove any section associations for the unit.

This PR includes a script to remove the section association for a give set of unit(s).

## Links

https://codedotorg.atlassian.net/browse/TEACH-1161

## Testing story

Validating running the script locally for units with no sections and units with a section associated. The units were marked in_development before running the script, hence from end user perspective, they were already not showing up in the section dashboard. Upon running this script, there was same end user experience.

## Deployment strategy

Regular DTP

## Follow-up work

Follow up work is captured https://codedotorg.atlassian.net/browse/TEACH-1161

## Privacy

N/A

## Security

N/A

## Caching

N/A

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/bin/curriculum/delete_section_association.rb
+++ b/bin/curriculum/delete_section_association.rb
@@ -1,0 +1,58 @@
+#!/usr/bin/env ruby
+
+require 'json'
+require 'optparse'
+
+# Given a unit name, the script will iterate over all the sections that
+# have the unit assigned and remove that association by setting assigned
+# unit for those sections to null.
+
+def parse_options
+  OpenStruct.new.tap do |options|
+    options.unit_names = nil
+    options.dry_run = false
+
+    opt_parser = OptionParser.new do |opts|
+      opts.banner = "Usage: #{$0} [options]"
+
+      opts.separator ""
+
+      opts.on('-u', '--unit_names UnitName1,UnitName2', Array, 'Unit names to remove all section associations') do |unit_names|
+        options.unit_names = unit_names.map(&:strip)
+      end
+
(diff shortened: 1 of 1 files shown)


bin/curriculum/delete_section_association.rb
```diff
@@ -0,0 +1,58 @@
...
+  puts "Number of sections that have unit #{unit_nm} associated: #{assigned_sections.count}."
+
+  if options.dry_run
+    puts "Skipping section update due to dry run mode."
+  elsif assigned_sections.count > 0
+    puts "Removing all section assignments."
+    assigned_sections.each do |section|
+      section.script_id = nil
+      section.save!
+    end
```
davidsbailey: this should be ok for small number of sections. if the script is taking a long time to run or using a lot of memory you could try:
```suggestion
    assigned_sections.update_all(script_id: nil)
```
if you try it in the rails console you should be able to see this issuing a single SQL query instead of N. 

https://api.rubyonrails.org/v3.2.3/classes/ActiveRecord/Relation.html#method-i-update_all

davidsbailey (approved):

bin/curriculum/delete_section_association.rb
vijayamanohararaj: Thanks for the suggestion!

PR #59372, opened 2024-06-24
Add aria labels to carousel buttons on Hour of Code /beyond page

TurnerRiley:
Adds aria labels to "Go" buttons in carousels on [hourofcode.com/beyond](https://hourofcode.com/us/beyond). Currently, each card in the carousels just has an external link button that says "Go" without additional info, so this PR provides that extra context.

ex_aria

## Links
Jira ticket: [here](https://codedotorg.atlassian.net/jira/software/c/projects/ACQ/boards/64?assignee=60d62161f65054006980bd71&selectedIssue=ACQ-1893)
Spreadsheet with aria labels: [here](https://docs.google.com/spreadsheets/d/1RZ6iutiOV8LTol1uooJ3m1GiHHPbpmssNSWkT87oTNI/edit?gid=822908068#gid=822908068)

## Testing story
Local testing.

### DIFF
--- a/pegasus/sites.v3/hourofcode.com/views/course_carousel.haml
+++ b/pegasus/sites.v3/hourofcode.com/views/course_carousel.haml
@@ -29,7 +29,7 @@
                   = tutorial[:platformtext]
                 %p.slide-longdescription= tutorial[:longdescription]
               .content-footer
-                %a.link-button{href: url, target: "_blank", aria: {label: tutorial[:buttonarialabel]}}
+                %a.link-button{href: url, target: "_blank", aria: {label: hoc_s(:module_label_visit_site, markdown: :inline, locals: {site_name: tutorial[:name]})}}
                   =I18n.t(:go)
 
     %button.swiper-nav-prev{class: id}
--- a/pegasus/sites.v3/hourofcode.com/views/course_carousel.haml
+++ b/pegasus/sites.v3/hourofcode.com/views/course_carousel.haml
@@ -29,7 +29,7 @@
                   = tutorial[:platformtext]
                 %p.slide-longdescription= tutorial[:longdescription]
               .content-footer
-                %a.link-button{href: url, target: "_blank"}
+                %a.link-button{href: url, target: "_blank", aria: {label: tutorial[:buttonarialabel]}}
                   =I18n.t(:go)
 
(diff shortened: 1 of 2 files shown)


kelbyhawn (approved):
I read the gsheet labels and this looks great! Did you try the following commands to try to seed locally?
`cd pegasus`
`bundle exec rake seed:sync`

TurnerRiley:
> I read the gsheet labels and this looks great! Did you try the following commands to try to seed locally? `cd pegasus` `bundle exec rake seed:sync`

That's the command that's not working for me locally unfortunately :/

bethanyaconnor:
I don't see this column in either the staging or production databases. Is there another step you need to do, maybe a migration?

pegasus/sites.v3/hourofcode.com/views/course_carousel.haml
```diff
@@ -29,7 +29,7 @@
                   = tutorial[:platformtext]
                 %p.slide-longdescription= tutorial[:longdescription]
               .content-footer
-                %a.link-button{href: url, target: "_blank"}
+                %a.link-button{href: url, target: "_blank", aria: {label: tutorial[:buttonarialabel]}}
```
bethanyaconnor: It's not as great of a solution as the one you've got, but would it be possible to do something like `aria: {label: "Go to #{tutorial[:name]"}` here instead to save the database migration?

kelbyhawn: Oh that's a good idea…

TurnerRiley: I considered that at first but was concerned about some of the overlapping names (particularly with Code.org's names). If this is okay to have a bit of overlap I can for sure do that and would be able to test it!

TurnerRiley: Actually looking more into it I think I may have jumped the gun with the concern of overalpping names because the overlapping ones would appear in different sections (of elementary, middle, and high schools) so I'll switch to that plan!

kelbyhawn (approved):
🎉

bethanyaconnor (commented):
Yay thank you for updating!

PR #59384, opened 2024-06-25
Add Open Graph title to certificates show page

bethanyaconnor:
Tentatively resolves https://codedotorg.atlassian.net/browse/ACQ-2002. The LinkedIn share doesn't work with localhost so I couldn't actually test that bit, but I did confirm the `og:title` tag is set correctly.

Screenshot 2024-06-25 at 8 38 36 AM

I'll test the LinkedIn share when this hits staging or production. I just didn't want to spin up an adhoc for such a small change. 😅


kelbyhawn (approved):
Thanks for updating this! 🎉

PR #59394, opened 2024-06-25
warn when using "wait until I am on" steps in Safari

davidsbailey:
discussion of https://github.com/code-dot-org/code-dot-org/pull/59358 revealed a problem with "wait until I am on" steps in Safari, that could be having widespread effect on flaky UI tests. I didn't investigate the root cause, but instead I'm adding a breadcrumb for anyone who runs into this while investigating flaky UI tests.

## screenshots

warning is visible in cucumber log output in Safari:
Screenshot 2024-06-25 at 12 38 42 PM


hannahbergam (approved):
LGTM! Thanks for the add!

PR #59396, opened 2024-06-25
Lab2/Python Lab: Support getting and saving predict responses

molly-moen:
Follow up to #59346. Now that we have predict settings saved to the level, we can get and save user's predict responses to level sources. This is done as follows:
- We save level sources the same way we always have, via the milestone api, with a `program` field added ([old code](https://github.com/code-dot-org/code-dot-org/blob/ac09e2b2e293e322d3ab7c5aead9681aafdd5f4e/apps/src/code-studio/initApp/loadApp.js#L100-L111) from `loadApp.js`).
- I created a new api, `/user_levels/level_sources` that gets the user's most recent level source data for a given script and level. Previously we always got this information on page load on the server, so there was no api that just gave us the level source for a user, script and level.

To plug everything together, this PR also includes the basic initial implementation of using these apis. If a lab2 level is free response predict level, we now show a text input field in the instructions panel. When the user clicks run (on Python Lab only), we save the predict response via the milestone api. I also made predict levels read only (if we are not in start or exemplar edit mode). For now you can submit multiple answers, configuring that logic is a follow-up.

This PR is best viewed ignoring whitespace changes.

### Screenshot

Screenshot 2024-06-26 at 9 50 49 AM

## Links

- jira ticket: [CT-635](https://codedotorg.atlassian.net/browse/CT-635)

## Testing story
Tested locally that we can get and save predict responses (free response only for now), and other lab2 labs are unaffected. I also added unit tests for the new controller action.

## Follow-up work
I will follow this up with a full implementation of free response and multiple choice questions, including all the input fields. There is also follow-up work to do to disable the run button until the user has answered the question, and to lock the answer if the level is configured to do so. The tasks for this work are enumerated in the spec linked above.

## PR Checklist:

- [ ] Code is well-commented


apps/src/codebridge/types.ts
```diff
@@ -48,7 +48,7 @@ export type SetProjectFunction = (project: ProjectType) => void;
 export type SetConfigFunction = (project: ConfigType) => void;
 export type ResetProjectFunction = () => void;
 export type OnRunFunction = (
-  runTexts: boolean,
+  runTests: boolean,
```
molly-moen, line 51: Just fixing a typo

apps/src/pythonlab/PythonlabView.tsx
```diff
@@ -74,6 +76,21 @@ const defaultConfig: ConfigType = {
 const PythonlabView: React.FunctionComponent = () => {
   const [config, setConfig] = useState<ConfigType>(defaultConfig);
   const {source, setSource, resetToStartSource} = useSource(defaultProject);
+  const isPredictLevel = useAppSelector(
+    state => state.lab.levelProperties?.predictSettings?.isPredictLevel
+  );
+
+  const onRun = (
```
sanchitmalhotra126, line 84: General question (maybe for curriculum/product) - do we only want to submit the predict response when the user actually runs the level? Or would it be better to have a separate "submit" button as part of the UI in the instructions panel? Just wondering if the current behavior is by design or just due to limitations at the time

apps/src/code-studio/progressRedux.ts
```diff
@@ -364,46 +367,70 @@ export function navigateToNextLevel(): ProgressThunkAction {
...
+    sendReportHelper(appType, TestResults.ALL_PASS, dispatch, getState);
+  };
+}
 
-    // The server does not appear to use the user ID parameter,
-    // so just pass 0, like some other milestone posts do.
-    const userId = 0;
+export function sendPredictLevelReport(appType: string): ProgressThunkAction {
+  return (dispatch, getState) => {
+    const predictResponse = getState().lab.predictResponse;
```
sanchitmalhotra126: minor: would it be cleaner to have the caller pass in the predictResponse string rather than fetching from lab state? I know we access to the full redux store here but wondering if there's value in keeping these two slices as decoupled as possible

apps/src/lab2/lab2Redux.ts
```diff
@@ -65,6 +66,9 @@ export interface LabState {
   levelProperties: LevelProperties | undefined;
   // If this lab should presented in a "share" or "play-only" view, which may hide certain UI elements.
   isShareView: boolean | undefined;
+  // User's response for the level, if the level is a predict level. It is an empty string if this
+  // is not a predict level or if the user has not yet submitted a response.
+  predictResponse: string;
```
sanchitmalhotra126, line 71: Future consideration - wonder if we should start thinking about breaking up lab2Redux as it continues to grow (like separating level-specific/non-user specific fields from user-specific, e.g. lab2LevelRedux, lab2ProjectRedux, lab2UserLevelRedux (?)), similar to how codebridge already has lab2Project. Definitely not necessary now since there aren't too many fields, but just a thought

sanchitmalhotra126 (approved):
Looks good!

apps/src/pythonlab/PythonlabView.tsx
molly-moen: Good question, I'll check with them but leave as-is for now

apps/src/code-studio/progressRedux.ts
molly-moen: Yeah I debated this. My motivation was to make it as easy as possible to re-use this method across different lab2 labs without having to know where it's stored. But I agree it's a little messy this way, I'll update it to pass in the value.

apps/src/lab2/lab2Redux.ts
molly-moen: Yeah I also thought about this, but since it was only one field it didn't make much sense to split this out. But I think it makes sense to refactor our redux store if it gets much bigger.

PR #59400, opened 2024-06-25
Fixes flakiness of 'I join the section'.

wilkie:
`Given I am on` in conjunction with `to load a new page` is somewhat squirmy. It has a small chance to confuse page loads that happen afterward, mostly in Safari it seems.

Certainly affects (these join a bunch of sections increasing the chance of failure):

* `teacher_tools_teacher_dashboard_teacher_dashboard_assessments2`
* `teacher_tools_pairing`

Might affect (These all use the rule to join a section sparingly):

* `teacher_tools_teacher_dashboard_teacher_dashboard_code_review_groups`
* `teacher_tools_cached_level_page`
* `teacher_tools_assignment_on_overview_pages`
* `star_labs_gamelab_libraries`
* `star_labs_applab_libraries`

Take this recent safari failure on `teacher_tools_teacher_dashboard_teacher_dashboard_assessments2`:

image
image

On [sauce labs](https://app.saucelabs.com/tests/f3a703a22b3d4cd1af44b670dc920350#345) we see this progression:

image
image
image

This is a bit of a tricky one, but I'm starting to see patterns in the Matrix when looking at these flaky tests now. Do not be fooled by all the green. We are racing from WAAAAAY before.

This one centers on the pending join to a section somehow racing the normal flow and just suddenly... uh... interrupting??? I think Safari is particularly keen on not interrupting / cancelling async `POST` requests via JS that we tend to do for our UI test steps... they just keep happening in the background if we are not careful.

Specifically, here, the steps above (what the heck) show us that the joining of a section was not complete before it went ahead to the level itself. It happened to join exactly when it was trying to also press the complete button ... so it then goes and mistakes the interrupted page load as the one it is looking for because it was momentarily 'complete' on a different `:root` for a hot second.

Ok. Why?

This centers around the `I join the section` metarule:

```
And(/^I join the section$/) do
  page_load(true) do
    steps <<~GHERKIN
      Given I am on "#{@section_url}"
      And I click selector ".btn.btn-primary" once I see it
    GHERKIN
  end
end
```

Now, look at `page_load`:

```
def page_load(wait = true, blank_tab: false)
  if wait
    root = @browser.find_element(css: ':root')
    ...
    yield
    wait_until {element_stale?(root)}
    ...
  end
end
```

Oops! If I change the page _during_ the `page_load` block, I will obviously have an _immediately_ stale `:root` if I expect multiple actions to load a new page.

So `I join a section` is pretty pretty borked! Let's fix it.

## Solution

The fix is to make the `page_load` responsibilities atomic. We wrap only the first navigation in the `page_load` block and then follow it with the `page_load` version of the click rule. This means it waits for each page load instead of waiting for _any_ page load.

## Links

- [Slack Thread](https://codedotorg.slack.com/archives/C0T0PNTM3/p1719349518319439?thread_ts=1719344611.541919&cid=C0T0PNTM3)

## Testing story

The existing UI tests still pass via Drone CI: (Safari is the problem but that is difficult to test locally... I'm just gonna hope!!)

```
Chrome_teacher_tools_teacher_dashboard_teacher_dashboard_assessments2 succeeded (2:18 minutes, 1 passed)
```

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/dashboard/test/ui/features/step_definitions/section_management_steps.rb
+++ b/dashboard/test/ui/features/step_definitions/section_management_steps.rb
@@ -153,9 +153,11 @@ And(/^I join the section$/) do
   page_load(true) do
     steps <<~GHERKIN
       Given I am on "#{@section_url}"
-      And I click selector ".btn.btn-primary" once I see it
     GHERKIN
   end
+  steps <<~GHERKIN
+    And I click selector ".btn.btn-primary" once I see it to load a new page
+  GHERKIN
 end
 
 And(/^I attempt to join the section$/) do


lfryemason (approved):

wilkie:
Full suite did not pass, but the relevant ones did:

```
Safari_teacher_tools_teacher_dashboard_teacher_dashboard_assessments2 succeeded (1:57 minutes, 1 passed)
```

The UI tests failed with another flaky test: `Safari_teacher_tools_rubrics_teacher_view_of_rubric` which does not use this rule.

PR #59403, opened 2024-06-25
AI Chat: disable update/save/publish buttons until changes have been made

sanchitmalhotra126:
This just checks if the existing `findChangedProperties` method returns any items, and if so, enables the update buttons.

https://github.com/code-dot-org/code-dot-org/assets/85528507/99773b4e-1569-4f73-b436-1928ee173285

## Links

https://codedotorg.atlassian.net/browse/LABS-744

## Testing story

Tested locally on a Gen AI pilot level and allthethings.

### DIFF
--- a/apps/src/aichat/views/modelCustomization/PublishNotes.tsx
+++ b/apps/src/aichat/views/modelCustomization/PublishNotes.tsx
@@ -7,6 +7,7 @@ import {
   saveModelCard,
   publishModel,
   selectHasFilledOutModelCard,
+  selectHavePropertiesChanged,
 } from '@cdo/apps/aichat/redux/aichatRedux';
 import Button from '@cdo/apps/componentLibrary/button/Button';
 import {FontAwesomeV6IconProps} from '@cdo/apps/componentLibrary/fontAwesomeV6Icon';
@@ -38,6 +39,7 @@ const PublishNotes: React.FunctionComponent = () => {
   const isReadOnly = useSelector(isReadOnlyWorkspace) || isDisabled(visibility);
   const saveInProgress = useAppSelector(state => state.aichat.saveInProgress);
   const currentSaveType = useAppSelector(state => state.aichat.currentSaveType);
+  const havePropertiesChanged = useAppSelector(selectHavePropertiesChanged);
 
   const onSave = useCallback(() => {
     dispatch(saveModelCard());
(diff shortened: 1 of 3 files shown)


bencodeorg (approved):
Deliciously simple!

apps/src/aichat/views/modelCustomization/PublishNotes.tsx
```diff
@@ -120,7 +122,12 @@ const PublishNotes: React.FunctionComponent = () => {
               ? spinnerIconProps
               : {iconName: 'upload'}
           }
-          disabled={isReadOnly || !hasFilledOutModelCard || saveInProgress}
+          disabled={
```
fisher-alice, line 125: There's an edge case when a user fills out an unpublished model card completely and clicks on 'save'. The 'publish' button is also disabled which may be confusing. Could add to `(!havePropertiesChanged && modelCardInfo.isPublished)` to condition.

fisher-alice: This is probably out of scope of this PR, but when the 'Publish' button is disabled, the 'Ready to publish' notification is still displayed at the top of the panel which may also be a little confusing.
We could remove this notification when the 'Publish' button is disabled - maybe as a follow-up? Currently, we either display the `PublishOkNotification` or the `CompleteToPublishNotification`.

fisher-alice (approved):
Nice! Just added a minor suggestion for an edge case.

apps/src/aichat/views/modelCustomization/PublishNotes.tsx
sanchitmalhotra126: Oh good call! Yep let me take another look at this and can file a follow-up if needed.

sanchitmalhotra126: Hmm yeah this is a little confusing - I added your first condition which makes sense for the first publish, but if the user then changes some information and saves, the publish button is disabled. I think it might make sense to always have the Publish button enabled (if all fields are filled out)? Even if there haven't been any updates, we'll skip the save, but we'll still change to the presentation view.

I might merge this is as-is and do a fast follow to implement this behavior since it'll involve some other redux-related changes.

Edit: will clarify the correct behavior with curriculum/product on slack

PR #59404, opened 2024-06-25
Adding Amplitude events for CAP parent consent

daynew:
In Amplitude, we have CPA funnel graphs for when students see a metric and then what actions they take. Getting parent permission is the final step in the CPA process so we need to have an event for that.

One of the challenges is that we want to log the student's UserId because the parent is the one actually clicking a link and granting permission. That is why the UserId associated with the PermissionRequest is logged.

* Converts JSX files related to CPA parental consent to TSX files.
* Logs either a PermissionGranted or PermissionExpired event on the page parent's visit to grant permission for their child.

## Links
[Jira](https://codedotorg.atlassian.net/browse/P20-953)

## Testing story
* Unit tests
* Manually tested on localhost and verified the Amplitude events were there.

## Follow-up work
* Figure out how to use the `studentId` in Amplitude.

## PR Checklist:

- [ ] Code is well-commented


wilkie (approved):
Looks great! Thank you for taking the time to convert to TypeScript, too. Wonderful! 👍

PR #55353, opened 2023-12-11
Update google blockly callout script for Sprite Lab

molly-moen:
A couple minor updates to the script added in #54642.

- Run the script for all sprite lab levels (GamelabJr levels with game id 64, because other level types are children of Game Lab Jr)
- Allow for non-existent long instructions
- Update the block count math when a block has a categorized toolbox. In that case, the block count starts with any instructions blocks, then goes to the main workspace blocks, then goes to the toolbox blocks (otherwise it's toolbox blocks, instructions, main workspace).
- Improved logging while running the script.

## Links
- jira ticket: [CT-151](https://codedotorg.atlassian.net/browse/CT-151)

## Testing story
Tested by running locally. The script updates 321 levels, so it was not feasible to manually check each one. I did a visual pass over the updates and spot checked many levels by comparing the cdo and google blockly versions to ensure the callouts were in the same location.

There are 16 levels that could not be updated by the script due to blocks already having ids or invalid ids. I will follow up with curriculum to see if we need to manually update those files.

## Deployment strategy

I will follow the same strategy I used before to run this script. Once it is on levelbuilder I will:
- Find a time when it is ok to do an out of band scoop (last time I did 4 pm pacific, verified with content editors)
- Do a manual levelbuilder scoop (`L`)
- Run the script on levelbuilder
- Do another manual scoop (this should only contain the files updated by the script)
- Run `DTS` to deploy all the changes

## PR Checklist:

- [ ] Tests provide adequate coverage
- [ ] Privacy and Security impacts have been assessed
- [ ] Code is well-commented
- [ ] New features are translatable or updates will not break translations
- [ ] Relevant documentation has been added or updated
- [ ] User impact is well-understood and desirable
- [ ] Pull Request is labeled appropriately
- [ ] Follow-up work items (including potential tech debt) are tracked and linked

PR #54642, opened 2023-11-01
Script to update callouts for Google Blockly labs

molly-moen:
This script updates callouts on Dance and Poetry levels to work with Google Blockly (Bounce and Flappy had no callouts to update). Callouts are written to point to specific css selectors, and those selectors changed between CDO and Google Blockly. This script covers callouts that point to toolbox blocks, workspace blocks and categories in the toolbox. The script creates a duplicate version of the old callout, with the new element id. It's fine to create a duplicate because we won't see both ids at the same time, we are either running cdo or google blockly for a given page view. 

The script also adds ids to the relevant blocks in the toolbox/main workspace. CDO Blockly automatically created those ids, but Google Blockly does not. The ids in CDO Blockly are created via a 1-based indexing system, starting with toolbox blocks, then going to any blocks in the instructions, then ending with workspace blocks. I have verified that the new ids generated by the script are equivalent to the old ids. If there were any previous ids on the blocks we keep those, and do not add new ids if the id already existed on another block. We also fail on any level where the block we want to add an id to already has an id (that does not match what we expect). We are doing this because we discovered some levels where the old ids were not on the blocks we expected, but those blocks were the ones that the callouts should have been on.

I will run the script on the levelbuilder environment once it has been merged and deployed to levelbuilder. When I run it locally it changes the levels `created_at` field, which is not a deal breaker, but running it on levelbuilder will avoid this and will also avoid any potential merge conflicts with the level changes. The script changes 94 levels. It should not be run until [this change](https://github.com/code-dot-org/code-dot-org/pull/54637) is in, which fixes callouts on Google Blockly levels.

## Links

- jira ticket: [https://codedotorg.atlassian.net/browse/CT-56](https://codedotorg.atlassian.net/browse/CT-56)
- [slack thread](https://codedotorg.slack.com/archives/C05DK21DAHK/p1698960891300809) with curriculum about inconsistent ids

## Testing story
I have run this script locally and verified that the callouts are changed correctly for each level that is touched.

## Follow-up work
Once we run this on levelbuilder for dance and poetry we can update the script to run on sprite lab levels and run it against those as well. Dance and Poetry are currently broken, whereas Sprite Lab is not broken until we switch over to Google Blockly.
Tasks for running the script:
- [dance/poetry](https://codedotorg.atlassian.net/browse/CT-150)
- [sprite lab](https://codedotorg.atlassian.net/browse/CT-151)

## PR Checklist:

- [ ] Tests provide adequate coverage
[...]

PR #59412, opened 2024-06-26
Update google blockly callout script for remaining Blockly labs

mikeharv:
This updates our script to fix callouts that are broken for Google Blockly. This was last used to update Sprite Lab levels: https://github.com/code-dot-org/code-dot-org/pull/55353 and was originally used for Dance/Poetry: https://github.com/code-dot-org/code-dot-org/pull/54642 We didn't use the script for Minecraft because there were so few levels to update that it was easier to do it without the script.

The following changes are made:

* We will process Blockly levels of all remaining record types.
* We will attempt to match callout element ids that don't specify a class, e.g. `[block-id="1"]`.
* Any id including "block-id" will be copied to the new callout as "data-id"
* Any of these elements that specifically target a block's edit button (ie. function definitions) will be updated.
* If we don't have a class to determine if the block is on the workspace or flyout, we will iterate through all blocks to find it.

Updating the callouts is a necessary step for any Blockly lab migration. We are anticipating Maze and Star Wars soon, so those are the most urgent. However, as the curriculum team has no current plans to update/create Artist and Play Lab levels at this time, it feels safe (and more efficient) to take care of those here as well.

When the script ran previously, we looked for outdated Blockly classes `.svgFlyoutGroup` and `.svgGroup` which were updated to `.blocklyFlyout` and `.blocklySvg`. Interestingly, this convention of writing callout ids is only found in labs that we've already migrated (Dance, Sprite Lab, Poetry). In all older labs, we just targeted the element by block-id alone.

## Testing story

While there are too many levels changed to manually test them all, I did spot check many, paying special attention to the most complex selectors (edit buttons). There is one `Blockly` level which fails to update and is not changed. This level is actually an App Lab level, so I'm ignoring the failure.
Screenshots below show callouts working on both CDO and Google Blockly and a comparison of the element id field for the callouts as shown on the level edit page.

**Categories**
imageimage
image

**Block with id only**
image
image
image

**Function Edit button on block with id**
image
image
image
## Deployment strategy

I will follow the same strategy @molly-moen used before to run this script. Once it is on levelbuilder I will:

- Find a time when it is ok to do an out of band scoop (likely early in the morning eastern time)
- Do a manual levelbuilder scoop (L)
- Run the script on levelbuilder as a dry run to verify the expected changes
- Edit the script file on the levelbuilder machine to set `DRY_RUN` to false
- Run the script on levelbuilder
- Do another manual scoop (this should only contain the files updated by the script)
- Run DTS to deploy all the changes

## Follow-up work

I will follow-up with the curriculum to team for any levels that cannot be updated due to blocks already having ids or invalid ids.

## PR Checklist:

- [ ] Code is well-commented


bin/oneoff/fix_broken_google_blockly_callouts.rb
```diff
@@ -108,21 +132,36 @@ def add_new_callouts(ids_to_fix, broken_callouts, callouts)
...
+def new_id_for_block(new_css_class, id, element_id)
+  if new_css_class
+    ".#{new_css_class} [data-id=\"#{id}\"]"
+  else
+    # For the general_block_id_matcher, we simply replace block-id with data-id.
+    # We copy the old element so we don't lose other trailing selector information.
+    modified_id = element_id.dup
+    modified_id.sub!('block-id', 'data-id')
+    # Special case to modify selection of function call block's edit button
+    modified_id.sub!('.blocklyIconGroup', 'g>.blocklyFieldRect')
```
mikeharv, line 162: In CDO Blockly, the edit button could be identified by a child svg on the block with the class `.blocklyIconGroup`.
image

With Google Blockly, that class isn't used. The most accurate alternative class to identify the button is `.blocklyFieldRect`, which is found on the `rect` grandchild:
image

While this technically changes which element we are pointing to, there's no visual difference because the two elements have the same size and position.

bin/oneoff/fix_broken_google_blockly_callouts.rb
```diff
@@ -16,8 +16,20 @@ def fix_callouts
   else
     puts "THIS IS A FULL RUN"
   end
-  puts "***SPRITE LAB***"
-  process_levels(GamelabJr.where(game_id: 64))
+  puts "***PROCESSING ALL BLOCKLY LEVELS***"
```
molly-moen, line 19: what are "blockly" levels?

molly-moen (approved):
One question for my information, otherwise looks good!

bin/oneoff/fix_broken_google_blockly_callouts.rb
mikeharv: These are engineer-owned levels that cannot be edited on levelbuilder. You can see them listed here: https://levelbuilder-studio.code.org/levels?name=&level_type=Blockly&script_id=&owner_id=&commit=Search

They include:
* All Star Wars "game" levels
* All Jigsaw levels
* The oldest Maze/Artist/Play Lab levels

Our other level model classes extend the Blockly class, but the records in the database are still 1:1. Processing all Blockly levels doesn't handle sub-types, nor does Maze handle Karel or Studio handle StarWarsGrid. 

We don't use Blockly levels for Dance/Sprite Lab/Poetry/Minecraft, which is why we haven't had to worry about them yet.

molly-moen: thanks!

PR #59414, opened 2024-06-26
Filter system prompt/retrievals for profanity in aichat

fisher-alice:
This PR adds WebPurify profanity filtering for system prompts and retrieval contexts in `aichat` when a user updates the model customizations.

There were a couple options I was considering:
- Add an API call to check the system prompt/retrieval contexts [within `saveAiCustomization`](https://github.com/code-dot-org/code-dot-org/blob/staging/apps/src/aichat/redux/aichatRedux.ts#L173) before saving the project source file (`main.json`).
- Adding a check for `aichat` system prompt/retrieval contexts when the project source file is saved.

I went with the second option since it will be more simple to add other labs to the list of labs which opt in to having source files be filtered.

In `files_api.rb`, I updated the `put_file` method so that it takes an optional `project_type` parameter. If the project type is in the `LABS_TO_CHECK_FOR_PROFANITY` list, then the source will be filtered via WebPurify.
For `aichat` specifically, we want to check the system prompt and retrieval contexts for profanity only (PII filtering may be added later). Model card info is not included in the api call to Sagemaker.

If the saving error is due to profanity, then the notification message includes this information. Otherwise, a generic save error notification is displayed. Note that the current AI customizations are not modified if there was a saving error because of the risk of false positives. The notification will include whether profanity was found in the system prompt, retrieval contexts, or both.

I added a DCDO flag that can remove `aichat` from the list of labs that are checked for profanity to address @dancodedotorg's [concern](https://docs.google.com/document/d/1cYS7fa6q_fmfWrwI_fANSgSFC7mUjbQyhrYgM_mKNjA/edit?disco=AAABQQbzHOk) about being able to test existing prompts/retrievals in current lessons and make sure these are not flagged for profanity.

## After update
Screencast video showing error notification message due to profanity in system prompt, retrieval contexts and then both.

https://github.com/code-dot-org/code-dot-org/assets/107423305/41c0a008-bd14-4701-be22-1ebf57979c30

Screenshot of save error notification message if not due to profanity (hard coded):
Screenshot 2024-06-26 at 12 51 42 PM

## Links
[jira](https://codedotorg.atlassian.net/browse/LABS-791)

## Testing story
I tested locally in a handful of /allthethings and /gen-ai-customizing-pilot-v1 `aichat` levels with system prompts/retrieval contexts that included and excluded profane words. I also confirmed that PII (emails, phone numbers) are not yet filtered in `aichat`.
I hard-coded a 503 server error and confirmed that a general saving error notification message was displayed.
Screenshot 2024-06-28 at 4 20 31 PM

I also tested an App Lab project locally and still received an error when profanity was included in a library when attempting to publish.

## PR Checklist:

- [ ] Code is well-commented


apps/src/aichat/redux/aichatRedux.ts
```diff
@@ -237,18 +237,27 @@ export const onSaveComplete =
...
-  );
-  // Notify the UI that the save is complete.
-  dispatch(endSave());
-};
+export const onSaveFail =
+  (e: Error) => (dispatch: AppDispatch, getState: () => RootState) => {
+    const {savedAiCustomizations} = getState().aichat;
+    const isProfanityError = e.message.includes('profanity');
+    if (isProfanityError) {
+      dispatch(setAllAiCustomizations(savedAiCustomizations));
```
sanchitmalhotra126: Should we be force-resetting a user's project if profanity is detected? I could imagine that if a user has done a bunch of work without saving and then tries to save, but WebPurify incorrectly detects profanity, then they might unintentionally lose a lot of work. Would it be better to just leave the project as is?

apps/src/aichat/redux/aichatRedux.ts
```diff
@@ -237,18 +237,27 @@ export const onSaveComplete =
...
+    }
+    const errorMessage = isProfanityError
+      ? 'Profanity detected in system prompt or retrieval context(s) and cannot be updated. Please try again.'
```
sanchitmalhotra126: I'm sure we'll take another pass at these error strings before launch, but wondering if it makes sense to not specify which fields? Technically this could include model card fields as well right? (Edit: realized you're specifically checking system prompt and retrieval in the put_file call, but also wondering if it's worth just checking everything)

apps/src/lab2/projects/SourcesStore.ts
```diff
@@ -1,16 +1,20 @@
...
 const {getTabId} = require('@cdo/apps/utils');
 
 export interface SourcesStore {
   load: (key: string) => Promise<ProjectSources>;
 
-  save: (key: string, sources: ProjectSources) => Promise<Response>;
+  save: (
+    key: string,
+    sources: ProjectSources,
+    appType?: ProjectType
```
sanchitmalhotra126, line 17: nit: should the arg be `projectType` for consistency?

Also generally might be good to call this property `projectType`/`project_type` throughout, for consistency (I know these terms are super overloaded, but I see "project_type" referenced a few times in `files_api.rb` as well)

dashboard/legacy/middleware/files_api.rb
```diff
@@ -413,13 +415,21 @@ def put_file(endpoint, encrypted_channel_id, filename, body)
...
     # the files are only being used by a single user.
-    if endpoint == 'libraries' && file_type != '.java'
+    if (endpoint == 'libraries' && file_type != '.java') || LABS_TO_CHECK_FOR_PROFANITY.include?(app_type)
       begin
-        share_failure = ShareFiltering.find_failure(body, request.locale)
+        if app_type == 'aichat'
+          source = JSON.parse(body)['source']
+          source_json = JSON.parse(source)
+          text_to_check = source_json['systemPrompt'] + ' ' + source_json['retrievalContexts'].join(' ')
+          share_failure = ShareFiltering.find_failure(text_to_check, request.locale)
```
sanchitmalhotra126: Would it be simpler to just check the entire `body`? That way we check the model card fields too? (Edit: or maybe not - would be curious how curriculum/product feel)

dashboard/legacy/middleware/files_api.rb
```diff
@@ -428,11 +438,17 @@ def put_file(endpoint, encrypted_channel_id, filename, body)
...
       # Once we have a better geocoding solution in H1, we should start filtering for addresses again.
       # Additional context: https://codedotorg.atlassian.net/browse/STAR-1361
-      if share_failure && share_failure[:type] != "address"
+      if app_type != 'aichat' && share_failure && share_failure[:type] != "address"
         details_key = share_failure.type == ShareFiltering::FailureType::PROFANITY ? "profaneWords" : "pIIWords"
         details = {details_key => [share_failure.content]}
         return json_bad_request(details)
       end
+      # For now, just check for profanity in aichat.
+      if app_type == 'aichat' && share_failure && share_failure[:type] == ShareFiltering::FailureType::PROFANITY
```
sanchitmalhotra126: Any harm in also checking PII? Might be something we need to clarify with curriculum/product, but since I imagine we'd also eventually be checking for PII in `aichat_controller` would it make sense to do it here as well (and not have to have this separate branch?)

Alternatively, if we want to check profanity only (that might be useful for other labs too), would be good to check the whole list rather than aichat specifically. We could also move the conditional up into the previous block, maybe something like

```
if share_failure && share_failure[:type] != "address"
  details_key = share_failure.type == ShareFiltering::FailureType::PROFANITY ? "profaneWords" : "pIIWords"
  details = {details_key => [share_failure.content]}
  if !project_type || (check_for_profanity?(project_type) && details_key == "profaneWords")
    return json_bad_request(details)
  end
end
```

dashboard/legacy/middleware/files_api.rb
```diff
@@ -413,13 +415,21 @@ def put_file(endpoint, encrypted_channel_id, filename, body)
     end
 
     # Block libraries with PII/profanity from being published.
+    # Block main.json file from aichat lab flagged with profanity from being saved.
     #
     # Javalab's "backpack" feature uses libraries to allow students to share code
     # between their own projects -- skip this check for .java files, since in this use case
     # the files are only being used by a single user.
-    if endpoint == 'libraries' && file_type != '.java'
+    if (endpoint == 'libraries' && file_type != '.java') || LABS_TO_CHECK_FOR_PROFANITY.include?(app_type)
```
sanchitmalhotra126: Just for consistency, might be good to have a function that checks the project type (similar to `moderate_type?` at the bottom of this file): https://github.com/code-dot-org/code-dot-org/blob/d971cd2ff302dc6088c253cd071ae3911e3b83ed/dashboard/legacy/middleware/files_api.rb#L1076-L1078

apps/src/lab2/projects/SourcesStore.ts
```diff
@@ -59,7 +69,13 @@ export class RemoteSourcesStore implements SourcesStore {
       this.firstSaveTime = this.firstSaveTime || timestamp;
       this.currentVersionId = versionId;
     } else {
-      throw new Error(response.status + ' ' + response.statusText);
+      const reader = await response.json();
+      const details = reader.details;
+      const statusText =
+        details && details.profaneWords
+          ? `${response.statusText} (profanity)`
```
sanchitmalhotra126: Hmm, this got me thinking about ways to avoid having to inspect the response and append a specific string here in the Lab2 layer. For example if we have other specific error types that different labs want to look out for, we could end up with a lot of special cases here. This would be out of scope for this PR, but one thing we could do is update `sourcesApi` to use the `HttpClient` put method, which will throw a `NetworkError` if a request fails (see [here](https://github.com/code-dot-org/code-dot-org/blob/d971cd2ff302dc6088c253cd071ae3911e3b83ed/apps/src/util/HttpClient.ts#L19)). That `NetworkError` type stores a reference to the response object, which then would be passed to AI Chat in the onSaveFail error handler. AI Chat could then inspect the response and decide how to handle it appropriately. 

For now, another option could be to just change the `Error` thrown here to a `NetworkError` and pass the response (i.e. `throw new NetworkError(response.status + ' ' + statusText, response);`). Then AI Chat can read the body JSON and show the appropriate error message.

sanchitmalhotra126 (commented):
Nice! A few comments / questions, but the general approach looks good to me.

apps/src/lab2/projects/SourcesStore.ts
fisher-alice: Sure - will update!

dashboard/legacy/middleware/files_api.rb
fisher-alice: Maybe we can wait till Sam gets back to add other fields. In the [jira ticket](https://codedotorg.atlassian.net/browse/LABS-791), she specifically requested that the system prompt and retrievals be filtered - I think bc these fields are sent to Sagemaker while model card info is not. With the tendency of WebPurify to have false positives, we can always follow-up with updating so that entire ` body` is checked.

dashboard/legacy/middleware/files_api.rb
fisher-alice: >Any harm in also checking PII? Might be something we need to clarify with curriculum/product, but since I imagine we'd also eventually be checking for PII in aichat_controller would it make sense to do it here as well (and not have to have this separate branch?)

Similar to response to comment above, I think perhaps we should first implement profanity checking, and then follow-up with PII when Sam gets back.

>Alternatively, if we want to check profanity only (that might be useful for other labs too), would be good to check the whole list rather than aichat specifically. 

Sounds good - thanks for the suggestion. Will update!

dashboard/legacy/middleware/files_api.rb
fisher-alice: Sure! Will update.

apps/src/aichat/redux/aichatRedux.ts
fisher-alice: ~~Text was suggested by Sam in jira ticket so I think perhaps we can leave string as-is for now, and we can iterate on this when Sam's back. Okay with you?~~
Now that we're probably leaving user work as-is, I think it would definitely be helpful to know which field was tagged as profane so will update this.

apps/src/aichat/redux/aichatRedux.ts
fisher-alice: ~~Good question! I'll post in Slack...~~
Actually, considering the risk of false positives, I agree that it would be better to just leave project as-is. Thanks!
@dancodedotorg lmk if you're okay to leave the user's work which was tagged as profane as-is (not saved, but still displayed in the workspace.

apps/src/lab2/projects/SourcesStore.ts
fisher-alice: Thanks! I'll update `Error` to `Network` error - appreciate the suggestion!

dashboard/legacy/middleware/files_api.rb
fisher-alice: Looking into what happens when both profanity and PII are included in the inspected text. If both are included, only PII will be reported according to https://github.com/code-dot-org/code-dot-org/blob/0be01bc8a5b927bf6b221b882dae97739463fe8c/lib/cdo/share_filtering.rb#L136.

fisher-alice: So I think I may update this method in `share_filtering` since we're interesting in filtering for profanity only at this point.

fisher-alice: Update: added `find_profanity_failure` in `share_filtering.rb` and updated the checking in `files_api.rb` https://github.com/code-dot-org/code-dot-org/pull/59414/commits/3ba5cb44226da04f29aa19f818af37fdacccd267#diff-7f7b9ffe903479398ea459bc7a01594b25b89baaa2c8fdda6aeb85deb7f9aacb

dashboard/legacy/middleware/files_api.rb
```diff
@@ -413,13 +413,24 @@ def put_file(endpoint, encrypted_channel_id, filename, body)
...
     # between their own projects -- skip this check for .java files, since in this use case
     # the files are only being used by a single user.
-    if endpoint == 'libraries' && file_type != '.java'
+    if (endpoint == 'libraries' && file_type != '.java') || profanity_project_type?(project_type)
+      text_to_check = body
+      if project_type == 'aichat'
+        source = JSON.parse(body)['source']
+        source_json = JSON.parse(source)
+        text_to_check = source_json['systemPrompt'] + ' ' + source_json['retrievalContexts'].join(' ')
+      end
```
sanchitmalhotra126: minor: might be nice to also pull this into a function to limit the amount of aichat-specific code here? Maybe something like (probably a less verbose way to name this)
```
text_to_check = get_profanity_project_text(project_type, body) if project_type
```
then later in a separate function
```
def get_profanity_project_text(project_type, body)
  if project_type == 'aichat'
    ... return system prompt & retrieval
  end
  body
end
```
That way we could also add other lab-specific parsing there if/when needed?

apps/src/aichat/redux/aichatRedux.ts
```diff
@@ -248,18 +249,40 @@ export const onSaveNoop =
...
-      notificationType: 'error',
-      timestamp: Date.now(),
-    })
-  );
-  // Notify the UI that the save is complete.
-  dispatch(endSave());
-};
+export const onSaveFail =
+  (e: NetworkError) => (dispatch: AppDispatch, getState: () => RootState) => {
+    e.response.json().then(body => {
```
sanchitmalhotra126: One thing to check - does `response.json()` still work correctly (i.e. doesn't throw an error) for a different type of error, like a 500 internal server error?

apps/src/aichat/views/AichatView.tsx
```diff
@@ -96,8 +97,10 @@ const AichatView: React.FunctionComponent = () => {
     projectManager.addSaveSuccessListener(() => {
       dispatch(onSaveComplete());
     });
-    projectManager.addSaveFailListener(() => {
-      dispatch(onSaveFail());
+    projectManager.addSaveFailListener((e: Error) => {
+      if (e instanceof NetworkError) {
+        dispatch(onSaveFail(e));
+      }
```
sanchitmalhotra126: I think it would still be good to show an error message even if we don't receive a `NetworkError` - could we pass in the error to `onSaveFail` and then check the error type there to determine what string to display?

dashboard/legacy/middleware/files_api.rb
```diff
@@ -428,8 +439,12 @@ def put_file(endpoint, encrypted_channel_id, filename, body)
...
       # Once we have a better geocoding solution in H1, we should start filtering for addresses again.
       # Additional context: https://codedotorg.atlassian.net/browse/STAR-1361
-      if share_failure && share_failure[:type] != "address"
-        details_key = share_failure.type == ShareFiltering::FailureType::PROFANITY ? "profaneWords" : "pIIWords"
+      if share_failure && share_failure[:type] != ShareFiltering::FailureType::ADDRESS
+        if share_failure[:type] == ShareFiltering::FailureType::PROFANITY
+          details_key = "profaneWords"
+        else
+          details_key = "pIIWords"
+        end
```
sanchitmalhotra126: Is this the same logic as before (with the ternary)? Any reason to change it (unless i'm missing something)?

lib/cdo/share_filtering.rb
```diff
@@ -112,6 +112,25 @@ def self.find_pii_failure(text, exceptions: false)
...
+  # Returns both the error type and the offending text snippet.
+  #
+  # If the check is successful and there were no offenses, the function
+  # will return `nil`.
+  #
+  # @param [String] text The text to search through.
+  # @param [String] locale a two-character ISO 639-1 language code
+  # @param [Hash] A set of text to replace before performing a profanity check.
+  # @return [ShareFailure, nil]
+  def self.find_profanity_failure(text, locale, profanity_filter_replace_text_list = {}, exceptions: false)
```
sanchitmalhotra126, line 125: Nice!

sanchitmalhotra126 (approved):
Looks good, nice work! A few minor suggestions, but nothing blocking.

dashboard/legacy/middleware/files_api.rb
fisher-alice: Sounds good! And can resolve a lint warning I got 'Use the return of the conditional for variable assignment and comparison.' - thanks!

dashboard/legacy/middleware/files_api.rb
fisher-alice: Oh yeah! I had changed this checking with `aichat`, but then updated to this logic comparing with my first change - https://github.com/code-dot-org/code-dot-org/pull/59414/commits/3ba5cb44226da04f29aa19f818af37fdacccd267#diff-7f7b9ffe903479398ea459bc7a01594b25b89baaa2c8fdda6aeb85deb7f9aacb.  So yes - same logic! Will revert.

apps/src/aichat/redux/aichatRedux.ts
fisher-alice: Oh - good call out! I added a catch to the promise chain to handle this and tested locally. Thanks!
Screenshot 2024-06-28 at 4 20 31 PM

sanchitmalhotra126 (approved):
LGTM!

PR #59368, opened 2024-06-24
Schoology Account Linking Bug

nicklathe:
This PR addresses a bug where users LTI launching from Schoology aren't taken into the expected account linking flow.

https://docs.google.com/document/d/1122s4C6M3HW_cj2DF3QqDTCA9BFLAEwRemsy8SMZD4Q/edit#heading=h.e9dpt2pc7d0

- Update `sign_out` to specify the `current_user` resource (see [devise docs](https://www.rubydoc.info/github/plataformatec/devise/main/Devise%2FControllers%2FSignInOut:sign_out))
- Change the cache key to the key stored in the session

## PR Checklist:

- [ ] Tests provide adequate coverage
- [ ] Privacy and Security impacts have been assessed
- [ ] Code is well-commented
- [ ] New features are translatable or updates will not break translations
- [ ] Relevant documentation has been added or updated
- [ ] User impact is well-understood and desirable
- [ ] Pull Request is labeled appropriately
- [ ] Follow-up work items (including potential tech debt) are tracked and linked

PR #59418, opened 2024-06-26
 fix(lms): only lock out users if session is populated with context

stephenliang:
This PR updates the lock out criteria to ensure we only lock out if the session is instantiated with the appropriate context. It also improves the specificity by only locking out new LTI users.

## Links

- [jira ticket](https://codedotorg.atlassian.net/browse/P20-1015)

## Testing story

[Tested by Carl
](https://codedotorg.slack.com/archives/C061VNB3YE9/p1719425018081999?thread_ts=1719413948.321039&cid=C061VNB3YE9)

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/dashboard/app/controllers/lti/v1/account_linking_controller.rb
+++ b/dashboard/app/controllers/lti/v1/account_linking_controller.rb
@@ -18,8 +18,8 @@ module Lti
         # clear the session, so we need to store cache key first, sign them out,
         # then set it again in their now empty session.
         if current_user
-          partial_registration_cache_key = PartialRegistration.cache_key(current_user)
-          sign_out
+          partial_registration_cache_key = session[PartialRegistration::SESSION_KEY]
+          sign_out current_user
           session[PartialRegistration::SESSION_KEY] = partial_registration_cache_key
         end
 
--- a/dashboard/app/controllers/lti/v1/account_linking_controller.rb
+++ b/dashboard/app/controllers/lti/v1/account_linking_controller.rb
@@ -18,8 +18,8 @@ module Lti
         # clear the session, so we need to store cache key first, sign them out,
         # then set it again in their now empty session.
         if current_user
-          partial_registration_cache_key = session[PartialRegistration::SESSION_KEY]
-          sign_out current_user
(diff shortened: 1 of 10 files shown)


dashboard/app/controllers/lti/v1/account_linking_controller.rb
```diff
@@ -18,8 +18,8 @@ def finish_link
         # clear the session, so we need to store cache key first, sign them out,
         # then set it again in their now empty session.
         if current_user
-          partial_registration_cache_key = session[PartialRegistration::SESSION_KEY]
-          sign_out current_user
+          partial_registration_cache_key = PartialRegistration.cache_key(current_user)
```
carl-codeorg: Was this change accidentally created by branching off a slightly outdated staging? Changing to pulling the key from the session was part of [this fix](https://github.com/code-dot-org/code-dot-org/pull/59368) for Schoology logouts

dashboard/lib/policies/lti.rb
```diff
@@ -219,4 +223,8 @@ def self.feedback_available?(user)
   def self.lti_registration_in_progress?(session)
     PartialRegistration.in_progress?(session) && PartialRegistration.get_provider(session) == AuthenticationOption::LTI_V1
   end
+
+  def self.account_linking?(session, user)
+    session[:lms_landing].present? && Policies::Lti.only_lti_auth?(user) && !user.lms_landing_opted_out
```
carl-codeorg: Since this is referencing another class method in the same class, I think it can just be `only_lti_auth?` without specifying Policies::Lti

dashboard/app/controllers/lti/v1/account_linking_controller.rb
stephenliang: Yep that's a rebase issue, fixed this

dashboard/lib/policies/lti.rb
stephenliang: Good catch!

carl-codeorg (approved):
Tested locally and it fixes the bug. Also ran through linking for new and synced users locally, as well as continuing with new accounts. Seems good to me!

PR #58944, opened 2024-05-28
Add start over button for Aichat levels

fisher-alice:
This PR adds a start over button for aichat levels. Starting over means resetting the model customizations to the default levelbuilder settings including the model card information. When the user clicks the start over button in the model customizations panel header (rotating arrows icon), a dialog will be displayed.
When they click on 'Start Over':
- model customizations and model card info will be reset to default levelbuilder settings,
- chat messages will be cleared and session id will be reset to `undefined`,
- an update notification that model customizations are now set to default settings will be displayed in chat workspace.

A report of this event will also be sent to Amplitude and Statsig.

## After update

Screenshot of notification message after user clicks on 'Start Over' in the dialog:
Screenshot 2024-05-29 at 3 12 20 PM

Screencast video of starting over in aichat:

https://github.com/code-dot-org/code-dot-org/assets/107423305/43d27560-21cc-42f9-9a85-73a72deb6178

Screencast video of dialog messages for different labs: aichat, python, and music:

https://github.com/code-dot-org/code-dot-org/assets/107423305/f567ba83-01bc-4f6b-8d2a-465e68619266

Screenshot of Amplitude/Statsig event in local dev env:
Screenshot 2024-05-29 at 12 21 19 PM

## Links
[jira](https://codedotorg.atlassian.net/browse/LABS-808)

## Testing story
I locally tested levels in /allthethings for aichat, python, and music labs.
I confirmed that saved customized settings in the 3 model customization tabs were reset to default settings after clicking on the start over button.

## Follow-up work
Create an `aichat/utils.ts` file includes `getNewMessageId`, `getCurrentTimestamp`, etc.

## PR Checklist:

- [ ] Tests provide adequate coverage
- [ ] Privacy and Security impacts have been assessed
- [ ] Code is well-commented
- [ ] New features are translatable or updates will not break translations
- [ ] Relevant documentation has been added or updated
- [ ] User impact is well-understood and desirable
- [ ] Pull Request is labeled appropriately
- [ ] Follow-up work items (including potential tech debt) are tracked and linked

PR #59425, opened 2024-06-26
AI Chat: always enable publish button if model card is complete

sanchitmalhotra126:
Fast follow to https://github.com/code-dot-org/code-dot-org/pull/59403 (thanks @fisher-alice for catching this edge case). Always enable the publish button (as long as model card is complete) even if there haven't been any changes made. I felt that this was the simplest solution since, even if there aren't any changes to save, the Publish button still functions as a way to transition the user to the Presentation View, and we can still show the "Ready to Publish" banner without it being too confusing.

https://github.com/code-dot-org/code-dot-org/assets/85528507/67cd06f8-a0aa-4f87-9844-5a42e4682d5b

## Testing story

Tested on Gen AI pilot model card level.

### DIFF
--- a/apps/src/aichat/views/AichatView.tsx
+++ b/apps/src/aichat/views/AichatView.tsx
@@ -25,13 +25,14 @@ import aichatI18n from '../locale';
 import {
   addNotification,
   clearChatMessages,
-  endSave,
   onSaveComplete,
   onSaveFail,
+  onSaveNoop,
   resetToDefaultAiCustomizations,
   selectAllFieldsHidden,
   setStartingAiCustomizations,
   setViewMode,
+  updateAiCustomization,
 } from '../redux/aichatRedux';
 import {getNewMessageId} from '../redux/utils';
 import {AichatLevelProperties, Notification, ViewMode} from '../types';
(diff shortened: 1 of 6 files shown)


apps/src/aichat/views/AichatView.tsx
```diff
@@ -163,6 +164,8 @@ const AichatView: React.FunctionComponent = () => {
 
   const resetProject = useCallback(() => {
     dispatch(resetToDefaultAiCustomizations(levelAichatSettings));
+    // Save the customizations to the user's project.
+    dispatch(updateAiCustomization());
```
fisher-alice, line 168: Thanks for adding this! Reminded me of your comment about the start over string - https://github.com/code-dot-org/code-dot-org/pull/58944#discussion_r1625151333. Is this a good time to move string?

fisher-alice (approved):
Thanks for the quick follow-up and checking in with curriculum.

apps/src/aichat/views/AichatView.tsx
sanchitmalhotra126: Oh good call! Will do

fisher-alice: Thank you!

PR #59429, opened 2024-06-26
mark computer-vision and csd2-2024 as ready to be translated

davidsbailey:
per [slack discussion](https://codedotorg.slack.com/archives/C045UAX4WKH/p1719438916277409?thread_ts=1719357431.904909&cid=C045UAX4WKH), we're not marking most 2024 courses as ready to translate any time soon. only these two are ready to translate.

## Testing story

existing test coverage, just to make sure nothing is broken

## Follow-up work

See the same slack thread above for ideas about moving this into a levelbuilder setting, especially since the curriculum team is generally managing the curriculum launch process now.


mgc1194 (approved):
Adding two new scripts to `ADDITIONAL_I18N_UNITS`.
Those scripts will be picked up by the i18n sync

davidsbailey:
Thank you Mario!

PR #59432, opened 2024-06-26
AI Chat: fix start over error

sanchitmalhotra126:
Fix an issue with Start Over in AI Chat that was caused by trying to write to a readonly property (`selectedModelId`). For some reason, this wasn't happening/reproducing locally, potentially due to differences in how redux/redux-toolkit runs in dev mode. However I was able to repro with a local production build and verify this fix.

## Links

https://codedotorg.atlassian.net/browse/LABS-894

## Testing story

Tested locally in the Gen AI pilot. Used a local production build to verify the repro and fix.

### DIFF
--- a/apps/src/aichat/redux/aichatRedux.ts
+++ b/apps/src/aichat/redux/aichatRedux.ts
@@ -424,9 +424,12 @@ const aichatSlice = createSlice({
       }
 
       // Make sure model ID is valid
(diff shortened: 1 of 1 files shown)


fisher-alice (approved):
Thanks for taking care of this!

Just a quick note that I was able to repro the bug locally without a local production build on `gen-ai-customizing-finetuning-explore-L5.level` before your fix.

PR #59437, opened 2024-06-27
Students in cpa lockout flow cannot change age or us state

juanmanzojr:
Students in cpa lockout flow cannot change age or us state.

Also adds ability to create a clever/google oauth account within a UI test.

Screenshot 2024-06-27 at 10 35 34 AM

## Links
- jira ticket: [here](https://codedotorg.atlassian.net/jira/software/c/projects/P20/boards/69?assignee=712020%3Af0e2716f-09f9-4686-8490-9f1f9a08bb06&selectedIssue=P20-952)

## Testing story
UI Tests within `dashboard/test/ui/features/platform/policy_compliance.feature`

## Follow up Work
Implementing backend so students in CPA flow cannot force update to us_state and age

### DIFF
--- a/dashboard/app/views/devise/registrations/edit.html.haml
+++ b/dashboard/app/views/devise/registrations/edit.html.haml
@@ -5,7 +5,7 @@
 - require 'queries/user'
 - @page_title = t('activerecord.attributes.user.edit_header')
 - no_email = current_user.no_personal_email? || params[:noEmail] == 'true'
-- studentInLockoutFlow = Policies::ChildAccount.underage?(current_user) && !Policies::ChildAccount::ComplianceState.permission_granted?(current_user)
+- studentInLockoutFlow = !!Policies::ChildAccount.lockout_date(user)
 
 = render_parental_permission_banner(current_user, request)
 
--- a/dashboard/app/views/devise/registrations/edit.html.haml
+++ b/dashboard/app/views/devise/registrations/edit.html.haml
@@ -74,7 +74,7 @@
       - if studentInLockoutFlow
         %em= t('user.update_field_parent_permission_required')
         %br/
-      = f.select :age, User::AGE_DROPDOWN_OPTIONS
+      = f.select :age, User::AGE_DROPDOWN_OPTIONS, {}, disabled: studentInLockoutFlow
     - if experiment_value('gender', request)
       .field
         = f.label :gender_student_input, t('signup_form.gender'), class: "label-bold"
@@ -86,7 +86,7 @@
       - if studentInLockoutFlow
         %em= t('user.update_field_parent_permission_required')
         %br/
-      = f.select :us_state, us_state_options, {include_blank: t('school_info.select_state')}, required: true
+      = f.select :us_state, us_state_options, {include_blank: t('school_info.select_state')}, disabled: studentInLockoutFlow, required: true
       = f.hidden_field :country_code, value: country_code
   - if resource.teacher?
     .field
@@ -311,7 +311,6 @@
       dependentStudentsCount: Queries::User.dependent_students_count(current_user.id),
       personalAccountLinkingEnabled: @personal_account_linking_enabled,
       lmsName: Queries::Lti.get_lms_name_from_user(current_user),
-      studentInLockoutFlow: studentInLockoutFlow,
     }.to_json
   }
 %script{src: webpack_asset_path('js/devise/registrations/edit.js'), data: script_data}
--- a/dashboard/app/views/devise/registrations/edit.html.haml
+++ b/dashboard/app/views/devise/registrations/edit.html.haml
@@ -5,7 +5,7 @@
 - require 'queries/user'
 - @page_title = t('activerecord.attributes.user.edit_header')
 - no_email = current_user.no_personal_email? || params[:noEmail] == 'true'
-- studentInLockoutFlow = !!Policies::ChildAccount.lockout_date(user)
+- studentInLockoutFlow = !!Policies::ChildAccount.lockout_date(current_user)
 
 = render_parental_permission_banner(current_user, request)
 
--- a/dashboard/app/views/devise/registrations/edit.html.haml
+++ b/dashboard/app/views/devise/registrations/edit.html.haml
@@ -300,7 +300,6 @@
   script_data = {
     edit: {
       userAge: current_user.age,
-      userState: current_user.us_state,
       userType: current_user.user_type,
       isAdmin: current_user.admin,
       isOauth: current_user.oauth?,
@@ -310,7 +309,7 @@
       isCleverStudent: current_user.clever_student?,
       dependentStudentsCount: Queries::User.dependent_students_count(current_user.id),
       personalAccountLinkingEnabled: @personal_account_linking_enabled,
-      lmsName: Queries::Lti.get_lms_name_from_user(current_user),
+      lmsName: Queries::Lti.get_lms_name_from_user(current_user)
     }.to_json
   }
 %script{src: webpack_asset_path('js/devise/registrations/edit.js'), data: script_data}
--- a/dashboard/app/views/devise/registrations/edit.html.haml
+++ b/dashboard/app/views/devise/registrations/edit.html.haml
@@ -5,6 +5,7 @@
 - require 'queries/user'
 - @page_title = t('activerecord.attributes.user.edit_header')
 - no_email = current_user.no_personal_email? || params[:noEmail] == 'true'
+- studentInLockoutFlow = Policies::ChildAccount.underage?(current_user) && !Policies::ChildAccount::ComplianceState.permission_granted?(current_user)
 
 = render_parental_permission_banner(current_user, request)
 
@@ -70,6 +71,9 @@
   - if !resource.teacher?
     .field
       = f.label t('signup_form.age'), class: "label-bold"
+      - if studentInLockoutFlow
+        %em= t('user.update_field_parent_permission_required')
+        %br/
       = f.select :age, User::AGE_DROPDOWN_OPTIONS
     - if experiment_value('gender', request)
       .field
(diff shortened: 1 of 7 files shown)


dashboard/app/views/devise/registrations/edit.html.haml
```diff
@@ -5,6 +5,7 @@
 - require 'queries/user'
 - @page_title = t('activerecord.attributes.user.edit_header')
 - no_email = current_user.no_personal_email? || params[:noEmail] == 'true'
+- studentInLockoutFlow = Policies::ChildAccount.underage?(current_user) && !Policies::ChildAccount::ComplianceState.permission_granted?(current_user)
```
daynew:
```suggestion
- studentInLockoutFlow = !!Policies::ChildAccount.lockout_date(user)
```
Alternatively, I think this method does the same thing. It should return the date in which a user will be locked out, which is essentially the same as determining that they are in the "grace_period".

You should double-check this suggestion by making sure your UI test still passes.
 
One use case to consider is `U13 + Colorado + Clever`. I don't think we want to lock the age and state, right?

dashboard/test/ui/features/platform/policy_compliance.feature
```diff
@@ -264,3 +264,27 @@ Feature: Policy Compliance and Parental Permission
 
     # And it should tell me that the request was granted
     And element "#permission-status" contains text "Granted"
+  
+  Scenario: Student in the CPA flow cannot change their age or state
+    Given I create an authorized teacher-associated under-13 student in Colorado named "Tandy" after CPA exception
```
daynew: Some other test cases to consider:

* Over-13 and in Colorado?\
* U13, Colorado, Clever-only

daynew (changes requested):
I have some test-cases I'd like you to double-check before I can give approval.

dashboard/test/ui/features/platform/policy_compliance.feature
juanmanzojr: Ill add these

dashboard/app/views/devise/registrations/edit.html.haml
juanmanzojr: Correct, we are not locking out age and state for clever users.

carl-codeorg: During the bugbash, it was confirmed that we want to prevent U13 + CO + LTI users from editing age and state. If we consider LTI and Clever to be in the same bucket, wouldn't we want to lock out age and state for Clever users?

dashboard/test/ui/features/platform/policy_compliance.feature
carl-codeorg: What is the distinction between Clever users and LTI users that makes us want to lock the state and age dropdowns for LTI but not for Clever?

dashboard/app/views/devise/registrations/edit.html.haml
```diff
@@ -70,6 +71,9 @@
   - if !resource.teacher?
     .field
       = f.label t('signup_form.age'), class: "label-bold"
+      - if studentInLockoutFlow
+        %em= t('user.update_field_parent_permission_required')
+        %br/
```
wilkie, line 75: why don't we just disable it here instead of passing the state to the JavaScript to disable it there?

dashboard/app/views/devise/registrations/edit.html.haml
```diff
@@ -79,6 +83,9 @@
       - location = Geocoder.search(request.ip).try(:first)
       - country_code = location&.country_code.to_s.upcase
       = f.label :us_state, t('activerecord.attributes.user.us_state'), class: 'label-bold'
+      - if studentInLockoutFlow
+        %em= t('user.update_field_parent_permission_required')
+        %br/
       = f.select :us_state, us_state_options, {include_blank: t('school_info.select_state')}, required: true
```
wilkie: if we do this (and for age as well) do we get to not pass things to JavaScript to disable?

```suggestion
      = f.select :us_state, us_state_options, {include_blank: t('school_info.select_state')}, required: true, disabled: !!Policies::ChildAccount.lockout_date(current_user)
```

wilkie (approved):
this looks great! thanks!

dashboard/test/ui/features/platform/policy_compliance.feature
daynew: > What is the distinction between Clever users and LTI users that makes us want to lock the state and age dropdowns for LTI but not for Clever?

I don't think we want to lock it for LTI either. Clever and LTI should be treated the same.

dashboard/app/views/devise/registrations/edit.html.haml
daynew: > During the bugbash, it was confirmed that we want to prevent U13 + CO + LTI users from editing age and state. If we consider LTI and Clever to be in the same bucket, wouldn't we want to lock out age and state for Clever users?

Yeah you are right @carl-codeorg, I was wrong about this feedback.

dashboard/app/views/devise/registrations/edit.html.haml
```diff
@@ -5,6 +5,7 @@
 - require 'queries/user'
 - @page_title = t('activerecord.attributes.user.edit_header')
 - no_email = current_user.no_personal_email? || params[:noEmail] == 'true'
+- studentInLockoutFlow = !!Policies::ChildAccount.lockout_date(current_user)
```
daynew:
```suggestion
- studentInLockoutFlow = Policies::ChildAccount.underage?(current_user) && !Policies::ChildAccount::ComplianceState.permission_granted?(current_user)
```

I'm sorry, I think you had this code right originally. When I saw `studentInLockoutFlow`, I immediately though we were talking about the "CPA lockout flow", not the "Account linking lockout flow". I'm sorry for the back and forth on this.

daynew (changes requested):

dashboard/app/views/devise/registrations/edit.html.haml
wilkie: I'll pull this down and double check. It would be interesting if our tests aren't covering it.

wilkie: Yeah the original was correct. It's basically what the logic a little further down page does when deciding whether or not account linking is locked. It's a little redundant, so I might consider moving that logic up to the top of the page and using it in both places. Will create a commit and push it to this feature branch.

wilkie:
I've fixed up that logic and added better documentation around it so that there is less confusion about what these lines mean. I'll run the tests locally ahead of the CI and see if this fixes those issues.

wilkie:
They seem to be passing locally, so fingers are crossed.

wilkie:
```
[platform/policy_compliance] UI tests for Chrome_platform_policy_compliance succeeded (6:22 minutes, 17 passed)
```

wilkie:
Local manual testing seems to line up with my expectations as well.

dashboard/app/views/devise/registrations/edit.html.haml
```diff
@@ -6,6 +6,19 @@
...
+-# We determine if the student is potentially locked by looking at their age
+-# If they are older (or there is no policy for them) they are unlocked
+-# This ignores them being explicitly unlocked by parental permission so we
+-# can show the 'Granted' status of that permission later.
+- potentially_locked = Policies::ChildAccount.underage?(current_user)
+-# The student is in a 'lockout' flow if they are potentially locked out and not unlocked
+- studentInLockoutFlow = potentially_locked && !Policies::ChildAccount::ComplianceState.permission_granted?(current_user)
+-# We need to also account for the case when the US State is not specified
+-# All students are locked out of account settings features until they specify these
+- locked = studentInLockoutFlow || current_user.country_code.nil? || current_user.us_state.nil?
```
daynew:
```suggestion
- locked = studentInLockoutFlow || ((current_user.country_code.nil? || current_user.us_state.nil?) && !isUSA)
```
We only need to ask for the US State and/or lock account linking if the request is coming from the USA, right?

dashboard/app/views/devise/registrations/edit.html.haml
```diff
@@ -79,7 +95,10 @@
       - location = Geocoder.search(request.ip).try(:first)
       - country_code = location&.country_code.to_s.upcase
       = f.label :us_state, t('activerecord.attributes.user.us_state'), class: 'label-bold'
-      = f.select :us_state, us_state_options, {include_blank: t('school_info.select_state')}, required: true
+      - if studentInLockoutFlow
+        %em= t('user.update_field_parent_permission_required')
+        %br/
+      = f.select :us_state, us_state_options, {include_blank: t('school_info.select_state')}, disabled: studentInLockoutFlow, required: true
```
daynew:
```suggestion
      - if !isUSA
        = f.select :us_state, us_state_options, {include_blank: t('school_info.select_state')}, disabled: studentInLockoutFlow, required: true
```
This was missed in the original PR where the field was added, but we shouldn't show this input to students outside the USA.

dashboard/app/views/devise/registrations/edit.html.haml
wilkie: That makes sense.

dashboard/app/views/devise/registrations/edit.html.haml
wilkie: The logic feels weird... why is it `!isUSA` ... shouldn't it be `if isUSA` to render the state selection?

dashboard/app/views/devise/registrations/edit.html.haml
```diff
@@ -70,17 +70,22 @@
...
-      = f.select :us_state, us_state_options, {include_blank: t('school_info.select_state')}, required: true
-      = f.hidden_field :country_code, value: country_code
+    - if @is_usa
+      .field
+        = f.label :us_state, t('activerecord.attributes.user.us_state'), class: 'label-bold'
+        - if @student_in_lockout_flow
+          %em= t('user.update_field_parent_permission_required')
+          %br/
+        = f.select :us_state, us_state_options, {include_blank: t('school_info.select_state')}, disabled: @student_in_lockout_flow, required: true
+        = f.hidden_field :country_code, value: country_code
```
daynew:
```suggestion
        = f.hidden_field :country_code, value: @country_code
```
When we were pairing, did you make this change already?

daynew (approved):

dashboard/app/views/devise/registrations/edit.html.haml
wilkie: Hmm I did but did not push it.

PR #59382, opened 2024-06-25
P20-1007: Update CAP Amplitude events

artem-vavilov:
## Summary
- Combined the Parental Request Modal events triggered by banner clicks and those shown daily.
- Added events reporting of the CAP partial lockout widget.
- Removed redundant tracking of the consent status changes.
## Links
- JIRA ticket: [P20-1007](https://codedotorg.atlassian.net/browse/P20-1007)

PR #59438, opened 2024-06-27
/lockout Amplitude metrics

daynew:
As a PM responsible for the Colorado Privacy Act, I want to usage metrics of the LockoutPanel so I can measure how effective it is.

When a student is locked out due to the Colorado Privacy Act (CPA), they are redirected to *studio.code.org/lockout*. This page only has one React component, LockoutPanel.jsx. The main goal of this PR is to add Amplitude metrics to all the interactive components so we can measure how effective it is. One big change that also happened with this component, is making similar to our other Parent Permission Request widgets where the UI is update dynamically and there is error handling when API requests fail.

These changes borrow/steal heavily from [LockoutLinkedAccounts.jsx](https://github.com/code-dot-org/code-dot-org/pull/59382/files#diff-215ae7d0a710b6895daa1a51180e6d79b1e280760556559869c1e265ff22f794). There will be a post-launch JIRA task to consolidate all the various Parent Permission Request widgets.

* [Converts LockoutPanel to TypeScript](https://github.com/code-dot-org/code-dot-org/pull/59438/commits/6fe8c44b038d54da7ce1c123ac502c3380c3db17)
* LockoutPanel doesn't require the page to be reloaded to update the UI.
* LockoutPanel displays an error if the API call to send a Parent Permission Request fails.
* LockoutPanel reports Amplitude metrics.

### New Metrics
```
CAP_LOCKOUT_SHOWN: 'CAP Lockout Shown',
CAP_LOCKOUT_SIGN_OUT: 'CAP Lockout Sign Out',
CAP_LOCKOUT_EMAIL_SUBMITTED: 'CAP Lockout Email Submitted',
CAP_LOCKOUT_EMAIL_UPDATED: 'CAP Lockout Email Updated',
CAP_LOCKOUT_EMAIL_RESEND: 'CAP Lockout Email Resend',
```
## Links
[Jira](https://codedotorg.atlassian.net/browse/P20-1014)

## Testing story
* Manually tested on localhost
* React Storybook

## Video
https://github.com/code-dot-org/code-dot-org/assets/1372238/6be8a913-326e-473e-88bb-3962217f0a33

## PR Checklist:

- [ ] Code is well-commented


apps/src/templates/sessions/LockoutPanel.tsx
```diff
@@ -37,22 +52,50 @@ export default function LockoutPanel(props) {
...
+  };
 
-    const form = document.getElementById('lockout-panel-form');
-    form.submit();
+  const handleUpdate = () => {
+    reportEvent(EVENTS.CAP_PARENT_EMAIL_UPDATED, {
+      inSection: currentUser.inSection,
+      consentStatus: currentUser.childAccountComplianceState,
+    });
   };
```
artem-vavilov: There are two issues:
1. The form ignores validation errors and counts them as successes. Because of this, we will report an event even if it didn't actually happen:
https://github.com/code-dot-org/code-dot-org/blob/3aa9278211cc8e4b7e04a04aa7f2446e5272b116/dashboard/app/controllers/policy_compliance_controller.rb#L67-L70 https://github.com/code-dot-org/code-dot-org/blob/5cc0038e840532e70d2dc79d86e9bdc7863d9239/dashboard/lib/forms/child_account/parental_permission_request.rb#L16-L22
2. When users click `Update` for the same email, it's technically the `Resend` event and not the`Update`

artem-vavilov: Please check [LockoutLinkedAccounts.jsx](https://github.com/code-dot-org/code-dot-org/pull/59382/files#diff-215ae7d0a710b6895daa1a51180e6d79b1e280760556559869c1e265ff22f794), where I had to deal with these issues as well:
https://github.com/code-dot-org/code-dot-org/blob/2811284a88f0e353cc1edf9c9b540246376c0dfd/apps/src/templates/policy_compliance/LockoutLinkedAccounts.jsx#L73-L100

daynew: I stole a lot of code from [LockoutLinkedAccounts.jsx](https://github.com/code-dot-org/code-dot-org/pull/59382/files#diff-215ae7d0a710b6895daa1a51180e6d79b1e280760556559869c1e265ff22f794). I will create a JIRA to consolidate all of the various CAP Lockout widgets.

apps/src/templates/sessions/LockoutPanel.tsx
```diff
@@ -0,0 +1,466 @@
...
+                      })}
+                    </em>
+                  </p>
+                )}
+              </div>
+            </div>
+          </div>
+        )}
+
+        {status !== ChildAccountComplianceStates.PERMISSION_GRANTED && (
```
wilkie: This condition is the same as the one directly above it. That seems correct though redundant. Is this just to avoid the indent? Maybe we can pull out the form rendering to its own function.

wilkie (approved):
Looks good. I look forward to refactoring this with the other lockout widget as you noted.

I'm unsure if the events have the right data. I think so. The only thing that jumps out is that the user id is not sent.

The comment I have about a possible slight refactor could just happen later when the overall refactor happens.

👍

apps/src/templates/sessions/LockoutPanel.tsx
```diff
@@ -0,0 +1,466 @@
...
+  const pendingPromptParts = pendingPrompt.split('{pendingEmail}');
+
+  const tokenElement = document.querySelector('meta[name="csrf-token"]');
+  let csrfToken = '';
+  if (tokenElement) {
+    const content = tokenElement.attributes.getNamedItem('content');
+    if (content) {
+      csrfToken = content.value;
+    }
+  }
```
artem-vavilov: I believe we don't need this since we started using `parentalPermissionRequestReducer`

apps/src/templates/sessions/LockoutPanel.tsx
```diff
@@ -0,0 +1,466 @@
...
+/**
+ * This panel represents the page that is displayed to accounts that are being
+ * locked from accessing certain parts of the site until they get parental
+ * permission. This panel gives a form to request (or see details about a
+ * pending request for) parental permission.
+ */
+const LockoutPanel: React.FC<LockoutPanelProps> = props => {
+  const reportEvent = (eventName: string, payload: object = {}) => {
+    analyticsReporter.sendEvent(eventName, payload, PLATFORMS.AMPLITUDE);
+  };
```
artem-vavilov, line 34: Due to the error `SyntaxError: Unexpected token '<', "<!DOCTYPE "... is not valid JSON`, the `currentUser` data is not set in the Redux store and is absent from the Amplitude events:
https://github.com/code-dot-org/code-dot-org/blob/9e79cdbafb97d34fe8f1281c53c3645c93e46216/apps/src/templates/currentUserRedux.js#L240-L244

### Error




### Event reporting
Screenshot 2024-06-30 at 20 54 31

### Event data

artem-vavilov: The fix: https://github.com/code-dot-org/code-dot-org/pull/59497

artem-vavilov (approved):
Nice work! 🙌

wilkie:
Finally manually tested it and it looks good here!

PR #59442, opened 2024-06-27
P20-987: Add ability to mock DCDO via cookie for testing

artem-vavilov:
## Summary

## Links
- Related PR: https://github.com/code-dot-org/code-dot-org/pull/59219
- Revert PR: https://github.com/code-dot-org/code-dot-org/pull/59440
- UI test: https://github.com/code-dot-org/code-dot-org/pull/59637


lib/cdo/http_cache.rb
```diff
@@ -58,6 +58,8 @@ class HttpCache
     'onetrust_cookie_scripts',
     # Feature flag for the Colorado Privacy Act (CPA)
     'cpa_experience',
+    # Allows DCDO settings via cookies. See: Rack::CookieDCDO
+    'DCDO',
```
Hamms: I _really_ don't like the idea of allowing this cookie in production when we only need it for test. If we're not able to base the conditional logic on the `CDO.use_cookie_dcdo` config value, we should at the very least have it be inspecting the environment.

I would ideally like for it to be able to be based on that config value, though. Do we know why exactly the last attempt didn't work? A quick test on an adhoc indicates that we should be able to read the correct value at this point in the startup process; is it possible the problem was with the `require` and using a constant, and we can keep the `if CDO.use_cookie_dcdo` logic in place but just hardcode the name of the cookie?

artem-vavilov: I believe that if this part `require 'cdo/rack/cookie_dcdo'` didn't work, we would see an error during the server run.
However, instead of an error, the DCDO cookies test just failed.
Therefore, the issue is most likely with `CDO.use_cookie_dcdo`.

Hamms: That does indeed seem most likely, but I think we should figure it out for sure. If we aren't able to use `CDO.use_cookie_dcdo` here, we'll need to find something else to base the conditional on.

Are there any experiments you'd like to run on the test server to attempt to figure this out? I could try to find a time this week to make some nondisruptive local modifications before a build runs. Even just something like,

```diff
diff --git a/lib/cdo/http_cache.rb b/lib/cdo/http_cache.rb
index 39483751ddd..1abb9e284e0 100644
--- a/lib/cdo/http_cache.rb
+++ b/lib/cdo/http_cache.rb
@@ -152,6 +152,10 @@ class HttpCache
       storage_id,
     ].concat(default_cookies)
 
+    if env.to_s == 'test'
+      CDO.log.info("CDO.use_cookie_dcdo: #{CDO.use_cookie_dcdo.inspect}")
+    end
+
     {
       pegasus: {
         behaviors: [
```

artem-vavilov: [Allows DCDO cookies only in the test env](https://github.com/code-dot-org/code-dot-org/pull/59442/commits/c83a9067523cfd391c235347e2112e6acc85a315)

Hamms: To clarify: I think inspecting the environment is a reasonable last-resort solution, but it really should be a last resort. I can easily imagine a situation in the future where developers want to enable this functionality for another environment, and I'd like them to be able to do so by just setting `use_cookie_dcdo` for that environment to true rather than needing to go update manual environment checks.

If we do have to resort to checking the environment, we'll need to put a nice big comment in there explaining for future developers exactly why checking `use_cookie_dcdo` was not an option for us, and to do that we need to understand why it didn't work. And hopefully by doing that investigation, we'll figure out how to get it to work and can avoid this last resort entirely.

Hamms (approved):
LGTM! Thanks for working through all the different iterations of this with me. Looking forward to figuring out what's up with that UI test! :crossed_fingers:

PR #59446, opened 2024-06-27
Ensure the time for the next tests is unfrozen

artem-vavilov:


wilkie (approved):
Very reasonable change to combat some potential flakiness.

PR #59448, opened 2024-06-27
Multi-worktree seeding support

etaderhold:
This PR adds the capability to set a `DASHBOARD_DIR=...` environment variable that will be used by the `rake:seed` task as a root directory for levelbuilder-generated content.

It also updates the cookbook to checkout a `staging` worktree on `production-daemon` so that this can be used to implement the new "seed production from staging" workflow.

## Links
[Design doc](https://docs.google.com/document/d/1RKrKOGL0DBAfG5YGrMcwcF2jrozvj0d50Go11QC9jwQ/edit#heading=h.umozcosknila)

## Testing story

I tested this on an adhoc.

The cookbook successfully created a separate `staging` worktree on the adhoc machine when it was initially set up. I attempted to have it do a sparse checkout of only the `dashboard/config` directory where levelbuilder content is written, but it failed for some reason and I haven't (yet) been able to debug it.

All the curriculum content seemed to seed perfectly well from the normal location in the initial setup where the new `DASHBOARD_DIR` variable is omitted.

I manually modified some levelbuilder files in the staging worktree and then ran `rake seed:all DASHBOARD_DIR=~/staging/dashboard/`

Results:

> https://adhoc-multi-worktree-seed-studio.cdn-code.org/s/csd1-2023/lessons/1/levels/1/page/1
> image

> https://adhoc-multi-worktree-seed-studio.cdn-code.org/s/csd1-2023
> image

## PR Checklist:

- [ ] Code is well-commented


dashboard/app/models/data_doc.rb
```diff
@@ -50,11 +50,11 @@ def remove_serialization
   end
 
   # creates and deletes records to match all the seed files
-  def self.seed_all
+  def self.seed_all(dashboard_root = '.')
```
davidsbailey, line 53: It looks like someone calling one of these methods from a cronjob or other bin script could have a bad default value here or below. It could be worth doing a spotcheck of ways to reach methods with `dashboard_root = '.'` from the bin directory and/or keeping an eye out for cronjob errors in honeybadger after you merge this.

dashboard/lib/policies/level_files.rb
```diff
@@ -32,8 +32,8 @@ def self.level_name_from_path(path)
 
     # Return a glob which can be used either to find the specified level file
     # by name if provided, or all level files if not.
-    def self.level_file_glob(level_name)
-      level_name ? "config/levels/**/#{level_name}.level" : 'config/levels/**/*.level'
+    def self.level_file_glob(level_name, root_dir = '.')
```
davidsbailey, line 35: How about providing `Rails.root` as the default here? it looks like this method could be reached from outside of the dashboard dir: https://github.com/code-dot-org/code-dot-org/blob/ab4507cf4832b04634599017460053250ae864ec/bin/oneoff/import_concept_difficulty_tags.rb#L42

davidsbailey:
Nice solution Eric! I spotted just one small issue inline.

Code changes LGTM, except the cookbooks which I'm relying on an infra team member to review.

Question about the overall approach -- when we discussed this previously, there was some question as to whether we'd want to use the staging branch (which could contain untested commits made by developers) or the levelbuilder branch (any dev changes that reach here have been tested). OTOH, stuff on the levelbuilder branch has never been seeded into another environment and so in that sense is untested. Can you share your thinking on the tradeoffs here, and whether a best-of-both-worlds approach might be possible?

davidsbailey (approved):

dashboard/lib/policies/level_files.rb
etaderhold: My goal with these was to leave the behavior unchanged for any other callers that currently don't pass parameters. I assumed that relative paths would be resolved relative to the current working directory rather than `Rails.root`. Was that assumption mistaken?

davidsbailey: Good point, I don't know the answer but it looks like you're making a best-effort attempt to preserve existing behavior, and that seems like the right level of effort here. if we want import_concept_difficulty_tags.rb to keep working over time then we should add test coverage next time we need to use it.

sureshc:
I’m excited about this! I haven’t had time to fully digest the changes. Got 15 minutes today in the office to walk through this in real-time?

etaderhold:
> I’m excited about this! I haven’t had time to fully digest the changes. Got 15 minutes today in the office to walk through this in real-time?

Yes, let's!

cookbooks/cdo-repository/recipes/default.rb
```diff
@@ -61,3 +61,27 @@
     not_if {File.exist? worktree_path}
   end
 end
+
+# Create sparse checkout of staging content directories on production-daemon
+# for the purpose of content seeding.
+if node.name == 'production-daemon'
```
sureshc: Eventually, every installation of our application except levelbuilder and test would likely be subscribing to the curriculum content committed to our main branch (which we call `staging`), and node names are a little brittle.

Maybe we should switch this login to test that we're running on a daemon  AND that the environment type is `production`.  I think we could start with this and then expand the array later to include `adhoc`:

``` ruby
if (node['cdo-apps']['daemon'] && %w[production].include?(node.chef_environment))
```

cookbooks/cdo-repository/recipes/default.rb
```diff
@@ -61,3 +61,27 @@
...
+  worktree_path = File.join(home_path, 'staging')
```
sureshc, line 68: Maybe call this `curriculum-content` to be more clear about what this other work tree is doing on the file system? I think that would require updating the git command we're using to create the work tree to pass separate arguments to the `git worktree` command for the path (`curriculum-content`) and the commit/branch (`staging`).

sureshc:
Writing down what I discussed in person with @etaderhold ...

I think our long term goal is to store and version curriculum content separately from our source code. I suggest a couple of naming changes to take us in that direction and to make it clear to other people reading this code or browsing the file system:
1)  Name the work tree directory  `curriculum-content` (or `staging-curriculum-content`)
2)  Change the name of the seeding method argument from `dashboard_root` to `curriculum_content` (or `curriculum_content_dir` or `curriculum_content_root`).

cookbooks/cdo-repository/recipes/default.rb
etaderhold: Done.

cookbooks/cdo-repository/recipes/default.rb
etaderhold: Looks like configuring a worktree with a different path than the branch name is possible but would require a more complicated git command (and another round of testing with spinning up a new adhoc environment).

sureshc (approved):

PR #59451, opened 2024-06-27
Change log event method

nicklathe:
The Metrics::Events.log_event Statsig wrapper was refactored to set the user to nil if empty in the method parameters. However, it was discovered this causes issues in the Statsig dashboard. The wrapper was refactored to include a [Metrics::Events.log_event_with_session](https://github.com/code-dot-org/code-dot-org/blob/staging/dashboard/lib/metrics/events.rb#L49) for this case. 

Changes:
* Use `log_event_with_session` instead of `log_event`, and pass in session as argument

## Links
[Jira](https://codedotorg.atlassian.net/browse/P20-1018)

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/dashboard/app/controllers/lti/v1/dynamic_registration_controller.rb
+++ b/dashboard/app/controllers/lti/v1/dynamic_registration_controller.rb
@@ -92,7 +92,8 @@ module Lti
           metadata = {
             lms_name: platform[:name],
           }
-          Metrics::Events.log_event(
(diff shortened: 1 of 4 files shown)


carl-codeorg (approved):
LGTM - only suggestion would be the same thing Wilkie suggested in my PR. In the tests for these methods, you could expect a log with the statsig stable id property

stephenliang (approved):

PR #59452, opened 2024-06-27
Fix translated code-block's format in long_instructions

mgc1194:
### The issue: 
Blocks in instructions do not render properly. It doesn’t completely break the experience because students can still access the appropriate documentation page, but it doesn’t look as neat as the instruction in English.

### The reason:

The translated content has the code backticks escaped: [\`if-statement\`(#64B5F6)] vs [\\\`if-statement\\\`(#64B5F6)]

**Why are backticks escaped?** Backtics are naturally escaped in strings, however, the back 
slash escaping backticks, also get escaped when downloading translations from crowdin, resulting in a double scped backtick (\\\\`).
We do not redact code blocks in course_content (we can redact code blocs and we do so in other strings). A process that would prevent this format issue during the download process.
We could start redacting these code blocks, however, translations already exist and we would have to re-translate all strings containing code blocks (very time-consuming).

I opted for a less elegant solution reformating existing translations.

This PR:
- Adds a private method (`unescape_codeblocks`) that unescapes backticks. 
- Uses the `unescape_codeblocks` in the `localized_long_instructions` method.
- Additionally add some documentation in undocumented methods.

## Links

- jira ticket: [P20-853](https://codedotorg.atlassian.net/browse/P20-853)

## Testing story

Screenshot 2024-06-27 at 12 36 01

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/dashboard/app/models/levels/blockly.rb
+++ b/dashboard/app/models/levels/blockly.rb
@@ -1008,6 +1008,7 @@ class Blockly < Level
   # @param text [String] the text to unescape.
   # @return [String] the text with unescaped backticks.
   private def unescape_codeblocks(text)
+    return text unless text
     text.gsub('\\`', '`')
   end
 end
--- a/dashboard/app/models/levels/blockly.rb
+++ b/dashboard/app/models/levels/blockly.rb
@@ -465,7 +465,7 @@ class Blockly < Level
     localized_blockly_in_text(unescaped_codeblocks)
   end
 
-  # Processes and localizes the TRANSALTIONTEXT from i18n start libraries content.
+  # Processes and localizes the TRANSLATIONTEXT from i18n start libraries content.
   # @param start_libraries [String] JSON-encoded string representing an array of library objects.
   # Each library object should contain a name and source field.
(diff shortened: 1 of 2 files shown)


dashboard/app/models/levels/blockly.rb
```diff
@@ -457,11 +457,18 @@ def localized_failure_message_override
...
+  # Retrieve the localized property for "long_instructions",
+  # @return [String] the localized long_instructions property
   def localized_long_instructions
     localized_long_instructions = get_localized_property("long_instructions")
-    localized_blockly_in_text(localized_long_instructions)
+    unescaped_codeblocks = unescape_codeblocks(localized_long_instructions)
+    localized_blockly_in_text(unescaped_codeblocks)
   end
 
+  # Processes and localizes the TRANSALTIONTEXT from i18n start libraries content.
```
wilkie: typo TRANSLATIONTEXT
```suggestion
  # Processes and localizes the TRANSLATIONTEXT from i18n start libraries content.
```

wilkie (approved):
looks good. one typo. thanks for the great work!

PR #59455, opened 2024-06-27
[Google Blockly] clear procedure maps on start over

mikeharv:
Following the Maze migration bug bash, @molly-moen reported an issue where the procedure map wasn't being reset when a user clicks "Start Over". 

1. create a "do something" function
2. Click start over
3. Go to create another function, the default name is now "do something2", implying that the function names didn't get wiped on start over
4. Close the editor, pull out "do something2" and the in-workspace editor pops up!

This problem wasn't noticeable with Sprite Lab because the Version History "Start Over" triggers a full page reload. It's also not easily reproducible in Dance or Minecraft as those labs don't use the modal function editor.

Fortunately, there's a straight-forward fix. When we clear each of the three student workspaces, we can manually clear each of their procedure maps. Following this step, we'll reload the start blocks as normal which will re-populate any expected procedures defined by the start blocks. Now, previously-used procedure names are available again:

https://github.com/code-dot-org/code-dot-org/assets/43474485/b570a258-99c7-4c03-ac3e-cbc47cab28ad

## Links

Slack thread:
- https://codedotorg.slack.com/archives/C05HZFH7BED/p1719508995108649

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/blockly/googleBlocklyWrapper.ts
+++ b/apps/src/blockly/googleBlocklyWrapper.ts
@@ -827,14 +827,23 @@ function initializeBlocklyWrapper(blocklyInstance: GoogleBlocklyInstance) {
 
   // Google Blockly labs also need to clear separate workspaces for the function editor.
   blocklyWrapper.clearAllStudentWorkspaces = function () {
-    Blockly.getMainWorkspace().clear();
-    const functionEditorWorkspace = Blockly.getFunctionEditorWorkspace();
-    if (functionEditorWorkspace) {
-      functionEditorWorkspace.clear();
-    }
-    if (Blockly.getHiddenDefinitionWorkspace()) {
-      Blockly.getHiddenDefinitionWorkspace().clear();
-    }
+    // Disable Blockly events to prevent unnecessary event mirroring
+    Blockly.Events.disable();
+
+    const studentWorkspaces = [
+      Blockly.getMainWorkspace(),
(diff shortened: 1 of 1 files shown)


apps/src/blockly/googleBlocklyWrapper.ts
```diff
@@ -827,14 +827,23 @@ function initializeBlocklyWrapper(blocklyInstance: GoogleBlocklyInstance) {
...
+    const studentWorkspaces = [
+      Blockly.getMainWorkspace(),
+      Blockly.getFunctionEditorWorkspace(),
+      Blockly.getHiddenDefinitionWorkspace(),
+    ];
+
+    studentWorkspaces.forEach(workspace => {
+      if (workspace) {
+        workspace.clear();
+        workspace.getProcedureMap().clear();
```
molly-moen, line 842: will `getProcedureMap` always return a non-null value?

mikeharv: Yes, it should always have a map, even if that map's size is 0.

mikeharv: https://github.com/google/blockly/blob/develop/core/workspace.ts#L111

This is where it's initialized in the constructor.

molly-moen (approved):
Nice!

PR #59463, opened 2024-06-27
Fixing issue where locked out CPA students could not login

daynew:
Bug: Users who have account linking locked could not sign back in.
Cause: The `check_account_linking_lock` was being called for all OAuth actions (such as signing in), not just the account linking actions.
Fix: Correct a misspelling so `check_account_linking_lock` is only called for account linking actions.

## Testing story
* Create a CPA student through Google or Microsoft. Sign out. Sign in.

## Screenshot
Here is a picture of what the user would see when trying to login for a 2nd time:
image

## PR Checklist:

- [ ] Tests provide adequate coverage
- [ ] Privacy and Security impacts have been assessed
- [ ] Code is well-commented
- [ ] New features are translatable or updates will not break translations
- [ ] Relevant documentation has been added or updated
- [ ] User impact is well-understood and desirable
- [ ] Pull Request is labeled appropriately
- [ ] Follow-up work items (including potential tech debt) are tracked and linked

PR #59471, opened 2024-06-28
Validate account linking permission only when trying to link an account

daynew:
Bug: Users who have account linking locked could not sign back in.
Cause: The check_account_linking_lock was being called for all OAuth controller actions (such as signing in), not just the account linking actions.
Fix: Limit the account linking lock redirect to only when the user is attempting to link an account.

## Links
* [Original "fix"](https://github.com/code-dot-org/code-dot-org/pull/59463)  which was reverted.

## Testing story
* Added unit test which reproduces the bug if "fix" in this PR is missing.
* Manually tested on localhost (see before and after).

### Before
https://github.com/code-dot-org/code-dot-org/assets/1372238/2ecd2f75-dcaf-4ca2-b875-bbbd5dcbce38

### After
https://github.com/code-dot-org/code-dot-org/assets/1372238/2ccf0743-ad92-4c08-9b5a-369c65367cc4

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/dashboard/app/controllers/omniauth_callbacks_controller.rb
+++ b/dashboard/app/controllers/omniauth_callbacks_controller.rb
@@ -11,7 +11,7 @@ class OmniauthCallbacksController < Devise::OmniauthCallbacksController
   # TODO: figure out how to avoid skipping CSRF verification for Powerschool
   skip_before_action :verify_authenticity_token, only: :powerschool
 
-  before_action :check_account_linking_lock, on: %i[connect_provider link_accounts]
+  before_action :check_account_linking_lock
 
   # Note: We can probably remove these once we've broken out all providers
   BROKEN_OUT_TYPES = [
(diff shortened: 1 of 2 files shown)


dashboard/app/controllers/omniauth_callbacks_controller.rb
```diff
@@ -11,7 +11,7 @@ class OmniauthCallbacksController < Devise::OmniauthCallbacksController
   # TODO: figure out how to avoid skipping CSRF verification for Powerschool
   skip_before_action :verify_authenticity_token, only: :powerschool
 
-  before_action :check_account_linking_lock, on: %i[connect_provider link_accounts]
```
daynew, line 14: Note that `on:` is actually intended to be `only:`. However, because `on:` is not a valid key, this `before_action` was happening for all controller actions. That's why  `check_account_linking_lock` was called when the user was simply logging in with OAuth.

wilkie (approved):
I should let Carl also approve, but this looks great! Thank you thank you thank you!

I still find it really odd that calling this check before all actions is correct to do. It is hard to understand why the 'before' situation did what it did... what path did it take? It seems it had to have called a provider specific route with the conditions set to connect instead of sign in and getting denied and that just locked them out.

This controller is certainly worth another look in terms of refactoring, and this is such a good start. A lot cleaner already.

carl-codeorg:
Looking good; I'll approve in a bit but just taking the time to run through some LTI account linking scenarios locally

carl-codeorg (approved):
I've run through the relevant test cases for LTI account linking and everything looks good. Thanks for tackling this, Dayne!

PR #59472, opened 2024-06-28
Datablock Storage: script to migrate old projects from Firebase en-masse

snickell:
Implements a mass-migration script for converting old Firebase projects to Datablock Storage.

From `bin/migrate_firebase_projects`:
```
# Script to migrate Firebase projects to DatablockStorage en-masse
# takes a file with a channel_id per line, and migrates them.
#
# Usage:
# bundle exec bin/migrate_firebase_projects firebase-channel-ids.txt
#
# This will create multiple files that log success/failure/skips:
# - firebase-channel-ids.txt.success
# - firebase-channel-ids.txt.fail
# - firebase-channel-ids.txt.exceptions
# - firebase-channel-ids.txt.skipped
#
# Using these files, if the script is interrupted, it can be restarted
# and will not attempt to re-do any already processed channel_ids.
# Multiple copies of the script should not be run at once.
#
# Once migration to DatablockStorage is complete, the script should be deleted.
#
```


bin/migrate_firebase_projects
```diff
@@ -0,0 +1,482 @@
...
+
+    is_shared_table = FIND_SHARED_TABLES ? get_is_shared_table(table_name, records) : nil
+
+    datablock_table = {
+      project_id: project_id,
+      table_name: table_name,
+      columns: columns,
+      is_shared_table: is_shared_table,
+      created_at: Time.now,
+      updated_at: Time.now
```
cnbrenci: should we set these to the same value (i.e. now = Time.now, set both created_at and updated_at to now) so we guarantee they aren't different?

cnbrenci: yeah I think we should. I'm imagining an assumption in the future of "if the time is the same it hasn't been updated since creation" not holding true. Other than that, I think it's ready to go in!

cnbrenci (approved):
magnificent!

PR #59476, opened 2024-06-28
[Google Blockly] Allow spaces when naming variables and parameters

mikeharv:
Yesterday, during the Maze migration bug bash, @molly-moen made the following observation regarding https://studio.code.org/s/course4/lessons/16/levels/6:

> The level instructions say to add a parameter called “nectar units” but we don’t seem to allow spaces in parameter names. If you try to add a space it closes the modal.
> If you start in cdo with a parameter with spaces it does transfer over correctly, so it might be a problem with the modal.

This branch introduces changes that make it possible to add spaces for both variables and parameters on Google Blockly:
image
image
image

While investigating this, I learned:

- Both versions of Blockly support spaces in function names and variable names. This is another good reason that we should probably allow spaces. (The fact that level instructions say you can is the first good reason.)
- Spaces can already be added to function names with either version of Blockly. In CDO, you use a text field on the modal, and in Google, you use the name field on the block itself. Spaces are fine in both places.
- For parameters, CDO Blockly supports spaces because it uses another text field on the modal. However, the new Google Blockly implementation uses a modal dialog. If you press “space” while the dialog is open, it is closed immediately. :bangbang:
- Currently, spaces cannot practically be used in variables in either version of Blockly (although they are technically supported). This is because both versions of Blockly use the same prompt described above, which closes immediately upon pressing space.

It's important to understand that `Blockly.customSimpleDialog` is actually just `FeedbackUtils.showSimpleDialog` with some custom mapping of options. This dialog was designed for other purposes, such as to display the “Clear Puzzle” confirmation dialog between levels. The modal dialog it creates includes a key down handler that dismisses the modal when enter or space are pressed. In a Blockly context, this prevents spaces from being added to variable/parameter names. We want to maintain the experience for non-Blockly contexts.

As a team, we considered several options:
1. Redesigning the new parameter authoring experience to not use `Blockly.customSimpleDialog`.
2. Creating a new component that is specifically tailored for the options we need in Blockly contexts and stop using the feedback dialog.
3. **Updating the feedback dialog to support an option to ignore spacebar presses.**
4. Revising the level instructions to remove the reference to a multi-word parameter name.

We opted to start with option 3 above with the possibility of pursuing option 2 later. This is the path of least resistance that doesn't negatively impact the curriculum or student experience.

After weaving the option through, I also decided to refactor and rename the two main helpers that call `Blockly.customSimpleDialog`. Instead of having a growing list of parameters, each now uses a keyed set of properties in an options object. 

Here's a breakdown of how variables/parameters/functions are renamed or deleted. This branch only changes two scenarios, marked with *.

|Task|CDO Blockly|Google Blockly|
|-|-|-|
|Rename variable|Dialog|**Dialog***|
|Remame parameter|Modal Editor UI|**Dialog***|
|Rename function|Modal Editor UI|Block field|
|Delete variable|n\a|n\a|
|Delete parameter|Dialog|Dialog|
|Delete function|Dialog|Dialog|

In terms of spaces in names, the following illustrates where it will be supported:
||CDO Blockly|Google Blockly|
|-|-|-|
|Variable|No|**Yes***|
|Parameter|Yes|Yes|
|Function|Yes|Yes|

In order to make it so that CDO Blockly variables can use spaces, we would need to update the fork, so that is unlikely to happen.

As an added bonus to this work, Sprite Lab variable watchers have increased potential as player-facing game elements. Because a student can create a variable with spaces, they can also create a watcher with a multi-word label - without resorting to camel case:
image

## Links

Slack thread: https://codedotorg.slack.com/archives/C05HZFH7BED/p1719522385377469
Bug bash doc: https://docs.google.com/document/d/14_fjbaK7hbsvmw7O5uJwSJbC7tW9TSxQCbXxk12IO0U/edit

## Follow-up work

Decouple `Blockly.customSimpleDialog` from `FeedbackUtils.showSimpleDialog`: https://codedotorg.atlassian.net/browse/CT-665

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/blockly/addons/cdoFieldParameter.ts
+++ b/apps/src/blockly/addons/cdoFieldParameter.ts
@@ -20,6 +20,13 @@ import {ExtendedWorkspaceSvg, ProcedureBlock} from '../types';
 const RENAME_PARAMETER_ID = 'RENAME_VARIABLE_ID';
 const DELETE_PARAMETER_ID = 'DELETE_VARIABLE_ID';
 
+interface ParameterPromptOptions {
+  promptText: string; // Description text for window prompt
+  confirmButtonLabel: string; // Label of confirm button, e.g., "Rename"
+  callback: (newName: string) => void; // Callback with text of new parameter name
+  isDeleteDialog: boolean; // True for delete dialog; False for rename dialog
+  defaultText?: string; // Default input text for window prompt
+}
 export default class CdoFieldParameter extends GoogleBlockly.FieldVariable {
   /**
    * Handle the selection of an item in the parameter dropdown menu.
@@ -41,7 +48,7 @@ export default class CdoFieldParameter extends GoogleBlockly.FieldVariable {
             }),
             confirmButtonLabel: commonI18n.rename(),
             callback: this.renameSelectedParameter.bind(this),
-            prompt: true,
+            isDeleteDialog: false,
             defaultText: oldVar,
           });
           break;
@@ -53,7 +60,7 @@ export default class CdoFieldParameter extends GoogleBlockly.FieldVariable {
             }),
             confirmButtonLabel: commonI18n.delete(),
             callback: this.deleteSelectedParameter.bind(this),
-            prompt: false,
+            isDeleteDialog: true,
           });
           break;
         default:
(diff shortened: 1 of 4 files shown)


apps/src/blockly/addons/cdoFieldParameter.ts
```diff
@@ -239,31 +241,31 @@ export default class CdoFieldParameter extends GoogleBlockly.FieldVariable {
...
    */
-  static modalDialogName = function (
-    promptText: string,
-    confirmButtonLabel: string,
-    callback: (newName: string) => void,
-    prompt: boolean,
-    isDangerCancel: boolean,
-    defaultText?: string
-  ) {
+  static parameterPrompt = function (options: {
```
molly-moen: mind making a type for the options object? Up to you if you want to keep the javadoc above since the parameters would be documented in a type.

apps/src/blockly/addons/cdoFieldParameter.ts
```diff
@@ -239,31 +241,31 @@ export default class CdoFieldParameter extends GoogleBlockly.FieldVariable {
...
-      bodyText: promptText,
-      prompt,
-      promptPrefill: defaultText,
-      cancelText: confirmButtonLabel,
-      isDangerCancel,
+      bodyText: options.promptText,
+      prompt: options.prompt,
+      promptPrefill: options.defaultText,
+      cancelText: options.confirmButtonLabel,
+      isDangerCancel: !options.prompt,
```
molly-moen: why can we make this inference?

apps/src/blockly/addons/cdoFieldParameter.ts
```diff
@@ -239,31 +241,31 @@ export default class CdoFieldParameter extends GoogleBlockly.FieldVariable {
...
       confirmText: commonI18n.cancel(),
       onConfirm: null,
-      onCancel: callback,
+      onCancel: options.callback,
+      disableSpaceClose: !!prompt,
```
molly-moen: same question, why can we make this inference?

molly-moen (commented):
Looks good, couple comments/questions!

apps/src/blockly/addons/cdoFieldParameter.ts
mikeharv: We basically have two options: rename (a prompt for a name) and delete (not a prompt). If it's not a prompt, we want the delete button to be red ("danger"). 
If it's a rename prompt, we want to disable closing on space, but if it's a delete we can let it do the default.

mikeharv: @molly-moen I wonder if it the inferences would make more sense if I chose a different name like `isDelete` for this option?

mikeharv: I pushed a change with that idea. Let me know if it feels clearer. I can also add comments. 
I was trying to avoid having multiple mutually exclusive boolean props since there's really only two scenarios we're supporting. The variable dialogs are simpler because they're both renaming prompts.

molly-moen: That's much clearer! I was having a hard time figuring out the assocation between prompt and this being a delete dialog.

molly-moen (approved):
Nice!

PR #59481, opened 2024-06-28
AI Chat: design polish

sanchitmalhotra126:
Some more design tweaks & polish:
- rounded corners & light grey border for panels
- white background for presentation view
- some scrollbar tweaks
- padding around entire lab

see screenshots at end of description

## Testing story

Tested on a few AI Chat levels with varying instructions, tabs, etc.

Also verified that this doesn't impact Music Lab (due to changes in PanelContainer and Instructions)

## Follow-up work

A few items on my mind
- Right now, some panels have a fixed height to make sure they constrain to their container but are still scrollable. It would be ideal for the height to automatically grow into the available space and then become scrollable beyond that, but that's a little tricky to implement (which is why we used fixed height for now).
- As per figma mocks, panel headers should be grey with black text. However I think this might require including typography styles in the panel headers, which will affect Music Lab, so I want to split that out into a separate change.
- Working on a separate change to update the chat bubble styles with the intention of creating shared reusable components.

Screenshots:

**Standard Level**
Screenshot 2024-06-28 at 10 52 19 AM

**With Edit/Presentation Toggle**
Screenshot 2024-06-28 at 10 52 42 AM

**Long instructions (scrollable)**
Screenshot 2024-06-28 at 10 51 33 AM

**Presentation View**
Screenshot 2024-06-28 at 10 53 06 AM

**Presentation View expanded (scrollable)**
Screenshot 2024-06-28 at 10 54 32 AM

**With ongoing chat (scrollable)**
Screenshot 2024-06-28 at 10 58 13 AM


apps/src/aichat/views/presentation/presentation-view.module.scss
```diff
@@ -1,9 +1,13 @@
 @import 'color';
 
 .modelCardTitle {
-    background-color: $neutral_light;
```
fisher-alice, line 4: Would you mind confirming that @markabarrett that he wants a white background for the model card title in presentation view? I don't see title included the figma mockups (I may be missing this). Thanks!

fisher-alice (approved):
Thanks for taking care of this! Just left a question for Mark to confirm background color of presentation view title.

apps/src/aichat/views/presentation/presentation-view.module.scss
sanchitmalhotra126: Thanks for catching this! Going to merge this is as-is for now and follow-up with Mark on further tweaks once he's back.

PR #59482, opened 2024-06-28
Lab2: flag to hide extra links button

sanchitmalhotra126:
Small change to enable hiding the Extra Links button via a query param (`?hide-extra-links=true`). This is useful mostly for UI/design work where it's helpful to hide the button for screenshots, etc.

### DIFF
--- a/apps/src/lab2/views/LabViewsRenderer.tsx
+++ b/apps/src/lab2/views/LabViewsRenderer.tsx
@@ -29,6 +29,7 @@ import ExtraLinks from './ExtraLinks';
 import {useAppSelector} from '@cdo/apps/util/reduxHooks';
 import {getAppOptionsViewingExemplar} from '../projects/utils';
 import NoExemplarPage from './components/NoExemplarPage';
+import {queryParams} from '@cdo/apps/code-studio/utils';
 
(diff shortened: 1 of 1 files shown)


molly-moen (approved):
Nice!

fisher-alice (approved):
What a great idea! So helpful for PR screenshots.

breville:
I think it might also be useful to have a one-click way to dismiss it, the same way a single click on the `developer` text in the upper-right corner of the window dismisses it.

molly-moen:
> I think it might also be useful to have a one-click way to dismiss it, the same way a single click on the `developer` text in the upper-right corner of the window dismisses it.

@breville how would we implement that without getting in the way of levelbuilders clicking the button to open the modal? We wouldn't want the button to disappear for them.

breville:
> > I think it might also be useful to have a one-click way to dismiss it, the same way a single click on the `developer` text in the upper-right corner of the window dismisses it.
> 
> @breville how would we implement that without getting in the way of levelbuilders clicking the button to open the modal? We wouldn't want the button to disappear for them.

It's a good question.  Maybe we could have a small option to permanently dismiss the button (until page reload) inside the modal?

PR #59485, opened 2024-06-28
Add join table for workshops and course offerings

TurnerRiley:
Creates a join table between `pd_workshops` and `course_offerings` to represent the PL workshop topics for "Build your own workshops" (where `course_offerings` will be the representation of a given topic for the PL workshop). This relation will be many-to-many because multiple PL workshops can have the same topic and a given PL workshop can have multiple topics. To be clearer about the table's use,

## Links
Jira ticket: [here](https://codedotorg.atlassian.net/jira/software/c/projects/ACQ/boards/64?assignee=60d62161f65054006980bd71&selectedIssue=ACQ-2021)
Spec: [Step 2C in the doc](https://docs.google.com/document/d/16IsvtmDFDIDzjn3vv1bIskRkET9ivC_l9LyUXpN0Fno/edit#heading=h.4schd43y1qoc)

## Testing story
Successfully migrated locally:
success creates

After migration, table shows up in dashboard db:
shows in db

### DIFF
--- a/dashboard/db/migrate/20240628202031_create_join_table_workshop_course_offering.rb
+++ b/dashboard/db/migrate/20240628202031_create_join_table_workshop_course_offering.rb
@@ -1,0 +1,9 @@
+class CreateJoinTableWorkshopCourseOffering < ActiveRecord::Migration[6.1]
+  def change
+    create_join_table :pd_workshops, :course_offerings, table_name: :pl_workshop_topics do |t|
+      t.index :pd_workshop_id
+      t.index :course_offering_id
+      t.timestamps
+    end
+  end
+end
--- a/dashboard/db/migrate/20240628202031_create_join_table_workshop_course_offering.rb
+++ b/dashboard/db/migrate/20240628202031_create_join_table_workshop_course_offering.rb
@@ -1,9 +1,0 @@
-class CreateJoinTableWorkshopCourseOffering < ActiveRecord::Migration[6.1]
-  def change
-    create_join_table :pd_workshops, :course_offerings, table_name: :pl_workshop_topics do |t|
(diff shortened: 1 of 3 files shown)


dashboard/db/migrate/20240628202031_create_join_table_workshop_course_offering.rb
```diff
@@ -0,0 +1,9 @@
+class CreateJoinTableWorkshopCourseOffering < ActiveRecord::Migration[6.1]
+  def change
+    create_join_table :pd_workshops, :course_offerings, table_name: :pl_workshop_topics do |t|
```
bethanyaconnor: This is not something I'm going to be a stickler about, but I am curious why you didn't like the default `course_offerings_workshops` table name? Rails really favors convention, so I'm nervous when we change it.

I would support, when we add the Rails model associations, doing something like
```
has_and_belongs_to_many :topics, join_table: course_offerings_workshops
```
in pd_workshops.rb, if that would make things clearer!

bethanyaconnor: (that syntax will need to be tweaked ... will also need to include the `class_name` at the very least. [Doc](https://api.rubyonrails.org/v7.1.3.4/classes/ActiveRecord/Associations/ClassMethods.html#method-i-has_and_belongs_to_many))

TurnerRiley: Oh I just wasn't sure if it wouldn't be clear or not. That sounds good to me, I'll update this to not rename!

bethanyaconnor (approved):
Thanks for updating!

PR #59490, opened 2024-06-28
allow for multiple images

kobryan0619:
This update to the upload image page will enable levelbuilders to batch upload multiple files at once and get all the corresponding images.code.org URLS

After picking files but before uploading:
image

After uploading:
image

## Follow-up work

- Add a "Copy URL" button to the right of each URL to make it easier for levelbuilders to take these URLs and use them
-  Build a page that allows the management of these files by levelbuilders so that they won't lose URLs and can reuse images more easily

## PR Checklist:

- [ ] Code is well-commented


apps/src/lib/levelbuilder/lesson-editor/uploadImage.module.scss
```diff
@@ -1,6 +1,15 @@
+.imageGrid {
+  display: grid;
+  grid-template-columns: repeat(
+    auto-fill,
+    minmax(100px, 1fr)
+  ); /* Adjust based on your needs */
```
snickell: Not important but not sure what "your needs" here refers to

apps/src/lib/levelbuilder/lesson-editor/UploadImageForm.jsx
```diff
@@ -26,18 +27,22 @@ export default function UploadImageForm() {
...
     const formData = new FormData();
-    formData.append('file', e.target.files[0]);
-    setFormDataForImage(formData);
-    setTempImageUrl(URL.createObjectURL(e.target.files[0]));
+    let tempImageUrlList = [];
+    for (let i = 0; i < e.target.files.length; i++) {
+      formData.append('file', e.target.files[i]);
+      tempImageUrlList.push(URL.createObjectURL(e.target.files[i]));
+    }
+    setFormDataForImages(formData);
```
snickell: Right now we are storing a list of image files in a FormData object stored in component state (`[formDataForImages, setFormDataForImages]`), but we actually don't use this object as FormData, we parse it as a list and create sub-form datas from it. Given how its used, all else being equal, I would prefer that this was stored as an array of File (`File[]`) rather than as form data that coincidentally only has Files as members.

apps/src/lib/levelbuilder/lesson-editor/UploadImageForm.jsx
```diff
@@ -50,9 +55,17 @@ export default function UploadImageForm() {
       });
   };
 
+  const saveImagesToS3 = () => {
+    for (const image of formDataForImages.entries()) {
```
snickell: Mentioned above, but since we're just looping over the `formDataForImages` to create the `new FormData()` we actually submit below, perhaps instead of `formDataForImages` it should be `listOfImageFiles` or such. I don't hate this way, but it does slightly obscure what's happening.

snickell (approved):
Yay!

For bonus points in a followup PR it might be really nice UX wise to make the image URLs copyable by clicking a little button next to them (or just clicking the URL itself is a nice UX but less discoverable... gives a big click target, and as long as you indicate what happened its pretty discoverable).

apps/src/lib/levelbuilder/lesson-editor/UploadImageForm.jsx
snickell: Want to add that if you didn't feel like changing this, I would not look askance at all.

wilkie:
Looks fairly nice. You are uploading one file at a time, but I do worry about errors as you batch these uploads. For instance, what happens if one of your images you are uploading is larger than 2MB, as this would return an error status? Would you know which were uploaded and which errored?

You may want an array of errors alongside your array of URLs so you can render each error in the batch.

kobryan0619:
> Looks fairly nice. You are uploading one file at a time, but I do worry about errors as you batch these uploads. For instance, what happens if one of your images you are uploading is larger than 2MB, as this would return an error status? Would you know which were uploaded and which errored?
> 
> You may want an array of errors alongside your array of URLs so you can render each error in the batch.

Good catch @wilkie - this was a test case we didn't try ourselves!  After testing it out, I think it still visually shows the user which ones were uploaded and which ones had an error.  Below is a screenshot where two of the 4 images were larger than 2MB. 

image

snickell:
❤️ the clipboard button

PR #59492, opened 2024-06-28
P20-1019: Add `providers` param to CAP Account Settings widget events

artem-vavilov:
## Links
- JIRA ticket: [P20-1019](https://codedotorg.atlassian.net/browse/P20-1019)

### DIFF
--- a/dashboard/app/views/devise/registrations/edit.html.haml
+++ b/dashboard/app/views/devise/registrations/edit.html.haml
@@ -149,6 +149,7 @@
       permission_status: @permission_status,
       user_email: current_user.hashed_email,
       in_section: JSON.dump(current_user.student? ? current_user.sections_as_student.present? : nil),
+      providers: JSON.dump(current_user.providers.filter_map(&:presence)),
     },
   }
 
--- a/apps/src/sites/studio/pages/devise/registrations/edit.js
+++ b/apps/src/sites/studio/pages/devise/registrations/edit.js
@@ -128,6 +128,9 @@ $(document).ready(() => {
         inSection={JSON.parse(
           lockoutLinkedAccountsMountPoint.getAttribute('data-in-section')
         )}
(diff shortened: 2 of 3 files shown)


daynew (approved):

dashboard/app/views/devise/registrations/edit.html.haml
```diff
@@ -149,6 +149,7 @@
       permission_status: @permission_status,
       user_email: current_user.hashed_email,
       in_section: JSON.dump(current_user.student? ? current_user.sections_as_student.present? : nil),
+      providers: JSON.dump(current_user.providers.filter_map(&:presence)),
```
artem-vavilov, line 152: https://github.com/code-dot-org/code-dot-org/blob/adae83a09e532c6849c051a34caf5e2fb111d834/dashboard/app/models/user.rb#L2642-L2648

wilkie: I believe this means you will have a `'sponsered'` provider for teacher-made secret word / password accounts. That works to cover that case, too! nice!

wilkie: Wait, I'm wrong because a normal secret word/picture `sponsored` account would still be `migrated` but have an empty list of providers. Is it just a lack of providers that you look for in that case?

wilkie: `current_user.providers.append(current_user.sponsored? && 'sponsored').filter_map(&:presence),` 🤭

wilkie (approved):

dashboard/app/views/devise/registrations/edit.html.haml
artem-vavilov: Doesn't the absence of the providers mean the account is "sponsored" (word/picture)?

PR #59501, opened 2024-07-01
Tidy Projects: remove unused canShare prop from PersonalProjectsTable + tendrils

Erin007:
There was a big effort (https://github.com/code-dot-org/code-dot-org/pull/57473) to remove the ability for students to publish projects to a public gallery for safety and customer support time management (see [spec](https://docs.google.com/document/d/1vU68tzXG72Z80MvM6xYEDwF5hLDf7V097P-0dDfpouc/edit)). When that work was done there was a lingering prop `canShare` in the `PersonalProjectsTable` that was no longer being used.  I removed the `canShare` prop from the component, the parent component, the js + haml entrypoint to the gallery, and the corresponding test and storybook entries. 

## Follow-up work

There is additional clean up work to continue removing references to "publishing" e.g. https://codedotorg.atlassian.net/browse/LABS-574. This felt like a small, contained start to cleaning up that space that I could take on in the time I had without needing to consult with other engineers or Product. 😁 

## PR Checklist:

- [ ] Tests provide adequate coverage
- [ ] Privacy and Security impacts have been assessed
- [ ] Code is well-commented
- [ ] New features are translatable or updates will not break translations
- [ ] Relevant documentation has been added or updated
- [ ] User impact is well-understood and desirable
- [ ] Pull Request is labeled appropriately
- [ ] Follow-up work items (including potential tech debt) are tracked and linked

PR #59506, opened 2024-07-01
Tidy Projects: update ProjectsGallery switcher 

Erin007:
More Project Gallery changes on my tidying rampage 😅 (https://github.com/code-dot-org/code-dot-org/pull/59501)

I modernized the `ProjectsGallery` component in a two ways: 
- the component now uses TypeScript
- I swapped out the old gallery toggles for the `SegmentedButton` component from the component library
    - one less usage of Radium 🎉   
    - no more in-line styles
    - more consistency across the site 

BEFORE 

https://github.com/code-dot-org/code-dot-org/assets/12300669/e17246c5-844c-4eb1-b762-3b8097626cf2

AFTER 

https://github.com/code-dot-org/code-dot-org/assets/12300669/0d6d482a-554a-47d0-8072-a7f88b0b551c

## PR Checklist:

- [ ] Code is well-commented


apps/src/templates/projects/GallerySwitcher.jsx
Erin007, line 1: Deleting the empty file, left over from this change: https://github.com/code-dot-org/code-dot-org/commit/74652d43c3c9039409141aa2424f34b2eac82669

apps/src/templates/projects/ProjectsGallery.tsx
```diff
@@ -0,0 +1,78 @@
...
+        tabContent: <PublicGallery limitedGallery={limitedGallery} />,
+      },
+    ];
+    return tabs;
+  }, [limitedGallery]);
+
+  const handleOnChange = useCallback(
+    (value: string) => {
+      setSelectedTab(value);
+      selectGallery(value);
```
molly-moen: why don't we use the redux value `selectedGallery` rather than the state value?

molly-moen (commented):
One question about state vs redux, otherwise looks good!

apps/src/templates/projects/ProjectsGallery.tsx
```diff
@@ -0,0 +1,78 @@
...
+  limitedGallery: boolean;
+}
+
+const ProjectsGallery: React.FunctionComponent<ProjectsGalleryProps> = ({
+  limitedGallery,
+}) => {
+  const [selectedTab, setSelectedTab] = useState(Galleries.PRIVATE);
+
+  const galleryTabs = useMemo(() => {
+    const tabs = [
```
thomasoniii: super minor style thing, but just noting that this could get smooshed into just

```
useMemo(() => ([ /* that tabs array */ ]) )
```

and spare the declared variable and explicit return.

thomasoniii (approved):
lgtm 🚀

apps/src/templates/projects/ProjectsGallery.tsx
Erin007: Can I do that without converting the projects redux to typescript? I got muddled with connect v selectors and didn't want to have to define the full type of projects state. But maybe I'm making it harder than it needs to be...?

molly-moen: You can manually type it as part of the call to useSelector, something like:
```
useSelector((state: {projects: {selectedGallery: string}) => state.projects.selectedGallery)
```
You only need to define the part of the state type you are using.

Erin007: aha! Thank you!

apps/src/templates/projects/ProjectsGallery.tsx
```diff
@@ -18,7 +20,12 @@ interface ProjectsGalleryProps {
 const ProjectsGallery: React.FunctionComponent<ProjectsGalleryProps> = ({
   limitedGallery,
 }) => {
-  const [selectedTab, setSelectedTab] = useState(Galleries.PRIVATE);
+  const dispatch = useAppDispatch();
+
+  const selectedGallery = useSelector(
```
molly-moen, line 26: nit: mind adding a comment that this redux store isn't typed yet, which is why we can't use `useAppSelector`?

molly-moen (approved):
Nice!

fisher-alice:
Thanks so much for the refactor! 🎉

PR #59457, opened 2024-06-27
Adding Build Your Own Workshop to workshop creation page

hannahbergam:
This PR adds the new option, and ensures that the subject dropdown and 'Is this a Code.org paid workshop?' dropdowns don't appear when it's selected. 

This also means it shows up on the course list for Facilitator permissions:

Screenshot 2024-06-27 at 11 42 50 AM

Screenshot 2024-06-27 at 11 43 04 AM

and I can then add that account from the facilitator dropdown:
Screenshot 2024-06-27 at 11 50 48 AM

I will add unit tests for the workshop dashboard in a followup PR, but here is the photo evidence for now:

Screenshot 2024-06-27 at 11 31 14 AM

(to see the fields when a different course is selected)
Screenshot 2024-06-27 at 11 31 43 AM

Screenshot 2024-06-27 at 11 31 49 AM

Additionally, I verified that a user can create a workshop with that course selected:
Screenshot 2024-06-27 at 1 53 01 PM

## Links

- spec: []()
- jira ticket: https://codedotorg.atlassian.net/browse/ACQ-2023
- https://codedotorg.atlassian.net/browse/ACQ-2018

## PR Checklist:

- [ ] Tests provide adequate coverage
- [ ] Privacy and Security impacts have been assessed
- [ ] Code is well-commented
- [ ] New features are translatable or updates will not break translations
- [ ] Relevant documentation has been added or updated
- [ ] User impact is well-understood and desirable
- [ ] Pull Request is labeled appropriately
- [ ] Follow-up work items (including potential tech debt) are tracked and linked

PR #59510, opened 2024-07-01
Turn off all email sends for workshop if 'Build Your Own' selected

TurnerRiley:
This PR suppresses emails sent for 'Build Your Own' workshops. I originally followed the pattern used for emails that can be suppressed where the mailer function returns before actually sending the email if the given condition is met ([example](https://github.com/code-dot-org/code-dot-org/blob/staging/dashboard/app/mailers/pd/workshop_mailer.rb#L121)). Since the spec says, "turn off all email sends", I thought it was calling for no emails to be sent relating to 'Build Your Own' workshops. However, the phrase "(but the field can stay on the create page)" and [this discussion with Hannah](https://docs.google.com/document/d/16IsvtmDFDIDzjn3vv1bIskRkET9ivC_l9LyUXpN0Fno/edit?disco=AAABQUJsroU) makes me think we want to just suppress reminder emails the same way they would be if the facilitator set that email field to `suppress_email = true`.

Assuming this is the case, I thought it would make more sense to have a solution similar to [Hannah's PR](https://github.com/code-dot-org/code-dot-org/pull/59457/files) that ensured the `funded` workshop form field doesn't show up when 'Build Your Own' workshop is selected. This way, `suppress_email` is set to `true` and the field for it doesn't show up at all if the user selects 'Build Your Own'.

If we actually do want to suppress all emails related to 'Build Your Own' workshops, I can do a follow-up PR (while still keeping this functionality so that the form doesn't show the email question for 'Build Your Own' workshops).

### 'Build Your Own' doesn't show the suppress emails question
no email

### Other courses still show the suppress emails question
shows_emails

## Links
Jira ticket: [here](https://codedotorg.atlassian.net/jira/software/c/projects/ACQ/boards/64?assignee=60d62161f65054006980bd71&selectedIssue=ACQ-2025)
Spec: [item 6](https://docs.google.com/document/d/16IsvtmDFDIDzjn3vv1bIskRkET9ivC_l9LyUXpN0Fno/edit#heading=h.4schd43y1qoc)
PR that adds 'Build Your Own' workshops: [here](https://github.com/code-dot-org/code-dot-org/pull/59457)

## Testing story
Local testing and updating a unit test.

### DIFF
--- a/apps/src/code-studio/pd/workshop_dashboard/components/workshop_form.jsx
+++ b/apps/src/code-studio/pd/workshop_dashboard/components/workshop_form.jsx
@@ -490,28 +490,30 @@ export class WorkshopForm extends React.Component {
               <HelpBlock>{validation.help.virtual}</HelpBlock>
             </FormGroup>
           </Col>
-          <Col sm={5}>
-            <FormGroup validationState={validation.style.suppress_email}>
-              <ControlLabel>
-                Enable workshop reminders?
-                <HelpTip>
-                  <p>
-                    Choose if you'd like automated 10-day and 3-day pre-workshop
-                    reminders to be sent to your participants.
-                  </p>
-                </HelpTip>
-              </ControlLabel>
-              <SelectSuppressEmail
-                onChange={this.handleSuppressEmailChange}
-                value={this.state.suppress_email || false}
-                readOnly={
-                  this.props.readOnly ||
-                  MustSuppressEmailSubjects.includes(this.state.subject)
-                }
-              />
-              <HelpBlock>{validation.help.suppress_email}</HelpBlock>
-            </FormGroup>
-          </Col>
+          {!isBuildYourOwnWorkshop && (
+            <Col sm={5}>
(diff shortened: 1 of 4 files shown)


hannahbergam (commented):
@dmcavoy I know your spec says 'turn off email reminders (No 10 day, 3 day, or post workshop email)'. Should we also be turning off workshop change notifications or and enrollment confirmations?

TurnerRiley:
> @dmcavoy I know your spec says 'turn off email reminders (No 10 day, 3 day, or post workshop email)'. Should we also be turning off workshop change notifications or and enrollment confirmations?

Oh good point! The intention may have only been for the emails that can be suppressed with the user's choice on the form. That makes more sense with then hiding that field. I'll update this to only the emails that are suppressed and if we in fact want to shut off all emails I can retroactively add that in after!

hannahbergam (approved):
YESSSS so elegant so clean. Thank you!

PR #58887, opened 2024-05-23
Enable `import/order` for Applab Directories

Hamms:
Follow-up to https://github.com/code-dot-org/code-dot-org/pull/57247

For just representing a single one of our app types, this is a relatively big one! As usual, please check for any manual ordering logic or comments that need to be preserved.

## PR Checklist:

- [ ] Tests provide adequate coverage
- [ ] Privacy and Security impacts have been assessed
- [ ] Code is well-commented
- [ ] New features are translatable or updates will not break translations
- [ ] Relevant documentation has been added or updated
- [ ] User impact is well-understood and desirable
- [ ] Pull Request is labeled appropriately
- [ ] Follow-up work items (including potential tech debt) are tracked and linked

PR #59511, opened 2024-07-01
Add somre more folders to eslint import order rules

lfryemason:

## PR Checklist:

- [ ] Code is well-commented


Hamms (commented):
Do the circular dependency errors you were seeing represent actual new circular dependencies, or are they just the result of the existing circular dependencies showing up in a different order? See https://github.com/code-dot-org/code-dot-org/pull/58887#issuecomment-2135327537 for an example

lfryemason:
> Do the circular dependency errors you were seeing represent actual new circular dependencies, or are they just the result of the existing circular dependencies showing up in a different order? See [#58887 (comment)](https://github.com/code-dot-org/code-dot-org/pull/58887#issuecomment-2135327537) for an example

They are because the order of imports change. I assume this is because the existing circular dependencies don't cause problems when they are imported in a specific order, but this would take time to confirm.

Hamms:
I think it's worth taking that time! Updating the circular dependencies tracker at `apps/circular_dependencies.json` should be pretty straightforward

lfryemason:
> circular_dependencies

I didn't know about this `circular_dependencies.json` list. I removed the other places I disabled the import rule and also added some clarifying text to the output so it is easier to find for the next person.

apps/webpack.config.js
```diff
@@ -109,7 +109,7 @@ const nodePolyfillConfig = {
         if (!circularDependenciesSet.has(pathString)) {
           compilation.errors.push(
             new Error(
-              `Circular Dependency Checker : New Circular Dependency found : ${pathString}`
+              `Circular Dependency Checker : A new Circular Dependency found.\nKnown circular dependencies can be found in 'apps/circular_dependencies.json'\n Circular dependency: ${pathString}.`
```
Hamms, line 112: :+1:

Hamms (approved):
Yeah, this circular dependencies functionality is relatively new. Thanks for taking the time to make it more discoverable!

This changeset LGTM, once that merge conflict gets resolved. Thank you!

PR #59512, opened 2024-07-01
Predict Levels: lock answer and run button appropriately

molly-moen:
This PR adds the functionality to lock predict answers and enable/disable the run button appropriately. The logic is:
If the level allows multiple tries:
- The run button becomes enabled once the user has provided an answer (there is some text for a free response level, or a multiple choice option has been selected)

If the level does not allow multiple tries:
- Same as above for the run button becoming enabled.
- Once the user clicks run, the answer becomes "locked", i.e. the user cannot edit it. This is true after reloading the page/switching levels as well.

Example screen recording for a free response level that does not allow multiple tries:

https://github.com/code-dot-org/code-dot-org/assets/33666587/177adcbc-5281-484a-9b86-4fbca810c245

As part of this work I did a little redux refactoring. Now that there are 2 predict settings, a selector, and an extra reducer, I made a `predictLevel` slice. I also converted the `sendPredictLevelReport` method into an official `asyncThunk` in order to respond to its `fulfilled` action in the predict level slice. 

## Links

- jira ticket: [CT-638](https://codedotorg.atlassian.net/browse/CT-638)
- For reference: [Redux docs on extraReducers](https://redux-toolkit.js.org/api/createSlice#extrareducers)

## Testing story
Tested locally. These changes should be isolated to Python Lab as it's the only lab using the new predict levels.

## Follow-up work
See the spec linked above for the rest of the planned work on this feature.

## PR Checklist:

- [ ] Code is well-commented


apps/src/lab2/redux/predictLevelRedux.ts
```diff
@@ -0,0 +1,62 @@
...
+
+// REDUCER
+const predictSlice = createSlice({
+  name: 'predictLevel',
+  initialState,
+  reducers: {
+    setPredictResponse(state, action: PayloadAction<string>) {
+      state.response = action.payload;
+    },
+    setHasSubmittedPredictResponse(state, action: PayloadAction<boolean>) {
```
ebeastlake: Non-blocking nit: Would it make sense to simplify these names, like `setResponse` and `setSubmitted`? (I'm not attached to this, and I'm not even sure it's better. I wonder if `setHasSubmittedPredictResponse` is getting into "hard to parse" territory.)

apps/src/codebridge/ControlButtons/index.tsx
```diff
@@ -27,6 +27,13 @@ const ControlButtons: React.FunctionComponent = () => {
   const hasNextLevel = useAppSelector(
     state => nextLevelId(state) !== undefined
   );
+  const hasPredictResponse = useAppSelector(
+    state => !!state.predictLevel.response
+  );
+  const isPredictLevel = useAppSelector(
+    state => state.lab.levelProperties?.predictSettings?.isPredictLevel
+  );
+  const disableRunAndTest = loading || (isPredictLevel && !hasPredictResponse);
```
ebeastlake, line 36: Can you explain why we don't need to consider whether `predictAnswerIsLocked` here?

apps/src/lab2/redux/predictLevelRedux.ts
```diff
@@ -0,0 +1,62 @@
...
+  },
+  extraReducers: builder => {
+    builder.addCase(sendPredictLevelReport.fulfilled, state => {
+      state.hasSubmittedResponse = true;
+    });
+    builder.addCase(setLoadedPredictResponse, (state, action) => {
+      // We have loaded a predict response from the server.
+      // If the response is not empty, we consider the user to have submitted a response.
+      state.response = action.payload;
+      state.hasSubmittedResponse = !!action.payload;
```
ebeastlake, line 51: Do we have validation on submitting something like a series of spaces somewhere? Could students lock themselves out of a predict level by submitting empty text, and would we want to protect against that? (Maybe we don't want to.)

ebeastlake (approved):
A couple small questions, but generally, this looks good to me!

apps/src/lab2/redux/predictLevelRedux.ts
molly-moen: I want the names to be understandable when importing into another file, so I think I want to keep `setPredictResponse` as-is. Looking at the code again, I actually am not using `setHasSubmittedPredictResponse` anywhere because that value is always set via on of the extra reducers. So I'll just delete that one :)

apps/src/codebridge/ControlButtons/index.tsx
molly-moen: The answer can't be locked if we don't have a predict response.

apps/src/lab2/redux/predictLevelRedux.ts
molly-moen: In our old version of predict levels, a student can submit a bunch of spaces and it will lock the answer. So for starters I think we can reproduce that behavior, but see if we want to improve on it as a follow up.

apps/src/codebridge/ControlButtons/index.tsx
ebeastlake: Ah, right. Okay. It seems like this disables the "run and test" button on predict levels _before_ you've submitted a response. But you can continue running and testing once you've submitted a response (even if it's un-editable). Thanks!

ebeastlake:
Thanks for addressing questions, @molly-moen!

PR #59513, opened 2024-07-01
Calculate CPA delete date from lockout date instead of created_at

carl-codeorg:
The delete date on the lockout page was being calculated based on created_at + 7 days, because during phase 1 the rules only applied to newly created accounts. This changes the controller to calculate the date from the lockout date instead.

## Links

[JIRA](https://codedotorg.atlassian.net/browse/P20-1023)

## Testing story

Tested locally. Example user has:
- `created_at: "2024-06-27 01:50:10.000000000 +0000"`
- `"child_account_compliance_lock_out_date"=>"2024-07-01T11:49:22.800-07:00"`

Screenshot 2024-07-01 at 12 01 21 PM

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/dashboard/app/controllers/sessions_controller.rb
+++ b/dashboard/app/controllers/sessions_controller.rb
@@ -74,7 +74,7 @@ class SessionsController < Devise::SessionsController
     end
 
     # Determine the deletion date as the lockout date of the account + 7 days
-    @delete_date = Date.parse(current_user.child_account_compliance_lock_out_date).since(7.days)
+    @delete_date = DateTime.parse(current_user.child_account_compliance_lock_out_date).since(7.days)
 
     # Find any existing permission request for this user
     # Students might have issued a few requests. We render the latest one.
--- a/dashboard/app/controllers/sessions_controller.rb
+++ b/dashboard/app/controllers/sessions_controller.rb
@@ -73,8 +73,8 @@ class SessionsController < Devise::SessionsController
       @disallowed_email = current_user.hashed_email
     end
 
-    # Determine the deletion date as the creation time of the account + 7 days
-    @delete_date = current_user.created_at.since(7.days)
+    # Determine the deletion date as the lockout date of the account + 7 days
+    @delete_date = Date.parse(current_user.child_account_compliance_lock_out_date).since(7.days)
 
     # Find any existing permission request for this user
     # Students might have issued a few requests. We render the latest one.
--- a/dashboard/test/controllers/sessions_controller_test.rb
+++ b/dashboard/test/controllers/sessions_controller_test.rb
@@ -333,6 +333,7 @@ class SessionsControllerTest < ActionController::TestCase
 
     before do
       Policies::ChildAccount::ComplianceState.stubs(:locked_out?).with(user).returns(user_is_locked_out)
+      user.update(child_account_compliance_lock_out_date: DateTime.now)
 
       sign_in user
     end
@@ -347,9 +348,9 @@ class SessionsControllerTest < ActionController::TestCase
       _(assigns[:disallowed_email]).must_equal ''
     end
 
-    it 'assigns @delete_date with 7 day after user creation' do
+    it 'assigns @delete_date with 7 day after user lockout date' do
       get :lockout
-      _(assigns[:delete_date]).must_equal user.created_at.since(7.days)
+      _(assigns[:delete_date]).must_equal DateTime.parse(user.child_account_compliance_lock_out_date).since(7.days)
     end
 
     it 'renders lockout page' do


dashboard/app/controllers/sessions_controller.rb
```diff
@@ -73,8 +73,8 @@ def lockout
       @disallowed_email = current_user.hashed_email
     end
 
-    # Determine the deletion date as the creation time of the account + 7 days
-    @delete_date = current_user.created_at.since(7.days)
+    # Determine the deletion date as the lockout date of the account + 7 days
+    @delete_date = Date.parse(current_user.child_account_compliance_lock_out_date).since(7.days)
```
stephenliang: Is it possible for `child_account_compliance_lock_out_date` to be nil? `Date.parse` will raise an error if so

wilkie: They should have a lockout date if they are rendering the lockout page. It is set when their lockout state is set, which is what the ComplianceState.locked_out? is looking for.

carl-codeorg: They should be redirected away if they aren't actually locked out, which should always be accompanied by having a `child_account_compliance_lock_out_date`. I think it might be possible to get it to be nil if their lockout status was set manually instead of using the lock_out method? But in the wild that shouldn't happen.

wilkie: The only problem is that `Date.parse` drops the time and timezone information and I have to think about whether or not this is a problem.

wilkie: You may want `DateTime.parse` but I'm not entirely sure... but that aligns more with the original `created_at.since` which would give you a `DateTime` and not just a `Date`

```
[development] dashboard > Date.parse(u.child_account_compliance_lock_out_date).iso8601
=> "2024-06-30"
[development] dashboard > DateTime.parse(u.child_account_compliance_lock_out_date).iso8601
=> "2024-06-30T16:52:58-07:00"
```

carl-codeorg: We could also do a DateTime, but I figured we only need date level precision, and it's an approximation for display purposes, not the actual mechanism for deciding when to delete

wilkie: Yep, just requires me to lose a few braincells to think about timezone implications, since it technically introduces them.

I think it is ok because the timezone as recorded (in mine it is -7) is the same that they want to actually see. Tested with a timestamp that would be the next day in UTC and it parsed as expected to local time just fine.

So, all good!

carl-codeorg: @wilkie it looks like the panel is [setting it to UTC](https://github.com/code-dot-org/code-dot-org/blob/staging/apps/src/templates/sessions/LockoutPanel.tsx#L250). The lockout date appears to be set based on server time, which is likely US eastern time. Should we make any changes here, or would you rather I just change it to use a DateTime, and preserve the previous logic as much as possible?

wilkie: It's probably safest and most dogmatic to just push the most accurate information (`DateTime`) out and decide on its actual representation on the front-end.

Looks like you need to set the user property in the controller test. easy peasy.

dashboard/test/controllers/sessions_controller_test.rb
```diff
@@ -333,6 +333,7 @@ class LtiAccountLinkingSignInPageTest < ActionDispatch::IntegrationTest
 
     before do
       Policies::ChildAccount::ComplianceState.stubs(:locked_out?).with(user).returns(user_is_locked_out)
+      user.update(child_account_compliance_lock_out_date: DateTime.now)
```
wilkie, line 336: You can also do (above):

```
let(:user) {create(:parent_managed_student, child_account_compliance_lock_out_date: DateTime.now.iso8601)}
```

wilkie: I do think regardless that the field needs to be a string in order to be parsable. hmm. Like, `DateTime.parse(DateTime.now)` I think errors out.

wilkie: You would need to repeat it below as well for the `let(:user) {create(:student, email: user_email)}` line, so maybe your way is better for the sake of brevity.

carl-codeorg: I could swap it to using `let` if you have a strong preference. Otherwise I'll just leave it as-is to benefit from the our json string bag working as intended

dashboard/app/controllers/sessions_controller.rb
carl-codeorg: Agree on the dogmatic approach. I normally like to just do everything UTC on the backend then convert later, which is why it feels off to me to be setting it in the server timezone on the backend then converting to UTC in the frontend

stephenliang (approved):

PR #59514, opened 2024-07-01
hide pop up when user toggles back to V1

kobryan0619:
Prevents the pop-up from popping up IMMEDIATELY after a user switches back to V1.  The pop-up will show up when the user re-loads the page while on V1 view.

The video below describes the behavior - sound on for my narration :) 

https://github.com/code-dot-org/code-dot-org/assets/24883357/10fe64c6-abf8-44a7-9b0f-9ef6cc6ba64b

## Links

[Ticket](https://codedotorg.atlassian.net/jira/software/projects/TEACH/boards/65?selectedIssue=TEACH-1089)

### DIFF
--- a/apps/src/templates/sectionProgressV2/SectionProgressSelector.jsx
+++ b/apps/src/templates/sectionProgressV2/SectionProgressSelector.jsx
@@ -31,17 +31,13 @@ function SectionProgressSelector({
   const [hasJustSwitchedToV1, setHasJustSwitchedToV1] = React.useState(false);
 
   const onShowProgressTableV2Change = useCallback(() => {
-    console.log('showProgressTableV2: ', showProgressTableV2);
     if (showProgressTableV2) {
       setHasJustSwitchedToV1(true);
-      console.log('should indicate a user switched to V1 from V2');
-      // do not show pop-up
     }
     const shouldShowV2 = !showProgressTableV2;
     new UserPreferences().setShowProgressTableV2(shouldShowV2);
     setShowProgressTableV2(shouldShowV2);
     setHasJustSwitchedToV2(true);
-    console.log('should indicate a user switched to V2 from V1.... maybe');
 
     if (shouldShowV2) {
       analyticsReporter.sendEvent(EVENTS.PROGRESS_V2_VIEW_NEW_PROGRESS, {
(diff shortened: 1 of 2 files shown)


dju90:
User experience looks great, thanks Kaitie!

apps/src/templates/sectionProgressV2/SectionProgressSelector.jsx
```diff
@@ -25,10 +25,15 @@ function SectionProgressSelector({
   setShowProgressTableV2,
   progressTableV2ClosedBeta,
   sectionId,
+  hasSeenProgressTableInvite,
 }) {
   const [hasJustSwitchedToV2, setHasJustSwitchedToV2] = React.useState(false);
+  const [hasJustSwitchedToV1, setHasJustSwitchedToV1] = React.useState(false);
```
lfryemason: Are `hasJustSwitchedToV2` and `hasJustSwitchedToV1` basically the same? Could we just use one and name it `hasJustToggledViews` or something?

kobryan0619: Good point.  In my mind, I liked the clarity of knowing which thing was toggled, but at the same time, I think it was false clarity since we never unset the opposite state when toggling.  I made this change.

apps/src/templates/sectionProgressV2/SectionProgressSelector.jsx
```diff
@@ -100,11 +101,14 @@ function SectionProgressSelector({
...
-    if (!disableModal) {
+    if (disableModal || hasJustToggledViews) {
+      return;
+    }
+    if (!hasSeenProgressTableInvite) {
       return (
         <InviteToV2ProgressModal
           sectionId={sectionId}
-          setHasJustSwitchedToV2={setHasJustSwitchedToV2}
+          setHasJustSwitchedToV2={setHasJustToggledViews}
```
kobryan0619, line 111: I did keep the prop name for `InviteToV2ProgressModal` the same since I think it made the code for InviteToV2ProgressModal slightly more readable.

apps/src/templates/sectionProgressV2/SectionProgressSelector.jsx
```diff
@@ -113,7 +117,7 @@ function SectionProgressSelector({
   return (
     <div className={styles.pageContent}>
       {displayV2 && (
-        <ProgressBanners hasJustSwitchedToV2={hasJustSwitchedToV2} />
+        <ProgressBanners hasJustSwitchedToV2={hasJustToggledViews} />
```
kobryan0619, line 120: Same note here re: property name staying the same.

lfryemason (approved):

PR #56874, opened 2024-02-28
Add content to icon key popup

kobryan0619:
Adds content to the pop-up.

image

Pop-up with validation
image

Left to do:
- Get grey box around each icon.
- Make the pop-up scrollable
- Write tests

## Links

- [Figma](https://www.figma.com/file/gCktlnafGYQSNJafD3FUgA/Progress-Tracking?node-id=2177%3A34456&mode=dev)

## Testing story

Tested locally.  Writing tests are coming in a future PR.

## PR Checklist:

- [ ] Tests provide adequate coverage
- [ ] Privacy and Security impacts have been assessed
- [ ] Code is well-commented
- [ ] New features are translatable or updates will not break translations
- [ ] Relevant documentation has been added or updated
- [ ] User impact is well-understood and desirable
- [ ] Pull Request is labeled appropriately
- [ ] Follow-up work items (including potential tech debt) are tracked and linked

PR #59517, opened 2024-07-01
Python Lab: Graph Modal 

Erin007:
[CT-656](https://codedotorg.atlassian.net/browse/CT-656)
Relevant [discussion](https://codedotorg.slack.com/archives/C05HZFH7BED/p1719858332949839?thread_ts=1719847255.575499&cid=C05HZFH7BED) 

This gives the user an icon to click on images output to the console that will pop up a modal containing the image. 

https://github.com/code-dot-org/code-dot-org/assets/12300669/60504d6c-26eb-4b76-9057-0018b0cdb82e

## Follow-up work
- Making the pop-up draggable and/or resizable, [CT-668](https://codedotorg.atlassian.net/browse/CT-668)
- I think there's potential, tangential follow-up work to broadly add a X close icon to the AccessibleDialog component

## PR Checklist:

- [ ] Code is well-commented


apps/src/templates/accessible-dialogue.module.scss
```diff
@@ -24,3 +24,47 @@
   padding: 1rem;
   overflow: auto;
 }
+
```
Erin007, line 27: We currently have [11 use cases](https://github.com/search?q=repo%3Acode-dot-org%2Fcode-dot-org+AccessibleDialog&type=code) of the `AccessibleDialog` component. Most of those render an X close icon and  then [duplicate the same css](https://github.com/search?q=repo%3Acode-dot-org%2Fcode-dot-org%20.xCloseButton%20&type=code) in individual scss modules. Let's bake the X icon into the component. 
I started down that path by co-locating the X icon styling with the rest of the styles for the accessible dialog.

Erin007: @levadadenys Is a Dialog/Modal like this in the works for the Design System Component Library? I think it would be great to standardize styling and de-duplicate code across these use cases, but don't want to do throw away work if there is an new and improved component already planned. Thanks!

Erin007: @code-dot-org/teacher-tools looks like ya'll own a handful of the other use cases for `AccessibleDialog` with an X close icon styled the same way (e.g. https://github.com/code-dot-org/code-dot-org/pull/56874) . Any initial objections if I try to consolidate?

apps/src/codebridge/Console/index.tsx
```diff
@@ -54,11 +58,27 @@ const Console: React.FunctionComponent = () => {
...
+                  alt="matplotlib_image"
+                />
+                <Button
+                  color={buttonColors.black}
+                  disabled={false}
+                  icon={{iconName: 'up-right-from-square', iconStyle: 'solid'}}
+                  isIconOnly={true}
+                  onClick={() => setGraphModalOpen(true)}
+                  size="xs"
+                  type="primary"
```
molly-moen, line 73: Should we add an aria-label to this button to make it more accessible?

apps/src/codebridge/Console/index.tsx
```diff
@@ -54,11 +58,27 @@ const Console: React.FunctionComponent = () => {
...
+                />
+                {graphModalOpen && (
```
molly-moen, line 76: is it possible to have more than one modal open?

molly-moen (commented):
Couple minor questions, looks good!

apps/src/codebridge/Console/index.tsx
Erin007: No, because there's no way to open a second without closing the first given that you have to click dismiss or click outside the first modal to access the "open in new window" icon of the second which will close the first.

Erin007: I can imagine a future world where it might be nice to have multiple resizable & draggable popups that don't darken the background so a student could compare graphs side by side; but that felt like a bridge we could cross if/when we come to it.

molly-moen (approved):
🎉

PR #59518, opened 2024-07-01
UnitGroupUnit table entries must be deleted when unit is deleted

vijayamanohararaj:
Issue: When a unit that belongs to a unit group is deleted, the corresponding entries in UnitGroupUnit table are not deleted. As a result, loading the unit group through course overview page fails with the invalid reference.

Fix: Include a dependency to delete corresponding UnitGroupUnit entry during Unit deletion.

## Links

JIRA https://codedotorg.atlassian.net/browse/TEACH-1161

## Testing story

Unit test, drone 
Validated the change locally in level builder mode.
 - Deleted the unit
 - Confirmed corresponding entry from UnitGroupUnit table was deleted
 - Loading the course overview page was successful.

## Deployment strategy

DTP

## Follow-up work

Captured in same JIRA task

## Privacy

N/A

## Security

N/A

## Caching

N/A

## PR Checklist:

- [x] Code is well-commented

### DIFF
--- a/dashboard/app/models/unit.rb
+++ b/dashboard/app/models/unit.rb
@@ -63,8 +63,8 @@ class Unit < ApplicationRecord
   has_one :plc_course_unit, class_name: 'Plc::CourseUnit', foreign_key: 'script_id', inverse_of: :script, dependent: :destroy
   belongs_to :wrapup_video, class_name: 'Video', optional: true
   belongs_to :user, optional: true
-  has_many :unit_group_units, foreign_key: 'script_id'
-  has_many :unit_groups, through: :unit_group_units, dependent: :destroy
+  has_many :unit_group_units, foreign_key: 'script_id', dependent: :destroy
+  has_many :unit_groups, through: :unit_group_units
   has_one :course_version, as: :content_root, dependent: :destroy
 
   scope :with_associated_models, -> do
--- a/dashboard/app/models/unit.rb
+++ b/dashboard/app/models/unit.rb
@@ -64,7 +64,7 @@ class Unit < ApplicationRecord
   belongs_to :wrapup_video, class_name: 'Video', optional: true
   belongs_to :user, optional: true
   has_many :unit_group_units, foreign_key: 'script_id'
-  has_many :unit_groups, through: :unit_group_units
+  has_many :unit_groups, through: :unit_group_units, dependent: :destroy
   has_one :course_version, as: :content_root, dependent: :destroy
 
   scope :with_associated_models, -> do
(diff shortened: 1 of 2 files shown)


dashboard/app/models/unit.rb
```diff
@@ -64,7 +64,7 @@ class Unit < ApplicationRecord
   belongs_to :wrapup_video, class_name: 'Video', optional: true
   belongs_to :user, optional: true
   has_many :unit_group_units, foreign_key: 'script_id'
-  has_many :unit_groups, through: :unit_group_units
+  has_many :unit_groups, through: :unit_group_units, dependent: :destroy
```
davidsbailey: please move `dependent: :destroy` to previous line. this looks like it will destroy the unit_group, whereas we want to destroy only the unit_group_unit.

dashboard/test/models/unit_test.rb
```diff
@@ -2448,6 +2448,45 @@ class MigratedScriptCopyTests < ActiveSupport::TestCase
...
+
+    # delete unit in unit group
+    unit_id = unit_in_course.id
+    unit_in_course.destroy
+
+    assert Unit.find_by(id: unit_id).nil?
+
+    # Course version is associated to unit group and shouldn't be deleted
+    assert CourseVersion.find_by(id: course_version.id)
+    assert UnitGroupUnit.find_by(id: unit_gp_unit.id).nil?
```
davidsbailey, line 2488: please also assert that the UnitGroup still exists

vijayamanohararaj: added

dashboard/app/models/unit.rb
vijayamanohararaj: Thanks for the catch!! I'll go over the docs to understand why the unit tests were passing.

vijayamanohararaj: image

May be this is why the unit group wasn't getting deleted. I don't fully understand the first part of the sentence, around `belongs_to` through.

davidsbailey: yes, that could be it! 

> the association on the join model must be a belongs_to

That is referring to these associations: https://github.com/code-dot-org/code-dot-org/blob/757b18217d88bb9f02efb1aabcc028ece0e051d3/dashboard/app/models/unit_group_unit.rb#L22-L23

davidsbailey: as opposed to has_one, has_many, etc

davidsbailey (approved):

PR #59523, opened 2024-07-01
Fix Violations of and Reenable `Style/FetchEnvVar`

Hamms:
> Suggests `ENV.fetch` for the replacement of `ENV[]`. `ENV[]` silently fails and returns `nil` when the environment variable is unset, which may cause unexpected behaviors when the developer forgets to set it. On the other hand, `ENV.fetch` raises KeyError or returns the explicitly specified default value.

Fix applied automatically with `bundle exec rubocop --autocorrect-all --only Style/FetchEnvVar`

This one seems pretty straightforward; it replaces all existing instances of `ENV['foo']` with `ENV.fetch('foo', nil)`, which is slightly more verbose but significantly more explicit. A win in my book!

What do y'all think; anyone have any objections to or concerns with enabling this rule?

## Links

- https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/FetchEnvVar
- https://docs.rubocop.org/rubocop/cops_style.html#stylefetchenvvar

## Testing story

Relying on existing tests to verify that none of these changes result in a difference in behavior.

## PR Checklist:

- [ ] Code is well-commented


sureshc:
This is a huge improvement. The defect in the `rds:delete_cluster` rake task related to missing environment variables had a very high risk.

We’re changing some parts of our monolith repository that are not exercised in our continuous integration builds. For example, `lib/cdo/aws/cloudfront.rb` probably only gets exercised with the `adhoc:cdn:start` rake task. One of the cookbooks we’re modifying here only runs on gateway.

It’s probably worth provisioning a full stack or at least a “CDN” adhoc on this feature branch to test some of the modifications here.

Hamms:
Oh, to clarify: this does not actually result in any change to behavior. The auto-correction replaces all existing instances of `ENV['foo']` with `ENV.fetch('foo', nil)`, which will continue to silently return `nil` when the value is not present in the environment, but will just make it more obvious to anyone reading the code that that's what it's doing. Note in the block removed from `.rubocop_todo.yml` that it says `This cop supports safe autocorrection`; that's an indication that the code changes won't actually affect any functionality

It sounds like we probably want to update the behavior of `rds:delete_cluster`, which we should do in a follow-up PR. Because the changes in this PR will not result in any different behavior, I don't think it's necessary to do any manual testing.

snickell (approved):
LGTM, though worth noting that the mass replacement of `ENV['blah']` with `ENV.fetch('blah', nil)` renders the advantages of ENV.fetch moot (namely the raising of a KeyError). That said, I recognize it'd be pretty hard to poke through this many instances to see which ones /should/ raise a KeyError.

I guess the main reason to enable this is to make future users of ENV think through if they want a default value, or an exception?

Hamms:
Yes, exactly; the long-term goal of enabling the rule is to make it more explicit to anyone trying to reason about `ENV` in the future to know exactly what's going on. The short-term goal of the autocorrection is to allow us to enable the rule without changing anything about the way the codebase currently works. This is a `Style/` cop, after all.

I think it would make sense as follow-up work for us to evaluate our existing uses to figure out in each case whether the current behavior is intentional or the product of ignorance, but that kind of improvement is outside the scope of this change.

Thanks for thinking this through with me!

PR #59525, opened 2024-07-02
Clean up existing ProjectUseDatablockStorage rows before creating a new one

snickell:
We allow multiple ProjectUseDatablockStorage rows with the same project_id (oops), so we need to clean up old ones before creating a new one.


cnbrenci (approved):
👁️👄

PR #59527, opened 2024-07-02
Adds a multi-select dropdown to Build Your Own Workshop for topics

hannahbergam:
This PR adds a new custom multi-select dropdown to the workshop dashboard for the topic selection. 

I am using dummy data for now and will pipe through the self-paced PL topics from levelbuilder in the next followup ticket. This ticket is only covering the UI and will not cover or test saving either. 

I know this design is far far from ideal. Unfortunately, the design system component is a form, which cannot be nested within an existing form. I used some of the design and stying from the similar work done in the amazon future engineer form. In an ideal world I would have the time to create an alternate design component that is not a form component that could be shared, but that is out of scope for this project. 

I would've also ideally used a stylesheet, but this form already has in-file styles so I matched the existing format.

Screenshot 2024-07-01 at 8 39 41 PM

## Links

- spec: []()
- jira ticket: https://codedotorg.atlassian.net/browse/ACQ-2019

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/code-studio/pd/workshop_dashboard/components/workshop_form.jsx
+++ b/apps/src/code-studio/pd/workshop_dashboard/components/workshop_form.jsx
@@ -52,6 +52,10 @@ import CourseSelect from './CourseSelect';
 import SubjectSelect from './SubjectSelect';
 import MapboxLocationSearchField from '../../../../templates/MapboxLocationSearchField';
 import ModuleSelect from './ModuleSelect';
+import SingleCheckbox from '@cdo/apps/code-studio/pd/form_components/SingleCheckbox';
+import FontAwesome from '@cdo/apps/templates/FontAwesome';
+import color from '@cdo/apps/util/color';
+import classnames from 'classnames';
 
 // Default to today, 9am-5pm.
 const placeholderSession = {
@@ -60,6 +64,15 @@ const placeholderSession = {
   startTime: '9:00am',
   endTime: '5:00pm',
 };
+
+const ALL_PL_TOPICS = [
+  'Test Self Paced PL Topic',
+  'Another Test Self Paced PL Topic',
+  'A Third Test Self Paced PL Topic',
+  'And a Final Very Long Test Self Paced PL Topic That Takes Up a Line',
+];
+
+const INPUT_HEIGHT = 34;
 
 // When selecting whether a workshop is virtual through the UI,
 // a user is really selecting two things:
@@ -139,6 +152,7 @@ export class WorkshopForm extends React.Component {
       virtual: false,
       suppress_email: false,
       third_party_provider: null,
+      plTopics: [],
     };
 
     if (props.workshop) {
@@ -720,6 +734,22 @@ export class WorkshopForm extends React.Component {
     return value;
   };
 
+  // Selects the given value in the topic dropdown
+  handleSelect = event => {
+    const value = Object.keys(event)[0];
+    const isChecked = event[Object.keys(event)[0]];
+
+    let updatedTopics;
+    if (isChecked) {
+      // Add checked item into applied filters
+      updatedTopics = [...this.state.plTopics, value];
+    } else {
+      // Remove unchecked item from applied filters
+      updatedTopics = this.state.plTopics.filter(item => item !== value);
+    }
+    this.setState({plTopics: updatedTopics});
+  };
+
   handleLocationChange = event => {
     const location = event && event.target && event.target.value;
     this.setState({location_address: location});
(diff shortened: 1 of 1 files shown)


apps/src/code-studio/pd/workshop_dashboard/components/workshop_form.jsx
```diff
@@ -61,6 +65,15 @@ const placeholderSession = {
...
 };
 
+const ALL_PL_TOPICS = [
+  'Test Self Paced PL Topic',
+  'Another Test Self Paced PL Topic',
+  'A Third Test Self Paced PL Topic',
+  'And a Final Very Long Test Self Paced PL Topic That Takes Up a Line',
+];
+
+const INPUT_HEIGHT = 34;
```
TurnerRiley, line 76: Nit: since this is only used in one spot in the in-file styling, I imagine it's fine to just set that height style down below to 34

apps/src/code-studio/pd/workshop_dashboard/components/workshop_form.jsx
```diff
@@ -720,6 +734,22 @@ export class WorkshopForm extends React.Component {
     return value;
   };
 
+  // Selects the given value in the topic dropdown
+  handleSelect = event => {
```
TurnerRiley: Nit: Would you mind adding a more descriptive function name please? Just in case we end up adding other Selects

apps/src/code-studio/pd/workshop_dashboard/components/workshop_form.jsx
```diff
@@ -720,6 +734,22 @@ export class WorkshopForm extends React.Component {
...
+    const value = Object.keys(event)[0];
+    const isChecked = event[Object.keys(event)[0]];
```
TurnerRiley: Nit: could this be `const isChecked = event[value];` ?

TurnerRiley (approved):
Nice job pulling this together despite the component issues!! Just a couple nits but none of them are blocking

apps/src/code-studio/pd/workshop_dashboard/components/workshop_form.jsx
```diff
@@ -61,6 +65,15 @@ const placeholderSession = {
   endTime: '5:00pm',
 };
 
+const ALL_PL_TOPICS = [
```
bethanyaconnor, line 69: nit: not worth a rebuild but can you add a TODO here that these are hardcoded? Not worth it if you'd have to run Drone again though

apps/src/code-studio/pd/workshop_dashboard/components/workshop_form.jsx
```diff
@@ -720,6 +734,22 @@ export class WorkshopForm extends React.Component {
...
+    const value = Object.keys(event)[0];
+    const isChecked = event[Object.keys(event)[0]];
+
+    let updatedTopics;
+    if (isChecked) {
+      // Add checked item into applied filters
+      updatedTopics = [...this.state.plTopics, value];
+    } else {
+      // Remove unchecked item from applied filters
+      updatedTopics = this.state.plTopics.filter(item => item !== value);
```
bethanyaconnor, line 749: does `filter` copy the list? if not, I think you'll need a spread here too

hannahbergam: Because we're starting with the plTopics list, filter is just removing anything that should be unchecked. But maybe I'm misunderstanding your comment?

apps/src/code-studio/pd/workshop_dashboard/components/workshop_form.jsx
hannahbergam: Updated! Thanks!

apps/src/code-studio/pd/workshop_dashboard/components/workshop_form.jsx
hannahbergam: Goooood call. Updated

apps/src/code-studio/pd/workshop_dashboard/components/workshop_form.jsx
hannahbergam: Added a TODO! Thanks!

apps/src/code-studio/pd/workshop_dashboard/components/workshop_form.jsx
bethanyaconnor: Got myself all turned around with immutability. React state should be immutable, but `filter` creates [a shallow copy](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter) of the array so I think you're good!

bethanyaconnor (approved):

apps/src/code-studio/pd/workshop_dashboard/components/workshop_form.jsx
```diff
@@ -1117,6 +1148,71 @@ export class WorkshopForm extends React.Component {
               )}
             </Col>
           </Row>
+          <Row>
```
bethanyaconnor, line 1151: just curious if this should be inside the conditional?

hannahbergam: Good q! I just double checked the formatting without the 'build your own' selected and it looks fine. That said I can pull it inside in my next PR that pipes in the pl offerings! Just to get this one merged:)

PR #59530, opened 2024-07-02
P20-893: Fix duplication of CAP user events

artem-vavilov:
## Summary
- Fixed duplication of CAP user events
- Fixed the `state_before` to set it to nil when it is not present

### Duplicates Removal
```ruby
%w[grace_period_start account_locking permission_granting].each do |event_name|
  events = CAP::UserEvent.where(policy: 'cpa', name: event_name)
  original_event_ids = events.group(:user_id).pluck('min(id)')
  events.where.not(id: original_event_ids).in_batches.delete_all
end
```

### Events correction
```ruby
%w[grace_period_start account_locking].each do |event_name|
  CAP::UserEvent.where(policy: 'cpa', name: event_name).in_batches.update_all(state_before: nil)
end
```

## Links
- JIRA ticket: [P20-893](https://codedotorg.atlassian.net/browse/P20-893)
- Related PR: https://github.com/code-dot-org/code-dot-org/pull/58568

### DIFF
--- a/dashboard/app/models/cap/user_event.rb
+++ b/dashboard/app/models/cap/user_event.rb
@@ -28,6 +28,7 @@ module CAP
       GRACE_PERIOD_START = 'grace_period_start'.freeze,
       ACCOUNT_LOCKING = 'account_locking'.freeze,
       ACCOUNT_PURGING = 'account_purging'.freeze,
+      COMPLIANCE_REMOVING = 'compliance_removing'.freeze,
     ].freeze
 
     enum policy: POLICIES.index_by(&:underscore), _suffix: true
@@ -37,12 +38,5 @@ module CAP
 
     validates :name, presence: true
     validates :policy, presence: true
-
-    before_save :ensure_state_is_set
-
-    private def ensure_state_is_set
-      self.state_before ||= user.child_account_compliance_state
-      self.state_after ||= user.child_account_compliance_state
-    end
   end
 end
--- a/dashboard/app/models/user.rb
+++ b/dashboard/app/models/user.rb
@@ -291,8 +291,6 @@ class User < ApplicationRecord
 
   after_save :save_email_reg_partner_preference, if: -> {share_teacher_email_reg_partner_opt_in_radio_choice.present?}
 
-  after_save :log_cap_event, if: -> {properties_previous_change&.dig(1, 'child_account_compliance_state')}
-
   after_create if: -> {Policies::Lti.lti? self} do
     Services::Lti.create_lti_user_identity(self)
   end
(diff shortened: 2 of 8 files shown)


wilkie:
Duplicate deletion query makes sense!

Regarding:

```
%w[grace_period_start account_locking].each do |event_name|
  CAP::UserEvent.where(policy: 'cpa', name: event_name).in_batches.update_all(state_before: nil)
end
```

Oddly there is a single recent case where we go from `s` to `p` and also many `s` to `l` in production which would become `nil` to `p`/`l` respectively. Hmm. Should we deal with the `s` events in general, too? Or is this also doing this for us as a bonus? Though there will still be some `s` to `g` events out there leftover.

wilkie (approved):
The changes to add the new event, remove pedantic setting of before state, and explicitly logging the cap_user_event outside of the update flow of the User record all make sense.

The tests are very comprehensive! Thank you! I can more or less rely on those.

artem-vavilov:
> Oddly there is a single recent case where we go from `s` to `p` and also many `s` to `l` in production which would become `nil` to `p`/`l` respectively. Hmm. Should we deal with the `s` events in general, too? Or is this also doing this for us as a bonus? Though there will still be some `s` to `g` events out there leftover.

The ideal solution would be to remove the "s" (request_sent) state from all students, as it is no longer in use.
However, it would be acceptable to simply nullify this state in the events.
The user compliance state "s" will be replaced with one of the existing states eventually.

wilkie:
Running the queries:

Checking via:

```
%w[grace_period_start account_locking permission_granting].each do |event_name|
  events = CAP::UserEvent.where(policy: 'cpa', name: event_name)
  original_event_ids = events.group(:user_id).pluck('min(id)')
  puts "#{event_name}"
  puts "- count: #{events.count}"
  puts "- unique ids: #{original_event_ids.count}"
  puts "- events.where.not(id: original_event_ids): #{events.where.not(id: original_event_ids).count}"
end
```

Before:

```
grace_period_start
- count: 31
- unique ids: 30
- events.where.not(id: original_event_ids): 1
account_locking
- count: 196
- unique ids: 116
- events.where.not(id: original_event_ids): 80
permission_granting
- count: 112
- unique ids: 44
- events.where.not(id: original_event_ids): 68
```

After:

```
grace_period_start
- count: 31
- unique ids: 31
- events.where.not(id: original_event_ids): 0
account_locking
- count: 116
- unique ids: 116
- events.where.not(id: original_event_ids): 0
permission_granting
- count: 44
- unique ids: 44
- events.where.not(id: original_event_ids): 0
```

wilkie:
Second query:

Checking with:

```
%w[grace_period_start account_locking].each do |event_name|
  puts "#{event_name} number with state_before not nil: #{CAP::UserEvent.where(policy: 'cpa', name: event_name).where.not(state_before: nil).count}"
end
```

Before:

```
grace_period_start number with state_before not nil: 30
account_locking number with state_before not nil: 116
```

After:

```
grace_period_start number with state_before not nil: 0
account_locking number with state_before not nil: 0
```

PR #59531, opened 2024-07-02
AI Chat: remove fixed height, allow contents to grow to fill size

sanchitmalhotra126:
More design tweaks. This change specifically removes the explicit heights from the customization panel, chat window, and presentation panel in favor of some flexbox + height: 100% finagling to allow the container to fill whatever height it is while always keeping bottom content pinned and visible.

I did have to make one change in the DSCO Tabs component to allow passing a custom classname for the specific `<Tab>` views. This is because we need a way to assign a height of `100%` to the div that wraps the specific tab content.

Also, went ahead and replaced the `PublishStatus` component with the new DSCO Alert component now that it's available, as per designs.

**Before**

https://github.com/code-dot-org/code-dot-org/assets/85528507/25646189-9eaf-437f-badc-fb3ffbe3fdc1

**After**

https://github.com/code-dot-org/code-dot-org/assets/85528507/00ea0d55-dcc2-4c04-848c-a3ab98ea7423

## Links

https://codedotorg.atlassian.net/browse/LABS-855

## Testing story

Tested on a Gen AI pilot level with a variety of screen sizes & scrollable content.


apps/src/componentLibrary/tabs/_TabPanel.tsx
```diff
@@ -6,15 +6,23 @@ interface TabPanelProps {
   id: string;
   /* The ID of the button element that controls this panel's visibility. */
   labelledBy: string;
+  className?: string;
```
sanchitmalhotra126: @levadadenys specifically want to run this change by you. This allows consumers to pass in a custom class name for the container that wraps this specific tab. I added this because I needed a way to assign `height: 100%` to the `div` that wrapped the tab. Let me know if you think this would be an issue. I checked other usages of `Tabs` and saw only one other (LandingPage.jsx) which seems to only use the tabs portion (not the content) so this shouldn't affect it.

apps/src/aichat/views/modelCustomization/PublishNotes.tsx
```diff
@@ -54,52 +52,50 @@ const PublishNotes: React.FunctionComponent = () => {
     animationType: 'spin',
   };
 
+  const [alertText, type]: [string, AlertProps['type']] = hasFilledOutModelCard
+    ? ['Ready to publish', 'success']
+    : ['In order to publish, you must fill out a model card', 'warning'];
```
sanchitmalhotra126, line 57: Noting that this means we lose the previous formatting (bolding "must") - I'm going to open a separate task to look into the possibility of generally formatting text in the Alert component since this will affect the chat window too (spoke to Denys briefly about this).

thomasoniii (approved):
lgtm 🚀

apps/src/aichat/views/modelCustomization/PublishNotes.tsx
thomasoniii: Just suggesting here (I don't know if the ticket isn't filed yet or my jira-fu just sucks), but my first guess at the easiest way to do this would just be to add a new attribute to `Alert` - (arbitrarily named) `Content`, and have _that_ just be any old react component.

Any attempts to bake in formatting to a component like this will tend to turn into a rat's nest of weird special cases and progressively obscure syntax as we attempt to handle all the weird cases that we'll inevitably come up with once the functionality exists. But by just handing in a component, we can kick all the formatting questions externally. So then this code could be something like this:

```
const [AlertContent, type]: [string, AlertProps['type']] = hasFilledOutModelCard
    ? [<span>Ready to publish</span?, 'success']
    : [<span>In order to publish, you <StrongText>must</StrongText> fill out a model card</span>, 'warning'];
// honestly, I don't remember offhand if we can embed jsx in this manner or if it'd need to be returned from a helper function/component, but I'll leave that as an exercise for the reader. ;-)

// and then
{!isReadOnly && <Alert Content={AlertContent} type={type} size="s" />}

// and finally, in Alert.tsx, given a new prop of "Content"
// change line 104 to this:
{ Content ? <Content/> : <span className={moduleStyles.alertText}>{text}</span> }
// which will use the new Content prop if it exists, or fall back on text if it doesn't.
```

sanchitmalhotra126: Yep, this makes sense to me! And nope I hadn't filed the ticket yet :) Curious what @levadadenys thinks about allowing arbitrary content into the alert (don't think it should be an issue, but I'm wondering if the design team has opinions otherwise).

apps/src/aichat/views/model-customization-workspace.module.scss
```diff
@@ -9,17 +9,23 @@
...
   }
 
   .tabPanels {
-    padding: 16px;
+    padding: 16px 0;
+    flex: 1;
+    overflow: hidden;
+  }
+
+  .tabPanel {
```
levadadenys: instead of adding className - can be simplified here to just styling the nearest level div (see screenshot), which will save us some lines of repeated code
image

apps/src/aichat/views/modelCustomization/PublishNotes.tsx
levadadenys: Yeah, agree, if adding support of custom html content - it can be implemented in a way @thomasoniii suggested.

However, I've spoken with design team last week and they said that `Alert` is already important/urgent enough to keep it as is, meaning we're not planning to support any bold text or custom html content in `Alerts` for now, sorry for not sharing it earlier, thought it was already discussed with everyone.

If it's crutial, might be worth discussing with design team (@markabarrett) once again.

apps/src/componentLibrary/tabs/_TabPanel.tsx
levadadenys: Adding this prop LGTM overall, thanks for adding!
In this specific case I think we can use my styles suggestion above, however for the future I think we should keep it in case we'll need some custom styles for specific tab/number of tabs(but not all) panels.

One small nit: Would be awesome if we also udpate Tabs Chagnelog here (can be [0.2.1]) just to keep logging the updates to hte component

levadadenys (commented):
Have a couple comments, but LGTM overall!
Thanks a lot for connecting Design System to new screen!

apps/src/aichat/views/model-customization-workspace.module.scss
sanchitmalhotra126: Ah good call, much simpler - will do!

apps/src/aichat/views/modelCustomization/PublishNotes.tsx
sanchitmalhotra126: Gotcha, thanks for the context! I'll chat with Mark/Sam directly.

apps/src/componentLibrary/tabs/_TabPanel.tsx
sanchitmalhotra126: Sounds good! In that case, just to keep this change more targeted, I can revert the design system changes and use your suggestion directly. If you still think it's worthwhile having a custom class name for the tab element, I can split that off into a separate change.

PR #59532, opened 2024-07-02
AI Chat: fix auto-scroll behavior

sanchitmalhotra126:
Fix a bug where the chat would auto-scroll on any update in the customization window (opening dropdown, typing text, changing slider, etc) instead of only when new messages or notifications appear.

**Before**

https://github.com/code-dot-org/code-dot-org/assets/85528507/da581a93-4ede-46e1-87e7-61a6e652c881

**After**

https://github.com/code-dot-org/code-dot-org/assets/85528507/20964704-afb3-41db-859b-82127c2f2541

## Testing story

Tested on pilot levels.

### DIFF
--- a/apps/src/aichat/views/ChatWorkspace.tsx
+++ b/apps/src/aichat/views/ChatWorkspace.tsx
@@ -23,6 +23,7 @@ const ChatWorkspace: React.FunctionComponent = () => {
   );
 
   const messages = useSelector(selectAllMessages);
+
   const isWaitingForChatResponse = useSelector(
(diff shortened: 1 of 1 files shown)


bencodeorg (approved):
Nice! Do you know if this has been here for a while, or is this a regression since our chat history refactoring?

sanchitmalhotra126:
> Nice! Do you know if this has been here for a while, or is this a regression since our chat history refactoring?

Not sure actually? @dancodedotorg spotted it and mentioned it in Slack, but I don't know if it's been around longer

PR #59535, opened 2024-07-02
Show predict answer to teachers

molly-moen:
For new predict levels, we need to show the solution to authorized teachers (I also am showing it to levelbuilders for ease of use for levelbuilders/us). Similar to how we display it in the old version, we put the solution in the "For Teachers Only" tab. If we end up doing away with this tab we can consider where else this could go. 

While working on this I realized I could re-use the same "solution" field for both multiple choice and free response fields, rather than having two different fields. That refactor is included in this PR.

The component that displays the solution lives in `lab2` since we hope to use this new pattern for other lab2 labs. The "For Teachers Only" tab is defined by Code Bridge, and now just imports the new component.

### Free Response
Screenshot 2024-07-02 at 11 20 29 AM

### Multiple choice, multiple answers
Screenshot 2024-07-02 at 11 06 06 AM

### Multiple choice, 1 answer
Screenshot 2024-07-02 at 11 58 41 AM
(In testing I removed the "Teachers should see this extra info" for teacher's only text on the level, which is why it's not present in this screenshot).

## Links

- jira ticket: [CT-640](https://codedotorg.atlassian.net/browse/CT-640)

## Testing story
Tested locally. The new component is only used by code bridge.

## PR Checklist:

- [ ] Code is well-commented


apps/src/lab2/levelEditors/predictSettings/MultipleChoiceFields.tsx
```diff
@@ -28,7 +28,7 @@ const MultipleChoiceFields: React.FunctionComponent<
     }
     setPredictSettings({
       ...predictSettings,
-      multipleChoiceAnswers: newCorrectAnswers,
+      solution: newCorrectAnswers.join(','),
```
sanchitmalhotra126, line 31: Any concerns that if the order is changed for multiple choice answers, the solution might be checked incorrectly?

apps/src/lab2/views/components/PredictSolution.tsx
```diff
@@ -0,0 +1,63 @@
...
+      const solutions = predictSettings.solution.split(',');
+      const formattedSolutions: JSX.Element[] = [];
+      for (const solution of solutions) {
+        const index = predictSettings.multipleChoiceOptions?.indexOf(solution);
+        if (index !== undefined && index !== -1) {
+          // Insert at index so the solutions show up in the same order as the options.
+          formattedSolutions[index] = (
+            <span key={index}>
+              <span className={moduleStyles.multipleChoiceLetter}>
+                {String.fromCharCode(65 + index)}.
```
sanchitmalhotra126, line 36: neat!

apps/src/lab2/views/components/PredictSolution.tsx
```diff
@@ -0,0 +1,63 @@
...
+    }
+    return solutionToDisplay;
+  };
+
+  return (
+    <div className={moduleStyles.predictSolutionContainer}>
+      <Typography semanticTag="h1" visualAppearance="heading-sm">
+        {commonI18n.answer()}
+      </Typography>
+      <Typography semanticTag="p" visualAppearance="body-three">
```
sanchitmalhotra126: nit: I think this can just be `<BodyThreeText>`

apps/src/lab2/views/components/PredictSolution.tsx
```diff
@@ -0,0 +1,63 @@
...
+        }
+      }
+      solutionToDisplay = <>{formattedSolutions}</>;
+    }
+    return solutionToDisplay;
+  };
+
+  return (
+    <div className={moduleStyles.predictSolutionContainer}>
+      <Typography semanticTag="h1" visualAppearance="heading-sm">
```
sanchitmalhotra126: Similarly I think this can be `<Heading1 visualAppearance="heading-sm">` if preferred

apps/src/lab2/views/components/PredictSolution.tsx
```diff
@@ -0,0 +1,63 @@
...
+const PredictSolution: React.FunctionComponent = () => {
+  const predictSettings = useAppSelector(
+    state => state.lab.levelProperties?.predictSettings
+  );
+
+  if (!predictSettings?.solution) {
+    return null;
+  }
+
+  const getFormattedSolution = () => {
```
sanchitmalhotra126, line 19: minor: might be worth putting this in a `useMemo` that only runs when `predictSettings` updates, but not necessary

sanchitmalhotra126 (approved):
LGTM! A few minor things, nothing blocking

apps/src/lab2/levelEditors/predictSettings/MultipleChoiceFields.tsx
molly-moen: We don't actually validate their answers anywhere. But we could avoid issues with ordering pretty easily by doing an includes rather than checking equality.

apps/src/lab2/views/components/PredictSolution.tsx
molly-moen: Good catch, I forgot that was an option.

apps/src/lab2/views/components/PredictSolution.tsx
molly-moen: I don't think this is complicated enough to warrant memoization, lmk if that is incorrect!

PR #25506, opened 2018-10-19
Projects: Dance Party in Project Gallery

Erin007:
This PR adds a Dance Party section to the top of the Public Gallery, which will show if the DCDO flag `dance_projects` is set to true. Checking for the DCDO flag and passing its value to the projects page was set up in #25498. 

screen shot 2018-10-30 at 3 12 38 pm

For now I set the default thumbnail image for Dance Party to nothing just to ensure that the projects show up correctly.  There's a separate tracked work item to add Dance Party thumbnails (@islemaster, I think you're planning to do this?)

PR #59536, opened 2024-07-02
Tidy Projects: remove more unused props

Erin007:
Similar to https://github.com/code-dot-org/code-dot-org/pull/59501

This PR removes 2 unused props from the Project Gallery components.

1.) `includeDance`
Years ago when Dance Party was released we had a specific timeline of when different Dance features could be released, including when dance type projects would show in the project gallery. To accomplish this Dance Party projects were hidden behind a DCDO flag (https://github.com/code-dot-org/code-dot-org/pull/25506). That time is long since past! Dance projects are now always in the gallery. The `includeDance` prop in the projects components was no longer being used, so I removed it. 

2.) `limitedGallery` 
When anyone could publish to the public gallery we would sometimes see outbreaks of scandalous images in App Lab and Game Lab projects. Featured "safe" projects were on the "front page" of the gallery, but if we detected naughty projects we implemented a way to "limit" the gallery by hiding the view more links for those project types (https://github.com/code-dot-org/code-dot-org/pull/24643). However, even with those safeguards, it was still too risky and time-intensive to moderate a public gallery where students could publish their projects; so we now have a "Featured Projects" gallery where all of the projects are curated internally and known to be safe (https://github.com/code-dot-org/code-dot-org/pull/57473 + [spec](https://docs.google.com/document/d/1vU68tzXG72Z80MvM6xYEDwF5hLDf7V097P-0dDfpouc/edit)). Therefore we no longer need the ability to limit the gallery so I removed this unused prop. 

## PR Checklist:

- [ ] Code is well-commented


molly-moen (approved):

fisher-alice:
Thanks for the detailed PR description!

PR #59542, opened 2024-07-02
fix(lms): do not style the teacher and student button for lti

stephenliang:
When SSOing from LTI, the user type is implied through the roster and therefore, the user does not need to select the user type. As such, we removed the student and teacher buttons but the finish signup script attempts to style these buttons. This causes a JS error which causes the submit event listener to not be attached. Adding a safety check to prevent styling if the buttons do not exist.

## Links

- [slack thread](https://codedotorg.slack.com/archives/CFTFD6BPV/p1719949268433889)

## Testing story

1. Create new teacher on Canvas
2. SSO Login first time (not roster synced)
3. Attempt to create account with school info
4. Select the no school option

Expected: Fields blank

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/sites/studio/pages/devise/registrations/_finish_sign_up.js
+++ b/apps/src/sites/studio/pages/devise/registrations/_finish_sign_up.js
@@ -171,6 +171,11 @@ $(document).ready(() => {
 
   // Style selected user type button to show it has been clicked
   function styleSelectedUserTypeButton(value) {
+    // For LTI, the user type is implied from the LMS so the buttons will not appear
+    if (!teacherButton || !studentButton) {
+      return;
+    }
+
     if (value === 'teacher') {
       teacherButton.classList.add('select-user-type-button-selected');
       studentButton.classList.remove('select-user-type-button-selected');


carl-codeorg (approved):

hannahbergam (commented):
Thanks for this! I think it would be good to test this with a student integration just to make sure there weren't any other gotchas if possible

mgc1194 (commented):
Awesome! Changes seem pretty straightforward. 
Just a couple questions/comments;
- Would it be reasonable to add test cases?
- Could we get a before-and-after screenshot of the error in case it shows in the console?
- Also, could you please link the slack thread as a link for reference?

stephenliang:
> Thanks for this! I think it would be good to test this with a student integration just to make sure there weren't any other gotchas if possible

Yes, just ran it with a student and it works without errors!

> Would it be reasonable to add test cases?

Yes, unfortunately these scripts are in the legacy non-react codebase so we have limited testing options. However, @alex-m-brown will be working on some E2E tests that can cover this scenario.

> Could we get a before-and-after screenshot of the error in case it shows in the console?

Here is the error (after this change, the error does not appear)

```
sign_up:16 Uncaught TypeError: Cannot read properties of null (reading 'classList')
    at c (_finish_sign_upwp3d5e56878b9cca22e8f9.min.js:1:1770)
    at _finish_sign_upwp3d5e56878b9cca22e8f9.min.js:1:1182
    at HTMLDocument.<anonymous> (_finish_sign_upwp3d5e56878b9cca22e8f9.min.js:1:1191)
    at h (application-0b73561263f8145fb2e5c9652b71af5375383bfa327999f59d80ad0803ab3bba.js:2:8541)
    at Object.fireWith [as resolveWith] (application-0b73561263f8145fb2e5c9652b71af5375383bfa327999f59d80ad0803ab3bba.js:2:9310)
    at Function.ready (application-0b73561263f8145fb2e5c9652b71af5375383bfa327999f59d80ad0803ab3bba.js:2:11129)
    at HTMLDocument.a (application-0b73561263f8145fb2e5c9652b71af5375383bfa327999f59d80ad0803ab3bba.js:1:1133)
    at HTMLDocument.nrWrapper (sign_up:16:25173)
```

> Also, could you please link the slack thread as a link for reference?

Yes, updated the description

PR #59546, opened 2024-07-03
fix(jest): mock out imageUtils to prevent a race condition

stephenliang:
The animation reducer uses the file uploader callback which throws due to our implementation of Image. The image api uses live browser callbacks to render out which causes random tests to fail when the callback is fired. Instead, mock out the imageUtil library as images are tested in the integration test library.

This wasn't captured in the development of jest as the staging server runs slower than my machine giving a higher chance of this happening due to timing.

## Testing story

Ran fix on the staging server, passed 5/5 times after fix.

## PR Checklist:

- [ ] Code is well-commented


Erin007 (approved):
Sneaky! Great find & fix, Stephen.

PR #59547, opened 2024-07-03
Datablock Storage: Don't double json-encode kvps

cnbrenci:
In the original implementation of kvps in datablock storage, we accidentally double json-encoded kvp values in the DB. This PR includes:

- A migration that json-decodes one level on the datablock_storage_kvps value column.
- Fixes datablock_storage_controller and datablock_storage_kvp to do the correct json encodings when getting and setting these values.

While migrations applied to prod can timeout after only 30s, this migration only updates 1500 rows and we expect it to finish in time.


snickell (approved):
"\\"lgtm\\""

PR #59595, opened 2024-07-03
Ceara/aitt 661 differentiation experiment flag and initial FAB/container

cearachew:
Adds an experiment flag for the differentiation chat, adds a mount point to level pages, and renders a floating action button with a dummy container for chat if the user is a teacher and the ai-differentiation experiment flag is enabled.

Screenshot 2024-07-03 at 12 59 25 PM

Screenshot 2024-07-03 at 12 59 43 PM

## Links

Jira ticket: https://codedotorg.atlassian.net/browse/AITT-661

## Testing story

Added a basic unit test for the FAB

## PR Checklist:

- [ ] Code is well-commented


apps/src/aiDifferentiation/AIDiffContainer.tsx
davidsbailey, line 1: naming nits: 
* AI should be written `Ai` as in `AiDiffContainer` since abbreviations are treated as their own words. this is consistent with the spelling of `aiDifferentiation` and ai-differentiation`
* I see 4 different ways you are spelling the same concept: "ai differentiation", "differentiation", "ai diff" and "differ". I'd really recommend limiting this to "ai differentation"(`AiDifferentiation`) plus at most one shortened version such as "ai diff" (`AiDiff`, etc). this is so that it is easy to search and read the code later. 
 
I think we have some of this in a style guide but I'm forgetting where.

apps/src/aiDifferentiation/AIDiffFloatingActionButton.tsx
```diff
@@ -0,0 +1,36 @@
...
+  const handleClick = () => {
+    setIsOpen(!isOpen);
+  };
+
+  return (
+    <div id="fab-contained">
+      <button
+        id="ui-floatingActionButton"
+        className={style.floatingActionButton}
+        // style={{backgroundImage: `url(${aiFabIcon})`}}
```
davidsbailey: remove dead line of code?

apps/src/aiDifferentiation/ai-differentiation.module.scss
```diff
@@ -0,0 +1,105 @@
+@import 'color';
+@import 'font';
+@import '@cdo/apps/componentLibrary/common/styles/mixins';
+
+
+.aiTutorContainer {
```
davidsbailey: can you safely remove the references to aiTutor from this file?

apps/src/sites/studio/pages/lessons/show.js
```diff
@@ -138,6 +141,15 @@ function prepareExpandableImageDialog() {
   );
 }
 
+function displayDifferentiationChat() {
+  const differFabMountPoint = document.getElementById(
+    'differentiation-fab-mount-point'
```
davidsbailey: recommendation: standardize on spelling, per earlier comment

davidsbailey (commented):
Nice job getting this working!

apps/src/sites/studio/pages/lessons/show.js
cearachew: Do you think just calling it "AiDiff" is fine? Differentiation is such a long word that I never spell right on the first try

davidsbailey: yes, I think AiDiff is fine. `Aidiff` is also an option with some precedent, see Ailab and Aichat

davidsbailey: the reason student labs use those spellings is so that they can do a case-insensitive search for "ailab" and not miss things spelled in snake or kebab case. I think that makes sense for something like a lab name since the config for each lab gets super spread out, but might be overkill for this case. you can decide, I just vote for consistency

davidsbailey (approved):
Thanks for the updates!

PR #59596, opened 2024-07-03
Allow teachers to reset new predict levels

molly-moen:
Implement the teacher reset button for lab2 predict levels. I was able to reuse api for this feature after adding another case to support resetting the new predict levels as well and multi and free response levels.

The original logic for this feature on the frontend was in [ContainedLevelResetButton](https://github.com/code-dot-org/code-dot-org/blob/staging/apps/src/templates/instructions/ContainedLevelResetButton.jsx) and [codeStudioLevels.resetContainedLevel](https://github.com/code-dot-org/code-dot-org/blob/staging/apps/src/code-studio/levels/codeStudioLevels.js). I could not directly reuse `ContainedLevelResetButton` because it relies on app options, and I wanted the new component to be hooked into the new `predictLevelRedux`. Therefore I made a new version of the component.

The logic to handle reset is mostly contained in `predictLevelRedux`, which handles initiating the api call, resetting the appropriate predict redux fields, and initiating a progress refresh. Because this feature is pretty redux-heavy, I ended up breaking our existing pattern of `InstructionsPanel` not using redux at all. Here's a link to the [Slack discussion](https://codedotorg.slack.com/archives/C044AGTKW3D/p1719957796624449) on this, and the [follow up task](https://codedotorg.atlassian.net/browse/CT-671) I filed.

## Screen recording

https://github.com/code-dot-org/code-dot-org/assets/33666587/ecd2ff3b-6acb-4471-9b3c-2e9edb3e5902

Note that the progress bubble gets reset as well as the components in the lab.

## Screenshot of error case

Screenshot 2024-07-11 at 9 40 21 AM

## Links
- jira ticket: [CT-639](https://codedotorg.atlassian.net/browse/CT-639)

## Testing story
Tested locally and added unit tests for the api change.

## Follow-up work
The old delete button was red. The new component library button does not have a red option. I started a [slack discussion](https://codedotorg.slack.com/archives/C0532LCK125/p1720026189728519) on this, for now I'm using the "secondary" button styling as suggested in the thread. There may be work to do in the future to update the styling if we get a red option. [Follow-up task](https://codedotorg.atlassian.net/browse/CT-681) is here.

## PR Checklist:

- [ ] Code is well-commented


apps/src/lab2/views/components/predict.module.scss
```diff
@@ -27,3 +27,9 @@
   background-color: $light_white;
   border-radius: 5px;
 }
+
+.resetError {
+  color: $red;
```
ebeastlake: I wonder if we should use a DSCO alert instead of red text.

dashboard/app/controllers/user_levels_controller.rb
```diff
@@ -29,7 +29,13 @@ def delete_predict_level_progress
...
     level = Level.find(params[:level_id])
     return head :not_found, text: 'Level not found' unless level
-    return head :bad_request, text: "Clearing progress on level type #{level.type} is not supported" unless ['Multi', 'FreeResponse'].include?(level.type)
+    is_allowed_level_type = ['Multi', 'FreeResponse'].include?(level.type)
+    # Some lab2 levels have a predict_settings field that indicates they are predict levels,
+    # rather than being a separate multi or free response level. We allow clearing progress on these levels.
+    if level.uses_lab2? && level.predict_settings && level.predict_settings["isPredictLevel"]
+      is_allowed_level_type = true
+    end
+    return head :bad_request, text: "Clearing progress on level type #{level.type} is not supported" unless is_allowed_level_type
```
ebeastlake: Something like this might be slightly clearer/more readable?
```
def delete_predict_level_progress
  ...

  unless clearable_level_type?(level)
    return head :bad_request, text: "Clearing progress on level type #{level.type} is not supported"
  end

  UserLevel.where(user_id: current_user.id, script_id: script.id, level: level.id).destroy_all
  head :ok
end

# Some lab2 levels have a predict_settings field that indicates they are predict levels. We allow clearing progress on these levels as well as separate multi or free-response levels.
private def clearable_level_type?(level)
  ['Multi', 'FreeResponse'].include?(level.type) || (level.uses_lab2? && level.predict_settings&.dig("isPredictLevel"))
end
```

ebeastlake (approved):
A couple of small things, but looks great overall! I filed a follow-up task based on the thread you linked if you want to add it to the PR description: https://codedotorg.atlassian.net/browse/CT-681

apps/src/lab2/views/components/predict.module.scss
molly-moen: Good idea!

dashboard/app/controllers/user_levels_controller.rb
molly-moen: good idea!

apps/src/lab2/views/components/Instructions.tsx
```diff
@@ -160,8 +160,13 @@ interface InstructionsPanelProps {
 }
 
 /**
- * Renders the instructions panel view. This is a separate component so that it can be
- * used without the Lab2 redux integration if necessary.
+ * Renders the instructions panel view. This was initially set up as a separate component
```
sanchitmalhotra126, line 169: Thanks for adding the comment!

dashboard/app/controllers/user_levels_controller.rb
```diff
@@ -58,4 +61,10 @@ def get_level_source
   private def user_level_params
     params.require(:user_level).permit(:best_result, :submitted)
   end
+
+  # Some lab2 levels have a predict_settings field that indicates they are predict levels.
+  # We allow clearing progress on these levels as well as multi or free-response levels.
+  private def clearable_level_type?(level)
+    ['Multi', 'FreeResponse'].include?(level.type) || (level.uses_lab2? && level.predict_settings&.dig("isPredictLevel"))
```
sanchitmalhotra126: nit: is it worth moving `uses_lab2 && predict_settings&.dig("isPredictLevel")` to some helper function on `level.rb` if we're checking this in a bunch of places?

sanchitmalhotra126 (approved):
LGTM!

dashboard/app/controllers/user_levels_controller.rb
molly-moen: I actually did this in a [separate pr](https://github.com/code-dot-org/code-dot-org/pull/59599/files#diff-86af1cb288d4d48f970630d0f21ee11f2dd971239bafe270eafce13691d1d0caR882). Depending on build times I'll either add that in here or do a fast follow to use that function!

molly-moen: included it here since drone failed on me 😄

PR #59599, opened 2024-07-03
Update text response summary api with new predict levels

molly-moen:
We show a summary page on the teacher dashboard of all text responses for students in the section. Previously this would show contained level responses as well as text match and free response level responses. We need to also add in responses from the new predict levels in Python Lab.

To do this I updated the `unit.text_response_levels_without_cache` function to include those levels. That function is called by [unit.text_response_levels](https://github.com/code-dot-org/code-dot-org/blob/85a25087c8482062831d2eb03442f9ad59e117f0/dashboard/app/models/unit.rb#L782), which is called by [api_controller.section_text_responses](https://github.com/code-dot-org/code-dot-org/blob/85a25087c8482062831d2eb03442f9ad59e117f0/dashboard/app/controllers/api_controller.rb#L492). Since we are saving responses in the same way for the new levels (in level sources), that function was all that needed to be updated for everything to "just work".

Now a response from a python lab level will show up in the teacher dashboard (this example was made after making some local changes to scripts to make a python script assignable)

Screenshot 2024-07-03 at 3 21 14 PM

## Links

- jira ticket: [CT-666](https://codedotorg.atlassian.net/browse/CT-666)

## Testing story
Tested locally and added unit tests. Also verified that this view for an existing course that uses the old levels is unaffected.

## PR Checklist:

- [ ] Code is well-commented


ebeastlake (approved):

dashboard/app/models/unit.rb
```diff
@@ -788,11 +788,17 @@ def text_response_levels_without_cache
...
       script_level.levels.map do |level|
-        next if level.contained_levels.empty? ||
+        # We are looking for two types of text response levels: levels with contained levels
+        # that are text response levels, and predict levels that are free response (predict
+        # levels are used by lab2).
+        is_not_contained = level.contained_levels.empty? ||
           TEXT_RESPONSE_TYPES.exclude?(level.contained_levels.first.class)
+        is_not_predict_free_response = !level.predict_level? || level.properties.dig('predict_settings', 'questionType') != 'freeResponse'
+        next if is_not_contained && is_not_predict_free_response
+        predict_level = level.predict_level? ? level : level.contained_levels.first
```
ebeastlake: No strong feelings: Would `text_response_level` be a better variable name since the comment above suggests that contained levels and predict levels are mutually exclusive?

molly-moen: good call, will update

sanchitmalhotra126 (approved):

PR #59600, opened 2024-07-03
[ProgressV2] Move expanded lessons and choice levels to redux.

lfryemason:
As part of the addition of a floating header, it will be much simpler to have expanded lessons and expanded choice levels in the redux store instead of in component state. This PR moves:
* `expandedChoiceLevels` from `ExpandedProgressDataColumn.jsx` to the redux store.
* `expandedLessonIds` from `SectionProgressV2.jsx` to the redux store.

All references to those structures was moved to redux. This means moving some from passed in props to redux references. Tests are also updated to put use redux.

## Links
[TEACH-940](https://codedotorg.atlassian.net/browse/TEACH-940)

## PR Checklist:

- [ ] Code is well-commented


apps/src/templates/sectionProgressV2/ExpandedProgressDataColumn.jsx
```diff
@@ -119,12 +98,7 @@ function ExpandedProgressDataColumn({
       <caption hidden={true}>
         {i18n.progressForLesson({lessonName: lesson.title})}
       </caption>
-      <ExpandedProgressColumnHeader
-        lesson={lesson}
-        removeExpandedLesson={removeExpandedLesson}
-        expandedChoiceLevels={expandedChoiceLevels}
-        toggleExpandedChoiceLevel={toggleExpandedChoiceLevel}
-      />
+      <ExpandedProgressColumnHeader lesson={lesson} />
```
lfryemason, line 101: This is a great example of why I wanted to do this. Dramatically reduces the number of props that are just passed through.

apps/src/templates/sectionProgress/sectionProgressRedux.js
```diff
@@ -38,6 +50,32 @@ export const addDataByUnit = data => ({
   data,
 });
 
+export const loadExpandedLessonsFromLocalStorage = (scriptId, sectionId) => ({
+  type: LOAD_EXPANDED_LESSONS_FROM_LOCAL_STORAGE,
+  scriptId,
```
etaderhold: Could we use `unit` rather than `script` more often throughout this code?

etaderhold (approved):
I like it! Does this have any performance implications?

lfryemason:
> I like it! Does this have any performance implications?

This is actually a performance improvement because connecting to the redux store will prevent component re-renders unless the redux value changes. [SO thread](https://stackoverflow.com/questions/61283239/performance-connecting-to-the-redux-store-vs-passing-data-through-props)

For the most part this will be minimal, but there are a couple components that just passed the prop from parent to child that will see (hopefully) significantly fewer re-renders.

PR #59603, opened 2024-07-04
Piping self paced pl course offerings into build your own workshop topics dropdown

hannahbergam:
This PR populates the topics list for the 'Build Your Own Workshop' option. It sets up the backend controller action to bring in only the data we needed (this seemed preferable to reusing the curriculum quick assign options but having to filter out the long list of other course types). 

I added a test (and mocked the server response) to cover this new functionality, and also verified it works with our actual PL course list (see pic below).

Screenshot 2024-07-03 at 4 55 19 PM

## Links

- spec: []()
- jira ticket: https://codedotorg.atlassian.net/browse/ACQ-2020

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/code-studio/pd/workshop_dashboard/components/workshop_form.jsx
+++ b/apps/src/code-studio/pd/workshop_dashboard/components/workshop_form.jsx
@@ -65,14 +65,7 @@ const placeholderSession = {
   endTime: '5:00pm',
 };
 
-// TODO pipe in PL topics via ACQ-2020
-const ALL_PL_TOPICS = [
-  'Test Self Paced PL Topic',
-  'Another Test Self Paced PL Topic',
-  'A Third Test Self Paced PL Topic',
-  'And a Final Very Long Test Self Paced PL Topic That Takes Up a Line',
-];
-
+let ALL_PL_TOPICS = {};
 const INPUT_HEIGHT = 34;
 
 // When selecting whether a workshop is virtual through the UI,
@@ -185,6 +178,7 @@ export class WorkshopForm extends React.Component {
       this.loadAvailableFacilitators(props.workshop.course);
     }
 
+    this.loadPlCourseOfferings();
     this.loadRegionalPartners();
 
     return initialState;
@@ -197,9 +191,11 @@ export class WorkshopForm extends React.Component {
     if (this.loadWorkshopRequest) {
       this.loadWorkshopRequest.abort();
     }
-
     if (this.loadRegionalPartnersRequest) {
       this.loadRegionalPartnersRequest.abort();
+    }
+    if (this.loadPlCoursesRequest) {
+      this.loadPlCoursesRequest.abort();
     }
   }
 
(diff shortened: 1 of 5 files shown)


apps/src/code-studio/pd/workshop_dashboard/components/workshop_form.jsx
```diff
@@ -1193,18 +1199,20 @@ export class WorkshopForm extends React.Component {
                       aria-labelledby="dropdownMenuButton"
                     >
                       <ul style={styles.listItems}>
-                        {ALL_PL_TOPICS.map(label => (
+                        {Object.values(ALL_PL_TOPICS).map(object => (
```
TurnerRiley: Nit: could `object` be renamed to `topic`?

apps/test/unit/code-studio/pd/workshop_dashboard/components/workshopFormTest.js
```diff
@@ -560,6 +560,46 @@ describe('WorkshopForm test', () => {
...
+    const courseField = wrapper.find('#course').first();
+    courseField.simulate('change', {
+      target: {name: 'course', value: 'Build Your Own Workshop'},
+    });
+    expect(wrapper.find('#topics')).to.have.lengthOf(1);
+    wrapper.find('#dropdownMenuButton').first().simulate('click');
+    // A user can select either the label or checkbox, so we expect 2 for each here
+    expect(wrapper.find('#myPlTestTopic')).to.have.lengthOf(2);
+    expect(wrapper.find('#mySecondTopic')).to.have.lengthOf(2);
+  });
```
TurnerRiley, line 601: Great test with checking for both label and checkbox!

dashboard/app/controllers/course_offerings_controller.rb
```diff
@@ -35,6 +35,12 @@ def quick_assign_course_offerings
     render :ok, json: offerings.to_json
   end
 
+  def self_paced_pl_course_offerings
+    return head :bad_request unless current_user
+    offerings = CourseOffering.assignable_course_offerings(current_user).select {|co| co.get_participant_audience == 'teacher' && co.instruction_type == 'self_paced'}
+    render :ok, json: offerings.map(&:summarize_self_paced_pl).to_json
```
TurnerRiley: Nit: Not sure if this is even a case that could happen but wondering if there should be safe chaining after `offerings` in the case that the query returns no results

TurnerRiley (approved):
Looking great!

dashboard/app/controllers/course_offerings_controller.rb
hannahbergam: Yesss good call. Added!

apps/src/code-studio/pd/workshop_dashboard/components/workshop_form.jsx
hannahbergam: Good call- more clear. Updated!

PR #59605, opened 2024-07-04
Adds ability to set remote location and uses it to refactor and fix flaky GDPR tests.

wilkie:
This fixes a flaky GDPR test. Essentially, the call to update whether or not the user has seen and dismissed the GDPR dialog happens asynchronously. The page refresh might race that update and reload before that user flag was set and perhaps interrupt that network request altogether.

When this happens, the modal will reappear, which fails the test.

This adds an admittedly strange fix: updating the source of truth for the modal when the request completes and having the test look for that flag being set before reloading the page. This is not the most ideal way to test this outside of just waiting a few seconds before reloading since Selenium/Webdriver does not yet offer a consistent/reliable way to just wait for a network request.

An alternative would be to only close the dialog when the request is complete and look for that, but that's not very convenient for the general user who expects the dialog to just instantly go away.

## GeolocationOverride Middleware

This adds some middleware that can fake the location of the request via a cookie. This is based off of the same mechanism and same general configuration of Artem's DCDO mocking middleware.

This felt appropriate since it needs to alter the request headers before the application takes control and a cookie was used to allow it to be session-based which can help in both development manual testing and our automated acceptance testing.

## Links

- jira ticket: [P20-959](https://codedotorg.atlassian.net/browse/P20-959)

## Testing story

Added a cucumber step to mock out being in Europe as a refactor for the GDPR tests while I was in there. We had similar issues with testing the CAP/CPA work, so it felt like a good time to sneak in a nice testing feature.

Added a cucumber step to look for script data. This is not a user-facing step, so it is worded as "it is observed" since it is from the point of view of the application. Not ideal.

## Deployment strategy

This contains rack middleware that overrides the `REMOTE_ADDR` field with a cookie. This will trick Geocoder into giving us a location based around the IP we set in that field. Using a cookie lets us temporary set our session to mock our location which will help us test region-locked features a little more easily and will let us mock such users in our UI testing.

This is demonstrated in the "I'm in Europe" cucumber step which establishes the session as coming from that region and then using that to more fluidly test the GDPR dialog experience.

## PR Checklist:

- [ ] Code is well-commented


wilkie:
Lots of other flaky tests causing failures for me, but my tests pass:

```
[xteam/gdpr_dialog] UI tests for Chrome_xteam_gdpr_dialog succeeded (2:23 minutes, 5 passed)
```

```
[xteam/gdpr_dialog] UI tests for Safari_xteam_gdpr_dialog succeeded (2:14 minutes, 5 passed)
```

```
[xteam/gdpr_dialog] UI tests for Firefox_xteam_gdpr_dialog succeeded (2:32 minutes, 5 passed)
```

apps/src/templates/GDPRDialog.jsx
```diff
@@ -22,8 +22,13 @@ export default class GDPRDialog extends Component {
 
   handleYesClick = () => {
     this.setState({isDialogOpen: false});
-    $.post(`/dashboardapi/v1/users/accept_data_transfer_agreement`, {
+    $.post('/dashboardapi/v1/users/accept_data_transfer_agreement', {
       user_id: this.props.currentUserId,
+    }).then(() => {
+      const gdprDataScript = document.querySelector('script[data-gdpr]');
+      let gdprData = JSON.parse(gdprDataScript.dataset['gdpr']);
```
carl-codeorg: nit: could be a const

lib/cdo/rack/geolocation_override.rb
```diff
@@ -0,0 +1,24 @@
...
+
+    def call(env)
+      # Forcibly update the REMOTE_ADDR to any given by the X_REMOTE_ADDR header
+      override = Rack::Request.new(env).cookies[KEY]
+      env['REMOTE_ADDR'] = override if override
+
+      # Coerce Geocoder to turn an internal ip to localhost so it will consider
+      # it a locale with a 'RD' country code.
+      if Geocoder.search(env['REMOTE_ADDR']).try(:first)&.data&.[]('bogon')
+        env['REMOTE_ADDR'] = '127.0.0.1'
```
carl-codeorg, line 17: Slightly confused about this block - what is an RD country code?

carl-codeorg (approved):
LGTM, had one small comment and a question about the block that forces 127.0.0.1 for bogon IPs - assuming it must be i18n-related since it's locale-specific.

lib/cdo/rack/geolocation_override.rb
mgc1194: RD is the country code for the development environment.

mgc1194 (approved):

PR #59607, opened 2024-07-04
[Design2-129] DSCO - Colors - Add ThemeContext

levadadenys:
### [Design2-129] DSCO - Colors - Add ThemeContext

* added ThemeContext.tsx

Upon working on Design System Color variables I investigated the possibility of adding support of switching Theme, which with this soultion turns out to be pretty clean and simple in my opinion. We haven't yet designed the switch themes user flow, but once it'll be designed and we have all the DSCO colors ready for dev - we'll be ready to implement it!

This solution supports using one theme in scope of the other theme (by nesting ThemeProviders).

## Links

- jira ticket: [Design2-129](https://codedotorg.atlassian.net/browse/DESIGN2-129)

## PR Checklist:

- [ ] Code is well-commented


apps/src/componentLibrary/common/contexts/ThemeContext.tsx
```diff
@@ -0,0 +1,65 @@
...
+export const ThemeProvider = ({children}: {children: ReactNode}) => {
+  const [theme, setTheme] = useState<Theme>('Light');
+
+  const toggleTheme = () => {
+    setTheme(prevTheme => (prevTheme === 'Light' ? 'Dark' : 'Light'));
+  };
+
+  return (
+    <ThemeContext.Provider value={{theme, toggleTheme, setTheme}}>
+      <div data-theme={theme}>{children}</div>
```
sanchitmalhotra126, line 66: ~~Curious why we're using the `data-theme` attribute? I'm assuming children would just grab the value out of the context directly, but is there another use for this?~~ Edit: makes more sense after seeing https://github.com/code-dot-org/code-dot-org/pull/59609 - nice solution!

sanchitmalhotra126 (commented):
Nice! One question but otherwise this approach looks good to me

apps/src/componentLibrary/common/contexts/ThemeContext.tsx
levadadenys: Thanks for bringing it up! Added an additional comment just before return statement explaining this

apps/src/componentLibrary/common/contexts/ThemeContext.tsx
```diff
@@ -0,0 +1,69 @@
...
+  theme: Theme;
+  /** Toggle between themes (light/dark)*/
+  toggleTheme: () => void;
+  /** In case we'll have more than 2 themes - we'll be able to update theme value via this callback */
+  setTheme: (theme: Theme) => void;
+}
+
+/**
+ * Handles multiple themes logic.
+ * Can ThemeProvider can be stacked, meaning that you can have different theme in existing themeContext by
```
kelbyhawn: Minor update here:
```suggestion
 *ThemeProvider can be stacked, meaning that you can have different theme in existing themeContext by
```

kelbyhawn (approved):
One small typo on a comment, but otherwise looks great! 🎉

sanchitmalhotra126 (approved):

levadadenys:
Merging as is for now since for now it won’t be used anyways, there’re no user flows and designs for this functionality yet.
Though once we're ready - come back to it.

Upon further discussion there's a question to be solved in the future. Since we don't have one react entry point (e.g. `App.tsx`) and routing is handled by rails - we'll need to wrap every page's top level React component with ThemeContext, and also we'll need to investigate and decide optimal solution on whether to save theme value to the localStorage(keep theme preferences by browser) or to the db(backend endpoint) (keep theme preferences by user)

https://codedotorg.slack.com/archives/C043ZKQ4BS6/p1720458134634489

PR #59612, opened 2024-07-07
Datablock Storage: fix migration edge cases

snickell:
- If we see key conflicts due to MySQL making 'key ' == 'key', replace keys and table names with trailing spaces with trailing _s.
- If we detect a known chatapp pattern (keys starting with takenusername and userTaken) do not import them
- Truncate tables to max of 10 (datablock storage limit), prioritizing keeping non-shared/current tables
- When current_table name conflicts with a local table, drop the local table (they are usually empty)

### DIFF
--- a/bin/migrate_firebase_projects
+++ b/bin/migrate_firebase_projects
@@ -21,7 +21,7 @@
 # TODO: post-firebase-cleanup, remove this script #56994
 
 # If DRY_RUN: run SQL in a transaction but rollback before committing
-DRY_RUN = true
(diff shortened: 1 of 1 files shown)


snickell:
As of the last commit, this goes to completion with no errors on a local dry run:
```
bundle exec bin/migrate_firebase_projects exception-Duplicate_entry_for_key_datablock_storage_kvps_PRIMARY__key_ends_with_a_space.channel_ids
```

cnbrenci (approved):

PR #59614, opened 2024-07-08
Tidy Projects: Delete ReportAbusePopUp

Erin007:
Students no longer have the ability to publish projects to a public gallery; instead the gallery is populated with an internally curated collection of "Featured Projects" (https://github.com/code-dot-org/code-dot-org/pull/57473). Because the gallery projects are hand-selected we are reasonably confident that they do not contain abusive content, which means we no longer need to support reporting abuse from the thumbnail of `ProjectCards` in the gallery. 

This PR deletes "dead" code leftover from the above linked change.  The`showReportAbuseHeader` prop was not being used, the `renderHeader` function was no longer being called, and there was no remaining code path to show the `ReportAbusePopUp`. I deleted it all.  🗑️ 

## PR Checklist:

- [ ] Code is well-commented


molly-moen (approved):
🧹

bencodeorg (approved):
Nice! Could be a follow-up, but could we also remove the Rails-side code as well? I think it'd just be this (and the route in routes.rb):

https://github.com/code-dot-org/code-dot-org/blob/5c9fe455c01a14abea867a86697c2d96d3845e4e/dashboard/app/controllers/report_abuse_controller.rb#L36-L60

fisher-alice (approved):
Thanks for the clean-up! You rock!

PR #59615, opened 2024-07-08
[JIMT][CT-444]draggable-file-tabs - draggable file tabs

thomasoniii:
Adds draggable file tabs to CodeBridge.

Just click and drag on a file in the tabs view to slide it around and re-order it.

Note - the individual files still maintain an `open` attribute, and we probably want to deprecate and refactor that in the future to use the `openFiles` array as a single source of truth, but let's circle back to that.

For now, a file's `open` attribute determines if it's in the tab array, and `openFiles` determines the order within the file.

## Links

- jira ticket: [CT-444](https://codedotorg.atlassian.net/browse/CT-444)

## PR Checklist:

- [ ] Code is well-commented


apps/src/codebridge/FileTabs/Sortable.tsx
```diff
@@ -0,0 +1,24 @@
+import {useSortable} from '@dnd-kit/sortable';
+import {CSS} from '@dnd-kit/utilities';
+import React from 'react';
+
+interface SortableProps {
+  id: string;
+  children: React.ReactNode;
+}
+
+const Sortable: React.FunctionComponent<SortableProps> = ({id, children}) => {
```
molly-moen: would it be worth adding a comment explaining the purpose of this component? Also is this the best location for it since it's not file tabs specific (and I'm guessing we will reuse it when we do drag and drop in the file nav?)

apps/src/codebridge/utils/getOpenFiles.ts
```diff
@@ -0,0 +1,20 @@
+import {ProjectType} from '@codebridge/types';
+
+import {shouldShowFile} from './fileUtils';
+import {sortFilesByName} from './sortFilesByName';
+
+export const getOpenFiles = (project: ProjectType) => {
+  if (project.openFiles) {
```
molly-moen, line 7: can you file a follow-up task to use `openFiles` as the single source of truth?

molly-moen (commented):
Looks good! I pulled and tried it out and noticed when you drag a tab you can see two versions of the tab, one static one in the original location, and then the one you are dragging. I don't remember exactly how I did it, but in my [prototype](https://codedotorg.slack.com/archives/C06JELCAPT9/p1710869464479659) I just had the single dragging tab. Do you have a strong opinion on which version to use? I like the single tab better because you don't have two of the same tab at once.

apps/src/codebridge/FileTabs/Sortable.tsx
thomasoniii: I'm not convinced it's the best spot for it, but I'm going to defer work on moving it/explaining it further until the file browser DnD is implemented.

apps/src/codebridge/utils/getOpenFiles.ts
thomasoniii: Filed as https://codedotorg.atlassian.net/browse/CT-699

thomasoniii:
> Looks good! I pulled and tried it out and noticed when you drag a tab you can see two versions of the tab, one static one in the original location, and then the one you are dragging. I don't remember exactly how I did it, but in my [prototype](https://codedotorg.slack.com/archives/C06JELCAPT9/p1710869464479659) I just had the single dragging tab. Do you have a strong opinion on which version to use? I like the single tab better because you don't have two of the same tab at once.

I think it's better to have the placeholder, but I also twiddled it a little bit so that the existing tab in the list is much more transparent, to hopefully emphasize that it's a placeholder value. At the same time, I'll totally defer to product/design's desires on the matter.

On the topic, I haven't got a clue how you implemented it to remove it from the array anyway, so if we go that route I'd definitely like to know the magic knob to turn to make it work. 😅

PR #59620, opened 2024-07-08
Add code-studio to import order rule

lfryemason:

## PR Checklist:

- [ ] Code is well-commented


apps/webpack.config.js
```diff
@@ -80,6 +80,7 @@ const circularDependenciesSet = new Set(circularDependencies);
 // as we see our known circular dependencies, we're gonna remove them from our list. That way,
 // we can report at the end if any circular dependencies have been cleaned up.
 let seenCircles = new Set();
+let numUnresolvedCircles = 0;
```
lfryemason: Adding some extra context to output to make debugging easier.

Hamms: Would you mind extracting this change into its own PR? I like the new functionality, but I do want to keep it separate from the cleanup work that's supposed to have zero side effects. I'll also be interested to hear what @thomasoniii thinks about it!

apps/src/code-studio/pd/workshop_dashboard/components/survey_results/text_responses.jsx
```diff
@@ -1,8 +1,8 @@
+import he from 'he';
+import _ from 'lodash';
 import PropTypes from 'prop-types';
 import React from 'react';
 import {Well} from 'react-bootstrap'; // eslint-disable-line no-restricted-imports
-import _ from 'lodash';
-import he from 'he';
```
Hamms, line 5: Total aside, but: looks like we have a worthy challenger to the Go programming language in the "least-searchable name of all time" competition

Hamms (commented):
Wooo, `code-studio`! Definitely feels like the most substantial of the remaining directories :tada: 

I'd ideally like to see them each in their own PR, but both the updates to circular dependency functionality and the import reorderings look good to me individually.

Hamms (approved):
LGTM; thank you!

PR #59623, opened 2024-07-08
Update Maze UI tests to use Google Blockly

mikeharv:
We have a bunch of UI tests that check various site functionality which assume that we can pass a simple Maze level. Because of differences in how each version of Blockly handles block ids, we need to update these tests and their associated levels before we can migration Maze to Google Blockly.

Google Blockly uses a different method to assign ids and uses a different attribute on the svg element. It also can potentially re-use an id of a toolbox block when it is dragged to the workspace.

**Review of key differences**
||CDO Blockly|Google Blockly|
|-|-|-|
|ID attribute|`block-id`|`data-id`|
|ID values| assigned sequentially (1, 2, 3,...), unless hard-coded|string of random characters, unless hard-coded|
|ID reuse for new blocks from flyout|New block ids are assigned sequentially, which makes them predicatable|New block ids will match the flyout block's id unless it's already in use. Subsequent copies of the same block will still have unique/random ids.|

For example a block might have `block-id="4"` in CDO Blockly, but `data-id="[{UiJ(GM._~}=@l[Iog!"` in Google Blockly. By adding an `id=` attribute to the block XML, we can override the ID that is used. For test readability, I opted to use descriptive ids like "moveWest" instead of the seemingly arbitrary integers that CDO Blockly would generate. (This contrasts with how our callouts script modifies XML automatically, since it's not smart enough to come up with descriptive ids.)

In most cases, the changes described above were sufficient to get the test steps working. However, in cases where a user needs to drag two of the same block out, it gets a little tricky. For example, several tests use a level where you are expected to connect to "move west" blocks. After dragging the first "move west" block to the "when run" block, you end up with two "move west" blocks that both have the same `"moveWest"` id (one in the toolbox, one on the workspace):
image

When we would generate drag code for the second "move west" block, the id selector was finding the workspace block and re-dragging it, rather than finding the toolbox block as expected. This is because Google Blockly's workspace blocks show up in the DOM earlier than the flyout blocks. Essentially, `$("[data-id='moveWest']")` finds the wrong block. As a fix for this, we can append `:last` to the selector for the block to be dragged. This shouldn't affect any tests still running CDO Blockly because there should only be one block in the list. 

## Deployment strategy

This PR will immediately cause UI tests to use Google Blockly - ahead of the migration. There should be no impact to users, but we can expect to need to update baselines for the eyes tests.

Once Maze has been migrated (and we are confident a revert will not be needed), we can remove the `?blocklyVersion=google` param from each URL.

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/dashboard/config/levels/custom/maze/2-3 Maze 1.level
+++ b/dashboard/config/levels/custom/maze/2-3 Maze 1.level
@@ -1,7 +1,8 @@
 <Maze>
   <config><![CDATA[{
+  "published": true,
   "game_id": 25,
-  "created_at": "2014-05-07T19:06:41.000Z",
+  "created_at": "2022-02-16T19:37:58.000Z",
   "level_num": "custom",
   "user_id": 19,
   "properties": {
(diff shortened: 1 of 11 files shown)


dashboard/config/levels/custom/maze/2-3 Maze 1.level
```diff
@@ -30,19 +31,30 @@
...
   },
-  "published": true,
   "notes": "",
+  "audit_log": "[{\"changed_at\":\"2024-07-08 15:13:58 -0400\",\"changed\":[\"toolbox_blocks\",\"recommended_blocks\",\"solution_blocks\",\"hint_prompt_attempts_threshold\",\"contained_level_names\"],\"changed_by_id\":1,\"changed_by_email\":\"mike@code.org\"}]",
   "level_concept_difficulty": {
     "sequencing": 2
   }
 }]]></config>
   <blocks>
+    <start_blocks>
```
molly-moen, line 50: Did these levels have implicit "when_run" blocks before?

molly-moen (approved):
Looks good, one question on the level configuration

dashboard/config/levels/custom/maze/2-3 Maze 1.level
mikeharv: Yup!

PR #59624, opened 2024-07-08
Prevent students in CPA flow from changing state and age backend

juanmanzojr:
Follow up to this [PR](https://github.com/code-dot-org/code-dot-org/pull/59437). Prevents students who are in CPA flow from changing their `us_state` and/or `age` in the backend.

## Links

- jira ticket: [here](https://codedotorg.atlassian.net/jira/software/c/projects/P20/boards/69?assignee=712020%3Af0e2716f-09f9-4686-8490-9f1f9a08bb06&selectedIssue=P20-1021)

## Testing story
New test within `user_test.rb`

### DIFF
--- a/dashboard/app/models/user.rb
+++ b/dashboard/app/models/user.rb
@@ -2774,8 +2774,7 @@ class User < ApplicationRecord
 
   private def enforce_age_or_state_update
     # Create copy of user to mock the user's state before an update.
-    user_before_update = User.new(attributes.deep_merge(changed_attributes))
-
+    user_before_update = User.new(attributes.merge(changed_attributes))
     potentially_locked = Policies::ChildAccount.underage?(user_before_update)
     # The student is in a 'lockout' flow if they are potentially locked out and not unlocked
     if potentially_locked && !Policies::ChildAccount::ComplianceState.permission_granted?(user_before_update)
--- a/dashboard/app/models/user.rb
+++ b/dashboard/app/models/user.rb
@@ -305,6 +305,8 @@ class User < ApplicationRecord
     self.gender = Policies::Gender.normalize gender
   end
 
+  before_validation :student_in_lockout_flow?, on: :update
+
   validate :validate_us_state, if: :should_validate_us_state?
 
   before_validation on: [:create, :update], if: -> {gender_teacher_input.present? && will_save_change_to_attribute?('properties')} do
@@ -2762,6 +2764,24 @@ class User < ApplicationRecord
     # us_state is only a required field if the User lives in the US.
     return false unless %w[US RD].include? country_code
     new_record? || us_state_changed?
+  end
+
+  def student_in_lockout_flow?
+    return unless should_validate_us_state?
+    return unless birthday_changed? || us_state_changed?
+
+    # Create copy of user to mock the user's state before an update.
+    user_before_update = dup
+
+    # If us_state or age is changed, set to previous value before the update to be used to check if the user is in the lockout flow.
+    user_before_update.us_state = properties_was['us_state'] if us_state_changed?
+    user_before_update.birthday = birthday_was if birthday_changed?
+
+    potentially_locked = Policies::ChildAccount.underage?(user_before_update)
+    # The student is in a 'lockout' flow if they are potentially locked out and not unlocked
+    if potentially_locked && !Policies::ChildAccount::ComplianceState.permission_granted?(user_before_update)
+      errors.add(:us_state, "LOCKOUT FLOW")
+    end
   end
 
   private def ai_tutor_feature_globally_disabled?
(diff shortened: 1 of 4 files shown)


dashboard/app/models/user.rb
```diff
@@ -2764,6 +2766,26 @@ def should_validate_us_state?
     new_record? || us_state_changed?
   end
 
+  def student_in_lockout_flow?
+    return unless user_type == TYPE_STUDENT
```
wilkie:
```suggestion
    return unless user.student?
```

dashboard/app/models/user.rb
```diff
@@ -2764,6 +2766,26 @@ def should_validate_us_state?
     new_record? || us_state_changed?
   end
 
+  def student_in_lockout_flow?
```
wilkie: I would make the validation function private since you do not intend it to be called on its own

dashboard/app/models/user.rb
```diff
@@ -305,6 +305,8 @@ class User < ApplicationRecord
     self.gender = Policies::Gender.normalize gender
   end
 
+  before_validation :student_in_lockout_flow?, on: :update
```
wilkie: Let's put the gates in their own method and change the name of the validation function to make its purpose, perhaps.

The name `student_in_lockout_flow?` seems to indicate that it is returning a boolean true/false and that it is informational rather than a validation.

```suggestion
  before_validation :enforce_age_or_state_update, on: :update, if: :should_check_age_or_state_update?
```

dashboard/app/models/user.rb
```diff
@@ -2764,6 +2766,26 @@ def should_validate_us_state?
     new_record? || us_state_changed?
   end
 
+  def student_in_lockout_flow?
+    return unless user_type == TYPE_STUDENT
+    return unless %w[US RD].include? country_code
+    return unless birthday_changed? || us_state_changed?
```
wilkie: I think it might be good to move these returns to their own function, perhaps, something like `should_check_age_or_state_update?` or similar to separate concerns. It's debatable. It's nice to know that a function that is called for every `User` update is not likely to run... and a good way to do that is separate the concern by putting the rule for when it runs or not into its own place and _that_ gets called every update, but provides a better guarantee that the function doing the destructive work is limited. Then this function only has one and only one purpose: to error out if something is done that isn't allowed.

dashboard/app/models/user.rb
wilkie: It should also not be a `?` function since that implies it returns a boolean. I think the name should be updated to reflect its purpose... I recommended above something like `enforce_age_or_state_update`

dashboard/app/models/user.rb
wilkie: If that feels like over-engineering, you can just introduce the block in the `if` via an anonymous function, but I feel like the logic is complicated enough to warrant a named function. But, to compare, it would be something like:

```
before_validation :enforce_age_or_state_update, on: :update, if: lambda do
  return false unless user.student?
  return false unless %w[US RD].include? country_code
  birthday_changed? || us_state_changed?
end
```

dashboard/app/models/user.rb
```diff
@@ -2764,6 +2766,26 @@ def should_validate_us_state?
...
+
+    # Create copy of user to mock the user's state before an update.
+    user_before_update = dup
+
+    # If us_state or age is changed, set to previous value before the update to be used to check if the user is in the lockout flow.
+    user_before_update.us_state = properties_was['us_state'] if us_state_changed?
+    user_before_update.birthday = birthday_was if birthday_changed?
```
wilkie:
```ruby
user_before_update.attributes = {
  us_state: property_before_save('us_state'),
  birthday: birthday_was
}
```

Or, I haven't quite tested this interaction, but I think this works instead of the dup and setting each property, just:

```ruby
user_before_update = User.new(attributes.deep_merge(changed_attributes))
```

juanmanzojr: Ill give it a try Thanks Wilkie!

dashboard/config/locales/en.yml
```diff
@@ -45,6 +45,7 @@ en:
               invalid: "is invalid. Note that each link can only be used once to reset your password, and only the most recently sent link will work."
             unlock_token:
               invalid: "is invalid or expired, possibly because your account has already been unlocked."
+            lockout_flow: "cannot be updated when in lockout flow."
```
wilkie: The error should indicate what you need to do, so let's say "cannot be updated because you need parent or guardian permission first."

wilkie:
A couple of flaky UI tests happen to fail, but also one relevant one:

image

image

Seems like it is triggering the block unintentionally on the partially locked out case. I probably am wrong about how to properly build the 'old' version of the User. I will look into this by manually pulling down the branch and/or pair on it at a more reasonable hour.

wilkie (approved):
LGTM! 👍 Thanks for seeing this through and for the great work!

PR #59632, opened 2024-07-08
AI Chat: add tooltips to customization fields

sanchitmalhotra126:
Adds DSCO tooltips to AI customization fields.

Screenshot 2024-07-10 at 9 35 56 AM
Screenshot 2024-07-10 at 9 36 06 AM

Using tab navigation:

Screenshot 2024-07-10 at 9 36 16 AM

## Links

https://codedotorg.atlassian.net/browse/LABS-809

## Testing story

Tested on pilot levels.


apps/src/aichat/views/modelCustomization/SetupCustomization.tsx
```diff
@@ -78,25 +79,38 @@ const SetupCustomization: React.FunctionComponent = () => {
...
+            size: 's',
+            tooltipId: 'selected-model-tooltip',
+            direction: 'onLeft',
+            className: styles.tooltip,
+          }}
+          tooltipOverlayClassName={styles['full-width']}
+        >
+          <div className={styles['full-width']}>
+            <SimpleDropdown
+              labelText="Selected model:"
```
fisher-alice: Just curious if there is a way we can add a tooltip to the label text and not both the label text and ropdown. Thanks!
cc: @levadadenys

sanchitmalhotra126:
After discussion with product/curriculum, I added icons to the right of labels as hover targets for the tooltips. This is hopefully less disruptive and clearer.

apps/src/aichat/views/modelCustomization/SetupCustomization.tsx
sanchitmalhotra126: Updated this to hide the default label, so we can display a separate label element with the icon. That way the tooltip doesn't appear when hovering the actual dropdown.

apps/src/aichat/views/presentation/ModelCardRow.tsx
```diff
@@ -7,14 +7,16 @@ import moduleStyles from './model-card-row.module.scss';
...
+  titleIcon?: string;
   expandedContent: string | string[];
+  tooltipText?: string;
 }
 
 const ModelCardRow: React.FunctionComponent<ModelCardRowProps> = ({
   title,
   titleIcon,
   expandedContent,
+  tooltipText, // TODO: Use tooltip text
```
sanchitmalhotra126, line 19: Planning on following up with a separate change to display the tooltips in the Presentation view. This is because we'll need to do a bit of refactoring with the `CollapsibleSection` to accommodate the tooltip info icon.

apps/src/aichat/views/model-customization-workspace.module.scss
```diff
@@ -95,3 +96,30 @@
...
+  from {
+    opacity: 0;
+  }
+  to {
+    opacity: 1;
+  }
+}
+
+.tooltip {
+  animation: fade-in 0.3s;
```
fisher-alice, line 114: As discussed via [Slack](https://codedotorg.slack.com/archives/C06FELVTV0Q/p1720624390509899?thread_ts=1720476078.818049&cid=C06FELVTV0Q)
could we left-align the text in the tooltips since the descriptions are a bit lengthy?

apps/src/aichat/views/modelCustomization/FieldLabel.tsx
```diff
@@ -0,0 +1,44 @@
...
+      <WithTooltip
+        tooltipProps={{
+          text: tooltipText,
+          size: 's',
+          tooltipId: `${id}-tooltip`,
+          direction: 'onRight',
+          className: styles.tooltip,
+        }}
+      >
+        <div>
```
fisher-alice: Could we make these buttons so that they are clickable and tab-navigable?

fisher-alice (commented):
Just a couple minor suggestions based on product/design considerations. Thanks for dong this!

apps/src/aichat/views/model-customization-workspace.module.scss
sanchitmalhotra126: Done!

apps/src/aichat/views/modelCustomization/FieldLabel.tsx
sanchitmalhotra126: Good call, done!

fisher-alice: Awesome! Just noting that when keyboard-navigating, the tooltip is immediately displayed when the user navigates to the icon so being clickable is not necessary - this seems right to me! Just confirming with  @samantha-code.

fisher-alice (approved):
Thanks for the iteration on this including the keyboard navigation of the tooltips! 🎉

apps/src/aichat/views/modelCustomization/FieldLabel.tsx
sanchitmalhotra126: Yeah I think that also sounds right to me? If we want the tooltip to appear only after clicking/pressing enter, that's separate functionality we'd have to implement in the Tooltip component. I think we'd still want to wrap it in a `<button>` though so it's a tab-navigation target.

PR #59633, opened 2024-07-08
Add a general aws_access? function to guard features that require AWS in development

wilkie:
This adds the `CDO.aws_access?` method that should be used to indicate that the running web server should enable AWS-only features.

This also adds this guard to a few places that are generally blocking for non-AWS-authenticated developers:

* AI metadata validation for Learning Goals (breaks seeding)
* The video generation feature in dance party
* PDF checking for link generation on lesson pages

## Links

- jira initiative: [P20-766](https://codedotorg.atlassian.net/browse/P20-766)

## PR Checklist:

- [ ] Code is well-commented


lib/cdo.rb
```diff
@@ -290,6 +290,11 @@ def ci_webserver?
       running_web_application? && ENV['CI']
     end
 
+    # The web server instance has some intention to use AWS credentials?
+    def aws_access?
+      rack_env?(:production) || !!ENV['AWS_PROFILE']&.include?('cdo')
```
snickell, line 295: This part is the only thing that made me sit down and think:
```
!!ENV['AWS_PROFILE']&.include?('cdo')
```

Pros: great simplicity and therefore predictability, works for everyone who uses a stock setup
Cons: keying off the AWS profile name set in the env var is a tiny bit hmm, e.g. if you configure the AWS profile as named 'default', you don't even need an env var set.

In an ideal world, I wish there was a "duck typing" way to sniff if we're connected to a "CDO-like" AWS instance. I don't think its worth doing a big amount of work to do this, but if it was easy, I think its a little better than keying off a specific way we currently instruct people to setup AWS. I would suggest seeing if we can access one of the AWS secret names we know we define in all environments.

I guess the philosophical question here is if `aws_access?` means "do we have AWS access" or "is there a chance the person wanted to use AWS access?". e.g. if I have AWS_PROFILE=cdo setup, and I'm on a plane without internet (can't access AWS), should it just quietly fallback to non-AWS mode?

I think what you've got here is safe and simple, so overall thinking this through, I think it might be the right tradeoff. We can always complexify this later.

snickell (approved):
I'm liking everything I'm seeing 100%, only one comment, and its more a "sit and think" rather than coming to a different conclusion.

lib/cdo.rb
wilkie: This is a fair point and I think I'm indeed playing it a little quick and dirty here, and I'm definitely looking for guidance on a better approach, if any.

For one, the potential follow-up is the adding of a `CDO.aws_emulated?` guard as well which allows the local fallback to some AWS emulation running alongside your development environment via MinIO/Localstack, etc. This complicates things because that is technically access to some kind of "AWS" but not the same as access to the AWS service proper. In my experience, that hasn't felt very confusing because you still have things that Localstack, etc, cannot do that needs the main guard (cookie signing, restricted content, etc) and then the fallback for things the emulation does pretty well.

Back to your main point... Yes, it is potentially better to try to 'touch' AWS somehow to prove the credentials are valid and then cache that value. Something like what `./bin/aws_access` does where we just use that as a source of truth once and continue on. I wonder if that process is quick enough to not make short scripts annoying, but I suspect it is. I might try to do that (and still assume the `production` environment is always AWS-connected)

PR #59634, opened 2024-07-08
Lab2: minor container style cleanup

sanchitmalhotra126:
Two small style cleanups on the lab2 container and some individual containers to fix some cross-lab style quirks. None of this is user facing and only manifests when there are multiple different lab types in the same progression (e.g. in the lab2 showcase allthethings lesson), but should keep things tidy as we add new labs and update styles.
- Removed the `#codeApp` wrapper since it functionally wasn't doing anything (the lab2 container uses `position: fixed` so it ignores all of `#codeApp`'s positioning)
- Change the top spacing from 52px to 50px - the height of the header is 50px, and I believe we chose 52px to give a small 2px gap between the header and the lab content. However, this means that individual labs don't have control over the background color that appears in that small spacing, which is especially apparent when moving from a dark mode lab to a light mode lab, and vice versa. For example, when starting the lab2 all the things progression on level 1 (Music Lab), and then going to level 2 (AI Chat) you'll see a small black bar above the lab in the 2px space:
  - Screenshot 2024-07-08 at 3 41 43 PM
  - And vice versa:
  - Screenshot 2024-07-08 at 3 42 03 PM
  - By changing it to 50px, we can instead move the 2px spacing inside the lab, where the lab's own background color will fill the space according to its desired theme.

## Testing story

Tested on allthethings Lab2 showcase progression

### DIFF
--- a/dashboard/app/views/levels/_lab2.html.haml
+++ b/dashboard/app/views/levels/_lab2.html.haml
@@ -1,5 +1,4 @@
-#codeApp{style: "position: absolute; top: 60px; bottom: 10px; left: 10px; right: 10px"}
-  #lab2-container{style: "height: 100%"}
+#lab2-container
 
 - content_for(:head) do
   %script{src: webpack_asset_path("js/#{js_locale}/music_locale.js")}
--- a/apps/src/lab2/views/Lab2Wrapper.module.scss
+++ b/apps/src/lab2/views/Lab2Wrapper.module.scss
@@ -3,8 +3,8 @@
 
 .labContainer {
   position: fixed;
-  top: 52px;
-  bottom: 2px;
+  top: $top-spacing;
+  bottom: 0;
   left: 0;
   right: 0;
   font-size: 13px;
--- a/apps/src/lab2/views/common.scss
+++ b/apps/src/lab2/views/common.scss
@@ -14,3 +14,5 @@
   left: 0;
   z-index: 80;
 }
+
+$top-spacing: 50px;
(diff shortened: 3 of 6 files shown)


dashboard/app/views/levels/_lab2.html.haml
```diff
@@ -1,5 +1,4 @@
-#codeApp{style: "position: absolute; top: 60px; bottom: 10px; left: 10px; right: 10px"}
```
molly-moen, line 1: there's a bunch of styles in `application.scss` (and other places) that are keyed on `#codeApp`, do we not need those?

sanchitmalhotra126: Yeah as far as I can tell none of these are really affected/used by Lab2? Since the Lab2 container uses `position: fixed`, it seems to ignore all the margin and offset applied to `#codeApp` and the other usages seem to be lab-specific (like Applab, oceans, jigsaw, etc). StudioApp/code-studio.js do some of manipulation with the #codeApp element but again shouldn't affect Lab2 since we don't use those classes.

That being said if it's safer to be consistent, we can definitely still name the top level container #codeApp, shouldn't have any effect on the lab2 container either way.

thomasoniii (approved):
lgtm 🚀

dashboard/app/views/levels/_lab2.html.haml
molly-moen: If we don't use the styles seems good to remove it!

molly-moen (approved):

PR #59637, opened 2024-07-09
P20-987: Add UI tests for DCDO cookies mocking

artem-vavilov:
Due to an issue on the DTT server causing the test to fail, we decided to deploy https://github.com/code-dot-org/code-dot-org/pull/59442 without the UI test and then run it locally to diagnose the issue.

## Links
- Related task: https://github.com/code-dot-org/code-dot-org/pull/59442


Hamms (approved):
I mentioned this in slack, but repeating here:

This change looks good to me, but because we still haven't been able to actually test this UI test, we could consider extracting just the changes to `test_dcdo.haml` and `config.ru` into another PR which when merged into test should allow us to run the UI test locally and verify that it won't fail before merging.

It's *probably* not necessary, and it wouldn't be a big deal even if we do need to revert and try again. But there have been enough surprises with this one that I'm feeling paranoid. It's your call, though!

Hamms (approved):
I locally kicked off a run of this UI test in saucelabs against the actual test server, and with a few tweaks in place to accommodate the actual test server not yet having the `#fetched_dcdo_value` id, it now passes!

image

Thank you for all your diligence on this one

PR #59638, opened 2024-07-09
AI Tutor: bulk response generator

Erin007:
[CT-673](https://codedotorg.atlassian.net/browse/CT-673)
[Problem summary](https://docs.google.com/document/d/1bel4ih2kztxxluwzjuNMK8Y30-xPvrEJtcSohLpyoJ4/edit)

MVP of a tool that would allow levelbuilders to upload a csv of student inputs and bulk generate AI Tutor responses for the purposes of iterating an testing. 

If the user does not have levelbuilder and ai tutor access permissions, they will see a warning message and everything will be disabled. 

Screenshot 2024-07-11 at 12 24 14 PM

If the user has the correct permissions, they can follow the sequence of button clicks to generate and download a csv of bulk AI Tutor responses. 

https://github.com/code-dot-org/code-dot-org/assets/12300669/f486b14e-787d-4fbe-b4e7-5460352af21b

As an example for the proof of concept I used this [reference](https://cs-people.bu.edu/dgs/courses/cs460/problem_sets/common_errors.shtml) to make a [small sample of 20 common Java compilation errors](https://docs.google.com/spreadsheets/d/1BeUT3u_mMFdEz7fOFvNWI0ukw5WcX9OGIn2S3YOYjh8/edit?gid=285183436#gid=285183436), ran it through the AI Tutor Response Generator and got these [responses](https://docs.google.com/spreadsheets/d/1rAN0oeXzP7K1i5hKbRB4eLW4WR6WFYOd7nzLH6Dj-EI/edit?gid=735410839#gid=735410839) back.

## Follow-up work

This is a true MVP! We'll live with it for a bit and see what changes make sense or if we even want to keep it.  

My guesses: 
- Friendly error messages for when things don't work as expected. Since this is the first rev of a tool that will be used internally and used initially by _maybe_ 3 people all of whom I can directly troubleshoot with I don't want to invest too much in building this out (yet)
         - column names aren't as expected
         - data isn't formatted as expected
         - upload fails
         - response from ai fails 
         - etc
- The ability to include `levelId` so we can use it pull level instructions and tests for validated levels on the backend to include in the payload to AI 
- The ability to iterate on system prompts 
- The ability to swap out different AI models
- A (truncated?) table to preview the AI responses 
- Ability to [stream](https://www.papaparse.com/docs#config) in data from the csv 
- Some kind of limiting of how many responses you can request from AI at a time 

## PR Checklist:

- [ ] Code is well-commented


apps/src/lib/levelbuilder/ai-tutor/AITutorResponseGenerator.tsx
```diff
@@ -0,0 +1,115 @@
...
+      <div style={{marginBottom: '10px'}}>
+        <Button
+          text="Get AI Tutor Responses"
+          onClick={getAIResponses}
+          disabled={!dataUploaded}
+        />
+        <span>
+          {responseCount} of {data.length}
+        </span>
+      </div>
```
Erin007: - [x] add pending state and spinner to button

dashboard/config/routes.rb
```diff
@@ -63,6 +63,8 @@
 
     resources :images, only: [:new]
 
+    get "/ai_tutor/generate_test_responses", to: "ai_tutor#generate_test_responses"
```
Erin007, line 67: - [x] should be blocked for everyone except levelbuilders

apps/src/lib/levelbuilder/ai-tutor/AITutorResponseGenerator.tsx
```diff
@@ -0,0 +1,115 @@
...
+  const aiResponded = dataUploaded && data.length === responseCount;
+
+  return (
+    <div>
+      <h2>Generate AI Tutor Responses</h2>
+      <p>
+        This allows you to upload a CSV of student inputs that will be sent to
+        AI Tutor. AI Tutor responses will then be saved and you can download the
+        resulting updated CSV. Your CSV must include a `studentInput` column to
+        succeed.
```
Erin007: - [ ] add some kind of reminder that we're hitting a real endpoint, $$ and what a reasonable number of test cases might be and/or add a limit to the number of rows that can be uploaded

apps/src/lib/levelbuilder/ai-tutor/AITutorResponseGenerator.tsx
```diff
@@ -0,0 +1,115 @@
...
+      </p>
+      <div style={{marginBottom: '10px'}}>
```
Erin007: - [ ] extract into styles module

apps/src/lib/levelbuilder/ai-tutor/AITutorResponseGenerator.tsx
```diff
@@ -0,0 +1,115 @@
...
+  };
+
+  const askAI = async (prompt: AIInteraction) => {
+    const chatApiResponse = await getChatCompletionMessage(
+      prompt.studentInput,
+      []
+    );
+    prompt.aiResponse = chatApiResponse.assistantResponse;
+    setResponseCount(prevResponseCount => prevResponseCount + 1);
+  };
```
Erin007, line 64: - [ ] error handling

apps/src/aiTutor/views/AITutorSuggestedPrompts.tsx
```diff
@@ -9,11 +9,12 @@ import {EVENTS} from '@cdo/apps/lib/util/AnalyticsConstants';
...
 
+import {genericCompilation, genericValidation} from '../constants';
 import {askAITutor} from '../redux/aiTutorRedux';
 
 const QuickActions = {
-  [ActionType.COMPILATION]: "Why doesn't my code compile?",
-  [ActionType.VALIDATION]: "Why aren't my tests passing?",
+  [ActionType.COMPILATION]: genericCompilation,
+  [ActionType.VALIDATION]: genericValidation,
 };
```
Erin007: - [x] move to its own PR

apps/package.json
```diff
@@ -289,6 +290,7 @@
     "react-motion": "^0.5.2",
     "react-onclickoutside": "~5.11.1",
     "react-paginate": "^6.3.0",
+    "react-papaparse": "^4.4.0",
```
Erin007, line 299: How do we feel about adding new packages? fwiw we used this same [package]((https://github.com/code-dot-org/ml-playground/blob/07f5171195dae794ada82e3e5b1176a84e292659/package.json#L64) for AI Lab

apps/src/lib/levelbuilder/ai-tutor/AITutorResponseGenerator.tsx
```diff
@@ -0,0 +1,115 @@
...
+      </div>
+
+      <div>
+        <Button
+          text=" Download CSV"
+          onClick={downloadCSV}
+          disabled={!aiResponded}
+        />
+      </div>
+    </div>
```
Erin007, line 139: Would it be useful to have a (truncated?) preview table in addition to ability to download?

apps/src/lib/levelbuilder/ai-tutor/AITutorResponseGenerator.tsx
```diff
@@ -0,0 +1,115 @@
...
+  };
+
+  const importCSV = () => {
+    if (csvFile) {
+      Papa.parse<AIInteraction>(csvFile, {
+        complete: updateData,
+        header: true,
+      });
+    }
+  };
```
Erin007, line 44: - maybe we want to stream these? depending on how many we think there might be

apps/package.json
molly-moen: I think we are generally fine with it? That one looks widely used and has been updated in the last year, which both seem like good signs.

dashboard/app/controllers/ai_tutor_controller.rb
```diff
@@ -0,0 +1,4 @@
+class AiTutorController < ApplicationController
+    def generate_test_responses
```
molly-moen: is this needed?

dashboard/app/models/user.rb
```diff
@@ -1518,6 +1518,10 @@ def verified_teacher?
     permission?(UserPermission::AUTHORIZED_TEACHER)
   end
 
+  def levelbuilder?
```
molly-moen, line 1521: I can't believe this didn't exist yet!

molly-moen (approved):
Looks good!

dashboard/app/controllers/ai_tutor_controller.rb
Erin007: yeah, it's one of those things Rails is opinionated about. When you have the route it expects the controller to exist and there to be a method that matches the name of the action in the route. I guess I could push some of the logic, like permissions checking into the the controller so it's not an empty method, but I think it's fine.

dashboard/app/models/user.rb
Erin007: same!

PR #59642, opened 2024-07-09
P20-1030 Refactor i18n unit tests to minitest/spec

alex-m-brown:
Converted unit tests in `test_18n_script_utils.rb` from minitest/unit to minitest/spec format. Tests reformatted:

- `to_crowdin_yaml`
- `log_error`
- `fix_yml_file`
- `to_js_locale`

## Links

- Jira Ticket: [P20-1030](https://codedotorg.atlassian.net/browse/P20-1030)

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/bin/test/i18n/test_i18n_script_utils.rb
+++ b/bin/test/i18n/test_i18n_script_utils.rb
@@ -122,19 +122,15 @@ describe I18nScriptUtils do
     let(:provided_yaml_file_data) {"---\nen-US:\n  data\n"}
     let(:expected_yaml_output) {%Q["en-US":\n  data\n]}
 
-    before do
-      FileUtils.mkdir_p File.dirname(provided_yaml_file_path)
-      File.write provided_yaml_file_path, provided_yaml_file_data
-    end
-
     it 'yml file should be reformatted correctly' do
+      File.expects(:read).with(provided_yaml_file_path).returns(provided_yaml_file_data)
+      File.expects(:write).with(provided_yaml_file_path, expected_yaml_output)
       _(fix_yml_file)
-      expect(File.read(provided_yaml_file_path)).must_equal expected_yaml_output
     end
   end
 
   describe '.to_js_locale' do
-    let(:to_js_locale) {I18nScriptUtils.to_js_locale(locale)}
+    let(:to_js_locale) {described_class.to_js_locale(locale)}
 
     let(:locale) {'en-US'}
     let(:expected_locale_string) {'en_us'}
--- a/bin/test/i18n/test_i18n_script_utils.rb
+++ b/bin/test/i18n/test_i18n_script_utils.rb
@@ -122,10 +122,15 @@ describe I18nScriptUtils do
     let(:provided_yaml_file_data) {"---\nen-US:\n  data\n"}
     let(:expected_yaml_output) {%Q["en-US":\n  data\n]}
 
+    before do
+      FileUtils.mkdir_p File.dirname(provided_yaml_file_path)
+      File.write provided_yaml_file_path, provided_yaml_file_data
+    end
+
     it 'yml file should be reformatted correctly' do
-      File.expects(:read).with(provided_yaml_file_path).returns(provided_yaml_file_data)
-      File.expects(:write).with(provided_yaml_file_path, expected_yaml_output)
-      _(fix_yml_file)
+      assert_changes -> {File.read(provided_yaml_file_path)}, from: provided_yaml_file_data, to: expected_yaml_output do
+        fix_yml_file
+      end
     end
   end
 
--- a/bin/test/i18n/test_i18n_script_utils.rb
+++ b/bin/test/i18n/test_i18n_script_utils.rb
@@ -1,34 +1,6 @@
 require_relative '../test_helper'
 require_relative '../../i18n/i18n_script_utils'
 
-class I18nScriptUtilsTest < Minitest::Test
-  def test_to_crowdin_yaml
-    assert_equal "---\n:en:\n  test: \"#example\"\n  'yes': 'y'\n", I18nScriptUtils.to_crowdin_yaml({en: {'test' => '#example', 'yes' => 'y'}})
-  end
-
-  def test_error_logging
-    expected_error_class = 'expected_error_class'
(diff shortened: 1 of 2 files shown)


bin/test/i18n/test_i18n_script_utils.rb
```diff
@@ -120,6 +103,47 @@ def test_to_js_locale_returns_formated_js_locale
...
+  end
+
+  describe '.to_js_locale' do
+    let(:to_js_locale) {I18nScriptUtils.to_js_locale(locale)}
+
+    let(:locale) {'en-US'}
+    let(:expected_locale_string) {'en_us'}
+
+    it 'locale string should be formatted correctly' do
+      _(to_js_locale).must_equal expected_locale_string
```
mgc1194, line 144: I see there are tests using different patterns. I typically would prefer a more explicit pattern where you can see the method being called, but probably there are benefits of this patter I'm not aware of. I'll do some research.

mgc1194 (approved):
Great job, thanks for updating these tests, much more readable now.

bin/test/i18n/test_i18n_script_utils.rb
```diff
@@ -120,6 +103,47 @@ def test_to_js_locale_returns_formated_js_locale
...
+    end
+
+    it 'yml file should be reformatted correctly' do
+      _(fix_yml_file)
+      expect(File.read(provided_yaml_file_path)).must_equal expected_yaml_output
+    end
+  end
+
+  describe '.to_js_locale' do
+    let(:to_js_locale) {I18nScriptUtils.to_js_locale(locale)}
```
wilkie: sometimes you are using `described_class` and sometimes the class name itself. let's pick one, but your choice which.

bin/test/i18n/test_i18n_script_utils.rb
wilkie: this pattern works well for when you are instantiating classes so you can alter parameters within context blocks. it isn't being shown off very well here because this is a static module, but it is fine and maybe nice to be consistent and use it here as well, I think.

bin/test/i18n/test_i18n_script_utils.rb
```diff
@@ -120,6 +103,47 @@ def test_to_js_locale_returns_formated_js_locale
...
+    let(:expected_yaml_output) {%Q["en-US":\n  data\n]}
+
+    before do
+      FileUtils.mkdir_p File.dirname(provided_yaml_file_path)
+      File.write provided_yaml_file_path, provided_yaml_file_data
+    end
+
+    it 'yml file should be reformatted correctly' do
+      _(fix_yml_file)
+      expect(File.read(provided_yaml_file_path)).must_equal expected_yaml_output
```
wilkie: I think we lost a stub on the read call here. we should do what the original tests do and just set up call expectations for read and write methods (so we don't actually write a file) and just call the method after setting those up.

bin/test/i18n/test_i18n_script_utils.rb
```diff
@@ -120,6 +103,47 @@ def test_to_js_locale_returns_formated_js_locale
...
+
+  describe '.fix_yml_file' do
+    let(:fix_yml_file) {described_class.fix_yml_file(provided_yaml_file_path)}
+
+    let(:provided_yaml_file_path) {'provided_yaml_file_path'}
+    let(:provided_yaml_file_data) {"---\nen-US:\n  data\n"}
+    let(:expected_yaml_output) {%Q["en-US":\n  data\n]}
+
+    before do
+      FileUtils.mkdir_p File.dirname(provided_yaml_file_path)
```
wilkie, line 126: let's avoid actual disk access and stub out the file write. see my comment about setting up call expectations

artem-vavilov: We use temporary (fake) dir for these tests:
https://github.com/code-dot-org/code-dot-org/blob/9ce543143363aa180b228392ee32d4c007a20b83/bin/test/i18n/test_i18n_script_utils.rb#L7-L9

bin/test/i18n/test_i18n_script_utils.rb
artem-vavilov: A benefit of having a testing method in a separate function is that the returned value is preserved on the first call, allowing for consistent results. Additionally, it makes it easier to modify a single method if changes are needed, rather than updating each test individually, adhering to the DRY (Don't Repeat Yourself) principle.

bin/test/i18n/test_i18n_script_utils.rb
```diff
@@ -120,6 +103,47 @@ def test_to_js_locale_returns_formated_js_locale
...
+      File.write provided_yaml_file_path, provided_yaml_file_data
+    end
+
+    it 'yml file should be reformatted correctly' do
+      _(fix_yml_file)
+      expect(File.read(provided_yaml_file_path)).must_equal expected_yaml_output
+    end
```
artem-vavilov: Let's try to use `assert_changes` here
```suggestion
    it 'yml file should be reformatted correctly' do
      assert_changes -> {File.read(provided_yaml_file_path)}, from: provided_yaml_file_data, to: expected_yaml_output do
        fix_yml_file
      end
    end
```

alex-m-brown: Should this test use what I originally had or what is in the latest commit with the File.expects calls?

bin/test/i18n/test_i18n_script_utils.rb
wilkie: ah good call. yep. thank you.

bin/test/i18n/test_i18n_script_utils.rb
wilkie: what you had originally since you have the FakeFS

wilkie: and the assert_changes that Artem is suggesting

wilkie: apologies for the cognitive whiplash. i missed the FakeFS which would mock the disk usage appropriately already.

wilkie (approved):
LGTM! 👍 

Thanks for cleaning up our tests! Great work!

PR #59646, opened 2024-07-09
[Google Blockly] update challenge level test

mikeharv:
This updates an existing UI test which relies on dragging blocks to complete a level in two different ways.

|State|Blocks|Modal|
|-|-|-|
|Initial|image|image|
|Imperfect Solution (extra block)|image|image|
|Perfect solution|image|image|

Like our other block-dragging tests, this one has been relying on `jquery-simulate` to perform a calcuated drag operation. This works fine in either version of Blockly when we are dragging a block from the flyout to the main workspace. However, for a reason I can't determine, the simulate drag is _not_ working on blocks already in the workspace. This could be a bug with Blockly. I used my browser console on the Blockly Playground site to load jQuery and the simulate plugin and also wasn't able to drag a block there. To do this, I asked ChatGPT for help writing a quick script to do this. I was able to get jQuery to return a jQuery object, but the simulate code didn't work. For reference (in case we want to submit a ticket to Blockly), here is the script I used:

```
function loadSimulatePlugin() {
    var script = document.createElement('script');
    script.src = 'https://cdnjs.cloudflare.com/ajax/libs/jquery-simulate/1.0.1/jquery.simulate.min.js';
    script.onload = function() {
        console.log('jquery-simulate plugin loaded');
        // Now you can use the simulate method
        // Example usage:
        $('[data-id="val"]').simulate('click');
    };
    document.head.appendChild(script);
}

// Step 1: Load jQuery
var script = document.createElement('script');
script.src = 'https://code.jquery.com/jquery-3.6.0.min.js';
script.onload = function() {
    console.log('jQuery loaded');
    // Step 2: Load jquery-simulate plugin
    var simulateScript = document.createElement('script');
    simulateScript.src = 'https://cdnjs.cloudflare.com/ajax/libs/jquery-simulate/1.0.1/jquery.simulate.min.js';
    simulateScript.onload = function() {
        console.log('jquery-simulate plugin loaded');
    };
    document.head.appendChild(simulateScript);
};
document.head.appendChild(script);
```

While looking for a workaround, I came across what I believe is actually a simpler and more reliable way manipulating blocks during tests - the actual Blockly APIs. Below is a comparison of the generated code we'd run for a step where one block is moved to another block on the same workspace. (Illustrated in the table above as the change in blocks between rows 1 and 2)

**Old jQuery Code** (comments added to PR for clarity)
```
// Calculate the horizontal drag distance
var drag_dx = $("[data-id='5']").filter(function (index) {
    return $(this).parents(':hidden').length === 0;
}).offset().left - $("[data-id='7']:last").filter(function (index) {
    return $(this).parents(':hidden').length === 0;
}).offset().left;

// Calculate the vertical drag distance
var drag_dy = $("[data-id='5']").filter(function (index) {
    return $(this).parents(':hidden').length === 0;
}).offset().top - $("[data-id='7']:last").filter(function (index) {
    return $(this).parents(':hidden').length === 0;
}).offset().top;

// Simulate a drag using the calculated distances (plus an extra 50px vertically to account for the height of the block)
$("[data-id='7']:last").simulate('drag', {handle: 'corner', dx: drag_dx + 0, dy: drag_dy + 50, moves: 5});
```

**New Blockly API Code**
```
var workspace = Blockly.getMainWorkspace();
var blockToMove = workspace.getBlockById('stoneTurn');
var targetBlock = workspace.getBlockById('stoneMoveTop');
targetBlock.nextConnection.connect(blockToMove.previousConnection);
```

The new code is simpler and easier to read.

## Links

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/dashboard/config/levels/custom/maze/courseC_maze_programming3.level
+++ b/dashboard/config/levels/custom/maze/courseC_maze_programming3.level
@@ -1,7 +1,8 @@
 <Maze>
   <config><![CDATA[{
+  "published": true,
   "game_id": 25,
-  "created_at": "2016-07-14T20:50:32.000Z",
+  "created_at": "2022-02-16T19:37:58.000Z",
   "level_num": "custom",
   "user_id": 19,
   "properties": {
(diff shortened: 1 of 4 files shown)


molly-moen (approved):
Much cleaner!

PR #59647, opened 2024-07-09
Set up relationship between Course Offerings and Workshops for join table

TurnerRiley:
Following [this PR](https://github.com/code-dot-org/code-dot-org/pull/59485) which added the join table between `course_offerings` and `pd_workshops`, this adds the many-to-many relationship definitions for course offerings and workshops. I used [the user.rb](https://github.com/code-dot-org/code-dot-org/blob/staging/dashboard/app/models/user.rb#L230) join table relation line and [the workshop.rb](https://github.com/code-dot-org/code-dot-org/blob/staging/dashboard/app/models/pd/workshop.rb#L42) join table relation line as reference since they represent a many-to-many relation with a join table that includes `pd_workshop`.

## Links
Jira ticket: [here](https://codedotorg.atlassian.net/jira/software/c/projects/ACQ/boards/64?assignee=60d62161f65054006980bd71&selectedIssue=ACQ-2048)

### DIFF
--- a/dashboard/app/models/course_offering.rb
+++ b/dashboard/app/models/course_offering.rb
@@ -34,6 +34,8 @@ class CourseOffering < ApplicationRecord
   has_many :course_versions, -> {where(content_root_type: ['UnitGroup', 'Unit'])}
   belongs_to :self_paced_pl_course_offering, class_name: 'CourseOffering', optional: true
 
+  has_and_belongs_to_many :pd_workshops, class_name: 'Pd::Workshop', join_table: :course_offerings_pd_workshops, association_foreign_key: 'pd_workshop_id'
+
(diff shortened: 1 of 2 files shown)


dashboard/app/models/course_offering.rb
```diff
@@ -34,6 +34,8 @@ class CourseOffering < ApplicationRecord
   has_many :course_versions, -> {where(content_root_type: ['UnitGroup', 'Unit'])}
   belongs_to :self_paced_pl_course_offering, class_name: 'CourseOffering', optional: true
 
+  has_and_belongs_to_many :pd_workshops, class_name: 'Pd::Workshop', join_table: :course_offerings_pd_workshops, association_foreign_key: 'pd_workshop_id'
```
bethanyaconnor, line 37: nit: are all these parameters needed?

bethanyaconnor (approved):
Curious if all the parameters are needed, but not a big deal either way. Looks good to me!

dashboard/app/models/course_offering.rb
TurnerRiley: Offline discussion: We felt some of the parameters had to stay to handle the namespace difference of `workshop` vs. `pd_workshop`, but the `association_foreign_key: 'course_offering_id'` line wasn't necessary in `workshop.rb`

PR #59649, opened 2024-07-09
[AI Chat] Update input/output of new fine-tuned models

fisher-alice:
This PR updates the formatting for the input and output of the 3 new fine-tuned models that were added recently (Pirate, Arithmo, and Karen) by https://github.com/code-dot-org/code-dot-org/pull/59284.

For Arithmo and Karen, the format for the input includes only one user message (user input - assistant response). These model formats do not accommodate conversation history as the base model, Pirate, and Biomistral do.

See:
- [akjindal53244/Arithmo-Mistral-7B · Hugging Face](https://huggingface.co/akjindal53244/Arithmo-Mistral-7B) 
- [FPHam/Karen_TheEditor_V2_CREATIVE_Mistral_7B · Hugging Face](https://huggingface.co/FPHam/Karen_TheEditor_V2_CREATIVE_Mistral_7B)

For Pirate, the input is formatted as the base model, but a stopping character is used to help with output quality.
See [phanerozoic/Mistral-Pirate-7b-v0.3 · Hugging Face](https://huggingface.co/phanerozoic/Mistral-Pirate-7b-v0.3)
Also, the output needs to be formatted/cleaned up as jargon/extraneous text or code is still sometimes generated.

## After update

Screencast video of 3 fine-tuned models:

https://github.com/code-dot-org/code-dot-org/assets/107423305/0a22e2a8-38e6-4f72-9823-42edf6a6cd84

Screenshot of a couple of responses from Karen model:
Screenshot 2024-07-09 at 11 59 10 AM

## Links
[jira - pirate formatting](https://codedotorg.atlassian.net/browse/LABS-903)
[jira - arithmo formatting](https://codedotorg.atlassian.net/browse/LABS-905)
[jira - karen formatting](https://codedotorg.atlassian.net/browse/LABS-904)

## Testing story
I tested locally in aichat levels. I did notice that sometimes responses from the Pirate model could be inappropriate (e.g., referring to women in disrespectful manner). Perhaps more testing can be done to add a base system prompt for the model @samantha-code @tshaffercodeorg.

As noted above the input format for Arithmo and Karen models only include one user message. In this update, the system prompt and retrievals are included along with the user message, but testing can be done to see if the system prompt and retrievals actually have an impact on the model response.

## PR Checklist:

- [ ] Code is well-commented


dashboard/app/controllers/aichat_controller.rb
```diff
@@ -29,7 +29,7 @@ def chat_completion
   private def get_response_body
     # Check for profanity
     locale = params[:locale] || "en"
-    filter_result = ShareFiltering.find_failure(params[:newMessage][:chatMessageText], locale)
+    filter_result = ShareFiltering.find_profanity_failure(params[:newMessage][:chatMessageText], locale)
```
fisher-alice, line 32: This method was added recently by https://github.com/code-dot-org/code-dot-org/pull/59414.

dashboard/app/helpers/aichat_sagemaker_helper.rb
```diff
@@ -1,62 +1,122 @@
...
+        []
+      end
     {
       inputs: inputs,
       parameters: {
         temperature: aichat_params[:temperature].to_f,
         max_new_tokens: MAX_NEW_TOKENS,
-        top_p: TOP_P
+        top_p: TOP_P,
+        stop: stopping_strings,
```
fisher-alice, line 42: Thanks @tshaffercodeorg for pointing me to this param. Seems to help with latency even if extra text still seems to be generated.

dashboard/app/helpers/aichat_sagemaker_helper.rb
```diff
@@ -1,62 +1,126 @@
...
-    inputs += aichat_params[:retrievalContexts].join(" ") if aichat_params[:retrievalContexts]
-    inputs = SENTENCE_BEGIN_TOKEN + wrap_as_instructions(inputs)
-    all_messages.each do |msg|
-      if msg[:role] == USER
-        inputs += wrap_as_instructions(msg[:chatMessageText])
-      elsif msg[:role] == ASSISTANT
-        # Note that each assistant message in the conversation history is followed by
-        # the end-of-sentence token but a begin-of-sentence token is not required.
-        inputs += msg[:chatMessageText] + SENTENCE_END_TOKEN
+    selected_model_id = aichat_params[:selectedModelId]
```
bencodeorg, line 29: Might be time to start writing some tests here, particularly for the base model to ensure we're not regressing anything?

fisher-alice: Good call! Will do.

dashboard/app/helpers/aichat_sagemaker_helper.rb
```diff
@@ -1,62 +1,126 @@
...
+    # Add system prompt and retrieval contexts if available to inputs as part of instructions that will be sent to model.
+    inputs = ""
+    inputs = aichat_params[:systemPrompt] + " " unless aichat_params[:systemPrompt].empty?
+    inputs << aichat_params[:retrievalContexts].join(" ") if aichat_params[:retrievalContexts]
+    if selected_model_id == FINE_TUNED_MODELS[:ARITHMO]
```
sanchitmalhotra126: I'd recommend breaking these out into separate files to handle processing for each specific model. Especially as we add more models, maintaining a long if-else chain here could get unwieldy. For example, maybe something like
```
def get_model_processor(model)
  case model
  when 'gen-ai-arithmo2-mistral-7b'
    Arithmo2Processor
  when 'gen-ai-mistral-7b-inst-v01'
    MistralInstProcessor
  # ... etc
  else
    # maybe a default or throw an error?
  end
end

# Then you can use the given processor to format the input, output, and get custom stop strings

processor = get_model_processor(selectedModelId)

inputs = processor.format_inputs(systemPrompt, retrievalContexts, stored_messages, new_message)

...

parameters: {
  ...,
  stop: processor.get_stop_strings,
}

...

response = processor.format_output(generated_text)
```

where a single processor just needs to implement the specific functions (i.e. `format_inputs`, `format_outputs`, `get_stop_strings` or something)

That way the logic for each model stays contained to its specific file, and this file can stay relatively simple.

Edit: also, if certain input/output processing is common across some models, we could have some hierarchy to the processor classes. Ex - maybe all of these classes extend from some base `MistralProcessor` class so that BioMistral, Mistral, and Pirate Mistral can use the same input formatting function. Then for example later on if we add Llama or Claude or something, we can create a new base processor class.

sanchitmalhotra126 (commented):
Nice work! One main comment on code organization, but the model-specific implementations look good to me. However I do want to call out -

> For Arithmo and Karen, the format for the input includes only one user message (user input - assistant response). These model formats do not accommodate conversation history as the base model, Pirate, and Biomistral do.

This is definitely something we should clear with curriculum and product. I'm not sure what assumptions are baked into the pilot curriculum currently around model conversation history, but this would be a significant difference from how Mistral and BioMistral work currently.

fisher-alice:
> Nice work! One main comment on code organization, but the model-specific implementations look good to me. However I do want to call out -
> 
> > For Arithmo and Karen, the format for the input includes only one user message (user input - assistant response). These model formats do not accommodate conversation history as the base model, Pirate, and Biomistral do.
> 
> This is definitely something we should clear with curriculum and product. I'm not sure what assumptions are baked into the pilot curriculum currently around model conversation history, but this would be a significant difference from how Mistral and BioMistral work currently.

Yes +1 on this - was planning to reach out to curriculum/product team soon so just posted in [Slack](https://codedotorg.slack.com/archives/C06FELVTV0Q/p1720641928634009). Thanks!

dashboard/app/helpers/aichat_sagemaker_helper.rb
fisher-alice: Great idea! Will update.

fisher-alice: @sanchitmalhotra126  - I ended up creating a base class `MistralProcessor`, and extending this class for all the other processors including `ArithmoProcessor` and `KarenProcessor` (for the `get_stop_strings` method).

A couple things I learned:
- I needed to `require_relative` first `mistral_processor`.
- I was unable to use:
```
  Dir[File.expand_path('./ai_model_processors/*.rb', __FILE__)].each {|file| require_relative file}
```
Not sure if the 2nd item is related to 1st but just wanting to note.

Anyways - lmk what you think! Thanks so much - I think it really helps clean up the code.

dashboard/app/helpers/ai_model_processors/karen_processor.rb
```diff
@@ -0,0 +1,25 @@
...
+    inputs = KarenProcessor::CHAT_ML_BEGIN_TOKEN + KarenProcessor::SYSTEM + KarenProcessor::NEWLINE + instructions + KarenProcessor::CHAT_ML_BEGIN_TOKEN + KarenProcessor::NEWLINE
+    inputs << (KarenProcessor::CHAT_ML_BEGIN_TOKEN + KarenProcessor::USER + KarenProcessor::NEWLINE + KarenProcessor::KAREN_PRETEXT + new_message[:chatMessageText] + KarenProcessor::CHAT_ML_END_TOKEN + KarenProcessor::NEWLINE)
+    inputs << (KarenProcessor::CHAT_ML_BEGIN_TOKEN + KarenProcessor::ASSISTANT)
+    inputs
+  end
+  
+  def format_model_output(generated_text)
+    assistant_response = generated_text.split(CHAT_ML_BEGIN_TOKEN + ASSISTANT).last
+    # Remove extraneous characters at end of assistant response from Karen model.
+    assistant_response.split("}").first
```
sanchitmalhotra126: nit: could chain this? `generated_text.split(...).last.split("}").first`

But also should we have any checks if the response doesn't include these characters for some reason?

dashboard/app/helpers/aichat_sagemaker_helper.rb
```diff
@@ -1,62 +1,75 @@
...
     create_sagemaker_client.invoke_endpoint(
-      endpoint_name: endpoint_name, # required
+      endpoint_name: selected_model_id, # required
       body: input.to_json, # required
       content_type: "application/json"
     )
   end
 
-  def self.get_sagemaker_assistant_response(sagemaker_response)
+  def self.get_sagemaker_assistant_response(sagemaker_response, selected_model_id)
```
sanchitmalhotra126, line 68: minor: not your change, so doesn't have to be part of this PR, but any reason `format_inputs_for_sagemaker_request` and `get_sagemaker_assistant_response` have to be two separate functions? could we just have one function that calls the API and formats the response and returns that to the caller? Wondering because that way we also wouldn't need to pass in `selected_model_id` here again.

dashboard/app/helpers/aichat_sagemaker_helper.rb
```diff
@@ -1,62 +1,75 @@
...
         max_new_tokens: MAX_NEW_TOKENS,
-        top_p: TOP_P
+        top_p: TOP_P,
+        stop: stopping_strings,
       }
     }
   end
 
-  def self.request_sagemaker_chat_completion(input, endpoint_name)
+  def self.get_model_processor(selected_model_id)
```
sanchitmalhotra126, line 47: 🎉

sanchitmalhotra126 (approved):
LGTM! Nice work!

dashboard/app/helpers/ai_model_processors/karen_processor.rb
fisher-alice: I separated this so that I could add a comment about removing extraneous characters in contrast to `format_model_output` in `MistralProcessor` [here](https://github.com/code-dot-org/code-dot-org/blob/3615a2171fb6a54fa91dce6c88e7cf1e2da980db/dashboard/app/helpers/mistral_processor.rb#L35).

And I believe if either of these delimiters are not found, this will return the `generated_text` string. So maybe a check isn't needed?

dashboard/app/helpers/aichat_sagemaker_helper.rb
fisher-alice: Yes - that makes sense! I'm going to do a quick follow-up to add unit tests so I will go ahead and combine `request_sagemaker_chat_completion` and `get_sagemaker_assistant_response` which currently both pass `selected_model_id` - thanks!

dashboard/app/helpers/aichat_sagemaker_helper.rb
fisher-alice: Hey @bencodeorg - @samantha-code would like to test these fine-tuned models soon so I'll do a quick follow-up to add unit tests. thanks!

fisher-alice:
A note that I moved the model processors directly into `helpers` directory since having them in a new subdirectory `ai_model_processors` was causing loading errors in the drone build.
Ideally, I'd like to keep these in a subdirectory. Posted [here](https://codedotorg.slack.com/archives/C0T10HG6N/p1720796350657319) in #dashboard-dev and seeing if this can be resolved. If so, will include in the follow-up PR.

dashboard/app/helpers/aichat_sagemaker_helper.rb
fisher-alice: Follow-up with tests: https://github.com/code-dot-org/code-dot-org/pull/59760

dashboard/app/helpers/ai_model_processors/karen_processor.rb
fisher-alice: FYI - in the follow-up PR https://github.com/code-dot-org/code-dot-org/pull/59760, I discovered a couple small bugs in input formatting for this model so I believe the second part of the output formatting is no longer needed. https://github.com/code-dot-org/code-dot-org/blob/15265dda6e2ce70dbb50de1f158f5a2ca3a1629e/dashboard/app/helpers/karen_processor.rb#L21

PR #59653, opened 2024-07-09
Support free response/multiple choice summary page for lab2 levels

molly-moen:
Teacher tools: tagging you on this PR since you own this feature. The PR should not change any existing behavior, just updates for lab2!

We have a per-level summary page for levels that are either free response/multiple choice or have a contained free response/multiple choice level. This shows a per-section summary of student responses, either as a graph for multiple choice or a list of responses. There is a button for teachers in the instructions panel that takes them to that page, along with a summary of how many students have responded.

We want to support this flow for lab2 as well, which uses new predict level configuration rather than a contained level (see linked spec for details). To do this, I updated the summary page to either use the old method of loading the settings, or to get the predict settings from the `predict_settings` object if it exists. The responses are stored in the same way for both level types, so no changes were needed there. 

I also updated how we store multiple choice responses for predict levels to match the old way, which is to store a list of the indices of the response(s) rather than the response strings.

The summary page displays the question and correct answer all via haml. For predict levels I added two divs that we inject the question and correct answer (defined as react components) into for predict levels. I was able to reuse the same component that we use to display the solution to teachers (I updated that component to have a non-redux variant), but made a new component to show the question as we don't want or need the question to have editable inputs.

Finally, I made a new button that links to the summary page from the lab2 instructions panel. Only teachers see this on predict levels. I haven't added the summary of student responses yet, I will cover that as a follow up.

### Old levels
Button in instructions
Screenshot 2024-07-09 at 1 22 00 PM

Free Response answers
Screenshot 2024-07-09 at 1 23 25 PM

Multiple Choice answers
Screenshot 2024-07-09 at 1 23 32 PM

### Lab2 version
Button in instructions
Screenshot 2024-07-09 at 1 30 47 PM

Free response
Screenshot 2024-07-09 at 1 24 17 PM

Multiple choice
Screenshot 2024-07-09 at 1 24 24 PM

## Links

- [spec](https://docs.google.com/document/d/1lrpc06mMV8CiYTFfx5ahKZ887qBEJ5AybHPn1Pd8JM0/edit#heading=h.n24aegko5o63)
- jira ticket: [CT-667](https://codedotorg.atlassian.net/browse/CT-667)

## Testing story
Tested locally that old and new predict/contained levels show the summary page correctly, and other lab2 non-predict levels are unaffected. I also updated the unit test for the Multi Response summary component.

## Follow-up work
Two follow up items:
- Show "x/y students answered" in the instructions panel. Adding this probably requires adding a new api since this information was previously added in haml. [Task](https://codedotorg.atlassian.net/browse/CT-678) to add this.
- You can navigate to and from the summary page in lab2 by clicking the button on the header, then clicking "back to level" on the summary page. However, if you click a bubble in the header on the summary page it doesn't seem like anything happens, because we are trying to change levels without a page reload but aren't actually in a lab2 level. [Task](https://codedotorg.atlassian.net/browse/CT-679) to fix this.

## PR Checklist:

- [ ] Code is well-commented


apps/src/templates/levelSummary/SummaryPredictQuestion.tsx
```diff
@@ -0,0 +1,48 @@
...
+    <div>
+      <EnhancedSafeMarkdown markdown={question} />
+      {predictSettings.questionType === PredictQuestionType.FreeResponse ? (
+        <textarea
+          placeholder={predictSettings.placeholderText}
+          style={{height: predictSettings.freeResponseHeight || 20}}
+          readOnly={true}
+        />
+      ) : (
+        predictSettings.multipleChoiceOptions?.map((option, index) => {
```
sanchitmalhotra126, line 30: Could this be shared with `PredictQuestion` at all? I noticed there's some similar stuff with how the multiple choice answers get displayed, but also not sure if there's much beyond that.

apps/src/templates/levelSummary/SummaryPredictQuestion.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+interface SummaryPredictQuestionProps {
+  question: string;
+  predictSettings: LevelPredictSettings;
+}
+
+const SummaryPredictQuestion: React.FunctionComponent<
+  SummaryPredictQuestionProps
+> = ({question, predictSettings}) => {
+  console.log(predictSettings);
```
sanchitmalhotra126: nit: remove?

apps/src/templates/levelSummary/SummaryPredictQuestion.tsx
```diff
@@ -0,0 +1,48 @@
...
+  return (
+    <div>
+      <EnhancedSafeMarkdown markdown={question} />
+      {predictSettings.questionType === PredictQuestionType.FreeResponse ? (
+        <textarea
+          placeholder={predictSettings.placeholderText}
+          style={{height: predictSettings.freeResponseHeight || 20}}
+          readOnly={true}
```
sanchitmalhotra126, line 27: For my understanding - why do we need this readonly textarea at all? I guess just to show the placeholder text?

apps/src/templates/levelSummary/MultiResponses.jsx
```diff
@@ -48,25 +48,53 @@ const multiChartData = (data, highlightCorrect = false) => [
 ];
 
 const MultiResponses = ({scriptData, showCorrectAnswer = false}) => {
-  const answers = scriptData.level.properties.answers;
+  // Get the number of answers and a list of correct answers.
+  // The correct answers will be the letters for the correct answers, ex. ['A', 'C']
+  const [numAnswers, correctAnswers] = useMemo(() => {
```
sanchitmalhotra126, line 53: Minor: could this be consolidated to something like
```
  let answers = [];
  let correctIndices = [];

  if (scriptData.level.properties.answers) {
    answers = scriptData.level.properties.answers;
    correctIndices = answers
      .filter(answer => answer.correct)
      .map(answer => answers.indexOf(answer));

  } else if (scriptData.level_properties.predict_settings) {
    const predictSettings = scriptData.level.properties.predict_settings;
    answers = predictSettings.multipelChoiceOptions;
    correctIndices = predictSettings.solution
      .split(',')
      .map(answer => answers.indexOf(answer));
  }

  const correctAnswers = correctIndices.map(index =>
    String.fromCharCode(65 + index)
  );

  return [answers.length, correctAnswers];
```

sanchitmalhotra126 (approved):
LGTM overall!

Out of scope for this PR but asking the perennial Lab2 question 😄 would it be beneficial for us to have this in a new component that isn't routed through the existing `summary` HAML/JS entrypoint? I could imagine Lab2 displaying a summary view (via the LabViewsRenderer?) that displays the predict question/solution for the current level, and makes an AJAX call to fetch the summary responses, without a full page reload. We could still use the existing `SummaryResponses` component even (maybe with a bit of refactoring so we don't need all the "is predict level" checks inside the component)

molly-moen:
> LGTM overall!
> 
> Out of scope for this PR but asking the perennial Lab2 question 😄 would it be beneficial for us to have this in a new component that isn't routed through the existing `summary` HAML/JS entrypoint? I could imagine Lab2 displaying a summary view (via the LabViewsRenderer?) that displays the predict question/solution for the current level, and makes an AJAX call to fetch the summary responses, without a full page reload. We could still use the existing `SummaryResponses` component even (maybe with a bit of refactoring so we don't need all the "is predict level" checks inside the component)

Yeah it would be better to have this all in a component rather than going back to the server, but it was just too much work to get that working right now. I filed a task to investigate this! [CT-680](https://codedotorg.atlassian.net/browse/CT-680)

apps/src/templates/levelSummary/SummaryPredictQuestion.tsx
molly-moen: good catch 🤦‍♀️

apps/src/templates/levelSummary/SummaryPredictQuestion.tsx
molly-moen: Yeah I was trying to match the existing behavior, which shows the readonly textarea. I don't have a strong opinion on whether we should show it or not.

apps/src/templates/levelSummary/SummaryPredictQuestion.tsx
molly-moen: Yeah it was different enough that I didn't share code (`PredictQuestion` displays radio buttons/checkboxes for example). It didn't seem worth extracting out a helper for the one line conversion from index to letter, but I can do that if you think it's valuable!

apps/src/templates/levelSummary/MultiResponses.jsx
molly-moen: good idea!

apps/src/templates/levelSummary/SummaryPredictQuestion.tsx
sanchitmalhotra126: Not necessary if there's not much to be shared!

PR #59655, opened 2024-07-09
One off script to update name_suffix for all CSD microbit levels to nil 

vijayamanohararaj:
Context: When cloning CSD for SY 24-25, we hit an issue were all microbit levels were overwritten by circuit playground ones. The underlying root cause was that both these levels in 2023 version of the curriculum had the same prefix, and only differed by the suffix., ie., circuit playground levels had _2023 suffix, while microbit had _mb2022 suffix.

During cloning, both these suffixes were replaced with _2024 initially, leading to the circuit playground levels getting overwritten. To fix the issues, the microbit units were deleted and recloned with _mb_2024 suffix. 

Issue: The level conflict issue could potentially happen again when cloning for SY 25-26 if both units are cloned with 2025 as suffix (which is very likely). 

Fix: The oneoff script in this PR is to avoid the issue when cloning for SY25-26 by setting name_suffix property of the micro bit levels to nil. [If the name_suffix is nil, during cloning, only the _2024 part of the microbit suffix would get replaced](https://github.com/code-dot-org/code-dot-org/blob/9b807edfc2d632baf01bcf4f25b473a5150c94b0/dashboard/app/models/levels/level.rb#L885C15-L885C24), hence preserving the unique name between circuit playground and microbit for all future clones.

## Links

JIRA: https://codedotorg.atlassian.net/browse/TEACH-1161

## Testing story

Local environment validation

## Deployment strategy

Manual execution only on levelbuilder environment, this would update level files which would seed in other environments

## Follow-up work

None

## Privacy

N/A

## Security

N/A

## Caching

N/A

## PR Checklist:

- [x] Code is well-commented

### DIFF
--- a/bin/oneoff/data_fix/remove_level_name_suffix_CSD_microbit_2024.rb
+++ b/bin/oneoff/data_fix/remove_level_name_suffix_CSD_microbit_2024.rb
@@ -24,8 +24,7 @@ def remove_level_name_suffix
     if level.name_suffix == "_mb_2024"
       puts "Processing level [#{level.name}] with name_suffix [#{level.name_suffix}]"
 
-      level.update(name_suffix: nil)
-      level.save!
+      level.update!(name_suffix: nil)
     else
       puts "Ignoring level [#{level.name}] with name_suffix [#{level.name_suffix}] not matching _mb_2024"
     end
--- a/bin/oneoff/data_fix/remove_level_name_suffix_CSD_microbit_2024.rb
+++ b/bin/oneoff/data_fix/remove_level_name_suffix_CSD_microbit_2024.rb
@@ -1,0 +1,32 @@
+#!/usr/bin/env ruby
+
+# https://codedotorg.atlassian.net/browse/TEACH-1161
+# When cloning CSD for SY 24-25, we hit an issue were all microbit levels were overwritten
+# by circuit playground ones. The underlying root cause was that both these levels in 2023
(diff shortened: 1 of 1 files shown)


bin/oneoff/data_fix/remove_level_name_suffix_CSD_microbit_2024.rb
```diff
@@ -0,0 +1,35 @@
...
+def remove_level_name_suffix
+  raise unless Rails.application.config.levelbuilder_mode
+
+  levels_to_update = Level.where("name LIKE ?", "%\\_mb\\_2024")
+  levels_to_update.each do |level|
+    if level.name_suffix == "_mb_2024"
+      puts "Processing level [#{level.name}] with name_suffix [#{level.name_suffix}]"
+
+      level.update(name_suffix: nil)
+      level.save!
```
davidsbailey: I'd like to make sure we understand exactly what will happen here if the save fails. as it is written, `update` would silently return something falsy, and then I'm not 100% sure what would happen on `save!`.

my recommendation would be to change this to:
```suggestion
      level.update!(name_suffix: nil)
```

I checked that `update!` should still trigger the `after_save` callback which writes the file: https://guides.rubyonrails.org/active_record_callbacks.html#available-callbacks 

alternatively, you could find a way to manually verify that the code you have here does the right thing when saving fails.

davidsbailey (approved):
Great explanation in comments! LGTM after addressing one issue below.

PR #59661, opened 2024-07-09
chore: remove `istanbul-traverse-loader` dep

stephenliang:
Addresses [dependabot 760](https://github.com/code-dot-org/code-dot-org/security/dependabot/760)

The `istanbul-traverse-loader` is used to generate coverage information when `yarn test:unit` is supplemented with `COVERAGE=1`. This was used for the karma based tests, however jest has its own coverage system which marks the need for the webpack based coverage collector to be obsoloete for a majority of unit tests. (There are 16 karma unit tests and 90+ karma integration tests).

Since karma is also on a path to deprecation with an upcoming browser based test solution, remove the istanbul loader as its value has diminished (versus the cost of maintaining it)

## PR Checklist:

- [ ] Code is well-commented


wilkie (approved):
LGTM 👍

PR #59663, opened 2024-07-09
P20-987: Use `Rack::CookieDCDO` in Pegasus

artem-vavilov:
We need to include the `Rack::CookieDCDO` middleware in Pegasus, as it runs on a different process than the dashboard
## Links
- Related task: https://github.com/code-dot-org/code-dot-org/pull/59442

### DIFF
--- a/pegasus/config.ru
+++ b/pegasus/config.ru
@@ -50,6 +50,12 @@ if CDO.image_optim
   use Rack::Optimize
 end
 
+if CDO.use_cookie_dcdo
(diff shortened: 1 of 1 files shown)


Hamms (approved):
LGTM, thank you!

A bit more context, for anyone watching: I tested this change manually on the test server between builds by updating the code locally and restarting the pegasus service.

Also, minor point of clarification, since my earlier explanation was ambiguous: Dashboard and Pegasus for our managed environments run on different *processes*, but on the same server

PR #59664, opened 2024-07-09
chore(deps): upgrade webpack 4 style loaders to webpack 5

stephenliang:
The Code.org code base uses some webpack 4 style loaders which were deprecated and merged into the [webpack 5 asset modules feature](https://webpack.js.org/guides/asset-modules/). These loaders are no longer maintained and have some critical dependency issues. To mitigate these issues, migrate the loaders to the webpack 5 feature.

**What was removed?**

* url-loader: Used to base64 encode an image. This is now automatically done via `asset/inline` and therefore the dependency can be removed after swapping to `asset/inline`.
* file-loader: This is not used in the code base except in unit tests. Jest handles this automatically, and therefore the dependency can be removed.
* raw-loader: This is not used in the code base, and therefore the dependency can be removed.

This addresses a majority of the issues in [dependabot 325](https://github.com/code-dot-org/code-dot-org/security/dependabot/325) but doesn't resolve it completely.

## Testing story

There is only one usage of `url-loader` in the code base, the app lab turtle. To test this, I inspected the production base64 encoding of the turtle and compared it with localhost. Both turtles produced the same output.

## Deployment strategy

Merge this at night on a standalone DTT to ensure no eyes mismatches as this updates image generation in bulk

## PR Checklist:

- [ ] Code is well-commented


apps/src/applab/applabTurtle.js
```diff
@@ -1,6 +1,6 @@
-import applabCommands from './commands';
+import turtleImageSrc from '../../static/applab/723-location-arrow-toolbar-48px-centered.png';
 
-import turtleImageSrc from '!!url-loader!../../static/applab/723-location-arrow-toolbar-48px-centered.png';
```
wilkie, line 3: Wow... it is only used in one place? Good find.

apps/webpack.config.js
```diff
@@ -293,17 +293,7 @@ const WEBPACK_BASE_CONFIG = {
         // this file when asset digests are turned off, it will return a
         // 404 because it thinks the hash is a digest and it won't
         // be able to find the file without the hash. :( :(
```
wilkie: This comment is unnecessary now I believe

wilkie (commented):
I'm a little intrigued and probably a little ignorant. You only found one use of `url-loader` yet this rule seems to apply to a large number of images. What will be the broad difference in behavior for those images? Are we concerned about, as this section was concerned about, the behavior for large files?

stephenliang:
> I'm a little intrigued and probably a little ignorant. You only found one use of `url-loader` yet this rule seems to apply to a large number of images. What will be the broad difference in behavior for those images? Are we concerned about, as this section was concerned about, the behavior for large files?

There is no fundamental difference, in webpack v4, `url-loader` was applied to the images which converted the following regex to base64 encoded inline images:

```
        test: /\.(png|jpg|jpeg|gif|svg)$/,
        include: [
          p('static'),
          p('src'),
          p('test'),
          p('../dashboard/app/assets/images'),
        ],
```

In webpack v5, the equivalent is provided in [webpack asset modules](https://webpack.js.org/guides/asset-modules/) and is documented:


> asset/inline exports a data URI of the asset. Previously achievable by using url-loader.

The images should have the same base64 encoding prior to this change and after this change, and I'll probably merge this at night on a standalone DTT to ensure no eyes mismatches.

wilkie (approved):
LGTM! 👍

stephenliang:
I'm targeting to merge this July 17th after the DTP and will monitor the DTT for baseline changes.

stephenliang:
[No baseline changes were detected when doing the eyes test](https://codedotorg.slack.com/archives/C03CM903Y/p1721257205606979) - going to merge this in now!

PR #59666, opened 2024-07-09
Fix Violations of and Reenable `Style/InverseMethods`

Hamms:
This one is interesting, but I think overall an improvement: 

> Check for usages of not (`not` or `!`) called on a method when an inverse of that method can be used instead.

Fix applied automatically with `bundle exec rubocop --autocorrect-all --only Style/InverseMethods`. I then also ran `bundle exec rubocop --autocorrect-all --only Style/SymbolProc` to apply some additional cleanup enabled by the first pass.

The change seems to be mostly things like updating `foo.select({|x| !x.bar?})` to `foo.reject({|x| x.bar?})` (and then further updating to just `foo.reject(&:bar?)`), or `!foo.any?(&:bar?)` to `foo.none?(&:bar?)`. I think it's consistently made the resulting code more readable, but as with all `Style/` cops it does come down to personal preference and I'm open to pushback if anyone has other preferences.

What do y'all think?

## Links

- https://docs.rubocop.org/rubocop/cops_style.html#styleinversemethods
- https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/InverseMethods

## Testing story

Note that this change is not completely guaranteed to be safe:

> This cop is unsafe because it cannot be guaranteed that the method and its inverse method are both defined on receiver, and also are actually inverse of each other.

I visually inspected each change and they all seem valid to me, but I'm relying on existing tests to catch any exceptions.
## PR Checklist:

- [ ] Code is well-commented


bencodeorg:
hmm this feels a little more controversial to me...it reminds me of how rubocop insists on the use of `unless` instead of `if !`, which I find turns my brain in knots sometimes (@cat5inthecradle and I were chatting about this recently, maybe he can offer an opinion as well here)?

The !any vs. none case makes more sense to me, trying to wrap my head around why...maybe because with the `select` situation you're trying to keep in your brain "what does the set of elements I've selected represent, as I might do something with them in the future" vs. the !any vs. none case is just returning a boolean, which is easier to wrap your head around?

Hamms:
Yeah, this one's definitely pretty subjective; `select! vs reject` for list manipulation clicks for me in the same way that `assert! vs reject` does for unit tests, but I still occasionally struggle with both `if! vs unless` and `!any vs none` :shrug:

For comparison, I'll put together a version of this PR that uses the `InverseBlocks` configuration option to ignore `select! vs reject`, so it only enforces use of the simpler inverse methods; that should make for a much smaller diff, too. Might be a sweet spot between following Ruby standards and accommodating the preferences of our team as demonstrated through the code we've written.

Hamms:
closing in favor of https://github.com/code-dot-org/code-dot-org/pull/61199

PR #59668, opened 2024-07-10
Add `unlock_token` to Users

Hamms:
In preparation for eventually adding the ability to lock out user accounts on repeated failed authentication attempts, we first add a column with an index on the Users table which will allow users to unlock their accounts by clicking a link in an email we'll send them.

Because we are on MySQL 8, [adding a column is now an instant operation](https://dev.mysql.com/doc/refman/8.0/en/innodb-online-ddl-operations.html#online-ddl-column-operations-table), even on a table as large as our Users table. Unfortunately, [adding an index is not](https://dev.mysql.com/doc/refman/8.0/en/innodb-online-ddl-operations.html#online-ddl-index-operations-table).

## Testing story

 I did some manual testing on a clone of the production database using an adhoc, and was able to confirm that adding the new column takes less than a second and adding the index for that column takes about 35 minutes. During that time, the table is still available for reads and writes; only schema changes are locked.

This means we can safely apply the change in production without needing to use `gh-ost`, but we will need to execute the query to add the index outside of the context of our Rails migrations, which have a 30-second limit.

## Follow-up work

Once the change has been fully deployed, all servers should have a new column and all servers except production should have an index on that column. To bring production in line with the rest of our pipeline, we will after deployment kick off a `CREATE UNIQUE INDEX` query from `production-console`:

```bash
nohup mysql-client-dashboard-writer "CREATE UNIQUE INDEX index_users_on_unlock_token ON users (unlock_token)" &
```

## Links

Work-in-progress PR to fully enable the feature:
- https://github.com/code-dot-org/code-dot-org/pull/57515

Devise documentation:
- https://github.com/heartcombo/devise/wiki/How-To:-Add-:lockable-to-Users

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/dashboard/db/migrate/20240321204728_add_devise_lockable_to_users.rb
+++ b/dashboard/db/migrate/20240321204728_add_devise_lockable_to_users.rb
@@ -29,7 +29,7 @@ class AddDeviseLockableToUsers < ActiveRecord::Migration[6.1]
   end
 
   def self.down
-    remove_column :users, :unlock_token
     remove_index :users, :unlock_token
+    remove_column :users, :unlock_token
   end
 end
--- a/dashboard/db/migrate/20240321204728_add_devise_lockable_to_users.rb
+++ b/dashboard/db/migrate/20240321204728_add_devise_lockable_to_users.rb
@@ -11,17 +11,20 @@ class AddDeviseLockableToUsers < ActiveRecord::Migration[6.1]
     # column; we therefore implement both in the properties blob.
     add_column :users, :unlock_token, :string
 
-    # On MySQL 8.0, adding an index does not lock the table but it does still
-    # take quite a while to apply successfully; see
-    # https://dev.mysql.com/doc/refman/8.0/en/innodb-online-ddl-operations.html#online-ddl-column-syntax-notes
+    # Attempting to create this index via a migration will fail on a table as
+    # large as the one we have in production. To avoid that, we plan to execute
+    # the index creation query manually after the rest of this migration has
+    # completed.
     #
+    # Specifically: on MySQL 8.0, adding an index does not lock the table but
+    # it does still take quite a while to apply successfully (see
+    # https://dev.mysql.com/doc/refman/8.0/en/innodb-online-ddl-operations.html#online-ddl-column-syntax-notes).
     # When tested against a clone of our production database, this operation
     # took approximately 35 minutes, much longer than the 30 seconds migrations
-    # will wait before erroring out. To avoid this migration failing in
-    # production, we plan to execute the index creation query manually after
-    # the rest of this migration has completed.
+    # will wait before erroring out. To create the index without the time limit
+    # enforced by migrations, we should run this command from a bash shell on
+    # production-console:
     #
-    # Specifically, we should from a bash shell on production-console run:
     #     nohup mysql-client-dashboard-writer "CREATE UNIQUE INDEX index_users_on_unlock_token ON users (unlock_token)" &
     unless Rails.env.production?
       add_index :users, :unlock_token, unique: true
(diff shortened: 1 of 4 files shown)


jordan-springer (approved):
This is cool. Will this field be surfaced in redshift as well? I'm sure we would find it useful

Hamms:
I don't think new fields get surfaced in redshift unless we specifically enable them. It's not clear to me what kind of value you'd get out of this field, though; can you expand a bit on the expected usefulness?

cat5inthecradle:
Our DMS migration config, "tasks.yml", only operates with an exclusion model. All columns in a table are included in the export unless explicitly excluded.

I would suggest excluding it, as it is either an ephemeral value and useless, or it's an ephemeral value that can only be used to analyze whether a particular user has triggered a lockout, or to count how many users have ever triggered a lockout.

We currently exclude a lot of similar columns when migrating to the non-pii schema: https://github.com/code-dot-org/code-dot-org/blob/2216f618b5cf54959d51c97eeedd6c08d6ea5d3b/aws/dms/tasks.yml#L120

But we do not exclude those columns for the pii schema: https://github.com/code-dot-org/code-dot-org/blob/2216f618b5cf54959d51c97eeedd6c08d6ea5d3b/aws/dms/tasks.yml#L162

cat5inthecradle (approved):
LGTM with one nit.

We should aim to deploy this on a less busy day (later next week?). Does this need to be in a standalone DTP, or are we good to ship it alongside regular daily changes?

dashboard/db/migrate/20240321204728_add_devise_lockable_to_users.rb
```diff
@@ -0,0 +1,35 @@
...
+
+    # On MySQL 8.0, adding an index does not lock the table but it does still
+    # take quite a while to apply successfully; see
+    # https://dev.mysql.com/doc/refman/8.0/en/innodb-online-ddl-operations.html#online-ddl-column-syntax-notes
+    #
+    # When tested against a clone of our production database, this operation
+    # took approximately 35 minutes, much longer than the 30 seconds migrations
+    # will wait before erroring out. To avoid this migration failing in
+    # production, we plan to execute the index creation query manually after
+    # the rest of this migration has completed.
```
cat5inthecradle: My more well written comment got deleted. I was going to suggest hoisting this bit of information up to the top of the comment. That way our future selves get to read the critical 'what' before digging into the 'why'.

dashboard/db/migrate/20240321204728_add_devise_lockable_to_users.rb
```diff
@@ -0,0 +1,35 @@
...
+    # Specifically, we should from a bash shell on production-console run:
+    #     nohup mysql-client-dashboard-writer "CREATE UNIQUE INDEX index_users_on_unlock_token ON users (unlock_token)" &
+    unless Rails.env.production?
+      add_index :users, :unlock_token, unique: true
+    end
+  end
+
+  def self.down
+    remove_column :users, :unlock_token
+    remove_index :users, :unlock_token
```
sureshc, line 35: Should we be removing the index first before removing the column?

Hamms: Oh, good catch!

Hamms:
> Our DMS migration config, "tasks.yml", only operates with an exclusion model. All columns in a table are included in the export unless explicitly excluded.

Oh dang! I didn't realize that; thank you for the clarification! It does seem like a potentially dangerous design; is there anything we could do to better highlight this to developers who are adding new fields?

> Does this need to be in a standalone DTP, or are we good to ship it alongside regular daily changes?

I think it's fine to deploy this alongside regular changes; but yes, we'll want to do it during a day that is otherwise low on noise so we can easily fit in the async query we need to run.

Hamms:
Comments addressed; please take another look!

And yes, for clarity: I plan to deploy this sometime next week, on a day during which things are otherwise relatively quiet and I will be available post-DTP to run the command to add the index manually.

PR #59672, opened 2024-07-10
update step mode and simpledrag tests

mikeharv:
This adds hard-coded block ids to a few level blocks and updates associated UI tests to use those ids explicitly so that the test can run on Google Blockly.

The Step Mode test works exactly as it used to. The Simple Drag test works the same way but now connects to different blocks rather than two copies of the same block. This is because a second block of the same type would have a randomly generated ID which would make it impossible to track within the workspace.

## Links

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/dashboard/test/ui/features/star_labs/step_mode.feature
+++ b/dashboard/test/ui/features/star_labs/step_mode.feature
@@ -2,31 +2,31 @@
 Feature: Step Mode
 
 Scenario: Step Only - Failure
-  Given I am on "http://studio.code.org/s/step/lessons/1/levels/1"
+  Given I am on "http://studio.code.org/s/step/lessons/1/levels/1?blocklyVersion=google"
     And I wait for the page to fully load
   Then element "#runButton" is hidden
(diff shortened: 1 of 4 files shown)


dashboard/test/ui/features/star_labs/step_mode.feature
```diff
@@ -2,31 +2,31 @@
...
-  Given I am on "http://studio.code.org/s/step/lessons/1/levels/1"
+  Given I am on "http://studio.code.org/s/step/lessons/1/levels/1?blocklyVersion=google"
     And I wait for the page to fully load
   Then element "#runButton" is hidden
     And element "#resetButton" is hidden
     And element "#stepButton" is visible
     And element "#stepButton" is not disabled
 
-  When I drag block "1" to block "5"
+  When I drag block "moveForward" to block "startBlock"
```
mikeharv, line 12: The `"moveForward" id was already added to this block in a previous PR as the block definition is shared across levels.

molly-moen (approved):

PR #59673, opened 2024-07-10
Use translated welcome emails

bethanyaconnor:
Uses the translated welcome emails! 🥳 

Screenshot 2024-07-16 at 12 02 49 PM

I've created [a task](https://codedotorg.atlassian.net/browse/ACQ-2071) to clean up more of the code in a follow-up.

### DIFF
--- a/lib/test/cdo/test_mailjet.rb
+++ b/lib/test/cdo/test_mailjet.rb
@@ -57,7 +57,7 @@ class MailJetTest < Minitest::Test
     MailJet.update_contact_field(mock_contact, 'field_name', 'field_value')
   end
 
-  def test_send_template_email
+  def test_send_template_email_default_locale
     to_email = 'fake.email@test.xx'
     to_name = 'Fake Name'
     from_address = 'test@code.org'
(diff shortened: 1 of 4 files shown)


lib/test/cdo/test_mailjet.rb
```diff
@@ -89,7 +91,46 @@ def test_send_template_email
         messages[0][:TemplateID] == template_id
     end
 
-    MailJet.send_template_email(mock_contact, email_config)
+    MailJet.send_template_email(mock_contact, email_config, 'en-US')
+  end
+
+  def test_send_template_email_locoalized
```
TurnerRiley: Nit: spelling of localized

TurnerRiley (approved):
LGTM! Out of curiosity, will Mailjet handle any translations? And if so, will we then have to retroactively add in the template ids down below for different locales?

bethanyaconnor:
> Out of curiosity, will Mailjet handle any translations? And if so, will we then have to retroactively add in the template ids down below for different locales?

There's a MailJet <-> Crowdin integration that we're utilizing. We would need to add template IDs for different locales if we decided to use the translations for those locales.

lib/test/cdo/test_mailjet.rb
bethanyaconnor: woops thank you!

PR #59683, opened 2024-07-10
Submittable python lab levels: part 1

molly-moen:
Part 1 of making submittable python levels. Submittable levels work as follows:
 1. The level is designated as submittable in the level configuration
 2. The continue/finish button says "submit" instead of continue/finish.
 3. The user must run their code at least once for the submit button to be enabled.
 4. When the user clicks submit, their code is marked as submitted and is locked for edits. The user can "unsubmit" their code (the submit button becomes an unsubmit button in this case) in order to edit again.
 5. The progress bubble turns purple when their code is submitted.

This PR covers most of the work for this. The work not included is the logic to ensure the user has run their code before submitting, and adding a confirmation modal on submit/unsubmit. These will be covered in a follow-up PR.

## Links

- jira ticket: [CT-621](https://codedotorg.atlassian.net/browse/CT-621)

## Testing story
Tested that you can submit and unsubmit a submittable python lab level, and other levels are unchanged.

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/code-studio/progressRedux.ts
+++ b/apps/src/code-studio/progressRedux.ts
@@ -411,7 +411,7 @@ export const sendSubmitReport = createAsyncThunk<
   const extraPayload = {
     submitted: payload.submitted,
   };
-  sendReportHelper(
+  await sendReportHelper(
     payload.appType,
     TestResults.CONTAINED_LEVEL_RESULT,
     thunkAPI.dispatch,
--- a/apps/src/code-studio/progressRedux.ts
+++ b/apps/src/code-studio/progressRedux.ts
@@ -87,11 +87,15 @@ export interface ProgressState {
   unitCompleted: boolean | undefined;
 }
 
-export interface MilestoneReport {
+export interface MilestoneReport extends OptionalMilestoneData {
   app: string;
   result: boolean;
   testResult: number;
+}
+
+interface OptionalMilestoneData {
   program?: string;
+  submitted?: boolean;
 }
 
 const initialState: ProgressState = {
@@ -384,12 +388,35 @@ export const sendPredictLevelReport = createAsyncThunk<
     state: RootState;
   }
 >('progress/sendPredictLevelReport', async (payload, thunkAPI) => {
+  const extraPayload = {
+    program: payload.predictResponse,
+  };
   sendReportHelper(
     payload.appType,
     TestResults.CONTAINED_LEVEL_RESULT,
     thunkAPI.dispatch,
     thunkAPI.getState,
-    payload.predictResponse
+    extraPayload
+  );
+});
+
+export const sendSubmitReport = createAsyncThunk<
+  void,
+  {appType: string; submitted: boolean},
+  {
+    dispatch: AppDispatch;
+    state: RootState;
+  }
+>('progress/sendSubmitReport', async (payload, thunkAPI) => {
+  const extraPayload = {
+    submitted: payload.submitted,
+  };
+  sendReportHelper(
+    payload.appType,
+    TestResults.CONTAINED_LEVEL_RESULT,
+    thunkAPI.dispatch,
+    thunkAPI.getState,
+    extraPayload
   );
 });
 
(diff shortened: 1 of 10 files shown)


apps/src/code-studio/progressRedux.ts
```diff
@@ -415,17 +450,15 @@ function sendReportHelper(
   // The server does not appear to use the user ID parameter,
```
ebeastlake, line 451: Unrelated to this PR but 😱 😅

apps/src/code-studio/progressRedux.ts
```diff
@@ -415,17 +450,15 @@ function sendReportHelper(
...
   // so just pass 0, like some other milestone posts do.
   const userId = 0;
+  extraData = extraData || {};
```
ebeastlake, line 454: Is it possible for a program property with an empty or falsey value in the data object sent to the server when it's not desired? Given the way `sendReportHelper` is used, it looks like not?

apps/src/code-studio/progressRedux.ts
```diff
@@ -384,12 +388,43 @@ export const sendPredictLevelReport = createAsyncThunk<
...
+    payload.appType,
+    result,
+    thunkAPI.dispatch,
+    thunkAPI.getState,
+    extraPayload
+  );
+  // Submit status isn't properly updated by just saving the status code, so re-query
+  // user progress to force the bubble to update.
+  thunkAPI.dispatch(
+    queryUserProgress(thunkAPI.getState().currentUser.userId.toString())
```
ebeastlake, line 428: It's a bummer we have to requery this (vs. optimistic updates or something), but it looks like it's a common pattern.

apps/src/codebridge/ControlButtons/index.tsx
```diff
@@ -55,6 +69,26 @@ const ControlButtons: React.FunctionComponent = () => {
...
+    if (isSubmittable) {
+      onSubmit();
+    } else if (hasNextLevel) {
+      onContinue();
+    } else {
+      onFinish();
+    }
+  };
+
+  const getNavigationButtonText = () => {
```
ebeastlake: You could combine these if you wanted:
```
const getNavigationButtonProps = () => {
  if (isSubmittable) {
    return {
      text: hasSubmitted ? commonI18n.unsubmit() : commonI18n.submit(),
      onClick: onSubmit,
    };
  } else if (hasNextLevel) {
    return {
      text: commonI18n.continue(),
      onClick: onContinue,
    };
  } else {
    return {
      text: commonI18n.finish(),
      onClick: onFinish,
    };
  }
};

const {text, onClick} = getNavigationButtonProps();

<Button
  text={text}
  onClick={onClick}
/>
```

apps/src/code-studio/progressRedux.ts
```diff
@@ -87,11 +87,15 @@ export interface ProgressState {
...
-export interface MilestoneReport {
+export interface MilestoneReport extends OptionalMilestoneData {
   app: string;
   result: boolean;
   testResult: number;
+}
+
+interface OptionalMilestoneData {
   program?: string;
+  submitted?: string;
```
ebeastlake, line 99: Why a string and not a boolean?

ebeastlake (commented):
A couple minor questions, but looks good to me!

apps/src/code-studio/progressRedux.ts
molly-moen: It shouldn't be possible (if we use typescript everywhere we call this function, which we currently do 😅). We also only send a report right now with a program for predict levels, and we only allow submitting a response when you have some response filled in.

apps/src/code-studio/progressRedux.ts
molly-moen: yeah...not sure what's going on there 😬

apps/src/code-studio/progressRedux.ts
molly-moen: Yeah in theory it should be possible, but the progress store is so complicated I couldn't figure out how to get the data plumbed through correctly. I justified this because a user should press submit rarely, probably only one time per submit level.

apps/src/code-studio/progressRedux.ts
molly-moen: That's what the server expects 🤷‍♀️ it didn't work when I sent a boolean. I can add a comment explaining this.

apps/src/codebridge/ControlButtons/index.tsx
molly-moen: good idea!

ebeastlake (approved):

PR #59686, opened 2024-07-10
chore(deps): remove dangling uglifyjs plugin

stephenliang:
The webpack config uses terser to minify and the uglify plugin appears to be dangling, therefore remove the unused dependencies.

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/package.json
+++ b/apps/package.json
@@ -31,7 +31,6 @@
     "storybook:deploy": "./script/deploy-storybook.sh"
   },
   "resolutions": {
-    "uglifyjs-webpack-plugin/uglify-js": "3.5.1",
     "firmata": "^2.2.0",
     "@types/react": "^18.0.28",
     "@types/react-dom": "^18.0.11"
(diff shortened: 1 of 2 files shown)


wilkie (approved):
LGTM! 👍 

Interestingly, it seems that the `ejs-webpack-loader` is one of only two things explicitly requiring `uglify-js` itself. The others are optional.

The other one is `grunt-contrib-uglify`. How necessary is that library now that you have removed the webpack plugin?

stephenliang:
> LGTM! 👍
> 
> Interestingly, it seems that the `ejs-webpack-loader` is one of only two things explicitly requiring `uglify-js` itself. The others are optional.
> 
> The other one is `grunt-contrib-uglify`. How necessary is that library now that you have removed the webpack plugin?

I did a quick test and yarn builds without `grunt-contrib-uglify` (the other ones like `grunt-contrib-watch` cause yarn to fail). I'll create a task to check if `grunt-contrib-uglify` can be removed.

PR #58999, opened 2024-05-30
Delete all datablock storage tables and kvps for user on account deletion

cnbrenci:
Fixes #57004 Delete all datablock storage tables and kvps for a given user on account deletion

PR #59692, opened 2024-07-10
Revert "Delete all datablock storage tables and kvps for user on account deletion"

snickell:
Reverts code-dot-org/code-dot-org#58999

The way we're using projects in tests here is causing other PRs to break (e.g. https://drone.cdn-code.org/code-dot-org/code-dot-org/10033/1/6): https://codedotorg.slack.com/archives/C0T0PNTM3/p1720624805278149


snickell:
FYI @cnbrenci

PR #59693, opened 2024-07-10
[Google Blockly] Feedback - find empty container blocks

mikeharv:
CDO Blockly has a custom feature that detects "empty" `if` and `repeat` blocks by searching for blocks with statement inputs that have no connected block. We had been assuming this was only used for CS in Algebra, but a failing UI test revealed it's also used in Maze and potentially other labs. This branch adds the missing functionality and updates the test file containing the relevant test. 

Original CDO Blockly code, used as a reference:
https://github.com/code-dot-org/blockly/blob/v4.1.0/core/utils/utils.js#L914-L936

The test is now passing with Google Blockly:
https://app.saucelabs.com/tests/983932d8a2ca4532bc5bb17828093e2c

This feature is tested in the second scenario of the maze2 test file. The other scenarios only needed to be updated with regard to block ids, similar to other recent PRs.

## Links

## PR Checklist:

- [ ] Code is well-commented


dashboard/test/ui/features/star_labs/maze2.feature
```diff
@@ -41,11 +41,11 @@ Feature: Complete a simple maze level
...
-    Then I drag block "1" to block "5"
-    And I drag block "4" to block "6"
-    And I drag block "1" into first position in repeat block "7"
-    And I drag block "2" to block "8"
-    And I drag block "1" to block "9"
+    Then I drag block "moveForward" to block "topBlock"
+    And I drag block "repeatForever" to block "moveForward"
+    And I drag block "turnLeft" into first position in repeat block "repeatForever"
+    And I drag block "moveForward" into first position in repeat block "repeatForever"
+    And I drag block "moveForward" to block "turnLeft"
```
mikeharv, line 48: You'll notice that we need several `move forward` blocks for this program:
image

However, we can only reliably predict the id for the first instance, placed by line 44 above. For this reason, I opted to place the `turn left` block inside the loop, and then add the second `move forward` block _above it_.

This results in the correct program but subtly changes the order in which the blocks are brought out. (Put another way, if the second `move forward` block were brought out first, we wouldn't be able to select it in order to place the `turn left` block below it.)

For a test like this where we're really interested in seeing the correct **feedback**, I'd suggest it could be more straight forward to simply provide a full program serialization and load it using Blockly APIs. Out of scope for now.

molly-moen (approved):

mikeharv:
Now includes https://github.com/code-dot-org/code-dot-org/pull/59702

PR #59695, opened 2024-07-10
Revert "Revert "Delete all datablock storage tables and kvps for user on account deletion""

snickell:
Restores: https://github.com/code-dot-org/code-dot-org/pull/58999

Which we reverted because it broke other tests in that "accessing projects causes SQL timeouts" sort of way:
```
deletes_all_of_user's_project_kvps", "DeleteAccountsHelperTest", 1420.216571637]
 test_Datablock_Storage:_hard-deletes_all_of_user's_project_kvps#DeleteAccountsHelperTest (1420.22s)
Minitest::UnexpectedError:         Sequel::DatabaseDisconnectError: Mysql2::Error::ConnectionError: Lost connection to MySQL server during query
            test/testing/projects_test_utils.rb:14:in `ensure in block in with_channel_for'
```

This type of issue usually occurs because of using both ActiveRecord /and/ "direct SQL" to access the projects table. Usually its caused by getting a project, but in this case, we're directly doing non-activerecord sql(`DASHBOARD_DB[:datablock_storage_tables]`) and I wonder if we're causing it that way.... this might just go away if we rewrite using activerecord (even if we keep it as raw sql, if we use activerecord's sql connection rather than DASHBOARD_DB)

### DIFF
--- a/dashboard/test/helpers/delete_accounts_helper_test.rb
+++ b/dashboard/test/helpers/delete_accounts_helper_test.rb
@@ -50,7 +50,6 @@ class DeleteAccountsHelperTest < ActionView::TestCase
     end
 
     # Skip real Firebase operations
-    # TODO: unfirebase, write a version of this for Datablock Storage: #57004
     # TODO: post-firebase-cleanup, switch to the datablock storage version: #56994
     FirebaseHelper.stubs(:delete_channel)
     FirebaseHelper.stubs(:delete_channels)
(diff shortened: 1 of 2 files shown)


snickell:
I've gone ahead and nixed the Firebase code paths in https://github.com/code-dot-org/code-dot-org/pull/59824, leaving this PR as a reminder we still need to figure out how to safely merge this without breaking the other tests.

cnbrenci (approved):
noice

PR #59697, opened 2024-07-10
limit ai evaluation requests per project for teachers and students

davidsbailey:
Finishes https://codedotorg.atlassian.net/browse/AITT-257. 

## description

When a student reaches the limit, we store an error code in the database, but there is no visible error to the teacher. they are simply prompted to request AI evaluation for the project: 
Screenshot 2024-07-11 at 3 24 34 PM

if a teacher reaches the limit, we display a warning and take away the option to request more ai evaluation:
Screenshot 2024-07-11 at 3 13 48 PM

### limits

Before using the limits of 5 suggested there, I did some digging in the database and found the following results:

* pilot students evaluate their own project:
  * 1.2 times on average (for projects evaluated at least once)
  * 10 times max
* pilot teachers evaluate each student project
  * 1.1 times on average (for projects evaluated at least once)
  * 6 times max

details can be found [here](https://docs.google.com/document/d/1VwV9jyCo3Uc6KeSAM14-BxNDNQyQbNokZnSuu29bs8A/edit). 

based on all of this, I'm proposing a limit of 5 times for students to ai-evaluate their own project (because the teacher can easily rerun ai when they are ready to), and a limit 10 times for teachers to request ai evaluation for their a given student project (better to err on the side of allowing teachers to get their work done, and the average number of 1.1 evals per project seems low enough for now). 

## Testing story

* new unit tests for ruby and js
* manual end-to-end testing shown in screenshots

## Future work

In the future we might want to tweak the limits or change this into more sophisticated rate limiting system, but we can tackle that when the need arises.


apps/test/unit/templates/rubrics/RubricContainerTest.jsx
```diff
@@ -706,12 +709,104 @@ describe('RubricContainer', () => {
...
+      0
+    );
+    expect(wrapper.text()).to.include(
+      i18n.aiEvaluationStatus_teacher_limit_exceeded({
+        limit: RubricAiEvaluationLimits.TEACHER_LIMIT,
+      })
+    );
+    expect(wrapper.find('Button').at(0).props().disabled).to.be.true;
+  });
+
```
Nokondi, line 811: Since these are new tests, can you do them in RTL?

Nokondi (commented):
As always, everything looks great! My only question was whether we should use RTL for the new unit tests.

apps/test/unit/templates/rubrics/RubricContainerTest.jsx
davidsbailey: Done!

PR #59699, opened 2024-07-10
chore(deps): upgrade to react 17

stephenliang:
This PR upgrades our react dependency to react 17. React 17 is an incremental upgrade which has no major incompatible changes from React 16. This is an incremental upgrade to lessen the load to upgrade to React 18 (and keep the PRs small).

 Primary changes were:

* The enzyme adapter for React 16 was changed to the React 17 version
* React 17 has a [different event listener methodology](https://blog.saeloun.com/2021/07/08/react-17-event-delagation/) which caused a different event propagation mechanism to occur. Certain features in the Code.org code base relies on a click event to open or close components and a separate click event listener to toggle on/off when clicking outside of the component. In React 16, the event will NOT bubble to the click event listener but in React 17, the event listener will bubble to the click event listener resulting in a duplicate event handling (this manifests as a double open or double close). This was fixed by stopping event propagation and explicitly handling the open/close rather than relying on the click listener.

## Links

- [React 17 release notes](https://legacy.reactjs.org/blog/2020/10/20/react-v17.html)
- [Event listener propagation](https://blog.saeloun.com/2021/07/08/react-17-event-delagation/)

## Testing story

1. Full UI test suite via Drone
2. We will do a quick bug bash as well

## Follow-up work

* React 18 upgrade to follow after React 17

## PR Checklist:

- [ ] Code is well-commented


mgc1194 (commented):
Drone tests passed.
Thank you so much for the link to the documentation. I understood the difference between React 16 vs 17. Why in our case attaching to the DOM instead of the document results in double-clicking is not yet super clear to me.
Are the few react components affected by this behaviour?

stephenliang:
> Drone tests passed. Thank you so much for the link to the documentation. I understood the difference between React 16 vs 17. Why in our case attaching to the DOM instead of the document results in double-clicking is not yet super clear to me. Are the few react components affected by this behaviour?

Previously events (like click) don't get bubbled to the top level so the event listener won't capture the event. However with the event changes in React 17, the events do get bubbled up to the top so the event listener will observe the click. So for all cases of where we have "I want to close this dialog if I click outside", what we actually want is "I want to close this dialog if I click outside excluding the control"). In react 16, the control was excluded (the component for which the event originated from) and in react 17, the control was included. This caused clicks on the control to be counted as "outside" and therefore result in a double close or double open.

carl-codeorg (approved):
LGTM

mgc1194 (approved):

PR #59704, opened 2024-07-10
AI Chat: add tooltips to presentation view

sanchitmalhotra126:
Add tooltips to the Presentation View (model card page). I pulled out the tooltip info icon into its own component so it could be shared across the different pages. Also a made a change `CollapsibleSection` where I removed the specific `titleVisualAppearance`/`titleIcon`/etc props in favor of passing in custom header content that consumers can style however they want. This way, the presentation view can pass in a custom header that includes an icon, text, and an icon with a tooltip without needing specific support for each distinct element inside `CollapsibleSection`. And other consumers of `CollapsibleSection` (like the edit AI chat settings components) can continue just passing in text.

https://github.com/code-dot-org/code-dot-org/assets/85528507/08712279-98cc-4874-ad04-ecd77ea29a9e

## Testing story

Tested on Gen AI pilot levels.

### DIFF
--- a/apps/src/aichat/views/InfoTooltipIcon.tsx
+++ b/apps/src/aichat/views/InfoTooltipIcon.tsx
@@ -1,0 +1,40 @@
+import classNames from 'classnames';
+import React from 'react';
+
+import FontAwesomeV6Icon from '@cdo/apps/componentLibrary/fontAwesomeV6Icon/FontAwesomeV6Icon';
+import {TooltipProps, WithTooltip} from '@cdo/apps/componentLibrary/tooltip';
+
+import styles from './info-tooltip-icon.module.scss';
+
+interface InfoTooltipIconProps {
+  id: string;
+  tooltipText: string;
+  direction?: TooltipProps['direction'];
+}
+
+const InfoTooltipIcon: React.FunctionComponent<InfoTooltipIconProps> = ({
+  id,
+  tooltipText,
+  direction,
+}) => {
+  return (
+    <WithTooltip
+      tooltipProps={{
+        text: tooltipText,
+        size: 's',
+        tooltipId: `${id}-tooltip`,
+        direction,
+        className: classNames(
+          styles.tooltip,
+          direction && styles[`tooltip-${direction}`]
+        ),
+      }}
+    >
+      <button type="button" className={styles.iconButton}>
+        <FontAwesomeV6Icon iconName={'info-circle'} className={styles.icon} />
+      </button>
+    </WithTooltip>
+  );
+};
+
+export default InfoTooltipIcon;
--- a/apps/src/aichat/views/info-tooltip-icon.module.scss
+++ b/apps/src/aichat/views/info-tooltip-icon.module.scss
@@ -1,0 +1,27 @@
+@import 'color.scss';
+@import '../../mixins.scss';
+
+@keyframes fade-in {
+  from {
+    opacity: 0;
+  }
+  to {
+    opacity: 1;
+  }
+}
+
+.tooltip {
+  animation: fade-in 0.3s;
+  &-onRight {
+    text-align: left;
+  }
+}
+
+.iconButton {
+  @include remove-button-styles;
+  font-size: 13px;
+}
+
+.icon {
+  color: $light_gray_500;
+}
--- a/apps/src/aichat/views/model-customization-workspace.module.scss
+++ b/apps/src/aichat/views/model-customization-workspace.module.scss
@@ -97,35 +97,12 @@
   margin: 10px 0;
 }
 
-.fieldLabel {
-  margin: 0;
-}
-
-@keyframes fade-in {
-  from {
-    opacity: 0;
-  }
-  to {
-    opacity: 1;
-  }
-}
-
-.tooltip {
-  animation: fade-in 0.3s;
-  text-align: left;
-}
-
 .fieldLabelContainer {
   display: flex;
   gap: 6px;
-  align-items: flex-start;
+  align-items: baseline;
+}
 
-  .iconButton {
-    @include remove-button-styles;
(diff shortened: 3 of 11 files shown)


apps/src/aichat/views/InfoTooltipIcon.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+import styles from './info-tooltip-icon.module.scss';
+
+interface InfoTooltipIconProps {
+  id: string;
+  tooltipText: string;
+  direction?: TooltipProps['direction'];
+}
+
+const InfoTooltipIcon: React.FunctionComponent<InfoTooltipIconProps> = ({
```
bencodeorg, line 15: This is slick! Is there anything like this in the design system @levadadenys ? I bet this would be nice to have for other teams as well.

apps/src/aichat/views/InfoTooltipIcon.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+import FontAwesomeV6Icon from '@cdo/apps/componentLibrary/fontAwesomeV6Icon/FontAwesomeV6Icon';
+import {TooltipProps, WithTooltip} from '@cdo/apps/componentLibrary/tooltip';
+
+import styles from './info-tooltip-icon.module.scss';
+
+interface InfoTooltipIconProps {
+  id: string;
+  tooltipText: string;
+  direction?: TooltipProps['direction'];
```
bencodeorg, line 12: Does this technically only support `onRight` currently?

apps/src/templates/CollapsibleSection.tsx
```diff
@@ -1,35 +1,20 @@
...
-  VisualAppearance as TypographyElementVisualAppearance,
-} from '@cdo/apps/componentLibrary/typography/types';
-import Typography from '@cdo/apps/componentLibrary/typography/Typography';
 
 import moduleStyles from './collapsible-section.module.scss';
 
 interface CollapsibleSectionProps {
-  title: string;
   children: React.ReactNode;
-  titleSemanticTag?: TypographyElementSemanticTag;
```
bencodeorg, line 15: Nice :)

bencodeorg (approved):
Slick!

apps/src/aichat/views/InfoTooltipIcon.tsx
sanchitmalhotra126: Well so currently only `onRight` will receive custom styling (`text-align: left`) but otherwise the direction will still determine where the tooltip appears.

apps/src/aichat/views/model-customization-workspace.module.scss
```diff
@@ -97,35 +97,12 @@
...
-}
-
 .fieldLabelContainer {
   display: flex;
   gap: 6px;
-  align-items: flex-start;
-
-  .iconButton {
-    @include remove-button-styles;
+  align-items: baseline;
```
fisher-alice, line 103: Just curious - why did you update `align-items` value to `baseline`?

apps/src/aichat/views/info-tooltip-icon.module.scss
```diff
@@ -0,0 +1,27 @@
...
+    opacity: 0;
+  }
+  to {
+    opacity: 1;
+  }
+}
+
+.tooltip {
+  animation: fade-in 0.3s;
+  &-onRight {
```
fisher-alice, line 15: Since info tooltip text is probably more verbose than general tooltips, what do you think of text being aligned to left no matter the direction of tooltip?

apps/src/aichat/views/InfoTooltipIcon.tsx
fisher-alice: +1 on having info tooltip handy for other teams. @markabarrett

apps/src/aichat/views/InfoTooltipIcon.tsx
bencodeorg: Ah got it!

apps/src/templates/CollapsibleSection.tsx
```diff
@@ -38,48 +23,20 @@ const CollapsibleSection: React.FunctionComponent<CollapsibleSectionProps> = ({
   const toggleCollapsed = useCallback(() => {
     setCollapsed(!collapsed);
   }, [collapsed, setCollapsed]);
-  const hasTitleIcon = titleIcon !== undefined;
 
   return (
     <>
-      <div className={moduleStyles.titleRow}>
```
fisher-alice, line 45: 🎉 🧹

fisher-alice (approved):
So nice! Just left a couple questions.

apps/src/aichat/views/model-customization-workspace.module.scss
sanchitmalhotra126: Good question, just a small tweak after I moved the `font-size` attribute from the `<i>` element to outer the `<button>` element. Previously, while the font-size for the icon was correct, the `button` had a larger font size and thus was aligning a little weirdly (which is why I used `flex-start` in the previous PR). With this change, `baseline` aligns the items more correctly.

apps/src/aichat/views/info-tooltip-icon.module.scss
sanchitmalhotra126: Hmm, we could but might be good to wait until we have other tooltips before making that assumption? I could imagine for some shorter tooltips that appear above/below that we still may want to use center alignment.

apps/src/aichat/views/model-customization-workspace.module.scss
fisher-alice: Thanks for the explanation!

PR #59710, opened 2024-07-11
Getting pl topics to save to the new table

hannahbergam:
I was able to get the topics saving to the database, and am able to see them loading on the front end form when one goes to edit a workshop!!!

I also am using this PR to do a few extra things and cleanup: 

First, I started utilizing the new Checkbox design component.
This also allowed me to delete some custom styling from the workshop_form file.
I created a constant on the front end for 'Build Your Own Workshops' and updated all the front end files that use it. 
I also renamed 'plTopics' to course_offerings to stay consistent with the table and make the saving logic more clear/obvious.
Finally, I added validation (on both the front end and the back end) so that a user can't save a BYOW course without selecting at least one PL topic.

I added back end tests to verify that a user can create and update Build Your Own workshop types, and to test that the validation/error messaging works if a user tries to create a workshop of that type but without any PL topics. This involved creating a new factory workshop and adding tests to the controller and model files.

Screenshot 2024-07-10 at 8 02 41 PM

## Links

- spec: []()
- jira ticket: https://codedotorg.atlassian.net/browse/ACQ-2022

## PR Checklist:

- [ ] Code is well-commented


apps/src/code-studio/pd/constants.js
```diff
@@ -21,3 +21,5 @@ export const PrivacyDialogMode = utils.makeEnum(
   'TEACHER_APPLICATION',
   'PRINCIPAL_APPROVAL'
 );
+
+export const BuildYourOwnWorkshop = 'Build Your Own Workshop';
```
TurnerRiley: Oh shoot I ended up also adding one too here. I'm happy to remove mine in one of my followup tasks! Do you think it makes more sense to have the constant at the level of `/code-studio/pd/` or /code-studio/pd/workshop_dashboard` ?

apps/src/code-studio/pd/workshop_dashboard/components/workshop_form.jsx
```diff
@@ -1160,71 +1169,78 @@ export class WorkshopForm extends React.Component {
...
-                    <button
-                      style={{...styles.fullWidth, ...styles.topicsButton}}
-                      className="btn btn-secondary dropdown-toggle"
-                      id="dropdownMenuButton"
-                      type="button"
-                      data-toggle="dropdown"
-                      aria-haspopup={true}
-                      aria-label="topics dropdown"
+            <FormGroup validationState={validation.style.course_offerings}>
+              {this.state.course === BuildYourOwnWorkshop && (
```
TurnerRiley: Should this line be outside the FormGroup?

apps/src/code-studio/pd/workshop_dashboard/components/workshop_form.jsx
```diff
@@ -1333,14 +1349,11 @@ const styles = {
     fontWeight: 100,
   },
   listItems: {
-    margin: '0 10px 0 0',
+    margin: '10px',
```
TurnerRiley: Nit: I think this can just be `margin: 10` now (definitely not worth a rebuild if this is the only change)

TurnerRiley (approved):
LGTM! I feel like maybe another pair of eyes on the backend stuff would be good since we paired on it (feels like a conflict of interest haha)

apps/src/code-studio/pd/constants.js
hannahbergam: oh gosh my bad! And oh I don't care either way- I just got excited when I saw that pd shared constants file and threw it in there.

hannahbergam: Is yours merged yet? I can move this one and update the imports

apps/src/code-studio/pd/workshop_dashboard/components/workshop_form.jsx
hannahbergam: Moved this and the <row!

apps/src/code-studio/pd/constants.js
hannahbergam: I moved mine to that workshopConstants file!

apps/src/code-studio/pd/workshop_dashboard/components/workshop_form.jsx
hannahbergam: Good catch- updated!

apps/src/code-studio/pd/constants.js
hannahbergam: LOL my bad- I hadn't merged staging yet today. Okay I ended up leaving the longer version to match the backed constant, resolved the merge conflict and updated the file you're using it in!

dashboard/app/controllers/api/v1/pd/workshops_controller.rb
```diff
@@ -278,6 +296,23 @@ def potential_organizers
...
+    course_offering_ids_to_remove.each do |course_offering_id|
+      @workshop.course_offerings.delete(course_offering_id)
+    end
+
+    new_course_offering_ids.each do |course_offering_id|
+      course_offering = CourseOffering.find_by(id: course_offering_id)
+      # Return if we don't find a corresponding/valid Course Offering
+      next unless course_offering
+      @workshop.course_offerings << course_offering
+    end
```
bethanyaconnor: not 100% sure this would work, but you might be able simplify this to 
```
@workshop.course_offerings = supplied_course_offering_ids.map {|id| CourseOffering.find_by(id: course_offering_id) }.compact
```

dashboard/test/controllers/api/v1/pd/workshops_controller_test.rb
```diff
@@ -437,6 +437,35 @@ class Api::V1::Pd::WorkshopsControllerTest < ActionController::TestCase
...
+        suppress_email: false,
+        sessions_attributes: [
+          {
+            start: session_start,
+            end: session_end
+          }
+        ]
+      }
+
+    workshop = create :pd_workshop, funded: false, course: Pd::Workshop::COURSE_BUILD_YOUR_OWN, subject: nil, course_offerings: [] << (create :course_offering)
```
bethanyaconnor, line 463: nit: can you use the `build_your_own` factory here?

dashboard/test/factories/workshop_factories.rb
```diff
@@ -340,5 +340,13 @@
       num_facilitators {2}
       each_session_hours {7}
     end
+
+    factory :build_your_own do
```
bethanyaconnor: nit: would it be clearer to name this something like `build_your_own_workshop`? I don't feel that strongly though if you think this name is better!

dashboard/test/models/pd/workshop_test.rb
```diff
@@ -1492,6 +1492,11 @@ class Pd::WorkshopTest < ActiveSupport::TestCase
     assert workshop.valid?
   end
 
+  test 'Build Your Own Workshop type is valid' do
```
bethanyaconnor: I'm curious what this is testing? As is it seems like it's testing the factory itself?

hannahbergam: Just testing that it's possible to make a valid BYO workshop (pass the backend validation). But if I'm going to move the other test to use these params, I can delete this one.

dashboard/test/factories/workshop_factories.rb
hannahbergam: Removing this entirely!

dashboard/app/controllers/api/v1/pd/workshops_controller.rb
hannahbergam: I got a version of this working with filter_map! TIL Rails magic will go through and delete join table associations without me having to do it explicitly:) Thanks!

dashboard/test/controllers/api/v1/pd/workshops_controller_test.rb
hannahbergam: I ended up deleting the factory!

bethanyaconnor (approved):
🎉

apps/src/code-studio/pd/constants.js
TurnerRiley: Yep mine is merged!

TurnerRiley: Thanks for updating mine too!

PR #59718, opened 2024-07-11
Add text and image for progress launch banner

bethanyaconnor:
Adds text and image for the progress teacher homepage banner. I'm still waiting on a link and whether this needs to be behind a DCDO flag, but this change will get the banner ready for translations.


TurnerRiley (approved):
LGTM!

PR #59721, opened 2024-07-11
Fix console errors from expanding metadata about divs in tr tags

lfryemason:
Should be no visible changes.

The only change is that an expanded cell uses a <td> tag instead of <div> and the <td> only contains <div> tags.

## Links
[TEACH-1183](https://codedotorg.atlassian.net/browse/TEACH-1183)

## PR Checklist:

- [ ] Code is well-commented


kobryan0619 (approved):
Looks good!

PR #59722, opened 2024-07-11
Added redirect for marketing banner.

kobryan0619:
The deliverable here is just a URL that can be given to the marketing team for an announcement banner on the site when a teacher is logged in.  This banner should allow a user to try the new view by redirecting them to the V2 version of the first section in their section list.  If they have not created a section yet, it will redirect them to a support article. 

The URL is: studio.code.org/teacher_dashboard/sections/first_section_progress

## Links

[Ticket](https://codedotorg.atlassian.net/jira/software/projects/TEACH/boards/65)

## Testing story

Tested locally and wrote a test for it.

### DIFF
--- a/apps/src/templates/sectionProgressV2/SectionProgressSelector.jsx
+++ b/apps/src/templates/sectionProgressV2/SectionProgressSelector.jsx
@@ -1,8 +1,9 @@
 import _ from 'lodash';
 import PropTypes from 'prop-types';
-import React, {useCallback} from 'react';
+import React, {useCallback, useEffect, useState} from 'react';
 import {connect} from 'react-redux';
 
+import {queryParams} from '@cdo/apps/code-studio/utils';
 import Link from '@cdo/apps/componentLibrary/link';
 import DCDO from '@cdo/apps/dcdo';
 import {EVENTS} from '@cdo/apps/lib/util/AnalyticsConstants';
@@ -27,7 +28,24 @@ function SectionProgressSelector({
   sectionId,
   hasSeenProgressTableInvite,
 }) {
-  const [hasJustToggledViews, setHasJustToggledViews] = React.useState(false);
+  const [hasJustToggledViews, setHasJustToggledViews] = useState(false);
+
+  useEffect(() => {
+    const params = queryParams('view');
+    if (params === 'v2') {
+      setShowProgressTableV2(true);
+      setHasJustToggledViews(true);
+    }
+  }, [setShowProgressTableV2, setHasJustToggledViews]);
+
+  const removeQueryParams = () => {
+    const url =
+      window.location.protocol +
+      '//' +
+      window.location.host +
+      window.location.pathname;
+    window.history.pushState({path: url}, '', url);
+  };
 
   const onShowProgressTableV2Change = useCallback(() => {
     const shouldShowV2 = !showProgressTableV2;
@@ -43,6 +61,7 @@ function SectionProgressSelector({
       analyticsReporter.sendEvent(EVENTS.PROGRESS_V2_VIEW_OLD_PROGRESS, {
         sectionId: sectionId,
       });
+      removeQueryParams();
     }
   }, [showProgressTableV2, setShowProgressTableV2, sectionId]);
 
@@ -80,7 +99,14 @@ function SectionProgressSelector({
   // If the user has not selected manually the v1 or v2 table, show the DCDO defined default.
   // If a user has selected manually, show that version.
   const isPreferenceSet = showProgressTableV2 !== undefined;
-  const displayV2 = isPreferenceSet
+  const params = queryParams('view');
+
+  // If there is a url pram, use that param to determine to show V2.
+  const displayV2FromUrl = params === 'v2';
+
+  const displayV2 = displayV2FromUrl
+    ? true
+    : isPreferenceSet
     ? showProgressTableV2
     : DCDO.get('progress-table-v2-default-v2', false);
 
--- a/apps/src/templates/sectionProgressV2/SectionProgressSelector.jsx
+++ b/apps/src/templates/sectionProgressV2/SectionProgressSelector.jsx
@@ -35,6 +35,7 @@ function SectionProgressSelector({
     if (params === 'v2') {
       setShowProgressTableV2(true);
       setHasJustToggledViews(true);
+      new UserPreferences().setShowProgressTableV2(true);
     }
   }, [setShowProgressTableV2, setHasJustToggledViews]);
 
@@ -104,11 +105,11 @@ function SectionProgressSelector({
   // If there is a url pram, use that param to determine to show V2.
   const displayV2FromUrl = params === 'v2';
 
-  const displayV2 = displayV2FromUrl
-    ? true
-    : isPreferenceSet
-    ? showProgressTableV2
-    : DCDO.get('progress-table-v2-default-v2', false);
+  const displayV2 =
+    displayV2FromUrl ||
+    (isPreferenceSet
+      ? showProgressTableV2
+      : DCDO.get('progress-table-v2-default-v2', false));
 
   const toggleV1OrV2Link = () => (
     <div className={styles.toggleViews}>
--- a/dashboard/app/controllers/teacher_dashboard_controller.rb
+++ b/dashboard/app/controllers/teacher_dashboard_controller.rb
@@ -8,6 +8,15 @@ class TeacherDashboardController < ApplicationController
     view_options(full_width: true, no_padding_container: true)
   end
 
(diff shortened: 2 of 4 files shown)


apps/src/templates/sectionProgressV2/SectionProgressSelector.jsx
```diff
@@ -27,7 +28,24 @@ function SectionProgressSelector({
...
+    }
+  }, [setShowProgressTableV2, setHasJustToggledViews]);
+
+  const removeQueryParams = () => {
+    const url =
+      window.location.protocol +
+      '//' +
+      window.location.host +
+      window.location.pathname;
+    window.history.pushState({path: url}, '', url);
```
lfryemason, line 48:
```suggestion
      '${window.location.protocol}//
      ${window.location.host}${window.location.pathname}';
    window.history.pushState({path: url}, '', url);
```
nit: I've been preferring this kind of string concatenation because I think it makes it clearer what the end result will look like. Totally optional if you disagree though.

apps/src/templates/sectionProgressV2/SectionProgressSelector.jsx
```diff
@@ -80,7 +99,14 @@ function SectionProgressSelector({
...
+  const params = queryParams('view');
+
+  // If there is a url pram, use that param to determine to show V2.
+  const displayV2FromUrl = params === 'v2';
+
+  const displayV2 = displayV2FromUrl
+    ? true
+    : isPreferenceSet
     ? showProgressTableV2
     : DCDO.get('progress-table-v2-default-v2', false);
```
lfryemason:
```suggestion
  const displayV2 = displayV2FromUrl || (
    isPreferenceSet
    ? showProgressTableV2
    : DCDO.get('progress-table-v2-default-v2', false));
```

I think this is slightly clearer, but another optional nit.

apps/src/templates/sectionProgressV2/SectionProgressSelector.jsx
```diff
@@ -27,7 +28,24 @@ function SectionProgressSelector({
   sectionId,
   hasSeenProgressTableInvite,
 }) {
-  const [hasJustToggledViews, setHasJustToggledViews] = React.useState(false);
+  const [hasJustToggledViews, setHasJustToggledViews] = useState(false);
+
+  useEffect(() => {
+    const params = queryParams('view');
+    if (params === 'v2') {
+      setShowProgressTableV2(true);
```
lfryemason, line 36: Do we also need to add anything from else from `onShowProgressTableV2Change` below?

Should we set the UserPreference or emit metrics when this happens?

dashboard/config/routes.rb
```diff
@@ -7,6 +7,9 @@
   # Redirect studio.code.org/courses to code.org/students
   get "/courses", to: redirect(CDO.code_org_url("/students"))
 
+  # Redirect studio.code.org/mysection to most recent section section
+  get '/teacher_dashboard/sections/firstSectionProgress', to: "teacher_dashboard#redirect_to_newest_section"
```
lfryemason:
```suggestion
  get '/teacher_dashboard/sections/first_section_progress', to: "teacher_dashboard#redirect_to_newest_section"
```
Looks like these should be underscore_case. Also, It might make sense to run the name by Bryan if you haven't already

dashboard/app/controllers/teacher_dashboard_controller.rb
```diff
@@ -8,6 +8,15 @@ def show
     view_options(full_width: true, no_padding_container: true)
   end
 
+  def redirect_to_newest_section
+    if current_user.sections_instructed.empty?
+      redirect_to "https://support.code.org/hc/en-us/articles/25195525766669-Getting-Started-New-Progress-View"
+    else
+      section_id = current_user.sections_instructed.order(created_at: :desc).first.id
```
lfryemason, line 15: Very nice

dashboard/config/routes.rb
```diff
@@ -7,6 +7,9 @@
   # Redirect studio.code.org/courses to code.org/students
   get "/courses", to: redirect(CDO.code_org_url("/students"))
 
+  # Redirect studio.code.org/mysection to most recent section section
```
lfryemason: I don't think the path on this comment is right?

dashboard/test/controllers/teacher_dashboard_controller_test.rb
```diff
@@ -43,4 +43,22 @@ class TeacherDashboardControllerTest < ActionController::TestCase
     get :show, params: {section_id: cotaught_section.id}
     assert_response :success
   end
+
+  test 'redirect_to_newest_section: redirects to support URL if no sections instructed' do
```
lfryemason, line 47: Beautiful

dashboard/test/controllers/teacher_dashboard_controller_test.rb
```diff
@@ -43,4 +43,22 @@ class TeacherDashboardControllerTest < ActionController::TestCase
...
+
+    assert_redirected_to 'https://support.code.org/hc/en-us/articles/25195525766669-Getting-Started-New-Progress-View'
+  end
+
+  test 'redirect_to_newest_section: redirects to newest section progress page if sections instructed' do
+    sign_in @section_owner
+
+    get :redirect_to_newest_section
+
+    newest_section = @section_owner.sections_instructed.order(created_at: :desc).first
```
lfryemason: Instead of getting the id using the same logic here, would it be possible to know which section was created last and call that section ID directly? Like it is, this doesn't actually test whether it retrieves the correct section, just whether it uses this same logic

lfryemason (commented):
I was feeling picky today, nothing to do with your code :). This looks great

dashboard/config/routes.rb
kobryan0619: @dju90 thoughts on the formatting of the URL or the name?  This would be easy to change also if you had an entirely different vision :)

dju90: Since the URL/name redirects immediately, I don't have a strong opinion what it should be. `first_section_progress` sounds good, as does `newest_section_progress` if you're looking for an alternative

kobryan0619: Great - let's go with `first_section_progress`

dashboard/config/routes.rb
kobryan0619: You're right - that is dated.  I will update.

apps/src/templates/sectionProgressV2/SectionProgressSelector.jsx
kobryan0619: Ah, yeah, I do think we want to have a set UserPreferences...  I will add that.  The current metrics include a `Section New Progress Viewed` amplitude event.  @dju90 are there other metrics you want logged from this redirect from marketing or does marketing provide you all that data?

apps/src/templates/sectionProgressV2/SectionProgressSelector.jsx
kobryan0619: This one was rougher for me to like :)

If it is optional to you, I think I will stick with the original here.

apps/src/templates/sectionProgressV2/SectionProgressSelector.jsx
lfryemason: Still need the UserPreference

lfryemason (approved):
Approving assuming you will add the UserPreference line

apps/src/templates/sectionProgressV2/SectionProgressSelector.jsx
dju90: There's Google Analytics showing clicks on this marketing banner, so no other amplitude metrics needed at this time.

dju90: > Still need the UserPreference

What's the status of this?

apps/src/templates/sectionProgressV2/SectionProgressSelector.jsx
```diff
@@ -27,7 +28,25 @@ function SectionProgressSelector({
...
+      setHasJustToggledViews(true);
+      new UserPreferences().setShowProgressTableV2(true);
```
kobryan0619, line 38: @dju90 this should be resetting the UserPreferences.  I should have commented on the thread that this was included.

apps/src/templates/sectionProgressV2/SectionProgressSelector.jsx
kobryan0619: See below for where the update UserPreferences is included.

PR #59723, opened 2024-07-11
Wrap MailJet call in a begin/rescue on sign up

bethanyaconnor:
Wrap the MailJet call in a begin/rescue. We've seen a couple of sporadic MailJet errors, normally transient issues. When those happen, it creates a weird user experience where the account is created but the user isn't logged in. 

As I was testing the translated welcome emails, I realized I didn't want to test with production keys, but that leaves the possibility of errors in production (most likely, an incorrect template ID). This solves that as _most_ testing can be done without production keys, but I can feel confident that if there is a production-specific issue, it won't create a lot of user pain.

### DIFF
--- a/dashboard/app/controllers/registrations_controller.rb
+++ b/dashboard/app/controllers/registrations_controller.rb
@@ -126,7 +126,17 @@ class RegistrationsController < Devise::RegistrationsController
     if current_user && current_user.errors.blank?
       if current_user.teacher?
         if MailJet.enabled? && request.locale != 'es-MX'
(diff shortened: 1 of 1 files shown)


hannahbergam (approved):
Makes sense to me!

bethanyaconnor:
I should've mentioned -- my testing here was that I mocked an error in the `Mailjet` module. I found a bug in my error handling, fixed it, then confirmed sign up continued as usual. (I love finding bugs in my testing, builds confidence). I also confirmed that without the error handling I added here, the user sees a 500.

PR #59724, opened 2024-07-11
[JIMT][CT-576] - implemented file downloads

thomasoniii:
Adds a download button to the kebab menu in the file browser.

Screenshot 2024-07-11 at 1 01 26 PM

Just hit the kebab menu on a file, and you'll get a new download prompt.

Notes and errata:
- I poked around for a library to use and finally opted for `js-file-download`. It hasn't been updated _in years_, which is bad, but it just implements a fairly standard function that's in myriad stackoverflow responses to this question. So I figured we'd be better off using this old library instead of just manually implementing the same function. Can definitely change this if we have something preferred.
- There's some security in that the `downloadFile` function has an arbitrary list of file types that we can download. So as to prevent the user from downloading a `dmg`, even if one is uploaded. We need to make sure that this list is complete, and possibly move it elsewhere if it needs to be better configured.
- There's no support for download files. Maybe we want that? It's doable.
- There's also no support for uploading files (yet). There are some security considerations that we need to get right about it, but it should be doable in a later rev.

## Links

- jira ticket: [CT-576](https://codedotorg.atlassian.net/browse/CT-576)

## PR Checklist:

- [ ] Code is well-commented


apps/src/codebridge/FileBrowser/index.tsx
```diff
@@ -290,6 +299,20 @@ export const FileBrowser = React.memo(() => {
     [newFolder, project.folders]
   );
 
+  const downloadFile: FilesComponentProps['downloadFile'] = useMemo(
+    () => fileId => {
+      const file = project.files[fileId];
+      // NOTE - this is arbitrary and maybe we need more? Also, maybe this should be elsewhere or more easy to configure?
+      const downloadableFileType = new Set(['txt', 'html', 'css', 'py', 'csv']);
```
molly-moen: could we reference the `editableFileTypes` list from the config instead?

apps/src/codebridge/FileBrowser/index.tsx
```diff
@@ -290,6 +299,20 @@ export const FileBrowser = React.memo(() => {
...
+      if (downloadableFileType.has(file.language)) {
+        fileDownload(file.contents, file.name);
+      } else {
+        alert(codebridgeI18n.downloadFileError({extension: file.language}));
```
molly-moen: can you file a task to replace all the `alerts` in this file with our own dialogs?

molly-moen (approved):
Looks good, one suggestion!

apps/src/codebridge/FileBrowser/index.tsx
thomasoniii: I went one better and removed the dialog entirely. 😆 

Upon reflection, it was silly to have the `download file` menu item always appear and pop up an alert if the file couldn't be downloaded, so I retooled it to only display the `download file` option if the file's type is one of the one's in the config's `editableFileTypes`.

PR #59731, opened 2024-07-11
[CodeBridge] Rename source to start_sources

mikeharv:
From [ticket](https://codedotorg.atlassian.net/browse/CT-644):
> When we added start sources to web lab 2 and python lab we started using `source` rather than `start_sources` in the level configuration. Java Lab uses `start_sources`. In order to make things simpler to eventually migrate Java Lab levels to the same system, we should rename this field in Python Lab and Web Lab 2. Since there's few levels right now and they are all test levels it's easier to do this sooner rather than later.

This renames the property in the 10 existing PythonLab levels. It renames the property in code where we would read this property and when we save start blocks.

## Process used
1. I first manually edited the levels and re-seeded them (to avoid extra diff for the audit log, etc.). After reloading one of the levels, I confirmed we were falling back to the default source.
2. After renaming the `levelProperties` key, I confirmed that we were once again using the (renamed) start sources.
3. I also renamed the value that gets returned by the levelbuilder save button and confirmed that start mode still works.

Once this merges, I would recommend all CodeBridge engineers re-seed these custom levels before modifying them using levelbuilder/start mode. If the levels aren't reseeded locally, changes will potentially re-introduce the deprecated `source` property. (`start_sources` would still be added as expected, but it won't overwrite the existing `source` in your locally seeded copy.

## Links

https://codedotorg.atlassian.net/browse/CT-644

## Testing story

I didn't see any test coverage that expects a particular spelling for this property.

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/codebridge/hooks/useInitialSources.ts
+++ b/apps/src/codebridge/hooks/useInitialSources.ts
@@ -17,7 +17,7 @@ import {useAppSelector} from '@cdo/apps/util/reduxHooks';
 export const useInitialSources = (defaultSources: ProjectSources) => {
   const labInitialSources = useAppSelector(state => state.lab.initialSources);
   const levelStartSource = useAppSelector(
-    state => state.lab.levelProperties?.source
+    state => state.lab.levelProperties?.startSources
   );
   const exemplarSources = useAppSelector(
     state => state.lab.levelProperties?.exemplarSources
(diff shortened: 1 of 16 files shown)


molly-moen (commented):
Looks good! There's also a backend change needed to make sure template levels work, and you'll want to update the property value in pythonlab and web lab 2. I think [this commit](https://github.com/code-dot-org/code-dot-org/commit/328e781d5c5f3b2cb35baca71925aa7c43e9b921) points to the code to change.

molly-moen (approved):
Nice!

PR #59733, opened 2024-07-11
Datablock Storage: Don't enforce MAX_TABLE_COUNT if adding a shared table.

### LINKED ISSUE #59732: datasets controller can't add new shared tables because they exceed max table constraint
Thanks @kakiha11 for bringing this to my attention! We had discovered this bug in seeding the shared tables locally, but didn't realize that it also was blocking new dataset upload.  Because of that, we didn't prioritize getting the fix in right away. I have prepped a fix!

cnbrenci:
Fixes #59732 

The MAX_TABLE_COUNT constraint applies to applab projects to limit the number of data tables students can create and import in their project. Since we store shared tables (from the Data Library) alongside other project tables in the DB with a unique project Id, they were being erroneously subject to the same constraint when new shared tables are added to the data library by level builders.

Changes in this PR:
- Do not apply MAX_TABLE_COUNT constraint if table being created is a shared table. (Note, this is when a new shared table is added by a curriculum writer through levelbuilder. The constraint is still in place for students importing shared tables to their projects.)
- Add test coverage to validate that adding shared datasets through levelbuilder will work

Note: This will also fix local seeding of shared tables.

### DIFF
--- a/dashboard/test/controllers/datasets_controller_test.rb
+++ b/dashboard/test/controllers/datasets_controller_test.rb
@@ -34,4 +34,20 @@ class DatasetsControllerTest < ActionController::TestCase
 
     assert_equal JSON.parse(@test_manifest.to_json), DatablockStorageLibraryManifest.instance.library_manifest
   end
+
+  test 'update: can exceed max table count' do
+    original_max_table_count = DatablockStorageTable::MAX_TABLE_COUNT
+    DatablockStorageTable.const_set(:MAX_TABLE_COUNT, 1)
+
+    csv_data = <<~CSV
+      id,name,age
+      1,fluffy,7
+    CSV
+
+    post :update, params: {dataset_name: 'cats', csv_data: csv_data}
+    post :update, params: {dataset_name: 'dogs', csv_data: csv_data}
+
+  ensure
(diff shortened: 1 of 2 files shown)


snickell:
I don't think we can do this, because what happens if they add a row to the table, converting it into not a shared table?

snickell:
I got confused here for a bit, just to clarify for future pepes:

This removes the constraint when a curriculum author DEFINES shared tables.

This does /not/ remove the constraint for a student adding shared tables to their project.

snickell (approved):
Looks good, sorry @kakiha11 🙈

dashboard/test/controllers/datasets_controller_test.rb
```diff
@@ -34,4 +34,24 @@ class DatasetsControllerTest < ActionController::TestCase
 
     assert_equal JSON.parse(@test_manifest.to_json), DatablockStorageLibraryManifest.instance.library_manifest
   end
+
+  test 'update: can exceed max table count' do
```
fisher-alice: Nit: maybe add to test name '...in levelbuilder mode' or '...if created by a curriculum author'

fisher-alice (approved):
Thanks for the test name/comment update!

PR #59737, opened 2024-07-11
Added UI test for metadata

kobryan0619:
Adds UI test for metadata feature of progress view (V2)

Checks:

- Expand all and collapse all work
- Checks that you can open and close a specific student
- Checks that after a student has completed work, the data shows integers for when it was completed and how long it took.


apps/src/templates/sectionProgressV2/LessonDataCell.jsx
```diff
@@ -82,10 +82,16 @@ function LessonDataCell({
     return (
       <div className={styles.lessonDataCellExpanded}>
         {lessonCellUnexpanded}
-        <div className={classNames(styles.gridBox, styles.gridBoxMetadata)}>
+        <div
+          id={'ui-test-time-spent-' + lesson.relative_position}
```
lfryemason: This might be fine, but this isn't enough to identify a specific LessonDataCell. You need lesson and student

dashboard/test/ui/features/teacher_tools/teacher_dashboard/teacher_dashboard_progress_v2.feature
```diff
@@ -28,6 +28,52 @@ Scenario: Teacher can open and close Icon Key and details
...
+
+  # Toggle to V2 progress view
+  And I wait until element "h6:contains(Icon Key)" is visible
+  And I wait until element "#ui-test-progress-table-v2" is visible
+
+  # Can see and open menu with more options
+  And I wait until element "#ui-see-more-options-dropdown" is visible
+  Then I click selector "#ui-see-more-options-dropdown"
+  And I wait until element "div:contains(Expand all student rows)" is visible
+  And I wait until element "div:contains(Collapse all student rows)" is visible
```
lfryemason, line 53: Nit: this doesn't need to be wait until. Both will load at the same time so the second can is just `and _ is visible` or whatever it is.

dashboard/test/ui/features/teacher_tools/teacher_dashboard/teacher_dashboard_progress_v2.feature
```diff
@@ -28,6 +28,52 @@ Scenario: Teacher can open and close Icon Key and details
...
+
+  # Can click on more options and it responds appropriately
+  Then I click selector "#ui-test-expand-all"
+  And I wait until element "div:contains(Last Updated)" is visible
+  And I wait until element "div:contains(Time Spent)" is visible
```
lfryemason, line 58: Same here

dashboard/test/ui/features/teacher_tools/teacher_dashboard/teacher_dashboard_progress_v2.feature
```diff
@@ -28,6 +28,52 @@ Scenario: Teacher can open and close Icon Key and details
...
+  And I wait until "#ui-test-time-spent-44" contains one or more integers
+  Then I click selector "#ui-see-more-options-dropdown"
+  Then I click selector "#ui-test-collapse-all"
+  And element "div:contains(Time Spent)" does not exist
+  And element "div:contains(Last Updated)" does not exist
+
+  # Can click on individual row and it opens with lesson and level data
+  Then I click selector "#ui-test-student-row-unexpanded-Sally"
+  And I wait until element "div:contains(Last Updated)" is visible
+  And I wait until element "div:contains(Time Spent)" is visible
```
lfryemason, line 70: Same here

lfryemason (approved):
Approving assuming that the first comment is expected and the nits are non-blocking

dashboard/test/ui/features/teacher_tools/teacher_dashboard/teacher_dashboard_progress_v2.feature
kobryan0619: Ugh... this is where I feel like I get nervous - I know "technically" we don't need the "wait until" BUT it seems like sometimes what causes some flaky tests is inconsistent loading (thus my preference to use "wait until" more often than not).  I am wondering if you see any harm in defaulting to it - thoughts?

apps/src/templates/sectionProgressV2/LessonDataCell.jsx
kobryan0619: Yeah, I knew that, but also had a really hard time finding a truly unique identifier for this.  After you comment here I did some research on the harm caused by not having a truly unique identifier and it seems accessibility related.  For that reason, I found a way to move this identifier to the classNames instead of ID and that seems to work.

PR #59741, opened 2024-07-11
[CodeBridge] always use level source in start mode

mikeharv:
From [ticket](https://codedotorg.atlassian.net/browse/CT-645):
> After https://github.com/code-dot-org/code-dot-org/pull/59312 , if you open a template-backed level in start mode you will always see the source of the template level. Editing and saving changes will save to the original level, but when you reload you will see the template level sources again.

This is happening due to a recent back-end change that _always_ prioritizes the start sources from the template level if they are there. In `summarize_for_lab2_properties` , we check whether the level has a template. if it does, we always use those sources. This back-end method can’t tell that we’re in start mode, so instead we can create `startSources` and `templateSources` at this point. Then on the front end, we can pick and choose based on 1) whether or not template sources exist and 2) whether or not we’re in start mode.

With these changes, we should now be using the following sources:
- level view with project -> project sources
- level view without project/progress -> template sources if they exist, or level sources
- exemplar mode -> template sources if they exist, or level sources
- start mode -> level sources

I made a judgement call for exemplar mode. I would assume that in a template-backed progression that the start blocks are typically unset in the progression levels, and it makes sense to base the exemplar on the start code that students will see.

I confirmed that we see the expected source code upon initial page load and after clicking the reset button.

Level view:
image

Exemplar Mode:
image

Start Mode:
image

## Links

https://codedotorg.atlassian.net/browse/CT-645

## PR Checklist:

- [ ] Code is well-commented


apps/src/codebridge/hooks/useInitialSources.ts
```diff
@@ -19,34 +19,47 @@ export const useInitialSources = (defaultSources: ProjectSources) => {
...
   const initialSources = useMemo(() => {
     const startSources = projectStartSource || defaultSources;
+    const templateSources = templateStartSource;
 
     if (isStartMode) {
       return startSources;
     }
     if (isEditingExemplar || isViewingExemplar) {
       // If we are viewing exemplars sources and have no exemplar, we show a fallback
       // page from LabViewsRenderer. We fall back to start sources for editing.
```
molly-moen: nit: update this comment to include reference to template sources?

apps/src/codebridge/hooks/useInitialSources.ts
```diff
@@ -19,34 +19,47 @@ export const useInitialSources = (defaultSources: ProjectSources) => {
   const levelStartSource = useAppSelector(
```
molly-moen, line 29: I can't comment higher, but can you update the comment above to include the new logic with template sources?

dashboard/app/models/levels/level.rb
```diff
@@ -844,9 +844,10 @@ def summarize_for_lab2_properties(script, script_level = nil, current_user = nil
     properties_camelized[:appName] = game&.app
     properties_camelized[:useRestrictedSongs] = game.use_restricted_songs?
     properties_camelized[:usesProjects] = try(:is_project_level) || channel_backed?
-    if try(:project_template_level).try(:start_sources) || try(:start_sources)
-      # Override start sources with project template level sources if they exist
-      properties_camelized['startSources'] = try(:project_template_level).try(:start_sources) || try(:start_sources)
+
+    properties_camelized['startSources'] = try(:start_sources)
```
molly-moen: do we need this anymore? I think `properties.camelize_keys` will camelize start_sources correctly and put it in properties_camelized.

molly-moen (commented):
Looks good, couple small comments!

dashboard/app/models/levels/level.rb
mikeharv: Confirmed that it works fine without it!

apps/src/codebridge/hooks/useSource.ts
```diff
@@ -53,8 +56,19 @@ export const useSource = (defaultSources: ProjectSources) => {
   );
 
   const resetToStartSource = useCallback(() => {
-    setSource(levelStartSource || (defaultSources.source as MultiFileSource));
-  }, [defaultSources.source, levelStartSource, setSource]);
+    // When resetting in start mode, we always use the level start source.
+    setSource(
+      (!isStartMode && templateStartSource) ||
```
mikeharv, line 61: @molly-moen Note this change in behavior when resetting. I think there's an open question around what the "initial" sources for start mode should be when resetting the workspace.

Previously, we would use the template sources, which was obviously a bug. Here, I propose using the level's (most recently saved) start sources, which matches the other labs I'm familiar with.

molly-moen: It makes sense to me to reset to the most recent start sources!

molly-moen (approved):

PR #59745, opened 2024-07-11
Python Lab Submittable Levels part 2

molly-moen:
Add the following features to python lab submittable levels:
- Do not allow users to submit until they have run their code at least once this session.
- Show a confirmation modal on submit/unsubmit

### Screen recording
https://github.com/user-attachments/assets/0bd09a79-fe6f-49d5-ac90-a258ef3b8f63

## Links

- jira ticket: [CT-621](https://codedotorg.atlassian.net/browse/CT-621)

## Testing story
Tested locally. This change only impacts python lab. 

## Follow-up work
This behavior matches our existing submittable level behavior in legacy labs. There are some potential improvements we could consider:
- We don't have any error handling if submit or unsubmit fail. In the case of submit, if we fail we still navigate to the next level, if there is one. I filed a [follow up task](https://codedotorg.atlassian.net/browse/CT-684) to improve this. We may want to do some sort of unified error handling for our various reporting handlers in progress redux (explained in the ticket).
- There's no explanation on the page as to why submit is disabled when you load the page. I will check with curriculum to see if we want some sort of hover tip to explain this.

## PR Checklist:

- [ ] Code is well-commented


apps/src/codebridge/ControlButtons/index.tsx
```diff
@@ -53,22 +59,49 @@ const ControlButtons: React.FunctionComponent = () => {
...
+      dialogMessage
+    );
+  };
+
+  const handleSubmit = () => {
     dispatch(
       sendSubmitReport({appType: appType || '', submitted: !hasSubmitted})
     );
+    // Go to the next level if we have one and we just submitted.
+    // TODO: If onContinue starts to update progress, make sure we don't override the submitted status.
```
ebeastlake, line 83: Do we need to file a ticket for this? Who would determine whether `onContinue` starts to update progress? Or is this a "if you come to this code and modify `onContinue`, remember to..."?

apps/src/codebridge/ControlButtons/index.tsx
```diff
@@ -53,22 +59,49 @@ const ControlButtons: React.FunctionComponent = () => {
...
         ? (data as PermissionResponse)
         : {permissions: []};
       onRun(runTests, dispatch, parsedPermissions.permissions, source);
+      setHasRun(true);
     } else {
       dispatch(appendSystemMessage("We don't know how to run your code."));
     }
   };
 
+  // We disabled navigation is we are still loading, or if this is a submittable level,
```
ebeastlake: nit: "We disable navigation if"?

ebeastlake (approved):
👍

ebeastlake:
```
There's no explanation on the page as to why submit is disabled when you load the page. I will check with curriculum to see if we want some sort of hover tip to explain this.
```
This isn't one of the things we checked with them on today, is it? It might be worth filing a placeholder to do this before we both disappear on vacation for a week 😊

apps/src/codebridge/ControlButtons/index.tsx
molly-moen: Good call out! We already have a ticket to cover updating progress appropriately: https://codedotorg.atlassian.net/browse/CT-526, I added a note to make sure we don't break submittable levels.

molly-moen:
> ```
> There's no explanation on the page as to why submit is disabled when you load the page. I will check with curriculum to see if we want some sort of hover tip to explain this.
> ```
> 
> This isn't one of the things we checked with them on today, is it? It might be worth filing a placeholder to do this before we both disappear on vacation for a week 😊

@ebeastlake thanks for noticing this! I was going to ask about this but didn't write it on my list 🤦‍♀️ . Just asked and they do want a hover tip, I filed a [task](https://codedotorg.atlassian.net/browse/CT-687)

PR #59746, opened 2024-07-11
Added better error message for save timeout

lfryemason:
Saving a unit seems to frequently hit a 504 timeout error, but still saves correctly. This just makes the error message a bit nicer so that levelbuilders don't panic when they see the error.

## Links
https://codedotorg.atlassian.net/browse/TEACH-1186

## PR Checklist:

- [ ] Code is well-commented


davidsbailey (approved):
Thank you Liam for jumping on this. Nice test!

PR #59754, opened 2024-07-12
Use versioned URLs for two PL units to fix broken links

bethanyaconnor:
It looks like the unversioned URLs go to curriculum that are in development, so we'll need to use the versioned URLs for now. As-is, they send the user to a link that says they don't have permissions.

Slack context: https://codedotorg.slack.com/archives/CFGAVL2CA/p1720795423895819?thread_ts=1720704620.747079&cid=CFGAVL2CA

## PR Checklist:

- [ ] Code is well-commented


kelbyhawn (approved):
Will these go back to the unversioned links once it's live, or are the versioned links the preferred ones?

bethanyaconnor:
The unversioned links don't work here because there's an in development unversioned curriculum. I made a [task](https://codedotorg.atlassian.net/browse/ACQ-2068) to sort that out but figured it could be a next week problem.

bethanyaconnor:
In general, the unversioned links are preferred because they reduce maintenance cost, but there are edge cases, like this one, where they don't work. :/

PR #59755, opened 2024-07-12
Update some versioned curriculum links on code.org

bethanyaconnor:
- The /s/coursea-2023/standards URLs can't use the automatic redirects, so I'm just updating them manually
- I think we want /s/interactive-games-animations over the versioned link. As Moshe pointed out, signed out users are automatically redirected anyway!


kelbyhawn (approved):
🎉

PR #59762, opened 2024-07-12
delete learning goal ai evals and feedbacks when user account is purged

davidsbailey:
Finishes https://codedotorg.atlassian.net/browse/AITT-426. The requirement is to wipe any PII for any user who either deletes their account or explicitly requests data to be wiped, and the easiest way for us to do that is to delete all ai-eval-related objects related to the user, whether they are a teacher or a student. See [slack](https://codedotorg.slack.com/archives/CFTFD6BPV/p1720740468262979) for discussion. 

## Testing story

new unit tests

## Follow-up work

If we add any more database tables,  remember to create a work item to add a similar method to delete them. CC @cearachew for differentiation chat (I've added a note to the tech spec)

### DIFF
--- a/lib/cdo/delete_accounts_helper.rb
+++ b/lib/cdo/delete_accounts_helper.rb
@@ -323,11 +323,17 @@ class DeleteAccountsHelper
   end
 
   def delete_rubric_ai_evaluations(user_id)
+    # This finds all RubricAiEvaluation records which evalute this user's work,
+    # whether they were implicitly requested by the student when they submitted
+    # the project, or explicitly requested by their teacher.
     as_student = RubricAiEvaluation.where(user_id: user_id)
     as_student_count = as_student.count
     as_student.each(&:destroy)
     @log.puts "Deleted #{as_student_count} RubricAiEvaluation as student" if as_student_count > 0
 
+    # We've already deleted all evaluations of this user's work as a student.
+    # Any remaining evaluations requested by this user are requests made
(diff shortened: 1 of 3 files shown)


lib/cdo/delete_accounts_helper.rb
```diff
@@ -322,6 +322,30 @@ def delete_ai_tutor_interactions(user_id)
     @log.puts "Deleted #{count} AI Tutor Interactions" if count > 0
   end
 
+  def delete_rubric_ai_evaluations(user_id)
+    as_student = RubricAiEvaluation.where(user_id: user_id)
+    as_student_count = as_student.count
+    as_student.each(&:destroy)
+    @log.puts "Deleted #{as_student_count} RubricAiEvaluation as student" if as_student_count > 0
+
+    as_teacher = RubricAiEvaluation.where(requester_id: user_id)
```
wilkie, line 337: technically, the requester is the student when it happens via the Submit button in a level, but this still makes sense to do.

davidsbailey: Good point! I've updated the comments to give more context about what user and requester mean. Ready for review!

davidsbailey:
@Nokondi @cearachew heads up that we need to do this for any database fields we add that might contain PII. I'm not really sure how this applies to differentiation since it is unclear to me if we think that whatever teachers type there would count as PII.

wilkie (approved):

PR #59763, opened 2024-07-12
Enable Manual User Session Expiration

Hamms:
Add the ability for users to log themselves out of all other browser sessions.

Specifically, we extend Devise with a new `expire_all_sessions!` method, which extends the existing logic to invalidate sessions after a password change to also invalidate them based on an arbitrary token. We then add a new controller route `/expire_other` which will invalidate all existing sessions and create a new session in the current browser, the result being that they will be logged out of every browser other than the one from which they initiated the operation. Finally, we add a "Manage Other Sessions" section to the Edit Account Details page with a button which will send a request to that route:

image

I've included a simple DCDO flag we can flip to disable this section in the UI if we need to for any reason, with the intent that we will remove the flag once the feature has been stable in production for long enough.

Very open to any and all feedback, particularly regarding the copy and the visual presentation of the form in general.

## Links

Based on / Inspired by https://makandracards.com/makandra/53562-devise-invalidating-all-sessions-for-a-user

## Testing story

Tested manually on multiple devices; also added some basic unit testing. I experimented with adding some more advanced integration testing, but ultimately decided it was more trouble than it was worth. Happy to revisit that experiment if folks think more robust testing would be valuable here.

## PR Checklist:

- [ ] Code is well-commented


Hamms:
@code-dot-org/i18n Would y'all prefer that we flip the DCDO flag to hide this feature from users until we've had the chance to collect some translations for the new strings?

daynew (approved):
Thank you for providing the links explaining how Devise works with Sessions and salts. This PR makes a lot of sense and was easy to understand.

One initial concern I had was that existing users would be logged out. But after reading the provides URLs and studying the code, I'm confident existing users will not be affected by this.

daynew:
> @code-dot-org/i18n Would y'all prefer that we flip the DCDO flag to hide this feature from users until we've had the chance to collect some translations for the new strings?

I think it makes sense to get at least a few of our top languages first. You should probably reach out in the #international channel to coordinate.

dashboard/app/views/devise/registrations/edit.html.haml
```diff
@@ -164,6 +164,15 @@
...
       %button.btn#change-user-type-button{disabled: true}= t('signup_form.user_type_button')
 
+- unless DCDO.get('disable-expire-other-sessions', false) # TODO: remove flag once feature is stable
+  %div#expire-other-sessions
+    %hr
+    %h2= t('user.expire_other_sessions')
+    %p= t('user.expire_other_sessions_description')
+    = form_for(resource, as: resource_name, url: expire_other_path, html: {method: :delete}) do |expire_other_form|
+      .farSide
+        = expire_other_form.submit t('user.expire_other_sessions_button'), class: 'btn btn-warning'
```
daynew, line 174: Have you tested what happens if a Code.org employee using assume identity clicks this button? We probably want to know what happens if our CX team clicks this button when trying to help students/teachers.

Hamms:
> One initial concern I had was that existing users would be logged out

Oh shoot, good call out! The original implementation I linked to will indeed preserve existing session, but it actually turns out that the little `::` separator I added to my implementation would indeed invalidate existing sessions. [I removed the separator](https://github.com/code-dot-org/code-dot-org/pull/59763/commits/b8f6baade3acf487a65f841d4dc6cea07678982c) and validated that signing in on the staging branch then switching to this feature branch no longer logs me out. Thanks!

dashboard/app/views/devise/registrations/edit.html.haml
Hamms: I had not, good idea! This wasn't a use case I had in mind, and it reveals some potentially-undesirable behavior.

An admin assuming identity as a user can indeed navigate to and click this button. When they do so, the user account whose identity is being assumed will be updated with a new session expiration token, resulting in the user being signed out of all sessions as expected:
image

Unfortunately, because we then [sign the current user account back in](https://github.com/code-dot-org/code-dot-org/blob/856e0a5c80aee917ca51eb69e2ed30d352d6b6e7/dashboard/app/controllers/sessions_controller.rb#L59) in order to preserve the session from which they initiated the expiration, the admin will no longer be assuming identity and will just be directly signed in to the user's account. Note the disappearance of the red "assuming identity" banner:
image

I'll see if I can figure out a better way to preserve the existing session

Hamms: This actually highlighted a deeper underlying issue, which is that the "expire all sessions and sign the user back in"  approach resulted in the loss of all existing miscellaneous session data, including not only assumed identity but also things like callouts seen, pairings, etc.

[I added some logic to copy all that over](https://github.com/code-dot-org/code-dot-org/pull/59763/commits/2ba44a721820aa30ba885b60189bf37778b0768c), and verified that with this change in place, an admin assuming identity as a user will still be in "assumed identity" mode after expiring that user's sessions.

PR #59766, opened 2024-07-12
Move Devise Customization into Subdirectory

Hamms:
In preparation for some upcoming work-in-progress changes to [enable automatic lockout for teachers](https://github.com/code-dot-org/code-dot-org/pull/57515) and [allow users to log out of all existing sessions](https://github.com/code-dot-org/code-dot-org/pull/59763), we'd like to move our existing customization to Devise functionality into a subdirectory and associated Module so that the upcoming functionality can be organized into the same place.

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/dashboard/config/initializers/devise.rb
+++ b/dashboard/config/initializers/devise.rb
@@ -337,9 +337,9 @@ Devise.setup do |config|
   #   manager.intercept_401 = false
   #   manager.default_strategies(:scope => :user).unshift :some_external_strategy
   # end
-  require 'custom_devise_failure'
+  require 'devise/custom_failure'
   config.warden do |manager|
-    manager.failure_app = CustomDeviseFailure
+    manager.failure_app = Devise::CustomFailure
   end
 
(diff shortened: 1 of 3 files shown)


sureshc:
Are there any tests that exercise the existing logic?

sureshc (approved):
Might be worth having someone in Platform Team take a glance?

Hamms:
> Are there any tests that exercise the existing logic?

None that I could find; I tested manually, but might check to see if it would be quick to add one

Hamms:
I checked; I'm guessing the reason there isn't already a test for this is that it looks like it would actually be quite a pain to add one. @code-dot-org/platform, what do you all think; is it worth adding a test for this functionality or do you think it's already adequately covered by existing tests?

carl-codeorg (approved):
The code being moved is pretty small, and has existed for 10 years. If it's been around that long without specific unit tests, we're probably fine.

PR #59767, opened 2024-07-12
AI Tutor: add systemPrompt to AI Tutor Tester

Erin007:
[CT-686](https://codedotorg.atlassian.net/browse/CT-686)
Follow up to https://github.com/code-dot-org/code-dot-org/pull/59638

There are 3 related updates here: 

1.) I renamed `AITutorBulkResponseGenerator` to `AITutorTester` for the sake of brevity 

2.) I built a little table that explains what columns can be uploaded in the csv 
Screenshot 2024-07-12 at 3 19 56 PM

3.) `systemPrompt` is now a supported column and if it's populated in the uploaded data, we'll override the default system prompt which will allow us to more easily iterate on system prompts. 

**Follow Ups:** 
- Next up I'm going to add the ability for levelbuilders to add `levelId` to the uploaded data which will enable us to fetch level instructions and test files for specific levels in the context of the Tester
- Ultimately, I think we'll want to split out and concatenate different parts of the system prompt so each can be customized. Something like: 
`system_prompt = base_student_facing_prompt + lab(or course?)_specific_prompt + level_specific_prompt` etc but let's see how far this gets us for now 😅  

--------

- [ ] Code is well-commented


apps/src/aiTutor/redux/aiTutorRedux.ts
```diff
@@ -84,6 +84,7 @@ export const askAITutor = createAsyncThunk(
     const chatApiResponse = await getChatCompletionMessage(
       formattedQuestion,
       storedMessages,
+      '',
```
molly-moen, line 87: should this be undefined rather than an empty string?

molly-moen (approved):
Looks good!

PR #59770, opened 2024-07-12
Show summary of student responses for predict levels

molly-moen:
On our old contained levels we show a little summary of the number of student responses to teachers in the instructions panel, if they are viewing a specific section. This reproduces this behavior for the new predict levels. Previously this information was set by the server on load, so I added a new api that provides this summary.

I also made the "view student responses" button a little smaller since this ended up taking up two lines in the instructions panel.

After this is in, we should have parity with the old contained levels with one exception! That is telling students if their multiple choice answer was right or wrong. I discussed this with curriculum and we don't want this feature for new labs, because students are meant to make a prediction, then discuss with the class, so having that immediate feedback isn't useful.

I plan to investigate making predict levels editable after students make a prediction after this, which is a potential new feature.

### Contained level version
Screenshot 2024-07-12 at 2 49 08 PM

### New version
Screenshot 2024-07-12 at 3 17 28 PM

If we make the instructions panel wider it will fit on one line
Screenshot 2024-07-12 at 3 16 54 PM

## Links

- jira ticket: [CT-678](https://codedotorg.atlassian.net/browse/CT-678)

## Testing story
Tested locally and added unit test for the new api.

## PR Checklist:

- [ ] Code is well-commented


apps/src/lab2/views/components/PredictSummary.tsx
```diff
@@ -17,18 +21,53 @@ const PredictSummary: React.FunctionComponent = () => {
     `viewAs=${ViewType.Instructor}`
   );
   const summaryUrl = document.location.pathname + SUMMARY_PATH + params;
+  const currentSectionId = useSelector(
+    (state: {teacherSections: {selectedSectionId: number}}) =>
+      state.teacherSections.selectedSectionId
+  );
+  const currentLevelId = useAppSelector(state => getCurrentLevel(state)?.id);
+  const [responseCount, setResponseCount] = React.useState<number | null>(null);
+  const [numStudents, setNumStudents] = React.useState<number | null>(null);
```
sanchitmalhotra126: super minor - but just for consistency, separate import for `useState` since you're already importing `useEffect`?

apps/src/lab2/views/components/PredictSummary.tsx
```diff
@@ -17,18 +21,53 @@ const PredictSummary: React.FunctionComponent = () => {
     `viewAs=${ViewType.Instructor}`
   );
   const summaryUrl = document.location.pathname + SUMMARY_PATH + params;
+  const currentSectionId = useSelector(
+    (state: {teacherSections: {selectedSectionId: number}}) =>
+      state.teacherSections.selectedSectionId
+  );
+  const currentLevelId = useAppSelector(state => getCurrentLevel(state)?.id);
```
sanchitmalhotra126: any reason not to use `state.progress.currentLevelId`?

apps/src/lab2/views/components/PredictSummary.tsx
```diff
@@ -17,18 +21,53 @@ const PredictSummary: React.FunctionComponent = () => {
...
+  const [responseCount, setResponseCount] = React.useState<number | null>(null);
+  const [numStudents, setNumStudents] = React.useState<number | null>(null);
+
+  useEffect(() => {
+    if (currentSectionId && currentLevelId) {
+      getSectionSummary(currentSectionId, currentLevelId).then(response => {
+        if (response.response.ok) {
```
sanchitmalhotra126: `HttpClient` which I think this function is using should throw if the response is not `ok`. So I think you'd need to add a catch here if you want to handle the error case (or try/catch if using async/await).

apps/src/lab2/views/components/predict-summary.module.scss
```diff
@@ -0,0 +1,35 @@
...
+}
+
+.responses {
+  display: flex;
+}
+
+.responseIcon {
+  padding-right: 4px;
+
+  :global(.fa) {
```
sanchitmalhotra126, line 17: for my understanding - what does this do?

apps/src/lab2/views/components/PredictSummary.tsx
molly-moen: yep that was just a miss on my part!

apps/src/lab2/views/components/PredictSummary.tsx
molly-moen: good catch!

apps/src/lab2/views/components/predict-summary.module.scss
molly-moen: this applies styling to the `fa` class specifically, rather than turning it into a css module class (which adds extra stuff to the class name so it is unique). We need this to style the font awesome icon. Since it's inside a css module-ized class it should be safe to do. We are doing something similar in the [old version](https://github.com/code-dot-org/code-dot-org/blob/0bf7d37763005917509eeac86abc6d788a09170a/apps/src/templates/levelSummary/summary-entry-point.module.scss#L40) of this (which is how I learned to do it)

apps/src/lab2/views/components/PredictSummary.tsx
molly-moen: also just a miss on my part, updating!

apps/src/lab2/views/components/PredictSummary.tsx
```diff
@@ -17,18 +20,56 @@ const PredictSummary: React.FunctionComponent = () => {
...
+  const currentSectionId = useSelector(
+    (state: {teacherSections: {selectedSectionId: number}}) =>
+      state.teacherSections.selectedSectionId
+  );
+  const currentLevelId = useAppSelector(state => state.progress.currentLevelId);
+  const [responseCount, setResponseCount] = useState<number | null>(null);
+  const [numStudents, setNumStudents] = useState<number | null>(null);
+
+  useEffect(() => {
+    if (currentSectionId && currentLevelId) {
```
sanchitmalhotra126, line 32: I think you would still want a `catch` here since `getSectionSummary` would throw any non-`NetworkError` right? Which I guess would be rare, but it could be like a `TypeError` if something goes wrong on the client side?

sanchitmalhotra126 (approved):
One comment/clarification but otherwise LGTM!

apps/src/lab2/views/components/PredictSummary.tsx
molly-moen: Good call, will update

PR #59779, opened 2024-07-15
Add more folders to eslint import order rule

lfryemason:

## PR Checklist:

- [ ] Code is well-commented


apps/src/p5lab/P5Lab.js
```diff
@@ -2,66 +2,82 @@ import $ from 'jquery';
...
+var dropletConfig = require('./gamelab/dropletConfig');
+var p5GroupWrapper = require('./P5GroupWrapper');
 var P5LabView = require('./P5LabView');
+var p5SpriteWrapper = require('./P5SpriteWrapper');
+var P5Wrapper = require('./P5Wrapper');
+var reducers = require('./reducers');
+
+// Disabling import order in order to maintain require statements placement
+// Require statements can change behavior based on the order they are called.
+// This might be safe to remove but needs investigation whether any behavior is changed by order.
```
Hamms: Please update this comment to be more specific to this particular file. Why has this particular set of imports been preserved in their original order when so many others have been updated?

apps/src/p5lab/P5Wrapper.js
```diff
@@ -1,12 +1,20 @@
+import p5 from '@code-dot-org/p5';
+
 import {getStore} from '@cdo/apps/redux';
+
+// Disabling import order in order to maintain p5 import first.
+// This might be safe to remove but needs investigation whether any behavior is changed by order.
```
Hamms, line 6: Why is the p5 import in particular necessary to come first?

apps/src/sites/studio/pages/code-studio.js
```diff
@@ -9,24 +9,34 @@
...
+// disable import/order rule to import consoleShim after setting store.
+// This might be safe to remove but needs investigation whether any behavior is changed by order.
+/* eslint-disable import/order*/
 import getScriptData from '@cdo/apps/util/getScriptData';
 
 const store = getStore();
 store.dispatch(setRtlFromDOM());
-
 // Shim window.console to be safe in IE
 require('@cdo/apps/code-studio/consoleShim')(window);
```
Hamms, line 33: We actually might not need this shim at all; we officially no longer support IE

apps/src/sites/studio/pages/init/loadApplab.js
```diff
@@ -1,3 +1,6 @@
+// Disabling import order in order to maintain window setting order
```
Hamms, line 1: :+1:

apps/src/sites/studio/pages/init/loadCraft.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable import/order */
```
Hamms, line 2: Please add an explanatory comment

Hamms (approved):
It's exciting to be so close to the end of this one! Great job of keeping at it.

I have a couple of comments and questions about some manual rule disablings, but the vast majority of the changes look good to me! I'm happy to see the comments dealt with in a follow-up PR, too, since they represent such a statistical minority of changes and we're going to have some merge conflicts to navigate.

apps/src/p5lab/P5Wrapper.js
lfryemason: `This might be safe to remove but needs investigation whether any behavior is changed by order.`

With such a big PR I want to minimize the number of possible unintended side effect and changing when window.p5 is defined could be a side effect.

apps/src/p5lab/P5Lab.js
lfryemason: That's a good question, I'm not sure why it didn't complain about the imports it did move. I disabled more of the file, specifically all of the imports after the first require.

apps/src/sites/studio/pages/code-studio.js
lfryemason: This should be in a separate PR if we want to remove it.

apps/src/p5lab/P5Lab.js
Hamms: Can you clarify in the comment what exactly is complaining and how exactly it's complaining? It's still not at all clear to me why we aren't just fixing the order of the imports here like we fixed the order of the requires.

Hamms: If the issue is just that the auto-correction logic gets tripped up a bit (which I have encountered), it's usually pretty straightforward to apply the organization manually

apps/src/p5lab/P5Wrapper.js
Hamms: Ahh, that makes sense. Could you update the comment to highlight the `window.p5` assignment rather than just the importing of p5?

apps/src/sites/studio/pages/code-studio.js
Hamms: Yes, agreed. It would probably be cleanest to remove it in a prerequisite PR so this one could be applied cleanly, but I think taking care of it in a fast follow-up should be just fine.

PR #59780, opened 2024-07-15
AI Tutor: use levelId in AITutorTester

Erin007:
[CT-689](https://codedotorg.atlassian.net/browse/CT-689)
[CT-682](https://codedotorg.atlassian.net/browse/CT-682)
[CT-690](https://codedotorg.atlassian.net/browse/CT-690)

Continuation of https://github.com/code-dot-org/code-dot-org/pull/59767

This work allows the user to include `levelId` as a column in the AI Tutor Tester uploaded csv. The backend will then use that `levelId` to fetch the level's instructions and test files where applicable. As a result we no longer need to pass around `levelInstructions` on the frontend because we fetch them from the backend now. 

Along the way I tackled a couple of other issues: 
- `actionType` (generalChat, validation or compilation) isn't relevant to AI. The question and system prompt dictate the AI's response. I decided to no longer pass around `actionType` in the request for the AI response. 
- Relatedly, even in general chat scenarios, the AI often needs access to the student's code and test files (if they are available) to be able to answer effectively so we default to sending those along in all scenarios.

## Testing story
I verified locally that I can still get AI responses for general chat, validation and compilation scenarios within the context of a level locally. 

Here's the flow from the csv upload including `levelId` using a small sample of [validation prompts](https://docs.google.com/spreadsheets/d/1BeUT3u_mMFdEz7fOFvNWI0ukw5WcX9OGIn2S3YOYjh8/edit?gid=81872701#gid=81872701).

https://github.com/user-attachments/assets/a62f0970-3e3f-40de-bd6e-e41fb6dee410

## Follow-up work
- The `actionType` distinctions are somewhat artificial on the backend since a user could (and often does) ask a compilation or test-related question in the "general chat" scenario. `actionType` is still currently used for suggested prompts and analytics, so I kept it around in UI components and logging for now; but, we may want to reconsider this as we refine AI Tutor use cases.

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/dashboard/app/controllers/openai_chat_controller.rb
+++ b/dashboard/app/controllers/openai_chat_controller.rb
@@ -116,14 +116,16 @@ class OpenaiChatController < ApplicationController
       return render(status: :bad_request, json: {message: "Couldn't find level with id=#{level_id}."})
     end
 
-    if level.validation.values.empty?
-      return render(status: :bad_request, json: {message: "There are no test files associated with level id=#{level_id}."})
-    else
-      test_file_contents = ""
-      level.validation.each_value do |validation|
-        test_file_contents += validation["text"]
+    test_file_contents = ""
+    if !!level.validation
+      if level.validation.values.empty?
+        return render(status: :bad_request, json: {message: "There are no test files associated with level id=#{level_id}."})
+      else
+        level.validation.each_value do |validation|
+          test_file_contents += validation["text"]
+        end
       end
-      return test_file_contents
     end
+    return test_file_contents
   end
 end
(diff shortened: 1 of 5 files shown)


dashboard/app/controllers/openai_chat_controller.rb
```diff
@@ -22,18 +22,20 @@ def chat_completion
     # If the content is profane, we skip sending to OpenAI and instead hardcode a warning response on the front-end.
     return render(status: :ok, json: {safety_status: filter_result.type, flagged_content: filter_result.content}) if filter_result && filter_result.type == 'profanity'
 
+    # Fetch the level-specific context: instructions and test files if applicable
+    level_id = params[:levelId]
+
+    level_instructions = ""
+    level_instructions = get_level_instructions(level_id) if !!level_id
```
bencodeorg, line 29: Nit: I think the !! are redundant when used in if statements

bencodeorg (approved):
LGTM!

dashboard/app/controllers/openai_chat_controller.rb
Erin007: you're totally right! thanks! I'm shuffling things around in a follow up; I'll address there.

PR #59781, opened 2024-07-15
Fix can't delete remixed featured project bug

fisher-alice:
This PR fixes a bug that was reported by a user in this [ZenDesk ticket](https://codeorg.zendesk.com/agent/tickets/499844) - a user remixed a featured project from the recently updated public project gallery, but was unable to delete it. In addition, a remixed project should be editable but it was read-only.

When a project is remixed, the source project's data is retrieved in `channel_token.rb`:
https://github.com/code-dot-org/code-dot-org/blob/206b563078f3c3f8d817218f808b815c73e3428a/dashboard/app/models/channel_token.rb#L88-L93

Because the keys from the project's source data are symbols and not strings, when the data is merged with the remixed data values, both key-value pairs are included in the data. When the project is created and `value` is assigned, the symbol's values are used.

I came up with 2 different approach which both resolve this issue, but decided to go with the 2nd approach:
~~1. Stringify the project keys: https://github.com/code-dot-org/code-dot-org/pull/59781/commits/9d7540a56a971ab4d2e8c17ef4c5a246f2b9b560~~
2. Use symbols for the updated data values. https://github.com/code-dot-org/code-dot-org/pull/59781/commits/d1a0a8cab7097acf27b50109a64eb9cf176daa39

~~I am unsure which approach is preferred but I am leaning towards the 2nd approach. Open to discussion!~~

After discussion, opted to symbolize keys.
~~and also updated `merged_row_value` in `projects.rb` to symbolize keys so that duplidate keys no longer are returned.~~ UPDATE: did not update `merged_row_value` this had impacts outside of the scope of this PR.

## Before update

Screenshot of remixed featured project in personal projects gallery- Delete option is unavailable.
Screenshot 2024-07-15 at 2 37 00 PM

Screenshot of remixed featured project - project is view-only.
Screenshot 2024-07-15 at 2 37 14 PM

## After update

Screencast of user remixing lab2 project and then able to delete. And then remixing a legacy project and deleting.

https://github.com/user-attachments/assets/4d3ade95-7a74-4ef2-9eb5-5f7e5c82e5bc

## Links
[jira](https://codedotorg.atlassian.net/browse/LABS-898)

## Testing story
Testing locally using both a teacher and student account with both legacy and lab2 type featured projects.
I also created new legacy/lab2 projects and tested that I could export an App Lab project successfully.

## PR Checklist:

- [ ] Code is well-commented


dashboard/app/models/channel_token.rb
```diff
@@ -87,14 +87,12 @@ def self.find_channel_token(level, user_storage_id, script_id)
   def self.create_channel(ip, project, data: {}, src: nil, type: nil, remix_parent_id: nil, standalone: true, level: nil)
     if src
       data = project.get(src)
-      data['name'] = "Remix: #{data['name']}"
-      data['hidden'] = false
-      data['frozen'] = false
+      data.merge!(name: "Remix: #{data['name']}", hidden: false, frozen: false)
```
cnbrenci, line 90: I would lean towards the approach of consistency with keeping the keys as symbols, if that's what's coming out of `project.get(src)` 

How is `#{data['name']}` working here? Won't that return `nil` if the data contains :name (not 'name')?

```ruby
[development] dashboard > data
=> {:name=>"blah"}
[development] dashboard > data['name']
=> nil
```

cnbrenci:
This is a really sneaky bug. Great sleuthing! I was wondering if there might be other place in the code where this same bug exists and it looks like there might be one in projects_controller.rb around like 533 where we're getting the project data the same way, and then overwriting some of the fields using string keys. I wonder if that's broken too?

Not a blocker for this PR, but just wanted to mention that to potentially log separately.

dashboard/app/models/channel_token.rb
```diff
@@ -87,14 +87,12 @@ def self.find_channel_token(level, user_storage_id, script_id)
...
     end
 
     timestamp = Time.now
     project.create(
-      data.merge('createdAt' => timestamp, 'updatedAt' => timestamp),
+      data.merge(createdAt: timestamp, updatedAt: timestamp),
```
cnbrenci, line 95: If data doesn't get set in the `if src` block above, then we'd be merging with whatever `data` object was passed in. Could that one have string keys? Would that be OK in this case because create() favors symbols?

dashboard/app/models/channel_token.rb
cnbrenci: (I mean obviously it works because you demo'd it and it did, but.. how tho?)

fisher-alice: > I would lean towards the approach of consistency with keeping the keys as symbols, if that's what's coming out of `project.get(src)`

Yes, that sounds good to me too - thanks!

> How is #{data['name']} working here? Won't that return nil if the data contains :name (not 'name')?

Great question! So `Project.merged_row_value` is defined as:
https://github.com/code-dot-org/code-dot-org/blob/9b4270890b89017ca99b58abff7732707ec4eb3b/dashboard/legacy/middleware/helpers/projects.rb#L361-L376

so that the `value` JSON is merged with the other values from the row data as well as a couple other values. But the keys are stored as strings! Therefore, the project `name` is able to be accessed as `data['name']`.

I think I'll go ahead and symbolize the keys of `values` so there aren't duplicate key names.

dashboard/app/models/channel_token.rb
fisher-alice: It looks like `ChannelToken.create_channel` is called on when a project is remixed, when a new project [is created](https://github.com/code-dot-org/code-dot-org/blob/d42c6bf7f87863f1c24374e04f42fac3061ea1aa/dashboard/app/controllers/projects_controller.rb#L302), or when a project [is exported](https://github.com/code-dot-org/code-dot-org/blob/d42c6bf7f87863f1c24374e04f42fac3061ea1aa/dashboard/app/controllers/projects_controller.rb#L523). For the last two cases, `data` is assigned `initial_data` defined [here](https://github.com/code-dot-org/code-dot-org/blob/d42c6bf7f87863f1c24374e04f42fac3061ea1aa/dashboard/app/controllers/projects_controller.rb#L612) and the keys are symbols.

fisher-alice:
> This is a really sneaky bug. Great sleuthing! I was wondering if there might be other place in the code where this same bug exists and it looks like there might be one in projects_controller.rb around like 533 where we're getting the project data the same way, and then overwriting some of the fields using string keys. I wonder if that's broken too?
> 
> Not a blocker for this PR, but just wanted to mention that to potentially log separately.

Good call! I'll update here as well since I symbolized the keys in `Projects.get`. Interestingly, the `export_create_channel` doesn't seem to be called when an App Lab project is exported. In the code base, the endpoint is referenced in `ExporterTest.js`, but when I exported an App Lab locally, this function was not called. I did test that the values are accessed as expected with the URL `http://localhost-studio.code.org:3000/projects/applab/<channel-id>/export_create_channel`.

UPDATE: did not make change here to limit scope of this PR.

dashboard/app/models/channel_token.rb
cnbrenci: Ahh, sweet! Thank you for checking that and the thorough explanation!

dashboard/app/models/channel_token.rb
cnbrenci: Ohhh, I see now. Yeah, I think symbolizing the keys when they get json parsed makes it a lot easier to understand.

cnbrenci: Super helpful explanation, thank you!

cnbrenci (approved):

dashboard/legacy/middleware/helpers/projects.rb
```diff
@@ -362,7 +362,7 @@ def most_recent_project_type(type)
   # 'createdAt', 'updatedAt', 'publishedAt' and 'projectType' merged from the
   # corresponding database row values.
   def self.merged_row_value(row, channel_id:, is_owner:)
-    JSON.parse(row[:value]).merge(
+    JSON.parse(row[:value], symbolize_names: true).merge(
```
fisher-alice: There are 2 unit tests that are failing in `test_projects.rb`. I can fix them by:
- accessing thumbnailUrl with symbol in `test_project_create_succeeds_with_valid_thumbnail_url`
- for `test_derive_project_type` I need to change the order of the two hashes that are being merged in merged_row_value in projects.rb. https://github.com/code-dot-org/code-dot-org/blob/ec6d13f6bd3c5e63569074303dc9d04260fad2a2/dashboard/legacy/middleware/helpers/projects.rb#L364

Making further changes in a legacy file is making me nervous so I'm going to revert PR to draft and discuss a bit further with folks..

dashboard/app/models/channel_token.rb
fisher-alice, line 95: This change allows the `createdAt` and `updatedAt` values to be updated with `Time.now` if project is remixed (so that the values are not those from the original/src project).
Before this change, both the string and symbole keys existed in `data` and the symbol keys have precedence

dashboard/legacy/middleware/helpers/projects.rb
fisher-alice: Discussed with @cnbrenci and we agreed to keep changes as minimal as possible.

cnbrenci (approved):
🥳

PR #59782, opened 2024-07-15
P20-876: Use Google Noto for fallback fonts

artem-vavilov:
Code.org uses the Metropolis font for rendering extended Latin characters. However, users who access the website in non-Latin extended languages (e.g., Farsi, Arabic, Chinese, Korean, etc.) face inconsistent font rendering. In some cases, users' browsers would default to system fonts, leading to varied experiences or, worse, question marks or empty boxes if the font was unavailable.

To resolve this issue and provide a uniform experience, Code.org now employs Google Noto as a fallback font. The system is designed to prioritize Metropolis, and if Metropolis cannot render a character, it seamlessly falls back to Noto. This ensures that all characters are rendered consistently, regardless of the language, thus enhancing the overall user experience for Code.org visitors worldwide.

## Screenshots
The Arabic-specific character `؟` is replaced by the corresponding character from the `Noto Sans Arabic` font.
| Metropolis | Metropolis + Noto | Noto |
| ---------- | ------------------ | ----- |
| metropolis | metropolis+noto | noto |

## Links
- JIRA task: [P20-876](https://codedotorg.atlassian.net/browse/P20-876)


levadadenys (approved):
LGTM! 😍 Thanks for this update!

cc @breville and @sanchitmalhotra126 so that you'll also have a context on this

wilkie:
This more or less looks good. I'm curious about the mechanism. If 'Noto Sans' is listed first in the order, as it seems to be here, how do we know it will effectively fallback to the regional glyphs if there isn't an explicit glyph range for the region specific ones? It's surprisingly difficult to find information about this. Did you find a good source on how to use the regional fonts as fallbacks?

artem-vavilov:
> This more or less looks good. I'm curious about the mechanism. If 'Noto Sans' is listed first in the order, as it seems to be here, how do we know it will effectively fallback to the regional glyphs if there isn't an explicit glyph range for the region specific ones? It's surprisingly difficult to find information about this. Did you find a good source on how to use the regional fonts as fallbacks?

When a specific character (e.g., special symbols or certain Unicode characters) is missing from the primary font (Metropolis), the browser will replace only that missing character with the next available font in the fallback list.

In the screenshots, you can see how the Arabic-specific character `؟` is replaced by the corresponding character from the `Noto Sans Arabic` font.

shared/css/font.scss
```diff
@@ -6,9 +6,13 @@
...
 
-$metropolis-font: 'Metropolis', sans-serif;
+$metropolis-font: 'Metropolis';
+$noto-sans-fonts: 'Noto Sans', 'Noto Sans Arabic', 'Noto Sans Armenian', 'Noto Sans Bengali', 'Noto Sans SC',
+                  'Noto Sans TC', 'Noto Sans Devanagari', 'Noto Sans Georgian', 'Noto Sans Hebrew', 'Noto Sans JP',
+                  'Noto Sans Kannada', 'Noto Sans Khmer', 'Noto Sans KR', 'Noto Sans Myanmar', 'Noto Sans Sinhala',
+                  'Noto Sans Tamil', 'Noto Sans Telugu', 'Noto Sans Thai', 'Noto Sans Thaana';
 
-$main-font: $metropolis-font;
+$main-font: $metropolis-font, $noto-sans-fonts, sans-serif;
```
artem-vavilov, line 15: The script [tools/scripts/convertScssToJs.js](https://github.com/code-dot-org/code-dot-org/pull/59782/files#diff-693ecc71cca8b8c59b3b731cb0a295f1a4bfd8ca95571b9e059812a19deac452) required some adjustments to properly parse newline-split lines and combine multiple CSS variables into one.
## Generated `apps/src/fontConstants.js`

### Before
```js
module.exports = {
  "metropolis-font": "'Metropolis'",
  "main-font": "$metropolis-font, $noto-sans-fonts, sans-serif",
};
```

### After
```js
module.exports = {
  "metropolis-font": "'Metropolis'",
  "noto-sans-fonts": "'Noto Sans', 'Noto Sans Arabic', 'Noto Sans Armenian', 'Noto Sans Bengali', 'Noto Sans SC', 'Noto Sans TC', 'Noto Sans Devanagari', 'Noto Sans Georgian', 'Noto Sans Hebrew', 'Noto Sans JP', 'Noto Sans Kannada', 'Noto Sans Khmer', 'Noto Sans KR', 'Noto Sans Myanmar', 'Noto Sans Sinhala', 'Noto Sans Tamil', 'Noto Sans Telugu', 'Noto Sans Thai', 'Noto Sans Thaana'",
  "main-font": "'Metropolis', 'Noto Sans', 'Noto Sans Arabic', 'Noto Sans Armenian', 'Noto Sans Bengali', 'Noto Sans SC', 'Noto Sans TC', 'Noto Sans Devanagari', 'Noto Sans Georgian', 'Noto Sans Hebrew', 'Noto Sans JP', 'Noto Sans Kannada', 'Noto Sans Khmer', 'Noto Sans KR', 'Noto Sans Myanmar', 'Noto Sans Sinhala', 'Noto Sans Tamil', 'Noto Sans Telugu', 'Noto Sans Thai', 'Noto Sans Thaana', sans-serif",
};
```

breville:
Does this also work for Pegasus pages (i.e. https://code.org)?

shared/css/font.scss
```diff
@@ -6,9 +6,13 @@
 // Try to avoid use of inline styles if you're creating new component).
 // Keeping that in mind -
 
-$metropolis-font: 'Metropolis', sans-serif;
+$metropolis-font: 'Metropolis';
+$noto-sans-fonts: 'Noto Sans', 'Noto Sans Arabic', 'Noto Sans Armenian', 'Noto Sans Bengali', 'Noto Sans SC',
```
stephenliang, line 10: Does this font family list include all of the languages supported by Code.org?

shared/css/fonts/noto-sans.scss
```diff
@@ -0,0 +1,75 @@
...
+ * 800 – Extra Bold (Ultra Bold)
+ * 900 – Black (Heavy)
+ */
+
+ // Latin, Cyrillic and Greek:
+ // az-AZ, bs-BA, ca-ES, co-FR, cs-CZ, da-DK, de-DE, el-GR, en-GB, en-US, es-ES, es-MX, et-EE, eu-ES, fi-FI, fil-PH,
+ // fr-FR, ga-IE, gl-ES, haw-HI, hr-HR, hu-HU, id-ID, is-IS, it-IT, kk-KZ, ky-KG, lt-LT, lv-LV, mi-NZ, mk-MK, ms-MY,
+ // mt-MT, nl-NL, nn-NO, no-NO, pl-PL, pt-BR, pt-PT, ro-RO, ru-RU, se-FI, sk-SK, sl-SI, sm-WS, sq-AL, sr-SP, sv-SE,
+ // tg-TJ, tr-TR, uk-UA, uz-UZ, vi-VN, zu-ZA
+@import url('https://fonts.googleapis.com/css2?family=Noto+Sans:ital,wdth,wght@0,62.5..100,100..900;1,62.5..100,100..900&display=swap');
```
stephenliang, line 21: Would these imports create a network request to Google to fetch the fonts or would these be served by Code.org? If it does go to Google we should monitor Zendesk and eyes for font download failures

shared/css/font.scss
artem-vavilov: Correct, this list covers all the supported languages by Code.org.
I also added comments about which font supports which language: https://github.com/code-dot-org/code-dot-org/blob/128bdd7b39ddcc8eefec4257cf3c0412f55a521a/shared/css/fonts/noto-sans.scss#L16-L21

shared/css/fonts/noto-sans.scss
stephenliang: Looking at metropolis, we base64 encode the font so we self-deliver whereas in this one it pulls from the Google Font CDN. This is probably fine to start with, but we may need to look into self-serving them due to school network restrictions

artem-vavilov: The font import is served by Google Fonts servers, not Code.org. Therefore, it will not impact Code.org's network load.

artem-vavilov: The font import is served by Google Fonts servers, not Code.org. Therefore, it will not impact Code.org's network load.

artem-vavilov:
> Does this also work for Pegasus pages (i.e. https://code.org/)?

@breville https://github.com/code-dot-org/code-dot-org/pull/59782/commits/1e46a1e1d3a7315d6a7804805d451a544431725c

shared/css/fonts/noto-sans.scss
artem-vavilov: cc @code-dot-org/infrastructure

stephenliang: My comment was mostly regarding how schools sometimes block certain domains which would result in the font being blocked. However, I think the net result would be better (a blocked font would result in the same behavior we have today) so I think this is good and we'll see how it plays out.

stephenliang (approved):

wilkie (approved):

sanchitmalhotra126 (approved):

shared/css/fonts/noto-sans.scss
breville: +1 to looking into whether we can self-host these files, which has always been a best practice for us, due to difficulty in some environments.

breville (approved):

shared/css/fonts/noto-sans.scss
artem-vavilov: @code-dot-org/infrastructure Should we self-host these fallback fonts, or is it okay to source them from Google?
As Stephen mentioned, the worst-case scenario is that a blocked font would result in the same behavior we experience today.

artem-vavilov: @code-dot-org/infrastructure Should we self-host these fallback fonts, or is it okay to source them from Google? As Stephen mentioned, the worst-case scenario is that a blocked font would result in the same behavior we experience today.

artem-vavilov: cc @Hamms  @snickell

stephenliang: I think we can self-host them but we can do self-hosting as a separate fast follow Jira task (no need to block on this task IMO). We would download the noto fonts and store them at the same place we do for Metropolis.

shared/css/font.scss
levadadenys: Oh wow, one more reason to avoid using styles in js 😄 
Thanks for updating!

PR #59784, opened 2024-07-15
Adding student names to the free responses.

kobryan0619:
Adds student name toggles to summary view.  Note that this lives behind a flag `cfu-show-names-enabled`

View without flag turned on:

image

View with flag turned on (also showing messages sent)

https://github.com/user-attachments/assets/e6419955-8e69-477e-b2a3-d208ec387c2e

## Links

[Jira](https://codedotorg.atlassian.net/jira/software/projects/TEACH/boards/65?selectedIssue=TEACH-1129)
[Figma](https://www.figma.com/design/i3smpK0bu5Tnc1JgmmMfTI/Check-For-Understanding-Summaries?node-id=2349-2606&m=dev)

## Testing story

Tests coming soon!


apps/src/templates/levelSummary/summary.module.scss
```diff
@@ -75,7 +75,7 @@ label.sectionSelector {
   float: right;
 }
 
-[dir="rtl"] .toggleContainer {
+[dir='rtl'] .toggleContainer {
```
kobryan0619, line 78: These quote changes happened on autosave... assuming this is OK (although distracting for reviewers)

dashboard/app/views/levels/_summary.html.haml
```diff
@@ -11,6 +11,7 @@
     {
       user_id: response.user_id,
       text: response.level_source&.data,
+      student_name: User.find_by(id: response.user_id)&.name,
```
kobryan0619: I am looking for extra scrutiny here.  I didn't see anywhere else in our codebase where this approach was used.  BUT I had a really hard time finding a better way to get this data in here.  I also considered matching up student names and ids in the react component (essentially saying "hey, you have a list of `student names` in this section, match them to the `student ids` connected to the `responses`"

I really wanted to get this data in earlier though.  I'd love to hear how we feel about this approach or any other way that might be more aligned to best practices.

apps/src/templates/levelSummary/SummaryResponses.jsx
```diff
@@ -49,13 +52,17 @@ const SummaryResponses = ({
...
+      analyticsReporter.sendEvent(
+        eventName,
+        {
+          levelId: level.id,
+          levelName: level.name,
+          levelType: level.type,
+          sectionSelected: !!selectedSection,
+          ...scriptData.reportingData,
+        },
+        PLATFORMS.BOTH
```
kobryan0619, line 64: Add this via Bryan's request.  Hoping it is ok to add now for all events in this component... @dju90

apps/src/templates/levelSummary/summary.module.scss
```diff
@@ -98,23 +105,39 @@ label.sectionSelector {
...
+        margin: 0;
+        font-size: 14px;
+        font-weight: bold;
+        padding-left: 12px;
+      }
+    }
   }
 
   .studentAnswer {
-    padding: 12px;
```
kobryan0619, line 110: Moved these to the "interior" to ensure the wrapping kept the names included in the wrapping.

dju90:
Looks great!

apps/src/lib/util/AnalyticsConstants.js
```diff
@@ -154,6 +154,10 @@ const EVENTS = {
   SUMMARY_PAGE_NEXT_LEVEL_CLICKED: 'Summary Page Next Level Clicked',
   SUMMARY_PAGE_BACK_TO_LEVEL_CLICKED: 'Summary Page Back To Level Clicked',
 
+  // Check for understanding
+  CFU_NAMES_TOGGLED_ON: 'Summary Page Names Toggled On',
```
lfryemason, line 158: I appreciate you adding these as you go. I forgot and will have to add them in afterwards :(

lfryemason: And thank you for clarifying with Bryan which we need.

apps/src/templates/levelSummary/summary.module.scss
```diff
@@ -98,23 +105,39 @@ label.sectionSelector {
...
-    margin-bottom: 16px;
     // TODO: Update to use #E6F6F8 once the phase-2 brand colors are in.
-    background-color: $lightest_cyan;
```
etaderhold, line 113: Move/remove this TODO comment as appropriate?

dashboard/app/views/levels/_summary.html.haml
etaderhold: Try this? I think it would probably work just as well.

```suggestion
      student_name: response.user.name,
```

apps/src/templates/levelSummary/summary.module.scss
```diff
@@ -98,23 +105,39 @@ label.sectionSelector {
     column-gap: 48px;
     padding-top: 24px;
     padding-bottom: 12px;
+
+    .studentResponseBlock {
+      margin-bottom: 16px;
```
lfryemason, line 116: Can you confirm that these don't get split between columns (e.g. text split between two columns or name and text on two separate columns) when there's a really long response?

We may need to add `display: inline-block` here

apps/src/templates/levelSummary/FreeResponseResponses.jsx
```diff
@@ -1,18 +1,27 @@
...
-          <div className={styles.studentAnswerInterior}>
-            <p>{response.text}</p>
-            <ResponseMenuDropdown response={response} />
+        <div key={response.user_id} className={styles.studentResponseBlock}>
+          <div className={styles.studentAnswer}>
+            <div className={styles.studentAnswerInterior}>
+              <p>{response.text}</p>
+              <ResponseMenuDropdown response={response} />
+            </div>
+            {DCDO.get('cfu-show-names-enabled', false) && (
```
lfryemason: Do we want to put this under the same or different DCDO flag to pinning? I understand I named mine incorrectly to show names, but we may just want/need the one. What do you think?

dashboard/app/views/levels/_summary.html.haml
etaderhold: Also...any thoughts about putting family names in here too? More of a product question than an engineering one.

apps/src/templates/levelSummary/FreeResponseResponses.jsx
```diff
@@ -1,18 +1,27 @@
...
+              <div className={styles.studentName}>
+                {showStudentNames && <p>{response.student_name}</p>}
```
lfryemason: Let's move `showStudentNames &&` outside the studentName div so we aren't creating an empty div when we don't need to. Unless we need it for margins?

apps/src/templates/levelSummary/SummaryResponses.jsx
```diff
@@ -91,6 +98,15 @@ const SummaryResponses = ({
   // only when the policy allows it for that user.
   const showAnswerToggle = scriptData.answer_is_visible && isMulti;
 
+  const toggleNames = () => {
+    if (showStudentNames) {
+      logEvent(EVENTS.CFU_NAMES_TOGGLED_OFF);
+    } else {
+      logEvent(EVENTS.CFU_NAMES_TOGGLED_ON);
+    }
+    setShowStudentNames(!showStudentNames);
```
lfryemason:
```suggestion
    setShowStudentNames((prevShowStudentNames => !prevShowStudentNames);
```
This prevents a race condition from spamming the toggle button back and forth.

We may also want to [debounce](https://www.geeksforgeeks.org/lodash-_-debounce-method/) the logging so it doesn't get spammed, but I think that's probably not too important/necessary.

apps/src/templates/levelSummary/SummaryResponses.jsx
```diff
@@ -115,13 +131,26 @@ const SummaryResponses = ({
...
 
-        {/* Section dropdown */}
-        {hasSections && (
-          <label className={styles.sectionSelector}>
-            {i18n.responsesForClassSection()}
-            <SectionSelector reloadOnChange={true} />
-          </label>
-        )}
+        <div className={styles.headerContainer}>
+          {/* Section dropdown */}
```
lfryemason, line 135: Remove?

dashboard/app/views/levels/_summary.html.haml
lfryemason: This is a very good first step. I think there is a slightly better approach that doesn't require the extra lookup to find the user based on the ID.

The `@responses` here is a list of UserLevels ([see where it's set](https://github.com/code-dot-org/code-dot-org/blob/staging/dashboard/app/controllers/script_levels_controller.rb#L177))

A [user level](https://github.com/code-dot-org/code-dot-org/blob/staging/dashboard/app/models/user_level.rb#L35) has a :user on it

I think we should then be able to do `response.user.name` (similar to [here](https://github.com/code-dot-org/code-dot-org/blob/staging/dashboard/app/models/peer_review.rb#L268))

The only reason this is better is because we don't need to do an extra `User.find_by` query. Although I am by no means a ruby expert, so if you want to check me or want more context, I'd ask Eric

apps/src/templates/levelSummary/FreeResponseResponses.jsx
lfryemason: Also I don't think student_name should be in a `<p>` tag since it's only one name (`<p>` for paragraph). Could we remove?

apps/src/templates/levelSummary/FreeResponseResponses.jsx
kobryan0619: Ah yeah, I was thinking the same flag but when I saw your flag name I thought maybe different... I could see them getting done at slightly different times and then us wanting two different flags.  

That being said, that does make for more variations of testing... let's do one flag.  I will remove mine and just stick with yours (with its current name)

dashboard/app/views/levels/_summary.html.haml
kobryan0619: That's it - I tried examining the full "response" but didn't see user connected to it.  It looks like it might have been because there is a connection to `UserLevel` and a `response` which I missed - I will research this a bit more as I learn more about the nuts / bolts / magic of ruby.

kobryan0619: And, re "family names" - "name" includes full name which is what the spec wants so we should be good with that!

kobryan0619: Thanks for the written explanation!  I do want to chat a bit more about how you found this tomorrow in our 1-1.  I was looking for something like this (... for probably way too long...) but despite my sleuthing this didn't show up on my radar.  Thanks for the digging - this better.

apps/src/templates/levelSummary/summary.module.scss
kobryan0619: Yeah, good point, I believe we are updating this color to purple anyway.  Will remove.

apps/src/templates/levelSummary/summary.module.scss
kobryan0619: Yep - it was actually breaking earlier which caused some moving around of CSS.  I took a video of the long test below to show it working.

https://github.com/user-attachments/assets/8a74709f-7265-46de-897e-2eea2c186c0b

apps/src/templates/levelSummary/SummaryResponses.jsx
kobryan0619: Style wise, these comments are kinda all over the file.  I didn't add this one in but rather it was moved from somewhere else.  I feel good leaving this in given the style of the rest of the file.

apps/src/templates/levelSummary/FreeResponseResponses.jsx
kobryan0619: This was my solution for getting the names to show and un-show without re-locating the text responses (shifting them up and down as the names got entered). 

For the `<p>` tag, I will still need to wrap this in something (a `<div>`?).  Is there a reason in your mind that a `<div>` would be better than `<p>`?

kobryan0619:
Note: I added tests for one of the two changed components.  I think the second test could benefit from a larger change over to Jest which I'd like to put into a separate PR.

dashboard/app/views/levels/_summary.html.haml
etaderhold: By default Rails doesn't know there's anything special about this `response.user_id` compared to any other numeric field. A valid way to look up the user would then be exactly what you did originally: `User.find_by(id: response.user_id)` (or its shorter cousin `User.find(response.user_id)`).

However this `find` syntax is a bit cumbersome. Rails gives a way to define [associations](https://guides.rubyonrails.org/association_basics.html) between models. These let you declare the relationships between tables in your database schema and translate them into properties you can call on the model objects (`response.user` in this case).

It's generally a good bet that when a database table has a `user_id` someone has probably already set up the necessary association linking the models together in this way. Might as well try using it!

Liam already pointed to the code where this happens in this case. I noticed something a bit unusual there: the `optional: true` at the end of the association declaration. This is saying that the `User` query might not actually return anything, and that Rails should return `nil` instead of raising an error in this case. Probably to handle soft-deleted users? Best to use `response.user&.name` to avoid some Honeybadger errors (and also handle this case on the front end so the UI says something like "name not found" instead of "written by nil" or whatever).

In terms of queries, Rails has "lazy loading" which means that when you load a `UserLevel` it won't query for the associated `User` or other associated objects until you actually ask for some field from that object.

Look at this Rails console output:
image

We can see a few things here. First, Rails doesn't query the `users` table when I first load the `UserLevel`...instead it waits until I call `response.user.name`. Second, when I call `response.user.family_name` it doesn't query the database again. The results from the prior query for that user are still stored in memory. Third, Rails issues the exact same query for `response.user.name` as it does for `User.find_by(id: response.user_id).name`. So, going through the Rails association might save you a query if some other part of the code already is looking at something in `response.user` during this web request. Even if not, you're no worse off.

Fourthly, I also made a mistake in my `find_by` query, passing in `response.id` instead of `response.user_id`. Going through the defined association instead of the lower-level query can save bugs too. :-)

etaderhold: > And, re "family names" - "name" includes full name which is what the spec wants so we should be good with that!

I don't actually think this is true. See my Rails console screenshot in my longer comment from a moment ago. `name` is the original name field, what we now refer to as "display name" in the UI. `family_name` is separate and you'll need to query for both if you want to display both.

kobryan0619: Youre right - thanks for checking me here!  This is updated

dashboard/app/views/levels/_summary.html.haml
```diff
@@ -11,6 +11,8 @@
     {
       user_id: response.user_id,
       text: response.level_source&.data,
+      student_display_name: response.user.name,
+      student_family_name: response.user.family_name
```
lfryemason:
```suggestion
      student_display_name: response.user&.name,
      student_family_name: response.user&.family_name
```
See Eric's response for why we need the optional component.

apps/src/templates/levelSummary/FreeResponseResponses.jsx
```diff
@@ -2,13 +2,19 @@ import PropTypes from 'prop-types';
...
+import DCDO from '@cdo/apps/dcdo';
 import i18n from '@cdo/locale';
 
 import ResponseMenuDropdown from './ResponseMenuDropdown';
 
 import styles from './summary.module.scss';
 
-const FreeResponseResponses = ({responses}) => {
+const FreeResponseResponses = ({responses, showStudentNames}) => {
+  const constructStudentName = response =>
```
lfryemason, line 13: Not necessarily for this PR, but we should really have a utils function that gets called each time we need the combined name instead of doing it manually. If we already have one, let's use it here, if not it's probably some tech debt we should fix at some point.

apps/src/templates/levelSummary/FreeResponseResponses.jsx
```diff
@@ -17,15 +23,24 @@ const FreeResponseResponses = ({responses}) => {
...
-                    setHiddenResponses(prevHidden => [...prevHidden, userId])
-                  }
-                />
+            <div key={response.user_id} className={styles.studentResponseBlock}>
+              <div className={styles.studentAnswer}>
+                <div className={styles.studentAnswerInterior}>
+                  <p>{response.text}</p>
+                  <ResponseMenuDropdown
+                    response={response}
+                    hideResponse={userId =>
```
lfryemason, line 32: Sorry for all the merge conflicts earlier :\

lfryemason (approved):
Approving assuming you add the `&`s.

dashboard/app/views/levels/_summary.html.haml
kobryan0619: Thanks so much for taking time to write this up, Eric.  This was so helpful in filling in the blanks!

apps/src/templates/levelSummary/FreeResponseResponses.jsx
kobryan0619: I will make a ticket for this!

PR #59789, opened 2024-07-15
[Google Blockly] Angle Helper and Angle Text Input fields

mikeharv:
## Context

This PR migrates the Angle Helper from CDO Blockly and implements it for Angle Text Input fields.

### About the Angle Helper (CDO Blockly)
CDO Blockly has an Angle Helper which can be initialized within the editor of various fields. Its primary use is to help students select a valid angle number (0-359) in order to program the artist to face the desired direction. It can show up inside 3 field types: Angle Text Input, [Angle Dropdown](https://github.com/code-dot-org/code-dot-org/pull/59816), and [Text Input](https://github.com/code-dot-org/code-dot-org/pull/59845). Each of these three field types has an associated `turn` block and a `point to` block.

### About Angle Text Input Fields (CDO Blockly)
These are the most straight-forward angle field option. Users can enter a number or use the angle helper to choose one. If this is a `turn` block, the angle field checks the current value of the `'DIR'` field to decide whether the angle should increase in a clockwise or counter-clockwise fashion. For `point to` blocks, it always turns to the right.
imageimageimage

3 months ago, we added Blockly's plugin for angle fields: https://github.com/code-dot-org/code-dot-org/pull/57826
This field features a completely different UI but was useful for us to get many Artist levels to point where they wouldn't immediately crash on Google Blockly. After sharing this quick solution with the curriculum team, it became clear that it would not meet our needs. We were going to need to migrate our full Angle Helper and use that instead. This means that this branch will revert some of the work linked above.

## Approach
The original Angle Helper is defined in CDO Blockly here:
https://github.com/code-dot-org/blockly/blob/v4.1.0/core/ui/fields/angle_helper.js

The new version has been added to our addons/ directory with a couple noticeable changes:
* The files has been updated to use modern class structures and TypeScript
* We no longer rely on the `goog.math.Vec2` class. (https://google.github.io/closure-library/api/goog.math.Vec2.html) This required re-creating a few vector methods.
* Fixes a bug when drawing arcs for angles > 360. While the Anger Helper can only select values from 0-359, anything can be entered in the text field, and the angle helper will attempt to animate it. Below, you can see a couple arcs that CDO Blockly fails to draw correctly. We now normalize the values  used to draw the arc to be within the expected range. This is done with the modulus operation (`%`) at the start of the new `describeArc` method.

**Before/CDO**
image
image

**After/Google**
image
image

In this branch, my goal was to find the path of least resistance to test that the Angle Helper was working as expected within a block field. Therefore, I added it directly to the Angle Text Input field class, similar to https://github.com/code-dot-org/blockly/commit/12311ee6483d5f74f0515192624b4886a68fb396. (In CDO Blockly, it is now added to the parent Text Input field class. This is handled in a follow-up branch. https://github.com/code-dot-org/code-dot-org/pull/59845)

It is worth noting that for Google Blockly, the Angle Text Input field extends the Number field which extends the Text Input field. In CDO Blockly, Number fields don't exist; rather it's just a Text Input field with custom value validation. It might make sense to rename this field `CdoAngleNumber`, but it wouldn't be worth doing that until we're off CDO Blockly completely.

## Links

## Follow-up work

- https://github.com/code-dot-org/code-dot-org/pull/59816
- https://github.com/code-dot-org/code-dot-org/pull/59845

## PR Checklist:

- [ ] Code is well-commented


apps/src/blockly/addons/cdoAngleHelper.ts
```diff
@@ -0,0 +1,533 @@
...
+  arcColour?: string;
+  angle?: number;
+  height?: number;
+  width?: number;
+  snapPoints?: number[];
+  onUpdate?: () => void;
+  enableBackgroundRotation?: boolean;
+}
+
+class Vector {
```
molly-moen: can we split up this file a bit since it's quite long? Looks like there's 2 classes and some helper methods, so we could probably split it into at least 3 files?

molly-moen: could be done as follow-up: can you add comments explaining to this file? Specifically around the purpose of the classes and the math that we are doing.

apps/src/blockly/addons/cdoAngleHelper.ts
```diff
@@ -0,0 +1,533 @@
...
+  private arc_: SVGPathElement | null;
+  public svg_: SVGSVGElement | null;
+  private rect_: DOMRect | null;
+  private animationInterval_: number | undefined;
+  private mouseMoveWrapper_: ((e: MouseEvent) => void) | null;
+  private mouseUpWrapper_: ((e: MouseEvent) => void) | null;
+  private mouseDownWrapper_: ((e: MouseEvent) => void) | null;
+
+  constructor(direction: string, options: AngleHelperOptions = {}) {
+    this.lineColour_ = '#4d575f';
```
molly-moen: can we import this color from `apps/src/util/color.js` instead of hard-coding it?

apps/src/blockly/addons/cdoAngleHelper.ts
```diff
@@ -0,0 +1,533 @@
...
+    this.strokeWidth_ = 3;
```
molly-moen, line 71: nit: constants for the magic numbers and `'turnRight'`?

apps/src/blockly/addons/cdoAngleHelper.ts
```diff
@@ -0,0 +1,533 @@
...
+    return this.picker_.angle;
+  }
+
+  init(svgContainer: HTMLElement): void {
+    this.svg_ = Blockly.utils.dom.createSvgElement(
+      'svg',
+      {
+        xmlns: 'http://www.w3.org/2000/svg',
+        'xmlns:html': 'http://www.w3.org/1999/xhtml',
+        'xmlns:xlink': 'http://www.w3.org/1999/xlink',
```
molly-moen: why do we need the `xmlns` values?

apps/src/blockly/addons/cdoAngleHelper.ts
```diff
@@ -0,0 +1,533 @@
...
+        stroke: this.arcColour_,
+        fill: this.arcColour_,
+        'fill-opacity': '0.3',
+        'stroke-width': this.strokeWidth_,
+      },
+      this.svg_
+    ) as SVGPathElement;
+
+    for (let angle = 0; angle < 360; angle += this.background_.tickSpacing) {
+      let markerSize;
```
molly-moen, line 241: this section would be an especially good candidate for comments on what this logic is doing

apps/src/blockly/addons/cdoAngleHelper.ts
```diff
@@ -0,0 +1,533 @@
...
+      },
+      this.svg_
+    ) as SVGLineElement;
+
+    this.picker_.handle = Blockly.utils.dom.createSvgElement(
+      'circle',
+      {
+        cx: this.picker_.handleCenter!.x,
+        cy: this.picker_.handleCenter!.y,
+        fill: '#a69bc1',
```
molly-moen: another candidate for importing a color constant

apps/src/blockly/addons/cdoAngleHelper.ts
```diff
@@ -0,0 +1,533 @@
...
+    startAngle %= 360;
+    endAngle %= 360;
+    const vector = center.clone().add(new Vector(radius, 0));
+    const start = Vector.rotateAroundPoint(vector, center, startAngle);
+    const end = Vector.rotateAroundPoint(vector, center, endAngle);
+
+    const largeArcFlag = Math.abs(startAngle - endAngle) > 180 ? '1' : '0';
+    const sweepFlag = endAngle - startAngle < 0 ? '0' : '1';
+
+    const d = [
```
molly-moen: nit: better variable name?

apps/src/blockly/addons/cdoFieldAngleTextInput.ts
```diff
@@ -1,44 +1,117 @@
...
+import GoogleBlockly from 'blockly/core';
 
-import {CLOCKWISE_TURN_DIRECTION} from '../constants';
+import {ExtendedBlockSvg} from '../types';
+
+import CdoAngleHelper from './cdoAngleHelper';
 
 interface AngleTextInputOptions {
   directionTitle: string; // Ex. 'DIR'
   direction: string; // Ex. 'turnRight'
```
molly-moen: will the direction always be some specific string like `turnRight` or `turnLeft`? If so, you can make this a more specific type with this syntax:
```
direction: 'turnRight' | 'turnLeft';
```

molly-moen (approved):
My comments are all for prettifying, so they can be done as follow ups! The old code in cdo looks like it was pretty uncommented, I'm assuming we will not be replacing this custom functionality with core blockly at any point we should make the new version more maintainable.

apps/src/blockly/addons/cdoAngleHelper.ts
mikeharv: For now, I'm moving the vector class out to a separate file and explaining its purpose/source in comments.

apps/src/blockly/addons/cdoAngleHelper.ts
mikeharv: There are unfortunately a ton of little magic numbers throughout this file, most of which are only used once. Ideally, some of these values would be computed based on a set scale for the entire component. Since everything is static, I'm inclined to leave them as is for now, rather define a whole bunch of constants with super context-specifc names.

apps/src/blockly/addons/cdoAngleHelper.ts
mikeharv: They were in the original file, but it doesn't seem like they're needed. Removed!

apps/src/blockly/addons/cdoAngleHelper.ts
molly-moen: Makes sense. I do see `'turnRight'` a few times, would that one be worth a constant?

mikeharv: Yes, I've done that here: https://github.com/code-dot-org/code-dot-org/pull/59789/commits/de25569f59b3bb6188e9d7a7e2b87922c7ac73de#diff-2340c7f9b234c61841daecd90ec745870b3bab1c80d511f68b6b07dedfcac3b8R79

I'll double-check to make sure I didn't miss any instances!

PR #59797, opened 2024-07-16
Upgrade TypeScript to 5.5.3

snickell:
Upgrades these to the latest version:
- TypeScript to 5.5.3
- ts-loader to 9.5.1
- @typescript-eslint/eslint-plugin to 7.16.1


apps/src/blockly/utils.ts
```diff
@@ -35,7 +35,7 @@ export const TRUTHY_DEFAULT = (attributeValue: xmlAttribute) =>
 export function readBooleanAttribute(
   xmlElement: Element,
   attribute: string,
-  callback = FALSEY_DEFAULT
+  callback: (attributeValue: xmlAttribute) => boolean = FALSEY_DEFAULT
```
snickell, line 38: TS 5.5.3 got too smart and was auto-typing callback as:
`(attributeValue: xmlAttribute) => attributeValue is "true"`

Note the `is "true"` which is overly specific (compared to boolean!), and caused type errors when TRUTHY_DEFAULT was passed in as the callback. Explicitly typing callback to widen its return signature back to boolean fixes the issue.

stephenliang (approved):

PR #59799, opened 2024-07-16
DatablockStorage.addSharedTable(): don't refresh view until async callback

### LINKED ISSUE #58456: Import table sometimes doesn't auto-refresh to show new table
We probably need to look at the JSX logic for when it calls a refresh

Image

snickell:
We were calling `refreshCurrentDataView()` before DatablockStorage.addSharedTable() had returned. On local, this seemed to work, but with prod latency, this resulted in the Data browser "Import" button not seeming to work until you refreshed.

Fixes [#58456](https://github.com/code-dot-org/code-dot-org/issues/58456)

Now the calls are sequential:
image


cnbrenci (approved):
Sweet!

PR #59814, opened 2024-07-16
Add amplitude events for metadata

kobryan0619:
Added amplitude events for:

- Opening all and expanding all meta data
- Opening and closing a single row of meta data

Video below.

https://github.com/user-attachments/assets/8ac49e97-6259-4427-91ef-aef3003274b5

Note: You will see some warnings in the console about styling of the dropdown.  This is because we are overriding some styles from the pop-up menu.  We know an action menu is on the table for our design system that we can use for this component when it is done OR repurpose what Liam is working on for hiding and pinning student responses in the summary view.

## Links

[Ticket](https://codedotorg.atlassian.net/jira/software/projects/TEACH/boards/65?selectedIssue=TEACH-1182)

### DIFF
--- a/apps/src/templates/sectionProgressV2/SectionProgressV2.jsx
+++ b/apps/src/templates/sectionProgressV2/SectionProgressV2.jsx
@@ -77,9 +77,7 @@ function SectionProgressV2({
           {i18n.lessonsIn()}
 
           <UnitSelectorV2 className={styles.titleUnitSelectorDropdown} />
-          {expandedMetadataEnabled && (
-            <MoreOptionsDropdown sectionId={sectionId} />
-          )}
+          {expandedMetadataEnabled && <MoreOptionsDropdown />}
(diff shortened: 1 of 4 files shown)


apps/src/templates/sectionProgressV2/SectionProgressV2.jsx
```diff
@@ -77,7 +77,9 @@ function SectionProgressV2({
           {i18n.lessonsIn()}
 
           <UnitSelectorV2 className={styles.titleUnitSelectorDropdown} />
-          {expandedMetadataEnabled && <MoreOptionsDropdown />}
+          {expandedMetadataEnabled && (
+            <MoreOptionsDropdown sectionId={sectionId} />
```
lfryemason: Can we get this from Redux instead of passing in a prop?

lfryemason (approved):

PR #59816, opened 2024-07-16
[Google Blockly] Implement Angle Helper for Angle Dropdown fields

mikeharv:
This PR implements the [Angle Helper](https://github.com/code-dot-org/code-dot-org/pull/59789) for Angle Dropdown fields.

## About the Angle Helper (CDO Blockly)
CDO Blockly has an Angle Helper which can be initialized within the editor of various fields. Its primary use is to help students select a valid angle number (0-359) in order to program the artist to face the desired direction. It can show up inside 3 field types: Angle Text Input, Angle Dropdown, and [Text Input](https://github.com/code-dot-org/code-dot-org/pull/59845). Each of these three field types has an associated `turn` block and a `point to` block.

### About Angle Dropdown Fields (CDO Blockly)
In both repos, the Angle Dropdown field extends a standard Dropdown field. When `showEditor_` is called, the dropdown menu options are populated within the dropdown div. We essentially hijack this div and insert the angle helper in beside it. The angle helper supports "snap options" which mean it can only be used to select values that are included by the menu generator. This makes for complex and nuanced UI. When the angle helper is used, we immediately set the value on the field and mirror the selection onto the menu options. If a user hovers of a menu option, it is highlighted, but we also need to animate the angle helper to preview the selection.

https://github.com/user-attachments/assets/c0365d07-0c7c-4618-8210-e89f14db6b1d

Like other angle fields, if this is a `turn` block, the angle field checks the current value of the `'DIR'` field to decide whether the angle should increase in a clockwise or counter-clockwise fashion. For `point to` blocks, it always turns to the right.

|image|image|
|-|-|

3 months ago, we added Blockly's plugin for angle fields: https://github.com/code-dot-org/code-dot-org/pull/57826
This field features a completely different UI but was useful for us to get many Artist levels to point where they wouldn't immediately crash on Google Blockly. Because Blockly's angle field extend Field Number, they don't support dropdown menus. Currently, "Angle Dropdown fields" on Google Blockly are treated like basic angle (number) fields, except that the would-be menu options are used to set the only valid values on the field editor. After sharing this quick solution with the curriculum team, it became clear that it would not meet our needs. We were going to need to migrate our full Angle Helper and use that instead. This means that this branch will revert some of the work linked above.

## Approach
Similar to https://github.com/code-dot-org/code-dot-org/pull/59789, this required reverting some earlier work that got Blockly's Angle Field plugin working. That plugin's field class explicitly extends the Number field, which isn't what we want for a dropdown.

To get the menu and angle helper to show up as expected requires some custom CSS. To facilitate this, I added the class `fieldAngleDropDownContainer` to the dropdown's content div. When a different field's editor is opened, that class is automatically cleaned up by Blockly.

The biggest challenges in this branch were mirroring the angle value selection two ways (_menu option hover to angle helper_ and _angle helper to menu_). Mainline Blockly doesn't expect either of these to be necessary and so there are no APIs to really help here. For example,  there is no way to change which option is selected (with a `✓`), because selecting a value immediately closes the menu.

### Mirroring angle helper to the dropdown menu

The angle helper has an `onUpdate` property which can be used to help update the menu. The requirement here is that when an angle is chosen, we should show the check mark next to it in the menu and stop highlighting the current option if there is one. Blockly's `Menu` class is mostly private, so I needed to use a query selector to find all `.blocklyMenuItem` elements and manually modify their classes.

### Previewing angles when hovering over menu options
In CDO Blockly, the `goog.events` library is used to detect highlighting: 
https://github.com/code-dot-org/blockly/blob/v4.1.0/core/ui/fields/field_angle_dropdown.js#L89-L101
My first attempt to replicate this was just to listen for `'mouseover'` events. This worked as expected but was an incomplete solution. Blockly has built-in keyboard navigation of dropdown menu options using the arrow keys. We also have enabled the keyboard navigation which supports dropdown menu traversal using WASD. I was able to listen for W and S key presses, but unfortunately blockly swallows `ArrowUp` and `ArrowDown` events, so I couldn't detect them.
https://github.com/google/blockly/blob/8cad97a214f23311a24143b9e18ce3b9a5d6c0a2/core/menu.ts#L420-L452

Overriding the `Menu` class isn't supported by Blockly and I don't know what side effects this swallowing behavior is intended to prevent. As a workaround, I found the [MutationObserver API](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver), which allows us to watch for changes being made to the DOM tree (e.g. new classes).  I set up a watch to check for elements that gain the class `.blocklyMenuItemHighlight.` When this occurs, we respond by finding the angle value in its inner node and animate the angle helper accordingly. Here's a sample menu item from the DOM (simplified):

```
<div id="blockly-17" class="blocklyMenuItem blocklyMenuItemHighlight" role="option" aria-selected="true" aria-disabled="false">
   <div class="blocklyMenuItemContent">
      <div class="blocklyMenuItemCheckbox" />
      90
   </div>
</div>
```

By doing this, we can extend the expected mirroring functionality to any potential means of highlighting a menu option:

|Mousing over|Arrow keys|W/S (Keyboard navigation plugin)|
|-|-|-|
|image|image|image|

## Links

## PR Checklist:

- [ ] Code is well-commented


apps/src/blockly/addons/cdoFieldAngleDropdown.ts
```diff
@@ -1,118 +1,177 @@
...
-    }
-    super.fromXml(element);
+  private initializeAngleHelper(): void {
+    const container = this.createAngleHelperContainer();
+    const sourceBlock = this.getSourceBlock() as Block;
+    this.angleHelper = new Blockly.AngleHelper(this.getDirection(), {
+      onUpdate: this.updateDropdownMenuOptions.bind(this),
+      snapPoints: this.getOptions().map(option => parseInt(option[1])),
+      arcColour: (sourceBlock as ExtendedBlockSvg)?.style.colourPrimary,
+      angle: parseInt(`${this.getValue()}`),
```
molly-moen, line 95: why do we parseInt a stringified number? (Does getValue() return a number?)

apps/src/blockly/addons/cdoFieldAngleDropdown.ts
```diff
@@ -1,118 +1,177 @@
...
+   * Update the dropdown menu options to reflect the selected angle value.
+   * @param {string | number} angleValue The selected angle value.
+   * @private
    */
-  toXml(element: Element) {
-    if (this.config) {
-      element.setAttribute('config', this.config);
+  private updateDropdownMenuOptions(): void {
+    const angleValue = this.angleHelper?.getAngle() as number;
+    this.setValue(`${angleValue}`);
```
molly-moen: nit: convert `angleValue` to a string when we initialize the value so we don't have to repeatedly convert it to a string in this method?

apps/src/blockly/addons/cdoFieldAngleDropdown.ts
```diff
@@ -1,118 +1,177 @@
...
+  }
+
+  /**
+   * Observe the menu element for changes and animate the angle helper based on the
+   * highlighted option's value.
+   * @private
+   */
+  private animateAngleHelperOnMenuOptionHighlight() {
+    // In order to preview dropdown a highlighted menu option, we detect changes to its
+    // class and animate the angle helper based on its value. A user might do this my
```
molly-moen: typo: by instead of my

apps/src/blockly/addons/cdoFieldAngleDropdown.ts
```diff
@@ -1,118 +1,177 @@
...
-  toXml(element: Element) {
-    if (this.config) {
-      element.setAttribute('config', this.config);
+  private updateDropdownMenuOptions(): void {
+    const angleValue = this.angleHelper?.getAngle() as number;
+    this.setValue(`${angleValue}`);
+    // Blockly has no public methods for getting dropdown menu items, so this
+    // is a necessary hack.
+    const menuItems =
+      Blockly.DropDownDiv.getContentDiv().querySelectorAll('.blocklyMenuItem');
```
molly-moen, line 112: could we accidentally get the wrong/extra menu items with this selector?

molly-moen (commented):
Looks good, a couple nits and one clarifying question!

apps/src/blockly/addons/cdoFieldAngleDropdown.ts
mikeharv: This was to satisfy TypeScript. In practice, `this.getValue()` actually returns a string, but Blockly says it could also be null. `parseInt` expect a string. In theory, `parseInt('null')` is okay as it returns `NaN`, but we can't really hit that case.

apps/src/blockly/addons/cdoFieldAngleDropdown.ts
mikeharv: I don't believe so. Blockly only ever has one dropdown menu open at a time. It's literally the same div that gets moved around and repurposed. When we're on this type of field, we want to potentially update all of the menu items it contains.

molly-moen (approved):

PR #59827, opened 2024-07-17
Datablock Storage: port current_table cron scripts from Firebase

### LINKED ISSUE #56993: Implement cron-updating tables (daily weather, spotify, etc)
In Datablock Storage, unlike in the firebase version, we don't distinguish between "shared_tables" and "current_tables". Shared tables are just a pointer to the real dataset, which is stored in the DatablockStorageTable table with a project_id of 0.

There are 3 "current tables", which are a result of 3 different cron jobs running (from `bin/cron/applab_datasets`), fetching data, and updating tables.

snickell:
This PR implements `DatablockStorageTable.update_shared_table(table_name, records)`, and uses it to port the cron jobs that update "current tables" from Firebase to Datablock Storage.

Port these cron scripts from Firebase to datablock storage:
- [x] Convert covid19 cron job: updates shared table_name `'COVID-19 Cases per US State'`
- [x] Convert daily_weather cron job: updates shared table_name `'Daily Weather'`
- [x] Convert spotify cron job: updates 5x different tables `'Top 50 Worldwide'`, `'Top 50 USA'`, `'Viral 50 Worldwide'`, `'Viral 50 USA'`

Tested for each cron script:
- Created records are valid and matches REST fetched record data
- Updates existing stale data records
- Creates table if table itself doesn't exist
- Updates if all records are deleted
- Updates if only some records are deleted
- Updates `DatablockStorageTable.lastUpdated` time

Fixes #56993

### DIFF
--- a/bin/cron/applab_datasets/covid19
+++ b/bin/cron/applab_datasets/covid19
@@ -1,9 +1,11 @@
 #!/usr/bin/env ruby
 
+require_relative '../only_one'
+abort 'Script already running' unless only_one_running?(__FILE__)
+
 # This script fetches covid19 data and uploads it to a datablock storage shared_table.
 require_relative '../../../dashboard/config/environment'
 require_relative '../../../deployment'
-require 'cdo/only_one'
 require 'net/http'
 require 'csv'
 require 'date'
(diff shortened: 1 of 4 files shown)


snickell:
The `covid19` script is ported, tested, and works, BUT the data source is permanently stale. If curriculum approves, I believe we should delete this cron job in a follow-up PR, but might have to wait until @kakiha11 is back from vacation to get an answer so think we should proceed with the "useless but correct port" in this PR.

More details: https://github.com/code-dot-org/code-dot-org/issues/56993#issuecomment-2232227567

cnbrenci (approved):
WOOOOO!

PR #59837, opened 2024-07-17
AI Tutor: concat system prompt + default to Python 

Erin007:
https://codedotorg.atlassian.net/browse/CT-693
https://codedotorg.atlassian.net/browse/CT-696

In a world where we are support multiple programming languages or conditionally supporting one (ie. while we decide on Python v. Java for piloting) our system prompt needs to be flexible. Also, it shouldn't be the responsibility of the OpenAI controller to build the system prompt. This PR moves the work of creating a level-specific system prompt to the new `AiTutorSystemPromptHelper`.  This change sets us up to try out AI Tutor in the new Python progression, because if AI Tutor is being used outside the context of a CSA lesson, we'll default to referencing Python in the system prompt. 

Follow up logged to update the base system prompt in s3 and go back to reading it from there once it's more solidified, https://codedotorg.atlassian.net/browse/CT-697. 

## PR Checklist:

- [ ] Code is well-commented


dashboard/app/helpers/aitutor_system_prompt_helper.rb
```diff
@@ -0,0 +1,81 @@
+module AitutorSystemPromptHelper
+  def self.get_base_system_prompt
```
Erin007, line 14: Ultimately we'll move this back to s3, but during this testing/iterating phase I want it to be in a place that can be easily referenced

dashboard/app/helpers/aitutor_system_prompt_helper.rb
```diff
@@ -0,0 +1,81 @@
...
+    unit = Unit.find(script_id)
+    unless unit
+      return render(status: :bad_request, json: {message: "Couldn't find unit with id=#{script_id}."})
+    end
+
+    system_prompt =
+      get_base_system_prompt +
+      get_language_specific_system_prompt(unit) +
+      get_level_instructions(level) +
+      get_validated_level_test_file_contents(level)
```
Erin007: I like this set up because it makes it super easy to also add in level-specific customizations if we store them as properties on a level in the future.

apps/src/lib/levelbuilder/ai-tutor/AITutorTesterSampleColumns.tsx
```diff
@@ -50,6 +50,10 @@ const AITutorTesterSampleColumns: React.FC = () => {
         </tbody>
       </table>
       <br />
+      <p>
+        * If you don't upload a levelId, you need to upload a systemPrompt. The
+        default system prompt is created using level-specific context.
+      </p>
```
Erin007: - [ ] add scriptId here too!

dashboard/app/helpers/aitutor_system_prompt_helper.rb
```diff
@@ -0,0 +1,81 @@
...
+  def self.get_level_instructions(level)
+    level_instructions = level.properties["long_instructions"]
+    "\n Here are the student instructions for this level: #{level_instructions}"
+  end
+
+  def self.get_validated_level_test_file_contents(level)
+    test_file_contents = ""
+    if level.validation
+      if level.validation.values.empty?
+        return render(status: :bad_request, json: {message: "There are no test files associated with level id=#{level_id}."})
```
bencodeorg: For my understanding, why don't we return an error if a level has no validation (but do if its values are empty)?

bencodeorg (approved):
Nice!

dashboard/app/helpers/aitutor_system_prompt_helper.rb
Erin007: My understanding that it's expected that there would be non-validated JavaLab levels, but if the level _is_ validated and we can't find the test files then something went wrong.

bencodeorg: Got it, thanks!

dashboard/test/controllers/openai_chat_controller_test.rb
```diff
@@ -65,6 +59,10 @@ class OpenaiChatControllerTest < ActionController::TestCase
   params: {},
   response: :bad_request
 
+  # test level not found
+
+  # test script/unit not found
+
```
Erin007: - [ ] add test cases

dashboard/app/helpers/aitutor_system_prompt_helper.rb
```diff
@@ -0,0 +1,103 @@
...
+  def self.get_level(level_id)
+    level = nil
+    if level_id
+      level = begin Level.find(level_id)
+      rescue ActiveRecord::RecordNotFound
+        Honeybadger.notify(exception,
+            error_message: 'Invalid level_id in AI Tutor system prompt helper',
+            context: {
+              level_id: level_id,
+              user_id: current_user.id
```
Erin007, line 79: If there's no `level_id` that's totally fine, we just won't append the level-specific additions to the system prompt. If there is a `level_id` but we can't find the associated `level` still build the system prompt without erroring, and log to HB so we can hopefully diagnose the mis-match. Same with unit.

PR #59845, opened 2024-07-17
[Google Blockly] migrate field helpers and implement with angle helper for number fields

mikeharv:
## Context

This PR implements the [Angle Helper](https://github.com/code-dot-org/code-dot-org/pull/59789) for Number fields.

### About the Angle Helper (CDO Blockly)
CDO Blockly has an Angle Helper which can be initialized within the editor of various fields. Its primary use is to help students select a valid angle number (0-359) in order to program the artist to face the desired direction. It can show up inside 3 field types: Angle Text Input, [Angle Dropdown](https://github.com/code-dot-org/code-dot-org/pull/59816), and Text Input. Each of these three field types has an associated `turn` block and a `point to` block.

### About Text Input Fields (CDO Blockly)
Text Inputs are a basic field type with no additional editor UI. They can optionally use a custom validator to ensure the value is a valid number. The common `math_number` block is an example: image

In Artist, we'll often have variants of a block with number inputs to make it possible for students to program with variables and complex expressions. If a `math_number` block is placed inside a `turn` or `point to` block like this, we show the angle helper:
image

This means that the number field in the purple math block needs to have awareness of its parent block and be able to determine if it should show the angle helper. To do this a concept of "field helpers" was added to CDO Blockly. Text Input fields look for these extra options on the parent block by traversing the output connection of the block: https://github.com/code-dot-org/blockly/commit/0dcc1a08b0ff54bcc69c870c46abf1f695ee9b6d

If Angle Helper options are found on the parent block, we'll use them to show it. For example, the options of the `turn` block indicate that the `'DIR'` field's value specifies the turning direction. So the connected math block will look there to determine if the angle helper should go clockwise or counter-clockwise.
https://github.com/code-dot-org/blockly/commit/e0e998a6f548dc74f7f2620f1753d504bbb718f7
image

These blocks do not currently show an angle picker. It should be noted that blocks like `math_number` now use the new `FieldNumber` class.

## Approach
Field helpers require a custom `Connection` method which I added via the wrapper. This is similar to other patching we've done with `ExtendedBlock`and `ExtendedInput`. 

To keep things "simple", I opted to migrate the field helpers concept in a 1:1 way. I created a new `CdoFieldNumber` class (which can sit between `GoogleBlockly.FieldNumber` and `CdoFieldAngleText`). This extension of the Number field class will be able to read these options from the connection and show the angle helper. To show the angle helper, I was able to copy several methods out of the `CdoFieldAngleTextInput`.

Finally, by setting the recently-created `CdoFieldAngleTextInput` to to extend `CdoFieldNumber`, we can remove all of the redundant methods. Instead of looking at the connection for the options, for `CdoFieldAngleTextInput` we will expect to be set in the constructor.

With all of these changes, we can now show the Angle Helper on `math_number` blocks when they are connected to a `turn` block:
image

The angle helper does not if the block is anywhere else:
image

There are no regressions for the Angle Text Input Fields that were added in the [base branch](https://github.com/code-dot-org/code-dot-org/pull/59789):
image

## Links

## PR Checklist:

- [ ] Code is well-commented


apps/src/blockly/addons/cdoFieldAngleTextInput.ts
```diff
@@ -24,94 +21,30 @@ export default class CdoFieldAngleTextInput extends GoogleBlockly.FieldNumber {
...
-   * @private
+   * @override
    */
-  private getDirection(): string {
+  getDirection(): string {
     let direction = this.direction;
     if (!direction && this.directionFieldName) {
       direction = this.getSourceBlock()?.getFieldValue(this.directionFieldName);
     }
+    console.log({direction});
```
molly-moen: remove this?

apps/src/blockly/addons/cdoFieldNumber.ts
```diff
@@ -0,0 +1,162 @@
...
+    config?: FieldNumberConfig
+  ) {
+    super(value, min, max, precision, validator, config);
+    this.angleHelper = null;
+  }
+
+  /**
+   * If this field is attached to a block whose output connection is attached to a
+   * connection that has the specified field helper, get the options for that
+   * field helper.
```
molly-moen, line 54: I'm having a hard time parsing this comment, although I'm not sure if there's a way to make it clearer. Maybe describing an example block configuration that this applies to?

apps/src/blockly/addons/cdoFieldNumber.ts
```diff
@@ -0,0 +1,162 @@
...
+
+  protected showEditor_(e?: Event, quietInput?: boolean): void {
+    super.showEditor_(e, quietInput);
+    if (this.shouldShowAngleHelper()) {
+      this.initializeAngleHelper();
+    }
+  }
+
+  /**
+   * Initialize the angle helper.
```
molly-moen: nit: explain what the angle helper is, and that we only use it if a certain field helper is specified on the parent block?

apps/src/blockly/addons/cdoFieldNumber.ts
```diff
@@ -0,0 +1,162 @@
...
+    const angleHelperDiv = this.createAngleHelperContainer();
+    Blockly.WidgetDiv.getDiv()?.appendChild(angleHelperDiv);
+    this.angleHelper.init(angleHelperDiv);
+  }
+
+  /**
+   * Get the direction from either the hardcoded setting or the direction field.
+   * @returns {string} The direction value.
+   */
+  getDirection(): string {
```
molly-moen: Is this only used when there's an angle helper? If so, it might be useful to rename this or update the comment to explain that.

apps/src/blockly/addons/cdoFieldNumber.ts
```diff
@@ -0,0 +1,162 @@
...
+      this.getSourceBlock() as ExtendedBlockSvg
+    ).style.colourPrimary;
+    div.style.padding = '5px';
+    div.style.paddingBottom = '0px';
+    Blockly.utils.dom.addClass(div, 'blocklyAngleHelperContainer');
+    return div;
+  }
+
+  /**
+   * Override to handle value update and animate the angle helper accordingly.
```
molly-moen: nit: `animate the angle helper accordingly, if it exists`.

apps/src/blockly/googleBlocklyWrapper.ts
```diff
@@ -462,6 +475,41 @@ function initializeBlocklyWrapper(blocklyInstance: GoogleBlocklyInstance) {
...
+   * @param {string} fieldHelper the field helper to retrieve. One of
+   *        Blockly.BlockFieldHelper
+   * @param {*} options for this helper
+   * @return {!Blockly.Input} The input being modified (to allow chaining).
+   */
+  extendedInput.addFieldHelper = function (
+    fieldHelper: string,
+    options: FieldHelperOptions
+  ) {
+    // if (this.type !== Blockly.INPUT_VALUE) {
```
molly-moen: remove this?

molly-moen (commented):
Looks good! Mostly nit-picking about comments 😄

apps/src/blockly/addons/cdoFieldNumber.ts
mikeharv: I had a hard time writing it! adding an example which hopefully makes it clearer?

apps/src/blockly/googleBlocklyWrapper.ts
mikeharv: Oh, thanks! I meant for this to be here, but I needed to correct the way the type check was written.

apps/src/blockly/addons/cdoFieldNumber.ts
molly-moen: that's clearer, thanks!

molly-moen (approved):

PR #59855, opened 2024-07-17
Lab2: Add lifecycle notifier

sanchitmalhotra126:
This adds a mechanism to Lab2 for components to hook into lifecycle events in the lab2 system, specifically through level transitions. This allows components to get notified when specific steps in the level change process occur, without depending on redux state changes to signal these events.

The motivation for adding this came from trying to implement analytics reporting for Music Lab on script levels; currently, Music Lab only reports session analytics on standalone projects, but in extending this to script levels, I found it challenging to manage the analytics session state purely through redux state transitions. I believe we've also dealt with similar challenges before, for example where the progress system's level ID updates before the new level is loaded, which requires lab components to know the internals of when specific pieces of data update. We've gotten around these today just by knowing the specifics of the state transitions involved, but I thought it would be good to have a dedicated mechanism to hook into these events explicitly, without needing to make assumptions based on the type of data that's updating. Another use case of this could be latency tracking and metrics reporting.

I also made some updates to `browserNavigation.js` to explicitly store the level ID in the browser history stack, rather than computing the level ID from the URL. This way we can also more easily publish the lab2 lifecycle event when the level is changing due to a back button press.

In this change, nothing is actually registered with this notifier; I'll follow up with the analytics changes music lab that will use these events.

## Testing story

Tested with a few different lab2 progressions (music intro, allthethings lab2 showcase, Gen AI pilot), as well as some legacy progressions to make sure nothing regressed.


apps/src/lab2/utils/notifyLevelChange.ts
```diff
@@ -0,0 +1,13 @@
+import Lab2Registry from '../Lab2Registry';
+
+import {LifecycleEvent} from './LifecycleNotifier';
+
+// Function for external components to notify the Lab2 lifecycle notifier that the level is changing.
+export default function notifyLevelChange(
```
sanchitmalhotra126, line 6: Not sure if this file is overkill; I mostly felt weird about importing `Lab2Registry` and `LifecycleNotifier` in files outside of lab2 (like `progressRedux` and `browserNavigation`)

apps/src/code-studio/browserNavigation.js
```diff
@@ -26,41 +27,19 @@ export function canChangeLevelInPage(currentLevel, newLevel) {
...
-    }
-
-    const values = path.split('/', sublevelNumberStringIndex + 1);
-    if (values.length < levelNumberStringIndex + 1) {
-      return;
-    }
-
-    const levelNumber = Number(values[6]);
-    if (!Number.isInteger(levelNumber) || levelNumber <= 0) {
+    const levelId = event.state?.levelId;
```
breville, line 34: Nice simplification!

breville (approved):
The lifecycle notification system should help simplify lab code and is great to see.  The simplifications in the browser navigation code are also great!

apps/src/code-studio/header.js
```diff
@@ -107,7 +107,9 @@ header.build = function (
 
   // Set up a navigation handler, in case we contain levels that don't
   // require a page reload when switching between them.
-  setupNavigationHandler(lessonData);
+  if (lessonData.levels.some(level => level.uses_lab2)) {
```
sanchitmalhotra126, line 110: Added a change to only setup the browser navigation handler for progressions with lab2 levels. This is because a few UI tests were failing due to the "levelId" being added to the browser history stack.

For context: the UI test step for "I get redirected to" [checks that the state is `null`](https://github.com/code-dot-org/code-dot-org/blob/0d41b2aa59eb58abcd6ddc9ef092ae9e1e9d9e5f/dashboard/test/ui/features/step_definitions/steps.rb#L1317), but I feel like that may not be necessary? In any case though, I figured this would be an easier fix, rather than reworking core UI test steps for now.

apps/src/code-studio/browserNavigation.js
fisher-alice: +1

apps/src/lab2/utils/LifecycleNotifier.tsx
```diff
@@ -0,0 +1,49 @@
...
+class LifecycleNotifier {
+  private listeners: {
+    [event in LifecycleEvent]?: ((...args: callbackArgs[event]) => void)[];
+  };
+
+  constructor() {
+    this.listeners = {};
+  }
+
+  addListener<T extends LifecycleEvent>(
```
fisher-alice, line 34: Just curious - where is this function used?

sanchitmalhotra126: Nowhere yet! Will be used in an upcoming PR for analytics reporting.

fisher-alice: Oh gotcha - thanks!

PR #59857, opened 2024-07-18
Refactors use of Phaser's isLoading to prevent flaky tests

wilkie:
This `Safari_hour_of_code_hour_of_code_finish` test will repeatably fail:

```
dashboard UI tests failed with Safari_hour_of_code_hour_of_code_finish (0:50 minutes), retrying (1/5, flakiness: 0.25)...
dashboard UI tests failed with Safari_hour_of_code_hour_of_code_finish (0:22 minutes), retrying (2/5, flakiness: 0.25)...
dashboard UI tests failed with Safari_hour_of_code_hour_of_code_finish (0:24 minutes), retrying (3/5, flakiness: 0.25)...
dashboard UI tests failed with Safari_hour_of_code_hour_of_code_finish (0:21 minutes), retrying (4/5, flakiness: 0.25)...
```

All cucumber test instances show the same behavior... it fails quickly with a JavaScript issue:

image

This JavaScript error seems to indicate that some part of that chain has not been initialized yet. Looking at the code for the various Minecraft labs, the `Craft` class initiates the loading of the Phaser application shortly after the page initializes. However, the chain might not be there immediately when the "`Then I wait until the Minecraft game is loaded`" step happens.

When you look at SauceLabs, you will see it doesn't quite load the page very quickly which is a second piece of evidence.

The code does try to allow this, but not every Minecraft lab has the logic. You'll see from the diffs that many have this check:

```
return (
  Craft.gameController &&
  Craft.gameController.game &&
  Craft.gameController.game.load &&
  !Craft.gameController.game.load.isLoading
);
```

But one (`simple/craft.js`) has:

```
return (
  Craft.gameController &&
  Craft.gameController.game &&
  !Craft.gameController.game.load.isLoading
);
```

Which will likely fail if `game` is there but somehow `game.load` isn't.

When you look at the [Phaser docs](https://phaser.io/docs/2.6.2/Phaser.Loader.html#hasLoaded) there is a `hasLoaded` function which seem much more reliable and lets us write the function in the affirmative instead of negating. This lets us use the more concise operator-chaining form. When you look at the Phaser code, `isLoading` and `hasLoaded` will be set at the same time to opposite values. (They are only both `false` initially.)

**Reason**: The reason this test is flaky must be that `game.load` might be `null`.

When the Minecraft lab is started, it will create a `GameController`. This is defined in our `craft` repository [here](https://github.com/code-dot-org/craft/blob/main/src/js/game/GameController.js).

That `craft` repo will immediately create the `game` property specifically [here](https://github.com/code-dot-org/craft/blob/65ccc4e9428c6ccac865ba056f97ea6eb69ee6e7/src/js/game/GameController.js#L51C5-L51C34).

This means that `Craft.gameController.game` will be valid every time `Craft.gameController` is valid... which means that the second `?.` in `Craft.gameController?.game?.load` is redundant, but not going to hurt.

Now, is it possible for `Craft.gameController.game` to be valid but `...game.load` to not? That requires us to look at `Phaser`.

In Phaser, `Game` has a `load` property initially set to `null` as seen [here](https://github.com/phaserjs/phaser/blob/dd39f9ab08d57fa1bacd1287ccadb03fb3151267/src/core/Game.js#L176).

The `load` property is set only when the `boot` is called on the `Game` as seen [here](https://github.com/phaserjs/phaser/blob/dd39f9ab08d57fa1bacd1287ccadb03fb3151267/src/core/Game.js#L553).

The game isn't booted until the Phaser state manager does so... it will not boot the game until the game loop starts and will just queue state changes (and the 'start' is a state that gets queued) until that happens. It is unclear when the game loop starts exactly, but it is enough evidence to support the idea that Phaser likely waits until some of its own assets load or the page is in some state until it starts the game loop (which in turn creates the loader we require.)

The flakiness is therefore us racing Phaser's state management and the game loop.

## Solution Explanation

The `isLoading` parameter is strange. It starts out `false` on a page load and therefore our `phaserLoaded` function returns `true` even when the assets are not loaded. So, replacing `!...isLoading` with `hasLoaded` seems correct, yet, it will fail our integration tests that run those levels because they rely on `phaserLoaded` being `true` even before Phaser has "loaded".

Honestly, we do not care if phaser is currently loading... we care that the Phaser libraries exist. So, we can replace:

```
return (
  Craft.gameController &&
  Craft.gameController.game &&
  Craft.gameController.game.load &&
  !Craft.gameController.game.load.isLoading
);
```

with simply:

```
return !!Craft.gameController?.game?.load;
```

since that `load` is a property of an initialized `Phaser.Game` class.

Again, because we never actually care about the case where `isLoading` is `true`, these two statements are equivalent in function.

To be safe, if we really really care about capturing the `isLoading` case, we would need:

```
return Craft.gameController?.game?.load && !Craft.gameController.game.load.isLoading;
```

But, interestingly, I think this is a mistake. If there is ever a situation where `isLoading` is `true` (whenever Phaser is loading some asset), the run button would not work... even though from what I can tell about Phaser and our use of it, it shouldn't matter. Basically, Phaser is fully capable of scheduling the runtime around asset loading to make it feel seemless.

I think the diff provided is the correct intention of the code.

## Links

- jira ticket: [LABS-928](https://codedotorg.atlassian.net/browse/LABS-928)

## Testing story

Will run with `[test all browsers]`.

## PR Checklist:

- [ ] Code is well-commented


wilkie:
The levels integration tests timeout for some reason... so I guess I broke something even though my manual testing of those levels work.

```
1844 | ✖ Craft walk above action plane
1845 | Chrome Headless 125.0.6422.141 (Linux x86_64)
1846 | Error: Timeout of 20000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
1847 |  
1848 | ✖ Craft fall into hole
1849 | Chrome Headless 125.0.6422.141 (Linux x86_64)
1850 | Error: Timeout of 20000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
1851 |  
1852 | ./craft/movement.js
1853 | ✖ walk forward next to sheep
1854 | Chrome Headless 125.0.6422.141 (Linux x86_64)
1855 | Error: Timeout of 20000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
1856 |  
1857 | ✖ walk forward turn right
1858 | Chrome Headless 125.0.6422.141 (Linux x86_64)
1859 | Error: Timeout of 20000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
1860 |  
1861 | ✖ walk forward with too few blocks to reach sheep
1862 | Chrome Headless 125.0.6422.141 (Linux x86_64)
1863 | Error: Timeout of 20000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
```

**Update**: Yes, these tests and the intention of the run button rely, confusingly, on the `phaserLoaded` function returning `true` before Phaser has started "loading" game assets.

That is, the `phaserLoaded` function returns ultimately `!isLoading` and `isLoading` is also `false` when it hasn't started to load anything.

wilkie:
Local runs of those integration tests also pass. Strange.

wilkie:
Can't get all tests to pass because of other flakiness.

However, all unit and integration tests and relevant UI tests pass:

```
[hour_of_code/hour_of_code_finish] UI tests for Safari_hour_of_code_hour_of_code_finish succeeded (2:29 minutes, 3 passed)
```

fisher-alice (approved):
Thanks for the detailed explanation of your investigation!

I merged `staging` since this was first opened a while back, and then merged with latest `staging` again after https://github.com/code-dot-org/code-dot-org/pull/61801 was merged (with the check that `Craft` is defined before calling `phaseLoaded` in mind).

PR #59861, opened 2024-07-18
[Google Blockly] add translations for controls_for_counter block

mikeharv:
The `controls_for_counter` block is a custom version of the loop that is hard-coded to use a "counter" variable, as opposed to separate variable dropdown field:

image
 
For standard blocks provided by Blockly, translations are also provided. For custom blocks, we define the the block strings ourselves. For CDO Blockly, we would define custom block strings right in the CDO Blockly repo:

https://github.com/code-dot-org/blockly/blob/v4.1.0/i18n/locales/en-US/core.json#L45-L46

For more recently created blocks, and all custom block text in Google Blockly labs, these strings are in this main repo. if you scan the edited `turtle/blocks.js` file, you will notice lots of strings provided with `msg.*()` instead of `blockly.Msg.*`.

For these strings, I've chained the old and new string references with `||` for a couple reasons. The new English strings would be shown to all users until it can be translated. Until then, it makes sense to keep preference for the old dictionary reference since Artist isn't migrated yet. Once it is migrated, the old dictionary entry will be unavailable and we will fall back to whatever translations of the new string are available.

 I've double-checked that all remaining strings using the Blockly `Msg` dictionary  in this file are also present in mainline Google Blockly's dictionary. This means we shouldn't need to provide redundant translations for them.  Furthermore, no other blocks appear to be missing any expected text when flipping Blockly versions locally.

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/i18n/turtle/en_us.json
+++ b/apps/i18n/turtle/en_us.json
@@ -42,6 +42,8 @@
   "drawShapeWithSideLength": "draw shape with side length",
   "drawSticker": "draw sticker",
   "drawStickerWithSize": "draw sticker with size",
+  "controlsForInputFromToBy": "from %1 to %2 count by %3",
+  "controlsForInputWith": "for",
   "forward": "forward",
   "heightParameter": "height",
   "hideTurtle": "hide artist",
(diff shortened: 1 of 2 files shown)


breville (approved):
Side note: in the future, would it be good to translate this as a single string that includes the `for` as well?

mikeharv:
> Side note: in the future, would it be good to translate this as a single string that includes the `for` as well?

Possibly? These two label fields are currently separated by another with the text "counter" for the variable. I'm not sure what advantage there was at the time to three separate fields. For this migration, I'm trying to be as minimally intrusive as possible.

PR #59862, opened 2024-07-18
Fix comments on import/order

lfryemason:
Follow up to https://github.com/code-dot-org/code-dot-org/pull/59779

## PR Checklist:

- [ ] Code is well-commented


Hamms (approved):
Thank you for the follow-up! Could we also apply a similar change to `apps/src/p5lab/P5Wrapper.js` and some of the other places we've added comments that explain the "what" of what we're doing but don't go into detail on the "why"?

PR #59871, opened 2024-07-18
Fix Aichat session matching bug

fisher-alice:
This PR fixes a logging bug for the `AichatSessions` table. Currently, a new session is created whenever a new message is added to stored messages. This results in multiple table entries with duplicate messages. A new session should be created only when the chat history is cleared or model customizations are updated which resets the conversation history. This regression is due to the removal of a frontend util function which stripped extra fields from chat messages. https://github.com/code-dot-org/code-dot-org/pull/59341/files

The bug was resolved by updating the `matches_existing_session` method in `aichat_controller` so that only the message fields 'role', 'status', and 'chatMessageText' are compared. 

## Before update
Note that the session id is incremented by one in the third request although the model customizations were not reset.

https://github.com/user-attachments/assets/66699155-8ab6-4e52-9c09-34c2b7e23b8f

## After update

Session id remains the same between 2nd and 3rd requests as expected.

https://github.com/user-attachments/assets/9c9abdaf-2df9-4ded-a829-706010858cbe

## Links
[jira](https://codedotorg.atlassian.net/browse/LABS-933)

## Testing story
I added a unit test that checks if another field is updated from frontend stored messages (like `timestamp`), the stored messages are still considering matching as long as `chatMessageText`, `role`, and `status` values match.

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/dashboard/app/controllers/aichat_controller.rb
+++ b/dashboard/app/controllers/aichat_controller.rb
@@ -128,9 +128,15 @@ class AichatController < ApplicationController
       end
     end
 
-    if params[:aichatModelCustomizations] != JSON.parse(session.model_customizations) ||
-        params[:storedMessages] != JSON.parse(session.messages)
+    if params[:aichatModelCustomizations] != JSON.parse(session.model_customizations)
       return false
+    end
+    
+    # Compare stored messages excluding timestamps
+    sessions_stored_messages = JSON.parse(session.messages).map { |message| message.except('timestamp') }
+    frontend_stored_messages = params[:storedMessages].map { |message| message.except('timestamp') }
+    if sessions_stored_messages != frontend_stored_messages
+      return false      
     end
 
     true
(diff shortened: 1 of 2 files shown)


dashboard/app/controllers/aichat_controller.rb
```diff
@@ -128,10 +128,16 @@ def chat_completion
...
     end
 
-    if params[:aichatModelCustomizations] != JSON.parse(session.model_customizations) ||
-        params[:storedMessages] != JSON.parse(session.messages)
+    if params[:aichatModelCustomizations] != JSON.parse(session.model_customizations)
       return false
     end
+    
+    # Compare stored messages excluding timestamps
+    sessions_stored_messages = JSON.parse(session.messages).map { |message| message.except('timestamp') }
```
bencodeorg: Thanks for doing this! Could we be explicit about the keys we want to include (rather than excluding)? Thinking something like:

`JSON.parse(session.messages).map{ |message| message.slice('chatMessageText', 'role', 'status') }`

I can do this as a follow-up as well, but would be great to add a test that an extra key (say, a timestamp 😬 ) doesn't generate a new session.

bencodeorg (approved):
Not sure if you're looking for active feedback but I think this is worth merging in the short term (one suggestion)!

dashboard/app/controllers/aichat_controller.rb
fisher-alice: Sure - will update!

PR #59872, opened 2024-07-18
[Aichat] Log notifications/refactor (not to be merged)

fisher-alice:
Note that this PR is not to be merged as-is, but a starting point for discussing whether we should refactor how we log Aichat data on the backend.

## Reflection
After this initial work in adding notifications to logging and reflection, I think that it would be worth the effort in updating the Aichat table especially now that we are logging notifications because we will be implementing the teacher view of chat history soon (and maybe eventually student view as well).

I believe the original impetus for grouping messages in sessions is that we wanted to indicate to teachers which messages were contained within a conversation history with the chatbot. Thus, if a teacher requested student chat history, we could give them a list of user/assistant messages with the session id indicating which messages were part of the same conversation.

Since then, we have received a [clear design](https://www.figma.com/design/tjkMqlAXi2iTqpUHSljB3A/Gen-AI-Unit?node-id=0-1&t=7tLuOglW2k3cxP3p-0) of how we’ll be implementing teacher view of student chat history, and Dan has shared how a teachers would be assessing student work in [this Slack thread](https://codedotorg.slack.com/archives/C06FELVTV0Q/p1719419287543449?thread_ts=1719356890.606439&cid=C06FELVTV0Q):

“from an assessment perspective, student updates to the chatbot should appear in the chat history so a teacher can tell when students made updates and how they responded to those updates. For example, if the level says to test out a message, then update the prompt so the chatbots responses are [X] or [Y], then the teacher would need to see where in the log a student made that update.”

Based on Dan’s comment, it may be more important for a teacher to be able to view how a student responds to assistant messages via updating model customizations or the model card information than necessarily when a session begins or ends. So it seems like logging all student updates to the model via notifications would be a good call. I'm thinking of the scenario when the student is working on the model card information and may save/update several times and not necessarily submit a chat message. With notifications being displayed in the student chat history, a teacher will know when a new ‘session’ begins. In the case of a student clearing chat messages, a hidden clear chat notification can be added (included this draft PR).

Also, for a couple of the fine-tuned models (Arithmo and Karen), they don’t accommodate conversation history; they require a single request/response format. So that the 'session' concept doesn't apply for these models. And we’ll be most likely adding more fine-tuned models in the future, some which may also not handle history.

Managing sessions adds complexity/logic to the code on both the frontend and backend. A bug was introduced within the checking of whether a new session should be created or an existing session should be updated when fields were added in chat messages. See https://github.com/code-dot-org/code-dot-org/pull/59871. When notifications are added, the chat message fields are different from those of user/assistant messages (‘chatMessageText’ vs ‘text’) so that more logic will need to be added to accommodate these changes.
Also - if all notifications are logged including when model updates are saved, and multiple updates are saved at the same time, then a ‘session’ would include multiple consecutive model update notifications which seems a bit arbitrary. This leads to a question of whether each model update notification should begin a new session, etc. 

We discussed earlier this week that an alternative could be that the logging could be similar to how chat items are stored on the frontend - log each individual chat item. I think this approach may make more sense, especially to folks looking at the codebase in the future as it would simplify the code.
But open to discussion!

## Description
This PR would lay the groundwork for allowing teacher view of student chat history including model update and other notifications as requested by curriculum/product.
Currently, we only log chat messages from the user and assistant on the backend in the `AichatSessions` table, and do not log notifications or model updates.

This PR adds logging of these other Chat Items and includes a bit of refactoring of types.
On the frontend:
- Since model updates are a type of notification, I removed the `ModelUpdate` type as this simplified the code a bit. I first thought of doing this since the `Notification` displayed when `getResetModelNotification` is called resets the model customizations and is a model update.
- Instead of removing model updates, I added an optional field titled `hidden` so that we can hide notifications from display.
- I added a hidden clear chat history notification when a user clicks on the 'Clear chat' button since this resets the session and may be useful information for a teacher when viewing student chat history.

On the backend:
- I added another endpoint in`aichat_controller` when a user updates/saves model customizations. The reason is that a user could update the model card information and update the model customizations several times before submitting a chat message. This user history would NOT be captured if we only log notifications when submitting a chat message.

## After update
Screenshots:

log_notifications - payload
Screenshot 2024-07-18 at 4 52 59 PM

log_notifications - response: 
Screenshot 2024-07-18 at 4 53 07 PM

chat_completion - payload:
Screenshot 2024-07-18 at 4 54 17 PM

Screencast video showing dev console request to both `chat_completion` and `log_notification` (payload and response) when a user updates model customizations and when a user submits a chat message.
Notice that notifications are now included in `storedMessages` that is sent to the backend.

https://github.com/user-attachments/assets/92e0023e-3044-40a6-9bc7-7a2365a83af6

Screenshot of 'messages' column from `AichatSessions` table which includes notification messages:
Screenshot 2024-07-18 at 5 08 25 PM

## Links
[jira](https://codedotorg.atlassian.net/browse/LABS-798)
[Slack thread discussion](https://codedotorg.slack.com/archives/C06FELVTV0Q/p1719356890606439)

## PR Checklist:

- [ ] Code is well-commented


bencodeorg:
Wondering if this could be split up a bit...could the `log_notifications` updates be separated out?

fisher-alice:
> Wondering if this could be split up a bit...could the `log_notifications` updates be separated out?

Oh yeah, definitely will do. But wanted to implement this with existing sessions table first and discuss with folks before proceeding forward. Lmk what you think!
~~At this point, I'm not feeling strongly either direction (keeping table as-is or refactoring). So curious what you and others think.~~ cc: @sanchitmalhotra126

bencodeorg:
Got it, will take a closer look and give overall thoughts tomorrow!

fisher-alice:
> Got it, will take a closer look and give overall thoughts tomorrow!

After some reflection, I do think it may be worth refactoring the table. I updated the PR description with some thoughts. Thanks!

PR #59877, opened 2024-07-18
Datablock Storage: shared projects work for unlogged-in users

### LINKED ISSUE #59339: Signed out users can't access datablock storage on projects they don't own
see slack https://codedotorg.slack.com/archives/C05DMS18MEX/p1718997900072409

snickell:
Allow shared student projects to access limited methods on datablock_storage_controller without requiring login. This allows projects that use data blocks to be shared, which matches the firebase design.

We only allow access to the methods that are actually invoked by data blocks.

Fixes #59339 


cnbrenci (approved):
get it girl

PR #59879, opened 2024-07-18
Datablock Storage: only allow up to 3x concurrent fetch() connections from data blocks

### LINKED ISSUE #58766: Datablock Storage: limit blocks to 5 concurrent IO operations (fetches)
We'd like to limit the number of parallel datablock storage requests a student can issue against our server, to limit the chances of impacting other students.

snickell:
We rate limit students to 600 datablock storage requests in 60s. After this limit, requests are dropped, and a warning message is raised. This provides a nice hard limit.

In addition, we'd like to "soft limit" the number of concurrent requests they can hold agains the server at any given time. With HTTP/1, browsers tended to limit requests to 6 at once, but with HTTP/2 (which we use) a browser will permit HUNDREDS of concurrent requests. This allows a single student who writes an infinite loop to significantly impact database availability for up to 60 seconds.

We don't want to impact use of the datasets UI, so we bottleneck access from within commands.js (the block specific bindings to the storage layer) rather than in the storage layer itself.

Fixes https://github.com/code-dot-org/code-dot-org/issues/58766

### DIFF
--- a/apps/jest.config.js
+++ b/apps/jest.config.js
@@ -234,7 +234,7 @@ const config = {
 
   // An array of regexp pattern strings that are matched against all source file paths, matched files will skip transformation
   transformIgnorePatterns: [
-    'node_modules/(?!vmsg|@code-dot-org/johnny-five|@code-dot-org/js-interpreter|blockly/core)',
+    'node_modules/(?!vmsg|@code-dot-org/johnny-five|@code-dot-org/js-interpreter|blockly/core|p-queue)',
   ],
 
   // An array of regexp pattern strings that are matched against all modules before the module loader will automatically return a mock for them
--- a/apps/jest.config.js
+++ b/apps/jest.config.js
@@ -26,7 +26,7 @@ Object.entries(APPLICATION_ALIASES).forEach(([alias, localPath]) => {
 });
 
 /**
- * List of node_modules to transform with babel-jest (default is not to transform)
+ * List of node_modules to transform with babel-jest.
  *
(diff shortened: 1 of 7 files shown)


cnbrenci (approved):

snickell:
Adding @stephenliang to take a look at the jest.config.js changes required to use the `p-queue` and `p-timeout` packages (which use ESM imports/exports in its distributed JS, which jest doesn't like).

stephenliang (approved):

apps/jest.config.js
```diff
@@ -25,6 +25,21 @@ Object.entries(APPLICATION_ALIASES).forEach(([alias, localPath]) => {
   jestAliases[`${alias}/(.*)`] = `${localPath}/$1`;
 });
 
+/**
+ * List of node_modules to transform with babel-jest (default is not to transform)
```
stephenliang: default is to transform, right?

snickell: hoo boy, we might have a double negation at play here, and its ambiguous which layer I'm speaking of 😆 

From the perspective of "node modules", and the `nodeModulesToTransform` property this is a comment on, we've configured our default to be "do not transform" (i.e. we don't apply babel to node_modules/ packages unless they're explicitly in this new list).

On the other hand, this property is used to configure the jest property `transformIgnorePatterns`, where the default is "transform unless listed".

snickell: I just removed the `default is not to transform` language, since it didn't really add anything and causes confusion.

PR #59884, opened 2024-07-19
Big playspace: constrain playspace to browser size

breville:
This builds on the experiment by @bencodeorg in https://github.com/code-dot-org/code-dot-org/pull/59145.  The size constraints have been adjusted and, most notably, resizing the browser now causes the size constraints on the playspace to be recalculated.

### Before

Before this change, some responsive CSS would attempt to size the playspace appropriately for the browser size, but as soon as the playspace was manually resized, this CSS would be overridden and no longer do anything.  

And presumably since the introduction of labs, resizing the playspace would not be constrained to the browser size, and resizing the browser would not recalculate the constraints on the playspace size.  As a result, making the window shorter could hide the Run button, and making the playspace too wide could also hide the Run button.  

All of that is shown here:

https://github.com/user-attachments/assets/94f28769-a739-401e-a284-a8c97d34eb1b

### After

After this change, and with the experiment turned on, the responsive CSS does the initial layout, but nothing after that.

Resizing the browser causes the size constraints on the playspace to be recalculated, and the playspace might indeed shrink.  And resizing the playspace will be constrained so that it can't go big enough to hide the Run button.  

All of that is shown here:

https://github.com/user-attachments/assets/b6343ee6-6af8-4087-8218-a8e07493e0f5

The calculation of the maximum playspace width has been slightly updated from the previous implementation.  Rather than using `75%` of the window's inner height, it uses `(100% - 160px)` of the window's inner height, assuming that other elements (like the header) will remain fixed height regardless of the window size.

That said, `160px` of additional "padding" might not be optimal, and so an additional URL parameter may be provided to set that.  In the video above, we used `bigPlayspacePadding=220`.

The maximum width of the big playspace has been set to `50%` of the window's inner width.

Some care was needed to avoid an infinite recursion between `onResize` and `resizeVisualization` since they can trigger each other.

The experiment continues to work for Sprite Lab levels, but now works for Dance Party levels too.

Finally, the `bigPlayspace` experiment is still being used, but it can now also be enabled for a single page by adding the `bigPlayspace=1` URL parameter.

## Further improvements

An ideal layout system for these labs could make more significant improvements, such as:
- a single source of truth for layout, versus a combination of responsive CSS, vanilla JS, and React.
- unified prioritization of components, so that core components are guaranteed to be on screen, and components that don't fit are collapsed/scrollable appropriately.
- responsive layout that takes into account not just aspect ratio but also number of pixels given to different areas, with scaling used when appropriate to still show sufficient content on small displays.
- support for a wider range of aspect ratios, including portrait orientation.


mikeharv:
Seems like a definite improvement! I wonder if, while we're doing this, we'd want to try to prevent the Finish button from being hidden behind the language dropdown. I think there's an argument that could be made that it is just as important as the Run button in terms of allowing students to progress between levels.

breville:
> Seems like a definite improvement! I wonder if, while we're doing this, we'd want to try to prevent the Finish button from being hidden behind the language dropdown. I think there's an argument that could be made that it is just as important as the Run button in terms of allowing students to progress between levels.

Interesting.  We can certainly learn more from this experiment about the appropriate padding to add to each level type to avoid an overlap.  It sounds like we could rethink the display of the language dropdown in general.  But I think we can do each of them without further changes to this particular PR?

apps/src/StudioApp.js
```diff
@@ -98,15 +98,26 @@ var codegen = require('./lib/tools/jsinterpreter/codegen');
...
+
 const MIN_WIDTH = 1400;
 const DEFAULT_MOBILE_NO_PADDING_SHARE_WIDTH = 400;
-export const MAX_VISUALIZATION_WIDTH = experiments.isEnabled(
-  experiments.BIG_PLAYSPACE
-)
-  ? 0.75 * window.innerHeight
-  : 400;
+const DEFAULT_VISUALIZATION_WIDTH = 400;
+export const MAX_VISUALIZATION_WIDTH = 400;
```
ebeastlake, line 120: Is this still used somewhere? I don't see where it's imported in this PR.

apps/src/StudioApp.js
```diff
@@ -98,15 +98,26 @@ var codegen = require('./lib/tools/jsinterpreter/codegen');
 var copyrightStrings;
 
 /**
- * The minimum width of a playable whole blockly game.
+ * Store experiment parameters.
  */
+const isBigPlayspaceExperiment = experiments.isEnabledAllowingQueryString(
```
ebeastlake: nit: `isBigPlayspaceEnabled`? Or `isBigPlayspaceExperimentEnabled`?

apps/src/StudioApp.js
```diff
@@ -98,15 +98,26 @@ var codegen = require('./lib/tools/jsinterpreter/codegen');
...
+  experiments.BIG_PLAYSPACE
+);
+const bigPlaySpacePadding = queryParams('bigPlayspacePadding') || 160;
```
ebeastlake, line 106: Would you need to reload the page to modify the padding here? If so, and that's undesirable, could we make this a function like `getMaxResizableVisualizationWidth` to fix that?

ebeastlake (approved):

ebeastlake:
Some minor feedback, but generally, this seems like a nice improvement! 🎉

sanchitmalhotra126 (approved):
LGTM!

apps/src/StudioApp.js
breville: Yes, Dance Party still uses it here.  
https://github.com/code-dot-org/code-dot-org/blob/3d4b6c1e827fd8a3a46c2e3d237ada73f2c5d5ef/apps/src/dance/constants.ts#L1-L7

apps/src/StudioApp.js
breville: Good call.  Let's use `isBigPlayspaceEnabled` since it can still apply if this graduates from being an experiment.

apps/src/StudioApp.js
breville: For now, yes.  The idea is that we can try different values in different situations to see what works best, before we move this out of being an experiment.  I'm not quite sure how a function would change this?

ebeastlake: While we're experimenting, my original comment doesn't matter. It seems good for now!

PR #59892, opened 2024-07-19
Remove advocacy.code.org from pegasus router and other misc deprecations

kelbyhawn:
Removes advocacy.code.org from the pegasus router and also removes other advocacy-related things from the pegasus directory.

## Links
Jira ticket: [ACQ-2087](https://codedotorg.atlassian.net/browse/ACQ-2087)

## Testing story
Making sure drone passes before merge


bethanyaconnor (approved):
🎉

PR #59898, opened 2024-07-19
Add Music Lab libraries to translation pipeline

bencodeorg:
Adds Music Lab "libraries" to the localization pipeline. A library includes the set of instruments (for the "Play Notes" block), "kits" (for the "Play Drums" block), "packs" (for the "Play Sounds" block, hopefully this is the right name -- these are often artist-specific), and individual sounds within each of these to the localization pipeline. More detail on the general approach in the dev doc linked below, but it is heavily inspired by [our localization approach for the "animation manifest"](https://github.com/code-dot-org/code-dot-org/tree/staging/bin/i18n/resources/apps/animations) that is used in Sprite Lab and Game Lab. These changes get the appropriate strings into the translation pipeline in order to get translations generated, but does not consume them.

One bit of complexity is that we have a new version of our `launch2024` library in the works, currently named `launch2024-v2`. We'll eventually substitute the contents of `launch2024-v2` for `launch2024`, so the changes here support a mapping between the v2 version of the library (where we source the strings) and an output file without the v2 suffix. The `launch2024-v2` file also has a change that we need to avoid localizing song lyrics and artist names (a `skip_localization` key) that isn't present in the existing `launch2024` file.

## Links

- Jira: https://codedotorg.atlassian.net/browse/LABS-614
- [Dev doc](https://docs.google.com/document/d/1nXTflYDo-VMca0PF0osKNvzjO4BB1-y-TE9VpgwxN3g/edit)

## Testing story

I've been using our test instance in CrowdIn to test these changes -- in addition to adding unit tests in line with what exists for the animation manifest, I've manully tested the full pipeline manually, which is comprised of:
- "sync in": extracting strings from the libraries to translate -- generates JSON files of strings to be translated.
- "sync up": move these strings to our test CrowdIn instance for translation.
- (manual steps in CrowdIn to simulate translation of strings)
- "sync down": copy down translations to my machine.
- "sync out": copy translations to the S3 bucket (note: I've stored some of these fake translations at `cdo-curriculum/media/musiclab-test...`, but exclude the `-test` suffix in this PR).

## Follow-up work

I have a (hopefully small) separate PR that I'll follow up with that makes changes to our Music Lab front-end code to consume these translations.


bin/i18n/resources/apps/musiclab_libraries/sync_in.rb
```diff
@@ -0,0 +1,67 @@
...
+            manifest = JSON.parse(
+              AWS::S3.download_from_bucket(
+                'cdo-curriculum',
+                "media/musiclab/#{library_filename}"
+              )
+            )
+
+            strings_to_translate = {}
+
+            manifest['instruments'].each do |instrument|
```
bencodeorg, line 34: I should note that a vast majority of the changes in this PR are boilerplate/tests/largely inspired by the animation manifest translation approach -- this (hopefully not too complicated) bit is the main logic that's new.

bin/i18n/resources/apps/musiclab_libraries.rb
```diff
@@ -0,0 +1,40 @@
...
+    module Apps
+      module MusiclabLibraries
+        DIR_NAME = 'musiclab_libraries'
+        # Allows temporarily sourcing input strings from a new version of our Music Lab
+        # launch library (music-library-launch2024-v2), which will eventually
+        # replace music-library-launch2024.
+        # Once we've overwritten the contents of music-library-launch2024 with
+        # what's in music-library-launch2024-v2, we can remove this
+        # and replace it with a simple array of library filenames
+        # (ie, %w[music-library-intro2024 music-library-launch2024-v2])
```
breville: I think it might be worth keeping the mapping system in place, since it's quite possible we'll have future library updates that we want to pre-translate like this.

bin/test/i18n/resources/apps/musiclab_libraries/test_sync_in.rb
```diff
@@ -0,0 +1,31 @@
...
+  end
+
+  describe '#process' do
+    let(:process) {described_instance.process}
+
+    let(:i18n_source_file_path_intro) {CDO.dir('i18n/locales/source/musiclab_libraries/music-library-intro2024.json')}
+    let(:i18n_source_file_path_launch_v2) {CDO.dir('i18n/locales/source/musiclab_libraries/music-library-launch2024.json')}
+
+    it 'prepares the i18n source file' do
+      described_instance.expects(:get_translation_strings).returns({'lady_gaga_born_this_way/vocals_verse_3': 'Verse 3'}).twice
```
breville: Can we do this with a sound in one of our default packs, rather than a commercial artist one?

bin/i18n/resources/apps/musiclab_libraries/sync_in.rb
breville: Not necessary for this PR, but it could be interesting to think about how much could be factored into common code, especially if we decided to use S3 to store translations for other situations.

bin/i18n/resources/apps/musiclab_libraries.rb
bencodeorg: Cool -- tangentially, do you have any concern about the stability of the v2/other in development files? I guess my concern about using this approach longer term is whether we might end up with partial/bad content in the translation pipeline as we're doing active development on a new library.

bin/test/i18n/resources/apps/musiclab_libraries/test_sync_in.rb
bencodeorg: Sorry this is just an overly detailed example, this stub could be anything. I can put something more generic to make that more clear.

bin/i18n/resources/apps/musiclab_libraries.rb
breville: We're fairly careful about this file, testing it regularly as we add content to it.  Certainly we might tweak some strings before release, but I think we can deal with those cases as they arise, similar to how existing loc'ed strings sometimes get changed.

breville (approved):

bin/test/i18n/resources/apps/musiclab_libraries/test_sync_in.rb
```diff
@@ -0,0 +1,31 @@
...
+
+      process
+
+      assert_equal %Q[{\n  "lady_gaga_born_this_way/vocals_verse_3": "Verse 3"\n}], File.read(i18n_source_file_path_intro)
```
breville: I'm not sure whether this should pass because the intro library does not have this sound in it.

breville: Oh, we stub the return value [here](https://github.com/code-dot-org/code-dot-org/pull/59898/files#diff-bfc8266100987232892051139758936c76e4c9b099d608c245903a67251270b2R23).  All good.

bin/test/i18n/resources/apps/musiclab_libraries/test_sync_in.rb
sanchitmalhotra126: Looks like this is stubbing the `get_translation_strings` function right? Any way we can also test the logic inside that function to make sure we're grabbing the right values and computing IDs correctly?

sanchitmalhotra126 (approved):
Looks great! Nice solution and thorough E2E testing! Just one comment on an additional unit test, but that can also come as a follow-up.

bin/test/i18n/resources/apps/musiclab_libraries/test_sync_in.rb
bencodeorg: Yep that's right, and a good callout -- I was torn on whether it was simple enough to skip testing the guts of the function...I'll try and circle back and add a couple if I have a bit of time.

PR #59903, opened 2024-07-20
Revert "Datablock Storage: only allow up to 3x concurrent fetch() connections from data blocks (#59879)"

snickell:
Reverts #59879

(Generated by a manual `git revert commit aedc5b014e4c604a721cc67334118efaddaa46e8`)

This appears to have generated test failures, see: https://codedotorg.slack.com/archives/C0T0PNTM3/p1721436717644599?thread_ts=1721432041.055659&cid=C0T0PNTM3

`node_modules/p-queue/index.js` contains a variety of "modern JS", including in this case use of the # characters in its to denote javascript private fields (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_properties). I believe `#pending = 0;` is being minified as "#p", and its causing issues in Safari (?).

### DIFF
--- a/apps/jest.config.js
+++ b/apps/jest.config.js
@@ -25,21 +25,6 @@ Object.entries(APPLICATION_ALIASES).forEach(([alias, localPath]) => {
 });
 
 /**
- * List of node_modules to transform with babel-jest.
- *
- * Packages that include ESM import/export statements in their distributed JS
- * will need to be added to this list.
- */
-const nodeModulesToTransform = [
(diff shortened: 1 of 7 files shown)


breville (approved):
Thanks for this!  Discussion [here](https://codedotorg.slack.com/archives/C0T0PNTM3/p1721437458566339?thread_ts=1721432041.055659&cid=C0T0PNTM3).

PR #59915, opened 2024-07-22
Use the component library Button component in CurriculumCatalogCard

bethanyaconnor:
Use the new component library button component on CurriculumCatalogCard and ExpandedCurriculumCatalogCard! When I was making a version for the congrats page, I realized it's generally easier to work with the component library button.

I did keep the old button around for the case where we use the prop `styleAsText`. I don't think the component library button supports that but I could be wrong. Obviously happy to change if so!

Some screenshots of CurriculumCatalogCard:

Screenshot 2024-07-22 at 11 08 44 AM

Screenshot 2024-07-22 at 11 09 14 AM

Screenshot 2024-07-22 at 11 08 54 AM

~There should not be major visual diffs as a result of this change. It's possible that a couple pixels shift in eyes tests but anything more than that is unexpected.~ There may be eyes diffs to the X button on the expanded card. No other major eyes diffs are expected, though it's possible for other elements to shift a couple pixels.


apps/src/templates/curriculumCatalog/CurriculumCatalogCard.jsx
```diff
@@ -284,43 +284,41 @@ const CustomizableCurriculumCatalogCard = ({
...
               <Button
-                color={Button.ButtonColor.neutralDark}
-                type="button"
                 onClick={onQuickViewClick}
-                aria-label={quickViewButtonDescription}
+                ariaLabel={quickViewButtonDescription}
                 text={i18n.quickView()}
                 className={`${style.buttonFlex} ${style.quickViewButton}`}
+                type="secondary"
+                color="black"
```
kelbyhawn: I left a similar comment on one of Turner's PRs, but Erin Bond made a good case (https://github.com/code-dot-org/code-dot-org/pull/59924#discussion_r1687241002) for using the `buttonColors` constant on this component.

apps/src/templates/curriculumCatalog/ExpandedCurriculumCatalogCard.jsx
```diff
@@ -311,7 +311,7 @@ const ExpandedCurriculumCatalogCard = ({
                   src={recommendedSimilarCurriculum.image || defaultImageSrc}
                   alt={recommendedSimilarCurriculum.display_name}
                 />
-                <Button
+                <DeprecatedButton
```
kelbyhawn: I think you can use the [Link component](https://code-dot-org.github.io/cdo-styleguide/?path=/docs/designsystem-link--docs) in place of this!

kelbyhawn (commented):
Yay for more DSCO updates! 🎉 Left a couple comments, but should be good to go once those are updated.

apps/src/templates/curriculumCatalog/ExpandedCurriculumCatalogCard.jsx
bethanyaconnor: It looks like the Link component only supports anchor tags, right? Semantically, shouldn't this be a button?

kelbyhawn: Are these the links here?

Screenshot 2024-07-24 at 9 04 57 AM

bethanyaconnor:
In updating to use `buttonColors`, I realized that the close button on the `ExpandedCurriculumCatalogCard` was broken. I fixed it but this change might cause an eyes diff. IMO the change in this PR looks better (the button is aligned to the right), but open to push back if folks feel this PR should truly be a no-op.


Local:
Screenshot 2024-07-24 at 12 12 00 PM


Production:
Screenshot 2024-07-24 at 12 11 51 PM

apps/src/templates/curriculumCatalog/ExpandedCurriculumCatalogCard.jsx
bethanyaconnor: Yep!

bethanyaconnor: ooo I guess because it's going to a new location on the same page, it might be okay as an anchor tag 🤔

kelbyhawn: That's what I was thinking originally, it's def a trickier use case! Let's go with the anchor tag since that's how it's behaving to a user, even if a button technically could work too.

kelbyhawn:
I'm fine keeping the updated button position!

kelbyhawn (approved):
🎉

PR #59917, opened 2024-07-22
Kaitie/add names to free response

kobryan0619:
Fixes issues raised in this [revert](https://github.com/code-dot-org/code-dot-org/pull/59908) from the "show answer" button.

Here's what the show names toggle looks like:

image

And here's what the "show answer" toggle looks like:

image

Note: I would like to update this toggle to the DSCO toggle, this will be done in a later PR.

## Links
[Figma](https://www.figma.com/design/i3smpK0bu5Tnc1JgmmMfTI/Check-For-Understanding-Summaries?node-id=2345-108&m=dev)

Note: Once merged in, I will let the DOTD know and go through any eyes tests with them together.


apps/src/templates/levelSummary/SummaryResponses.jsx
```diff
@@ -115,31 +131,46 @@ const SummaryResponses = ({
...
+                isToggledOn={showCorrectAnswer}
+                onToggle={() => {
+                  setShowCorrectAnswer(!showCorrectAnswer);
+                }}
+                label={i18n.showAnswer()}
+                expands="summary-correct-answer"
+              />
+            </div>
+          )}
+          {isFreeResponse && DCDO.get('cfu-pin-hide-enabled', false) && (
```
lfryemason, line 156: optional nit: move under `{/* Free response visualization */}`

lfryemason: Could you include a picture when it has both toggles or will showAnswerToggle and isFreeResponse never both be true?

kobryan0619: As far as I can tell, both should never be true.  The "Answer" to Free Response is always in the "Teachers Only" section.  AND for multiple choice, we never show names.

image

lfryemason (approved):

PR #59918, opened 2024-07-22
PythonLab: draggable graph popup 

Erin007:
Follow up to https://github.com/code-dot-org/code-dot-org/pull/59517
[CT-668](https://codedotorg.atlassian.net/browse/CT-668) partial

This PR makes the graph popup in PythonLab draggable. Only one graph is able to be open at a time. 

https://github.com/user-attachments/assets/19e8782d-a3dd-4d96-8fcb-247b41093b89

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/codebridge/Console/GraphModal.tsx
+++ b/apps/src/codebridge/Console/GraphModal.tsx
@@ -1,8 +1,11 @@
-import React from 'react';
+import React, {useState} from 'react';
+import Draggable, {DraggableEventHandler} from 'react-draggable';
 
-import AccessibleDialog from '@cdo/apps/templates/AccessibleDialog';
+import CloseButton from '@cdo/apps/componentLibrary/closeButton/CloseButton';
+import i18n from '@cdo/locale';
 
-import moduleStyles from '@cdo/apps/templates/accessible-dialogue.module.scss';
+import styles from './console.module.scss';
+import dialogStyles from '@cdo/apps/templates/accessible-dialogue.module.scss';
 
 /**
  * Renders a modal that displays data visualizations from Python Lab console output.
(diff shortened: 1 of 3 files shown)


Erin007:
We don't preserve positioning between graph pop ups. If you drag graph 1 to a new position, and then open graph 2, graph 2 will be in the center rather than the position graph 1 ended up in. This seems reasonable and expected to me, but wanted to flag in case anyone thinks users will be grumpy about having to re-move subsequent graphs if they have a strong position preference. 

https://github.com/user-attachments/assets/9b83f14a-6ded-4e11-b0f5-ca4d83429625

molly-moen (approved):
This is awesome! I think having the pop-up start in the same spot is reasonable for now. One thing to maybe file a follow-up for is to answer the question of if we want any sort of border/shadow/etc on the modal to make it look more modal-y? Not a blocker here though.

thomasoniii (approved):
lgtm 🚀 

But a few notes for follow up...

- how likely is it for the user to pop out a graph, re-position it, and then pop out another graph? I think that's the use case to decide whether we need to worry about it maintaining position. If that's an "unexpected" operation, then having it move to the default location could even be beneficial (it's emphasizing it's a _new_ graph), but if the user might constantly pop open things in a row, then constant re-positioning would probably be annoying.
- What _is_ the use case for popping up the graph? If it's to position it outside of the console to compare graphs side-by-side, then I can see the immediate utility in having multiple pop ups visible at once. If that's not the use case, what is it?
- I think it should use dnd-kit instead of react-draggable, but I'm not so set on it that we need to block on it. But a follow up ticket would be keen. Or, alternatively, we should axe dnd-kit from the filetabs view and replace it with react-draggable. But codebridge is a relatively small and self contained bit of the app, so only using a single dnd library would be awesome. (or if there's a reason why we need two, we should document it)

Erin007:
@molly-moen & @thomasoniii Thanks for the feedback! I filed 2 follow up tickets: 
1.) [CT-668](https://codedotorg.atlassian.net/browse/CT-668) Graph Modal 2.0 which is basically "resolve the open questions about what changes we actually want and _then_ go make them" 
2.) [CT-704](https://codedotorg.atlassian.net/browse/CT-704) which is to consolidate down to one package for supporting dragging functionality in the UI, turns out we actually have at least 3 and as long as we're standardizing we should pick + stick across components

PR #59924, opened 2024-07-22
Accessibility: use updated Button component in PersonalProjectsTableActionsCell

Erin007:
[A11Y-97](https://codedotorg.atlassian.net/browse/A11Y-97)

Updating more buttons to use the `Button` component from the DSCO component library both for aesthetics and because the newer button is accessible. There are slight visual improvements, and the functionality remains the same. 

BEFORE 
Screenshot 2024-07-22 at 1 05 06 PM

AFTER 
Screenshot 2024-07-22 at 1 30 22 PM

Tested that saving and cancelling still work as expected. 

https://github.com/user-attachments/assets/72c9b88d-10d6-471d-a15a-a0beec3c2489

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/templates/projects/PersonalProjectsTableActionsCell.jsx
+++ b/apps/src/templates/projects/PersonalProjectsTableActionsCell.jsx
@@ -2,12 +2,12 @@ import PropTypes from 'prop-types';
 import React, {Component} from 'react';
 import {connect} from 'react-redux';
 
+import Button, {buttonColors} from '@cdo/apps/componentLibrary/button';
 import PopUpMenu, {MenuBreak} from '@cdo/apps/lib/ui/PopUpMenu';
 import i18n from '@cdo/locale';
 
 import NameFailureDialog from '../../code-studio/components/NameFailureDialog';
 import color from '../../util/color';
-import Button from '../Button';
 import FontAwesome from '../FontAwesome';
 import QuickActionsCell from '../tables/QuickActionsCell';
 
@@ -93,19 +93,20 @@ export class PersonalProjectsTableActionsCell extends Component {
         {isEditing && (
           <div>
             <Button
-              __useDeprecatedTag
               onClick={this.onSave}
-              color={Button.ButtonColor.brandSecondaryDefault}
               text={i18n.save()}
+              size="s"
+              color={buttonColors.purple}
               disabled={isSaving}
-              className="ui-projects-rename-save"
+              id="ui-projects-rename-save"
+              className={moduleStyles.buttonMargin}
             />
-            <br />
             <Button
-              __useDeprecatedTag
               onClick={this.onCancel}
-              color={Button.ButtonColor.gray}
               text={i18n.cancel()}
+              size="s"
+              type="secondary"
+              color={buttonColors.gray}
             />
           </div>
         )}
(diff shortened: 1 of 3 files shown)


apps/src/templates/projects/personal-projects-table-actions-cell.module.scss
```diff
@@ -1,3 +1,7 @@
 .xIcon {
       padding-right: 5px;
 }
+
+.buttonMargin {
+      margin-right: 5px;
```
kelbyhawn: Can you update this to `8px`? It'll fit better with the 16px grid system we've been implementing 😄
```suggestion
      margin-right: 8px;
```

apps/src/templates/projects/PersonalProjectsTableActionsCell.jsx
```diff
@@ -93,19 +93,20 @@ export class PersonalProjectsTableActionsCell extends Component {
         {isEditing && (
           <div>
             <Button
-              __useDeprecatedTag
               onClick={this.onSave}
-              color={Button.ButtonColor.brandSecondaryDefault}
               text={i18n.save()}
+              size="s"
+              color={buttonColors.purple}
```
kelbyhawn, line 99: I think you can get rid of the `buttonColors` here. This is how I've been setting the color, and it works like this on Storybook.
```suggestion
              color="purple"
```

apps/src/templates/projects/PersonalProjectsTableActionsCell.jsx
```diff
@@ -93,19 +93,20 @@ export class PersonalProjectsTableActionsCell extends Component {
...
             />
-            <br />
             <Button
-              __useDeprecatedTag
               onClick={this.onCancel}
-              color={Button.ButtonColor.gray}
               text={i18n.cancel()}
+              size="s"
+              type="secondary"
+              color={buttonColors.gray}
```
kelbyhawn, line 109: Same thing here:
```suggestion
              color="gray"
```

molly-moen (approved):
🎉

kelbyhawn (commented):
Left a couple comments, but this looks great! So excited to see updates like this 🎉

fisher-alice (approved):
I defer to @kelbyhawn for the style improvements, but love that you added the DSCO Button and now they are accessible! 🎉

apps/src/templates/projects/personal-projects-table-actions-cell.module.scss
Erin007: There's a 16px grid system in play! 🤩 This is why I love tagging folks from other teams on PRs. Thanks @kelbyhawn

apps/src/templates/projects/PersonalProjectsTableActionsCell.jsx
Erin007: 🤔 `buttonColors.purple` and `"purple"` end up being equivalent, but I think we want to use `buttonColors` since it's easier to search and update later. For example if we ever want to change "purple" to something more semantic. The Buttons [story](https://github.com/code-dot-org/code-dot-org/blob/09601303feea4a6f2a2a2b55ab51dd551b5db39e/apps/src/componentLibrary/button/Button.story.tsx#L84) uses `buttonColors` and so do the majority of `Button` usages that I see with a quick search.

kelbyhawn: Oh that's a great point! I was looking at the buttons on the Storybook site (not the code file) and it was using the `"purple"` by itself so I used that, but I'll make a note to start using `buttonColors`.

PR #59930, opened 2024-07-22
Script to allow DotD to seed curriculum content to production from staging branch

etaderhold:
This PR includes a script to seed the latest curriculum content from the staging branch, and an option in the DotD script to call that script on production-daemon.

A couple of things to be aware of/check me on:
1. I currently have the DotD script calling this new script through `production-daemon`. Is that the right system to run this sort of periodic task from, or would `production-console` or something else be more appropriate?
1. The script aims to restart the web servers in order to populate the in-memory content caches with the new curriculum content. I copied the `restart_service` line from [build.rake](https://github.com/code-dot-org/code-dot-org/blob/8180d74e376deda92eeeff9d4adc2057cd473da1/lib/rake/build.rake#L154) after tracing back from the Slack message preceding the restart during the DTP process. During the DTP we see _n_ new workers start up with the new code, and then the old _n_ workers shut down once the new ones start. I want to verify that running this same line from this new script on `production-daemon` (or whatever) will achieve that same result.

## Testing story

I tested this script on an adhoc environment. I set up the initial environment toward the end of last week.

This morning I ran `ssh -t gateway.code.org ssh -t adhoc-seed-then-restart adhoc/bin/seed-from-staging` (analogous to the command the `dotd` script will run in the production environment). The result was that content from last night's levelbuilder content scoop (such as [this change](https://github.com/code-dot-org/code-dot-org/commit/0592ec14b61d00f73941c263e3712e509a262b68#diff-7a565d5442abb41410b1eeebe7402ff98499a99da9f55e1e9265ed9d642b749c) — see it live [here](https://adhoc-seed-then-restart-studio.cdn-code.org/levels/48506)) were visible in the adhoc environment.

By the way when I ran the `bundle exec rake adhoc:start_inactive_instance RAILS_ENV=adhoc` to restart the adhoc environment after the idle weekend, requests to the dashboard web server on this environment had 502 Bad Gateway errors. Once I ran the seed/restart script the web server was back up and running as expected. I therefore have reason to believe the `restart_service` line is working fine in the adhoc context, but still want extra scrutiny on whether that's the way to get the graceful restart in the production context.

## PR Checklist:

- [ ] Code is well-commented


cat5inthecradle:
>The script aims to restart the web servers in order to populate the in-memory content caches with the new curriculum content. I copied the restart_service line from [build.rake](https://github.com/code-dot-org/code-dot-org/blob/8180d74e376deda92eeeff9d4adc2057cd473da1/lib/rake/build.rake#L154) after tracing back from the Slack message preceding the restart during the DTP process. During the DTP we see n new workers start up with the new code, and then the old n workers shut down once the new ones start. I want to verify that running this same line from this new script on production-daemon (or whatever) will achieve that same result.

We do not restart our frontend EC2 web servers during a DTP, we completely replace them with brand new servers. The code you referenced simply restarts the dashboard service on production-daemon. I'm not sure what you're referring to with 'workers' unless you're talking about delayed_job.

bin/dotd
```diff
@@ -303,6 +304,9 @@ def dotd_menu(dotd_name)
         @logger.info 'kills tests'
       end
       menu.choice(:DTP) {do_dtp(dotd_name)}
+      menu.choice(:STP) do
+        run_on 'production_daemon', 'production/bin/seed-from-staging'
+      end
```
cat5inthecradle, line 313: nitpick on naming, is 'seed from staging' more broad than what we're specifically doing?

bin/seed-from-staging
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env ruby
+
+require_relative '../deployment'
+require 'cdo/rake_utils'
+
+Dir.chdir(dashboard_dir) do
+  staging_dir = '../../staging'
```
cat5inthecradle: This code expects there to be a staging directory that is a git repository. Can you add a comment to this script describing the prerequisites here? If that directory is not present, will this simply error out with an appropriate message?

cat5inthecradle:
It's hard to say whether production-daemon or production-console is the better choice here - but DOTD commands generally run on 'console'

bin/dotd
etaderhold: I think it captures the concept pretty specifically, but am open to a better phrase.

bin/seed-from-staging
etaderhold: Yes the script would previously exit with an exception if the directory was not found. I have a better error message now.

etaderhold (commented):
> We do not restart our frontend EC2 web servers during a DTP, we completely replace them with brand new servers. The code you referenced simply restarts the dashboard service on production-daemon. I'm not sure what you're referring to with 'workers' unless you're talking about delayed_job.

By "workers" I just meant the production Rails web servers. Apologies for the imprecise terminology. Is my "application code engineer" showing? :-)

Anyway, what I meant to do here was call whatever code makes the below screenshots happen, so that our servers have caches primed with the new content. Can you give me a pointer to that? If we have something that simply restarts the servers instead of replacing them, without causing any availability issues, that should also work for this purpose.

image

---------------------
image

cat5inthecradle:
We don't have a regular use case of restarting either the EC2 web servers or their specific application services. We have at times used a fan-out script for maintenance. The screenshots you shared are an effect of changing [the AMI specified in the auto-scaling group's launch configuration](https://github.com/code-dot-org/code-dot-org/blob/fe415228405edc335d24ad07d7b489ee02183dbe/aws/cloudformation/components/ami.yml.erb#L153) during a DTP, which causes the auto-scaling group to replace all instances.

The thing we might be looking for here is "[instance refresh](https://docs.aws.amazon.com/autoscaling/ec2/userguide/asg-instance-refresh.html)".

etaderhold:
> We don't have a regular use case of restarting either the EC2 web servers or their specific application services. We have at times used a fan-out script for maintenance. The screenshots you shared are an effect of changing [the AMI specified in the auto-scaling group's launch configuration](https://github.com/code-dot-org/code-dot-org/blob/fe415228405edc335d24ad07d7b489ee02183dbe/aws/cloudformation/components/ami.yml.erb#L153) during a DTP, which causes the auto-scaling group to replace all instances.
> 
> The thing we might be looking for here is "[instance refresh](https://docs.aws.amazon.com/autoscaling/ec2/userguide/asg-instance-refresh.html)".

Instance refresh sounds like a fine concept. I'll admit I don't even know where to begin translating the examples in that AWS document into code running in our Rails environment. Would you or someone on your team be able to help me with that?

cat5inthecradle:
I've created https://codedotorg.atlassian.net/browse/INF-1373, and infra will discuss it tomorrow.

bin/dotd
sureshc: I'd prefer `seed-curriculum-content-from-staging-branch`
It's longer but explains exactly what it's doing.

sureshc:
> It's hard to say whether production-daemon or production-console is the better choice here - but DOTD commands generally run on 'console'

Our long term goal is for deployment of updated curriculum content to be an automated process triggered by the merging of validated, tested curriculum content to the main branch, so I think `production-daemon` is the place where we want to manually execute this process.

etaderhold:
In light of the difficulty getting an automatable instance refresh working, I'd like to move forward with this seeding process using a manual instance refresh through the AWS console. I have put instructions to this effect in the DotD script in this PR and would like to test the procedure during my upcoming DotD shift.

bethanyaconnor (approved):

bin/dotd
```diff
@@ -303,6 +304,13 @@ def dotd_menu(dotd_name)
         @logger.info 'kills tests'
       end
       menu.choice(:DTP) {do_dtp(dotd_name)}
+      menu.choice(:STP) do
+        if run_on 'production_daemon', 'production/bin/seed-curriculum-content-from-staging-branch'
+          puts 'Content from staging branch successfully seeded to production database'
+        else
+          puts 'Content from staging branch was not successfully seeded to production database.'
```
cat5inthecradle, line 311: Can this say "why"?

bin/seed-curriculum-content-from-staging-branch
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env ruby
+
+require_relative '../deployment'
+require 'cdo/aws/ec2_autoscaling'
+require 'cdo/rake_utils'
+
+# This script looks for a separate git worktree at ../../staging that has the
```
cat5inthecradle, line 7: @sureshc have we automated or documented the creation of this worktree?

cat5inthecradle (approved):

bin/dotd
etaderhold: I'm expecting the output of the `rake seed` to show up in the terminal and provide more clarity. If it doesn't we can address that in a follow-up.

PR #59934, opened 2024-07-22
Make it possible to edit predict level code after submitting an answer

molly-moen:
We had a request from curriculum to see if we could support the following scenario for predict levels
- Predict levels always start out read only.
- Once a user submits their response, the predict level code becomes editable _if_ the level has that setting enabled. By default it stays read only.

With our new model for predict levels, this ended up being pretty easy! It was a matter of adding the checkbox on the levelbuilder side, and using that configuration in our `isReadOnlyWorkspace` method.

As part of this PR I also noticed that when editing start sources for a predict level, we were disabling the run button until the levelbuilder wrote an answer, which is unnecessary. I added logic to enable the run button in start mode for predict levels always.

In addition, level 7 of the all the things python lab progression was using the outdated `source` field, so I updated that to `start_sources`.

## Links
- jira ticket: [CT-646](https://codedotorg.atlassian.net/browse/CT-646)

## Testing story
Tested locally that the new readonly logic works as expected, and does not impact other labs.

## PR Checklist:

- [ ] Code is well-commented


apps/src/lab2/lab2Redux.ts
```diff
@@ -288,15 +288,25 @@ export const isReadOnlyWorkspace = (state: RootState) => {
...
     state.lab.levelProperties?.predictSettings?.isPredictLevel || false;
+  let isReadonlyPredictLevel = isPredictLevel;
+  if (isPredictLevel) {
+    const isEditableAfterSubmit =
+      state.lab.levelProperties?.predictSettings?.editableAfterSubmit || false;
+    const hasSubmittedPredictResponse = state.predictLevel.hasSubmittedResponse;
+    // If the predict level is not editable after submit or the user has not submitted a response,
+    // the predict level is read only.
+    isReadonlyPredictLevel =
+      !isEditableAfterSubmit || !hasSubmittedPredictResponse;
```
mikeharv, line 300: Could we make it clearer that "editable" here refers to the code and the answer to the predict question? I could see this being done by renaming the level property `editableAfterSubmit` to `codeEditableAfterSubmit`. Since that would be a pain, we could also do a similar rename to the locally scoped `isEditableAfterSubmit` on line 293 above, or even just add a comment.

apps/src/codebridge/ControlButtons/index.tsx
```diff
@@ -53,7 +55,9 @@ const ControlButtons: React.FunctionComponent = () => {
   const hasSubmitted = useAppSelector(
     state => getCurrentLevel(state)?.status === LevelStatus.submitted
   );
-  const disableRunAndTest = loading || (isPredictLevel && !hasPredictResponse);
+  const isStartMode = getAppOptionsEditBlocks() === START_SOURCES;
+  const disableRunAndTest =
+    loading || (!isStartMode && isPredictLevel && !hasPredictResponse);
```
mikeharv: nit: This might be more readable if we make one more helper variable, like:

```
  const awaitingPredictSubmit = !isStartMode && isPredictLevel && !hasPredictResponse;
  const disableRunAndTest = loading || awaitingPredictSubmit;
```

mikeharv (approved):
Looks good! A couple small suggestions, neither are blocking

apps/src/lab2/lab2Redux.ts
molly-moen: it's pretty easy to rename it at this point, I'll rename the level property!

apps/src/codebridge/ControlButtons/index.tsx
molly-moen: good call, will update

apps/src/codebridge/ControlButtons/index.tsx
```diff
@@ -53,7 +55,12 @@ const ControlButtons: React.FunctionComponent = () => {
   const hasSubmitted = useAppSelector(
     state => getCurrentLevel(state)?.status === LevelStatus.submitted
   );
-  const disableRunAndTest = loading || (isPredictLevel && !hasPredictResponse);
+  const isStartMode = getAppOptionsEditBlocks() === START_SOURCES;
```
sanchitmalhotra126, line 58: Minor: not required for this change, but wondering if it would be good to have a function in `utils` that just returns `isStartMode` directly

sanchitmalhotra126 (approved):
LGTM!

apps/src/codebridge/ControlButtons/index.tsx
molly-moen: good call, it probably would. I think we set it up like this since we would probably need to get different edit blocks values, but we check start sources enough that it would be good to have a dedicated method.

PR #59936, opened 2024-07-22
Another speculative flaky test fix for pl_landing_page.feature

kelbyhawn:
Another try at fixing the flaky `pl_landing_page.feature` UI test. Bethany discovered there were 2k+ workshops on the test machine so I deleted all of them in hopes that it will stop timing out during this test. 

~@bethanyaconnor and I tried adding a step that deletes the PM account and workshop as the last step of this test scenario to prevent the issue above from happening again, but kept getting an error that the Workshop ID didn't exist. @hannahbergam and I paired on this, and she discovered that the workshops seem to be automatically deleted on Cucumber; we checked the tables locally and on `test` via a SQL query and it had a `deleted_at` date for every workshop in the column which seems to suggest that the workshops are being deleted after the tests run:~ Ignore! See comments.

Screenshot 2024-07-30 at 4 14 34 PM

## Links
Jira ticket: [ACQ-2057](https://codedotorg.atlassian.net/browse/ACQ-2057)

## Testing story
Waited for drone to pass w/ the [test all browsers] commit message.


bethanyaconnor:
@kelbyhawn those timestamps match up with when you bulk deleted a lot of objects on the test server. Do you have more recent runs that show that the workshops are being automatically deleted?

bethanyaconnor:
Maybe instead of trying to delete workshops we could use a different RP for each run? Ideally we'd still clean up the same way the application test does but we might be able to get away with not doing that if we're not trying to load all the workshops for each program manager

kelbyhawn:
> @kelbyhawn those timestamps match up with when you bulk deleted a lot of objects on the test server. Do you have more recent runs that show that the workshops are being automatically deleted?

Oh you're right. Here's a screenshot of the bottom of the table in `test` which still looks like the bulk deletion, but when I do `created_at` it's showing ones from yesterday 🤔 @hannahbergam maybe they aren't getting automatically deleted on test? (Also very excited about learning some SQL queries!)

Screenshot 2024-07-31 at 8 41 46 AM

kelbyhawn:
For more context, this is similar to what we were seeing locally yesterday, which seems to suggest they're getting deleted automatically locally, and the ui test couldn't find the workshop because it was already deleted (the SauceLabs error is the HTML part):

Screenshot 2024-07-31 at 9 00 15 AM

kelbyhawn:
@bethanyaconnor I'm able to delete the workshop now, and it seems like the errors were coming from trying to delete the PM 🤔 Would this be sufficient? Passing test [here](https://app.saucelabs.com/tests/3d113a29f1b44aadbe56c4ba85ae12a5) and commit w/ just the workshop deletion here: https://github.com/code-dot-org/code-dot-org/pull/59936/commits/6d984fd5567f05b8511943db084dcb3cb3fa835f. 

Screenshot 2024-07-31 at 10 05 19 AM

bethanyaconnor:
Yes just deleting the workshops is probably sufficient!

bethanyaconnor (approved):
Woohoo! Looks good after Drone passes!

PR #59810, opened 2024-07-16
Create Account Button A/B Test - Add Sign in button event

kelbyhawn:
Adds/updates the following events for the Create Account button experiment:
- Adds an event to the Sign in button
- Updates the property for the header/hamburger links to the url
- Adds a `headerType` property to the `Signed Out Header Navigation Shown` event
  - `large desktop`
    - all header links, no hamburger
  - `small desktop`
    - shows the truncated header links and partial hamburger
   - `tablet`
      - shows the full hamburger but buttons are outside (when in the test bucket)
  -  `mobile`
      - everything is in the hamburger (when in the test bucket)

## Links
Jira ticket: [ACQ-2072](https://codedotorg.atlassian.net/browse/ACQ-2072)

## Testing story
Tested locally

PR #59942, opened 2024-07-23
Fix browser null error on LTI launch

carl-codeorg:
[An event listener was recently added ](https://github.com/code-dot-org/code-dot-org/pull/59810/files)that depends on signInButton existing. In at least one case (launching from the LMS as a roster synced user), the button doesn't exist, so a JS error is thrown trying to add an event listener to a null object. This adds a guard to ensure the listener is only added if the button exists.

## Links

[JIRA](https://codedotorg.atlassian.net/browse/P20-1056)

## PR Checklist:

- [ ] Code is well-commented


stephenliang (approved):

kelbyhawn (approved):
Thanks for fixing this Carl!

PR #59944, opened 2024-07-23
Removing reference to rbenv-doctor

unlox775-code-dot-org:
This is a followup PR to [#59869](https://github.com/code-dot-org/code-dot-org/pull/59869), as we decided we just don't need to run rbenv-doctor.


unlox775-code-dot-org:
Accidentally included a wrong file.  Reverted changes to logrotate.erb in https://github.com/code-dot-org/code-dot-org/pull/59946

PR #59951, opened 2024-07-23
Accessibility: update homepage to use DSCO Button

Erin007:
This PR updates 3 components used on the student and teacher homepages to use the accessible DSCO button.

1.) `TopCourse` 

BEFORE 
Screenshot 2024-07-25 at 5 10 41 PM

AFTER 
Screenshot 2024-07-25 at 5 19 24 PM

2.) `CourseCard`

BEFORE 
Screenshot 2024-07-25 at 5 12 10 PM

AFTER
Screenshot 2024-07-25 at 5 12 22 PM

3.) `ImageResourceCard`

BEFORE 
Screenshot 2024-07-25 at 5 13 28 PM

AFTER
Screenshot 2024-07-25 at 5 00 32 PM

While I was in here, I also extracted `TopCourse` styles into their our module, which allowed me to [remove another usage of Radium](https://codedotorg.atlassian.net/browse/XTEAM-204) 🎉 

## Follow up 
This was already a side quest for me that I need to put down for now, but extracting styles for `CourseCard` and `ImageResourceCard` would be nice-to-have clean up. We also now have a Component Library [Card](https://github.com/code-dot-org/code-dot-org/tree/staging/apps/src/componentLibrary/card) component; it would be good to investigate what it would take to use that in place of these components (and `ResourceCard` too). 

## PR Checklist:

- [ ] Code is well-commented


molly-moen (approved):
🎉

apps/src/templates/studioHomepages/CourseCard.jsx
```diff
@@ -24,19 +24,22 @@ class CourseCard extends Component {
     const {title, description, link, isRtl} = this.props;
 
     return (
-      <a href={link} style={styles.card}>
+      <div style={styles.card}>
```
Erin007, line 27: Note this now means that only the button, not the whole card is clickable. Nested <a> tags is an anti-pattern and the storybook tests got very grumpy about it.

PR #59960, opened 2024-07-23
Python Lab: More improvements to parsing error messages

molly-moen:
Follow up to #58958
Some cleanup to make handling error messages and stack traces a bit cleaner in Python Lab. The background for this change is by default pyodide puts all files in the `/home/pyodide` folder in the virtual filesystem. I was then parsing our error messages and removing references to `/home/pyodide`, to make it look like user's files are at the root of the filesystem. However, as I was looking at validation tests I found another case where I would have to do this, and wondered if there was a better way.

I came up with the following solution, and confirmed with curriculum it would make sense to students. Since the student files appear under a `Files` header, I named the default folder `Files`. This is similar behavior to a filesystem on a computer: your downloads folder has a "Downloads" header at the top. Now the filesystem and stack traces look as follows:

### File system
Screenshot 2024-07-23 at 1 57 51 PM

### Stack Trace
Screenshot 2024-07-23 at 2 02 26 PM

By making this change I was able to remove some of the code to customize error messages. I also discovered there was a setting in [pyodide](https://pyodide.org/en/stable/usage/api/js-api.html#pyodide.runPythonAsync) where we could set the name of the main file, rather than it being `<exec>` by default. So I could also remove the custom code to replace `<exec>` with `main.py`.

We still need to adjust the line numbers in the stack trace for any lines coming from `main.py`. This is because we add custom code before the student's code in order to apply our custom patches. We also remove the internal pyodide stack trace, which is confusing.

## Testing story
Tested locally with a variety of errors to ensure the messages still looked correct.

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/pythonlab/pyodideWebWorker.js
+++ b/apps/src/pythonlab/pyodideWebWorker.js
@@ -1,6 +1,9 @@
 import {DEFAULT_FOLDER_ID} from '@codebridge/constants';
 import {loadPyodide, version} from 'pyodide';
 
+import {MAIN_PYTHON_FILE} from '@cdo/apps/lab2/constants';
+
+import {HOME_FOLDER} from './pythonHelpers/constants';
 import {
   getUpdatedSourceAndDeleteFiles,
   importPackagesFromFiles,
@@ -22,7 +25,7 @@ async function loadPyodideAndPackages() {
       `/blockly/js/pyodide/${version}/pythonlab_setup-0.0.1-py3-none-any.whl`,
     ],
     env: {
-      HOME: '/Files/',
+      HOME: `/${HOME_FOLDER}/`,
     },
   });
   self.pyodide.setStdout(getStreamHandlerOptions('sysout'));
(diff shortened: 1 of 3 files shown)


snickell (approved):
- Using `/Files` is really clever, and I really like that it both makes UX-sense, and removes one more thing we have to munge.
- Also appreciate dropping `exec` in the same light

As far as munging line numbers (sure would be nice not to have to munge.... and this gets us much closer!), have you considered creating a `sitecustomize.py` module? 

https://docs.python.org/3/library/site.html#module-sitecustomize

I'm not sure what things you have to load currently, but if they aren't imports that are actually used in `main.py`, and just "we want to always run this python when we start" that might work to replace the hack.

On a normal system, this would be a thing the distro did and would be in a directory like `site-packages/sitecustomize.py`. I'm not 100% sure how that would map for this weird embedded python distro, but maybe its doable?

ebeastlake:
Nice work! Captured potential follow up: https://codedotorg.atlassian.net/browse/CT-702

ebeastlake (approved):

PR #59962, opened 2024-07-24
P20-1055: CAP data migration Phase 1

artem-vavilov:
Created one-off script to populate the CAP columns with data extracted from the JSON values `child_account_compliance_state` and `child_account_compliance_state_last_updated`.

`CAP` - Child Account Policy
https://github.com/code-dot-org/code-dot-org/blob/180cf779efa5c7bc0918ca17ca0a3dd0e4933fb6/dashboard/config/initializers/inflections.rb#L22

## Links
- JIRA task: [P20-1055](https://codedotorg.atlassian.net/browse/P20-1055)
- Migration PR: https://github.com/code-dot-org/code-dot-org/pull/60218
- Phase 2 PR: https://github.com/code-dot-org/code-dot-org/pull/59963


dashboard/app/models/user.rb
```diff
@@ -50,10 +50,13 @@
 #  races                    :string(255)
 #  primary_contact_info_id  :integer
 #  unlock_token             :string(255)
+#  cap_state                :string(1)
+#  cap_state_date           :datetime
```
mgc1194: Could you please give some detail about why you chose to create these columns of al the properties? Can you give an example of how they can be used to delete a locked-out user?

dashboard/test/models/user_test.rb
```diff
@@ -5455,4 +5455,24 @@ def hide_scripts_in_sections(section1, section2)
...
+    end
+
+    it 'assigns "cap_state_date" with data from "child_account_compliance_state"' do
+      expected_cap_state_date = Time.now.change(usec: 0)
+
+      assert_changes -> {user.reload.cap_state_date}, from: nil, to: expected_cap_state_date do
+        user.update!(child_account_compliance_state_last_updated: expected_cap_state_date)
+      end
+    end
+  end
```
mgc1194: These tests will also be removed:
`  # TODO(P20-1055): Remove in "CAP data migration Phase 2" once new CAP columns are populated`
See PR #59963

dashboard/app/models/user.rb
artem-vavilov: We need them to identify students locked out for more than 7 days. Additionally, we can use these fields to identify students in the CAP flow, which includes those in the grace period or locked-out states.
https://github.com/code-dot-org/code-dot-org/blob/95eead1b0dd4656e003f96203517e58fd2bb48af/dashboard/lib/queries/child_account.rb#L4-L17

mgc1194: How long would it take to iterate through all users by querying the user object such as in:
`User.where("cap_state = 'l'").find_each`?

dashboard/app/models/user.rb
```diff
@@ -50,10 +50,13 @@
 #  races                    :string(255)
 #  primary_contact_info_id  :integer
 #  unlock_token             :string(255)
+#  cap_state                :string(1)
```
mgc1194: Child Accoun Policy state (`cap_state`) refers to the current compliance status of a children's account.

Child accounts are defined per State (US State) law in `state_policies`:
https://github.com/code-dot-org/code-dot-org/blob/032f53af60a78e57580343329769f9226c0197dc/dashboard/lib/policies/child_account.rb#L144-L151

Possible values defined in:
https://github.com/code-dot-org/code-dot-org/blob/778859277f8bdaa46fa4665ac7da2b9b1a2db1b2/lib/cdo/shared_constants.rb#L652-L658

dashboard/app/models/user.rb
artem-vavilov: Since the `cap_state` column is indexed, the query shouldn't take longer than a few milliseconds

artem-vavilov: Since the `User` model uses the Paranoia gem, which adds a filter by the `deleted_at` column, the only issue could arise if the MySQL query optimizer decides to use `index_users_on_deleted_at` instead of `index_users_on_cap_state_and_cap_state_date`. This can be fixed in two ways: either exclude the filter by the `deleted_at` column from the query or force the query optimizer to use the required index (not a good solution).

dashboard/app/models/user.rb
mgc1194: I added this note because the term `cap_state` was confusing at first since it relates to US States (as in territories), but we are referring to the status of the compliance. 
I'd consider disambiguating by using the term `cap_status` (`ComplianceStatus`, `CHILD_ACCOUNT_COMPLIANCE_STATUSES`, etc)

artem-vavilov: cc @dayne @wilkie

bin/oneoff/backfill_data/user_cap_state_columns
```diff
@@ -0,0 +1,24 @@
...
+progress_bar = ProgressBar.create(
+  total: nil,
+  title: 'Migrating CAP data',
+  format: '%t[%c]: |%B| %a',
+  unknown_progress_animation_steps: %w[>-- ->- -->]
+)
+
+User.select(:id).where(cap_state: nil).where("properties->>'$.child_account_compliance_state' IS NOT NULL").find_in_batches do |users_batch|
+  User.where(id: users_batch).update_all <<~SQL.strip
+    cap_state = IF(properties->>'$.child_account_compliance_state' = 's', 'l', properties->>'$.child_account_compliance_state'),
```
artem-vavilov: https://github.com/code-dot-org/code-dot-org/pull/59192/files#diff-3ff006839352d79e94cd9df80af678928faa6c72047f52c73d55c809979dfa92R652-R658

dashboard/test/models/user_test.rb
```diff
@@ -5455,4 +5455,24 @@ def hide_scripts_in_sections(section1, section2)
...
+
+    it 'assigns "cap_state" with data from "child_account_compliance_state"' do
+      expected_cap_state = Policies::ChildAccount::ComplianceState::LOCKED_OUT
+
+      assert_changes -> {user.reload.cap_state}, from: nil, to: expected_cap_state do
+        user.update!(child_account_compliance_state: expected_cap_state)
+      end
+    end
+
+    it 'assigns "cap_state_date" with data from "child_account_compliance_state"' do
```
daynew:
```suggestion
    it 'assigns "cap_state_date" with data from "child_account_compliance_state_last_updated"' do
```

dashboard/app/models/user.rb
daynew: @mgc1194 I agree that it's a little confusing, but we have been consistent about using `us_state` when talking about USA states. I think we should keep it as `_state` because we should feel free to use the word "state" when used in the appropriate context.

daynew (approved):
My biggest concern is whether we are selecting configuring the indexes correctly since I'm not experienced at that. However, I'm willing to ship this and then see what happens in Prod.

mgc1194 (approved):

dashboard/db/migrate/20240723202827_add_cap_state_and_cap_state_date_to_users.rb
```diff
@@ -0,0 +1,13 @@
+class AddCAPStateAndCAPStateDateToUsers < ActiveRecord::Migration[6.1]
+  def up
+    add_column :users, :cap_state, :string, limit: 1
+    add_column :users, :cap_state_date, :datetime
+    add_index :users, %i[cap_state cap_state_date]
```
Hamms: Unfortunately, this index can't get applied as part of this migration in production; we have a 30-second time limit on ActiveRecord queries, and even with the new faster and non-locking MySQL 8 indexes, it takes longer than 30 seconds to add an index to a table the size of our production users table.

We'll have to find another way to do this; one option is to add an `unless production` check to this migration and to add the index manually. See https://github.com/code-dot-org/code-dot-org/pull/59668 for an example

dashboard/app/models/user.rb
Hamms: I'm curious to know what the performance implications would be of extracting and adding an index to just the `cap_state` column, and leaving the `cap_state_date` column in the blob. Do we have any idea of approximately how many users we expect to be in a locked-out state at any given time?

bin/oneoff/backfill_data/user_cap_state_columns
```diff
@@ -0,0 +1,24 @@
...
+
+progress_bar = ProgressBar.create(
+  total: nil,
+  title: 'Migrating CAP data',
+  format: '%t[%c]: |%B| %a',
+  unknown_progress_animation_steps: %w[>-- ->- -->]
+)
+
+User.select(:id).where(cap_state: nil).where("properties->>'$.child_account_compliance_state' IS NOT NULL").find_in_batches do |users_batch|
+  User.where(id: users_batch).update_all <<~SQL.strip
```
Hamms: Mentioned this in slack, repeating it here too:

I believe we have historically run into problems when trying to batch update users in large transactions, although I don't remember the details. I think we will at a minimum want to use a single isolated transaction for each row update and add some error handling logic, and will certainly want to run some tests with this script against a clone of the production database. Tagging @sureshc who has more recent familiarity with data migration scripts

artem-vavilov: > I'm curious to know what the performance implications would be of extracting and adding an index to just the cap_state column, and leaving the cap_state_date column in the blob. Do we have any idea of approximately how many users we expect to be in a locked-out state at any given time?

@Hamms We don't expect a large number of users to be affected initially, as this will only impact Colorado students under 13 for 7 days. However, the number of locked-out users could increase significantly in the future as we add more US states.

dashboard/app/models/user.rb
artem-vavilov: We don't expect a large number of users to be affected initially, as this will only impact Colorado students under 13 for 7 days. However, the number of locked-out users could increase significantly in the future as we add more US states.

dashboard/db/migrate/20240723202827_add_cap_state_and_cap_state_date_to_users.rb
artem-vavilov: https://github.com/code-dot-org/code-dot-org/pull/59962/commits/4deef87ca772d108e8e794495d618c99fbd7c0b2

coryamanda (approved):
Seen by RED team, thanks for the notification.

cat5inthecradle (changes requested):
Please update [tasks.yml](https://github.com/code-dot-org/code-dot-org/blob/119c27e2e9e7ea1ddea0a9c2f8df37d4e3683944/aws/dms/tasks.yml#L141) to exclude new PII columns from the non-pii schema.

The current configuration in tasks.yml removes the `properties` column before exporting the users table, which effectively excludes this data. Now that this data is in it's own columns, they will need to be explicitly excluded from the non-pii schema.

EDIT: After Slack discussion, none of these columns are PII. So while it might be noise to include them, it's not a privacy concern.

cat5inthecradle (approved):

sureshc:
This is not well documented, but we have a policy of placing Rails database migrations in their own Pull Request and fully deploying that schema change to all environments before merging changes that use the new schema. Let’s place the migration, the updated `schema.rb` and the updated User model with just the new annotation in a separate pull request.

sureshc:
It might be useful to spell out what “CAP” stands for in the Summary of this Pull Request.

artem-vavilov:
> It might be useful to spell out what “CAP” stands for in the Summary of this Pull Request.

`CAP` means Child Account Policy
https://github.com/code-dot-org/code-dot-org/blob/180cf779efa5c7bc0918ca17ca0a3dd0e4933fb6/dashboard/config/initializers/inflections.rb#L22

dashboard/db/migrate/20240723202827_add_cap_state_and_cap_state_date_to_users.rb
artem-vavilov: https://github.com/code-dot-org/code-dot-org/pull/60083
https://github.com/code-dot-org/code-dot-org/blob/a3155bc0a2b7f9cc453e9462331912f4d82f83e1/dashboard/db/migrate/20240723202827_add_cap_state_and_cap_state_date_to_users.rb#L6-L21

bin/oneoff/backfill_data/user_cap_state_columns
```diff
@@ -0,0 +1,30 @@
...
+batch_size = 100_000
+total_updated = 0
+
+progress_bar_title = proc {|num_updates = 0| "Processed[%c/%C]: |%W| Updated: #{total_updated += num_updates} %a"}
+progress_bar = ProgressBar.create(total: max_id, format: progress_bar_title.call)
+
+(1..max_id).step(batch_size) do |start_id|
+  finish_id = [start_id + batch_size.pred, max_id].min
+
+  User.select(:id).where(cap_state: nil).where("properties->>'$.child_account_compliance_state' IS NOT NULL").find_in_batches(start: start_id, finish: finish_id) do |users_batch|
```
sureshc: There are >100M rows in this table and likely only the most recent rows have a value in the `child_account_compliance_state` property). Can we safely assume that only students who have signed in within the last 2 years will have a value set? `last_sign_in_at` is an indexed column. Maybe we can do something like this:

``` ruby

cutoff_date = 2.year.ago
User.
  select(:id).
  where('last_sign_in_at > ?', cutoff_date).
  where(cap_state: nil).
  where("properties->>'$.child_account_compliance_state' IS NOT NULL").
  in_batches(of: 1000) do |relation|
    relation.update_all <<~SQL.strip
       cap_state = IF(properties->>'$.child_account_compliance_state' = 's', 'l', properties->>'$.child_account_compliance_state'),
       cap_state_date = CAST(properties->>'$.child_account_compliance_state_last_updated' AS DATETIME)
     SQL
  end
```

sureshc: We should provision an adhoc environment and configure it to use a clone of the production database to test this backfill script.

artem-vavilov: Any Colorado student under 13 is covered by the Colorado Privacy Act policy, regardless of when they were created.

bin/oneoff/backfill_data/user_cap_state_columns
```diff
@@ -0,0 +1,30 @@
...
+# This script migrates CAP data from properties to separate indexed columns.
+
+require 'ruby-progressbar'
+require_relative '../../../dashboard/config/environment'
+
+max_id = User.maximum(:id)
+batch_size = 100_000
+total_updated = 0
+
+progress_bar_title = proc {|num_updates = 0| "Processed[%c/%C]: |%W| Updated: #{total_updated += num_updates} %a"}
```
sureshc: In addition to a live progress bar, we should adopt the best practices for logging successful and failed batches from this data clean up script: https://github.com/code-dot-org/code-dot-org/blob/staging/bin/oneoff/data_fix/fix_invalid_users

bin/oneoff/backfill_data/user_cap_state_columns
sureshc: > Any Colorado student under 13 is covered by the Colorado Privacy Act policy, regardless of when they were created.

Are there Colorado students where the `child_account_compliance_state` property is NOT NULL that have not signed in over the last 2 years? Did we fill in that value for all Colorado students regardless of whether they have signed in?

bin/oneoff/backfill_data/user_cap_state_columns
artem-vavilov: The script is straightforward and doesn't require logging failed batches.
If the script is interrupted, it can always be restarted from where it stopped due to this condition: `where(cap_state: nil)`

artem-vavilov: cc @Hamms

bin/oneoff/backfill_data/user_cap_state_columns
artem-vavilov: It's a great idea, Suresh! Thank you!

bin/oneoff/backfill_data/user_cap_state_columns
sureshc: We have inadvertently corrupted production data in the past with data update scripts, so careful logging of the changes that a back fill script has made including detailed error logging of each failed batch is important. The `fix_invalid_users` script has some of the best logging / tracking logic that we've implemented in recent years. And because we're using `update_all` we'll have limited information about which records were modified. We should log the User ID of each row we've successfully updated, and the User ID of each row we attempted to update and failed, along with the reason.

bin/oneoff/backfill_data/user_cap_state_columns
```diff
@@ -0,0 +1,30 @@
...
+progress_bar = ProgressBar.create(total: max_id, format: progress_bar_title.call)
+
+(1..max_id).step(batch_size) do |start_id|
+  finish_id = [start_id + batch_size.pred, max_id].min
+
+  User.select(:id).where(cap_state: nil).where("properties->>'$.child_account_compliance_state' IS NOT NULL").find_in_batches(start: start_id, finish: finish_id) do |users_batch|
+    User.where(id: users_batch).update_all <<~SQL.strip
```
sureshc: There can be performance tradeoffs when updating records in batches vs individually. Some testing on a clone of the production database cluster can help us assess that. Separately from performance, are we intentionally wanting to use the features of the `update_all` method such as:

- Do NOT update `created_at` / `updated_at` attributes
- Do NOT invoke callbacks

I think the answer is that we DO want that functionality. These aren't application updates to these records. We're just moving data to a new schema.

bin/oneoff/backfill_data/user_cap_state_columns
sureshc: Let's also `rescue` each attempt to `update_all` a batch, so that we can log it and move on to the next batch even if one set of updates fails.

bin/oneoff/backfill_data/user_cap_state_columns
artem-vavilov: What data corruption are you referring to? Because the worst that can happen with this script is that some records might not be updated.

artem-vavilov: What data corruption are you referring to? The worst that can happen with this script is that some records might not be updated. However, this is not an issue because the script will exclude the already updated records when rerun

bin/oneoff/backfill_data/user_cap_state_columns
artem-vavilov: https://github.com/code-dot-org/code-dot-org/pull/59962/commits/ba87048a1cb6fd5835dea764fbe9dfac8d5617ca

bin/oneoff/backfill_data/user_cap_state_columns
artem-vavilov: https://github.com/code-dot-org/code-dot-org/pull/59962#discussion_r1697584729

bin/oneoff/backfill_data/user_cap_state_columns
sureshc: Hopefully the script and its logic don't have any defects, but when we're modifying production data in a critical table and particularly when we don't have unit tests for the logic that finds and updates records, we need to be very careful. Logging each batch and each record (`users.id`) that we've successfully updated and that we've failed to update is important.

bin/oneoff/backfill_data/user_cap_state_columns
```diff
@@ -0,0 +1,40 @@
...
+users.select(
+  :id,
+  :cap_state,
+  :cap_state_date,
+  "properties->>'$.child_account_compliance_state' AS original_cap_state",
+  "properties->>'$.child_account_compliance_state_last_updated' AS original_cap_state_date"
+).find_each(batch_size: 10_000) do |user|
+  next if user.cap_state || user.original_cap_state.blank?
+
+  user.update_columns(
```
Hamms: I believe we do want to make sure each of these updates happens in an individual transaction; does `update_columns` do that for us, or do we need to explicitly wrap it in an `ActiveRecord::Base.transaction` block?

bin/oneoff/backfill_data/user_cap_state_columns
```diff
@@ -0,0 +1,40 @@
...
+    cap_state: user.original_cap_state == 's' ? Policies::ChildAccount::ComplianceState::LOCKED_OUT : user.original_cap_state,
+    cap_state_date: user.original_cap_state_date.presence && DateTime.parse(user.original_cap_state_date)
+  )
+
+  progress_bar.format update_progress_title.call
```
Hamms: We will also need to log data about successful updates as well as unsuccessful ones; in the event of any kind of unexpected outcomes, it's critical that we have all the information we need to get our users back into their original state. At a minimum, this means we need to record the ids of all affected users as well as the original data values in the affected columns before the update is applied

Hamms (commented):
Couple little nits, but this is looking a lot closer! Will be interested to hear what @sureshc has to say about the latest version.

One new thing I would like to see is support for a "dry run" option; or even better, an "actually perform the update" option like: https://github.com/code-dot-org/code-dot-org/blob/afabbe66990d87fbbc7a0e0d683a4339be555773/bin/oneoff/data_fix/fix_invalid_users#L36-L38

bin/oneoff/backfill_data/user_cap_state_columns
artem-vavilov: Since we use InnoDB with [autocommit](https://dev.mysql.com/doc/refman/8.0/en/server-system-variables.html#sysvar_autocommit) mode enabled (`SELECT @@autocommit;`), each SQL statement forms a single transaction on its own. See: https://dev.mysql.com/doc/refman/8.0/en/innodb-autocommit-commit-rollback.html#:~:text=In%20InnoDB,%20all%20user%20activity%20occurs%20inside%20a%20transaction

bin/oneoff/backfill_data/user_cap_state_columns
artem-vavilov: Do we need to track all affected users individually, since, in this case, we can easily identify them by the statement `cap_state IS NOT NULL`?

Hamms: Yes, we do; among other things, we will need to be able to distinguish between users who arrived at that state via this script and users who got there by any other method.

bin/oneoff/backfill_data/user_cap_state_columns
Hamms: Great, thanks!

bin/oneoff/backfill_data/user_cap_state_columns
artem-vavilov: I agree we would need it if we were going to use it multiple times. However, in this case, we only need to copy-paste data from one column to another once, and we won't need to touch it again. I would also prefer to delete the script immediately after the data is populated.

bin/oneoff/backfill_data/user_cap_state_columns
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env ruby
+
+# This script migrates CAP (Child Account Policy) data from properties to separate indexed columns.
+
+require 'ruby-progressbar'
+require_relative '../../../dashboard/config/environment'
+
+# @see https://github.com/code-dot-org/code-dot-org/pull/52244
+CAP_RELEASE_DATE = DateTime.parse('2023-06-09')
```
daynew: I queried our database, and the oldest `current_sign_in_at` is `2023-06-29 18:23:10.000+00:00` where `child_account_compliance_state` is set, so this date should be fine. 👍

daynew:
I ran the script on a clone of our production database:
```
Processed[16288377/16288377]: |==== 100 ====| Updated: 5157 Time: 03:20:04
```
The 5157 updates is very close to the 5150 count I get when I do a redshift query for our 1 day stale data.

I also did a spot check on a user and confirmed that the `cap_state` and `cap_state_date` attributes now have values.

bin/oneoff/backfill_data/user_cap_state_columns
artem-vavilov: https://github.com/code-dot-org/code-dot-org/pull/59962/commits/31f5b2c2a232523a12bcc7f7e25ad60925a1cfe3

bin/oneoff/backfill_data/user_cap_state_columns
```diff
@@ -0,0 +1,91 @@
...
+  when 'skipped'
+    total_successes += 1
+    progress_bar.log "[#{Time.now.utc.iso8601}] Skipped updating User[#{user.id}]: #{message}".yellow
+  when 'failed'
+    total_failures += 1
+    progress_bar.log "[#{Time.now.utc.iso8601}] Failed to update User[#{user.id}]: #{message}".red
+  end
+
+  log_data[status][user.id] = message
+  File.write log_path, JSON.pretty_generate(log_data)
```
wilkie: Maybe I'm being a little pedantic for a oneoff script, but is it going to be a problem generating this thousands of entries wide dictionary repeatedly? I could see it getting ever so slightly slower as it goes haha. Maybe it is more reasonable to just do an appended log / CSV? You're storing 'message' here as a JSON string anyway, I believe.

artem-vavilov: https://github.com/code-dot-org/code-dot-org/pull/59962/commits/46951522ab5aeef954df90dd51dcc10d19e5f5b7

bin/oneoff/backfill_data/user_cap_state_columns
```diff
@@ -0,0 +1,88 @@
...
+    progress_bar.log "[#{Time.now.utc.iso8601}] Successfully updated User[#{user.id}]: #{message}".green
+  when 'skipped'
+    total_successes += 1
+    progress_bar.log "[#{Time.now.utc.iso8601}] Skipped updating User[#{user.id}]: #{message}".yellow
+  when 'failed'
+    total_failures += 1
+    progress_bar.log "[#{Time.now.utc.iso8601}] Failed to update User[#{user.id}]: #{message}".red
+  end
+
+  log_csv.add_row [status, user.id, message]
```
Hamms: Does this actually write the row data out to the file immediately upon being called? My understanding of Ruby CSV is that it will store the content in memory until the handler is closed, at which point everything will be written out. It is critical for our needs that the log data for each individual user being updated gets persisted somewhere before any change is made to the actual database data. If for example something goes wrong and we need to kill the script mid-execution, we will need to know exactly what operations the script completed before being killed.

Because managing the exact timing of file writes can be tricky from within Ruby, and because we care a lot more about performance and reliability of this log data than we do about it being nicely packaged, it might be best to keep our implementation simple. One approach would be to just write the data we care about to `stdout`, and we can make sure to persist that output to a file when we run it. See for example the implementation of https://github.com/code-dot-org/code-dot-org/blob/staging/bin/oneoff/data_fix/fix_invalid_users

bin/oneoff/backfill_data/user_cap_state_columns
```diff
@@ -0,0 +1,88 @@
...
+    ).
+    where(cap_state: nil).
+    where("properties->>'$.child_account_compliance_state' IS NOT NULL").
+    where("properties->>'$.child_account_compliance_state' != ''")
+
+  cap_users.find_each do |user|
+    cap_state = user.original_cap_state == 's' ? Policies::ChildAccount::ComplianceState::LOCKED_OUT : user.original_cap_state
+    cap_state_date = user.original_cap_state_date.presence && DateTime.parse(user.original_cap_state_date).utc
+
+    if options[:skip_update]
```
Hamms: :+1:

bin/oneoff/backfill_data/user_cap_state_columns
artem-vavilov: https://github.com/code-dot-org/code-dot-org/pull/59962/commits/fe88013842bdeba4c4f95f365130a5cd8daea228

Gemfile
```diff
@@ -188,7 +188,7 @@ gem 'highline', '~> 1.6.21'
 
 gem 'honeybadger', '>= 4.5.6' # error monitoring
 
-gem 'newrelic_rpm', '~> 6.14.0', group: [:staging, :development, :production] # perf/error/etc monitoring
+gem 'newrelic_rpm', '~> 6.14.0', require: false, group: [:staging, :development, :production] # perf/error/etc monitoring
```
artem-vavilov: https://github.com/code-dot-org/code-dot-org/blob/d9caabe9ac007e1c99ba398db5c6c99ce4221e38/dashboard/config/application.rb#L212-L214

wilkie (approved):
looks great! thanks so much! 👍

Gemfile
Hamms: This feels like a potentially very broad change; I'm definitely not opposed to us exploring it, but I am opposed to us trying to do it as a small piece of a PR that's largely focused elsewhere.

Based on my manual testing of this script, I think it's plenty performant without this optimization; I vote we leave as-is for now, and consider this for a follow-up

bin/oneoff/backfill_data/user_cap_status_columns
```diff
@@ -54,30 +60,21 @@ end
...
-    if options[:skip_update]
-      user.assign_attributes(cap_status: cap_status, cap_status_date: cap_status_date)
-    else
-      user.update_columns(cap_status: cap_status, cap_status_date: cap_status_date)
+  cap_users.ids.each do |user_id|
+    unless options[:skip_update]
+      User.where(id: user_id).update_all <<-SQL.squish
+        cap_status = IF(properties->>'$.child_account_compliance_state' = 's', 'l', properties->>'$.child_account_compliance_state'),
+        cap_status_date = CAST(properties->>'$.child_account_compliance_state_last_updated' AS DATETIME)
+      SQL
```
Hamms: I really don't like using `update_all` for this; is there a specific reason to do so, rather than going back to

```ruby
 User.find(user_id).update_columns(cap_status: cap_status, cap_status_date: cap_status_date)
```

bin/oneoff/backfill_data/user_cap_status_columns
```diff
@@ -54,30 +60,21 @@ end
...
     end
 
-    log_operation.call(success_status, user, user.as_json(only: %i[cap_status cap_status_date]))
+    log_operation.call(success_status, user_id)
```
Hamms: We do indeed want to log data for the fields we're trying to update, not just the user's id

bin/oneoff/backfill_data/user_cap_status_columns
```diff
@@ -1,8 +1,14 @@
 #!/usr/bin/env ruby
+# frozen_string_literal: true
 
 # This script migrates CAP (Child Account Policy) data from properties to separate indexed columns.
 
 require 'ruby-progressbar'
+require 'honeybadger/ruby'
+
+# Disable Honeybadger breadcrumbs to avoid memory leaks.
+Honeybadger.config.set(:'breadcrumbs.enabled', false)
```
Hamms: Great investigation on this, by the way!

Hamms (commented):
I tested an earlier version of this script against a clone of the production database in an adhoc. It finished successfully in ~20 minutes, updating 5175 user records:

```
Finished at: 2024-08-05T23:30:54Z
275.31user 5.62system 19:14.12elapsed 24%CPU (0avgtext+0avgdata 3871572maxresident)k
0inputs+4424outputs (0major+1065555minor)pagefaults 0swaps
```

Although that was before the latest batch of tweaks, which I have some additional feedback about. But we are super close!

Gemfile
artem-vavilov: https://github.com/code-dot-org/code-dot-org/pull/59962/commits/1ad136ecf01555113a5c2a25836ff9ea92513e7d

bin/oneoff/backfill_data/user_cap_status_columns
artem-vavilov: `User.where(id: user_id).update_all` achieves the same result as `User.find(user_id).update_columns` but without loading and initializing the model instance. This prevents memory leaks caused by the serialization of `properties` (`292.28 MB in dashboard/app`).

Hamms: Ahh, gotcha. That seems reasonable; thanks for clarifying! Can you add a comment explaining that reasoning? I'd like to capture the details for anyone reading this code in the future

bin/oneoff/backfill_data/user_cap_status_columns
artem-vavilov: https://github.com/code-dot-org/code-dot-org/pull/59962/commits/756a73d082adff1d2335ee691bf4987a49bcbf2a

Hamms (approved):
LGTM! Thanks for all your hard work and persistence with this one.

I tested the latest version on an adhoc against a clone of the production db; it ran successfully in ~20 minutes

```
Finished at: 2024-08-07T01:19:50Z
263.94user 5.78system 19:05.92elapsed 23%CPU (0avgtext+0avgdata 3723752maxresident)k
0inputs+3960outputs (0major+1028700minor)pagefaults 0swaps
```

@sureshc do you have any remaining thoughts or concerns with this one? I definitely want to make sure we haven't missed anything

sureshc (approved):

PR #52244, opened 2023-06-07
Adding URL for parents to approve their child's Code.org account

daynew:
This creates a URL Colorado parents can visit to grant permission for their child to have a Code.org account. A separate PR will do the work of sending these links in an email.

Notes
* The CSS styling doesn't match what is in the UI mocks. Code.org is currently transitioning to a new style and I have implemented this feature using the new Component Library components.

### Permission was successfully granted
Box with text thanking the parent for granting their child permission.

### The permission token no longer exists
Box with text explaining that the link they clicked is no longer valid and the child needs to submit another request.

## Links
* [JIRA](https://codedotorg.atlassian.net/browse/P20-135)

## Testing story
* Unit tests
* [Storybook](http://localhost:9001/?path=/story/childaccountconsent--valid-token)
* Manually tested bin/dashboard-server
  * http://localhost-studio.code.org:3000/policy_compliance/child_account_consent
  * http://localhost-studio.code.org:3000/policy_compliance/child_account_consent?token=parent_permission_uuid_here

You can easily generate your own test parent permission token:
1. `bin/dashboard-console`
2. `include FactoryBot::Syntax::Methods`
3. `permission = create :parental_permission_request`
4. `permission.uuid`

## Deployment strategy
This is not hidden behind a feature flag because it is a new URL and we won't encounter issues if people find it.
## PR Checklist:

- [ ] Tests provide adequate coverage
- [ ] Privacy and Security impacts have been assessed
- [ ] Code is well-commented
- [ ] New features are translatable or updates will not break translations
- [ ] Relevant documentation has been added or updated
- [ ] User impact is well-understood and desirable
- [ ] Pull Request is labeled appropriately
- [ ] Follow-up work items (including potential tech debt) are tracked and linked

PR #59963, opened 2024-07-24
P20-1055: CAP data migration Phase 2

artem-vavilov:
Replaced user JSON properties `child_account_compliance_state`, `child_account_compliance_state_last_updated ` and `child_account_compliance_lock_out_date` with new indexed columns `cap_status` and `cap_status_date`.

## Links
- JIRA task: [P20-1055](https://codedotorg.atlassian.net/browse/P20-1055)
- Migration PR: https://github.com/code-dot-org/code-dot-org/pull/60218
- Phase 1 PR: https://github.com/code-dot-org/code-dot-org/pull/59962
- Related PR: https://github.com/code-dot-org/code-dot-org/pull/52244

### DIFF
--- a/dashboard/app/models/user.rb
+++ b/dashboard/app/models/user.rb
@@ -109,10 +109,6 @@ class User < ApplicationRecord
   #   us_state: A 2 letter code United States state code the user has given us.
   #   country_code: The country the user was in when they told us their
   #     us_state.
-  #   child_account_compliance_state: The state of a user's compliance with our
-  #     child account policy.
-  #   child_account_compliance_state_last_updated: The date the user became
-  #     compliant with our child account policy.
   #   ai_rubrics_disabled: Turns off AI assessment for a User.
   #   ai_rubrics_tour_seen: Tracks whether user has viewed the AI rubric product tour.
   #   lti_roster_sync_enabled: Enable/disable LTI roster syncing for a User.
@@ -149,9 +145,6 @@ class User < ApplicationRecord
     gender_student_input
     gender_teacher_input
     gender_third_party_input
-    child_account_compliance_state
-    child_account_compliance_state_last_updated
-    child_account_compliance_lock_out_date
     us_state
     country_code
     family_name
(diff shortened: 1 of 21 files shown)


bin/oneoff/backfill_data/user_cap_state_columns
mgc1194, line 1: Double-check that we want to delete the on-off script. Although git already keeps track of changes in the codebase, sometimes we want to keep these kind of scripts for the record.

dashboard/app/models/user.rb
```diff
@@ -147,9 +143,6 @@ class User < ApplicationRecord
     gender_student_input
     gender_teacher_input
     gender_third_party_input
-    child_account_compliance_state
-    child_account_compliance_state_last_updated
-    child_account_compliance_lock_out_date
```
mgc1194, line 154: Could you link to the PR that added these properties? I'm pretty sure there's documentation about these deleted columns, so I'd like to avoid asking questions I could read about :)

artem-vavilov: https://github.com/code-dot-org/code-dot-org/pull/52244

artem-vavilov: @daynew Do you happen to know anything about the documentation for these fields?

bin/oneoff/backfill_data/user_cap_state_columns
artem-vavilov: cc @daynew @wilkie

dashboard/app/models/user.rb
daynew: Nothing besides what is in the comments of this code.

bin/oneoff/backfill_data/user_cap_state_columns
daynew: @mgc1194 is right, let's keep the oneoff script for now. It will also serve as a good example for future engineers who might want to do something similar.

daynew (approved):
Besides keeping the oneoff script, these changes look good to me.

daynew:
We need to make we don't merge these changes until https://github.com/code-dot-org/code-dot-org/pull/58790 is run. We need the locked_out_date so we can backfill the lockout events.

PR #59969, opened 2024-07-24
CFU - Added in event tracking for summary view

kobryan0619:
Video showing event logging

https://github.com/user-attachments/assets/a23775b4-cca1-4522-905a-a8495ef59d6d

## Links
[Spec](https://docs.google.com/document/d/1Q942bLqf8dI08w8vwCzmk13RXD3zMFTBywAUQzpNDdc/edit#heading=h.qi8xxyq72bct)

### DIFF
--- a/apps/src/templates/levelSummary/SummaryResponses.jsx
+++ b/apps/src/templates/levelSummary/SummaryResponses.jsx
@@ -66,6 +66,13 @@ const SummaryResponses = ({
     [scriptData, selectedSection]
   );
 
+  const eventData = {
+    levelId: scriptData.level.id,
+    levelName: scriptData.level.name,
+    curriculumUmbrella: scriptData.reportingData.curriculumUmbrella,
+    unitId: scriptData.reportingData.unitId,
+  };
+
   useEffect(() => {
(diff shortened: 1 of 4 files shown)


apps/src/templates/levelSummary/SummaryResponses.jsx
```diff
@@ -66,6 +66,13 @@ const SummaryResponses = ({
     [scriptData, selectedSection]
   );
 
+  const eventData = {
```
kobryan0619: I did initially try getting this data all in redux in the FreeResponseResponses component but it got quite messy - it seemed like it might need to come from a few different redux stores.  I chose this strategy instead as all the data we needed was available in the scriptData.   I am still getting the hang of "when to pass through data vs when to get it from redux" so any additional considerations / reasoning would be appreciated.

lfryemason: This looks great. My only suggestion would be to wrap it in `React.useMemo` so that we only recalculate it when `scriptData` changes

lfryemason (approved):
Looks great, one non-blocking nit that I would like to see.

PR #59973, opened 2024-07-24
Python Lab: friendlier validation output

molly-moen:
Note: this PR should not be merged until [this package PR](https://github.com/code-dot-org/pythonlab-packages/pull/3) is in.

This PR uses the new `unittest_runner` package added in the linked PR above. It extracts out the logic for running unit tests, and makes validation tests have friendlier output. Student-written unit tests are unchanged (so far, we don't have a use case in data science for student-written unit tests).

Now validation tests can be written with a docstring, and that docstring will be the only thing students see when looking at test output. If there is no docstring, we fall back to the test name (default behavior). For example, a test can be written like so:
```
  def test_factorial_zero(self):
      """ Test that factorial of 0 should be 1"""
      self.assertEqual(factorial(0), 1, "Factorial of 0 should be 1")
```

And the test will appear to students as `Test that factorial of 0 should be 1` rather than `test_factorial_zero`.

### Before
Screenshot 2024-07-24 at 11 33 28 AM

### After
Screenshot 2024-07-24 at 11 27 33 AM

### With a failure
Screenshot 2024-07-24 at 11 27 16 AM
Note that the `/home/pyodide` in the stack trace is being fixed by #59960.

## Links

- jira ticket: [CT-658](https://codedotorg.atlassian.net/browse/CT-658)

## Testing story
Tested locally. This change only impacts python lab.

## Follow-up work
We are in discussions about how we will show validation results to students for the pilot, so I kept this very simple for now.

## PR Checklist:

- [ ] Code is well-commented


apps/src/pythonlab/pythonHelpers/scripts.ts
```diff
@@ -1,16 +1,16 @@
...
-
-def discover_and_run(start_dir, pattern):
-    loader = unittest.TestLoader()
-    test_suite = loader.discover(start_dir, pattern)
-    runner = unittest.TextTestRunner(verbosity=2)
-    result = runner.run(test_suite)
-
-discover_and_run('.', '${filePattern}')
+from unittest_runner import run_student_tests
+run_student_tests('test*.py')
```
snickell, line 14: 👍, nice using the module to keep python code mostly in python packages, I think the future will thank you for this

apps/src/pythonlab/pyodideWebWorker.js
```diff
@@ -20,6 +20,7 @@ async function loadPyodideAndPackages() {
       'numpy',
       'matplotlib',
       `/blockly/js/pyodide/${version}/pythonlab_setup-0.0.1-py3-none-any.whl`,
+      `/blockly/js/pyodide/${version}/unittest_runner-0.0.1-py3-none-any.whl`,
```
snickell, line 25: I wouldn't mind linking to the unittest_runner repo here since its outside the mono-repo, I can imagine somebody struggling to figure out where this code came from in several years.

snickell: Also... is there a reason to keep our python packages outside the mono-repo? I could see it hard finding a home for them, but there's no particular reason we have to keep "non ruby/js" code elsewhere. Having them in-repo would allow PRs like this one to be a single PR, so you can make lock-step changes to both sides of the "python wall" at the same time, and would improve searching. I don't think its a factor for this PR, but might be worth discussing out of band?

snickell (approved):
Appreciate putting the python code in a module so we can test it, edit it with syntax highlighting etc.

ebeastlake (approved):

apps/src/pythonlab/pyodideWebWorker.js
molly-moen: > I could see it hard finding a home for them

This was my main reason, afaik we don't have any python code in our repo, and this code would exist in some space other than apps or dashboard. I am happy to put them in the main repo if we have a good idea as to where that would be, but it seemed cleaner for now to have them split out. Agree it's worth discussing!

molly-moen: Also added link back to repo, good idea!

PR #59974, opened 2024-07-24
Make dsl editor fit content

kakiha11:
This change will allow the dsl editor field to auto fit it's content. This should work on all browsers except Firefox.

dsleditor

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/dashboard/app/views/levels/editors/fields/_dsl_fields.html.haml
+++ b/dashboard/app/views/levels/editors/fields/_dsl_fields.html.haml
@@ -4,7 +4,11 @@
   $lighter_cyan: #a6daed;
   $lightest_cyan: #d9eff7;
 
-  .script_text textarea { width: 100%; height: 100px}
+  .script_text textarea {
+    field-sizing: content;
+    max-width: 100%;
+    min-width: 95%;}
+
   .markdown {
     border: 5px solid gray;
     border-radius: 5px;
--- a/dashboard/app/views/levels/editors/fields/_dsl_fields.html.haml
+++ b/dashboard/app/views/levels/editors/fields/_dsl_fields.html.haml
@@ -6,8 +6,9 @@
 
   .script_text textarea {
     field-sizing: content;
-    max-width: 100%;
-    min-width: 95%;}
+    width: 100%;
+    box-sizing: border-box;
+  }
 
   .markdown {
     border: 5px solid gray;


dashboard/app/views/levels/editors/fields/_dsl_fields.html.haml
```diff
@@ -4,7 +4,11 @@
   $lighter_cyan: #a6daed;
   $lightest_cyan: #d9eff7;
 
-  .script_text textarea { width: 100%; height: 100px}
+  .script_text textarea {
+    field-sizing: content;
+    max-width: 100%;
+    min-width: 95%;}
```
snickell: I was confused by the need for setting a width related property, so I tried it out to check what was going on. `width: 100%` results in the right several pixels being truncated, which is a result of a manually specified "focus boxshadow" (the blue blur around it when focused), combined with a parent element with class=collapse having `overflow: hidden` set. Sort of a combination of ughs makes us do this ugh, which is common.

All that said, I think it'd be better and clearer to accomplish the same goal using just the direct `width` property (rather than setting a narrow mix and max) so like:
```
.script_text textarea {
  field-sizing: content;
  width: 98%;
}
```

(btw, 98% worked for me without cutting it off, and getting as close as possible makes the right edge less visually ragged, not that it matters much).

Also, ideally the `}` should be on the next line, normally a linter would enforce that but we gotta self-lint in HAML.

snickell (approved):

dashboard/app/views/levels/editors/fields/_dsl_fields.html.haml
wilkie: Need to have it include its own padding in its calculation of `100%`... not all browsers will have the same scrollbar width. I suggest using `border-box` sizing.

```suggestion
    width: 100%;
    box-sizing: border-box;}
```

wilkie: If you want to not allow resizing by width larger than the bounding box, keeping the `max-width` is reasonable.

kakiha11: Wilkie! Thanks for this suggestion :)

kakiha11: Thanks for playing around with this! I was following the guidance from the MDN docs to not use `width` and instead use `max-width` and `min-width` but this looks better!

PR #59981, opened 2024-07-24
Part 1 - View Aichat student chat history: teacher view student history tabs/toggle

fisher-alice:
This PR is part 1 in implementing teacher chat view history for aichat levels. A list of related follow-up tasks are outlined in [this doc](https://docs.google.com/document/d/1GXz3RwQPjYWhUK84xW-fKImWujVyB8wIE2MmTp5dtQs/edit).
~~Note that this PR will NOT be merged as-is since the student history is not retrieved and displayed yet.~~
I added an experiment so that this PR can be merged to `staging`. To enable the experiment, use 
'?enableExperiments=view_chat_history' at the end of the url on an aichat level. You can disable the experiment by adding
'?disableExperiments=view_chat_history' to the end of the url.

This PR updates the aichat workspace as follows:
- When a teacher clicks on the Teacher Panel, selects a section, and then selects a student, two tabs are displayed in the chat workspace: 'View [student name]'s chat history' and 'Test student model'.
- In the view mode, the chat message editor is NOT displayed and the 'Clear chat' button is disabled.
- In the test mode, the model customizations are still view only, but the teacher can test the chatbot with the student customizations. Clear chat and the message editor are available.
- When a student either clicks on 'Me' in the student list or clicks on 'Student' in the Teacher Panel toggle, the two tabs are no longer displayed in the workspace.

## After update

Screencast video:

https://github.com/user-attachments/assets/f795f3e5-cd49-4278-90ef-6777da315a0d

UPDATED TAB LABELS:

https://github.com/user-attachments/assets/dcb8111d-8152-491b-9677-845aae5ce33a

## Links
[jira](https://codedotorg.atlassian.net/browse/LABS-937)

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/aichat/types.ts
+++ b/apps/src/aichat/types.ts
@@ -59,6 +59,11 @@ export enum ViewMode {
   PRESENTATION = 'presentation-mode',
 }
 
+export enum WorkspaceTeacherViewTab {
+  STUDENT_CHAT_HISTORY = 'viewStudentChatHistory',
+  TEST_STUDENT_MODEL = 'testStudentModel',
+}
+
 export interface AichatLevelProperties extends LevelProperties {
   /**
    * Initial AI chat customizations set by the level.
--- a/apps/src/aichat/types.ts
+++ b/apps/src/aichat/types.ts
@@ -59,11 +59,6 @@ export enum ViewMode {
   PRESENTATION = 'presentation-mode',
 }
 
-export enum WorkspaceTeacherViewTab {
-  STUDENT_CHAT_HISTORY = 'viewStudentChatHistory',
-  TEST_STUDENT_MODEL = 'testStudentModel',
-}
-
 export interface AichatLevelProperties extends LevelProperties {
   /**
    * Initial AI chat customizations set by the level.
--- a/apps/src/aichat/views/ChatWorkspace.tsx
+++ b/apps/src/aichat/views/ChatWorkspace.tsx
@@ -16,6 +16,7 @@ import CopyButton from './CopyButton';
 import UserChatMessageEditor from './UserChatMessageEditor';
 
 import moduleStyles from './chatWorkspace.module.scss';
+import {ChatItem} from '../types';
 
 interface ChatWorkspaceProps {
   onClear: () => void;
@@ -53,6 +54,18 @@ const ChatWorkspace: React.FunctionComponent<ChatWorkspaceProps> = ({
 
   const dispatch = useAppDispatch();
 
+  const showWaitingAnimation = () => {
+    if (isWaitingForChatResponse) {
+      return (
+        <img
+          src="/blockly/media/aichat/typing-animation.gif"
+          alt={'Waiting for response'}
+          className={moduleStyles.waitingForResponse}
+        />
+      );
+    }
+  };
+
   const studentShortName = 'Sam';
   const tabs = [
     {
@@ -63,7 +76,13 @@ const ChatWorkspace: React.FunctionComponent<ChatWorkspaceProps> = ({
     {
       value: 'testStudentModel',
       text: 'Test student model',
-      tabContent: <div>Test student model here.</div>,
+      tabContent: (
+        <TestModel
+          items={items}
+          showWaitingAnimation={showWaitingAnimation}
+          conversationContainerRef={conversationContainerRef}
+        />
+      ),
     },
   ];
 
@@ -78,7 +97,7 @@ const ChatWorkspace: React.FunctionComponent<ChatWorkspaceProps> = ({
     onChange: handleOnChange,
     type: 'secondary',
     tabsContainerClassName: moduleStyles.tabsContainer,
-    tabPanelsContainerClassName: moduleStyles.tabPanels,
+    tabPanelsContainerClassName: moduleStyles.conversationArea,
   };
 
   const onCloseWarningModal = useCallback(
@@ -86,32 +105,11 @@ const ChatWorkspace: React.FunctionComponent<ChatWorkspaceProps> = ({
     [dispatch]
   );
 
-  const showWaitingAnimation = () => {
-    if (isWaitingForChatResponse) {
-      return (
-        <img
-          src="/blockly/media/aichat/typing-animation.gif"
-          alt={'Waiting for response'}
-          className={moduleStyles.waitingForResponse}
-        />
-      );
-    }
-  };
-
   return (
     <div id="chat-workspace-area" className={moduleStyles.chatWorkspace}>
       {showWarningModal && <ChatWarningModal onClose={onCloseWarningModal} />}
       <Tabs {...tabArgs} />
-      <div
-        id="chat-workspace-conversation"
-        className={moduleStyles.conversationArea}
-        ref={conversationContainerRef}
-      >
-        {items.map((item, index) => (
-          <ChatItemView item={item} key={index} />
-        ))}
-        {showWaitingAnimation()}
-      </div>
+
       <UserChatMessageEditor />
       <div className={moduleStyles.buttonRow}>
         <Button
@@ -124,6 +122,31 @@ const ChatWorkspace: React.FunctionComponent<ChatWorkspaceProps> = ({
         />
         <CopyButton />
       </div>
+    </div>
+  );
+};
+
+interface TestModelProps {
+  conversationContainerRef: React.RefObject<HTMLDivElement>;
+  items: ChatItem[];
+  showWaitingAnimation: () => React.ReactNode;
+}
+
+const TestModel: React.FunctionComponent<TestModelProps> = ({
+  items,
+  showWaitingAnimation,
+  conversationContainerRef,
+}) => {
+  return (
+    <div
+      id="chat-workspace-conversation"
+      className={moduleStyles.conversationArea}
+      ref={conversationContainerRef}
+    >
+      {items.map((item, index) => (
+        <ChatItemView item={item} key={index} />
+      ))}
+      {showWaitingAnimation()}
     </div>
   );
 };
--- a/apps/src/aichat/views/ChatWorkspace.tsx
+++ b/apps/src/aichat/views/ChatWorkspace.tsx
@@ -1,4 +1,4 @@
-import React, {useCallback, useRef, useEffect, useState} from 'react';
+import React, {useCallback, useEffect, useMemo, useRef, useState} from 'react';
 import {useSelector} from 'react-redux';
 
 import {
@@ -42,9 +42,6 @@ const ChatWorkspace: React.FunctionComponent<ChatWorkspaceProps> = ({
   onClear,
 }) => {
   const [viewMode, setViewMode] = useState<string | null>(null);
-  const [selectedStudentName, setSelectedStudentName] = useState<string | null>(
-    null
-  );
 
   const {showWarningModal, isWaitingForChatResponse} = useAppSelector(
     state => state.aichat
@@ -71,17 +68,16 @@ const ChatWorkspace: React.FunctionComponent<ChatWorkspaceProps> = ({
     }
   }, [messagesString, isWaitingForChatResponse]);
 
-  useEffect(() => {
-    let selectedStudentName = null;
+  const selectedStudentName = useMemo(() => {
     if (viewAsUserId) {
       const selectedStudent = Object.values(students).find(
         student => student.id === viewAsUserId
       );
       if (selectedStudent) {
-        selectedStudentName = getShortName(selectedStudent.name);
+        return getShortName(selectedStudent.name);
       }
     }
-    setSelectedStudentName(selectedStudentName);
+    return null;
   }, [viewAsUserId, students]);
 
   useEffect(() => {
--- a/apps/src/aichat/views/ChatWorkspace.tsx
+++ b/apps/src/aichat/views/ChatWorkspace.tsx
@@ -120,9 +120,11 @@ const ChatWorkspace: React.FunctionComponent<ChatWorkspaceProps> = ({
     {
       value: 'viewStudentChatHistory',
       text:
-        selectedTab === WorkspaceTeacherViewTab.STUDENT_CHAT_HISTORY
-          ? `${selectedStudentName}'s chat history (view only)`
-          : `View ${selectedStudentName}'s chat history`,
+        `${selectedStudentName}'s chat history` +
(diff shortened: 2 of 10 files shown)


apps/src/aichat/views/ChatWorkspace.tsx
```diff
@@ -50,12 +70,31 @@ const ChatWorkspace: React.FunctionComponent<ChatWorkspaceProps> = ({
...
+    if (viewAsUserId) {
+      const selectedStudent = Object.values(students).find(
+        student => student.id === viewAsUserId
+      );
+      if (selectedStudent) {
+        selectedStudentName = getShortName(selectedStudent.name);
+      }
+    }
+    setSelectedStudentName(selectedStudentName);
+  }, [viewAsUserId, students]);
```
thomasoniii: This is a double rendering anti-pattern.

Get rid of the `selectedStudentName` local variable and make it a `useMemo call instead.

```
const selectedStudentName = useMemo( () => {
    if (viewAsUserId) {
      const selectedStudent = Object.values(students).find(
        student => student.id === viewAsUserId
      );
      if (selectedStudent) {
        return getShortName(selectedStudent.name);
      }
    }
    return null
}, [viewAsUserId, students]);

File: apps/src/aichat/views/ChatWorkspace.tsx [hunk 2]
```diff
@@ -50,12 +70,31 @@ const ChatWorkspace: React.FunctionComponent<ChatWorkspaceProps> = ({
...
+    // or from when viewing workspace as a participant, default to the student chat history view.
+    if (
+      viewAsUserId &&
+      (!viewMode || viewMode === WORKSPACE_VIEW_MODE.PARTICIPANT)
+    ) {
+      setViewMode(WORKSPACE_VIEW_MODE.STUDENT_CHAT_HISTORY);
+    } else if (!viewAsUserId) {
+      setViewMode(WORKSPACE_VIEW_MODE.PARTICIPANT);
+    }
+  }, [viewAsUserId, viewMode]);
```
thomasoniii: This has the same render trap by setting the local state via a `useEffect`, but the fix is a little trickier since `setViewMode` is _also_ set via a callback. In this case, I think it should probably be in the viewMode initializer.

```
const [viewMode, setViewMode] = useState<string>( () => {
    if (
      viewAsUserId &&
      (!viewMode || viewMode === WORKSPACE_VIEW_MODE.PARTICIPANT)
    ) {
      return WORKSPACE_VIEW_MODE.STUDENT_CHAT_HISTORY;
    } else if (!viewAsUserId) {
      return WORKSPACE_VIEW_MODE.PARTICIPANT;
    }
});
```

With the caveat that this would then _not_ update if the `viewAsUserId` value changes. If that is needed and `viewAsUserId` can change, then it might be simpler to just leave the existing logic in place. I bet it can still be simplified, but I'm not sure how off the top of my head.

Regardless, the conditional in here may also be able to be simplified - since it's always initialized to a value, I dropped the `| null` typing, and I _think_ the `!viewMode ||` part of the conditional can come out, but I wasn't completely sure of that.

Anyway, def check my work and make sure it still works before just blindly dropping it in. 😉

thomasoniii (approved):
couple of notes about getting rid of some `useEffect` calls to prevent double renders, but overall lgtm 🚀

apps/src/aichat/views/ChatWorkspace.tsx
fisher-alice: Thanks Jim! Will update.

apps/src/aichat/views/ChatWorkspace.tsx
fisher-alice: Hey Jim - thanks for the comment!

We do need to update if `viewAsUserId` changes. If `viewAsUserId` is `null`, then the teacher is viewing lab as a participant: 'http://localhost-studio.code.org:3000/s/allthethings/lessons/47/levels/1?section_id=4&user_id&viewAs=Instructor' and we don't want to show the two tabs.

I added the `!viewMode || ` art of the conditional so that if the URL is loaded for the first time and the `user_id` param is included in the URL, then we do want to default to the chat history view so I think I'll leave thing as-is.

apps/src/aichat/views/ChatWorkspace.tsx
```diff
@@ -19,21 +22,36 @@ import moduleStyles from './chatWorkspace.module.scss';
...
   onClear: () => void;
 }
+interface Students {
+  [index: number]: {
+    id: number;
+    name: string;
+  };
+}
+
+const WORKSPACE_VIEW_MODE = {
```
sanchitmalhotra126: We should use an enum / union type for this so we can type the state value

Also - this feels a little overloaded to me since it determines both which tab is being shown as well as whether to show the tabs or not? I think it would be good to decouple those two sets of values - maybe if we can compute whether to show the tabs or not based off of `viewAsUserId` as mentioned below, these values can just determine which tab is being shown?

apps/src/aichat/views/ChatWorkspace.tsx
```diff
@@ -50,12 +68,30 @@ const ChatWorkspace: React.FunctionComponent<ChatWorkspaceProps> = ({
...
+    return null;
+  }, [viewAsUserId, students]);
 
-  const onCloseWarningModal = useCallback(
-    () => dispatch(setShowWarningModal(false)),
-    [dispatch]
-  );
+  useEffect(() => {
+    // If a teacher is viewing workspace as a student when level first loads (user_id param included in url)
+    // or from when viewing workspace as a participant, default to the student chat history view.
```
sanchitmalhotra126: A little confused by this - when you say "viewing the workspace as a participant" are you referring to the teacher clicking the "Student" toggle under "View as" in the teacher panel? Or the `WORKSPACE_VIEW_MODE.PARTICIPANT` referring to the state of this specific view?

In either case though - couldn't we just determine which workspace view mode to show purely from `viewAsUserId` (i.e. if `viewAsUserId` is defined, show the tabs, if not only show the chat)?

apps/src/aichat/views/ChatWorkspace.tsx
```diff
@@ -69,23 +105,78 @@ const ChatWorkspace: React.FunctionComponent<ChatWorkspaceProps> = ({
...
+    iconName: 'lock',
+    iconStyle: 'solid',
+  };
+
+  const tabs = [
+    {
+      value: 'viewStudentChatHistory',
+      text: `View ${selectedStudentName}'s chat history`,
+      tabContent: (
+        <div>Viewing {selectedStudentName}'s chat history - TODO</div>
```
sanchitmalhotra126, line 129: If this is going out to teacher(s) during our current pilot, we should probably have a more official message here and/or just hide these tabs behind an experiment flag until the pilot ends. Might be good to double check with @samantha-code

apps/src/aichat/views/ChatWorkspace.tsx
```diff
@@ -98,4 +189,37 @@ const ChatWorkspace: React.FunctionComponent<ChatWorkspaceProps> = ({
   );
 };
 
+interface TestModelProps {
+  conversationContainerRef: React.RefObject<HTMLDivElement>;
+  items: ChatItem[];
+  showWaitingAnimation: () => React.ReactNode;
+}
+
+const TestModel: React.FunctionComponent<TestModelProps> = ({
```
sanchitmalhotra126: Minor: I know "Test Model" is the name of this tab but it looks like a little confusing to me as a component name in this file for some reason 🙂 maybe something like `ModelChat` or `ChatWithModel` or something similar? It's also possible that this inner component could be the same for both viewing chat history and the current chat since it's just rendering a list of chat items.

apps/src/aichat/views/ChatWorkspace.tsx
```diff
@@ -98,4 +189,37 @@ const ChatWorkspace: React.FunctionComponent<ChatWorkspaceProps> = ({
...
+      {items.map((item, index) => (
+        <ChatItemView item={item} key={index} />
+      ))}
+      {showWaitingAnimation()}
+    </div>
+  );
+};
+
+const MAX_NAME_LENGTH = 15;
+const getShortName = (studentName: string): string => {
```
sanchitmalhotra126: nit: move this to a util file?

sanchitmalhotra126 (commented):
Thanks for getting started on this! A few comments, nothing major

apps/src/aichat/views/ChatWorkspace.tsx
fisher-alice: I wasn't planning to merge this PR until we had the actual chat history in place. We could use an experiment flag though. Do you have a preference?

apps/src/aichat/views/ChatWorkspace.tsx
fisher-alice: Yeah - I wasn't super excited about this name. Let's go with ChatWithModel'!

apps/src/aichat/views/ChatWorkspace.tsx
sanchitmalhotra126: Oh sure yep, I think either way is fine. If multiple people are planning on working on this, it might be good to have the experiment flag just so we don't have to branch off of PRs, but I'm also fine keeping it unmerged until the pilot is over.

apps/src/aichat/views/ChatWorkspace.tsx
fisher-alice: Good points! Decoupled, renamed, and cleaned up. Thanks!

apps/src/aichat/views/ChatWorkspace.tsx
fisher-alice: Hi again - I refactored and made thing a lot more simple. Thanks again.

apps/src/aichat/views/ChatWorkspace.tsx
fisher-alice: // If a teacher is viewing workspace as a student when level first loads (user_id param included in url)
    // or from when viewing workspace as a participant, default to the student chat history view.

I think what I mean was:
- when a teacher was viewing level, clicked on the Teacher Panel, and then clicked on 'Me'
- or when a teacher was on a level with just the level path.

Either way, I agree it's easier to just use `viewAsUserId` to show tabs or not. Just pushed the refactor. Thanks!

apps/src/aichat/views/ChatWorkspace.tsx
fisher-alice: Oh yeah - good point. I can add an experiment flag then since there are a handful of tasks that need to be completed by hopefully different folks and are a bit intertwined.

apps/src/aichat/types.ts
```diff
@@ -59,6 +59,11 @@ export enum ViewMode {
   PRESENTATION = 'presentation-mode',
 }
 
+export enum WorkspaceTeacherViewTab {
```
sanchitmalhotra126: Minor - this could just stay in `ChatWorkspace.tsx` since it's only used internally by that component

apps/src/aichat/views/ChatWorkspace.tsx
```diff
@@ -69,23 +106,81 @@ const ChatWorkspace: React.FunctionComponent<ChatWorkspaceProps> = ({
...
-        {items.map((item, index) => (
-          <ChatItemView item={item} key={index} />
-        ))}
-        {showWaitingAnimation()}
-      </div>
-      <UserChatMessageEditor />
+      {experiments.isEnabled(experiments.VIEW_CHAT_HISTORY) && viewAsUserId && (
+        <Tabs {...tabArgs} />
+      )}
+      {(!experiments.isEnabled(experiments.VIEW_CHAT_HISTORY) ||
```
sanchitmalhotra126: nit: ternary for this so you don't have to check the conditions twice?
```
{experiments.isEnabled(experiments.VIEW_CHAT_HISTORY) && viewAsUserId ? <Tabs .../> : <ChatWithModel .../>}
```

apps/src/aichat/views/ChatWorkspace.tsx
```diff
@@ -50,13 +65,35 @@ const ChatWorkspace: React.FunctionComponent<ChatWorkspaceProps> = ({
...
+      !experiments.isEnabled(experiments.VIEW_CHAT_HISTORY),
+    [selectedTab]
   );
 
+  useEffect(() => {
+    // If we are viewing as a student, default to the student chat history tab if tab is not yet selected.
+    if (viewAsUserId && !selectedTab) {
+      setSelectedTab(WorkspaceTeacherViewTab.STUDENT_CHAT_HISTORY);
+    } else if (!viewAsUserId) {
+      setSelectedTab(null);
```
sanchitmalhotra126, line 98: minor - do we need to actually null out the tab when the viewAsUserId is undefined? We're not going to show the tabs anyway right, so does the value actually matter?

Also, if a teacher switches from teacher view to participant view and then back to teacher view, do we want to show them the tab they were previously on, or reset to student chat history? Looks like here we will reset back to student chat history but without this useEffect, we would stay on the tab they were on previously

sanchitmalhotra126 (approved):
LGTM! Thanks for adding the experiment flag!

apps/src/aichat/views/ChatWorkspace.tsx
fisher-alice: Yeah - I intentionally set this so that the student chat history would be the first thing they see vs. testing the student model. Lm ask product what they prefer. Thanks!

fisher-alice: Checked with @samantha-code and curriculum and they both prefer having student chat history viewed as default after switching back to teacher view.

apps/src/aichat/views/ChatWorkspace.tsx
```diff
@@ -69,23 +111,84 @@ const ChatWorkspace: React.FunctionComponent<ChatWorkspaceProps> = ({
...
+  const iconValue: FontAwesomeV6IconProps = {
+    iconName: 'lock',
+    iconStyle: 'solid',
+  };
+
+  const tabs = [
+    {
+      value: 'viewStudentChatHistory',
+      text:
+        `${selectedStudentName}'s chat history` +
```
fisher-alice, line 123: Hey @sanchitmalhotra126, just a FYI that I updated this as requested by Mark Barrett after your most recent review.

PR #59982, opened 2024-07-24
Turn off delayed_job retries and log failed jobs

cat5inthecradle:
First, this adds a FailedJobCount metric, helpful for alerting. It also removes failed jobs from the JobCount metric.

Second, it turns off `delayed_job` retries, which will have us only using `ActiveJob` retries. We're seeing jobs with massive retry timeouts and lots of retry attempts because delayed_job is excessively retrying per it's [default config](https://stackoverflow.com/questions/75212317/does-rails-activejob-retry-on-work-with-delayed-job-gem/75298587#75298587).

- [Stack Overflow Post about double retry logic](https://stackoverflow.com/questions/75212317/does-rails-activejob-retry-on-work-with-delayed-job-gem/75298587#75298587)

## Testing story

TODO - tests may fail, but regardless I need to add tests for the new metric.

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/bin/cron/report_activejob_metrics
+++ b/bin/cron/report_activejob_metrics
@@ -12,7 +12,7 @@ def push_queue_length_metrics
     [
       {
         metric_name: 'JobCount',
-        value: Delayed::Job.count,
+        value: Delayed::Job.where(failed_at: nil).count,
         unit: 'Count',
         timestamp: Time.now,
         dimensions: [
--- a/bin/cron/report_activejob_metrics
+++ b/bin/cron/report_activejob_metrics
@@ -6,23 +6,29 @@ require_relative '../../dashboard/config/environment'
 
 METRICS_NAMESPACE = 'code-dot-org/ActiveJob'
 
-def push_queue_length_metric
-  job_count = Delayed::Job.count
-
+def push_queue_length_metrics
   Cdo::Metrics.push(METRICS_NAMESPACE,
+    # Same me trics as "dashboard/app/jobs/application_job.rb"
     [
       {
-        # Same metric as "dashboard/app/jobs/application_job.rb"
         metric_name: 'JobCount',
-        value: job_count,
+        value: Delayed::Job.count,
         unit: 'Count',
         timestamp: Time.now,
         dimensions: [
           {
-            name: 'Environment',
-            value: CDO.rack_env
+            name: 'Environment', value: CDO.rack_env
           }
         ]
+      },
+      {
+        metric_name: 'FailedJobCount',
+        value: Delayed::Job.where.not(failed_at: nil).count,
+        unit: 'Count',
+        timestamp: Time.now,
+        dimensions: [
+          {name: 'Environment', value: CDO.rack_env},
+        ],
       }
     ]
   )
@@ -51,7 +57,7 @@ def push_oldest_job_age_metric
 end
 
 def main
-  push_queue_length_metric
+  push_queue_length_metrics
   push_oldest_job_age_metric
 end
 
--- a/bin/cron/report_activejob_metrics
+++ b/bin/cron/report_activejob_metrics
@@ -11,7 +11,7 @@ def push_queue_length_metrics
     # Same me trics as "dashboard/app/jobs/application_job.rb"
     [
       {
-        metric_name: 'JobCount',
+        metric_name: 'PendingJobCount',
         value: Delayed::Job.where(failed_at: nil).count,
         unit: 'Count',
         timestamp: Time.now,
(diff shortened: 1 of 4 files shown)


cat5inthecradle:
I'm working on getting my local env setup for unit tests, so I haven't verified that these test changes are valid.

bin/cron/report_activejob_metrics
```diff
@@ -6,23 +6,29 @@ require_relative '../../dashboard/config/environment'
 
 METRICS_NAMESPACE = 'code-dot-org/ActiveJob'
 
-def push_queue_length_metric
-  job_count = Delayed::Job.count
-
+def push_queue_length_metrics
   Cdo::Metrics.push(METRICS_NAMESPACE,
+    # Same me trics as "dashboard/app/jobs/application_job.rb"
```
davidsbailey:
```suggestion
    # Same metrics as "dashboard/app/jobs/application_job.rb"
```

dashboard/app/jobs/concerns/active_job_metrics.rb
```diff
@@ -31,18 +31,27 @@ module ActiveJobMetrics
...
-    Cdo::Metrics.push(
-      METRICS_NAMESPACE, [
+    Cdo::Metrics.push(METRICS_NAMESPACE,
+      # Same metrics as "bin/cron/report_activejob_metrics"
+      [
         {
-          # Same metric as "bin/cron/report_activejob_metrics"
           metric_name: 'JobCount',
-          value: Delayed::Job.count,
+          value: Delayed::Job.where(failed_at: nil).count,
```
davidsbailey: what do you think about renaming this metric to PendingJobCount, to hopefully clarify that it does not include failed jobs? here and above

davidsbailey (approved):

wilkie (approved):
This looks good! That was good sleuthing and some strange defaults on the part of `Delayed::Job`, but what can you do.

cat5inthecradle:
@wilkie @davidsbailey I simplified one of the tests to get this passing, I'd prefer something more robust but I spent a little too much time banging my head against the wall on this. With approval I think it's good enough to ship, and I'll see if I can get some advice for improvement in Slack.

dashboard/test/jobs/application_job_test.rb
```diff
@@ -11,29 +11,16 @@ def perform
...
-    TestableJob.perform_later
-  end
-
-  test 'enqueued jobs log JobCount, WaitTime, ExecutionTime, and TotalTime' do
-    # after_enqueue metrics
-    Delayed::Job.stubs(:count).returns(1)
+  test 'enqueued jobs log PendingJobCount FailedJobCount WaitTime ExecutionTime and TotalTime' do
+    # TODO: actually assert the expected values for job counts
+    # When implementing, I ran into a lot of difficulty mocking
+    # the Delayed::Job.where(failed_at: nil) calls accurately.
```
davidsbailey: I don't recall ever having to mock these kinds of calls -- rather I have always relied on the database transaction to isolate my test case, and then done something like this: https://github.com/code-dot-org/code-dot-org/blob/43d2f4652e17d6919979541e8ad54438db0e3413/dashboard/test/controllers/programming_classes_controller_test.rb#L180-L183

davidsbailey (approved):
sorry the tests are giving you so much trouble! agree with getting this merged with reduced test coverage as a practical tradeoff. one alternative idea for how to implement the tests inline.

dashboard/test/jobs/application_job_test.rb
cat5inthecradle: I believe that `perform_later` during tests does not use the database, but maybe it should?

davidsbailey: hmm good point, I mistook this for something resembling active record.

now that you mention it, I believe we are not even using the delayed job adapter during unit tests, rather we are using:
* https://api.rubyonrails.org/classes/ActiveJob/QueueAdapters/TestAdapter.html
* https://api.rubyonrails.org/classes/ActiveJob/TestHelper.html

~~It looks like the test adapter has `ActiveJob::QueueAdapters::TestAdapter.enqueued_jobs`, in case that's helpful?~~

dashboard/app/jobs/concerns/active_job_metrics.rb
```diff
@@ -31,18 +31,27 @@ module ActiveJobMetrics
...
           unit: 'Count',
           timestamp: Time.now,
           dimensions: [
             {name: 'Environment', value: CDO.rack_env},
           ],
         },
+        {
+          metric_name: 'FailedJobCount',
+          value: Delayed::Job.where.not(failed_at: nil).count,
```
davidsbailey: ohh, I see now why you want to stub the `Delayed::Job.where` call, so that you can get the right value here. one idea is to introduce a helper method
```
def get_failed_jobs
  Delayed::Job.where.not(failed_at: nil).count
end
```
and just stub that method instead.

cat5inthecradle: That works! I added helper methods and a quick test for each. That should be making it's way through Drone now.

PR #59985, opened 2024-07-24
Sanitize introjs UI strings by wrapping them with React

davidsbailey:
Finishes https://codedotorg.atlassian.net/browse/AITT-612

Per discussion in [slack](https://codedotorg.slack.com/archives/CFTFD6BPV/p1721852673658879), the normal way to guard against potentially malicious HTML is to render it via React. today we are rendering some introjs strings without react, which exposes us to an XSS vulnerability (see screenshots). Fortunately, introjs also provides the option to pass a React component instead, which will take care of sanitizing the string.

## Screenshots

To see how this change addresses the issue, I temporarily added some malicious HTML to one of the product tour translations:
```diff
-  "rubricTourStepOneText": "Launch your AI Teaching Assistant from the bottom left corner of the screen.",
+  "rubricTourStepOneText": "Launch your AI Teaching Assistant from the bottom left corner of the screen. <img src=\"bogus\" onerror=\"alert('XSS')\" />",
```

### before

Screenshot 2024-07-25 at 1 55 20 PM

### after

Screenshot 2024-07-25 at 1 56 25 PM

## Testing story

* verified this protects against malicious HTML, as shown in screenshots
* added validation to ensure introjs step `intro` values are always wrapped in React components
* manually verified that the product tour still looks the same with this change, including step 3 which includes some legit HTML

### DIFF
--- a/apps/src/templates/rubrics/productTourHelpers.js
+++ b/apps/src/templates/rubrics/productTourHelpers.js
@@ -1,40 +1,61 @@
+import React from 'react';
+
 import i18n from '@cdo/locale';
 import evidenceDemo from '@cdo/static/ai-evidence-demo.gif';
+
+// intro.js-react allows a string or a react component for the intro prop.
+// Providing a string that was written by a translator is risky, because it
+// could contain malicious HTML. This helper method wraps the string in a react
+// component, which will take care of sanitizing the string.
+const sanitize = unsafe => <span>{unsafe}</span>;
 
 export const INITIAL_STEP = 0;
 export const STEPS = [
   {
     element: '#ui-floatingActionButton',
     title: i18n.rubricTourStepOneTitle(),
-    intro: i18n.rubricTourStepOneText(),
+    intro: sanitize(i18n.rubricTourStepOneText()),
   },
   {
     element: '#tour-ai-assessment',
     title: i18n.rubricTourStepTwoTitle(),
-    intro: i18n.rubricTourStepTwoText(),
+    intro: sanitize(i18n.rubricTourStepTwoText()),
   },
   {
     element: '#tour-ai-evidence',
-    title: i18n.rubricTourStepThreeTitle(),
+    title: i18n.rubricTourStepThreeTitle() + ' sanitized',
     position: 'top',
-    intro: `<p>${i18n.rubricTourStepThreeText()}</p><img src=${evidenceDemo}>`,
+    intro: (
+      <span>
+        <p>{i18n.rubricTourStepThreeText()}</p>
+        <img src={evidenceDemo} alt="evidence highlighting example" />
+      </span>
+    ),
   },
   {
     element: '#tour-ai-confidence',
     title: i18n.rubricTourStepFourTitle(),
-    intro: i18n.rubricTourStepFourText(),
+    intro: sanitize(i18n.rubricTourStepFourText()),
   },
   {
     element: '#tour-evidence-levels',
     title: i18n.rubricTourStepFiveTitle(),
-    intro: i18n.rubricTourStepFiveText(),
+    intro: sanitize(i18n.rubricTourStepFiveText()),
   },
   {
     element: '#tour-ai-assessment-feedback',
     title: i18n.rubricTourStepSixTitle(),
-    intro: i18n.rubricTourStepSixText(),
+    intro: sanitize(i18n.rubricTourStepSixText()),
   },
 ];
+
+STEPS.forEach((step, index) => {
+  if (typeof step.intro === 'string') {
+    throw new Error(
+      `Step ${index} intro has type 'string'. Please wrap it in a react component or a call to sanitize().`
+    );
+  }
+});
 
 // Dummy props for product tour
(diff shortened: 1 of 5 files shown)


stephenliang (approved):

apps/src/templates/rubrics/productTourHelpers.js
```diff
@@ -1,41 +1,62 @@
...
   {
     element: '#tour-ai-assessment',
     title: i18n.rubricTourStepTwoTitle(),
-    intro: i18n.rubricTourStepTwoText(),
+    intro: sanitize(i18n.rubricTourStepTwoText()),
   },
   {
     element: '#tour-ai-evidence',
-    title: i18n.rubricTourStepThreeTitle(),
+    title: i18n.rubricTourStepThreeTitle() + ' sanitized',
```
stephenliang: Should the `sanitized` suffix be in i18n?

apps/src/templates/rubrics/productTourHelpers.js
```diff
@@ -1,41 +1,62 @@
...
     position: 'top',
-    intro: `<p>${i18n.rubricTourStepThreeText()}</p><img src=${evidenceDemo}>`,
+    intro: (
+      <span>
+        <p>{i18n.rubricTourStepThreeText()}</p>
+        <img src={evidenceDemo} alt="evidence highlighting example" />
```
stephenliang: Should the alt text also be i18n?

apps/src/templates/rubrics/productTourHelpers.js
```diff
@@ -1,41 +1,62 @@
+import React from 'react';
+
 import i18n from '@cdo/locale';
 import evidenceDemo from '@cdo/static/ai-evidence-demo.gif';
 
+// intro.js-react allows a string or a react component for the intro prop.
+// Providing a string that was written by a translator is risky, because it
+// could contain malicious HTML. This helper method wraps the string in a react
+// component, which will take care of sanitizing the string.
+const sanitize = unsafe => <span>{unsafe}</span>;
```
stephenliang: I'm not 100% sure if this will work, but using a react fragment instead of a span might trigger React's virtual DOM to parse the unsafe without having to add a dummy element, which should lighten the DOM

```suggestion
const sanitize = unsafe => <>{unsafe}</>;
```

stephenliang (approved):

apps/src/templates/rubrics/productTourHelpers.js
davidsbailey: oops! reverted. this was debug code

apps/src/templates/rubrics/productTourHelpers.js
davidsbailey: Yes! good catch

apps/src/templates/rubrics/productTourHelpers.js
davidsbailey: Fragments seem to work! I verified that the product tour still renders correctly, and that this approach still escapes malicious HTML.

apps/src/templates/rubrics/productTourHelpers.js
```diff
@@ -1,41 +1,62 @@
...
     element: '#tour-ai-assessment-feedback',
     title: i18n.rubricTourStepSixTitle(),
-    intro: i18n.rubricTourStepSixText(),
+    intro: sanitize(i18n.rubricTourStepSixText()),
   },
 ];
 
+STEPS.forEach((step, index) => {
+  if (typeof step.intro === 'string') {
+    throw new Error(
```
stephenliang: nit: I suspect this would cause either the tours/the whole page to crash? I wonder if a we did have a malicious string we could possibly blank it out instead of throwing since the i18n could result in a denial of service if this happened. Looks good though!

stephenliang (approved):

apps/src/templates/rubrics/productTourHelpers.js
wilkie: Yes, this should crash the main thread. I'm fairly sure that this would happen regardless of the localized string and instead just ensures that you have wrapped the string in the `sanitize` call. Seems reasonable, but could also be something you write a test for instead, considering we export `STEPS`.

davidsbailey: Good points, I've moved this to a unit test. if you forget to sanitize like this
```diff
diff --git a/apps/src/templates/rubrics/productTourHelpers.js b/apps/src/templates/rubrics/productTourHelpers.js
index 5adc88ca99e..cc2b7780ce5 100644
--- a/apps/src/templates/rubrics/productTourHelpers.js
+++ b/apps/src/templates/rubrics/productTourHelpers.js
@@ -14,7 +14,7 @@ export const STEPS = [
   {
     element: '#ui-floatingActionButton',
     title: i18n.rubricTourStepOneTitle(),
-    intro: sanitize(i18n.rubricTourStepOneText()),
+    intro: i18n.rubricTourStepOneText(),
   },
```
the test will catch it like this
Screenshot 2024-07-25 at 8 54 34 PM

PR #59990, opened 2024-07-25
[Google Blockly] remove logic_compare type check

mikeharv:
The `logic_compare` block returns a boolean after comparing values from two connected blocks. In CDO Blockly, these can be any types, even mismatched:

image

Mainline Google Blockly provides this same block but with an additional check to ensure the types match:
https://github.com/google/blockly/blob/ebb56b2ce814f113b469f4cf4fb7f76d7a3c4d99/blocks/logic.ts#L599-L647

The new default behavior is that on any change, child blocks with mismatched types will get unplugged:

This is problematic because the block is explicitly used to compare string characters with numbers within a hidden function definition used in 100+ Artist Binary levels. Updating the levels isn't an option either because students will already have these blocks saved in their level sources. Start blocks shown below with CDO Blockly:

image

With Google Blockly's tighter type checking, these child blocks get disconnected:
image

Worse still, these blocks are hidden from students, so they'd have no way of even fixing them.

Fortunately, the change is as simple as extending this `onchange` function and making it do nothing.

With these changes, the block now accepts child blocks regardless of type:
image

## PR Checklist:

- [ ] Code is well-commented


apps/src/blockly/addons/extensions/logic_compare.ts
```diff
@@ -0,0 +1,31 @@
...
+  onchange: (this: CompareBlock) => void;
+};
+
+/**
+ * This override intentionally removes dynamic type validation for the left
+ * and right sides of a logic_compare (? == ?) block.
+ * Artist binary levels and existing student work use this block to compare
+ * characters from strings (ex. '01010101') with the numbers 0 and 1.
+ */
+const LOGIC_COMPARE_ONCHANGE_MIXIN = {
```
molly-moen, line 14: will removing this check cause any issues (thinking about all our blockly-related console errors)? Or do we work around any such issues?

mikeharv: It shouldn't. The original check only serves to unplug blocks with mismatching types. Most other blocks do not have an `onchange` method.

Which console errors are you thinking about?

molly-moen: I didn't know about a specific one, but I was wondering if blockly auto-disconnects them because if they stay connected something bad happens (such as a console error). If not, disregard this!

mikeharv: Oh, I see!

No, nothing bad happens. We still generate valid code and everything works as far as I can tell. :)

molly-moen (approved):

PR #59993, opened 2024-07-25
Remove inline attribute from blocks in levels

mikeharv:
In Course 4, and some older versions of CSF, there is an “Artist Binary” lesson: https://studio.code.org//s/course4/lessons/18/levels/2

The `draw` block is a function call block with a parameter named `binary`. Currently, the second (parameter) input is not inline with CDO Blockly:
image (210)

 However, I noticed that it _is_ inline with Google Blockly"
image

What’s strange here is that the block XML for the level suggests that the inputs should be inline:
image (212)

This makes me believe that it’s technically a bug with CDO Blockly since we seem to be ignoring `inline="true"`.

The block still works exactly the same way, but looks a little different. When we flip Artist, perhaps in August, this would change for any version of this lesson.
This proposes removing every instance of `inline="true"`  from these levels. This doesn't cause any regressions because CDO Blockly is ignore the attribute. This would get us to a place where the Google Blockly block inputs are not inline either:

image (213)

This branch includes similar changes in many levels, not all of which are in Artist Binary. Some of the others come from Play Lab lessons, etc. and for these I spot-checked to confirm that we never use inline inputs.

While discussing this issue on [Slack](https://codedotorg.slack.com/archives/C060J3A3BBK/p1721864106889939), it was pointed out that these lessons may not get enough use to make this work a strict requirement. However, it extends to other lessons beyond Artist Binary and the fix is straight-forward and low risk. There was also an open-ended question about potentially re-using this lesson in the upcoming version of CSF, so it feels appropriate to just make sure it will work the same way it always has.

Note that the .level file edited in the first commit shows additional changes. These are not significant and are just the result of editing the level directly with levelbuilder tools so that I could see the changes right away. The others were edited in my editor and require re-seeding to manually test.

## Links
Slack thread: https://codedotorg.slack.com/archives/C060J3A3BBK/p1721864106889939

## PR Checklist:

- [ ] Code is well-commented


molly-moen (approved):
Skimmed these, looks good!

PR #59995, opened 2024-07-25
Remove new teacher email and mailer

bethanyaconnor:
Removes the old teacher welcome email and all related code! 🚮 I'm relying on tests, plus a simple grep, to make sure nothing was missed here.


kelbyhawn (approved):
🗑️ 🔥

PR #60000, opened 2024-07-25
Move DPA  and cookie links into privacy directory

kelbyhawn:
Moves the following pages in to the `/privacy` directory:
- https://code.org/dpa-mt
- https://code.org/dpa-ny
- https://code.org/dpa
- https://code.org/cookies

## Links
Jira ticket: [ACQ-2111](https://codedotorg.atlassian.net/browse/ACQ-2111)

## Testing story
Tested locally


dmcavoy:
Was wondering for these if we should make a dpa directory so it would be code.org/privacy/dpa and code.org/privacy/dpa/mt etc

kelbyhawn:
> Was wondering for these if we should make a dpa directory so it would be code.org/privacy/dpa and code.org/privacy/dpa/mt etc

@dmcavoy done!

hannahbergam (approved):
These look good to me!

PR #60007, opened 2024-07-25
[Artist/Maze] Get all code from student+initialization blocks

mikeharv:
This makes sure that our code generation is complete and accurate for Artist/Maze on Google Blockly for two scenarios:
1. Running student code, with or without code from a level's initialization blocks
2. Running code from a level's solution blocks in the background to create the target image (Artist only)

This work was initially scoped to just Artist since Maze has already been migrated. However, I found that our approach to code generation used there wasn't working in certain Artist levels. The important characteristics of these levels/projects are:
* Function call blocks on the main workspace
* Function definition blocks on the hidden workspace
* At least one variable is used only within hidden functions (and not on the main workspace)

To my knowledge, the third bullet isn't something we do in any Maze level, but it does happen in Artist Binary levels where start blocks include definitions like this:

image

What's important above is that the `draw` function uses the variables `bit` and `i`, which will only be used on the hidden workspace. (There is also a `binary` variable due to the function parameter which will be used on both workspaces.)

In this case, if we only ever initialize the generator on the main workspace it never supplies the code with `var bit; var i;` and so the JS interpreter hits a runtime due to undefined variables. The solution ended up being a little tricky - at least it was tricky to come up with it. 

Previously, for Maze, we get the needed code from `blocksToCode` which initializes the generator on the main workspace only: https://github.com/code-dot-org/code-dot-org/blob/de8090c481f683af1ff734a66a871d66fddd347f/apps/src/blockly/addons/cdoGenerator.ts#L40-L62
Now, we use the new util `getAllGeneratedCode` which gets the finished generated code for each workspace that we have and then concatenates it with any code from initialization blocks. 
For CDO Blockly, this also changes things slightly, although the result is the same. Previously, CDO Blockly would add the initalization blocks to the list of student blocks and generate code for all of them at once. Now it will also generate code for the initialization blocks separately (handled in StudioApp) and add that code string to generated code string for the student blocks. Altering this flow slightly allows the util to work for either Blockly version and take the same one parameter (the string `extraCode` for the initialization blocks code).

For solution blocks, we need to handle generating the code behind the scenes, because the blocks are not rendered. For this, we can use a single headless workspace and we only need to initialize the generator once. This is handled by `getCodeFromBlockXmlSource`.

### Bee example
This shows a bee level being completed, demonstrating that we can still execute complex student code correctly (no regressions):
image

### Artist Initialization blocks
This CDO Blockly screenshot shows that we can still execute complex Artist code correctly, including for initialization blocks (which set the color to white for this level):
image

image

The same level, now working with Google Blockly:
image

### Artist Binary example
The dark yellow squares on the canvas are evidence that the solution blocks are working as expected:
image

The white squares show that we successfully executed the students code, which happens to match the solution image:
image

Mid-execution with CDO Blockly to demonstrate no regressions:
image

## PR Checklist:

- [ ] Code is well-commented


apps/src/blockly/addons/cdoUtils.ts
```diff
@@ -764,3 +764,39 @@ export function toolboxWithoutIds(
...
+// The students blocks are considered to be any on the main or hidden workspaces.
+export function getAllGeneratedCode(extraCode?: string) {
+  let code = extraCode || '';
+
+  [Blockly.getHiddenDefinitionWorkspace(), Blockly.getMainWorkspace()].forEach(
+    workspace => {
+      if (workspace) {
+        Blockly.getGenerator().init(workspace);
+        const blocks = workspace.getTopBlocks(true);
+        const blocksCode = [] as Block[];
```
molly-moen: nit: you can write this as 
```
const blocksCode: Block[] = [];
```
to avoid needing to cast the value

apps/src/blockly/addons/cdoUtils.ts
```diff
@@ -764,3 +764,39 @@ export function toolboxWithoutIds(
   removeIdsFromBlocks(toolboxDom);
   return Blockly.Xml.domToText(toolboxDom);
 }
+
+// Sets the lab code based on the student's blocks and any extra (e.g. initialization) code.
+// The students blocks are considered to be any on the main or hidden workspaces.
+export function getAllGeneratedCode(extraCode?: string) {
+  let code = extraCode || '';
```
molly-moen, line 771: do we need a `\n` after `extraCode`?

molly-moen (approved):
Looks good!

apps/src/blockly/addons/cdoUtils.ts
mikeharv: It doesn't seem like it. In the example level with the moon and stars above, this is the code we end up with:

```
"\n\n\n\n\n\n\nTurtle.penColour('#ffffff', 'block_id_draw-color');\n\n\n\nfunction draw_a_star() {\n    executionInfo.checkTimeout(); if (executionInfo.isTerminated()){return;}\n  for (var count = 0; count < 8; count++) {\n      executionInfo.checkTimeout(); if (executionInfo.isTerminated()){return;}\n    Turtle.moveForward(25, 'block_id_,tQB)s0?x}Be~N-pJzb2');\n    Turtle.moveBackward(25, 'block_id_f1I=YG6VeM5.6XcQP?~t');\n    Turtle.turnRight(45, 'block_id_cNG}r$iGr2!s2%ebX1nA');\n  }\n}\n\n\n\ndraw_a_star();\n\n\n"
```

which amounts to this:
```







Turtle.penColour('#ffffff', 'block_id_draw-color');



function draw_a_star() {
    executionInfo.checkTimeout(); if (executionInfo.isTerminated()){return;}
  for (var count = 0; count < 8; count++) {
      executionInfo.checkTimeout(); if (executionInfo.isTerminated()){return;}
    Turtle.moveForward(25, 'block_id_,tQB)s0?x}Be~N-pJzb2');
    Turtle.moveBackward(25, 'block_id_f1I=YG6VeM5.6XcQP?~t');
    Turtle.turnRight(45, 'block_id_cNG}r$iGr2!s2%ebX1nA');
  }
}



draw_a_star();


```

I think we're already super generous with new lines.

PR #60008, opened 2024-07-25
[Google Blockly] override procedure methods

mikeharv:
I noticed that some Artist levels weren't loading due to issue with function parameters:

image
image

Evidently, this was due to the solution blocks of a level which contained a function with parameters. These blocks are loaded onto an unrendered workspace, and for some reason the function definition hadn't been initialized properly with a procedure model. At the time when we would try to get the variables, the model did exist however. I was able to get past the crash by patching `getProcedureModel` so that it would find the model before returning it, if it hadn't already been assigned.

Because the function in question was also part of the start blocks, I was able to see a new related issue that resulted in child blocks being disconnected from the parameter input:

image

During deserialization and paramater updates, all parameter inputs are apparently removed from the block temporarily and then re-added based on the up-to-date model. The child blocks are able to be restored based on `.prevParams_` which is populated during initialization. For some reason, we were refreshing the inputs on these blocks before managing to initialize the block with a model. Without no model, there were no parameters to keep track off. The second fix here is make sure that we have a list of previous parameters before we take the inputs apart. If we don't, we can find them on the model. If for some reason the model didn't exist yet, the first fix described above allows us to find it.

Both of these changes together allow the page to load, without errors and without any blocks being removed from their expected inputs.

The root cause isn't currently known. There's a lot of complexity in our multiple workspaces, rendered and not, hidden and visible, and solution blocks add one more wrinkle. That being said, these changes should carry much risk, other than the need to maintain the overrides. Both can be thought of as double-checking that we have the right values before attempting to process with them.

## PR Checklist:

- [ ] Code is well-commented


apps/src/blockly/customBlocks/googleBlockly/proceduresBlocks.ts
```diff
@@ -276,6 +277,42 @@ GoogleBlockly.Extensions.register(
...
+
+      /**
+       * Makes sure that if we are updating the parameters before any move events
+       * have happened, the args map records the current state of the block. Does
+       * not remove entries from the array, since blocks can be disconnected
+       * temporarily during mutation (which triggers this method).
+       */
+      syncArgsMap_: function (this: ProcedureBlock) {
+        // If we haven't yet stored the previous parameters, do so now. This would
+        // normally happen when we or initialize the procedure block with a model
```
molly-moen: nit: remove or from `when we or initialize`?

apps/src/blockly/customBlocks/googleBlockly/proceduresBlocks.ts
```diff
@@ -276,6 +277,42 @@ GoogleBlockly.Extensions.register(
...
+        // or update its parameters.
+        if (!this.prevParams_.length) {
+          this.prevParams_ = [...this.getProcedureModel().getParameters()];
```
molly-moen: could `this.getProcedureModel()` technically return `null` here still if we don't find a procedure model?

molly-moen (commented):
One null check question, otherwise looks good!

apps/src/blockly/customBlocks/googleBlockly/proceduresBlocks.ts
mikeharv: Technically, yes, it could, although I haven't found a way to hit that. This `syncArgsMap_` function is only called when parameters are updated, and the models should at least exist. Happy to add a null check. I think I can just update the spread part to `...(this.getProcedureModel()?.getParameters() || [])`

molly-moen: yep might as well add the null check

molly-moen (approved):

PR #60011, opened 2024-07-25
Renames And I Wait For the Page To Fully Load to be more specific

hannahbergam:
The step definition 'and I wait fo the page to fully load' is used very widely across our .feature files (as evidenced by the files changed count here). However, I think the name doesn't quite capture what the step does. It waits for the run button, checks for a user header, and dismisses instructions if they exist. While this is super useful, it doesn't work on pages that aren't on a lab (that example, has no run button), and will fail if engineers try to use it elsewhere. 

This PR renames the step (and updates the files that call it) to 'and I wait for the lab page to fully load' so that an engineer will be less likely to try (and fail) to use it elsewhere.

I did this via a find and replace in vscode. 

Screenshot 2024-07-25 at 10 49 31 AM

## Links

- spec: []()
- jira ticket: https://codedotorg.atlassian.net/browse/ACQ-2131

## PR Checklist:

- [ ] Code is well-commented


TurnerRiley (approved):
🎉

molly-moen (approved):
Nice!

PR #60012, opened 2024-07-25
Redeliver "Firehose" Application Event Records to Redshift

sureshc:
Script that iterates through batches of "Firehose" application event records that failed delivery to Redshift and redelivers those batches by truncating record Properties whose length exceeded the maximum length of the destination Redshift column and which caused the entire batch to fail.

The script iterates in the intermediate S3 bucket for the Firehose stream through Objects that contain batches of application event records that previously failed delivery to Redshift. Specifically, it fetches and inspects error manifest objects that identify batches that failed delivery to Redshift.

This script attempts to validate all Properties against the schema of the destination Redshift `analysis.events` table. Specifically validate that:
- Required Properties are populated for destination columns that are NOT nullable
- Properties have the correct datatype for the destination column. Convert Properties with a "complex" value (datatype is an Object) to the JSON string representation, since our client applications emitting these event records sometimes inadvertently publish a JSON object in a string field.
- Enforce byte length constraints in Properties where the destination column is `CHAR` ("character") or `VARCHAR` (varying character) by accounting for multibyte characters.
- Ensure that multibyte characters are not inserted into `CHAR` ("character") columns.

Here's the schema of the destination Redshift table:

``` sql
CREATE TABLE analysis.events (
    created_at timestamp without time zone NOT NULL ENCODE raw,
    environment character(128) NOT NULL ENCODE raw,
    study character(128) NOT NULL ENCODE raw,
    study_group character(128) ENCODE raw,
    device character(1024) ENCODE raw,
    uuid character(128) ENCODE raw,
    user_id integer ENCODE raw,
    script_id integer ENCODE raw,
    level_id integer ENCODE raw,
    project_id character(128) ENCODE raw,
    event character(128) NOT NULL ENCODE raw,
    data_int integer ENCODE raw,
    data_float double precision ENCODE raw,
    data_string character(4096) ENCODE raw,
    data_json character varying(65535) ENCODE raw
)
DISTSTYLE EVEN;
```

## Testing story

Provision an adhoc and configure with Redshift credentials to test the script to a test Redshift table: `analysis.application_events`

These [CloudWatch Logs](https://us-east-1.console.aws.amazon.com/cloudwatch/home?region=us-east-1#logsV2:log-groups/log-group/adhoc-syslog/log-events/i-03e2aa317bdd42439$3FfilterPattern$3DFIREHOSE$26start$3D1728544200000$26end$3D1728550800000) are from a test run that reprocessed all failed Firehose events that originally failed delivery to Redshift on 2024-05-09.

[This is the redelivery status log CSV](https://docs.google.com/spreadsheets/d/1j9UiANUqbHZLdAuspeyCHYE7L65NdCH_8don5ZKkMIo/edit?gid=76621162#gid=76621162) generated from the above reprocess test.

## Follow-up work

1.  Modify `/lib/cdo/firehose.rb` to validate application event records before publishing them to Firehose
2. Port the JSON to Redshift validation functionality to Javascript so that we can validate application events published by our browser applications.
3.  Implement a Lambda that executes on the Firehose stream that validates records.
4. Modify the client and web application server logic that is publishing invalid records.

## PR Checklist:

- [ ] Code is well-commented


lib/test/cdo/test_json_to_redshift_table_validator.rb
```diff
@@ -0,0 +1,318 @@
+require_relative '../test_helper'
+require 'json'
+require 'cdo/json_to_redshift_table_validator'
+require 'active_support'
+require 'active_support/core_ext/string/multibyte'
+
+class JSONtoRedshiftTableValidatorTest < Minitest::Test
+  def setup
+    @redshift_schema = {
```
cat5inthecradle, line 9: Is this schema defined elsewhere in the codebase? If this needs to be kept in sync with that, then it might be good to include a comment connecting the duplicated code.

sureshc: This schema is only used for these unit tests. It happens to be the same schema for the only code we currently have that uses the JSON-to-Redshift validator (the Firehose redelivery script), but over time this schema will evolve independently and does not need to be explicitly linked to other parts of our codebase.

PR #60013, opened 2024-07-25
apps organization: make legacySharedComponents folder and move Button into it

Erin007:
In general the apps directory isn’t well organized, which makes grokking our frontend time-consuming and confusing. If you're developing on the frontend, you need to routinely ask questions like: 
- Does this component or one similar already exist? 
- If it does, where is it? 
- Oh, wait, which of these 3 similar components should I use? 

For example, Slack history is riddled with questions like this: 

Screenshot 2024-07-25 at 2 44 16 PM

The good news is we're in the process of building out a robust component library based on a Design System. 🎉 

In the meantime, we're living in a hybrid state where only some planned components are in the new DSCO component library and those new components are only being used in some places so far. 

This PR is an initial step to reorganizing our apps directory to make it more clear from reading the codebase which components should be used, which should be updated and which should be avoided. I've done 3 main things: 

1.) Made a `legacySharedComponents` folder in the apps directory
2.) Wrote a little README about what should live in this new folder
3.) Moved our old `Button` component into the `legacySharedComponents` 

If we like this direction, we can move additional old components here as we sunset them.


apps/src/applab/ExternalRedirectDialog.jsx
```diff
@@ -7,7 +7,7 @@ import {studio} from '@cdo/apps/lib/util/urlHelpers';
 import Dialog, {Body} from '@cdo/apps/templates/Dialog';
 import i18n from '@cdo/locale';
 
-import Button from '../templates/Button';
```
ebeastlake, line 10: If there's nothing else in this folder, should the new path just be...
```
'../legacySharedComponents/Button'
```
?

thomasoniii (approved):
Looks good. Awesome start.

Two potential follow up tasks -
- I'm in favor of `Foo/index.tsx` style components, and I think it'd be nice here because then it could just be imported as `legacySharedComponents/Button` (and not `button/Button`). I don't know how much argument that's gonna generate, so we may need a broader discussion.
- I'd also say that generic components should _always_ be referenced via an absolute path (`@cdo/apps/legacySharedComponnets/button/Button` in this case) because individual app structure is more likely to vary and it'll lead to needless refactoring. I get that it's not changed here because it's the existing way it was imported, and there's a case to be made to do that mod in a follow up PR so as to compartmentalize work. Just noting it so we can hopefully follow up on it and make that change as well.

apps/src/applab/ExternalRedirectDialog.jsx
ebeastlake: It looks like @thomasoniii said something similar below 😊

apps/test/unit/templates/ButtonTest.js
```diff
@@ -4,7 +4,7 @@ import sinon from 'sinon'; // eslint-disable-line no-restricted-imports
 
 import Button from '@cdo/apps/templates/Button';
 
-import style from '@cdo/apps/templates/button.module.scss';
```
levadadenys, line 7: nit: Should we also create `legacySharedComponents` folder in `test/unit` and move this test file there?

levadadenys (commented):
😍 LGTM in general! Thanks a lot for this update, it should make life a lot easier! 🎉 
Just one question about organizing tests

kelbyhawn (approved):
I love an update like this!! 🎉

PR #60014, opened 2024-07-25
[Google] move number validators to cdoUtils (+ show "invisible" blocks in start mode)

mikeharv:
While trying to grab a screenshot of some blocks for another branch, I realized that "invisible" blocks are always moved to the hidden workspace, even if we're in start mode. This makes it impossible for them to be managed by levelbuilders outside of editing the XML on the level edit page. After adding a quick `isStartMode` check to fix this, I was hit by an error relating to `wrapNumberValidatorsForLevelBuilder`. The number validators are part of the `FieldTextInput` class in CDO Blockly and are used to ensure that students don't enter invalid values in number blocks. They are wrapped for levelbuilders so that the string `'???'` can be used where a number would normally be expected. This is a common practice for default field values in the toolbox and/or start blocks.
These validators are not part of the Google Blockly class of the same name, likely because there's now a separate `FieldNumber` class to use for this sort of thing. We _could_ add these validators ourselves if we extended `GoogleBlockly.FieldTextInput`, but they don't really make sense to be there any more. Instead I moved the functions to `cdoUtils`. The CDO Blockly versions (in the CDO wrapper) just call the originals. The Google Blockly versions (in `cdoUtils.ts`) do the same things. 
Starting with the levelbuilder example, I updated each place where we call or access one of these functions off of `FieldTextInput` to use `cdoUtils` instead.
I quickly found other instances of these validators being used, most often being supplied as validators when creating a new instance of the field class. Interestingly, some of these instances were happening in Google Blockly blocks. Because the functions were undefined, the validators weren't working. For example, loops in the recently migrated K1 Maze, and the `set score` block in Flappy (migrated in 2021) have both been accepting strings since the labs were flipped:
imageimage

The validators for all of these blocks are now working as expected.

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/blockly/addons/cdoUtils.ts
+++ b/apps/src/blockly/addons/cdoUtils.ts
@@ -213,7 +213,9 @@ export function moveHiddenBlocks(
 
   source.blocks.blocks.forEach(block => {
     const {invisible, procedureId} = block.extraState || {};
-    const hideBlock = procedureTypesToHide.includes(block.type) || invisible;
+    const hideBlock =
+      procedureTypesToHide.includes(block.type) ||
+      (invisible && !Blockly.isStartMode);
     const destination = hideBlock ? hiddenDefinitionSource : mainSource;
     destination.blocks.blocks.push(block);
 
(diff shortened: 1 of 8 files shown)


apps/src/blockly/addons/cdoUtils.ts
```diff
@@ -443,6 +445,37 @@ export function getCode(workspace: WorkspaceSvg, getSourceAsJson: boolean) {
...
+/**
+ * Ensure that only a nonnegative integer may be entered.
+ * @param {string} text The user's text.
+ * @returns {?string} A string representing a valid int, or null if invalid.
+ * @static
+ */
+export function nonnegativeIntegerValidator(text: string): string | null {
+  const n = numberValidator(text);
+  if (n) {
+    return String(Math.max(0, Math.floor(parseFloat(n))));
```
molly-moen, line 476: Am I correct in reading that this turns negative numbers into 0? If so can you update the description? I read the description as if the number is negative we return `null`.

molly-moen (commented):
Looks good, one question on how a function works

molly-moen (approved):

apps/src/blockly/addons/cdoUtils.ts
mikeharv: Add a clarifying example to the description! It will return `null` if the input isn't a valid number, ex. `'abc'`.

PR #60016, opened 2024-07-25
Kaitie/new teachers automatically get v2

kobryan0619:
This PR defaults new teachers to progress view V2.

Note: because this changed what newly created teachers see for the progress view, we had to modify some UI tests as well.

## Links
[
Ticket](https://codedotorg.atlassian.net/jira/software/projects/TEACH/boards/65?selectedIssue=TEACH-1213)

## Testing story

Tested locally and added one test.


dashboard/app/models/user.rb
```diff
@@ -362,6 +364,17 @@ def save_email_reg_partner_preference
...
 
+  # Puts teachers directly into the progress table v2 view when new account is created.
+  def save_show_progress_table_v2
+    if email.present?
+      user = User.find_by_email_or_hashed_email(email)
+      if user&.teacher?
+        user.show_progress_table_v2 = true
+        user.save!
+      end
+    end
```
etaderhold: Would this work instead? I don't love looking the user up all over again from an instance method.
```suggestion
    if teacher?
      show_progress_table_v2 = true
      save!
    end
```

kobryan0619: Good point!  This new code mostly works, I just had to modify it a bit.   We will see if it passes drone!  I did check the ruby tests that were failing before and it seemed to pass those.

etaderhold (approved):

dashboard/app/models/user.rb
```diff
@@ -355,13 +357,21 @@ def save_parent_email_preference
...
       user.share_teacher_email_regional_partner_opt_in = DateTime.now
       user.save!
     end
   end
 
+  # Puts teachers directly into the progress table v2 view when new account is created.
+  def save_show_progress_table_v2
+    if teacher?
+      self.show_progress_table_v2 = true
+    end
```
kobryan0619, line 374: @etaderhold I had to remove the `save!` here to get it to pass drone.  With `save!` multiple tests failed - specifically tests related to calling `update_school_info` - which is odd because i had not touched anything there.  That being said, without `saving`, everything appears to work.  I think this is because we are saving after this method is called - but I cannot find out where.

kobryan0619: Note to my future self: We went with removing the save but then changing this to run `before_create` to ensure the changes did get saved at some point. This shows the order in which[ these callbacks occur](https://api.rubyonrails.org/v7.1.3.4/classes/ActiveRecord/Callbacks.html).

PR #60021, opened 2024-07-25
Enable strongly consistent reads in the test DynamoDBAdapter

carl-codeorg:
One of the tests for setting a DCDO value has been deemed flaky:

```
  test 'storing hashes' do
    to_store = {
      'b' => 'yo dude',
      'c' => [1, 2, 3]
    }
    key = 'random'
    DCDO.set(key, to_store)
    assert_equal DCDO.get(key, nil), to_store
  end
```

I doubt that the test could possibly flake locally or in staging, since the in-memory cache adapter is used, meaning all operations will complete synchronously. The test environment, however, uses DynamoDB, which is eventually consistent by default. This means that read operations that occur in quick succession after a write operation may return stale results.

This PR adds a new param to the `DynamoDBAdapter` to enable strongly consistent reads. Since strongly consistent reads consume double the normal amount of read units, this will increase our DynamoDB usage. I don't think this will be a big deal since we're using provisioned capacity, and we appear to be way over-provisioned on the test DCDO table:

Screenshot 2024-07-25 at 2 03 39 PM

## Links

[JIRA](https://codedotorg.atlassian.net/browse/P20-1064)

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/lib/dynamic_config/adapters/dynamodb_adapter.rb
+++ b/lib/dynamic_config/adapters/dynamodb_adapter.rb
@@ -5,9 +5,10 @@ require 'oj'
 
 class DynamoDBAdapter
   # @param table_name [String] the name of the dynamodb table to use
-  def initialize(table_name)
+  def initialize(table_name, consistent_read: false)
     @table_name = table_name
     @client = Aws::DynamoDB::Client.new
+    @consistent_read = consistent_read
   end
 
   # @param key [String]
@@ -16,7 +17,8 @@ class DynamoDBAdapter
     resp = @client.get_item(
       {
         table_name: @table_name,
-        key: {'data-key' => key}
+        key: {'data-key' => key},
+        consistent_read: @consistent_read
       }
     )
     return nil if resp.item.nil?
--- a/lib/dynamic_config/environment_aware_dynamic_config_helper.rb
+++ b/lib/dynamic_config/environment_aware_dynamic_config_helper.rb
@@ -6,11 +6,14 @@ module EnvironmentAwareDynamicConfigHelper
       # Use the memory adapter if we're running unit tests, but not if we're
       # running the web application server.
       adapter = MemoryAdapter.new
-    elsif rack_or_rails_env == 'production' || managed_test_server?
+    elsif rack_or_rails_env == 'production'
       # Production and the managed test system web application servers
       # (test.code.org / studio.code.org) use DynamoDB.
       cache_expiration = 30
       adapter = DynamoDBAdapter.new(identifier)
+    elsif managed_test_server?
+      cache_expiration = 30
+      adapter = DynamoDBAdapter.new(identifier, consistent_read: true)
     else
       # Everything else writes out to the local filesystem
       file_path = "#{dashboard_dir(identifier)}_temp.json"


stephenliang:
What do you think of embracing eventual consistency rather than swapping over to strong consistency? The reason is that strong consistency still has a chance where it will not produce the correct output. Instead, since [consistency will generally be achived in less than a second](https://docs.aws.amazon.com/whitepapers/latest/comparing-dynamodb-and-hbase-for-nosql/consistency-model.html#:~:text=The%20eventual%20consistency%20option%20is,usually%20reached%20within%20a%20second.), could we do a retry mechanism with a timeout of 5-10 seconds? 

This would also allow test and production to have the same environmental setup which should help reduce branching

carl-codeorg:
> What do you think of embracing eventual consistency rather than swapping over to strong consistency? The reason is that strong consistency still has a chance where it will not produce the correct output. Instead, since [consistency will generally be achived in less than a second](https://docs.aws.amazon.com/whitepapers/latest/comparing-dynamodb-and-hbase-for-nosql/consistency-model.html#:~:text=The%20eventual%20consistency%20option%20is,usually%20reached%20within%20a%20second.), could we do a retry mechanism with a timeout of 5-10 seconds?
> 
> This would also allow test and production to have the same environmental setup which should help reduce branching

Yeah my first thought was just to add a single retry on a slight delay, but it felt like it could artificially increase test times if the retry is needed consistently.

I'd agree that we should embrace eventual consistency overall with Dynamo, but the scope of use is pretty limited in this case - just for testing DCDO flags, where the consistent read might help improve flakiness. In prod, it doesn't matter since we don't expect flags to flip instantaneously. I've never used the strong consistency option in Dynamo, actually - does it really have a chance to produce a stale output if we're synchronously doing a put, then a get, and the put succeeded?

daynew (commented):
Are there other features which use DynamoDB? I'm worried this changes the behavior of them. Then bugs which are caused by eventual consistency will no longer be caught on test.

carl-codeorg:
> Are there other features which use DynamoDB? I'm worried this changes the behavior of them. Then bugs which are caused by eventual consistency will no longer be caught on test.

This doesn't change config for anything except the `EnvironmentAwareDynamicConfigHelper`

stephenliang:
> > What do you think of embracing eventual consistency rather than swapping over to strong consistency? The reason is that strong consistency still has a chance where it will not produce the correct output. Instead, since [consistency will generally be achived in less than a second](https://docs.aws.amazon.com/whitepapers/latest/comparing-dynamodb-and-hbase-for-nosql/consistency-model.html#:~:text=The%20eventual%20consistency%20option%20is,usually%20reached%20within%20a%20second.), could we do a retry mechanism with a timeout of 5-10 seconds?
> > This would also allow test and production to have the same environmental setup which should help reduce branching
> 
> Yeah my first thought was just to add a single retry on a slight delay, but it felt like it could artificially increase test times if the retry is needed consistently.
> 
> I'd agree that we should embrace eventual consistency overall with Dynamo, but the scope of use is pretty limited in this case - just for testing DCDO flags, where the consistent read might help improve flakiness. In prod, it doesn't matter since we don't expect flags to flip instantaneously. I've never used the strong consistency option in Dynamo, actually - does it really have a chance to produce a stale output if we're synchronously doing a put, then a get, and the put succeeded?

The test does a rapid write and read, so probably an operation that occurs in <10 ms which is unlikely to have replicated to other nodes. IIRC, the strong consistency model waits for a majority of replicas to have the written value but if you do a read you'll still have a chance where the test fails if a selected replica doesn't have the updated value (hence strong consistency but not completely consistent). The chance of flakiness is reduced but not eliminated. 

If we do a 5-10 second timeout with a 1 second retry (so 10 retries at max), it should cover all cases short of an AWS brown out or outage.

wilkie:
strong consistency ensures write-then-read order is consistent. anything else is 'eventual' or only read-order consistent. it's, like, the whole point 😅 unless dynamo is just using the terms wrong for some reason.

carl-codeorg:
> The test does a rapid write and read, so probably an operation that occurs in <10 ms which is unlikely to have replicated to other nodes. IIRC, the strong consistency model waits for a majority of replicas to have the written value but if you do a read you'll still have a chance where the test fails if a selected replica doesn't have the updated value (hence strong consistency but not completely consistent). The chance of flakiness is reduced but not eliminated.

Oh interesting. My guess would have been that the master node would always return the latest result, even if it hasn't propagated to all the nodes, in strong consistency mode. 

> If we do a 5-10 second timeout with a 1 second retry (so 10 retries at max), it should cover all cases short of an AWS brown out or outage.

Alright, I'll take a look at how a retry would look for this test since folks have misgivings about messing with the Dynamo call.

wilkie:
> My guess would have been that the master node would always return the latest result, even if it hasn't propagated to all the nodes, in strong consistency mode.

As a person who taught distributed systems at the graduate level... ME TOO. I would love to know why this wouldn't be true... are they claiming propagation delay as not part of their consistency window?

wilkie:
I'm only being the academically the other side on this because it seems reasonable to make the test suite consistent unless we really care about this as an integration test at that level of compatibility. The other option for tests that care about handling eventual consistency (reads not reflected by recent writes) would be to consistently mock unordered responses. Don't mind it either way I suppose though... just my two cents

Hamms:
Great investigation, thank you!

In the interest of minimizing differences between the way our production and test environments are configured, I'm also in favor of embracing eventual consistency and adding a retry mechanism to the test.

carl-codeorg:
Closing without merging because:

- DynamoDB strong consistency is apparently actually eventual consistency in disguise
- The majority opinion is that changing the consistency setting for the DCDO call on test could result in escaped defects due to lack of environmental parity

PR #60026, opened 2024-07-26
[JIMT]defer-codebridge-clickhandler - add the document event listener at the next tick of the event loop

thomasoniii:
React 17 added in #59699 changed how clickhandlers are attached to the dom.

Further explanation is here: https://blog.saeloun.com/2021/07/08/react-17-event-delagation/

In turn, this broke the pop up menus in the file browser in codebridge. Here's what happened:

React 17 changed where react’s event listeners are tacked on (to the app root vs the document), and even that only manifests as an issue if other event handlers are tacked on outside of react’s knowledge.

So the old scenario is that a click event on the document might fire before react’s event handlers, and that click handler might call stopPropagation itself, and that in turn would prevent React’s handlers from firing.

here’s what was happening here - the codebridge button click is showing a pop up menu, and I was attaching a click handler to the document to allow clicks away on the window to close the popup. So my click handler itself was attaching a new click handler to the doc.

My document click handler didn’t call stopPropagation  so it shouldn’t manifest any problems.

Prior to react 17, it’d fire any non-react handlers, then the react handler (including mine!), which would attach a new handler to the document. Since the document’s click handlers had already been invoked, it was ignored on this pass through the loop and all worked well.

After react 17, my handler would wake up and attach to the document. Then, because the invocation order is different, the document click handler that I just attached would wake up and fire, immediately re-closing the pop up.

Stopping propagation would fix it (because it’d prevent the newly attached handler from firing), but a better fix is to just defer the attachment of the new click handler to the next tick of the event loop so it isn’t called as part of this invocation cycle and is just present in the future.

All of this taken together makes me wonder if we’ve got similar patterns elsewhere in the code where new click handlers were attached on the fly and we just need to defer it to the next tick so it doesn’t get muddied up.

#59699 added several `stopPropagation` calls to get around the issue by preventing any clickhandlers attached at the document level from firing at all. Indeed, that would work in this case as well. However, simply deferring the addition of the clickhandler to the next tick also resolved the issue and would not prevent any other external click handlers from firing.

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/codebridge/PopUpButton/PopUpButton.tsx
+++ b/apps/src/codebridge/PopUpButton/PopUpButton.tsx
@@ -39,6 +39,8 @@ export const PopUpButton = ({
       setIsOpen(oldIsOpen => {
         const newIsOpen = !oldIsOpen;
         if (newIsOpen) {
+          // React 17 changed the location where clickhandlers are added, so we want to defer adding the close
+          // handler until the next tick of the event loop, otherwise it'll fire immediately and re-close the pop up.'
           setTimeout(
             () => document.addEventListener('click', setIsOpenFalse),
             0
--- a/apps/src/codebridge/PopUpButton/PopUpButton.tsx
+++ b/apps/src/codebridge/PopUpButton/PopUpButton.tsx
@@ -39,7 +39,10 @@ export const PopUpButton = ({
       setIsOpen(oldIsOpen => {
         const newIsOpen = !oldIsOpen;
         if (newIsOpen) {
-          document.addEventListener('click', setIsOpenFalse);
+          setTimeout(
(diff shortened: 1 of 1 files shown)


apps/src/codebridge/PopUpButton/PopUpButton.tsx
```diff
@@ -39,7 +39,10 @@ export const PopUpButton = ({
       setIsOpen(oldIsOpen => {
         const newIsOpen = !oldIsOpen;
         if (newIsOpen) {
-          document.addEventListener('click', setIsOpenFalse);
+          setTimeout(
```
molly-moen, line 44: mind adding a comment explaining why we need this timeout?

molly-moen (approved):
Nice fix!

PR #60031, opened 2024-07-26
Revert "Revert "Remove inline attribute from blocks in levels""

mikeharv:
Reverts code-dot-org/code-dot-org#60027

Redoes https://github.com/code-dot-org/code-dot-org/pull/59993

The first attempt at this work caused a UI test to become flaky. The Lesson Edit Page test checks that we can add an Artist level to a lesson plan. The steps for that test make the assumption that the search results for that level have completed based on the string `"Artist"` showing up in any cell of the table. The assumption was based on the observation that no Artist levels were in the initial view. However, the initial lesson shows the most recently updated levels by default. This means this test was destined to break as soon as an Artist level was edited.

The solution here is to update this test to look for the string `"Standalone_Artist_1"` to determine that we have finished loading the results. This is less risky because we're unlikely to update this level, which means it should not get promoted to the initial view.

Slack threads:
Initial work proposal and eventual test fix: https://codedotorg.slack.com/archives/C060J3A3BBK/p1722014901565169?thread_ts=1721864106.889939&cid=C060J3A3BBK
Test failure and revert thread: https://codedotorg.slack.com/archives/C0T0PNTM3/p1722010208479609

### DIFF
--- a/dashboard/test/ui/features/teacher_tools/levelbuilder/lesson_edit_page.feature
+++ b/dashboard/test/ui/features/teacher_tools/levelbuilder/lesson_edit_page.feature
@@ -71,7 +71,7 @@ Feature: Using the Lesson Edit Page
     And I press ".fa-search" using jQuery
     # We will know the search has completed after the following step, because we
     # confirmed earlier that there were no Artist levels in the initial view.
-    And I wait until element ".uitest-level-dialog-content td" contains text "Artist"
+    And I wait until element ".uitest-level-dialog-content td" contains text "Standalone_Artist_1"
     And element ".uitest-level-dialog-content td .fa-plus" is visible
     And I press ".uitest-level-dialog-content td .fa-plus:first" using jQuery
     And I click selector ".save-add-levels-button"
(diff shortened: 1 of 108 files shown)


dashboard/test/ui/features/teacher_tools/levelbuilder/lesson_edit_page.feature
```diff
@@ -71,7 +71,7 @@ Feature: Using the Lesson Edit Page
     And I press ".fa-search" using jQuery
     # We will know the search has completed after the following step, because we
     # confirmed earlier that there were no Artist levels in the initial view.
-    And I wait until element ".uitest-level-dialog-content td" contains text "Artist"
+    And I wait until element ".uitest-level-dialog-content td" contains text "Standalone_Artist_1"
```
molly-moen, line 74: Can you update the comments in this scenario? We have a comment above and here referring to "Artist", we should update to explain why we think search succeeded if we see "Standalone_Artist_1".

davidsbailey, line 74: nice fix Mike!

davidsbailey (approved):

molly-moen (approved):
One request for updating comments on that test, otherwise looks good!

PR #59538, opened 2024-07-02
Skip enroll form for Build Your Own workshops

TurnerRiley:
This PR skips the enrollment form for Build Your Own workshops and instead auto-enrolls the user upon visiting the enrollment link and sends them to the My PL page with a success message!

### Successfully enrolls user in Build Your Own workshop
Once the user visits the enrollment link, it:
- Skips the enroll form
- Enrolls the user in the workshop
- Sends them to the My PL page
- Shows a success message
- Logs the successful enrollment event

User flow demo:

https://github.com/code-dot-org/code-dot-org/assets/56283563/6af0695a-5d54-4a19-af56-17b00ae21de5

Enrollment shows up in DB:
entry in db

Note: I inserted a basic success message as a placeholder with in-file styles since Hannah is working on the actual Successful Enrollment Dialog. I [created a ticket](https://codedotorg.atlassian.net/jira/software/c/projects/ACQ/boards/64?assignee=60d62161f65054006980bd71&selectedIssue=ACQ-2051) to delete the placeholder message and styles and swap it for using the new dialog once it's merged.

### Still shows the form for other types of workshops
https://github.com/code-dot-org/code-dot-org/assets/56283563/107366ac-ca4c-4066-83c7-8704781d4a13

### Back end vs. front end school_info association
As part of this, the enrollment's required backend association with school_info had to be deleted since we didn't need school info for Build Your Own enrollments. However, obtaining school info in the enroll form is still enforced on the front-end for workshops that still use school info:

school is still required on front end for form

## Links
Jira ticket: [here](https://codedotorg.atlassian.net/jira/software/c/projects/ACQ/boards/64?assignee=60d62161f65054006980bd71&selectedIssue=ACQ-2026)
Spec: [item 7A](https://docs.google.com/document/d/16IsvtmDFDIDzjn3vv1bIskRkET9ivC_l9LyUXpN0Fno/edit#heading=h.4schd43y1qoc)

## Testing story
Local testing and updating unit tests.

## Follow-up work
Use new Enrollment Success Dialog once merged.

PR #60033, opened 2024-07-26
Make last_name not required for enrollment creation

TurnerRiley:
When setting up the auto-enrollment process for the Build Your Own workshop, we had to [get rid of the back-end presence validation](https://github.com/code-dot-org/code-dot-org/pull/59538/files#diff-b628772b45f6dd4a03f6f4842ced36ec4e6098ebf40de42217067d06fed1c1d4L68) for `school_info` for each enrollment created (as the user would submit that info in the enroll form we were skipping for Build Your Own workshops). Recently, I found out that the `last_name` field has the same issue, so this PR removes that requirement. Filling in the Last Name field in the enroll form is still required on the front-end, this just allows us to skip the enroll form and not raise any errors on the back-end when creating the enrollment for Build Your Own workshops.

### Can auto-enroll in a Build Your Own workshop with no last name set

https://github.com/user-attachments/assets/c1527949-11bd-4758-8e54-12d93029bf80

### Can display an enrollment that has no last name with no errors
can have no last name

## Links
Jira ticket: [here](https://codedotorg.atlassian.net/jira/software/c/projects/ACQ/boards/64?assignee=60d62161f65054006980bd71&selectedIssue=ACQ-2130)

## Testing story
Local testing and updating unit tests.

### DIFF
--- a/dashboard/app/models/pd/enrollment.rb
+++ b/dashboard/app/models/pd/enrollment.rb
@@ -55,10 +55,6 @@ class Pd::Enrollment < ApplicationRecord
 
   validates_presence_of :first_name, unless: :deleted?
 
-  # Some old enrollments, from before the first/last name split, don't have last names.
-  # Require on all new enrollments.
-  validates_presence_of :last_name, unless: -> {deleted? || created_before_name_split?}
-
   validates_presence_of :email, unless: :deleted?
   validates_confirmation_of :email, unless: :deleted?
   validates_email_format_of :email, allow_blank: true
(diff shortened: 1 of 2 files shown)


hannahbergam (approved):
Makes sense to me! I assume this change is going to be permanent for BYO?

TurnerRiley:
> Makes sense to me! I assume this change is going to be permanent for BYO?

Yep! Since BYO will skip the enroll form, if a user doesn't have a last name then they just won't have a last name in the workshop info display. Other workshops will still require a last name on the front-end form!

PR #60037, opened 2024-07-26
Fix Violations of and Reenable `Style/Lambda`

Hamms:
Fix applied automatically with `bundle exec rubocop --autocorrect-all --only Style/Lambda`. I then also autocorrected some resulting `Layout/` violations (mostly related to whitespace) and also made some manual fixes; replaced some instances of `{}` block syntax with `do/end` and added explicit parentheses to some method calls that now need them.

This is an interesting one:

> (by default) checks for uses of the lambda literal syntax for single line lambdas, and the method call syntax for multiline lambdas. It is configurable to enforce one of the styles for both single line and multiline lambdas as well.

By default, it encourages you to use the "literal" syntax for lambdas whose body consists of only a single line of code, and to use the more-verbose "lambda" syntax for more complex lambdas. Ie:

```ruby
# bad
f = lambda { |x| x }
f = ->(x) do
  x
end

# good
f = ->(x) { x }
f = lambda do |x|
  x
end
```

It is a little odd to be explicitly enforcing a slight inconsistency; this approach does represent a much smaller delta relative to our current codebase than does standardizing around the "lambda" syntax, but is a bit larger than standardizing around the "literal" syntax would be. See the linked PRs below for examples of both alternatives.

Because this is a `Style/` cop, I think we do have a lot of latitude to decide what style we personally prefer, though, and if anyone would like to make an argument for an EnforcedStyle other than the default, I'm open to hearing it. What do folks think? Anyone have any preferences?

## Links

Documentation:
- https://docs.rubocop.org/rubocop/cops_style.html#stylelambda
- https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/Lambda

Alternative `EnforcedStyle` options:
- https://github.com/code-dot-org/code-dot-org/pull/59958
- https://github.com/code-dot-org/code-dot-org/pull/59947

## Testing story

This cop supports safe autocorrection, so this should be guaranteed to not result in any change to functionality. Relying on existing tests to verify that none of my manual changes did, either.

## PR Checklist:

- [ ] Code is well-commented


cat5inthecradle (approved):
I'd say either the default rule or no rule is the best option. I think they're all ugly, so I'll defer to the ruby devs :P

I'm curious, is there a rule that complains about "unnecessary multi-line lambdas" that just have a single return? Those can be annoying while developing when you're still determining how many lines your lambda will have.

Hamms:
> I think they're all ugly, so I'll defer to the ruby devs :P

Lol, yeah, I'd say that's basically my opinion here, too; all the options are kinda gross, but adding an override to either disable or configure the rule is ever so slightly more gross than going with the default.

> is there a rule that complains about "unnecessary multi-line lambdas" that just have a single return

Do you mean something like:

```ruby
# good
f = lambda do |x|
  {
    foo: x
  }
end
```

If so, this rule would allow that and would warn on something like,

```ruby
# bad
f = ->(x) do
  {
    foo: x
  }
end
```

Somewhat unexpectedly, it also allows this in-between implementation:

```ruby
# good
f = lambda {|x|
  {
    foo: x
  }
}
```

davidsbailey (approved):
makes sense to me!

PR #60040, opened 2024-07-26
Create Account Button A/B Test - Fix Statsig config load issue

kelbyhawn:
Moves the Statsig config code into the event listener on `userHeaderStatsigReporter.js` so the user is put into a bucket right away. Previously this was outside of the event listener and not firing on the first load of a page. The Statsig code was loading after our code causing the button not to show up unless they visited another page (if in the test bucket).

This should also prevent _Uninitialized_ and _Not started_ rules from appearing 🙌 

**Things started working as expected once I moved the code into the event listener:**
Screenshot 2024-07-26 at 2 15 07 PM

## Links
Jira ticket: [ACQ-2134](https://codedotorg.atlassian.net/browse/ACQ-2134)

## Testing story
Tested locally on incognito browser.

### DIFF
--- a/apps/src/userHeaderStatsigReporter/userHeaderStatsigReporter.js
+++ b/apps/src/userHeaderStatsigReporter/userHeaderStatsigReporter.js
@@ -1,15 +1,5 @@
 import statsigReporter from '../lib/util/StatsigReporter';
 
-const isInCreateAccountButtonExperiment = statsigReporter.getIsInExperiment(
-  'create_account_button',
-  'showCreateAccountButton',
-  false
-);
-
-export function showCreateAccountButton() {
-  return isInCreateAccountButtonExperiment;
-}
-
 document.addEventListener('DOMContentLoaded', function () {
   const createAccountButton = document.querySelector('#create_account_button');
   const createAccountButtonDesktop = document.querySelector(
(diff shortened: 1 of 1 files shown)


apps/src/userHeaderStatsigReporter/userHeaderStatsigReporter.js
```diff
@@ -19,6 +9,16 @@ document.addEventListener('DOMContentLoaded', function () {
   const hamburgerButtons = document.querySelector('#hamburger-sign-up-buttons');
   const signUpPage = window.location.pathname.includes('/users/sign_up');
 
+  const isInCreateAccountButtonExperiment = statsigReporter.getIsInExperiment(
+    'create_account_button_2',
```
kelbyhawn, line 13: Updated this to work w/ V2 of the experiment.

apps/src/userHeaderStatsigReporter/userHeaderStatsigReporter.js
```diff
@@ -19,6 +9,16 @@ document.addEventListener('DOMContentLoaded', function () {
...
+    'showCreateAccountButton',
+    false
+  );
+
+  function showCreateAccountButton() {
```
bethanyaconnor, line 18: is this function still necessary? (though if you don't need to make another update, it's probably not worth updating this)

bethanyaconnor (approved):

apps/src/userHeaderStatsigReporter/userHeaderStatsigReporter.js
kelbyhawn: Probably not, but to your point, drone is almost done so I'll leave it 😅

PR #60044, opened 2024-07-26
log thumbs up / down feedback for ai evals to cloudwatch

davidsbailey:
Starts https://codedotorg.atlassian.net/browse/AITT-410:
* logs to cloudwatch when thumbs up / down are pressed

### screenshots (updated)

verified by running a snippet of the code directly on production-console that I can get events to show up in cloudwatch:
Screenshot 2024-07-30 at 3 21 03 PM
https://us-east-1.console.aws.amazon.com/cloudwatch/home?region=us-east-1#metricsV2?graph=~(metrics~(~(~(expression~'m1*2fSUM*28METRICS*28*29*29~label~'Expression1~id~'e1))~(~'AiRubric~'LearningGoalAiEvaluationFeedback~'Environment~'production~'Unit~'interactive-games-animations-2023~'Approval~'thumbs_down~'Lesson~'11~(id~'m1))~(~'...~'thumbs_up~'.~'.~(id~'m2))~(~'.~'.~'.~'development~'Approval~'thumbs_down~(id~'m3))~(~'...~'thumbs_up~(id~'m4)))~view~'timeSeries~stacked~false~region~'us-east-1~stat~'Average~period~300~start~'-PT1H~end~'P0D)&query=~'*7bAiRubric*2cApproval*2cEnvironment*7d*20approval

## Testing story

- manual verification shown in screenshots
- new unit tests

## Follow-up work

* configure CloudWatch to alarm on this metric
* make sure we log an event as soon as thumbs down is pressed, even if the user does not go on to submit the resulting form

### DIFF
--- a/dashboard/app/jobs/concerns/ai_rubric_metrics.rb
+++ b/dashboard/app/jobs/concerns/ai_rubric_metrics.rb
@@ -53,11 +53,12 @@ class AiRubricMetrics
       [
         {
           metric_name: 'LearningGoalAiEvaluationFeedback',
-          value: feedback.ai_feedback_approval ? 1 : 0,
+          value: 1,
           dimensions: [
             {name: 'Environment', value: CDO.rack_env},
             {name: 'Unit', value: lesson&.script&.name || ''},
             {name: 'Lesson', value: lesson&.relative_position&.to_s || ''},
+            {name: 'Approval', value: feedback.ai_feedback_approval ? 'thumbs_up' : 'thumbs_down'}
           ],
           unit: 'Count',
         }
--- a/dashboard/app/jobs/concerns/ai_rubric_metrics.rb
+++ b/dashboard/app/jobs/concerns/ai_rubric_metrics.rb
@@ -46,8 +46,6 @@ class AiRubricMetrics
   # Log thumbs up / down feedback for AI evaluation
   # @param [LearningGoalAiEvaluationFeedback] feedback
   def self.log_feedback(feedback)
-    lesson = feedback.learning_goal_ai_evaluation&.learning_goal&.rubric&.lesson
-
     Cdo::Metrics.push(
       AI_RUBRIC_METRICS_NAMESPACE,
       [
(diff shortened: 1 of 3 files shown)


dashboard/app/jobs/concerns/ai_rubric_metrics.rb
```diff
@@ -43,6 +43,29 @@ def self.log_metric(metric_name:, agent: nil, value: 1)
...
+          value: 1,
+          dimensions: [
+            {name: 'Environment', value: CDO.rack_env},
+            {name: 'Unit', value: lesson&.script&.name || ''},
+            {name: 'Lesson', value: lesson&.relative_position&.to_s || ''},
+            {name: 'Approval', value: feedback.ai_feedback_approval ? 'thumbs_up' : 'thumbs_down'}
+          ],
+          unit: 'Count',
+        }
+      ]
```
cat5inthecradle: We should think about demensions carefully here. Excess or poorly designed dimensions can create noise and increase cost. I initially wrote a novel here, but [this section in the docs](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/cloudwatch_concepts.html#Dimension) explains it clearly, so make sure you're happy with these dimensions. I don't have a perfect mental model, so I think my only concern is whether changes to Units could throw off the lesson numbers and impact historic metrics - like if a lesson gets added or removed from the middle of a Unit.

For other reader reference, here's how this looks as data starts to get populated:
image

cat5inthecradle (approved):

dashboard/app/jobs/concerns/ai_rubric_metrics.rb
davidsbailey: per offline discussion, we do not want to log unit and lesson because we just plan to alert on the (already low) total volume of thumbs up and thumbs down events. if we want to do deeper analysis, we can look at the data in redshift and even set up a tableau dashboard if we want to. additionally, including these dimensions would come 
with some cost, because aggregating across all possible values for a dimension in cloudwatch is difficult.

PR #60003, opened 2024-07-25
Setup pycall

davidsbailey:
## PR Checklist:

- [ ] Tests provide adequate coverage
- [ ] Privacy and Security impacts have been assessed
- [ ] Code is well-commented
- [ ] New features are translatable or updates will not break translations
- [ ] Relevant documentation has been added or updated
- [ ] User impact is well-understood and desirable
- [ ] Pull Request is labeled appropriately
- [ ] Follow-up work items (including potential tech debt) are tracked and linked

PR #60048, opened 2024-07-29
pycall.rb + pdm: call python from ruby

snickell:
This PR is  a continuation of @davidsbailey 's work in https://github.com/code-dot-org/code-dot-org/pull/60003 and allows Python code to be invoked easily from within Ruby. Ruby and Python have very similar object systems, and [pycall.rb](https://github.com/mrkn/pycall.rb) allows bridging the two. Invocations are very fast and can be generally treated as "function calls" in terms of performance.

[pycall.rb](https://github.com/mrkn/pycall.rb) permits a possible architecture where ruby and python code run in the same monolith. This should be considered an experiment for now, but **if it proves reliable** in practice, I would advocate this "side by side in the monolith" approach **become the primary way we use python**.

This PR uses [pdm](https://pdm-project.org/) for python dependency management, virtualenv management, and python interpreter installation. pdm is similar to bundle + rbenv or yarn + nvm.

For more information on how we might structure python code in our repo, see: **[python/README.md](https://github.com/code-dot-org/code-dot-org/blob/seth/setup-pycall/python/README.md#how-to-structure-your-python-code)**

### Example

Ruby code using the python `math` package:
```
pyimport 'math'
puts "Python says the square root of 25 is: #{math.sqrt(25)}"
```

Ruby runs python code from [/python/pycdo/pycdo/test_module/test_func.py](https://github.com/code-dot-org/code-dot-org/blob/seth/setup-pycall/python/pycdo/pycdo/test_module/test_func.py):
```
pyfrom 'pycdo.test_module`, import: :test_func
puts test_func()
# => "Ruby can call Python!"
```

See [pycall_test.rb](https://github.com/code-dot-org/code-dot-org/blob/seth/setup-pycall/dashboard/test/lib/pycall_test.rb) and [python/README.md](https://github.com/code-dot-org/code-dot-org/blob/seth/setup-pycall/python/README.md#examples-of-calling-python-from-ruby) for more examples.

### Architectural advantages

As we've started to do AI work, we've discovered that python has more and better libraries of interest (compared to ruby). This has led to python being the obvious language for writing much of our AI code in, and we only expect the gap to grow over time. 

As of 2024, we have decided not to use a microservice architecture, but in two cases pragmatism has necessitated auxiliary microservices. Both were needed for the same reason: support domain-specific must-use programming languages that would be too expensive to do in ruby. The [javabuilder](https://github.com/code-dot-org/javabuilder) service was required in order to host `javac`, and AI services like [aiproxy](https://github.com/code-dot-org/aiproxy) have been needed to utilize python's ai-adjacent packages.

Javabuilder is a small body of code, highly contained and not anticipated to grow, so it represents an exception rather than the rule. However, since we expect AI work to continue growing, and anticipate that we'll continue to see a gap between ruby and python in supporting AI libraries, I am concerned with a growing fork between "we write ruby here" and "we write python there".

With pycall.rb, we can keep our [Majestic Monolith](https://signalvnoise.com/svn3/the-majestic-monolith/) while easily taking advantage of amazing python libraries.

### Everyday advantages

Directly calling python from ruby can make programming easier (vs multiple services):
1. Its easy to pass ruby callbacks into python
    - This also lowers the need for python to store state / access a DB. That in turn increases re-use of Rails model code. See: [comment with a callback example](https://github.com/code-dot-org/code-dot-org/pull/60048#issuecomment-2261895323)
2. A single ruby+py PR lets you review and atomically deploy changes that require modification to both sides of the code.
3. The high-performance reduces the need for "fat python code": its cheap to do lots of invocations rather than shifting code into python. Wanna call python code in a for loop? No need to shift the loop to python, just call it from a ruby for loop.
    - This could **reduce the amount of python we write** to accomplish the same goal.
4. Its easy to casually take advantage of Python libraries, outside core AI, in cases where its only "helpful" not needed. Would `scikit-learn` make a random graph easier to generate? Use it!
    - This could **increase the number of projects python helps us accomplish**.

#### CPU

On my system, you can do **4 million function calls/second** from ruby to python:
```
Benchmark.measure {
  1_000_000.times { math.sqrt(5) }
}.total
# => 0.2446950000000007
```

Conclusion: function calls from ruby => python are only 10x slower than ruby => ruby function calls, they can be treated essentially the same way.

#### Resident RAM

1. load `irb` _without_ our environment
7. `ps -o pid,rss` => irb process using `42.6 MB` RSS
8. `require ./lib/cdo/pycall.rb` to load pycall
9. `ps -o pid,rss` => irb process using `49.7 MB` RSS
10. `PyCall.import_module('math')` to force use of python
11. `ps -o pid,rss` => irb process using `56.2 MB` RSS

56.2 - 42.6 => **pycall + python take 13 MB of RAM** (compare to python3 running by itself which takes 12.5MB of RAM).

Conclusion: we can expect each Ruby process to take a baseline extra 13 MB of RAM + ram taken by whatever python packages we want to use (e.g. with openai, numpy and boto3: 62MB)

#### Memory leaks

I tested on Linux with `MALLOC_CHECK_=1 irb` as well as valgrind's memcheck, played with importing a couple dozen python modules and doing lots of function calls. Did not get report of any memory leaks or other memory issues.

### Thread Safety

We use threads in our puma configuration, and PyCall is not thread-safe. The good news is that its SO not thread safe that its very obvious if you make a mistake: it crashes instantly if ever used from a thread. I believe these issues should be addressed in a follow-up PR.

1. This is not an issue for our `delayed_job` ActiveJob workers in prod, which use processes instead of threads.
  - This does not work with the :async workers we currently have [set as the default in :development](https://github.com/code-dot-org/code-dot-org/blob/a76e2afb255bbd6ebd327420ffb765c3ad5d4315/dashboard/config/environments/development.rb/#L80-L84).    On the one hand, the async workers are nice because they don't require launching `bin/delayed_job start` to use worker features. On the other hand, it'd be nice if our stock config worked with Python features without requiring a locals.yml tweak. Ideally, we could start delayed_jobs along with dashboard server. This is out of scope for this PR and would be an interesting subject for a future PR.
2. This means you cannot use PyCall from Rails controllers as-is.
  - I have some work toward remedying this issue here: https://github.com/code-dot-org/code-dot-org/pull/60267 and https://github.com/snickell/pycall_thread, but it still has a critical bug that I need to find before we can use it.

## pdm

pdm solves both the python interpreter version problem, the virtualenv problem, and the dependency versioning problem. 

PDM:
- Functionality is similar to bundle:
  - `bundle install` => `pdm install`
    - `pdm install` will install a python interpreter if needed to fulfill the `requires-python` clause in pyproject.toml
    - `pdm install` will install packages into the venv, including deps of sub-packages like pycdo
  - `bundle exec ____` => `pdm run ____`
  - `pdm add` or `pdm add --dev` can be used to add deps and dev-deps to the pyproject.toml
- permits local sub-packages to be defined, which allows dependencies to be specified in `/python/subpackage/pyproject.toml` files rather than all mixed together in a top-level `/pyproject.toml`.
- follows [PEP 621](https://peps.python.org/pep-0621/) for its dependency metadata, which means future package managers should be compatible with it as well.
  - uses a lockfile `pdm.lock`, which allows the `pyproject.toml` dependencies to focus on what's actually compatible and not try to double as an imperfect lockfile (ala requirements.txt)

## Changes in PR

- Adds the [pycall gem](https://github.com/mrkn/pycall.rb)
- Includes a basic test that we can invoke python code from within ruby, including accessing python libraries
- Uses pdm for managing the python virtualenv as well pip dependencies.
- [lib/cdo/pycall.rb](https://github.com/code-dot-org/code-dot-org/blob/seth/setup-pycall/lib/cdo/pycall.rb) is called at rails init and uses pdm to direct pycall toward the correct python version/virtualenv
- Adds a [/pyproject.toml](https://github.com/code-dot-org/code-dot-org/blob/seth/setup-pycall/pyproject.toml) for defining python our major python packages
  -/pyproject.toml also specifies a Python version: 3.12 so we start off as fresh as possible
- Extends our Dockerfile (including drone) to install pdm
  - Adds a new Chef cookbook: `cdo-python`
    - installs pdm to /usr/local/bin
- Extends `rake install` and `rake build` to run `pdm install` (equivalent to `bundle install`)
  - this will download (and build if necessary) a version of python matching `/pyproject.toml`
  - this will create a virtual environment and install deps in it
- Updates SETUP.md to include instructions for installing pdm.
- Creates a package structure in `/python` showing how to create sub-packages, and add dependencies to them.
  - Example package: [/python/pycdo](https://github.com/code-dot-org/code-dot-org/tree/seth/setup-pycall/python/pycdo)
- Extends `lint.rake` (lint everything) and `lint.rb` (lint git commit hooks) to run linting on Python code in /python using [Ruff](https://github.com/astral-sh/ruff)

### Things for a followup PR:

- Fill in the sketch of how to use python started by `/python`:
  - System for running tests for all sub-packages of `/python` and laddering up the results
  - More extension /python/README.md, in particular advice on how to break work into packages
- Example of using python from a ruby controller/model

## Post Merge: post on #developers with install instructions

Currently
- Mac users: `brew install pdm`
- Linux users:
  - `sudo pip3 install --prefix=/usr/local --upgrade pdm`
- Then both mac & linux users, from the project root: 
  - `pdm install --dev`: should create a venv in .venv
  - `pdm run python3 --version` and confirm its 3.12.x

## TODO

- [x] verify shelling out to `aws` cli command works
- [ ] get a linux user to test the linux instructions
- [ ] get at least one more mac user to test mac instructions
- [x] write chef dep install for pdm
- [x] test chef install on adhoc
- [x] write `/python/README.md`
- [x] Right before merge: tag docker build with a version number, push to hub, and update .drone.yml with the version number (currently using tag `codedotorg/code-dot-org:pycall-pdm`), probably will use `codedotorg/code-dot-org:1.10.0` depending on when we merge
- [ ] don't forget: send post merge instructions to #developers

Co-authored-by: dave@code.org


davidsbailey:
Thank you for the detailed write up and supporting evidence! One thought is that it would be nice for some of this to live in a README.

> Essentially the pyenv / Pipfile in the root of our repo defacto specifies the python interpreter and packages available in all of our invocations and usage - the same as rbenv specifies the ruby interpreter for our whole repo.

I agree with this statement, however a couple of things to clarify / emphasize to help identify potential problems:
* to be slightly more specific, this affects any python invocation _within our main repo after loading our rails environment_
* this includes many/most rake tasks, cronjobs and oneoff scripts
* this could affect the version of python used if a rake task were to call into aws-cli. I had trouble finding examples of this since we generally use the aws-sdk gem, but mentioning in case infra team can think of a time we use the cli (credit to Suresh for mentioning this vector some time ago)

> confirm its 3.12.x

I 'm seeing the test machine is running python 3.8.10. apologies if i missed this, but is there something we need to do to make sure this doesn't break things on chef-managed environments?

snickell:
> Thank you for the detailed write up and supporting evidence! One thought is that it would be nice for some of this to live in a README.

I was just thinking about how to do this, would love your thoughts. If this was a "python repo" I feel like I'd put the info in the `/README.md` itself. But we don't even do that for ruby stuff, in some ways the documentation for how to use bundle etc is kinda in the `SETUP.md`. I was thinking we should have a `README.md` targeted toward python, it sounds like you're thinking similar? 

I was thinking either `docs/PYTHON.md` or `/python/README.md`. Either way, I would link to them from SETUP.md along with a snippet on using `pipenv install`. 

I would imagine once we go to next steps, the Python README would also include information on our conventions for structuring modules and using Python from ruby. I was going to include a brief section on that in this PR, but mostly it would say "TBD in a future PR".

That said, I would default to imagining we would use python in two ways:
1. Well-defined larger python feature modules, possibly with their own Pipfiles, probably with their own test systems (learning from our /apps mistakes), possibly located in `/python`? 🤷 
2. Little python .py snippets included directly beside .rb code, and included directly from .rb. Think "glue code".

> 
> > Essentially the pyenv / Pipfile in the root of our repo defacto specifies the python interpreter and packages available in all of our invocations and usage - the same as rbenv specifies the ruby interpreter for our whole repo.

> I agree with this statement, however a couple of things to clarify / emphasize to help identify potential problems:
> 
> * to be slightly more specific, this affects any python invocation _within our main repo after loading our rails environment_
> * this includes many/most rake tasks, cronjobs and oneoff scripts
> * this could affect the version of python used if a rake task were to call into aws-cli. I had trouble finding examples of this since we generally use the aws-sdk gem, but mentioning in case infra team can think of a time we use the cli

Good points, I'll add these.

> > confirm its 3.12.x
> 
> I 'm seeing the test machine is running python 3.8.10. apologies if i missed this, but is there something we need to do to make sure this doesn't break things on chef-managed environments?

You're 100% right, I still need to do the chef work and probably test it in an adhoc before this could be merged.

-Seth

davidsbailey:
Per offline discussion, I like the idea of having a top-level `python/` directory, which could contain the README, pipfiles, and (not in this PR) any python files and corresponding tests. python tests could be run entirely in python and invoked directly by a rake task, without going through pycall.

snickell:
Per offline discussion with @davidsbailey, we wanted to check if pycall.rb supports python=>ruby callbacks... and it does! This could make a number of patterns easier (vs doing callbacks in a multiple-service model), and reduces the need for python to access the DB.

```
require './lib/cdo/pycall.rb'

ruby_callback = lambda do |arg|
  "ruby_callback('#{arg}') was invoked"
end

PyCall.exec <<~PYTHON
  def py_run_callback(cb):
      return "Python says: " + cb("hi from py")
PYTHON

py_run_callback = PyCall.eval('py_run_callback')
response = py_run_callback.call(ruby_callback)

puts "Ruby says: #{response}"
# => Ruby says: Python says: ruby_callback('hi from py') was invoked
```

snickell:
Huh:
image

We don't have to get this exactly perfectly right in this PR, buuuuuuut.....

davidsbailey:
> Huh: image
> 
> We don't have to get this exactly perfectly right in this PR, buuuuuuut.....

I'm not super familiar with the other libraries. What is your takeaway from the graph?

davidsbailey:
A quick read of the pdm docs suggests it addresses a few shortcomings of pipenv and poetry which I'm not sure apply to us. This seems like an easy thing to change later, so I'm kind of leaning toward sticking with pipenv just given the steps you've already done to get it working plus no clear current need for one of the other libraries (please correct me if I'm mistaken there).

snickell:
I'm inclined to stay the course, but it was enough to prompt me to do a brief experiment with the other two. I converted our Pipfile to the other two's formats (they both share a pyproject.toml, but poetry is totally custom fields whereas pdm follows pep 621 fields) and played around with them, in particular focusing on env creation which is good in pipenv but not perfect.

Conclusion: stick with pipenv unless I get sick of trying to make pyenv work with chef (this has been hard actually, the way pyenv wants to be installed is really not a good fit for chef) and at some point it looks easier to switch to pdm just to get passed the chef bit.

Poetry:
- still doesn't follow [PEP 621](https://peps.python.org/pep-0621/), so its not anymore "future proof" standards-wise than pipenv as far as how deps are specified
- does /not/ make setting up and using a consistent python version easy, MUCH harder than pipenv or pdm, which means more confusion, more complex CI, and more steps non-python users have to go through to make python work.
- Conclusion: I don't think poetry is a good fit for us even if its popular, it'd be hard to integrate transparently into a `rake install` task as for as self-setup

PDM:
- does follow [PEP 621](https://peps.python.org/pep-0621/) for deps, which means even if pdm wasn't the package manager of the future, the package manager of the future could use our pre-existing dep specs. same way yarn/npm are reasonably interchangable. 🤷, kinda cool.
- has an absolutely bulletproof python installation + virtualenv system that doesn't require pyenv to be installed, reducing complexity and is really well designed for docker/chef style integration. whereas pipenv+pyenv tends to require building python, pdm somehow has a huge archive of prebuilt python binaries, which makes it really fast in a docker/chef context. it also handles python version upgrades better venv-wise (I believe with pipenv we'll have to tell devs to destroy their venv manually when we change python major versions)
- support for "nested packages": Pipenv doesn't provide a story for having our python modules be separate packages, with separately specified deps, separate pyproject.tomls, etc. Pipenv specifically does /not/ try to address the case of making packages, and it doesn't have well designed support for even including local packages.
- Conclusion: really cool, worth keeping an eye on, still not as popular as pipenv but growing, I might have used it if I was starting from scratch, and an off-chance I'll switch to pdm if pyenv continues being so difficult to integrate with chef.


> seems like an easy thing to change later

Its an easy change now, you know how these things go as far actually ending up being done later 🤣, but yeah, theoretically not hard later either. I'm imagining this baked in 10 years from now, and seemed worth spending another few hours in 2024 to lower the unchangeable pain in 2034.

snickell:
@davidsbailey: besides "cut and run" pragmatism on not spending too much time trying to make pyenv (for pipenv) work in chef, the one serious feature that pdm might give us is the ability to (if and when we want) declare python modules as truly separate packages.

This seems relevant in light of our upcoming (probable?) decision to split apps/ into separate packages: which in historical hindsight they probably always should have been. Are we potentially re-making the same architectural mistake with Python in 2024 that we made with JS in 201x?

PDM would give us a reasonable story for sub-packages, that have their own separately specified deps and test systems, pipenv does not.

Now, Ruby doesn't give us that either, so I'm not saying its critical, but its something that came to light as I experimented with actually putting modules in `/python` and seeing how it felt to use and define them.

davidsbailey:
makes sense to me, I'm open to trying pdm! thanks for doing the research to rule out poetry.

snickell:
I gave up on trying to make Chef work, and I'm experimenting with a full conversion over to PDM here: https://github.com/code-dot-org/code-dot-org/pull/60196

I really like the ability to put the deps in sub-packages rather than putting everything in a huge `/Pipfile`, you can see that working in `/python/pycdo/pyproject.toml` which is a non-toplevel python project that pulls in its own deps, which are then laddered up to `/pyproject.toml`

snickell:
And it looks like we /are/ causing issues for the `aws` command, this is not hypothetical:
```
ImportError: cannot import name 'DEFAULT_CIPHERS' from 'urllib3.util.ssl_' (/Users/seth/src/code-dot-org/.venv/lib/python3.12/site-packages/urllib3/util/ssl_.py)
Traceback (most recent call last):
  File "/opt/homebrew/bin/aws", line 19, in <module>
    import awscli.clidriver
  File "/opt/homebrew/Cellar/awscli/2.17.21/libexec/lib/python3.11/site-packages/awscli/clidriver.py", line 22, in <module>
    import botocore.session
  File "/opt/homebrew/Cellar/awscli/2.17.21/libexec/lib/python3.11/site-packages/awscli/botocore/session.py", line 27, in <module>
    import botocore.client
  File "/opt/homebrew/Cellar/awscli/2.17.21/libexec/lib/python3.11/site-packages/awscli/botocore/client.py", line 16, in <module>
    from botocore import UNSIGNED, waiter, xform_name
  File "/opt/homebrew/Cellar/awscli/2.17.21/libexec/lib/python3.11/site-packages/awscli/botocore/waiter.py", line 17, in <module>
    from botocore.docs.docstring import WaiterDocstring
  File "/opt/homebrew/Cellar/awscli/2.17.21/libexec/lib/python3.11/site-packages/awscli/botocore/docs/__init__.py", line 15, in <module>
    from botocore.docs.service import ServiceDocumenter
  File "/opt/homebrew/Cellar/awscli/2.17.21/libexec/lib/python3.11/site-packages/awscli/botocore/docs/service.py", line 14, in <module>
    from botocore.docs.client import ClientDocumenter, ClientExceptionsDocumenter
  File "/opt/homebrew/Cellar/awscli/2.17.21/libexec/lib/python3.11/site-packages/awscli/botocore/docs/client.py", line 16, in <module>
    from botocore.docs.example import ResponseExampleDocumenter
  File "/opt/homebrew/Cellar/awscli/2.17.21/libexec/lib/python3.11/site-packages/awscli/botocore/docs/example.py", line 13, in <module>
    from botocore.docs.shape import ShapeDocumenter
  File "/opt/homebrew/Cellar/awscli/2.17.21/libexec/lib/python3.11/site-packages/awscli/botocore/docs/shape.py", line 19, in <module>
    from botocore.utils import is_json_value_header
  File "/opt/homebrew/Cellar/awscli/2.17.21/libexec/lib/python3.11/site-packages/awscli/botocore/utils.py", line 36, in <module>
    import botocore.httpsession
  File "/opt/homebrew/Cellar/awscli/2.17.21/libexec/lib/python3.11/site-packages/awscli/botocore/httpsession.py", line 21, in <module>
    from urllib3.util.ssl_ import (
```

Setting the PYTHON bin would not be a problem, because that's hardcoded into the aws cli command. The problem is that PyCall is requiring us to manually set `PYTHON_PATH`, because its following symlinks and not finding the top-level site-packages directory automatically. Setting PYTHON_PATH /is/ inherited by the aws cli's python interpreter, and the 3.12 packages on that path are causing issues with its built-in venv.

snickell:
Doing something like this at the end of pycall.rb seems to fix the issues with launching `aws` cli:
```
# Do an import to ensure the Python interpreter is initialized.
pyimport 'pycdo'

# Now restore the PYTHONPATH so we don't mess python3-using apps launched from
# our Ruby processes (like the aws cli)
ENV.delete('PYTHONPATH')
```

However, I think we might have to re-do it post-fork. I can't figure out where/how the tests fork, so its not working as-is. Specifically, if I add this to pycall.rb, when I run `bundle exec spring testunit ./test/lib/pycall_test.rb` I get this several times:

```
Python path configuration:
  PYTHONHOME = '/opt/homebrew/opt/python@3.12/Frameworks/Python.framework/Versions/3.12:/opt/homebrew/opt/python@3.12/Frameworks/Python.framework/Versions/3.12'
  PYTHONPATH = (not set)
  program name = '/opt/homebrew/Cellar/python@3.11/3.11.9_1/Frameworks/Python.framework/Versions/3.11/Resources/Python.app/Contents/MacOS/Python'
  isolated = 0
  environment = 1
  user site = 1
  safe_path = 0
  import site = 1
  is in build tree = 0
  stdlib dir = '/opt/homebrew/opt/python@3.12/Frameworks/Python.framework/Versions/3.12/lib/python3.11'
  sys._base_executable = '/opt/homebrew/Cellar/python@3.11/3.11.9_1/Frameworks/Python.framework/Versions/3.11/bin/python3.11'
  sys.base_prefix = '/opt/homebrew/opt/python@3.12/Frameworks/Python.framework/Versions/3.12'
  sys.base_exec_prefix = '/opt/homebrew/opt/python@3.12/Frameworks/Python.framework/Versions/3.12'
  sys.platlibdir = 'lib'
  sys.executable = '/opt/homebrew/Cellar/awscli/2.17.21/libexec/bin/python'
  sys.prefix = '/opt/homebrew/opt/python@3.12/Frameworks/Python.framework/Versions/3.12'
  sys.exec_prefix = '/opt/homebrew/opt/python@3.12/Frameworks/Python.framework/Versions/3.12'
  sys.path = [
    '/opt/homebrew/opt/python@3.12/Frameworks/Python.framework/Versions/3.12/lib/python311.zip',
    '/opt/homebrew/opt/python@3.12/Frameworks/Python.framework/Versions/3.12/lib/python3.11',
    '/opt/homebrew/opt/python@3.12/Frameworks/Python.framework/Versions/3.12/lib/python3.11/lib-dynload',
  ]
Fatal Python error: init_fs_encoding: failed to get the Python codec of the filesystem encoding
Python runtime state: core initialized
ModuleNotFoundError: No module named 'encodings'

Current thread 0x00000001f6ae4c00 (most recent call first):
  <no Python frame>
```

What's odd is the tests still pass? 🤷

snickell:
PyCall /also/ sets PYTHONHOME even if we don't. By unsetting both of them, it seems to not interfere with aws cli AND we don't get warnings on pycall_test, at the end of pycall.rb we now do:
```
# Now unset the PYTHONPATH & PYTHONHOME so we don't mess up python3-using apps
# launched from our Ruby processes (like the aws cli)
ENV.delete('PYTHONPATH')
ENV.delete('PYTHONHOME')
```

snickell:
This PR has been switched from using `pipenv` to now using `pdm`.

## reasons for switching from pipenv to pdm

pdm has all the same basic upsides as pipenv, except it is not blessed by the PyPa. However, it /does/ follow PEP 621, unlike pipenv, which makes it potentially more future compatible with other python package managers. pdm is relatively young, but its usage is growing.

1. Practical reason: pyenv (required for pipenv) was a pain to get working on chef, whereas pdm was very easy
2. Architectural reason: pdm lets us split our python code into sub-packages, which can define their own deps, testing systems, etc.
3. pdm frequently doesn't have to compile python to get an exact right version, which makes it less annoying to use in a dockerfile or chef

For posterity, before I remove this from the PR's description, here's our description of why we were using pipenv:

```
## old pipenv section

pipenv was selected because it is officially blessed by PyPa (the pip people) and Python.org to be Python's next packaging approach. Its relatively early days for pipenv, but its recently picked up momentum and does appear to be in practice solidifying its blessed position as the default packaging approach for modern python.

pipenv solves both the python interpreter version problem, the virtualenv problem, and the dependency versioning problem. It shares a lot in common with ruby's bundle+rbenv, and I believe is heavily inspired by the combination:
- `bundle install` => `pipenv install`
- `bundle exec ____` => `pipenv run ____`

- pipenv uses a lockfile approach, which allows the Pipfile to focus on what's actually compatible and not try to double as an imperfect lockfile (ala requirements.txt)
- pipenv includes builtin virtualenv support, similar to ruby's `bundle` (in fact, appears to potentially be modeled off it?)
  - the virtualenv is connected to a repo's directory and is "in effect" anywhere inside that directory, much like `bundle`
  - From the commandline `pipenv run _____` can be used to run things within the virtualenv, or `pipenv shell` can be used to open a shell where the virtualenv has taken over the path
  - pyenv shims can be optionally used (similar to rbenv shims) so that when inside the dir, `which python` will always result in the repo python version and packages. This PR does not set these up, but the setup is compatible with them.
- `pipenv install` has similar functionality to `yarn install` in that it both installs what's defined in the lockfile, and can be used to add new packages to the Pipfile and break the lock to update packages
- pipenv uses pyenv to specify an exact python version, which will be built if system python is not a suitable version. built pythons are located in ~/.pyenv and linked into the virtual env, which is activated based on directory.
```

snickell:
We now end pycall.rb with the following clause, which means we don't need to worry about shelling out to python-using cli utils like `aws` (now tested as working):
```
ENV.delete('PYTHONPATH')
ENV.delete('PYTHONHOME')
ENV.delete('PYTHON')
```

This renders the following section from the PR description moot, and I have now removed it:
```
### Effect on Python in sub-shells

Due to pycall internals, directing pycall toward our virtual env requires setting the `PYTHON` and `PYTHONPATH` env vars. This would be inherited by any spawned sub-shells, if they were to invoke python3 they would end up using the Python and packages from our virtualenv.

- This affects any python invocation _within our main repo after loading our rails environment_
- This includes many/most rake tasks, cronjobs and oneoff scripts
- This could affect the version of python used if a rake task were to call into aws-cli. I had trouble finding examples of this since we generally use the aws-sdk gem, but mentioning in case infra team can think of a time we use the cli (credit to Suresh for mentioning this vector some time ago)

Essentially the pyenv / Pipfile in the root of our repo defacto specifies the python interpreter and packages available in all of our invocations and usage - the same as rbenv specifies the ruby interpreter for our whole repo.

Our ubuntu Dockerfile does currently include a python3 intrepeter, which will be aliased by the pyenv interpreter. However, the installation doesn't even include pip, and system dependencies hardcode the system interpreter and should not be affected.
```

lib/cdo/python_venv.rb
```diff
@@ -0,0 +1,69 @@
...
+    # --frozen-lockfile: don't update the lockfile
+    # --check: fail if lockfile is not up-to-date
+    flags = if ENV['CI']
+              '--dev --frozen-lockfile --check'
+            elsif rack_env? :production
+              '--prod --frozen-lockfile'
+            elsif rack_env? :test
+              '--dev --frozen-lockfile --check'
+            else # staging, adhocs, local environments
+              '--dev'
```
snickell, line 16: @code-dot-org/infrastructure lets say we have two PRs that both modify the package deps, and they both merge and are included in the same staging build. `pdm install` will be run during `rake build`, and this will generate a modified `pdm.lock` (because we aren't running with `--frozen-lockfile` in staging). Will this modified `pdm.lock` be committed similar to how schema.rb changes get recommitted? Do I need to do something intentionally to get that?

lib/cdo/pycall.rb
```diff
@@ -0,0 +1,33 @@
...
+
+# Do an import to ensure the Python interpreter is initialized.
+pyimport 'pycdo'
+
+# Now unset PYTHON, PYTHONPATH & PYTHONHOME so we don't mess up python3-using apps
+# launched from our Ruby processes (like the aws cli)
+# see: https://github.com/code-dot-org/code-dot-org/pull/60048#issuecomment-2267510208
+ENV.delete('PYTHONPATH')
+ENV.delete('PYTHONHOME')
+ENV.delete('PYTHON')
```
snickell, line 37: We spawn our python interpreter, and then unset PYTHON env vars (including those we don't set but are set by PyCall). This makes stuff like the AWS cli happy if spawned from our process.

pyproject.toml
```diff
@@ -0,0 +1,22 @@
+[project]
+requires-python = "==3.12.*"
+
+dependencies = [
+    # CDO python packages, from /python:
+    "pycdo @ file:///${PROJECT_ROOT}/python/pycdo",
```
snickell, line 6: See [/python/pycdo/pyproject.toml](https://github.com/code-dot-org/code-dot-org/blob/seth/setup-pycall/python/pycdo/pyproject.toml), which includes deps on outside packages. These are laddered up 
 by pdm and included in the overall venv for code-dot-org.

The idea (refined by practice in future PRs) is that we'll have a `/python/pycdo` package with its own deps as a sort of equivalent of lib/cdo. This is where you might put small python snippets that don't represent large features.

Parallel to it, we'll have feature packages for our larger python-using features, e.g. we might have `/python/aitutor/pyproject.toml`. Each python sub-package will define its own deps, tests, etc.

cookbooks/cdo-apps/libraries/cdo_apps.rb
```diff
@@ -17,7 +17,9 @@ def setup_app(app_name)
     execute "setup-#{app_name}" do
       command "bundle exec rake #{app_name}:setup_db --trace"
       cwd app_root
-      environment env.merge(node['cdo-apps']['bundle_env'])
+      environment env.merge(node['cdo-apps']['bundle_env'], {
+        HOME: "/home/#{user}"
+      })
```
snickell, line 22: This chef recipe executes `bundle exec rake dashboard:setup_db`, which in turn loads pycall.rb, which does a `pdm run which python` to find the venv. However, despite running the execute with `user user`, chef does NOT set the env vars so its like running with `su blah` rather than `su - blah`. PDM reads the `HOME` env var to decide where to store config, and finds it cannot read `/root/.config` and bails out.

The fix? Just set HOME before shelling out to run `bundle exec rake dashboard:setup_db`

cookbooks/cdo-apps/libraries/cdo_apps.rb
```diff
@@ -17,7 +17,7 @@ def setup_app(app_name)
     execute "setup-#{app_name}" do
       command "bundle exec rake #{app_name}:setup_db --trace"
       cwd app_root
-      environment env.merge(node['cdo-apps']['bundle_env'])
+      environment env.merge(node['cdo-apps']['bundle_env'], {HOME: "/home/#{user}"})
       live_stream true
       user user
       group user
```
snickell, line 23: This chef recipe executes `bundle exec rake dashboard:setup_db`, which in turn loads pycall.rb, which does a `pdm run which python` to find the venv. This breaks before this change.

Why? despite running the execute with user user, chef does NOT load the user's env vars, so its like running with `su blah` rather than `su - blah`. PDM reads the HOME env var to decide where to store config, finds it cannot read /root/.config and bails out.

The fix? Just set HOME before shelling out to run bundle exec rake dashboard:setup_db

lib/cdo/python_venv.rb
davidsbailey: there are two things going on which seem relevant to this question:

* staging updating schema_cache.yml (if needed) on every build: https://github.com/code-dot-org/code-dot-org/blob/f1e4094a2b14d3e76bd4edbd7c2b42495f47a57a/lib/rake/build.rake#L62-L84
* nightly staging content scoop does not appear to be explicitly excluding schema.rb or any other non-content-related files. you can see from that file's [history](https://github.com/code-dot-org/code-dot-org/commits/staging/dashboard/db/schema.rb) that it sometimes gets updated as part of the staging scoop

lib/cdo/rake_utils.rb
```diff
@@ -131,6 +132,10 @@ def self.run_bundle_command(*args)
     end
   end
 
+  def self.python_venv_install
```
davidsbailey, line 135: I would vote for making this its own build rake target, so that python developers do not have to run all of `rake build` in order to regenerate the venv

davidsbailey (commented):
just one comment after running through setup

lib/cdo/rake_utils.rb
snickell: This ruby method just runs `pdm install`. I was thinking this would be sort of like running `yarn install`: active JS developers probably run the command directly rather than via rake? LMK if that doesn't make sense to you, happy to add a direct rake target too.

davidsbailey: yes, this makes sense to leave as is then, thanks!

python/README.md
```diff
@@ -0,0 +1,139 @@
...
+  - example: `pdm run ipython`: start ipython
+  - example: `pdm run pytest`: run pytest in the current dir
+- `source .venv/bin/activate`: activate the python virtual environment created by pdm (see `pdm venv activate` for exact command)
+
+## pycall.rb: how to invoke python code from rails
+
+- Ruby => Python function calls are very fast: microseconds not milliseconds
+- Ruby callbacks can be passed into Python functions
+- Python packages, specified in sub-packages like `/python/pycdo/pyproject.toml`, can be imported in Ruby using `pyimport('pipname')`
+- Python code can be added to pycdo, e.g. /python/pycdo/mynewfeature.py, or by creating a new toplevel package like /python/myfeature. See [how to structure your python code](#how-to-structure-your-python-code).
```
davidsbailey, line 45: given that we have not yet solved the threading issues, perhaps there should be a warning here saying only to access python code from ruby via active job or unit tests for now?

pdm.lock
davidsbailey, line 1: our yarn.lock file lives in /apps. any reason pdm.lock shouldn't live inside /python?

.gitignore
```diff
@@ -93,10 +93,14 @@ npm-debug.log
 /pegasus/config/newrelic.yml
 /pegasus/.sass-cache
 /pegasus/sites.v3/*/public/css/generated
+/pybin/
```
davidsbailey: can this be removed now? I don't see anything here in my local filesystem after running through the setup steps and pycall_test.rb.

.gitignore
```diff
@@ -93,10 +93,14 @@ npm-debug.log
...
 /tools/scripts/brokenLinkChecker/node_modules/
 /i18n/crowdin/*
 /i18n/locales/*/codeorg-markdown/
 /i18n/locales/*/curriculum_content/
 /i18n/locales/*/piskel/
 /i18n/locales/*/standards/
 /i18n/locales/*/docs/
+
+# Specifies the user-specific path to the venv for pdm
+.pdm-python
```
davidsbailey: super tiny nit: please add back a trailing newline, these can cause headaches when omitted.

davidsbailey (approved):

pdm.lock
snickell: pdm.lock has to be parallel to pyproject.toml, and that in turn determines the root of our venv and thereby what paths `pdm run` works from. While I would have preferred to keep files out of the root dir, it seemed somehow clearer to match the semantics for `bundle exec` which is also whole repo. This links to python_venv.rb, which also depends on the "whole repo" thing. This could be altered, but this way was sort of the "natural" way.

Another thought is that (post-threading issues, I have a good start on those here btw: https://github.com/snickell/pycall_thread) you can use pyimport in Ruby paths, so its not just a /python thing.

python/README.md
snickell: I also added a comment to `locals.yml.default` about switching the queue_adapter to delayed_job:
```
# Uncomment to use `delayed_job` to fulfill ActiveJob requests. This is useful
# for matching prod, and is also currently required to use ActiveJobs that
# invoke Python via PyCall. These will SEGV if run with the default async
# adapter, because it uses threads instead of processes.
#
# If you enable `delayed_job`, you will need to manually run `bundle exec bin/delayed_job start`
# from the dashboard/ directory to start the workers.
#
# active_job: { queue_adapter: delayed_job }
```

snickell: Oof, a new rabbit hole, have to make that actually configurable via config.yml.erb.

.drone.yml
```diff
@@ -189,7 +189,7 @@ steps:
 
   - name: ui-tests
     # see: docker/dockerfiles/Dockerfile
-    image: codedotorg/code-dot-org:1.9.4
+    image: codedotorg/code-dot-org:pycall-pdm
```
cat5inthecradle: Need to push a new image after this is merged to go back to version numbering (and/or merge my docker refactor pr)

SETUP.md
```diff
@@ -251,6 +252,9 @@ Note: Virtual Machine Users should check the [Alternative note](#alternative-use
     1. If your PATH is missing `~/.rbenv/shims`, the next two commands might not work. Edit your .bashrc to include the following line:
        `export PATH="$HOME/.rbenv/bin:~/.rbenv/shims:$PATH"`, then run `source .bashrc` for the change to take effect (as seen in [this github issue](https://github.com/rbenv/rbenv/issues/877)).
     1. `rbenv rehash`
+1. Install pdm, which will be used later by `rake install` to install python
+    1. `sudo pip3 install --prefix=/usr/local --upgrade pdm`
+        - alternatively, if you prefer pipx and have it configured path-wise: `pipx install pdm`
```
cat5inthecradle, line 257: This hasn't been working for me, not sure if it's the instructions or my environment.
image

python/README.md
davidsbailey: Oof indeed! for that rabbit hole, you should be able to set a CDO variable in this file and then pass it here: https://github.com/code-dot-org/code-dot-org/blob/90410de82aae9fe48f0165acbc7d5dcf87b1e205/dashboard/config/application.rb#L223

if pycall works only on the main thread, then one idea would be to wrap our access to it (calls to `pyimport`,  etc) with a ruby check for `Thread.current == Thread.main`, followed by a helpful error message explaining the requirements for running pycall rather than an opaque seg fault.

I don't want to go too crazy with protections given our intent to solve the threading issue, so feel free to put this off though.

davidsbailey: or perhaps lib/cdo/pycall.rb could simply provide a method
```
def safe_to_run_pycall?
  Thread.main == Thread.current
end
```
which developers can then use in their active jobs before calling into pycall.

.drone.yml
snickell: 💯, got it on the checklist

SETUP.md
snickell: 1. Which ubuntu? Maybe I can replicate in docker
2. Does `/usr/local/bin/pdm` exist? 
  - If so, is `/usr/local/bin` in your path?
    - If so, maybe you just need a `zsh rehash` to rescan the PATH?
3. What about `/usr/local/lib/python3.8/dist-packages/pdm`
  - it might be other than python3.8 in the path, esp if not on ubuntu 20.04: `ls /usr/local/lib/`
 4. If all else fails: `find /usr/local | grep pdm`, on 20.04 clean docker mine produces:
 ```
 circleci@75027a115ad9:/usr/local$ find /usr/local | grep pdm
 /usr/local/lib/python3.8/dist-packages/pdm-2.17.3.dist-info
/usr/local/lib/python3.8/dist-packages/pdm-2.17.3.dist-info/WHEEL
/usr/local/lib/python3.8/dist-packages/pdm-2.17.3.dist-info/RECORD
/usr/local/lib/python3.8/dist-packages/pdm-2.17.3.dist-info/INSTALLER
/usr/local/lib/python3.8/dist-packages/pdm-2.17.3.dist-info/licenses
/usr/local/lib/python3.8/dist-packages/pdm-2.17.3.dist-info/licenses/LICENSE
/usr/local/lib/python3.8/dist-packages/pdm-2.17.3.dist-info/METADATA
/usr/local/lib/python3.8/dist-packages/pdm-2.17.3.dist-info/entry_points.txt
/usr/local/lib/python3.8/dist-packages/pdm
/usr/local/lib/python3.8/dist-packages/pdm/project
SNIP
/usr/local/lib/python3.8/dist-packages/pdm/resolver/core.py
/usr/local/lib/python3.8/dist-packages/pdm/resolver/__init__.py
/usr/local/lib/python3.8/dist-packages/pdm/resolver/providers.py
/usr/local/bin/pdm
```

cnbrenci (approved):
Love to see it! 🚀

snickell:
Docker tagged, .drone.yml update with new tag

SETUP.md
fisher-alice: I ran into the same thing, and it was a path issue. Thanks Seth!

PR #60053, opened 2024-07-29
[Aichat] Add ChatEvent type

fisher-alice:
This PR adds to the `aichat` lab a new `ChatEvent` type that includes chat messages, notifications, model updates, and general `Aichat` user events. This will include when the user clears chat history with the 'Clear chat' button and when a student loads the `aichat` level.

All of these `aichat` user events will be viewable by the teacher in the teacher view of student chat history. https://github.com/code-dot-org/code-dot-org/pull/59981

There is no impact on the UI.

## Links
Step 1 of [jira](https://codedotorg.atlassian.net/browse/LABS-940)

## Testing story
I checked locally on `aichat` levels in /allthethings to confirm there is no impact on the user experience.

## PR Checklist:

- [ ] Code is well-commented


apps/src/aichat/types.ts
```diff
@@ -5,41 +5,44 @@ import {Role} from '../aiComponentLibrary/chatItems/types';
 // TODO: Update this once https://codedotorg.atlassian.net/browse/CT-471 is resolved
 export type AichatInteractionStatusValue = string;
 
-export interface ChatItem {
+export interface AichatEvent {
   // UTC timestamp in milliseconds
   timestamp: number;
+  description?: string;
```
sanchitmalhotra126: What is `description` used for / planned to be used for?

apps/src/aichat/types.ts
```diff
@@ -5,41 +5,44 @@ import {Role} from '../aiComponentLibrary/chatItems/types';
...
 
-export interface Notification extends ChatItem {
+export interface Notification extends AichatEvent {
   id: number;
   text: string;
   notificationType: 'error' | 'success';
 }
 
-// Type Predicates: checks if a ChatItem is a given type, and more helpfully,
+export type ChatItem = ChatMessage | ModelUpdate | Notification;
```
sanchitmalhotra126: Can we add some comments to explain the differences between `ChatItem`s and `AichatEvent`s?

apps/src/aichat/redux/aichatRedux.ts
```diff
@@ -65,6 +66,8 @@ export interface AichatState {
   chatMessagePending?: ChatMessage;
   // Denotes whether we are waiting for a chat completion response from the backend
   isWaitingForChatResponse: boolean;
+  // Student events viewd by a teacher user in chat workspace
+  allStudentAichatEvents: AichatEvent[];
```
sanchitmalhotra126: nit: maybe `studentChatHistory: AichatEvent[]` might be clearer? or even `viewAsUserChatHistory` if we wanted to get even more generic

sanchitmalhotra126 (commented):
Nice to see this coming in! A few comments

apps/src/aichat/types.ts
fisher-alice: This will be the text to display in the teacher view of student chat history for events that are not chat messages. Example: 'User cleared chat history.'

apps/src/aichat/types.ts
fisher-alice: How does this sound:
```
// ChatItems are types of messages that are displayed for the user in the chat workspace.
// ChatEvents that are not ChatItems are NOT displayed to the user. However, they
// are displayed when a teacher views a student's chat history in the chat workspace.
// An example of a ChatEvent that is not a ChatItem is 'The user cleared the chat workspace.'
```

apps/src/aichat/redux/aichatRedux.ts
fisher-alice: I like `studentChatHistory`! Will update.

apps/src/aichat/types.ts
fisher-alice: I can add a comment including this.

sanchitmalhotra126: Ah I see - it might be better from a translations/presentation perspective to have discrete event types for each these scenarios (e.g. `ClearEvent extends AichatEvent`) especially if we don't anticipate having a lot of them? Alternatively, we could keep it flexible but in that case we should maybe store a string key here instead of the description directly.

fisher-alice: That makes sense. I like the idea of discrete event types since I agree there won't be many of them.

apps/src/aichat/views/AichatEventView.tsx
```diff
@@ -43,17 +43,19 @@ function formatModelUpdateText(update: ModelUpdate): string {
 }
 
 /**
- * Renders AI Chat {@link ChatItem}s using common AI design components.
+ * Renders AI Chat {@link Chatevent}s using common AI design components.
```
cnbrenci: nit: should this say AichatEvent?

apps/src/aichat/types.ts
```diff
@@ -5,41 +5,51 @@ import {Role} from '../aiComponentLibrary/chatItems/types';
...
   text: string;
   notificationType: 'error' | 'success';
 }
 
-// Type Predicates: checks if a ChatItem is a given type, and more helpfully,
+// ChatItems are types of messages that are displayed for the user in the chat workspace.
+// ChatEvents that are not ChatItems are NOT displayed to the user. However, they
+// are displayed when a teacher views a student's chat history in the chat workspace.
+// An example of a ChatEvent that is not a ChatItem is 'The user cleared the chat workspace.'
+export type ChatItem = ChatMessage | ModelUpdate | Notification;
```
cnbrenci: The comment is really helpful, but I think there are ways in code to make the distinction more clear.  Maybe as simple as renaming it something like VisibleAichatEvents? 

Or give AichatEvent a static property like `IsVisibleInStudentChatLog` defaulting to `false`, and overridden to `true` in `ChatMessage`, `ModelUpdate` and `Notification` 

or a helper method on AiChatEvent for the same type of thing `IsVisibleInStudentChatLog`  that checks type might be easier to avoid extraneous stuff being serialized to the db when saving these events

apps/src/aichat/types.ts
fisher-alice: Actually - to make the view of the `AichatEvent` more simple,  I opted for storing a string key instead. Lmk what you think.

apps/src/aichat/types.ts
fisher-alice: `VisibleAichatEvents` is a bit long and `ChatItem` is already used quite a bit now in the code base including the name of the shared subdirectory `chatItems/aiComponentLibrary` so I prefer not to change the name.
I would also like to avoid adding another property to `AichatEvent`.
Do you mind explaining a little more in detail about the helper method? Would we still be able to [type arrays](https://github.com/code-dot-org/code-dot-org/blob/92af97d9256167104efb1e135de4e9fffd576b6f/apps/src/aichat/redux/aichatRedux.ts#L64) as `ChatItem` like `chatItemsCurrent`? Student chat history will be stored in an`AichatEvents` array separately. Thanks!

fisher-alice: FYI - I tweaked the comment a bit.

fisher-alice: Yay! After chatting with @cnbrenci and @sanchitmalhotra126, we reached a consensus on renaming `AichatEvent` as `ChatEvent`, adding optional fields: `hideForParticipants` and `description` (which will use keyed strings), and removing `ChatItem`. 
Turns out Sanchit created the `chatItems` subdirectory and AiTutor is not using `ChatMessage` as of today. So we felt more leeway to changing the name.

Thanks to you both for taking time to chat with me on this! ❤️

cnbrenci: Thanks for facilitating the discussion! I learned more about TS too along the way :) 

Also to note, we decided to leave the db table name as-is (aichatEvent) since it makes sense to have that specificity at the db level but it's a bit redundant to keep the 'ai' prepended in the context of the code.

cnbrenci (approved):
🎉

apps/src/aichat/types.ts
fisher-alice: @sanchitmalhotra126 - will merge now to unblock myself but feel free to make comments and can address in follow-up tasks. Thanks!

apps/src/aichat/types.ts
```diff
@@ -1,45 +1,52 @@
...
+} as const;
 
 // TODO: Update this once https://codedotorg.atlassian.net/browse/CT-471 is resolved
 export type AichatInteractionStatusValue = string;
 
-export interface ChatItem {
+export interface ChatEvent {
   // UTC timestamp in milliseconds
   timestamp: number;
+  hideForParticipants?: true;
```
sanchitmalhotra126, line 16: Minor (as a follow-up) - let's add a comment here explaining why this field is optional but type `true` instead of boolean (doesn't have to defined, but if defined it must be true, i.e. we don't want `hideForParticipants: false`)

apps/src/aichat/views/ChatEventView.tsx
```diff
@@ -64,12 +67,22 @@ const ChatItemView: React.FunctionComponent<ChatItemViewProps> = ({item}) => {
...
+        type="success"
+        onClose={() => dispatch(removeUpdateMessage(event.id))}
+        size="s"
+      />
+    );
+  }
+
+  if (event.descriptionKey) {
+    return (
+      <Alert
```
sanchitmalhotra126, line 83: I'm sure we'll iterate on this when we actually start displaying chat history events - I was thinking that we might want to display the non-alert events (like clear/load) differently from notifications? Probably something we can run by Mark/Sam in a later PR

sanchitmalhotra126 (commented):
Nice work! Glad we could talk through all the typing/naming and appreciate the insight from both you and @cnbrenci! I know this has been merged but LGTM - just left a couple comments for follow-up consideration.

apps/src/aichat/types.ts
fisher-alice: Good call! Will do in this follow-up PR: https://github.com/code-dot-org/code-dot-org/pull/60110

apps/src/aichat/views/ChatEventView.tsx
fisher-alice: Sounds good - when we get to point of fetching student chat history on the frontend - will definitely consult with Mark/Sam. Thanks!

PR #60055, opened 2024-07-29
Adds ErrorBoundary and renders a friendlier error message on internal JS error.

wilkie:
Introduces an error boundary that prevents global page crashes when something is wrong within the rubric container and renders the error message within.

Ideally, nobody would ever see this, but this at least provides some place for errors to be caught within the AI rubric context and not have that leak out.

image

image

## Links

- jira ticket: [AITT-494](https://codedotorg.atlassian.net/browse/AITT-494)

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/i18n/common/en_us.json
+++ b/apps/i18n/common/en_us.json
@@ -2382,6 +2382,7 @@
   "rubricNextLearningGoal": "Go to next learning goal",
   "rubricPreviousLearningGoal": "Go to previous learning goal",
   "rubricAiHeaderText": "AI Teaching Assistant",
+  "rubricAiInternalError": "We're sorry, but there is an error on our end.",
   "rubricTourNextButtonText": "Next Tip",
   "rubricTourStepOneTitle": "Getting Started with Your AI Teaching Assistant",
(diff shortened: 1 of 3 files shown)


davidsbailey:
Hey Wilkie, I saw your link to this PR. I am wondering, the goal of the error boundary is to make the rest of the page still usable if something goes wrong inside TA. At first glance it looks hard to use the rest of the page in the screenshot. do you think the error boundary could be moved up to the FAB? open to other ideas here.

wilkie:
I could do that, but it seemed reasonable to provide the context that there was an error to the user that might expect the rubric modal to exist. They can still close the modal by clicking on the button still.

davidsbailey:
ah, It didn't occur to me to click the TA button. I like the idea of using the box to show the error text, but I'm worried users will make the same mistake I did and will have to reload the page, defeating somewhat the purpose of the error boundary which is to make sure the rest of the page remains functional.  if there was an easy way to keep the `xmark` button that would be ideal!

wilkie:
Ah! Yes, I think the 'x' button can be there. Let me do that.

wilkie:
image

The 'x' closes the window.

davidsbailey (approved):
looks great! thanks for the follow-up.

PR #60056, opened 2024-07-29
Fix applab design mode layout for RTL languages

cnbrenci:
Design mode in applab was not reflowing correctly, resulting in the design toolbox rendering behind the design workspace and being unusable. 

Here is a screenshot with highlighted incorrectly laid out elements:
image

 
Fixed:

https://github.com/user-attachments/assets/45903d93-b597-46f7-a11c-c9a8db539878

## Links

- jira ticket: [Applab: design toolbox is not visible for RTL languages](https://codedotorg.atlassian.net/browse/LABS-920)


apps/src/applab/DesignModeHeaders.jsx
```diff
@@ -84,23 +85,26 @@ export default class DesignModeHeaders extends React.Component {
         display: this.props.isToolboxVisible ? 'flex' : 'none',
         justifyContent: 'space-between',
         width: 270,
-        borderRight: '1px solid gray',
-        float: 'left',
+        borderLeft: this.props.isRtl ? '1px solid gray' : '',
```
fisher-alice, line 88: Non-blocking but if you have time, these could be put in a separate css module - kinda busy in this react file.

fisher-alice (approved):
Nice fix! Interesting that it's only been reported recently. Was there a regression somewhere along the line?
Left a minor, non-blocking suggestion. Thanks for taking care of this!

apps/src/applab/DesignModeHeaders.jsx
cnbrenci: Yeeeeaaahhhh, I decided against it cause it would take more time than it's worth rn 😶‍🌫️

PR #59875, opened 2024-07-18
Use new survey link for Build Your Own workshops

TurnerRiley:
We are going to use a special post-workshop survey link for the new Build Your Own workshops. This PR updates where we obtain that survey link to return the new one for Build Your Own workshops, and still returns a Foorm otherwise.

This was tricky to test because the new survey page doesn't work locally since it is unfinished (there is a placeholder live now on production at [studio.code.org/form/professional_learning_educator_post_workshop_survey](https://studio.code.org/form/professional_learning_educator_post_workshop_survey) for teachers and [studio.code.org/form/professional_learning_facilitator_post_workshop_survey](https://studio.code.org/form/professional_learning_facilitator_post_workshop_survey) for facilitators), so I was making sure it sent to the right URL (which locally shows a sad bee). In addition, I don't have Foorm set up locally so I was testing that it sent to the proper URL and showed Foorm errors in the console. I felt comfortable not doing more to check Foorm since I didn't touch anything with Foorm itself and I didn't want to spend too much time on this ticket (but I'm open to pushback if this feels necessary).

### Post-workshop survey link on My PL page after workshop ends
Non-Build Your Own uses the foorm:
(I don't have Foorms set up locally but this shows that it's sending to the right url and has errors expecting values for Foorm)
right url

Build Your Own sends user to `/form/professional_learning_educator_post_workshop_survey`:
(this page is currently not ready so locally it shows a sad bee, but on production there is a placeholder at [/form/professional_learning_educator_post_workshop_survey](https://studio.code.org/form/professional_learning_educator_post_workshop_survey))
form link educator

#### `exit_survey` email
The survey is linked here in the email:
teacher survey email

For a non-BYO workshop, clicking that link sends the user to that page with the Foorm errors as expected:
teacher email non byo shows other

For a BYO workshop, clicking that link sends the user to the sad bee page with the right url:
teacher survey byo sad bee

#### `facilitator_post_workshop` email
The survey is linked here in the email:
facilitator email survey link

For a non-BYO workshop, clicking that link sends the user to that page with the Foorm errors as expected:
facilitator email sends to foorm non byo

For a BYO workshop, clicking that link sends the user to the sad bee page with the right url:
facilitator email sends to special byo

## Links
Jira ticket: [here](https://codedotorg.atlassian.net/jira/software/c/projects/ACQ/boards/64?assignee=60d62161f65054006980bd71&selectedIssue=ACQ-2058)

## Testing story
Local testing and adding unit tests.

PR #60057, opened 2024-07-29
Fix safe chaining for new workshop surveys and multiple survey links

TurnerRiley:
Following [this HoneyBadger error](https://app.honeybadger.io/projects/3240/faults/109837158), it looked like the safe chaining [here](https://github.com/code-dot-org/code-dot-org/blob/staging/dashboard/app/controllers/pd/workshop_daily_survey_controller.rb#L104) wasn't working as expected. The `get_workshop_by_enrollment_or_course_and_subject` method actually returns `false` if it doesn't get a result (rather than `nil` which is what the safe chaining is waiting for). I dug a bit deeper and actually realized this is later called within `new_general_foorm` so it doesn't really make sense to call it here since it can return a `render` if there's an error. Since I was only using this to get the workshop, I just updated it to be an actual query.

In addition, we were seeing lots of [sad bees](https://codedotorg.slack.com/archives/C04540KNGDQ/p1722271354567879) for post-workshop emails that had been sent out before the different post-survey routes [were consolidated](https://github.com/code-dot-org/code-dot-org/pull/59875/files#diff-3fffe83c5f87f14c4c59c00fc3556305f7d4a28ee7748b688700228b3c64e70d) from both `workshop_post_survey?enrollment_code=[:enrollment_code]` and `workshop_survey/post/:enrollment_code` to just `workshop_survey/post/:enrollment_code` (meaning any email that was still pointing to `workshop_post_survey?enrollment_code=[:enrollment_code]` would have an invalid survey link). This PR adds that route back in but just points it to `workshop_survey/post/:enrollment_code`. We can remove this line as cleanup in the future since it's a tiny change and doesn't feel worth keeping (if we kept it, we would have to figure out all the workshops affected and re-send the post-workshop email with the new url).

## Links
Honeybadger link: [here](https://app.honeybadger.io/projects/3240/faults/109837158)
Slack discussion about sad bees: [here](https://codedotorg.slack.com/archives/C04540KNGDQ/p1722271354567879)

## Testing story
- Checked that tests still passed for obtaining the right post-workshop link
- Locally tested that both versions of the post-workshop link show the same page (note: Foorm is not set up locally for me so this is the page I'm expecting to see when I navigate to a new Foorm):

With `workshop_post_survey?enrollment_code=[:enrollment_code]`:
workshop_post_survey

With `workshop_survey/post/:enrollment_code`:
new url

### DIFF
--- a/dashboard/app/controllers/pd/workshop_daily_survey_controller.rb
+++ b/dashboard/app/controllers/pd/workshop_daily_survey_controller.rb
@@ -101,8 +101,8 @@ module Pd
     # GET /pd/workshop_survey/post/:enrollment_code
     # If Build Your Own, redirect to its specific survey link. Otherwise, use Foorm.
     def new_post
-      course = get_workshop_by_enrollment_or_course_and_subject(enrollment_code: params[:enrollment_code], course: nil, subject: nil)&.course
-      if course == COURSE_BUILD_YOUR_OWN
+      workshop = get_workshop_by_enrollment_or_course_and_subject(enrollment_code: params[:enrollment_code], course: nil, subject: nil)
+      if workshop && workshop.course == COURSE_BUILD_YOUR_OWN
         redirect_to CDO.studio_url SURVEY_LINKS[:COURSE_BUILD_YOUR_OWN_TEACHER], CDO.default_scheme
       else
         # If the post-survey has been already completed, will redirect to thanks page.
--- a/dashboard/app/controllers/pd/workshop_daily_survey_controller.rb
+++ b/dashboard/app/controllers/pd/workshop_daily_survey_controller.rb
@@ -101,8 +101,8 @@ module Pd
     # GET /pd/workshop_survey/post/:enrollment_code
     # If Build Your Own, redirect to its specific survey link. Otherwise, use Foorm.
     def new_post
-      workshop = get_workshop_by_enrollment_or_course_and_subject(enrollment_code: params[:enrollment_code], course: nil, subject: nil)
-      if workshop && workshop.course == COURSE_BUILD_YOUR_OWN
+      course = Pd::Enrollment.find_by(code: params[:enrollment_code], user: current_user)&.workshop&.course
+      if course == COURSE_BUILD_YOUR_OWN
         redirect_to CDO.studio_url SURVEY_LINKS[:COURSE_BUILD_YOUR_OWN_TEACHER], CDO.default_scheme
       else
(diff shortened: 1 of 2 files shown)


hannahbergam (approved):
LGTM!

dashboard/app/controllers/pd/workshop_daily_survey_controller.rb
```diff
@@ -101,8 +101,8 @@ def new_general_foorm(survey_names:, day:, subject: FOORM_GENERAL_SURVEY_SUBJECT
     # GET /pd/workshop_survey/post/:enrollment_code
     # If Build Your Own, redirect to its specific survey link. Otherwise, use Foorm.
     def new_post
-      course = get_workshop_by_enrollment_or_course_and_subject(enrollment_code: params[:enrollment_code], course: nil, subject: nil)&.course
-      if course == COURSE_BUILD_YOUR_OWN
+      workshop = get_workshop_by_enrollment_or_course_and_subject(enrollment_code: params[:enrollment_code], course: nil, subject: nil)
```
bethanyaconnor: if this is `false`, what should happen? Should the else branch of the if/else below still happen?

TurnerRiley: Yep!

dashboard/config/routes.rb
```diff
@@ -803,6 +803,7 @@
       get 'workshop_daily_survey/day/:day', to: 'workshop_daily_survey#new_daily_foorm'
       get 'workshop_pre_survey', to: 'workshop_daily_survey#new_pre_foorm'
       post 'workshop_survey/submit', to: 'workshop_daily_survey#submit_general'
+      get 'workshop_post_survey', to: 'workshop_daily_survey#new_post'
```
hannahbergam, line 806: Can you link the old pr where those two routes were consolidated?

TurnerRiley: Oh yes sorry forgot to link that (it's in the PR description now): https://github.com/code-dot-org/code-dot-org/pull/59875/files#diff-3fffe83c5f87f14c4c59c00fc3556305f7d4a28ee7748b688700228b3c64e70d

hannahbergam: Thanks! So enrollment.rb is still set up to work in this old flow? Sorry just trying to make sure I understand the old pr considering it's kind of a partial revert/patch

TurnerRiley: Yep they both just point to the same `workshop_daily_survey` method!

PR #60059, opened 2024-07-29
Create folder for new Sign Up flow

kelbyhawn:
Creates a `signUpFlow` folder for the new Sign Up flow. We're going to add this to the `src` directory instead of in `templates` since this is a major feature.

## Links
Jira ticket: [ACQ-2144](https://codedotorg.atlassian.net/browse/ACQ-2144)

### DIFF
--- a/apps/src/signUpFlow/accountType.jsx
+++ b/apps/src/signUpFlow/accountType.jsx
@@ -1,0 +1,1 @@
+// Placeholder for the new account type selection page


hannahbergam (approved):
I could go either way on whether or not we need the word 'Flow' here, but that's a nit so approving!

TurnerRiley (approved):
LGTM!

bethanyaconnor:
This would be the first time in a while we've put something in the src/ directory over src/templates. I feel a bit mixed on that but I'll defer to you all there.

hannahbergam:
> This would be the first time in a while we've put something in the src/ directory over src/templates. I feel a bit mixed on that but I'll defer to you all there.

I feel weird about that templates directory... do we know what would ideally be in there? To me it seems like a directory for things that are reused across the site, but that's not how we currently organize those files😂

bethanyaconnor:
I feel weird about the templates directory too but I'm not convinced I'd want everyone to put their folders in the root directory either...

hannahbergam:
> I feel weird about the templates directory too but I'm not convinced I'd want everyone to put their folders in the root directory either...

Another option is to make a team folder. I have concerns about that though as our team names and configurations/codebase coverage change fairly frequently. 

Also, the other files I see in that directory seem to be on par with a medium sized collection of work (see flappy, aiTutor, aichat, weblab). Is there another 'in between' file name that could make sense here if we want to start moving our files from the templates directory?

kelbyhawn:
Bethany and I just talked about this in our 1:1 and we're going to keep it in the `src` directory. I'll add note about this in the PR description above too!

PR #60065, opened 2024-07-29
Speculative fix to send Facilitator post-workshop email for Build Your Own workshops

TurnerRiley:
Now that all the "Build Your Own workshop" content is in production, I was able to start testing some of it (since I wasn't able to test a lot of email and Foorm stuff locally). When testing, I noticed that the Facilitator post-workshop survey wasn't sending for "Build Your Own workshop" workshops. This adds "Build Your Own" workshop as a course type that sends emails.

## Testing story
Adding unit test.


hannahbergam (approved):
Thank you for adding this test!

PR #60069, opened 2024-07-29
Music: add additional Amplitude analytics

sanchitmalhotra126:
Adds some new analytics logging for Music Lab via Amplitude:
- Instead of just logging on standalone projects, we will now record analytics on all level types (i.e. script levels).
- In addition to our existing properties (user ID, user type, sign in state), we will now track the following properties on every Amplitude event
  - Level Type ("Standalone Project" or "Level")
  - View Mode ("Edit", "View", or "Share")
  - Channel ID
  - Level Path (ex: "/s/music-intro-2024/lessons/1/levels/4")
  - Script Name (ex: "music-intro-2024")
- Added events for clicking buttons on the share/play mobile player view ("remixFromShareView", "shareFromShareView", "viewCodeFromShareView").

This PR also uses the lifecycle hooks from https://github.com/code-dot-org/code-dot-org/pull/59855 to manage the analytics sessions when switching between levels in a progression.

### Examples from Amplitude

(these all include channel ID but it's separated due to alphabetization, so I only included a screenshot for the first one)

User properties for events on a script level:
Screenshot 2024-07-29 at 3 35 35 PM
Screenshot 2024-07-29 at 3 35 40 PM

User properties for a teacher viewing a level as a student:
Screenshot 2024-07-29 at 3 46 46 PM

User properties on a standalone project:
Screenshot 2024-07-29 at 3 48 48 PM

User properties on the share/play view for a standalone project:
Screenshot 2024-07-29 at 3 49 45 PM

User properties viewing another user's project:
Screenshot 2024-07-29 at 3 52 01 PM

Events & user properties for button clicks on the mobile share/play view.
Screenshot 2024-07-29 at 3 50 06 PM
Screenshot 2024-07-29 at 3 50 25 PM
Screenshot 2024-07-29 at 3 51 03 PM

## Links

https://codedotorg.atlassian.net/browse/LABS-901

## Testing story

Tested locally with Amplitude


apps/src/music/analytics/AnalyticsReporter.ts
```diff
@@ -116,36 +122,65 @@ export default class AnalyticsReporter {
...
+  setProjectProperty(
+    property: (typeof trackedProjectProperties)[number],
+    value: string | number | undefined
+  ) {
+    if (!this.session) {
+      this.log('No session in progress');
+      return;
+    }
+
+    const identifyEvent = new Identify();
```
fisher-alice, line 174: Just curious - before this update, I think there was one `Identify` instance being used within this class which was initialized in the constructor. Why are you now creating a new `Identify` instance every time you set a project property?

apps/src/music/views/hooks/useUpdateAnalytics.tsx
```diff
@@ -0,0 +1,114 @@
+import {useEffect} from 'react';
```
fisher-alice: Why is this a .tsx file and not a .ts file similar to `useUpdatePlayer.ts`?

apps/src/music/analytics/AnalyticsReporter.ts
sanchitmalhotra126: Yeah, I realized there were some issues with using a single Identify object where certain properties weren't being captured consistently. Amplitude's docs always seem to show using a single identify object per identify call, so I just went with that pattern here.

apps/src/music/views/hooks/useUpdateAnalytics.tsx
sanchitmalhotra126: Good catch :) should be a `.ts` file!

apps/src/music/analytics/AnalyticsReporter.ts
fisher-alice: Oh interesting - thanks!

apps/src/music/analytics/AnalyticsReporter.ts
```diff
@@ -50,6 +50,14 @@ interface BlockStats {
   maxTriggerBlocksWithCode: number;
 }
 
+interface Session {
```
fisher-alice, line 53: Nice clean up!

fisher-alice (approved):
Cool implementation! Also nice that you resolved the identify issues observer earlier.

apps/src/music/views/hooks/useUpdateAnalytics.ts
```diff
@@ -0,0 +1,114 @@
...
+    );
+
+    // Also attach a listener to end the session if necessary when the level is reloaded without
+    // changing levels, which can happen when the view as user is changed via the Teacher Panel.
+    lifecycleNotifier.addListener(LifecycleEvent.LevelLoadStarted, endSession);
+
+    // TODO: the 'beforeunload' callback is advised against as it is not guaranteed to fire on mobile browsers. However,
+    // we need a way of reporting analytics when the user navigates away from the page. Check with Amplitude for the
+    // correct approach.
+    window.addEventListener('beforeunload', endSession);
```
breville, line 61: Is this tracked as a follow-up work item?

breville (approved):
Excellent description.  Thanks for such a comprehensive implementation.

apps/src/music/views/hooks/useUpdateAnalytics.ts
sanchitmalhotra126: Hmm not sure, I suspect it's been lost in our backlog. I'll create a work item.

PR #60076, opened 2024-07-30
Hot Module Replacement: don't write to disk

snickell:
- Don't write files to disk when doing `yarn start`
  - Avoid lots of .hot-update files filling the disk
  - Try to improve `yarn start` perf on WSL
- Clean `apps/build/package` before doing a `yarn start`: now that we don't overwrite this data on disk, this avoids confusion from old vs new code if you accidentally end up on :3000. This way the JS just won't be found, vs confusing random past version (last time you did `yarn build`).


snickell:
@cnbrenci wanna test if this patch makes `yarn start` faster on WSL? One thing to consider in testing is that the disk cache makes a BIG difference on Mac.

If I run `yarn start` from a cold disk cache (doing other stuff whatever) it takes about a minute. If I immediately cancel and do another `yarn start` back to back, it takes about 30s. I'm not sure how to flush the disk read cache on WSL with the complexity of having both windows and ubuntu potentially involved. My suggestion would be to run `yarn start` twice in each condition (with this branch and without) and take the second time in each case as a baseline (basically compare with disk cache to with disk cache numbers).

cnbrenci:
On staging
1:47
1:52

On seth/diskless-hmr
aborted due to warnings @ 2:06
aborted due to warnings @ 3:06

looking into (node:754390) [DEP0111] DeprecationWarning: Access to process.binding('http_parser') is deprecated.

snickell:
Well that is certainly not promising 😅

snickell:
Wait and it got slower on the second runs? Sooooo confuzzled 🤪.

Would you time on this branch and staging?
1. `HOT=0 yarn start`
2. A rebuild with `yarn start` (i.e. not initial startup time, but change a jsx or whatever and see how long that takes)

cnbrenci:
> Wait and it got slower on the second runs? Sooooo confuzzled 🤪.
> 
> Would you time on this branch and staging?
> 
> 1. `HOT=0 yarn start`
> 2. A rebuild with `yarn start` (i.e. not initial startup time, but change a jsx or whatever and see how long that takes)

staging `HOT=0 yarn start`, change jsx & save
1:56, 10 sec rebuild

seth/diskless-hmr `HOT=0 yarn start`
aborted due to warnings at 2:12

breville (approved):
Assuming this clears out the older `*hot-update*` files in `apps/build/package/js/`, this seems great.

snickell:
Yup, just re-confirmed, it cleans out the hot-update files (the whole build package, which actually makes sense so you don't have conflicting code from a build vs served from memory) when you run `yarn start`

PR #60081, opened 2024-07-30
Add aichat logging controller action

fisher-alice:
Following up on the creation of the db table `AichatEvents` by https://github.com/code-dot-org/code-dot-org/pull/60052, this PR adds a a new aichat controller action `log_aichat_event` that will log to the new table and new api endpoint. The sending of a request for an assistant response from Sagemaker will be decoupled from and the logging of aichat user events. More context in [this doc](https://docs.google.com/document/d/1OJCDc9t5DPu08gt8U52q8aa7dInUYvlzd1TvJq2PEzY/edit).

Note that we will continue to log to 'AichatSessions' during this transition until we confirm that the logging to 'AichatEvents' is successful on production. Included in follow-up work will be to remove the logging to the 'AichatSessions' table.

## After update

In order to test this controller action, I added some test front-end code in [this separate, working branch](https://github.com/code-dot-org/code-dot-org/tree/alice/log-aichat-events).

Screencast video of dev network tab showing the request/response to the new

https://github.com/user-attachments/assets/c37ed41c-b7ce-4c59-b58b-c0776b44733f

 
## Links
[jira](https://codedotorg.atlassian.net/browse/LABS-939)

## Testing story
I tested locally on aichat levels in /allthethings. See After update section above.

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/dashboard/app/controllers/aichat_controller.rb
+++ b/dashboard/app/controllers/aichat_controller.rb
@@ -1,4 +1,8 @@
 ROLES_FOR_MODEL = %w(assistant user).freeze
+AICHAT_ENDPOINTS = {
+  chat_completion: 'chat_completion',
+  log_aichat_event: 'log_aichat_event'
+}.freeze
 
 class AichatController < ApplicationController
   include AichatSagemakerHelper
@@ -12,7 +16,7 @@ class AichatController < ApplicationController
   # POST /aichat/chat_completion
   def chat_completion
     return render status: :forbidden, json: {} unless AichatSagemakerHelper.can_request_aichat_chat_completion?
-    unless has_required_params?
+    unless has_required_params?(AICHAT_ENDPOINTS[:chat_completion])
       return render status: :bad_request, json: {}
     end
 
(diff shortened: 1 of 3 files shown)


dashboard/app/controllers/aichat_controller.rb
```diff
@@ -84,16 +119,28 @@ def chat_completion
     {messages: messages}
   end
 
-  private def has_required_params?
-    begin
-      params.require([:newMessage, :aichatModelCustomizations, :aichatContext])
-    rescue ActionController::ParameterMissing
-      return false
+  private def has_required_params?(endpoint)
```
sanchitmalhotra126: nit: would it be simpler to just check the params inside each function directly since we're checking a different set of params for each? is there much value in having the shared function?

dashboard/app/controllers/aichat_controller.rb
```diff
@@ -26,6 +30,37 @@ def chat_completion
     render(status: :ok, json: response_body)
   end
 
+  # params are newAichatEvent: AichatEvent, aichatContext: {currentLevelId: number; scriptId: number; channelId: string;}
+  # POST /aichat/log_aichat_event
```
sanchitmalhotra126: nit: maybe just `/aichat/log_event` since we already have `aichat` in the path?

dashboard/app/controllers/aichat_controller.rb
```diff
@@ -26,6 +30,37 @@ def chat_completion
...
+
+    context = params[:aichatContext]
+    event = params[:newAichatEvent]
+
+    project_id = nil
+    if context[:channelId]
+      _, project_id = storage_decrypt_channel_id(context[:channelId])
+    end
+
+    logged_event = AichatEvent.create(
```
sanchitmalhotra126: Do we need to handle a save failure? I think if we use .`create!` it should throw an exception if something goes wrong.

I also see in our codebase that we use `.new` followed by `.save!` but not sure if we have a preference

sanchitmalhotra126 (approved):

dashboard/app/controllers/aichat_controller.rb
fisher-alice: What do you think of `aichat/log_chat_event`?

dashboard/app/controllers/aichat_controller.rb
fisher-alice: Oh yeah - I think I initially had the params being very similar but they've diverged so checking directly sounds good. Don't have a preference either way.

fisher-alice: I think I'll keep the helper function for `chat_completion` bc a little more complex.

dashboard/app/controllers/aichat_controller.rb
fisher-alice: Good call! I am also logging a jira ticket for myself to write dashboard unit tests for this and the fetch chat events controller action for a follow-up. https://codedotorg.atlassian.net/browse/LABS-952

dashboard/app/controllers/aichat_controller.rb
fisher-alice: Will go with `aichat/log_chat_event` since we renamed events on the front-end `ChatEvent`. Thanks!

PR #60062, opened 2024-07-29
DSCO swap: replace `ToggleSwitch` with DSCO `Toggle` in `AdvancedSettingToggles`

Erin007:
### Description

We have beautiful new Design System components. Let's use them! This PR updates the "Advanced Settings" area of the Section Edit area of the Teacher Dashboard to use the DSCO `Toggle` component. Functionality is the same as before, but the visuals are now more consistent with the rest of the page and new styles across the site. 

**Before** 

Screenshot 2024-07-29 at 4 18 32 PM

**After**  (I didn't add new toggles, I just artificially made all the possible toggles show to ensure they worked well together)

Screenshot 2024-07-29 at 3 45 07 PM

### Follow up

There are 3 threads of work related to this PR: 

1.) Efforts to deprecate legacy components in favor of using Design System Components, tracked for Toggle in [XTEAM-440](https://codedotorg.atlassian.net/browse/XTEAM-440) 
- Swap out other usages of `ToggleSwitch` for `Toggle` in SectionAccessToggle and CodeReviewGroupsStatus
- remove `ToggleSwitch` when it's fully deprecated

2.) Updating frontend tests to use React Testing Library, tracked [here](https://docs.google.com/spreadsheets/d/1OWwkmfAdoZkFc6apW2qtcjPmPaCv00jUvTCpOkZc-dc/edit?gid=1456256702#gid=1456256702)

3.) AI Tutor, tracked [CT-707](https://codedotorg.atlassian.net/browse/CT-707)
- Update info text of "Enable AI Tutor" toggle to no longer specifically reference CSA 
- Update logic for `aiTutorAvailable` in `SectionsSetUpContainer` to no longer specifically reference CSA 
- add UI tests for "Enable AI Tutor" toggle [CT-571](https://codedotorg.atlassian.net/browse/CT-571)

## PR Checklist:

- [ ] Tests provide adequate coverage
- [ ] Privacy and Security impacts have been assessed
- [ ] Code is well-commented
- [ ] New features are translatable or updates will not break translations
- [ ] Relevant documentation has been added or updated
- [ ] User impact is well-understood and desirable
- [ ] Pull Request is labeled appropriately
- [ ] Follow-up work items (including potential tech debt) are tracked and linked

PR #60089, opened 2024-07-30
DSCO swap: replace `ToggleSwitch` with DSCO `Toggle` in `CodeReviewGroupsStatusToggle`

Erin007:
### Description 

Similar to https://github.com/code-dot-org/code-dot-org/pull/60062 and https://github.com/code-dot-org/code-dot-org/pull/60086

**Before** 
Screenshot 2024-07-30 at 4 03 10 PM

**After** 
Screenshot 2024-07-30 at 4 05 59 PM

### Follow up

Up next is deleting `ToggleSwitch`.


molly-moen (commented):
Looks good, although it looks like the unit test is dependent on the old component :(

molly-moen (approved):

PR #60090, opened 2024-07-30
Codebridge Version History: Part 1

molly-moen:
Part 1 of adding version history to Codebridge. The version history components and logic lives entirely in lab2 so this component could be easily added to other lab2 labs.

This PR covers adding basic UI and functionality for version history for just the "restore" option, not view. It includes:
- Updating the project manager, sources store, and sources api to handle restoring a specific version. 
- New version history button that opens a dropdown of versions
- Replacing the start over button in codebridge with the version history button.

The most interesting part of this PR is the project manager updates. When we restore an old version in our legacy labs, we call the [restore](https://github.com/code-dot-org/code-dot-org/blob/b0c54c23af8515cd2770cd14f7a60ef2e64d1b18/dashboard/legacy/middleware/files_api.rb#L696) api, then reload the page to get the updated version. We obviously don't want to force a page reload for this in lab2. `restore` does not return the newly-restored version. To avoid updating the backend, I had the project manager restore do the following:
- Ensure any previously enqueued saves get saved
- Call restore
- Load the sources again, which will be the newly-restored version. 
We use `restore` rather than loading and saving the previous version because we have [logic on the backend](https://github.com/code-dot-org/code-dot-org/blob/b0c54c23af8515cd2770cd14f7a60ef2e64d1b18/dashboard/legacy/middleware/files_api.rb#L460) that checks that we are saving the most recent version of the code. 

As I was doing this I ran into a couple issues:

We did not ever clear the `sourcesToSave` variable. This was fine when the only way new changes would come in was via `save`. However, now we can get an update via `restore`, which does not require a `save` as `restore` handles the S3 update. This meant that even if we had already saved whatever was in `sourcesToSave`, if we didn't make a change to the code after restoring, when we try to flush saves on page change we would see an old `sourcesToSave` that is different from the newly-restored code (stored in `lastSave`), and would overwrite the newly-restored code. To fix this I have us clear `sourcesToSave` after save, or if we try to save and find the sources have not changed.

I also found that our logic to decide if we should create a new version is broken. We always create a new version rather than creating a new one every 15 minutes. This wasn't noticeable before because we never listed versions in lab2. I've filed a [follow-up task](https://codedotorg.atlassian.net/browse/CT-711) for this.

### Restore a specific Version
https://github.com/user-attachments/assets/3e8359c2-c9a2-4fd0-b450-4717c0973508

### Start Over
https://github.com/user-attachments/assets/c19ccbeb-a484-4765-9cc6-bbb96f0815d8

## Links

- jira ticket: [CT-453](https://codedotorg.atlassian.net/browse/CT-453)

## Testing story
Tested locally and added some additional project manager unit tests. I verified that Music Lab (also uses the project manager) is unchanged.

## Follow-up work
- General polish task: error handling, a confirmation on start over, better load behavior: [CT-712](https://codedotorg.atlassian.net/browse/CT-712)
-  View a previous version: [CT-713](https://codedotorg.atlassian.net/browse/CT-713). We need to decide how exactly we will do this (either a diff view, open in a new tab, some third option), so this ticket covers prototyping and making a decision on how we will implement view.

## PR Checklist:

- [ ] Code is well-commented


apps/src/util/hooks/useOutsideClick.ts
```diff
@@ -0,0 +1,25 @@
...
+
+    document.addEventListener('click', handleClick);
+
+    return () => {
+      document.removeEventListener('click', handleClick);
+    };
+  }, [ref, callback]);
+
+  return ref;
+}
```
thomasoniii, line 25: Nice hook!

thomasoniii (approved):
lgtm 🚀 

Only minor note would be that the callbacks in `VersionHistoryDropdown` should probably be `useCallback`, but realistically I'm not gonna stress about it.

apps/src/lab2/projects/ProjectManager.ts
```diff
@@ -248,6 +244,11 @@ export default class ProjectManager {
     this.publishHelper(false);
   }
 
+  async getVersionList() {
+    // TODO: Error handling
```
sanchitmalhotra126, line 250: Just checking, do we have a JIRA for this?

apps/src/lab2/projects/ProjectManager.ts
```diff
@@ -489,6 +493,35 @@ export default class ProjectManager {
     }
   }
 
+  private async loadHelper(versionId?: string, isPreview?: boolean) {
```
sanchitmalhotra126: nit: `loadSources` since it returns sources?

apps/src/lab2/projects/ProjectManager.ts
```diff
@@ -489,6 +493,35 @@ export default class ProjectManager {
...
+    let sources: ProjectSources | undefined;
+    try {
+      sources = await this.sourcesStore.load(this.channelId, versionId);
+      // If we are previewing a project, we don't want to save the source.
+      if (!isPreview) {
+        this.lastSource = JSON.stringify(sources);
+      }
```
sanchitmalhotra126: Checking my understanding, does this assume that the lab will save the sources if `isPreview` is false? Doesn't look like the project manager actually performs a save either way right?

sanchitmalhotra126: Also, are there any usages of `loadHelper` with `isPreview = true` or would that be in a different change?

apps/src/lab2/projects/ProjectManager.ts
```diff
@@ -113,6 +92,23 @@ export default class ProjectManager {
     return {sources, channel};
   }
 
+  async restore(versionId: string): Promise<ProjectSources | undefined> {
```
sanchitmalhotra126: nit: mind adding a doc explaining what this function does? mostly so it's clear that it's performing a project overwrite (not just retrieving)

Also would `restoreSources` or `revertSources` be a bit clearer?

apps/src/lab2/views/components/versionHistory/VersionHistoryButton.tsx
```diff
@@ -0,0 +1,75 @@
...
+  const [versionList, setVersionList] = useState<ProjectVersion[]>([]);
+
+  const isReadOnly = useAppSelector(isReadOnlyWorkspace);
+  const toggleVersionHistory = (
+    e: React.MouseEvent<HTMLButtonElement> | React.MouseEvent<HTMLAnchorElement>
+  ) => {
+    setVersionList([]);
+    const projectManager = Lab2Registry.getInstance().getProjectManager();
+    if (!isVersionHistoryOpen && projectManager) {
+      projectManager.getVersionList().then(versionList => {
```
sanchitmalhotra126, line 41: Optional - could potentially optimize this to store the version list in state after its loaded once, and only clear out when the project/level changes?

apps/src/lab2/views/components/versionHistory/VersionHistoryDropdown.tsx
```diff
@@ -0,0 +1,111 @@
...
+        } else {
+          // TODO: handle no source
+          console.error('no source found!');
+        }
+        closeDropdown();
+      });
+    }
+  };
+
+  const startOver = () => {
```
sanchitmalhotra126: Should this use the existing Start Over dialog?

apps/src/util/hooks/useOutsideClick.ts
sanchitmalhotra126: +1!

sanchitmalhotra126 (commented):
Awesome! Cool to see all these features coming into Lab2! A few comments/questions

Minor general UI thing - wonder if it might be nice to include the blackout/fade or some transition when restoring version history so it's obvious that code has changed? Not blocking though, and I'm sure we'll iterate on design as we keep adding features

apps/src/lab2/projects/ProjectManager.ts
molly-moen: yep! https://codedotorg.atlassian.net/browse/CT-712

apps/src/lab2/projects/ProjectManager.ts
molly-moen: Yeah I added this because we'll need it for the view feature but it isn't used anywhere yet.
I will make this clearer, we aren't ever saving the code. When all is said and done there will be 3 cases for this function:
1. We are loading the most recent sources on level load (what existed before). 
2. We just restored a previous version and are now loading the most recent sources to send back to the lab.
3. We are previewing a specific version (haven't done this end to end yet).

For 1 and 2, the sources have already been saved and we are sending them to the lab. When the lab calls `save`, we want to compare the new code to what we loaded so we don't trigger an unnecessary save (that's the purpose of `lastSource`. For 3 we don't want the project manager to track this code at all, as it's for a preview state.

apps/src/lab2/projects/ProjectManager.ts
molly-moen: Will update!

apps/src/lab2/views/components/versionHistory/VersionHistoryButton.tsx
molly-moen: The version list may change during a given session, so I think it makes sense to get the most recent list when you open the dropdown.

apps/src/lab2/views/components/versionHistory/VersionHistoryDropdown.tsx
molly-moen: Good idea, I'll add that to the follow up ticket.

apps/src/lab2/projects/ProjectManager.ts
molly-moen: good call

molly-moen:
> Minor general UI thing - wonder if it might be nice to include the blackout/fade or some transition when restoring version history so it's obvious that code has changed? Not blocking though, and I'm sure we'll iterate on design as we keep adding features

@sanchitmalhotra126 I added that to the follow up task!

molly-moen:
@thomasoniii I added `useCallback`, lmk if I missed any you were thinking were necessary!

apps/src/lab2/projects/ProjectManager.ts
molly-moen: I updated the comments here, lmk if this makes sense!

apps/src/lab2/projects/ProjectManager.ts
```diff
@@ -489,6 +495,44 @@ export default class ProjectManager {
     }
   }
 
+  /**
+   * Load the sources for this project. If a versionId is provided, load that version, otherwise
+   * load the latest version.
+   * @param versionId Optional version id to load. If not provided, the latest version is loaded.
+   * @param isPreview Optional boolean to indicate if we are previewing a project. If we are previewing,
```
sanchitmalhotra126, line 502: Thanks for the comments! This makes sense - just as an alternate idea, we could also maybe have two separate functions, one that just loads the project for a given version and another that calls load and saves it to `lastSources`? Something like

```
private async loadSources(versionId?: string): Promise<ProjectSources | undefined> {
   ... existing load code, without storing in this.lastSource
}

private async loadAndStoreSources(versionId?: string) {
   const sources = await loadSources(versionId);
   this.lastSource = JSON.stringify(sources);
   return sources;
}
```

Then `load` and `restore` can call `loadAndStoreSources` while any previewing APIs can just call `loadSources()`? Might be overkill, but just as a thought to reduce ambiguity. This works too though, so not blocking

sanchitmalhotra126 (approved):
LGTM!

apps/src/lab2/projects/ProjectManager.ts
molly-moen: Good idea! I might add that in the follow-up to do view-only, since that will be the use case for just loading sources.

PR #60097, opened 2024-07-31
log thumbs-down click without waiting for survey submission

davidsbailey:
Finishes https://codedotorg.atlassian.net/browse/AITT-410. https://github.com/code-dot-org/code-dot-org/pull/60044  started logging thumbs up/down clicks. This PR makes sure we log those events as soon as the thumbs down is clicked, without waiting for the survey to be submitted.

This PR may be easiest to review in the following chunks:
* code changes: https://github.com/code-dot-org/code-dot-org/pull/60097/files/fe8c584766c3ef2ab348729b027d1b4f0399c96f
* refactoring: https://github.com/code-dot-org/code-dot-org/pull/60097/files/fe8c584766c3ef2ab348729b027d1b4f0399c96f..ce7301db95f6789150f3d6b7908bfc753e52cb7d
* new tests: https://github.com/code-dot-org/code-dot-org/pull/60097/files/ce7301db95f6789150f3d6b7908bfc753e52cb7d..09f005b1eea504b04f8ba846cda2db2df2b5b3a1

### screenshots

https://github.com/user-attachments/assets/9955f2da-f2dc-4ff7-a4d4-b01ff98381f2

## Testing story
* new unit test coverage
* manual testing shown in screenshots


apps/test/unit/templates/feedback/LevelFeedbackEntryTest.jsx
```diff
@@ -9,7 +9,7 @@ import {UnlocalizedTimeAgo as TimeAgo} from '@cdo/apps/templates/TimeAgo';
 import color from '@cdo/apps/util/color';
 import i18n from '@cdo/locale';
 
-const DEFAULT_FEEDBACK = {
+const NO_FEEDBACK = {
```
wilkie: Is this name change necessary? If so `NO_FEEDBACK` is confusing since it looks like it has feedback.

davidsbailey: I can switch it back to DEFAULT_FEEDBACK if that's clearer! Is there other wording that would help indicate not just the default, but the absence of feedback?

apps/test/unit/templates/rubrics/AiAssessmentBoxTest.jsx
```diff
@@ -15,11 +15,7 @@ import {RubricUnderstandingLevels} from '@cdo/generated-scripts/sharedConstants'
 import i18n from '@cdo/locale';
 
 async function wait() {
-  for (let _ = 0; _ < 10; _++) {
-    await act(async () => {
-      await Promise.resolve();
-    });
-  }
+  await act(async () => Promise.resolve());
```
wilkie: You may have to wait more than once, in this case, if you're trying to get through a few `await` lines. The `act` will wait until Promises resolve... but basically, they might resolve into another Promise. Firefox for some reason generally needs one extra `act` over Chrome (although admittedly I think I'm the only one that runs JS tests on Firefox).

I wonder how much Jest changes this if it isn't targeting a specific browser implementation... haven't tried it thoroughly since then. 🤷

apps/test/unit/templates/feedback/LevelFeedbackEntryTest.jsx
wilkie: I'm confused! Isn't there feedback here?

Also, I didn't quite understand the connection between this change and the overall update you're doing.

apps/src/templates/rubrics/AiAssessmentBox.jsx
```diff
@@ -136,6 +136,8 @@ export default function AiAssessmentBox({
     );
   };
 
+  const [aiFeedbackId, setAiFeedbackId] = useState(null);
+
   const {aiFeedback, setAiFeedback} = useContext(AiAssessmentFeedbackContext);
```
wilkie: If you wanted, you could place the `aiFeedbackId` and `setAiFeedbackId` into the feedback context and then pull it out when you need it. This does mean defining it a level up the tree alongside the feedback state itself.

So, you could add that:

```jsx
const [aiFeedbackId, setAiFeedbackId] = useState(null);
```

to `LearningGoals.jsx` (and I guess :( `LearningGoal.jsx` too)

And then in `LearningGoals.jsx` where the Provider is attached, you then:

```jsx
<AiAssessmentFeedbackContext.Provider
  value={{aiFeedback, setAiFeedback, aiFeedbackId, setAiFeedbackId}}
>
```

You should be able to do the normal:

```jsx
const {aiFeedback, setAiFeedback} = useContext(AiAssessmentFeedbackContext);
```

When you don't need the other things and then, at the places where you need it, you can use:

```jsx
const {aiFeedback, setAiFeedback, aiFeedbackId, setAiFeedbackId} = useContext(AiAssessmentFeedbackContext);
```

It doesn't really matter, but I do appreciate placing related things into the Context together. Though the network calls make me want to use Redux or a React Reducer... but I don't think it is _that_ necessary.

What do you think?

apps/test/unit/templates/feedback/LevelFeedbackEntryTest.jsx
davidsbailey: > I'm confused! Isn't there feedback here?

no, prior to thumbs up or thumbs down being clicked, we are in a state of not having any feedback. LearningGoalAiEvaluation is the evaluation provided by AI. LearningGoalAiEvaluationFeedback is what's meant by "feedback" here, and represents the user's opinion of the AI evaluation, as indicated by thumbs up / thumbs down click, plus any additional info from the checkboxes + "other" textbox which show up when you click thumbs down. 

thus when the page loads and AI evaluation is present but the user has not yet clicked and thumbs up / down button, we are in a state of "no feedback". not sure if that helps clear things up? 

> Also, I didn't quite understand the connection between this change and the overall update you're doing.

we had local definitions of `thumbsupval`, `thumbsdownval`, and `-1` scattered throughout the code, so I did a refactoring step to centralize these all as constants. Please see links in the PR description to see code, test and refactoring diffs separately.

apps/test/unit/templates/rubrics/AiAssessmentBoxTest.jsx
davidsbailey: I agree that this might not work in all cases, but since this is limited to this one file, I thought it might be ok? If we ran into a problem when adding another test, the test writer could always work around it by calling `wait()` a second time, perhaps commenting on what each one is waiting for. this seems better to me than blindly waiting 10 times. this also reads more simply: "wait one network request, then update the DOM". I suppose i could add that in a comment!

I don't think we'll have to worry about other browsers, now that we are running the tests in Jest.

apps/src/templates/rubrics/AiAssessmentBox.jsx
davidsbailey: good question! I looked into this a bit because I actually tried to move all that was in AiAssessmentFeedbackContext down into AiAssessmentBox, before realizing that LearningGoals needs to be able to reset it when we switch learning goals. since the higher component (LearningGoals) needs to access this state, AFAIK there's no way to move the state further down, so keeping the state up high (and using a context) makes sense.

for why I used state in `AiAssessmentBox`, my thought was to keep the state in the lowest common ancestor of all the components that need it. However, now I am wondering if this state actually should also be reset when we switch learning goals? I don't currently see a problem, but I think I am just getting away with it because setting the feedback back to `NO_FEEDBACK` is enough to hide the feedback form. 

Conclusion: since I may be hiding a latent bug, and we've already gone through the trouble of extracting `AiAssessmentFeedbackContext`, I think it makes sense to take your suggestion and move this to `AiAssessmentFeedbackContext`. 👍

apps/test/unit/templates/rubrics/AiAssessmentBoxTest.jsx
wilkie: Agreed!

apps/test/unit/templates/feedback/LevelFeedbackEntryTest.jsx
wilkie: I understand the changes and refactoring you have done to the AI feedback components. I'm not sure about this particular change to the LevelFeedbackEntry component test file. This component seemed somewhat distant from the AI assessment logic... although we _do kinda_ touch the keep working data.

I was presuming that this was student progress feedback, and from my understanding it has "`Great work`" as the student feedback, so it was a little hard to understand why this was `NO_FEEDBACK` and felt like maybe an accidental find/replace.

davidsbailey: oh jeez, I was totally unaware that i had touched an unrelated file. you are correct, I will revert!

wilkie: No worries! Easy peasy lemon squeezy.

wilkie:
This is looking really great! I really like and appreciate the refactoring on the AI feedback components. I don't mind what you decide about where to put that feedback ID state, but I think you might be right as I look at it that this id might need to be nullified when the learning goal is switched.

Either way, I like everything else and will be happy to approve it after whatever updates you add!

wilkie (approved):

PR #60110, opened 2024-07-31
[Aichat] Log chat events on the frontend

fisher-alice:
This is a follow-up to https://github.com/code-dot-org/code-dot-org/pull/60081 which added an aichat controller action for logging chat events.

This PR adds `postLogChatEvent` in the renamed `aichatApi.ts` that makes a call to the new `aichat` controller action to log chat events. This function is called by a singleton instance of  the `ChatEventLogger`  that helps manage the queue of chat events to be logged to the backend. Thanks to @sanchitmalhotra126 for helping me think through the design of this class.
This logger ensures that chat events are logged in sequence and multiple chat events are not sent to be logged at the same time. 

Ths following `ChatEvent` are logged:
- user and assistant `ChatMessage` (successful and errored)
- all `ModelUpdate`
- save failure `Notification` due to profanity (other save failure errors are NOT logged)
- model customizations set to default setting `Notification`
- `ChatEvent`: 'LOAD_LEVEL' and 'CLEAR_CHAT'.

## After update

Screencast video of the dev network tab while user is submitting a chat message, making model customizations, clearing chat history, changing levels, and then reloading the browser.

https://github.com/user-attachments/assets/d731595d-ae81-4bee-bf2c-954fbb45a5ef

Screencast video showing log request for a user messaged marked as profanity and the saving of a system prompt that includes profanity.

https://github.com/user-attachments/assets/6cd45ad9-9fb0-4a9f-94c1-7d61a73a2a1a

## Links
[jira](https://codedotorg.atlassian.net/browse/LABS-940)

## Testing story
I tested locally in aichat levels at /allthethings and added unit tests for `ChatEventLogger`.

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/aichat/chatEventLogger.ts
+++ b/apps/src/aichat/chatEventLogger.ts
@@ -13,7 +13,7 @@ export default class ChatEventLogger {
 
   private static instance: ChatEventLogger;
 
-  private constructor() {
+  constructor() {
     this.queue = [];
     this.sendingInProgress = false;
   }
--- a/apps/src/aichat/chatEventLogger.ts
+++ b/apps/src/aichat/chatEventLogger.ts
@@ -42,11 +42,9 @@ export default class ChatEventLogger {
       const loggerPayload = this.queue.shift(); // Remove the first element from the queue.
       if (loggerPayload) {
         const {chatEvent, aichatContext} = loggerPayload;
-        let logResponse;
         this.sendingInProgress = true;
         try {
-          logResponse = await postLogChatEvent(chatEvent, aichatContext);
-          console.log('logChatEventResponse', logResponse);
+          await postLogChatEvent(chatEvent, aichatContext);
           this.sendingInProgress = false;
         } catch (error) {
           Lab2Registry.getInstance()
--- a/apps/src/aichat/chatEventLogger.ts
+++ b/apps/src/aichat/chatEventLogger.ts
@@ -1,0 +1,64 @@
+import Lab2Registry from '@cdo/apps/lab2/Lab2Registry';
+
+import {postLogChatEvent} from './aichatCompletionApi';
+import {ChatEvent, AichatContext} from './types';
+
+interface ChatEventLoggerPayload {
+  chatEvent: ChatEvent;
+  aichatContext: AichatContext;
+}
+export default class ChatEventLogger {
+  private chatEventsQueue: ChatEventLoggerPayload[];
+  private sendingChatEventInProgress: boolean;
+
+  private static _instance: ChatEventLogger;
+
+  constructor() {
+    this.chatEventsQueue = [];
+    this.sendingChatEventInProgress = false;
+  }
+
+  public static getInstance(): ChatEventLogger {
+    if (ChatEventLogger._instance === undefined) {
+      ChatEventLogger.create();
+    }
+    return ChatEventLogger._instance;
+  }
+
+  public static create() {
+    ChatEventLogger._instance = new ChatEventLogger();
+  }
+
+  logChatEvent(chatEvent: ChatEvent, aichatContext: AichatContext) {
+    this.chatEventsQueue.push({chatEvent, aichatContext});
+    this.sendChatEvent();
+  }
+
+  async sendChatEvent() {
+    // Send aichat events to the server to be logged.
+    while (
+      this.chatEventsQueue.length > 0 &&
+      !this.sendingChatEventInProgress
+    ) {
+      const chatEventLoggerPayload = this.chatEventsQueue.shift(); // Remove the first element from the queue.
+      if (chatEventLoggerPayload) {
+        const {chatEvent, aichatContext} = chatEventLoggerPayload;
+        let logChatEventResponse;
+        this.sendingChatEventInProgress = true;
+        try {
+          logChatEventResponse = await postLogChatEvent(
+            chatEvent,
+            aichatContext
+          );
+          console.log('logChatEventResponse', logChatEventResponse);
+          this.sendingChatEventInProgress = false;
+        } catch (error) {
+          Lab2Registry.getInstance()
+            .getMetricsReporter()
+            .logError('Error in aichat event logging request', error as Error);
+          return;
+        }
+      }
+    }
+  }
+}
--- a/apps/src/aichat/chatEventLogger.ts
+++ b/apps/src/aichat/chatEventLogger.ts
@@ -1,6 +1,6 @@
 import Lab2Registry from '@cdo/apps/lab2/Lab2Registry';
 
-import {postLogChatEvent} from './aichatCompletionApi';
+import {postLogChatEvent} from './aichatApi';
 import {ChatEvent, AichatContext} from './types';
 
 interface ChatEventLoggerPayload {
--- a/apps/src/aichat/chatEventLogger.ts
+++ b/apps/src/aichat/chatEventLogger.ts
@@ -13,7 +13,7 @@ export default class ChatEventLogger {
 
   private static _instance: ChatEventLogger;
 
-  constructor() {
+  private constructor() {
     this.chatEventsQueue = [];
     this.sendingChatEventInProgress = false;
   }
--- a/apps/src/aichat/chatEventLogger.ts
+++ b/apps/src/aichat/chatEventLogger.ts
@@ -25,11 +25,11 @@ export default class ChatEventLogger {
     return ChatEventLogger.instance;
   }
 
-  public static create() {
+  public static create(): void {
     ChatEventLogger.instance = new ChatEventLogger();
   }
 
-  logChatEvent(chatEvent: ChatEvent, aichatContext: AichatContext) {
+  public logChatEvent(chatEvent: ChatEvent, aichatContext: AichatContext) {
     this.queue.push({chatEvent, aichatContext});
     if (!this.sendingInProgress) {
       this.sendChatEvent();
@@ -50,7 +50,6 @@ export default class ChatEventLogger {
           Lab2Registry.getInstance()
             .getMetricsReporter()
             .logError('Error in aichat event logging request', error as Error);
-          return;
         }
       }
     }
--- a/apps/src/aichat/chatEventLogger.ts
+++ b/apps/src/aichat/chatEventLogger.ts
@@ -3,55 +3,51 @@ import Lab2Registry from '@cdo/apps/lab2/Lab2Registry';
 import {postLogChatEvent} from './aichatApi';
 import {ChatEvent, AichatContext} from './types';
 
-interface ChatEventLoggerPayload {
+interface LoggerPayload {
   chatEvent: ChatEvent;
   aichatContext: AichatContext;
 }
 export default class ChatEventLogger {
-  private chatEventsQueue: ChatEventLoggerPayload[];
-  private sendingChatEventInProgress: boolean;
+  private queue: LoggerPayload[];
+  private sendingInProgress: boolean;
 
-  private static _instance: ChatEventLogger;
+  private static instance: ChatEventLogger;
 
   private constructor() {
-    this.chatEventsQueue = [];
-    this.sendingChatEventInProgress = false;
+    this.queue = [];
+    this.sendingInProgress = false;
   }
 
   public static getInstance(): ChatEventLogger {
-    if (ChatEventLogger._instance === undefined) {
+    if (ChatEventLogger.instance === undefined) {
       ChatEventLogger.create();
     }
-    return ChatEventLogger._instance;
+    return ChatEventLogger.instance;
   }
 
   public static create() {
-    ChatEventLogger._instance = new ChatEventLogger();
+    ChatEventLogger.instance = new ChatEventLogger();
   }
 
   logChatEvent(chatEvent: ChatEvent, aichatContext: AichatContext) {
-    this.chatEventsQueue.push({chatEvent, aichatContext});
-    this.sendChatEvent();
+    this.queue.push({chatEvent, aichatContext});
+    if (!this.sendingInProgress) {
+      this.sendChatEvent();
+    }
   }
 
-  async sendChatEvent() {
+  private async sendChatEvent() {
     // Send aichat events to the server to be logged.
-    while (
-      this.chatEventsQueue.length > 0 &&
-      !this.sendingChatEventInProgress
-    ) {
-      const chatEventLoggerPayload = this.chatEventsQueue.shift(); // Remove the first element from the queue.
-      if (chatEventLoggerPayload) {
-        const {chatEvent, aichatContext} = chatEventLoggerPayload;
-        let logChatEventResponse;
-        this.sendingChatEventInProgress = true;
+    while (this.queue.length > 0) {
+      const loggerPayload = this.queue.shift(); // Remove the first element from the queue.
+      if (loggerPayload) {
+        const {chatEvent, aichatContext} = loggerPayload;
+        let logResponse;
+        this.sendingInProgress = true;
         try {
-          logChatEventResponse = await postLogChatEvent(
-            chatEvent,
-            aichatContext
-          );
-          console.log('logChatEventResponse', logChatEventResponse);
-          this.sendingChatEventInProgress = false;
+          logResponse = await postLogChatEvent(chatEvent, aichatContext);
+          console.log('logChatEventResponse', logResponse);
+          this.sendingInProgress = false;
         } catch (error) {
           Lab2Registry.getInstance()
             .getMetricsReporter()
--- a/apps/src/aichat/redux/aichatRedux.ts
+++ b/apps/src/aichat/redux/aichatRedux.ts
@@ -321,12 +321,53 @@ const dispatchSaveFailNotification = (
   dispatch(endSave());
 };
 
+// This thunk adds a chat event to chatEventsCurrent (displayed in current chat workspace)
+// if hideForParticipants != true and then logs the event to the backend for all chat events
+// except notifications with includeInHistory != true.
+export const addChatEvent =
+  <T extends ChatEvent>(chatEvent: T) =>
+  (dispatch: AppDispatch, getState: () => RootState) => {
+    if (!chatEvent.hideForParticipants) {
+      dispatch(addEventToChatEventsCurrent(chatEvent));
+    }
+    // Log chat event to backend.
+    const state = getState() as RootState;
+    const aichatContext: AichatContext = {
+      currentLevelId: parseInt(state.progress.currentLevelId || ''),
+      scriptId: state.progress.scriptId,
+      channelId: state.lab.channel?.id,
+    };
+    // Other than notifications that are not included in chat history (save errors not due to profanity),
+    // log the chat event to backend.
+    if (
+      !isNotification(chatEvent) ||
+      (isNotification(chatEvent) && chatEvent.includeInChatHistory)
+    ) {
+      // If a model update, log only the updated value for temperature and selected model id.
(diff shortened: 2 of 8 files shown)


apps/src/aichat/types.ts
```diff
@@ -13,6 +13,8 @@ export type AichatInteractionStatusValue = string;
 export interface ChatEvent {
   // UTC timestamp in milliseconds
   timestamp: number;
+  // This field is optional but when it is defined, it must be set to `true`.
```
fisher-alice, line 16: Added comment - addresses https://github.com/code-dot-org/code-dot-org/pull/60053#discussion_r1698884071 @sanchitmalhotra126

cnbrenci: Can you include the why? (to avoid bloating all ChatEvents since only a small subset will have this field set to true)

fisher-alice: I think I do in the line below.

cnbrenci: OH sorry! I didn't see it under the comment dialog 🤦🏻‍♀️

fisher-alice: Lmk if you'd like me to add more details. Thanks!

cnbrenci: No that is perfect

apps/src/aichat/chatEventLogger.ts
```diff
@@ -0,0 +1,64 @@
...
+      if (chatEventLoggerPayload) {
+        const {chatEvent, aichatContext} = chatEventLoggerPayload;
+        let logChatEventResponse;
+        this.sendingChatEventInProgress = true;
+        try {
+          logChatEventResponse = await postLogChatEvent(
+            chatEvent,
+            aichatContext
+          );
+          console.log('logChatEventResponse', logChatEventResponse);
```
fisher-alice: Logging response for now.

apps/src/aichat/chatEventLogger.ts
```diff
@@ -0,0 +1,64 @@
...
+  public static create() {
+    ChatEventLogger._instance = new ChatEventLogger();
+  }
+
+  logChatEvent(chatEvent: ChatEvent, aichatContext: AichatContext) {
+    this.chatEventsQueue.push({chatEvent, aichatContext});
+    this.sendChatEvent();
+  }
+
+  async sendChatEvent() {
```
sanchitmalhotra126: nit: this can be `private`, right?

apps/src/aichat/chatEventLogger.ts
```diff
@@ -0,0 +1,64 @@
...
+    return ChatEventLogger._instance;
+  }
+
+  public static create() {
+    ChatEventLogger._instance = new ChatEventLogger();
+  }
+
+  logChatEvent(chatEvent: ChatEvent, aichatContext: AichatContext) {
+    this.chatEventsQueue.push({chatEvent, aichatContext});
+    this.sendChatEvent();
```
sanchitmalhotra126: I think we would want to check `this.sendingChatEventInProgress` here, if i'm following? The idea is that if the sendChatEvent loop is not running, we should start it, but if it is running, we just need to add the event to the queue and it should get picked up on the next iteration.

apps/src/aichat/chatEventLogger.ts
```diff
@@ -0,0 +1,64 @@
+import Lab2Registry from '@cdo/apps/lab2/Lab2Registry';
+
+import {postLogChatEvent} from './aichatApi';
+import {ChatEvent, AichatContext} from './types';
+
+interface ChatEventLoggerPayload {
```
sanchitmalhotra126: nit: can this just be `LoggerPayload`?

apps/src/aichat/chatEventLogger.ts
```diff
@@ -0,0 +1,64 @@
...
+  chatEvent: ChatEvent;
+  aichatContext: AichatContext;
+}
+export default class ChatEventLogger {
+  private chatEventsQueue: ChatEventLoggerPayload[];
+  private sendingChatEventInProgress: boolean;
```
sanchitmalhotra126: nit: similarly here, maybe just `queue` and `sendInProgress` for simplicity?

apps/src/aichat/chatEventLogger.ts
```diff
@@ -0,0 +1,64 @@
+import Lab2Registry from '@cdo/apps/lab2/Lab2Registry';
+
+import {postLogChatEvent} from './aichatApi';
+import {ChatEvent, AichatContext} from './types';
+
+interface ChatEventLoggerPayload {
+  chatEvent: ChatEvent;
+  aichatContext: AichatContext;
+}
+export default class ChatEventLogger {
```
sanchitmalhotra126, line 10: I know we haven't added any unit tests for AI Chat yet (😅) but this might a good first one. Since this deals with some tricky async logic and timing, it would be good to ensure it's working as intended

apps/src/aichat/redux/aichatRedux.ts
```diff
@@ -209,13 +211,24 @@ export const onSaveComplete =
...
+          ? updatedValue
+          : 'N/A';
       dispatch(
-        addModelUpdate({
-          id: getNewMessageId(),
-          updatedField: typedProperty,
-          updatedValue: currentAiCustomizations[typedProperty],
-          timestamp: Date.now(),
-        })
+        logChatEvent({
```
sanchitmalhotra126: I think it might be good to reorganize this a little bit. We could have a single thunk that all events are routed through, which logs the event, and adds to the current set of events if necessary. That way we're not at risk of forgetting to call `logChatEvent` in every place that we also add a chat event to the window. For example:

```
const addEvent = <T extends ChatEvent>(event: T) =>
  (dispatch: AppDispatch, getState: () => RootState) => {
    // Log all events except for notifications that are not included in history (i.e. non-profanity save errors).
    if (
      !isNotification(event) ||
      (isNotification(event) && event.includeInHistory) // Let's assume we add some `includeInHistory` flag to Notification, so we don't need to check the profanity message string directly
    ) {
      const {progress, lab} = getState();
      const aichatContext: AichatContext = {
        currentLevelId: parseInt(progress.currentLevelId || ''),
        scriptId: progress.scriptId,
        channelId: lab.channel?.id,
      };
      ChatEventLogger.getInstance().logChatEvent(event, aichatContext);
    }

    // Add to current chat window if visible.
    if (!event.hideForParticipants) {
      dispatch(addEventToCurrentChat(event)); // addEventToCurrentChat (can be named whatever) would be a new redux action to replace addChatMessage/addModelUpdate/addNotification, that adds the event to "chatEventsCurrent"
    }
  };
```

Then, we can replace all usages of `addChatMessage`, `addModelUpdate`, and `addNotification` with simply `addEvent`. `addChatMessage` and `addModelUpdate` are only used internally in this file as it is; `addNotification` is currently used externally only for the reset model notification, which we could just create a dedicated thunk for

apps/src/aichat/chatEventLogger.ts
fisher-alice: Oh yeah - that makes sense - will update. Thanks!

apps/src/aichat/redux/aichatRedux.ts
fisher-alice: Nice idea! Updated. Don't think I need `addNotification` as can use `addChatEvent` for the reset model notification also. Pushed the updates and tested locally and confirmed logging is as expected. Lmk what you think. Thanks!

apps/src/aichat/chatEventLogger.ts
fisher-alice: If you're okay with this, I'm going to log this as a separate jira ticket and can do as a fast follow-up. https://codedotorg.atlassian.net/browse/LABS-958

fisher-alice: Unit tests added in https://github.com/code-dot-org/code-dot-org/pull/60191.

apps/src/aichat/chatEventLogger.ts
```diff
@@ -0,0 +1,60 @@
...
+    // Send aichat events to the server to be logged.
+    while (this.queue.length > 0) {
+      const loggerPayload = this.queue.shift(); // Remove the first element from the queue.
+      if (loggerPayload) {
+        const {chatEvent, aichatContext} = loggerPayload;
+        let logResponse;
+        this.sendingInProgress = true;
+        try {
+          logResponse = await postLogChatEvent(chatEvent, aichatContext);
+          console.log('logChatEventResponse', logResponse);
```
sanchitmalhotra126: nit: remove?

apps/src/aichat/redux/aichatRedux.ts
```diff
@@ -288,7 +287,12 @@ export const onSaveFail =
           if (body?.details?.profaneWords?.length > 0 && flaggedProperties) {
             errorMessage = `Profanity detected in the ${flaggedProperties} and cannot be updated. Please try again.`;
           }
-          dispatchSaveFailNotification(dispatch, errorMessage);
+          const includeInChatHistory = flaggedProperties ? true : false;
+          dispatchSaveFailNotification(
+            dispatch,
+            errorMessage,
+            includeInChatHistory
```
sanchitmalhotra126: nit: `!!flaggedProperties`?

apps/src/aichat/redux/aichatRedux.ts
```diff
@@ -362,15 +368,16 @@ export const submitChatContents = createAsyncThunk(
         .logError('Error in aichat completion request', error as Error);
 
       thunkAPI.dispatch(clearChatMessagePending());
-      thunkAPI.dispatch(addChatMessage({...newMessage, status: Status.ERROR}));
+      const erroredUserChatMessage = {...newMessage, status: Status.ERROR};
+      thunkAPI.dispatch(addChatEvent(erroredUserChatMessage));
```
sanchitmalhotra126: nit: any reason to not keep this inlined?

apps/src/aichat/redux/aichatRedux.ts
```diff
@@ -395,23 +402,59 @@ export const submitChatContents = createAsyncThunk(
     }
 
     thunkAPI.dispatch(clearChatMessagePending());
-    chatApiResponse.messages.forEach(message =>
-      thunkAPI.dispatch(addChatMessage({...message, timestamp: Date.now()}))
-    );
+    chatApiResponse.messages.forEach(message => {
+      const successfulChatMessage = {...message, timestamp: Date.now()};
+      thunkAPI.dispatch(addChatEvent(successfulChatMessage));
```
sanchitmalhotra126: nit: same here, keep inlined?

apps/src/aichat/redux/aichatRedux.ts
```diff
@@ -395,23 +402,59 @@ export const submitChatContents = createAsyncThunk(
...
   initialState,
   reducers: {
-    addChatMessage: (state, action: PayloadAction<ChatMessage>) => {
-      state.chatEventsCurrent.push(action.payload);
-    },
-    addModelUpdate: (state, action: PayloadAction<ModelUpdate>) => {
-      state.chatEventsCurrent.push(action.payload);
-    },
-    addNotification: (state, action: PayloadAction<Notification>) => {
+    addEventToChatEventsCurrent: (state, action: PayloadAction<ChatEvent>) => {
```
sanchitmalhotra126, line 450: nice!

apps/src/aichat/types.ts
```diff
@@ -33,6 +35,10 @@ export interface Notification extends ChatEvent {
   id: number;
   text: string;
   notificationType: 'error' | 'success';
+  // This field is optional but when it is defined, it must be set to `true`.
+  // This allows the notification to not be included in chat history as default
+  // without having to add an extra field.
+  includeInChatHistory?: true;
```
sanchitmalhotra126: We're using a type of `true` here but allowing a `boolean` when creating an error notification in `dispatchSaveFailNotification`, right? Would be good to be consistent either way - in `onSaveFail`, we are using `flaggedProperties` to determine whether `includeInChatHistory` is true or false, so it seems like we do want to allow this value to be either `true` or `false`, I think?

apps/src/aichat/views/AichatView.tsx
```diff
@@ -108,6 +114,13 @@ const AichatView: React.FunctionComponent = () => {
         studentAiCustomizations,
       })
     );
+    dispatch(
+      addChatEvent({
+        timestamp: Date.now(),
+        descriptionKey: 'LOAD_LEVEL',
+        hideForParticipants: true,
+      } as ChatEvent)
```
sanchitmalhotra126: minor: why do we need this cast?

sanchitmalhotra126 (commented):
Looks good! Just a few minor questions

apps/src/aichat/redux/aichatRedux.ts
fisher-alice: Oh yeah - I think I iterated on this a bit. Can keep inlined.

apps/src/aichat/views/AichatView.tsx
fisher-alice: Not needed! Removed.

apps/src/aichat/types.ts
fisher-alice: Oh yes! I first defined this, then went back and added the check for `flaggedProperties` on a later update so will update to `boolean`. Thanks and good catch!

apps/src/aichat/chatEventLogger.ts
fisher-alice: Removed logging.

apps/src/aichat/redux/aichatRedux.ts
```diff
@@ -288,7 +287,12 @@ export const onSaveFail =
           if (body?.details?.profaneWords?.length > 0 && flaggedProperties) {
             errorMessage = `Profanity detected in the ${flaggedProperties} and cannot be updated. Please try again.`;
           }
-          dispatchSaveFailNotification(dispatch, errorMessage);
+          const includeInChatHistory = flaggedProperties ? true : false;
+          dispatchSaveFailNotification(
+            dispatch,
+            errorMessage,
+            !!includeInChatHistory
```
sanchitmalhotra126: nit: do you need the separate `includeInChatHistory` local variable? can this just be `!!flaggedProperties`?

apps/src/aichat/redux/aichatRedux.ts
```diff
@@ -362,15 +368,18 @@ export const submitChatContents = createAsyncThunk(
         .logError('Error in aichat completion request', error as Error);
 
       thunkAPI.dispatch(clearChatMessagePending());
-      thunkAPI.dispatch(addChatMessage({...newMessage, status: Status.ERROR}));
+      thunkAPI.dispatch(
+        addChatEvent({...newUserMessage, status: Status.ERROR} as ChatEvent)
```
sanchitmalhotra126: do we need the cast here?

apps/src/aichat/redux/aichatRedux.ts
```diff
@@ -362,15 +368,18 @@ export const submitChatContents = createAsyncThunk(
...
-        timestamp: Date.now(),
-      };
-      thunkAPI.dispatch(addChatMessage(assistantChatMessage));
+      thunkAPI.dispatch(
+        addChatEvent({
+          role: Role.ASSISTANT,
+          status: Status.ERROR,
+          chatMessageText: 'error',
+          timestamp: Date.now(),
+        } as ChatEvent)
```
sanchitmalhotra126: Here as well?

apps/src/aichat/redux/aichatRedux.ts
```diff
@@ -395,23 +404,59 @@ export const submitChatContents = createAsyncThunk(
...
+        const {updatedField, updatedValue} = chatEvent;
+        // Only log updated value for temperature and selected model id - free text values are not logged.
+        const updatedValueToLog =
+          updatedField === 'temperature' || updatedField === 'selectedModelId'
+            ? updatedValue
+            : 'N/A';
+        chatEvent = {
+          ...chatEvent,
+          updatedValue: updatedValueToLog,
+        } as ModelUpdate;
```
sanchitmalhotra126: Same here, why is the cast needed? Just curious because I would have expected that the `isModelUpdate` function would have already narrowed the type

apps/src/aichat/views/AichatView.tsx
```diff
@@ -179,6 +192,13 @@ const AichatView: React.FunctionComponent = () => {
 
   const onClear = useCallback(() => {
     dispatch(clearChatMessages());
+    dispatch(
+      addChatEvent({
+        timestamp: Date.now(),
+        descriptionKey: 'CLEAR_CHAT',
+        hideForParticipants: true,
+      } as ChatEvent)
```
sanchitmalhotra126: Same here as well

sanchitmalhotra126 (approved):
LGTM! Just a few minor things, nothing blocking

apps/src/aichat/redux/aichatRedux.ts
fisher-alice: Sure!

apps/src/aichat/redux/aichatRedux.ts
fisher-alice: I added bc I am getting a TS error without the cast.
Screenshot 2024-08-07 at 4 13 21 PM

apps/src/aichat/redux/aichatRedux.ts
fisher-alice: TS error here as well:
`Object literal may only specify known properties, and 'role' does not exist in type 'ChatEvent'.ts(2353)`

apps/src/aichat/redux/aichatRedux.ts
fisher-alice: Without cast: `Object literal may only specify known properties, and 'updatedValue' does not exist in type 'ChatEvent'.ts(2353)`

apps/src/aichat/views/AichatView.tsx
fisher-alice: Will remove here. No cast needed. I think I got in the habit of casting from cases above 😅

apps/src/aichat/redux/aichatRedux.ts
```diff
@@ -395,23 +403,58 @@ export const submitChatContents = createAsyncThunk(
...
+    chatApiResponse.messages.forEach(message => {
+      thunkAPI.dispatch(addChatEvent({...message, timestamp: Date.now()}));
+    });
+  }
+);
+
+// This thunk adds a chat event to chatEventsCurrent (displayed in current chat workspace)
+// if hideForParticipants != true and then logs the event to the backend for all chat events
+// except notifications with includeInHistory != true.
+export const addChatEvent = createAsyncThunk(
```
sanchitmalhotra126: Oh one quick thing (sorry I didn't notice this before) - does this need to be an async thunk? Is there any async work  happening in this function?

apps/src/aichat/redux/aichatRedux.ts
sanchitmalhotra126: Ah interesting, does it work if you use a generic in the `addChatEvent` function signature?
```
<T extends ChatEvent>(chatEvent: T, thunkAPI) => { ... }
```

fisher-alice: Oh yeah - that worked. Thanks!

apps/src/aichat/redux/aichatRedux.ts
fisher-alice: Good catch! Nope, the async stuff is now in `ChatEventLogger`. Updated!

PR #60111, opened 2024-07-31
Python Lab: allow everyone to run code

molly-moen:
We can now remove the permissions check on Python Lab that only allowed levelbuilders to run Python code!

If we decide in the future that we want to have permissions checks on other labs (i.e. if we migrate Java Lab to Code Bridge) we can add the call back to check permissions, but since it wasn't used it didn't make sense to keep that API call.

## Links

- jira ticket: [CT-513](https://codedotorg.atlassian.net/browse/CT-513)

## Testing story
Tested that any user can now run Python code.

## PR Checklist:

- [ ] Code is well-commented


ebeastlake (approved):
LGTM! 🎉

PR #60112, opened 2024-07-31
Update promote map buttons

dmcavoy:
The state fact sheets were out of date that were being used on the code.org/promote map. A better more current version of these sheets is to send people to the State of CS pages for their states. This updates to point at the State of CS report instead of landscape report and state fact sheets. It then deprecates the unused resources.

Screenshot 2024-07-31 at 1 33 45 PM

@kelbyhawn is there an easy way to redirect everything in a certain directory? I want to redirect all the state fact sheets to the State of CS page on advocacy.code.org


kelbyhawn:
> @kelbyhawn is there an easy way to redirect everything in a certain directory? I want to redirect all the state fact sheets to the State of CS page on advocacy.code.org

I'm not sure 🤔 I'll investigate this tomorrow.

dmcavoy:
@bethanyaconnor wondering if there are pdf generation things we can remove with this change too?

kelbyhawn:
@dmcavoy to redirect everything in the `state-facts` directory you can add this code to the `state_sheets.rb` [file](https://github.com/code-dot-org/code-dot-org/blob/staging/pegasus/routes/state_sheets.rb):
```
get '/advocacy/state-facts/*' do
  redirect "https://advocacy.code.org/stateofcs/"
end
```

It's probably ok to remove the existing code in that file too.

bethanyaconnor:
Looks like there are a couple generation pieces that can be removed:
- At least [this function](https://github.com/code-dot-org/code-dot-org/blob/c26d909f1f3476baf02e9b2ac8c979b611ed11b1/bin/content-push#L43) in `content-push`, possibly more
- Some or all of [generate_pdfs](https://github.com/code-dot-org/code-dot-org/blob/staging/pegasus/rake/generate_pdfs.rake)

After this, do we even need `cdo_state_promote` anymore? If not, we can remove some logic around that table and drop the table itself.

dmcavoy:
I think we still need cdo_state_promote for a couple facts for the map. Hopefully can remove those later

dmcavoy:
@kelbyhawn @bethanyaconnor can you take another look?

kelbyhawn (approved):
🎉

PR #60117, opened 2024-07-31
only include school and district in workshop organizer registration confirmation if provided

dmcavoy:
Will only show school and district information in workshop organizer registration confirmation if they exist

### DIFF
--- a/dashboard/app/views/pd/workshop_mailer/_teacher_info.html.haml
+++ b/dashboard/app/views/pd/workshop_mailer/_teacher_info.html.haml
@@ -14,11 +14,12 @@ Teacher Info:
     %td Email:
     %td
       = @enrollment.email
-  %tr
-    %td District:
-    %td
-      = @enrollment.school_district_name
-  %tr
-    %td School:
-    %td
-      = @enrollment.school_name
+  - if @enrollment.school_district_name && @enrollment.school_name
+    %tr
+      %td District:
+      %td
+        = @enrollment.school_district_name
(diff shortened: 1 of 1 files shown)


dashboard/app/views/pd/workshop_mailer/_teacher_info.html.haml
```diff
@@ -14,11 +14,12 @@ Teacher Info:
...
       = @enrollment.email
-  %tr
-    %td District:
-    %td
-      = @enrollment.school_district_name
-  %tr
-    %td School:
-    %td
-      = @enrollment.school_name
+  - if @enrollment.school_district_name && @enrollment.school_name
```
TurnerRiley, line 17: Is there a situation where we would want one and not the other? Like should the logic be to show the school district if the school district exists and show the school name if the school name exists, rather than show both only if both are present?

TurnerRiley (approved):
LGTM! Had a quick product question

dashboard/app/views/pd/workshop_mailer/_teacher_info.html.haml
dmcavoy: Hmmm good question. I was assuming that in order to have one you must have both because of NCES matching.... I don't know that its that important that one is there if the other is blank so will keep as is for now

TurnerRiley: Sounds good!

PR #60119, opened 2024-07-31
Music: make SoundCache a singleton

sanchitmalhotra126:
Make SoundCache a singleton, so that it can persist between level changes even if Music Lab is unmounted. Currently Music Lab is always rendered in the background if present in a lab2 progression, partially because we wanted the cache to persist between levels. Now the sounds will still stay cached even if we don't always have Music Lab rendered in the background. 

Note that Music Lab will still stay rendered in the background as of this PR - there is one more necessary change related to analytics lifecycle management before we can flip the flag to not have Music Lab render in the background.

Example: (I manually set Music Lab to not render in the background (`backgroundMode: false`) locally only)

https://github.com/user-attachments/assets/b62a5f11-f592-4f22-aeb5-e660bc106889

## Links

https://codedotorg.atlassian.net/browse/LABS-925

## Testing story

Tested on the intro progression and verified sounds are not reloaded once cached even when switching to a panels level and back.

### DIFF
--- a/apps/src/music/player/SoundCache.ts
+++ b/apps/src/music/player/SoundCache.ts
@@ -6,12 +6,21 @@ import {baseAssetUrlRestricted} from '../constants';
 import {LoadFinishedCallback} from '../types';
 
 class SoundCache {
+  private static instance: SoundCache;
+
+  public static getInstance() {
+    if (!SoundCache.instance) {
+      SoundCache.instance = new SoundCache();
+    }
+    return SoundCache.instance;
+  }
+
   private readonly audioContext: AudioContext;
   private readonly metricsReporter: LabMetricsReporter;
   private audioBuffers: {[id: string]: AudioBuffer};
   private hasLoadedSignedCookies: boolean;
 
-  constructor(
+  private constructor(
     audioContext: AudioContext = new AudioContext(),
     metricsReporter: LabMetricsReporter = Lab2Registry.getInstance().getMetricsReporter()
   ) {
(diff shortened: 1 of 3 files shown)


apps/src/music/player/SoundCache.ts
```diff
@@ -6,12 +6,21 @@ import {baseAssetUrlRestricted} from '../constants';
 import {LoadFinishedCallback} from '../types';
 
 class SoundCache {
+  private static instance: SoundCache;
```
fisher-alice, line 9: Nit: use underscore for `instance`

fisher-alice (approved):
NIce! Singletons can be so handy.

breville (approved):

apps/src/music/player/SoundCache.ts
sanchitmalhotra126: Personal preference, I usually don't use underscores for private members in typescript since we already have the `private` modifier. But can change if you think it's clearer!

fisher-alice: Not a personal preference, but saw it used in `Lab2Registry.ts` so thought we could be consistent. I have it used in my current PR https://github.com/code-dot-org/code-dot-org/blob/e7eeb24045b16bd27f9bd8aa7fb59aad7866edd8/apps/src/aichat/chatEventLogger.ts.
I can remove in my current PR for consistency.

sanchitmalhotra126: Gotcha - yeah I don't have a strong opinion either way. I'll merge this is as and maybe we can revisit as part of a broader code style discussion?

PR #60121, opened 2024-07-31
Add retry and timeout to flaky DCDO test

carl-codeorg:
Adds a 1-second retry and 10-second timeout to a flaky DCDO test. Previously, [I had a PR up](https://github.com/code-dot-org/code-dot-org/pull/60021) to correct this same issue by making the call to DynamoDB strongly consistent, but the majority of reviewers did not like that solution.

This retry loop is a bit more basic of a solution, but carries no risk of affecting other tests or reducing parity with production configuration. Note that this test is only flaky in the test env (which uses DynamoDB to back DCDO), so we won't be able to observe that the test is truly fixed until it's been merged and no longer produces flaky results over time.

## Links

[JIRA](https://codedotorg.atlassian.net/browse/P20-1064)

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/dashboard/test/lib/dcdo_test.rb
+++ b/dashboard/test/lib/dcdo_test.rb
@@ -3,10 +3,28 @@ require 'dynamic_config/dcdo'
 require 'timeout'
 
 class DCDOTest < ActiveSupport::TestCase
+  setup do
+    DCDO.unstub(:get)
+  end
+
   test 'basic set and get' do
-    assert_nil DCDO.get('key', nil)
-    DCDO.set('key', 'okay')
-    assert_equal DCDO.get('key', nil), 'okay'
+    key = SecureRandom.hex
+    assert_nil DCDO.get(key, nil)
+    DCDO.set(key, 'okay')
+
+    result = nil
+    Timeout.timeout(10) do
+      loop do
+        result = DCDO.get(key, nil)
+        if result.present?
+          break
+        else
+          sleep 1
+        end
+      end
+    end
+
+    assert_equal result, 'okay'
   end
 
   test 'returns default if key is not stored' do
@@ -18,7 +36,7 @@ class DCDOTest < ActiveSupport::TestCase
       'b' => 'yo dude',
       'c' => [1, 2, 3]
     }
-    key = 'random'
+    key = SecureRandom.hex
     DCDO.set(key, to_store)
 
     result = nil
--- a/dashboard/test/lib/dcdo_test.rb
+++ b/dashboard/test/lib/dcdo_test.rb
@@ -1,5 +1,6 @@
 require 'test_helper'
 require 'dynamic_config/dcdo'
+require 'timeout'
 
 class DCDOTest < ActiveSupport::TestCase
   test 'basic set and get' do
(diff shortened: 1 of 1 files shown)


artem-vavilov:
We would see the error "DynamoDBAdapter does not support clear" if we use DynamoDB for unit testing:
https://github.com/code-dot-org/code-dot-org/blob/a6859c54dd045d99396375ab85c11388452141d0/lib/dynamic_config/adapters/dynamodb_adapter.rb#L78-L80

https://github.com/code-dot-org/code-dot-org/blob/a6859c54dd045d99396375ab85c11388452141d0/dashboard/test/test_helper.rb#L84

dashboard/test/lib/dcdo_test.rb
```diff
@@ -19,7 +20,20 @@ class DCDOTest < ActiveSupport::TestCase
     }
     key = 'random'
     DCDO.set(key, to_store)
-    assert_equal DCDO.get(key, nil), to_store
```
artem-vavilov, line 22: The test "basic set and get" is quite similar, is it also flaky?
https://github.com/code-dot-org/code-dot-org/blob/a6859c54dd045d99396375ab85c11388452141d0/dashboard/test/lib/dcdo_test.rb#L6-L10

artem-vavilov: I have a suspicion that the issue might be in `DCDO.stubs(:get)`, which we use frequently in other test cases.

We could try unstubbing the method `DCDO.get` before each test:
```ruby
setup do
  DCDO.unstub(:get)
end

test 'basic set and get' do 
  assert_nil DCDO.get('key', nil) 
  DCDO.set('key', 'okay') 
  assert_equal DCDO.get('key', nil), 'okay' 
end 
```

or test a new instance of `DCDO` that is not stubbed:
```ruby
setup do
  @dcdo = DCDO.dup
end

test 'basic set and get' do
  assert_nil @dcdo.get('key', nil) 
  @dcdo.set('key', 'okay') 
  assert_equal @dcdo.get('key', nil), 'okay' 
end
```

artem-vavilov: We could also ensure that the `DCDO` will not be rewritten in any other place by using a unique key for each test:
```ruby
test 'basic set and get' do
  dcdo_key = SecureRandom.hex
  dcdo_value = 'okay'

  assert_changes -> {DCDO.get(dcdo_key, nil)}, from: nil, to: dcdo_value do
    DCDO.set(dcdo_key, dcdo_value)
  end
end
```

carl-codeorg: > The test "basic set and get" is quite similar, is it also flaky?
> 
> https://github.com/code-dot-org/code-dot-org/blob/a6859c54dd045d99396375ab85c11388452141d0/dashboard/test/lib/dcdo_test.rb#L6-L10

Good catch - I'm assuming it also has the potential to be flaky even though it hasn't been called out yet, so I added the retry there as well.

carl-codeorg: > I have a suspicion that the issue might be in `DCDO.stubs(:get)`, which we use frequently in other test cases.
> 

Good idea; I added the unstub

carl-codeorg:
closing for now - don't want to take up reviewers time on this if there's a possible better solution

PR #60123, opened 2024-07-31
Set MailJet to sed CAP section warning emails

mgc1194:
### This PR does:
- Edits the `send_template_email` method to support Variables
- Adds a new method `send_cap_section_warning_email` that sends a CAP Section Warnign Email to the specified user, warning about the specified sections.
- Adds the development `template_id` to `mailjet_constants.rb`

## Links

- jira ticket: [P20-1072](https://codedotorg.atlassian.net/browse/P20-1072)
- [MailJet documentation](https://dev.mailjet.com/email/guides/send-api-v31/#use-templating-language)

## Testing story
Steps taken:
1. Added my locals.yml with the development values for 

```
mailjet_api_key: 
mailjet_secret_key: 
mailgun_api_key:
```
2. started `dashboard-console`
3. Looked for a user with a valid email: `u = User.find()`
4. Imported MailJet module to the console: `require 'cdo/mailjet'`
5. Invoke the added method:
```
MailJet.send_cap_section_warning_email(mario, {capSections: [{ Name: 'Section 1', Link: 'https://example.com/section1' }, { Name: 'Section 2', Link: 'https://example.com/section 2' }]})
```
6. Received email. Note that the links do not show up as there's a bug with the font size. Ticket submitted [here](https://app.mailjet.com/support/ticket/3125148)
Screenshot 2024-08-07 at 15 15 26

## PR Checklist:

- [ ] Code is well-commented
- [ ] Pull Request is labelled appropriately


lib/cdo/mailjet.rb
```diff
@@ -51,6 +51,47 @@ def self.create_contact_and_send_welcome_email(user, locale = 'en-US')
...
+  #     - the user is not a teacher.
+  #
+  # Workflow:
+  #   1. Checks if the feature is enabled.
+  #   2. Verifies the presence of a valid user ID.
+  #   3. Ensures the user has a teacher role.
+  #   4. Finds or creates a contact using the user's email and name.
+  #   5. Sends a template email with the specified sections in the specified locale.
+  def self.send_cap_section_warning_email(user, sections, locale = 'en-US')
+    return unless enabled?
```
artem-vavilov, line 67: Is it okay to skip sending this warning if Mailjet is disabled for some reason?
Shouldn't we use another mailer instead?

lib/cdo/shared_constants/mailjet_constants.rb
```diff
@@ -17,6 +17,18 @@ module MailJetConstants
...
+    cap_section_warning: {
+      template_id: {
+        staging: {
+          default: 6_190_124,
+        },
+        development: {
+          default: 6_189_697,
+        }
+      },
+      from_address: 'hadi_partovi@code.org',
```
daynew:
```suggestion
      from_address: 'noreply@code.org',
```

lib/cdo/shared_constants/mailjet_constants.rb
```diff
@@ -17,6 +17,18 @@ module MailJetConstants
...
+      from_name: 'Hadi Partovi',
```
daynew:
```suggestion
      from_name: 'Code.org',
```

daynew (approved):
Looks good! I will do a follow-up PR which will create an ActiveJob wrapper for these emails and also move this email to our ApplicationMailers codebase.

lib/cdo/shared_constants/mailjet_constants.rb
```diff
@@ -17,6 +17,18 @@ module MailJetConstants
...
       from_name: 'Hadi Partovi',
+    },
+    cap_section_warning: {
+      template_id: {
+        staging: {
+          default: 6_190_124,
+        },
+        development: {
+          default: 6_189_697,
+        }
```
daynew: What about `production`?

lib/test/cdo/test_mailjet.rb
```diff
@@ -154,6 +156,31 @@ def test_create_contact_and_send_welcome_email
...
+    sign_up_time = Time.now.to_datetime
+
+    user = mock
+    user.stubs(:id).returns(1)
+    user.stubs(:email).returns(email)
+    user.stubs(:name).returns('Fake Name')
+    user.stubs(:teacher?).returns(true)
+    user.stubs(:created_at).returns(sign_up_time)
+
+    sections = {'sections' => [
```
daynew: Should this be `'sections'` or `'capSections'`? The documentation is not consistent with the test.

lib/cdo/mailjet.rb
```diff
@@ -51,6 +51,47 @@ def self.create_contact_and_send_welcome_email(user, locale = 'en-US')
...
+  #     - the user is nil or does not have a valid ID.
+  #     - the user is not a teacher.
+  #
+  # Workflow:
+  #   1. Checks if the feature is enabled.
+  #   2. Verifies the presence of a valid user ID.
+  #   3. Ensures the user has a teacher role.
+  #   4. Finds or creates a contact using the user's email and name.
+  #   5. Sends a template email with the specified sections in the specified locale.
+  def self.send_cap_section_warning_email(user, sections, locale = 'en-US')
```
daynew:
```suggestion
  def self.send_cap_section_warning_email(user, variables, locale = 'en-US')
```
`variables` might be a better name because it is being passed through directly to the `variables` part of the template. Or, instead the sections could be an array, and then in this method you create`{sections: sections}` and pass that to the template. Imagine how we would support a different variable, like `school_name`.

artem-vavilov, line 80:
```suggestion
  def self.send_cap_section_warning(user, sections, locale: 'en-US')
```

artem-vavilov: Ruby has very useful `optional arguments` that can be used precisely in such situations:
```suggestion
  def self.send_cap_section_warning(user, sections, locale: 'en-US')
```

lib/cdo/mailjet.rb
```diff
@@ -51,6 +51,47 @@ def self.create_contact_and_send_welcome_email(user, locale = 'en-US')
...
+    return unless enabled?
+
+    return unless user&.id.present?
+    return unless user.teacher?
```
artem-vavilov:
```suggestion
    raise ArgumentError, 'the user must be a teacher' unless user.teacher?
```

lib/cdo/mailjet.rb
```diff
@@ -51,6 +51,47 @@ def self.create_contact_and_send_welcome_email(user, locale = 'en-US')
...
+  # Workflow:
+  #   1. Checks if the feature is enabled.
+  #   2. Verifies the presence of a valid user ID.
+  #   3. Ensures the user has a teacher role.
+  #   4. Finds or creates a contact using the user's email and name.
+  #   5. Sends a template email with the specified sections in the specified locale.
+  def self.send_cap_section_warning_email(user, sections, locale = 'en-US')
+    return unless enabled?
+
+    return unless user&.id.present?
```
artem-vavilov:
```suggestion
    raise ArgumentError, 'the user must be persisted' unless user&.persisted?
```

lib/cdo/mailjet.rb
artem-vavilov: Wouldn't it be better to create a single universal method called `send` instead of a separate method for each mail?
```ruby
def self.send(template_name, user, *vars, locale: :'en-US')
  template_config = EMAILS[template_name]
  raise ArgumentError, "template config for `#{template_name}` does not exist" unless template_config
  
  contact = find_or_create_contact(user.email, user.name)
    
  send_template_email(
    contact,
    template_config,
    locale,
    vars
  )
end
```

mgc1194: Definitely agreed.
I have modified the method to be more strict and expect a list of sections since they are the main subject of this email. I think we should add validation to check sections have Name and Link.
Another option, once we have more functionality added to the mailer, is passing Section objects and forming the links in this method (instead of explicitly passing names and links for each section).
What do you think @daynew ?

lib/cdo/mailjet.rb
mgc1194: Is it preferred to raise an error or silently fail if the user is not a teacher? I was just following the patters of the other emails.

lib/cdo/mailjet.rb
mgc1194: I see the benefits of having a universal method. Currently, we are just using it for two types of email but there's a risk it'll grow as we migrate some other mailers to MailJet.
To not block the CAP work, I propose creating a ticket to abstract the universal "`send`" method.

lib/cdo/mailjet.rb
artem-vavilov: I believe that if there is an attempt in the code to send the email to someone who shouldn't receive it,
it's better to be aware of the issue and fix it rather than letting the mistake go unnoticed.

lib/cdo/mailjet.rb
```diff
@@ -51,6 +51,45 @@ def self.create_contact_and_send_welcome_email(user, locale = 'en-US')
...
+  #     - the user is nil or does not have a valid ID.
+  #     - the user is not a teacher.
+  #
+  # Workflow:
+  #   1. Checks if the feature is enabled.
+  #   2. Verifies the presence of a valid user ID.
+  #   3. Ensures the user has a teacher role.
+  #   4. Finds or creates a contact using the user's email and name.
+  #   5. Sends a template email with the specified sections in the specified locale.
+  def self.send_cap_section_warning_email(user, sections = [], locale = 'en-US')
```
artem-vavilov: 1. There's no need to include `email` in the method name, as it unnecessarily lengthens it without adding meaningful context, given that MailJet sends only emails.
2. Instead, I would suggest adding `teacher` to the name, as it provides specific and valuable context.
3. Please use the optional keyword argument style for the `locale` argument to make the method more flexible and adaptable, especially if we need to add another optional argument in the future.
4. The `sections` argument shouldn't be optional, as sections are required for the email.
```suggestion
  def self.send_teacher_cap_section_warning(user, sections, locale: 'en-US')
    raise ArgumentError, 'sections must not be empty' if sections.blank?
```

lib/cdo/mailjet.rb
```diff
@@ -51,6 +51,45 @@ def self.create_contact_and_send_welcome_email(user, locale = 'en-US')
...
+  #
+  # Parameters:
+  # user     - a single User object.
+  # sections - an Array of sections, each with a 'Name' and a 'Link'.
+  #            Example format:
+  #             [
+  #               { Name: 'Section 1', Link: 'https://example.com/section1' },
+  #               { Name: 'Section 2', Link: 'https://example.com/section2' }
+  #             ]
+  # locale   - (optional) the locale for the email template. Defaults to 'en-US'.
```
artem-vavilov: Please use the [YARD](https://rubydoc.info/gems/yard) style documentation syntax, as it's commonly used in the project. This ensures consistency and clarity across the entire codebase.
```suggestion
  # Sends a warning to a teacher about sections with students affected by the Child Account Policy (CAP).
  #
  # @note Nothing if the function exits early due to any of the following conditions:
  #   - the feature is not enabled.
  #   - the user is nil or does not have a valid ID.
  #   - the user is not a teacher.
  #
  # @param user [User] The teacher to be warned.
  # @param sections [Array<Hash>] The list of CAP-affected sections. 
  #   Each section is represented as a hash with keys: 
  #   - `:Name` [String] The name of the section.
  #   - `:Link` [String] The URL link to the section.
  # @param locale [String] The locale to use for the email. Defaults to 'en-US'.
  #
  # @return [void]
```

lib/cdo/mailjet.rb
mgc1194: Would we want to log a honeybadger error instead?

artem-vavilov: We can handle this outside of this method if needed, as the method should not be responsible for mistakes made by the caller.

lib/cdo/mailjet.rb
```diff
@@ -110,33 +140,23 @@ def self.add_to_contact_list(contact, list_id)
     )
   end
 
-  def self.send_template_email(contact, email_config, locale = 'en-US')
+  def self.send_template_email(contact, email_config, locale = 'en-US', variables = {})
```
artem-vavilov:
```suggestion
  def self.send_template_email(contact, email_config, locale = 'en-US', variables: {})
```

artem-vavilov (approved):
Good job! 🙌🏻

lib/cdo/mailjet.rb
wilkie: Or, at least, should we raise some kind of error if we're on production to be able to tell that the mail is not going out for whatever reason?

wilkie (approved):
Looks good! I don't mind either way you address capturing unsent emails due to `enabled?` being false. Maybe we push a record to honeybadger.

PR #60130, opened 2024-07-31
apps reorg: sort a handful of components 

Erin007:
More shuffling around in the junk drawer that is the apps directory 😅 

- `Dialog` and `DialogButtons` now live in "legacySharedComponents": current preferred components are `StylizedBaseDialog` or `AccessibleDialog`, which now both live in "sharedComponents".  DSCO Dialogs are slated for autumn. 

- `FontAwesome` now lives in "legacySharedComponents", preferred component is DSCO `FontAwesomeV6Icon`. There are currently 156 places that use the old `FontAwesome` component. 

- `LegacyButton` now lives in "legacySharedComponents", preferred component is DSCO `Button`.`LegacyButton` is used in 5 places and work to deprecate is tracked https://codedotorg.atlassian.net/browse/XTEAM-430. 

- `PendingButton` now lives in "legacySharedComponents". preferred component is DSCO `Button` which has a pending state with a spinner. `PendingButton` is used in 6 places in the Data Browser and work to deprecate is tracked https://codedotorg.atlassian.net/browse/XTEAM-446

- `Notification` and `SearchBar` now live in "sharedComponents" since they are used widely but do not have DSCO counterparts. 

- `TeacherNavigationBar.story` now lives in the "teacherDashboard" folder since it should sit alongside the component it is the Story for and it was just floating around in templates. 

The majority of the diff is updated paths in files that reference theses components.


apps/src/legacySharedComponents/Button.jsx
sanchitmalhotra126, line 1: Should we add `@deprecated` tags to these components that have newer preferred replacements? Edit: and which components to use instead?

sanchitmalhotra126: Thanks! one more nitpicky request if it's not too much work 😄 could we move the deprecated tags to the JSDocs on the actual class/function, so we get the nice strikethrough to show up in the IDE whenever the component is used elsewhere? Just serves as an extra visual reminder when seeing these components in other classes.

i.e.
```
/**
 * @deprecated use DSCO button instead
 */
class Button extends React.Component { ... }
```
and shows up as
Screenshot 2024-08-02 at 2 36 26 PM

Erin007: updated ✔️ thanks for filling me in! I didn't know JSDocs could do that!

sanchitmalhotra126 (approved):
🎉 awesome reorganization! Thanks for doing this!

PR #60134, opened 2024-07-31
Removing the CPA exceptions added by p20-937

daynew:
During the CPA Live Site issue in May 2024 we found that we weren’t requiring new Email, Facebook, and Microsoft accounts to declare their US State. This resulted in users suddenly being locked out when they told us their US State. We added an exception for these users so classrooms would not be interrupted. Now that it is Summer, and most schools are not in session, we are going to remove the exception.

## Links
[Jira](https://codedotorg.atlassian.net/browse/P20-942)

## Testing story
* Unit tests

## PR Checklist:

- [ ] Code is well-commented


artem-vavilov (approved):
Good job! 🙌🏻

PR #60136, opened 2024-07-31
Speculative fix for the Create Account button experiment loading issue

kelbyhawn:
Adds logic to run the Statsig Reporter event listener once the document is ready. I based this off of what @breville suggested [here](https://codedotorg.slack.com/archives/C046G4TRLEN/p1722373489753019?thread_ts=1722290836.545479&cid=C046G4TRLEN). 

Hopefully this will fix the issue on production where the Create Account button isn't showing up unless another page is visited first. I added console log statements so we can check this on production since we're unable to add breakpoints.

**Making sure the function runs after the document has loaded:** 

https://github.com/user-attachments/assets/9c2cc017-aa3b-4693-9acc-7b540d003d7b

## Links
Jira ticket: [ACQ-2173](https://codedotorg.atlassian.net/browse/ACQ-2173)
Slack convo: [here](https://codedotorg.slack.com/archives/C046G4TRLEN/p1722373489753019?thread_ts=1722290836.545479&cid=C046G4TRLEN)

## Testing story
Tested locally to see the console.log statements and checked to make sure the experiment was still working in Statsig:
Screenshot 2024-07-31 at 4 06 49 PM

### DIFF
--- a/apps/src/userHeaderStatsigReporter/userHeaderStatsigReporter.js
+++ b/apps/src/userHeaderStatsigReporter/userHeaderStatsigReporter.js
@@ -6,6 +6,7 @@ if (document.readyState !== 'loading') {
 } else {
   document.addEventListener('DOMContentLoaded', function () {
     console.log('Document is loading');
+    console.log(`Document ready state: ${document.readyState}`);
     runStatsigReporter();
   });
 }
--- a/apps/src/userHeaderStatsigReporter/userHeaderStatsigReporter.js
+++ b/apps/src/userHeaderStatsigReporter/userHeaderStatsigReporter.js
@@ -5,7 +5,7 @@ if (document.readyState !== 'loading') {
   runStatsigReporter();
 } else {
   document.addEventListener('DOMContentLoaded', function () {
-    console.log('Document is loading');
+    console.log('DOM content has loaded');
     console.log(`Document ready state: ${document.readyState}`);
     runStatsigReporter();
   });
--- a/apps/src/userHeaderStatsigReporter/userHeaderStatsigReporter.js
+++ b/apps/src/userHeaderStatsigReporter/userHeaderStatsigReporter.js
@@ -1,6 +1,17 @@
 import statsigReporter from '../lib/util/StatsigReporter';
 
-document.addEventListener('DOMContentLoaded', function () {
+if (document.readyState !== 'loading') {
+  console.log('Document is ready');
+  runStatsigReporter();
+} else {
+  document.addEventListener('DOMContentLoaded', function () {
+    console.log('Document is loading');
+    runStatsigReporter();
+  });
+}
+
+function runStatsigReporter() {
+  console.log('Running Statsig Reporter');
   const createAccountButton = document.querySelector('#create_account_button');
   const createAccountButtonDesktop = document.querySelector(
     '#create_account_button.desktop'
@@ -49,4 +60,4 @@ document.addEventListener('DOMContentLoaded', function () {
     // Hide the Sign in and Create account buttons in the hamburger
     hamburgerButtons ? (hamburgerButtons.style.display = 'none') : null;
   }
-});
+}


apps/src/userHeaderStatsigReporter/userHeaderStatsigReporter.js
```diff
@@ -1,6 +1,18 @@
 import statsigReporter from '../lib/util/StatsigReporter';
 
-document.addEventListener('DOMContentLoaded', function () {
+if (document.readyState !== 'loading') {
+  console.log('Document is ready');
```
breville, line 4: Yes, if we see this particular log output, it should serve as confirmation that we are making the correct fix.

apps/src/userHeaderStatsigReporter/userHeaderStatsigReporter.js
```diff
@@ -1,6 +1,18 @@
...
+  runStatsigReporter();
+} else {
+  document.addEventListener('DOMContentLoaded', function () {
+    console.log('Document is loading');
+    console.log(`Document ready state: ${document.readyState}`);
```
breville, line 9: Since this is happening in the `DOMContentLoaded` handler, it will presumably always be `"complete"`, which might be interesting to confirm, but doesn't seem crucial.

apps/src/userHeaderStatsigReporter/userHeaderStatsigReporter.js
```diff
@@ -1,6 +1,18 @@
...
+    runStatsigReporter();
+  });
+}
+
+function runStatsigReporter() {
+  console.log('Running Statsig Reporter');
```
breville, line 15: This should appear immediately after one of the above newly-added `console.log` calls, so should be interesting to confirm, but again, doesn't seem crucial.

breville (approved):

apps/src/userHeaderStatsigReporter/userHeaderStatsigReporter.js
```diff
@@ -1,6 +1,18 @@
 import statsigReporter from '../lib/util/StatsigReporter';
 
-document.addEventListener('DOMContentLoaded', function () {
+if (document.readyState !== 'loading') {
+  console.log('Document is ready');
+  runStatsigReporter();
+} else {
+  document.addEventListener('DOMContentLoaded', function () {
+    console.log('Document is loading');
```
bethanyaconnor: nit: Is this accurate? The document content has loaded at this point, yeah?

bethanyaconnor (approved):
🤞

apps/src/userHeaderStatsigReporter/userHeaderStatsigReporter.js
kelbyhawn: The console says it's `interactive` so technically it's still loading? 😅 I can remove this if you think it's superfluous!

apps/src/userHeaderStatsigReporter/userHeaderStatsigReporter.js
kelbyhawn: I added this and the one commented on [here](https://github.com/code-dot-org/code-dot-org/pull/60136/files#r1699240205) for extra context as the page is loading. I figured in this case it's better to over-explain than under-explain.

apps/src/userHeaderStatsigReporter/userHeaderStatsigReporter.js
bethanyaconnor: According to [this documentation](https://developer.mozilla.org/en-US/docs/Web/API/Document/readyState#interactive), the document has loaded but subresources haven't. I think my brain doesn't like the word "loaded" on one line followed by "loading" on the next line. I think my preference would be to either remove this line or change it to say "DOM Content Loaded" or something similar. Presumably this line is just to denote when the listener starts?

kelbyhawn: That makes sense! I'll update it to say "DOM content has loaded".

PR #59191, opened 2024-06-12
Update MySQL to 8.0.37

Hamms:
We are currently attempting to install version 8.0.36, which was the latest version available at the time we assembled the change. Unfortunately, that attempt is currently failing with the error:

```
no candidate version available for mysql-client, mysql-client-8.0
```

It appears that version 8.0.36 is no longer provided by the MySQL apt repositories we're using, and they are now only providing version 8.0.37, which represents a minor security update:

```bash
ubuntu@test:~$ apt-cache madison mysql-client
mysql-client | 8.0.37-1ubuntu20.04 | http://repo.mysql.com/apt/ubuntu focal/mysql-8.0 amd64 Packages
mysql-client | 8.0.37-0ubuntu0.20.04.3 | http://us-east-1.ec2.archive.ubuntu.com/ubuntu focal-updates/main amd64 Packages
mysql-client | 8.0.37-0ubuntu0.20.04.3 | http://security.ubuntu.com/ubuntu focal-security/main amd64 Packages
mysql-client | 8.0.19-0ubuntu5 | http://us-east-1.ec2.archive.ubuntu.com/ubuntu focal/main amd64 Packages
ubuntu@test:~$ apt-cache madison mysql-client-8.0
mysql-client-8.0 | 8.0.37-0ubuntu0.20.04.3 | http://us-east-1.ec2.archive.ubuntu.com/ubuntu focal-updates/main amd64 Packages
mysql-client-8.0 | 8.0.37-0ubuntu0.20.04.3 | http://security.ubuntu.com/ubuntu focal-security/main amd64 Packages
mysql-client-8.0 | 8.0.19-0ubuntu5 | http://us-east-1.ec2.archive.ubuntu.com/ubuntu focal/main amd64 Packages
```

Presumably 8.0.36 was pulled because of the security vulnerabilities fixed by 8.0.37, although I so far have been unable to find an official announcement to that effect.

Regardless, picking up this security patch seems like a good idea which will also unblock the pipeline.

## Testing story

Tested on [an adhoc](https://adhoc-mysql-8-0-37.cdn-code.org) to verify that with this change in place we are again able to successfully install mysql from apt.

## PR Checklist:

- [ ] Tests provide adequate coverage
- [ ] Privacy and Security impacts have been assessed
- [ ] Code is well-commented
- [ ] New features are translatable or updates will not break translations
- [ ] Relevant documentation has been added or updated
- [ ] User impact is well-understood and desirable
- [ ] Pull Request is labeled appropriately
[...]

PR #60137, opened 2024-07-31
chef: drop mysql package repo and use ubuntu packages instead

snickell:
The mysql apt repo was very quickly removing old versions, and we didn't trust them to maintain stability. As a result, our exact version pin in our `cdo-mysql` chef recipe broke a couple times. This breakage manifested as AMIBuilder breaking, which resulted in a DTP which appeared to succeed but didn't contain new git changes "it didn't take".

This PR switches our chef recipe to fetch `mysql-client`, `mysql-client-8.0` and `mysql-server` from upstream ubuntu versions, and removes the old mysql sources.list as well as previously installed mysql packages.

Previous breakage: https://github.com/code-dot-org/code-dot-org/pull/59191
Current breakage discussed here: https://codedotorg.slack.com/archives/C0T0PNTM3/p1722448127109449


Hamms (approved):
This LGTM and seems to work as expected, although looking at our persistent managed servers this morning it does appear as though they have all been mysteriously updated to MySQL 8.0.39! It's possible that means some of the removal logic in here has been rendered unnecessary, but I think I'd still like to deploy this change with that logic in place just in case.

I tested this change on two adhocs; specifically, I spun up a new adhoc based on staging which failed to provision because it could not install MySQL 8.0.37 as desired. I then merged this branch in, updated the `cdo-mysql` cookbook, manually ran `chef-client`, and verified that not only did that execution succeed but a full regular build also succeeded therafter. I ran the same test on an adhoc from earlier this week which had originally been created with MySQL 8.0.37 but has like our managed servers since been updated to MySQL 8.0.39. Again, this change applied cleanly and a subsequent regular build went fine.

I think we can proceed with shepherding this change out to all of our servers, and formalizing the unplanned spontaneous update to 8.0.39.

Hamms:
I also took snapshots of the current state of both [staging](https://us-east-1.console.aws.amazon.com/ec2/home?region=us-east-1#SnapshotDetails:snapshotId=snap-01d706757effe68af) and [test](https://us-east-1.console.aws.amazon.com/ec2/home?region=us-east-1#SnapshotDetails:snapshotId=snap-064e5cefebf3243cb) so we can try to figure out the root cause after applying the fix

cat5inthecradle (approved):

PR #60145, opened 2024-08-01
Analytics: GA4 updates

breville:
This PR began with adding a single new GA4 (Google Analytics 4) custom event (to report music pack selection) but has turned into a wholesale removal of legacy UA (Universal Analytics) support, a reduction in the set of custom events we report, and an update to what we do report.

First, a few notes on debugging:

- Google Analytics has a [DebugView](https://support.google.com/analytics/answer/7201382) mode which allows you to see the events coming in from your own browser.
- One way to turn this on in your own browser is using the  [Google Tag Assistant Companion](https://chromewebstore.google.com/detail/tag-assistant-companion/jmekfmbnaedfebfnmakmokmlfpblbfdm) Chrome extension.
- Since the pages are tagged, https://tagassistant.google.com/ seems to work even without the Chrome extension, is quite useful for making sure each page is reporting events properly, and appears to ensure that events show up in the DebugView.

The tidy-up began with reducing how much of Google Analytics we load.  We were still loading both UA and GA4, and it seemed that we were still getting UA custom events, particular in conjunction with the "Collect Universal Analytics events" tag setting that is turned on.  Now, we only load the recommended snippet for GA4.

The tidy-up continued with removing our UA code, leaving only GA4 support in place.

We this change, we also started using the `hourofcode.com` data stream's specific measurement ID.

The "Custom dimensions and metrics" section of [this page](https://support.google.com/analytics/answer/10271001) was quite helpful in understanding the difference in reporting custom events between UA and GA4.

From there, some more drastic tidy-up is being proposed in this PR:
- No more legacy `ga()` calls.
- Existing `trackEvent()` calls are updated to the GA4 format.
- Removal of a _lot_ of legacy custom event calls that we don't seem to be analyzing any more.

This leaves a relatively small set of custom event calls in place, and it's worth noting that the familiar ones have new names and contents, so Google Analytics reports will need to account for this transition.  To take inventory, the intact custom events (with `categoryValue` followed by `actionName`) are:

- `activity`, `activity_lines_of_code`
- `puzzle`, `puzzle_attempt`
- `puzzle`, `puzzle_success`
- `puzzle`, `puzzle_completed`
- `dance`, `dance_play_song`
- `music`, `music_session_start`
- `music`, `music_pack_selected`
- `learn`, `learn_start`
- `learn`, `learn_start_[grade]`
- `share`, `share_open_dialog`
- `share`, `share_copy_url`

Some of these have additional parameters.  These parameters need to be registered in the Google Analytics interface as "custom dimensions" before they will be properly recorded for future report querying.  There is also a limited number of these custom dimensions.  So far, fairly generic names have been used, such as `path` and `value`.  

### Example

Going back to the original event in question, music pack selection, it's handy to look at a screenshot of the DebugView.  

Here is the event:

Screenshot 2024-08-01 at 11 07 06 AM

We can see our long-standing dimensions coming through:

Screenshot 2024-08-01 at 11 07 48 AM

We can see `eventCategory` now coming in as a parameter:

Screenshot 2024-08-01 at 11 08 03 AM

And we see the event-specific `value` also coming in as a parameter:

Screenshot 2024-08-01 at 11 08 09 AM


dju90:
Out of curiosity, how did you determine which custom events we aren't analyzing any more? My reference point - one of the Event Categories that Hadi/marketing really cares about is "HoC_Song" (where the Event Label is the name of the song, such as shawnmendes_ificanthaveyou) - this allows them to pitch/follow-up with artists how many plays their song is getting. I don't see that Event Category in the list above. I also did an inventory earlier this year of all our event categories and got different teams to weigh in on which they considered important to their work - dm me and I'll share it with you.

breville:
> Out of curiosity, how did you determine which custom events we aren't analyzing any more? My reference point - one of the Event Categories that Hadi/marketing really cares about is "HoC_Song" (where the Event Label is the name of the song, such as shawnmendes_ificanthaveyou) - this allows them to pitch/follow-up with artists how many plays their song is getting. I don't see that Event Category in the list above. I also did an inventory earlier this year of all our event categories and got different teams to weigh in on which they considered important to their work - dm me and I'll share it with you.

This is exactly what I was hoping to hear.  Let's talk through what's needed.  FWIW, `"HoC_Song"` is still tracked in the above list, but as `dance_play_song`.

apps/src/StudioApp.js
```diff
@@ -1372,18 +1372,11 @@ StudioApp.prototype.onReportComplete = function (response) {
...
-      response.level_attempts
-    );
-  }
-
-  if (response.share_failure) {
-    trackEvent('Share', 'Failure', response.share_failure.type);
+    trackEvent('puzzle', 'puzzle_completed', {
+      path: response.level_path,
+      value: response.level_attempts,
+    });
```
thomasoniii, line 1377: There's a lot of magic strings hardwired throughout this PR. would it be worth the time to refactor them into external constants? It may be a diminishing return.

apps/src/code-studio/videos.js
```diff
@@ -400,15 +392,9 @@ videos.onYouTubeBlocked = function (youTubeBlockedCallback, videoInfo) {
   testImageAccess(
     youTubeAvailabilityEndpointURL(noCookie) + '?' + Math.random(),
     // Called when YouTube availability check succeeds.
-    function () {
-      // Track event in Google Analytics.
-      trackEvent('showvideo', 'startVideoYouTube', key);
-    },
-
+    function () {},
```
thomasoniii, line 395: Probably unneeded, but just asking - would it be worthwhile to name these functions?

w/o being sure of exactly how this is used, I know that you can get reports in logs that just say "anonymous function", whereas if it was named it might be a more useful report "testImageAccessTracker" or something.

Or those may not show up in this case making it moot. 😄

thomasoniii (approved):
couple of super minor questions, but otherwise lgtm 🚀

dju90:
Looking at something like https://github.com/code-dot-org/code-dot-org/blob/cf26e3b8d5dbf70437a35f2983c51acca8de785d/apps/src/templates/Notification.jsx#L96C7-L96C17, is GA event with the category "teacher_announcement", action name "click" maintained here?

breville:
> Looking at something like https://github.com/code-dot-org/code-dot-org/blob/cf26e3b8d5dbf70437a35f2983c51acca8de785d/apps/src/templates/Notification.jsx#L96C7-L96C17, is GA event with the category "teacher_announcement", action name "click" maintained here?

No, I removed them because I wasn't sure if anybody was still using them.  Only the events listed in the PR description are intact after the current state of this change.

That said, we can certainly take inventory of what we still need and continue reporting them, albeit in the new format.  But I think easiest to have that discussion outside of this PR.

apps/src/code-studio/videos.js
breville: Good point.  GA4 doesn't report function names, so this is probably alright.

apps/src/StudioApp.js
breville: Good point.  Might do this in a follow-up.

PR #60157, opened 2024-08-01
NAV - Create sidebar

kobryan0619:
This:

- Creates a sidebar for navigation
- Adds a dropdown with active sections (non-archived sections)
- Creates a "playground" page hidden behind DCDO and experiement flags (note: to test locally, `nav_test_v2?enableExperiments=teacher-local-nav-v2` at the end of your section URL (ex. `http://localhost-studio.code.org:3000/teacher_dashboard/sections/338/nav_test_v2?enableExperiments=teacher-local-nav-v2` where `388` is a valid section.

The video below shows it in action.  

https://github.com/user-attachments/assets/b8cd13d4-b7d2-4982-9602-4df7c086deed

A few notes about the video:

- Currently the "section progress" view is permanently in there.  It is not expected to update based on the dropdown yet.  
- The second half of the video shows the teacher dashboard working normally with these changes. 

## Links

[Ticket - drop down](https://codedotorg.atlassian.net/jira/software/projects/TEACH/boards/65?selectedIssue=TEACH-1225)
[Ticket - bar](https://codedotorg.atlassian.net/jira/software/projects/TEACH/boards/65?selectedIssue=TEACH-1224)
[Jira](https://www.figma.com/design/5ILfiJkPQpjskBwksMDeZl/Teacher-Dashboard?node-id=3106-3059&m=dev)

## Testing story

Tested locally (see video)

No tests yet... I feel like this is very much a first stab at this and will have clean up work (including tests) to follow. 

I also have not seen this pass drone (yet) so there may be more test updates to follow :)


apps/src/templates/teacherDashboard/TeacherDashboard.jsx
```diff
@@ -84,6 +87,10 @@ function TeacherDashboard({
     }
   });
 
+  const teacherNavV2Enabled =
```
kobryan0619: I considered putting this in a useEffect / useState but I actually didn't see any other examples of this currently in our codebase.  I would be interested in your decision tree here.  In theory, we really only should need this to be calculated once on load...

apps/src/templates/teacherDashboard/TeacherDashboard.jsx
```diff
@@ -99,7 +106,8 @@ function TeacherDashboard({
   // Include header components unless we are on the /login_info or /standards_report page.
   const includeHeader =
     location.pathname !== TeacherDashboardPath.loginInfo &&
-    location.pathname !== TeacherDashboardPath.standardsReport;
+    location.pathname !== TeacherDashboardPath.standardsReport &&
+    location.pathname !== TeacherDashboardPath.navTestV2;
```
kobryan0619, line 115: We don't want the header on this playground space.

apps/src/templates/teacherDashboard/TeacherDashboard.jsx
```diff
@@ -179,6 +187,19 @@ function TeacherDashboard({
...
           )}
         </Route>
+        {teacherNavV2Enabled && (
+          <Route
+            path={TeacherDashboardPath.navTestV2}
+            component={props => (
+              <div className={dashboardStyles.pageContainer}>
+                <TeacherNavigationBar />
+                <div className={dashboardStyles.content}>
+                  <SectionProgressSelector />
```
kobryan0619, line 210: This SectionProgressSelector is only intended to help visualize what it would look like in the future.  The focus of this PR is on the sidebar, but I thought the inclusion of this component here is a more realistic way to see what the page will look like.

apps/src/templates/teacherNavigation/teacher-navigation.module.scss
```diff
@@ -0,0 +1,21 @@
...
+  padding-left: 16px;
+  position: sticky;
+  top: 0;
+  width: 210px;
+}
+
+.sectionHeader {
+  margin-top: 0;
+  margin-bottom: 2px;
+  // TO DO: modify color of text after hearing from Molly
```
kobryan0619, line 20: This was intentionally left in.  This was also documented in the ticket as well.

dashboard/app/assets/stylesheets/application.scss
kobryan0619, line 1: Urgh... this is a large change because it auto-formatted quotes... AND this is probably the change you all need to see the most...

I couldn't figure out how to only commit the important change (not the quotes...) I might need help on this tomorrow...

BUT what you will eventually see in this file is a change to the `overflow: auto` for the `#teacher-dashboard`.  I removed the `overflow`.  I do want extra scrutiny on that decision.  As far as I can tell, it didn't change any behavior on the old teacher dashboard.  Removing this though allowed us to keep the sidebar "sticky" to the top of the screen rather than scroll off the page.

apps/src/templates/teacherDashboard/TeacherDashboard.jsx
lfryemason: I would agree, we only need to do it once. We could use `useEffect` or we can use `_.once()` ([docs](https://lodash.com/docs/4.17.15#once))

apps/src/templates/teacherDashboard/teacher-dashboard.module.scss
```diff
@@ -2,6 +2,18 @@
 @import 'font';
 @import 'style-constants';
 
+.pageContainer {
+  display: flex;
+  flex-direction: row;
+  min-height: 100vh;
```
lfryemason: 100vh will be the entire screen, not including the header. Does this need to be `100%` or remove the space for the global header?

apps/src/templates/teacherNavigation/TeacherNavigationBar.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+interface SectionsData {
+  [sectionId: number]: {
+    name: string;
+    hidden: boolean;
+  };
+}
+
+const TeacherNavigationBar: React.FunctionComponent = () => {
+  const sectionList = useSelector(
```
lfryemason:
```suggestion
  const sections = useSelector(
```
nit: mostly because it's not a list, it's an object.

apps/src/templates/teacherNavigation/teacher-navigation.module.scss
lfryemason: Could you link the ticket from the comment?

apps/src/templates/teacherNavigation/teacher-navigation.module.scss
```diff
@@ -0,0 +1,21 @@
...
+.sidebarContainer {
+  width: 242px;
+  flex: 0 0 242px;
+  background-color: $light_gray_50;
+}
+
+.sidebarContent {
+  padding-top: 20px;
+  padding-left: 16px;
+  position: sticky;
```
lfryemason, line 12: Very nice

dashboard/app/assets/stylesheets/application.scss
lfryemason: What you can do here is make the revert the file, make the changes you want and the `Save without formatting` in vscode (cmd+shift+p -> search for `save without formatting` -> select). We should also see why your formatter thinks all our scss files are wrong, I don't think it should

apps/src/templates/teacherNavigation/TeacherNavigationBar.tsx
```diff
@@ -0,0 +1,67 @@
...
+    setSectionArray(updatedSectionArray);
+
+    // Set initial selectedSectionId if not already set
+    if (updatedSectionArray.length > 0 && !selectedSectionId) {
+      setSelectedSectionId(updatedSectionArray[0].value);
+    }
+  }, [sectionList, selectedSectionId]);
+
+  return (
+    <div className={styles.sidebarContainer}>
```
lfryemason: Let's make this a `<nav>` tag to identify that it is a navigation bar.

lfryemason: You might also find this article helpful: https://www.w3.org/WAI/tutorials/menus/structure/

lfryemason (commented):
Really happy with how it looks so far.

apps/src/templates/teacherNavigation/teacher-navigation.module.scss
kobryan0619: To clarify, this was a documented within the ticket for this task.  It should be fixed before we move this ticket into the "done" column.  There was no separate ticket created for this change.

MollyPeredo: More info - this is a result of the DSCO team making updates to how colors are used in the new design system. I put more info in Slack (teacher-dashboard-lounge).

MollyPeredo:
Looks good to me! I am noticing that the focus state is being triggered for the dropdown and then persisting...is that a feature of the component?

lfryemason:
> Looks good to me! I am noticing that the focus state is being triggered for the dropdown and then persisting...is that a feature of the component?

Yes, this is from the component

apps/src/templates/teacherDashboard/TeacherDashboard.jsx
kobryan0619: I am adding a useEffect, however I am still a little unclear as to why this is actually better.  useState and useEffect feels less clean to me and less readable.  Researching it, I don't see a strong case for performance improvements.  I'd love 2 sentences more on why you think useState/useEffect is a better move here.

apps/src/templates/teacherDashboard/teacher-dashboard.module.scss
kobryan0619: Hmm... it did work with 100vh but I moved it to 100% to be safe.

lfryemason (approved):

PR #60159, opened 2024-08-01
Music: play tune block

breville:
This adds a new `play tune` block for internal evaluation.

It will appear in the toolbox when the `?play-tune-block=true` URL parameter is added.

It shares a lot with the `play drums` block and a little with the `play notes` block.  Beyond the shared `Keybed`, I didn't attempt to factor these too much at this stage.

Screenshot 2024-08-03 at 8 15 42 AM


dju90:
Excited to try it out!

apps/src/music/blockly/FieldTune.ts
```diff
@@ -0,0 +1,217 @@
...
+        this.fieldGroup_
+      );
+
+    this.updateSize_();
+  }
+
+  applyColour() {
+    const style = (this.sourceBlock_ as BlockSvg).style;
+    if (this.borderRect_) {
+      this.borderRect_.setAttribute('stroke', style.colourTertiary);
```
thomasoniii, line 87: 🇬🇧 😆

apps/src/music/blockly/FieldTune.ts
```diff
@@ -0,0 +1,217 @@
...
+  private disposeDropdown() {
+    this.newDiv = null;
+  }
+
+  private getTruncatedNotes(notes: number[]): string {
+    const allNotes = notes
+      .map(note => getNoteName(note))
+      .slice(0, MAX_DISPLAY_NOTES)
+      .join(', ');
+    return notes.length > MAX_DISPLAY_NOTES ? allNotes + '...' : allNotes;
```
thomasoniii, line 213: I don't think we should open this can of worms atm, but just noting that the hardwired ellipsis here may cause localization issues. It's not universally used in all languages, and in RTL languages this would end up showing up on the wrong side of the list.

I'm not sure the best approach. Break it out into a formatted template that can be rtl'ed? make it an actually localizable string? Use the unicode glyph? I dunno.

Here's a little info:
https://stackoverflow.com/questions/48718469/how-is-ellipsis-localized

Prob reasonable to defer, but wanted to note it.

apps/src/music/views/TunePanel.tsx
```diff
@@ -0,0 +1,218 @@
...
+  ]);
+
+  // Generate an array of notes.
+  const arrayOfNotes = Array.from(
+    {length: NUM_OCTAVES * 12},
+    (_, i) => i + START_OCTAVE * 12
+  );
+
+  // Generate an array containing tick numbers from 1..16.
+  const arrayOfTicks = Array.from({length: 16}, (_, i) => i + 1);
```
thomasoniii, line 152: It's probably not critical, but these two should probably be in a `useMemo`

thomasoniii (approved):
noted a few super minor nits, but otherwise lgtm 🚀

apps/src/music/blockly/FieldTune.ts
breville: Yes, Blockly [uses](https://developers.google.com/blockly/guides/create-custom-blocks/block-colour) `colour`!

sanchitmalhotra126 (commented):
This is awesome! So cool to finally see a proper piano roll editor in music lab 🙂

Code-wise, I'd be curious to see if we could consolidate some of the interfaces and views between the drum pattern panel and this UI, but that's definitely something that can be addressed as a follow-up, especially as we play around with this more.

PR #60162, opened 2024-08-01
Fix new version logic in lab2

molly-moen:
In old labs, we decide to make a new version based on the following logic:
- If we set a flag to force a new version to true, create a new version.
- Otherwise, if we haven't saved yet this session, make a new version
- Otherwise, if it has been more than 15 minutes since we last made a new version, make a new version

You can see the original logic [here](https://github.com/code-dot-org/code-dot-org/blob/f525839caca03a0ead96e1df09ab1a0628f9c3e6/apps/src/code-studio/initApp/project.js#L1098-L1105).

We tried to recreate this logic in lab2, but it did not work correctly. We were comparing to the last saved time, not the last new version time, and we had the boolean flipped. Therefore we were always creating a new version. This led to very long version history lists once I started working on version history in lab2.

I updated `SourcesStore` to use the correct logic. It got me tripped up a bit because the flag we send to the backend is `replace`, which is false is we should create a new version, and true if we should not create a new version (it indicates whether we should replace the existing version or not). That, along with tracking the incorrect save time, probably contributed to the bug before. I improved the function names and added comments to make this clearer. We also previously had a `replace` flag one could send, but that would only allow you to force replacing an existing version, not creating a new version. I changed this flag to be `forceNewVersion` instead, and updated the logic accordingly.

## Links

- jira ticket: [CT-711](https://codedotorg.atlassian.net/browse/CT-711)

## Testing story
Tested locally. The project manager is the only user of the `save` method, and it never used the previous `replace` flag, so it was safe to update it. Now we only create a new version on load and every 15 minutes.

## PR Checklist:

- [ ] Code is well-commented


breville (approved):

apps/src/lab2/projects/SourcesStore.ts
```diff
@@ -66,13 +66,23 @@ export class RemoteSourcesStore implements SourcesStore {
...
+    forceNewVersion = false
   ) {
     let options = undefined;
     if (this.currentVersionId) {
+      // If forceNewVersion is set to true, we will not replace the existing version (i.e., we will create
+      // a new version). Otherwise we check if we should replace the existing version based on the last new
+      // version saved in this session.
+      const shouldReplaceExistingVersion = forceNewVersion
+        ? false
+        : this.shouldReplaceExistingVersion();
```
mikeharv: I'm finding it a little bit confusing that we have a class method and a local boolean with the same name. Could we consider renaming one or the other? Perhaps the const above could be `versionReplacementRequired` or something?

Secondly, and more subjectively, I find ternaries harder to read than a simple `&&` chain. I wonder if `!forceNewVersion && this.shouldReplaceExistingVersion()` is more readable while being equivalent. No need to change this if you prefer the current way!

mikeharv (approved):
Nice fix! One suggestion for a rename, but otherwise LGTM

apps/src/lab2/projects/SourcesStore.ts
molly-moen: I'll rename it! 
Doing these booleans broke my brain (I think because `false` means `create a new version`). I personally found the ternary easier to understand while writing this because if `forceNewVersion` is true we don't care what `shouldReplaceExistingVersion` is. But I know others prefer chains in general, since they are equivalent I'll update it!

mikeharv: > if `forceNewVersion` is true we don't care what `shouldReplaceExistingVersion` is

This is actually exactly how I interpret the chain as well. For example, with `a && b` JS never bothers to evaluate `b` if `a` is false. I think others (maybe @ebeastlake?) have disagreed with which is more intuitive. Perhaps it's worth more discussion, but certainly not in any way that should hold up this PR!

molly-moen: updated to use a chain, it's definitely the more "standard" way 😄

apps/src/lab2/projects/SourcesStore.ts
```diff
@@ -73,16 +73,15 @@ export class RemoteSourcesStore implements SourcesStore {
       // If forceNewVersion is set to true, we will not replace the existing version (i.e., we will create
       // a new version). Otherwise we check if we should replace the existing version based on the last new
       // version saved in this session.
-      const shouldReplaceExistingVersion = forceNewVersion
-        ? false
-        : this.shouldReplaceExistingVersion();
-      if (!shouldReplaceExistingVersion) {
+      const replaceExistingVersion =
```
mikeharv, line 76: 👍

PR #60178, opened 2024-08-02
Music: report pack ID selection to Amplitude

sanchitmalhotra126:
Records analytics when a user chooses a pack in the pack dialog, as well as storing the currently selected pack whenever it changes. This also includes the situation where the user is opening a project that already has a pack selected, so we can track what pack they used for that session.

The pack selection is reported as a 'button click' event, and the final chosen pack is also reported as part of the session end payload in amplitude.

**Button click event:**
Screenshot 2024-08-02 at 2 01 12 PM

**Additional metadata in the session end event:**
Screenshot 2024-08-02 at 2 01 46 PM

## Links

https://codedotorg.atlassian.net/browse/LABS-901

## Testing story

Tested with standalone music projects, as well as script levels. For script levels (like the onboarding progression) where no pack ID is specified, no pack ID is reported with the final payload.

### DIFF
--- a/apps/src/music/analytics/AnalyticsReporter.ts
+++ b/apps/src/music/analytics/AnalyticsReporter.ts
@@ -50,20 +50,21 @@ interface BlockStats {
   maxTriggerBlocksWithCode: number;
 }
 
-interface Session {
-  startTime: number;
-  soundsUsed: Set<string>;
-  soundsPlayed: {[id: string]: number};
+interface CommonSessionFields {
   blockStats: BlockStats;
   featuresUsed: {[feature: string]: boolean};
+  soundsPlayed: {[id: string]: number};
+  selectedPack?: string;
 }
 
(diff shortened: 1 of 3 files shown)


apps/src/music/analytics/AnalyticsReporter.ts
```diff
@@ -181,6 +182,18 @@ export default class AnalyticsReporter {
...
 
+  setSelectedPack(packId: string | undefined) {
+    if (!this.session) {
+      this.log('No session in progress');
+      return;
+    }
+    this.session.selectedPack = packId;
+  }
+
+  onPackSelected(packId: string) {
```
sanchitmalhotra126, line 193: @breville as per Slack discussion, created a separate function for reporting pack ID selection that GA4 can also use

breville (approved):

PR #60183, opened 2024-08-02
Set up Teacher Page for new Sign Up flow

TurnerRiley:
Sets up the 'Finish Signing up' page for teachers. The page currently stores all inputs in `sessionStorage`.

### New finish sign-up page for teachers
Capture

#### Demo of using fields and seeing them in `sessionStorage`
https://github.com/user-attachments/assets/31bc8361-2992-4662-9384-a27031a529e4

#### Page scales with screen size
Scales to phone size

#### Right-to-left view
I didn't see a right-to-left view in the Figma, so let me know if any styles should be updated!
rtl_last

### Updates to school association styles
I updated some of the styles in the stylesheet for the school association content because a lot of its padding and spacing was not lining up with how the school association looks like in the sign-up flow Figma. The updates barely tweak the look of the old sign-up flow while fixing the look of the school association in the new sign-up flow. The alternative I saw was to pass down overriding styles to the relevant components but that got messy and convoluted quickly, so I figured this was simpler.

#### School association in old sign-up flow
old look

#### School association in new sign-up flow
new school ass

## Links
Jira ticket: [here](https://codedotorg.atlassian.net/jira/software/c/projects/ACQ/boards/64?assignee=60d62161f65054006980bd71&selectedIssue=ACQ-1987
Figma: [here](https://www.figma.com/design/88sbry7xnl1wcyrOgbgFtn/Sign-Up-Flow?node-id=2911-1878&t=vyYFZowJjLHB9ch4-0)
Spec: [item 6.1 in this doc](https://docs.google.com/document/d/1-Bxt7orcLUsc6sHR8dKYj3XmfMNs3pG7AK8j0xeVIG8/edit#heading=h.yl293hqzpe0s)

## Testing story

## Follow-up work
- Combining stylesheets with other new sign-up pages
- Linking the new sign-up pages together
- Adding functionality to the finish sign-up button


apps/src/templates/school-association.module.scss
```diff
@@ -54,7 +55,7 @@
   flex-direction: column;
 
   .padding {
-    padding-top: 0.75rem;
+    padding-top: 1.125rem;
```
TurnerRiley, line 54: These style changes align the school association section with the [sign-up flow Figma](https://www.figma.com/design/88sbry7xnl1wcyrOgbgFtn/Sign-Up-Flow?node-id=2911-2780&t=vyYFZowJjLHB9ch4-0) (keeping it super similar to the current look, so this will cause small eyes diffs but will allow both to look right)

apps/src/sites/studio/pages/devise/registrations/_finish_sign_up.js
```diff
@@ -236,7 +236,7 @@ $(document).ready(() => {
   function renderSchoolInfo() {
     if (schoolInfoMountPoint) {
       ReactDOM.render(
-        <div style={{padding: 10}}>
+        <div style={{padding: 22}}>
```
TurnerRiley, line 239: This div has its own in-line styling and `SchoolDataInputs` adds additional padding on top of that, so I combined them here and removed it from the stylesheet to make the school association look right on the new sign-up page (without this change, the school association had funky-looking spacing).

apps/src/signUpFlow/FinishTeacherAccount.tsx
```diff
@@ -0,0 +1,109 @@
...
+            {locale.stay_updated_email_opt_in_footnote()}
+          </BodyThreeText>
+        </div>
+      </div>
+      <div className={style.finishSignUpButtonContainer}>
+        <Button
+          className={style.finishSignUpButton}
+          color={buttonColors.purple}
+          type="primary"
+          onClick={() => console.log('FINISH SIGN UP')}
```
TurnerRiley, line 94: Currently just a placeholder `log` statement for the button. This will be filled in when the other sign-up pages are all put together!

apps/i18n/signup/en_us.json
bethanyaconnor, line 1: Can you ping Jorge to let him know a few more strings are coming in here?

apps/src/signUpFlow/FinishTeacherAccount.tsx
```diff
@@ -0,0 +1,109 @@
...
+  BodyThreeText,
+} from '@cdo/apps/componentLibrary/typography';
+import locale from '@cdo/apps/signup/locale';
+import SchoolDataInputs from '@cdo/apps/templates/SchoolDataInputs';
+
+import Checkbox from '../componentLibrary/checkbox/Checkbox';
+
+import style from './finish-teacher-account.module.scss';
+
+const USER_NAME_SESSION_KEY = 'username';
```
bethanyaconnor: tiny tiny nit: should this be `userName` as "user" and "name" are separated by underscores?

This is _definitely_ not worth changing if nothing else needs to change 😅

apps/src/signUpFlow/FinishTeacherAccount.tsx
bethanyaconnor, line 1: Should there be tests, particularly for storing the fields in session data?

TurnerRiley: Yes good point! I had thought there were tasks to add tests later but I'm now realizing those are UI tests so I'll ad some front-end tests!

TurnerRiley: As part of this, I made a constants file for session storage keys (and I pulled in the ones already being used for school association since I figured it could make sense to have school association as part of signup constants (since associating schools will usually happen surrounding account creation I imagine?). But I'm open to alternative methods if this doesn't seem right!

apps/src/templates/SchoolDataInputs.jsx
```diff
@@ -33,7 +33,7 @@ export default function SchoolDataInputs({
     detectedCountry === US_COUNTRY_CODE
   );
   const [isOutsideUS, setIsOutsideUS] = useState(
-    detectedCountry && detectedCountry !== US_COUNTRY_CODE
```
TurnerRiley, line 36: While testing, I found a bug that if `usIp` was passed in as `false`, `isOutsideUS` would still return `false` because it would process `detectedCountry` as `''` (meaning the school name field wouldn't automatically show up). This handles the case now for when `usIp` is `false`, rather than treating `false` and `undefined` as the same for `usIp`.

apps/src/signUpFlow/FinishTeacherAccount.tsx
```diff
@@ -0,0 +1,107 @@
...
+      <div className={style.inputContainer}>
+        <BodyTwoText visualAppearance={'heading-xs'}>
+          {locale.what_do_you_want_to_be_called()}
+        </BodyTwoText>
+        <TextField
+          name="userName"
+          onChange={onNameChange}
+          value={name}
+          className={style.nameInput}
+          placeholder={'Dr. Joanna'}
```
TurnerRiley: I figured since this was a name we wouldn't want to translate it, but it does also contain "Dr." in it so I'm unsure. Thoughts on this?

bethanyaconnor: This could come as a follow-up, but I'd translate this. "Dr. Joanna" isn't referring to a real person, just the general idea of a placeholder name

bethanyaconnor: But thinking on this, is this an appropriate placeholder for "userName"? It's more of a display name, no? (This might be something to address in this PR)

bethanyaconnor (approved):
One question about the placeholder but looks good to me!

apps/src/signUpFlow/FinishTeacherAccount.tsx
TurnerRiley: This was just the name in the Figma. @dmcavoy @moshebaricdo Thoughts on this?

moshebaricdo: @bethanyaconnor can you clarify about the username aspect? Since this field was intended for display name, I think it makes sense to use a placeholder that indicates this is where they enter what students are seeing and not necessarily their full name and/or a username like user123. I think in the student flow, Dani suggested something like "Coder" or "student," maybe we can do something similar here "Mr/s. Coder?"

bethanyaconnor: `username` is a property of users in our database that has specific requirements. It is not what is shown on the top right of the page. For example, the username for my fake teacher account is `bethany_teacher369`.

So what I'm getting at is that I'd like to see this input renamed to `displayName` or something similar, if possible!

bethanyaconnor: Sorry for not being clear -- I didn't connect this to the screenshots in the description and thought we were letting users choose their username on sign up 🤦‍♀️

TurnerRiley: Final decision was to use `displayName` over `userName` and to have the placeholder be `Ms. Coder` and translate it.

PR #60184, opened 2024-08-02
Report eyes diff as a different metric to the cloudwatch dashboard.

vijayamanohararaj:
Eyes diffs are not considered as errors and hence not reported as runner failures to cloud watch metrics. Although this is true in the ideal case, we have many instances where eyes diffs tend to be flaky and needs to be surfaced for appropriate prioritization.

Fix: When UI tests fail due to eyes diffs, emit that as a different metric for seeing trends in the dashboard. This would contain valid/expected diffs to start with, but once we do the work to ensure eyes diffs are accepted as part of the PRs, this would only report on flaky issues.

## Links

JIRA https://codedotorg.atlassian.net/browse/TEACH-1227

## Testing story

Validated the change by running eyes test locally and seeing metrics show up for an intentionally introduced diff
- [Cloudwatch dashboard with metrics emitted in local testing](https://us-east-1.console.aws.amazon.com/cloudwatch/home?region=us-east-1#metricsV2?graph=~(metrics~(~(~'Infrastructure~'runner_feature_eyes_diff~'environment~'development~'feature_name~'features*2feyes~'target_browser~'LocalBrowser~'test_type~'Eyes~'is_continuous_integration_run~'false))~sparkline~false~view~'table~stacked~false~region~'us-east-1~stat~'Average~period~300)&query=~'*7bInfrastructure*2cenvironment*2cfeature_name*2cis_continuous_integration_run*2ctarget_browser*2ctest_type*7d*20runner_feature_eyes_diff)
- [Diffs in applit tools](https://eyes.applitools.com/app/test-results/00000251679409808080/?accountId=mG4DoRfv0ntEgI3XIYBulrKxlZqrHtGwbP3V103ojvLFU110)

## Deployment strategy

Regular DTP

## Follow-up work

None

## Privacy

N/A

## Security

N/A

## Caching

N/A

## PR Checklist:

- [x] Code is well-commented

### DIFF
--- a/dashboard/test/ui/runner.rb
+++ b/dashboard/test/ui/runner.rb
@@ -343,10 +343,7 @@ def run_tests(env, feature, target_platform, arguments, log_prefix)
                         feature_name: feature.include?(".feature") ? feature.split(".feature")[0] : feature,
                         target_browser: target_platform}
     # Metrics for individual feature runs. They will be flushed once all of them run
-    unless eyes_succeeded
-      Infrastructure::Logger.put("runner_feature_eyes_diff", 1, extra_dimensions)
-    end
-
+    Infrastructure::Logger.put("runner_feature_eyes_diff", 1, extra_dimensions) unless eyes_succeeded
     Infrastructure::Logger.put("runner_feature_success", cucumber_succeeded ? 1 : 0, extra_dimensions)
     Infrastructure::Logger.put("runner_feature_failure", cucumber_succeeded ? 0 : 1, extra_dimensions)
     Infrastructure::Logger.put("runner_feature_execution_time", duration, extra_dimensions)
--- a/dashboard/test/ui/runner.rb
+++ b/dashboard/test/ui/runner.rb
@@ -343,6 +343,10 @@ def run_tests(env, feature, target_platform, arguments, log_prefix)
                         feature_name: feature.include?(".feature") ? feature.split(".feature")[0] : feature,
                         target_browser: target_platform}
     # Metrics for individual feature runs. They will be flushed once all of them run
+    unless eyes_succeeded
+      Infrastructure::Logger.put("runner_feature_eyes_diff", 1, extra_dimensions)
+    end
+
     Infrastructure::Logger.put("runner_feature_success", cucumber_succeeded ? 1 : 0, extra_dimensions)
     Infrastructure::Logger.put("runner_feature_failure", cucumber_succeeded ? 0 : 1, extra_dimensions)
     Infrastructure::Logger.put("runner_feature_execution_time", duration, extra_dimensions)
--- a/dashboard/test/ui/runner.rb
+++ b/dashboard/test/ui/runner.rb
@@ -344,8 +344,8 @@ def run_tests(env, feature, target_platform, arguments, log_prefix)
                         target_browser: target_platform}
     # Metrics for individual feature runs. They will be flushed once all of them run
     Infrastructure::Logger.put("runner_feature_eyes_diff", 1, extra_dimensions) unless eyes_succeeded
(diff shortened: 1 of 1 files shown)


dashboard/test/ui/runner.rb
```diff
@@ -343,6 +343,10 @@ def run_tests(env, feature, target_platform, arguments, log_prefix)
                         feature_name: feature.include?(".feature") ? feature.split(".feature")[0] : feature,
                         target_browser: target_platform}
     # Metrics for individual feature runs. They will be flushed once all of them run
+    unless eyes_succeeded
+      Infrastructure::Logger.put("runner_feature_eyes_diff", 1, extra_dimensions)
+    end
+
```
Hamms: I'd suggest either making this a one-liner like so:

```suggestion
      Infrastructure::Logger.put("runner_feature_eyes_diff", 1, extra_dimensions) unless eyes_succeeded
```

Or maintaining parallelism with the other logs by conditionally logging 0 or 1 like so:

```suggestion
      Infrastructure::Logger.put("runner_feature_eyes_diff", eyes_succeeded ? 1 : 0, extra_dimensions)
```

Hamms (approved):
I have one minor nitpick about implementation details, but overall this LGTM! Thanks

cat5inthecradle (dismissed):

dashboard/test/ui/runner.rb
cat5inthecradle: I definitely agree we should be logging 0's

vijayamanohararaj: Making the one liner sounds good to me. 

Can you pl help me see the value of logging 0's though. In case of no eyes diffs, those are reported as `runner_feature_success` event. Also, I'm curious how the 0's help in general, is there a cloudwatch dashboard method that can help with computing % by emitting both 0's and 1's for complementary events?

davidsbailey (approved):

dashboard/test/ui/runner.rb
cat5inthecradle: You're logging 0's for `runner_feature_failure` and `runner_feature_success`, so for consistency, shouldn't you do the same here? (or make those also follow the `log 1 unless condition` pattern?

cat5inthecradle: This is the inconsistency,

```
Infrastructure::Logger.put("runner_feature_eyes_diff", 1, extra_dimensions) unless eyes_succeeded
Infrastructure::Logger.put("runner_feature_success", cucumber_succeeded ? 1 : 0, extra_dimensions)
Infrastructure::Logger.put("runner_feature_failure", cucumber_succeeded ? 0 : 1, extra_dimensions)
```
 
Should you do:

```
Infrastructure::Logger.put("runner_feature_eyes_diff", 1, extra_dimensions) unless eyes_succeeded
Infrastructure::Logger.put("runner_feature_success", 1, extra_dimensions) if cucumber_succeeded
Infrastructure::Logger.put("runner_feature_failure", 1, extra_dimensions) unless cucumber_succeeded
```

vijayamanohararaj: The metrics for `runner_feature_success` and `runner_feature_failure` existed earlier, so I am not sure what the intention was behind following the 0/1 approach. From my research on the various methods that cloud watch provides to plot these metrics, I don't see a way to leverage these duplicate metrics.

So, I'm inclined to change those also to avoid publishing duplicate metrics. Sending another iteration with that update.

cat5inthecradle: Yeah, I think we'll get more value from just sending 1's when the event happens, and not sending otherwise.

cat5inthecradle (approved):

PR #60185, opened 2024-08-02
Fix flaky Java Lab tests

molly-moen:
This PR fixes two flaky Java Lab tests, `code_review_scenarios` and `code_review_finish_button`. They were flaky for two reasons:
1. We have a general issue with UI tests where if you log in, log out, then log in with another user you can end up still logged in as the first user. When this happened, we would try to access the teacher dashboard with a student account and time out due to a 404.
2. We were also seeing an issue where when trying to view another student's code review there would be no students to review. I believe the issue here was closing the code review groups creation modal before it finished saving, so I added a check that saving completed. There were also cases where only 1 student got added to the group; I added a check that the group exists before starting to add students to it.

The issue in 1 is the subject of a very long [slack thread](https://codedotorg.slack.com/archives/C0T0PNR0D/p1722444400850089). The TL;DR is that we think there is some sort of race condition happening. While Platform figures that out, we can add a wait between log out and log in to fix the issue.

After applying the two fixes, I ran each failing test 4 times. Both passed 4 times in a row. Before the fix, at least 1 out of every 3 runs would fail.

## Links
- jira tickets: [CT-706](https://codedotorg.atlassian.net/browse/CT-706), [CT-705](https://codedotorg.atlassian.net/browse/CT-705)

## Testing story
Tested using a saucelabs tunnel.

## Follow-up work
Platform has a [ticket](https://codedotorg.atlassian.net/browse/P20-1080) for a long-term fix for this race condition.

## PR Checklist:

- [ ] Code is well-commented


ebeastlake (approved):
Nice work, @molly-moen, and thanks so much for the help, @stephenliang and others!

dashboard/test/ui/features/step_definitions/steps.rb
```diff
@@ -1484,6 +1484,7 @@ def refute_bad_gateway_or_site_unreachable
   student_count.to_i.times do |i|
     add_student_step_list.push("Given I create a student named \"student_#{i}\"")
     add_student_step_list.push("And I join the section")
+    add_student_step_list.push("And I wait for 3 seconds")
```
stephenliang, line 1487: Do we know how often the bug affects other tests? Wondering if we should apply the wait globally as part of the `reset_session` logic. This can be a follow up though, looks good!

stephenliang (approved):

dashboard/test/ui/features/step_definitions/steps.rb
molly-moen: I'm not sure and didn't want to apply that broad of a change unless we were sure we needed it.

PR #59565, opened 2024-07-03
codemod chai and sinon to use jest equivalents

stephenliang:
This PR runs the jest codemod utility to bulk convert chai and sinon based unit tests to the equivalent jest syntax. For tests that failed, the tests are marked with an exception to allow continued usage of chai and sinon but new tests must be created using jest only.

The heuristic is as follows:

1. Run sinon and chai codemods
2. lint:fix
3. Run tests, mark those that failed
4. Restore failed tests from staging
5. Restrict chai and sinon from being imported as eslint rule
6. Any remaining sinon/chai due to failures are excluded (new tests must be written in jest syntax)

## Links

- [design](https://docs.google.com/document/d/1-U41r5O8Y404HeWHOAn2vxUI6G1lSJbtK7xi4j2eZsA/edit#heading=h.zfzb7h4ccnko)

## PR Checklist:

- [ ] Tests provide adequate coverage
- [ ] Privacy and Security impacts have been assessed
- [ ] Code is well-commented
- [ ] New features are translatable or updates will not break translations
- [ ] Relevant documentation has been added or updated
- [ ] User impact is well-understood and desirable
- [ ] Pull Request is labeled appropriately
- [ ] Follow-up work items (including potential tech debt) are tracked and linked

PR #60191, opened 2024-08-03
Add ChatEventLogger unit tests

fisher-alice:
This is a follow-up to https://github.com/code-dot-org/code-dot-org/pull/60110 which added `ChatEventLogger`.
This PR adds 2 unit tests for the `ChatEventLogger`.

I haven't added apps unit tests recently so initially used chai, but then used jest since I saw the 'eslint-disable-line no-restricted-imports' exception. Reference: https://github.com/code-dot-org/code-dot-org/pull/59565.

## Links
[jira](https://codedotorg.atlassian.net/browse/LABS-958)

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/test/unit/aichat/ChatEventLoggerTest.ts
+++ b/apps/test/unit/aichat/ChatEventLoggerTest.ts
@@ -22,9 +22,6 @@ describe('ChatEventLogger', () => {
       channelId: 'abc123',
     };
     chatEventLogger = ChatEventLogger.getInstance();
-    postLogChatEventSpy = jest
-      .spyOn(aichatApi, 'postLogChatEvent')
-      .mockResolvedValue({chat_event_id: 1, chat_event: userChatMessage});
   });
 
   afterEach(() => {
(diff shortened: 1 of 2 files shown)


apps/test/unit/aichat/ChatEventLoggerTest.ts
```diff
@@ -0,0 +1,55 @@
...
+  });
+
+  it('logChatEvent calls on postLogChatEvent', async () => {
+    chatEventLogger.logChatEvent(userChatMessage as ChatMessage, aichatContext);
+    expect(postLogChatEventSpy).toHaveBeenCalledTimes(1);
+  });
+
+  it('postLogChatEvent not called when sendingInProgress is true', async () => {
+    // Simulate a sending in progress state.
+    chatEventLogger.setSendingInProgress(true);
```
sanchitmalhotra126: Instead of manually setting `sendingInProgress` to true, could we stub the `postLogChatEvent` function to just wait for some time inside a promise? That would more realistically simulate a network request and allow us to actually test the logic of changing the flag while sending is in progress.

fisher-alice: Yes - good call! I knew there was probably a better way to test this. Thanks for your help!

apps/test/unit/aichat/ChatEventLoggerTest.ts
```diff
@@ -0,0 +1,66 @@
...
+      chatMessageText: 'hello',
+      status: 'OK',
+      timestamp: Date.now(),
+    };
+    aichatContext = {
+      currentLevelId: 123,
+      scriptId: 321,
+      channelId: 'abc123',
+    };
+    chatEventLogger = ChatEventLogger.getInstance();
```
sanchitmalhotra126, line 24: Would be good to create an instance of ChatEventLogger for testing (which would require making the constructor non-private), or mock the getInstance() method to return an instance created within the scope of this test. Just to prevent any weirdness if the actual ChatEventLogger singleton is used in other tests at some point.

sanchitmalhotra126 (approved):
Nice! Thanks for adding tests!

apps/test/unit/aichat/ChatEventLoggerTest.ts
fisher-alice: Make sense - will do. Thanks!

fisher-alice: I'll go ahead and merge as-is and then update in https://github.com/code-dot-org/code-dot-org/pull/60110.

fisher-alice: https://github.com/code-dot-org/code-dot-org/pull/60110/commits/46db50fd226107d81da81a52de11211de1977d2c

PR #60192, opened 2024-08-03
[JIMT][BETTER CODEBRIDGE DIALOGS] - removed window functions

thomasoniii:
Here are the tl;dr highlights:
- retools lab2 dialogs for stricter interface and easier additions
- adds new `GenericDialog` root component which accepts confirm, cancel, and neutral buttons.
- updates other uses of `showDialog` to the new interface
- adds a prompt type
- moves away from window functions to built in dialog prompts.

The existing `dialogControl.showDialog` function had the following signature:

```
    dialogControl.showDialog(
      dialogType: DialogType,
      callback: () => void,
      title?: string,
      message?: string,
      confirmText?: string
    )
```

but it was kinda confusing and limited.

For starters, a list of 5 unnamed positional arguments is difficult to keep track of. Secondly, it accepts a `dialogType` argument, which (to me at least) implies that all of these arguments can be handed into all dialog components. That's not actually true - for example, `SkipDialog` will ignore the `title`, `message`, and `confirmText` args and they shouldn't be supplied. Finally, there is simply more functionality that _could_ be exposed to these dialogs that isn't. There's no way to change the text on the `cancel` button, for example.

So it's retooled. Dialogs now have as their root a `GenericDialog`, which takes the following props:
```
  // you can hand in strings for title/message OR full react nodes for a more complicated display
  title : string | titleComponent;
  message : string | bodyComponent

  // there are three defined button types - confirm (far right), cancel (far left) and neutral (far right, to the left of confirm, if it's present)
  // each of these buttons can accept text to place on the button (default text will be used for cancel or confirm, but neutral requires it), as well as an optional callback to invoke when the button is clicked.
  // note that the callback is optional, and if not supplied, it will simply close the dialog.
  buttons : {
    confirm | cancel | neutral : { text? : string, callback? : function }
  }
```

`dialogControl.showDialog` is now more strictly typed to only allow in args for an appropriate dialog. For example, any dialog could be invoked as follows:

```
dialogControl.showDialog({
  dialogType : DialogType.GenericDialog,
  title : "This is a generic dialog",
  message : "Just a dialog to test with",
  buttons : {
    confirm : { text : "Confirm button", callback : () => console.log("I hit confirm") },
    cancel : { text : "cancel button", callback : () => console.log("I hit cancel") },
    neutral : { text: "My action is ambiguous. I need a good label", callback : () => console.log("I hit neutral) }
  }
});
```

Further, the canned dialog types (`SkipDialog` and `StartOverDialog` were pre-existing, and I've added `GenericAlert` and `GenericPrompt`) are strictly typed for only the args they care about. e.g.

```
// previously, you would do this for a skip dialog:
dialogControl.showDialog(DialogType.StartOver, resetProject);

// whereas now you do this:
dialogControl.showDialog({
  type: DialogType.StartOver,
  handleConfirm: resetProject,
});
```

The new `GenericAlert` is as follows:
```
dialogControl?.showDialog({
  type: DialogType.GenericAlert,
  title: message,
});
```

This dialog takes no action, it merely informs the user. So there is no need for any callbacks.

The new prompt dialog allows the user to type in a value:
```
dialogControl?.showDialog({  
  type: DialogType.GenericPrompt,
  title: codebridgeI18n.moveFilePrompt(),
  handleConfirm: destinationFolderName => {
    // the value typed into the prompt text field will be returned to the callback.
    console.log(`The user typed in ${destinationFolderName}`);
  },
});
```

And that's it. The rest of the PR is just all the plumbing needed to make this all work, and refactoring on existing calling of the function.

There are several future to-do items that I want to circle back to:
- The `type` argument to `showDialog` is superfluous - since different dialogs should accept different types of prompts, we can't _really_ just swap the type and get different ones. This should be refactored away and the `dialogControlContext` given separate functions for each type. (e.g., `showAlert`)
- Ideally, registration of different dialogs shouldn't be statically handled within the `DialogManger` itself and each dialog should be able to call a register function so it's more dynamic.
- You can add a button by adding an empty object to `buttons`, e.g. `buttons : { cancel : {} }` will add a button with the default text and no callback (which is totally valid), but it's clunky devX and should be re-considered.
- Handling responses via callbacks is a little clunky, so I'm going to retool it so it can be used synchronously (e.g. `const value = await dialogControl.showPrompt( { title : "Type in your name" } )`), but that's a more involved change and I wanted to get this in first.
- A discussion point on the above is whether we want to have the callbacks _at all_ or make all interactions async. Once I've got the interface in place, I feel like we should discuss further.
- There are some shortcomings in the prompt dialog especially - it's not autofocusing the text field, and the buttons aren't focusing yet either. That should be addressed.

But that's about it. Thoughts? I'm happy to set up calls with anyone interested to talk through it, but it's not as much going on as the size of the PR may imply. 

## PR Checklist:

- [ ] Code is well-commented


apps/src/aichat/views/AichatView.tsx
```diff
@@ -169,11 +166,14 @@ const AichatView: React.FunctionComponent = () => {
...
-  const dialogControl = useContext(DialogContext);
+  const dialogControl = useDialogControl();
 
   const onClickStartOver = useCallback(() => {
     if (dialogControl) {
-      dialogControl.showDialog(DialogType.StartOver, resetProject);
+      dialogControl.showDialog({
+        type: DialogType.StartOver,
+        handleConfirm: resetProject,
+      });
```
thomasoniii, line 184: The changes to the non-codebridge, non-dialog files are identical - just changing to the new named args and the new wrapper hook to provide the context.

apps/src/codebridge/FileBrowser/index.tsx
```diff
@@ -13,28 +13,30 @@ import {
...
-  DialogType,
-} from '@cdo/apps/lab2/views/dialogs/DialogManager';
+import {useDialogControl, DialogType} from '@cdo/apps/lab2/views/dialogs';
 import {useAppSelector} from '@cdo/apps/util/reduxHooks';
 
 import {FileBrowserHeaderPopUpButton} from './FileBrowserHeaderPopUpButton';
 import {
   downloadFileType,
   moveFilePromptType,
+  moveFileType,
```
thomasoniii, line 31: This is an example of something I want to get rid of by moving it to async - it needed restructuring to allow the prompted values to be accessed via a callback, and I opted to move them into separate functions instead of inside the callback. It's gonna be less work to unwind this later, it's just a little ugly atm.

apps/src/codebridge/FileBrowser/index.tsx
```diff
@@ -282,28 +284,79 @@ export const FileBrowser = React.memo(() => {
     setFileType,
   } = useCodebridgeContext();
   const isReadOnly = useAppSelector(isReadOnlyWorkspace);
+  const dialogControl = useDialogControl();
 
-  const newFolderPrompt: FilesComponentProps['newFolderPrompt'] = useMemo(
```
thomasoniii, line 286: this style of refactoring in this component is just breaking apart into a separate function that displays the prompt, and then the existing function which now takes the prompted value as an argument and other proceeds as normal.

But the gist of the change is that it's moving from `window.prompt` -> `dialogControl.showDialog({type : DialogType.GenericPrompt, ... })`

apps/src/lab2/views/dialogs/DialogControlContext.ts
```diff
@@ -0,0 +1,14 @@
...
+interface DialogControl {
+  closeDialog: closeDialogType;
+  showDialog: (args: TypedDialogProps) => void;
+}
+
+export const DialogControlContext = React.createContext<DialogControl>({
+  closeDialog: () => {},
+  showDialog: () => {},
+});
+export const useDialogControl = () => useContext(DialogControlContext);
```
thomasoniii, line 14: This file is really just moving the `DialogContext` out of `DialogManager` into a separate file, and renaming to `DialogControlContext` for clarity. It also gets a wrapper.

Also, `closeDialog` is now exposed as a function in the context as well.

apps/src/lab2/views/dialogs/DialogManager.tsx
```diff
@@ -45,85 +37,39 @@ const DialogManager: React.FunctionComponent<DialogManagerProps> = ({
...
-    () => null
-  );
-  const [dialogTitle, setDialogTitle] = useState<string | undefined>(undefined);
-  const [dialogMessage, setDialogMessage] = useState<string | undefined>(
-    undefined
-  );
-  const [dialogConfirmText, setDialogConfirmText] = useState<
-    string | undefined
-  >(undefined);
+  const [dialogArgs, setDialogArgs] = useState<AnyDialogType>();
```
thomasoniii, line 40: Since each type of dialog has different args, we just read 'em in generically and pass along the object as appropriate.

apps/src/lab2/views/dialogs/DialogManager.tsx
```diff
@@ -45,85 +37,39 @@ const DialogManager: React.FunctionComponent<DialogManagerProps> = ({
...
-    setOpenDialog(null);
-  }, [setOpenDialog]);
+  const closeDialog = useCallback(() => setOpenDialog(null), [setOpenDialog]);
 
-  const DialogView = openDialog && dialogCallback && DialogViews[openDialog];
+  // Allow the any because if it's NOT any, then line 63 with DialogView's args will toss an error.
+  // Keep this until we have a better solution. ¯\_(ツ)_/¯
+  // The typing on the `showDialog` function ensures the props are correct, so we're still safe'
+  // eslint-disable-next-line
+  const DialogView: any = openDialog && dialogArgs && DialogViews[openDialog];
```
thomasoniii, line 56: If anybody can make typescript happy w/o the explicit `any` cast, that'd be keen. 😠 

It's still type safe because the `showDialog` function is so strict, but it feels icky.

apps/src/lab2/views/dialogs/GenericAlertDialog.tsx
```diff
@@ -0,0 +1,15 @@
...
+  Pick<GenericDialogProps, 'title'>
+> &
+  Pick<GenericDialogProps, 'message'>;
+
+const GenericAlertDialog: React.FunctionComponent<GenericAlertDialogProps> = ({
+  title,
+  message,
+}) => <GenericDialog title={title} message={message} />;
+
+export default GenericAlertDialog;
```
thomasoniii, line 15: All the dialogs wrapper around `GenericDialog` and hand through the props they've received. This keeps all UI centralized in the one component and much easier to change.

apps/src/lab2/views/dialogs/GenericConfirmationDialog.tsx
```diff
@@ -1,50 +1,36 @@
...
+        text: confirmText,
+      },
+      cancel: {
+        callback: handleCancel,
+      },
+    }}
+  />
+);
 
 export default GenericConfirmationDialog;
```
thomasoniii, line 36: Likewise, the pre-existing components were just re-tooled to centralize around `GenericDialog`, and let that component handle all of the layout.

apps/src/lab2/views/dialogs/GenericDialog.tsx
```diff
@@ -0,0 +1,133 @@
...
+type GenericDialogBodyProps =
+  | {
+      message?: never;
+      bodyComponent?: React.ReactNode;
+    }
+  | {
+      message?: string;
+      bodyComponent?: never;
+    };
+
```
thomasoniii, line 32: typing is strict to allow a `title` OR `titleComponent` but not both.

Likewise a `message` or `bodyComponent` but not both.

apps/src/lab2/views/dialogs/GenericPrompt.tsx
```diff
@@ -0,0 +1,66 @@
...
+}) => {
+  return (
+    <TextField
+      name="prompt-field"
+      placeholder={placeholder}
+      value={prompt}
+      onChange={e => setPrompt(e.target.value)}
+    />
+  );
+};
```
thomasoniii, line 38: Is there a prop I can tack onto this thing to get it to span the container? I haven't figured it out yet.

apps/src/lab2/views/dialogs/types.ts
```diff
@@ -0,0 +1,41 @@
...
+  | (SkipDialogProps & {type: DialogType.Skip})
+  | (StartOverDialogProps & {type: DialogType.StartOver});
+
+export type AnyDialogType =
+  | GenericAlertDialogProps
+  | GenericConfirmationDialogProps
+  | GenericDialogProps
+  | GenericPromptProps
+  | SkipDialogProps
+  | StartOverDialogProps;
```
thomasoniii, line 41: just a new types file, mostly moving stuff out of ~dialogManager`. Has the strict typing used for the `showDialog` function.

molly-moen (approved):
This looks great! Big improvement

apps/src/lab2/views/dialogs/DialogManager.tsx
sanchitmalhotra126: Not 100% sure if this would work but -

Since it looks like all Dialogs' props extend `GenericDialogProps`, we should be able to keep the type on `DialogViews` as `{[key in DialogType]: React.FunctionComponent<GenericDialogProps>}`? That way `DialogView` can be treated as a `FunctionComponent<GenericDialogProps>`. Not sure if that'll play nicely with the type of `dialogArgs` but we might be able to use some parameterized types to handle that too.

thomasoniii: Didn't work. Or at least I couldn't get it working.

Opting to just get it merged and we can circle back to it. I hate this stupid any. :angry:

PR #60206, opened 2024-08-05
GenAi: Only render retrieval input area when not read-only.

cnbrenci:
This change removes the input area of the Retrieval tab in Gen AI Units when they are read-only.

Before the change (when retrieval is read-only):
image

After the change:
image

## Links

- jira ticket: [Feature Request: hide text inputs when retrieval is in Read Only](https://codedotorg.atlassian.net/browse/LABS-778)

## Testing story

Checked a level with and without read-only set.

### DIFF
--- a/apps/src/aichat/views/modelCustomization/RetrievalCustomization.tsx
+++ b/apps/src/aichat/views/modelCustomization/RetrievalCustomization.tsx
@@ -61,30 +61,33 @@ const RetrievalCustomization: React.FunctionComponent = () => {
   return (
     <div className={modelCustomizationStyles.verticalFlexContainer}>
       <div className={modelCustomizationStyles.customizationContainer}>
-        <div className={modelCustomizationStyles.inputContainer}>
-          <FieldLabel
-            label="Retrieval"
-            id="retrieval-input"
-            tooltipText="Retrieval lets you add new information for a chatbot to reference. Type in each retrieval statement into the text box, then click “Add” for each one."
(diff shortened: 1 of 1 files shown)


apps/src/aichat/views/modelCustomization/RetrievalCustomization.tsx
```diff
@@ -61,30 +61,33 @@ const RetrievalCustomization: React.FunctionComponent = () => {
...
-            disabled={!newRetrievalContext || isReadOnly}
-          />
-        </div>
+        {!isReadOnly && (
+          <>
+            <div className={modelCustomizationStyles.inputContainer}>
+              <FieldLabel
+                label="Retrieval"
+                id="retrieval-input"
+                tooltipText="Retrieval lets you add new information for a chatbot to reference. Type in each retrieval statement into the text box, then click “Add” for each one."
```
fisher-alice, line 70: Do you mind checking with product/curriculum to see if they would still want the Retrieval title and info icon at the top when read-only? Thanks!

cnbrenci: I had the same thought and checked with Sam already :) see the comments on the jira ticket here https://codedotorg.atlassian.net/browse/LABS-778

fisher-alice: Thanks - LGTM!

PR #60212, opened 2024-08-05
Remove Temporary DCDO Flag for new Lockable Feature

Hamms:
Now that it's been released and we know it's working, we're no longer worried about needing to quickly disable it.

## Links

Follow-up to https://github.com/code-dot-org/code-dot-org/pull/57515


cat5inthecradle (approved):
We don't have an authoritative list of DCDO flags, right? We simply check them from various places around the codebase? Meaning that the list of flags in dynamodb likely includes no-longer-referenced flags?

Hamms:
No authoritative list anywhere that I'm aware of. I believe we also didn't ever actually flip this switch, so there shouldn't even be a record for it in dynamodb.

But yes, any other DCDO flag that was once set in the database but has since been removed from the codebase is likely still set in the database. Would probably be a good idea for us to clean those up at some point

PR #60217, opened 2024-08-06
apps reorg: re-home /lib files, part 1

Erin007:
The `lib` directory of `apps` contained a subset of things, but there wasn't a clear reason for all of those thing to be grouped in there. Instead of maintaining a subfolder without cohesive contents, I'm going through the process of taking everything out of a `lib` and taking inventory of what we have. This is just the first pass. Once components are out of meaningless groupings like `lib` and `templates` we can then choose to re-group them more intentionally. In this PR: 

- I moved the entire `maker` directory out of `lib` one folder up into `src` and then deleted the newly empty `kits` directory from `lib` (side note from the annals of history: `kits` was a relic of a [plan](https://github.com/code-dot-org/code-dot-org/blob/staging/apps/docs/refactor.md#directory-hierarchy) from 8 years ago, but it appears as though `maker` was the only feature set that ever stuck to the "rules") 
- similarly, I moved the entire `levelbuilder` directory out of `lib` into `src`
- I moved `metrics` out of `levelbuilder` and into `src`
- I took the files in `lib/util` and sorted them into new homes 
        - `AnalyticsConstants`, `AnalyticsReporter`, `GoogleAnalyticsReporter`, `StatsigReporter`, and `firehose.js` now live in `apps/src/metrics/utils`

[CONTINUE HERE]


Erin007:
I broke this up into smaller PRs that were easier to reason about and review.

PR #60219, opened 2024-08-06
Delete mysql-metrics Cron Job

sureshc:
We no longer need the custom CloudWatch Metrics published by the `mysql-metrics` cron job because RDS Aurora MySQL. publishes most of these now (or equivalent metrics) and also RDS Performance Insights provides better analysis of database performance.

## PR Checklist:

- [ ] Code is well-commented


bin/cron/mysql-metrics
```diff
@@ -1,377 +0,0 @@
...
-abort 'Script already running' unless only_one_running?(__FILE__)
-
-require_relative '../../deployment'
-require 'cdo/db'
-require 'aws-sdk-cloudwatch'
-
-# Report details MySQL metrics to CloudWatch.
-# Intended to be run at 1-minute intervals.
-
-STATUS_CACHE = pegasus_dir('cache', 'mysql-status-cache.json')
```
Hamms, line 12: Looks like we can remove this line from this directory's gitignore: https://github.com/code-dot-org/code-dot-org/blob/b01bde73c4b4bddde9b61543a5c02e1a4b6c04c0/pegasus/cache/.gitignore#L13

Hamms (approved):
One nit, otherwise LGTM!

bin/cron/mysql-metrics
sureshc: Nice catch (cache?)! Should we automate deletion of this file? When it's no longer ignored, will it get auto-committed by our staging and levelbuilder content push jobs?

Hamms: Probably worth verifying, but I think those jobs are smart enough to only commit content from specific subdirectories. We should plan to clean it up anyway, though; no preferences on whether we do so manually or automatically.

PR #60221, opened 2024-08-06
pycall example: use llm-guard pip

snickell:
# THIS PR DOES NOT WORK IT WILL CRASH

## Replaced by #60335 which is a rebase

---

Short example of using the [llm-guard pip](https://pypi.org/project/llm-guard/). Eventually I'll expand this to include calling it from a rails controller.

### Add deps to pycdo/pyproject.toml and install them
1. `cd python/pycdo`
2. Add the pip into pycdo/pyproject.toml: `pdm add llm-guard`
3. Back to the project root: `cd ../..` 
4. Install the venv again: `pdm install`

### Create python source

See [changed files](https://github.com/code-dot-org/code-dot-org/pull/60221/files) for details, but its just vanilla python at this point:
```
➜  pycdo git:(seth/pycall-llmguard-demo) find . | grep genai | grep py$
./pycdo/genai/__init__.py
./pycdo/genai/llmguard.py
./tests/genai/__init__.py
./tests/genai/test_llmguard.py
```

Now that we've added source, you can run python tests. From /python/pycdo run:
```
pdm run pytest
```

Or if you want to try using them from, say, ipython:
```
pdm run ipython
```

## Alternative approach: create a new genai package

In this PR, I create `llmguard.py` inside the existing pycdo module, and add it to the deps for pycdo. 

If genai (your name here) is going to use more python code than just a little llmguard, I suggest creating a new top-level `genai` package parallel to pycdo. Its not much harder, and I'm happy to help with this, but I just started here with the simplest possible example.


snickell:
Replaced by https://github.com/code-dot-org/code-dot-org/pull/60335 which is a continuation of this PR with a nice rebase, and a better focus. This PR did its job: it shaped our Python.run API to the point that it is now updated in name, etc.

PR #60224, opened 2024-08-06
Extend AI Tutor Tester to get responses from  LLMGuard

Erin007:
[CT-721](https://codedotorg.atlassian.net/browse/CT-721)

For internal testing of the chat safety layer, it will more efficient to get bulk responses back rather than test inputs 1 by 1. In this PR I extend the functionality of the AI Tutor Tester to conditionally be ably to hit the LLMGuard endpoint based on selection from a dropdown.  The message safety check is currently hitting an EC2 instance that has been set up as an endpoint until we have a better long-term solution that will allow us to run Python alongside our Ruby code. This is a very lightweight, temporary solution to unblock testing of LLMGuard. 

The uploaded csv should have one column `studentInput` and the `aiResponse` column will be set as "ok" if LLMGuard deems the input "safe", "toxic" if it catches toxic content and "error" if something in the process goes wrong. 

https://github.com/user-attachments/assets/a84a223f-8304-4907-aff4-2b7669437385


Erin007:
cc: @sanchitmalhotra126 @fisher-alice

snickell (approved):
Assuming we do the followup work (e.g. https://github.com/code-dot-org/code-dot-org/pull/60221) and pull the EC2 instance out of the loop, I'm game to merge this now so @samantha-code can test.

Erin007:
@samantha-code merged ✅

PR #60225, opened 2024-08-06
apps reorg: move maker out of lib and delete kits subdirectory

Erin007:
The /lib directory of apps contains a subset of things, but there isn't a clear reason for all of those things to be grouped in there. Instead of maintaining a subfolder without cohesive contents, I'm going through the process of taking everything out of  /lib and taking inventory of what we have. This is just the first pass, but I wanted the diff to be manageable. 

In this PR I moved the entire maker directory out of /lib one folder up into /src and then deleted the newly empty /kits directory from /lib.  

Up next... 
- move /levelbuilder out of /lib 
- move /metrics out of /lib 
- inventory lib/tools and move contents to new better-suited homes 
- inventory lib/ui and move contents to new better-suited homes 
- inventory lib/util and move contents to new better-suited homes 

Once components are out of meaningless groupings like /lib and /templates we can then choose to re-group them more intentionally.


Erin007:
side note from the annals of history: kits was a relic of a [plan](https://github.com/code-dot-org/code-dot-org/blob/staging/apps/docs/refactor.md#directory-hierarchy) from 8 years ago, but it appears as though maker was the only feature set that ever stuck to the "rules"

PR #60227, opened 2024-08-06
Remove feedback banner for teachers who create accounts after V2 launch

kobryan0619:
Currently, when a new user signs up, they immediately see a feedback banner in the V2 view.  We don't want them to see that banner.  This PR removes that banner for those users.  All other users should see no change in behavior. 

## Links

[Slack convo](https://codedotorg.slack.com/archives/C06ERUABG01/p1722952839776999)
[Ticket](https://codedotorg.atlassian.net/jira/software/projects/TEACH/boards/65?selectedIssue=TEACH-1231)

## Testing story

- Modified unit test
- Tested locally
- Expecting to have eyes tests fail when merged in - will accept new baselines at that point and let DOTD know

### DIFF
--- a/apps/src/templates/sectionProgressV2/ProgressFeedbackBanner.jsx
+++ b/apps/src/templates/sectionProgressV2/ProgressFeedbackBanner.jsx
@@ -21,6 +21,7 @@ const ProgressFeedbackBanner = ({
   fetchProgressV2Feedback,
   createProgressV2Feedback,
   errorWhenCreatingOrLoading,
+  userCreatedAt,
 }) => {
   const [bannerStatus, setBannerStatus] = React.useState(BANNER_STATUS.UNSET);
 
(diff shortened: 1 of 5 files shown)


apps/src/templates/sectionProgressV2/ProgressFeedbackBanner.jsx
```diff
@@ -46,14 +47,18 @@ const ProgressFeedbackBanner = ({
       bannerStatus === BANNER_STATUS.UNSET &&
       progressV2Feedback
     ) {
-      if (progressV2Feedback.empty) {
+      if (
+        progressV2Feedback.empty &&
+        new Date(userCreatedAt) < new Date('2024-08-02')
```
lfryemason: Let's pull this constant into a variable and add a comment to explain why we need it

lfryemason (approved):
Looks good as long as we comment the constant

PR #60229, opened 2024-08-06
Enable generateSharedConstants to output typescript

sanchitmalhotra126:
Enable the `generateSharedConstants.rb` script to optionally output TypeScript files if specified. This will output a file that looks very similar to the JS version but with a .ts extension and `as const` assertions on the values. The latter is particularly useful for using these constants as types themselves (i.e. enforcing that an interface member is one of the values of a given constant).

For now this just updates the `apps/generated-scripts/sharedConstants` file to TypeScript. I figured we can manually convert other files if/when necessary.

Shared constants aren't stored in our repo, but once generated the TypeScript output looks like this:

Screenshot 2024-08-06 at 11 12 05 AM

## Testing story

Tested with some constant types in AI Chat and AI Tutor

### DIFF
--- a/apps/src/aiTutor/views/UserMessage.tsx
+++ b/apps/src/aiTutor/views/UserMessage.tsx
@@ -17,7 +17,7 @@ const PROFANITY_VIOLATION_USER_MESSAGE =
 const PII_VIOLATION_USER_MESSAGE =
   'This chat has been hidden because it contains personal information.';
 
-const statusToStyleMap = {
+const statusToStyleMap: {[status: string]: string} = {
   [Status.OK]: style.userMessage,
   [Status.UNKNOWN]: style.userMessage,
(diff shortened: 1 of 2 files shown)


apps/src/aiTutor/views/UserMessage.tsx
```diff
@@ -17,7 +17,7 @@ const PROFANITY_VIOLATION_USER_MESSAGE =
 const PII_VIOLATION_USER_MESSAGE =
   'This chat has been hidden because it contains personal information.';
 
-const statusToStyleMap = {
+const statusToStyleMap: {[status: string]: string} = {
```
sanchitmalhotra126, line 20: Needed to make the TS compiler happy

molly-moen (approved):
Neat!

PR #60238, opened 2024-08-06
Fixing the slack notification for group to use group specific formatting.

vijayamanohararaj:
Context: Currently any file that fails PDF generation (intermittent google API failures, sharing settings are incorrect, size too large) goes unnoticed until reported by a teacher. The goal of this task is to tag teacher tools on call whenever PDF generation fails, so it can be investigated in a proactive manner.

Issue: The original approach for this was to include <@teacher-tools-on-call> to tag the group, but it was not resulting in a notification from slack. 

Upon investigating further, found that the notification format for groups is different from users when sending messages programmatically. 

Fix: This change implements tagging user groups in slack following this documentation https://api.slack.com/reference/surfaces/formatting#mentioning-groups 

Logs from local validation
`I, [2024-08-06T15:33:22.069814 #646726]  INFO -- : [development] <!subteam^S07FB3XSAR5> Please follow instructions in https://docs.google.com/document/d/1mBY56DeAzrwTM3CVIOFho3azTi9mudE37ZQrVZXxaMA/edit#heading=h.axfu5or8cueg to troubleshoot`

Snippet from API documentation
image

## Links

JIRA: https://codedotorg.atlassian.net/browse/TEACH-1178

## Testing story

Local validation could only confirm the expected message was getting written. Need to further validate in staging once this change is merged.

## Deployment strategy

Regular deploy pipeline

## Follow-up work

None

## Privacy

N/A

## Security

N/A

## Caching

N/A

## PR Checklist:

- [x] Code is well-commented

### DIFF
--- a/dashboard/lib/services/curriculum_pdfs/resources.rb
+++ b/dashboard/lib/services/curriculum_pdfs/resources.rb
@@ -98,7 +98,6 @@ module Services
               "temporary directory contents: #{Dir.entries(pdfs_dir).inspect}",
               color: 'red'
             )
-            # Group ID: S07FB3XSAR5 - teacher-tools-on-call would be notified
             ChatClient.log(
               "Please follow instructions in https://docs.google.com/document/d/1mBY56DeAzrwTM3CVIOFho3azTi9mudE37ZQrVZXxaMA/edit#heading=h.axfu5or8cueg to troubleshoot",
               color: 'red',
@@ -223,9 +222,8 @@ module Services
             color: 'yellow'
           )
 
-          # Group ID: S07FB3XSAR5 - teacher-tools-on-call would be notified
           ChatClient.log(
-            "<!subteam^S07FB3XSAR5> Please follow instructions in https://docs.google.com/document/d/1mBY56DeAzrwTM3CVIOFho3azTi9mudE37ZQrVZXxaMA/edit#heading=h.axfu5or8cueg to troubleshoot",
+            "Please follow instructions in https://docs.google.com/document/d/1mBY56DeAzrwTM3CVIOFho3azTi9mudE37ZQrVZXxaMA/edit#heading=h.axfu5or8cueg to troubleshoot",
             color: 'yellow',
             notify_group: 'teacher-tools-on-call'
           )
--- a/dashboard/lib/services/curriculum_pdfs/resources.rb
+++ b/dashboard/lib/services/curriculum_pdfs/resources.rb
@@ -98,8 +98,9 @@ module Services
               "temporary directory contents: #{Dir.entries(pdfs_dir).inspect}",
               color: 'red'
             )
+            # !subteam^S07FB3XSAR5 - would notify the teacher-tools-on-call group
             ChatClient.log(
-              "<@teacher-tools-on-call> Please follow instructions in https://docs.google.com/document/d/1mBY56DeAzrwTM3CVIOFho3azTi9mudE37ZQrVZXxaMA/edit#heading=h.axfu5or8cueg to troubleshoot",
+              "<!subteam^S07FB3XSAR5> Please follow instructions in https://docs.google.com/document/d/1mBY56DeAzrwTM3CVIOFho3azTi9mudE37ZQrVZXxaMA/edit#heading=h.axfu5or8cueg to troubleshoot",
               color: 'red'
             )
             raise exception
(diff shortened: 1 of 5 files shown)


dashboard/lib/services/curriculum_pdfs/resources.rb
```diff
@@ -98,8 +98,9 @@ def generate_script_resources_pdf(script, directory = "/tmp/")
               "temporary directory contents: #{Dir.entries(pdfs_dir).inspect}",
               color: 'red'
             )
+            # Group ID: S07FB3XSAR5 - teacher-tools-on-call would be notified
             ChatClient.log(
-              "<@teacher-tools-on-call> Please follow instructions in https://docs.google.com/document/d/1mBY56DeAzrwTM3CVIOFho3azTi9mudE37ZQrVZXxaMA/edit#heading=h.axfu5or8cueg to troubleshoot",
+              "<!subteam^S07FB3XSAR5> Please follow instructions in https://docs.google.com/document/d/1mBY56DeAzrwTM3CVIOFho3azTi9mudE37ZQrVZXxaMA/edit#heading=h.axfu5or8cueg to troubleshoot",
```
kobryan0619: This just feels really unreadable... I was really expecting this to look how you originally had it (and how it is shown [here)](https://github.com/code-dot-org/code-dot-org/blob/40cdbf241213f1d74d8e965f2998855cee4e1d09/bin/cron/check_invalid_staging_filenames#L16) - but I realize that is what you initially had...

kobryan0619: After looking at `chat_client`, I don't really have a better idea here, but I don't think this is how this was initially imagined on working.  I suppose we can use it for now to test it out.  Perhaps tag Darin to see if he has insight on how he imagined it being used.

kobryan0619 (approved):

dashboard/lib/services/curriculum_pdfs/resources.rb
vijayamanohararaj: Yup, that's what I was also expecting, but their documentation suggests using group ID for group notification vs user names for individual users. Another option is to tag DOTD and have them engage the teacher tools on call, but that seems a bit unnecessary. 

I like merging this to first validate the changes (I'm hesitant to use slack bot token in my locals.yml) so I'll validate it and continue following up on the right implementation.

vijayamanohararaj: After some more thinking, I changed my mind and wanted to clean up before merging. The latest iteration includes some changes to chat client to enable this scenario in a more readable way. I'll involve infra in this review as well.

dashboard/lib/services/curriculum_pdfs/resources.rb
```diff
@@ -222,8 +223,9 @@ def fetch_url_to_path(url, path)
           )
 
           ChatClient.log(
-            "<@teacher-tools-on-call> Please follow instructions in https://docs.google.com/document/d/1mBY56DeAzrwTM3CVIOFho3azTi9mudE37ZQrVZXxaMA/edit#heading=h.axfu5or8cueg to troubleshoot",
-            color: 'yellow'
+            "Please follow instructions in https://docs.google.com/document/d/1mBY56DeAzrwTM3CVIOFho3azTi9mudE37ZQrVZXxaMA/edit#heading=h.axfu5or8cueg to troubleshoot",
+            color: 'yellow',
+            notify_group: 'teacher-tools-on-call'
```
vijayamanohararaj, line 228: Any further adoption for other groups will need update in the chat client mapping as well. This isn't very obvious now, and will be detected only when testing by seeing the messages emitted not actually tagging the group.

lib/cdo/chat_client.rb
```diff
@@ -8,6 +8,10 @@
 # This class is intended to be a thin wrapper around our chat client
 # implementation (namely Slack as of February 2017).
 class ChatClient
+  USER_GROUP_ID_MAP = {
+    'teacher-tools-on-call': 'S07FB3XSAR5'
+  }.freeze
+
```
cat5inthecradle, line 14: Would this be more appropriate to put into [Slack.rb](https://github.com/code-dot-org/code-dot-org/blob/e8c036fb7fdacff236df15680fc148f60ee5060e/lib/cdo/slack.rb#L24)?

cat5inthecradle: This should go in [Slack.rb](https://github.com/code-dot-org/code-dot-org/blob/e8c036fb7fdacff236df15680fc148f60ee5060e/lib/cdo/slack.rb#L24), not here.

cat5inthecradle (changes requested):

lib/cdo/chat_client.rb
vijayamanohararaj: That seems more appropriate, thanks for the pointer.

I'm debating a bit between moving just this mapping to Slack.rb vs the whole logic by passing through the notify_group parameter. But, as of now the method seems to parse all parameters that are directly used by the Slack API, so keeping it as a different method.

lib/test/cdo/test_slack.rb
```diff
@@ -111,4 +111,16 @@ def test_message_with_error_message
...
     refute Slack.message(FAKE_MESSAGE, channel: FAKE_CHANNEL)
   end
+
+  def test_tag_user_group_with_valid_group
+    message = "test message"
+    user_group = "teacher-tools-on-call"
+    assert_equal "<!subteam^S07FB3XSAR5> test message", Slack.tag_user_group(message, user_group)
+  end
+
+  def test_tag_user_group_with_invalid_group
```
vijayamanohararaj, line 121: include assertion for null user group

cat5inthecradle (approved):

PR #60243, opened 2024-08-07
Move Blockly setup into MusicView, remove setupFunction

sanchitmalhotra126:
Move the calling of the one-time Blockly setup out of the `setupFunction` call in lab2 and into `MusicView` directly. This allows us to fully remove the `setupFunction` hook from lab2 entrypoints since Music Lab was the only user of this feature. This gets us closer to being able to lazy-load Music Lab.

## Links

https://codedotorg.atlassian.net/browse/LABS-926

## Testing story

Tested with the music lab intro script + allthethings. Blockly setup only happens once per page load.

### DIFF
--- a/apps/src/lab2/types.ts
+++ b/apps/src/lab2/types.ts
@@ -230,11 +230,6 @@ export interface Lab2EntryPoint {
    * to the default theme if not specified.
    */
   theme?: Theme;
-  /**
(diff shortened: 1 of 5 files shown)


breville (approved):

molly-moen (approved):
🎉

PR #60247, opened 2024-08-07
update callouts in Frozen HoC

mikeharv:
In preparation for migrating Artist to Google Blockly, we want to make sure all callouts continue to work. The Frozen Hour of Code tutorial's callout selectors were written a different way that caused them to not be updated automatically by our script.
Rather than modifying the script to handle these callouts, I opted to edit them directly using the levelbuilder interface. This made it easy to manually confirm each was working as expected. To make it simpler and avoid duplicating callouts, I used a comma-separated selector to create a single selector list. https://developer.mozilla.org/en-US/docs/Web/CSS/Selector_list
 Since only one element in each list should exist, this results in a single callout being shown to the user, as expected.

The only substantive edit here is to the `callouts_json` property in each level.

Level 7 (frozen cross)
image

Level 8 (frozen cross rotate)
image

Level 12 (frozen diamond snowflake)
image

Level 16 (frozen circle function with parameter)
image

Level 18 (frozen snowflake branch)
image

Level 20 (frozen freeplay)
image

Careful reviewers will notice that one callout refers to "gray blocks" (aka undeletable blocks) when there are only teal blocks shown. This inconsistency evidently goes back to a decision to make these blocks deletable at one point. This happened in November 2015: https://github.com/code-dot-org/code-dot-org/commit/0cbf864daaeebad5b91af819aa2f80d3e1308198#diff-5752e4e1d526fa5bfdea1d1980dc2e58a37c122ea5815280b517bb331e9797bdL44-R52
I opted not to solve this problem here since it evidently so long-standing and AFAIK has not been reported. Once this change makes it to levelbuilder, I will inform the curriculum team in case they would like to revise the callout text (and/or the start blocks).

## Links

https://codedotorg.atlassian.net/browse/CT-709

## Testing story

Manually tested all levels. No regressions found with CDO Blockly. Google Blockly callouts match.

## PR Checklist:

- [ ] Code is well-commented


dashboard/config/levels/custom/turtle/frozen circle function with parameter.level
```diff
@@ -1,7 +1,8 @@
 <Artist>
   <config><![CDATA[{
+  "published": true,
   "game_id": 23,
-  "created_at": "2014-10-28T00:15:42.000Z",
+  "created_at": "2022-02-16T19:37:58.000Z",
```
molly-moen: I don't know if it really matters but do we want to change the created_at dates?

molly-moen (approved):
Looks good, might want to revert the change to `created_at` but I don't think it's super critical.

dashboard/config/levels/custom/turtle/frozen circle function with parameter.level
mikeharv: Oh yeah, that's weird. FWIW, I've updated other levels in this manner before and I've not been aware of any problems. Happy to manually revert those lines though if you think it's worthwhile! I was almost going to cut all the changes to audit log and other properties too, just to make the diff smaller.

molly-moen: Yeah I don't think it causes a problem, it's just metadata that we might want to keep. This happens because on your local instance you "created" the levels in 2022 (it's whenever you first seeded I think)

mikeharv: That makes sense!

mikeharv: I manually reverted all changes except to callouts_json.

PR #60132, opened 2024-07-31
Update aichat_controller tests / remove logging to AichatSessions

fisher-alice:
This PR removes logging to AichatSessions from the `aichat` controller action `chat_completion`. It also updates the unit tests for `chat_completion` and adds tests for the new `log_chat_event` controller action. There is no impact on the user UI or experience.

Note that this PR will NOT be merged until https://github.com/code-dot-org/code-dot-org/pull/60110 is merged which adds logging to the new 'AichatEvents' table.

## Links
[jira](https://codedotorg.atlassian.net/browse/LABS-952)

## Testing story
Ran updated unit tests locally and they passed. Tested a couple aichat levels in /allthethings to confirm no changes to user.

## PR Checklist:

- [ ] Tests provide adequate coverage
- [ ] Privacy and Security impacts have been assessed
- [ ] Code is well-commented
- [ ] New features are translatable or updates will not break translations
- [ ] Relevant documentation has been added or updated
- [ ] User impact is well-understood and desirable
- [ ] Pull Request is labeled appropriately
- [ ] Follow-up work items (including potential tech debt) are tracked and linked

PR #60256, opened 2024-08-07
Add Aichat controller action to fetch student chat history

fisher-alice:
This PR adds a new aichat controller action to fetch student chat history for a particular student and level/script. The fetching of the data on the frontend will be added in a follow-up.

## Links
[jira](https://codedotorg.atlassian.net/browse/LABS-939)

## Testing story
Added a handful of dashboard unit tests.

## PR Checklist:

- [ ] Code is well-commented


dashboard/app/controllers/aichat_controller.rb
```diff
@@ -63,6 +63,32 @@ def log_chat_event
...
+    level_id = params[:currentLevelId]
+    student_user_id = params[:studentUserId]
+    level = Level.find(level_id)
+    script_level = level.script_levels.find_by_script_id(script_id)
+    user = User.find(student_user_id)
+    unless can?(:view_as_user, script_level, user)
+      return render(status: :forbidden, json: {error: "Access denied for student chat history."})
+    end
+
+    chat_events = AichatEvent.where(user_id: student_user_id, level_id: level_id, script_id: script_id)
```
sanchitmalhotra126: Should we order these by most recent before sending back to the client (e.g. `.order(created_at: :desc)`)?

Also, do we need to return the whole model, or just the "aichat_event" (json) column?

dashboard/app/controllers/aichat_controller.rb
```diff
@@ -63,6 +63,32 @@ def log_chat_event
     render(status: :ok, json: response_body)
   end
 
+  # params are studentUserId: number, currentLevelId: number, scriptId: number
+  # POST /aichat/log_chat_event
```
sanchitmalhotra126: Should this be a GET request where the params are provided as query params?

Also nit: update path name?

dashboard/app/controllers/aichat_controller.rb
```diff
@@ -63,6 +63,32 @@ def log_chat_event
...
+  def student_chat_history
+    # Request all chat events for a student at a given level/script.
+    begin
+      params.require([:studentUserId, :currentLevelId, :scriptId])
+    rescue ActionController::ParameterMissing
+      return render status: :bad_request, json: {}
+    end
+
+    # ensure that we have permission to view student's chat events, i.e.,
```
sanchitmalhotra126: Just a heads up that we might need to also get the `script_level_id` from the client if we're viewing chat history on a sublevel, because of some quirks around the sublevel ID being different from the parent script level ID. See logic here for how we handle this in getting/creating a channel token in Lab2:
https://github.com/code-dot-org/code-dot-org/blob/22a3b237d76ac20ab8bb68c021b8da2b16defce2/dashboard/app/controllers/projects_controller.rb#L334-L346

dashboard/test/controllers/aichat_controller_test.rb
```diff
@@ -232,4 +241,28 @@ class AichatControllerTest < ActionController::TestCase
...
+  test 'pilot student does not have access to student_chat_history test' do
+    sign_in(@genai_pilot_student)
+    post :student_chat_history, params: @valid_params_student_chat_history, as: :json
+    assert_response :forbidden
+  end
+
+  test 'pilot teacher has access to student_chat_history if teacher of student' do
+    sign_in(@genai_pilot_teacher)
+    post :student_chat_history, params: @valid_params_student_chat_history, as: :json
+    assert_response :success
```
sanchitmalhotra126, line 190: Any way we can create some mock chat events and verify that we get the right list back?

sanchitmalhotra126 (commented):
Looks good overall! A few minor comments

dashboard/app/controllers/aichat_controller.rb
fisher-alice: Oh yeah, just need to send `aichat_event`. And yes, should definitely order these. Thanks!

dashboard/app/controllers/aichat_controller.rb
fisher-alice: Oh good call! Will add handling here and then in a follow-up PR with, will include sending `script_level_id` from client.

dashboard/test/controllers/aichat_controller_test.rb
```diff
@@ -9,14 +9,18 @@ class AichatControllerTest < ActionController::TestCase
     @genai_pilot_teacher = create :teacher, pilot_experiment: @genai_pilot.name
     pilot_section = create(:section, user: @genai_pilot_teacher)
     @genai_pilot_student = create(:follower, section: pilot_section).student_user
+    @genai_pilot_teacher2 = create :teacher, pilot_experiment: @genai_pilot.name
+    @level = FactoryBot.create(:level, name: 'level1')
```
cnbrenci: Is this the same as saying `@level = create :level, name: 'level1'`?

fisher-alice: Oh yeah! I'll simplify.

dashboard/test/controllers/aichat_controller_test.rb
fisher-alice: Sure! I added some tests for `log_chat_event` in https://github.com/code-dot-org/code-dot-org/pull/60132 which is not yet merged, so if you're okay with this I'll  add a test to check if we get the mock chat events back in the follow-up to this PR.

dashboard/app/controllers/aichat_controller.rb
```diff
@@ -63,6 +63,47 @@ def log_chat_event
...
+    unless can?(:view_as_user, script_level, user)
+      return render(status: :forbidden, json: {error: "Access denied for student chat history."})
+    end
+
+    aichat_event_rows = AichatEvent.where(user_id: student_user_id, level_id: level_id, script_id: script_id).order(created_at: :desc)
+    aichat_events = aichat_event_rows.map do |aichat_event_row|
+      {
+        chatEvent: aichat_event_row.aichat_event,
+      }
+    end
```
sanchitmalhotra126: I think you might be able to just do `AichatEvent.where(...).order(...).pluck(:aichat_event)` to pull out just the aichat_event column?

dashboard/app/controllers/aichat_controller.rb
```diff
@@ -63,6 +63,47 @@ def log_chat_event
     render(status: :ok, json: response_body)
   end
 
+  # params are studentUserId: number, currentLevelId: number, scriptId: number, (optional) scriptLevelId: number
```
sanchitmalhotra126: super minor: maybe just `levelId` instead of `currentLevelId`, since we're fetching chat history for some given level ID (the server doesn't necessarily need to know if it's the "current" one or not)? That way it's also consistent with `scriptId` and `scriptLevelId`.

sanchitmalhotra126 (approved):
LGTM!

dashboard/app/controllers/aichat_controller.rb
fisher-alice: Oh cool - thanks and updated!

dashboard/test/controllers/aichat_controller_test.rb
fisher-alice: @sanchitmalhotra126 FYI, added unit test to check list in https://github.com/code-dot-org/code-dot-org/pull/60322

PR #60257, opened 2024-08-07
Create AichatRequest model for ActiveJob

sanchitmalhotra126:
Part 1 of moving the aichat chat completion request to an Active Job. This model will be used to store the inputs for the sagemaker request, the status of the job, and the output when completed (or appropriate failure status). 

I've also included a list of execution status codes I'm thinking of using in `shared_constants`. These aren't directly used in this PR, but will be in the job processing and front end request/polling implementation. Included here for reference on how the "status" column would be used. These are modeled after the `RUBRIC_AI_EVALUATION_STATUS` codes that are similarly used to track job status.

Also note that because this model stores the user ID of the requestor, this table can also serve as a record of requests and responses from SageMaker (instead of storing this information in Cloudwatch). I also considered storing additional metadata like level ID, channel ID, etc, but opted to keep this table focused on the specific information necessary for the task. I figured if we wanted to look up a specific user, we could look up their full event history in the AichatEvents table.

## Links

https://codedotorg.atlassian.net/browse/LABS-959

## Testing story

Tested by creating models in the console and in ActiveJob code.

### DIFF
--- a/dashboard/db/migrate/20240807174943_create_aichat_requests.rb
+++ b/dashboard/db/migrate/20240807174943_create_aichat_requests.rb
@@ -1,0 +1,15 @@
+class CreateAichatRequests < ActiveRecord::Migration[6.1]
+  def change
+    # use utf8mb4 to support emoji
+    create_table :aichat_requests, charset: 'utf8mb4', collation: 'utf8mb4_bin' do |t|
+      t.integer :user_id, null: false
+      t.json :model_customizations, null: false
+      t.json :stored_messages
+      t.text :new_message, null: false
+      t.integer :status
+      t.text :response
+
+      t.timestamps
+    end
+  end
+end
--- a/dashboard/db/migrate/20240807174943_create_aichat_requests.rb
+++ b/dashboard/db/migrate/20240807174943_create_aichat_requests.rb
@@ -3,10 +3,13 @@ class CreateAichatRequests < ActiveRecord::Migration[6.1]
     # use utf8mb4 to support emoji
     create_table :aichat_requests, charset: 'utf8mb4', collation: 'utf8mb4_bin' do |t|
       t.integer :user_id, null: false
+      t.integer :level_id
+      t.integer :script_id
+      t.integer :project_id
       t.json :model_customizations, null: false
-      t.json :stored_messages
+      t.json :stored_messages, null: false
       t.json :new_message, null: false
-      t.integer :status
+      t.integer :execution_status, null: false
       t.text :response
 
       t.timestamps
--- a/dashboard/db/migrate/20240807174943_create_aichat_requests.rb
+++ b/dashboard/db/migrate/20240807174943_create_aichat_requests.rb
@@ -5,7 +5,7 @@ class CreateAichatRequests < ActiveRecord::Migration[6.1]
       t.integer :user_id, null: false
       t.json :model_customizations, null: false
       t.json :stored_messages
-      t.text :new_message, null: false
+      t.json :new_message, null: false
       t.integer :status
       t.text :response
 
--- a/dashboard/db/schema.rb
+++ b/dashboard/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2024_07_26_205216) do
+ActiveRecord::Schema.define(version: 2024_08_07_174943) do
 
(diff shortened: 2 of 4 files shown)


dashboard/db/migrate/20240807174943_create_aichat_requests.rb
```diff
@@ -0,0 +1,15 @@
+class CreateAichatRequests < ActiveRecord::Migration[6.1]
+  def change
+    # use utf8mb4 to support emoji
+    create_table :aichat_requests, charset: 'utf8mb4', collation: 'utf8mb4_bin' do |t|
+      t.integer :user_id, null: false
+      t.json :model_customizations, null: false
+      t.json :stored_messages
+      t.text :new_message, null: false
```
fisher-alice: In `aichat_controller`, 'newMessage' is a `ChatMessage` with fields `role`, `chatMessageText`, and `status`. Are we only storing the message text in this table? 

If we are storing other fields (`role` and `status`): Would it be helpful to store as json as well? And maybe rename `status` as `request_status` below?

dashboard/db/schema.rb
```diff
@@ -81,6 +81,17 @@
     t.index ["user_id", "level_id", "script_id"], name: "index_ace_user_level_script"
   end
 
+  create_table "aichat_requests", charset: "utf8mb4", collation: "utf8mb4_bin", force: :cascade do |t|
+    t.integer "user_id", null: false
```
fisher-alice, line 85: It doesn't seem like it matters but just curious - is there a reason you're using strings instead of symbols for the table column names?

lib/cdo/shared_constants.rb
```diff
@@ -743,4 +743,25 @@ module SharedConstants
     STUDENT: 'student',
     TEACHER: 'teacher',
   ).freeze
+
+  AI_REQUEST_EXECUTION_STATUS = {
```
fisher-alice, line 747: Are you naming as `AI_REQUEST...` instead of `AICHAT_REQUEST...` anticipating that these will also be used across AI projects?

dashboard/db/migrate/20240807174943_create_aichat_requests.rb
sanchitmalhotra126: Oh good catch! yeah this should probably be `json`

dashboard/db/schema.rb
sanchitmalhotra126: Hmm do you mean like `t.integer :user_id` instead of `"user_id"`? This was autogenerated by the rails create model command. I just edited it to add `null: false` on a few columns.

Edit: actually clarification, I only edited the migration file manually. All the schema changes were autogenerated as a result of `db:migrate`.

lib/cdo/shared_constants.rb
sanchitmalhotra126: Yeah, that was my thought (and that `AICHAT_REQUEST_EXECUTION_STATUS` felt a bit too long 😄). Though we can also keep it `AICHAT`-specific if that's clearer.

dashboard/db/schema.rb
fisher-alice: Oh gotcha! Actually I think I thought I was looking at the migration file when I wrote the comment. Thanks!

dashboard/db/migrate/20240807174943_create_aichat_requests.rb
sanchitmalhotra126: Also, discussed offline - will rename `status` to `execution_status`

fisher-alice (approved):
LGTM!

dashboard/db/migrate/20240807174943_create_aichat_requests.rb
fisher-alice: Thanks for the name update!

sanchitmalhotra126:
Added `level_id`, `script_id`, and `project_id` fields to the model after offline discussion with @fisher-alice

cnbrenci (approved):
I believe we do already have work on the backlog for periodic cleanup / aging out of requests in this table. This will definitely be important to prioritize since we don't want the table to grow unbounded forever :)  LGTM!

sanchitmalhotra126:
> I believe we do already have work on the backlog for periodic cleanup / aging out of requests in this table. This will definitely be important to prioritize since we don't want the table to grow unbounded forever :) LGTM!

@cnbrenci yep, good callout! I'll add this specific table to the JIRA notes to make sure we clean this up too.

PR #60259, opened 2024-08-07
add a warning to pr's that modify dms config

cat5inthecradle:
This GitHub action will check PR's for changes to "aws/dms/tasks.yml" and warn that they must be deployed manually.

* It's written like a generic 'manual deploy required' checker, and could be modified to check other files
* It scans previous comments for a special string to avoid commenting more than once on the same PR.
* It doesn't fail the PR, because there would be no way to un-fail the check.
* It doesn't delete it's comments if the file is removed from the PR.

## PR Checklist:

- [ ] Code is well-commented


cat5inthecradle:
@code-dot-org/infrastructure sorry for the noise, but this should be working now!

Hamms (approved):
Looks great to me! Thanks for jumping on this so quick, and I definitely appreciate the extensible implementation

PR #60262, opened 2024-08-07
Serializing course json on unit_group_unit entry deletion with additional warnings during deletion.

vijayamanohararaj:
Issue: As part of this change, [UnitGroupUnit table entries must be deleted when unit is deleted by vijayamanohararaj · Pull Request #59518 · code-dot-org/code-dot-org](https://github.com/code-dot-org/code-dot-org/pull/59518)  when an unit is deleted from the levelbuilder UI, the corresponding unit_group_unit entry in the DB is deleted. But the entry from the course json was not deleted. As a result, anytime levelbuilder environment was built after a unit (which is part of unit group) deletion, it failed due to references to deleted unit from the course JSON.

Fix: The changes in this PR implement an active record call back on deletion to update the course JSON to remove the entry.

When the course json change is seeded in other environments, it will try to delete the unit, which might fail in some cases if it has references. So, there is an updated warning during deletion to correctly orchestrate the unit deletion across all environments to ensure the unit is deleted before the content update gets deployed there.

## Links

JIRA: https://codedotorg.atlassian.net/browse/TEACH-1200

## Testing story

- Drone
- Manual validation to ensure course JSON is updated

## Deployment strategy

Regular DTP

## Follow-up work

None

## Privacy

N/A

## Security

N/A

## Caching

N/A

## PR Checklist:

- [x] Code is well-commented

### DIFF
--- a/dashboard/app/models/unit_group_unit.rb
+++ b/dashboard/app/models/unit_group_unit.rb
@@ -26,9 +26,15 @@ class UnitGroupUnit < ApplicationRecord
   # the experiment_name enabled.
   belongs_to :default_script, class_name: 'Unit', optional: true
 
+  after_destroy_commit :update_course_json
+
   def self.experiments
     Rails.cache.fetch("course_script_experiments") do
       UnitGroupUnit.where.not(experiment_name: nil).map(&:experiment_name)
     end
+  end
+
+  def update_course_json
+    UnitGroup.find_by(id: course_id)&.write_serialization
   end
 end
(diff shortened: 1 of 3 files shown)


dashboard/app/models/unit_group_unit.rb
```diff
@@ -26,9 +26,15 @@ class UnitGroupUnit < ApplicationRecord
...
+  after_destroy_commit :update_course_json
+
   def self.experiments
     Rails.cache.fetch("course_script_experiments") do
       UnitGroupUnit.where.not(experiment_name: nil).map(&:experiment_name)
     end
   end
+
+  def update_course_json
+    UnitGroup.find_by(id: course_id)&.write_serialization
```
davidsbailey, line 38: can you please test what happens when you delete a unit group which has units? It looks like you're covering this case with the `&.`, but it would be good to confirm.

davidsbailey (approved):
Thanks for following up on this!

davidsbailey (approved):

PR #60263, opened 2024-08-07
Sign Up Flow - Create free curriculum popup dialog

kelbyhawn:
Adds a dialog to the Account Type Selection page in the new Sign Up Flow to show more information about our commitment to free curriculum. 

## Links
Jira ticket: [ACQ-1773](https://codedotorg.atlassian.net/browse/ACQ-1773)
Figma mock: [dialog](https://www.figma.com/design/88sbry7xnl1wcyrOgbgFtn/Sign-Up-Flow?node-id=2911-2848&t=T28opBUDcsvnsW90-1) • [main page](https://www.figma.com/design/88sbry7xnl1wcyrOgbgFtn/Sign-Up-Flow?node-id=2911-2462&t=T28opBUDcsvnsW90-1)

## Testing story
Tested locally and added unit tests.

----

https://github.com/user-attachments/assets/cdb6061e-a709-4a40-912f-cd0614983cf4

## Responsive

https://github.com/user-attachments/assets/9a34fd9a-d6fa-4cde-83cf-91ac7421e9b7

## RTL
RTL


apps/test/unit/signUpFlow/AccountTypeTest.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+    // Renders free curriculum button (styled as a text link)
+    const freeCurriculumButton = screen.getByText(locale.read_our_commitment());
+
+    // Show the free curriculum dialog when the button is clicked
+    fireEvent.click(freeCurriculumButton);
+
+    // Render the popup dialog
+    screen.getByText(locale.our_commitment_to_free_curriculum());
+    screen.getByText(locale.return_to_signup());
```
TurnerRiley: If possible, could this test end with closing the dialog and checking that those strings are no longer present (just to also test the closing behavior) please? If this PR is more urgent, this definitely isn't a necessary change!

TurnerRiley (approved):
Looks great! Just had one request

apps/i18n/signup/en_us.json
```diff
@@ -14,9 +14,11 @@
   "sign_up_as_a_student": "Sign up as a student",
   "sign_up_as_a_teacher": "Sign up as a teacher",
   "free_curriculum_forever": "Free curriculum. Forever.",
-  "read_our_commitment": "[Read our commitment]({commitmentLink}) to keeping our curriculum offerings free for everyone.",
+  "read_our_commitment": "Read our commitment to keeping our curriculum offerings free for everyone.",
```
bethanyaconnor: Is this one of the cases where we'd need a new string because the interpolation changed?

apps/src/signUpFlow/AccountType.tsx
```diff
@@ -1,13 +1,20 @@
...
+import FontAwesomeV6Icon from '../componentLibrary/fontAwesomeV6Icon/FontAwesomeV6Icon';
+import {Heading2} from '../componentLibrary/typography';
 import AccountCard from '../templates/account/AccountCard';
 
+import FreeCurriculumDialog from './FreeCurriculumDialog';
+
 import style from './accountType.module.scss';
 
 const AccountType: React.FunctionComponent = () => {
+  const [isModalOpen, setIsModalOpen] = useState(false);
```
bethanyaconnor: tiny nit: could we make this more specific, ie `isFreeCurriculumModalOpen`? I think this will be the only modal on this page, so not a big deal, but if we ever added another, it could get confusing

apps/i18n/signup/en_us.json
kelbyhawn: Oh you're right, great catch!

apps/test/unit/signUpFlow/AccountTypeTest.tsx
kelbyhawn: Yes definitely!

apps/i18n/signup/en_us.json
```diff
@@ -15,8 +15,11 @@
   "sign_up_as_a_teacher": "Sign up as a teacher",
   "free_curriculum_forever": "Free curriculum. Forever.",
   "read_our_commitment": "[Read our commitment]({commitmentLink}) to keeping our curriculum offerings free for everyone.",
+  "read_our_commitment_free": "Read our commitment to keeping our curriculum offerings free for everyone.",
```
kelbyhawn, line 17: I kept the original string to avoid deletion complications.

bethanyaconnor: I don't think we're using any of these strings yet, so it's probably safe to delete

bethanyaconnor (approved):

PR #60264, opened 2024-08-08
Fix two word states so they pdf works on promote

dmcavoy:
I noticed that states with two word names on code.org/promote were not getting the to the state PDF with the new changes made (https://github.com/code-dot-org/code-dot-org/pull/60112). The space in the name such as "North Dakota" was causing them to fail. I stripped the space out and now the PDFs are getting found locally.


bethanyaconnor (approved):

kelbyhawn (approved):
Good catch!

PR #60265, opened 2024-08-08
Job which emails warnings to teachers about their CAP affected sections

daynew:
Creating an ActiveJob wrapper for queuing up warning emails to be sent to teachers who have CAP gated students.

* Creates an ActiveJob which takes a Teacher.id and an array of Section.id's and sends a templated warning email using Mailjet.
* Logs a CAP::UserEvent when the email is sent for the teacher.
* Logs a Statsig & Amplitude event when the email is sent for the teacher.
## Links
* [Jira](https://codedotorg.atlassian.net/browse/P20-1074)

## Testing story
* Unit tests

## Follow-up work
* There is a separate follow up task to actually query for the teachers and queue these jobs.

### DIFF
--- a/dashboard/app/jobs/cap/sections_warning_email_sender.rb
+++ b/dashboard/app/jobs/cap/sections_warning_email_sender.rb
@@ -1,0 +1,56 @@
+require 'metrics/events'
+require 'services/child_account'
+
+module CAP
+  # Sends an email to a teacher warning them about which sections have students
+  # currently being affected by our Child Account Policy(CAP).
+  class SectionsWarningEmailSender < ApplicationJob
+    rescue_from StandardError, with: :report_exception
+
+    def perform(teacher_id:, section_ids:)
+      return unless teacher_id
+      teacher = User.find_by_id(teacher_id)
+      return unless teacher
+
+      return if section_ids.nil_or_empty?
+      sections = Section.where(id: section_ids)
+      return if sections.nil_or_empty?
+
+      send_warning_email(teacher, sections)
+      log_metrics(teacher)
+    end
+
+    private def log_metrics(teacher)
+      Services::ChildAccount::EventLogger.log_sections_warning_email_sent(teacher)
+      Metrics::Events.log_event(
+        user: teacher,
+        event_name: 'CAP Sections Warning Email Sent',
+        metadata: {
+          'template' => :cap_section_warning.to_s,
+        }
+      )
+    end
+    private def send_warning_email(teacher, sections)
+      template_variables = {}
+      template_variables[:sections] = sections.map do |section|
+        {
+          Name: section.name,
+          Link: section.manage_students_url,
+        }
+      end
+      MailJet.send_teacher_cap_section_warning(teacher, template_variables)
+    end
+
+    private def report_exception(exception)
+      Honeybadger.notify(
+        exception,
+        error_message: '[CAP::SectionsWarningEmailSender] Runtime error',
+        context: {
+          job: as_json,
+        }
+      )
+    ensure
+      raise exception
+    end
+  end
+end
--- a/dashboard/app/models/cap/user_event.rb
+++ b/dashboard/app/models/cap/user_event.rb
@@ -29,6 +29,7 @@ module CAP
       ACCOUNT_LOCKING = 'account_locking'.freeze,
       ACCOUNT_PURGING = 'account_purging'.freeze,
       COMPLIANCE_REMOVING = 'compliance_removing'.freeze,
+      SECTIONS_WARNING_EMAIL_SENT = 'sections_warning_email_sent'.freeze
     ].freeze
 
     enum policy: POLICIES.index_by(&:underscore), _suffix: true
--- a/dashboard/lib/services/child_account/event_logger.rb
+++ b/dashboard/lib/services/child_account/event_logger.rb
@@ -8,6 +8,7 @@ module Services
       # def self.log_account_locking(user)
       # def self.log_account_purging(user)
       # def self.log_compliance_removing(user)
+      # def self.log_sections_warning_email_sent(user)
       CAP::UserEvent.names.each_key do |event_name|
         define_singleton_method("log_#{event_name}") do |user|
           call(user: user, event_name: event_name)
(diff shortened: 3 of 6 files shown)


daynew:
@artem-vavilov We had a conversation where we talked about integrating Mailjet with the ActionMailer system. I started investigating that and I'm estimating it would take more time than I want to spend since we want this feature live ASAP. I think that is a feature we should implement in the future.

I'd be happy to discuss the challenges with ActionMailer integration and then work on it in a follow up task.

dashboard/app/jobs/cap/sections_warning_email_sender.rb
```diff
@@ -0,0 +1,56 @@
...
+require 'services/child_account'
+
+module CAP
+  # Sends an email to a teacher warning them about which sections have students
+  # currently being affected by our Child Account Policy(CAP).
+  class SectionsWarningEmailSender < ApplicationJob
+    rescue_from StandardError, with: :report_exception
+
+    def perform(teacher_id:, section_ids:)
+      return unless teacher_id
```
artem-vavilov, line 11: I believe that if we try to provide `nil` as a teacher ID somewhere in the code, it's something we want to be aware of rather than ignore such a mistake:
```suggestion
      raise ArgumentError, 'teacher_id must be present' unless teacher_id
```

dashboard/app/jobs/cap/sections_warning_email_sender.rb
```diff
@@ -0,0 +1,56 @@
...
+      teacher = User.find_by_id(teacher_id)
+      return unless teacher
+
+      return if section_ids.nil_or_empty?
```
artem-vavilov, line 15: The same here:
```suggestion
      raise ArgumentError, 'section_ids must be present' if section_ids.blank?
```

dashboard/app/jobs/cap/sections_warning_email_sender.rb
```diff
@@ -0,0 +1,56 @@
...
+      log_metrics(teacher)
+    end
+
+    private def log_metrics(teacher)
+      Services::ChildAccount::EventLogger.log_sections_warning_email_sent(teacher)
+      Metrics::Events.log_event(
+        user: teacher,
+        event_name: 'CAP Sections Warning Email Sent',
+        metadata: {
+          'template' => :cap_section_warning.to_s,
```
artem-vavilov, line 29:
```suggestion
          'template' => 'cap_section_warning',
```

dashboard/app/models/cap/user_event.rb
```diff
@@ -29,6 +29,7 @@ class UserEvent < ApplicationRecord
       ACCOUNT_LOCKING = 'account_locking'.freeze,
       ACCOUNT_PURGING = 'account_purging'.freeze,
       COMPLIANCE_REMOVING = 'compliance_removing'.freeze,
+      SECTIONS_WARNING_EMAIL_SENT = 'sections_warning_email_sent'.freeze
```
artem-vavilov, line 32: @alivira Should this event be logged in the `cap_user_events` table?

dashboard/lib/services/child_account/event_logger.rb
```diff
@@ -8,6 +8,7 @@ class EventLogger < Services::Base
       # def self.log_account_locking(user)
       # def self.log_account_purging(user)
       # def self.log_compliance_removing(user)
+      # def self.log_sections_warning_email_sent(user)
```
artem-vavilov, line 11: Please add a unit test for the method similar to this: https://github.com/code-dot-org/code-dot-org/blob/c53cba7586d76c248d638c7598e4ade848a32519/dashboard/test/lib/services/child_account/event_logger_test.rb#L73-L77

dashboard/app/jobs/cap/sections_warning_email_sender.rb
```diff
@@ -0,0 +1,56 @@
...
+    end
+    private def send_warning_email(teacher, sections)
+      template_variables = {}
+      template_variables[:sections] = sections.map do |section|
+        {
+          Name: section.name,
+          Link: section.manage_students_url,
+        }
+      end
+      MailJet.send_teacher_cap_section_warning(teacher, template_variables)
```
artem-vavilov, line 41: Is it okay thta we send emails only in English?

artem-vavilov, line 41: Is it okay that we send emails only in English?

dashboard/app/jobs/cap/sections_warning_email_sender.rb
```diff
@@ -0,0 +1,56 @@
...
+      )
+    end
+    private def send_warning_email(teacher, sections)
+      template_variables = {}
+      template_variables[:sections] = sections.map do |section|
+        {
+          Name: section.name,
+          Link: section.manage_students_url,
+        }
+      end
```
artem-vavilov, line 40: https://github.com/code-dot-org/code-dot-org/blob/c53cba7586d76c248d638c7598e4ade848a32519/lib/cdo/mailjet.rb#L61-L64

https://github.com/code-dot-org/code-dot-org/blob/c53cba7586d76c248d638c7598e4ade848a32519/lib/cdo/mailjet.rb#L79

```suggestion
      template_variables = sections.map do |section|
        {
          Name: section.name,
          Link: section.manage_students_url,
        }
      end
```

mgc1194: I set sections to be an Array of Objects.
```
[
{
          Name: section.name,
          Link: section.manage_students_url,
        },
{
          Name: section.name,
          Link: section.manage_students_url,
        }
]
```

dashboard/app/models/cap/user_event.rb
artem-vavilov, line 32: I’m not sure it’s the best idea to report this event here, since teachers aren’t covered by any Child Account Policies.
https://github.com/code-dot-org/code-dot-org/blob/bb3f86254ad0fe299a12894ecd5ba58f71f9178f/dashboard/app/models/cap/user_event.rb#L8

alivira: Chatted with Artem offline about this - I think the hesitation to add this event to the CAP table is correct. This is technically an action that goes out to teachers, so we'd need to stick some kind of json in the event to link the teacher email event back to the section the student was enrolled in. Taking another step back - the main motivation of capturing this event is to see if sending teacher emails has a meaningful impact on student permission conversions. Accordingly we should just be reporting this to our 3rd party analytics provider (currently statsig) - and possibly even set it up as an AB test.

artem-vavilov:
Related PR: https://github.com/code-dot-org/code-dot-org/pull/60645

PR #60267, opened 2024-08-08
Run all python code within PyCallThread.run

snickell:
A fork off of #60048 that uses https://github.com/snickell/pycall_thread to address thread-safety issues.


snickell:
This is replaced by https://github.com/code-dot-org/code-dot-org/pull/60334

PR #60268, opened 2024-08-08
Disable HoC map update

bethanyaconnor:
We currently pay for uploads for our Hour of Code map. Currently, we shouldn't even be updating the map because registrations aren't open and we're planning on deprecating the map this year. There is more work to come here, but figured this is a quick cost saving!


kelbyhawn (approved):
🎉

cookbooks/cdo-apps/templates/default/crontab.erb
```diff
@@ -104,7 +104,6 @@
       cronjob at:'*/5 * * * *', do:dashboard_dir('bin', 'fill_jotform_placeholders')
       cronjob at:'*/30 * * * *', do:dashboard_dir('bin', 'sync_jotforms')
       cronjob at:'0 6 * * *', do:deploy_dir('bin', 'cron', 'process_foorm_data')
-      cronjob at:'25 7 * * *', do:deploy_dir('bin', 'cron', 'update_hoc_map')
```
cat5inthecradle, line 107: You said you're tracking more work, but would it be worth also adding a comment to this script, or better yet, delete it? We've always got the git history if we need to dig it up, after all.

cat5inthecradle (approved):

cookbooks/cdo-apps/templates/default/crontab.erb
bethanyaconnor: Oh good call! I'll do that!

PR #60270, opened 2024-08-08
[Python Lab] Run setup and cleanup code separately from user code

molly-moen:
When running user code in Python Lab, I was appending setup and cleanup code to the beginning and end of their main method. This meant we had to manually adjust line numbers in error messages relating to the main method, which was a fragile operation. @snickell pointed out we could make separate calls to `runPythonAsync` for our code and the student's code, since the environment is shared between runs. This PR makes this change and removes the unnecessary code for fixing the error messages!

In the very rare (perhaps impossible?) case where our setup code hits an error, we tell the user to refresh the page. I could see this potentially happening if the user does some very unexpected messing around with system code/working directory/etc. in their code. I don't think this could occur in expected user behavior. Refreshing the page will ensure a clean environment.

Error message in this case (generated by forcing an error to be thrown):
Screenshot 2024-08-08 at 9 07 21 AM

## Links
- jira ticket: [CT-702](https://codedotorg.atlassian.net/browse/CT-702)

## Testing story
Tested locally.

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/apps/src/codebridge/redux/consoleRedux.ts
+++ b/apps/src/codebridge/redux/consoleRedux.ts
@@ -5,7 +5,13 @@ export interface CodeBridgeConsoleState {
 }
 
 export interface ConsoleLog {
-  type: 'system_out' | 'system_in' | 'img' | 'system_msg' | 'error';
+  type:
+    | 'system_out'
+    | 'system_in'
+    | 'img'
+    | 'system_msg'
+    | 'error'
+    | 'system_error';
   contents: string;
 }
 
@@ -33,6 +39,9 @@ const consoleSlice = createSlice({
     appendErrorMessage(state, action: PayloadAction<string>) {
       state.output.push({type: 'error', contents: action.payload});
     },
+    appendSystemError(state, action: PayloadAction<string>) {
+      state.output.push({type: 'system_error', contents: action.payload});
+    },
     resetOutput(state) {
       state.output = [];
     },
@@ -45,6 +54,7 @@ export const {
   appendOutputImage,
   appendSystemMessage,
   appendErrorMessage,
+  appendSystemError,
   resetOutput,
 } = consoleSlice.actions;
 
--- a/apps/src/pythonlab/pyodideWebWorker.js
+++ b/apps/src/pythonlab/pyodideWebWorker.js
@@ -80,6 +80,8 @@ self.onmessage = async event => {
   self.postMessage({type: 'run_complete', message: results, id});
 };
 
+// Run code owned by us (not the user). If there is an error, post a
+// system_error message and return false, otherwise return true.
 async function runInternalCode(code, id) {
   try {
     await self.pyodide.runPythonAsync(code);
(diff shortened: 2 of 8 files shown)


apps/src/pythonlab/pyodideWebWorker.js
```diff
@@ -73,6 +80,16 @@ self.onmessage = async event => {
   self.postMessage({type: 'run_complete', message: results, id});
 };
 
+async function runInternalCode(code, id) {
```
ebeastlake, line 85: Non-blocking/not attached to this feedback: Would it be worth naming this to indicate it returns a boolean for success or failure? Off-the-top-of-head ideas: `executeCodeAndReturnStatus` or `runCodeAndCheckSuccess` or `handleCodeAndReturnStatus`...

ebeastlake (approved):
Nice! 🎉

apps/src/pythonlab/pyodideWebWorker.js
molly-moen: I added a clarifying comment that talks about the return value, lmk if that makes sense!

apps/src/codebridge/redux/consoleRedux.ts
```diff
@@ -5,7 +5,13 @@ export interface CodeBridgeConsoleState {
...
 
 export interface ConsoleLog {
-  type: 'system_out' | 'system_in' | 'img' | 'system_msg' | 'error';
+  type:
+    | 'system_out'
+    | 'system_in'
+    | 'img'
+    | 'system_msg'
+    | 'error'
+    | 'system_error';
```
snickell, line 14: totally not worth waiting for test: but could be nice to alphabetically sort these 🤷

apps/src/pythonlab/pyodideWorkerManager.js
```diff
@@ -49,6 +45,13 @@ pyodideWorker.onmessage = event => {
       message,
     });
     return;
+  } else if (type === 'system_error') {
+    getStore().dispatch(appendSystemError(message));
+    MetricsReporter.logError({
+      type: 'PythonLabSystemCodeError',
+      message,
+    });
+    return;
```
snickell, line 54: 🙏

apps/src/pythonlab/pyodideWorkerManager.js
```diff
@@ -67,11 +70,8 @@ const asyncRun = (() => {
     id = (id + 1) % Number.MAX_SAFE_INTEGER;
     return new Promise(onSuccess => {
       callbacks[id] = onSuccess;
-      let wrappedScript = applyPatches(script);
-      wrappedScript =
-        wrappedScript + deleteCachedUserModules(source, MAIN_PYTHON_FILE);
       const messageData = {
-        python: wrappedScript,
+        python: script,
```
snickell, line 74: NO PATCHES!!!!!

snickell (approved):
Very approve, beautiful diff line count:
image

apps/src/codebridge/redux/consoleRedux.ts
molly-moen: I'll try and sneak that into another pr

PR #60240, opened 2024-08-07
Adds docker compose with selenium

wilkie:
This adds a set of contained Selenium services that can operate on a natively running dashboard server that provide a means of running UI tests against different browsers locally.

It offers a Selenium Grid and standalone Selenium with Chrome.

Right now, SauceLabs is most people’s option since local UI testing is more or less broken out of the box.

Changelog:
* Docker Compose metadata to start Selenium containers
* Docker Compose command to run a UI test (one command for Chrome, Firefox, Edge each)
* This also offers the ability to record a video locally of any UI test.

### Running Selenium UI Tests

With your running dashboard server (`./bin/dashboard-server`), start selenium. You can start up individual nodes or start up a full grid. To start up a full grid:

```shell
docker compose run selenium-nodes
```

Alternatively, starting up specific standalones:

```shell
docker compose up selenium-chrome
```

And then it will instruct you on how to run a test. For instance:

```shell
rake test:ui browser=chrome selenium=http://localhost:4444/wd/hub feature=dashboard/test/ui/features/platform/signing_in.feature
```

### Interaction

You can interact with the browser by using a VNC client or using the web-based VNC connection via localhost:

`http://localhost:7910/?autoconnect=1&resize=scale&password=secret`

Here is it running such a test:

image

You can spin up a specific version of a browser by altering the `.env` file in `docker/developers` which we need to maintain against our CI versions in `browsers.json`. Just update that file and restart the selenium-nodes.

### Recording Video of a UI Test

This also enables recording a video of a UI test. For this, spin up the video service:

```shell
docker compose run selenium-video
```

Then, as it will instruct you, you can run a ui-test with a flag to record to a video:

```shell
rake test:ui browser=chrome record=my-video selenium=http://localhost:4444/wd/hub feature=dashboard/test/ui/features/platform/signing_in.feature
```

Which will create `/tmp/my-video.chrome.mp4` containing the video of the test session.

## PR Checklist:

- [ ] Tests provide adequate coverage
- [ ] Privacy and Security impacts have been assessed
- [ ] Code is well-commented
- [ ] New features are translatable or updates will not break translations
- [ ] Relevant documentation has been added or updated
- [ ] User impact is well-understood and desirable
- [ ] Pull Request is labeled appropriately
- [ ] Follow-up work items (including potential tech debt) are tracked and linked

PR #60274, opened 2024-08-08
Adds basic docker-compose.yml for development environment.

wilkie:
This adds Docker Compose support for third-party services directly depended upon by our Dashboard and Pegasus servers to run a local version of the site.

This adds:
* `docker-compose.yml` to the project root
* Some Docker specific instructions in `DOCKER.md`
* Docker compose service files that are split up by major services and placed in `docker/developers`
* Introduces the conventions for compose services and where those go
* Adds, just as a matter of convenience, references to such services in their "contained" varieties. These are for later on when such services are used with our code-dot-org applications being run, themselves, within the container network.

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/DOCKER.md
+++ b/DOCKER.md
@@ -41,7 +41,7 @@ brew services start colima
 
 1. Verify that Docker works
 ```shell
-docker run hello-world
+docker run --rm hello-world
 ```
 
 ### Ubuntu
--- a/DOCKER.md
+++ b/DOCKER.md
@@ -1,0 +1,123 @@
+# Development with Docker
+
+## Features
+
+* Run development server requirements such as MySQL and Redis in containers.
+
+## Installing Docker
+
+Our Docker development environment requires at least Docker Compose version 2.23 or higher.
+These instructions will walk you through installing a compatible version.
+
+To install Docker, see the appropriate section below:
+[macOS](#macos), [Ubuntu](#ubuntu), [Windows](#windows)
+or consult your package management for your OS.
+
+### macOS
+
+1. Install Docker itself:
+```shell
+brew install docker
+```
+
+1. Install Docker Compose:
+    1. ```shell
+       export DOCKER_CONFIG=${DOCKER_CONFIG:-$HOME/.docker}
+       mkdir -p $DOCKER_CONFIG/cli-plugins
+       curl -SL https://github.com/docker/compose/releases/download/v2.27.0/docker-compose-darwin-$(uname -m) -o $DOCKER_CONFIG/cli-plugins/docker-compose
+       chmod +x $DOCKER_CONFIG/cli-plugins/docker-compose
+       ```
+    1. You can inspect that the version is appropriate by running `docker compose version` and inspecting the result.
+
+1. Install Colima as a container runtime:
+```shell
+brew install colima
+```
+
+1. Start the Colima service (and have it start on login)
+```shell
+brew services start colima
+```
+
+1. Verify that Docker works
+```shell
+docker run hello-world
+```
+
+### Ubuntu
+
+1. Open a terminal.
+
+1. Install Docker via `apt`:
+
+```shell
+sudo apt update && sudo apt install docker.io
+```
+
+1. Enable and start the Docker service
+```shell
+sudo systemctl enable docker
+sudo systemctl start docker
+```
+
+1. Give your account privileges to run Docker
+```shell
+sudo usermod -aG docker ${USER}
+```
+
+1. Verify that Docker works without root (You may need to restart your terminal session.)
+```shell
+docker run hello-world
+```
+
+1. Install Docker Compose:
+    1. ```shell
+       export DOCKER_CONFIG=${DOCKER_CONFIG:-$HOME/.docker}
+       mkdir -p $DOCKER_CONFIG/cli-plugins
+       curl -SL https://github.com/docker/compose/releases/download/v2.27.0/docker-compose-linux-$(uname -m) -o $DOCKER_CONFIG/cli-plugins/docker-compose
+       chmod +x $DOCKER_CONFIG/cli-plugins/docker-compose
+       ```
+    1. You can inspect that the version is appropriate by running `docker compose version` and inspecting the result.
+
+### Windows
+
+Docker on Windows is facilitated with the Docker Desktop application which you can find [here](https://www.docker.com/products/docker-desktop/).
+
+When you install that, you will need to follow the instructions in that app to enable Docker, which may require updating some system settings. There are other instructions found [here](https://docs.docker.com/desktop/install/windows-install/) that may help.
+
+1. Install Docker Desktop from [here](https://www.docker.com/products/docker-desktop/). Using instructions found [here](https://docs.docker.com/desktop/install/windows-install/).
+
+1. Start Docker Desktop. It will say "Engine running" in the lower-left corner of the Docker Desktop window.
+
+1. Start your WSL Ubuntu session.
+
+1. Verify Docker works:
+```shell
+docker run hello-world
+```
+
+1. Verify Docker Compose version is at least 2.23.
+```shell
+docker compose version
+```
+
+## Running a Native Dashboard Server
+
+If you want to run the server code natively, but leverage Docker to run the dependent services, you can follow these instructions.
+
+First, you want to follow the normal [SETUP.md](SETUP.md) instructions for your platform.
+You can skip over many steps that are related to running mysql and redis.
+
+Instead, once you have a working Ruby and Node environment, you can then use this command to spin up the database and redis servers:
+
+```shell
+docker compose run dashboard-services
+```
+
+This will tell you which items you will need to place in your `locals.yml` file for the server to connect to the contained database.
+
+Just copy those lines into your `locals.yml` and start your Dashboard server via:
+
+```shell
+./bin/dashboard-server
+```
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,14 @@
+# Our entry into our docker development environment is described here.
+
+# For an overview of how to use the Docker Development Environment see `DOCKER.md`
+
+# It is split among a few different files that are then merged together via
+# the following `include` section.
+
+include:
+  # Services the dashboard server depends on
+  - path: ./docker/developers/docker-compose.mysql.yml
+  - path: ./docker/developers/docker-compose.redis.yml
+
+  # The dashboard server
+  - path: ./docker/developers/docker-compose.dashboard.yml
--- a/docker/developers/docker-compose.dashboard.yml
+++ b/docker/developers/docker-compose.dashboard.yml
@@ -1,0 +1,32 @@
+include:
+  - path: ./docker-compose.mysql.yml
+  - path: ./docker-compose.redis.yml
+  - path: ./docker-compose.networks.yml
+
+services:
+  # Runs all the dependent services when the dashboard server is running natively
+  dashboard-services:
+    image: alpine
+    depends_on:
+      mysql:
+        condition: service_healthy
+      redis:
+        condition: service_started
+    command: >
+      /bin/sh -c "
+        echo 'Dashboard services running.' &&
+        echo &&
+        echo 'To shut down all containers:' &&
+        echo 'docker compose stop' &&
+        echo &&
+        echo 'To run the native dashboard server:' &&
+        echo './bin/dashboard-server' &&
+        echo &&
+        echo 'Be sure to have the following in your locals.yml to link to the contained services:' &&
+        echo &&
+        echo 'db_writer: mysql://root:password@127.0.0.1/' &&
+        echo 'db_endpoint_writer: 127.0.0.1' &&
+        echo 'db_credential_admin:' &&
+        echo '  username: user' &&
+        echo '  password: password'
+      "
--- a/docker/developers/docker-compose.dashboard.yml
+++ b/docker/developers/docker-compose.dashboard.yml
@@ -23,10 +23,20 @@ services:
         echo './bin/dashboard-server' &&
         echo &&
         echo 'Be sure to have the following in your locals.yml to link to the contained services:' &&
+        echo 'These will allow you to connect to the database with: ./bin/mysql-client-admin' &&
         echo &&
-        echo 'db_writer: mysql://root:password@127.0.0.1/' &&
-        echo 'db_endpoint_writer: 127.0.0.1' &&
+        echo 'db_writer: mysql://root:password@127.0.0.1:${MYSQL_NATIVE_PORT:-3306}/' &&
+        echo 'db_endpoint_writer: 127.0.0.1:${MYSQL_NATIVE_PORT:-3306}' &&
         echo 'db_credential_admin:' &&
-        echo '  username: user' &&
+        echo '  username: root' &&
+        echo '  password: password' &&
+        echo 'db_endpoint_proxy_reporting: 127.0.0.1:${MYSQL_NATIVE_PORT:-3306}' &&
+        echo 'db_endpoint_proxy_reader: 127.0.0.1:${MYSQL_NATIVE_PORT:-3306}' &&
+        echo 'db_credential_reader:' &&
+        echo '  username: root' &&
+        echo '  password: password' &&
+        echo 'db_endpoint_proxy_writer: 127.0.0.1:${MYSQL_NATIVE_PORT:-3306}' &&
+        echo 'db_credential_writer:' &&
+        echo '  username: root' &&
         echo '  password: password'
       "
--- a/docker/developers/docker-compose.dashboard.yml
+++ b/docker/developers/docker-compose.dashboard.yml
@@ -25,6 +25,11 @@ services:
         echo 'Be sure to have the following in your locals.yml to link to the contained services:' &&
         echo 'These will allow you to connect to the database with: ./bin/mysql-client-admin' &&
         echo &&
+        echo '# Redis connections' &&
+        echo 'geocoder_redis_url: redis://localhost:${REDIS_NATIVE_PORT}/0/geocoder' &&
+        echo 'session_store_server: redis://localhost:${REDIS_NATIVE_PORT}/0/session' &&
+        echo &&
+        echo '# MySQL connections' &&
         echo 'db_writer: mysql://root:password@127.0.0.1:${MYSQL_NATIVE_PORT:-3306}/' &&
         echo 'db_endpoint_writer: 127.0.0.1:${MYSQL_NATIVE_PORT:-3306}' &&
         echo 'db_credential_admin:' &&
--- a/docker/developers/docker-compose.mysql.yml
+++ b/docker/developers/docker-compose.mysql.yml
@@ -1,0 +1,41 @@
+include:
+  - path: ./docker-compose.networks.yml
+
+services:
+  # Docker-managed database
+  mysql-base: &mysql-base
+    hostname: db
+    image: mysql/mysql-server:8.0
+    # You can add default options via the 'command':
+    #command:
+    # - '--log-error-verbosity=3'
+    environment:
+      MYSQL_DATABASE: 'db'
+      MYSQL_USER: 'user'
+      MYSQL_PASSWORD: 'password'
+      MYSQL_ROOT_PASSWORD: 'password'
+      MYSQL_ROOT_HOST: '%'
+    volumes:
+      - mysql_data:/var/lib/mysql
+    # Many default ulimits are just too high
+    # And, so, it will consume way too much memory (16GB+)
+    ulimits:
+      nofile:
+        soft: 65536
+        hard: 65536
+
+  mysql-contained:
+    <<: *mysql-base
+    expose:
+      - "3306"
+    networks:
+      cdo_network_test:
+      cdo_network:
+
+  mysql:
+    <<: *mysql-base
+    ports:
+      - "3306:3306"
+
+volumes:
+  mysql_data:
--- a/docker/developers/docker-compose.mysql.yml
+++ b/docker/developers/docker-compose.mysql.yml
@@ -24,6 +24,7 @@ services:
         soft: 65536
         hard: 65536
 
+  # Supports a Dockerized Dashboard server
   mysql-contained:
     <<: *mysql-base
     expose:
@@ -32,10 +33,11 @@ services:
       cdo_network_test:
       cdo_network:
 
+  # Runs a database visible to the native machine
   mysql:
     <<: *mysql-base
     ports:
-      - "3306:3306"
+      - "${MYSQL_NATIVE_PORT:-3306}:3306"
 
 volumes:
   mysql_data:
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -200,6 +200,7 @@ updates:
     directories:
     - "/aws/offsite/aurora-snapshots-fargate"
     - "/docker/dockerfiles"
+    - "/docker/developers"
     schedule:
       interval: "weekly"
       day: "sunday"
--- a/bin/mysql-client-admin
+++ b/bin/mysql-client-admin
@@ -5,11 +5,13 @@ require 'uri'
 
 raise 'please define CDO.db_credential_admin' unless CDO.db_credential_admin
 credentials = CDO.db_credential_admin
+host = CDO.db_endpoint_writer.split(':')[0]
+port = CDO.db_endpoint_writer.split(':')[1] || 3306
 db_connection_settings = URI::Generic.build(
   scheme: 'mysql',
   userinfo: credentials['username'] + ':' + credentials['password'],
-  host: CDO.db_endpoint_writer,
-  port: 3306,
+  host: host,
+  port: port,
   path: '/'
 )
 MysqlConsoleHelper.run(db_connection_settings, ARGV.join(' ').strip)
--- a/bin/mysql-client-dashboard-reader
+++ b/bin/mysql-client-dashboard-reader
@@ -5,11 +5,13 @@ require 'uri'
 
 raise 'please define CDO.db_endpoint_proxy_reader' unless CDO.db_endpoint_proxy_reader
 credentials = CDO.db_credential_reader
+host = CDO.db_endpoint_proxy_reader.split(':')[0]
+port = CDO.db_endpoint_proxy_reader.split(':')[1] || 3306
 db_connection_settings = URI::Generic.build(
   scheme: 'mysql',
   userinfo: credentials['username'] + ':' + credentials['password'],
-  host: CDO.db_endpoint_proxy_reader,
-  port: 3306,
+  host: host,
+  port: port,
   path: '/' + CDO.dashboard_db_name
 )
 MysqlConsoleHelper.run(db_connection_settings, ARGV.join(' ').strip)
--- a/bin/mysql-client-dashboard-reporting
+++ b/bin/mysql-client-dashboard-reporting
@@ -5,11 +5,13 @@ require 'uri'
 
 raise 'please set CDO.db_endpoint_proxy_reporting' unless CDO.db_endpoint_proxy_reporting
 credentials = CDO.db_credential_reader
+host = CDO.db_endpoint_proxy_reporting.split(':')[0]
+port = CDO.db_endpoint_proxy_reporting.split(':')[1] || 3306
 db_connection_settings = URI::Generic.build(
   scheme: 'mysql',
   userinfo: credentials['username'] + ':' + credentials['password'],
-  host: CDO.db_endpoint_proxy_reporting,
-  port: 3306,
+  host: host,
+  port: port,
   path: '/' + CDO.dashboard_db_name
 )
 MysqlConsoleHelper.run(db_connection_settings, ARGV.join(' ').strip)
--- a/bin/mysql-client-dashboard-writer
+++ b/bin/mysql-client-dashboard-writer
@@ -5,11 +5,13 @@ require 'uri'
 
 raise 'please define CDO.db_endpoint_writer' unless CDO.db_endpoint_writer
 credentials = CDO.db_credential_writer
+host = CDO.db_endpoint_proxy_writer.split(':')[0]
+port = CDO.db_endpoint_proxy_writer.split(':')[1] || 3306
 db_connection_settings = URI::Generic.build(
   scheme: 'mysql',
   userinfo: credentials['username'] + ':' + credentials['password'],
-  host: CDO.db_endpoint_writer,
-  port: 3306,
+  host: host,
+  port: port,
   path: '/' + CDO.dashboard_db_name
 )
 MysqlConsoleHelper.run(db_connection_settings, ARGV.join(' ').strip)
--- a/bin/mysql-client-pegasus-reader
+++ b/bin/mysql-client-pegasus-reader
@@ -5,11 +5,13 @@ require 'uri'
 
 raise 'please define CDO.db_endpoint_proxy_reader' unless CDO.db_endpoint_proxy_reader
 credentials = CDO.db_credential_reader
+host = CDO.db_endpoint_proxy_reader.split(':')[0]
+port = CDO.db_endpoint_proxy_reader.split(':')[1] || 3306
 db_connection_settings = URI::Generic.build(
   scheme: 'mysql',
   userinfo: credentials['username'] + ':' + credentials['password'],
-  host: CDO.db_endpoint_proxy_reader,
-  port: 3306,
+  host: host,
+  port: port,
   path: '/' + CDO.pegasus_db_name
 )
 MysqlConsoleHelper.run(db_connection_settings, ARGV.join(' ').strip)
--- a/bin/mysql-client-pegasus-reporting
+++ b/bin/mysql-client-pegasus-reporting
@@ -5,11 +5,13 @@ require 'uri'
 
 raise 'please define CDO.db_endpoint_proxy_reporting' unless CDO.db_endpoint_proxy_reporting
 credentials = CDO.db_credential_reader
+host = CDO.db_endpoint_proxy_reporting.split(':')[0]
+port = CDO.db_endpoint_proxy_reporting.split(':')[1] || 3306
 db_connection_settings = URI::Generic.build(
   scheme: 'mysql',
   userinfo: credentials['username'] + ':' + credentials['password'],
-  host: CDO.db_endpoint_proxy_reporting,
-  port: 3306,
+  host: host,
+  port: port,
   path: '/' + CDO.pegasus_db_name
 )
 MysqlConsoleHelper.run(db_connection_settings, ARGV.join(' ').strip)
--- a/bin/mysql-client-pegasus-writer
+++ b/bin/mysql-client-pegasus-writer
@@ -5,11 +5,13 @@ require 'uri'
 
 raise 'please define CDO.db_endpoint_writer' unless CDO.db_endpoint_writer
 credentials = CDO.db_credential_writer
+host = CDO.db_endpoint_proxy_writer.split(':')[0]
+port = CDO.db_endpoint_proxy_writer.split(':')[1] || 3306
 db_connection_settings = URI::Generic.build(
   scheme: 'mysql',
   userinfo: credentials['username'] + ':' + credentials['password'],
-  host: CDO.db_endpoint_writer,
-  port: 3306,
+  host: host,
+  port: port,
   path: '/' + CDO.pegasus_db_name
 )
 MysqlConsoleHelper.run(db_connection_settings, ARGV.join(' ').strip)
--- a/docker/developers/.env
+++ b/docker/developers/.env
@@ -1,0 +1,3 @@
+# The port to spin up a mysql server that is visible to the system (the default is 3306)
+# (This does not affect the port for the fully dockerized mysql container, that is always 3306)
+MYSQL_NATIVE_PORT: 3307
--- a/docker/developers/.env
+++ b/docker/developers/.env
@@ -1,3 +1,6 @@
 # The port to spin up a mysql server that is visible to the system (the default is 3306)
 # (This does not affect the port for the fully dockerized mysql container, that is always 3306)
 MYSQL_NATIVE_PORT: 3306
+
+# Similarly, the native port for redis
+REDIS_NATIVE_PORT: 6379
--- a/docker/developers/.env
+++ b/docker/developers/.env
@@ -1,3 +1,3 @@
 # The port to spin up a mysql server that is visible to the system (the default is 3306)
 # (This does not affect the port for the fully dockerized mysql container, that is always 3306)
-MYSQL_NATIVE_PORT: 3307
+MYSQL_NATIVE_PORT: 3306
--- a/docker/developers/docker-compose.networks.yml
+++ b/docker/developers/docker-compose.networks.yml
@@ -1,0 +1,5 @@
+networks:
+  cdo_network:
+    driver: bridge
+  cdo_network_test:
+    driver: bridge
--- a/docker/developers/docker-compose.redis.yml
+++ b/docker/developers/docker-compose.redis.yml
@@ -1,0 +1,21 @@
+include:
+  - path: ./docker-compose.networks.yml
+
+services:
+  # Docker-managed redis server
+  redis-base: &redis-base
+    hostname: redis
+    image: redis:latest
+
+  redis-contained: 
+    <<: *redis-base
+    expose:
+      - "6379"
+    networks:
+      cdo_network_test:
+      cdo_network:
+
+  redis:
+    <<: *redis-base
+    ports:
+      - "6379:6379"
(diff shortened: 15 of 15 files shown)


DOCKER.md
```diff
@@ -0,0 +1,123 @@
...
+```shell
+docker run hello-world
+```
+
+1. Verify Docker Compose version is at least 2.23.
+```shell
+docker compose version
+```
+
+## Running a Native Dashboard Server
```
cat5inthecradle, line 104: I'd suggest moving this section to the top - put the "Daily reference" details above the fold, and bury the first-time setup details a little bit.

DOCKER.md
```diff
@@ -0,0 +1,123 @@
...
+
+If you want to run the server code natively, but leverage Docker to run the dependent services, you can follow these instructions.
+
+First, you want to follow the normal [SETUP.md](SETUP.md) instructions for your platform.
+You can skip over many steps that are related to running mysql and redis.
```
cat5inthecradle, line 109: It might be worth including a note for folks who are already running mysql and/or redis natively - they'll need to stop those services or the `docker compose run` will fail due to in-use ports.

docker/developers/docker-compose.mysql.yml
```diff
@@ -0,0 +1,41 @@
...
+  mysql-contained:
+    <<: *mysql-base
+    expose:
+      - "3306"
+    networks:
+      cdo_network_test:
+      cdo_network:
+
+  mysql:
+    <<: *mysql-base
```
cat5inthecradle, line 38: YAML Merge keys and anchors are a very new concept to me - would it be worth some comments to explain this magic or link to docs? These symbols aren't especially googleable.

This stack overflow post cleared it up for me, but maybe there's a better source.
https://stackoverflow.com/a/50072961

cat5inthecradle (approved):
LGTM and works on my machine!

Are the `*-contained` services used in this iteration? I'm not really sure what's going on with those container definitions.

DOCKER.md
```diff
@@ -0,0 +1,123 @@
+# Development with Docker
```
snickell, line 1: Not very important, and a distinct element of bikeshedding, but I'd advocate putting this in SETUP.md instead of creating another top-level .md file.

docker-compose.yml
```diff
@@ -0,0 +1,17 @@
+# Our entry into our docker development environment is described here.
+
+# For an overview of how to use the Docker Development Environment see `DOCKER.md`
```
snickell, line 3: 👍, very happy to see a link to the docs from here, I'd definitely notice this file and it would lead me the right place

dashboard/test/ui/utils/selenium_browser.rb
```diff
@@ -14,7 +14,17 @@ def self.local(browser: :chrome, headless: true)
       options[:options] = Selenium::WebDriver::Firefox::Options.new
       options[:options].headless! if headless
       options[:options].add_argument('window-size=1280,1024')
+    when :edge
+      options[:options] = Selenium::WebDriver::Edge::Options.new
+      options[:options].headless! if headless
+      options[:options].add_argument('window-size=1280,1024')
```
snickell: Should we reverse the logic and not add this in case of safari?

snickell: ....or maybe we want this for safari too? 🤷

docker/developers/docker-compose.selenium.yml
```diff
@@ -0,0 +1,347 @@
...
+        echo 'Selenium Chrome Started.' &&
+        echo &&
+        echo 'To see and interact with this browser session:' &&
+        echo '- http://localhost:7910/?autoconnect=1&resize=scale&password=secret (VNC port 5910)' &&
+        echo &&
+        echo 'To shut down all selenium nodes, use:' &&
+        echo 'docker compose down selenium-chrome-native' &&
+        echo &&
+        echo 'To run a UI test with a locally running dashboard server:' &&
+        echo 'bundle exec rake test:ui browser=chrome selenium=http://localhost:4444/wd/hub feature=dashboard/test/ui/features/platform/signing_in.feature'
```
snickell: I appreciate the cli UX nicety here, printing the HOWTO on startup like this is much more likely to get this feature used imo

snickell (approved):
So happy to see this, I'm working on testing it locally now, but gets a strong ✅ from me based on read-through.

DOCKER.md
snickell: (100% ignorable if you disagree btw, I would not want to block forward progress here on something silly and very debatable)

snickell:
When I started running this, I get a bunch of non-breaking warnings:
```
code-dot-org-3.12➜  code-dot-org git:(adds-docker-compose-basics) docker compose run dashboard-services
WARN[0000] The "SELENIUM_EDGE_VERSION" variable is not set. Defaulting to a blank string. 
WARN[0000] The "SELENIUM_CHROME_VERSION" variable is not set. Defaulting to a blank string. 
WARN[0000] The "SELENIUM_FIREFOX_VERSION" variable is not set. Defaulting to a blank string. 
WARN[0000] The "SELENIUM_HUB_VERSION" variable is not set. Defaulting to a blank string. 
WARN[0000] The "SELENIUM_EDGE_VERSION" variable is not set. Defaulting to a blank string. 
WARN[0000] The "SELENIUM_CHROME_VERSION" variable is not set. Defaulting to a blank string. 
WARN[0000] The "SELENIUM_FIREFOX_VERSION" variable is not set. Defaulting to a blank string. 
WARN[0000] The "SELENIUM_HUB_VERSION" variable is not set. Defaulting to a blank string. 
WARN[0000] The "SELENIUM_CHROME_VERSION" variable is not set. Defaulting to a blank string. 
WARN[0000] The "SELENIUM_EDGE_VERSION" variable is not set. Defaulting to a blank string. 
WARN[0000] The "SELENIUM_CHROME_VERSION" variable is not set. Defaulting to a blank string. 
WARN[0000] The "SELENIUM_FIREFOX_VERSION" variable is not set. Defaulting to a blank string. 
WARN[0000] The "SELENIUM_FIREFOX_VERSION" variable is not set. Defaulting to a blank string. 
WARN[0000] The "SELENIUM_HUB_VERSION" variable is not set. Defaulting to a blank string. 
WARN[0000] The "SELENIUM_CHROME_VERSION" variable is not set. Defaulting to a blank string. 
WARN[0000] The "SELENIUM_FIREFOX_VERSION" variable is not set. Defaulting to a blank string. 
WARN[0000] The "SELENIUM_EDGE_VERSION" variable is not set. Defaulting to a blank string. 
WARN[0000] The "SELENIUM_CHROME_VERSION" variable is not set. Defaulting to a blank string. 
WARN[0000] The "SELENIUM_EDGE_VERSION" variable is not set. Defaulting to a blank string. 
WARN[0000] The "SELENIUM_EDGE_VERSION" variable is not set. Defaulting to a blank string. 
WARN[0000] The "SELENIUM_FIREFOX_VERSION" variable is not set. Defaulting to a blank string. 
```

snickell:
## UPDATE: these are totally issues with my local environment, I've been playing with replacing aws-google, and I managed to break my local AWS credentials in terrible ways, I'd assume these are not general issues

I love the flow you've got here where it prints out suggested modifications to locals.yml.

Confusingly, and almost certainly because I'm starting with a non-clean starting point here, I actually got one of our Google Auth prompts, when I started dashboard-server and when I logged in as @code.org I got:
```
code-dot-org-3.12➜  code-dot-org git:(adds-docker-compose-basics) bin/dashboard-server 
[DEPRECATED] `Bundler.with_clean_env` has been deprecated in favor of `Bundler.with_unbundled_env`. If you instead want the environment before bundler was originally loaded, use `Bundler.with_original_env` (called at /Users/seth/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/rerun-0.14.0/lib/rerun/notification.rb:74)

13:34:36 [rerun] Dashboard launched
13:34:36 [rerun] Rerun (35366) running Dashboard (35382)
2024-09-04 13:34:36 -1000 Using rack adapter
13:34:38 [rerun] Watching /Users/seth/src/code-dot-org/lib, /Users/seth/src/code-dot-org/shared/middleware, /Users/seth/src/code-dot-org/dashboard/legacy, /Users/seth/src/code-dot-org/pegasus for **/*.{rb,ru,yml} (ignoring **/migrations/*.rb,test/**/*.rb) with Darwin adapter
/Users/seth/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/concurrent-ruby-1.2.3/lib/concurrent-ruby/concurrent/promises.rb:1268:in `raise':  (Aws::STS::Errors::AccessDenied)
Your Google ID does not have access to the requested AWS Role. Ask your administrator to provide access.
Role: arn:aws:iam::165336972514:role/admin/Developer
Email: seth.nickell@code.org
Google ID: 101547480200094416574
        from /Users/seth/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/concurrent-ruby-1.2.3/lib/concurrent-ruby/concurrent/promises.rb:1268:in `wait_until_resolved!'
        from /Users/seth/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/concurrent-ruby-1.2.3/lib/concurrent-ruby/concurrent/promises.rb:998:in `value!'
        from /Users/seth/src/code-dot-org/lib/cdo/secrets.rb:75:in `get!'
        from /Users/seth/.rbenv/versions/3.0.5/lib/ruby/3.0.0/delegate.rb:87:in `method_missing'
        from /Users/seth/src/code-dot-org/lib/cdo/secrets_config.rb:66:in `lookup'
        from /Users/seth/src/code-dot-org/lib/cdo/secrets_config.rb:153:in `block (2 levels) in lazy_load_secrets!'
        from /Users/seth/.rbenv/versions/3.0.5/lib/ruby/3.0.0/delegate.rb:320:in `__getobj__'
        from /Users/seth/src/code-dot-org/lib/cdo/lazy.rb:25:in `__getobj__'
        from /Users/seth/src/code-dot-org/lib/cdo/secrets_config.rb:165:in `block (2 levels) in lazy_load_secrets!'
        from /Users/seth/src/code-dot-org/lib/cdo/slack.rb:41:in `<class:Slack>'
        from /Users/seth/src/code-dot-org/lib/cdo/slack.rb:7:in `<top (required)>'
        from /Users/seth/src/code-dot-org/lib/cdo/chat_client.rb:6:in `require'
        from /Users/seth/src/code-dot-org/lib/cdo/chat_client.rb:6:in `<top (required)>'
        from /Users/seth/src/code-dot-org/lib/cdo/data/csv_to_sql_table.rb:2:in `require'
        from /Users/seth/src/code-dot-org/lib/cdo/data/csv_to_sql_table.rb:2:in `<top (required)>'
        from /Users/seth/src/code-dot-org/pegasus/data/static_models.rb:1:in `require'
        from /Users/seth/src/code-dot-org/pegasus/data/static_models.rb:1:in `<top (required)>'
        from /Users/seth/src/code-dot-org/lib/cdo/db.rb:4:in `require'
        from /Users/seth/src/code-dot-org/lib/cdo/db.rb:4:in `<top (required)>'
        from /Users/seth/src/code-dot-org/lib/cdo/poste.rb:2:in `require'
        from /Users/seth/src/code-dot-org/lib/cdo/poste.rb:2:in `<top (required)>'
        from /Users/seth/src/code-dot-org/dashboard/config/application.rb:2:in `require'
        from /Users/seth/src/code-dot-org/dashboard/config/application.rb:2:in `<top (required)>'
        from /Users/seth/src/code-dot-org/dashboard/config/environment.rb:2:in `require'
        from /Users/seth/src/code-dot-org/dashboard/config/environment.rb:2:in `<top (required)>'
        from /Users/seth/src/code-dot-org/dashboard/config.ru:7:in `require'
        from /Users/seth/src/code-dot-org/dashboard/config.ru:7:in `block in <main>'
        from /Users/seth/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/rack-2.2.9/lib/rack/builder.rb:125:in `instance_eval'
        from /Users/seth/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/rack-2.2.9/lib/rack/builder.rb:125:in `initialize'
        from /Users/seth/src/code-dot-org/dashboard/config.ru:1:in `new'
        from /Users/seth/src/code-dot-org/dashboard/config.ru:1:in `<main>'
        from /Users/seth/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/thin-1.8.2/lib/rack/adapter/loader.rb:33:in `eval'
        from /Users/seth/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/thin-1.8.2/lib/rack/adapter/loader.rb:33:in `load'
        from /Users/seth/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/thin-1.8.2/lib/rack/adapter/loader.rb:42:in `for'
        from /Users/seth/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/thin-1.8.2/lib/thin/controllers/controller.rb:170:in `load_adapter'
        from /Users/seth/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/thin-1.8.2/lib/thin/controllers/controller.rb:74:in `start'
        from /Users/seth/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/thin-1.8.2/lib/thin/runner.rb:203:in `run_command'
        from /Users/seth/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/thin-1.8.2/lib/thin/runner.rb:159:in `run!'
        from /Users/seth/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/thin-1.8.2/bin/thin:6:in `<top (required)>'
        from /Users/seth/.rbenv/versions/3.0.5/bin/thin:25:in `load'
        from /Users/seth/.rbenv/versions/3.0.5/bin/thin:25:in `<main>'
/Users/seth/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/aws-sdk-core-3.201.3/lib/seahorse/client/plugins/raise_response_errors.rb:17:in `call': Not authorized to perform sts:AssumeRoleWithWebIdentity (Aws::STS::Errors::AccessDenied)
        from /Users/seth/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/aws-sdk-core-3.201.3/lib/aws-sdk-core/plugins/checksum_algorithm.rb:111:in `call'
        from /Users/seth/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/aws-sdk-core-3.201.3/lib/aws-sdk-core/plugins/jsonvalue_converter.rb:16:in `call'
        from /Users/seth/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/aws-sdk-core-3.201.3/lib/aws-sdk-core/plugins/invocation_id.rb:16:in `call'
        from /Users/seth/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/aws-sdk-core-3.201.3/lib/aws-sdk-core/plugins/idempotency_token.rb:19:in `call'
        from /Users/seth/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/aws-sdk-core-3.201.3/lib/aws-sdk-core/plugins/param_converter.rb:26:in `call'
        from /Users/seth/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/aws-sdk-core-3.201.3/lib/seahorse/client/plugins/request_callback.rb:89:in `call'
        from /Users/seth/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/aws-sdk-core-3.201.3/lib/aws-sdk-core/plugins/response_paging.rb:12:in `call'
        from /Users/seth/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/aws-sdk-core-3.201.3/lib/seahorse/client/plugins/response_target.rb:24:in `call'
        from /Users/seth/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/aws-sdk-core-3.201.3/lib/seahorse/client/request.rb:72:in `send_request'
        from /Users/seth/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/aws-sdk-core-3.201.3/lib/aws-sdk-sts/client.rb:1645:in `assume_role_with_web_identity'
        from /Users/seth/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/aws-google-0.2.0/lib/aws/google.rb:171:in `refresh'
        from /Users/seth/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/aws-google-0.2.0/lib/aws/google/cached_credentials.rb:29:in `block in refresh_if_near_expiration'
        from /Users/seth/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/aws-google-0.2.0/lib/aws/google/cached_credentials.rb:27:in `synchronize'
        from /Users/seth/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/aws-google-0.2.0/lib/aws/google/cached_credentials.rb:27:in `refresh_if_near_expiration'
        from /Users/seth/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/aws-google-0.2.0/lib/aws/google/cached_credentials.rb:22:in `initialize'
        from /Users/seth/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/aws-google-0.2.0/lib/aws/google.rb:63:in `initialize'
        from /Users/seth/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/aws-google-0.2.0/lib/aws/google/credential_provider.rb:35:in `new'
        from /Users/seth/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/aws-google-0.2.0/lib/aws/google/credential_provider.rb:35:in `google_credentials_from_config'
        from /Users/seth/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/aws-google-0.2.0/lib/aws/google/credential_provider.rb:15:in `google_credentials'
        from /Users/seth/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/aws-sdk-core-3.201.3/lib/aws-sdk-core/credential_provider_chain.rb:13:in `block in resolve'
        from /Users/seth/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/aws-sdk-core-3.201.3/lib/aws-sdk-core/credential_provider_chain.rb:12:in `each'
        from /Users/seth/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/aws-sdk-core-3.201.3/lib/aws-sdk-core/credential_provider_chain.rb:12:in `resolve'
        from /Users/seth/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/aws-sdk-core-3.201.3/lib/aws-sdk-core/plugins/credentials_configuration.rb:73:in `block in <class:CredentialsConfiguration>'
        from /Users/seth/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/aws-sdk-core-3.201.3/lib/seahorse/client/configuration.rb:72:in `call'
        from /Users/seth/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/aws-sdk-core-3.201.3/lib/seahorse/client/configuration.rb:215:in `block in resolve_defaults'
        from /Users/seth/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/aws-sdk-core-3.201.3/lib/seahorse/client/configuration.rb:59:in `each'
        from /Users/seth/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/aws-sdk-core-3.201.3/lib/seahorse/client/configuration.rb:59:in `each'
        from /Users/seth/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/aws-sdk-core-3.201.3/lib/seahorse/client/configuration.rb:214:in `resolve_defaults'
        from /Users/seth/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/aws-sdk-core-3.201.3/lib/seahorse/client/configuration.rb:207:in `value_at'
        from /Users/seth/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/aws-sdk-core-3.201.3/lib/seahorse/client/configuration.rb:191:in `block in resolve'
        from /Users/seth/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/set-1.0.3/lib/set.rb:511:in `each_key'
        from /Users/seth/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/set-1.0.3/lib/set.rb:511:in `each'
        from /Users/seth/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/aws-sdk-core-3.201.3/lib/seahorse/client/configuration.rb:191:in `resolve'
        from /Users/seth/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/aws-sdk-core-3.201.3/lib/seahorse/client/configuration.rb:179:in `apply_defaults'
        from /Users/seth/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/aws-sdk-core-3.201.3/lib/seahorse/client/configuration.rb:152:in `build!'
        from /Users/seth/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/aws-sdk-core-3.201.3/lib/seahorse/client/base.rb:65:in `build_config'
        from /Users/seth/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/aws-sdk-core-3.201.3/lib/seahorse/client/base.rb:21:in `initialize'
        from /Users/seth/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/aws-sdk-secretsmanager-1.48.0/lib/aws-sdk-secretsmanager/client.rb:335:in `initialize'
        from /Users/seth/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/aws-sdk-core-3.201.3/lib/seahorse/client/base.rb:102:in `new'
        from /Users/seth/src/code-dot-org/lib/cdo/secrets.rb:32:in `block in client_promise'
        from /Users/seth/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/concurrent-ruby-1.2.3/lib/concurrent-ruby/concurrent/promises.rb:1593:in `evaluate_to'
        from /Users/seth/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/concurrent-ruby-1.2.3/lib/concurrent-ruby/concurrent/promises.rb:1776:in `block in on_resolvable'
        from /Users/seth/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/concurrent-ruby-1.2.3/lib/concurrent-ruby/concurrent/executor/ruby_thread_pool_executor.rb:359:in `run_task'
        from /Users/seth/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/concurrent-ruby-1.2.3/lib/concurrent-ruby/concurrent/executor/ruby_thread_pool_executor.rb:350:in `block (3 levels) in create_worker'
        from /Users/seth/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/concurrent-ruby-1.2.3/lib/concurrent-ruby/concurrent/executor/ruby_thread_pool_executor.rb:341:in `loop'
        from /Users/seth/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/concurrent-ruby-1.2.3/lib/concurrent-ruby/concurrent/executor/ruby_thread_pool_executor.rb:341:in `block (2 levels) in create_worker'
        from /Users/seth/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/concurrent-ruby-1.2.3/lib/concurrent-ruby/concurrent/executor/ruby_thread_pool_executor.rb:340:in `catch'
        from /Users/seth/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/concurrent-ruby-1.2.3/lib/concurrent-ruby/concurrent/executor/ruby_thread_pool_executor.rb:340:in `block in create_worker'
```

Not sure if this is related, it could be independent breakage.

DOCKER.md
```diff
@@ -0,0 +1,123 @@
...
+```shell
+brew install docker
+```
+
+1. Install Docker Compose:
+    1. ```shell
+       export DOCKER_CONFIG=${DOCKER_CONFIG:-$HOME/.docker}
+       mkdir -p $DOCKER_CONFIG/cli-plugins
+       curl -SL https://github.com/docker/compose/releases/download/v2.27.0/docker-compose-darwin-$(uname -m) -o $DOCKER_CONFIG/cli-plugins/docker-compose
+       chmod +x $DOCKER_CONFIG/cli-plugins/docker-compose
```
stephenliang, line 28: To avoid the docs from becoming stale, would it make sense to link to the docker compose installation documentation instead?

DOCKER.md
```diff
@@ -0,0 +1,123 @@
...
+```
+
+1. Start the Colima service (and have it start on login)
+```shell
+brew services start colima
+```
+
+1. Verify that Docker works
+```shell
+docker run hello-world
```
stephenliang:
```suggestion
docker run --rm hello-world
```

Adding the `--rm` flag will remove the container on termination, this will be useful for the hello world example since we are likely only doing this once.

DOCKER.md
```diff
@@ -0,0 +1,123 @@
...
+```
+
+### Ubuntu
+
+1. Open a terminal.
+
+1. Install Docker via `apt`:
```
stephenliang, line 51: Would it make sense to link to the Docker setup instructions instead of having them here? This would allow us to not have to keep updating them as these do change from time to time. Perhaps keeping only the things that is needed to run Code.org would be best left for this documentation.

docker/developers/docker-compose.dashboard.yml
```diff
@@ -0,0 +1,32 @@
...
+        echo &&
+        echo 'To shut down all containers:' &&
+        echo 'docker compose stop' &&
+        echo &&
+        echo 'To run the native dashboard server:' &&
+        echo './bin/dashboard-server' &&
+        echo &&
+        echo 'Be sure to have the following in your locals.yml to link to the contained services:' &&
+        echo &&
+        echo 'db_writer: mysql://root:password@127.0.0.1/' &&
```
stephenliang: Is this the actual password or a placeholder? If placeholder, perhaps `[your password here]` may help to disambiguate

docker/developers/docker-compose.dashboard.yml
```diff
@@ -0,0 +1,32 @@
+include:
+  - path: ./docker-compose.mysql.yml
+  - path: ./docker-compose.redis.yml
+  - path: ./docker-compose.networks.yml
+
+services:
+  # Runs all the dependent services when the dashboard server is running natively
+  dashboard-services:
+    image: alpine
```
stephenliang, line 9: Out of curiosity, did we consider `ubuntu-server` for consistency with our Production cluster or does `alpine` make the most sense?

docker/developers/docker-compose.mysql.yml
```diff
@@ -0,0 +1,41 @@
...
+    hostname: db
+    image: mysql/mysql-server:8.0
+    # You can add default options via the 'command':
+    #command:
+    # - '--log-error-verbosity=3'
+    environment:
+      MYSQL_DATABASE: 'db'
+      MYSQL_USER: 'user'
+      MYSQL_PASSWORD: 'password'
+      MYSQL_ROOT_PASSWORD: 'password'
```
stephenliang, line 16: Instead of having a defined password, would it make sense to enable "allow empty password"?

DOCKER.md
```diff
@@ -0,0 +1,123 @@
...
+1. Install Docker itself:
+```shell
+brew install docker
+```
+
+1. Install Docker Compose:
+    1. ```shell
+       export DOCKER_CONFIG=${DOCKER_CONFIG:-$HOME/.docker}
+       mkdir -p $DOCKER_CONFIG/cli-plugins
+       curl -SL https://github.com/docker/compose/releases/download/v2.27.0/docker-compose-darwin-$(uname -m) -o $DOCKER_CONFIG/cli-plugins/docker-compose
```
bencodeorg, line 27: I'm on an M1 mac and this didn't work, `uname -m` returns `arm64` and the releases are suffixed `aarch64`, which the internet tells me is effectively the same thing? ([release link for this version](https://github.com/docker/compose/releases/download/v2.27.0/docker-compose-darwin-aarch64))

DOCKER.md
```diff
@@ -0,0 +1,123 @@
...
+Our Docker development environment requires at least Docker Compose version 2.23 or higher.
+These instructions will walk you through installing a compatible version.
+
+To install Docker, see the appropriate section below:
+[macOS](#macos), [Ubuntu](#ubuntu), [Windows](#windows)
+or consult your package management for your OS.
+
+### macOS
+
+1. Install Docker itself:
```
bencodeorg, line 18: I also had Docker Desktop installed from a previous foray into Docker, which was the preferred version on the command line. Might be worth telling people to check that `which docker` returns the Homebrew installation (ie, `/opt/homebrew/bin/docker`)

DOCKER.md
Hamms: Lol, not to hop on the bikeshedding train or anything, but I have an in-between suggestion. I think the amount of docker-specific information that we plan to add as part of the follow-up PRs more than justifies having a standalone document independent of SETUP.md, but I agree I don't love adding yet another top-level file.

So, what if we just put this somewhere like `docker/developers/README.md`? Particularly since we're referencing it in a comment in the top-level `docker-compose.yml`, I don't think we'd lose much discoverability by moving it deeper into the directory structure.

docker/developers/docker-compose.mysql.yml
```diff
@@ -0,0 +1,41 @@
...
+    volumes:
+      - mysql_data:/var/lib/mysql
+    # Many default ulimits are just too high
+    # And, so, it will consume way too much memory (16GB+)
+    ulimits:
+      nofile:
+        soft: 65536
+        hard: 65536
+
+  mysql-contained:
```
Hamms, line 28: This might be my docker ignorance showing, but is it worth adding an explanation for why we're defining three different-but-related services here, and what exactly each of them is responsible for? Even just some comments like the ones in `docker-compose.selenium.yml` would be helpful!

Or is this a common pattern used in docker compose that I can read up on elsewhere?

docker/developers/docker-compose.dashboard.yml
Hamms: My understanding is that `alpine` is intended to be a minimal base image, and that the only purpose of this service is to depend on some other services and echo some output to the console. It makes sense to me that we wouldn't care as much about compatibility with production for this particular part of the local development environment as we would for the part that's actually *running* the dashboard server.

docker-compose.yml
```diff
@@ -0,0 +1,17 @@
...
+# It is split among a few different files that are then merged together via
+# the following `include` section.
+
+include:
+  # Services the dashboard server depends on
+  - path: ./docker/developers/docker-compose.mysql.yml
+  - path: ./docker/developers/docker-compose.redis.yml
+
+  # Selenium services for UI testing
+  - path: ./docker/developers/docker-compose.selenium.yml
```
Hamms: Is the selenium stuff added by this PR meant to be here, or is it all intended for https://github.com/code-dot-org/code-dot-org/pull/60240? It's certainly not a problem to do em all at once, I'm just curious if the MySQL and Redis services have any dependence on the Selenium one.

docker/developers/docker-compose.dashboard.yml
```diff
@@ -0,0 +1,32 @@
...
+        echo &&
+        echo 'To run the native dashboard server:' &&
+        echo './bin/dashboard-server' &&
+        echo &&
+        echo 'Be sure to have the following in your locals.yml to link to the contained services:' &&
+        echo &&
+        echo 'db_writer: mysql://root:password@127.0.0.1/' &&
+        echo 'db_endpoint_writer: 127.0.0.1' &&
+        echo 'db_credential_admin:' &&
+        echo '  username: user' &&
```
bencodeorg: Is this intentionally not `root`? Introduced a bit of confusion as our `mysql-client-admin` script uses these credentials and doesn't end up with full visibility into schema/tables created with the `db_writer` credentials.

Hamms (commented):
I really like the pattern this PR establishes as well as the broader plan to incrementally add functionality. I am fully on board for the mysql+redis components of this PR!

One thing I am worried about is the long-term maintainability of the current Selenium implementation. It's really impressive and runs way faster and smoother than I was expecting! But that is one intimidating yaml file, and a lot of the details it's configuring are not things I'd know what to do with if we wanted to update or tweak anything in the future. I'm not entirely sure what improvements I would want to see here, though. Maybe just breaking it up a bit into a few difference files and adding some more documentation?

Also, heads up @wilkie that the Engineering Excellence team is starting to look into some options for containerized and AWS-less development environments, and - in case it wasn't obvious - we all agreed that these PRs seem like a great place to start! Hope we aren't overwhelming ya with feedback :smile: Please let me know if you'd like us to swarm on responding to various feedback items, too! I'd be happy to start making some commits to this branch if you've got other work you'd rather focus on, but I definitely don't want to step on any toes if you've still got things you want to do with this!

DOCKER.md
```diff
@@ -0,0 +1,123 @@
...
+       export DOCKER_CONFIG=${DOCKER_CONFIG:-$HOME/.docker}
+       mkdir -p $DOCKER_CONFIG/cli-plugins
+       curl -SL https://github.com/docker/compose/releases/download/v2.27.0/docker-compose-darwin-$(uname -m) -o $DOCKER_CONFIG/cli-plugins/docker-compose
+       chmod +x $DOCKER_CONFIG/cli-plugins/docker-compose
+       ```
+    1. You can inspect that the version is appropriate by running `docker compose version` and inspecting the result.
+
+1. Install Colima as a container runtime:
+```shell
+brew install colima
```
bencodeorg, line 34: I ran into issues where the MySQL container would terminate randomly (observed consistently when running `bundle exec rake install`) -- we identified it as a "out of memory" issue in [this thread](https://codedotorg.slack.com/archives/C07LFGRN422/p1726176328741309).

It looks like Colima has a 2 GB memory limit by default. Unfortunately, it doesn't look like you pass options like these via `brew services start colima`, which leaves 3 slightly annoying possible recommendations:
1. Open `~/.colima/default/colima.yaml` and edit the memory config.
2. `colima template`, which opens the file to be edited on the command line.
3. run `colima start --memory 4`, which starts colima directly and also appears to modifies the template.

Worth noting that this was only an issue if I had the Selenium containers running at the same time. I'm not sure what the limit should be, but the 2 GB limit enforced by default by Colima is much lower than what Elijah saw on his system (I think Docker is given the system's memory by default).

docker/developers/docker-compose.mysql.yml
wilkie: Why? It doesn't make anything simpler, does it?

docker/developers/docker-compose.mysql.yml
wilkie: To support the native dashboard server, we need a service that exposes the ports on the host network and, for the case of a dashboard server inside a container, one that exposes ports on the virtual network. You cannot do both very easily... so the idea here is to define a network-less base and then two "real" services: `mysql` which is for connecting to from your native machine and `mysql-contained` which should be hidden neatly behind the dockerized Dashboard server in the final "full docker" PR.

dashboard/test/ui/utils/selenium_browser.rb
wilkie: We don't test safari locally. Though, that would be nice. Not sure exactly what you are advocating, however, Reversing what logic?

DOCKER.md
wilkie: I need help with the Mac instructions. I do not have a test machine and online ones do not play well with Docker for the most part. :(

DOCKER.md
wilkie: grrr that this happens. I guess it needs an extra instruction for arm-based Macs.

DOCKER.md
wilkie: Is there such documentation?

docker/developers/docker-compose.mysql.yml
Hamms: Oh, that makes sense; thanks for the explanation! I would definitely love to see some comments which make that motivation explicit; even if it's as simple as labeling the services as "for docker dashboard" and "for native dashboard"

DOCKER.md
wilkie: Macs will virtualize the memory for Docker, so you would have to configure Docker (via colima or docker desktop) to pre-allocate a certain amount. The database will consume exactly as much as the ulimit allows (including ALL of it) and that will kill the database service. It is a very annoying problem. It is interesting that the selenium containers are sharing that memory with the database.

DOCKER.md
```diff
@@ -0,0 +1,123 @@
...
+```
+
+1. Start the Colima service (and have it start on login)
+```shell
+brew services start colima
+```
+
+1. Verify that Docker works
+```shell
+docker run --rm hello-world
```
wilkie, line 44: I don't mind this `--rm` inclusion. It's generally true for the other `run` steps to automatically throw away the exited container.

docker-compose.yml
wilkie: Hmm yes... it really should have been in the other PR. 😵‍💫

wilkie: The mysql and redis services have no dependence on the selenium services at all. This is probably making the second PR harder to review. dang.

docker/developers/docker-compose.dashboard.yml
wilkie: That seems true that it could be `root` here.

DOCKER.md
stephenliang: It looks like the recommended approach is to install [docker desktop](https://docs.docker.com/compose/install/) which includes both docker and docker compose. The cask version of docker for Mac does install docker desktop.

wilkie:
I have moved the Selenium bits into PR 2 like they were supposed to be. This PR is just the basic docker compose stuff.

docker/developers/docker-compose.dashboard.yml
wilkie: In PR 4, you can see the fully contained dashboard server which will indeed use ubuntu to match the version.

docker/developers/docker-compose.mysql.yml
wilkie: I'm having some trouble knowing where such documentation should go. This is mostly just an educational thing related to informing anybody reading these how this piece of YAML trivia about how this non-standard implementation works. That documentation would be as such:

This uses YAML anchor expansion syntax, which is supported by Docker Compose, to help keep duplicated content cleanly in one place within the Docker Compose files. With this, you can name blocks of YAML like this:

```
my-service: &my-service-base
  hostname: foo
```

And then use it to pre-populate the block with the contents of that prior block before your additional items:

```
my-service-hosted:
  <<: *my-service-base
  network: host
```

This service will have the properties of its named block and then the additional fields. Any fields that exist in both will be overwritten when read top-down, so in the prior case, if we also specify `hostname` at the end, it will override it. If we specify it as the first line, it will be overwritten by the expanded properties instead.

wilkie:
To address some issues that might arise when you are trying to deploy the Dockerized database alongside a native database, I've amended it a little bit to allow you to set what database port you want. The `MYSQL_NATIVE_PORT` variable is used to define this.

For instance, you can do this:

```
MYSQL_NATIVE_PORT=3307 docker compose run --rm dashboard-services
```

And it will tell you:

```
Dashboard services running.

To shut down all containers:
docker compose stop

To run the native dashboard server:
./bin/dashboard-server

Be sure to have the following in your locals.yml to link to the contained services:
These will allow you to connect to the database with: ./bin/mysql-client-admin

db_writer: mysql://root:password@127.0.0.1:3307/
db_endpoint_writer: 127.0.0.1:3307
db_credential_admin:
  username: root
  password: password
db_endpoint_proxy_reporting: 127.0.0.1:3307
db_endpoint_proxy_reader: 127.0.0.1:3307
db_credential_reader:
  username: root
  password: password
db_endpoint_proxy_writer: 127.0.0.1:3307
db_credential_writer:
  username: root
  password: password
```

This will allow and enable all the `./bin/mysql-*`-style commands and the dashboard server itself to directly use this alternative port.

DOCKER.md
bencodeorg: I'm well qualified play the role of "well meaning but Docker clueless developer with a Mac" 😆

DOCKER.md
bencodeorg: Is it worth pursuing suggesting Docker Desktop as the way of managing Docker on Macs? I did see that the [memory issue](https://github.com/code-dot-org/code-dot-org/pull/60274#discussion_r1761661360) could potentially be managed via a nice slider in the Docker Desktop UI (although I'm not sure I'm totally following whether adjusting the amount of memory available to the VM will sort out the MySQL OOM errors I've encountered)?

PR #60275, opened 2024-08-08
make isCurrentLevelLab2 optional

mikeharv:
Script overview pages are showing the following error in the console:

image (214)

This is because the Teacher Panel expects the `isCurrentLevelLab2` property to be required. However, there's no level when we're on a script overview page. The solution is to not make this property required.

Slack thread: https://codedotorg.slack.com/archives/C044AGTKW3D/p1723133710098249


sanchitmalhotra126 (approved):
Thanks for the catch and quick fix!

breville (approved):
Thanks for fixing!

PR #60276, opened 2024-08-08
Adds docker compose for s3

wilkie:
## PR Checklist:

- [ ] Code is well-commented


docker-compose.yml
```diff
@@ -13,5 +13,40 @@ include:
...
 
+  # MinIO S3 Emulation
+  - path: ./docker/developers/docker-compose.minio.yml
+
   # The dashboard server
   - path: ./docker/developers/docker-compose.dashboard.yml
+
+services:
+  all-services:
+    image: alpine
```
stephenliang, line 24: Curious if we've considered `ubuntu-server` versus `alpine` for consistency?

lib/cdo/aws/s3.rb
```diff
@@ -87,6 +133,13 @@ def self.exists_in_bucket(bucket, key)
       create_client.head_object(bucket: bucket, key: key)
       return true
     rescue Aws::S3::Errors::NotFound, Aws::S3::Errors::Forbidden
+      if CDO.aws_s3_emulated?
```
stephenliang, line 136: Should we log out something here?

docker-compose.yml
wilkie: No no. This just prints out some text to the screen. A libc+busybox distro is all that is necessary.

lib/cdo/aws/s3.rb
wilkie: Oh, I could probably get rid of this, but this would, in theory, be used to redirect to the content on production to "backfill" the local buckets. But it is not implemented fully.

DOCKER.md
```diff
@@ -110,11 +110,19 @@ You can skip over many steps that are related to running mysql and redis.
 
 Instead, once you have a working Ruby and Node environment, you can then use this command to spin up the database and redis servers:
```
bencodeorg, line 111: add note about S3 emulation as well?

bencodeorg:
@wilkie would you be open to rebasing this onto the `adds-docker-compose-basics` branch rather than the Selenium one? I think we have a clearer path/motivation for merging this one, and I'm inclined not to block it on figuring out the nooks and crannies of the Selenium work?

docker/developers/install-s3.sh
```diff
@@ -0,0 +1,108 @@
...
+
+export AWS_ENDPOINT_URL=${AWS_S3_ENDPOINT_URL}
+
+# Determine all the buckets that exist in the s3 path and create those buckets
+cd s3
+for path in `ls ./*/ -d`; do
+  path=$(basename "${path}")
+  bucket_name=${path//_/-}
+
+  if [ ! $(aws s3api head-bucket --bucket ${bucket_name}) ]; then
```
bencodeorg, line 23: ...this may be incredibly obvious, but these are creating fake buckets via Minio, right? Is there any way to make that more clear to future readers?

config.yml.erb
```diff
@@ -266,6 +266,14 @@ google_client_secret:
 # Default AWS region for the application.
 aws_region: us-east-1
 
+# Set the AWS endpoint, typically to point to a local instance of
+# an AWS service. For instance, an emulated LocalStack solution.
+# When the endpoint is set, you can also set the access key, etc.
+aws_endpoint:
```
bencodeorg, line 270: Related to confusion about "what is emulated S3 and what is real S3", these are only used when `aws_s3_emulated` is true, right? Could we rename these keys to that effect?

docker/developers/s3/cdo_sound_library/hoc_song_meta/populate.rb
```diff
@@ -0,0 +1,102 @@
...
+        f.write(data)
+      end
+    end
+
+    to
+  end
+
+  # Hook into the populate call to produce the appropriate mp3 as well
+  def populate(path = nil)
+    data = super(path)
```
bencodeorg, line 45: A little bit of noob with Ruby modules/etc, why not just call download(path) directly here (and same with the first call to `populate` in `populate_all` below?

lib/cdo.rb
```diff
@@ -282,6 +282,23 @@ def ci_webserver?
       running_web_application? && ENV.fetch('CI', nil)
     end
 
+    # The web server instance has some intention to use AWS credentials?
+    def aws_access?
+      rack_env?(:production) || !!ENV['AWS_PROFILE']&.include?('cdo')
+    end
+
+    # The web server in connected to an emulated (non-production) AWS system.
+    def aws_emulated?
```
bencodeorg, line 291: `aws_emulated` is currently not really distinct from S3 emulation (yet), right? Maybe we drop this in the short term to avoid confusion?

lib/cdo/aws/cloudfront.rb
```diff
@@ -323,6 +318,18 @@ def self.signed_cookies(resource, expiration_date)
         ]
       }.to_json
 
+      # If we are emulating AWS at all, we just return mock cloudfront policy output
+      if CDO.aws_emulated? || CDO.aws_s3_emulated?
+        return {
+          'CloudFront-Policy-Emulated': Base64.encode64(policy).tr('+=/', '-_~')
```
bencodeorg, line 325: For my understanding, what does this do?

lib/cdo/aws/s3.rb
```diff
@@ -69,11 +79,47 @@ def self.create_client
...
+
+      base.new.populate(File.join(*relative_path)) if base && relative_path
+    end
+
     # Returns the value of the specified S3 key in bucket.
     # @param [String] bucket
     # @param [String] key
     # @return [String]
     def self.download_from_bucket(bucket, key, options = {})
+      if CDO.aws_s3_emulated?
```
bencodeorg, line 117: This pattern is very cool!

lib/cdo/aws/s3.rb
```diff
@@ -69,11 +79,47 @@ def self.create_client
       connect_v2!
     end
 
+    # Populates locally hosted buckets for development environments without full AWS access.
+    def self.populate_local_bucket(bucket, key)
+      return unless CDO.aws_s3_emulated?
+      return if exists_in_bucket(bucket, key)
+
+      # Determine the Populator class that can generate files for this bucket, if it
```
bencodeorg, line 87: This logic (ie, find the right path and class if they exist) feels more like a utility that it would be nice to house alongside the other populator code, rather than nested in here. I'm not sure of the ruby-preferred way to do that though?

bencodeorg (commented):
This is super cool to see, and the "populator" concept is really interesting! I can see how we'd extend it over time as we want to flesh out the experience of working with emulated  S3 going forward.

One general question I had was about how you think about managing all the little special sauces we're sprinkling in various places to support S3 emulation going forward...is it just unavoidable that we'll end up with some little snippets in various places to support it? Possible (but painful?) to centralize it a bit more?

docker/developers/s3/.gitignore
```diff
@@ -0,0 +1,5 @@
+*.*
```
bencodeorg, line 1: Based on the .gitignore / folder structure, I was expecting files to end up in the directories for a given emulated bucket...is that incorrect? I see things working as expected when I use the Minio UI to navigate to a given bucket, but the files aren't actually in the folders.

bencodeorg: (I should clarify that I do see mp3/json files being added the `cdo_sound_library` and `cdo_restricted` directories, just not files or sources like `main.json`)

docker/developers/install-s3.sh
wilkie: Sure. This is the normal `aws` command line usage to create buckets. It has nothing to do with MinIO outside of the fact that MinIO accepts the AWS CLI commands. If we want, we can of course link to the aws cli documentation: https://awscli.amazonaws.com/v2/documentation/api/latest/reference/index.html and specifically https://awscli.amazonaws.com/v2/documentation/api/latest/reference/s3api/index.html

docker/developers/s3/.gitignore
wilkie: The files that make up the local S3 buckets are in docker volumes. Take a look at the `minio` service description. The files in the paths are sourced from somewhere else, so have to be downloaded / generated into those directories, and then uploaded to our emulated "s3". The files in the directories are a cache of sorts... so if they exist already and you destroy your local S3 buckets, they will quickly populate again from them.

docker/developers/s3/cdo_sound_library/hoc_song_meta/populate.rb
wilkie: Because we are overriding the call, basically. The call usually "downloads" the file, but we want to do something after we download the file... here, we want to parse it.

The `download` call is probably best seen as a private or protected call. It duplicates a file that we have upstream via our `studio.code.org/api` calls to kinda lazily populate our local buckets... gets around the need for AWS access. The `populate` call is there to be overridable and is intended to be called when a request for that file happens.

lib/cdo.rb
wilkie: Sure! Go ahead and patch it.

lib/cdo/aws/cloudfront.rb
wilkie: It emulates the CloudFront signer policy needed to allow `/restricted/` (signed cookie) access to the bucket. MinIO nor LocalStack actually emulate CloudFront signing... but it is pretty easy to pretend to do it anyway. This does that.

This policy is checked when the `/restricted/` asset path gets a request and it should have a cookie with the right policy info. It just needs to match... so we can just spit it out as though it happened and it will still more or less effectively emulate signed bucket access.

It allows for this check to succeed: https://github.com/code-dot-org/code-dot-org/pull/60276/files/0965896c56ff2b421b95f073b5f6845f5fc5fe6e#diff-10cfca81b682d8a8836e2f5bf0ea04f83264368ad0df41214ea8ebd53be18ac9R86

lib/cdo/aws/s3.rb
wilkie: "ruby-preferred" depends on you! what do you think it should be?

lib/cdo/aws/s3.rb
wilkie: It's pretty neat. I did not want to add a seeding process or track what files exist and such... just wanted a way to, as things are requested, to copy them over and treat the local s3 buckets (at least in terms of the static assets we store there) as a kind of cache... but still have them work just the same way as production.

Now, we could really consider fixing all of the hardcoded URLs to those assets in our level data... the dev site is still going to hit the production buckets for assets more often than not, which can pose a problem for running it in limited environments (or just geographically far from our data centers) and/or where this technique could help facilitate offline deployments.

docker/developers/s3/.gitignore
bencodeorg: Gotcha! Another q then -- why do the `cdo_v3_files` and `cdo_v3_sources` directories exist then? I see they get used when deciding which minio S3 buckets to create in `install-s3.sh`, but do they have another purpose (eventually house `populate.rb` file(s))?

wilkie: Yeah, it tells that script to make that bucket, and then path would contain a `README.md` to explain to devs what that bucket is for, what limitations exist in the dev environment, etc, I'd think. And any configuration / populators that make the bucket work.

docker/developers/install-s3.sh
```diff
@@ -0,0 +1,108 @@
...
+aws s3api put-object --bucket cdo-sound-library --key hoc_song_meta/ --content-length 0
+
+if ! aws s3api head-bucket --bucket cdo-restricted; then
+  echo "Creating s3 bucket cdo-restricted"
+  aws s3api create-bucket --bucket cdo-restricted --object-lock-enabled-for-bucket
+else
+  echo "Creating s3 bucket cdo-restricted [EXISTS]"
+fi
+
+if ! aws s3api head-bucket --bucket cdo-ai; then
```
bencodeorg, line 96: What is the distinction between buckets like these (no associated directory) and buckets like `cdo_v3_files` (has associated directory)? These mostly? look like what happens in the loop above? I also see a little bit of special sauce for `cdo-sound-library`, is that the only one that is supposed to be doing something special, and the rest are created in the same way?

docker/developers/docker-compose.minio.yml
```diff
@@ -0,0 +1,113 @@
...
+      - minio1_storage:/data1
+      - minio2_storage:/data2
+      - minio3_storage:/data3
+      - minio4_storage:/data4
+    command: server --console-address ":3001" /data1 /data2 /data3 /data4
+
+  minio:
+    <<: *minio-base
+    environment: &minio-env
+      MINIO_VOLUMES: "http://localhost:9000/data{1...4}"
```
bencodeorg, line 21: Will using port 9000 here conflict with our use of port 9000 for webpack dev server here?

https://github.com/code-dot-org/code-dot-org/blob/093b289ffc52a81d1e69c83610d776da9436fbeb/apps/webpack.config.js#L33

lib/cdo/aws/s3.rb
```diff
@@ -47,7 +47,17 @@ def initialize(message = nil)
     # the credentials specified in the CDO config.
     # @return [Aws::S3::Client]
     def self.connect_v2!
-      self.s3 ||= Aws::S3::Client.new
+      self.s3 ||= if CDO.aws_s3_emulated?
```
bencodeorg, line 50: An interesting finding this morning...TIL we have a second S3 client for "middleware" (files/sources/etc):

https://github.com/code-dot-org/code-dot-org/blob/093b289ffc52a81d1e69c83610d776da9436fbeb/dashboard/legacy/middleware/helpers/bucket_helper.rb#L18

S3 emulation seems to be working though, even without the ternary like we've got here. [I think that makes sense given the hierarchy of credentials listed here](https://rubydoc.info/gems/aws-sdk-s3/1.98.0/Aws%2FS3%2FClient:initialize). Do you think it makes sense to put this ternary in place everywhere we use S3, or rely on the defaults to do the work for us everywhere?

docker/developers/install-s3.sh
bencodeorg: Oh, there's a sneaky little `exit` on line 33 I missed. So do we need the rest of this script (lines 34 onward)?

lib/cdo/aws/s3.rb
bencodeorg: Oh nvm, got turned around. This uses the same S3 client as the one where your changes are, so we're good!

docker/developers/s3/cdo_sound_library/hoc_song_meta/populate.rb
```diff
@@ -0,0 +1,102 @@
...
+      end
+
+      # Ensure we write the audio data to the cdo-restricted bucket
+      put('cdo-restricted', "restricted/#{file_name}", -> {File.read(file_path)})
+    end
+
+    data
+  end
+
+  def populate_all
```
bencodeorg, line 93: As written, this never gets called...right?

PR #60285, opened 2024-08-08
NAV - add tests for sidebar

kobryan0619:
Added unit test for sidebar.  Updated small style changes including:

- Color of header text
image

- Width of dropdown (ensure it goes across the sidebar)
image

## Links

[Figma](https://www.figma.com/design/5ILfiJkPQpjskBwksMDeZl/Teacher-Dashboard?node-id=4121-18427&m=dev)


apps/test/unit/templates/teacherDashboard/teacherNavigation/TeacherNavigationBarTest.tsx
```diff
@@ -0,0 +1,63 @@
...
+      <Provider store={store}>
+        <TeacherNavigationBar />
+      </Provider>
+    );
+
+    const classLabel = screen.getByText(i18n.classSections());
+    expect(classLabel).toBeDefined();
+
+    const dropdown = screen.getByRole('combobox');
+    expect(dropdown).toBeDefined();
```
lfryemason: We shouldn't need this line because `getByText` already throws an error if it isn't found. Same with classLabel above and options below
```suggestion
```

lfryemason (approved):
Approved with one comment

apps/test/unit/templates/teacherDashboard/teacherNavigation/TeacherNavigationBarTest.tsx
kobryan0619: I have always felt uneasy leaving `screen.getByRole` as the sole test that something is in the document.  I understand, technically, it is just as useful as using an assertion, but assertions feel so much more readable to me and the intention feels clear.  This is probably in part a hold over from chai. 

Have you seen any standards as a team that would prefer one over another?  I noticed a lot of Denys' components to follow the pattern of getting component and then asserting they are there.  Since he has been the most consistent contributor of TypeScript components with tests I was following his lead.  @levadadenys do you have any insights about conversations around this through your work?

Again, happy to agree to disagree (re: readability and clarity), just trying to get a sense of the bigger trend/sentiment in the org.

PR #60287, opened 2024-08-08
Change JWKS per prod and non-prod env

nicklathe:
Currently, we use the same JWKS for prod and non-prod environments. This PR updates those JWKS to be separate between environments via the following:

- Configure public JWKS in config.yml.erb for all non-prod environments
- Override a different public JWKS in config/production.yml.erb for prod
- Remove dashboard/app/public/oauth/jwks.json

There was discussion about putting the JWKS in secrets manager, as we do the private key. However in order to add a key to the JWKS array, we'd first have to pull the existing JWKS object from Secrets Manager. Our current IAM polices don't allow either EngineeringFullAccess or EngineeringReadOnly to read/pull secrets that are used on our production servers. As a result, the decision was made to store the JWKS object in the appropriate config files.

**Follow-up work**
- Once this change has been deployed, we will need to update the non-prod environments with the new private key, via the normal secret creation/update flow.
- Currently both the production (currently used) JWK and the new non-prod JWK are configured to return at the JWKS endpoint. This is so we don't break the non-prod envs during this key rotation. Once the new private key has been pushed to secrets manager and is live (via a DTP), we can subsequently remove the production JWK from config.yml.erb.

## Links
[Jira](https://codedotorg.atlassian.net/browse/P20-336)

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/bin/generate-jwks
+++ b/bin/generate-jwks
@@ -1,24 +1,33 @@
 #!/usr/bin/env ruby
 
 # This script is meant to be run locally on a dev machine (not a server), and it
-# is used to generate new JSON Web Keys (JWK) to be added to the
-# JSON Web Key Set (JWKS) exposed at /oauth/jwks. It generates both the public
-# key and the private key. The public key will be added to the JWKS in
-# dashboard/public/oauth/jwks.json. It will print the private key and
-# key id (kid) that is to be used for updating the secret at
-# CDO.jwk_private_key_data.
-
+# is used to generate new JSON Web Keys (JWKS). It generates both the public key
+# and the private key. The public key will be formated in YAML, to simplify adding
+# it to the approprate config files (see below). The private key will be formated
+# as JSON, which is how it should be stored in Secrets Manager.
+#
+# Here are the locations the public key should be added
+# - For Production, the public key should be added to config/production.yml.erb
+# - For Non-Prod, the public key should be added to config.yml.erb
+#
 # When you include 'dev' as an argument, it will write the JWK and the private
 # key to your locals.yml. This should be done once for local development.
+#
+# You can find more information about the JWKS implimentation in docs/jwks.md
 
 require 'jwt'
 require 'json'
+require 'yaml'
+require 'active_support/core_ext/hash'
 
-allowed_args = ["dev", "prod"].freeze
-# default to 'prod'
-arg = ARGV.first || "prod"
+allowed_args = ["dev", "prod", "non-prod"].freeze
 
-puts "Usage: 'dev' or 'prod' as arguments...\n" unless allowed_args.include? arg
+@arg = ARGV.first
+
+unless allowed_args.include?(@arg)
+  puts "Usage: 'dev', 'prod', or 'non-prod' as arguments...\n"
+  return
+end
 
 params = {use: 'sig', alg: 'RS256'}
 jwk = JWT::JWK.new(OpenSSL::PKey::RSA.new(2048), params, kid_generator: JWT::JWK::Thumbprint)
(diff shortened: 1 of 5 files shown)


config.yml.erb
```diff
@@ -616,7 +616,20 @@ cdo_amplitude_api_key:
...
-jwks_data: <%=File.read("#{File.expand_path(File.dirname(__FILE__))}/dashboard/public/oauth/jwks.json")%>
+jwks_data:
+  keys:
+    - kty: "RSA"
+      n: "o-pwMF3Hf7iy_Rev-XEf4aPxCYxYiOevIXZVrQuQPVU8JE80XcNlrq8az8c3afp4G79nAVRWDxWTsYQXTNnr-HEBBklO_w5M-h6wIgMCnA9VDJtATgIlqNJ5hIQ37U_1Mmw3oLYJVXaKhkjD-GNFTF0r6rKjeAOp4sWoKTn83r2Tyi8ZGM2l3IQ9NYsurFrAT7DALu01V5EsM_sdBezc6r-LkGyTkIQqS4qDV7qesJE9cUe2FVXfnedWLeYBCsq-tIh50XVMB1FMFa97kN3UfS3rigQYAq-LkDc3e7rHB0DqduHJNZKiHBoYOra484N4nmDYj-cSoBkDzhyu84ni4w"
+      e: "AQAB"
+      use: "sig"
+      alg: "RS256"
+      kid: "yNt3MOdUenNV_bn6IKboK2d8iJBL2hDJ_RoMhUtO33Q"
+    - kty: "RSA"
```
nicklathe, line 621: Temporarily include the prod public key in the JWKS object for non-prod environments. This is so we don't break those envs during this key rotation, and can be removed once the private key has been applied in those envs.

cat5inthecradle (approved):

docs/jwks.md
```diff
@@ -14,19 +14,24 @@ allows us to start using a new key when signing JWTs, while still allowing users
...
    the [secrets process](../config/secrets.md)
-1. After the next deployment, code will start using the new private key when
+1. After a following DTP, code will start using the new private key when
    signing JWTs, and the `kid` in those JWTs will point to the new JWK
 
 After you are sure that any JWTs issued with the old key have expired, you can:
 
-1. Remove the old JWK from the `keys` array in the JWKS object at
-   `dashboard/public/oauth/jwks.json`
+1. Remove the old JWK from the `keys` array in the appropriate config file`
```
Hamms: nit:
```suggestion
1. Remove the old JWK from the `keys` array in the appropriate config file
```

docs/jwks.md
```diff
@@ -14,19 +14,24 @@ allows us to start using a new key when signing JWTs, while still allowing users
 to validate JWTs issued with the old key. To generate a new JWK, refer to the
 following steps:
 
-1. Run `bin/generate-jwks` locally on your dev machine (not on a server)
-1. This will push the new JWK to the JWKS array in
-   `dashboard/public/oauth/jwks.json`
-1. Open and merge a PR with the change
+1. Run `bin/generate-jwks <env>` locally on your dev machine (not on a server)
```
Hamms: Worth specifying here that `<env>` should be one of `'dev', 'prod', or 'non-prod'`?

bin/generate-jwks
```diff
@@ -1,24 +1,33 @@
...
-# CDO.jwk_private_key_data.
-
+# is used to generate new JSON Web Keys (JWKS). It generates both the public key
+# and the private key. The public key will be formated in YAML, to simplify adding
+# it to the approprate config files (see below). The private key will be formated
+# as JSON, which is how it should be stored in Secrets Manager.
+#
+# Here are the locations the public key should be added
+# - For Production, the public key should be added to config/production.yml.erb
+# - For Non-Prod, the public key should be added to config.yml.erb
```
Hamms: Might be worth incorporating a bit of an explanation of these config files into this instruction; something like,

```suggestion
# The public key used by default should be defined in `config.yml.erb`, for use in
# non-production environments. For the production environment, a production-specific
# public key override should be added to `config/production.yml.erb`.
```

Hamms (approved):
I have a couple of nitpicky suggestions for possible improvements, but regardless this LGTM. Thanks!

carl-codeorg (approved):
LGTM

docs/jwks.md
nicklathe: Thanks for calling this out, I had this as a passing thought as well, but kept moving. I'll add this clarification.

unlox775-code-dot-org (approved):

PR #60288, opened 2024-08-08
Use new DCSO TextField in school association

TurnerRiley:
When the new school association was developed, the DCSO `TextField` component wasn't ready yet. Now that it is, this PR implements it. It ends up looking nearly identical but simplifies the styles needed a bit and will make it easier to incorporate school association elsewhere like in building the new sign-up flow.

This PR also fixes a tiny bug I found with the school association in whether it defaults country to the U.S. or not.

### Old look
AAAAAAAAAAAAAAAAAAAA

### New look (nearly identical)
BBBBBBBBBBBBBBBBBBBB

### Now uses a placeholder for the zip code
zip placeholder

## Testing story


apps/src/templates/SchoolDataInputs.jsx
```diff
@@ -33,7 +33,7 @@ export default function SchoolDataInputs({
     detectedCountry === US_COUNTRY_CODE
   );
   const [isOutsideUS, setIsOutsideUS] = useState(
-    detectedCountry && detectedCountry !== US_COUNTRY_CODE
+    (detectedCountry || usIp === false) && detectedCountry !== US_COUNTRY_CODE
```
TurnerRiley, line 36: While testing, I found a bug that if `usIp` was passed in as `false`, `isOutsideUS` would still return `false` because it would process `detectedCountry` as `''` (meaning the school name field wouldn't automatically show up). This handles the case now for when `usIp` is `false`, rather than treating `false` and `undefined` as the same for `usIp`.

apps/i18n/common/en_us.json
```diff
@@ -1156,6 +1156,7 @@
   "enterYourAnswerHere": "Enter your answer here",
   "enterYourName": "Enter Your Name",
   "enterYourSchoolZip": "Enter your school's zip code:",
+  "enterYourSchoolZipPlaceholder": "30305",
```
bethanyaconnor: I don't think this needs to be translated. Though I'm curious where this zip came from and what best practices are here 🤔

apps/src/templates/SchoolDataInputs.jsx
bethanyaconnor, line 36: nit: use `!usIp`

bethanyaconnor: If `usIp` is `false` shouldn't `isOutsideUS` still be false?

TurnerRiley: I set it to `=== false` because if `usIp` is `undefined` then we do want to default to the U.S. I believe, but if it's `false` then we know it's explicitly not the U.S. Does that seem okay or I can also find another way to restructure the if/else's?

apps/i18n/common/en_us.json
TurnerRiley: Oh good point! I just copied the zip from the Figma, but I can make it something else too if we want something more generic like "00000"?

apps/src/templates/SchoolDataInputs.jsx
TurnerRiley: I believe we want the flipped behavior, so if `usIp` is `false` then we know they're outside of the U.S. so `isOutsideUS` should be true (and vice versa). But if `usIp` is `undefined` then we default to `isOutsideUS` being `false`

apps/i18n/common/en_us.json
bethanyaconnor: Let's do that and remove it from this file! It can just be hardcoded

apps/src/templates/SchoolDataInputs.jsx
bethanyaconnor: ugh right right, double negatives are hard

bethanyaconnor: I think I'm getting tripped up on needing separate behavior for undefined, which is a bit of a code smell

bethanyaconnor: okay okay, so it sounds like the original bug is that this line was true when `detectedCountry === ''`, right? Could we change this to just be `!!detectedCountry`?

bethanyaconnor: (Sorry, I'm just trying to understand what's going on at this point. A lot of logic here)

TurnerRiley: Super fair! It's just that right now `usIp` isn't really doing anything since we are prioritizing `detectedCountry`, but then if we have no `detectedCountry` then regardless of if `usIp` is `true`, `false`, or `undefined`, it will still cause `isOutsideUS` to be `false`. So we could also remove `usIp` and just always default to U.S. if no `detectedCountry` is present?

apps/i18n/common/en_us.json
moshebaricdo: LOL that was my old zip, we should def change it to just 00000 😭

moshebaricdo:
LGTM! Thank you for making these updates!!

apps/src/templates/SchoolDataInputs.jsx
TurnerRiley: Yes for sure! The main issue is that `''` is considered `false` when it's checking the first `detectedCountry` on line 36. So if `sessionStorage.getItem(SCHOOL_COUNTRY)` is empty, then `detectedCountry` is only relying on `(usIp ? US_COUNTRY_CODE : '')`. If `usIp` is `true`, then `detectedCountry && detectedCountry !== US_COUNTRY_CODE` returns `false` for `isOutsideUS` (which is what we want). However, if `usIp` is `false` (for sure outside of the U.S.) or `undefined` (unclear if outside U.S.), `(usIp ? US_COUNTRY_CODE : '')` will be evaluated as `''` in both cases. So `usIp` doesn't change the outcome (i.e. if `usIp` is `false`, we aren't treating it like we know that the user is for sure outside of the U.S.): if we have no `detectedCountry`, then `isOutsideUS` is always `false`.

So I think we can either update the logic to actually rely on `usIp` or remove `usIp` alltogether!

bethanyaconnor: ooo I think I see. So we should probably be defaulting to assuming outside the US if `usIp` is false, but we're not doing that.

And then there's also the question of how we want to treat `usIp` as undefined, right?

bethanyaconnor: I think I'm convinced but would love @hannahbergam to take a look when she's back. I think this PR is blocking you so I'll approve and you can merge, then we can follow up if Hannah has feedback

bethanyaconnor (approved):

apps/src/templates/SchoolDataInputs.jsx
TurnerRiley: Yep that's right! Sounds good, thank you!

PR #60300, opened 2024-08-09
seth/pretty-in-py: Python.run API

snickell:
Followup to the big PyCall PR, this creates a longer term Python API atop the PyCall library.

This is in turn extended by this PR which backs the hollow Python.run API with the real deal, PyCallThread: https://github.com/code-dot-org/code-dot-org/pull/60334


lib/python/venv.rb
```diff
@@ -0,0 +1,71 @@
+require 'pty'
+
```
snickell, line 2: This file is just a rename of `/lib/cdo/python_venv.rb`, but the indentation change to make it Python::Venv instead of PythonVenv caused git to think its a new file.

dashboard/test/lib/python_test.rb
```diff
@@ -0,0 +1,84 @@
+require 'test_helper'
+require 'python/python'
+
```
snickell, line 3: This file is mostly a rename of `pycall_test.rb` with the addition of putting python calls inside a `Python.run()` block, which made git think it was a new file not a rename.

dashboard/test/lib/python_test.rb
```diff
@@ -0,0 +1,84 @@
...
+  test 'pyimport a module from pycdo' do
+    # From /python/pycdo/test_module/example_func.py
+    result = Python.run do
+      pyfrom 'pycdo.test_module', import: :example_func
+      example_func
+    end
+    assert_equal 'Ruby can call Python!', result
+  end
+
```
snickell, line 21: The three following tests concerning `Python::ThreadSafetyError` are new:

PR #60307, opened 2024-08-09
Adds Dockerfile and docker commands to contain the dashboard server

wilkie:
This adds the `Dockerfile` and a set of convenience commands to control the running Dashboard server completely within Docker.

The setup instructions (in `DOCKER.md`) will guide you, but it is just:

```
FIXUID=${id -u} docker compose build web selenium-video
docker compose run install-all
docker compose run build
docker compose up web
```

This will start a dashboard server at `http://localhost-studio.code.org:3000` also accessible via port `3000` on your host machine. It starts all relevant services.

The `install-all` command runs essentially the `rake install` script (along with bundle install and friends) and the `build` command runs `rake build`. These are essential parts of the process that aren't quick (nor quickly finish if they aren't needed), but they are required as part of our development process.

The `web` service encapsulates our `./bin/dashboard-server` command while forcing some config parameters to values compatible with the use of the services running alongside such as MySQL and Redis, among others. This is done via environment variables as seen in `docker/developers/docker-compose.dashboard.yml`.

## PR Checklist:

- [ ] Code is well-commented

### DIFF
--- a/DOCKER.md
+++ b/DOCKER.md
@@ -112,7 +112,7 @@ then skip ahead to the next section and all other sections referring to the "Nat
 For the fully contained setup, you need to build the Dashboard code-dot-org Docker image:
 
 ```shell
-FIXUID=$(id -u) docker compose build web test selenium-video
+FIXUID=$(id -u) docker compose build web selenium-video
 ```
 
 Then we install all the necessary libraries and such into our environment:
@@ -134,6 +134,69 @@ docker compose up web
 ```
 
 And then navigate to `http://localhost-studio.code.org:3000`.
+
+To shell into the development environment:
+
+```
+docker compose run shell
+```
+
+## Running tests and other interactions with the Dashboard Server
+
+From outside of the container, you can access most operations from convenient
+Docker services. For instance, you can run any rake command without a local Ruby
+environment via:
+
+```
+docker compose run rake <task>
+```
+
+For instance `docker compose run rake install` will run the `rake install` command
+within the running `web` container. Of course, this is a common operation which
+we've provided as part of the `install-all` command above.
+
+For running rake within the `dashboard` or `pegasus` paths, use `dashboard-rake` or
+`pegasus-rake`, respectively, instead.
+
+### UI Testing
+
+You can run the UI tests with the `ui-test` service like so where any argument
+is interpreted as the feature file you want to run:
+
+```
+docker compose run ui-test dashboard/test/ui/features/platform/signing_in.feature
+```
+
+This will run it via the running `web` container (starting it in the background
+if you haven't started it already) and run it on the local Selenium Grid service.
+
+It will tell you how to view the visuals of the running test via a local URL, most
+likely: `http://localhost:7910/?autoconnect=1&resize=scale&password=secret` which
+you can also use to interact with the test browser. Different browsers you are
+using will run in different ports. This one (`7910`) is Chrome, which is the
+default browser target.
+
+You can run on specific browsers by appending the browser name to the service:
+
+```
+docker compose run ui-test-firefox dashboard/test/ui/features/platform/signing_in.feature
+docker compose run ui-test-edge dashboard/test/ui/features/platform/signing_in.feature
+```
+
+You can also record a video of the testing session via the `record-ui-test-...` options:
+
+```
+docker compose run record-ui-test-edge dashboard/test/ui/features/platform/signing_in.feature
+```
+
+This will create a `ui-test.edge.mp4` file in your local `/tmp` path. To change the name of
+the video file use the `record=` environment variable:
+
+```
+record=my-video docker compose run record-ui-test-edge dashboard/test/ui/features/platform/signing_in.feature
+```
+
+This will create `/tmp/my-video.edge.mp4` instead.
 
 ## Running a Native Dashboard Server
 
--- a/DOCKER.md
+++ b/DOCKER.md
@@ -101,6 +101,40 @@ docker run hello-world
 docker compose version
 ```
 
+## Running a fully Docker Serup
+
+These steps will use Docker Compose to run everything entirely within Docker. This is the easiest
+path since it relegates all responsibility for the development environment to Docker.
+
+If you want to run a native server and just use Docker to handle the services, such as the database,
+then skip ahead to the next section and all other sections referring to the "Native Dashboard Server".
+
+For the fully contained setup, you need to build the Dashboard code-dot-org Docker image:
+
+```shell
+FIXUID=$(id -u) docker compose build web test selenium-video
+```
+
+Then we install all the necessary libraries and such into our environment:
+
+```
+docker compose run install-all
+```
+
+Then build:
+
+```
+docker compose run build
+```
+
+Now we can start the dashboard server:
+
+```
+docker compose up web
+```
+
+And then navigate to `http://localhost-studio.code.org:3000`.
+
 ## Running a Native Dashboard Server
 
 If you want to run the server code natively, but leverage Docker to run the dependent services, you can follow these instructions.
--- a/docker/developers/Dockerfile
+++ b/docker/developers/Dockerfile
@@ -1,0 +1,191 @@
+# Web Environment
+# ----------------
+
+# Ubuntu 20.04
+FROM ubuntu:focal AS cdo-web
+
+ARG USERNAME=cdodev
+
+ARG DEBIAN_FRONTEND=noninteractive
+ENV TZ=America/Los_Angeles
+
+# Additional required packages
+RUN apt-get update \
+    && apt-get -y install --no-install-recommends ca-certificates \
+    && apt-get -y install --no-install-recommends \
+    autoconf gcc g++ bison build-essential libssl-dev libyaml-dev libreadline-dev \
+    zlib1g-dev libncurses-dev libffi-dev git git-lfs parallel debian-archive-keyring
+
+# Set the locale
+RUN apt-get -y install locales
+RUN sed -i '/en_US.UTF-8/s/^# //g' /etc/locale.gen && \
+    locale-gen
+ENV LANG en_US.UTF-8  
+ENV LANGUAGE en_US:en  
+ENV LC_ALL en_US.UTF-8 
+
+# Set the non-root user for the container and switch
+ARG UID
+ARG GID
+RUN apt-get -y install sudo
+RUN groupadd -g ${GID} ${USERNAME} \
+    && useradd -r -u ${UID} -g ${USERNAME} --shell /bin/bash --create-home ${USERNAME} \
+    && echo "${USERNAME} ALL=NOPASSWD: ALL" >> /etc/sudoers \
+    && chown -R ${USERNAME} /usr/local
+USER ${USERNAME}
+
+# Install ImageMagick
+RUN sudo -E apt-get -y install --no-install-recommends libmagickwand-dev imagemagick
+
+# MySQL Dependencies
+RUN sudo -E apt-get -y install --no-install-recommends cmake wget libaio-dev
+
+# Install MySQL Native Client
+RUN sudo -E apt-get -y install --no-install-recommends mysql-client-8.0 libmysqlclient-dev
+
+# Install rbenv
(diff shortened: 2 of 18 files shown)


DOCKER.md
```diff
@@ -101,6 +101,40 @@ docker run hello-world
...
+These steps will use Docker Compose to run everything entirely within Docker. This is the easiest
+path since it relegates all responsibility for the development environment to Docker.
+
+If you want to run a native server and just use Docker to handle the services, such as the database,
+then skip ahead to the next section and all other sections referring to the "Native Dashboard Server".
+
+For the fully contained setup, you need to build the Dashboard code-dot-org Docker image:
+
+```shell
+FIXUID=$(id -u) docker compose build web test selenium-video
```
stephenliang: Curious why is the `FIXUID` necessary?

DOCKER.md
```diff
@@ -101,6 +101,40 @@ docker run hello-world
...
+```
+
+Then we install all the necessary libraries and such into our environment:
+
+```
+docker compose run install-all
```
stephenliang, line 121: Is this possible to run as part of the entrypoint to the docker container (to avoid the need to run this command)? Sort of like how Rails does seeding

docker/developers/Dockerfile
```diff
@@ -0,0 +1,203 @@
+# Web Environment
+# ----------------
+
+# Ubuntu 20.04
+FROM ubuntu:focal AS cdo-web
```
stephenliang:
```suggestion
FROM ubuntu:20.04 AS cdo-web
```

docker/developers/Dockerfile
```diff
@@ -0,0 +1,203 @@
...
+
+# Additional required packages
+RUN apt-get update \
+    && apt-get -y install --no-install-recommends ca-certificates \
+    && apt-get -y install --no-install-recommends \
+    autoconf gcc g++ bison build-essential libssl-dev libyaml-dev libreadline-dev \
+    zlib1g-dev libncurses-dev libffi-dev git git-lfs parallel debian-archive-keyring
+
+# Set the locale
+RUN apt-get -y install locales
```
stephenliang, line 20: This comment applies to all the `apt-get install` that we do, could we inline all of the dependencies in a single large `apt-get install` so we can take advantage of the layer and cache it (avoid having to constantly run the installs if a layer mismatches?)

docker/developers/Dockerfile
```diff
@@ -0,0 +1,203 @@
...
+# Install Ruby. Also replace the system ruby (required for RubyMine debugging).
+ARG RUBY_VERSION=3.0.5
+RUN rbenv install ${RUBY_VERSION} \
+    && echo -n '\n# rbenv init\neval "$(rbenv init -)"\n' >> ~/.bashrc \
+    && rbenv global ${RUBY_VERSION} \
+    && sudo rm /usr/bin/ruby \
+    && sudo ln -s /home/${USERNAME}/.rbenv/versions/${RUBY_VERSION} /usr/bin/ruby
+
+# Install nvm
+ARG NODE_VERSION=18.16.0
```
stephenliang, line 63: I don't believe we are pinned to a specific minor version of node, so we can use `18` for `NODE_VERSION` ( I didn't see the equivalent for NVM though)

```suggestion
ARG NODE_VERSION=18
```

docker/developers/Dockerfile
```diff
@@ -0,0 +1,203 @@
...
+
+# Install Yarn
+ARG YARN_VERSION=1.22.19
+RUN . $NVM_DIR/nvm.sh \
+    && npm i -g yarn@${YARN_VERSION} \
+    && sudo mkdir -p /opt/base-nvm \
+    && sudo chown -R ${USERNAME} /opt/base-nvm \
+    && cp -r ${NVM_DIR}/* /opt/base-nvm/. \
+    && rm -f /opt/base-nvm/README.md \
+    && sudo chown -R ${USERNAME} /opt/base-nvm
```
stephenliang, line 81: Could you explain what the `base-nvm` piece is for?

docker/developers/Dockerfile
```diff
@@ -0,0 +1,203 @@
...
+    curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v${NVM_VERSION}/install.sh | bash \
+    && . $NVM_DIR/nvm.sh \
+    && nvm install $NODE_VERSION \
+    && nvm alias default $NODE_VERSION \
+    && nvm use default
+
+# Install Yarn
+ARG YARN_VERSION=1.22.19
+RUN . $NVM_DIR/nvm.sh \
+    && npm i -g yarn@${YARN_VERSION} \
```
stephenliang, line 76: For yarn, I believe we should be using `corepack enable` instead of installing it via npm. The Code.org instructions (under install node.js) as well as [official Yarn docs](https://yarnpkg.com/getting-started/install) seem to indicate this.

docker/developers/Dockerfile
```diff
@@ -0,0 +1,203 @@
...
+# Add Ruby binaries to path
+RUN echo -n "\n# Add Ruby binaries on path\nexport PATH=\$PATH:/home/${USERNAME}/.rbenv/versions/${RUBY_VERSION}/bin\n" >> ~/.profile
+
+# Install node-pre-gyp (required for Web packaging)
+RUN sudo -E apt install -y --no-install-recommends node-pre-gyp
+
+# Install python2 which might be required by some dependencies during their compilation
+RUN sudo -E apt install -y --no-install-recommends python2
+
+# Make temporary directory and do a bundle install
```
stephenliang, line 92: I wonder for the steps before this point, if we should have that publish a `cdo-base` image as it is mostly OS environment setup which would allow it to be cached (and skip all the install steps). If cpu architecture is going to cause issues, we can probably defer that to an optimization later on. I do think it would be good to minimize runtime on OS environment setup if we can though.

docker/developers/Dockerfile
```diff
@@ -0,0 +1,203 @@
...
+ARG BUNDLER_VERSION=2.3.22
+RUN sudo mkdir -p /app/src
+COPY Gemfile /app/src/.
+COPY Gemfile.lock /app/src/.
+COPY .ruby-version /app/src/.
```
stephenliang, line 97:
```suggestion
COPY Gemfile Gemfile.lock .ruby-version /app/src/.
```

docker/developers/Dockerfile
```diff
@@ -0,0 +1,203 @@
...
+    && eval "$(rbenv init -)" \
+    && gem install bundler -v ${BUNDLER_VERSION} \
+    && RAILS_ENV=development bundle config set --local without staging production test levelbuilder \
+    && RAILS_ENV=development bundle install \
+    && sudo cp -r /home/${USERNAME}/.rbenv /opt/base-rbenv \
+    && sudo chown -R ${USERNAME} /opt/base-rbenv
+
+# Install python and pdm
+RUN sudo apt update
+RUN sudo -E apt-get -y install python3-pip python3-venv
```
stephenliang, line 109: For the apt pieces, can we inline these in one of the first `apt-get` steps?

docker/developers/docker-compose.commands.yml
```diff
@@ -0,0 +1,211 @@
...
+  build-js:
+    <<: *command-isolated-base
+    volumes:
+      - nvm:/home/cdodev/.nvm
+      - ../..:/app/src
+    working_dir: /app/src/apps
+    command: /bin/bash -l -c '../docker/developers/run.sh yarn build'
+
+  # Runs a server that watches for changes made to JS code and rebuilds on update
+  build-js-server:
```
stephenliang, line 101: Should this be `run-js-server`?

docker/developers/install-local-nvm.sh
```diff
@@ -0,0 +1,39 @@
...
+if [ ! -e ${HOME}/.nvm/package.json ]; then
+  cp -r /opt/base-nvm/* ${HOME}/.nvm/.
+fi
+
+# Initialize NVM
+. ${HOME}/.nvm/nvm.sh
+
+# Determine the desired node version
+NODE_VERSION=`jq -r .engines.node < apps/package.json | sed "s/[\^]//g"`
+echo " - Installing node v${NODE_VERSION}"
```
stephenliang, line 25: The semver in `package.json` is for the highest minor of 18, instead of reading the version from `package.json` we should probably look at [.nvmrc](https://github.com/code-dot-org/code-dot-org/blob/staging/.nvmrc). If we change directories to `apps`, we can run `nvm install`, or we can read `.nvmrc` to get the version. I also think we can unpin the version in `.nvmrc` but that is a separate issue.

docker/developers/install-local-nvm.sh
```diff
@@ -0,0 +1,39 @@
...
+# Install the correct version of NPM
+NPM_VERSION=`jq -r .engines.npm < apps/package.json | sed "s/[\^]//g"`
+echo " - Installing npm v${NPM_VERSION}"
+npm install -g npm@${NPM_VERSION}
+
+# Install Yarn
+YARN_NPM_VERSION=1.22.19
+YARN_VERSION=`jq -r .packageManager < apps/package.json | sed "s/^yarn@//" | sed "s/[\^]//g"`
+echo " - Installing yarn v${YARN_NPM_VERSION} (bootstrapped from npm)"
+npm i -g yarn@${YARN_NPM_VERSION}
```
stephenliang, line 39: Similar to the Dockerfile comment, the yarn should probably be installed via `corepack` not using `npm install`

docker/developers/install-local-nvm.sh
```diff
@@ -0,0 +1,39 @@
...
+
+# Determine the desired node version
+NODE_VERSION=`jq -r .engines.node < apps/package.json | sed "s/[\^]//g"`
+echo " - Installing node v${NODE_VERSION}"
+nvm install ${NODE_VERSION}
+nvm alias default ${NODE_VERSION}
+nvm use default
+
+# Install the correct version of NPM
+NPM_VERSION=`jq -r .engines.npm < apps/package.json | sed "s/[\^]//g"`
```
stephenliang, line 31: I believe we can preserve the semver syntax, otherwise this would pin us to an older version when we want the latest minor on the same major.

```suggestion
NPM_VERSION=`jq -r .engines.npm < apps/package.json`
```

docker/developers/Dockerfile
wilkie: `SETUP.md` specifies a specific version, so the Dockerfile does, too.

docker/developers/Dockerfile
wilkie: I would rather not.

docker/developers/docker-compose.commands.yml
wilkie: But it builds? It is the `build-js` step but as a sustained server. It's a little confusing to have it be `run`, no?

docker/developers/Dockerfile
wilkie: If you do that, adding a new dependency would mean redoing that entire layer and any layer below it. Do you really care about that in the case of a fresh build where you would do it anyway? Or is there an argument to placing the install dependencies in places around their context and commenting them within this Dockerfile.

I think I would rather them be scattered.

Eventually, the "base" image of the Dockerfile will be uploaded and not built by each contributor. The only step a dev contributor might want to make would be the very last step of rewriting the container user UID to match their system user outside.

docker/developers/install-local-nvm.sh
wilkie: I want every developer to have the same version of npm! Ideally, the same version as production. How to make this happen?

docker/developers/install-local-nvm.sh
wilkie: Why?

wilkie: I'd have to install corepack, then, as well, just to install yarn. This ecosystem is just madness, honestly.

DOCKER.md
wilkie: So that the docker container user has the same id as the host user, so when it creates files the host user owns them. you do this on the build for the image so that you do not need root to run the image.

PR #60308, opened 2024-08-09
Add new users to welcome series list

bethanyaconnor:
Instead of sending a welcome email when teachers sign up for our site, we're going to instead add them a "contact list", which will be configured to send the user a series of welcome emails.

I left the previous methods around as we'll need them for other scenarios.

See the [product spec](https://docs.google.com/document/d/11ZhI9YvbJwMICH4St9TJSHoIY-BLluJxt30YJsw0hHk/edit?usp=gmail_thread#heading=h.qi8xxyq72bct) for more details!

### DIFF
--- a/dashboard/config/locales/en.yml
+++ b/dashboard/config/locales/en.yml
@@ -972,7 +972,7 @@ en:
     privacy_notice: "Privacy Notice"
     accept_label: 'I agree to the <a href="%{tos_url}" target="_blank">Terms of Service</a> and <a href="%{privacy_url}" target="_blank">Privacy Policy</a>.'
     accept_label_markdown: I agree to the [Terms of Service](%{tos_url}) and [Privacy Policy](%{privacy_url}).
-    accept_tos: By signing up for Code.org, you agree to our [Terms of Service](%{tos_url}) and [Privacy Policy](%{privacy_url}), and acknowledge that you will receive a series of 6 onboarding emails over 2 weeks.
+    accept_tos: By signing up for Code.org, you agree to our [Terms of Service](%{tos_url}) and [Privacy Policy](%{privacy_url}), and acknowledge that you will receive a series of onboarding emails over the next few weeks.
     accept_new_tos: By signing up for Code.org, you agree to our [Terms of Service](%{tos_url}) and [Privacy Policy](%{privacy_url}).
     accept: "Accept"
     print: "Print"
--- a/dashboard/config/locales/en.yml
+++ b/dashboard/config/locales/en.yml
@@ -972,7 +972,8 @@ en:
     privacy_notice: "Privacy Notice"
     accept_label: 'I agree to the <a href="%{tos_url}" target="_blank">Terms of Service</a> and <a href="%{privacy_url}" target="_blank">Privacy Policy</a>.'
     accept_label_markdown: I agree to the [Terms of Service](%{tos_url}) and [Privacy Policy](%{privacy_url}).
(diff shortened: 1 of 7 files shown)


dashboard/config/locales/en.yml
```diff
@@ -972,7 +972,8 @@ en:
     privacy_notice: "Privacy Notice"
     accept_label: 'I agree to the <a href="%{tos_url}" target="_blank">Terms of Service</a> and <a href="%{privacy_url}" target="_blank">Privacy Policy</a>.'
     accept_label_markdown: I agree to the [Terms of Service](%{tos_url}) and [Privacy Policy](%{privacy_url}).
-    accept_tos: By signing up for Code.org, you agree to our [Terms of Service](%{tos_url}) and [Privacy Policy](%{privacy_url}).
+    accept_tos: By signing up for Code.org, you agree to our [Terms of Service](%{tos_url}) and [Privacy Policy](%{privacy_url}), and acknowledge that you will receive a series of 6 onboarding emails over 2 weeks.
```
bethanyaconnor: @dmcavoy should we make this statement a bit vaguer? Something like "you will receive a series of onboarding emails over the next few weeks"?

dmcavoy: Yes lets make it more vague for now. Thank you!

dashboard/config/locales/en.yml
bethanyaconnor, line 1: We're changing the existing string so that we don't lose translations. The new string will appear on the dialog that shows for users who have not accepted our terms of service, which is much lower frequency.

hannahbergam: And that use of the updated accept_tos string will be in a future PR?

lib/cdo/mailjet.rb
```diff
@@ -65,7 +62,7 @@ def self.find_or_create_contact(email, name)
     return contact if contact&.id.present?
 
     Mailjet::Contact.create(
-      is_excluded_from_campaigns: true,
+      is_excluded_from_campaigns: "false",
```
hannahbergam: I'm curious about why true wasn't a string but false is a string

lib/cdo/mailjet.rb
```diff
@@ -108,6 +105,23 @@ def self.delete_contact(email)
     end
   end
 
+  def self.add_to_contact_list(contact, list_id)
+    return unless enabled?
+    return if contact.nil?
+
+    Mailjet.configure do |config|
```
hannahbergam: Could this somehow be abstracted? Or will we sometimes be using different keys/secrets/versions?

hannahbergam (commented):
I left a couple questions:)

dashboard/config/locales/en.yml
bethanyaconnor: It's already being used in the sign up flow

lib/cdo/mailjet.rb
bethanyaconnor: 1. It's different between contacts and emails 🫠 But I think it could likely be abstracted somewhat, let me play with it

dashboard/config/locales/en.yml
hannahbergam: Ahhh okok thanks!

lib/cdo/mailjet.rb
hannahbergam: omg the API version 3 vs 3.1 🤦🏻‍♀️

hannahbergam (approved):
Yay! Thanks for all the spelunking!

PR #60310, opened 2024-08-09
Remove engineering careers page

bethanyaconnor:
We've moved this page to [github](https://github.com/code-dot-org).  I don't see this page linked anywhere and we've already asked TJ to point to our new github version, so I'm going full deprecation! 🔥 

How I confirmed deletions:
- I grepped for `engineering_` and confirmed that the only instances of that string in the pegasus directory were in translation files, so all of those strings can be deleted
- I grepped for `careers/engineering` and found nothing, so all the images can be deleted
- I searched for any other uses of the view or style sheet and found none, so I deleted those


TurnerRiley (approved):
LGTM!

Erin007:
Acquisitions is deleting things left and right and I love it! Thanks @bethanyaconnor 

yay

PR #60315, opened 2024-08-09
Use migrated Blockly levels for i18n test

mikeharv:
The i18n UI test feature uses a couple of Artist levels to check translations of toolbox categories and function names. Some test steps need to be updated in order to continue working with Google Blockly. 

## Previous approach with Maze
When updating similar UI tests affecting Maze, we were able to just add some explicit block ids to the levels and update the steps. The tests would then run with `blocklyVersion=google` ahead of the Maze migration. We then removed the URL param after Maze was stable on Google Blockly (without a DCDO flag).

This approach doesn't quite work for these i18n tests. The tests use URL which append a locale like `s/{lesson and level}/lang/es-MX`. This redirects to `s/{lesson and level}?lang=es-MX`. If we add the Blockly param, it gets lost during the redirect.

## New approach for i18n tests
To get these tests passing with Google Blockly, we can switch them over to a similar level for a different (migrated) lab. I used an existing Poetry level for toolbox categories. For function names, I found a very similar Bee level from Course 3 and added it to allthethings. (I opted not to clone it to ensure that the existing translations for the level would remain in tact.

||Old level|New level|
|-|-|-|
|Toolbox Categories|image|image|
|Function names|image|image|

I didn't find a great way to keep using the same Artist levels. I considered trying use `s/{lesson and level}?lang=es-MX&blocklyVersion=google`, but unfortunately this first `lang` param doesn't seem to change the language on its own (unlike the `/lang/` route). I also considered first using `/lang/` and then manually reloading the level with the block param, but this felt messy. 

One advantage of the chosen approach is that we won't need to go back and re-edit these tests after the Artist migration is complete. In any case, it shouldn't matter which lab we use since we are testing Blockly more than we are testing Artist. 

### RTL Marker issue
After switching up the levels as described above, I noticed that the function name test was still [failing](https://cucumber-logs.s3.amazonaws.com/circle/13161/Chrome_foundations_i18n_output.html?versionId=gH7kSAkKDvaR3zfubAIh7Rhjd9smmcc9) in Arabic. 
image
Specifically I was seeing:
```
expected: "احصل على ٥"
     got: "احصل على ٥‏"
```

It was tricky to determine the problem, but I eventually determined that the string from the actual block ended with an invisible RTL mark: https://unicode-explorer.com/c/200F

I think that Blockly adds this character to help with cases where a string might contain both LTR and RTL characters (even though our translation here only uses RTL). In any case, it feels safer enough to ignore this character without compromising the purpose of the test. To handle this, I added a new RTL specific test step which checks that one string _includes_ the other, rather than requiring strict string equality. In this case, the string with the RTL mark includes the string without it, which is hopefully strong enough evidence that we managed to translate the name of the function.

## PR Checklist:

- [ ] Code is well-commented


dashboard/config/levels/custom/maze/2-3 Bee Functions 2.level
```diff
@@ -1,5 +1,6 @@
 <Karel>
   <config><![CDATA[{
+  "published": true,
```
mikeharv, line 3: I used the levelbuilder interface to edit the levels, but I manually reverted `created_at` on line 5 so that we don't lose that meta-data.

dashboard/test/ui/features/foundations/i18n.feature
```diff
@@ -121,27 +117,25 @@ Scenario: Minecraft:Agent tutorial in Arabic (RTL)
   Then element ".csf-top-instructions p" has "ar-SA" text from key "data.short_instructions.MC_HOC_2017_01_RETRY"
 
 Scenario: Translated function names in Arabic
-  Given I am on "http://studio.code.org/s/allthethings/lessons/3/levels/3/lang/ar-SA"
+  Given I am on "http://studio.code.org/s/allthethings/lessons/4/levels/6/lang/ar-SA"
   And I wait for the lab page to fully load
   # Toolbox call block is translated
-  Then element "[block-id=6] .blocklyText" has "ar-SA" text from key "data.function_definitions.2-3 Artist Functions 4.draw a square.name"
+  Then element "[data-id='toolboxCallBlock'] .blocklyText" has "ar-SA" RTL text from key "data.function_definitions.2-3 Bee Functions 2.get 5.name"
```
mikeharv, line 123: This is the first of 3 instances of a new "RTL text" step. This works essentially the same way, but ignores an invisible RLM character that can be found on the block text with Google Blockly.

dashboard/config/levels/custom/maze/2-3 Bee Functions 2.level
```diff
@@ -1,5 +1,6 @@
 <Karel>
```
molly-moen, line 1: Is this level used in a real script? I'm wondering if it would be safer to clone it for the purposes of unit testing (and use the clone in `allthethings`), so any changes made to the level don't cause unexpected test failures.

molly-moen (commented):
One question on the new allthethings level, otherwise looks good!

dashboard/config/levels/custom/maze/2-3 Bee Functions 2.level
mikeharv: The level comes from Courses 1-4, which are not updated regularly. There are also other levels in allthethings that also show up in these courses. I considered cloning the level too, I believe that the cloned level wouldn't automatically have the same translations available. If the cloned level shows English, it defeats the point of the test.

molly-moen: Makes sense!

molly-moen (approved):

PR #60320, opened 2024-08-09
Add ability to stop an in-progress python program

molly-moen:
This PR adds the ability to stop an in progress Python program in Python Lab. As part of this I also did some improvements to how we load Python Lab, to make the overall load experience smoother for users. Now while a program is running, the run and test buttons are replaced with a "stop" button.

Pyodide has a way to [interrupt execution](https://pyodide.org/en/stable/usage/keyboard-interrupts.html) of an in-progress program. However, this method relies of using `SharedArrayBuffer`, which required a [secure context](https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts). I tried to get this set up, but it caused other things (such as font awesome icons) to break.

Instead of using this method, I just terminate the current worker and restart. The downside of this is we have to re-initialize pyodide, which takes between 5-30 seconds depending on the computer. Previously, much of this load time was noticeable to users on first run, as it takes running code for the initialization to complete (this is a black box for me, it's just what I've noticed with pyodide). 

However, after #60270 I realized we could warm up our environment, because we now run setup code before running user code. As part of this work, I moved running that setup code from when the user clicks the run button to on initialization, and then at the end of each run to set up for the next run. This means for most users their first run will be as fast as their subsequent runs, which happen instantly.

I also added some signaling that the setup was happening as a part of this work, especially since on stop it would happen again and slow down their runtime. When we are loading pyodide now the test and run buttons are disabled, and we put a message in the console explaining that the environment is being set up.

## Screen Recordings
These were all filmed on a speedy macbook on localhost, which is faster than prod.
### Old load/run experience
https://github.com/user-attachments/assets/cba03cd3-c575-4fe0-a298-10fb3e6d733d

### Normal Execution
Note that run/test go away while the program is running and are replaced with a stop button

https://github.com/user-attachments/assets/2fef38bc-7a8e-43ca-8e0b-947a1b255b47

### Stopping a program
https://github.com/user-attachments/assets/a9525730-e1de-4b4a-ab04-0295d57f92ae

### New startup behavior
https://github.com/user-attachments/assets/169a4651-e9ed-448d-b18e-b6b2d2ceba90

## Links

- jira tickets: [CT-612](https://codedotorg.atlassian.net/browse/CT-612), [CT-613](https://codedotorg.atlassian.net/browse/CT-613)

## Testing story
Tested locally. 

## Follow-up work
We discussed on [slack](https://codedotorg.slack.com/archives/C06JELCAPT9/p1723230006341119) whether we might want to replace the console text with some sort of spinner/progress bar. This is tracked for post pilot in [CT-729](https://codedotorg.atlassian.net/browse/CT-729)

## PR Checklist:

- [ ] Code is well-commented


snickell:
From [MDN on secure context](https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts):
> Locally-delivered resources such as those with http://127.0.0.1 URLs, http://localhost and http://*.localhost URLs (e.g. http://dev.whatever.localhost/), and file:// URLs are also considered to have been delivered securely.

I've been confused because I could have sworn that I used all these "not allowed" features in dev servers before without localhost. We're having these unusual problems because of our `localhost-studio` approach! Its a good reminder how KISS tends to be the winning strategy over long time periods.

I think we're implicitly avoiding using lots of great modern web features due to the challenges of running HTTPS in dev.

Maybe this could all be fixed by switching from localhost-studio to `studio.localhost` ?

I've opened an issue for me to investigate: https://github.com/code-dot-org/code-dot-org/issues/60328

apps/src/codebridge/ControlButtons/control-buttons.module.scss
```diff
@@ -3,16 +3,19 @@
...
 }
 
-.firstControlButton {
+.centerButton {
   grid-column-start: 2;
 }
 
+.runButton {
+  margin-right: 8px;
+}
```
snickell, line 17: zero important just curious: what trouble was `column-gap` causing?

apps/src/codebridge/ControlButtons/index.tsx
```diff
@@ -87,13 +91,24 @@ const ControlButtons: React.FunctionComponent = () => {
 
   const handleRun = (runTests: boolean) => {
     if (onRun) {
-      onRun(runTests, dispatch, source);
+      setIsRunning(true);
```
snickell, line 94: I was curious why the "Run" button feels a very tiny bit laggy (it doesn't disappear quickly after click)? I don't think you should change this for this PR fwiw. I assumed that run was being gated on a slow operation that had to start, but this looks like a pretty instant state change. 🤷 Is react rendering taking that long to kick in? I wouldn't have even noticed the lag except the UX is so slick in general, the fading buttons etc are sweet.

apps/src/pythonlab/pyodideWebWorker.js
```diff
@@ -35,15 +35,23 @@ async function loadPyodideAndPackages() {
...
 }
 
 let pyodideReadyPromise = null;
 let pyodideGlobals = null;
 async function initializePyodide() {
+  let promiseWasNull = false;
   if (pyodideReadyPromise === null) {
     pyodideReadyPromise = loadPyodideAndPackages();
+    self.postMessage({type: 'loading_pyodide'});
+    promiseWasNull = true;
```
snickell: unimportant nit, 100% could wontfix, what about:
```
const promiseWasNull = pyodideReadyPromise === null
if (promiseWasNull) {
  pyodideReadyPromise = loadPyodideAndPackages();
  self.postMessage({type: 'loading_pyodide'});
}
```

One line of code shorter, and uses one less mutable variable.

apps/src/pythonlab/pyodideWorkerManager.js
```diff
@@ -10,59 +10,77 @@ import {setAndSaveProjectSource} from '@cdo/apps/lab2/redux/lab2ProjectRedux';
...
+      getStore().dispatch(
+        appendSystemMessage('Loading your python environment...')
+      );
+      getStore().dispatch(setLoadingCodeEnvironment(true));
+      return;
+    } else if (type === 'loaded_pyodide') {
+      getStore().dispatch(appendSystemMessage('Your environment is ready!'));
+      getStore().dispatch(setLoadingCodeEnvironment(false));
+      return;
+    } else {
```
snickell, line 68: I only suggest this when an if/then chain is getting really long, but at a certain point I think its a candidate for a `switch` statement. This clarifies at first glance that only the `type` will be used for varying conditions, and thus clarifies the relative simplicity of the long stanza.

Also, switch statements have a way of avoiding future maintainers years down the road from ADDING complexity by if/then-ing on complex conditions and stretching the simplicity of the message forking code.

The only except I see here is the merging of sysout and syserr, I would argue these would be well served by extracting the contents of the block and having two separate one-liner sysout and syserr case statements.

snickell: Not necessarily a candidate idea for this PR or particularly important, more a general thought.

snickell (approved):
An elegant+clever technical solution to a UX problem, with high code quality (as always).

My only high level thought is UX-related, and probably a consideration for a future PR. Should we perhaps keep the console empty by default, and use a different mechanism for signaling "setting up" and "now set up" status?

Even with experience, increased "console noise" makes it harder to visually see problems.... now imagine that each console line takes deciphering because you're just getting started. Don't you love JS apps (none found here lol) where you open the JS console, and its dead empty unless there's a problem?

I would advocate shifting this status signaling to the run button itself, which also attaches the status visually next to the thing affected, making the linkage clear.

For example, have a "disabled" state where Run is visible but its highly faded out and disabled. If you hover the button in the faded state, it tells you that its setting up the python environment, possibly with a small animated "progress spinner" next to the button.

It'd help to have the Run button's "active" state more visible active by 25%, which I think would be a good improvement independently, its a little subtle atm, esp stuck at the bottom of the screen its a little visually hard to find.

apps/src/codebridge/ControlButtons/control-buttons.module.scss
molly-moen: I changed this because I switched from having 1 column per button to having one column for run/test and one for continue. The run/test column is replaced by the stop button when we switch states. After moving to one column, I needed a margin on the run button so. the buttons weren't squished together.

apps/src/pythonlab/pyodideWorkerManager.js
molly-moen: I'll update this as a follow up!

molly-moen: https://codedotorg.atlassian.net/browse/CT-730

molly-moen:
@snickell @breville I filed a [follow up ticket](https://codedotorg.atlassian.net/browse/CT-732) to cover improving the load state, because this PR is already pretty big.

apps/src/codebridge/ControlButtons/index.tsx
molly-moen: It doesn't feel laggy to me locally, did you feel like it was laggy in the video or did you pull and try it out?

apps/src/pythonlab/pyodideWebWorker.js
molly-moen: updated!

